PR #2366, opened 2016-02-17
AO3-4496 Cache chapter counts

james-beal:
Cache the counts of the number of chapters a work has.

https://otwarchive.atlassian.net/browse/AO3-4496
 .

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -6,6 +6,7 @@ class Work < ActiveRecord::Base
   include Pseudable
   include Searchable
   include WorkStats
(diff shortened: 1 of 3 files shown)


app/models/work.rb
```diff
@@ -684,6 +691,7 @@ def multipart?
 
   after_save :update_complete_status
   def update_complete_status
+    # self.chapters.posted.count ( not self.number_of_posted_chapter , here be dragons )
```
james-beal: I realise now that this looks like a function call not an explaination.

PR #2367, opened 2016-02-17
AO3-4495 Cache the bookmark counts

zz9pzza:
https://otwarchive.atlassian.net/browse/AO3-4495


app/models/bookmark.rb
```diff
@@ -86,6 +86,15 @@ class Bookmark < ActiveRecord::Base
   # Use the current user to determine what works are visible
   scope :visible, visible_to_user(User.current_user)
 
+  before_destroy :invalidate_bookmark_count
+  after_save :invalidate_bookmark_count
+
+  def invalidate_bookmark_count
+    if self.bookmarkable_type == 'Work'
+      Work.find(self.bookmarkable_id).invalidate_work_bookmark_count
```
cesy: So we're caching bookmark counts on works but not series?

cesy:
Code looks sensible but could do with review from someone who's more experienced with caching than me.

sarken:
I think this looks good, but I share Cesy's question about why we're just caching the bookmark count for  works instead of works, series, and external works.

zz9pzza:
I am only caching works as that is the query that is impacting the database.

lib/work_bookmark_count_caching.rb
```diff
@@ -0,0 +1,15 @@
+module WorkBookmarkCountCaching
+  def key_for_bookmark_counting_public
```
shalott: I hate to be picky but I find the naming here confusing. Can I suggest for clarity and consistency: 

BookmarkCountCaching/bookmark_count_caching.rb
public_bookmarks_count
key_for_public_bookmarks_count
invalidate_public_bookmarks_count
/v1/public_bookmarks_count/#{self.id}

shalott:
Other than the naming I have no issue with only caching the query that's actually causing slowdowns, makes sense since work bookmarks are the most heavily used.

shalott:
Looks good to me!

PR #2368, opened 2016-02-18
AO3-4496 Cache chapter counts

zz9pzza:
Cache the counts of the number of chapters a work has.

https://otwarchive.atlassian.net/browse/AO3-4496


scottsds:
This looks good to me. I pulled it down and tested locally. The chapter_edit.feature tests that this bit of code is working correctly.

PR #2369, opened 2016-02-19
Ao3 4497 Cache the list of tags associated with the work

zz9pzza:
https://otwarchive.atlassian.net/browse/AO3-4497

### DIFF
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -160,8 +160,9 @@ class Tag < ActiveRecord::Base
   end
 
   def flush_work_cache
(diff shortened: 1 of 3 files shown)


sarken:
Silly question! Why do we need to cache the tags in the blurb when they're already in the part of the blurb that's cached? Is that caching not working for some reason?

zz9pzza:
I am not 100% certain that it is working, I can say that those queries are very popular and the caching here is actually done in the model so any other calls to that will see a speed up too.

This call is the 8% of the mysql queries and the query taking most time right now.

PR #2376, opened 2016-02-27
AO3-4500 Sometimes ec.external_author_name can be nil

zz9pzza:
https://otwarchive.atlassian.net/projects/AO3/issues/AO3-4500

https://otw.16bugs.com/projects/4911/bugs/359269

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -155,7 +155,7 @@ module ApplicationHelper
         external_creatorships = creation.external_creatorships.select {|ec| !ec.claimed?}
         external_creatorships.each do |ec|
           archivist_pseud = pseuds.select {|p| ec.archivist.pseuds.include?(p)}.first
-          archivists[archivist_pseud] = ec.external_author_name.name
+          archivists[archivist_pseud] = ec.external_author_name.nil? ? nil : ec.external_author_name.name
         end
       end
 


sarken:
I'd really like to see an automated test for this.

zz9pzza:
From a programming point of view the code is just more defensive. It seems "unfair" not to allow the fix in because the original code didn't have a test.

https://github.com/zz9pzza/otwarchive/blob/AO3-4500/app/helpers/application_helper.rb#L163

Has a test for the archivists[pseud].nil? beiing null

( I will say that I have no way to test this apart from hot fixing it on live )

sarken:
That test isn't for quite the same thing. For works imported by an archivist, a proper external creator byline would look like "CREATOR [archived by ARCHIVIST]." The existing test handles the case where ARCHIVIST is nil -- it makes sure the byline just says "CREATOR" instead of the rather odd "CREATOR [archived by]." Here we're dealing with CREATOR being nil and leaving us with a byline that says only "[archived by ARCHIVIST]."

However! Since we talked about this in Slack and clarified that we don't know what causes `ec.external_author_name` to be nil, we can let the test go for now. We should note on the issue that this is fixing a symptom (the feed giving an error) rather than the root cause (`ec.external_author_name` being nil, which appears to manifest as a byline that says only "[archived by ARCHIVIST]"). That way, we'll know we need to keep digging until we figure out how to reproduce (and prevent) the nil `ec.external_author_name`... and _that_ fix will be the one that needs a test. :P

PR #2377, opened 2016-02-27
AO3-4501 Import notes

ariana-paris:
https://otwarchive.atlassian.net/browse/AO3-4501

- Adds "notes" to all the places we collect importing data
- Tweaks "Tags" section of the import form so the override checkbox is in the form with the tags it overrides
- Also removed unused tags from import tests and expands the descriptions of individual field tests


zz9pzza:
This looks reasonable to me. I am willing to put a ready to merge on if no one else is free to review.

app/views/works/_work_form_tags.html.erb
```diff
@@ -1,9 +1,18 @@
 <fieldset class="work meta">
   <legend><%= ts('Tags') %></legend>
   <h3 class="landmark heading"><%= ts('Tags') %></h3>
+
   <p class="note"><%= ts('Tags are comma separated, %{max} characters per tag.', :max => ArchiveConfig.TAG_MAX) %></p>
 
   <dl>
+    <% import ||= false
```
sarken: So the good news is, this is what I was going to suggest when I saw the earliest version of moving the "override tags" option! (When I saw it, it wasn't part of the dl, but above the note.) The bad news is now that it covers the notes as well as the tags, I think it's better to keep it where it was. /o\ If you're adding notes but not tags, there's a very good chance you're just going to ignore the Tags section of the form completely and thus miss that it also contains an option that affects notes. At least the top section of the form is kind of a general preferences area with things you have to fill in no matter what, so there's a _slightly_ better chance you might see it (especially if you've used the form before).

ariana-paris: I've had another go at formatting the form by putting the notes into the main form which might improve the logical flow.

sarken: I think the notes were much better off in their own fieldset -- they don't make much sense in a fieldset labeled Tags that has tag-specific instructions.

sarken:
Looks good!

CristinaRO:
Already grey, please update whenever convenient.

sarken:
Marking Needs Coder Action just so we know this is waiting on conflict resolution.

ariana-paris:
Conflicts resolved!

app/views/works/new_import.html.erb
```diff
@@ -91,16 +98,28 @@
...
+        <%= check_box_tag "override_tags" %>
+        <%= label_tag "override_tags", ts("Set the following tags and/or notes on all works, overriding whatever the importer finds in the content.") %>
+      </dd>
     </dl>
   </fieldset>
 
-  <%= render :partial => 'work_form_tags', :locals => {:include_blank => true} %>
+  <%= render :partial => 'work_form_tags', locals: { include_blank: true, import: true } %>
+
+  <fieldset>
```
sarken: If you have a second, could you fix the indenting here? `<legend>` has extra indenting and lines 118 and 119 should both be indented two more spaces. (Line 67 above could probably use a `ts()` as well, but I'm not particularly concerned given the future of `ts()`.)

Going to mark ready to merge, though, since it's probably not worth holding it up for either of those things.

ariana-paris:
Update indenting and `ts()` around a label (some day we'll have to hunt them all down but not today)

PR #2379, opened 2016-02-28
AO3-4503 Remove tag feeds from the NAS

zz9pzza:
https://otwarchive.atlassian.net/browse/AO3-4503


zz9pzza:
This needs coordination with a bit of systems work. 

The old tag directory needs to be put to one side.

I would also ensure that the squid cache cached the feeds.

app/models/work.rb
```diff
@@ -186,8 +186,19 @@ def check_for_invalid_chapters
   after_update :adjust_series_restriction
   
   after_save :expire_caches
-  
+
+  def self.rss_work_summary_key(id)
+    "/v1/rss_data_tags/#{id}/summary"
+  end
```
shalott: Are these deliberately in the same space as the tag keys despite using a different id field? What I mean is, you can theoretically have a work and tag with the same id, which would mean you had rss_data_tags/[work id]/whatever and rss_data_tags/[tag id] and if you wanted to wipe out everything under rss_data_tags/[work id] you would as a result take out everything for the tag that accidentally shares the id number. This seems clearly wrong so I'm wondering what I'm missing?

app/models/tag.rb
```diff
@@ -354,6 +354,9 @@ def async(method, *args)
 
   # Class methods
 
+  def self.tag_key_for_feeds(id)
```
shalott: Also, naming issue here: for consistency would suggest self.tag_feeds_key(id)

app/controllers/tags_controller.rb
```diff
@@ -90,23 +88,34 @@ def show
...
-    else
-      redirect_to tag_works_path(:tag_id => @tag.to_param) and return
+      { path: path, works: works, tag: tag }
+    end
+    redirect = hash[:path]
+    @works = hash[:works]
+    @tag = hash[:tag]
+    unless redirect.nil? 
+      redirect_to path and return
     end
```
shalott: Doesn't this need to be redirect_to hash[:path]? I'm not sure how this is working: is there a test that is actually exercising this code? If not I suggest writing the test first, because I am pretty sure it should fail, and then fixing the code after. 

Also, why are you using this separate variable redirect? Why not just unless hash[:path].nil? And setting the @works and @tag is pointless if there is a redirect, isn't it? I would suggest moving that to after the unless check.

zz9pzza: The tests are in other_b/subscriptions_fandoms.feature and I remember them failing while I was developing

shalott: You would need to test for this specifically with a feed that causes a redirect and it has to be actually cached so it has to be the second time the feed is fetched; are you doing so anywhere? Because if you just get the cached value of hash, the code inside should NOT be re-executed in which case the "path" variable never gets set and redirect_to path should fail. (Actually I would guess the redirect should fail anyway because path is a local variable inside the hash block?)

zz9pzza: What do you mean by "You would need to test for this specifically with a feed that causes a redirect" ?

shalott: What I mean is, the line "redirect_to path and return" needs to be exercised by a test and is not, because I am pretty sure it SHOULD be failing given the way you have it:

```
hash = Rails.cache.fetch(Tag.tag_feeds_key(params[:id])) do
  ....
  { path: path, works: works, tag: tag }
end
```

Here you are setting a variable "hash". But the variable "path" itself is ONLY set inside the cached block--which will not run if the value is retrieved from the cache. You do set `redirect = hash[:path]` but then you don't actually USE the variable "redirect", you use the variable "path" which has not been set. 

So when you do `redirect_to path and return` where the value is coming out of the hash, you should be getting a variable not set failure. Since you are NOT getting that failure out of the tests, that tells me the tests aren't exercising this line, and you want a test that does exercise that line. That line gets invoked when redirect (which is equal to hash[:path]) is not nil AND when the value comes out of the cache, so you need to create a test case where you are redirecting via a cached value.

zz9pzza: Thanks, I understand now, I will try and write a test that goes though that path.

app/controllers/tags_controller.rb
```diff
@@ -105,21 +103,32 @@ def show
...
+    hash = Rails.cache.fetch(Tag.tag_feeds_key(params[:id])) do
+      path = nil
+      works = nil 
+      begin
+        tag = Tag.find(params[:id])
+      rescue ActiveRecord::RecordNotFound
+        raise ActiveRecord::RecordNotFound, "Couldn't find tag with id '#{params[:id]}'"
+      end
+      tag = tag.merger if !tag.canonical? && tag.merger
+      # Temp for testing
```
shalott: Noting this is still here even though it's supposed to only be temporary

zz9pzza: I can take the bit out that says it's only temporary.

app/controllers/tags_controller.rb
```diff
@@ -105,21 +103,32 @@ def show
...
-      @tag = Tag.find(params[:id])
-    rescue ActiveRecord::RecordNotFound
-      raise ActiveRecord::RecordNotFound, "Couldn't find tag with id '#{params[:id]}'"
-    end
-    @tag = @tag.merger if !@tag.canonical? && @tag.merger
-    # Temp for testing
-    if %w(Fandom Character Relationship).include?(@tag.type.to_s) || @tag.name == 'F/F'
-      if @tag.canonical?
-        @works = @tag.filtered_works.visible_to_all.order('created_at DESC').limit(25)
+    hash = Rails.cache.fetch(Tag.tag_feeds_key(params[:id])) do
```
shalott: Suggest a comment above this saying that the hash you are building and/or getting from the cache is tag, path, and works, and what each element is used for

app/models/tag.rb
```diff
@@ -368,6 +368,9 @@ def async(method, *args)
 
   # Class methods
 
+  def self.tag_feeds_key(id)
+    "/v1/rss_data_tags/#{id}"
```
shalott: are these paths going to be user-visible or are they intended to only ever be seen by readers? why rss_data_tags -- any reason not to just have rss_tags or tags_rss? Why is the v1 in the url, would we ever have multiple versions active? (Not that we shouldn't, just want to know the reasoning.)

zz9pzza: So the paths will not be seen by users.

We have v1 so if we need to change this code we don't need to worry about clearing the cache. I prefer having a format for cache keys.

app/views/tags/feed.atom.builder
```diff
@@ -6,10 +6,10 @@ atom_feed do |feed|
     unless work.unrevealed?
       feed.entry work do |entry|
         entry.title work.title
```
shalott: Why are you caching summary but not title? would it maybe make more sense to cache/expire the entire entry together? When does expire_caches get called -- will it definitely change whenever the authors are changed? Would like to see a test that checks the feed before/after an author change, title change, summary change.

zz9pzza: Because at that point you have done the title in @works already which is cached so you don't gain anything and increase the complexity.

app/controllers/tags_controller.rb
```diff
@@ -112,7 +116,7 @@ def feed
         raise ActiveRecord::RecordNotFound, "Couldn't find tag with id '#{params[:id]}'"
       end
       tag = tag.merger if !tag.canonical? && tag.merger
-      # Temp for testing
+      # F/F is currently a special case...
```
shalott: Ok but why? If it's just because you need it to be canonical in testing then that is not a good solution. Why aren't you just making it canonical in the test?

app/models/user.rb
```diff
@@ -183,7 +183,10 @@ def direct_filters
 
   def expire_caches
     if login_changed?
-      self.works.each{ |work| work.touch }
+      self.works.each do |work| 
+        work.touch
+        work.expire_caches
```
shalott: It seems to me that work.touch should expire its caches--I'd expect it to. Can you make that happen rather than having this manually call the method as well as touching the work?

app/controllers/tags_controller.rb
zz9pzza: My understanding is that it isn't a change for test but adding the ability for F/F to have a feed when it wouldn't normally ?

zz9pzza:
@shalott I think this has the change you wanted ?

app/controllers/tags_controller.rb
```diff
@@ -105,21 +103,36 @@ def show
...
-      @tag = Tag.find(params[:id])
-    rescue ActiveRecord::RecordNotFound
-      raise ActiveRecord::RecordNotFound, "Couldn't find tag with id '#{params[:id]}'"
-    end
-    @tag = @tag.merger if !@tag.canonical? && @tag.merger
-    # Temp for testing
-    if %w(Fandom Character Relationship).include?(@tag.type.to_s) || @tag.name == 'F/F'
-      if @tag.canonical?
-        @works = @tag.filtered_works.visible_to_all.order('created_at DESC').limit(25)
+    # Construct a hash that can be cached. The has contains
```
sarken: Totally minor, but I think you meant "hash contains" and not "has contains"

app/models/tag.rb
zz9pzza: To me clear these are memcache keys.

sarken:
Aaaand it's grey again, but we might want to hold off until @shalott is available to confirm her concerns were addressed.

sarken:
Should this be https://otwarchive.atlassian.net/browse/AO3-4503? It looks like AO3-4502 is an epic intended to cover multiple issues like this, but I am not 100% sure.

zz9pzza:
I think you are right sarken....

tickinginstant:
Why did you decide to switch from caching the rendered XML to caching the objects and re-rendering the XML each time? Is the time to re-render the XML negligible?

Also, I believe this change results in 41 cache reads per page view -- is the performance hit from doing 41 smaller reads vs. 1 larger read acceptable?

zz9pzza:
The main issue is that the way this is done assumes that we have a shared filesystem between the app servers. This is not the case as we avoid single points of failure so the cache will only come in to effect if you happen to hit the same application server. The chances for that are fairly low.

We want to be able to make the number of works returned variable. And we would like to be able to have feeds for all tag types and maybe even searches in the future.

tickinginstant:
I understand that you need to avoid using the filesystem. I'm wondering why you didn't just wrap a `Rails.cache.fetch(key)` block around the `atom_feed` block.

PR #2384, opened 2016-03-01
AO3-4508 validate kudos

zz9pzza:
https://otwarchive.atlassian.net/browse/AO3-4508

### DIFF
--- a/features/comments_and_kudos/kudos.feature
+++ b/features/comments_and_kudos/kudos.feature
@@ -124,3 +124,26 @@ Feature: Leave kudos
       And the email should contain "guest"
       And the email should contain "Awesome Story"
       And the email should contain "Another Awesome Story"
+
+  Scenario: pseud creation and playing and kudos
+
+    Given I am logged in as "myname1"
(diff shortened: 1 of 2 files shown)


features/comments_and_kudos/kudos.feature
```diff
@@ -124,3 +124,26 @@ Feature: Leave kudos
       And the email should contain "guest"
       And the email should contain "Awesome Story"
       And the email should contain "Another Awesome Story"
+
+  Scenario: pseud creation and playing and kudos
```
sarken: Good news! The pseud (username) use case is actually already covered as part of the scenario on line 67, so you can remove this if you like. Otherwise, if we'd like to keep it around to make it clearer this scenario is indeed covered, I'd suggest shortening it a bit:

```
Scenario: Kudos are left in the format pseud (username) when the username and default pseud do not match

    Given I am logged in as "myself" with password "password"
      And "myself" creates the default pseud "Me"
    When I view the work "Awesome Story"
      And I press "Kudos ♥"
    Then I should see "Me (myself) left kudos on this work!"
```

PR #2322, opened 2016-01-23
AO3-4460 Update rest-client gem

zz9pzza:
bundle update --source rest-client

https://otwarchive.atlassian.net/browse/AO3-4460

PR #2319, opened 2016-01-23
AO3-4447 update the resque/redis/resque-scheduler ( take 2 )

zz9pzza:
https://otwarchive.atlassian.net/browse/AO3-4447

PR #2385, opened 2016-03-03
AO3-4509 Add gem security to tests

zz9pzza:
https://otwarchive.atlassian.net/browse/AO3-4509
So we will either need to update a gem or add it to an exceptions list

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -129,6 +129,7 @@ end
 group :test, :development do
   gem 'pry'
   gem 'whiny_validation'
+  gem 'bundler-audit'
 end
 
 group :test, :development, :staging  do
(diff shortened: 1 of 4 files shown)


cesy:
Can we add some commentary about the ones we've chosen to ignore and why, please? Either in the code or in the comments in Jira, just somewhere that people can trace back if necessary. This looks like a good check in principle, though.

zz9pzza:
They are the ones which we fail on right now, so if we didn't ignore them then the tests would fail.

https://github.com/otwcode/otwarchive/pull/2322
https://github.com/otwcode/otwarchive/pull/2319

Are the ones I think

cesy:
Great, so if we merge this first, we can take them out after those two PRs get merged, or if we merge this last, we can take them all out. Neat! Seems good to go, then. It's small enough it's not a significant performance hit on Travis or anything.

zz9pzza:
Yes

shalott:
Looks good to me as well

PR #2388, opened 2016-03-04
AO3-4510 Allow comments to fall out of the cache

zz9pzza:
The cache key on comments is too complex to work out, so just stick a maximum cache time of a week.

https://otwarchive.atlassian.net/browse/AO3-4510

### DIFF
--- a/app/views/comments/_single_comment.html.erb
+++ b/app/views/comments/_single_comment.html.erb
@@ -14,7 +14,7 @@
     <% elsif single_comment.hidden_by_admin? && !logged_in_as_admin? %>
       <p class="message"><%= ts("(This comment is under review by an admin and is currently unavailable.)") %></p>
     <% else %>
-      <% cache("v1"+single_comment, :expires_in => 1.week ) do %>
+      <% cache(single_comment, :expires_in => 1.week ) do %>
         <% # FRONT END, update this if a.user comes in %>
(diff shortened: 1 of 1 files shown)


app/views/comments/_single_comment.html.erb
```diff
@@ -14,7 +14,7 @@
     <% elsif single_comment.hidden_by_admin? && !logged_in_as_admin? %>
       <p class="message"><%= ts("(This comment is under review by an admin and is currently unavailable.)") %></p>
     <% else %>
-      <% cache single_comment do %>
+      <% cache(single_comment, :expires_in => 1.week ) do %>
```
shalott: Er, but doesn't this mean that no edits on the comment will appear until it has expired one week later? That's definitely not going to go over well with users.

zz9pzza: No this is just so changes which currently never show up such as user name changes or icon changes. Show up eventually rather than never. We have 800GB of memcached right now and effectively some things where we have no cache cleating never fall out of the cache this is just putting a maximum time for the comments to live in the cache. Any cache invalidation we do would still work it just means that if the cache invalidation is not complete ( it is not ) then there is a maximum time to live.

shalott: aha got it! OK, looks good!

PR #2389, opened 2016-03-05
AO3-4511 use the new ruby hash syntax

zz9pzza:
http://effectif.com/ruby/update-your-project-for-ruby-19-hash-syntax

find . -name \*.rb -exec perl -p -i -e 's/([^:]):(\w+)\s*=>/\1\2:/g' {} \;

https://otwarchive.atlassian.net/browse/AO3-4511

### DIFF
--- a/app/controllers/admin/admin_invitations_controller.rb
+++ b/app/controllers/admin/admin_invitations_controller.rb
@@ -12,22 +12,22 @@ class Admin::AdminInvitationsController < ApplicationController
   def create
     @invitation = current_admin.invitations.new(params[:invitation])
     if @invitation.invitee_email.blank?
-      flash[:error] = t('no_email', :default => "Please enter an email address.")
-      render :action => 'index'      
+      flash[:error] = t('no_email', default: "Please enter an email address.")
+      render action: 'index'      
     elsif @invitation.save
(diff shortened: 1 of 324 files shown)


zz9pzza:
I would expect to have to throw this branch away and repeat when we actually want to do this ( if we do ) as it will go grey almost instantly.

This is more to see if we want to do it.

shalott:
On balance, I too would like to swap over to this syntax completely, but this will probably cause literally every other open pullreq and every branch that everyone is working on to develop conflicts. I might suggest instead that we ask all coders for lets say the next six months to update all the syntax in every file they have commited in a pullreq (since they are touching those files anyway), and then run the swap on whatever is left afterwards which hopefully by that point will not include files that anyone is working on?

zz9pzza:
I would be happy with that, it will increase the likelihood of merge conflicts during that time so I would be tempted just to bite the bullet if we can get down to say 1 page of pull requests.

What should I do with this pull request, it is really here to start a discussion

shalott:
I would put the code that you used to make the change into the Atlassian issue and then close this pullreq for now, and put the proposed policy change (of updating the hash syntax in all files we touch as coders) into the agenda for meeting.

PR #2390, opened 2016-03-05
AO3-4512 More caching of tags

zz9pzza:
https://otwarchive.atlassian.net/browse/AO3-4512

I would look at the sharing links to make sure all is well.

### DIFF
--- a/app/controllers/chapters_controller.rb
+++ b/app/controllers/chapters_controller.rb
@@ -25,6 +25,7 @@ class ChaptersController < ApplicationController
   # GET /work/:work_id/chapters/:id
   # GET /work/:work_id/chapters/:id.xml
   def show
+    @tag_groups = @work.tag_groups
     if params[:view_adult]
       session[:adult] = true
     elsif @work.adult? && !see_adult?
@@ -55,7 +56,7 @@ class ChaptersController < ApplicationController
       @comments = @chapter.comments.reviewed
 
       @page_title = @work.unrevealed? ? ts("Mystery Work - Chapter %{position}", :position => @chapter.position.to_s) :
-        get_page_title(@work.fandoms.string,
+        get_page_title(@tag_groups["Fandom"][0].name,
           @work.anonymous? ? ts("Anonymous") : @work.pseuds.sort.collect(&:byline).join(', '),
           @work.title + " - Chapter " + @chapter.position.to_s)
 
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -206,8 +206,9 @@ class WorksController < ApplicationController
   # GET /works/1
   # GET /works/1.xml
   def show
+    @tag_groups = @work.tag_groups
     @page_title = @work.unrevealed? ? ts("Mystery Work") :
-      get_page_title(@work.fandoms.size > 3 ? ts("Multifandom") : @work.fandoms.string,
+      get_page_title(@tag_groups["Fandom"].size > 3 ? ts("Multifandom") : @tag_groups["Fandom"][0].name,
         @work.anonymous? ?  ts("Anonymous")  : @work.pseuds.sort.collect(&:byline).join(', '),
         @work.title)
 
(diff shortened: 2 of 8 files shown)


app/controllers/works_controller.rb
```diff
@@ -206,8 +206,9 @@ def drafts
   # GET /works/1
   # GET /works/1.xml
   def show
+    @tag_groups = @work.tag_groups
     @page_title = @work.unrevealed? ? ts("Mystery Work") :
-      get_page_title(@work.fandoms.size > 3 ? ts("Multifandom") : @work.fandoms.string,
+      get_page_title(@tag_groups["Fandom"].size > 3 ? ts("Multifandom") : @tag_groups["Fandom"][0].name,
```
shalott: I'd agree with Hound on changing this one

shalott:
Actually, I would add, this needs automated tests!

zz9pzza:
What additional tests do you think it needs, the current tests found lots of errors so if you can think of more tests then \o/

zz9pzza:
@shalott Is this how you want it to be ?

app/controllers/works_controller.rb
```diff
@@ -206,10 +206,24 @@ def drafts
...
+    @tag_groups = @work.tag_groups
+    if @work.unrevealed?
+      @page_title = ts("Mystery Work")
+    else
+      page_title_inner = ""
+      page_creator = ""
+      if @work.anonymous?
+        page_creator = ts("Anonymous")
+      else
+        page_creator = @work.pseuds.sort.collect(&:byline).join(', ')
```
shalott: Super minor: better to sort after we get the bylines, no?

zz9pzza:
@shalott I think that is what you mean. I will update the other pull based on this branch.

shalott:
Yep, that's right. On the tests, can you point me at the current tests that are covering this code and I'll see if I can think of anything else to add?

zz9pzza:
Well 
works/work_edit_tags.feature  "Scenario: Edit tags on a work"
works/work_edit.feature "Scenario: You can't edit a work unless you're logged in and it's your work"

Sorry its been a whole week since I wrote this but I do remember lots of failures until I got the cache keys to what I think is correct. 

( I also had the approach of computing both the result from the cache and doing it the old way and checking that the two methods gave the same answer ).

If you can think of a test then \o/

app/controllers/chapters_controller.rb
```diff
@@ -55,7 +56,7 @@ def show
       @comments = @chapter.comments.reviewed
 
       @page_title = @work.unrevealed? ? ts("Mystery Work - Chapter %{position}", :position => @chapter.position.to_s) :
-        get_page_title(@work.fandoms.string,
+        get_page_title(@tag_groups["Fandom"][0].name,
           @work.anonymous? ? ts("Anonymous") : @work.pseuds.sort.collect(&:byline).join(', '),
```
shalott: OK, having looked over the tests, the only thing I think needs coverage for this pullreq is the page title. There's page_title.feature but it's not actually checking for anything (including that the user preference or anonymity works at all :P) I suggest:

`Scenario: Page title should respect user preference

    And I go to my preferences page
    And I fill in "Browser page title format" with "FANDOM - AUTHOR - TITLE"
    And I press "Update"
    And I post the work "New Story" with fandom "Stargate"
  When I view the work "New Story"
  Then the page title should include "Stargate - author - New Story"
  
Scenario: Page title should change when tags are edited
    And I post the work "New Story" with fandom "Stargate"
  When I view the work "New Story"
  Then the page title should include "Stargate"
  When I edit the work "New Story"
    And I fill in "Fandoms" with "Harry Potter"
    And I post
  When I view the work "New Story"
  Then the page title should include "Harry Potter"
    And the page title should not include "Stargate"
  
`

zz9pzza:
tests added @shalott \o/

zz9pzza:
Thank you

PR #2394, opened 2016-03-11
AO3-4499 Backdate the work to fix chapter dates on Chapter Index

ariana-paris:
Follow-on to https://otwarchive.atlassian.net/browse/AO3-4499 whereby the change didn't seem to change very much.

### DIFF
--- a/features/importing/work_import.feature
+++ b/features/importing/work_import.feature
@@ -64,6 +64,26 @@ Feature: Import Works
     Then I should see "Preview"
       And I should see "2010-01-11"
 
+  Scenario: Importing a new multichapter work with backdating
+    Given basic tags
+    And the following activated user exists
+      | login          | password    |
+      | cosomeone      | something   |
+    And I am logged in as "cosomeone" with password "something"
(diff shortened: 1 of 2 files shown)


features/importing/work_import.feature
```diff
@@ -64,6 +64,26 @@ Feature: Import Works
...
+    Given basic tags
+    And the following activated user exists
+      | login          | password    |
+      | cosomeone      | something   |
+    And I am logged in as "cosomeone" with password "something"
+    And I set my time zone to "UTC"
+    When I go to the import page
+    And I fill in "urls" with
+         """
+         http://rebecca2525.dreamwidth.org/3506.html
```
zz9pzza: Given that this feels like it is fixing a BOT right now, so not to worry immediately. But shouldn't we use the shiny new adt test account ?

ariana-paris: I agree that the test works should be recreated on the Dreamwidth account, though I'd rather do that separately at this stage.

PR #2395, opened 2016-03-12
AO3-4516 Make remaining absolute urls, relative paths

zz9pzza:
https://otwarchive.atlassian.net/browse/AO3-4516


app/helpers/application_helper.rb
```diff
@@ -126,7 +126,7 @@ def non_anonymous_byline(creation)
   end
 
   def pseud_link(pseud)
```
sarken: This is used in `byline`, which is used in a lot of places, some of which need the full URL -- have you checked RSS feeds, downloads, emails, and share code to make sure that they still have the full URL? I just did a spot check and while the share code looks okay, an HTML download of a work has a broken link for the byline, giving me only `<a href="/users/testy2/pseuds/testy2" rel="author">testy2</a>` when it needs to be the full URL.

app/helpers/tags_helper.rb
```diff
@@ -76,7 +76,7 @@ def link_to_tag_works(tag, options = {})
   end
 
   def link_to_tag_with_text(tag, link_text, options = {})
-    link_to_with_tag_class(@collection ? collection_tag_url(@collection, tag) : tag_url(tag), link_text, options)
+    link_to_with_tag_class(@collection ? collection_tag_path(@collection, tag) : tag_path(tag), link_text, options)
```
sarken: I'm _pretty_ sure these are okay, but while you're checking on the byline change, it would be a good idea to also do a quick check of tag links in emails, RSS feeds, downloads, and the share code just to be sure.

zz9pzza:
Thanks sarken I will look at that and fix it, I can't see an obvious way of testing the content of the downloads. We might be able to do it for html. If we had multiple urls we supported, would the correct way be to have a set url at the front ( eg a download generated via a "insecure or secure" url should have the canonical url for the site be from the archive config "APP_URL"

hlieberman:
For what it's worth, the branch that's currently on test.a.o seems to have solved a mixed-content issue that happens with the user icons off of the s3 bucket when you expand comments on a work.

zz9pzza:
I have done the downloads. I have checked the rss and that those need to be absolute.The share code is absolute.

app/helpers/tags_helper.rb
```diff
@@ -76,7 +76,11 @@ def link_to_tag_works(tag, options = {})
   end
 
   def link_to_tag_with_text(tag, link_text, options = {})
-    link_to_with_tag_class(@collection ? collection_tag_path(@collection, tag) : tag_path(tag), link_text, options)
+    if options[:full_path]
```
sarken: Would it be possible to do something like this to make it DRYer?
```
def link_to_tag_with_text(tag, link_text, options = {})
  collection_url_for_tag = options[:full_path] ? collection_tag_url(@collection, tag) : collection_tag_path(@collection, tag)
  link_to_with_tag_class(@collection ? collection_link_for_tag : tag_path(tag), link_text, options)
end
```

My other question is do we _definitely_ always want `tag_path`, or should it be `tag_url` when `options[:full_path]`?

app/helpers/works_helper.rb
```diff
@@ -134,7 +134,7 @@ def download_url_for_work(work, format)
   # Generates a list of a work's tags and details for use in feeds
   def feed_summary(work)
     tags = work.tags.group_by(&:type)
-    text = "<p>by #{byline(work, visibility: 'public')}</p>"
+    text = "<p>by #{byline(work, { visibility: 'public', full_path: true})}</p>"
```
sarken: Can you add a space between `true` and the `}`?

app/helpers/works_helper.rb
```diff
@@ -152,7 +152,7 @@ def feed_summary(work)
         else
           type.pluralize
         end
-        text << "<li>#{label}: #{tags[type].map{ |t| link_to_tag_works(t) }.join(', ')}</li>"
+        text << "<li>#{label}: #{tags[type].map{ |t| link_to_tag_works(t,{full_path: true}) }.join(', ')}</li>"
```
sarken: I think there needs to be a space after the `,` and before the `}`

app/helpers/application_helper.rb
```diff
@@ -90,15 +90,16 @@ def byline(creation, options={})
     if creation.respond_to?(:anonymous?) && creation.anonymous?
       anon_byline = ts("Anonymous")
       if (logged_in_as_admin? || is_author_of?(creation)) && options[:visibility] != 'public'
-        anon_byline += " [".html_safe + non_anonymous_byline(creation) + "]".html_safe
+        anon_byline += " [".html_safe + non_anonymous_byline(creation,options[:only_path]) + "]".html_safe
```
sarken: Missing space between the `,` and `options[:only_path]`

app/helpers/application_helper.rb
```diff
@@ -284,15 +285,15 @@ def params_without(name)
...
+    span = content_tag(:span, max, id: "#{field_id}_counter", class: "value", "data-maxlength" => max, "aria-live" => "polite", "aria-valuemax" => max, "aria-valuenow" => field_id)
+    content_tag(:p, span + ts(' characters left'), class: "character_counter")
   end
 
   # expand/contracts all expand/contract targets inside its nearest parent with the target class (usually index or listbox etc)
   def expand_contract_all(target="index")
-    expand_all = content_tag(:a, ts("Expand All"), :href=>"#", :class => "expand_all", "target_class" => target, :role => "button")
-    contract_all = content_tag(:a, ts("Contract All"), :href=>"#", :class => "contract_all", "target_class" => target, :role => "button")
-    content_tag(:span, expand_all + "\n".html_safe + contract_all, :class => "actions hidden showme", :role => "menu")
+    expand_all = content_tag(:a, ts("Expand All"), href:"#", class: "expand_all", "target_class" => target, role: "button")
```
sarken: Missing space between `href:` and `"#"`

app/helpers/application_helper.rb
```diff
@@ -284,15 +285,15 @@ def params_without(name)
...
+    contract_all = content_tag(:a, ts("Contract All"), href:"#", class: "contract_all", "target_class" => target, role: "button")
```
sarken: Missing space between `href:` and `"#"`

app/helpers/tags_helper.rb
sarken: Actually, ignore my question about DRYness -- it's so much easier to read the way it is, I think we can probably leave it. >.> 

However, the question about path vs. url remains.

zz9pzza: If we want the full url then we should se the option to be full_path otherwise not :)

sarken: What I mean is you're using `tag_path` even when `full_path` is true.

zz9pzza: your right I will look at that

shalott:
The code looks okay to me reading it over but I am definitely leery of whether the existing tests adequately cover this change. Would there be any way to crawl the live production AO3 site to find all the instances of "http://(www.)?archiveofourown" that currently exist so we can check them?

zz9pzza:
"The code looks okay to me reading it over but I am definitely leery of whether the existing tests adequately cover this change. Would there be any way to crawl the live production AO3 site to find all the instances of "http://(www.)?archiveofourown" that currently exist so we can check them?"

Not easily and I would point out that we don't need to rely on this code immediately we can start out with a secure.archiveofourown.org which we can examine and look at before the migration to https by default.

So in short I can't reliably crawl ao3.org.

PR #2399, opened 2016-03-19
AO3-4737 Start using strong parameters

zz9pzza:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-4737

## Purpose

Rails 4 has moved the matter of mass assignment protection from the model into the controller. Instead of specifying accessible or protected attributes in the model, Rails now requires us to inform the controller which params it should allow through from the outside world into our application.
strong_parameters is a gem that will help ease us into the Rails 4 way, by changing mass assignment protection model by model and controller by controller. Fewer conflicts, less code rot!

## Testing

I would check that tag_wrangling guidelines pages still allow changes to be made.


CristinaRO:
@zz9pzza  Would you please paste the output of the added script? (`topological_active_record.rb`)
That's the part that took me the most puzzling out, and I can't run it locally because of database shenanigans, so I'd like to see what it says. :)

As far as I can understand it, the big picture story goes like this:

Rails 4 has moved the matter of mass assignment protection from the model into the controller. Instead of specifying accessible or protected attributes in the _model_, Rails now requires us to inform the _controller_ which params it should allow through from the outside world into our application.

When (not if...) we switch to Rails 4, we could either do it as one big painful switch, or we could start building it up in ways that are compatible with Rails 3, so we can build our immunity / compatibility in steps.

`strong_parameters` is a gem that will help ease us into the Rails 4 way, by changing mass assignment protection model by model and controller by controller. Fewer conflicts, less code rot!

zz9pzza:
The script topological_active_record.rb comes straight from the gem home page ( https://github.com/rails/strong_parameters ) which also has a good description which matches what you say.

When I run it we get:

WranglingGuideline
TranslationNote
TagSet
OwnedTagSet
TagSetNomination
FreeformNomination
FandomNomination
RelationshipNomination
CharacterNomination
CastNomination
TagNomination
Role
PromptRestriction
OwnedSetTagging
PotentialMatchSettings
KnownIssue
InviteRequest
Feedback
PromptMeme
GiftExchange
Collection
CollectionProfile
WorkSearch
TagSearch
PseudSearch
BookmarkSearch
Search
Warning
UnsortedTag
Relationship
Rating
Media
Freeform
Fandom
Character
Category
Banned
ApiKey
AdminPostTag
AdminBanner
AdminActivity
AbuseReport
Admin
User
LogItem
Tag
TagSetAssociation
SetTagging
MetaTagging
CommonTagging
FilterCount
WranglingAssignment
Reading
WorkSkin
Skin
SkinParent
Preference
Profile
Pseud
ChallengeSignup
PotentialMatch
Offer
Request
PotentialPromptMatch
Prompt
TagSetOwnership
CollectionParticipant
FavoriteTag
FannishNextOfKin
ExternalAuthor
ExternalAuthorName
UserInviteRequest
Invitation
RolesUser
ExternalWork
CollectionPreference
Series
AdminSetting
Language
ArchiveFaq
Question
Question::Translation
ArchiveFaq::Translation
AdminPost
AdminPostTagging
Locale
Work
StatCounter
WorkLink
ChallengeClaim
ChallengeAssignment
Subscription
Gift
RelatedWork
SerialWork
Chapter
Kudo
Comment
InboxComment
ExternalCreatorship
Creatorship
Bookmark
CollectionItem
Tagging
FilterTagging
Audited::Adapters::ActiveRecord::Audit

So I have started on the wrangling_guidelines.

app/controllers/wrangling_guidelines_controller.rb
```diff
@@ -72,4 +72,13 @@ def destroy
 
     redirect_to(wrangling_guidelines_url)
   end
-end
\ No newline at end of file
+
+   private
```
CristinaRO: This looks misaligned, at least on the website. Would you please check that the tabs in the file are made of 2 spaces, and that they're in the correct number?

app/controllers/wrangling_guidelines_controller.rb
```diff
@@ -72,4 +72,13 @@ def destroy
...
+  def wrangling_guidelines_params
```
CristinaRO: Style note: it makes more sense to name the method `wrangling_guideline_params`, as in general we're talking about one generic wrangling guideline, or one specific wrangling guideline being updated.

Content note:
- we already have an admin only `before` filter, so having an `if` branch doesn't seem to add any value in this method
- this code is not functionally equivalent to what we had before; before we were protecting `content_sanitizer_version ` from being updated through the form, now we're permitting everything, including any future attributes to be added to the model.

The right way to implement strong params is to list out the params that we really want to permit to be set through the form, in this case `title` and `content`.

shalott:
Is there an issue for this that describes something of how to test it?

zz9pzza:
How to test this is not clear to me, these posts look relevant however I don't claim to be a rails expert. More than one person says "Testing strong params basically feels like "testing Rails stuff" which is largely unnecessary as it's already well tested." however it is more along have we configured it correctly ?

http://stackoverflow.com/questions/24292162/how-to-test-strong-params-with-rspec
https://blog.pivotal.io/labs/labs/rails-4-testing-strong-parameters

sarken:
Does this need coder action or has the coder actioned the review? I'm taking off the "Awaiting review" label because this has clearly had some degree of review; it's just not immediately clear if it's waiting for @CristinaRO or @shalott to finish the review or @zz9pzza to do something.

app/controllers/wrangling_guidelines_controller.rb
```diff
@@ -29,47 +28,52 @@ def manage
...
     end
   end
 
   # reorder FAQs
   def update_positions
     if params[:wrangling_guidelines]
-      @wrangling_guidelines = WranglingGuideline.reorder(params[:wrangling_guidelines])       
+      @wrangling_guidelines = WranglingGuideline.reorder(params.permit(:utf8, :commit, wrangling_guidelines: [])[:wrangling_guidelines])
       flash[:notice] = ts('Wrangling Guidelines order was successfully updated.')
-    elsif params[:wrangling_guideline]
```
zz9pzza: As far as I can see this path is not used in the code. I removed it in branch to increase the coverage however I think here is a better place as I don't want to add strong parameters to a bit of code which I don't believe is used and I can't test

shalott:
James, I'm sorry, I really don't get what this code is doing based on reading this pullreq, which makes it hard to review. I get that strong_parameters is trying to make sure that we don't allow undesired editing of attributes, but I don't see what effect it is having here. If the only point here is you're trying to stick in the gem and have it work and not blow anything up, then the question for me  is whether a test exists for wrangling guidelines that covers the existing, desired behavior which should remain unchanged. If so, then please link me to that test & what file it is in. If not, that test should be written.

zz9pzza:
features/tags_and_wrangling/wrangling_guidelines.feature

shalott:
OK, the code is covered by https://github.com/otwcode/otwarchive/blob/master/features/tags_and_wrangling/wrangling_guidelines.feature so I think we're good.

app/controllers/wrangling_guidelines_controller.rb
```diff
@@ -29,47 +28,52 @@ def manage
...
-      render :action => 'edit'
+      render action: 'edit'
     end
   end
 
   # reorder FAQs
   def update_positions
     if params[:wrangling_guidelines]
-      @wrangling_guidelines = WranglingGuideline.reorder(params[:wrangling_guidelines])       
+      @wrangling_guidelines = WranglingGuideline.reorder(params.permit(:utf8, :commit, wrangling_guidelines: [])[:wrangling_guidelines])
```
shalott: OK, so now am not clear if this is actually desirable behavior. This method should not be used to mass-edit wrangling guidelines, right? It should only be used to change the positions. So wouldn't you want to use strong_parameters here to permit only position information? Also why is :utf8 being permitted here?

zz9pzza: So I set test to error if we pass anything in that is not allowed.

We pass in both utf8 and commit so if we want the test to pass we need to allow it.

However we then only use wrangling_guidelines

shalott: But that's like defeating the actual underlying purpose. IMO what you should rather do is delete utf8 and commit out of the params before you permit on them if you are relying on them never being used.

zz9pzza: But we are only using wrangling_guidelines I could remove the other bits but I think it would be counter intuitive. The raising of the error only happens in test. In production strong_params will get rid of params we do not permit.

shalott: right, but the point is you ARE permitting these params. If you don't want them, don't permit them. I don't think it's counter-intuitive at all, that's the whole idea -- I mean, you could just permit! everything, and it would all work instantly, but you'd be breaking the point of the gem/Rails change, which is that it is a bad idea to permit attributes unless you explicitly want and need them. utf8 and commit are just housekeeping params that are not required past this point so deleting them seems to me the best thing to do. 

And again, I think in this method, even within wrangling_guidelines we should only be permitting position information. We wouldn't want someone submitting new text to multiple wrangling guidelines here, right?

zz9pzza: We are permitting utf8 and commit but then we are throwing them away. I am willing to remove them if you prefer but it is less efficient that way.

Additionally wrangling_guidelines is just an array of int's and we are limiting it to an array.

sarken:
Can you make an issue for this and update the pull request's subject/description accordingly? Thanks!

zz9pzza:
@sarken Is that sufficient ?

sarken:
Yep, pretty good! The only thing I'd ask is that you also add the testing info to the JIRA issue, since QA are the folks who need it and they don't generally look at the pull requests.

Thanks!

app/controllers/wrangling_guidelines_controller.rb
```diff
@@ -9,7 +8,7 @@ def index
 
   # GET /wrangling_guidelines/1
   def show
-    @wrangling_guideline = WranglingGuideline.find(params[:id])
+    @wrangling_guideline = WranglingGuideline.find(wrangling_guideline_id)
```
sarken: I'm just starting to do some reading, but everything I've read so far continues to use the old `find(params[:id])` approach here while switching to strong parameters for `new` and `update_attributes`. Do you have a link that explains why it's a good idea to the new approach for `find` as well?

app/controllers/wrangling_guidelines_controller.rb
sarken: Sooooo, for what it's worth, we're not doing [mass assignment](http://guides.rubyonrails.org/v3.2/security.html#mass-assignment) here -- mass assignment is when you're using a parameters hash to update the attributes of a single object (or to create a single object with attributes assigned from such a hash). With `reorder`, we're only updating one attribute (`position`) of multiple `WranglingGuideline` objects. (The relevant code is `list_item.update_attribute(:position, index + 1)}` in `lib/sortable_list`.)

So, anyway, I'm pretty sure the reason we're having trouble here is we _shouldn't_ be trying to use strong parameters here at all. I checked our Brakeman report, and indeed this isn't flagged as something that has a mass assignment vulnerability. I've also tried using `update_positions` to change the titles of wrangling guidelines, and I've had no success.

kamilogorek:
Everything looks good! 👍

PR #2404, opened 2016-03-23
AO3-4522 More accessible "Read more" links on homepage

sarken:
https://otwarchive.atlassian.net/browse/AO3-4522

This applies the concept on https://www.w3.org/TR/WCAG20-TECHS/ARIA7.html. Alas, we must hash rocket our ARIA roles.


scottsds:
Looks good :+1:

PR #2405, opened 2016-03-24
AO3-4523 Fix collapse/expand fandoms link on dashboard

sarken:
https://otwarchive.atlassian.net/browse/AO3-4523

Both the Expand and Collapse options were showing on the fandoms list. We only want the one that applies to show.

### DIFF
--- a/public/stylesheets/site/2.0/09-roles-states.css
+++ b/public/stylesheets/site/2.0/09-roles-states.css
@@ -42,7 +42,7 @@ span.unread, .replied, span.claimed, .actions span.defaulted {
   font-weight: 700;
 }
 
-.hidden, body .hidden, .actions .hidden {
+.hidden, body .hidden, .actions .hidden, .actions a.hidden {
   display: none;
 }
 


public/stylesheets/site/2.0/09-roles-states.css
```diff
@@ -42,7 +42,7 @@ span.unread, .replied, span.claimed, .actions span.defaulted {
   font-weight: 700;
 }
 
-.hidden, body .hidden, .actions .hidden {
+.hidden, body .hidden, .actions .hidden, .actions a.hidden {
```
sarken: Our actions are usually lists like `<ul class="actions"><li class="hidden"><a href="">Link</a></li></ul>`, but on the dashboard fandoms lists, it's `<p class="actions"><a href="">Link</a> <a href="" class="hidden">Link</a></p>` because we only ever show one link at a time, so a list isn't very appropriate. 

Alas, this meant the `display: inline-block` we use on `.actions a` farther down our cascade was overriding the `display: none` here. Adding another selector with some heft will fix that.

PR #2407, opened 2016-03-26
AO3-4512 fix 500 on test

zz9pzza:
New format, new version of cache key

https://otwarchive.atlassian.net/browse/AO3-4512


zz9pzza:
There are some more bits coming along

PR #2409, opened 2016-04-01
AO3-4528 - Add byline steps to archivist tests

ariana-paris:
https://otwarchive.atlassian.net/browse/AO3-4528

Following recent excitement with the bylines on imports getting stuck in memcache, I noticed the archivist tests don't check for the presence of the archivist byline. This wouldn't have made any difference in the stuck memcache scenario since the code was fine, but might be useful in case something does break the byline code in future.


sarken:
Can you add a link to the issue, please? Thanks!

ariana-paris:
Doh. Added link to issue.

zz9pzza:
I can't see the bit which is not a change of name/email address in this :( sorry

features/importing/archivist.feature
```diff
@@ -10,37 +9,38 @@ Feature: Archivist bulk imports
...
   
-  Given I have an archivist "alice_ttlg"
+  Given I have an archivist "archivist"
     And the default ratings exist
-  When I am logged in as "alice_ttlg"
+  When I am logged in as "archivist"
     And I import the work "http://rebecca2525.livejournal.com/3562.html"
   Then I should see "We have notified the author(s) you imported works for"
     And I should see "Importing Test"
+    And I should see "rebecca2525 [archived by archivist]"
```
ariana-paris: Since the naming changes might be obscuring this, here is where it's testing for the presence of the archivist byline on a work whose author doesn't have an AO3 account

features/importing/archivist.feature
```diff
@@ -49,80 +49,82 @@ Feature: Archivist bulk imports
...
     | ao3   | ao3testing@dreamwidth.org |
-    And I have an archivist "heathercook"
+    And I have an archivist "archivist"
     And the default ratings exist
-  When I am logged in as "heathercook"
+  When I am logged in as "archivist"
     And I import the work "http://ao3testing.dreamwidth.org/593.html"
   Then I should see import confirmation
+    And I should see "ao3"
+    And I should not see "[archived by archivist]"
```
ariana-paris: And here is where it checks that a work imported for an author who already has an AO3 account only mentions the author and not the archivist

zz9pzza: Should there be a test that it has the correct byline and not just not the wrong one ?

shalott: FYI, james, it is testing that in the I should see "ao3" line (that's the name of the user)

shalott:
Looks good to me

sarken:
Merged, but I want to check -- is the linked issue a bug that actually exists and needs to be resolved? Or did it turn out to be a false alarm, and we just added tests to make sure it never happens for real?

ariana-paris:
Good point about the issue - it was indeed a false alarm due to memcached being stuck

sarken:
Thanks for updating the issue!

PR #2281, opened 2016-01-03
AO3-4420 Correct pluralisation on kudos emails

zz9pzza:
https://otwarchive.atlassian.net/browse/AO3-4420

Test by having a user receive a kudos from one person and then from two

PR #2410, opened 2016-04-04
AO3-4536 guest email blacklist

shalott:
See https://otwarchive.atlassian.net/browse/AO3-4536

Allows blacklisting of emails so they can't be used to post guest/anonymous comments.


app/models/admin_blacklisted_email.rb
```diff
@@ -0,0 +1,36 @@
...
+    self.email = AdminBlacklistedEmail.canonical_email(self.email) if self.email
+  end
+  
+  # Produces a canonical version of a given email reduced to its simplest form
+  # This is what we store in the db, so that if we subsequently check a submitted email,
+  #  we only need to clean the submitted email the same way and look for it.
+  def self.canonical_email(email_to_clean)
+    canonical_email = email_to_clean.downcase
+    canonical_email.strip!
+    canonical_email.sub!('@googlemail.','@gmail.')
```
zz9pzza: I could have a domain gmail.catbus.co.uk and this would be effected.

There are more than .com though....

app/models/admin_blacklisted_email.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    # strip periods from gmail addresses
+    if (matchdata = canonical_email.match(/(.+)\@gmail\.com/))
+      canonical_email = matchdata[1].sub('.', '') + "@gmail.com"
```
zz9pzza: Can we have some tests for the canonicalisation.

"james.b.e.al.@gmail.com".match(/(.+)\@gmail\.com/)[1].sub('.', '')+ "@gmail.com"
 => "jamesb.e.al.@gmail.com" 

 "james.b.e.al.@gmail.com".match(/(.+)\@gmail\.com/)[1].gsub('.', '')+ "@gmail.com"
 => "jamesbeal@gmail.com"

app/controllers/admin/blacklisted_emails_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+    end
+  end
+
+  def new
+    @admin_blacklisted_email = AdminBlacklistedEmail.new
+  end
+
+  def create
+    @admin_blacklisted_email = AdminBlacklistedEmail.new(params[:admin_blacklisted_email])
+
```
zz9pzza: There isn't a test to make sure that the email address being added is not already in the system ?

app/validators/email_blacklist_validator.rb
```diff
@@ -0,0 +1,10 @@
+class EmailBlacklistValidator < ActiveModel::EachValidator
+  def validate_each(record,attribute,value)
+    if AdminBlacklistedEmail.is_blacklisted?(value)
+      record.errors[attribute] << (options[:message] || I18n.t('email_blacklisted', default: "has been blocked at the owner's request. That means it can't be used in guest comments. Please check the address to make sure it's yours to use and contact AO3 Support if you have any questions."))
```
zz9pzza: Using the default option for I18n.t will putting things in to the the app we plan to use translation more difficult. I am not convinced the key is in the format we would expect ? 

https://github.com/otwcode/otwarchive/pull/2281/files might be useful ( if 'email_blacklisted' is a full key then default could be used usefully )

app/views/admin/blacklisted_emails/_form.html.erb
```diff
@@ -0,0 +1,21 @@
+<%= form_for(@admin_blacklisted_email) do |f| %>
+  <% if @admin_blacklisted_email.errors.any? %>
+    <div id="error_explanation">
+      <h2><%= pluralize(@admin_blacklisted_email.errors.count, "error") %> prohibited this admin_blacklisted_email from being saved:</h2>
```
zz9pzza: There is a different way to do this if we head to i18n....

http://guides.rubyonrails.org/i18n.html#pluralization

zz9pzza:
@sarken can you comment on the front end bits ?

sarken:
@zz9pzza I'm working on the front end, just haven't had time to finish

zz9pzza:
Thanks, merging this would mean I wouldn't have to worry about trying to do difficult things in the mail server as a stop gap.

app/models/admin_blacklisted_email.rb
shalott: Actually no it wouldn't, it would only affect googlemail.catbus.co.uk. But anyway I'll put in the .com as that is the most common domain suffix and possibly the only working one now. I do point out, the goal here is to prevent trolls from successfully sending emails to a user who doesn't want them. We will only be entering email addresses that users write to us to ask to put into the blacklist. If a user complains of trolling or spam and they have created some weird email system for themselves or have an easily spoofed email address (more likely actually would be that they have a "catchall" email address where [anything]@userdomain.com goes to them) then ultimately we may have to tell them we can't stop people spoofing them.

app/models/admin_blacklisted_email.rb
shalott: I do have tests (in the spec) but missed this variant, whoops. Any other suggestions to test?

zz9pzza: Your right, I was reading thought the pull slowly and I hand't got that far. Sorry.

spec/models/admin_blacklisted_email_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+    before(:each) do
+      @existing_email = FactoryGirl.create(:admin_blacklisted_email, email: "foobar@gmail.com")
+    end
+    
+    it "match themselves" do
+      expect(AdminBlacklistedEmail.is_blacklisted?("foobar@gmail.com"))
+    end
+    
+    it "match variants" do
+      expect(AdminBlacklistedEmail.is_blacklisted?("FOOBAR@gmail.com")).to be_truthy
```
zz9pzza: Maybe also upper case the domain name ?

spec/models/admin_blacklisted_email_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+      expect(AdminBlacklistedEmail.is_blacklisted?("foobar+baz@gmail.com")).to be_truthy
```
zz9pzza: foobar+bar+fred@gmail.com works.

spec/models/admin_blacklisted_email_spec.rb
zz9pzza: I would also put a googlemain.com version of each test and a non google version in.

app/controllers/admin/blacklisted_emails_controller.rb
shalott: There is, it's validated for uniqueness in the model and there's a db constraint as well.

shalott:
Have added more tests, fixed the bug james found, and updated the internationalization. Also got rid of the new/form entirely, errors take you back to the original form on the index page.

config/locales/en-US.yml
```diff
@@ -3,6 +3,10 @@
 # Use this as the base for the locale file of your language.
 
 "en-US":
+  admin:
+    blacklist:
+      emails_found: "%{emails_count} emails found"
```
zz9pzza: Minor wibble if we are doing this it would be ( I think ) from ( http://guides.rubyonrails.org/i18n.html#pluralization ) also see https://github.com/otwcode/otwarchive/pull/2281

blacklist:
  emails_found:
     one: an email found
     other: "%{emails_count} emails found"

zz9pzza:
I have made a comment in the i18n section but given how much work we need to do I would merge as is if needed if sarken ok's front end bit.

shalott:
Grrr, attempting to merge sarken's changes effed up the whole thing, closing this

PR #2415, opened 2016-04-05
AO3-4538 Resize news post images to fix screen

sarken:
https://otwarchive.atlassian.net/browse/AO3-4538


sarken:
james reviewed, but we decided to wait to merge until around the time he gets up to avoid a naked staging environment, which would interfere with testing.

PR #2295, opened 2016-01-10
AO3-4430 update best_in_place

zz9pzza:
https://otwarchive.atlassian.net/browse/AO3-4430

https://github.com/bernat/best_in_place

Comes from : https://github.com/otwcode/otwarchive/commit/bfb492dc93436839f969d998b40f339c60317104

I am guessing auto complete things to test ( I would also test the ajax kudos and comment bits just in case )

PR #2431, opened 2016-04-09
AO3-4430 remove best_in_place

zz9pzza:
Tests pass without best in place.

https://otwarchive.atlassian.net/browse/AO3-4430

Closes : https://github.com/otwcode/otwarchive/pull/2295


shalott:
I think we should also remove the best_in_place.js javascript from the public/javascripts folder?

shalott:
All good!

PR #2433, opened 2016-04-10
AO3-4544 comment bookmarks

shalott:
https://otwarchive.atlassian.net/browse/AO3-4544

Adds "Comment And Bookmark" option to the top-level comment form to streamline the process of doing both things.


sarken:
I think this might be a real failure -- it relates to bookmark_exists?, which is used here

shalott:
The bug is fixed but I'm leaving this 'needs coder action' until after I finish discussing with Sarken and bingeling re the UI

PR #2435, opened 2016-04-11
AO3-4547 Allow banner updates WITHOUT turning banner back on

sarken:
https://otwarchive.atlassian.net/browse/AO3-4547

We want to make it so admins can fix typos or update fundraising totals in admin banners without turning the banner back on instead of having to ask poor james to do it from the console.

### DIFF
--- a/app/controllers/admin/banners_controller.rb
+++ b/app/controllers/admin/banners_controller.rb
@@ -4,7 +4,7 @@ class Admin::BannersController < ApplicationController
 
   # GET /admin/banners
   def index   
-    @admin_banners = AdminBanner.order("id DESC").paginate(:page => params[:page])
+    @admin_banners = AdminBanner.order("id DESC").paginate(page: params[:page])
   end
 
   # GET /admin/banners/1
--- a/app/controllers/admin/banners_controller.rb
+++ b/app/controllers/admin/banners_controller.rb
@@ -43,7 +43,10 @@ class Admin::BannersController < ApplicationController
   def update
     @admin_banner = AdminBanner.find(params[:id])
 
-    if @admin_banner.update_attributes(params[:admin_banner])
+    if @admin_banner.update_attributes(params[:admin_banner]) && params[:admin_banner_minor_edit]
+      flash[:notice] = ts('Updating banner for users who have not already dismissed it. This may take some time.')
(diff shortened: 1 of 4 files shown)


app/controllers/admin/banners_controller.rb
```diff
@@ -43,7 +43,10 @@ def create
   def update
     @admin_banner = AdminBanner.find(params[:id])
 
-    if @admin_banner.update_attributes(params[:admin_banner])
+    if @admin_banner.update_attributes(params[:admin_banner]) && params[:admin_banner_minor_edit]
```
sarken: There may be a neater way to do this. If so, I am all ears!

features/other_a/banner_general.feature
```diff
@@ -90,3 +90,30 @@ Scenario: Admin can delete a banner and it will no longer be shown to users
...
+  Then I should not see "This is a minor update (Do not turn the banner back on for users who have dismissed it)"
+
+Scenario: Admin can make minor changes to the text of an active banner without turning it back on for users who have already dismissed it
+  Given there are no banners
+    And an admin creates an active banner
+    And I am logged in as "banner_tester_3" with password "nobanners"
+    And I set my preferences to turn off the banner showing on every page
+    And an admin makes a minor edit to the active banner
+  When I am logged in as "banner_tester_3" with password "nobanners"
+  Then I should not see "This is some banner text!"
```
shalott: This is nitpicky, but ideally you'd have the actual text in one file. What I mean is, either your earlier step should be "And an admin adds "This is some banner text!" to the active banner as a minor edit" and change the step definition to take custom text, or you should have this step should be "Then I should not see the minor edit" and make the step definition in the banner_steps file look for the text.

app/controllers/admin/banners_controller.rb
shalott: Actually, doing it this way you'll be updating_attributes twice if it is not a minor edit. I suggest you flip it, ie:

```
if !@admin_banner.update_attributes(params[:admin_banner])
  render action: 'edit'
elsif params[:admin_banner_minor_edit]
  minor edit code here
else
  normal success code here
end
```

sarken:
Tests updated, btw! I forgot to comment, whoops.

PR #2436, opened 2016-04-11
AO3-3597 Let admins edit work languages

sarken:
https://otwarchive.atlassian.net/browse/AO3-3597

Pull the language field out into partial and plop it down in the edit_tags page for admins


shalott:
Looks good!

PR #2142, opened 2015-07-14
AO3-3719 Return to the restricted work after login

zz9pzza:
This makes it more explicit in the code I hope, the login url will look like 

http://ao3.org/login?restricted=true&return_to_url=%2Fworks%2F17

https://otwarchive.atlassian.net/browse/AO3-3719

PR #2437, opened 2016-04-11
AO3-3719 Tests to show login after visiting a restrict work does the right thing

zz9pzza:
This works locally ( it would be nice if it failed as it does in beta )....

This is the test for this issue : https://github.com/otwcode/otwarchive/pull/2142

https://otwarchive.atlassian.net/browse/AO3-3719


features/step_definitions/user_steps.rb
```diff
@@ -267,4 +268,11 @@ def get_series_name(age, classname, name)
 Then /^I should get confirmation that I changed my username$/ do
   step(%{I should see "Your user name has been successfully updated."})
 end
- 
\ No newline at end of file
+ 
+When(/^I am login in to the current page using "([^"]*)" as the username and  "([^"]*)" as the password$/) do |username, password|
```
shalott: should be "I login to the current page"

features/step_definitions/work_steps.rb
```diff
@@ -502,3 +508,11 @@
 Then /^the work "([^\"]*)" should be deleted$/ do |work|
   assert !Work.where(title: work).exists?
 end
+
+Then /^I remmeber the url$/ do
+  @remmeber = current_url
```
shalott: needs to be corrected to "remember" in all places

shalott:
The fact that this does not fail on local is a problem since it means we don't know why it is failing, so we have no way of testing the fix. We need to know why this is passing in test (and dev?) and get it to fail properly before we use the fix IMO.

features/step_definitions/user_steps.rb
ariana-paris: In fact, I would even argue that there is no need for an additional step, since this is almost identical to `I am logged in as "([^\"]*)" with password "([^\"]*)"` on line 39, which could be extended to have a `(to the current page)` clause that turns off the `visit login_url` line. For example:

zz9pzza:
I would point out that test and dev are not production. Additionally these are tests which are not a bad thing.... I will comment on the other pull

redsummernight:
We'll combine this with #2142.

PR #2438, opened 2016-04-18
AO3-4548 Allow for the reordering of FAQ Questions

scottsds:
Resolves: https://otwarchive.atlassian.net/browse/AO3-4548


app/controllers/questions_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+  # GET /archive_faq/:archive_faq_id/questions/manage
+  def manage
+    @questions = @archive_faq.questions.order('position')
+  end
+
+  # fetch archive_faq these questions belong to from db
+  def load_archive_faq
+    @archive_faq = ArchiveFaq.find_by_slug(params[:archive_faq_id])
+    unless @archive_faq.present?
+      flash[:error] = ts("Sorry, we couldn't find the ArchiveFaq you were
```
sarken: I think as long as we're writing the error message out ourselves, we should use the more natural terminology -- FAQ -- instead of the technical name (ArchiveFAQ).

app/controllers/questions_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+    if params[:questions]
+      @archive_faq = ArchiveFaq.find_by_slug(params[:archive_faq_id])
+      @archive_faq.reorder(params[:questions])
+      flash[:notice] = ts("Question order has been successfully updated.")
+    elsif params[:question]
+      params[:question].each_with_index do |id, position|
+        Question.update(id, :position => position + 1)
+        (@questions ||= []) << Question.find(id)
+      end
+      flash[:notice] = ts("Question order has been successfully updated.2")
```
sarken: Is the 2 at the end intentional?

scottsds:
I cleared up those two points you had, @sarken

app/controllers/archive_faqs_controller.rb
```diff
@@ -18,7 +18,20 @@ def index
 
   # GET /archive_faqs/1
   def show
-    @archive_faq = ArchiveFaq.find_by_slug(params[:id])
+    @questions = []
+    if params[:language_id] == "en"
+      @archive_faq = ArchiveFaq.find_by_slug(params[:id])
```
shalott: This line gets called for both conditions, better to move it above the "if/else"

app/views/archive_faqs/_question_answer_fields.html.erb
```diff
@@ -42,6 +42,10 @@
       <dd>
         <%= form.text_field :screencast %>
       </dd>
+      <% if Globalize.locale.to_s != "en" %>
```
sarken: Can you switch the order of these so the checkbox is before the label? 
```
<dt><%= form.check_box :is_translated %></dt>
<dd><%= form.label :is_translated, ts("Question translated") %></dd>
```

app/views/questions/manage.html.erb
```diff
@@ -0,0 +1,41 @@
+<div id="manage-questions">
+
+  <h4 class="heading"><%= ts("Reorder questions") %></h4>
+
+  <p id="drag" class="showme"><%= ts("Drag questions to change their order.") %><p>
```
sarken: Should be `</p>` at the end

app/views/questions/manage.html.erb
```diff
@@ -0,0 +1,41 @@
...
+  <noscript><p><%= ts("Enter new question numbers.") %><p></noscript>
```
sarken: Should be `</p>` at the end

app/views/archive_faqs/_question_answer_fields.html.erb
scottsds: Done.

shalott:
The code now looks good to me although could use a test for the non-javascript version?

sarken:
Just poking and seeing how the tests are coming!

scottsds:
Sorry, I missed both @shalott and @sarken 's comments in emails. I'll try to take a look at this this upcoming weekend and get it pushed back to git so we can get this in for Docs/Support!

scottsds:
I've added the cucumber test for actually reordering the questions inside of the FAQ. This should be good to go!

features/admins/admin_reorder_faq_questions.feature
```diff
@@ -0,0 +1,48 @@
+@admin
+Feature: Admin Actions to re-order FAQs
+  As an an admin
+  I want to be able to re-order the FAQs
+
+Scenario: Re-order the FAQs
```
sarken: Just a nitpick and a suggestion! 

The suggestion: This could be shortened up a bit so it just checks the bits that are relevant to reordering FAQs -- we have a step to create several FAQs in one go, as well as a path definition that should go right to the FAQ reorder page.

The nitpick: It doesn't really _affect_ anything in the grand scheme of things, but in Cucumbers, `Given` is usually for set up steps (like `Given I am logged in as an admin`), `When` is for steps that take an action, and `Then` is for steps that check result of the action you took in the `When`s, so we might have something like:

```
Given I am logged in as an admin
  And 3 Archive FAQs exist
When I go to the FAQ reorder page
Then I should see "1. Number 1 Question" 
  And I should see etc
  And I should see etc
When I fill in etc
  And I fill in etc
  And I press "Update Positions"
  And I follow etc
Then etc
```

scottsds:
Made the changes requested! There is no easy way to get to the Reorder questions page via a path, so we're still doing that the long way.

features/admins/admin_reorder_faq_questions.feature
```diff
@@ -4,36 +4,16 @@ Feature: Admin Actions to re-order FAQs
...
-    And I fill in "archive_faq_questions_attributes_2_anchor" with "whatisao33"
-  When I press "Post"
-  Then I should see "ArchiveFaq was successfully updated"
-  When I follow "Edit"
     And I follow "Reorder Questions"
     # First confirm the current order of the questions
-    And I should see "1. What is AO3?"
-    And I should see "2. This is a second question"
-    And I should see "3. This is a third question"
+    And I should see "1. Number 1 Question."
```
sarken: This should be a `Then`, since as you point out, we're confirming that something is working as expected. Other than that, if Travis says we're good, we're good!

db/migrate/20160416163754_add_is_translated_check_to_questions.rb
```diff
@@ -0,0 +1,9 @@
+class AddIsTranslatedCheckToQuestions < ActiveRecord::Migration
```
sarken: Our schema file's version is `20160706031054`, so I think this migration needs to be regenerated with a newer date in order to run.

PR #2442, opened 2016-04-28
AO3-4359 Mass import bookmarks

ariana-paris:
https://otwarchive.atlassian.net/browse/AO3-4359

Adds bookmark importing capability to the mass import API. Open Doors sometimes import sites which include works that are links to stories on another site which is not being imported (and therefore hasn't granted permission to scrape in the story contents). This will allow those works to be imported as bookmarks on the Archive, referencing the story's currently location rather than importing it.

In order to achieve this, I have also:
- refactored the existing code so that the authentication handling and work imports are in separate files
- rationalised the API end points so that instead of `api/v1/import`, we now have `api/v1/works` and `api/v1/bookmarks` (paving the way for, say, `collections`, `users` or, idk, `videos` in future, whatever comes up next)
- added an `original_id` field to all the API requests so the remote site can pass in its own id for the story/bookmark it's importing and use that to match up the Archive's response to the relevant item (currently, the OD temporary sites rely on the responses being in the same order as the requests which is risky)
- converted `routes.rb` to use Ruby 1.9 hashes where possible


sarken:
*bats eyelashes* Can you include a link to the issue in the message at the top, please? (And edit the issue to have 0.9.34 as the release for now, as long as you're there?) Thanks!

ariana-paris:
Doh, I have no explanation for why I keep forgetting to put the ticket in the pull request.

app/controllers/api/v1/base_controller.rb
```diff
@@ -14,6 +14,28 @@ def restrict_access
...
+        unless archivist && archivist.is_archivist?
+          status = :forbidden
+          errors << "The 'archivist' field must specify the name of an Archive user with archivist privileges."
+        end
+
+        if import_items.nil? || import_items.empty?
+          errors << "No items to import were provided."
+        elsif import_items.size >= ArchiveConfig.IMPORT_MAX_WORKS_BY_ARCHIVIST
+          errors << "This request contains too many items to import. A maximum of #{ ArchiveConfig.IMPORT_MAX_WORKS_BY_ARCHIVIST }" +
+            "items can be imported in one go by an archivist."
```
sarken: Really minor complaint -- can this be rephrased to "at one time"/"at once"?

app/controllers/api/v1/bookmarks_controller.rb
```diff
@@ -0,0 +1,117 @@
...
+
+    render status: status, json: { status: status, messages: messages, bookmarks: bookmarks_responses }
+  end
+
+  private
+
+  # Set messages based on success and error flags
+  def response_message(messages)
+    if @some_success && @some_errors
+      messages << "At least one bookmark was not created. Please check the bookmark array for further information."
```
sarken: Will an archivist know what the "bookmark array" is? (I ask mostly because I'm not looking at the page and I don't know.)

ariana-paris: Good point - that's an implementation detail that doesn't need to be in the message. The developer who implements the API will presumably turn the JSON array into something more user friendly, so I'll rephrase that as "individual bookmark messages".

app/controllers/api/v1/bookmarks_controller.rb
```diff
@@ -0,0 +1,117 @@
+class Api::V1::BookmarksController < Api::V1::BaseController
```
shalott: Shouldn't this be ExternalBookmarksController? We may well want to add API for regular bookmarks at some point

spec/api/api_bookmarks_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+
+    it "should create bookmarks associated with the archivist" do
+      pseud_id = @user.default_pseud.id
+      post "/api/v1/bookmarks/import",
+           { archivist: @user.login,
+             bookmarks: [ bookmark, bookmark ]
+           }.to_json,
+           valid_headers
+      bookmarks = Bookmark.find_all_by_pseud_id(pseud_id)
+      assert_equal bookmarks.count, 2
```
shalott: I'm confused by these two last specs. I assumed from the first one that duplicate bookmarks wouldn't be created, and the spec was therefore testing that it would return 200 even if only one bookmark were created. Then the second one seems to expect both to be created? In which case the first spec doesn't seem to be testing "when only some are created". 

My gut feeling is an exact duplicate should ideally not be created? So bookmarks.count should be 1 not 2?

ariana-paris: Good point about the "some" bookmarks spec. When I originally wrote the works controller, I returned different HTTP responses for these different states (all, none or some created). But as they're application errors, not HTTP errors, I now think 200 is the code to use. Anyway, I copy-pasted those into the bookmarks spec and forgot to clean up. I'll remove that test altogether.

The second one does expect duplicates to be created because there doesn't seem to be a de-duping mechanism in the Bookmark model. If I bookmark the same work multiple times, I get multiple bookmarks (even though only one is listed under the work, interestingly). This may indeed be a unintended flaw in the Bookmark model, but it feels as if fixing it is beyond the scope of this particular ticket which aims to reproduce the web behaviour in the API (though a separate one which fixes the model and this test sounds like a good idea!)

app/controllers/api/v1/bookmarks_controller.rb
ariana-paris: With all my focus on the Open Doors use case, I hadn't thought of the two bookmark types possibly being used in future - I'll rename this and have a think about whether the URL end-point needs to be changed as well. (i.e. would it still make sense to POST both external and internal bookmarks to `api/v1/bookmarks`, or should there be a separate `api/v1/external-bookmarks`, and if so what happens to GET requests when we get around to them? Should both types be returned by a call to `bookmarks`?)

spec/api/api_bookmarks_spec.rb
shalott: FWIW, my inclination would be it's more urgent to avoid creating the duplicates here because eg if an import fails halfway you would ideally like to be able to re-run without worrying about creating duplicate bookmarks?

app/controllers/api/v1/bookmarks_controller.rb
shalott: Hmm, I was thinking about External Works, I realize now. And the actual bookmark object is the same, so all bookmark creation makes sense in BookmarksController, but where the process of creating external works belongs... IDK, probably it should be here after all and just made conditional on whether the url passed is for an AO3 work or not, that seems most parallel to the existing process since External Works controller just makes the new external work object and passes off the actual creation to the Bookmarks controller via the form.

spec/api/api_bookmarks_spec.rb
ariana-paris: Good point. I'll see if it's easy to do without delving too intimately into the Bookmark model - for example, by searching for a bookmark owned by the archivist which already has that URL (sounds easy enough but we'll see!)

app/controllers/api/v1/bookmarks_controller.rb
```diff
@@ -0,0 +1,128 @@
...
+  respond_to :json
+
+  def create
+    archivist = User.find_by_login(params[:archivist])
+    external_bookmarks = params[:bookmarks]
```
shalott: here is where I think the naming is confusing me -- these are just bookmarks and they happen to be of external works. Can we just call them bookmarks throughout? it looks to me like this code would actually work just fine for a NON external work too.

ariana-paris: I'm not sure it would work for the other types as is, because the `external_bookmark` method in the API `bookmark_controller` coerces the JSON input into a structure with an `external` hash in it, which IIRC is the structure passed back by the normal `bookmark_controller` when creating a bookmark for an external work. I think this brings the `external=` virtual attribute in the `Bookmark` model into play. 

Having said that, I didn't inspect the other bookmark types, and it's been a while since I unpicked how bookmarks are created, so you might well be right. I'll rename that variable to make it less external-centric which will make it easier to adapt this if we have a use case for creating other types of bookmarks in future.

app/controllers/api/v1/bookmarks_controller.rb
```diff
@@ -0,0 +1,127 @@
...
+      messages << "None of the bookmarks were created. Please check the individual bookmark results for further information."
+    else
+      messages << "All bookmarks were successfully created."
+    end
+    messages
+  end
+
+  # Returns a hash
+  def import_bookmark(archivist, params)
+    bookmark_request = external_bookmark(archivist, params)
```
shalott: Sorry to be nitpicky some more: it feels to me like even here it makes more sense to have this remain generic (eg "build_bookmark_request(user, params)") because the only place where it actually becomes specific to external works is in the currently named external_bookmark method and bookmark_errors (which is already named generically) -- and in those two places, I would suggest adding a comment explaining that for the moment this ONLY works for bookmarks for external works. There is basically no difference except if the "external" stuff is passed AFAIK.

app/controllers/api/v1/bookmarks_controller.rb
```diff
@@ -0,0 +1,127 @@
...
+    end
+
+    status = :ok if errors.empty?
+    [status, errors]
+  end
+
+  # Map Json request to Bookmark request
+  def external_bookmark(archivist, params)
+    {
+      pseud_id: archivist.default_pseud.id,
```
shalott: and last question here: what kind of cleaning is happening on this user input and where? I might suggest also a test or two in the spec of submitting bad/malicious content to make sure it doesn't get passed through.

app/controllers/api/v1/bookmarks_controller.rb
ariana-paris: As discussed previously, I've added some error handling. As a general rule, I like to delegate as much as possible to the back-end when writing an API, as the model should be performing essential data validation before anything goes into the database, and this theoretically ensures consistency between the UI and the API.

However, the ExternalWorks model doesn't check for the presence of a fandom and returns a weird message for a missing author, so I've intercepted some invalid requests in the API controller to return helpful messages from there instead, and added tests to check that this is happening.

shalott:
Looks good to me!

PR #2445, opened 2016-05-04
AO3-4544 comment bookmarks

shalott:
Allow user to create a bookmark directly from commenting.

https://otwarchive.atlassian.net/browse/AO3-4544

### DIFF
--- a/app/helpers/bookmarks_helper.rb
+++ b/app/helpers/bookmarks_helper.rb
@@ -13,6 +13,11 @@ module BookmarksHelper
     end
   end  
   
+  def bookmark_exists?(bookmarkable)
+    return false unless logged_in?
+    Bookmark.where(:bookmarkable_id => bookmarkable.id, :bookmarkable_type => bookmarkable.class.name.to_s, :pseud_id => current_user.pseuds.collect(&:id)).exists?
+  end
+  
   # returns just a url to the new bookmark form
   def get_new_bookmark_path(bookmarkable)
     return case bookmarkable.class.to_s
--- a/app/helpers/bookmarks_helper.rb
+++ b/app/helpers/bookmarks_helper.rb
@@ -13,7 +13,7 @@ module BookmarksHelper
     end
   end  
   
-  def bookmark_exists?(bookmarkable)
+  def current_user_bookmark_exists?(bookmarkable)
     return false unless logged_in? && bookmarkable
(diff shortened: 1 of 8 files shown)


sarken:
Do you mean you're going to submit a new pull request for this, or is this the one? Unfortunately, the tests appear to be failing.

Could you change it to "Also bookmark" instead of "Also Bookmark?" We're moving away from title case and question marks in labels.

app/views/comments/_comment_form.html.erb
```diff
@@ -40,7 +40,7 @@
         <% end %>
 
         <% if current_user.pseuds.count > 1 %>
-          <h4 class="heading"><%= ts("Comment as") %> <%= f.collection_select :pseud_id, current_user.pseuds, :id, :name, {:selected => (comment.pseud ? comment.pseud.id.to_s : current_user.default_pseud.id.to_s)}, :id => "comment_pseud_id_for_#{commentable.id}", :title => ts("Choose Name") %>
+          <h4 class="heading"><%= ts("Comment as") %> <%= f.collection_select :pseud_id, current_user.pseuds, :id, :name, {:selected: (comment.pseud ? comment.pseud.id.to_s  current_user.default_pseud.id.to_s)}, id: "comment_pseud_id_for_#{commentable.id}", title: ts("Choose Name") %>
```
sarken: The missing : here appears to be what's causing the test failures: `comment.pseud ? comment.pseud.id.to_s  current_user.default_pseud.id.to_s`

shalott:
Fixed the errors and a couple more Hound complaints, changed the label on the checkbox!

app/helpers/bookmarks_helper.rb
```diff
@@ -13,6 +13,11 @@ def bookmark_if_exists(bookmarkable)
     end
   end  
   
+  def bookmark_exists?(bookmarkable)
```
sarken: Pull request looks good! The name of this method is my only quibble -- bookmark_exists? makes me think "is there a bookmark for this?" rather than "does the current user have this bookmarked?" I'm not sure what else I would suggest, though. /o\

If you can think of something, that would be great, but if not, let me know and I can mark this ready to merge as it is.

shalott:
I went ahead and changed the method name to current_user_bookmark_exists? for clarity! Marking this as ready-to-merge since it's just a name change.

PR #2292, opened 2016-01-10
AO3-4427 Update mysql2 as far as rails will allow

zz9pzza:
I think if this loads its probably ok

https://otwarchive.atlassian.net/browse/AO3-4427

PR #2447, opened 2016-05-08
AO3-4566 Update Gems for Ruby upgrades

zz9pzza:
https://otwarchive.atlassian.net/browse/AO3-4566

Looking at this I would just check that things work as the things changing look as if they are used every where eg mysql, memcached etc


Gemfile
```diff
@@ -11,7 +10,7 @@ gem 'rails', '3.2.22.2'
 
 # Database
 # gem 'sqlite3-ruby', :require => 'sqlite3'
-gem 'mysql2', '0.3.10'
+gem 'mysql2', '0.3.20'
```
shalott: didn't we already change mysql2 to something higher? I might be totally misremembering...

shalott:
Gone grey

Gemfile
zz9pzza: We have a pull request for it but it has yet to be merged.

https://github.com/otwcode/otwarchive/pull/2292

sarken:
Grey again

shalott:
Gone grey again

PR #2448, opened 2016-05-08
AO3-4563 update ruby to 2.1.9

zz9pzza:
https://otwarchive.atlassian.net/browse/AO3-4563

I would do a smoke test and deploy this slowly to beta.

https://www.ruby-lang.org/en/news/2016/03/30/ruby-2-1-9-released/

As announced before, this is the last normal release of the Ruby 2.1 series. After this release we will never backport any bug fixes to 2.1 except security fixes. We recommend that you start planning to upgrade to Ruby 2.3 or 2.2.


shalott:
yep, it's got to be done and we'll just have to fix whatever breaks! FUN TIMES :P

PR #2449, opened 2016-05-09
AO3-4567 Minor fixes for ruby upgrades

zz9pzza:
https://otwarchive.atlassian.net/browse/AO3-4567

These are minor fixes pointed out by the ruby upgrades

### DIFF
--- a/app/controllers/devmode_controller.rb
+++ b/app/controllers/devmode_controller.rb
@@ -408,7 +408,6 @@ class DevmodeController < ApplicationController
       :authors => authors,
       :summary => summary,
       :fandoms => fandoms,
-      :ratings => ratings,
       :characters => characters,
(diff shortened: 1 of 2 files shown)


app/controllers/devmode_controller.rb
```diff
@@ -408,7 +408,6 @@ def create_work(name, authors, ratings, warnings, characters, relationships, fan
       :authors => authors,
       :summary => summary,
       :fandoms => fandoms,
-      :ratings => ratings,
```
sarken: Just for nitpicking's sake, I'd actually leave this one and remove the one on line 419 instead -- this one is more logically placed (it's with the other tag types).

shalott:
Looks good to me barring sarken's note

PR #2450, opened 2016-05-09
AO3-4564 upgrade to ruby 2.2.5

zz9pzza:
https://otwarchive.atlassian.net/browse/AO3-4564


shalott:
Gone grey, I assume there's no point updating this until after we do the 2.1.9 upgrade and test

shalott:
Tests failing on bundle install of raindrops -- incompatible?

sarken:
Grey, although that is somewhat expected.

shalott:
Since the Ruby 2.1.9 upgrade went fine I would suggest first merging (one at a time) the gem upgrades and then this. Marking it ready.

PR #2460, opened 2016-05-14
AO3-4532 Fixing it so Abuse reports go through. Removing language lis…

scottsds:
…ting from reports new page

Issue: https://otwarchive.atlassian.net/browse/AO3-4569
https://otwarchive.atlassian.net/browse/AO3-4570
https://otwarchive.atlassian.net/browse/AO3-4568

### DIFF
--- a/app/models/abuse_report.rb
+++ b/app/models/abuse_report.rb
@@ -55,7 +55,7 @@ class AbuseReport < ActiveRecord::Base
   end
 
   def send_report
-    return unless %w(staging production).include?(Rails.env)
+    #return unless %w(staging production).include?(Rails.env)
     reporter = AbuseReporter.new(
       title: url,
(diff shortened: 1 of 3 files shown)


app/models/abuse_report.rb
```diff
@@ -55,15 +55,15 @@ def email_and_send
   end
 
   def send_report
-    return unless %w(staging production).include?(Rails.env)
+    #return unless %w(staging production).include?(Rails.env)
```
sarken: I think you just commented this out for testing on your machine... we still want it in

sarken:
Issue: https://otwarchive.atlassian.net/browse/AO3-4569

Also, were you going to fix https://otwarchive.atlassian.net/browse/AO3-4568?

PR #2465, opened 2016-05-24
AO3-4575 Add activerecord-mysql-reconnect gem

zz9pzza:
https://otwarchive.atlassian.net/browse/AO3-4575


shalott:
Whoops, actually, doesn't this need a configuration? Per https://github.com/winebarrel/activerecord-mysql-reconnect?

shalott:
Do we really want 20 retries over 6 seconds? Wouldn't it be safer to do, say, 3 retries and then give up? I'm happy to be convinced otherwise, just want to know what the reason is for this configuration, since there's no commenting or info in the issue.

zz9pzza:
Its more of an intuition, and each try will be quick ?

shalott:
OK!

PR #2466, opened 2016-05-25
AO3-3496 Mark for Later should be immeadiate

zz9pzza:
https://otwarchive.atlassian.net/browse/AO3-3496

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -9,7 +9,7 @@ class WorksController < ApplicationController
   before_filter :check_user_status, :except => [ :index, :show, :navigate, :search, :collected ]
   before_filter :load_work, :except => [ :new, :create, :import, :index, :show_multiple, :edit_multiple, :update_multiple, :delete_multiple, :search, :drafts, :collected ]
   # this only works to check ownership of a SINGLE item and only if load_work has happened beforehand
-  before_filter :check_ownership, :except => [ :index, :show, :navigate, :new, :create, :import, :show_multiple, :edit_multiple, :edit_tags, :update_tags, :update_multiple, :delete_multiple, :search, :marktoread, :drafts, :collected ]
+  before_filter :check_ownership, :except => [ :index, :show, :navigate, :new, :create, :import, :show_multiple, :edit_multiple, :edit_tags, :update_tags, :update_multiple, :delete_multiple, :search, :marktoread, :removemarktoread, :drafts, :collected ]
   # admins should have the ability to edit tags (:edit_tags, :update_tags) as per our ToS
   before_filter :check_ownership_or_admin, :only => [ :edit_tags, :update_tags ]
   before_filter :log_admin_activity, :only => [ :update_tags ]
@@ -785,13 +785,23 @@ public
     end
   end
 
-  # marks a work to read later, or unmarks it if the work is already marked
+  # marks a work to read later
   def marktoread
     @work = Work.find(params[:id])
-    Reading.mark_to_read_later(@work, current_user)
+    Reading.mark_to_read_later(@work, current_user,true)
     read_later_path = user_readings_path(current_user, :show => 'to-read')
     if @work.marked_for_later?(current_user)
       flash[:notice] = ts("This work was <strong>added</strong> to your #{view_context.link_to('Marked for Later list', read_later_path)}.").html_safe
+    end
+    redirect_to(request.env["HTTP_REFERER"] || root_path)
+  end
+
+  def removemarktoread
+    @work = Work.find(params[:id])
+    Reading.mark_to_read_later(@work, current_user,false)
+    read_later_path = user_readings_path(current_user, :show => 'to-read')
+    unless @work.marked_for_later?(current_user)
+      flash[:notice] = ts("This work was <strong>removed</strong> to your #{view_context.link_to('Marked for Later list', read_later_path)}.").html_safe
     end
     redirect_to(request.env["HTTP_REFERER"] || root_path)
   end
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -788,8 +788,8 @@ public
   # marks a work to read later
   def marktoread
     @work = Work.find(params[:id])
-    Reading.mark_to_read_later(@work, current_user,true)
-    read_later_path = user_readings_path(current_user, :show => 'to-read')
+    Reading.mark_to_read_later(@work, current_user, true)
+    read_later_path = user_readings_path(current_user, show: 'to-read')
     if @work.marked_for_later?(current_user)
       flash[:notice] = ts("This work was <strong>added</strong> to your #{view_context.link_to('Marked for Later list', read_later_path)}.").html_safe
     end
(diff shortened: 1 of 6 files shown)


zz9pzza:
I can't see why travis didn't run, here is the run on my account

https://travis-ci.org/zz9pzza/otwarchive/builds/132877562

app/helpers/works_helper.rb
```diff
@@ -96,12 +96,16 @@ def marked_for_later?(work)
...
+    link_to ts("Mark as Read"), marktoread_work_path(work)
+  end
+
   def marktoread_link(work)
     link_to ts("Mark for Later"), marktoread_work_path(work)
   end
 
-  def markasread_link(work)
-    link_to ts("Mark as Read"), marktoread_work_path(work)
+  def removemarktoread_link(work)
```
sarken: Why are we adding this and keeping Mark as Read rather than changing `markasread_link(work)` to use the new path?

app/controllers/works_controller.rb
```diff
@@ -785,15 +785,23 @@ def update_multiple
...
     @work = Work.find(params[:id])
-    Reading.mark_to_read_later(@work, current_user)
-    read_later_path = user_readings_path(current_user, :show => 'to-read')
+    Reading.mark_to_read_later(@work, current_user, true)
+    read_later_path = user_readings_path(current_user, show: 'to-read')
     if @work.marked_for_later?(current_user)
-      flash[:notice] = ts("This work was <strong>removed</strong> from your #{view_context.link_to('Marked for Later list', read_later_path)}. It may take a while for changes to show up.").html_safe
-    else
-      flash[:notice] = ts("This work was <strong>added</strong> to your #{view_context.link_to('Marked for Later list', read_later_path)}. It may take a while for changes to show up.").html_safe
+      flash[:notice] = ts("This work was <strong>added</strong> to your #{view_context.link_to('Marked for Later list', read_later_path)}.").html_safe
```
sarken: I think we can remove the `<strong>` tags for both messages now that we're fixing the behavior that necessitated them.

app/views/works/_work_header_navigation.html.erb
```diff
@@ -51,10 +51,15 @@
       <%= link_to ts("Cancel Bookmark"), "#", :class => "hidden bookmark_form_placement_close" %>
     </li>
 
-    <% unless current_user.is_author_of?(@work) || current_user.try(:preference).try(:history_enabled) == false || marked_for_later?(@work) %>
-      <li class="mark"><%= marktoread_link(@work) %></li>
+    <% unless current_user.is_author_of?(@work) || current_user.try(:preference).try(:history_enabled) == false %>
+      <% if  marked_for_later?(@work) %>
```
sarken: There's an extra space after `if` here. We also don't need to repeat the `<li>` element, since one or the other will always be true:

```
<li class="mark">
  <% if marked_for_later?(@work) %>
    [Mark as Read]
  <% else %>
    [Mark for Later]
  <% end %>
</li>
```

And while I don't mind re-adding a mark as read option here, it should use the existing "Mark as Read" text, not the new "Remove mark for later" text. Also, if we're changing the option up here to show both states, we might want to consider changing the button at the bottom of the work to show them as well.

app/controllers/works_controller.rb
```diff
@@ -9,7 +9,7 @@ class WorksController < ApplicationController
   before_filter :check_user_status, :except => [ :index, :show, :navigate, :search, :collected ]
   before_filter :load_work, :except => [ :new, :create, :import, :index, :show_multiple, :edit_multiple, :update_multiple, :delete_multiple, :search, :drafts, :collected ]
   # this only works to check ownership of a SINGLE item and only if load_work has happened beforehand
-  before_filter :check_ownership, :except => [ :index, :show, :navigate, :new, :create, :import, :show_multiple, :edit_multiple, :edit_tags, :update_tags, :update_multiple, :delete_multiple, :search, :marktoread, :drafts, :collected ]
+  before_filter :check_ownership, :except => [ :index, :show, :navigate, :new, :create, :import, :show_multiple, :edit_multiple, :edit_tags, :update_tags, :update_multiple, :delete_multiple, :search, :marktoread, :removemarktoread, :drafts, :collected ]
```
sarken: Can we do `markasread` rather than `removemarktoread`?

app/helpers/works_helper.rb
zz9pzza: There are two paths, one to mark as read and one to remove the mark as read ?

sarken: Right now, there are three buttons:

- Mark as Read, which incorrectly uses the old path (line 99-101)
- Mark for Later, which correctly uses the old path (line 103-105)
- Remove mark for later, which uses the new path (line 107-109)

Why aren't we changing the existing Mark as Read to use the new path instead of adding a new one with different phrasing? Mark as Read is the same thing as removing from Marked for Later.

app/controllers/works_controller.rb
zz9pzza: But it is not, it is removemarktoreadlater

app/helpers/works_helper.rb
zz9pzza: Mark as Read, which incorrectly uses the old path (line 99-101)

Why do you think this is using the wrong path ?

app/controllers/works_controller.rb
sarken: The two things are functionally the same -- marking something as read is what we do to remove it from marked for later.

zz9pzza: This is no longer the case. We have a separate function for this.

sarken: But my point is, why are we keeping "Mark as Read" _and_ adding "Remove mark for later"? Why not just fix the existing one to work the way we want?

app/helpers/works_helper.rb
sarken: Because marking as read should remove the item from your marked for later list. Isn't this just adding it if it uses marktoread_work_path(work)?

app/controllers/works_controller.rb
zz9pzza: As it is cleaner.

zz9pzza: See now I have gone down a bit I understand what you mean

sarken:
@shalott Now that we have the functionality straightened out, would you mind giving this a look? Thanks!

app/controllers/works_controller.rb
shalott: I agree the naming here is not ideal. However, making "removemarktoread" into "markasread" will make it hard to distinguish from "marktoread" and also IMO the method names should match the user-facing names if at all possible. So I would suggest making marktoread -> mark_for_later and removemarktoread -> mark_as_read.

app/views/works/_work_header_navigation.html.erb
```diff
@@ -51,10 +51,15 @@
       <%= link_to ts("Cancel Bookmark"), "#", :class => "hidden bookmark_form_placement_close" %>
     </li>
 
-    <% unless current_user.is_author_of?(@work) || current_user.try(:preference).try(:history_enabled) == false || marked_for_later?(@work) %>
-      <li class="mark"><%= marktoread_link(@work) %></li>
+    <% unless current_user.is_author_of?(@work) || current_user.try(:preference).try(:history_enabled) == false %>
+     <li class="mark"> 
+      <% if marked_for_later?(@work) %>
```
shalott: Also, FWIW, this code will fix the bug with pressing the button multiple times causing a toggle effect, but I believe it will NOT fix the bug with the button appearing wrongly, because I believe that happens because Reading objects are written to the db in batches on a delay, and therefore "marked_for_later?" returns false even after the object has been marked for later.

zz9pzza: This patch moves mark to read later/mark as read to an immediate write to the database.

zz9pzza:
The running code can be seen at http://james2.archiveofourown.org/

app/views/works/_work_header_navigation.html.erb
shalott: Duh, just saw it, nevermind! So just the renaming then and it looks good to go.

zz9pzza:
Replaced by https://github.com/otwcode/otwarchive/pull/2481

PR #2468, opened 2016-05-28
Fix for AO3-4580 - inviting to hidden/anon collection causes work to become hidden/anon

shalott:
Fixes https://otwarchive.atlassian.net/browse/AO3-4580

Signed-off-by: astolat shalott <shalott@gmail.com>


sarken:
I am unfortunately not sure why this is failing, but marking Needs Coder Action.

shalott:
Yep a bunch of the tests need fixing because they expect the behavior we don't want (ie unapproved by user work being hidden/anon), can fix them Tuesday

PR #2471, opened 2016-05-28
AO3-4547 Allow banner updates WITHOUT turning banner back on

zz9pzza:
https://otwarchive.atlassian.net/browse/AO3-4547

Reverts otwcode/otwarchive#2470


sarken:
Note: We don't want to merge this until we find out what the current behavior when updating the banner is for logged out users.

PR #2472, opened 2016-05-30
AO3 4581 cache the counts in the wrangling sidebar.

zz9pzza:
https://otwarchive.atlassian.net/browse/AO3-4581

### DIFF
--- a/features/step_definitions/tag_set_steps.rb
+++ b/features/step_definitions/tag_set_steps.rb
@@ -144,6 +144,6 @@ end
 
 When(/^I flush the wrangling sidebar caches$/) do
   [Fandom, Character, Relationship, Freeform].each do |klass|
-    Rails.cache.delete("/wrangler/counts/sidebar/#{klass.to_s}")
+    Rails.cache.delete("/wrangler/counts/sidebar/#{klass}")
   end
 end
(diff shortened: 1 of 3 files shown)


CristinaRO:
Question for clarification: is this the kind of cache that only expires when the expiration time kicks in? So, in this case, the cached count will only expire every hour for each category?

zz9pzza:
That is correct it will only expire after an hour, The tag wrangling chairs were happy to accept two hours and willing to accept 12 hours.

CristinaRO:
If @LadyOscar has pre-emptively given her 🐹 of approval, this can go in.

features/step_definitions/tag_set_steps.rb
```diff
@@ -140,4 +140,10 @@
 When /^I view the tag set "([^\"]*)"/ do |tagset|
   tagset = OwnedTagSet.find_by_title!(tagset)
   visit tag_set_path(tagset)
-end
\ No newline at end of file
+end
+
+When(/^I flush the wrangling sidebar caches$/) do
```
sarken: This issue appears to be broken on test. When you submit a fix for it, can you please move this step definition into the proper file? Right now it's in the tag _set_ steps, but it's not related to tag sets -- it should be in https://github.com/otwcode/otwarchive/blob/master/features/step_definitions/tag_steps.rb, which is where wrangling stuff lives. Thanks!

PR #2473, opened 2016-06-01
AO3-4585 wrap creating a user in a transaction 

zz9pzza:
Try and have a transaction to ensure that the user is found ( all to do with mysql clusters )

https://otwarchive.atlassian.net/browse/AO3-4585

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -155,10 +155,12 @@ class UsersController < ApplicationController
 
       @user.activation_code = Digest::SHA1.hexdigest( Time.now.to_s.split(//).sort_by {rand}.join )
 
-      if @user.save
-        notify_and_show_confirmation_screen
-      else
-        render :action => "new"
+      @user.transaction do
+        if @user.save
+          notify_and_show_confirmation_screen
+        else
+          render :action => "new"
+        end
       end
     end
   end


shalott:
I'm confused here because according to the documentation save is automatically wrapped in a transaction so I don't see why this will help. Is there a query in the "notify_and_show" method that you are trying to get included in the transaction? Also if the transaction fails then as far as I can tell the user will fall through because there is no rescue to render the :new action?

zz9pzza:
So the save has a transaction around it.

Maxscale ( https://mariadb.com/products/mariadb-maxscale ) which is what we use to steer our queries to different servers should see the wrapping transaction and ensure that the read from the database will come from exactly the same database as the write happened to. There is a small delay ( about .5 ms ) between the write and data being visible on the other servers. If the data is used in a later write then it will be ordered however reads can miss.

This is honestly a best guess about what is causing the issue.

shalott:
OK, looking at notify_and_show, it's using @user, not a new query, so I'm guessing this is happening within the deliver! method which is presumably doing a lookup since it is being sent the @user.id. If so, would a simpler solution be just making that "deliver" so it gets queued up?

zz9pzza:
"Editing to add, I remember it was originally changed to deliver! to avoid this kind of request getting stuck behind a long queue of emails going out--if that is still an issue then I agree we can go ahead and try this solution, in which case I will mark it ready to merge, lmk."

So today we have a two stage mail server, the first stage will try and deliver the mail to the remote server immediately and fall back to a second machine which queues the mail if needed.Given we get a 404 I think this would help however I still don't guarantee it, I don't think it can hurt though

PR #2476, opened 2016-06-04
AO3-4587 Error 404 when trying to edit or delete a bookmark for a deleted work

zz9pzza:
https://otwarchive.atlassian.net/browse/AO3-4587


app/views/bookmarks/confirm_delete.html.erb
```diff
@@ -3,7 +3,7 @@
 
 <!--main content-->
 <%= form_for(@bookmark, :html => {:method => :delete, :class => "simple destroy"}) do |f| %>
-  <p class="caution notice"><%= ts('Are you sure you want to <strong><em>delete</em></strong> your bookmark of "%{bookmarkable}"?', :bookmarkable => @bookmark.bookmarkable.title).html_safe %></p>
+  <p class="caution notice"><%= ts('Are you sure you want to <strong><em>delete</em></strong> your bookmark of "%{bookmarkable}"?', :bookmarkable => (@bookmark.bookmarkable.nil? ? "This has been deleted, sorry!": @bookmark.bookmarkable.title)).html_safe %></p>
```
shalott: This defeats the purpose of having it be in ts() though and also inserting "This has been deleted" into this sentence would look very odd. Doesn't this option only come up ON the bookmark itself? Do we actually need to include the name of the bookmarkable object? I suggest we remove it.

sarken: This isn't on the bookmark itself, but rather a separate confirmation page for anyone who doesn't have JavaScript enabled, so it's relatively important to make sure users know what they're deleting (in case they have memory/cognitive issues, opened multiple delete windows, opened the window and walked away, etc). I'd suggest having two versions of the message, like:

```
<% if @bookmark.bookmarkable.nil? %>
  <%= ts('Are you sure you want to <strong><em>delete</em></strong> your bookmark of this deleted work?').html_safe %>
<% else %>
  <%= ts('Are you sure you want to <strong><em>delete</em></strong> your bookmark of "%{bookmarkable}"?', bookmarkable: @bookmark.bookmarkable.title).html_safe %>
<% end %>
```

shalott: This looks good to me too.

shalott:
Note this may also fix https://otwarchive.atlassian.net/browse/AO3-3788 and also fixes an issue where users who had a bookmark of a deleted work could not delete their accounts. 

James, if you can fix the hash syntax that would be great but given the issues I am marking this ready to merge

PR #2477, opened 2016-06-05
AO3-3870 Ensure that collection is both revealed and not anonymous before sending notifications

zz9pzza:
https://otwarchive.atlassian.net/browse/AO3-3870


shalott:
This looks good but looking at the issue, I think you want one more test here for the case where the collection is both anon and hidden, then is de-anoned first (subscriptions should NOT be sent out) and then is revealed (at which point subscriptions SHOULD be sent out).

app/models/collection_item.rb
```diff
@@ -188,7 +188,7 @@ def notify_of_status_change
         notify_of_reveal
       end
     end
-    if anonymous_changed?
```
elzj: If you take out "anonymous_changed?" then this will send out emails every time the collection item is saved (whether it was ever anonymous in the first place or not), because it's a simple after_update hook. That's probably not what you want.

zz9pzza: Then we will need a migration to say we have sent the emails out ?

scottsds:
Hi @zz9pzza , don't mean to bother, but did you ever end up figuring out to ensure that reveals were only sent once? No rush, I was just looking through things to review.

tickinginstant:
I extracted the tests from this PR and ran them on the current master version, without changing any of the CollectionItem code. It passed all three new tests, so I think this may have already been fixed.

If this is also working as desired on Beta, maybe this PR can be converted to be tests-only?

sarken:
This is definitely already fixed, yeah -- I just tried both cases mentioned in the issue on staging and they behaved as expected.

sarken:
I think #2513  is what fixed this, but I'm not 100% sure.

sarken:
Closing, since we've established this has been fixed. I'll poke LO about closing the issue.

PR #2478, opened 2016-06-05
AO3 4588 Make "Make tag non-canonical and unhook all associations" function admin only

zz9pzza:
https://otwarchive.atlassian.net/browse/AO3-4588


shalott:
Test failures are real and need to be fixed (possibly by fixing the tests?)

PR #2480, opened 2016-06-08
AO3-4581 cache wrangling counts. ( BOT )

zz9pzza:
Cache the Unsorted Tags, don't particularly like this in the view

https://otwarchive.atlassian.net/browse/AO3-4581


sarken:
I agree that it doesn't really go in the view. We should move it to the controller.

sarken:
Can you make the changes Hound commented about? Thanks!

zz9pzza:
Apologies I did do that last night but i must have forgotten to push them, here is a new version of that as I am using that dev area for something else.

PR #2481, opened 2016-06-08
AO3-3496 Mark for Later should be immediate ( take 2 ) 

zz9pzza:
Replaces https://github.com/otwcode/otwarchive/pull/2466

https://otwarchive.atlassian.net/browse/AO3-3496

I think this has the renames you wanted.


app/helpers/works_helper.rb
```diff
@@ -96,12 +96,12 @@ def marked_for_later?(work)
     reading && reading.toread?
   end
 
-  def marktoread_link(work)
-    link_to ts("Mark for Later"), marktoread_work_path(work)
+  def markasread_link(work)
```
sarken: This should be `mark_as_read_link(work)` now that we're using the underscores between each word.

app/models/reading.rb
```diff
@@ -14,9 +14,13 @@ def self.update_or_create(work, user)
   end
 
   # called from reading controller
-  def self.mark_to_read_later(work, user)
-    reading_json = [user.id, Time.now, work.id, work.major_version, work.minor_version, true].to_json
-    REDIS_GENERAL.sadd("Reading:new", reading_json)
+  def self.mark_to_read_later(work, user, toread)
```
sarken: Should we be calling this `mark_for_later` instead now, or would that involve changing too many things in too many places?

app/views/works/_work_header_navigation.html.erb
```diff
@@ -51,10 +51,15 @@
       <%= link_to ts("Cancel Bookmark"), "#", :class => "hidden bookmark_form_placement_close" %>
     </li>
 
-    <% unless current_user.is_author_of?(@work) || current_user.try(:preference).try(:history_enabled) == false || marked_for_later?(@work) %>
-      <li class="mark"><%= marktoread_link(@work) %></li>
+    <% unless current_user.is_author_of?(@work) || current_user.try(:preference).try(:history_enabled) == false %>
+     <li class="mark">
```
sarken: I think the indenting is off here -- this looks like one space instead of two. Also, line 57 will need to be `mark_as_read_link(@work)` to go with the change requested in [my comment above ](https://github.com/otwcode/otwarchive/pull/2481/files#r66373553)

app/models/reading.rb
zz9pzza: It would be mark as read if anything, but that name is being used to remove something from the to read table.

sarken: Okay! In that case, this only needs https://github.com/otwcode/otwarchive/pull/2481/files#diff-47cdfed9877703d7f3ff7b154fa65728R99

zz9pzza:
@sarken I think that is all you wanted ?

PR #2489, opened 2016-06-21
AO3-4128-more thorough progress fix

shalott:
https://otwarchive.atlassian.net/browse/AO3-4128

The progress report was NOT fixed. I've reworked the way it was being done, and also added tests. Now hopefully working for reals. Tests running clean. 

Signed-off-by: astolat shalott <shalott@gmail.com>

### DIFF
--- a/app/models/potential_match.rb
+++ b/app/models/potential_match.rb
@@ -3,10 +3,10 @@ class PotentialMatch < ActiveRecord::Base
   # We use "-1" to represent all the requested items matching
   ALL = -1
 
-  CACHE_PROGRESS_KEY = "potential_match_status_for_"
-  CACHE_SIGNUP_KEY = "potential_match_signups_for_"
-  CACHE_INTERRUPT_KEY = "potential_match_interrupt_for_"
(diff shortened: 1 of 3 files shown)


app/models/potential_match.rb
```diff
@@ -104,9 +104,14 @@ def self.generate_in_background(collection_id)
       required_types = settings.required_types.map {|t| t.classify}
 
       # treat each signup as a request signup first
-      collection.signups.find_each do |signup|
+      # because find_each doesn't give us a consistent order, but we don't necessarily
```
elzj: What ordering problem did you find with find_each? The Rails documentation says: "[Order] is automatically set to ascending on the primary key (“id ASC”) to make the batch ordering work." So I would expect that to do what you want?

shalott: Actually, in retrospect I realized the issue was likely because the job-restarting code was causing the process to start over and over on top of itself. However, I think this remains a better way of handling the progress (doesn't break if user changes name/pseud midstream), so would still rather get this in.

PR #2491, opened 2016-07-09
AO3-4597 actions spacing

pixelred:
https://otwarchive.atlassian.net/browse/AO3-4597

there was a tiny design quirk that always stood out to me on mobile...
this seems to fix it on both mobile and desktop

before: 
http://imgh.us/Screen_Shot_2016-07-09_at_10.33.38.png

after:
http://imgh.us/Screen_Shot_2016-07-09_at_10.32.37.png

it's my first time contributing to an open source project so i
apologise if i made any errors > <

### DIFF
--- a/public/stylesheets/site/2.0/04-region-dashboard.css
+++ b/public/stylesheets/site/2.0/04-region-dashboard.css
@@ -19,6 +19,7 @@ http://otwcode.github.com/docs/front_end_coding/patterns/supertypes
   height: auto;
   line-height: 2;
   padding: 0 0.5625em;
+  margin: 0;
   border: none;
   background: transparent;
   vertical-align: middle;
--- a/public/stylesheets/site/2.0/08-actions.css
+++ b/public/stylesheets/site/2.0/08-actions.css
@@ -25,7 +25,6 @@ ul.actions, p.actions, li.actions, span.actions, fieldset.actions, dd.actions {
   vertical-align: middle;
   display: inline-block;
   padding: 0.25em 0.75em;
-  margin-bottom:7px !important;
   white-space: nowrap;
   overflow: visible;
   position: relative;
@@ -97,6 +96,10 @@ ol.pagination, div.pagination, ol.year {
 
 .actions label.disabled {
   background: #ccc;
+}
+
+.actions li a, .actions li input, .actions li input[type="submit"], .actions li .current, .actions li label {
+  margin: 0.375em auto;
 }
 
 .delete a, span.delete {
--- a/public/stylesheets/site/2.0/08-actions.css
+++ b/public/stylesheets/site/2.0/08-actions.css
@@ -25,6 +25,7 @@ ul.actions, p.actions, li.actions, span.actions, fieldset.actions, dd.actions {
   vertical-align: middle;
   display: inline-block;
   padding: 0.25em 0.75em;
+  margin-bottom:7px !important;
   white-space: nowrap;
   overflow: visible;
   position: relative;
--- a/public/stylesheets/site/2.0/26-media-narrow.css
+++ b/public/stylesheets/site/2.0/26-media-narrow.css
@@ -102,12 +102,6 @@ body .narrow-shown {
   clear: both;
 }
 
-/* 08 actions */
-
-#main .actions a {
-  margin: auto;
-}
-
 /* 10 types and groups */
 
 .blurb dl.tags dt, .blurb dl.tags dd, dl.meta dt, dl.meta dd, .alphabet .listbox li, .media .listbox {


sarken:
Hi, virabird!

First, thank you for your interest in contributing to the Archive! We generally prefer that independent contributors work on approved issues already listed in [our JIRA bug tracker](https://otwarchive.atlassian.net/projects/AO3/issues), but this is an issue we'd like resolved, so I have created [a new issue](https://otwarchive.atlassian.net/browse/AO3-4597) for it and edited your pull request title and comment to include the issue number and link, per [our commit guidelines](https://github.com/otwcode/otwarchive/wiki/Commit-policy#Good_Practice). 

Unfortunately, there are a few issues that will need to be fixed before we can accept this pull request.

The first few points are just a matter of making the code meet [our front end coding standards](http://sarken.github.io/docs/front_end_coding/coding-standards) (while our front end documentation is a work in progress, the information on that page is accurate):

* We format our CSS with a space separating the property and value, like `margin-bottom: 7px !important;`.
* Additionally, to make the Archive layout responsive and accessible, we use relative units rather than pixels to specify sizes. With [our em scale](http://sarken.github.io/docs/front_end_coding/em-scale#standard-scale-our-scale), 7px would be roughly 0.5em; however, a slightly bigger margin of 0.643em would be more in line with our standards.
* We also refrain from using `!important` in our CSS, since it makes it harder for users to override our layout using browser stylesheets or skins. Because we use actions in so many places on the Archive, using the `!important` declaration here creates another issue: the new margin is applied in places we don't want want it, such as the site header.

However, simply changing to `margin-bottom: 0.643em;` won't work, as you probably noticed -- some buttons end up misaligned. I'd recommend the following instead, although I haven't had time to test it extensively, so you might find some issues with it that I have overlooked:

* Remove `margin-bottom:7px !important;`
* On line 102, after the styles for `.actions label.disabled`, add the following selectors and styles to make sure we're only targeting actions that are likely to run onto multiple lines (they should be in lists): `.actions li a, .actions li input, .actions li input, .actions li .current, .actions li label {
  margin: 0.375em auto;
}`
* In 04-region-dashboard.css, add `margin: 0;` to `#dashboard a, #dashboard span` to keep from overriding the existing margins on the sidebar links on user pages

If you run into any problems or have any questions, please let me know! I'm happy to help.

Thank you again for contributing,
Sarken
AD&T Co-chair and FrontEnd Lead

virabird:
hello! i'm very sorry about that... thank you for taking the time to explain all of this.

i added `.actions li a, .actions li input, .actions li input, .actions li .current, .actions li label { margin: 0.375em auto; }` to the **08-actions.css** file and it worked great up to a point, but once the window was resized to anything smaller than 42em the buttons clustered together again. changing `#main .actions a { margin: auto; }` in **26-media-narrow.css** to `#main .actions a {margin: 0.375em auto; }` seemed to resolve the problem. but targeting only .actions lists in this stylesheet didn't work without the use of !important.

i've been testing these changes on various pages and i've yet to find an issue with it;

sarken:
Hi again! 

I just realized I made a mistake in my selectors! I included plain `input` twice instead of doing one with the `type` attribute -- sorry about that! I should've said:
```css
.actions li a, .actions li input, .actions li input[type="submit"], .actions li .current, .actions li label {
  margin: 0.375em auto;
}
```

And great catch with the narrow stylesheet! I took a look around on my iPhone and a tiny Safari window on my laptop, and it seemed like the buttons were only missing the margins on the top of works -- is that what you experienced as well, or where there other places the margins were missing?

While you're right that changing the margin on that selector fixes the issue, I _think_ we might be able to do even better. It seems like `#main .actions a { margin: auto; }` is a relic from a long, long time ago, and we should be able to remove it altogether. Have you tried that, and if not, would you like to give it a shot and see how it looks?

virabird:
oh! no worries. i just tried `.actions li a, .actions li input, .actions li input[type="submit"], .actions li .current, .actions li label { margin: 0.375em auto; }` and then all together removed `#main .actions a { margin: auto; }` from the narrow stylesheet and from what i can see it works great! it was only a problem for the buttons above the works as you said, but this fixed it :  )

sarken:
Yay -- thank you so much! I've marked this ready to merge, although it may be a while before we actually do merge it (we have a bit of a backlog, as you can see). Would you like to be credited in our release notes as virabird or something else?

virabird:
np! thank you for all your help! i'm glad it worked out. if it's alright could we use my otw volunteer name? (dolleye) since i'm thinking of changing this username

sarken:
Absolutely! In fact, that's even better (and I didn't know you were a volunteer -- hi!)!

virabird:
awesome! 
hi to you too :D

PR #2493, opened 2016-07-14
AO3-4451 Error pages need to use new @media, be more responsive

sarken:
https://otwarchive.atlassian.net/browse/AO3-4451

Our error pages were still using the old 480px stylesheets. Now they use the new 62em and 42em ones... and when the 42em stylesheet is active, the image gets bumped down under the page text so it's a bit easier to read.


public/500.html
```diff
@@ -13,119 +13,124 @@
...
   </head>
-	<body>
-		<ul id="skiplinks"><li><a href="#main">Main Content</a></li></ul>
-    <!-- BEGIN header -->
-    <div id="header" class="region">
-      <h1 class="heading">
-        <a href="/"><span>Archive of Our Own</span><sup> beta</sup><img alt="AO3 logo - the letters A O 3 combined with arms raised in celebration, symbolizing the joy of fannish creation in the Archive" class="logo" src="/images/ao3_logos/logo_42.png?1362864128" /></a> 
-      </h1>
+  <body>
+    <div id="outer" class="wrapper">
```
sarken: FUNNY STORY. Somehow, the outer div we use on all the pages was missing from these, leading to chaos and mayhem once the new stylesheets were added. It is no longer missing.

public/502.html
```diff
@@ -14,120 +14,125 @@
...
+          </li>
+        </ul>
+        <div class="clear"></div>
+      </div>
+      <!-- END header -->
+      <div id="inner" class="wrapper">
+        <div id="main" class="system errors error-502 region">
+          <h2 class="heading">Error 502</h2>
+          <h3 class="heading">The page was responding too slowly.</h3>
+          <!--plain english suggestion: There are so many people using the archive right now, we can't show your page. -->
```
ariana-paris: Do we want to implement the plain English suggestion or remove the comment if we're not?

public/503.html
```diff
@@ -14,119 +14,124 @@
...
+          </li>
+        </ul>
+        <div class="clear"></div>
+      </div>
+      <!-- END header -->
+      <div id="inner" class="wrapper">
+        <div id="main" class="system errors error-503 region">
+          <h2 class="heading">Error 503</h2>
+          <h3 class="heading">The page was responding too slowly.</h3>
+          <!--plain english suggestion: There are so many people using the archive right now, we can't show your page. -->
```
ariana-paris: I see this one doesn't even get an explanation - but that's another ticket altogether!

PR #2494, opened 2016-07-16
AO3-3788 Manage collection items page should work w deleted bookmarks

sarken:
https://otwarchive.atlassian.net/browse/AO3-3788

If a user went to their Manage Collected Works' Approved page and there was a deleted bookmark there, the page would give a 500 error. Ditto a collection owner going to the collection's Manage Items's Approved page. And double ditto the Awaiting Approval page of a collection's Manage Items section, which errored not just because of the title, but because the `collection_item.item` was nil, so `collection_item.item_date.to_date` would error.


app/views/collection_items/_item_fields.html.erb
```diff
@@ -1,64 +1,73 @@
-<li class="collection item picture blurb group">
-  <div class="header module">
-  <% item_name = collection_item.item_type == "Bookmark" ? 
-            ts("Bookmark for %{title}", :title => collection_item.item.bookmarkable.title) : 
-            (collection_item.item.title + (!collection_item.item.posted? ? ts("(draft)") : "")) %>
+<% if collection_item.item_type == "Bookmark" && collection_item.item.present? && collection_item.item.bookmarkable.present? %>
```
sarken: This should be a helper. I know. Dishonor on me and my cow. I will fix it.

shalott:
This is grey but otherwise looks good to me.

app/views/collection_items/_item_fields.html.erb
```diff
@@ -1,64 +1,63 @@
...
+          <% collection_item.item_creator_pseuds.each do |pseud| %>
+            <%= pseud.byline %>
             <% if collection_item.collection.user_is_posting_participant?(pseud.user) %>
               <%= ts("(Member)")%>
             <% end %>
           <% end %>
-        <% # TODO: Make recipients and bylines links, separate recipients with space (currently "snow,astolat") %>
-        <% unless collection_item.recipients.blank? %>
-          <span class="recipients"><%= ts("for") %> <span class="user"><%= collection_item.recipients %></span></span>
+          <% # TODO: Make recipients and bylines links, separate recipients with space (currently "snow,astolat") %>
```
ariana-paris: Would calling `to_sentence` on `collection_item.recipients` take care of this TODO ?

features/collections/collectible.feature
```diff
@@ -112,3 +112,70 @@ Feature: Collectible items
     Then I should see "Bookmarks (0)"
       And I should not see "Tundra penguins"
 
+  Scenario: Bookmarks of deleted items are included on the collection's Manage
```
ariana-paris: Only a suggestion so don't worry if you don't have time, but these scenarios could be extracted into a separate "Deleted Items in Collections" feature with the Background:
```
   Given I have a collection "Various Penguins"
      And I am logged in as a random user
```
(no big deal though)

sarken: Alas, one of the scenarios needs to be moderated, so it'll take a bit more work to make the collection moderated after the Background step. I can do that in a separate pull request and refactor the existing tests in this as well, since I think they can also be reworked to share the same background.

sarken:
Here we go! I broke them out by collection type, so everything that gets added to an open collection is in one file, moderated in another, closed another, and multiple still another. (Even though I _could_ do "I have the moderated collection" as background in the multiple collection tests, I'm not going to -- you could just as easily test two open collections in that file.)

app/views/collection_items/_item_fields.html.erb
ariana-paris: Oh well /o\

PR #2511, opened 2016-08-14
AO3-4466 Reduce the one call to skins per page

zz9pzza:
https://otwarchive.atlassian.net/browse/AO3-4466


sarken:
So this will need coder action in order to avoid the problem where changing the default Archive skin breaks everything?

sarken:
Pokes @zz9pzza -- we reverted this because it broke changing the default skin on the Archive. Should we close this until that's fixed?

zz9pzza:
I would rather keep it open as the code is right but the database in production and test is wrong, we can work on finding out what is wrong on test by changing the default skin to the right thing on test and seeing what happens.

sarken:
Will we need to make that change on staging before this is merged or can it be done after?

zz9pzza:
I would rather do it before as we can get staging working after it breaks things ( and learn what to do on beta ). Then add this pull

sarken:
Okie dokie -- gonna mark this ready to merge, but also deferred so we don't accidentally merge it before we fix the skin issue on staging.

sarken:
Travis is failing with `Your Ruby version is 2.1.9, but your Gemfile specified 2.2.5`

Also, before we merge, have we definitely taken care of the default skin issue on staging that is referenced in the comments here?

zz9pzza:
I am going to say we probably should make so we can actually successfully change the skin. I don't think this needs it but I am feeling a bit conservative right now.

sarken:
Should we take it out of this release, then? (It's not a problem -- just want to make sure things are tidy on JIRA.)

zz9pzza:
Please, the devises pull will be stress enough for me on this release :)

sarken:
Okay, I'll give it the boot!

app/models/admin_setting.rb
```diff
@@ -38,14 +39,17 @@ class << self
...
 
   def self.default_skin
-    settings = current
-    if settings.default_skin_id.present?
-      Rails.cache.fetch("admin_default_skin") { settings.default_skin }
+    if self.first
+      self.first.default_skin_id ? self.first.default_skin : Rails.cache.fetch(default_skin_cache_key) { Skin.default }
     else
-      Skin.default
+      Rails.cache.fetch(default_skin_cache_key) { Skin.default }
```
tickinginstant: #3382 introduced caching for `Skin.default`, so wrapping it in another cache block is a bit redundant (both here and above on line 43).

In general, I'm not sure what the changes to this function are intended to accomplish. Currently, I think you're wrapping a cache block around something that's already cached, and deleting a cache block around something that's not otherwise cached. Is there something subtly wrong with the current caching set-up that I'm missing? Or did the changes just get obsoleted by #3382?

app/models/admin_setting.rb
```diff
@@ -38,14 +39,17 @@ class << self
...
     end
   end
 
+  def self.stats_updated_at
```
tickinginstant: I think this is already handled by the call to `delegate` above:
https://github.com/otwcode/otwarchive/blob/9dba531c689f9bcc1eac1a9e4ac50227bb0414e2/app/models/admin_setting.rb#L37-L39

app/models/admin_setting.rb
redsummernight: Yeah, #3382 added caching for the default skin, so the changes in this PR are no longer needed. I'll close this.

I'll keep [AO3-4466](https://otwarchive.atlassian.net/browse/AO3-4466) open though, because @zz9pzza's comment from the issue is still true:

> If you are logged in there is still one look up, so better but not completely finished.
>
>     ↳ app/helpers/skins_helper.rb:25:in `set_site_skin'
>     Skin Load (1.3ms) SELECT `skins`.* FROM `skins` WHERE `skins`.`id` = 1 LIMIT 1
>     ↳ app/helpers/skins_helper.rb:26:in `set_site_skin'
>     Cache read: views//v1/site_skin_1_2016-08-14 04:41:15 -0400
>     Cache read: skins_generation/site_skin

PR #2508, opened 2016-08-13
Fix for issue AO3-3947, with cucumber scenarios to test

tickinginstant:
https://otwarchive.atlassian.net/browse/AO3-3947

I added a check to CreationObserver so that it won't trigger the UserMailer's recipient_notification if the user has gift notifications turned off in their preferences. I also created three cucumber scenarios to test recipient notifications (corresponding to the three laid out in the description of issue AO3-3947).

(I hope this is okay! I tried to follow the guidelines in the wiki, but I've never tried to contribute code to an open-source project before.)

PR #2513, opened 2016-08-19
AO3-4108 (& AO3-4367) - Fix some problems with subscriptions and anonymous or unrevealed collections.

tickinginstant:
https://otwarchive.atlassian.net/browse/AO3-4108
https://otwarchive.atlassian.net/browse/AO3-4367

Attempts to fix AO3-4108 (and AO3-4367) by removing the code for author reveals from the CollectionItem and Collection objects, and rewriting a similar system within CreationObserver's before_update function.

The bug AO3-4108, where works in two anon collections send out author subscription emails when only one collection does author reveals, meant that CollectionItem's notify_of_author_reveal needed to be able to check whether the item was in another anonymous collection (i.e. whether it would still be anonymous after the author reveal). In other words, when queuing up author subscription emails, it would be really nice to access the *new* value of item.in_anon_collection (the value that takes into account the newly-de-anonymized collection). Unfortunately, item.in_anon_collection doesn't seem to be updated before notify_of_author_reveal is called, and I didn't want to duplicate work by calculating the value twice.

I eventually decided to try to move the logic for author reveal emails to CreationObserver's before_update function, where the other half of the creator subscriber code lives. (And it gets called just before the work's value of in_anon_collection is saved, so it's guaranteed to be up-to-date.) Instead of relying on a call from Collection, it checks whether the work's creator is currently visible, and then uses in_anon_collection_changed? and in_revealed_collection_changed? to tell whether the author is newly revealed (and thus whether emails should be sent out).

As a happy accident, the code that I wrote also seems to fix AO3-4367 (the bug where author reveals aren't sent out for a collection that goes straight from unrevealed to revealed, with no anon period).

I wrote up four new cucumber tests, for four scenarios:

1. A work is in two anonymous collections and one is revealed.
2. A work is in two hidden collections and one is revealed.
3. A work is in two collections, one hidden and one anonymous, and the anonymous collection is revealed.
4. A work is in two collections, one hidden and one anonymous, and the hidden collection is revealed.

In each of these scenarios, no author subscription emails should be sent. The original code actually passes two of these tests, because AO3-4367 means that author subscriptions aren't sent out in scenarios 2 and 4. But since my code seems to fix AO3-4367, I figured it would be good to add some tests to make sure that I wasn't missing any weird edge cases.

I also uncommented the two existing tests for AO3-4367.

### DIFF
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -416,17 +416,12 @@ class Collection < ActiveRecord::Base
   
   def reveal_collection_item_authors
     approved_collection_items.each { |collection_item| collection_item.update_attribute(:anonymous, false) }
-    send_author_reveal_notifications
   end
   
   def send_reveal_notifications
     approved_collection_items.each {|collection_item| collection_item.notify_of_reveal}
   end
   
-  def send_author_reveal_notifications
-    approved_collection_items.each {|collection_item| collection_item.notify_of_author_reveal}
-  end
-
   def self.sorted_and_filtered(sort, filters, page)
     pagination_args = {:page => page}
 
--- a/app/models/collection_item.rb
+++ b/app/models/collection_item.rb
@@ -188,9 +188,6 @@ class CollectionItem < ActiveRecord::Base
         notify_of_reveal
       end
     end
-    if anonymous_changed?
-      notify_of_author_reveal
-    end
   end
   
   after_destroy :expire_caches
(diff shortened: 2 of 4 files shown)


elzj:
Looks good to me! @sarken

sarken:
Yay, thanks, @elzj! 

And thank you for the fix, @tickinginstant -- we'll get this is in as soon as possible. We'll credit tickinginstant in the release notes, as we mentioned on PR #2508, but I forgot to ask what pronouns we should use for you. (We don't generally use them in release notes, but sometimes we do need them.)

tickinginstant:
Glad to hear it! And my pronouns are she/her.

PR #2514, opened 2016-08-21
AO3-4622 Remove broken hit count preferences

sarken:
https://otwarchive.atlassian.net/browse/AO3-4622

Adding the Deferred or Superseded label right out of the gate because this cannot be merged until there has been an announcement.


app/controllers/stats_controller.rb
```diff
@@ -27,13 +27,8 @@ def index
...
     sort_options = ""
     @sort = ""
-    if current_user.preference.hide_hit_counts
-      sort_options = %w(date kudos.count comments.count bookmarks.count subscriptions.count word_count)
-      @sort = sort_options.include?(params[:sort_column]) ? params[:sort_column] : "kudos.count"
-    else
-      sort_options = %w(hits date kudos.count comments.count bookmarks.count subscriptions.count word_count)
-      @sort = sort_options.include?(params[:sort_column]) ? params[:sort_column] : "hits"
-    end
+    sort_options = %w(hits date kudos.count comments.count bookmarks.count subscriptions.count word_count)
```
sarken: The note above this says "NOTE: Because we are going to be eval'ing the @sort variable later we MUST make sure that its content is checked against the whitelist of valid options" -- does that mean it needs to be done this way, or can it be reduced to just these two lines?

sarken: ...never mind.

ariana-paris:
Needs conflict resolution anyway.

sarken:
Finally!

sarken:
Resolved conflicts and updated, but will probably need to do it again before this gets merged.

redsummernight:
This has conflicts now, plus we should update the timestamp of the migration.

db/migrate/20200415010506_remove_hits_from_preferences.rb
```diff
@@ -0,0 +1,7 @@
+class RemoveHitsFromPreferences < ActiveRecord::Migration[5.1]
+  def change
+    remove_column :preferences, :hide_all_hit_counts
+    remove_column :preferences, :hide_private_hit_count
+    remove_column :preferences, :hide_public_hit_count
```
tickinginstant, line 5: Is this table large enough that this migration needs the `pt-online-schema-change` treatment? Or can it stay as is?

(Side note: I don't think this migration is reversible, because `remove_column` can only be reversed if you include the type of the column being removed. Not sure if it needs to be reversible, though.)

redsummernight: Let's make it reversible for good practice.

This table should be 2.5m (number of users). We were at ~2m when we ran this migration:

https://github.com/otwcode/otwarchive/blob/b3bdab95dec26a004df9ec450efac72251317eec/db/migrate/20190405191806_add_allow_cocreator_to_preference.rb

and we were fine. So I think we're fine without pt-online-schema-change.

redsummernight: It was pretty slow on staging so better safe than sorry: #3805.

PR #2515, opened 2016-08-22
AO3-4999 Chapter: Make it easier to see and add co-authors who are already on the work

sarken:
https://otwarchive.atlassian.net/browse/AO3-4999
Originally: https://otwarchive.atlassian.net/browse/AO3-3752

See comment for explanation: https://github.com/otwcode/otwarchive/pull/2515#issuecomment-298780380

Gives you handy checkboxes so you can add co-creators who are already on the work to a chapter ~instead of having them automatically added for you, without knowledge or warning~.

You can do this when creating or editing a chapter. 

Once a given co-creator is added, their checkbox will be disabled and you will not be able to remove them. 

If there are no co-creators already on the work, there should be no checkboxes.

On the chapter form, where it used to say "Current co-creators," it now says "Work co-creators" so you know they do not belong to the chapter.

The other change is everything on the form (for chapters and works) now says "creator" instead of "author."


shalott:
So the code looks good, but I have a couple concerns about the actual change that aren't discussed in the issue: 

One is, whether there remains any easy way to add a coauthor on a large multichapter work -- eg, if you've imported it from another site or posted in when they didn't have an account or (probably more common) if you're transferring it to a different user account or pseud of your own. 

Second, and this is actually  a question about what happens now since the situation can already be created, what impact does this have in terms of control? eg, does it mean this user can't edit the content or tags of the chapter in question? If the chapter gets deleted will they not get a copy?  If the coauthor deletes their account and all their works, will their sole chapters get wiped out leaving the coauthor with a partial work? I ask this because whatever can happen when you have an owner of a work but not of multiple chapters will happen much more often in this scenario, so before we do something that will increase the frequency, I think we should make sure we're actually happy with the behavior.

cjrecordvt:
> One is, whether there remains any easy way to add a coauthor on a large multichapter work...

One problem is that there isn't a consistent and reliable way to do it currently (let alone easy). The issue is fairly consistent that if you try to add an author to a multichapter work, some chapters won't get the author.

> Second...does it mean this user can't edit the content or tags of the chapter in question? If the chapter gets deleted will they not get a copy? If the coauthor deletes their account and all their works, will their sole chapters get wiped out leaving the coauthor with a partial work? ...

The user can't /edit that chapter's content. I'd have to test about /edit_tags, as I don't remember if the global attributes are locked. We're uncertain about receiving a deleted copy, as we've not received a question specifically (though it would answer some of the "My work was deleted and I didn't get a copy" tickets). The chapters aren't wiped out on account deletion of one contributor, but we've had one case where it did result in a chapter that had no one "owning" it.

sarken:
> eg, does it mean this user can't edit the content or tags of the chapter in question?

This is a fun question, the answer to which has changed several times over the lifespan of this pull request.

Initially, attempting to edit a chapter you were not a co-author of would give a 500 error: https://otwarchive.atlassian.net/browse/AO3-3176

This pull request fixed that for a while -- you no longer received a 500 error and instead were allowed to edit the chapter (it would not add you as a co-author). The two commits prefixed with AO3-3176 (9ea3419 and 5d7819a) added a test for that:
```
   Scenario: Editing a chapter even if you are not a co-creator
 
     Given I am logged in as "originalposter"
       And I post the work "OP's Work"
       And a chapter with the co-author "opsfriend" is added to "OP's Work"
     When I am logged in as "opsfriend"
       And I view the work "OP's Work"
       And I follow "Edit Chapter"
     Then I should see "Chapter 1"
       And I should see "Chapter by originalposter"
     When I follow "Edit Chapter"
       And I fill in "content" with "opsfriend was here"
       And I post the chapter
     Then I should see "opsfriend was here"
       And I should see "Chapter by originalposter"
```

However, then we found and fixed https://otwarchive.atlassian.net/browse/AO3-4664. As a result of _that_, the 500 error from AO3-3176 no longer happens -- instead you get an error message about the pseud you're trying to use, as described in https://otwarchive.atlassian.net/browse/AO3-4699. 

So tl;dr: you still can't currently edit a chapter you are not a co-author of. I think I see how to change that and reproduce the behavior in the test, though.

sarken:
Whoops, I'm sorry! I can't read my own issues.

The 500 error on trying to _post_ the edits still happens in the current code. In addition to the 500 error, there is a message at the top of the form telling you you can't use the pseud.

What happens with _this_ pull request is you wouldn't get the 500 error on trying to post -- you'd get the "can't use that pseud" message again. 

I think I have a fix that will allow you to edit the chapter.

sarken:
I've fixed the merge conflicts and made two small adjustments: displaying the co-author checkboxes as "pseud (username)" rather than just "pseud", and changed single quotes to double in two places.

CJ's already covered the "no good way to add a co-author to all chapters of a multichapter work" issue, but I did some poking to get definite answers to the other questions. They should be clearer than my previous replies :P

> does it mean this user can't edit the content or tags of the chapter in question? 

(Note: adding a co-author to a chapter also adds them to the work. When they're removed from the chapters, or their chapters are deleted from the work, they stay on the work.)

Right now, there's a 500 error when trying to edit a chapter you are not co-author of ([AO3-3176](https://otwarchive.atlassian.net/browse/AO3-3176)). Pull request #2627 fixes that, and any user who is listed as a co-author on the work can edit any chapter.

Chapters don't have tags, but any user who is listed as a co-author of a work can edit the work's tags.

> If the chapter gets deleted will they not get a copy? 

We don't email deleted chapters. However, if the work is deleted and the user is still listed as a work co-author, they will get the work deletion notice and a copy of the work.

> If the coauthor deletes their account and all their works, will their sole chapters get wiped out leaving the coauthor with a partial work? I ask this because whatever can happen when you have an owner of a work but not of multiple chapters will happen much more often in this scenario, so before we do something that will increase the frequency, I think we should make sure we're actually happy with the behavior.

If a co-author deletes their account while removing their name entirely, all of their chapters stay behind and can be edited by the remaining authors.

If a co-author deletes their account while orphaning, all of their chapters stay behind, but any chapter that _only_ they wrote is subject to the AO3-3176 bug if the remaining author(s) try editing.

features/step_definitions/work_steps.rb
```diff
@@ -216,19 +216,30 @@
   Work.tire.index.refresh
 end
 
+When /^a chapter with the co-author "([^\"]*)" is added to "([^\"]*)"$/ do |coauthor, work_title|
```
ariana-paris: As noted elsewhere, the `([^\"]*)` could be `([^"]*)` or `(.*)` but not a big deal as we have this all over the place

sarken:
Argle bargle. Let me know when we're ready and I'll update this then.

sarken:
I've merged this with master, which resulted in discovering the original issue was now fixed, yay! Except it was fixed in a weird way: strong parameters _broke_ the code that caused the original issue... fortunately, that code doesn't appear to be necessary, so this pull request removes that code in addition to implementing the interface changes planned for AO3-3752 and adding a whole heap of tests.

We haven't marked AO3-3752 closed yet, but this is the new issue for this pull request: https://otwarchive.atlassian.net/browse/AO3-4999

bingeling:
Is this a subset of AO3-3176? Anyway, it looks good. :)

sarken:
Yeah, it's actually the parent of AO3-3176 -- once I fixed this, AO3-3176 became necessary, but we wanted separate branches.

PR #2516, opened 2016-08-23
AO3-1259 Option to remove self as chapter co-creator for edit page

sarken:
https://otwarchive.atlassian.net/browse/AO3-1259

You could already remove yourself as a chapter co-creator, but it was hidden on the Manage Chapters page. Now it will be on the Edit page for an individual chapter. It will only show up if you are a co-creator of that particular chapter. It also now says "Chapter Co-Creator" instead of just "Author" because clarity is our friend.

I had to merge in the changes from #2515 so I could use the test steps I wrote for that... and so I could avoid merge conflicts with the tests.


sarken:
Please excuse me while I hunt down a fix for the caching bug the tests found. :|

sarken:
Closing whilst I have feels at this caching issue.

PR #2518, opened 2016-08-24
AO3-1259 Add "Remove Me As Chapter Co-Creator" to chapter edit page

sarken:
https://otwarchive.atlassian.net/browse/AO3-1259

You could already remove yourself as a chapter co-creator, but it was hidden on the Manage Chapters page. Now it will be on the Edit page for an individual chapter. It will only show up if you are a co-creator of that particular chapter. It also now says "Chapter Co-Creator" instead of just "Author" because clarity is our friend.

I had to merge in the changes from #2515 so I could use the test steps I wrote for that... and so I could avoid merge conflicts with the tests. The relevant changes are in 63e2546 (tests) and be759b7 (actual code).


app/controllers/chapters_controller.rb
```diff
@@ -91,10 +91,10 @@ def new
   # GET /work/:work_id/chapters/1/edit
   def edit
     if params["remove"] == "me"
-      @chapter.pseuds = @chapter.pseuds - current_user.pseuds
+      @chapter.authors_to_remove = current_user.pseuds
```
sarken: When you have a work with multiple creators and a chapter that _isn't_ by all the people on the work, under the chapter title will be a second byline saying "Chapter by creator." This part of the chapter is cached, and the cache wasn't expiring when a creator removed themselves from a chapter. This was happening because nothing _in_ the chapter itself was changing -- the creators for a chapter aren't in Chapter, but instead in Creatorship -- so the cache key, which is based on the chapter's updated_at date, wasn't expiring.

I tried adding a callback in the Chapter model to touch the chapter when pseuds were removed, but it was a damn lot of code. I tried touching the chapter using the before_destroy callback in the Creatorship model (here: https://github.com/otwcode/otwarchive/blob/master/app/models/creatorship.rb#L5), but surprise! That doesn't actually get called.

And that is how I ended up with this method -- setting authors_to_remove and then touching the chapter in the creation_observer.

sarken: I put it in the model now because (a) I realized that would work now and (b) Elz mentioned observers have recently gone bye-bye in Rails, so having fewer things in the observer seemed like something future!us would appreciate.

zz9pzza:
This has gone grey :(

sarken:
Something broke, but idk what yet

sarken:
The failure I was getting locally also happens on master, so we'll just wait and see what this says: https://travis-ci.org/sarken/otwarchive/builds/229900757

PR #2520, opened 2016-08-26
Attempt to set up test runs for code climate

elzj:
Not totally sure this will work for Cucumber, but giving it a shot. CODECLIMATE_REPO_TOKEN environment variable needs to be set separately.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,5 +1,8 @@
 ENV["RAILS_ENV"] ||= 'test'
 
+require "codeclimate-test-reporter"
+CodeClimate::TestReporter.start
+
 require File.expand_path("../../config/environment", __FILE__)
 #require File.expand_path('../../features/support/factories.rb', __FILE__)
 require 'rspec/rails'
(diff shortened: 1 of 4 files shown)


spec/spec_helper.rb
```diff
@@ -1,5 +1,8 @@
 ENV["RAILS_ENV"] ||= 'test'
 
+require "codeclimate-test-reporter"
+CodeClimate::TestReporter.start
+
```
zz9pzza: Under parallel tests we have https://github.com/codeclimate/ruby-test-reporter#installation

require 'simplecov'
require 'codeclimate-test-reporter'
SimpleCov.add_filter 'vendor'
SimpleCov.formatters = []
SimpleCov.start CodeClimate::TestReporter.configuration.profile

Then after all your tests run, in a rake task or as a build step do:

require 'simplecov'
require 'codeclimate-test-reporter'
CodeClimate::TestReporter::Formatter.new.format(SimpleCov.result)

zz9pzza:
We have this working now

PR #2523, opened 2016-08-29
AO3-4627 Add  viddertube.com to whitelist

zz9pzza:
https://otwarchive.atlassian.net/browse/AO3-4627


atpcm:
Per Garfield: "No, we do not need to allow flashvars. Our embed code is iframe based just like youtube's updated format."

cynsa:
Per Garfield: "No, we do not need to allow flashvars. Our embed code is iframe based just like youtube's updated format."

PR #2525, opened 2016-08-29
AO3-4628 Incompatible character encodings: ASCII-8BIT and UTF-8

ariana-paris:
https://otwarchive.atlassian.net/browse/AO3-4628

This fixes a problem which occurs when importing a work (both manually or automatically) whose metadata contains non-ASCII characters. When these tags reach the Sanitise gem without specifying an encoding, it bombs out because it expects pure ASCII or UTF-8, but gets something it thinks is ASCII but actually contains Unicode characters.

I also add tests which reproduce the problem if you remove the UTF-8 encoding on the string passed to Sanitise.

This is a blocker for the ongoing Ink Stained Fingers import that is just waiting for the works with this problem to be imported.

### DIFF
--- a/app/models/story_parser.rb
+++ b/app/models/story_parser.rb
@@ -440,7 +440,7 @@ class StoryParser
         story = eval("download_from_#{source.downcase}(location)")
       end
 
-      return story.force_encoding("UTF-8")
+      return story
     end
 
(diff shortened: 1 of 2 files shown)


zz9pzza:
Looks sane to me, waiting on travis.

PR #2526, opened 2016-08-30
AO3-3213 Fix tag unescaping for things like "a.k.a."

tickinginstant:
https://otwarchive.atlassian.net/browse/AO3-3213

Given a tag `A a.k.a. B`, the system for generating URLs turns that into `A a*d*k*d*a*d* B`. When that kind of tag name is passed to Tag.find_by_name, it first replaces all substrings `*s*` with `/`, then `*a*` with `&`, then `*d*` with `.`, etc. So when processing `A a*d*k*d*a*d* B`, it would first replace the `*a*` sequence in the middle to make `A a*d*k*d&d* B`, and then would replace the `*d*` sequence in the middle to make `A a.k*d&d* B`. So it would be incapable of looking up the tag that the user wanted.

This changes the substitution code to use a single substitution with a regex, so that it will go through the tag name from left to right, replacing each escape sequence with the appropriate character as it goes.

I also added a couple of cucumber tests.

### DIFF
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -503,7 +503,14 @@ class Tag < ActiveRecord::Base
   # Substitute characters that are particularly prone to cause trouble in urls
   def self.find_by_name(string)
     return unless string.is_a? String
-    string = string.gsub('*s*', '/').gsub('*a*', '&').gsub('*d*', '.').gsub('*q*', '?').gsub('*h*', '#')
+    string = string.gsub(
+      /\*[sadqh]\*/,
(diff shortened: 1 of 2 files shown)


sarken:
Oh, yay, this is an excellent fix to have -- thank you!

tickinginstant:
Glad to help! And thank you for reviewing it. :)

sarken:
(Adding the high priority label because it'll eliminate some work for AD&T and Tag Wrangling, so we should merge this ASAP.)

PR #2528, opened 2016-09-04
AO3-4633 An ugly way to ensure that the machine is quiet during the upgrade

zz9pzza:
https://otwarchive.atlassian.net/browse/AO3-4633

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -32,19 +32,19 @@ require 'bundler/capistrano'
 
 # deploy to different environments with tags
 require 'capistrano/ext/multistage'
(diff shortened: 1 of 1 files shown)


sarken:
How soon do we need this, and can you address Hound's comments?

zz9pzza:
It would make deployments cleaner, I havn't got the spare cycles to do the hound bits so it can wait a cycle or two.

zz9pzza:
@ariana-paris its a fair comment, I think maybe iterating through the web machines might be better but this should work and is simple.

PR #2529, opened 2016-09-06
AO3-3686 Keep prompts when deleting a pseud.

tickinginstant:
https://otwarchive.atlassian.net/browse/AO3-3686

When deleting a pseud, the pseud's challenge signups and challenge assignments are transferred to the user's default pseud, but the existing code doesn't update the `pseud_id` associated with any of the pseud's prompts. The pseud has a `:dependent => :destroy` relationship with any prompts that belong to the pseud, so when `pseud.destroy` is called, the prompts (but not the signups associated with those prompts) are deleted. This results in blank, invalid signups.

This pull request modifies the pseud deletion code to update each prompt's `pseud_id`, and keep the prompts from being destroyed. I also wrote some cucumber tests to check whether the prompts stuck around after the pseud was deleted.

(As a side note: Prompts aren't indexed by `pseud_id`. But prompts always have the same `pseud_id` as the signup they're associated with, thanks to a `before_save` hook, and challenge signups _are_ indexed by `pseud_id`. So I used a join with the `challenge_signups` table to try to make the update to `prompts.pseud_id` more efficient.)

### DIFF
--- a/app/models/pseud.rb
+++ b/app/models/pseud.rb
@@ -372,6 +372,16 @@ class Pseud < ActiveRecord::Base
   end
 
   def change_challenge_participation
+    # We want to update all prompts associated with this pseud, but although
+    # each prompt contains a pseud_id column, they're not indexed on it. That
+    # means doing the search Prompt.where(pseud_id: self.id) would require
+    # searching all rows of the prompts table. So instead, we do a join on the
+    # challenge_signups table and look up prompts whose ChallengeSignup has the
+    # pseud_id that we want to change.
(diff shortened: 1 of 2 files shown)


sarken:
@shalott: when you have a few minutes, could you give this a look? It looks basically good to me, but as our resident challenge expert, your expertise would be much appreciated!

shalott:
Looks good!

tickinginstant:
Thanks for reviewing it! I'm glad it looks okay. :)

PR #2531, opened 2016-09-13
AO3-4635 Attempt at some matching speed improvements.

tickinginstant:
https://otwarchive.atlassian.net/browse/AO3-4635

This pull request is a draft of some code intended to speed up the potential match generation system through various methods. My explanation of the changes and the effects got a bit long-winded, sorry!

# Speeding up Prompt.match

TagSet.with_type always triggers a database query, and it can be called up to five times per tag type (three times for one tag set, twice for the other) in a single call to Prompt.match. So I created a new function, TagSet.tag_ids_by_type, which creates a hash mapping from tag types to lists of tag IDs, and then caches the result to avoid duplicate computation. This makes it much faster to compare tag-sets. And since matching can only be run when signups are closed, there shouldn't be any danger of the tags changing out from under the matching code, so that kind of computation caching should be fine.

I also tried to make Prompt.match slightly more efficient by having it cache the full tag set (base tag set plus optional tag set) that it creates during Prompt.match. In addition, since it's inefficient for Prompt.match to load individual tagsets one-by-one (even if it's not doing it as frequently), I added some scopes with includes() statements to the ChallengeSignup class, to make it easier to ensure that all of the relevant data is loaded along with each challenge signup.

# Processing in batches, and indexing prompts

In the class PotentialMatcherConstrained, the new code processes all matches in batches. It divides all signups into batches of 100, and then for each pair of batches, it performs two computations: one to generate all signup matches when the signups in the first batch are acting as requests, and one to generate all signup matches when the signups in the second batch are acting as requests. Each batch is represented by a class PromptBatch, which builds an index in Ruby mapping from tag ids to lists of prompts, so that it's easier to look up "candidate" matches for a prompt -- prompts that are likely to produce prompt matches.

For each signup in the request batch, we iterate through its prompt requests, find candidates in the offer batch, and then try to build PotentialPromptMatch objects by calling Prompt.match. Once all of the prompt matches have been generated between that request signup and the offer batch, we group them by offer signup ID, and then try to build and save PotentialMatch objects for that request.

This has two advantages, at least in theory: it should reduce the number of SQL queries in most cases (since SQL is only used for loading batches and saving PotentialMatches), and it should reduce the number of calls to Prompt.match in many cases.

# Miscellaneous Notes

**Progress:** The way that I did the batching meant that each request was processed more than once (in fact, once for each offer batch), so it wasn't possible to keep the same username-based progress system. Instead, I switched over to showing only percents (but with two decimals of precision, so that for long-running matching tasks, the user has some indication that it's still going).

**Index Type:** PotentialMatcherConstrained only uses tags of a single tag type to look up "candidate" matches. Originally, I used the first required type as the "index type," but this isn't always ideal. For example, in a single-fandom ship-based exchange where the organizer requires a match on fandom (like my second experiment), *everyone* is going to have the same fandom marked in all of their prompts, so looking up prompts by that tag isn't particularly useful.

So I added a pre-processing step that tries to check, for each of the required tag types, whether it would make a "good" index type. If there are any good index types, it uses the first; otherwise, it falls back to using the first required type. This adds a little extra time to the matching, but in my experiments the slowdown is only 1-2%, and it should provide a significant improvement in, e.g. character-based single fandom exchanges where the mod lists "fandoms" as one of the required tag types.

**Batch Size:** I picked a batch-size of 100 somewhat arbitrarily -- I wanted it to be large enough to provide an improvement, but not too large (since each batch would also have to load all related prompts, tagsets, and tags). I think 100 is probably a good size for the tests that I was doing, where most signups had under 100 tags, but it might be necessary to go smaller for an exchange with really high limits. Maybe it would be good to adjust the batch sizes based on the challenge's prompt requirements?

**Limits on Matching:** In order to introduce batching the way that I did, I had to forego any limits on the matches, because batching affects the order of processing (and means that if you stop after reaching a minimum number of matches, you're not going to get an even distribution of matches).

In the tests that I ran with "sparse" matching graphs (relatively small ratio of PotentialMatches to ChallengeSignups), this didn't seem to be much of a problem. And when there are no required types, I have some special-case code in PotentialMatcherUnconstrained that samples random offer IDs to pair with each request (so that it doesn't generate the complete graph). But that's not the only way to have basically unconstrained matching. I have some ideas about maybe trying to add a fall-back to the original matching code if there are no "good" index types, but I'm not sure that would catch all of the tricky cases.

# Testing

I added some cucumber tests for generating signups for small exchanges, and then checking whether the correct PotentialMatches were generated. (In the process of writing them, I found and corrected a small bug in the original potential matching code, which could occasionally cause two copies of the same PotentialMatch to be generated.)

In order to tell whether I was making progress speed-wise, I also wrote some code to generate fake signups for a couple of test exchanges, to evaluate runtimes.

* Exchange A is a multi-fandom exchange that requires matching fandoms and 1 matching ship, and has ~150 participants.

* Exchange B is a single-fandom exchange that requires matching fandoms and 1 matching ship, and has ~75 participants. It also has "fandom" as one of the required tag types, has more prompts per signup, and has nearly twice as many potential matches as Exchange A (despite only having half as many participants). I believe that's why matching takes longer than the larger exchange.

* Exchange C is a multi-fandom exchange that requires matching fandoms and all matching characters, and has ~200 participants. It's also very sparse -- there's an average of about 1.5 potential matches per participant, and a lot of participants have no potential matches.

|                            | Exchange A  | Exchange B  | Exchange C  |
|----------------------------|-------------|-------------|-------------|
| original code              | 102 minutes | 240 minutes | 8.7 minutes |
| only Prompt.match changes  | 5.7 minutes | 8.5 minutes | 1.3 minutes |
| all changes                | 1.5 minutes | 2.0 minutes | 0.9 minutes |

To get the times, I used ruby's Benchmark.measure on a call to PotentialMatch.generate_in_background. (So all of the times include checking for invalid signups, potential match generation, and assignment generation.) I ran all of the experiments on a vagrant box on my laptop. Obviously, this doesn't necessarily indicate anything about how this code might perform on real data in production, and there's a fair amount of variation in how much of an improvement it might be, but it seems like a promising direction.


features/step_definitions/potential_match_steps.rb
```diff
@@ -0,0 +1,158 @@
...
+end
+
+When /^potential matches are generated for "([^\"]*)"$/ do |name|
+  collection = Collection.find_by_name(name)
+  PotentialMatch.generate_in_background collection.id
+end
+
+Then /^there should be no potential matches for "([^\"]*)"$/ do |name|
+  collection = Collection.find_by_name(name)
+  collection.potential_matches.count.should == 0
```
tickinginstant: I tried implementing this suggestion, but on my machine it made the test error out (saying there was no method `zero?` for the RSpec object). Should I change it, or is this Hound being overzealous about `== 0`?

shalott: Yeah, Hound is frequently overzealous, don't worry about it if you have any reason for ignoring a Hound comment, we use them more as reminders than a strict rule

shalott:
I will do a more detailed review line by line shortly, but this is fabulous, thank you! The speedup is sufficiently enormous that I don't think we need to worry about the lack of constraint, but if you had the chance to add one more benchmark test for a real-world problem case, we ran into major issues with a single-fandom challenge with the following configuration: 

4 requests and 4 offers required, but up to 6 and 10 allowed
Characters: 1-10 on requests, 1-10 for offers, any allowed 
Ships: 0-5 on requests, 0-10 on offers, any allowed, must be unique
Ratings:2-3 for requests, 2-5 for offers, any allowed
Additional tags: 1-2 for requests, 1-2 for offers
Matching requirements: 1 request, All characters, 1 freeform, 1 rating

There were only 100 participants but many people included huge numbers of characters in offers and roughly 3/4 of the participants matched each other and matched over and over and over. /o\ (Part of the issue was a background job auto-restarter run amok but the challenge only hit it because of the settings.) We've been considering lowering the limits as a quick way to avoid this in future, but I suspect with the speedup you've got here, we won't need to, so I'd love to confirm.

tickinginstant:
I've started generating some tests that fit that description, but I think I was reading your explanation wrong, so they're not showing the results that I'd expect (i.e. taking hours and hours to run the existing matching code). So if you don't mind answering, I have a few questions about the problematic case, so that I can try to write a test case that matches it better:

1. You mentioned ships in the signup requirements, but not the matching requirements. (Which is particularly weird since ships are the only tag required to be unique.) Did the exchange require matching on ships?

2. It sounds to me like you're saying that the set of PotentialMatches is dense (i.e. many PotentialMatches per person). I always thought that was one of the (relatively) good cases for the existing potential match generation code, since it means that the loop inside of generate_for_signup doesn't have to run very many times. I figured that the worst case for the current potential matching code would be having sparse matches (only a few PotentialMatches per person), but with one tag that almost everyone is using -- e.g. a "Fic" or "Art" freeform, or a fandom tag for a single-fandom exchange. That way, matching_signup_ids returns massive numbers of signups, and generate_for_signup has to go through almost all of them to find all of the matches.

   It's occurred to me that when you said "roughly 3/4 of the participants matched each other," you might mean matching_signup_ids, not PotentialMatches. Just to double-check -- do you mean matching_signup_ids?

3. When you say that the participants matched over and over and over, do you mean an average of ~5 prompt matches per PotentialMatch (so that every request matches with one offer), ~10 prompt matches per PotentialMatch (so that some requests match twice over), or ~30-60 prompt matches per potential match (so that every request prompt matches every single offer prompt)? I was aiming for 5-10, but I'm wondering whether I should have aimed higher.

shalott:
1. Having double-checked yes, and yes, it's weird. My guess is that the mods changed the matching requirements a few times during their issues trying to get things to work. There was also another weirdness I missed: there were actually 20 characters allowed in requests but only 10 in offers, so you could potentially request 11 characters and make one of your requests totally unmatchable. However, I don't think people had actually done that for the most part. 

2. To clarify, this challenge encountered its issues _before_ the constraints we recently put in. (Even the earlier straight cap of 50 was not actually working at all as it turned out the constraint was only in the config file, which is just used in the dev environment, and not loaded on the production site config engine.) So because most participants had at least some overlapping characters, matching_signup_ids did return huge numbers of signups, and then the loop ran exhaustively over all of them, generating truly spectacular numbers of PotentialPromptMatches and PotentialMatches. 

3. Because of the job restarter in the mix, it's hard to say what of the match volume was actually due to the real settings. It was generating enough db writes that the matching process was losing its db connection and dying along the way. Then the job restarter would restart it, to make matters worse the cleanup process was running the on-destroy callbacks on the prompt matches and took forever, so it was deleting matches while new ones were being generated--the job would die again and get restarted again, multiple times -- it was very special. So the real damage was mostly done by the restarting, but that only hit in the first place because the challenge was generating a LOT of prompt matches, so I would aim at that 30-60 range.

tickinginstant:
Thanks for the info! :) I can try to aim for 30-60 prompt matches per signup, but unfortunately, my code is probably not going to handle it well. Since there's at least one required tag type, it will generate all matches (and all PotentialPromptMatches for those matches). With 100 participants, 75 offers matching each recipient, and 30-60 prompt matches per pair, it's going to have to create between 200,000 and 500,000 PotentialPromptMatches. (Versus the 20,000 to 50,000 prompt matches that would be generated by the existing matching code, thanks to the restrictions.) If lots and lots of database writes are a problem -- and it sounds like they really are -- I don't see how any code that generates the full set of matches and prompt matches can avoid it.

Also, the time is likely to be a bit of a problem, too. I recently finished running a test with ~130,000 prompt matches for ~10,000 signup matches, and the whole thing took 2 hours and 17 minutes. It took ~30 minutes for the potential match generation code alone -- not ideal, but not too outrageous. Unfortunately, the ChallengeAssignment code took another ~110 minutes after that, because it had to process all 130K prompt matches. And since `PotentialMatch.<=>` has worst-case runtime quadratic in the number of prompts matches per signup match, doubling or tripling the number of prompt matches could cause the ChallengeAssignment process to take 4-9x as long, which is very far from ideal.

IMHO, it really does sound like it's a very good idea to limit the number of matches per person, which means that this pull request is probably not a good idea. On the plus side, my improvements to just Prompt.match will work just fine with the limits set by the existing code. It's not as fast as this pull request in many sparser cases, but in every single test I've done it's at least 5x faster than the existing matching code, and is frequently more like 10-20x faster. Would you like me to post a pull request for that?

Alternatively, if your ideal is to find all signup matches, and you're willing to compromise a little on the outcomes of the comparisons between different PotentialMatches, maybe it's not necessary to store all of the *prompt* matches. Would it be acceptable to, say, generate all of the prompt matches for a signup match, but only save the best match for each request? That would mean that in the terrible case, there'd only be around 6 prompt matches per PotentialMatch, for a total of ~45,000 prompt matches.

Or maybe it would be possible to limit the total number of PotentialPromptMatches per exchange, and then just stop saving new PotentialPromptMatches once the exchange hits its limit? You'd still compute all of the matching signups correctly, and you'd still be able to use `PotentialMatch.num_prompts_matched` to compare PotentialMatches -- you just wouldn't be able to break ties with the information from the prompt matches.

shalott:
Hm, rather than saving just the best one, what about not saving any of them and instead storing a numeric "quality" value on the potential match objects in a form x.y where x is the number of prompt matches and y is the number of tag matches in the best prompt match and using that for <=> purposes? (IME people are generally much more excited to have multiple requests to choose from rather than a single more-ideal request, so the # of prompt matches is a stronger measure of quality than the # of tag matches in a single prompt match.)

(BTW, if you don't mind adding your benchmarking/signup generating tool to the pullreq, perhaps under scripts/, that would be great!)

tickinginstant:
I think I was unclear, sorry. For my first suggestion, you could still keep track of the total number of prompts matches in the existing `PotentialMatch.num_prompts_matched` field, and compare based on that. But you'd only be saving a small number of PotentialPromptMatches per PotentialMatch (maybe just the best one, or maybe the best one associated with each request prompt), so that you wouldn't have quite as much information when breaking ties for num_prompts_matched.

I prefer your suggestion over my own -- it's very logical, it's easy-to-understand, and it's presumably a little easier on the database since it doesn't involve quite as many tables. (It does sound like it involves migrations, which I don't have much experience with, but I think I can figure it out.) I'm just worried that I explained myself poorly, so I figured I'd try to be a little clearer. :)

If I modify the code to reduce the number of PotentialPromptMatches saved, there's also a chance that I may be able to speed up the potential match generation code a little more for cases like the one you're worried about. A couple weeks back, I was experimenting with some more extreme changes -- ActiveRecords are (understandably) very slow compared to other Ruby classes, so I wrote some code to (1) load all of the data into non-Rails objects, (2) perform computations, and then (3) repackage the results as Rails objects so that they could be saved. It didn't seem to provide a substantial speed-up over the code in this pull request, so I figured that the slowness of the ActiveRecords wasn't significant, and I should focus on the code that worked in a more Rails-like way.

Out of curiosity, I recently brushed off the code from that branch and tried it out on the monster 130K example, and found that steps (1) & (2) took under a minute, and step (3) took the remaining 20-30 minutes. If saving is proportional to the number of objects created -- and it seems like it ought to be, unless I'm missing something -- then saving a fraction of the objects should mean that it runs in a fraction of the time.

The downside is that disabling saving in this pull request doesn't seem to result in any real speed-up. I'm not 100% sure why, but I suspect that some of it may be a result of the overhead associated with ActiveRecords. So I may end up wanting to submit another pull request, if that's all right.

As for my tools: A lot of the scripts are in rough shape, and sometimes my benchmarking involves diving into the code and adding random Benchmark.measure calls in interesting places. But I can try to clean some of it up to make it usable. (Some of my tests are based around parsing pages of Request Summaries that I downloaded from exchanges on AO3, and then making each participant's offers the same as their requests. I hope that's okay -- I wanted some real-looking data to play around with, but I've never been an exchange mod.)

shalott:
Oh, I'd forgotten the num_prompts_matched field already being in there. Yeah, so just adding a max_tags_matched field or similar I think would make a fine tiebreaker without saving any prompt matches. (Also, just to save you looking it up if you haven't already, `rails generate migration AddMaxTagsMatchedToPotentialMatches` max_tags_matched:int` will generate the up/down migrations for you, but it's straightforward to add a new field, it's migrating existing populated fields that is an issue.) 

By all means put in any pullreqs you'd like and I will review and compare if it's helpful. But honestly, this is already a huge improvement and I'd go for this code over the current, even with the issue for not-sparse challenges without the cap on potential matches. For one thing, even with zero constraints until just recently, and without all the db reads you've eliminated here, challenges have been completing fine albeit very slowly. People tolerate the process taking a while because in practice even after the automated matching is done the mods often have to or want to do massaging on human time frames.

And if we let people know that sparser challenges run this much faster, mods will almost certainly MAKE the challenges sparser -- ie, as a mod in an ideal world you would like your matching requirements as high as possible because that makes better matches, but when matching takes a very long time, you don't want to take the risk of winding up with half your participants unmatched and having to re-run matching. So currently, mods tend to be conservative about requirements. But if matching gets this much faster, we can encourage mods to start with much stricter requirements, and if the matching doesn't work out, there will be a low cost to reducing the requirements and trying again.

tickinginstant:
I'm most of the way through the update to only keep the max_tags_matched, and while I was working on it, I realized that I didn't actually really need PotentialPromptMatches anymore -- they're still useful as a concept, but there's no longer any need to store all of the information about, e.g. num_fandoms_matched for a pair of prompts. All the code needs to know is whether there's a match between the two prompts, and the total number of overlapping tags (which is a single value).

Thanks for the code for the migration for adding a field! That part does seem straightforward. But as far as migrations, wouldn't it also be necessary to compute max_tags_matched for existing PotentialMatch objects? (Although I guess that would be an after task, not a migration?) And since PotentialPromptMatches aren't being stored anymore, wouldn't that table eventually end up getting deleted, as well?

If you think that it would be better to still store all of the PotentialPromptMatches, I'm willing to stick with the current version of the code -- you have a much better idea about the production requirements than I do. :) But I don't think it would be very difficult to complete the changes, either. Would you like me to continue with them? And if so, would you rather I put them in a different pull request, or push the commits to this pull request?

shalott:
Yes, having poked around, I agree we should get rid of the class. And I would not bother with migrating the existing data, either, because all potential matches and potential prompt matches are cleaned up anyway as soon as assignments are sent out -- they are intended as ephemeral data, so there's no need to preserve them for the future. Any mod who needs/wants to re-run matching will clean up the old PMs and generate the max_tags_matched values in the re-run anyway. 

There might be challenges that are in an in-between state (pms generated, assignments not yet sent), but I think in any case we'll want to notify all active-challenge mods with a heads-up email before the deploy with some warning and ask them to either finish sending their assignments before we deploy or re-run matching afterwards. I will need to discuss with Systems but now I think about it, during migration we might want to wipe out the entire potential_matches table and reset the ids back to 0 in order to clean up the gargantuan numbers that were racked up by the job-restarter. 

I would say push commits here, let's keep it together since we'd want to deploy it together anyway, and that way I can review all the changes together.

tickinginstant:
All right, I made the changes to remove PotentialPromptMatches. I tried to make the migration reversible, but I ran into some difficulties testing the rollback because there was something kind of odd going on with the existing migrations. (One of them seems to have one too many digits in the date, which seems to make Rails a little confused about how to roll back the most recent migration.)

As a separate commit, I added in a directory of scripts for testing, which is why the line count of the pull request has pretty much doubled.

Also, sorry for all of the massive numbers of commits and Hound comments. I figured that I'd try to add all of the frozen_string_literal comments suggested by Hound -- even if they're literally useless at the moment, I'm assuming they'll be useful if there's ever an upgrade to Ruby 2.3. And then I think I was pushing commits before Hound finished posting comments, and things sort of spiraled from there. :(

shalott:
No worries about the Hound, am used to the giant inundation of comments, thankfully gmail collapses it all. ;)

app/models/potential_match_builder.rb
```diff
@@ -0,0 +1,61 @@
...
+
+    # If we've already matched ALL, we don't need to check the number of
+    # overlapping tags, since we've already got the maximum possible.
+    return if @max_tags_matched == ALL
+
+    # Compute the number of matching tags, and update max_tags_matched if
+    # necessary.
+    curr_tags_matched = request_prompt.count_tags_matched(offer_prompt)
+    @max_tags_matched = [@max_tags_matched, curr_tags_matched].max
+    @max_tags_matched = ALL if curr_tags_matched == ALL
```
shalott: Occurs to me that using 'ALL' for the purposes of @max_tags_matched privileges a match between short requests/offers. That is, A has R1 with one tag and R2 with seven tags. B has O1 with one tag matching R1. C has O2 with six tags matching R2. Match 1 between A and B will be considered better than match 2 between A and C, even though that one is actually much more specific. Since we only use this if the two prompts are a valid match by requirements, I would be inclined to try and save the absolute number here instead.

app/models/potential_matcher/prompt_tag_type_info.rb
```diff
@@ -0,0 +1,95 @@
+# frozen_string_literal: true
+
+# A class used for gathering information about the rough frequency of different
+# required tag types in all of the prompts in a collection. Used to figure out
+# what would be a good "index type" -- that is, a type that we can use to try
+# to look up matches with.
+class PromptTagTypeInfo
```
shalott: QA note: run this against all real-world challenges in the db on staging in console to see what index types result just as sanity check, looks good to me

app/models/potential_matcher/prompt_tag_type_info.rb
```diff
@@ -0,0 +1,95 @@
...
+  end
+
+  # A good type to use for indexing has a fair number of tags of that type (so
+  # that we aren't, e.g. trying to match prompts on fandom in a single-fandom
+  # exchange), and doesn't have very many people requesting "any" (so that we
+  # have a lot of constraints on who can match with whom).
+  def calculate_good_index_types
+    @good_index_types = @required_types.select do |type|
+      (@first_tags_of_type[type].size >= @limit_first_tags) &&
+        (@count_prompts_with_any[type] <= @limit_prompts_with_any)
```
shalott: Not sure but talking out a thought here: it might work better to do this check purely on requests rather than on all prompts. Typically lots more people offer "any" rather than requesting it, especially with anything below fandom (character, ship, rating, etc), because there are many people who use as their only restriction on offering what they have the _knowledge_ to do, but their requests are much more specific. 

So eg, for a single-fandom (where the mod required 'fandom' anyway) or a few-fandom challenge (eg, an MCU challenge with say six Marvel movies as fandoms), I think looking at characters it's very likely to exceed the limit on 'any', and thus wind up creating the candidate batches on fandom as the top required tag, even though you are still better off in that scenario creating the candidate batches based on character rather than on fandom. 

That said, that challenge would probably have pretty dense matching no matter what unless there was some other constraining tag type below character. I'm fine leaving this alone and seeing what results we get from real-world data.

app/models/potential_matcher/prompt_tag_type_info.rb
```diff
@@ -0,0 +1,95 @@
...
+
+  # Load information about the passed-in collection.
+  def load_collection_info
+    @total_prompts = @collection.prompts.count
+    @total_signups = @collection.signups.count
+    @settings = @collection.challenge.potential_match_settings
+    @required_types = @settings.required_types
+
+    @limit_first_tags = [10, @total_signups / 10].min
+    @limit_prompts_with_any = [10, @total_prompts / 10].max
```
shalott: A note here, move the numbers here into separate constants in config.yml, which you can then access here with "ArchiveConfig.[constant name]" eg ArchiveConfig.FIRST_TAGS_LIMIT etc. -- this will go into cfengine and can then be tweaked live if we find a better number

shalott:
OK, I confess I haven't line-by-lined all the scripts for generating the random exchanges, but will sit down and run them on my dev setup in the next few days and look them over some more, and everything else looks great to me. 

FYI, having talked to the team we would love to push this to staging soon: we have three deploys scheduled, and then I think this one will go next, early in October. Would you be available around then for consult if we turn up any issues?

tickinginstant:
I'm glad the code looks good so far! Hopefully the tests will also go well. :) I'll try to make the suggested changes soon.

As for early October, I should be available.

app/models/potential_matcher/prompt_tag_type_info.rb
tickinginstant: Hmm, interesting. I think part of the problem is that I approached "goodness" as a binary, when really it's a spectrum. If that MCU challenge had a lot of different freeforms, and no one requested Any Freeform, then characters would be a bad tag type to index on, and we wouldn't want it to be included in the set of "good" tag types (since we'd want it to be indexed on freeforms instead). But if there was nothing below characters, then characters would be a "good" tag type. (Or at least, probably better than fandoms.)

Maybe it would be better to compute a goodness "score," and pick the tag type with the best score? Something like ... prompts_with_any + prompts.size * (estimated_tags_per_prompt)^2 / estimated_unique_tag_count, to get a rough estimate of the number of prompts that would be considered "candidates" when indexing on that tag type? (Ignoring the division between requests and offers for simplicity's sake.) Then pick the tag type with the lowest score. That would mean that the character tag type would pretty much always score better than the fandom type on a single-fandom exchange, even with oodles of people offering any.

As you say, though, it might be easier to tune the heuristic on real-world data.

PR #2532, opened 2016-09-15
Update application.rb

scottsds:
This is a test. I don't see a template though...


scottsds:
This is a comment on the issue.

PR #2534, opened 2016-09-20
AO3-4639 Update the list of valid options

zz9pzza:
https://otwarchive.atlassian.net/browse/AO3-4639


sarken:
These failures are all related to https://otwarchive.atlassian.net/browse/AO3-4640 so I'm gonna merge this

PR #2539, opened 2016-09-21
AO3-4643 Deleting a subscription will leave an error in resque

zz9pzza:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-4643
## Purpose

This checks that we have a subscription before trying to use it to send notifications

## Testing

subscribe to a work.
update the work.
delete the subscription
After the notification would have been delivered go in to resque interface and see if you can find an error like:
Worker
ao3-app08:862 on MAILER at about 2 hours ago Retry or Remove
Class
UserMailer
Arguments
"batch_subscription_notification"
"62205766"
"[\"Chapter_18569068\"]"
Exception
ActiveRecord::RecordNotFound
Error
Couldn't find Subscription with id=62205766

### DIFF
--- a/app/mailers/user_mailer.rb
+++ b/app/mailers/user_mailer.rb
@@ -90,6 +90,8 @@ class UserMailer < BulletproofMailer::Base
   
   # Sends a batched subscription notification
   def batch_subscription_notification(subscription_id, entries)
+    # Here we use find_by_id so that if the subscription is not found 
+    # then the resque job does not error and we just silently fail.
     @subscription = Subscription.find_by_id(subscription_id)
     return if @subscription.nil?
     creation_entries = JSON.parse(entries)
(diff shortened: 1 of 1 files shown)


app/mailers/user_mailer.rb
```diff
@@ -90,7 +90,8 @@ def claim_notification(creator_id, claimed_work_ids, is_user=false)
   
   # Sends a batched subscription notification
   def batch_subscription_notification(subscription_id, entries)
-    @subscription = Subscription.find(subscription_id)
+    @subscription = Subscription.find_by_id(subscription_id)
```
sarken: Could you add a comment above this that explains, roughly, `# Use find_by_id (not find) so it returns nil if there's no subscription by that id`?

PR #2550, opened 2016-09-25
AO3-4505 Line breaks for challenge assignment notifications.

tickinginstant:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-4505

## Purpose

When sending out gift exchange assignment notifications, challenge mods can include a custom (plain-text) message. In the text version of the email, linebreaks seem to be preserved. But in the HTML version of the email (which most users probably see, since a lot of email clients can display HTML emails), the newlines aren't converted into `<br>` tags, so the mod's linebreaks disappear. This pull request updates the handling of the assignment notification for HTML emails to add a `<br>` tag between lines.

I added a cucumber test to ensure that linebreaks were being preserved, and a second test to double-check that my changes didn't break the handling of HTML entities. (Tags are stripped when the CollectionsController sanitizes the params, so I used an ampersand to verify that the text of the email was being escaped.)

## Testing

The bug report lays out steps for testing whether linebreaks work.

## Credit

tickinginstant, she/her


app/helpers/mailer_helper.rb
```diff
@@ -72,5 +72,15 @@ def to_plain_text(html)
...
+
+  # Reformat a string as HTML with <br> tags instead of newlines, but with all
+  # other HTML escaped.
+  # This is used for collection.assignment_notification, which already strips
+  # HTML tags (when saving the collection settings, the params are sanitized),
+  # but that still leaves other HTML entities.
+  def escape_html_and_create_linebreaks(html)
+    # Escape each line with h(), then join with <br>s and mark as html_safe to
+    # ensure that the <br>s aren't escaped.
+    html.split("\n").map { |x| h(x) }.join('<br>').html_safe
```
sarken: I think this could benefit from a slightly more descriptive variable name than `x`, like `line` or `line_of_text` or something.

tickinginstant: Good point! I'll make the change.

features/gift_exchanges/notification_emails.feature
```diff
@@ -0,0 +1,60 @@
...
+    Then I should see "Sign-up was successfully created."
+
+    When I am logged in as "participant2"
+      And I start signing up for "Holiday Swap"
+      And I press "Submit"
+    Then I should see "Sign-up was successfully created."
+
+    When I close signups for "Holiday Swap"
+      And I have generated matches for "Holiday Swap"
+      And all emails have been delivered
```
sarken: Will this work with the `And I have sent assignments for "Holiday Test"` step instead of these four lines?

tickinginstant: It looks like it should work, thanks!

PR #2551, opened 2016-09-27
AO3-3380 Handle when a work becomes anonymous after queuing subscriptions, but before sending subscriptions.

tickinginstant:
# Pull Request Checklist

_(Is there anything I need to update the JIRA issue with?)_
## Issue

https://otwarchive.atlassian.net/browse/AO3-3380
## Purpose

Before queuing up a subscription notification, the CreationObserver checks whether the work in question is anonymous (to make sure that subscription notifications won't inadvertently reveal anonymous authors). However, those checks only work if you post directly to an anonymous collection, so that the work has always been anonymous. It doesn't cover cases where someone accidentally forgets to add the collection when they first post the work, and then immediately adds it to the correct collection before batch subscription emails are sent.

This pull request adds an extra check to the UserMailer, to ensure that the work didn't suddenly become anonymous between the time the subscription notification was queued, and the time the subscription email is being sent out.
## Testing

The bug report has steps that can be used for testing this change.
## Credit

tickinginstant, she/her

### DIFF
--- a/app/mailers/user_mailer.rb
+++ b/app/mailers/user_mailer.rb
@@ -101,6 +101,14 @@ class UserMailer < BulletproofMailer::Base
       next if (creation.is_a?(Chapter) && !creation.work.try(:posted))
       next if creation.pseuds.any? {|p| p.user == User.orphan_account} # no notifications for orphan works
       # TODO: allow subscriptions to orphan_account to receive notifications
+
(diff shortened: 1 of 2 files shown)


sarken:
Just wanted to let you know the JIRA issue is fine and doesn't need updating!

tickinginstant:
Great, thanks! :)

PR #2553, opened 2016-09-28
AO3-4651 Abuse reports should only accept 5 reports per work

sarken:
https://otwarchive.atlassian.net/browse/AO3-4651

## Purpose

The check to make sure a work hadn't been reported to Abuse more than 5 times was broken. This fixes it and adds a test.

It's currently hot fixed on production.

## Testing

Attempt to report a work to Abuse 6 times. The 6th time should fail.


zz9pzza:
Looks like the hotfix applied.

zz9pzza:
It would be nice to have a test showing that you could report the work one time less than the maximum and it would work.

zz9pzza:
Looks good waiting on travis.

PR #2554, opened 2016-09-28
AO3-4636 turn off tag detection

ariana-paris:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4636
## Purpose

Currently, the importer always detects tags in the content of the stories it imports and adds those tags to the imported work. This sometimes leads to rubbish tags (e.g.: "Warnings: fluffy bunnies") and in some cases where it detects text that isn't a valid tag, breaks the import. The `Override...` checkbox on the import form (and its API equivalent) only allows you to override tags if you provide some other content, but not to ignore detected tags completely.

This PR adds a `detect_tags` option in addition to the existing `override_tags`, to prevent all tags from being detected at all when some tags are overridden.
## Testing

See Jira for testing notes

### DIFF
--- a/app/controllers/api/v1/works_controller.rb
+++ b/app/controllers/api/v1/works_controller.rb
@@ -186,7 +186,7 @@ class Api::V1::WorksController < Api::V1::BaseController
       import_multiple: "chapters",
       importing_for_others: true,
       do_not_set_current_author: true,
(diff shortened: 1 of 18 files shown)


sarken:
I haven't done any reviewing, but I'm going to mark this needs coder action because these are unfortunately not our regular failures.

shalott:
This looks good to me, the tests are solid!

PR #2557, opened 2016-09-28
AO3-4652 AO3-4654 AO3-4656 Assignment email fixes

sarken:
https://otwarchive.atlassian.net/browse/AO3-4652
https://otwarchive.atlassian.net/browse/AO3-4654
https://otwarchive.atlassian.net/browse/AO3-4656

## Purpose

AO3-4652: When you receive an assignment email in a gift exchange, it says "Rating: Choose Not To Use Archive Warnings" if (a) your recipient did not select a rating or (b) the challenge does not use rating tags. This makes it so "Rating:" is only included if the recipient specifies.

AO3-4654: All of the fields the requester explicitly set to "Any" should be listed in the email, like they are on the sign-up you can access by going through your dashboard.

AO3-4656: The link to your Assignments page should say "Assignments page" rather than "archive dashboard."

## Testing

Refer to JIRA.

### DIFF
--- a/app/views/user_mailer/challenge_assignment_notification.html.erb
+++ b/app/views/user_mailer/challenge_assignment_notification.html.erb
@@ -21,9 +21,9 @@
     <% chars = prompt.any_character ? "Any" : styled_tag_list(tag_groups["Character"]) %>
     <% ships = prompt.any_relationship ? "Any" : styled_tag_list(tag_groups["Relationship"]) %>
     <% ratings = prompt.any_rating ? "Any" : (tag_groups["Rating"] ? get_title_string(tag_groups["Rating"]) : nil) %>
-    <% warnings = tag_groups["Warning"] ? get_title_string(tag_groups["Warning"]) : nil %>
-    <% categories = tag_groups["Category"] ? get_title_string(tag_groups["Category"]) : nil %>
-    <% atags = styled_tag_list(tag_groups["Freeform"]) %>
+    <% warnings = prompt.any_warning ? "Any" : (tag_groups["Warning"] ? get_title_string(tag_groups["Warning"]) : nil) %>
+    <% categories = prompt.any_category ? "Any" : (tag_groups["Category"] ? get_title_string(tag_groups["Category"]) : nil) %>
+    <% atags = prompt.any_freeform ? "Any" : styled_tag_list(tag_groups["Freeform"]) %>
     <% otags = prompt.optional_tag_set ? styled_tag_list(prompt.optional_tag_set.tags) : nil %>
 
     <%= styled_divider %>
--- a/app/views/user_mailer/challenge_assignment_notification.html.erb
+++ b/app/views/user_mailer/challenge_assignment_notification.html.erb
@@ -53,7 +53,7 @@
   </p>
 
   <p>
-    You can look up this assignment from <%= style_link("your archive dashboard", user_assignments_url(@assigned_user)) %>.
+    You can look up this assignment from <%= style_link("your Assignments page", user_assignments_url(@assigned_user)) %>.
   </p>
 
   <% if @collection && !@collection.assignment_notification.blank? %>
--- a/app/views/user_mailer/challenge_assignment_notification.html.erb
+++ b/app/views/user_mailer/challenge_assignment_notification.html.erb
@@ -14,38 +14,38 @@
 
     <% def styled_tag_list(tags) %>
       <% return nil if !tags || tags.empty? %>
-      <% tags.map{|tag| style_link(tag.name, tag_works_url(tag))}.to_sentence.html_safe %>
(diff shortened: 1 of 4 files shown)


app/views/user_mailer/challenge_assignment_notification.html.erb
```diff
@@ -14,38 +14,38 @@
...
       <% return nil if !tags || tags.empty? %>
-      <% tags.map{|tag| style_link(tag.name, tag_works_url(tag))}.to_sentence.html_safe %>
+      <% tags.map { |tag| style_link(tag.name, tag_works_url(tag)) }.to_sentence.html_safe %>
     <% end %>
 
     <% fandoms = prompt.any_fandom ? "Any" : styled_tag_list(tag_groups["Fandom"]) %>
     <% chars = prompt.any_character ? "Any" : styled_tag_list(tag_groups["Character"]) %>
     <% ships = prompt.any_relationship ? "Any" : styled_tag_list(tag_groups["Relationship"]) %>
-    <% ratings = prompt.any_rating ? "Any" : get_title_string(tag_groups["Rating"]) %>
+    <% ratings = prompt.any_rating ? "Any" : (tag_groups["Rating"] ? get_title_string(tag_groups["Rating"]) : nil) %>
```
sarken: The relevant change, in case you're looking at the full diff rather than a specific commit.

app/views/user_mailer/challenge_assignment_notification.text.erb
```diff
@@ -8,21 +8,21 @@ Prompts:
...
 <% def tag_list(tags) %>
 <%  return nil if !tags || tags.empty? %>
-<%  tags.map{|tag| tag.name}.to_sentence.html_safe %>
+<%  tags.map { |tag| tag.name }.to_sentence.html_safe %>
 <% end %>
 <% fandoms = prompt.any_fandom ? "Any" : tag_list(tag_groups["Fandom"]) %>
 <% chars = prompt.any_character ? "Any" : tag_list(tag_groups["Character"]) %>
 <% ships = prompt.any_relationship ? "Any" : tag_list(tag_groups["Relationship"]) %>
-<% ratings = prompt.any_rating ? "Any" : get_title_string(tag_groups["Rating"]) %>
+<% ratings = prompt.any_rating ? "Any" : (tag_groups["Rating"] ? get_title_string(tag_groups["Rating"]) : nil) %>
```
sarken: The relevant change, in case you're looking at the full diff rather than a specific commit.

app/views/user_mailer/challenge_assignment_notification.html.erb
ariana-paris: Should the same logic apply to `warnings` and `categories` below? (i.e. is there a case where a prompt might stipulate "Any" warning or category and the email should reflect this - or are they special cases?)

sarken: I was kind of wondering that myself, but not enough to look into it beyond the original bug that presented itself to me on a platter. >.> 

But, in general, this whole thing seems very inconsistent. Any type of tag can have an "Any?" option, and any type of tag can be omitted (unless the mod of a give challenge has decided to make that tag mandatory, of course). I have no idea why they don't all have the same logic I put in here.

sarken: So the results of that were this issue: https://otwarchive.atlassian.net/browse/AO3-4654

app/views/user_mailer/challenge_assignment_notification.html.erb
```diff
@@ -14,46 +14,46 @@
...
     <% end %>
 
     <% fandoms = prompt.any_fandom ? "Any" : styled_tag_list(tag_groups["Fandom"]) %>
     <% chars = prompt.any_character ? "Any" : styled_tag_list(tag_groups["Character"]) %>
     <% ships = prompt.any_relationship ? "Any" : styled_tag_list(tag_groups["Relationship"]) %>
-    <% ratings = prompt.any_rating ? "Any" : get_title_string(tag_groups["Rating"]) %>
-    <% warnings = tag_groups["Warning"] ? get_title_string(tag_groups["Warning"]) : nil %>
-    <% categories = tag_groups["Category"] ? get_title_string(tag_groups["Category"]) : nil %>
-    <% atags = styled_tag_list(tag_groups["Freeform"]) %>
+    <% ratings = prompt.any_rating ? "Any" : (tag_groups["Rating"] ? get_title_string(tag_groups["Rating"]) : nil) %>
```
zz9pzza: Having multiple ? makes this hard to understand.

ratings = nil 
if prompt.any_rating then 
  ratings= "Any"
else 
  if tag_groups["Rating"]  then
    ratings = get_title_string(tag_groups["Rating"])
end

PR #2560, opened 2016-09-30
AO3-4616 Add delays for the skin tests.

tickinginstant:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-4616

## Purpose

The updated_at field only has resolution down to the second, not the millisecond, so if two updates to the same skin are performed within the same second, the cache key used to access the cached version of the data won't change between updates. But in order to ensure that caching always has the most recent version of the data, the cache key has to change whenever the underlying data changes.

This pull request adds a small delay to two of the automated skins tests, in an attempt to ensure that the cache key has changed when the data changes. This is a stopgap measure that will hopefully stop the intermittent Travis skin test failures until the larger caching problem is handled (likely by updating Rails).

## Testing

This is an update to the automated tests, so I don't think it really requires any manual QA. (Aside from possibly making Travis CI run the other_b tests over and over again to check whether this fixes the problem.)

## Credit

tickinginstant, she/her

### DIFF
--- a/features/other_b/skin.feature
+++ b/features/other_b/skin.feature
@@ -212,6 +212,8 @@ Feature: creating and editing skins
     And I submit
   Then I should see "Your preferences were successfully updated."
     And I should see "margin: auto 5%; max-width: 100%" within "style"
+    # Make sure that the creation/update cache keys are different:
+    And I wait 1 second
   When I edit the skin "Wide margins" with the wizard
     And I fill in "Work margin width" with "4.5"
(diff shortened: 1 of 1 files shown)


sarken:
This looks good! I'm re-running the test on Travis now and I'll kick it a few more times to make sure.

sarken:
I've run this a bunch of times and it hasn't failed, so I'm marking this ready to merge!

PR #2561, opened 2016-09-30
AO3-4343 Support emails say "sent_at" in footer

sarken:
https://otwarchive.atlassian.net/browse/AO3-4343

## Purpose

Stops displaying the words `sent_at` in the Support automated response email and instead starts displaying the date.

## Notes

This should fix it everywhere it was happening, not just the Support automated response, because the problem was the mailer layout that we use for all the user mailers. So it probably also fixes https://otwarchive.atlassian.net/browse/AO3-4657

### DIFF
--- a/app/views/layouts/mailer.html.erb
+++ b/app/views/layouts/mailer.html.erb
@@ -98,7 +98,7 @@
                           <%= t 'mailer.general.footer.general.html', support_url: root_url + "support", donate_url: donate_url %>
 
                           <% if content_for?(:sent_at) %>
-                            <p><%= t 'mailer.general.footer.sent_at', sent_at: :sent_at %></p>
+                            <p><%= t 'mailer.general.footer.sent_at', sent_at: yield(:sent_at).strip %></p>
                           <% end %>
 
                           <% if content_for?(:sent_by) %>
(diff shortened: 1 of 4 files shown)


app/views/layouts/mailer.html.erb
```diff
@@ -98,7 +98,7 @@
                           <%= t 'mailer.general.footer.general.html', support_url: root_url + "support", donate_url: donate_url %>
 
                           <% if content_for?(:sent_at) %>
-                            <p><%= t 'mailer.general.footer.sent_at', sent_at: :sent_at %></p>
+                            <p><%= t 'mailer.general.footer.sent_at', sent_at: yield(:sent_at).strip %></p>
```
sarken: The `.strip` was necessary because otherwise it includes any spaces used for indenting, e.g.

```
<% content_for :sent_at do %>
  <%= Time.now.to_s(:time_for_mailers) %>
<% end %>
```

I'd have included a nice little comment in the code if the mailer views weren't so darn sensitive about the spacing. :|

zz9pzza: I don't understand why the yield helps ?

PR #2563, opened 2016-09-30
AO3-4655 Fix optional tag display on prompts

sarken:
https://otwarchive.atlassian.net/browse/AO3-4655

## Purpose

Fixes the alignment of the optional tags in challenge sign up prompt blurbs. (I moved the prompt class to the index that contains the prompts, not the prompts themselves, to bring it in line with the rest of the site -- we want to be able to do `.prompt .blurb` to style things, not `.prompt` and pray it's only a blurb.)


app/views/prompts/_prompt_blurb.html.erb
```diff
@@ -1,5 +1,5 @@
 <% # expects "prompt", if index is passed will use that in title if none exists, if suppress_claims is true will ignore %>
-<li class="<% if is_author_of?(prompt) %>own <% end %>prompt blurb group" role="article">
+<li class="<% if is_author_of?(prompt) %>own <% end %> blurb group" role="article">
```
ariana-paris: This could probably be changed to have less erb verbiage, e.g.:
```
<% is_author_of?(prompt) ? "own" : "" %>
```
(or the same thing with an `if` statement

PR #2564, opened 2016-10-01
AO3-4660 Issue with removing something from an index.

zz9pzza:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-4660

## Purpose

Clean up an error message and add a diagnostic so we can go looking for the real issue

## Testing

I don't believe this needs testing ( Delete a work and ensure it is removed from feeds etc )

### DIFF
--- a/app/models/indexing/index_subqueue.rb
+++ b/app/models/indexing/index_subqueue.rb
@@ -120,8 +120,12 @@ class IndexSubqueue
   end
 
   def add_deletion_to_batch(id)
-    basics = { "_index" => klass.index_name, "_type" => klass.document_type, "_id" => id }
-    @batch << { delete: basics }.to_json
+    if klass.respond_to? :index_name 
+      basics = { "_index" => klass.index_name, "_type" => klass.document_type, "_id" => id }
+      @batch << { delete: basics }.to_json
+    else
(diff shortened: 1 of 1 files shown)


app/models/indexing/index_subqueue.rb
```diff
@@ -120,8 +120,12 @@ def add_document_to_batch(obj)
   end
 
   def add_deletion_to_batch(id)
-    basics = { "_index" => klass.index_name, "_type" => klass.document_type, "_id" => id }
-    @batch << { delete: basics }.to_json
+    if klass.respond_to? :index_name 
+      basics = { "_index" => klass.index_name, "_type" => klass.document_type, "_id" => id }
+      @batch << { delete: basics }.to_json
+    else
+      Rails.logger.error "Brownshirt: #{klass.class.name} passed to add_deletion_to_batch Unknown"
```
sarken: This isn't related the the functionality, but could you change this to another color shirt, or even "Browncoat," like in Firefly? "Brownshirt" usually refers to a Nazi, and I think it would be better to avoid including the term in our codebase, even though it's not an intentional reference.

sarken:
It's one of our normal test failures and this passed tests before, so I'm going to live dangerously~ and merge this.

PR #2566, opened 2016-10-05
AO3-1654 Make "n more comments in this thread" a link.

tickinginstant:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-1654

## Purpose

When a work has a comment thread that is very deeply nested, the thread doesn't display the most deeply-nested comments when you view the work -- instead, it displays the text `(%{count} more comments in this thread)`, and you can click on the `Thread` link for the parent comment to view the hidden comments. When all comments visible on the work's comments page are deleted, however, there are no `Thread` links to click on, so the deeply nested comments become inaccessible to anyone without a direct link.

As suggested in the bug report, this pull request makes the `%{count} more comments in this thread` text a link, making the nested comments accessible again.

## Testing

The bug report has some steps that can be used for testing.

One thing to keep in mind for trying to force the page to hide some deeply nested comments: if a comment has only one direct child, no matter how deep that comment is, the comment display code won't hide it. It'll only hide a comment's descendants if the comment's depth is more than 5, **and** the comment has more than one **direct** reply. (The QA testers probably already know this, but I had trouble with this when I tried to set up my own tests, so I figured I'd mention it just in case.)

### DIFF
--- a/app/views/comments/_comment_thread.html.erb
+++ b/app/views/comments/_comment_thread.html.erb
@@ -9,7 +9,7 @@
       <% if child_comments && child_comments.count > 0 %>
         <!-- cut off the recursion when we get too deep, but not if there's just one last measly comment left -->
         <% if depth >= ArchiveConfig.COMMENT_THREAD_MAX_DEPTH && child_comments.count > 1 %>
-          <li class="comment"><p>(<%= link_to ts("%{count} more comments in this thread", :count => comment.children_count), comment_url(comment) %>)</p></li>
+          <li class="comment"><p>(<%= link_to ts("%{count} more comments in this thread", count: comment.children_count), comment_url(comment) %>)</p></li>
         <% else %>
           <li>
             <%= render 'comments/comment_thread', comments: child_comments, depth: depth+1 %>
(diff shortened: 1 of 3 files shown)


app/views/comments/_comment_thread.html.erb
```diff
@@ -9,7 +9,7 @@
       <% if child_comments && child_comments.count > 0 %>
         <!-- cut off the recursion when we get too deep, but not if there's just one last measly comment left -->
         <% if depth >= ArchiveConfig.COMMENT_THREAD_MAX_DEPTH && child_comments.count > 1 %>
-          <li class="comment"><p>(<%= ts("%{count} more comments in this thread", :count => comment.children_count) %>)</p></li>
+          <li class="comment"><p>(<%= link_to ts("%{count} more comments in this thread", :count => comment.children_count), comment_url(comment) %>)</p></li>
```
sarken: Could you update this to use `count: comment.children_count` rather than the hash rocket?

features/comments_and_kudos/comments_delete.feature
```diff
@@ -47,4 +47,23 @@ Feature: Delete a comment
     Then I should see "Comment deleted."
       And I should see "(Previous comment deleted.)"
       And I should see "I didn't mean that"
-  
+
+  Scenario: Deleting higher-level comments in a deep comment thread should still allow readers to access the deeper comments.
+
+    Given I am logged in as "author"
```
sarken: If the name of the work creator's not important -- it doesn't seem to be -- I think you could do just:

```
Given the work "Testing"
And I am logged in as "commenter"
```

tickinginstant:
Thanks for the feedback! I made both changes.

PR #2569, opened 2016-10-06
AO3-4375 Update tips and tricks page

sarken:
https://otwarchive.atlassian.net/browse/AO3-4375

## Purpose

Updates the Tips & Tricks first login help banner to: https://docs.google.com/document/d/1qIIM4GUFJl7Su8h21oBgUMVR6tn0PbCZw7hLfOYWUtI

Also removes the dismiss button that was in the popup, 'cause that was weird and ugly and not worth making less ugly given it already exists on the banner.

### DIFF
--- a/app/views/home/first_login_help.html.erb
+++ b/app/views/home/first_login_help.html.erb
@@ -44,12 +44,12 @@
   <h3 class="heading" id="edit_profile">
     <%= ts('Editing Your Profile, Password, and Preferences') %>
   </h3>
-  <p><%= ts('To add your info, go to %{homepage}, select the 
+  <p><%= ts('To add your info, go to %{dashboard}, select the 
   "%{profile}" tab, and select "%{edit_profile}" from the range of 
   options at the end of the page. Here, you can enter some basic personal 
   information. It\'s also where to go to change your password. Refer to the 
   %{profile_faq} for more information.',
-  homepage: (link_to ts('your home page'), user_path(current_user)),
+  dashboard: (link_to ts('your Dashboard'), user_path(current_user)),
   profile: (link_to ts('Profile'), user_profile_path(current_user)),
   edit_profile: (link_to ts('Edit My Profile'), edit_user_path(current_user)), 
   profile_faq: (link_to ts('Profile FAQ'), archive_faqs_path + '/profile')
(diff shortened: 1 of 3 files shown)


app/views/home/first_login_help.html.erb
```diff
@@ -1,49 +1,233 @@
...
+
+  <h3 class="heading" id="tags">
+    <%= ts('Tags') %>
+  </h3>
+  <p><%= ts('All tags on the Archive—including those for fandoms, relationships, 
+  and characters—start out user-created. You can always use the existing tags by 
+  selecting from the autocomplete list. If you cannot find the tags you want to 
+  use, feel free to create new tags for your works. Behind the scenes, our tag 
+  wrangling team will match the tags up with their synonyms, so that people can 
+  find your work whether you tag it "<abbr>AMTDI</abbr>", "Aliens Made Them Do
```
ariana-paris: Should that `abbr` have a title - otherwise, maybe drop it?

app/views/home/first_login_help.html.erb
```diff
@@ -1,49 +1,233 @@
...
+  ).html_safe %></p>
+
+  <h3 class="heading" id="legal_stuff">
+    <%= ts('Terms of Service') %>
+  </h3>
+  <p><%= ts('You can keep up to date with the policies and procedures for the 
+  site listed in the current %{tos}, and find explanations for some of the more 
+  common questions in the %{tos_faq}. If you still have questions about a 
+  specific work, please %{contact_abuse}. If you have general policy questions, 
+  you can either contact Abuse or %{contact_support}.',
```
ariana-paris: should the "contact Abuse" text be a link too?

app/views/home/first_login_help.html.erb
```diff
@@ -1,49 +1,233 @@
...
+      </ul>
+    </li>
+    <li><%= link_to ts('Terms of Service'), '#legal_stuff' %></li>
+    <li><%= link_to ts('Support and Feedback'), '#support' %></li>
+  </ul>
+
+  <h3 class="heading" id="logging_in">
+    <%= ts('Logging In and Logging Out') %>
+  </h3>
+  <p><%= ts('To log in, locate the login link and fill in your Username and
```
ariana-paris: As this is text approved by several people, I assume this is how we want it, but the capitalisation looks a bit weird. (Maybe "fill in the Username and Password fields" would be better for next time)

sarken: I'll pass that along!

app/views/home/first_login_help.html.erb
sarken: I wouldn't use a title attribute here -- partly to make it match the tag, partly to make sure screenreaders don't confuse someone by reading "Aliens Made Them Do It, Aliens Made Them Do It, or Sex Pollen." But consistent use of the <abbr> tag is a good idea even without the title attribute, since it can be used for styling or JavaScript or whatever else down the road. (It's also Doc's policy to use the tag in most cases.)

app/views/home/first_login_help.html.erb
sarken: I think it's probably okay given there's already a link just a few words before. If it were a different section, I'd definitely say yes.

PR #2570, opened 2016-10-09
AO3-4667 Add a button to allow a work to be reindexed.

zz9pzza:
# Pull Request Checklist
## Issue

https://otwarchive.atlassian.net/browse/AO3-4667
## Purpose

This adds a button on a work that a tag wrangler or admin should be able to see which will send the work off to be reindexed.
## Testing

Unless we have any badly indexed works testing will be hard.

We could stop the workers, do some tag wrangling, clear the queue, start the workers and see the work was in the wrong places and then press the button ?


app/controllers/works_controller.rb
```diff
@@ -785,6 +785,17 @@ def update_multiple
     end
   end
 
+  # Reindex the work.
+  def reindex
+    if logged_in_as_admin? || permit?("tag_wrangler")
+      RedisSearchIndexQueue.queue_works([params[:id]], priority: :high)
+      flash[:notice] = ts('Work queued to be reindexed')
+    else
+      flash[:notice] = ts("Sorry, you don't have permission to access the page you were trying to reach.")
```
sarken: I'd go with something like "to perform this action" or "to reindex works" or something, since we're not really trying to access a page

app/views/works/_work_header_navigation.html.erb
```diff
@@ -1,6 +1,9 @@
 <!-- BEGIN navigation -->
 <h3 class="landmark heading"><%= ts("Actions") %></h3>
 <ul class="work navigation actions" role="menu">
+  <% if ( logged_in_as_admin? || permit?("tag_wrangler") ) %>
```
sarken: Ruby style-wise, I don't think there should be spaces inside parentheses.

app/views/works/_work_header_navigation.html.erb
```diff
@@ -1,6 +1,9 @@
...
+    <li class="reindex"><%= link_to ts("Reindex Work"), reindex_work_path(@work) %></li>
+  <% end %>
```
sarken: For readability, can you add a blank line between this and `<% if is_author_of?(@work) %>`?

app/views/works/_work_header_navigation.html.erb
```diff
@@ -1,6 +1,9 @@
 <!-- BEGIN navigation -->
 <h3 class="landmark heading"><%= ts("Actions") %></h3>
 <ul class="work navigation actions" role="menu">
+  <% if ( logged_in_as_admin? || permit?("tag_wrangler") ) %>
+    <li class="reindex"><%= link_to ts("Reindex Work"), reindex_work_path(@work) %></li>
```
sarken: I really want to add some styling to this like we do with the admin show/hide options, but I need a couple of hours or a day to think of what.

features/admins/admin_works.feature
```diff
@@ -5,6 +5,13 @@ Feature: Admin Actions for Works and Bookmarks
 
   Scenario: Can hide works
```
sarken: I think this should be "Scenario: Can reindex works."

features/tags_and_wrangling/tag_wrangling.feature
```diff
@@ -189,6 +189,10 @@ Feature: Tag wrangling
     Then I should see "Tag was updated"
       And I should see "Stargate Atlantis"
 
+    # A wrangler should see the reindex button
```
sarken: This test scenario is not the greatest the way it's currently written -- they're really only supposed to test one thing (lolsob) -- so if you have a minute to make at least this part of it its own scenario, like:

```
Scenario: Wrangler has option to reindex a work

Given the work "Indexing Issues"
  And I am logged in as a tag wrangler
When I view the work "Indexing Issues"
Then I should see "Reindex Work"
```

that'd be dandy. If not, it's an existing problem, with this particular test, so it's not worth holding up the code for it.

features/works/work_view.feature
```diff
@@ -5,6 +5,7 @@ Feature: View a work with various options
   Scenario: viewing a work in explicit View Full Work mode, with JavaScript turned off (Issue 2205)
   Given the chaptered work with 2 comments "Whatever"
   When I view the work "Whatever" in full mode
+  Then I should not see "Reindex Work"
```
sarken: This definitely needs to go into its own scenario, or perhaps two. (The wording for the steps here isn't at all accurate.)

```
Scenario: Regular logged-in user doesn't have the option to reindex a work
Given I am logged in
  And I have the work "Whatever"
When I view the work "Whatever"
Then I should not see "Reindex Work"

Scenario: Logged-out user doesn't have the option to reindex a work
Given the work "Whatever"
  And I am logged out
When I view the work "Whatever"
Then I should not see "Reindex Work"
```

app/views/works/_work_header_navigation.html.erb
sarken: Just poking about removing these parentheses again

app/views/works/_work_header_navigation.html.erb
sarken: After the style rules for `li.relationships a` in https://github.com/otwcode/otwarchive/blob/master/public/stylesheets/site/2.0/10-types-groups.css#L168, let's do

```
.reindex a {
  border-bottom: 3px solid #900;
}

to give the button a red bottom border, like the teal one on admin actions. I don't love using red on buttons, but it's good enough for now.

sarken:
Alas, this has conflicts now. /o\

sarken:
Unfortunately, this has merge conflicts and needs to be updated before we can hit the button.

PR #2571, opened 2016-10-09
AO3-4669 Button to fix common issues with users accounts

zz9pzza:
# Pull Request Checklist
- [ ] Have you added tests for any changed functionality?
## Issue

https://otwarchive.atlassian.net/browse/AO3-4669
## Purpose

This should fix some common issues with users. 

Subscription Page 500
Stats Page 500
Reindexing a User's Works
Reindexing a User's Bookmarks
## Testing

Some of this is going to be hard to test like works or bookmarks not being indexed. But I could find a user with a broken subscription and see if this fixes it for them.


app/controllers/users_controller.rb
```diff
@@ -46,6 +46,35 @@ def check_account_creation_status
     end
   end
 
+  # Run a set of corrections on a user
```
sarken: I think this needs moar comments -- up here it should at least give a vague overview of the problems we're fixing (like you did in the pull request description), and it wouldn't hurt to basically break the code up into paragraphs, with a comment before each one explaining, e.g. (and my apologies for including the +; I'm lazy)

```
 +    # Delete any subscriptions the user has to deleted items because this causes
 +    # the user's subscription page to error
 +    @subscriptions = @user.subscriptions.includes(:subscribable)
 +    @subscriptions.to_a.each do |sub| 
 +      if sub.name.nil? 
 +        sub.destroy
 +      end
 +    end 
 +
 +    # Do something with the revised at date for the user's works to fix something and then
 +    # reindex the user's works to make sure they show up on the user's works page
 +    @user.works.each do |work| 
 +      if work.revised_at.nil?
 +        work.save
 +      end
 +      IndexQueue.enqueue(w, :main)
 +    end
 +
 +    # comment
 +    @user.bookmarks.each do |bookmark|
 +      bookmark.update_index
 +    end
 +
 +    # comment
 +    @user.update_works_index_timestamp!
```

app/controllers/users_controller.rb
```diff
@@ -46,6 +46,35 @@ def check_account_creation_status
...
+  def check
```
sarken: The button says "Reindex User," but this appears to do more than that. I'm honestly not sure what the best solution is here -- rename the button or separate the actions. 

Renaming the button something more vague like "Troubleshoot Account" is certainly easiest, but it means running checks we don't necessarily need (although I don't think any of them are resource-intensive, so that's not a _terrible_ problem), and it's not particularly informative from an interface standpoint. 

Maybe the thing to do would be rename the button, add some text to the top of the page the button appears on saying "To do X, Y, and Z to this account, choose 'Button Name,'" and then include a note here in the controller to update the view if adding/removing functions.

app/controllers/users_controller.rb
```diff
@@ -46,6 +46,35 @@ def check_account_creation_status
...
+    unless logged_in_as_admin?
```
sarken: Should all of this code be in the admin_users_controller instead? https://github.com/otwcode/otwarchive/blob/master/app/controllers/admin/admin_users_controller.rb

app/controllers/users_controller.rb
zz9pzza: So in unix terms it would be an fsck however I don't think that is useful.

Sanity check, is ableist.

Check user, gives the wrong idea.

zz9pzza: I am happy enough with "Troubleshoot Account". Would you be happy with "To fix relatively potential common errors with the account choose "Troubleshoot Account" this should not have any negative effects"

app/controllers/users_controller.rb
sarken: I'm not sure about the model vs controller, so let's poll the audience! @shalott, do you have an opinion?  :D 

My overall feeling, though, is that if we _do_ move it to the model, each action should be a separate method -- one like clean_up_subscriptions, another like set_works_revised_at_date, another for reindex_users_works, another for reindex_users_bookmarks. (Bad names, but you get the idea.)

app/controllers/users_controller.rb
sarken: "Troubleshoot Account" it is! For the note, I was thinking something a little more detailed, e.g. `To fix common errors with this user's Subscriptions and Stats pages, and to reindex their works and bookmarks, choose "Troubleshoot Account."`.

app/models/user.rb
```diff
@@ -488,6 +488,35 @@ def remove_from_queue
...
+    # Delete any subscriptions the user has to deleted items because this causes
+    # the user's subscription page to error
+    @subscriptions = self.subscriptions.includes(:subscribable)
+    @subscriptions.to_a.each do |sub|
+      if sub.name.nil?
+        sub.destroy
+      end
+    end
+
+    # AKA "stats page unfucking". It is most often caused by the existence of works with nil revised_at dates.
```
sarken: I'd suggest something a little more meaningful to future coders than "stats page unfucking", like `Fix user stats page error caused by the existence of works with nil revised_at dates`

app/models/user.rb
```diff
@@ -488,6 +488,35 @@ def remove_from_queue
...
+    # reindex the user's works to make sure they show up on the user's works page
+    self.works.each do |work|
+      if work.revised_at.nil?
+        work.save
+      end
+      IndexQueue.enqueue(w, :main)
+    end
+
+    # Reindex a users bookmakes.
```
sarken: A couple of typos -- should be "user's" and "bookmarks"

app/views/admin/admin_users/show.html.erb
```diff
@@ -5,6 +5,7 @@
 
   <!--subnav-->
   <ul class="navigation actions" role="navigation">
+    <li><%= link_to ts('Troubleshoot Account'), check_user_path(@user) %> </li>
```
sarken: There's an extra space before the `</li>`

app/controllers/users_controller.rb
ariana-paris: I've found that a good rule of thumb for "model vs controller" is to ask the question "Would a full-blown API need access to this method?". If it's something that is only ever going to be relevant to the UI, then put it in the controller. Otherwise, put it in the model or a helper that an API controller can also get at.

I also agree that separate methods can help readability and granularity. Any modern compiler or interpreter worth its salt will inline them if required, so always go for human readability and potential reuse if it make sense.

shalott: Hm, I would say that all these methods should be separate even if invoked with a single controller button, and I am inclined to say they should actually only be in the admin_users controller precisely because we _don't_ want them exposed except to admins going through our main interface -- james, because that controller is for admin actions on users it is more appropriate IMO than including them directly in the users controller and I think it's important to keep a strong separation for the admin-only actions. The button itself is only appearing under admin_users view anyway so it also makes more sense for it to be an admin_users action than a users action I think.

shalott: Sorry, to clarify, what I mean is we should have `def clean_up_subscriptions(user)` and `def set_works_revised_at_date(user)` etc as private methods in the controller that are not exposed via routes, and then the method `troubleshoot_common` would call all of them. 

Although as I type this, I am thinking we might want to use these methods in more places after all, in which case they probably should go in the user model, and then troubleshoot_common would just call them from admin_users. Anyway, I think that is a debatable question, but I do feel strongly they should be in admin_users and not users.

zz9pzza: @shalott Your last paragraph seems contradictory to me.

My instinct is that these are operations on user accounts and therefore should be in the user model. I am more than happy to split them up in to different fixes. 

Can you clarify please.

shalott: Sorry, @zz9pzza, I was going back and forth because I could see it going either way. But if you want my final rec, I would say put the individual operations in the user model, and then put the controller action "troubleshoot_account" into the admin_users controller.

app/controllers/admin/admin_users_controller.rb
```diff
@@ -247,6 +247,18 @@ def send_notification
     redirect_to :action => :notify
   end
 
+  def check
+    @user = User.find_by_login(params[:id])
+    @user.fix_user_subscriptions
```
sarken: I think shalott had the right idea about renaming this action "troubleshoot" rather than "check" -- we're fixing rather than merely observing what's wrong (checking). The success message could be something like "User account troubleshooting complete." 

Also, there's a discrepancy in how the methods are named -- some use `users` and others `user`. I'd suggest going with `user` on all of them to avoid any confusion where `users` is assumed to be a plural rather than a possessive. 

For `check_users_work_dates`, I'd again suggest a verb other than `check` -- maybe `set_user_work_dates`, since we're setting work dates where it's currently nil.

app/views/admin/admin_users/show.html.erb
```diff
@@ -5,6 +5,7 @@
 
   <!--subnav-->
   <ul class="navigation actions" role="navigation">
```
sarken: I think we also wanted to add `<p class="note">To fix common errors with this user's Subscriptions and Stats pages, and to reindex their works and bookmarks, choose "Troubleshoot Account."<p>` on this page, probably above the <h3> for the form.

config/config.yml
```diff
@@ -356,6 +356,7 @@ ACTION_WARN: 6
 ACTION_RENAME: 7
 ACTION_PASSWORD_RESET: 8
 ACTION_NEW_EMAIL: 9
+ACTION_CHECK: 10
```
sarken: If the action is renamed "troubleshoot," this should probably change as well.

app/controllers/admin/admin_users_controller.rb
```diff
@@ -239,6 +239,18 @@ def send_notification
...
   end
 
+  def troubleshoot
+    @user = User.find_by_login(params[:id])
+    @user.fix_user_subscriptions
+    @user.reindex_user_works
+    @user.set_user_work_dates
+    @user.reindex_user_bookmarks
+    @user.create_log_item(options = { action: ArchiveConfig.ACTION_TROUBLESHOOT, admin_id: current_admin.id })
+    flash[:notice] = ts('User account checked.')
```
sarken: Yay, looking good! The only thing is the flash message got overlooked -- it should be something like "User account troubleshooting complete." or "User account troubleshooting successful."

After that, I think we're ready to go! :tada:

features/admins/admin_manage_users.feature
```diff
@@ -136,6 +136,14 @@ Feature: Admin Actions to manage users
     And the "Fannish next of kin's username" field should contain "harrysmom"
     And the "Fannish next of kin's email" field should contain "harrysmom@example.com"
 
+  Scenario: A user account is Troubleshoot 
+  Given the user "mrparis" exists and is activated
+    And I am logged in as an admin
+   When I go to the abuse administration page for "mrparis"
+    And I follow "Troubleshoot Account"
+   Then I should see "User account checked."
```
sarken: The test is failing because this should be looking for 'User account troubleshooting complete.'

app/views/admin/admin_users/show.html.erb
```diff
@@ -4,7 +4,9 @@
   <!--/descriptions-->
 
   <!--subnav-->
+  <p class="note">To fix common errors with this user's Subscriptions and Stats pages, and to reindex their works and bookmarks, choose "Troubleshoot Account."<p>
```
sarken: This needs to be `</p>` at the end. Also, I think this should be in <%= ts() %>?

app/helpers/users_helper.rb
```diff
@@ -160,36 +157,37 @@ def authors_header(collection, what = "People")
 
   def log_item_action_name(action)
     if action == ArchiveConfig.ACTION_ACTIVATE
-      t('users_helper.log_validated', :default => 'Account Validated')
+      t('users_helper.log_validated', default: 'Account Validated')
+    elsif action == ArchiveConfig.ACTION_CHECK
```
sarken: This should be changed to `ACTION_TROUBLESHOOT` and `Account Troubleshot`. (Yep, "troubleshot" is what the dictionary says is the past tense: http://www.merriam-webster.com/dictionary/troubleshoot) Also, I think it should go last -- the other items are arranged by the action number, like in the config file.

app/helpers/users_helper.rb
```diff
@@ -1,156 +1,153 @@
...
+      when 0
+        "0 #{what}"
+      when 1
+        "1 #{what.singularize}"
+      else
+        collection.total_entries.to_s + " #{what}"
       end
     else
-      %{ %d - %d of %d }% [
+      %( %d - %d of %d ) % [
```
sarken: The space between the `)` and the `%` doesn't seem right to me.

app/views/admin/admin_users/show.html.erb
zz9pzza: I would rather leave the ts in views until we know how we are going to do them.

sarken: Fine with me as long as it won't make it harder to track down the strings! (I assumed we were going to search for `ts('')` when it was time for us to make our switch.)

app/helpers/users_helper.rb
```diff
@@ -1,156 +1,153 @@
...
+      when 0
+        "0 #{what}"
+      when 1
+        "1 #{what.singularize}"
+      else
+        collection.total_entries.to_s + " #{what}"
       end
     else
-      %{ %d - %d of %d }% [
+      %( %d - %d of %d )% [
```
sarken: Ah-ha! I see now -- the % sign goes with the _square brackets_. Sorry, this was my bad and I told you to remove the wrong space. It should be `%( %d - %d of %d ) %[`. Then we'll be ready to merge.

app/helpers/users_helper.rb
```diff
@@ -160,36 +157,37 @@ def authors_header(collection, what = "People")
...
-		elsif action == ArchiveConfig.ACTION_PASSWORD_RESET
-      t('users_helper.log_password_change', :default => 'Password Changed')
-		elsif action == ArchiveConfig.ACTION_NEW_EMAIL
-      t('users_helper.log_email_change', :default => 'Email Changed')
+      t('users_helper.log_rename', default: 'Username Changed')
+    elsif action == ArchiveConfig.ACTION_PASSWORD_RESET
+      t('users_helper.log_password_change', default: 'Password Changed')
+    elsif action == ArchiveConfig.ACTION_NEW_EMAIL
+      t('users_helper.log_email_change', default: 'Email Changed')
+    elsif action == ArchiveConfig.ACTION_TROUBLESHOT
```
sarken: Wait, also: typo here -- it should be `ArchiveConfig.ACTION_TROUBLESHOOT` for the config item name. (The past tense below is correct.)

app/helpers/users_helper.rb
```diff
@@ -160,36 +157,37 @@ def authors_header(collection, what = "People")
...
+      t('users_helper.log_validated', default: 'Account Troubleshot')
```
sarken: Even if we're not currently using it, this should be `log_troubleshot` or something -- right now it duplicates another key from line 160

PR #2574, opened 2016-10-13
AO3-4668 Add a reindex button on a tag 

zz9pzza:
# Pull Request Checklist
- [ ] Have you added tests for any changed functionality?
- [ ] Have you updated the JIRA issue with the information below?
## Issue

https://otwarchive.atlassian.net/browse/AO3-4668
## Purpose

Reindex the works ( and the associated bookmarks ) that are either directly on a tag or on a synonyms of the tag.
## Testing

If we can find a broken tag we should be able to test it ?


app/views/tags/show.html.erb
```diff
@@ -9,10 +9,13 @@
           <li><%= link_to ts('Works'), {:controller => :works, :action => :index, :tag_id => @tag} %></li>
           <li><%= link_to ts('Bookmarks'), {:controller => :bookmarks, :action => :index, :tag_id => @tag} %></li>
         <% end %>
-    	  <% if can_wrangle? %>
+    	<% if can_wrangle? %>
       	  <li><%= link_to ts('Edit'), {:controller => :tags, :action => :edit, :id => @tag} %> </li>
       	  <li><%= tag_comment_link(@tag) %></li>
         <% end %>
+        <% if logged_in_as_admin? %>
+          <li><%= link_to ts('Reindex Tag'), reindex_tag_path(@tag) %> </li>
```
sarken: There's an extra space before the `</li>`

app/controllers/tags_controller.rb
```diff
@@ -24,32 +24,49 @@ def load_tag_and_subtags
     end
   end
 
+  def reindex
```
sarken: I've never had the pleasure of reindexing the works for a tag, but this looks reasonable! Is there someone else who could review it with a bit more authority?

zz9pzza: It is what the tag wranglers asked me to do ?

sarken: What I mean is, can someone else review the code?

app/controllers/tags_controller.rb
```diff
@@ -24,32 +24,49 @@ def load_tag_and_subtags
...
+    work_ids = @tag.work_ids
+    @tag.synonyms.each do |syn|
+      work_ids.push syn.work_ids
+    end
+    work_ids.flatten!
+    reindex_all_works(work_ids)
+    flash[:notice] = ts('Tag sent to be reindexed')
+    redirect_to(request.env['HTTP_REFERER'] || root_path) && return
+  end
+
```
shalott: The code looks reasonable to me but there is no test and this should have one.

zz9pzza: I can write a test that presses the button and sees if the right message is displayed but I don’t know how to break things so the fixing it does something useful.

features/tags_and_wrangling/tag_wrangling.feature
```diff
@@ -227,3 +227,32 @@ Feature: Tag wrangling
...
     Then I should not see "Sign Up"
       And I should see the tag wrangler listed as an editor of the tag
+
+  Scenario: A user can not see the reindex button on a tag page
+
+    Given the following typed tags exists
+        | name              | type         | canonical |
+        | Cowboy Bebop      | Fandom       | true      |
+      And I am logged in as a random user
+      When I view the tag "Cowboy Bebop"
```
sarken: Your Whens and Thens are indented too much in these tests -- they need to line up with the Givens, not the Ands.

PR #2575, opened 2016-10-16
AO3-4670 Extend test suite to cover admin activating users accounts

zz9pzza:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-4670

## Purpose

This extends the test suite and removes the controller action which allowed an admin to delete a users account which was not available in the user interface.

## Testing

Nothing to test

### DIFF
--- a/features/step_definitions/user_steps.rb
+++ b/features/step_definitions/user_steps.rb
@@ -275,7 +275,7 @@ Then /^I should get confirmation that I changed my username$/ do
   step(%{I should see "Your user name has been successfully updated."})
 end
  
-Then /^the user "([^"]*)" is activated$/ do |login|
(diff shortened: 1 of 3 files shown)


features/step_definitions/user_steps.rb
```diff
@@ -267,4 +274,8 @@ def get_series_name(age, classname, name)
 Then /^I should get confirmation that I changed my username$/ do
   step(%{I should see "Your user name has been successfully updated."})
 end
- 
\ No newline at end of file
+ 
+Then /^the user "([^"]*)" is activated$/ do |login|
```
sarken: My only suggestion would be to use "should be activated" rather than "is activated" since we're checking that something is true (like with "I should see 'some text'" or "one email should be sent")

PR #2578, opened 2016-10-16
AO3-4673 Add test for setting and unsetting skins

zz9pzza:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-4673

## Purpose

Adds a test to check that a logged in user can set a temporary skin

## Testing

Nothing to test.

### DIFF
--- a/features/other_b/skin.feature
+++ b/features/other_b/skin.feature
@@ -7,14 +7,42 @@ Feature: creating and editing skins
   When I am on skinner's preferences page
   Then "Default" should be selected within "preference_skin_id"
 
-  Scenario: A user should be able to choose a different public skin in their preferences
+  Scenario: User can set a skin for a session and then unset it
   Given basic skins
     And the approved public skin "public skin" with css "#title { text-decoration: blink;}"
-    And I am logged in as "skinner"
-  When I change my skin to "public skin"
-  When I am on skinner's preferences page
-  Then "public skin" should be selected within "preference_skin_id"
-    And I should see "text-decoration: blink;" within "style"
+    And the skin "public skin" is cached
+    And the skin "public skin" is in the chooser
+  When I am logged in as "skinner"
+    And I follow "public skin"
+  Then I should see "The skin public skin has been set. This will last for your current session."
+    And the page should have the cached skin "public skin"
+  When I follow "Default"
+  Then I should see "You are now using the default Archive skin again!"
+    And the page should not have the cached skin "public skin"
+
+  Scenario: Admin can cache a public skin
+  Given basic skins
+    And the approved public skin "public skin" with css "#title { text-decoration: blink;}"
+    And I am logged in as an admin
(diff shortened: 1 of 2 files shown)


features/other_b/skin.feature
```diff
@@ -433,3 +433,19 @@ Feature: creating and editing skins
     And I should see "only screen and (max-width: 42em), only screen and (max-width: 62em)"
   When I press "Use"
   Then the page should have a skin with the media query "only screen and (max-width: 42em), only screen and (max-width: 62em)"
+
+  Scenario: A user should be able to choose a temporary skin
+  Given basic skins
+    And the approved public skin "public skin" with css "#title { text-decoration: blink;}"
+   Then I am logged in as an admin
```
sarken: Is the reason we're doing these log in as admin/check "Cache"/etc steps to make sure the skin is cached so we can put it in the chooser? If so, I'd suggest setting it up as a step, like you did with "the skin  "([^\"]*)" is in the chooser":
```
When /^the skin "([^\"]*)" is cached$/ do |skin_name|
  skin = Skin.find_by_title(skin_name)
  skin.cached = true
  skin.save
  skin.cache!
end
```

I might also add two steps to make sure we're actually using (or not using) the cached skin rather than relying solely on the success message:
```
Then /^the page should have the cached skin "([^"]*)"$/ do |skin_name|
  skin = Skin.find_by_title(skin_name)
  page.should have_xpath("//link[contains(@href, '#{skin.skin_dirname}')]")
end

Then /^the page should not have the cached skin "([^"]*)"$/ do |skin_name|
  skin = Skin.find_by_title(skin_name)
  page.should_not have_xpath("//link[contains(@href, '#{skin.skin_dirname}')]")
end
```

Then I would do:
```
  Scenario: User can set a skin for a session and then unset it
  Given basic skins
    And the approved public skin "public skin" with css "#title { text-decoration: blink;}"
    And the skin "public skin" is cached
    And the skin "public skin" is in the chooser
  When I am logged in as "skinner"
    And I follow "public skin"
  Then I should see "The skin public skin has been set. This will last for your current session."
    And the page should have the cached skin "public skin"
  When I follow "Default"
  Then I should see "You are now using the default Archive skin again!"
    And the page should not have the cached skin "public skin"
```

zz9pzza: Is it better to do that kind of thing or step though the interface, which will give us better coverage ?

I am happy to add the extra tests

sarken: Adding coverage for the interface is a good idea, yep! I think the thing to do is use the shortened version I suggested for users and then set up a test for "Scenario: Admin can cache a public skin" specifically for testing caching the skin through the interface, and maybe even one for "Scenario: Admin can add a public skin to the chooser"

sarken:
Let's merge this and see how Coveralls feels about the addition of some new tests...

PR #2582, opened 2016-10-19
AO3-4678 Tests for destroying wrangling guidelines and assignments

sarken:
https://otwarchive.atlassian.net/browse/AO3-4678

## Purpose

Automated tests for
* admin removing a wrangler's wrangling permission (which leaves their assignments intact)
* admin removing an assignment from a wrangler (tag_wraglers #destroy)
* tag wrangler removing their own assignment (tag_wranglers #destroy)
* admin deleting a wrangling guideline (wrangling_guidelines #destroy)

Some of these use JavaScript, so if unlike my computer, Travis doesn't like them, we can comment them out :|

(Annoyingly, there are no success messages to check for after performing these actions.)


features/step_definitions/admin_steps.rb
```diff
@@ -170,6 +170,12 @@ def update_settings(settings)
   click_button("Post")
 end
 
+Given(/^the following language exists$/) do |table|
```
sarken: (I just moved this from the Whens to the Givens, since it starts with "Given")

zz9pzza:
"(Annoyingly, there are no success messages to check for after performing these actions.)" 

We could add the messages and test for them ?

There are some files which do not have a trailing new line.

zz9pzza:
I am happy to merge but wanted you to have a chance to look at comments.

sarken:
I think we should make a new issue for adding success messages (with a note to update the tests), since if we change any code instead of just tests, we'll want to wait longer to merge this because it'll need manual QA.

I'll edit the files here to add a newline -- I've never been able to figure out how to do it in TextWrangler.

PR #2589, opened 2016-10-21
AO3-4680 Add tests for Admin activities.

zz9pzza:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-4680

## Purpose

Increases coverage in Admin activities ( even if it is only two lines )

## Testing

Nothing to test


features/admins/admin_works.feature
```diff
@@ -93,6 +93,10 @@ Feature: Admin Actions for Works and Bookmarks
       And I should see "Mature"
       And I should see "Admin-Added Relationship"
       And I should see "Admin-Added Character"
+     When I follow "Activities"
+     Then I should see "View Admin Activity"
+     When I visit the last activities item
+     Then I should see "No Archive Warnings Apply"
```
sarken: I wasn't familiar with this page, and from the test alone, I wasn't very sure what information it should display, so I went and played around until I figured it out. Could you maybe change the `Then` to make it clearer that the page is expected to show the old data and not the new? Something like:

```
Then I should see "Old tags"
  And I should see "User-Added Fandom"
  And I should not see "Admin-Added Fandom"
```

features/admins/admin_manage_users.feature
```diff
@@ -195,6 +195,12 @@ Feature: Admin Actions to manage users
   Then I should see "User has been permanently suspended."
     And I should see "Suspended Permanently"
     And I should see "To the New Zealand penal colony with you."
+  Then I follow "Manage Users"
```
sarken: Should be `When`, since we're taking an action (following a link).

features/admins/admin_manage_users.feature
```diff
@@ -195,6 +195,12 @@ Feature: Admin Actions to manage users
...
+   And I fill in "query" with "mrparis"
+   And I press "Find"
+  Then I should see "1 user found"
+  Then I follow "Details"
```
sarken: Should be `When` since we're taking an action (following a link).

features/admins/admin_manage_users.feature
```diff
@@ -195,6 +195,12 @@ Feature: Admin Actions to manage users
...
+   And I should see "To the New Zealand penal colony with you."
```
sarken: Should be `Then` since we're checking that the results of a previous action are what we expect.

features/step_definitions/admin_steps.rb
```diff
@@ -399,4 +399,8 @@ def update_settings(settings)
   step %{I post the comment "I loved this" on the work "New Work" as a guest with email "#{email}"}
   step %{I should not see "has been blocked at the owner's request"}
   step %{I should see "Comment created!"}
-end
\ No newline at end of file
+end
+
+Then(/^I visit the last activities item$/) do
```
sarken: You were right in the scenario itself -- should be `When` because it's an action (visiting a page).

PR #2590, opened 2016-10-22
AO3-4688 Test gift exchange deletion

sarken:
https://otwarchive.atlassian.net/browse/AO3-4688

## Purpose

Tests that sign-ups and assignments are deleted when a gift exchange is deleted, and also ensures that the sign-up and assignment pages of exchange participants are still accessible after the exchange is deleted. 

These tests show https://otwarchive.atlassian.net/browse/AO3-3994 ("Deleting the gift exchange from a collection causes the Signups pages of participants to give Error 500") is fixed.

Commit 57fcbba contains the new tests; commit e381210 contains a bit of refactoring so we're not repeating ourselves so much when we just want to assume an exchange's assignments have been sent. Commit 86dc39a lets us delete a challenge without the going through the interface, since going to the settings page and clicking the button is already tested in its own scenario.

## Testing

No manual testing required.

### DIFF
--- a/features/gift_exchanges/challenge_giftexchange.feature
+++ b/features/gift_exchanges/challenge_giftexchange.feature
@@ -277,24 +277,17 @@ Feature: Gift Exchange Challenge
       And the email should link to myname1's user url
       And the email html body should link to the works tagged "Stargate Atlantis"
 
-  Scenario: User signs up for two gift exchanges at once #'
-    Given I am logged in as "mod1"
-      And I have created the gift exchange "Awesome Gift Exchange"
-      And I open signups for "Awesome Gift Exchange"
-      And everyone has signed up for the gift exchange "Awesome Gift Exchange"
-      And I have generated matches for "Awesome Gift Exchange"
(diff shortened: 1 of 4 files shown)


features/gift_exchanges/challenge_giftexchange.feature
```diff
@@ -277,24 +277,17 @@ Feature: Gift Exchange Challenge
...
-      And I open signups for "Awesome Gift Exchange"
-      And everyone has signed up for the gift exchange "Awesome Gift Exchange"
-      And I have generated matches for "Awesome Gift Exchange"
-      And I have sent assignments for "Awesome Gift Exchange"
-    Given I have created the gift exchange "Second Challenge" with name "testcoll2"
-      And I open signups for "Second Challenge"
-      And everyone has signed up for the gift exchange "Second Challenge"
-      And I have generated matches for "Second Challenge"
-      And I have sent assignments for "Second Challenge"
+  Scenario: User signs up for two gift exchanges at once and can use the Fulfill
```
sarken: Just improving the comments to better reflect the issue and the expected fix.

PR #2591, opened 2016-10-22
AO3-4675 API controller unit tests

ariana-paris:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4675

## Purpose

Adds Rspec tests to cover some areas left untested by the original tests. Also uncovered a couple of minor bugs (exception not caught, error message not returned).

Also standardises on double quotes throughout instead of a mix of single and double quotes.

## Testing

No manual testing required.

### DIFF
--- a/spec/api/api_base_controller_spec.rb
+++ b/spec/api/api_base_controller_spec.rb
@@ -1,59 +1,48 @@
-require 'spec_helper'
-require 'api/api_helper'
+require "spec_helper"
+require "api/api_helper"
 
 include ApiHelper
 
-describe 'Base API Controller' do
+describe "Base API Controller" do
+
+  before do
+    allow(User).to receive(:is_archivist?).and_return(true)
+  end
 
   before(:each) do
     @under_test = Api::V1::BaseController.new
   end
 
-  describe 'batch_errors with no archivist' do
-    it 'should return the "forbidden" status' do
-      status, _ = @under_test.instance_eval { batch_errors(nil, api_fields) }
-      assert_equal status, :forbidden
-    end
+  describe "batch_errors" do
 
-    it 'should return an error message' do
-      _, messages = @under_test.instance_eval { batch_errors(nil, api_fields) }
-      assert_equal "The 'archivist' field must specify the name of an Archive user with archivist privileges.", messages[0]
-    end
-  end
+    describe "with no archivist" do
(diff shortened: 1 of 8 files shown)


spec/api/api_base_controller_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  before(:each) do
+    @under_test = Api::V1::BaseController.new
+  end
+
+  describe "batch_errors" do
+
+    describe "with no archivist" do
+      it "should return the 'forbidden' status" do
+        status, _ = @under_test.instance_eval { batch_errors(nil, api_fields) }
```
ariana-paris: This `instance_eval` means that the test can exercise the private method `batch_errors` in the controller without having to make it public

spec/api/api_base_controller_spec.rb
```diff
@@ -0,0 +1,86 @@
+require "spec_helper"
+require "api/api_helper"
+
+include ApiHelper
+
+describe "Base API Controller" do
+
+  before do
+    allow(User).to receive(:is_archivist?).and_return(true)
```
ariana-paris: This fakes the `is_archivist?` call on `User` to always return `true` (so effectively, all users are archivists in these tests)

spec/api/api_base_controller_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+      user = @user
+      _, messages = @under_test.instance_eval { batch_errors(user, nil) }
+      assert_equal "No items to import were provided.", messages[0]
+    end
+
+    it "should return error messages with too many items to import" do
+      user = @user
+      loads_of_items = Array.new(210) { |_| api_fields }
+      _, messages = @under_test.instance_eval { batch_errors(user, loads_of_items) }
+      assert_equal "This request contains too many items to import. A maximum of 200 items can be imported at one time by an archivist.", messages[0]
```
zz9pzza: Should this pull in the config variable eg

assert_equal "This request contains too many items to import. A maximum of #{ArchiveConfig.IMPORT_MAX_CHAPTERS} items can be imported at one time by an archivist.", messages[0]

Or which ever config variable it should be ?

ariana-paris: Or even not bother to check the part of the message that includes it as it's testing that there is a message, not what it is. Done.

PR #2597, opened 2016-10-26
AO3-4696 Link fandom page back in

zz9pzza:
# Pull Request Checklist
## Issue

https://otwarchive.atlassian.net/browse/AO3-4696
## Purpose

The fandom page should be available from the tag page, if the tag is a fandom.
## Testing

Look on the tag page and check it is there on a fandom page and not on a different type of tag.


zz9pzza:
I need to add the test.

app/views/tags/show.html.erb
```diff
@@ -21,8 +21,13 @@
   <!--/descriptions-->
 
   <!--main content-->
-  <p><%= ts("This tag belongs to the %{tag_class} Category.", :tag_class => @tag.class::NAME) %>
+  <p>
   <% if @tag.canonical %>
+    <% if @tag.is_a?(Fandom) %>
+      This tag belongs to the Fandom Category, a list of <%= link_to ts('all the relationships is available'), fandom_path(@tag.name) %>.
```
sarken: I'd suggest making "list of all the relationships" the link text and leaving the other words out of the link (but keep them in the sentence, of course). Also, there's a comma splice going on here -- it should be two separate sentences: "This tag belongs to the Fandom Category. A list of all the relationships is available." 

However, now I have a question! Is it a list of relationship _tags_, e.g. Kirk/Spock, or a list of tags that are related to the fandom? Maybe it should be "A list of all the Relationship tags in this fandom is available" (which I think is what's going on here) or "A list of all tags related to this Fandom is available" for clarity.

app/views/tags/show.html.erb
```diff
@@ -23,6 +23,9 @@
   <!--main content-->
   <p><%= ts("This tag belongs to the %{tag_class} Category.", :tag_class => @tag.class::NAME) %>
   <% if @tag.canonical %>
+    <% if @tag.is_a?(Fandom) %>
+      A list of all the Relationship tags in this <%= link_to ts('fandom is available'), fandom_path(@tag.name) %>.
```
sarken: Should this be a `<%= ts() %>`?

Also, I'm afraid this link text won't quite work -- an accessible link describes the designate page even when taken out of context, and "fandom is available" doesn't mean much if you just say it to someone who's not looking at the whole sentence. It would be better if the words "Relationship tags in this fandom" were the link.

zz9pzza:
I suspect I could pull the tests out of this if we want to push it in earlier ( for the whole 3 lines )

sarken:
Looks good! I think we can leave the tests and the change combined and wait until the next deploy... we have a lot in this one already.

PR #2602, opened 2016-10-28
AO3-4692 multiple external byline

ariana-paris:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4692
## Purpose

Fixes a bug which caused only one external author created by an archivist to be displayed on imported works.
## Testing

This also removes duplication in the byline code.
- check that existing stories have the same byline they normally have (including Anonymous)
- check that importing a work as an archivist with an author and co-author who do not have accounts on the Archive displays both authors in the format `name1 [archived by archivist], name2 [archived by archivist]`

### DIFF
--- a/features/step_definitions/user_steps.rb
+++ b/features/step_definitions/user_steps.rb
@@ -45,7 +45,12 @@ end
 
 Given /^I am logged in as "([^\"]*)" with password "([^\"]*)"(?:( with preferences set to hidden warnings and additional tags))?$/ do |login, password, hidden|
   step("I am logged out")
-  find_or_create_user(login, password, hidden)
+  user = find_or_create_new_user(login, password)
+  if hidden.present?
+    user.preference.hide_warnings = true
+    user.preference.hide_freeform = true
+    user.preference.save
+  end
   visit login_path
   fill_in "User name", :with => login
   fill_in "Password", :with => password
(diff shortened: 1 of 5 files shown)


features/step_definitions/user_steps.rb
```diff
@@ -45,20 +45,7 @@
...
-    user.password = password
-    user.password_confirmation = password
-    user.save
-  end
-  if hidden.present?
-    user.preference.hide_warnings = true
-    user.preference.hide_freeform = true
-    user.preference.save
-  end
+  find_or_create_user(login, password, hidden)
```
zz9pzza: Is it better to move it ?

ariana-paris: Move it where?

ariana-paris:
Yes, there was a bit of duplication in the byline code and the tests so this was an opportunity to address that, while we're pruning...

features/step_definitions/user_steps.rb
```diff
@@ -148,40 +109,40 @@
...
   FactoryGirl.create(:work, authors: [author1, author2], posted: true, title: "Shared")
 end
 
 # WHEN
 
-When /^I follow the link for "([^\"]*)" first invite$/ do |login|		
-  user = User.find_by_login(login)		
-  invite = user.invitations.first		
-  step(%{I follow "#{invite.token}"})		
+When /^I follow the link for "([^"]*)" first invite$/ do |login|
```
zz9pzza: Why has this dropped the \ to be consistent ? 

The other regex's have it in the first string ( It does not need it but it is I feel more right ).

PR #2604, opened 2016-10-28
AO3-4474 italicised/justified text being clipped

pixelred:
https://otwarchive.atlassian.net/browse/AO3-4474

small issue, but the fix suggested in the description works like a charm... i added

`
padding: 0 0.25em;
`

to **#worskin** in **21-userstuff.css** and tested it in safari, chrome, firefox and opera. i tried resizing the window and zooming in and out to see if any misalignment issues or clipping would occur but it seemed just fine 👍 i clicked around on various documentation pages and user pages to see if the tweak caused any unexpected spacing issues, but again, everything looked ship shape. 

0.25em was more than enough space to fix the problem, since it was only clipping long letters like t and f and l. could someone else also check it out in their own browser to make sure the extra padding doesn't cause any unexpected spacing issues?

### DIFF
--- a/public/stylesheets/site/2.0/21-userstuff.css
+++ b/public/stylesheets/site/2.0/21-userstuff.css
@@ -146,6 +146,7 @@
 #workskin {
   font-size: 1.08em;
   margin: auto;
+  padding: 0 0.25em;
   max-width: 72em;
   overflow-x: auto;
   overflow-y: hidden;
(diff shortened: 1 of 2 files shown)


sarken:
Thank you! We'll definitely give this a poke in a couple of browsers when we merge it and deploy it to staging.

pixelred:
thank _you_ :' )

pixelred:
sweats;;;; um... i didn't mean to add that commit to this OTL

PR #2605, opened 2016-10-29
AO3-4705 Test for downloading unrevealed works

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4705
## Purpose

This tests that unrevealed works can't be downloaded. Unfortunately, I discovered they can be! So I made an issue (AO3-4706). 

Because my test does squat until the issue is fixed, I figured I'd also clean up the wip tests a bit and remove the guest downloading test, which is already covered in admins/admin_settings.feature.
## Testing

No testing needed.


features/works/work_download.feature
```diff
@@ -5,180 +5,172 @@
 Feature: Download a work
   @wip
```
ariana-paris: Any idea why these are marked with `@wip` and whether this is strictly necessary? I suppose once upon a time someone was running the tests manually using `~@wip" to exclude tests with this tag, but we don't do that anymore.

features/step_definitions/work_steps.rb
```diff
@@ -160,6 +160,13 @@
   end
 end
 
+Given /^the work "([^\"]*)" in the unrevealed collection "([^\"]*)"$/ do |work, collection|
```
ariana-paris: FYI, quote marks aren't special in Regex and therefore don't need to be escaped, so the placeholders could be `([^"]*)` (meaning any number of characters that are not a quote) or for that matter just `(.*)` (meaning any number of characters until you meet whatever is after this, which in this case will be the first quote after the placeholder)

features/works/work_download.feature
sarken: It looks like taking `@wip` out of the file leads to the tests running and thus failing, so I've put it back in and maybe some brave soul will be able to finish this some day.

ariana-paris: Github has no suitable emoji. Oh well, something to add to the list of things to fix some day.

PR #2606, opened 2016-10-29
AO3-2431 Make collection fandom count skip metatags

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-2431
## Purpose

The fandom counts for collections are a little inconsistent. The fandom count in the sidebar includes all metatags for all fandoms used on works in the collection; the list of fandoms doesn't. This pull request changes the fandom count in the sidebar (and the fandom count in the collection's blurb) to skip inherited taggings, to make it match the query used to generate the list of fandoms.
## Testing

The example given in the bug report no longer works, but there are very detailed steps on how to reproduce this issue in the bug report's comments.

### DIFF
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -309,15 +309,14 @@ class Collection < ActiveRecord::Base
   end
 
   def all_fandoms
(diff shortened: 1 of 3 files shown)


shalott:
Looks good!

PR #2608, opened 2016-10-30
AO3 4704 Extend tests for the tag_wranglings controller.

zz9pzza:
# Pull Request Checklist
## Issue

https://otwarchive.atlassian.net/browse/AO3-4704
## Purpose

Extend tests for the tag_wranglings controller.
## Testing

No manual testing required.

### DIFF
--- a/features/step_definitions/generic_steps.rb
+++ b/features/step_definitions/generic_steps.rb
@@ -2,10 +2,6 @@ When /^(?:|I )unselect "([^"]+)" from "([^"]+)"$/ do |item, selector|
   unselect(item, :from => selector)
 end
 
-When /^(?:|I )pry in to the system$/ do
-  require 'pry'
-  binding.pry
-end
 Then /^debug$/ do
   binding.pry
 end
--- a/features/step_definitions/generic_steps.rb
+++ b/features/step_definitions/generic_steps.rb
@@ -2,6 +2,10 @@ When /^(?:|I )unselect "([^"]+)" from "([^"]+)"$/ do |item, selector|
   unselect(item, :from => selector)
 end
 
+When /^(?:|I )pry in to the system$/ do
+  require 'pry'
+  binding.pry
+end
 Then /^debug$/ do
   binding.pry
 end
--- a/features/tags_and_wrangling/tag_wrangling_more.feature
+++ b/features/tags_and_wrangling/tag_wrangling_more.feature
@@ -163,17 +163,60 @@ Feature: Tag wrangling: assigning wranglers, using the filters on the Wranglers 
        And I press "Wrangle"
      Then I should not see "Cowboy Bebop"
 
-  Scenario: more wrangling....Relationships( Canonical edition )
+  Scenario: more wrangling....Relationships
     Given the following typed tags exists
         | name                                   | type         | canonical |
-        | October Daye Series - Seanan McGuire   | Fandom       | true      |
         | Tybalt (Toby Daye)                     | Character    | false     |
-        | October "Toby" Daye                    | Character    | true      |
-        | Toby Daye/Tybalt                       | Relationship | false     |
+        | October Daye                           | Character    | false     |
+        | Toby Daye/Tybalt                       | Relationship | true      |
+        | October Daye Series - Seanan McGuire   | Fandom       | false     |
       And a media exists with name: "Books & Literature", canonical: true
       And I am logged in as a random user
-      And I post the work "Brain Scratch" with fandom "Cowboy Bebop" with freeform 
+      And I post the work "The course of true love never did run smooth" with fandom "October Daye Series - Seanan McGuire" with character "Tybalt (Toby Daye)" with second character "October Daye" with relationship "Toby Daye/Tybalt"
+     When the tag wrangler "lain" with password "lainnial" is wrangler of "October Daye Series - Seanan McGuire"
+       And I go to the tagwranglings page
+       And I follow "Relationships by fandom (1)"
+       And I check the wrangling option for "Toby Daye/Tybalt"
+       And I fill in "fandom_string" with "October Daye Series - Seanan McGuire"
+       And I press "Wrangle"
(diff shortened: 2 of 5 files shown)


features/step_definitions/generic_steps.rb
```diff
@@ -2,6 +2,10 @@
   unselect(item, :from => selector)
 end
 
+When /^(?:|I )pry in to the system$/ do
+  require 'pry'
+  binding.pry
```
zz9pzza: Ha, really no its useful

sarken: Just wondering -- can you explain the difference between this step and the one below it, which is:
```
Then /^debug$/ do
    binding.pry
end
```
Basically, why do we need one with `require 'pry'` and one without?

Also, I think you need a blank line between these two steps.

features/tags_and_wrangling/tag_wrangling_more.feature
```diff
@@ -141,10 +141,83 @@ Feature: Tag wrangling: assigning wranglers, using the filters on the Wranglers
        And I check the canonical option for the tag "Faye Valentine is a sweetie"
        And I check the canonical option for the tag "Ed is a sweetie"
        And I press "Wrangle"
-     Then I should see "The following tags were successfully wrangled to Cowboy Bebop: Faye Valentine is a sweetie, Ed is a sweetie Wrangle Tags for Cowboy Bebop"
+     Then I should see "The following tags were successfully wrangled to Cowboy Bebop: Faye Valentine is a sweetie, Ed is a sweetie"
        And the "Faye Valentine is a sweetie" tag should be canonical
        And the "Ed is a sweetie" tag should be canonical
 
+  Scenario: more wrangling....Freeform
```
sarken: Could you make this a little more descriptive? What specifically are we testing here? What I see happening is we're making sure we can mass wrangle in the fandoms bin, but I can't tell:

- Why this test needs to create freeform tags
- Why this test needs to create works
- Why the test needs to create a media category

features/tags_and_wrangling/tag_wrangling_more.feature
```diff
@@ -141,10 +141,83 @@ Feature: Tag wrangling: assigning wranglers, using the filters on the Wranglers
...
+      And I post the work "Brain Scratch" with fandom "Cowboy Bebop" with freeform "Spike Spiegel is a sweetie"
+      And I post the work "Asteroid Blues" with fandom "Cowboy Bebop" with freeform "Faye Valentine is a sweetie"
+     When the tag wrangler "lain" with password "lainnial" is wrangler of "Cowboy Bebop"
+       And I go to the tagwranglings page
+       And I follow "Fandoms by media (2)"
+       And I check the wrangling option for "Cowboy Bebop"
+       And I press "Wrangle"
+     Then I should not see "Cowboy Bebop"
+
+  Scenario: more wrangling....Relationships
```
sarken: Again, I think we need a better description for the scenario. What I see happening is we're making sure a relationship can't be wrangled into a fandom that isn't a canonical tag, but I'm not sure

- Why the character tags are necessary
- Why the media tag is necessary
- Why the work is necessary

features/tags_and_wrangling/tag_wrangling_more.feature
```diff
@@ -141,10 +141,83 @@ Feature: Tag wrangling: assigning wranglers, using the filters on the Wranglers
...
+      And I post the work "The course of true love never did run smooth" with fandom "October Daye Series - Seanan McGuire" with character "Tybalt (Toby Daye)" with second character "October Daye" with relationship "Toby Daye/Tybalt"
+     When the tag wrangler "lain" with password "lainnial" is wrangler of "October Daye Series - Seanan McGuire"
+       And I go to the tagwranglings page
+       And I follow "Relationships by fandom (1)"
+       And I check the wrangling option for "Toby Daye/Tybalt"
+       And I fill in "fandom_string" with "October Daye Series - Seanan McGuire"
+       And I press "Wrangle"
+     Then I should see "The following names are not canonical fandoms: October Daye Series - Seanan McGuire."
+
+  Scenario: more wrangling....Relationships canonically
```
sarken: Same thing --  I think we need a better description for the scenario. What I see happening is we're making sure a relationship can be wrangled into a canonical fandom from the relationships bin, but I'm not sure

- Why the character tags are necessary
- Why the media tag is necessary
- Why the work is necessary

features/tags_and_wrangling/tag_wrangling_more.feature
```diff
@@ -141,10 +141,83 @@ Feature: Tag wrangling: assigning wranglers, using the filters on the Wranglers
...
+      And I post the work "The course of true love never did run smooth" with fandom "October Daye Series - Seanan McGuire" with character "Tybalt (Toby Daye)" with second character "October 'Toby' Daye" with relationship "Toby Daye/Tybalt"
+     When the tag wrangler "lain" with password "lainnial" is wrangler of "October Daye Series - Seanan McGuire"
+       And I go to the tagwranglings page
+       And I follow "Relationships by fandom (1)"
+       And I check the wrangling option for "Toby Daye/Tybalt"
+       And I fill in "fandom_string" with "October Daye Series - Seanan McGuire"
+       And I press "Wrangle"
+     Then I should see "The following tags were successfully wrangled to October Daye Series - Seanan McGuire: Toby Daye/Tybalt"
+
+  Scenario: more wrangling....Characters can be canonical
```
sarken: Again, I think we need a better description for the scenario -- something like "Tag wranglers can mass wrangle characters into canonical fandoms from the Characters bin." (It occurs to me 0-use tags _might_ not show up in the wrangling bin, in which case I'd suggest shortening this by posting one work with two character tags and not creating the character tags in the `Given the following typed tags exists` step.)

features/tags_and_wrangling/tag_wrangling_more.feature
zz9pzza: The truthful answer of writing a test to check the other path and collect the last line in the routines worth of coverage doesn't count ?

PR #2613, opened 2016-11-05
AO3-4721 - Perform partial match by email in admin user search

ariana-paris:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4721

## Purpose

Enabled admins to search for users by a partial email address.

## Testing

Log in as an admin, go to the Find Users page and search for, for example, "gmail". Currently, this will only return users whose pseud includes "gmail". Now, it will also return users who email address includes "gmail".

## Reviewing
This includes style changes. Review commit aa0e949 for the actual functionality changes


scottsds:
Looks good! 👍

PR #2615, opened 2016-11-06
AO3-4719 Refactor work search tests

zz9pzza:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-4719

## Purpose

This just refactors the works searching, there will be another pull along in a bit to actually add coverage.

## Testing

No manual testing

### DIFF
--- a/spec/controllers/works_controller_spec.rb
+++ b/spec/controllers/works_controller_spec.rb
@@ -107,6 +107,7 @@ describe WorksController do
             fake_login
             get :index, tag_id: @fandom.name
             expect(assigns(:works).items).to match_array([@work, @work2])
+            restore_login
           end
         end
 
--- a/spec/controllers/works_controller_spec.rb
+++ b/spec/controllers/works_controller_spec.rb
@@ -64,7 +64,7 @@ describe WorksController do
         allow(controller).to receive(:use_caching?).and_return(true)
       end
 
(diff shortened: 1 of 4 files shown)


spec/controllers/works_controller_spec.rb
```diff
@@ -97,13 +97,13 @@
...
 
-          xit "should not show restricted works to guests" do
+          it "should not show restricted works to guests" do
             get :index, tag_id: @fandom.name
             expect(assigns(:works).items).to include(@work)
             expect(assigns(:works).items).not_to include(@work2)
           end
 
-          xit "should show restricted works to logged-in users" do
+          it "should show restricted works to logged-in users" do
```
ariana-paris: This is the test that breaks the import specs (I enabled it locally to check and the `story_parser` tests failed). I suspect it's due to the `fake_login` method which amongst other things, stubs the `current_user` call to return a user who isn't an archivist, and then doesn't reset normal calls afterwards. I think this test needs to clean up after itself.

zz9pzza:
@ariana-paris @sarken it is unclear if you want me to move this to Background ?

ariana-paris:
No, the Background thing was just a suggestion. I only requested changes it because you needed to fix the test, which I guess you did by putting it back the way it was. Something for another day!

PR #2616, opened 2016-11-08
AO3-4692 - Preserve external authors when one author is claimed

ariana-paris:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4692

## Purpose

Initial testing on my fix for AO3-4692 revealed that claiming a work (including automatic claiming when one author on an imported work already has an AO3 account) removes all other external authors from a work. This pull request fixes that issue.

## Testing

See the issue for testing notes.

Note: this is currently broken on Test.


scottsds:
Looks good to go in 👍

PR #2617, opened 2016-11-10
AO3-4724 Make rspec tests more independant

zz9pzza:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-4724

## Purpose

This resets the state of memcached, redis, and elastic search between each rspec test.

## Testing

No Manual testing required

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,8 @@ cache: bundler
 sudo: required
 env:
   - TEST_GROUP="./script/check_syntax"
-  - TEST_GROUP="rspec spec --tag ~rspec_run_two"
-  - TEST_GROUP="rspec spec --tag  rspec_run_two"
+  - TEST_GROUP="rspec spec"
+  - TEST_GROUP="rspec spec"
   - TEST_GROUP="cucumber -f progress -r features features/admins"
   - TEST_GROUP="cucumber -f progress -r features features/bookmarks"
   - TEST_GROUP="cucumber -f progress -r features features/collections"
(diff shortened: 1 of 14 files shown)


zz9pzza:
I am defering this as it causes issues on codeship, so I am still looking

.travis.yml
```diff
@@ -3,7 +3,8 @@ cache: bundler
 sudo: required
 env:
   - TEST_GROUP="./script/check_syntax"
-  - TEST_GROUP="rspec spec"
+  - TEST_GROUP="rspec spec --tag ~rspec_run_two"
+  - TEST_GROUP="rspec spec --tag  rspec_run_two"
```
ariana-paris: We discussed this in chat, but I think splitting the Rspec tests into two runs because of a conflict between the StoryParser tests and a single call to `fake_login` method in the WorksController test feels like overkill. 

As far as I can tell, the StoryParser test is failing because the StoryParser internally calls `User.current_user` and then attempts to get that user's default Pseud, but the `fake_login` method has stubbed `User.current_user` to return a user with no default pseud.

However, I have some good news! The behaviour for restricted works is already covered in https://github.com/otwcode/otwarchive/blob/master/features/works/work_lock.feature#L9-L49 so you can fix this problem by removing the test on line 106 of spec/controllers/works_controller_spec.rb altogether and the problem will go away. 🎉

zz9pzza:
@ariana-paris Is this how you would like it ?

PR #2619, opened 2016-11-12
AO3-4232 "If you accept cookies from our site" note floats oddly on wide screens

pixelred:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4232

## Purpose

makes the "if you accept cookies from our site and you choose proceed..." note sit beneath the proceed and go back buttons on all screen sizes.

## Testing

didn't seem to cause alignment issues when clicking around on other pages, but it'd be nice if the reviewer could double check 👌

tried to make the selector as specific as possible

## Credit

cosette is fine!

she/her is fine

### DIFF
--- a/public/stylesheets/site/2.0/22-system-messages.css
+++ b/public/stylesheets/site/2.0/22-system-messages.css
@@ -18,6 +18,10 @@ System messages use the following colours:
   float: right;
 }
 
+p.message.footnote {
+  clear: both;
+}
+
 .datetime {
   font-family: monospace;
 }
--- a/public/stylesheets/site/2.0/22-system-messages.css
+++ b/public/stylesheets/site/2.0/22-system-messages.css
@@ -18,7 +18,7 @@ System messages use the following colours:
   float: right;
 }
 
-p.message.footnote {
+.message.footnote {
   clear: both;
 }
 


public/stylesheets/site/2.0/22-system-messages.css
```diff
@@ -18,6 +18,10 @@ System messages use the following colours:
   float: right;
 }
 
+.message.footnote {
```
sarken: Sorry it took so long to get back to you on this! I wanted to give this some thought for a couple reasons: first, we like to avoid using chained selectors like `.class.class` to avoid specificity issues; and second, while we don't currently use this combination of classes anywhere else that I can find either, it's so generic I can definitely see us doing so in the future, and in a much different context. 

So I think what I'd suggest is using `.actions + .footnote` as the selector, because it most accurately describes  and targets the problem we're trying to solve: actions and footnotes are both floated, so it's important to clear `.footnote` when its an adjacent sibling of `.actions`. I checked and this is the only place where this situation is currently occurring -- most of our footnotes are adjacent siblings of text areas or text inputs.

When we modify something based on the context it appears in, we try to stick it in a little "context" section below all the basic rules (you can see an example in [08-actions.css](https://github.com/otwcode/otwarchive/blob/master/public/stylesheets/site/2.0/08-actions.css)). This file's divided up a little weirdly, so I'd suggest adding
```
/* contexts */

.actions + .footnote {
 clear: both;
}
```
directly above the big "ADMIN ANNOUNCEMENT BANNERS" section.

Thank you, and my apologies again for the delay!

pixelred:
no need to apologise, sarken ✌️

front-end code requires careful consideration in many ways, so thank you for being so diligent :' )

i'll make these changes and poke around the site a bit to make sure there are no surprises. i spotted the contexts example in 08-actions (just to clarify, there's no need to write an actual context like 'addresses floating footnotes')?

sarken:
Great, thank you! And nope, no need to add detailed comments in the CSS -- if the selector's not clear enough, we can always check the file's history/blame and check the JIRA issue.

sarken:
Closing in favor of #2630

PR #2620, opened 2016-11-12
AO3-4708 No padding around drafted works in yellow container

pixelred:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4708

## Purpose

adds some padding to the drafts container. this addition fixes it but i'm not sure i placed it in the right stylesheet... i did some searching but i couldn't quite figure out where this selector should be 😞


sarken:
Closing, since discussion on the JIRA ticket reveals this was a problem with an unsupported skin.

PR #2624, opened 2016-11-19
AO3-4372 Change wording on Error 500 page to be more useful

cjrecordvt:
https://otwarchive.atlassian.net/browse/AO3-4732
This change would more proactively ask for user feedback on Error 500s and request more meaningful information. Link code taken from footer.erb. (No existing JIRA issue found.)


cjrecordvt:
screwed up and included the merges, Killing this request.

cjrecordvt:
Okay, no, it's only showing the one line changed.

sarken:
Also: do we want the link text to just say "contact Support" rather than repeat the footer text? I think we use plain old "contact Support" in most places.

cjrecordvt:
Even fixed the caps on "Contact"

public/500.html
```diff
@@ -97,7 +97,7 @@ <h3 class="landmark heading">Site Navigation</h3>
     <div id="main" class="system errors error-500 region">
       <h2 class="heading">Error 500</h2>
       <h3 class="heading">We're sorry, but something went wrong.</h3>
-      <p>We've been notified about this issue and we'll take a look at it shortly.</p>
+      <p>If you are receiving this error repeatedly, please <%= link_to ts('contact Support'), new_feedback_report_path %>. In the form, please include a link to the page you're trying to reach and how you're trying to reach this page.</p>
```
sarken: Since this is a plain old static page, the link here will need to be just regular HTML (preferably just "/support" rather than the full URL), not Ruby. 

Sorry, my fault for not explicitly saying that!

PR #2629, opened 2016-11-23
AO3-4738 Allow an Admin to change the default skin

zz9pzza:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-4738

## Purpose

Previously changing the default skin would bring the site to its knees as the code to compute the style went in to a recursive heap.

## Testing

Change the default skin and check that it is changed for guests and for new users who are created after the default is changed ( but not for existing users).

## References

https://otwarchive.atlassian.net/browse/AO3-4618
https://otwarchive.atlassian.net/browse/AO3-4466


app/models/skin.rb
```diff
@@ -298,31 +295,31 @@ def get_all_parents
...
-      ie_comment += "gte " if ie_condition.match(/or_higher/)
-      ie_comment += "IE"
-      ie_comment += " #{$1}" if ie_condition.match(/IE(\d)/)
-      ie_comment += "]>" + style + "<![endif]-->"
+      ie_comment = '<!--[if '
+      ie_comment += 'lte ' if ie_condition =~ /or_lower/
+      ie_comment += 'gte ' if ie_condition =~ /or_higher/
+      ie_comment += 'IE'
+      ie_comment += " #{Regexp.last_match(1)}" if ie_condition =~ /IE(\d)/
+      ie_comment += ']>' + style + '<![endif]-->'
```
zz9pzza: Your wrong as the return is the last evaluated expression....

sarken:
I haven't looked at this completely, but can you please keep the double quotes? That's what we're using (since it makes things easier when we use contractions and possessives in strings, which come up fairly frequently).

PR #2633, opened 2016-11-25
AO3-4493 Make long unsubscribe text wrap

pixelred:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4493

## Purpose

makes long unsubscribe text wrap inside button. 

for the text inside the unsubscribe buttons to wrap, the height must be set to auto and white-space must be set to normal... but if input[type="submit"] had been set to a height of 1.929em with a specific purpose in mind, we can also target only the buttons in the subscribe section with a selector like 

.subscription input[type="submit"]

that being said, i didn't notice any immediate changes in button heights when testing the change on other pages, but i'll continue to test different pages and check for possible height issues.

### DIFF
--- a/public/stylesheets/site/2.0/03-region-header.css
+++ b/public/stylesheets/site/2.0/03-region-header.css
@@ -1,4 +1,4 @@
-/*==REGION: HEADER is #1 of the 4 regions shown on every page 
+/*==REGION: HEADER is #1 of the 4 regions shown on every page
 http://otwcode.github.com/docs/front_end_coding/patterns/supertypes
 notice that CSS3 declarations sit are indented twice (four spaces) and always come last*/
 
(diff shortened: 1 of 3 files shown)


sarken:
I think the old `height` value was necessary for certain older browsers that did funny things with the box sizing on inputs. However, that doesn't appear to be the case with newer browsers we actively support, so this fix should be fine!

Since we're adding `height: auto` here, can you edit 03-region-header.css to remove it from `#header #search .button`?  Also, a quick poke suggests that it's safe to remove this in its entirety from 18-zone-searchbrowse.css (it was added to allow the "Sort and Filter" text on the filters to wrap in the same way you're doing with these buttons):

`.filters .submit input {
  height: 100%;
  min-height: 1.286em;
  white-space: normal;
}`

Once those two changes are made, it'll be all set to merge! I'll make a note on the issue to test this in as many browsers and screen sizes as possible, though, just to make sure it doesn't create any unforeseen problems.

pixelred:
done! (ﾉ◕ヮ◕)ﾉ*:･ﾟ✧

PR #2636, opened 2016-11-27
AO3-4524 Repairing broken/redirecting link to OTW FAQ

pixelred:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4524

## Purpose

the link to the OTW FAQ in the AO3 "about" page was reportedly broken (though it redirected me to the correct page). the link has been changed to lead the visitor directly to the OTW FAQ page.


pixelred:
think it's best if both about page tweaks are addressed in AO3-4529

PR #2637, opened 2016-11-27
AO3-4529 Minor changes to AO3 "About" page

pixelred:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4529

but the change for [AO3-4524](https://otwarchive.atlassian.net/browse/AO3-4524) is also in this commit, so maybe [this pull request](https://github.com/otwcode/otwarchive/pull/2636) can be closed and both issues can be addressed in this pull request? since they're both concerning the same page...

## Purpose

makes changes to AO3 about page requested in [AO3-4524](https://otwarchive.atlassian.net/browse/AO3-4524) and [AO3-4529](https://otwarchive.atlassian.net/browse/AO3-4529)

### DIFF
--- a/app/views/home/about.html.erb
+++ b/app/views/home/about.html.erb
@@ -15,5 +15,5 @@ commercial exploitation and legal challenge.") %></li>
 that seeks to promote scholarship on fanworks and practices.") %></li>
     </ul>
   </p>
-  <p><%= ts("You can find out more about the OTW and its projects at its website,") %> <%= link_to ts("transformativeworks.org"), "http://transformativeworks.org" %><%= ts(", and learn about how your financial support is vital to the continuation and expansion of the OTW's work on its") %> <%= link_to ts("FAQ page"), "http://transformativeworks.org/faq-page" %>.</p>
+  <p><%= ts("You can find out more about the OTW and its projects at its website,") %> <%= link_to ts("transformativeworks.org"), "http://transformativeworks.org" %><%= ts(", and learn about how your financial support is vital to the continuation and expansion of the OTW's work on its") %> <%= link_to ts("FAQ page"), "http://transformativeworks.org/faq" %> <%= ts(". If you have a media or research question, please contact the") %> <%= link_to ts("Communications team"), "http://transformativeworks.org/contact_us/concomm" %>.</p>
 </div>


sarken:
Assigning james because he mentioned having some thoughts!

app/views/home/about.html.erb
```diff
@@ -15,5 +15,5 @@ commercial exploitation and legal challenge.") %></li>
 that seeks to promote scholarship on fanworks and practices.") %></li>
     </ul>
   </p>
-  <p><%= ts("You can find out more about the OTW and its projects at its website,") %> <%= link_to ts("transformativeworks.org"), "http://transformativeworks.org" %><%= ts(", and learn about how your financial support is vital to the continuation and expansion of the OTW's work on its") %> <%= link_to ts("FAQ page"), "http://transformativeworks.org/faq-page" %>.</p>
+  <p><%= ts("You can find out more about the OTW and its projects at its website,") %> <%= link_to ts("transformativeworks.org"), "http://transformativeworks.org" %><%= ts(", and learn about how your financial support is vital to the continuation and expansion of the OTW's work on its") %> <%= link_to ts("FAQ page"), "http://transformativeworks.org/faq" %><%= ts(". If you have a media or research question, please contact the") %> <%= link_to ts("Communications team"), "http://transformativeworks.org/contact_us/concomm" %>.</p>
```
zz9pzza: I would like us to use arguments to the ts call to allow the translators the most flexibility we can give them. For example I think the first one should look something like:

ts("You can find out more about the OTW and its projects at its website, %{transformative_works} , and learn about how your financial support is vital to the continuation and expansion of the OTW's work on its %{faq}.", :transformative_works => link_to(ts("transformativeworks.org"), "http://transformativeworks.org") , :faq => link_to ( ts("FAQ page"), "http://transformativeworks.org/faq" ) )

Note this is completely untested and just so hopefully you can understand what I mean.

pixelred: i see what you mean 👍

should i try making these changes throughout the whole page?

zz9pzza: While I don't think it would be wasted effort. I am not going to hold you to doing it. I believe we are going to change how ts works in the future ( so the index key is explicitly given ) I can't guarantee that it will not need a rework of sorts later.

pixelred: i don't mind making these changes, and it's good practice since i'm still not used to working with ruby~

ariana-paris: I think whatever method we eventually adopt and how we convert the `ts` calls, we _will_ need the strings to all be interpolated as James suggests. As it's only a small file and this is the only "real" change, I think you can go for it if you have the time and want the practice!

pixelred:
(i tried to make the changes, but i'm still iffy with syntax so i apologise if i made a mistake)

PR #2634, opened 2016-11-26
AO3-4739 Extend tests for admin skin controller

zz9pzza:
# Pull Request Checklist

* [x] Have you read [How to write the perfect pull request](https://github.com/blog/1943-how-to-write-the-perfect-pull-request)?
* [x] Have you read through the [contributor guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [x] Have you added tests for any changed functionality?
* [x] Have you added the JIRA issue number as the *first* thing in your pull request title (eg: `AO3-1234 Fix thing`)
* [x] Have you updated the JIRA issue with the information below?

## Issue

https://otwarchive.atlassian.net/browse/AO3-4739
## Purpose

Extends coverage over admin skin controller.

## Testing

No manual testing required

PR #2638, opened 2016-11-27
AO3-4740 Remove featured skins

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4740

## Purpose

This removes the featured flag on a skin which ensures it is at the top of a listing of skins.

## Testing

Check that each of the three index pages still work.

# Pull Request Checklist


sarken:
Should this also have a migration to drop the "featured" column from the skins table?

Other than that, this looks good to me! But I'm going to stick the Deferred label on it so we don't merge it until after the purge.

sarken:
(Also just leaving a note that if #2634 has been merged by the time this goes in, it will need to be updated to remove the tests for featuring skins.)

zz9pzza:
@sarken I had forgotten to add the migration.

db/migrate/20161127084744_remove_featuredfrom_skins.rb
```diff
@@ -0,0 +1,9 @@
+class RemoveFeaturedfromSkins < ActiveRecord::Migration
+  def up
+    remove_column :skins, :featured
+  end
+
+  def down
+    add_column :skins:, :featured
```
sarken: Hound is right -- there's a `:` that doesn't belong here. Other than that, looks good!

zz9pzza:
@sarken Could you clarify what this needs ?

sarken:
Sure, sorry! After this pull request was created, we merged #2634, so this will need to remove the "An admin can feature and unfeature skin" test.

sarken:
Thanks -- looks good!

sarken:
Merge conflicts

db/migrate/20161127084744_remove_featuredfrom_skins.rb
```diff
@@ -0,0 +1,9 @@
+class RemoveFeaturedfromSkins < ActiveRecord::Migration
```
redsummernight: This migration is currently not working because it's not in Rails 5 format.

```
Caused by:
StandardError: Directly inheriting from ActiveRecord::Migration is not supported. Please specify the Rails release the migration was written for:
  class RemoveFeaturedfromSkins < ActiveRecord::Migration[4.2]
```

app/controllers/skins_controller.rb
```diff
@@ -30,13 +30,13 @@ def index
       end
     else
       if is_work_skin
-        @skins = WorkSkin.approved_skins.sort_by_recent_featured
+        @skins = WorkSkin.approved_skins
```
tickinginstant: I think maybe you should change this to another scope like `sort_by_recent`, instead of deleting the scope altogether. The `sort_by_recent_featured` scope ordered first by `featured`, then by `updated_at`. So it'd show all the featured skins, ordered by last update, and then all of the unfeatured skins (also ordered by last update). By deleting this, you're deleting both ordering constraints, which means that the skins will be arranged in whatever order the database defaults to. (Which I think will be increasing order by ID. Oldest first, instead of recently updated first.)

I suspect this is what's causing the automated test error in `features/other_b/skin_public.feature` (which doesn't appear to be intermittent).

app/controllers/skins_controller.rb
```diff
@@ -30,13 +30,13 @@ def index
...
         @title = ts('Public Work Skins')
       else
         if logged_in?
-          @skins = Skin.approved_skins.usable.site_skins.sort_by_recent_featured
+          @skins = Skin.approved_skins.usable.site_skins
```
tickinginstant: Also here.

app/controllers/skins_controller.rb
```diff
@@ -30,13 +30,13 @@ def index
...
         else
-          @skins = Skin.approved_skins.usable.site_skins.cached.sort_by_recent_featured
+          @skins = Skin.approved_skins.usable.site_skins.cached
```
tickinginstant: And here.

features/other_b/skin.feature
```diff
@@ -24,6 +24,21 @@ Feature: Non-public site and work skins
   When I go to the new skin page
     Then I should see "Sorry, you don't have permission"
 
+  Scenario: An admin can reject and unreject a skin
```
tickinginstant: This looks like a duplicate of this test from the admin skins feature:
https://github.com/otwcode/otwarchive/blob/ab4e40182099780635b7af479c249c68f3d2f721/features/admins/admin_skins.feature#L35-L48
I'm guessing the process of resolving merge conflicts went a bit awry.

db/migrate/20161127084744_remove_featuredfrom_skins.rb
```diff
@@ -0,0 +1,9 @@
+class RemoveFeaturedfromSkins < ActiveRecord::Migration[4.2]
```
sarken: Just a note to update the date on this before it's merged so there won't be any issues with the updated schema (#3643).

db/migrate/20200703142643_remove_featured_from_skins.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveFeaturedFromSkins < ActiveRecord::Migration[5.1]
+  def change
+    remove_column :skins, :featured, :boolean, default: false, null: false
```
tickinginstant, line 3: Do we need a temporary `ignored_columns=` for the dropped column?

sarken:
Alas, no, the purge still hasn't happened.

sarken:
We'll need to bump the migration time stamp on this before it's merged, but that's a problem for way-in-the-future us.

sarken:
This still needs work and the purge is still pending, so I'm going to close this for now and set the issue to In Progress. However, I know you're _quite_ busy, so feel free to unassign the issue and set it to To Do instead.

PR #2641, opened 2016-11-28
AO3-3359 Add search button to top of advanced search form

pixelred:
## Issue

https://otwarchive.atlassian.net/browse/AO3-3359

## Purpose

adds a search button to the top of the advanced search form, and removes top padding from the new button to preserve the balance/symmetry of the page. (tried to make sure the selector would only target that one submit button, since this styling isn't necessary for any other buttons i don't think...)

## Testing

the changes shouldn't affect any other forms (or submit buttons) on the site.

### DIFF
--- a/app/views/works/_search_form.html.erb
+++ b/app/views/works/_search_form.html.erb
@@ -2,6 +2,7 @@
   <fieldset>
     <legend><%= ts('Work Info') %></legend>
     <h3 class="landmark heading"><%= ts('Work Info') %></h3>
+    <p class="submit"><%= f.submit ts('Search') %></p>
     <dl>
       <dt>
         <%= f.label :query, ts('Any Field') %>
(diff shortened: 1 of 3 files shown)


sarken:
Assigning myself to this -- will take a look when I have a minute!

pixelred:
okay! ✌️

i read that AO3 uses CSS 2.1 except for certain CSS3 properties like box shadow... and first-of-type is a CSS3 selector. but i also wasn't sure how else to target that one button using CSS 2.1 without giving the button its own class or ID. also wasn't sure if i should use padding-top or just redefine the padding using 4 numbers.

pixelred:
okay! done ✌️ did i add too much info to the sandbox comment;;

sarken:
Oh, sorry! I meant leave the HTML change in the form and just put the CSS in the sandbox like the rule for `.edit_external_author ul ul` is in there. We do want to merge and deploy the change for users to use; we're just not quite sure how the "button at the top of a form" design pattern is going work in the long term, so the styles should go in sandbox for now.

pixelred:
ahh okay! sorry about that. if i understood correctly then: the html should be applied to the form, but the css should not be applied to 08-actions and should instead be left in the sandbox (that's what i've done in this last commit)

sarken:
Yep, you got it! :+1:

PR #2643, opened 2016-11-30
AO3-4716 Add bulk email search feature in admin section

ariana-paris:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4716

## Purpose

Open Doors need to know if authors in external archives they are processing already have Archive accounts. This feature adds a new Bulk Email Search form in the Admin section which allows you to paste in a list of emails and then view which ones are associated with Archive users, and download the results as a CSV file.

## Testing

The **Bulk Email Search** feature is a new entry in Admin section > **Manage Users**. When you paste a list of line-break separated email addresses into the form and press **Find**, you should see:
- a plain comma-separated list of email addresses that weren't found
- a paginated table of user accounts for all the email addresses on the list that were found (this is the same table that appears when doing a normal search by user or email)
- if you press **Download CSV** instead of **Find**, you will get all of the above and your browser should download an Excel-friendly CSV file containing all the original addresses with their AO3 logins if they have one, or a blank cell in the Login column if they don't

The CSV code had to be moved out of the ChallengeSignupsController, so it is worth checking that prompt meme and gift exchange CSVs can still be downloaded (existing challenges would be enough to test this)

### DIFF
--- a/app/controllers/admin/admin_users_controller.rb
+++ b/app/controllers/admin/admin_users_controller.rb
@@ -19,7 +19,7 @@ class Admin::AdminUsersController < ApplicationController
       if params[:download_button]
         header = [["Email", "Username"]]
         found = all_users.map { |u| [u.email, u.login] }
-        not_found = @not_found.map { |email| [email, ""]}
+        not_found = @not_found.map { |email| [email, ""] }
         send_csv_data(header + found + not_found, "bulk_user_search_#{Time.now.strftime('%Y-%m-%d-%H%M')}.csv")
         flash.now[:notice] = ts("Downloaded CSV")
       end
--- a/app/controllers/admin/admin_users_controller.rb
+++ b/app/controllers/admin/admin_users_controller.rb
@@ -1,4 +1,5 @@
 class Admin::AdminUsersController < ApplicationController
+  include ExportsHelper
 
   before_filter :admin_only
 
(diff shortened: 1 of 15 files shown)


features/support/paths.rb
```diff
@@ -213,7 +201,23 @@ def path_to(page_name)
       tag_wranglings_path
     when /^the "(.*)" fandom relationship page$/i
       fandom_path($1)
-      
+
+    # Admin Pages
```
sarken: It's so beautiful :sob:

app/views/admin/admin_users/bulk_search.html.erb
```diff
@@ -0,0 +1,45 @@
+<div class="admin">
+  <!--Descriptive page name, messages and instructions-->
+  <h2 class="heading"><%= ts("Bulk Email Search") %></h2>
+  <!--/descriptions-->
+
+  <!--main content-->
+  <div class="notice">
```
sarken: We usually put the notice class directly on the paragraph, but other than that, the front end stuff looks good to me!

sarken:
Two quick notes: I'm not marking this reviewed because it still needs someone to look at the backend. Also, it has unfortunately gone grey, but just in the tests.

app/controllers/admin/admin_users_controller.rb
```diff
@@ -1,12 +1,29 @@
...
   def index
     @roles = Role.assignable.uniq
     @role_values = @roles.map{ |role| [role.name.humanize.titlecase, role.name] }
     @role = Role.find_by_name(params[:role]) if params[:role]
-    @users = User.search_by_role(@role, params[:query], :inactive => params[:inactive], :page => params[:page])
+    @users = User.search_by_role(@role, params[:query], inactive: params[:inactive], page: params[:page])
+  end
+
+  def bulk_search
+    @roles = Role.assignable.uniq
```
shalott: Suggest moving this setting of @roles to a before filter

shalott:
Tests and all looks great to me!

PR #2644, opened 2016-12-03
AO3-4743 Fix error on external works page showing only duplicates

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4743

## Purpose

Currently going to http://test.archiveofourown.org/external_works as an admin gives a 500 error.
As does http://archiveofourown.org/external_works?show=duplicates as a user.
Once this is deployed those paths should return pages which show only the external works which are duplicates.
We want to increase test coverage of the external works controller from 53.66%.

## Testing

Visit http://archiveofourown.org/external_works?show=duplicates and should not get a 500 error

# Pull Request Checklist

* [ ] Have you added the JIRA issue number as the *first* thing in your pull request title (eg: `AO3-1234 Fix thing`)
* [ ] Have you updated the JIRA issue with the information below?


sarken:
The JIRA issue says this is just tests, but this pull request changes functionality. Can you please update the issue to explain (1) the existing behavior, (2) the new behavior, and (3) how to test it?

zz9pzza:
so the current page has duplicates in it, while the page called asking for duplicates has no duplicates. I could either fix the bug or write the test to test the broken functionality. I asked around and people preferred the bug fixed. I can do it the other way if people prefer ?

sarken:
You don't have to change what this pull request does -- just update the JIRA issue so we know what's being changed and how to test it.

sarken:
I tried updating the issue (and the pull request title, so I stop looking at it when I'm looking for test-only fixes ;)), but I think we need more information on how to manually test this. What steps currently make an external work show up twice (two people bookmarking it? someone bookmarking it and someone else listing it as a related work? something else)? Once we know that, we can take those steps and check that the external work only shows up once.

zz9pzza:
Do you want me to make the test, test the current broken functionality ?

Looking at the tests, two different users bookmarking the same work will cause the issue. I would happily either put that in the issue or reverse the fix and swap the test around.

zz9pzza:
The tests have two different users bookmarking the same work

sarken:
I don't think this is working. I pulled the code and tested it manually by bookmarking the same external URL multiple times, and I still see the exact same results when I go to http://sarkenalt.archiveofourown.org/external_works?show=duplicates and http://sarkenalt.archiveofourown.org/external_works. 

(Additionally, the `paths` file says `external_works?show=duplicates` is the page _without_ duplicates. I think that disagrees with what the controller says, and it's also somewhat unintuitive. I'd expect a page with the parameters `?show=duplicates` to show me duplicates.)

Just in case I was doing something wrong with my manual testing, I tried adding a couple "And show me the page" steps to the automated tests. Both pages show the exact same thing there: two works called External Changes by Zooey Glass. In fact, once I change the step definition from `! page.body.should =~ /#{first_example}.*#{second_example}/m` to the proper `page.body.should_not` the automated tests began to fail.

zz9pzza:
So what we have right now:

http://archiveofourown.org/external_works

Shows duplicates.

http://archiveofourown.org/external_works?show=duplicates gives a 500

But I will revisit it and have another look...

features/works/external_works.feature
```diff
@@ -0,0 +1,17 @@
+@works
+
+Feature: External works are not hosted on the Archive
+
+  Scenario: Can see an external works
```
sarken: I think you could do without the "an" here, since it's external works, plural. Other than that and Hound's comment about an extra space, :+1:

sarken:
Merge conflicts

app/controllers/external_works_controller.rb
```diff
@@ -21,7 +21,8 @@ def fetch
 
   def index
     if params[:show] == 'duplicates'
-      @external_works = ExternalWork.duplicate.order("created_at DESC").paginate(page: params[:page])
+      works = ExternalWork.all.group(:url).having("count(*) >1").order("created_at DESC")
```
tickinginstant: Since the root issue seems to be that the `duplicate` scope for `ExternalWork`s doesn't separate out `group` and `having` the way you did here, why not fix the scope instead of changing this code?

features/works/external_works.feature
```diff
@@ -0,0 +1,17 @@
...
+    Given basic tags
+      And I am logged in as "regular_user"
+      And I bookmark the external work "External Changes"
+      And I bookmark the external work "All changes"
+      And I am logged in as "less_regular_user"
+      And I bookmark the external work "External Changes"
+    When I go to the external works page
+    Then I should see "External Changes"
+      And I should see "All changes"
+    When I go to the external works with only duplicates page
```
tickinginstant: There's supposed to be a link to the duplicates page from the external works page, but only when logged in as admin. So having an admin load the external works page vs. a normal user should have slightly different results. Maybe the scenario should be rewritten to check both? e.g.
```
    When I am logged in as a random user
      And I go to the external works page
    Then I should see "External Changes"
      And I should see "All changes"
      But I should not see "Show duplicates"
    When I am logged in as an admin
      And I go to the external works page
    Then I should see "External Changes"
      And I should see "All changes"
      And I should see "Show duplicates (1)"
    When I follow "Show duplicates"
    Then I should see "External Changes"
      But I should not see "All changes"
```

app/controllers/external_works_controller.rb
zz9pzza: I don't know how to do that, @tickinginstant . I have looked at this and got something similar. However using count/size fails as the count(*) bit makes the sql invalid, unlike the earlier statement. I have worked out that if you use length the whole set is loaded in to memory which can then be counted. Which does work. I suspect size mostly worked as the whole query had been loaded in to memory before making size the same as length.

features/works/external_works.feature
zz9pzza: I think I have done as you wanted ?

app/controllers/external_works_controller.rb
tickinginstant: What I meant by fixing the scope is editing the `ExternalWork` scope to be:
```
scope :duplicate, -> { group(:url).having("count(DISTINCT id) > 1") }
```
For me, that fixes the issue with the SQL error when calling `ExternalWork.duplicate.count`.

Because it's a `group`/`count` combo, it returns a hash instead of a number, which makes the "Show duplicates" button display wrong. But it should work fine with the `will_paginate` gem, which [calls `size` on the hash returned by `count` in order to get the true size](https://github.com/mislav/will_paginate/blob/b6fb7b0db9cacb858ab7092a0447ff9db573b9fd/lib/will_paginate/active_record.rb#L68-L79). Retrieving tens or hundreds of thousands of `ExternalWork` URLs isn't super-efficient, but it's better than calling `length`, which will retrieve more columns from the database, and take more time to stuff the data into `ActiveRecord` objects.

zz9pzza: I had tried that, but I have tried again and I get a test failure: https://travis-ci.org/zz9pzza/otwarchive/jobs/556960242 on this branch: https://github.com/zz9pzza/otwarchive/tree/AO3-4742_2

zz9pzza: I get  Show duplicates ({\"http://zooey-glass.dreamwidth.org\"=>3}) which mirrors what I get in the console.

tickinginstant: Yes, as I said:

>Because it's a group/count combo, it returns a hash instead of a number, which makes the "Show duplicates" button display wrong.

But it's easy enough to fix the same way that the `will_paginate` gem handles it: by calling `count.size` instead of just `count` to figure out the true size. (And as I said in my previous comment, that's more efficient than calling `length` in the `external_works/index` view.)

app/models/external_work.rb
```diff
@@ -8,7 +8,8 @@ class ExternalWork < ApplicationRecord
 
   belongs_to :language
 
-  scope :duplicate, -> { group("url HAVING count(DISTINCT id) > 1") }
+  # Need to use count.size to get the number of records.
```
redsummernight:
```
# Use .duplicate.count.size for the number of records.
```

features/support/paths.rb
```diff
@@ -24,6 +24,10 @@ def path_to(page_name)
       search_people_path
     when /^the bookmarks page$/i
       bookmarks_path
+    when /^the external works page$/i
+      external_works_path
+    when /^the external works with only duplicates page$/i
```
redsummernight: "the external works page with only duplicates"

features/works/external_works.feature
```diff
@@ -0,0 +1,26 @@
+@works
+
+Feature: External works are not hosted on the Archive
```
redsummernight: This should live in features/bookmarks/bookmark_external.feature along with similar tests.

app/views/external_works/index.html.erb
```diff
@@ -2,7 +2,7 @@
 
 <!--navigation-->
 <% if logged_in_as_admin? %>
```
redsummernight: This button is shown only to admins, but the link itself is accessible to all. Should we lock that down too?

sarken: Yeah, definitely. There's probably not much chance of users finding it, but better safe than sorry.

redsummernight: We forgot about this thread, let's update this before merging.

sarken (approved):

PR #2645, opened 2016-12-03
AO3-4744 Remove autocomplete around external bookmarks

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4744

## Purpose

Clean up the external work bookmark code 

## Testing

Confirm that you can still create a book mark for an external work

# Pull Request Checklist

* [ ] Have you added the JIRA issue number as the *first* thing in your pull request title (eg: `AO3-1234 Fix thing`)
* [ ] Have you updated the JIRA issue with the information below?


app/views/bookmarks/_external_work_fields.html.erb
```diff
@@ -3,23 +3,10 @@
   <dl>
     <dt class="required"><%= ew.label :url, ts('URL') + '*' %></dt>
     <dd class="required" title="URL">
-      <% if !params[:url_from_external].blank? %>
-        <%= ew.text_field :url, autocomplete_options('external_work', :autocomplete_token_limit => 1, :value => params[:url_from_external]) %>
-      <% else %>
+      <% if params[:url_from_external].blank? %>		 
         <%= ew.text_field :url, autocomplete_options('external_work', :autocomplete_token_limit => 1) %>
```
sarken: If we're removing the autocomplete, we should also remove all of the `autocomplete_options` bits from this file -- they're used to set HTML attributes the autocomplete JavaScript relies on

app/views/bookmarks/_external_work_fields.html.erb
```diff
@@ -1,25 +1,9 @@
...
-          $j("#bookmark_external_url").blur(function() {
-            $j.ajax({
-              type: "get",
-              data: "external_work_url=" + $j("#bookmark_external_url").val(),
-              dataType: "script",
-              url: "<%= url_for(:controller => "external_works", :action => "fetch") %>"
-            })
-          })
-        <% end %>
+      <%= ew.text_field :url %>
```
sarken: We still want to set the value of the URL field using the params from the bookmarklet, so I think we need to do:

```
<% if !params[:url_from_external].blank? %>
  <%= ew.text_field :url, value: params[:url_from_external] %>
<% else %>
  <%= ew.text_field :url %>
<% end %>
```

sarken:
I'm looking at this again, and I completely misunderstood the issue. /o\ The autocomplete on all the _individual_ fields is just fine and should stay -- it's getting the title, creator, and summary based on the URL that isn't working. 

The reason it's not working is it's not firing at all in 99% of cases -- only on the rare occasion that something blocks the autocomplete JS from loading, but allows our other scripts through. It doesn't fire because (a) `blur` doesn't work with dynamically-inserted fields and (b) it's looking for the wrong ID.

The good news is this is a pretty straightforward fix. In the view file, find this:

`$j("#bookmark_external_url").blur(function() {`

Change it to this:

`$j("#bookmark-form").on("blur", "#bookmark_external_url_autocomplete", function() {`

It will work like a charm.

zz9pzza:
Shall I close this pull and let you re open one with the fix ?

sarken:
Sure, that would be fine! Actually, I'll just hit "Close and comment" as long as I'm here.

PR #2651, opened 2016-12-12
AO3-4747 Confirmation page for purging assignments when JS is disabled

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4747

## Purpose

Adds a confirmation page for purging gift exchange assignments when JavaScript is disabled.

## Testing

Refer to JIRA ticket for instructions.


app/controllers/challenge_assignments_controller.rb
```diff
@@ -181,11 +181,14 @@ def set
     end
   end
 
+  def confirm_purge
```
sarken: Go away.

PR #2652, opened 2016-12-13
AO3-4629 make download menu & chapter index menu sit lower on narrow screens

pupupuri:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4629

## Purpose

the download menu and chapter index menu were positioned at the top of the page on narrow screens; by removing `.chapter .secondary, .download .secondary { top: 3.5em; }` from **08-actions.css**, they sit naturally under their parent elements on all screen sizes. (tested on wide, medium, and narrow screen sizes in chrome, firefox, and safari).

## Testing

please go to a multichapter work and make sure that the chapter index menu sits below the chapter index button on all screen/window sizes, and that the download menu sits below the download button on all screen/window sizes.

### DIFF
--- a/public/stylesheets/site/2.0/08-actions.css
+++ b/public/stylesheets/site/2.0/08-actions.css
@@ -168,10 +168,6 @@ dl.stats + .landmark + .actions {
   float: none;
 }
 
-.chapter .secondary, .download .secondary {
-  top: auto;
-}
-
 .dashboard .landmark {
   clear: right;
 }
--- a/public/stylesheets/site/2.0/08-actions.css
+++ b/public/stylesheets/site/2.0/08-actions.css
@@ -72,7 +72,7 @@ p.submit, input.submit, dd.submit {
   padding: 0;
 }
 
-/*subtypes and modes 
+/*subtypes and modes
 test note: this pagination clear needs to be kept an eye on*/
 
 ol.pagination, div.pagination, ol.year {
@@ -168,10 +168,6 @@ dl.stats + .landmark + .actions {
   float: none;
 }
 
-.chapter .secondary, .download .secondary {
-  top: 3.5em;
-}
-
 .dashboard .landmark {
   clear: right;
 }
@@ -186,13 +182,13 @@ dl.stats + .landmark + .actions {
   padding: 0 0 0.15em 0;
 }
 
-.many p.actions { 
-  float: none; 
-  text-align: right; 
+.many p.actions {
+  float: none;
+  text-align: right;
 }
 
-.concise label.action { 
-  float: left; 
+.concise label.action {
+  float: left;
 }
 
 dd.any label.action {
--- a/public/stylesheets/site/2.0/08-actions.css
+++ b/public/stylesheets/site/2.0/08-actions.css
@@ -168,6 +168,10 @@ dl.stats + .landmark + .actions {
   float: none;
 }
 
+.chapter .secondary, .download .secondary {
+  top: auto;
+}
+
 .dashboard .landmark {
   clear: right;
 }


sarken:
Unfortunately, we end up with the opposite problem here -- on wide screens, the menus end up too low and too far to the right, making it hard to see that they've appeared. Here's a screen shot of the Downloads menu on a 13" screen:

ao3-4629_downloads

My screen is a little too small to show the problem with the Chapter Index in the normal font size, but if I make my browser's font size smaller, you can get the idea:

ao3-4629_chapters

pixelred:
hmm... okay. i'll look for a solution that addresses both of these

pixelred:
so the pop out menu... sits really far to the right because `.secondary` has `right: 2.5em;` but when you remove the right position it sits really far to the left instead. it sits at the very edge of the `.work` container. i'm still trying to figure out why `.secondary` are positioning themselves relative to `.work` and not to their parents `li`

pixelred:
it... oh? maybe cause they're set to `display:block`

pixelred:
okay! so... regarding the issue of the secondary menus sitting too far right on some large screen sizes

if you remove `right: 2.5em` and set them to `display: inline` instead of `display: block` on click, then they position themselves by their parent `<li>`, but... it doesn't look very cute when they're positioned in the middle of the div under their parent. it definitely looks cleanest when they're sitting in the far right corner.



this fix achieves the `make it not sit so far to the right` goal but it's not very pretty... and it requires more tweaking with the `top` property and the addition of `margin-left` so it gets a bit messy

the current commits fix the top positioning on mobile. but i'm not sure what to do about the wide screen too-far-right issue

sarken:
Have you tried using `li.download, li.chapter { position: relative; }` combined with zeroing out the top and bottom positions on `.chapter .secondary, .download .secondary` and setting their widths?

pixelred:
i tried it now and; it positions them well on wide screens, but on mobile it ends up like this

screen shot 2016-12-21 at 20 55 08

sarken:
If you do width: auto; for the menus in mobile, that should probably limit them to being only as wide as the parent element, which would stop them from hanging off the edge of the screen.

But it occurs to me that this isn't good for users with JavaScript disabled -- the menus are always open for them, which means they'd always be covering up other buttons.

pixelred:
the good news is that on mobile the chapter index is a link so the only open menu there is download - and if you add `margin-top: 3em` to it it looks like this

screen shot 2016-12-21 at 21 16 59

objectsforheads:
I had a go at this the other night - maybe this could help?

```
.navigation.work {
  position: relative; /* Set as a parent context */
}

.download {
  display: inline-block;
  /* display: inline fails when the buttons break into 2 lines because inline elements do not force new lines */
  position: relative; /* Set as secondary's closest parent context */
}

.download .secondary {
  display: block;
  min-width: 16.5rem; /* Because .download is our parent context, we need to set the width of the child secondary */
  /*
  The way DOM is rendered, we don't need to specify a top for .secondary to be set at the bottom
  We could be explicit and use top: 100%; but then we need to accommodate the margins because of display: inline-block;
  */
  top: initial;
  right: 0;
}

.chapter .secondary {
  display: block;
  top: initial;
}

/*
At narrow widths, the secondary menus run the risk of running off page
To prevent that, set the secondary menu at the middle of the bar while
ensuring that the secondary menu will shrink to accommodate smaller screens

This code should go into 'media-narrow'
Unless you want to go for a components based stylesheet structure
In which case, keep it here
*/
@media only screen and (max-width: 42em) {
  .download {
    position: initial; /* Unset as a parent context so that secondary is positioned relative to .navigation.work */
  }
  .download .secondary {
    left: 0; /* align to middle */
    right: 0; /* align to middle */
    min-width: initial; /* .download is no longer our parent context; we can unset min-width */
    max-width: 16.5rem; /* ~ the width in rems so that it'll scale even if we zoom in/out */
  }

  .chapter .secondary {
    left: 0; /* At lower resolutions, the menu can run off screen - set a left value to prevent this */
    right: 0; /* can also set a right value to center it */
  }
}

/*
Some adjustments to the chapter dropdown resolving some alignment issues in FF and some sizing @ small resolutions
This actually isn't related to this particular issue, but I thought I'd note it while I was here
*/

/* this is more selective than I'd like, but I can't recall if [name] is supported in CSS2 */
.chapter .secondary select {
  vertical-align: middle; /* FF is bumping this out of alignment w/the other elements. This is not a problem on Chrome */
  /*
  at lower resolutions, the element can stretch outside of its parent container,
  making the dropdown arrow disappear from view. I'm not going to presume the obviousness of
  this being a dropdown so this fixes it to make the arrow always visible

  The chapter title, if too long, will be cut off in either scenario - to accommodate this
  I would recommend a title attribute so that it shows up on hover for non-screen-readers
  touch devices are out of luck either way as well

  I can't recall a fix for making select options break onto a second line at the moment
  Alternatively, we could just replace the entire select input with a custom solution
  */
  max-width: 100%;
}
```

PR #2653, opened 2016-12-13
AO3-4617 Change login error message to include "reset password" link

pixelred:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4617

## Purpose

the error message a user sees when they enter the wrong login information has been changed to include a link to the new password page

## Testing

please make sure the old error message 

_The password or user name you entered doesn't match our records. Please try again or follow the 'Forgot password?' link below. If you still can't log in, please visit Problems When Logging In for help._ 

has been changed to

 _The password or user name you entered doesn't match our records. Please try again or <a href="http://archiveofourown.org/passwords/new">reset your password</a>. If you still can't log in, please visit <a href="http://archiveofourown.org/admin_posts/1277">Problems When Logging In</a> for help._

### DIFF
--- a/app/controllers/user_sessions_controller.rb
+++ b/app/controllers/user_sessions_controller.rb
@@ -44,13 +44,13 @@ class UserSessionsController < ApplicationController
             if @user_session.being_brute_force_protected?
               message = ts("Your account has been locked for 5 minutes due to too many failed login attempts.")
             else
-              message = ts("The password or user name you entered doesn't match our records. Please try again or <a href=\"http://archiveofourown.org/passwords/new\">reset your password</a>.".html_safe)
+              message = ts("The password or user name you entered doesn't match our records. Please try again or <a href=\"#{new_password_path}\">reset your password</a>. If you still can't log in, please visit <a href=\"#{admin_posts_path + '/1277'}\">Problems When Logging In</a> for help.".html_safe)
             end
           else
             message = ts("You'll need to activate your account before you can log in. Please check your email or contact support.")
           end
         else
-          message = ts("The password or user name you entered doesn't match our records. Please try again or <a href=\"http://archiveofourown.org/passwords/new\">reset your password</a>.".html_safe)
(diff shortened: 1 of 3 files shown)


app/controllers/user_sessions_controller.rb
```diff
@@ -41,17 +41,17 @@ def create
               message = ts("The password you entered has expired. Please click the 'Reset password' link below.")
             end
           elsif user.active?
-            if @user_session.being_brute_force_protected? 
-           
+            if @user_session.being_brute_force_protected?
```
pixelred: what 😥

app/controllers/user_sessions_controller.rb
```diff
@@ -41,17 +41,16 @@ def create
...
               message = ts("Your account has been locked for 5 minutes due to too many failed login attempts.")
             else
-              message = ts("The password or user name you entered doesn't match our records. Please try again or follow the 'Forgot password?' link below.  If you still can't log in, please visit <a href=\"http://archiveofourown.org/admin_posts/1277\">Problems When Logging In</a> for help.".html_safe)
+              message = ts("The password or user name you entered doesn't match our records. Please try again or <a href=\"http://archiveofourown.org/passwords/new\">reset your password</a>.".html_safe)
```
sarken: Whoops, sorry, I think the issue was unclear -- we just want to change the second sentence, but the third (about problems when logging in) can stay.

Also, we're going to be changing over to https sooner rather than later (fingers crossed!) so we don't want to include the `http://` bit in the URL. A better way to do this might be: `<a href=\"#{new_password_path}\">reset you password</a>` If you could change the other link to work like this, too, that would be great! It would probably be something like `<a href=\"#{admin_posts_path + "/1277"}\">link text</a>` (I haven't tried it and am just guessing at where all those quotation marks, slashes, and braces go, so you'll have to fiddle with it a bit).

pixelred: ahh i see. no worries. i'm on it :D

pixelred:
done :v: tested it and the new links work in my local environment

sarken:
Yay! This is ready to merge, then. (Also, just a heads up that you don't need to fix spacing and things in the /docs files, and if you have something that changes those files automatically, you don't need to commit the changes and include them in your pull request.)

pixelred:
ooh okay! thank you for the heads up :' )

PR #2654, opened 2016-12-15
AO3-4762 added test for controllers/invite_requests_controller (admin deleting someone from queue)

pixelred:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4762

### DIFF
--- a/features/other_a/invite_queue.feature
+++ b/features/other_a/invite_queue.feature
@@ -38,10 +38,10 @@ Feature: Invite queue management
   Scenario: An admin can delete people from the queue
     Given an invitation request for "invitee@example.org"
       And I am logged in as an admin
-      When I go to the manage invite queue page
+    When I go to the manage invite queue page
       And I follow "Delete"
     Then I should see "Request was removed from the queue."
-    And I should be on the manage invite queue page
+      And I should be on the manage invite queue page
 
   Scenario: Visitors can join the queue and check status when invitations are required and the queue is enabled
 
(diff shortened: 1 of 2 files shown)


features/other_a/invite_queue.feature
```diff
@@ -35,6 +35,13 @@ Feature: Invite queue management
     When I follow "Get an Invitation"
     Then I should see "Request an Invitation"
 
+  Scenario: An admin can delete people from the queue
+    Given an invitation request for "invitee@example.org"
+      And I am logged in as an admin
+      And I go to the manage invite queue page
```
sarken: Just a tiny nitpick here! Cucumbers usually have Given steps for setting up background, When steps for performing actions, and Then steps for describing results (https://cucumber.io/docs/reference#steps). It ultimately doesn't change how anything works, but for style, I'd suggest making this `When I go to the manage invite queue page` instead.

features/other_a/invite_queue.feature
```diff
@@ -35,6 +35,13 @@ Feature: Invite queue management
...
+      And I follow "Delete"
+    Then I should see "Request was removed from the queue."
```
sarken: This is good! I'd just suggest adding one more thing to make sure we're ending up on the right page: `And I should be on the manage invite queue page` Doing that should cover the redirect on line 53 of the controller.

pixelred:
that makes sense :v: thank you for the help!

how does it look?

sarken:
The steps look good! But one more style comment: Given, When, and Then steps should all be indented the same amount, and the And steps should be indented two spaces deeper. (It might be clearer if you look at the other scenarios in this file to see what I mean--I'm on my phone and can't illustrate at the moment.)

pixelred:
ahh you're right. i see it now. sorry about that /o\

pixelred:
oh; i also forgot the spacing between the scenario and the rest... i'll do that as soon as i'm on my computer

pixelred:
whoops... thought i added the space this morning...

PR #2157, opened 2015-08-23
AO3-3880 Downloads by calibre - updated

shalott:
UGH. The previous branch/pullreq became horribly confused after an ill-advised attempt to update via github desktop for the first time, where the pullreq wouldn't update with the resolution of the merge-preventing conflicts. Sob. I have created this new branch to resolve the conflicts and move forward. Below is the same notes from the original pullreq.  

RELEASE NOTES: Previously we've been creating mobi and epub downloads using a combination of home-grown and third party tools. We're now using the popular program calibre to generate these files. This should mean better compatibility and better metadata in these downloads, and will hopefully clear up a wide variety of issues with downloads. In addition, we have fixed issues where downloads didn't update even after a work was changed, and made some improvements in error handling and in the organization of the code that handles downloads. 

---

THIS REQUIRES CALIBRE TO BE INSTALLED somewhere in the PATH or else it will not work!

https://code.google.com/p/otwarchive/issues/detail?id=3888

Switch to using calibre for mobi/epub generation. 
Refactor downloadable functionality out of Work into Downloadable module
Switched from Iconv to ActiveSupport::Inflector.transliterate(string) for converting to ASCII for filenames
Added/edited lots of tests. 
Cleaned up deprecated calls. 

UPDATE:
Secured system call properly
Moved generation of downloads into Downloadable
Made sure downloads dir gets cleaned up appropriately (on work/owner changes). Fixes:
- https://code.google.com/p/otwarchive/issues/detail?id=3440
- https://code.google.com/p/otwarchive/issues/detail?id=3053
- https://code.google.com/p/otwarchive/issues/detail?id=3054

BUT WAIT THERE'S MORE:
Closed hole that allowed unrevealed works to be downloaded
https://code.google.com/p/otwarchive/issues/detail?id=3055

Removed html entities from download filenames entirely instead of just stripping the & and ;
https://code.google.com/p/otwarchive/issues/detail?id=2645

Fixed issue with not providing error message when the download file wasn't rendered yet
https://code.google.com/p/otwarchive/issues/detail?id=3024

NOW ADDED:
Downloads for downstream series works now get expired
https://code.google.com/p/otwarchive/issues/detail?id=2319

Also fixes 
issues with mobi downloads for dash at start of title 
https://code.google.com/p/otwarchive/issues/detail?id=3908

PR #2660, opened 2016-12-17
Testing downloads

cesy:
No JIRA issue - do you want me to create one?

## Purpose

Improves the download test slightly. The rest of the work should wait until after the major download upgrade, but this is useful progress that will help whoever picks it up then.

## Testing

None required.

## References

#2157 

## Credit

Cesy, she

### DIFF
--- a/features/works/work_download.feature
+++ b/features/works/work_download.feature
@@ -4,7 +4,7 @@
 
 Feature: Download a work
   @wip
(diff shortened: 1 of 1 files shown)


sarken:
Just a heads-up that we already have https://github.com/otwcode/otwarchive/pull/2605, which includes some refactoring, too.

PR #2663, opened 2016-12-18
AO3-4768 Extended tests for controllers/series_controller.rb

pixelred:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4768

## Purpose

test for title in pseud's series index page
test for creating co-authored series and removing self as co-author
organise/minimise # of lines in series_lock and series_order and series


features/other_b/series.feature
```diff
@@ -209,21 +213,50 @@ Feature: Create and Edit Series
...
-    Given I am logged in as "whoever" with password "whatever"  
+    Given I am logged in as "whoever" with password "whatever"
       And I add the work "grumble" to "32" series "penguins"
     When I go to whoever's series page
     Then I should see "penguins30"
     When I follow "Next"
     Then I should see "penguins0"
-  
+
+  Scenario: Removing self as co-author from co-authored series
```
sarken: This is good! However, I think we can shorten it a bit (and make it faster) by taking advantage of some of our more flexible step definitions and only checking the most necessary bits:

```
    Given I am logged in as "sun"
      And I set up the draft "Sweetie Bell" as part of a series "Ponies"
      And I add the co-author "moon"
      And I post the work without preview
    When I view the series "Ponies"
      And I follow "Remove Me As Author"
    Then I should see "You have been removed as an author from the series and its works."
      And I should see "by moon"
      And I should not see "by moon, sun"
    When I go to my works page
    Then I should not see "Sweetie Bell"
```

features/other_b/series.feature
```diff
@@ -135,6 +135,10 @@ Feature: Create and Edit Series
       And I view the series "Ponies"
     Then I should see "Sweetie Belle"
       And I should see "Starsong"
+    When I follow "Pointless Pseud"
```
sarken: The title check looks good! However, this is already a really, really long scenario (which is totally not your fault!) so instead of making it longer (which makes it harder to figure out what's broken if it starts to fail), I'd suggest moving the page title bit to its own scenario. 

The set up for creating a series with a pseud is actually a lot less involved than this existing scenario suggest. It would just be:
```
  Scenario: A pseud's series page contains the pseud in the page title
      And I add the pseud "Pointless Pseud"
      And I set up the draft "Sweetie Belle" as part of a series "Ponies"
      And I select "Pointless Pseud" from "work_author_attributes_ids_"
      And I press "Post"
```
followed by your When and Then steps. (You can omit the `I should see "Ponies"` and just go for checking the page title.)

If you want to go back through this "Three ways to add a work to a series: a user with more than one pseud" and try to shorten a few of the parts where it posts a work, that might be a good way to get more familiar with all the options we have for creating works in our Cucumber tests. However, it's totally optional! 

(Also, I'm done now -- sorry about the trigger-happy comment finger. ;))

pixelred:
okay :D i'll go through and minimise it where i can ✌️

pixelred:
i'm almost done with the Scenario: Post a series with a restricted work, then add a draft, then make the draft public and post it in series_lock :v:

features/other_b/series.feature
```diff
@@ -4,59 +4,53 @@ Feature: Create and Edit Series
...
-      And I select "Not Rated" from "Rating"
-      And I check "Choose Not To Use Archive Warnings"
-      And I fill in "Fandoms" with "My Little Pony"
-      And I check "series-options-show"
-      And I fill in "work_series_attributes_title" with "Ponies"
-      And I fill in "Work Title" with "Sweetie Belle"
-      And I fill in "content" with "First little pony is all alone"
+  Scenario: Creator adds a work to a new series when the work is first posted
+    Given I am logged in as "author"
+      And I set up the draft "Sweetie Belle"
```
zz9pzza: This can be replaced by

I set up the draft "Sweetie Belle" as part of a series "Ponies"

And the following line removed

features/other_b/series_lock.feature
```diff
@@ -4,84 +4,60 @@ Feature: Locked and partially locked series
...
+    Then I should see "Antiholidays"
+    When I am logged out
+     And I go to fandomer's series page
+    Then I should not see "Antiholidays"
+    When I am logged in as "reccer"
+      And I go to fandomer's series page
+    Then I should see "Antiholidays"
+    When I view the series "Antiholidays"
+    Then I should see "Humbug"
+    When I am logged out
```
sarken: Instead of `When I am logged out/And I am logged in as "fandomer"`, we can just do `When I am logged in as "fandomer"` -- the log in step takes care of logging out, too. :D

features/other_b/series_lock.feature
```diff
@@ -4,84 +4,60 @@ Feature: Locked and partially locked series
...
+      And I press "Preview"
+    Then I should see "Draft was successfully created."
+      And I should see "Part 2 of the Antiholidays series"
+    When I view the series "Antiholidays"
+    Then I should see "Works: 1"
+      And I should not see "Antivalentine"
+    When I view the work "Humbug"
+    Then I should not see "»" within "dd"
+    When I edit the work "Antivalentine"
+      And I uncheck "work_restricted"
```
sarken: This could be `And I unlock the work` (it's the same thing, but if we ever change the id of the checkbox, it makes it so there are fewer places we need to update the tests).

features/other_b/series.feature
```diff
@@ -4,59 +4,53 @@ Feature: Create and Edit Series
...
+    When I fill in "work_series_attributes_title" with "Ponies"
       And I press "Preview"
```
zz9pzza: Does pressing Preview and saving a preview gain any coverage ?

features/other_b/series.feature
```diff
@@ -4,59 +4,53 @@ Feature: Create and Edit Series
...
-      And I check "Choose Not To Use Archive Warnings"
-      And I fill in "Fandoms" with "My Little Pony"
-      And I select "Ponies" from "work_series_attributes_id"
-      And I fill in "Work Title" with "Starsong"
-      And I fill in "content" with "Second little pony want to make friends"
+
+  Scenario: Creator adds a work to an existing series when the work is first posted
+    Given I am logged in as "author"
+      And I post the work "Sweetie Belle" as part of a series "Ponies"
+      And I set up the draft "Starsong"
```
zz9pzza: Again you could use 

I set up the draft "Sweetie Belle" as part of a series "Ponies"

or 

I post "Sweetie Belle" as part of a series "Ponies"

Should cover the same ground ?

features/other_b/series.feature
```diff
@@ -78,112 +76,79 @@ Feature: Create and Edit Series
...
-      And I fill in "Fandoms" with "My Little Pony"
-      And I select "Pointless Pseud" from "work_author_attributes_ids_"
-      And I check "series-options-show"
-      And I fill in "work_series_attributes_title" with "Ponies"
-      And I fill in "Work Title" with "Sweetie Belle"
-      And I fill in "content" with "First little pony is all alone"
+  Scenario: Creator with multiple pseuds adds a work to a new series when the work is first posted
+    Given I am logged in as "author"
+      And I add the pseud "Pointless Pseud"
+      And I set up the draft "Sweetie Belle" using the pseud "Pointless Pseud"
```
zz9pzza: Again I can't see why this is not 

And I set up the draft "Sweetie Belle" using the pseud "Pointless Pseud"  as part of a series "Ponies"

features/other_b/series.feature
```diff
@@ -78,112 +76,79 @@ Feature: Create and Edit Series
...
-    When I go to the new work page
-      And I select "Not Rated" from "Rating"
-      And I check "Choose Not To Use Archive Warnings"
-      And I fill in "Fandoms" with "My Little Pony"
+
+  Scenario: Creator with multiple pseuds adds a work to an existing series when the work is first posted
+    Given I am logged in as "author"
+      And I add the pseud "Pointless Pseud"
+      And I post the work "Sweetie Belle" as part of a series "Ponies" using the pseud "Pointless Pseud"
+    When I set up the draft "Starsong" using the pseud "Pointless Pseud"
```
zz9pzza: And again ?

features/other_b/series_lock.feature
pixelred: that makes sense :' )

features/other_b/series_lock.feature
zz9pzza: This step is not needed as the following one includes it.

features/step_definitions/work_steps.rb
```diff
@@ -57,29 +56,30 @@
...
   end
   Work.tire.index.refresh
 end
 
-# Again, same regexp, it just creates a draft and not a posted 
+# Again, same regexp, it just creates a draft and not a posted
 # To test posting after preview, use: Given the draft "Foo"
 # Then use: When I post the work "Foo"
-# and the above step 
+# and the above step
```
zz9pzza: Can you update the regex to match the earlier one ?

features/other_b/series.feature
sarken: tbh, I don't think it does. I left it because it was already there, but I would be in favor of taking it out of all the scenarios and just straight up posting. If we want to test how series information looks when previewing, we can make a dedicated scenario for it instead of repeating it 57 times throughout multiple scenarios.

features/other_b/series.feature
sarken: I'm torn on this. It _would_ be simpler, but I think the point of this scenario is to check that the series is added when we type a name in the field, so it _might_ be a good idea to leave it so it's explicit that we're doing that. (The scenario after this is about testing that we can use the select field.) `I set up the draft "Sweetie Belle" as part of a series "Ponies"` does the exact same thing, but less obviously.

zz9pzza: If you feel that it is more readable this way I am happy to defer :)

sarken: I think it's hideous, honestly, especially since it uses the field id rather than the label (which I did because the label makes 0 sense out of context). :P 

Maybe the thing to do is change the Scenario description to say "Creator manually enters a series name to add a work to a new series when the work is first posted" and the other one can me "Creator selects an existing series name to add a work to an existing series when the work is first posted". That way, we can use the "as part of a series" bit.

Whatever @pixelred wants to do is fine by me!

features/step_definitions/work_steps.rb
pixelred: sorry; could you clarify what you meant here?

zz9pzza: This is a bit of feature creep however.

The comment says "Again, same regexp, it just creates a draft and not a posted"

But the regex only takes "title, fandom, freeform, category, collection, recipient" while the other one takes "title, fandom, character, character2, freeform, freeform2, category, collection, recipient, series, relationship, pseud" 

Given that I was probably the last person to extend the regex I would be happy to do it however you are working here right now. Does that make sense ?

pixelred: ahh okay! thank you for clarifying. i'll go ahead and make this change

pixelred:
passed :D

PR #2665, opened 2016-12-19
AO3-4753 – AdminSettingController Strong Params

kamilogorek:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4753

### DIFF
--- a/app/controllers/admin/settings_controller.rb
+++ b/app/controllers/admin/settings_controller.rb
@@ -1,5 +1,4 @@
 class Admin::SettingsController < ApplicationController
-
   before_filter :admin_only
 
   def index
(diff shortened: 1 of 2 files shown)


app/controllers/admin/settings_controller.rb
```diff
@@ -11,12 +10,23 @@ def index
...
   end
 
+  private
+  def admin_setting_params
+    params.require(:admin_setting).permit(
+      :account_creation_enabled, :invite_from_queue_enabled, :invite_from_queue_number,
+      :invite_from_queue_frequency, :days_to_purge_unactivated, :last_updated_by,
+      :invite_from_queue_at, :suspend_filter_counts, :suspend_filter_counts_at,
+      :enable_test_caching, :cache_expiration, :tag_wrangling_off, :guest_downloading_off,
+      :default_skin_id, :stats_updated_at, :disable_filtering, :request_invite_enabled,
```
zz9pzza: default_skin_id is set via the skin object ( and you can see it is not in the form that matches this controller ).

stats_updated_at show when a process was last run and should not be settable from this controller.

PR #2668, opened 2016-12-19
AO3-4757 – Remove redundant fallback for new user preferences

kamilogorek:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4757

### DIFF
--- a/app/controllers/preferences_controller.rb
+++ b/app/controllers/preferences_controller.rb
@@ -3,16 +3,15 @@ class PreferencesController < ApplicationController
   before_filter :check_ownership
   skip_before_filter :store_location
 
-  
   # Ensure that the current user is authorized to view and change this information
   def load_user
     @user = User.find_by_login(params[:user_id])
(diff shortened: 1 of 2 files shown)


kamilogorek:
One thing regarding this change. User has defaults set on creation time in `before_create` callback, including `preference`. Because of this, there's no way that `||` will be ever triggered.

I checked `git blame` to make sure that it hasn't been added before the model setup and it wasn't.

```
ab595706 (ambtus         2008-05-17 18:38:05 +0000  70)   before_create :create_default_associateds
ab595706 (ambtus         2008-05-17 18:38:05 +0000 306)     self.preference = Preference.new
```
```
249d732e (ambtus         2010-03-02 03:26:44 +0000 14)     @preference = @user.preference || Preference.create(:user_id => @user.id)
```

Therefore it's very safe to assume that it won't ever break. To be 101% sure, you can query prod database `SELECT * FROM users WHERE preference = null`

PR #2670, opened 2016-12-19
AO3-4759 – FeedbacksController Strong Params

kamilogorek:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4759

### DIFF
--- a/app/controllers/feedbacks_controller.rb
+++ b/app/controllers/feedbacks_controller.rb
@@ -36,7 +36,7 @@ class FeedbacksController < ApplicationController
 
   def feedback_params
     params.require(:feedback).permit(
-      :comment, :email, :summary, :user_agent, :category, :approved,
+      :comment, :email, :summary, :user_agent,
       :ip_address, :username, :language
     )
   end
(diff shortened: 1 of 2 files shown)


app/controllers/feedbacks_controller.rb
```diff
@@ -32,4 +32,13 @@ def load_support_languages
     @support_languages = Language.where(support_available: true).order(:name)
   end
 
+  private
+
+  def feedback_params
+    params.require(:feedback).permit(
+      :comment, :email, :summary, :user_agent, :category, :approved,
```
zz9pzza: I can't see category or approved in the form and therefore should they be let in ?

sarken: I'm not sure what approved is or was for, but category was from the old version of the Support form, and we've left the column in the database in case we need the old data.

sarken: Looking over the model code, `approved` is where we save the result of our spam check -- so it's definitely not something we want to let the user submit. It seems like it might be helpful to check the column names against the form fields that are in the view, although it probably is easier for our reviewers to spot things that _shouldn't_ be allowed than things that are missing.

PR #2671, opened 2016-12-19
AO3-4763 – PseudsController Strong Params

kamilogorek:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4763

### DIFF
--- a/app/controllers/pseuds_controller.rb
+++ b/app/controllers/pseuds_controller.rb
@@ -151,8 +151,8 @@ class PseudsController < ApplicationController
 
   def pseud_params
     params.require(:pseud).permit(
-      :user_id, :name, :description, :is_default, :icon_file_name, :icon_content_type,
-      :icon_file_size, :icon_updated_at, :icon_alt_text, :delta, :icon_comment_text
+      :user_id, :name, :description, :is_default, :icon, :icon_file_name, :icon_content_type,
+      :icon_file_size, :icon_alt_text, :icon_comment_text, :delta
     )
   end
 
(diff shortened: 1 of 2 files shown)


kamilogorek:
One of the features that was failing is in `features/collections/icon.feature` and not `features/other_a/pseud*`.

app/controllers/pseuds_controller.rb
```diff
@@ -146,4 +146,14 @@ def destroy
 
     redirect_to(user_pseuds_url(@user))
   end
+
+  private
+
+  def pseud_params
+    params.require(:pseud).permit(
+      :user_id, :name, :description, :is_default, :icon, :delete_icon, :icon_file_name,
```
zz9pzza: I am not convinced that the following need to be allowed though:

user_id,  :icon_file_name, :icon_content_type, :icon_file_size, :icon_alt_text, :delta

sarken: It does seem to work without `:icon_alt_text`, both in the sense that the pseud is created/updated *and* the specified alt text is saved, but I'm a little confused _why_ it works. Do you know, james? (I checked and taking out something like `:name` does cause a problem.)

zz9pzza: All I commented on is that it is not in the form :)

sarken: But it is -- right here: https://github.com/otwcode/otwarchive/blob/master/app/views/pseuds/_pseuds_form.html.erb#L48

zz9pzza: You are completely correct. I must of been thinking about the icon_file_name. sorry ( have to look at how it works now )

PR #2672, opened 2016-12-19
AO3-4764 – CollectionsController Strong Params

kamilogorek:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4764

### DIFF
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -169,23 +169,16 @@ class CollectionsController < ApplicationController
 
   def collection_params
     params.require(:collection).permit(
-      :name, :title, :email, :header_image_url, :description, :parent_id,
-      :parent_name, :challenge_id, :challenge_type, :icon, :icon_file_name,
-      :icon_content_type, :icon_file_size, :icon_alt_text, :icon_comment_text,
+      :name, :title, :email, :header_image_url, :description,
+      :parent_name, :challenge_type, :icon, :icon_file_name,
+      :icon_alt_text, :icon_comment_text,
       collection_profile_attributes: [
-        :id, :collection_id, :intro, :faq, :rules,
+        :id, :intro, :faq, :rules,
         :gift_notification, :assignment_notification
       ],
       collection_preference_attributes: [
-        :id, :collection_id, :moderated, :closed, :unrevealed, :anonymous,
+        :id, :moderated, :closed, :unrevealed, :anonymous,
         :gift_exchange, :show_random, :prompt_meme, :email_notify
-      ],
-      collection_items_attributes: [
-        :id, :collection_id, :item_id, :item_type, :user_approval_status,
-        :collection_approval_status, :anonymous, :unrevealed, :_destroy
-      ],
-      collection_participants_attributes: [
-        :id, :collection_id, :pseud_id, :participant_role, :_destroy
       ]
     )
   end
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -1,8 +1,8 @@
 class CollectionsController < ApplicationController
 
-  before_filter :users_only, :only => [:new, :edit, :create, :update]
-  before_filter :load_collection_from_id, :only => [:show, :edit, :update, :destroy, :confirm_delete]
-  before_filter :collection_owners_only, :only => [:edit, :update, :destroy, :confirm_delete]
+  before_filter :users_only, only: [:new, :edit, :create, :update]
+  before_filter :load_collection_from_id, only: [:show, :edit, :update, :destroy, :confirm_delete]
+  before_filter :collection_owners_only, only: [:edit, :update, :destroy, :confirm_delete]
   before_filter :check_user_status, only: [:new, :create, :edit, :update, :destroy]
   before_filter :validate_challenge_type
   cache_sweeper :collection_sweeper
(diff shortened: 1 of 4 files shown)


app/models/collection.rb
```diff
@@ -116,36 +114,36 @@ def parent_is_allowed
...
 
-  validates_presence_of :name, :message => ts("Please enter a name for your collection.")
-  validates_uniqueness_of :name, :case_sensitive => false, :message => ts('Sorry, that name is already taken. Try again, please!')
+  validates_presence_of :name, message: ts("Please enter a name for your collection.")
+  validates_uniqueness_of :name, case_sensitive: false, message: ts('Sorry, that name is already taken. Try again, please!')
   validates_length_of :name,
-    :minimum => ArchiveConfig.TITLE_MIN,
-    :too_short=> ts("must be at least %{min} characters long.", :min => ArchiveConfig.TITLE_MIN)
+    minimum: ArchiveConfig.TITLE_MIN,
+    too_shor: ts("must be at least %{min} characters long.", min: ArchiveConfig.TITLE_MIN)
```
sarken: Just wanted to point out that there are a couple places in here where it got changed from `too_short` to `too_shor` without the t, or from `too_long` to `too_lon` with the g missing.

app/controllers/collections_controller.rb
```diff
@@ -165,4 +165,22 @@ def destroy
     redirect_to(collections_url)
   end
 
+  private
+
+  def collection_params
+    params.require(:collection).permit(
+      :name, :title, :email, :header_image_url, :description,
```
sarken: @zz9pzza I think we can take out `:icon_file_name`. I experimented, and if we allow it, the user can add an `icon_file_name` field to the form and set whatever name they want, but if we take it off this list of allowed parameters and they try to do that, it just ignores what the user has entered and takes the filename from the icon. I'm pretty sure that's exactly what's supposed to happen with strong parameters.

sarken: Ooh! Sorry, one more thing -- we need to add `:delete_icon` or else we can't delete an icon that's been added to a collection.

features/collections/icon.feature
```diff
@@ -17,8 +17,10 @@ Feature: User icons
     And I press "Update"
   Then I should see the image "alt" text "Some test description"
 
-  Scenario: Add an icon to a collection
+  Scenario: Add and remove an icon to a collection
```
zz9pzza: It is worth noting that the test is useful as it is as it fails on an exception if we don't have :delete_icon

sarken: Yeah, that's definitely true -- where I get conflicted about its ~good-ness is that it will never tell us if we break deleting an icon in any other possible way, so it's giving us a false sense of security. /o\ If you'd prefer not to fight with it, that's fine -- I don't mind marking this ready to merge as it is. But if you could make an issue for improving the icon tests, that would be good, and I'll write the tests tomorrow.

sarken: 👍 Good luck!

features/step_definitions/icon_steps.rb
```diff
@@ -10,17 +10,35 @@
...
   step "I am editing a pseud"
   step %{I attach the file "test/fixtures/icon.gif" to "icon"}
-    step %{I press "Update"}
+  step %{I press "Update"}
 end
 
 ### WHEN
 
-When /^I add an icon to the collection$/ do
+When /^I add an icon to the collection "([^"]*)"$/ do |name|
```
sarken: \o/ and :tada: and :cake: for making it so we can specify the collection!

However, I'd suggest changing the name of this variable to `title`, since it's the collection title, not the collection name, that we use for `"Whatever" collection's page`.

features/step_definitions/icon_steps.rb
```diff
@@ -10,17 +10,35 @@
...
-    step %{I follow "Settings"}
-    step %{I attach the file "test/fixtures/icon.gif" to "collection_icon"}
-    step %{I press "Update"}
+  step %{I am on "#{name}" collection's page}
+  step %{I follow "Settings"}
+  step %{I attach the file "test/fixtures/icon.gif" to "collection_icon"}
+  step %{I press "Update"}
+end
+
+When /^I delete the icon from the collection "([^"]*)"$/ do |name|
```
sarken: Same thing -- I'd suggest changing the name of this variable to `title`, since it's the collection title, not the collection name, that we use for `"Whatever" collection's page`.

features/step_definitions/icon_steps.rb
```diff
@@ -10,17 +10,35 @@
...
+  step %{I am logged in as "moderator"}
+  step %{I am on "#{name}" collection's page}
+  step %{I follow "Settings"}
+  check("collection_delete_icon")
+  step %{I press "Update"}
+end
+
+Then /^the collection "([^"]*)" has an icon$/ do |name|
```
sarken: Since this is a `then` step and we're checking whether an expectation is met rather than performing an action, it would be better to phrase this as `the collection "([^"]*)" should have an icon`.  If it screams bloody murder about an ambiguous match due to a model-related pickle step, `the "([^"]*)" collection should have an icon` will probably work.

Also, I'd suggest looking up the collection by title rather than name, so it's in line with most of other collection-related steps. (We do `find_by_name` in 5 places, but `find_by_title` in 53.)

features/step_definitions/icon_steps.rb
```diff
@@ -10,17 +10,35 @@
...
+  collection = Collection.find_by_name!(name) 
+  assert !collection.icon_file_name.blank?
+end
+
+Then /^the collection "([^"]*)" has no icon$/ do |name|
```
sarken: Same as above, basically! Should be `the collection "([^"]*)" should not have an icon` (or `the "([^"]*)" collection should not have an icon`) and `Collection.find_by_title` rather than `name`.

PR #2674, opened 2016-12-20
AO3-4785 fix test where user's profile wasn't being destroyed properly

katieby:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4785

## Purpose

Fix tests for profile_controller to get 100% coverage.

## Testing

run `bundle exec rspec spec/controllers/profile_controller_spec.rb`

## Credit

cresenne, she/her

### DIFF
--- a/spec/controllers/profile_controller_spec.rb
+++ b/spec/controllers/profile_controller_spec.rb
@@ -3,7 +3,7 @@ require 'spec_helper'
 describe ProfileController do
   describe 'show' do
     it 'should be an error for a non existent user' do
-      get :show, user: 999_999_999_999
+      get :show, user_id: 999_999_999_999
 
       expect(response).to redirect_to(root_path)
       expect(flash[:error]).to eq "Sorry, there's no user by that name."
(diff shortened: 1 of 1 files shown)


spec/controllers/profile_controller_spec.rb
```diff
@@ -12,7 +12,8 @@
     it 'should create a new profile if one does not exist' do
       @user = FactoryGirl.create(:user)
       @user.profile.destroy
-      get :show, user: @user.id
+      @user.reload
+      get :show, user_id: @user
```
zz9pzza: I am aware we are just about to remove the bit that creates a new preference if the user doesn't have one.

https://github.com/otwcode/otwarchive/pull/2668/files#diff-2a02e520fe278dced5dea79dbb1c0ef4R14

What should we do instead ?

sarken:
Woohoo! I ran these through Codeship and Coveralls really quick to confirm the line was now covered, so this is ready to go -- thank you!

spec/controllers/profile_controller_spec.rb
sarken: (Just noting that this was a bit of a mix-up -- with both preferences and profile, we had our P-words confused. All is good!)

PR #2678, opened 2016-12-22
AO3-4722 WorksController 1 - import errors and search param cleansing

ariana-paris:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4722

## Purpose

This is the first of probably a short series of pull requests aiming to increase the WorksController coverage using Rspec instead of Cucumber. The new tests cover:

- The before_filter for work search parameter cleansing - I found a couple of potential bugs or omissions there, though I don't think they are currently affecting users so I left them alone to avoid scope creep
- The errors returned when performing imports

Creating some of the new tests uncovered some problems in the API tests (they were making all users in all tests into archivists) and a bug in the StoryParser itself, though one which would only ever affect the tests (no nil check on pseuds, which breaks if there is no current user – but in real life, you can't reach that code without a logged in user). These had to be fixed in order to have sensible WorksController tests, so they are also included in this pull request.

## Testing

These are mainly test changes, as well as one minor refactor in the WorksController and in the StoryParser. Those are both exercised by existing Cucumber tests that I did not touch, so their functionality hasn't changed.

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -517,14 +517,41 @@ class WorksController < ApplicationController
 
   # POST /works/import
   def import
+    import_errors = ""
+    # check to make sure we have some urls to work with
     @urls = params[:urls].split
 
-    import_errors = check_import_errors(params)
+    if @urls.nil? || @urls.empty?
+      import_errors = ts("Did you want to enter a URL?")
+    end
+
+    # is external author information entered when import for others is not checked?
+    if (params[:external_author_name].present? || params[:external_author_email].present?) && !params[:importing_for_others]
+      import_errors = ts("You have entered an external author name or e-mail address but did not select \"Import for others.\" " +
+                           "Please select the \"Import for others\" option or remove the external author information to continue.")
+    end
+
+    # is this an archivist importing?
+    if params[:importing_for_others] && !current_user.archivist
+      import_errors = ts("You may not import stories by other users unless you are an approved archivist.")
+    end
+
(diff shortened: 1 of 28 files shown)


app/controllers/works_controller.rb
```diff
@@ -517,45 +517,52 @@ def destroy
...
-      render(:new_import) && return
+    if @urls.nil? || @urls.empty?
+      import_errors = ts("Did you want to enter a URL?")
     end
 
     # is external author information entered when import for others is not checked?
     if (params[:external_author_name].present? || params[:external_author_email].present?) && !params[:importing_for_others]
-      flash.now[:error] = ts('You have entered an external author name or e-mail address but did not select "Import for others." Please select the "Import for others" option or remove the external author information to continue.')
-      render(:new_import) && return
+      import_errors = ts("You have entered an external author name or e-mail address but did not select \"Import for others.\" " +
```
ariana-paris: I agree that returning just one error so you can surprise the user by returning a different one once they've fixed that isn't brilliant. It should also be a separate method which returns the error messages so the API can use it as well, instead of fattening this controller and duplicating error handling in the API, and there are error conditions which aren't covered at all.

However, as the purpose of the pull request was to expand the coverage of the WorksController, I wanted to minimise the scope of what I changed. (I couldn't stand all the `render view && returns` though 😄 ). Yanking the importing code out of this controller and improving it will happen in AO3-4315, so I will address it in that branch when we have the spoons to review it.

PR #2680, opened 2016-12-26
AO3-4793 Add strong parameters to models where we don't think we are doing mass assignment ( low coverage )

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4793

## Purpose
The idea here is that both brakeman and ourselves do not think we are doing mass assignment in a number of controllers, It seem sensible to confirm this by turning on the use of strong parameters in the controllers. This way we will not get less surprises in the move to rails 4.

## Testing

No Manual testing

### DIFF
--- a/app/models/errors.rb
+++ b/app/models/errors.rb
@@ -1,0 +1,3 @@
+class Errors < ActiveRecord::Base
+  include ActiveModel::ForbiddenAttributesProtection
+end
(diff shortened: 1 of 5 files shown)


app/models/errors.rb
```diff
@@ -0,0 +1,3 @@
+class Errors < ActiveRecord::Base
```
zz9pzza: Should this be Errors or Error ?

PR #2681, opened 2016-12-26
AO3-4794  Add strong parameters to models where we don't think we are doing mass assignment 

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4794 
## Purpose

The idea here is that both brakeman and ourselves do not think we are doing mass assignment in a number of controllers, It seem sensible to confirm this by turning on the use of strong parameters in the controllers. This way we will not get less surprises in the move to rails 4.

## Testing

No manual testing

### DIFF
--- a/app/models/admin/user_creations_controller.rb
+++ b/app/models/admin/user_creations_controller.rb
@@ -1,0 +1,3 @@
+class Admin::UserCreationsController < ActiveRecord::Base
+  include ActiveModel::ForbiddenAttributesProtection
+end
--- a/app/models/api/v1/base_controller.rb
+++ b/app/models/api/v1/base_controller.rb
@@ -1,0 +1,3 @@
+class Api::V1::BaseController < ActiveRecord::Base
+  include ActiveModel::ForbiddenAttributesProtection
+end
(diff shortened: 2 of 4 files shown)


zz9pzza:
So I understand, however we are not 100% certain that we are right about which controllers do mass assignment. By turning on strong parameters on in chunks we reduce the risk that we might miss multiple controllers and have more work landed on us by surprise when the final switch is thrown and strong parameters is applied to all the controllers. I agree it's not a priority but I think it mitigates some if not a lot of risk.

sarken:
I don't think this is the best way to confirm the Brakeman report on mass assignment. When we add forbidden attribute protection to a random selection on controllers, we also run the risk of misleading results. For example, when we tried adding protection to the CollectionProfile model, we ended up with test failures... but only because we were trying to protect our models in the wrong order.

If we really think it's necessary to do this, we need to pay more attention to the sort order. However, as we saw with the Fandom model, even _that_ order isn't infallible.

I think the best approach might be the simplest one: implement strong parameters for the controllers that need it according to Brakeman, and then flip the switch to turn protection on globally. The tests will tell us then if we missed anything; we don't need to spend time trying to puzzle it out before then. (The test coverage will be more thorough by the end, too, so there will be less risk of the tests missing anything.)

PR #2683, opened 2016-12-26
AO3-4795 filtering by fandom in dashboard is broken

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4795

## Purpose

Fix the filtering in a user dashboard ( and hopefully in a collection )

## Testing

When viewing filtered results for a user or collection, and then sorting, the filter is removed.
Example:
Go to a user's dashboard and click on one of their fandoms. This shows you the filtered list of the works they have created in that fandom. The title of the page will for example be "1-3 works by ariana_paris in Testing".
In the right-hand Sort and filter bar, select a sort order (for example, "word count"), then click on the Sort and filter button.
Instead of getting the same 3 works, sorted by word count, you see the list of all the user's works sorted by word count. The Fandom filtering is gone.

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -102,8 +102,12 @@ class WorksController < ApplicationController
       end
 
       tag = @fandom || @tag
-      options[:filter_ids] ||= []
-      options[:filter_ids] << tag.id
+      # This strange dance is because there is an interaction between
+      # strong_parameters and dup, with out the dance 
+      # options[:filter_ids] << tag.id is ignored.
+      filter_ids = options[:filter_ids] || []
+      filter_ids << tag.id
+      options[:filter_ids] = filter_ids
     end
 
     options[:page] = params[:page]
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -103,7 +103,7 @@ class WorksController < ApplicationController
 
       tag = @fandom || @tag
       # This strange dance is because there is an interaction between
-      # strong_parameters and dup, with out the dance 
+      # strong_parameters and dup, without the dance 
       # options[:filter_ids] << tag.id is ignored.
       filter_ids = options[:filter_ids] || []
       filter_ids << tag.id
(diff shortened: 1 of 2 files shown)


features/users/user_dashboard.feature
```diff
@@ -124,3 +124,18 @@ Feature: User dashboard
     And I should not see "Oldest Series"
   Then I should see "Recent bookmarks"
     And I should see "Work 5" within "#user-bookmarks"
+
+  Scenario: Sorting within user or collection filtered results removes fandom filter
```
sarken: The scenario should really describe the _desired_ behavior, not the bug, e.g. "You can follow a fandom link from a user's dashboard and then use the filters to sort within that fandom." Also, since this doesn't explicitly test collections, it shouldn't mention collections.

features/users/user_dashboard.feature
```diff
@@ -124,3 +124,18 @@ Feature: User dashboard
...
+  Given a media exists with name: "TV Shows", canonical: true
+    And a fandom exists with name: "Star Trek", canonical: true
+    And a fandom exists with name: "Star Trek: Discovery", canonical: true
+  When I am logged in as "meatloaf"
```
sarken: I'd suggest making this "And," since being logged in and posting works is all background/setup for what we're testing, and not the main behavior.

features/users/user_dashboard.feature
```diff
@@ -124,3 +124,18 @@ Feature: User dashboard
...
+    And I post the work "Excellent" with fandom "Star Trek"
+    And I post the work "Even more Excellent" with fandom "Star Trek"
+    And I post the work "Exciting" with fandom "Star Trek: Discovery"
+  When I go to meatloaf's user page
```
sarken: This can stay a When, since it's the main behavior we're testing, but `When I follow "Star Trek"` should be an "And"

app/controllers/works_controller.rb
```diff
@@ -102,8 +102,12 @@ def index
       end
 
       tag = @fandom || @tag
-      options[:filter_ids] ||= []
-      options[:filter_ids] << tag.id
+      # This strange dance is because there is an interaction between
+      # strong_parameters and dup, with out the dance
```
zz9pzza: I don't know if it's a temporary hack or a permanent change. I guess it is dependant on whether the strong params gem and rails 4 implement it in the same way. I would add it to the list of things to test after the rails 4 upgrade.

sarken:
:+1: I will merge this and make a Task issue to remind us to check if we still need these changes after Rails 4.

PR #2684, opened 2016-12-27
AO3-4754 Add strong params to bookmark api

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4754

## Purpose

Add strong parameters

## Testing

No Manual testing required

### DIFF
--- a/app/models/api/v1/bookmarks.rb
+++ b/app/models/api/v1/bookmarks.rb
@@ -1,0 +1,3 @@
+class Api::V1::Bookmarks < ActiveRecord::Base
+  include ActiveModel::ForbiddenAttributesProtection
+end


app/models/api/v1/bookmarks.rb
```diff
@@ -0,0 +1,3 @@
+class Api::V1::Bookmarks < ActiveRecord::Base
```
ariana-paris: Having looked into this further, I don't think we need this dummy class. The model the bookmark API mass assigns to is the existing `Bookmark` model, so that's the one we need to protect. Adding the `include ActiveModel::ForbiddenAttributesProtection` to model/bookmarks.rb should do the trick (and will then also cover controllers/bookmarks_controller.rb)

PR #2688, opened 2016-12-29
AO3-4734 Remove vestigal tag_nominations_controller

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4734

## Purpose

Cleans up code that is no longer used.

## Testing

I would edit a challenge and ensure that the nominations for a tag set still work.


sarken:
Ready to merge, but can you please update the issue to reflect that this is removing code and not adding tests? You can copy from https://otwarchive.atlassian.net/browse/AO3-4746. Also, can you update the issue status to In Review?

PR #2689, opened 2016-12-29
AO3-4802 Increase coverage of opendoors/external_authors_controller.rb

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4802

## Purpose

Increase coverage of opendoors/external_authors_controller.rb

## Testing

No manual testing.

### DIFF
--- a/features/importing/archivist.feature
+++ b/features/importing/archivist.feature
@@ -225,3 +225,36 @@ Feature: Archivist bulk imports
       And I am logged in as "OpenDoors"
     When I go to the Open Doors tools page
     Then I should see "Update Redirect URL"
+
+  Scenario: Open Doors committee members can block an email address from having imports
+    Given I have an Open Doors committee member "OpenDoors"
+      And I have an archivist "archivist"
+      And the default ratings exist
+      And I am logged in as "OpenDoors"
+    When I go to the Open Doors tools page
+    Then I fill in "external_author_email" with "random@example.com"
+    When I submit with the 2nd button
+    Then I should see "We have saved and blocked the email address random@example.com"
+    When I am logged in as "archivist"
+      And I go to the import page
+      And I fill in "URLs*" with "http://ao3testing.dreamwidth.org/593.html"
+      And I fill in "Author Name*" with "ao3testing"
+      And I fill in "Author Email Address*" with "random@example.com"
+      And I check the 1st checkbox with id matching "importing_for_others"
+    And I press "Import"
+    Then I should see "Author ao3testing at random@example.com does not allow importing their work to this archive."
+
+  Scenario: Open Doors committee members can supply a new email address for an already imported work.
+    Given I have an Open Doors committee member "OpenDoors"
+      And I have an archivist "archivist"
+      And the default ratings exist
+      And I am logged in as "archivist"
+    When I import the work "http://ao3testing.dreamwidth.org/593.html" by "randomtestname" with email "random@example.com"
+      And the system processes jobs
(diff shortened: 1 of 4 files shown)


features/importing/archivist.feature
```diff
@@ -225,3 +225,36 @@ Feature: Archivist bulk imports
...
     When I go to the Open Doors tools page
     Then I should see "Update Redirect URL"
+
+  Scenario: Open Doors committee members can block an email address from having imports
+    Given I have an Open Doors committee member "OpenDoors"
+      And I have an archivist "archivist"
+      And the default ratings exist
+      And I am logged in as "OpenDoors"
+    When I go to the Open Doors tools page
+    Then I fill in "external_author_email" with "random@example.com"
```
sarken: This and line 236 below it should be `And` -- you're telling the test to do something, so they go with the `When` above

features/importing/archivist.feature
```diff
@@ -225,3 +225,36 @@ Feature: Archivist bulk imports
...
+    When I submit with the 2nd button
+    Then I should see "We have saved and blocked the email address random@example.com"
+    When I am logged in as "archivist"
+      And I go to the import page
+      And I fill in "URLs*" with "http://ao3testing.dreamwidth.org/593.html"
+      And I fill in "Author Name*" with "ao3testing"
+      And I fill in "Author Email Address*" with "random@example.com"
+      And I check the 1st checkbox with id matching "importing_for_others"
+    And I press "Import"
```
sarken: This should be indented like the other `And`s

features/importing/archivist.feature
```diff
@@ -225,3 +225,36 @@ Feature: Archivist bulk imports
...
+      And I have an archivist "archivist"
+      And the default ratings exist
+      And I am logged in as "archivist"
+    When I import the work "http://ao3testing.dreamwidth.org/593.html" by "randomtestname" with email "random@example.com"
+      And the system processes jobs
+      And I am logged in as "OpenDoors"
+      And I go to the Open Doors external authors page
+    Then I should see "random@example.com"
+    When I fill in "email" with "random_person@example.com"
+    When I submit
```
sarken: Should be `And`

features/importing/archivist.feature
```diff
@@ -225,3 +225,36 @@ Feature: Archivist bulk imports
...
+    Given I have an Open Doors committee member "OpenDoors"
+      And I have an archivist "archivist"
+      And the default ratings exist
+      And I am logged in as "OpenDoors"
+    When I go to the Open Doors tools page
+    Then I fill in "external_author_email" with "random@example.com"
+    When I submit with the 2nd button
+    Then I should see "We have saved and blocked the email address random@example.com"
+    When I am logged in as "archivist"
+      And I go to the import page
```
sarken: I think you can replace lines 239-244 with just `And I import the work "http://ao3testing.dreamwidth.org/593.html" by "ao3testing" with email "random@example.com"` here. It does check the "Post without previewing" option, though, so if for some reason that won't work, you could do this instead:
```
And I start to import the work "http://ao3testing.dreamwidth.org/593.html" by "ao3testing" with email "random@example.com"
And I press "Import"
```

sarken:
I've updated the issue status and given the issue description a bit more detail -- I know it's a pain, but we do need to be setting a good example (as you've reminded me about making separate issues for things :P ). If you could copy and paste from other "extend tests for whatever" issues in the future, that would be great.

PR #2692, opened 2016-12-31
AO3-4797 - Add Rspec tests to WorksController 

ariana-paris:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4797

## Purpose

This adds missing coverage to the index, collected, work_search, import, drafts and multiple actions to bring coverage for the WorksController around 91%.

## Testing

This is all automated tests, no change to functionality.


zz9pzza:
It's gone grey I am afraid

spec/controllers/works/importing_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+          params = { urls: urls, importing_for_others: false, import_multiple: "works" }
+          allow_any_instance_of(User).to receive(:is_archivist?).and_return(true)
+
+          get :import, params
+          expect(flash[:error]).to start_with "You cannot import more than #{max}"
+
+          allow_any_instance_of(User).to receive(:is_archivist?).and_call_original
+        end
+
+        it "should error when importing over the maximum number of chapters" do
```
zz9pzza: Can we have a test showing these work for IMPORT_MAX_WORKS_BY_ARCHIVIST-1

ariana-paris: (we discussed this in chat and decided it wouldn't tell us much on reflection)

sarken:
@annelu would you be interested in taking a look at this, since @zz9pzza thinks it could use a second reviewer? If not, no worries!

spec/controllers/works/default_rails_actions_spec.rb
```diff
@@ -289,75 +314,102 @@ def call_with_params(params)
 
       end
     end
+  end
+
+  describe "update" do
+    let!(:update_user) { create(:user) }
```
annelu: Just a nitpick, I don't think the `!` from these three `let`s are necessary, just relying on the late binding works fine.

spec/controllers/works/drafts_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+          get :drafts, user_id: drafts_user.login, pseud_id: drafts_user_pseud.name
+          expect(assigns(:works)).to include(other_pseud_work)
+          expect(assigns(:works)).not_to include(default_pseud_work)
+        end
+      end
+    end
+  end
+
+
+  describe "post_drafts" do
```
annelu: Super nitpick, but the action is `post_draft` 😅

spec/controllers/works/drafts_spec.rb
```diff
@@ -0,0 +1,123 @@
+# frozen_string_literal: true
+require "spec_helper"
+
+describe WorksController do
+  include LoginMacros
+
+  let!(:drafts_user_pseud) { create(:pseud, name: "Visiting User Pseud") }
```
annelu: Not super important, but this and the next `let` don't need the `!`, since they'll get invoked in the `before` block.

spec/controllers/works/importing_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+      expect(controller.send(:import_single, ["url1"], {})).to rescue # stop it whingeing about the response
+      expect(flash[:error]).to start_with "We couldn't successfully import that work, sorry: message"
+      allow_any_instance_of(StoryParser).to receive(:download_and_parse_story).and_call_original
+    end
+  end
+
+  # note: this non-action method renders a template before returning which Rspec can't process,
+  # so calling the method directly requires a rescue to swallow the Rspec error
+  describe "import_multiple" do
+    it "should display the correct error when a timeout occurs" do
```
annelu: Correct me if I'm wrong, but I don't think these are timeout errors, but rather url and other errors

spec/controllers/works/multiple_actions_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+    end
+  end
+
+  describe "confirm_delete_multiple" do
+    it "should return the works specified in the work_ids parameters" do
+      work1 = create(:work, authors: [multiple_works_user.default_pseud], posted: true)
+      work2 = create(:work, authors: [multiple_works_user.default_pseud], posted: true)
+      fake_login_known_user(multiple_works_user)
+      params = { commit: "Orphan", id: work1.id, work_ids: [work1.id, work2.id] }
+      post :confirm_delete_multiple, params
```
annelu: Should some expectations be tested here?

spec/controllers/works/multiple_actions_spec.rb
```diff
@@ -0,0 +1,112 @@
+# frozen_string_literal: true
+require "spec_helper"
+
+describe WorksController do
+  include LoginMacros
+  include RedirectExpectationHelper
+
+  let(:multiple_user_pseud) { create(:pseud) }
+  let!(:multiple_works_user) {
```
annelu: This `!` and the one on line 67 aren't necessary since the user is always invoked before the request.

annelu:
Looks great overall, just a few non-blocking nitpicks! 🎊

spec/controllers/works/importing_spec.rb
ariana-paris: :+1: Copy-paste from the test just above it, of course...

spec/controllers/works/multiple_actions_spec.rb
ariana-paris: Ahahaha. Yes, that would probably be a good idea.

spec/controllers/works/default_rails_actions_spec.rb
ariana-paris: Thanks, I think I originally had trouble with the variable being unbound or apparently bound to a different user, but then never changed it once the problem was resolved!

PR #2693, opened 2016-12-31
AO3-4771 Tests for the collection_participants_controller.

annelu:
Notes:
* The collection lookup in load_participant_and_collection using collection.find raises an ActiveRecord error if the collection isn't found; since there is a case handling this I assumed it wasn't the desired behavior, so I changed it to .where(id: the_id).first which will return nil if the collection isn't found. Then the handling works as expected.
* The participant.approve_membership! did not previously appear to work as written.

## Issue

https://otwarchive.atlassian.net/browse/AO3-4771

## Purpose

Rspec tests for the collection_participants_controller with small tweaks to the controller so functionality works as expected.

## Testing

There are two minor changes: 
* When updating participants, when changing a participant from invited to member, it should now work
* When a participant is deleted between when the edit page is opened and the edit is attempted, there will be a redirect to the archive index with a message that says "Which participant did you want to work with?"

## Credit

potatoesque, she/her

### DIFF
--- a/app/controllers/collection_participants_controller.rb
+++ b/app/controllers/collection_participants_controller.rb
@@ -53,8 +53,11 @@ class CollectionParticipantsController < ApplicationController
     end
     participants = CollectionParticipant.in_collection(@collection).for_user(current_user) unless current_user.nil?
     if participants.empty?
-      @participant = CollectionParticipant.new(collection: @collection, pseud: current_user.default_pseud,
-                        participant_role: CollectionParticipant::NONE)
+      @participant = CollectionParticipant.new(
+        collection: @collection,
+        pseud: current_user.default_pseud,
+        participant_role: CollectionParticipant::NONE
+      )
       @participant.save
       flash[:notice] = t('applied_to_join_collection', :default => "You have applied to join %{collection}.", :collection => @collection.title)
     else
--- a/app/controllers/collection_participants_controller.rb
+++ b/app/controllers/collection_participants_controller.rb
@@ -44,8 +44,6 @@ class CollectionParticipantsController < ApplicationController
     !@participant.is_owner? || (@collection.owners != [@participant.pseud]) || owners_required
   end
 
-
-
   ## ACTIONS
 
   def join
@@ -55,8 +53,8 @@ class CollectionParticipantsController < ApplicationController
     end
     participants = CollectionParticipant.in_collection(@collection).for_user(current_user) unless current_user.nil?
     if participants.empty?
-      @participant = CollectionParticipant.new(:collection => @collection, :pseud => current_user.default_pseud,
-                        :participant_role => CollectionParticipant::NONE)
+      @participant = CollectionParticipant.new(collection: @collection, pseud: current_user.default_pseud,
+                        participant_role: CollectionParticipant::NONE)
       @participant.save
       flash[:notice] = t('applied_to_join_collection', :default => "You have applied to join %{collection}.", :collection => @collection.title)
     else
--- a/app/controllers/collection_participants_controller.rb
+++ b/app/controllers/collection_participants_controller.rb
@@ -22,10 +22,10 @@ class CollectionParticipantsController < ApplicationController
 
   def load_participant_and_collection
     if params[:collection_participant]
-      @participant = CollectionParticipant.where(id: params[:collection_participant][:id]).first
+      @participant = CollectionParticipant.find_by_id(params[:collection_participant][:id])
       @new_role = params[:collection_participant][:participant_role]
     else
-      @participant = CollectionParticipant.where(id: params[:id]).first
+      @participant = CollectionParticipant.find_by_id(params[:id])
     end
 
     no_participant and return unless @participant
--- a/app/controllers/collection_participants_controller.rb
+++ b/app/controllers/collection_participants_controller.rb
@@ -14,27 +14,28 @@ class CollectionParticipantsController < ApplicationController
     redirect_to collection_participants_path(@collection)
     false
   end
-  
+
   def no_participant
     flash[:error] = t('no_participant', :default => "Which participant did you want to work with?")
     redirect_to root_path
   end
-  
+
   def load_participant_and_collection
     if params[:collection_participant]
-      @participant = CollectionParticipant.find(params[:collection_participant][:id])
+      @participant = CollectionParticipant.where(id: params[:collection_participant][:id]).first
       @new_role = params[:collection_participant][:participant_role]
     else
-      @participant = CollectionParticipant.find(params[:id])
+      @participant = CollectionParticipant.where(id: params[:id]).first
     end
+
     no_participant and return unless @participant
     @collection = @participant.collection
-  end    
+  end
 
   def allowed_to_promote
     @participant.user_allowed_to_promote?(current_user, @new_role) || not_allowed(@collection)
   end
-  
+
   def allowed_to_destroy
     @participant.user_allowed_to_destroy?(current_user) || not_allowed(@collection)
   end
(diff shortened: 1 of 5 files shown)


app/controllers/collection_participants_controller.rb
```diff
@@ -54,29 +55,29 @@ def join
...
     if participants.empty?
-      @participant = CollectionParticipant.new(:collection => @collection, :pseud => current_user.default_pseud, 
+      @participant = CollectionParticipant.new(:collection => @collection, :pseud => current_user.default_pseud,
                         :participant_role => CollectionParticipant::NONE)
       @participant.save
       flash[:notice] = t('applied_to_join_collection', :default => "You have applied to join %{collection}.", :collection => @collection.title)
     else
       participants.each do |participant|
         if participant.is_invited?
-          participant approve_membership!
```
annelu: I don't believe this previously worked; if someone else has more background, please let me know!

app/controllers/collection_participants_controller.rb
```diff
@@ -14,37 +14,38 @@ def owners_required
...
   def no_participant
     flash[:error] = t('no_participant', :default => "Which participant did you want to work with?")
     redirect_to root_path
   end
-  
+
   def load_participant_and_collection
     if params[:collection_participant]
-      @participant = CollectionParticipant.find(params[:collection_participant][:id])
+      @participant = CollectionParticipant.where(id: params[:collection_participant][:id]).first
```
annelu: Using `.find` in this line raises an ActiveRecord error if the participant isn't found, which then gives the user a 500. Since `no_participant` exists as a branch for this particular situation, I thought it'd make sense to just set a nil here and go down `no_participant` instead. Otherwise, there'd be no purpose in having `no_participant` and I'd recommend just getting rid of it. Open to either way, really.

app/controllers/collection_participants_controller.rb
ariana-paris: That does look very weird! RubyMine is confused by it and thinks `participant` is a method it can't find, so I think you're right.

app/controllers/collection_participants_controller.rb
ariana-paris: Ideally, we prefer not to make bug fixes in PRs whose original purpose is to add tests, as bug fixes require manual testing for sign-off so there's more overhead involved in getting them into production, whereas test-only PRs can be merged straight away. Something to bear in mind for next time, no need to change this now!

I agree it's best to let the code fall through to `no_participant` since it's there, and thank you for spotting the 500 error. These lines of code were already covered, probably by a Cucumber test, and the tests have passed so I think we can be confident that you haven't broken anything as well as making things more robust. 🎉

spec/controllers/collection_participants_controller_spec.rb
```diff
@@ -0,0 +1,464 @@
...
+    it_redirects_to(path)
+    expect(flash[:notice]).to eq notice
+  end
+
+  def it_redirects_to_with_error(path, error)
+    it_redirects_to(path)
+    expect(flash[:error]).to eq error
+  end
+
+  def it_redirects_to(path)
```
ariana-paris: Amusingly, I have added exactly this same method in one of my PRs too (copied from your previous PR, I think). However, I think this means we're going to use this quite a lot, so it might make sense to either put it at the end of the `spec_helper`, or maybe create another module like `LoginHelpers` for it with your notice and error variations above, so they can all be reused everywhere that tests controller actions with redirects.

spec/controllers/collection_participants_controller_spec.rb
```diff
@@ -0,0 +1,464 @@
...
+              FactoryGirl.create(
+                :collection_participant,
+                collection: collection,
+                pseud: user.default_pseud
+              )
+              user
+            end
+          end
+
+          it "displays the participants in the correct order" do
```
ariana-paris: Does this actually test whether the users are in a specific order, or just that the users are present? I think it will pass even if they're higgledy-piggledy.

spec/controllers/collection_participants_controller_spec.rb
```diff
@@ -0,0 +1,464 @@
...
+    end
+
+    before do
+      fake_login_known_user(user)
+    end
+
+    context "where there is no participant" do
+      it "displays an error and redirects to the index" do
+        put :update, params
+        expect(response).to have_http_status :redirect
```
ariana-paris: This might be a good place to use your `it_redirects_to_with_error` method (also applies to the other tests in this group that check for a redirect and flash error)

app/controllers/collection_participants_controller.rb
annelu: Ah ok, thanks, I'll open a separate bug ticket next time. Thanks!

spec/controllers/collection_participants_controller_spec.rb
annelu: Ah yeah, good catch. I wrote the description first, figured it wasn't so plausible to test the order, then forgot to change it. 👍

spec/controllers/works_controller_spec.rb
```diff
@@ -2,11 +2,7 @@
 
 describe WorksController do
   include LoginMacros
-
-  def it_redirects_to_user_login
-    expect(response).to have_http_status(:redirect)
-    expect(response).to redirect_to new_user_session_path
-  end
+  include RedirectExpectationHelper
```
annelu: I changed this across the board to use the helper, hope that's ok

ariana-paris: :+1: I'll change my tests to use your methods once this is merged too!

spec/controllers/collection_participants_controller_spec.rb
```diff
@@ -0,0 +1,442 @@
...
+        end
+      end
+
+      context "where the user is a collection maintainer" do
+        let(:user_role) { CollectionParticipant::MODERATOR }
+        context "where the participant is updated successfully" do
+          it "successfully updates and redirects to collection participants" do
+            put :update, params
+            it_redirects_to_with_notice(collection_participants_path(collection), "Updated #{participant.pseud.name}.")
+            expect(flash[:notice]).to eq "Updated #{participant.pseud.name}."
```
ariana-paris: I think this line is left over from before the change to the `it_redirects_to_with_notice`?

app/controllers/collection_participants_controller.rb
```diff
@@ -14,37 +14,36 @@ def owners_required
...
+
   def no_participant
     flash[:error] = t('no_participant', :default => "Which participant did you want to work with?")
     redirect_to root_path
   end
-  
+
   def load_participant_and_collection
     if params[:collection_participant]
-      @participant = CollectionParticipant.find(params[:collection_participant][:id])
```
zz9pzza: Can we use find_by_id rather than where ?

http://stackoverflow.com/questions/11161663/find-vs-find-by-vs-where

( soon to be find_by(id: )

app/controllers/collection_participants_controller.rb
zz9pzza: I found this one as well :) 👍

app/controllers/collection_participants_controller.rb
sarken: Just wanted to second this, since `find_by` is generally what the Rails style guide recommends: https://github.com/bbatsov/rails-style-guide#find_by

annelu: Sure, I just used `where` since I understand `find_by_id` is deprecated. No bigs if it'll be changed to `find_by` soon.

sarken:
Thanks! Putting Ready to Merge label back on.

PR #2694, opened 2017-01-01
AO3-4809 - Extend test coverage in Bookmarks Controller

scottsds:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4809

## Purpose

Extends tests for the Bookmarks Controller.

## Testing

2. In the "Search" field near the "Source Files on Master" heading, enter the name of the controller: bookmarks_controller
3. Note the number in the "Coverage" column of the table, it should be greater than 79.88%


sarken:
Marking needs coder action due to chat discussion

scottsds:
I have restored the code I deleted (whoopsies), this should be ready for a review now.

features/step_definitions/admin_steps.rb
```diff
@@ -70,6 +70,20 @@ def update_settings(settings)
   Language.find_or_create_by_short_and_name("FR", "Francais")
 end
 
+Given /^caching is enabled$/ do
```
sarken: I think these changes might have snuck in from somewhere else? It doesn't look like they're being used in the cukes.

zz9pzza: It looks that way.

scottsds: I was going to use them to further test the Controller but couldn't figure it out, but thought these step defs would be good to keep for future tests. I can pull them out if you'd like. Let me know.

sarken: Maybe take them out for now? We've kind of been mumbling about perhaps getting rid of these settings.

scottsds:
Removed the step defs that weren't used.

scottsds:
@sarken I have updated the PR to reflect the changes you requested regarding taking the logic out of the flash message. I have also made sure that the flash message is being automatically checked by a cuke. I left notes on the issue so that testers can manually test as well.

sarken:
Looks good to me!

PR #2695, opened 2017-01-01
AO3-4549 - Stop page title creation from causing a 500 error when fandom is empty

ariana-paris:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4549

## Purpose

It is possible to edit an existing work and save it without a fandom (see [AO3-4808](https://otwarchive.atlassian.net/browse/AO3-4808)). When this happens and the work's cached blurb expires, trying to view the work results in a 500 error because the code that generates the page title assumes there is always a fandom. 

## Testing

1. edit an existing work to remove the fandom
2. get someone with database admin rights to kill the Rails cache
3. check that the work is still accessible, does not have a fandom in the blurb and that the page title says "No fandom specified" where it would normally have the fandom.

## References

The underlying cause of works being saved without a fandom will be addressed in [AO3-4808](https://otwarchive.atlassian.net/browse/AO3-4808).

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -210,19 +210,18 @@ class WorksController < ApplicationController
   def show
     @tag_groups = @work.tag_groups
     if @work.unrevealed?
-      @page_title = ts('Mystery Work')
+      @page_title = ts("Mystery Work")
     else
-      page_title_inner = ''
-      page_creator = ''
-      if @work.anonymous?
-        page_creator = ts('Anonymous')
-      else
-        page_creator = @work.pseuds.collect(&:byline).sort.join(', ')
-      end
-      page_title_inner = if @tag_groups['Fandom'].size > 3
-                           ts('Multifandom')
+      page_creator = if @work.anonymous?
+                       ts("Anonymous")
+                     else
+                       @work.pseuds.map(&:byline).sort.join(", ")
+                     end
+      fandoms = @tag_groups["Fandom"]
+      page_title_inner = if fandoms.size > 3
+                           ts("Multifandom")
                          else
-                           @tag_groups['Fandom'][0].name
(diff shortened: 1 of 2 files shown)


app/controllers/works_controller.rb
```diff
@@ -210,19 +210,18 @@ def drafts
...
-      end
-      page_title_inner = if @tag_groups['Fandom'].size > 3
-                           ts('Multifandom')
+      page_creator = if @work.anonymous?
+                       ts("Anonymous")
+                     else
+                       @work.pseuds.map(&:byline).sort.join(", ")
+                     end
+      fandoms = @tag_groups["Fandom"]
+      page_title_inner = if fandoms.size > 3
```
ariana-paris: This precisely mimics the existing behaviour, though now I look at it, this means that: 
- if there are 4 fandoms, you will get "Multifandom"
- if you have 1 fandom, you will get that fandom
- if you have 2 or 3 fandoms, you will only get the first fandom
Maybe I should take this opportunity to change this line to `if fandoms.size > 1`

spec/controllers/works_controller_spec.rb
```diff
@@ -195,13 +195,13 @@ def call_with_params(params)
 
   describe "create" do
     before do
-      @user = FactoryGirl.create(:user)
+      @user = create(:user)
```
zz9pzza: Why have you stopped using FactoryGirl here ?

ariana-paris: Just succinctness. FactoryGirl::Syntax::Methods is loaded into the Rspec config in the spec_helper so the methods are available directly, and I noticed that we have both syntaxes throughout our tests (with and without `FactoryGirl.`). Since there appears to be no ambiguity using `create` vs `FactoryGirl.create`, I thought I'd standardise this particular spec file on the shorter version to remove visual clutter.

PR #2696, opened 2017-01-02
AO3-4804 Extend tests for challenge_assignments_controller

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4804

## Purpose

Increase coverage of challenge_assignments_controller

## Testing

No manual testing required.


features/gift_exchanges/challenge_giftexchange.feature
```diff
@@ -348,6 +348,46 @@ Feature: Gift Exchange Challenge
     Then I should see "myname1"
       And I should see "Fulfilled Story"
 
+  Scenario: A mod can default all incomplete assignments
```
sarken: It looks like the indenting on this went a little wonky -- the `Given` and `And`s here are right, but everything else is wrong. Specifically, each line starting with `When` or `Then` is indented 3 spaces and each line starting `And` is 5.

features/gift_exchanges/challenge_giftexchange.feature
```diff
@@ -348,6 +348,46 @@ Feature: Gift Exchange Challenge
...
+      And I fulfill my assignment
+   When I am logged in as "mod1"
+     And I go to the "Awesome Gift Exchange" assignments page
+     And I follow "Default All Incomplete"
+   Then I should see "All unfulfilled assignments marked as defaulting."
+     And I should see "Undefault myname2"
+     And I should see "Undefault myname3"
+     And I should see "Undefault myname4"
+
+  Scenario: User can default and undefault on their assignment
```
sarken: The moderator, not the user, does the undefaulting, so I think the scenario here should be either "User can default on their assignment, and mod can undefault them."

It looks like you're _also_ testing that the mod can assign a pinch hitter to a user who has defaulted. Because that's unrelated to undefaulting a user, I'd suggest making that its own scenario, even if it does make the test take a tiny bit longer to run.

Also, the indenting for this scenario is off -- all the `Given`, `When`, and `Then` are indented 3 spaces, and all the `And` are indented 5.

features/gift_exchanges/challenge_giftexchange.feature
```diff
@@ -348,6 +348,46 @@ Feature: Gift Exchange Challenge
...
+   Then I should see "Assignment updates complete!"
+     And I should not see "Undefault"
+      When I am logged in as "myname1"
+     And I go to my user page
+     And I follow "Assignments"
+     And I follow "Default"
+   Then I should see "We have notified the collection maintainers that you had to default on your assignment."
+   When I am logged in as "mod1"
+     And I go to the "Awesome Gift Exchange" assignments page
+     And I fill in the 2nd field with id matching ".*" with "nonexistent"
```
sarken: I gave it a try, and this should work just fine as `And I fill in "Pinch Hitter:" with "nonexistent"`. If it doesn't, though, it would help to add a comment above so we can understand what's being done here.

features/gift_exchanges/challenge_giftexchange.feature
```diff
@@ -348,6 +348,46 @@ Feature: Gift Exchange Challenge
...
+     And I press "Submit"
+   Then I should see "We couldn't find the user nonexistent to assign that to."
+   When I fill in the 2nd field with id matching ".*" with "myname1"
```
sarken: Likewise, this should work just fine if you use the label text and write `When I fill in "Pinch Hitter:" with "myname1"`.

PR #2697, opened 2017-01-02
AO3-4806 Remove unused methods

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4806

## Purpose

Clean up code in challenge_assignments_controller 

## Testing

Given we can find no references to the use of this code, I would hope automated testing should be sufficient.


sarken:
I think this is fine, but let's merge it with or after #2696 (which adds coverage to undefaulting) just to be 100% safe.

PR #2698, opened 2017-01-03
AO3-4811 Allow error to be reached in prompt controller, removing error 500

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4811

## Purpose

While working on the tests I discovered that it was impossible to get to one of the error messages, the app would give a 500 instead. While it should not be possible to get the error via the user interface this pull should make the system more robust and improve coverage slightly.

## Testing

No manual testing required.


annelu:
👍

PR #2699, opened 2017-01-03
AO3-4810 Extend tests for PromptsController

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4810

## Purpose

prompts_controller currently has 65.17% test coverage on Coveralls:
https://coveralls.io/builds/9474724/source?filename=app%2Fcontrollers%2Fprompts_controller.rb
We'd like to increase that, with an ultimate goal of 90% or greater.
How to test:
2. In the "Search" field near the "Source Files on Master" heading, enter the name of the controller: prompts_controller


spec/controllers/prompts_controller_spec.rb
```diff
@@ -0,0 +1,131 @@
...
+  describe 'create_prompt_offer' do
+    it 'should have no errors and redirect' do
+      signups = create(:challenge_signup)
+      prompt = signups.collection.prompts.first
+      signups.collection.challenge.signup_open = true
+      signups.collection.challenge.save
+      # Login as the signup owner
+      fake_login_known_user(Pseud.find(ChallengeSignup.in_collection(signups.collection).first.pseud_id).user)
+      post :create, collection_id: signups.collection.name, prompt_type: 'offer', prompt: {}
+      expect(response).to redirect_to("#{collection_signups_path(signups.collection)}/#{prompt.challenge_signup_id}/edit")
```
sarken: Hm -- I think I'm confused. If this is testing `create`, we only redirect to the edit signup path when the prompt is invalid... but if that's the case, we'd `expect(flash[:error]).to eq "That prompt would make your overall sign-up invalid, sorry."`. What are you testing here? A `context` would be helpful.

spec/controllers/prompts_controller_spec.rb
```diff
@@ -0,0 +1,131 @@
...
+      expect(flash[:notice]).blank?
+      expect(flash[:error]).blank?
+    end
+  end
+
+  describe 'update_prompt' do
```
sarken: Should be `describe "update" do`. We could also use a context here, like: `context "when prompt is valid" do`

spec/controllers/prompts_controller_spec.rb
```diff
@@ -0,0 +1,131 @@
...
+      put :update, collection_id: signups.collection.name, prompt_type: 'offer', prompt: {description: 'This is a description'}, id: prompt.id
+      expect(flash[:notice]).to eq 'Prompt was successfully updated.'
+      expect(flash[:error]).blank?
+      expect(response).to redirect_to("#{collection_signups_path(signups.collection)}/#{signups.id}")
+      new_prompt = signups.collection.prompts.first
+      expect(new_prompt.description).to eq("<p>This is a description</p>")
+   end
+  end
+
+  describe 'destroy' do
```
sarken: Description-wise, maybe something more like:
```
describe "destroy" do
  it "redirects with an error when sign-ups are closed" do
  end
  it "redirects with an error when it would make a sign-up invalid" do
  end
  it "deletes the prompt and redirects with a success message" do
  end
end
```

spec/controllers/prompts_controller_spec.rb
```diff
@@ -0,0 +1,131 @@
+require 'spec_helper'
+
+describe PromptsController do
+  include LoginMacros
```
sarken: I'd suggest also including `RedirectExpectationHelper` and then using that in all the examples that check that something redirects with an error.

spec/controllers/prompts_controller_spec.rb
```diff
@@ -1,18 +1,136 @@
...
     fake_login
   end
 
-  describe 'no_prompt' do
-    it 'should show an error and redirect' do
+  describe "no_challenge" do
+    it "should show an error and redirect" do
+      get :show, collection_id: collection.name
+      it_redirects_to collection_path(collection)
+      expect(flash[:error]).to eq "What challenge did you want to sign up for?"
```
sarken: With RedirectExpectationHelper, it should be possible to write one line to check anything that redirects with an error or notice. For an error, it would be: `it_redirects_to_with_error(collection_path(collection), "What challenge did you want to sign up for?")`

spec/controllers/prompts_controller_spec.rb
```diff
@@ -1,18 +1,136 @@
...
+      signups = create(:challenge_signup)
+      post :create, collection_id: signups.collection.name
+      it_redirects_to collection_signups_path(signups.collection) + "/new"
+      expect(flash[:error]).to eq "Please submit a basic sign-up with the required fields first."
+    end
+  end
+
+  describe "signups_closed" do
+    it "should show an error and redirect" do
+      signups = create(:challenge_signup)
```
sarken: You use this in a lot of places -- you should put `let(:signup) { create(:challenge_signup) }` at the top, too. (It should be changed to the singular because we're only working with one sign up, not multiple.)

spec/controllers/prompts_controller_spec.rb
```diff
@@ -1,18 +1,136 @@
...
+      prompt = signups.collection.prompts.first
+      post :edit, id: prompt.id, collection_id: signups.collection.name
+      it_redirects_to collection_path(signups.collection)
+      expect(flash[:error]).to eq "You can't edit someone else's sign-up!"
+    end
+  end
+
+  describe "new" do
+    context "when prompt_type is offer" do
+      it "should have no errors and redirect" do
```
sarken: I don't think there's a redirect here -- we're just loading the new prompt form in `new` and not submitting it or anything yet

spec/controllers/prompts_controller_spec.rb
```diff
@@ -1,18 +1,136 @@
...
+  end
+
+  describe "create" do
+    it "should have no errors and redirect" do
+      # Login as the signup owner
+      fake_login_known_user(Pseud.find(ChallengeSignup.in_collection(open_signup.collection).first.pseud_id).user)
+      post :create, collection_id: open_signup.collection.name, prompt_type: "offer", prompt: {}
+      it_redirects_to "#{collection_signups_path(open_signup.collection)}/"\
+                      "#{open_signup.collection.prompts.first.challenge_signup_id}/edit"
+      expect(flash[:notice]).blank?
```
sarken: It looks like there should be a success notice on this saying "Prompt was successfully added." So you should probably check for that and change the description to `it "should redirect with a success message"` or `it "should redirect with a notice"`.

spec/controllers/prompts_controller_spec.rb
```diff
@@ -1,18 +1,136 @@
...
+      expect(flash[:error]).blank?
+    end
+  end
+
+  describe "update" do
+    context "when prompt is valid" do
+      it "should have no errors and redirect" do
```
sarken: `it "should save the prompt and redirect with a success message"` is probably more representative of the important bits of code than the lack of error.

spec/controllers/prompts_controller_spec.rb
```diff
@@ -1,18 +1,136 @@
...
+        fake_login_known_user(Pseud.find(ChallengeSignup.in_collection(open_signup.collection).first.pseud_id).user)
+        put :update, collection_id: open_signup.collection.name, prompt_type: "offer",\
+                     prompt: { description: "This is a description" }, id: open_signup.collection.prompts.first.id
+        expect(flash[:notice]).to eq "Prompt was successfully updated."
+        expect(flash[:error]).blank?
+        it_redirects_to "#{collection_signups_path(open_signup.collection)}/#{open_signup.id}"
```
sarken: You might be able to use `it_redirects_to_with_notice(path, message)` here

spec/controllers/prompts_controller_spec.rb
zz9pzza: I am not sure about this bit...

spec/controllers/prompts_controller_spec.rb
```diff
@@ -1,18 +1,126 @@
...
+      post :create, collection_id: signup.collection.name
+      it_redirects_to_with_error(collection_signups_path(signup.collection) + "/new",\
+                                 "Please submit a basic sign-up with the required fields first.")
+    end
+  end
+
+  describe "signups_closed" do
+    it "should show an error and redirect" do
+      # Login as the signup owner
+      fake_login_known_user(Pseud.find(ChallengeSignup.in_collection(signup.collection).first.pseud_id).user)
```
annelu: What I would do here is instead of having `fake_login` in the before block up top and the `fake_login_known_user` here, is to have this in the `before` block on line 15:

```
let(:user) { create(:user) }
before do
  fake_login_known_user(user)
end
```

Then, for each context here you can just `let` user as `Pseud.find(ChallengeSignup.in_collection(signup.collection).first.pseud_id).user` or whatever it's supposed to be. It's still a line per context, but it seems like it'd adhere to the DRY principle a bit better.

Or, it would be fewer lines to set the default user in the larger context to be `Pseud.find(ChallengeSignup.in_collection(signup.collection).first.pseud_id).user` and `let` user to be `create(:user)` where it's different, but I think setting the default as `create(:user)` makes more sense.

annelu: I'd just `let` the challenge sign up user (`Pseud.find(ChallengeSignup.in_collection(signup.collection).first.pseud_id).user`) in the larger context, so that it can just be referred to with a more readable variable in the rest of the tests.  I don't think there's any reason not to do it, since `let` does late binding, so it won't slow down the tests by doing the lookup when it's not necessary.

zz9pzza: @annelu I keep on looking at this and I am don't know what you mean, sorry :(

annelu: Sorry, I should have been more clear. I'd just have a `let(:challenge_sign_up_user) { (Pseud.find(ChallengeSignup.in_collection(signup.collection).first.pseud_id).user) }` in the `describe PromptsController do` context, then just refer to `challenge_sign_up_user` in the rest of the tests, so it'd be a little more readable. Just a non-blocking style comment.

zz9pzza: @annelu I can't get this to work, would you be kind enough to show me what you mean ?

annelu: Yeah, so on line 15 instead of

```
 before do
    fake_login
  end
```

have something like

```
let(:user) { create(:user) }
before do
  fake_login_known_user(user)
end
```

Then have the `signups_closed` context look something like this:

```
describe "signups_closed" do
  let(:user) { Pseud.find(ChallengeSignup.in_collection(signup.collection).first.pseud_id).user }
  it "should show an error and redirect" do
    post :create, collection_id: signup.collection.name
    it_redirects_to_with_error(signup.collection, "Signup is currently closed: please contact a moderator for help.")
  end
end
```

And do that for every context where you want to log in as the signup owner, or some other user other than the one created up above in the larger context.

zz9pzza: Thank you, I have updated the tests to the point where following that pattern fails.

spec/controllers/prompts_controller_spec.rb
```diff
@@ -1,18 +1,127 @@
...
+      post :edit, id: prompt.id, collection_id: signup.collection.name
+      it_redirects_to_with_error(signup.collection, "You can't edit someone else's sign-up!")
+    end
+  end
+
+  describe "new" do
+    context "when prompt_type is offer" do
+      it "should have no errors" do
+        # Login as the signup owner
+        fake_login_known_user(Pseud.find(ChallengeSignup.in_collection(open_signup.collection).first.pseud_id).user)
```
annelu: Since `fake_login_known_user` is being called in the `before` block on line 17, the user could be let in all contexts that still contain the login. They're currently being logged in on lines 55, 67, 79, 93, 101, and 109. That'd prevent the logged in user from being stubbed twice for those `it` blocks.

zz9pzza: @annelu I did try for example:

  describe "new" do
    context "when prompt_type is offer" do
      let(:user) { Pseud.find(ChallengeSignup.in_collection(signup.collection).first.pseud_id).user }
      it "should have no errors" do
...

And

  describe "new" do
    let(:user) { Pseud.find(ChallengeSignup.in_collection(signup.collection).first.pseud_id).user }
    context "when prompt_type is offer" do

Removing the context block also fails in the same way

it errors with: 

  1) PromptsController new when prompt_type is offer should have no errors
     Failure/Error: expect(response).to have_http_status(200)
       expected the response to have status code 200 but it was 302
     # ./spec/controllers/prompts_controller_spec.rb:58:in `block (4 levels) in <top (required)>'

Which I think is not logging in properly

annelu: Weird, they pass for me. Does [this](https://github.com/zz9pzza/otwarchive/pull/3) work for you?

sarken:
It looks like @annelu's comments resolved the remaining issue, and someone's going to be working on this controller _very_ soon, so let's merge this!

PR #2700, opened 2017-01-05
AO3-4815 Elasticsearch overhaul

sarken:
# Issue

https://otwarchive.atlassian.net/browse/AO3-4815

# Details

The code is primarily concerned with:

* updating to Elasticsearch 2.x (we're running 2.3.3 on our dev server, but we can upgrade further if needed) 
* moving from the Tire gem to the official Elasticsearch gem (we have version 1.0.15 of the gem and would like to move up to 2.0)
* adding an interface for easily excluding things from work and bookmark listings
    * work page with Elasticsearch-based filters: http://sarken.archiveofourown.org/tags/Gen/works
    * bookmark page with Elasticsearch-based filters: http://sarken.archiveofourown.org/tags/Gen/bookmarks
* adding and updating the corresponding automated tests for these changes

The automated tests are not very helpful at the moment due to missing index errors. Additionally, using the include and exclude checkboxes and radio buttons on the work and bookmark pages does not appear to do anything. 

However, after using a previous version of the code on http://elzj78.archiveofourown.org/tags/Gen/works, I think most of the Elasticsearch code is working -- it's primarily a matter of incorporating the new interface. The one major exception is that there doesn't appear to be any code to handle exclusion filtering for bookmarks.

To make the transition as smooth as possible, we want to use rollout (https://github.com/fetlife/rollout) to allow us to create the new indexes and do a little bit of testing before enabling the new Elasticsearch code for all of our users. (And we'd of course like to remove any old code once the new is enabled.)

I hope that helps a little, and I'm sorry I don't have more information prepared. If there's something you'd like more detail about, please let me know and I will do my best to get it.


dondenoncourt:
## A couple of general questions and comments:

* Why do all the models associated with ElasticSearch still Tire?
    * `include Tire::Model::Search`
    * used in: bookmark, pseud, tag, work
  * I would think they'd `include Elasticsearch::Model`
  * although the following directly uses ES:
    * app/models/bookmark_search.rb|53 col 16| response = ElasticsearchSimpleClient.perform_count(Bookmark.index_name, 'bookmark', query)
    * app/models/indexing/index_subqueue.rb|107 col 5| ElasticsearchSimpleClient.send_batch(@batch)
    * app/models/work_search.rb|75 col 16| response = ElasticsearchSimpleClient.perform_count(Work.index_name, 'work', query)

* Some of the Cucumber tests fail with IndexMissingException
  *  I'm resisting the urge to create those indexes. 

* Note: In my last project where I converted from the Tire to to ElasticSearch gems, creation and population of an index from the model was easy (when all rows went into the ES index:
  * `Incident.__elasticsearch__.create_index! force: true`
  * `Incident.import`
* When the put to ES was conditional, I'd use callbacks to get/put JSON for associated objects on local setup:
  * `Scene.__elasticsearch__.create_index! force: true`
  * `Scene.where(updated_at: (DateTime.now-1.year)..(DateTime.now)).update_all(updated_at: DateTime.now.to_s)`

sarken:
It looks like we started to go directly to Elasticsearch in those files due to inconsistent bookmark and work counts appearing on the site:
- https://otwarchive.atlassian.net/browse/AO3-4070
- https://github.com/otwcode/otwarchive/pull/1831
- https://github.com/otwcode/otwarchive/pull/1812

(We used to display work and bookmark counts in two places, and one of those counts came from Elasticsearch and the other directly from the MySQL database. The MySQL count was faster to update, so users would get concerned/confused when they saw different numbers in different places.) 

I'm just guessing on these, but if we're not currently using Tire at all, it's possible we simply forgot to remove `include Tire::Model::Search` from the models. If we _are_ currently using Tire, maybe the plan was to remove it after we turned on the new search, since we were expecting a clean up pull request.

zz9pzza:
One comment is that i would like us to run both old and new systems at the same time so new indexes can be generated and we can migrate from using one system to the other for displays using rollout.

dondenoncourt:
* Here's my estimate on how much effort it might take to complete the elasticsearch upgrade:

1) Models should be modified to replace the Tire gem with the elasticsearch-rails gem
This may require adding or changing the mapping and as_indexed_json gems.  Data can then be imported into ElasticSearch there are 4 models: Bookmark, Pseud, Tag, and work.
Tests will need to be written to verify the mapping and Elasticsearch index creation.
2) Data needs to be populated from the relational database to Elasticsearch. This is a straightforward task as there are model methods available to perform this task.
3) Existing queries and filters may require modifications. They certainly need to have unit tests written and then discovered issues corrected.
4) Use of the rollout strategy should be tested. This would require the creation of a second Elasticsearch service.
I am not aware of the requirement for populating new data to the older Elasticsearch index. If that is a requirement, dual population of data will have to be enabled and tested.

Estimates:
1) 4 hours per model 16 hours
2) 4 hours
3) 16 hours
4) 16-32 hours
Total: 52-68 hours with a 60% confidence level. To get 80% I'd have to say 100 hours.

PR #2704, opened 2017-01-07
AO3-4819 Tests for skins controller

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4819

## Purpose

Extend tests for skins controller, plus do some minor refactoring and shuffling of the skin tests because daaaang, was that one long feature.

Should bring the skins_controller tests up to 93% and, as a bonus, the Work Skin model up to 100.

## Testing

Refer to JIRA.

### DIFF
--- a/features/admins/admin_skins.feature
+++ b/features/admins/admin_skins.feature
@@ -10,14 +10,14 @@ Feature: Admin manage skins
   Given basic skins
     And the approved public skin "public skin"
     And I am logged in as an admin
-   When I follow "Approved Skins"
+  When I follow "Approved Skins"
     And I check "Cache"
-   Then I press "Update" 
+  Then I press "Update" 
     And I should see "The following skins were updated: public skin"
-   When I follow "Approved Skins"
+  When I follow "Approved Skins"
     And I check "Uncache"
     And I press "Update"
-   Then I should see "The following skins were updated: public skin"
+  Then I should see "The following skins were updated: public skin"
 
   Scenario: Admin can add a public skin to the chooser and then remove it
   Given the approved public skin "public skin"
--- a/features/admins/admin_skins.feature
+++ b/features/admins/admin_skins.feature
@@ -100,10 +100,10 @@ Feature: Admin manage skins
     And I should not see "#title"
   Then the cache of the skin on "public skin" should expire after I save the skin
 
-  Scenario: Admins should be able to unapprove public skins, which should also remove
-  them from preferences
+  Scenario: Admins should be able to unapprove public skins, which should also
+  remove them from preferences
   Given "skinuser" is using the approved public skin "public skin" with css "#title { text-decoration: blink;}"
-    And I unapprove the skin "public skin"
+  When I unapprove the skin "public skin"
   Then I should see "The following skins were updated: public skin"
     And I should see "public skin" within "table#unapproved_skins"
   When I am logged in as "skinuser"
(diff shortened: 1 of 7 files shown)


features/admins/admin_skins.feature
```diff
@@ -0,0 +1,114 @@
...
+    And I fill in "Description" with "Blinky love (admin modified)"
+    And I submit
+  Then I should see an update confirmation message
+    And I should see "(admin modified)"
+    And I should see "#greeting.logged-in"
+    And I should not see "#title"
+  Then the cache of the skin on "public skin" should expire after I save the skin
+
+  Scenario: Admins should be able to unapprove public skins, which should also
+  remove them from preferences
```
zz9pzza: Does this work without a continuation \  or just putting it on the line above ?

features/other_b/skin.feature
```diff
@@ -106,7 +35,8 @@ Feature: creating and editing skins
     And I should see "#title {" within "style"
     And I should see "text-decoration: blink;" within "style"
 
-  Scenario: A user should be able to select one of their own non-public skins to use in their My Skins page
+  Scenario: A user should be able to select one of their own non-public skins to use in
+  their My Skins page
   Given I am logged in as "skinner"
```
zz9pzza: If it does work it makes it harder to see where the tests start. Maybe a blank line ?

features/other_b/skin.feature
```diff
@@ -491,16 +207,43 @@ Feature: creating and editing skins
...
-    And the approved public skin "public skin" with css "#title { text-decoration: blink;}"
-    And the skin "public skin" is cached
-    And the skin "public skin" is in the chooser
+  Scenario: A user should be able to delete a skin
+  Given I am logged in as "skinner"
+    And I create the skin "Ugly Skin"
+  When I go to "Ugly Skin" skin page
+    And I follow "Delete"
+  Then I should see "The skin was deleted."
+    And I should be on skinner's skins page
```
zz9pzza: Is it worth adding a check that the skin is not a visible on the skinners' skin's page ?

features/admins/admin_skins.feature
```diff
@@ -0,0 +1,114 @@
+@skins
+Feature: Admin manage skins
+
+  Scenario: Users should not be able to see the admin skins page
+  Given I am logged in as "skinner"
+  When I go to admin's skins page
+  Then I should see "I'm sorry, only an admin can look at that area"
+
+  Scenario: Admin can cache and uncache a public skin
```
zz9pzza: With a pull that ends up this big it would be worth doing the refactor as a separate pull I think so we don't have to think about all the things at once.

features/admins/admin_skins.feature
sarken: I can put it all on the same line -- it was like that and works just fine; I was just trying to keep it down to our desired line length.

zz9pzza: If it works leave it as it is, I do think I prefer a blank line between the Scenario and the test if we do this.

sarken: Yeah, I generally like a blank line between the two regardless -- whitespace is good! -- but we don't have a standard and I didn't want to add more changes.

features/other_b/skin.feature
sarken: I think that makes sense, yep!

PR #2706, opened 2017-01-08
AO3-4787 Add browserstack as a test platform

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4787

## Purpose

Allows us to target browserstack for tests. There are two use cases one is to get a set of video's/images of the site on various platforms. The second is to be able to test javascript issues on a real platform.

I don't expect to use the script/codeship_tests.sh  as a script on codeship, it is here as a reference as to what to cut and paste in to the correct section on codeship.

## Testing

No manual testing required.


zz9pzza:
Deferring until I have written more tests elsewhere.

redsummernight:
We'll keep using BrowserStack for manual testing, but we decided against adding to our CI setup.

PR #2708, opened 2017-01-11
AO3-4824 Improve tests in model for admin users

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4824

## Purpose

Improve the quality of the model tests for admin_users.

## Testing

No manual testing required.

### DIFF
--- a/factories/admin.rb
+++ b/factories/admin.rb
@@ -2,7 +2,7 @@ require 'faker'
 
 FactoryGirl.define do
   factory :admin do
-    login
+    login {generate(:login)}
     password "password"
     password_confirmation { |u| u.password }
     email
--- a/factories/admin.rb
+++ b/factories/admin.rb
@@ -2,7 +2,7 @@ require 'faker'
 
 FactoryGirl.define do
   factory :admin do
-    login {generate(:login)}
+    login { generate(:login) }
     password "password"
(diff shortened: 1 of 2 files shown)


annelu:
Hey, this looks generally good! The only comment I have is that the tests about login and password length seem to expect that the archive config is being used, when it's the Authlogic gem defaults that are being used instead since they never get overriden in the model. So the login minimum length is 3, max is 100; password min is 4 and has no max. (Source: http://www.rubydoc.info/github/binarylogic/authlogic/Authlogic/ActsAsAuthentic/Login/Config#validates_length_of_login_field_options-instance_method).

I don't know if the desired behavior is to use the archive config values, in which case I think we could do a code change to the `admin` model, or leave it as is, in which case I don't think we really need to test the Authlogic validation. What do you think?

zz9pzza:
So the idea here was to write tests that showed us how the system was configured and allowed a better understanding of the combination of authlogic and the admin model.

I had an expectation the authlogic would have been configured with the defaults but found that not to be the case. I thought it better to write tests that reflected reality rather than make a code change when we want to rip out auth logic and put in devises as you can see in #2710 

Opinions on #2710 would definitely be appreciated.

annelu:
Ah, if that's the case, should this PR be closed?

PR #2709, opened 2017-01-14
AO3-4825 Delete obsolete help files

SammieLouise:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4825

## Purpose

Removes help files that are no longer in use (ie they are not linked anywhere on the site)

Changes should not be visible from the site side. 

When investigating, Docs found that these files had been replaced by more specific help files, or there was no longer a pop-up link for the feature anymore at all. Some are replaced by a short line of text on the page itself.


zz9pzza:
The only file I could find a reference to was in one of the files getting removed.

/public/help/challenge-defaulted-assignments.html

PR #2710, opened 2017-01-15
AO3-4844 Migrate to Devise for admin users

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4844

## Purpose

Migrates us to using devises for admin users.

## Testing

This passes test which is positive. I would ask Betsy  in particular to do so as she has a old,old password encrypted form on test.

Well check that non admins can login and out and can't access admin only pages.
Check that admins can log in and out with their old password and that they can access admin pages.
Check that an admin can change their password.

I would do a backup of the test db in case we have to revert and the reverting fails.


app/controllers/application_controller.rb
```diff
@@ -152,9 +143,23 @@ def redirect_back_or_default(default = root_path)
     end
   end
 
+  def after_sign_in_path_for(resource)
```
zz9pzza: I will do when I get to the bit for user accounts.

app/controllers/admin/sessions_controller.rb
```diff
@@ -0,0 +1,8 @@
+# Namespaced Admin class
+class Admin
+ # Handle admin session authentication
+ class SessionsController < Devise::SessionsController
```
ariana-paris: Please add a space before `class SessionsController` so it's indented by two spaces.

app/models/admin.rb
```diff
@@ -1,18 +1,18 @@
...
-    if (ArchiveConfig.BCRYPT  || "true")  == "true" then
-      config.crypto_provider = Authlogic::CryptoProviders::BCrypt
-      config.transition_from_crypto_providers = [Authlogic::CryptoProviders::Sha512, Authlogic::CryptoProviders::Sha1]
-    else
-      config.crypto_provider = Authlogic::CryptoProviders::Sha512
-      config.transition_from_crypto_providers = [Authlogic::CryptoProviders::Sha1]
-    end
-  end
+
+  devise :database_authenticatable, \
```
ariana-paris: I'm pretty certain you don't need all these backslashes as the commas already indicate that the statement isn't finished

app/models/admin.rb
```diff
@@ -1,18 +1,15 @@
...
+
+  devise :database_authenticatable, :registerable, :encryptable,\
+    :validatable, password_length: ArchiveConfig.PASSWORD_LENGTH_MIN..ArchiveConfig.PASSWORD_LENGTH_MAX
   
   has_many :log_items
   has_many :invitations, :as => :creator
   has_many :wrangled_tags, :class_name => 'Tag', :as => :last_wrangler 
+
+  validates :login, presence: true, uniqueness: true, length: { in: ArchiveConfig.LOGIN_LENGTH_MIN..ArchiveConfig.LOGIN_LENGTH_MAX }
+  validates :email,  uniqueness: true
```
ariana-paris: Can you remove one of the spaces before `uniqueness:` as you only need one space after a comma?

app/controllers/admins/sessions_controller.rb
```diff
@@ -0,0 +1,25 @@
+class Admins::SessionsController < Devise::SessionsController
```
ariana-paris: I'm confused. Why is there both an `Admins::SessionsController` (plural Admins) and an `Admin::SessionsController` (singular Admin)? Can we get rid of this one as it's all by itself in a folder and doesn't even seem to be doing anything?

app/controllers/admin/sessions_controller.rb
```diff
@@ -0,0 +1,8 @@
+# Namespaced Admin class
```
ariana-paris: You probably don't need to specify this - let's assume people reading this code understand some Ruby

app/controllers/application_controller.rb
```diff
@@ -152,9 +143,23 @@ def redirect_back_or_default(default = root_path)
...
+    admin_users_path
+  end
+
+  def authenticate_admin!
+    if admin_signed_in?
+      super
+    else
+      redirect_to root_path, notice: "I'm sorry, only an admin can look at that area"
+      ## if you want render 404 page
```
ariana-paris: Will we ever want to do this, or could this be removed?

app/views/admins/sessions/new.html.erb
```diff
@@ -0,0 +1,11 @@
+<h2>Log in as Admin</h2>
```
ariana-paris: Following on from my comments about the commented out controller sitting alone in the `Admins` namespace, I think this view should be in the `admin` (singular) folder with the rest of the admin-related views, and where the `Admin::SessionsController` (singular) can hopefully find it.

spec/models/admin_spec.rb
```diff
@@ -1,73 +1,88 @@
...
+        raise_error(ActiveRecord::RecordInvalid, \
+                    "Validation failed: Login has already been taken")
     end
 
+    # Why does this get a double error ?
     it "is invalid if email already exists" do
-      expect(build(:admin, email: existing_user.email)).to be_invalid
+      expect { create(:admin, email: existing_user.email) }.to \
+        raise_error(ActiveRecord::RecordInvalid, \
+                    "Validation failed: Email has already been taken, Email has already been taken")
```
ariana-paris: Side-eyeing that repetitive error message. Is that really what it says when you validate one admin?

app/controllers/admin/sessions_controller.rb
zz9pzza: This comes directly from the generators. I can remove it if you want ?

app/controllers/application_controller.rb
zz9pzza: This is so we get the same behaviour as we currently have with authlogic.

spec/models/admin_spec.rb
zz9pzza: Why yes it does :)

Why it does is another question.

app/controllers/admins/sessions_controller.rb
zz9pzza: We need to have class Admins::SessionsController and this is where the generators put it and the contents are still the default from the generator. I could get rid of the comments if you want ?

app/controllers/application_controller.rb
zz9pzza: Do you mean the comment ?

zz9pzza:
Looking at this I can see that it is not creating users with bcrypt so I need to fix that.

app/models/admin.rb
```diff
@@ -1,18 +1,48 @@
...
-    config.transition_from_restful_authentication = true
-    if (ArchiveConfig.BCRYPT  || "true")  == "true" then
-      config.crypto_provider = Authlogic::CryptoProviders::BCrypt
-      config.transition_from_crypto_providers = [Authlogic::CryptoProviders::Sha512, Authlogic::CryptoProviders::Sha1]
-    else
-      config.crypto_provider = Authlogic::CryptoProviders::Sha512
-      config.transition_from_crypto_providers = [Authlogic::CryptoProviders::Sha1]
+
+  devise :database_authenticatable,
+         :registerable,
```
bingeling: I don't know how we currently handle Admin accounts, but :registerable means Admins can theoretically sign up themselves and also delete their own account? We may want to skip that for Admins. [devise_invitable](https://github.com/scambra/devise_invitable) may be worth a look, it's a gem to invite users via email, so only an existing admin can create a new admin.

I think we also need :rememberable if we allow Admins to stay signed in/use cookies.

app/models/admin.rb
```diff
@@ -1,18 +1,42 @@
...
+  # Not useful but https://github.com/plataformatec/devise/issues/511
+
+  alias :devise_valid_password? :valid_password?
+  def valid_password?(password)
+    begin
+      result = super(password)
+      # Now the common form is that we are using an authlogic method so lets
+      # test that on failure
+      return true if result
+      if Authlogic::CryptoProviders::BCrypt.matches?(self.encrypted_password,[password, self.password_salt].compact)
```
bingeling: I don't know enough about encryption to judge if this weakens our security or not. Your comment implies that you did your research and I trust you on that.

app/models/admin.rb
zz9pzza: I don't know if we want the :rememberable as we want admins to have to log in each time. I can also see an issue later when we would want Users to be able to stay logged in ( probably for a different amount of time that we might want to keep admins logged in for ).

I will remove registerable thank you.

bingeling: That makes sense.

db/migrate/20160706031054_move_admins_to_devise.rb
```diff
@@ -0,0 +1,46 @@
...
+
+    # Remove old authlogic field
+    remove_column :admins, :persistence_token
+
+    # Fields for Recoverable module
+    add_column :admins, :reset_password_token, :string, limit: 255
+    add_column :admins, :reset_password_sent_at, :timestamp
+
+    # Field for Rememberable module
+    add_column :admins, :remember_created_at, :timestamp
```
bingeling: If we don't use the rememberable module, I think this column (and also line 41) can be removed from the migration.

sarken:
I think there's a problem -- the tests are failing. The relevant bit in Travis is:
```
==  MoveAdminsToDevise: migrating =============================================
-- rename_column(:admins, :crypted_password, :encrypted_password)
   -> 0.0137s
-- rename_column(:admins, :salt, :password_salt)
   -> 0.0132s
-- remove_column(:admins, :persistence_token)
   -> 0.0105s
-- add_index(:admins, {:unique=>true})
rake aborted!
StandardError: An error has occurred, all later migrations canceled:
Mysql2::Error: Key column '{:unique=>true}' doesn't exist in table: CREATE  INDEX `index_admins_on_{:unique=>true}` ON `admins` (`{:unique=>true}`)
```

sarken:
I came here to merge this, but it seems to have acquired conflicts /o\ Can you update, please?

PR #2711, opened 2017-01-15
AO3-4808 Fix 500 error when a work is edited to have no fandom

ariana-paris:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4808

## Purpose

Following the PR for [AO3-4549](https://otwarchive.atlassian.net/browse/AO3-4549), which fixed a problem stopping the user from viewing a work that was saved with no fandom, this stops a user from saving a work with no fandom in the first place.

This also takes the opportunity to refactor the edit and create methods as they had diverged (creating this problem) and with them both having 100% coverage, this seemed a safe thing to do.

## Testing

1. Edit a work that was previously posted
2. Hit the "x" next to the only fandom tag to remove it
3. Choose "Post Without Preview"
4. Now instead of a 500 error (or a 404 on Staging), the form should reload with the error message:
```
Sorry! We couldn't save this work because:
Please add all required tags. Fandom is missing.
```

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -298,58 +298,66 @@ class WorksController < ApplicationController
 
   # POST /works
   def create
+    set_edit_form_fields
+
+    @work.ip_address = request.remote_ip
+    # If Edit or Cancel is pressed, bail out and display relevant form
+    if params[:edit_button]
+      render :new && return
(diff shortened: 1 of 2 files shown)


sarken:
james said he also looked at it, so that's good enough for me. Marking ready to merge.

PR #2713, opened 2017-01-16
AO3-4357  Allow downloads to be disabled

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4357

## Purpose

This pull removes the option to download works. ( we the caveat that the option still may be in the header on a cached blurb )

This pull had to be reverted as it needed updating with strong params after being merged.


sarken:
Still looks good! However, I think it would be a good idea to add a test for it ASAP to avoid dragging down the coverage we're trying so hard to increase.

sarken:
Added a quick cuke to make sure unchecking the "allow downloads" option takes the download section off works.

bingeling:
Test looks good too

PR #2714, opened 2017-01-21
AO3-4303 fix BOT, english and missing subject

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4303

## Purpose

The subject line needs to be fixed, and so does the phrase "as an creator," as mentioned in the previous comment.

## Testing

See if the issue has been fixed.


config/locales/en.yml
```diff
@@ -63,7 +63,7 @@ en:
         part2_work: "%{edit} to remove yourself as creator if you've been added in error or don't want to be listed as a creator."
         part2_series: "%{edit} to remove yourself as creator if you've been added in error or don't want to be listed as a creator."
       text:
-        part2_work: "Edit the work to remove yourself as an creator if you've been added in error or don't want to be listed as a creator: %{url}"
+        part2_work: "Edit the work to remove yourself as a creator if you've been added in error or don't want to be listed as a creator: %{url}"
```
sarken: It should be just "as creator" (no "an" or "a") to match the text of the HTML email. Also, `part2_series` needs to be updated as well.

PR #2717, opened 2017-01-22
AO3-4837 Add test for collection_profile_controller

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4837

## Purpose

Add test for collection_profile_controller

## Testing

No manual testing required

### DIFF
--- a/spec/controllers/collection_profile_controller_spec.rb
+++ b/spec/controllers/collection_profile_controller_spec.rb
@@ -1,10 +1,13 @@
 require 'spec_helper'
 
 describe CollectionProfileController do
-
-  it "checks to see if the collection exists" do
-    get :show, collection_id: "A non existent collection"
-    expect(response).to redirect_to(collections_path)
-    expect(flash[:error]).to eq "What collection did you want to look at?"
-  end 
+  describe "GET #show" do
+    context "collection does not exist" do
+      it "redirects and provides an error message" do
+        get :show, collection_id: "A non existent collection"
+        expect(response).to redirect_to(collections_path)
+        expect(flash[:error]).to eq "What collection did you want to look at?"
+      end
+    end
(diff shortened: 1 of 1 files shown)


spec/controllers/collection_profile_controller_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'spec_helper'
+
+describe CollectionProfileController do
+
+  it "checks to see if the collection exists" do
```
sarken: I am not an expert, but I think it's generally consider good practice to describe the action, context, and expected behavior like: 

```
describe CollectionProfileController do
  describe "GET #show" do
    context "collection does not exist" do
      it "redirects and provides an error message" do
        get :show, collection_id: "A non existent collection"
        expect(response).to redirect_to(collections_path)
        expect(flash[:error]).to eq "What collection did you want to look at?"
      end
    end
  end
end
```

So if someone were to extend this test, they could do:
```
describe CollectionProfileController do
  describe "GET #show" do
    context "collection does not exist" do
      it "redirects and provides an error message"
        get :show, collection_id: "A non existent collection"
        expect(response).to redirect_to(collections_path)
        expect(flash[:error]).to eq "What collection did you want to look at?"
      end
    end

    context "collection exists" do
      let(:collection_profile) { FactoryGirl.create etc }
      it "renders the show template" do
        get :show, collection_id: collection_profile.something
        expect(response).to render_template("show")
      end
    end
  end

  describe "PUT #update" do
    context "attributes are valid" do
      it "changes @collection_profile's attributes" do
        expect something
      end
    end
  end

end
```

sarken:
The rspecs have passed on Travis, so adding the ready to merge label! We can hit the button once everything's done running.

PR #2718, opened 2017-01-24
AO3-4821 SkinsController Strong Parameters

dondenoncourt:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4821

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,10 @@ gem 'test-unit', '~> 3.0'
 gem 'bundler'
 
 gem 'rails', '3.2.22.3'
-gem 'strong_parameters'
+
+# the published gem does not include fixes that are in Rails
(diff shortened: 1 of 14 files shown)


Gemfile
```diff
@@ -6,7 +6,10 @@ gem 'test-unit', '~> 3.0'
 gem 'bundler'
 
 gem 'rails', '3.2.22.3'
-gem 'strong_parameters'
+
+# the published gem does not include fixes that are in Rails
+# specifically https://github.com/rails/strong_parameters/issues/16
+gem 'strong_parameters', github: 'rails/strong_parameters'
```
zz9pzza: Looking at https://github.com/rubysec/bundler-audit/issues/66

This implies that we should be using https for the source ?

PR #2720, opened 2017-01-24
AO3-4827 Count the contents of all chapters when creating a work

ariana-paris:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4827

## Purpose

We recently noticed that multi chapter imports had a word count that only included the first chapter when imported without preview. Any action that resaves the work after initial posting corrects the word count, but this isn't something Open Doors want to have to do for automated imports. This pull request fixes that situation by adding up all the word counts for all the chapters on a work when it is first posted.

## Testing

See the Jira ticket

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -748,9 +748,13 @@ class Work < ActiveRecord::Base
   # Set the value of word_count to reflect the length of the chapter content
   # Called before_save
   def set_word_count
-    self.word_count = 0
-    chapters.each do |chapter|
-      self.word_count += chapter.set_word_count
+    if self.new_record?
+      self.word_count = 0
+      chapters.each do |chapter|
+        self.word_count += chapter.set_word_count
+      end
+    else
+      self.word_count = Chapter.select("SUM(word_count) AS work_word_count").where(:work_id => self.id, :posted => true).first.work_word_count
     end
   end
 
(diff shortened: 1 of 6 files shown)


app/models/work.rb
```diff
@@ -748,7 +748,10 @@ def chapter_total_display
   # Set the value of word_count to reflect the length of the chapter content
   def set_word_count
     if self.new_record?
-      self.word_count = self.chapters.first.set_word_count
+      self.word_count = 0
```
zz9pzza: Could we use the same code on both sides of the if statement and get rid of it completely ?

ariana-paris: I initially thought not as the original method relies on the chapters being in the database, which they're not in this `before_save` method if the work is brand new (has never been saved). Using only the code I've added, which uses the counts from the chapters associated with the work in memory, made one of the chapter tests fail, but that turned out to be because it was possible for a chapter to be saved with a nil word count. Once I fixed _that_ bug, both creating and editing works appears to return the right word counts. So once I've written some more tests around word counts, as I don't think we're covering everything, I think this will be ready to go. (I've set to Needs coder action)

zz9pzza: \o/ I was hoping you would choose the new path too :) +1

ariana-paris: Alas, it turned out that my initial hunch was correct. There appear to be two scenarios - one where the chapters are only stored in memory, and one where they are apparently not in memory and their word count needs to be retrieved from the database. The second occurs when editing works that have been imported: the word counts on those broke when I removed the line that does a database SUM, so I've had to put it back.

PR #2726, opened 2017-01-29
AO3-4765 Make Notice and Notes class the same width

pixelred:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4765

## Purpose

makes forms, notices, and notes class the same width.

## Testing

make sure that on pages like collections creations page or challenge editing page, the blue boxes are the same width as the grey form field. these are css-only changes.


public/stylesheets/site/2.0/07-interactions.css
```diff
@@ -21,15 +21,15 @@ fieldset {
...
+/*Guideline: Forms are written by lots of people, and, this is beta, change a LOT.
+I've written a load of possible different nests to give a reasonably/broadly consistent view whatever goes in,
 but it's probably not complete. */
 
 fieldset, form dl, fieldset dl dl, fieldset fieldset fieldset, fieldset fieldset dl dl, dd.hideme, form blockquote.userstuff {
   display: block;
   background: #ddd;
   border: 2px solid #f3efec;
-  margin: 0.643em;
+  margin: 0.643em auto;
```
sarken: I'm afraid this isn't a good idea. This doesn't _just_ affect the outermost edges of a form -- it also reduces the left and right margins on all sorts of things _inside_ forms, leading to a cramped or unbalanced appearance, depending on the form. 

For example, here's a prompt meme form with this change:
margin_auto

And here it is before this change:
margin_0643

The new version cuts back on the amount of empty space between the edges of the white portion of the form and the edges of the grey portion. 

In general, I'm not sure the notices and the forms _should_ be the same width. With the notices wider than the forms, it's an additional, subtle reinforcement of the relationship between the two elements, saying "this notice encapsulates everything beneath it."

pixelred: ahh, alright. if it's a feature and not a bug then perhaps this pull is unnecessary?

sarken: I think you're right. Sorry for not catching it when the issue was created! I'm going to close the issue and this pull request, but thank you for taking a shot at it.

PR #2728, opened 2017-01-31
AO3-4830 Extend tests for tag_set_nominations_controller

cyrilcee:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4830

## Purpose

This is my first code change to the otwarchive repository, so I'm hoping someone can help set me straight on any otw style guidelines or best practices that I've disregarded!

This is to increase code coverage of tag_set_nominations_controller.rb. Code coverage in coveralls will increase from [68.34%](https://coveralls.io/builds/9916720/source?filename=app%2Fcontrollers%2Ftag_set_nominations_controller.rb) to [96.98%](https://coveralls.io/builds/9900548/source?filename=app%2Fcontrollers%2Ftag_set_nominations_controller.rb). \o/

Previously, tag_set_nominations_controller.rb was covered primarily with Cucumber tests. This code change will add additional unit test coverage.

Other open TODO items I'd like some help with:
There were 3 error handling paths in the controller that I wasn't sure how to set up to test:
1. In TagSetNominationsController.load_tag_set:22 - Is it possible to test the case where OwnedTagSet.find returns a falsey value, but doesn't raise an ActiveRecord error?
2. In TagSetNominationsController.load_nomination:30 - Is it possible to test the case where TagSetNomination.find returns a falsey value, but doesn't raise an ActiveRecord error?
3. In TagSetNominationsController#index:82 - Is it possible to test the case where OwnedTagSet.find returns a falsey value, but doesn't raise an ActiveRecord error?

~~I also wasn't sure what params would trigger the flash notice in TagSetNominationsController.request_noncanonical_info. I did set up some specs to test this (line 825 and 992), but I'd have preferred to use params that could actually be triggered by a real user through the UI. If someone knows better params to use here to test this code path, that would be very great.~~ Addressing this in [AO3-4856](https://otwarchive.atlassian.net/browse/AO3-4856).

## Testing

2. In the "Search" field near the "Source Files on Master" heading, enter the name of the controller: tag_set_nominations_controller.rb
3. Note the number in the "Coverage" column of the table

## Credit

cyrilcee

she/her

### DIFF
--- a/spec/controllers/tag_set_nominations_controller_spec.rb
+++ b/spec/controllers/tag_set_nominations_controller_spec.rb
@@ -19,13 +19,12 @@ describe TagSetNominationsController do
 
   let(:random_user) { FactoryGirl.create(:user) }
 
-  #TODO: Create some tag_set_nomination and tag_nomination factories?
-
   describe 'GET index' do
     context 'user is not logged in' do
       it 'redirects and returns an error message' do
         get :index, user_id: moderator.login, tag_set_id: owned_tag_set.id
-        it_redirects_to_with_error(new_user_session_path, "Sorry, you don't have permission to access the page you were trying to reach. Please log in.")
+        it_redirects_to_with_error(new_user_session_path, "Sorry, you don't have permission to access the page you " +
+            "were trying to reach. Please log in.")
       end
     end
 
@@ -39,8 +38,8 @@ describe TagSetNominationsController do
 
         context 'user_id does not match logged in user' do
           it 'redirects and returns an error message' do
-            get :index, user_id: "invalid", tag_set_id: owned_tag_set.id
-            it_redirects_to_with_error(tag_sets_path, "You can only view your own nominations, sorry.")
+            get :index, user_id: 'invalid', tag_set_id: owned_tag_set.id
+            it_redirects_to_with_error(tag_sets_path, 'You can only view your own nominations, sorry.')
           end
         end
 
@@ -50,11 +49,11 @@ describe TagSetNominationsController do
           end
 
           it 'renders the index template' do
-            expect(response).to render_template("index")
+            expect(response).to render_template('index')
           end
 
           it 'returns expected tag set nominations' do
-            expect(assigns(:tag_set_nominations)).to include(tag_set_nomination)
+            expect(assigns(:tag_set_nominations)).to eq([tag_set_nomination])
           end
         end
       end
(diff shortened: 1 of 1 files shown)


sarken:
I think you're right that request_noncanonical_info can't be triggered through the UI -- it definitely wouldn't be the first time we've run across code we couldn't reach.

spec/controllers/tag_set_nominations_controller_spec.rb
```diff
@@ -0,0 +1,1547 @@
+require 'spec_helper'
+
+describe TagSetNominationsController do
+  include LoginMacros
+  include RedirectExpectationHelper
+
+  let(:tag_set_nomination) { FactoryGirl.create(:tag_set_nomination) }
+  let(:owned_tag_set) { tag_set_nomination.owned_tag_set }
+
+  let(:tag_nominator) { tag_nominator_pseud.user }
```
shalott: Minor note, these definitions are in reverse order, which is confusing to follow -- ie, would prefer to have tag_nominator_pseud defined first, followed by tag_nominator, ditto for mod_pseud before moderator.

spec/controllers/tag_set_nominations_controller_spec.rb
```diff
@@ -0,0 +1,1547 @@
...
+            let(:user) { random_user }
+
+            it 'redirects and returns an error message' do
+              get :index, tag_set_id: owned_tag_set.id
+              it_redirects_to_with_error(tag_sets_path, "You can't see those nominations, sorry.")
+            end
+          end
+        end
+
+        # TODO: how can OwnedTagSet.find not raise an error but still return falsey?
```
shalott: I think this is probably hitting the error message in load_tag_set rather than in load_nomination, so the error should be "What tag set did you want to nominate for?" ?

spec/controllers/tag_set_nominations_controller_spec.rb
```diff
@@ -0,0 +1,1547 @@
...
+    context 'user is logged in' do
+      context 'invalid params' do
+        before do
+          fake_login_known_user(tag_nominator)
+        end
+
+        # TODO: how can OwnedTagSet.find not raise an error but still return falsey?
+        xcontext 'no tag set' do
+          it 'redirects and returns an error message' do
+            get :show, id: tag_set_nomination.id, tag_set_id: nil
```
shalott: Ditto here

spec/controllers/tag_set_nominations_controller_spec.rb
cyrilcee: 👍 Will update!

spec/controllers/tag_set_nominations_controller_spec.rb
cyrilcee: This is actually testing the final conditional case of the index action (TagSetNominationsController.rb:82).

spec/controllers/tag_set_nominations_controller_spec.rb
cyrilcee: This one already has the right error message to test load_tag_set, no?

For this case (and most of the other cases I marked with TODOs), I marked the tests as pending because I'm not sure how it can get to this code path. The code is set up basically with:
```ruby
if OwnedTagSet.find(blah)
  do_a_thing
else
  flash[:error] = "Oh no."
end
```

But my understanding is that if OwnedTagSet.find fails, it would raise an ActiveRecord:RecordNotFound error, and never make it into the else code path. Is that wrong? Do you have suggestions for testing the else case in this situation?

cyrilcee:
@sarken Can I delete request_noncanonical_info from the controller then? Is there a reason to leave it?

sarken:
@cyrilcee I think we can delete it from the code, but just to be on the safe side, let's keep the code changes separate from the test changes. You can make an issue like potatoesque did here: https://otwarchive.atlassian.net/browse/AO3-4845

cyrilcee:
@sarken - Created a new [JIRA](https://otwarchive.atlassian.net/browse/AO3-4856) issue and [PR](https://github.com/otwcode/otwarchive/pull/2742) for removing request_noncanonical_info.

spec/controllers/tag_set_nominations_controller_spec.rb
cyrilcee: @shalott (Pinging you in case you didn't see my reply earlier!) 

There's a couple other places where I ran into this issue too, which I listed in the summary of this PR. I'd like to remove any code in tag_set_nominations_controller.rb that we aren't actually able to hit (I can do this in a separate branch), but I want to make sure I'm not missing something obvious.

spec/controllers/tag_set_nominations_controller_spec.rb
shalott: Oh right, sorry, you were getting the index. Looks like a code error then, and I have confirmed it is breaking on the actual archive. I would submit a bug report & then change the TODO to say "this is failing due to bug #"

spec/controllers/tag_set_nominations_controller_spec.rb
shalott: (Sorry for taking so long to reply, had sequence of self and child sick with cold/stomach flu.)

Yes, after I poked a bit more, I see you're right, the find is raising an error and we're just getting the 404 not found page. Which is not a bad thing necessarily but is also not what's intended in the code. I could go for either solution, catching the raised error so the more informative error message you're testing for here does get delivered & the test works, or stripping out the unreachable sections and testing that you get an expected 404 error. Either way the code itself needs to be fixed so I would say, write the test for whichever behavior you prefer (I slightly incline towards stripping the code myself), then submit the issue report and comment out the test with the issue #

shalott:
My small notes aside, this is nevertheless a big improvement over the current state and IMO can be merged even if the changes aren't made so am marking it Ready to Merge! @cyrilcee if you want to make the changes first and have review again, just @ me and will be happy to take another look.

spec/controllers/tag_set_nominations_controller_spec.rb
cyrilcee: Thanks for taking a look! I'll go ahead and add a JIRA issue to address this and the above issue. I'll probably pick it up next week, assuming no one else beats me to it.

PR #2731, opened 2017-02-01
AO3-4305 fix issues found on test

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4305

## Purpose

- All text emails need their spacing fixed -- there should be a blank line between paragraphs
- The pseud doesn't show up in the text email that's delivered to the user who did not delete the work
- In the text emails, there should probably be the word "at" between "Support" and the URL, or the URL should be in parentheses or something 
- In the text emails, it would also be nice if we put quotation marks around the work title, like we do in the email for the work being deleted by an admin


sarken:
Travis failed on something unrelated. I restarted it, but upon further consideration, I'm just going to merge this.

PR #1975, opened 2015-01-01
AO3-4159 - Add last login information to user 

zz9pzza:
https://code.google.com/p/otwarchive/issues/detail?id=4167
https://otwarchive.atlassian.net/browse/AO3-4159

PR #2732, opened 2017-02-02
AO3-4851 Update schema and structure files

scottsds:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4851

## Purpose

Updates are needed to our structure.sql and schema.rb files

## Testing

No testing needed.


sarken:
@zz9pzza is the auto_increment stuff here correct? If so, I think this is good to go

zz9pzza:
Doing this is painful....

I believe we should drop all the AUTO_INCREMENT's eg
AUTO_INCREMENT=484751139

For each of the migrations you have included you can delete the file associated.
eg
INSERT INTO schema_migrations (version) VALUES ('20151130183602');
means that

db/migrate/20151130183602_add_download_enabled_to_admin_settings.rb 

Can be deleted safely.

I recommend not making the following change as it will mean that any migrations created before march 2016 would need to have their number changed.

-ActiveRecord::Schema.define(:version => 20150901132832) do
 +ActiveRecord::Schema.define(:version => 201604030319571) do

sarken:
I think it's fine if we need to update migrations created before March 2016 -- that only applies to #1975, which is for next release, so we won't forget. (The other pull requests with migrations are: #2438 dated April 2016, #4622 dated August 2016, and #2638 dated November 2016.) 

The `AUTO_INCREMENT`s didn't look right so me, so I'm glad I asked, but... any idea why they're there?

zz9pzza:
The AUTO_INCREMENTs are based upon the number of rows that that databases has contained.

So the database this was generated has had some use in it's time :)

sarken:
Ooooooh. That... certainly would explain a problem or two that I've had using our schema.

scottsds:
So what do you both want me to do with this PR?

sarken:
I think you can just delete the `AUTO_INCREMENT`s  and remove all the migration files that are no longer needed

scottsds:
Do you mean physically edit the file and remove the lines about auto increment? Or will deleting my migration files (the actual .rb files) and then rerunning the two commands automatically fix this?

zz9pzza:
You end up having to do it by hand.

scottsds:
That's fine, I can make the changes here on Git then!

scottsds:
@zz9pzza and @sarken I have removed the auto increment stuff. Let me know if this needs anything else.

zz9pzza:
you need to git rm the migrations we no longer need.

scottsds:
Done!

sarken:
I think the schema's going to need to be updated again once we've deployed Devise, which had migrations.

scottsds:
@sarken @zz9pzza I ran the devise migrations and then did the schema and structure dump again. I'm assuming I need to delete all the AUTO INCREMENT stuff again?

PR #2715, opened 2017-01-21
AO3-4726 Extend tests for the comments controller

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4726

## Purpose

Extend coverage for the comments controller.

## Testing

No manual testing required

PR #2716, opened 2017-01-22
AO3-4836 500 error when approving moderated comments on work you didn't create

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4836

## Purpose

Testing the comments controller though up an issue and this should clean it up. Merge after #2734 (which was previously #2715)

## Testing

Refer to Jira, although this is also covered by an automated test.

PR #2734, opened 2017-02-05
AO3-4726 extend tests for the comment controller

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4726

## Purpose

Extend coverage for the comments controller. These are the two files rescued from https://github.com/otwcode/otwarchive/pull/2715

## Testing

No manual testing required


sarken:
@zz9pzza here's the comment from the previous pull request, as requested:

**tl;dr** I refactored this quite a bit and you can see the results at  https://github.com/otwcode/otwarchive/compare/master...sarken:AO3-4726?expand=1  -- please feel free to pull them in (and make more changes if you want, since I'm sure there's stuff that could be done better). I've added a bit more coverage, too.

However! I'm also going to walk through the process I went through, since it's more helpful to explain that just say "here, use this." 

This describe was the very first thing I noticed -- with controllers, it's really a good idea to break it down by action, like:

```
  describe "PUT #review_all" do
    xit "does x, y, z" do
    end
  end

  describe "PUT #approve" do
    it "does x, y, z" do
    end
  end

  describe "GET #hide_comments" do
    it "does x, y, z" do
    end
  end
```

However, before starting to rearrange, it might be a good idea to tackle the creation of unreviewed comments (and other objects):
```
 +      comment = FactoryGirl.create(:comment)
 +      comment.unreviewed = true
 +      comment.save
```
You _should_  be able to set the specific attribute on a comment by doing `comment = FactoryGirl.create(:comment, attribute: thing)`, but that doesn't work with the initial creation of the comment because the default _work_ that gets created for the comment doesn't have moderated commenting enabled. So we need a new factory:
```
  factory :unreviewed_comment, class: Comment do
    name { Faker::Name.first_name }
    content { Faker::Lorem.sentence(25) }
    email { Faker::Internet.email }
    commentable_type { "Work" }
    commentable_id { FactoryGirl.create(:work, moderated_commenting_enabled: true).id }
    pseud
    unreviewed true
  end
```
Then we can change all of those three-line instances to just `comment = create(:unreviewed_comment)`.

The next thing is just plain comments, where it's currently:
```
 +      comment = FactoryGirl.create(:comment)
 +      comment.save
```
The `save` is unnecessary -- creating it saves it -- and so is the `FactoryGirl`. So all of those should just become `comment = create(:comment)`. All of the other instances of `FactoryGirl.create` can become just `create`, too, and that will leave a few more things that can be refactored, e.g.:
```
 +      pseud = @current_user.default_pseud
 +      comment = FactoryGirl.create(:comment)
 +      comment.unreviewed = true
 +      comment.pseud_id = pseud.id
 +      comment.save
```
That can be just `comment = create(:unreviewed_comment, pseud_id: @current_user.default_pseud.id)`

It's probably a good idea to run the tests here just to make sure everything worked, and _then_ move on to rearranging it by controller action.

After rearranging it by controller action, the descriptions could use a rewrite. A lot of them are just too vague and big-picture -- they're more Cucumber-style than RSpec. Like Ariana said in chat, you really want to describe what the very specific lines of code you're testing are doing. For example:
```
  describe "POST #new" do
    it "errors if the commentable is not a valid tag" do
      post :new, tag_id: "Non existent tag"
      expect(flash[:error]).to eq "What did you want to comment on?"
    end

    it "renders the :new template if commentable is a valid admin post" do
      admin_post = create(:admin_post)
      post :new, admin_post_id: admin_post.id
    end

    it "renders the :new template if commentable is a valid tag" do
      fandom = create(:fandom)
      post :new, tag_id: fandom.name
    end

    it "renders the :new template if commentable is a valid comment" do
      comment = create(:comment)
      post :new, comment_id: comment.id
    end
  end
```
might be more accurate than the current examples, which say things like `create a comment on an admin post` -- we're not actually _creating_ anything with the new action, though. It just renders the form. `create` is the action that handles the actual making of things.

There were a couple examples that tested too many things at once and needed to be broken down, for example: 
```
 +    it "can cancel a comment reply" do
 +      comment = FactoryGirl.create(:comment)
 +      comment.save
 +      get :cancel_comment_reply, comment_id: comment.id
 +      expect(flash[:error]).to be_nil
 +      expect(response).to redirect_to(comment_path(comment, anchor: "comments"))
 +      get :cancel_comment_reply, comment_id: comment.id, show_comments: 'yes', random_option: 'no'
 +      expect(flash[:error]).to be_nil
 +      expect(response).to redirect_to(comment_path(comment, show_comments: 'yes', anchor: "comments"))
 +    end
```
The fact that you're calling the `cancel_comment_reply` twice there is a good hint it should be two things, like maybe:
```
  describe "GET #cancel_comment_reply" do
    let(:comment) { create(:comment) }

    context "with only valid params" do
      it "redirects to comment path with the comments anchor and without an error" do
        get :cancel_comment_reply, comment_id: comment.id
        expect(response).to redirect_to(comment_path(comment, anchor: "comments"))
      end
    end

    context "with valid and invalid params" do
      it "removes invalid params and redirects without an error to comment path with valid params and the comments anchor" do
        get :cancel_comment_reply, comment_id: comment.id, show_comments: 'yes', random_option: 'no'
        expect(response).to redirect_to(comment_path(comment, show_comments: 'yes', anchor: "comments"))
      end
    end
  end
```
Which brings us to the next thing! If you find yourself repeating things when you write your descriptions (or if you think someone else might want to test something _kind_ of similar), or if your descriptions are getting too long, you can start nesting `describe` or `context` (`context` is good mainly if your description starts with `when` or `with`; describe is good for other things) to make things a bit more organized. It will also let you make things a little DRYer -- you can use a `before` or a `let` for things that are repeated across several examples, e.g.:
```
  describe "GET #add_comment_reply" do
    context "when comment is unreviewed" do
      let(:comment) { create(:unreviewed_comment) }

      it "redirects logged out user to login path with an error" do
        get :add_comment_reply, comment_id: comment.id
        expect(response).to redirect_to(login_path)
        expect(flash[:error]).to eq "Sorry, you cannot reply to an unapproved comment."
      end

      it "redirects logged in user to root path with an error" do
        fake_login
        get :add_comment_reply, comment_id: comment.id
        expect(response).to redirect_to(root_path)
        expect(flash[:error]).to eq "Sorry, you cannot reply to an unapproved comment."
      end
    end

    context "when comment is not unreviewed" do
      let(:comment) { create(:comment) }

      it "redirects to the comment on the commentable without an error" do
        get :add_comment_reply, comment_id: comment.id
        expect(response).to redirect_to(work_path(comment.ultimate_parent, show_comments: true, anchor: "comment_#{comment.id}"))
      end

      it "redirects to the comment on the commentable with the reply form open and without an error" do
        get :add_comment_reply, comment_id: comment.id, id: comment.id
        expect(response).to redirect_to(work_path(comment.ultimate_parent, add_comment_reply_id: comment.id, show_comments: true, anchor: "comment_#{comment.id}"))
      end
    end
  end
```
After all that, it seemed like a good idea to expand some of the more obvious examples a tiny bit, like making sure destroy actually deletes the comment and approve actually changes the approved status. It also occurred to me that we have `RedirectExpectionHelper`, which let me use a single line to replace things like:
```
 +      expect(response).to redirect_to(root_path)
 +      expect(flash[:error]).to eq "Sorry, you cannot reply to an unapproved comment."
```
And, lastly, I realized I could DRY it up a whole lot more by setting `comment` and `unreviewed` comment at the top instead of creating them wherever they were needed:
```
  let(:comment) { create(:comment) }
  let(:unreviewed_comment) { create(:unreviewed_comment) }
```
There is probably more stuff that could be done, since I am not an expert, but that's some stuff to keep in mind for next time! http://betterspecs.org has some good general guidelines, and https://everydayrails.com/2012/04/07/testing-series-rspec-controllers.html is helpful because it lets you see controller specs slowly come together.

spec/controllers/comments_controller_spec.rb
```diff
@@ -0,0 +1,247 @@
...
+
+      it "redirects to the comment on the commentable with the reply form open and without an error" do
+        get :add_comment_reply, comment_id: comment.id, id: comment.id
+        expect(flash[:error]).to be_nil
+        expect(response).to redirect_to(work_path(comment.ultimate_parent, add_comment_reply_id: comment.id, show_comments: true, anchor: "comment_#{comment.id}"))
+      end
+    end
+  end
+
+  describe "GET #unreviewed" do
```
annelu: Could we add a test for this action where the `comment_id` param doesn't correspond to an actual comment, so that it hits the `if @commentable.nil?` block on line 165 in comments_controller?

spec/controllers/comments_controller_spec.rb
```diff
@@ -0,0 +1,247 @@
...
+  describe "POST #new" do
+    it "errors if the commentable is not a valid tag" do
+      post :new, tag_id: "Non existent tag"
+      expect(flash[:error]).to eq "What did you want to comment on?"
+    end
+
+    it "renders the :new template if commentable is a valid admin post" do
+      admin_post = create(:admin_post)
+      post :new, admin_post_id: admin_post.id
+      expect(response).to render_template("new")
```
annelu: Could we check that the `assigns(:name)` is set correctly for each work?

spec/controllers/comments_controller_spec.rb
```diff
@@ -0,0 +1,247 @@
...
+
+  describe "GET #cancel_comment_edit" do
+    it "redirects to the comment on the commentable without an error" do
+      get :cancel_comment_edit, id: comment.id
+      expect(flash[:error]).to be_nil
+      expect(response).to redirect_to(work_path(comment.ultimate_parent, show_comments: true, anchor: "comment_#{comment.id}"))
+    end
+  end
+
+  describe "GET #destroy" do
```
annelu: Could we have another test where `@comment.destroy_or_mark_deleted` is stubbed to return false, so we can hit lines 301-302?

spec/controllers/comments_controller_spec.rb
zz9pzza: We don't get that we would throw a 500 for an unfound record ?

Failures:

  1) CommentsController GET #unreviewed On an invalid id it errors
     Failure/Error: @commentable = Comment.find(params[:comment_id])

     ActiveRecord::RecordNotFound:
       Couldn't find Comment with id=999999
     # /home/vagrant/.rvm/gems/ruby-2.2.5/gems/activerecord-3.2.22.3/lib/active_record/relation/finder_methods.rb:344:in `find_one'
     # /home/vagrant/.rvm/gems/ruby-2.2.5/gems/activerecord-3.2.22.3/lib/active_record/relation/finder_methods.rb:315:in `find_with_ids'
     # /home/vagrant/.rvm/gems/ruby-2.2.5/gems/activerecord-3.2.22.3/lib/active_record/relation/finder_methods.rb:107:in `find'
     # /home/vagrant/.rvm/gems/ruby-2.2.5/gems/activerecord-3.2.22.3/lib/active_record/querying.rb:5:in `find'
     # ./app/controllers/comments_controller.rb:132:in `load_commentable'

The fix might be:

https://github.com/otwcode/otwarchive/pull/2716/files but I am not feeling well enough to be sure.

PR #2737, opened 2017-02-05
AO3-4853 Improve cache busting for a single comment. ( Make tests and coverage more reliable )

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4853

## Purpose

This should both make the tests more reliable and also make so we are less likely to have comments not appearing.

## Testing

Create a comment and ensure it is visible.

### DIFF
--- a/app/views/comments/_single_comment.html.erb
+++ b/app/views/comments/_single_comment.html.erb
@@ -14,7 +14,7 @@
     <% elsif single_comment.hidden_by_admin? && !logged_in_as_admin? %>
       <p class="message"><%= ts("(This comment is under review by an admin and is currently unavailable.)") %></p>
     <% else %>
-      <% cache(single_comment, :expires_in => 1.week ) do %>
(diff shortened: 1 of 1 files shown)


sarken:
This looks good! However, now that you've investigated the failing test, found the cause, and fixed it, could you update the JIRA issue to describe the code change? (Someone who is wondering, "Huh, when and why did we change the comment cache key?" should be able to find the answer by searching JIRA.)

app/views/comments/_single_comment.html.erb
```diff
@@ -14,7 +14,7 @@
     <% elsif single_comment.hidden_by_admin? && !logged_in_as_admin? %>
       <p class="message"><%= ts("(This comment is under review by an admin and is currently unavailable.)") %></p>
     <% else %>
-      <% cache(single_comment, :expires_in => 1.week ) do %>
+      <% cache(key: "/v1/single_comment/#{single_comment.id}/#{single_comment.edited_at}/#{single_comment.updated_at}", expires_in: 1.week ) do %>
```
sarken: If you have a second before we get around to merging this, I just realized there's an extra space before the `)` that could use removal.

PR #2740, opened 2017-02-08
AO3-4835 Fix unreachable branch in owned_tag_sets_controller

annelu:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4835

## Purpose

Previously when tag_set is not found by load_tag_set, it would raise an ActiveRecord error and return a 500. Now, it redirects to the tag set path and displays a flash notice (the previously unreachable branch).

## Testing

Navigate to /tag_sets/{some tag set id that doesn't exist} (i.e. /tag_sets/12345). It should redirect to /tag_sets and have a flash notice that says "What Tag Set did you want to look at?"

## Credit

potatoesque, she/her

### DIFF
--- a/spec/controllers/owned_tag_sets_controller_spec.rb
+++ b/spec/controllers/owned_tag_sets_controller_spec.rb
@@ -104,6 +104,13 @@ describe OwnedTagSetsController do
   end
 
   describe "show" do
+    context "where tag set is not found" do
+      it "redirects and displays a notice" do
+        get :show, id: 12345
+        it_redirects_to_with_notice(tag_sets_path, "What Tag Set did you want to look at?")
+      end
+    end
+
     context "where tag set is found" do
       let(:visible) { false }
       let(:tag) { create(:character) }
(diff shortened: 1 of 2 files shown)


spec/controllers/owned_tag_sets_controller_spec.rb
```diff
@@ -104,6 +104,13 @@
   end
 
   describe "show" do
+    context "where tag set is not found" do
+      it "redirects and displays a notice" do
+        get :show, id: 12345
```
annelu: I'll do it if it's the convention, but it feels kind of silly?

sarken: I think it's fine as is -- the _ is more helpful if you're dealing with a number like 1000000000 and your need it to be easily read by humans. It doesn't serve much purpose when making up a random id for a test.

sarken:
Travis is still running, but this looks good to the human eye!

PR #2743, opened 2017-02-17
AO3-4858 Speed up posting, hopefully work around posting timing out.

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4858

## Purpose

Removes the rails automatic caching 

## Testing

Post works, I would not expect things to be quicker on test particularly as the slow down only happens when two or more people post to the same large tag.

Make sure that works turn up in tags as quickly or slowly as they did before.

I would deploy this with a bit of down time I suspect as I think I would be suspicious of the migration and two systems working in different ways.


app/models/tag.rb
```diff
@@ -23,6 +23,68 @@ class Tag < ActiveRecord::Base
   # the order is important, and it is the order in which they appear in the tag wrangling interface
   USER_DEFINED = ['Fandom', 'Character', 'Relationship', 'Freeform']
 
+  def is_this_a_large_tag(length_of_time_in_cache,how_long_it_takes_to_compute_size)
```
sarken: For succinctness, maybe something more like `def is_large_tag(time_in_cache, time_to_compute_size)`?

app/models/tag.rb
```diff
@@ -23,6 +23,68 @@ class Tag < ActiveRecord::Base
...
+    if length_of_time_in_cache < 40 || how_long_it_takes_to_compute_size < ( ArchiveConfig.TAGINGS_COUNT_DEFAULT_FOR_LARGE_TAGS || 4 )
```
sarken: I don't think there should be a space in these parentheses, but also, what's the significance of `4` here? And I think it should be `TAGGINGS` with two Gs, not `TAGINGS` with one G.

app/models/tag.rb
```diff
@@ -23,6 +23,68 @@ class Tag < ActiveRecord::Base
...
+        return
+      end
+    end
+    return if self.large_tag
+    self.large_tag = true
+    self.save
+  end
+
+  def taggings_count_expiry(count)
+    expiry_time = count/(ArchiveConfig.TAGINGS_COUNT_CACHE_DIVISOR || 2000 )
```
sarken: Two Gs in "taggings." Why are we setting a fallback of 2000 rather than some other number here?

app/models/tag.rb
```diff
@@ -23,6 +23,68 @@ class Tag < ActiveRecord::Base
...
+    [[expiry_time, (ArchiveConfig.TAGINGS_COUNT_MIN_TIME||3)].min, (ArchiveConfig.TAGINGS_COUNT_MIN_TIME||60)].max
```
sarken: Pretty much the same as above! "Taggings" should have two Gs and there should probably be a comment somewhere explaining why the fallback values are 3 and 60 here.

app/models/tag.rb
zz9pzza: How about check_if_large_tag as its not returning a value ?

app/models/tag.rb
```diff
@@ -23,6 +23,68 @@ class Tag < ActiveRecord::Base
...
+    end
+  end
+
+  def taggings_count
+    cache_read = Rails.cache.read(self.taggings_count_cache_key)
+    return cache_read unless cache_read.nil?
+    time_start = Time.now.to_i
+    real_value = self.taggings.length
+    time_end = Time.now.to_i
+    if (real_value > (ArchiveConfig.TAGINGS_COUNT_MIN_CACHE_COUNT || 1000)) || (time_end - time_start > (ArchiveConfig.TAGINGS_COUNT_MAX_ALLOWED_TIME || 6))
```
sarken: Same thing -- two Gs, and this could use an explanation of where the fallback values are coming from.

app/models/tag.rb
```diff
@@ -23,6 +23,68 @@ class Tag < ActiveRecord::Base
...
+        self.taggings_count_cache = real_value
+        self.save
+      end
+    end
+    real_value
+  end
+
+  def check_if_taggings_count_needs_updating
+    cache_read = Rails.cache.read(self.taggings_count_cache_key)
+    if cache_read.nil? || (cache_read < (ArchiveConfig.TAGINGS_COUNT_MIN_CACHE_COUNT || 1000))
```
sarken: Ditto the Gs and the need for an explanation.

features/tags_and_wrangling/tag_wrangling_more.feature
```diff
@@ -24,7 +24,7 @@ Feature: Tag wrangling: assigning wranglers, using the filters on the Wranglers
     Then I should see "Edit first fandom Tag"
     
     # assigning media to a fandom
-    When I fill in "Media" with "TV Shows"
+    When I fill in "tag[media_string]" with "TV Shows"
```
sarken: Just wondering why this needs to change. Also, as Ariana mentioned in our discussion, we're definitely going to want some tests to make sure that caching is working right.

zz9pzza: So this is because in the past the tag wasn't properly given the "Uncategorized Fandoms" so now it is taking tag[media_string] as the field if you look at 

view-source:http://test.archiveofourown.org/tags/Arrow%20(TV%202012)/edit for example

app/models/tag.rb
```diff
@@ -23,6 +23,69 @@ class Tag < ActiveRecord::Base
...
+    else
+      if self.taggings_count_cache != real_value
+        self.taggings_count_cache = real_value
+        self.save
+      end
+    end
+    real_value
+  end
+
+  def check_if_taggings_count_needs_updating
```
ariana-paris: This isn't checking if the count needs updating, it's actually updating the count if it needs updating. `update_taggings_count` might suffice - which in turn begs the question what this method is doing that isn't or couldn't be covered already in `taggings_count`

app/models/tag.rb
```diff
@@ -23,6 +23,69 @@ class Tag < ActiveRecord::Base
...
+    cache_read = Rails.cache.read(self.taggings_count_cache_key)
+    if cache_read.nil? || (cache_read < (ArchiveConfig.TAGGINGS_COUNT_MIN_CACHE_COUNT || 1000))
+      self.taggings_count
+      return
```
ariana-paris: You don't need to explicitly return here as the method doesn't do anything else

lib/tasks/tag_tasks.rake
```diff
@@ -38,7 +37,7 @@ namespace :Tag do
   desc "Delete unused tags"
   task(:delete_unused => :environment) do
     deleted_names = []
-    Tag.find(:all, :conditions => {:canonical => false, :merger_id => nil, :taggings_count => 0}).each do |t|
+    Tag.find(:all, :conditions => {:canonical => false, :merger_id => nil, :taggings_count_cache => 0}).each do |t|
```
ariana-paris: Good opportunity to kill some rocket hashes

spec/models/tag_spec.rb
```diff
@@ -11,6 +11,56 @@
     User.current_user = nil
   end
 
+  context "checking count caching" do
+
+    before(:each) do
+      # Set the minimail amount of time a tag can be cached for.
```
ariana-paris: `minimum`

spec/models/tag_spec.rb
```diff
@@ -11,6 +11,56 @@
...
+      ArchiveConfig.TAGGINGS_COUNT_MIN_TIME = 1
+      # Set so that we need few uses of a tag to start caching it.
+      ArchiveConfig.TAGGINGS_COUNT_CACHE_DIVISOR = 2
+      # Set the minimum number of uses needed for caching is possible
```
ariana-paris: should that be "if possible" ?

spec/models/tag_spec.rb
```diff
@@ -11,6 +11,56 @@
...
+      ArchiveConfig.TAGGINGS_COUNT_MIN_CACHE_COUNT = 3
+      @fandom_tag = FactoryGirl.create(:fandom)
+    end
+
+    it "A small tag should be uneffected by caching" do
```
ariana-paris: "unaffected". Also, the reason Rspec methods start with "it" is so it can produce a nice output with complete sentences like "it should not cache a small tag"

spec/models/tag_spec.rb
```diff
@@ -11,6 +11,56 @@
...
+      work = FactoryGirl.create(:work, fandom_string: @fandom_tag.name)
+      work.save
+      @fandom_tag.reload
+      expect(@fandom_tag.taggings_count_cache).to eq 1
+      expect(@fandom_tag.taggings_count).to eq 1
+      expect(@fandom_tag.large_tag).not_to be_truthy
+    end
+
+    it "A tag will start to be cached when its used" do
```
ariana-paris: "it's used" (short for "it is")

spec/models/tag_spec.rb
```diff
@@ -11,6 +11,56 @@
...
+      work.save
+      @fandom_tag.reload
+      # This value should be cached and wrong
+      expect(@fandom_tag.taggings_count_cache).to eq ArchiveConfig.TAGGINGS_COUNT_MIN_CACHE_COUNT+1
+      expect(@fandom_tag.taggings_count).to eq ArchiveConfig.TAGGINGS_COUNT_MIN_CACHE_COUNT+1
+      expect(@fandom_tag.large_tag).not_to be_truthy
+    end
+
+    it "A tag used enough will become a big tag" do
+      (1..40*ArchiveConfig.TAGGINGS_COUNT_CACHE_DIVISOR-1).each do |try|
```
ariana-paris: What Hound said

app/models/tag.rb
```diff
@@ -23,6 +23,80 @@ class Tag < ActiveRecord::Base
...
+    end
+  end
+
+  def taggings_count
+    cache_read = Rails.cache.read(self.taggings_count_cache_key)
+    return cache_read unless cache_read.nil?
+    time_start = Time.now.to_i
+    real_value = self.taggings.length
+    time_end = Time.now.to_i
+    # here we cache the value we have more than AGGINGS_COUNT_MIN_CACHE_COUNT uses or
```
ariana-paris: Typo: TAGGINGS...

spec/models/tag_spec.rb
```diff
@@ -11,6 +11,55 @@
...
+      work = FactoryGirl.create(:work, fandom_string: @fandom_tag.name)
+      work.save
+      @fandom_tag.reload
+      # This value should be cached and wrong
+      expect(@fandom_tag.taggings_count_cache).to eq ArchiveConfig.TAGGINGS_COUNT_MIN_CACHE_COUNT + 1
+      expect(@fandom_tag.taggings_count).to eq ArchiveConfig.TAGGINGS_COUNT_MIN_CACHE_COUNT + 1
+      expect(@fandom_tag.large_tag).not_to be_truthy
+    end
+
+    it 'that a tag is big depending on the number of uses it has' do
```
ariana-paris: `it "flags a tag as large based on its number of uses"` flows a bit better.

Also, as a general note, we're trying to standardise on double quotes rather than single quotes for strings.

spec/models/tag_spec.rb
```diff
@@ -11,6 +11,55 @@
...
+      # Set so that we need few uses of a tag to start caching it.
+      ArchiveConfig.TAGGINGS_COUNT_CACHE_DIVISOR = 2
+      # Set the minimum number of uses needed for before caching is started.
+      ArchiveConfig.TAGGINGS_COUNT_MIN_CACHE_COUNT = 3
+      @fandom_tag = FactoryGirl.create(:fandom)
+    end
+
+    it 'should not cache tags which are not used much' do
+      work = FactoryGirl.create(:work, fandom_string: @fandom_tag.name)
+      work.save
```
ariana-paris: I'm pretty certain the point of using `create` is that FactoryGirl will save the item and persist it to the database (as opposed to `build`, which only constructs it in memory). Also, you don't need to specify `FactoryGirl`, fyi.

app/models/tag.rb
```diff
@@ -23,6 +23,80 @@ class Tag < ActiveRecord::Base
   # the order is important, and it is the order in which they appear in the tag wrangling interface
   USER_DEFINED = ['Fandom', 'Character', 'Relationship', 'Freeform']
 
+  def check_if_large_tag(time_in_cache, time_to_count_sec)
+    # A large tag is one which is periodically refreshed by a background task.
+    # The task is defined in config/resque_schedule.yml
+    # This refreshes the large tags every 30 minutes. It seemed sensible to 
+    # only allow tags which stay in the cache longer than the perodic refresh
```
sarken: Do you mean "periodic refresh"?

ariana-paris:
A couple of typos which should be fixed but marking ready to merge so we can move quickly when we're ready.

PR #2746, opened 2017-02-19
AO3-4864 Cache media and fandom pages

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4864

## Purpose

This caches the fandom pages to reduce the strain on the servers when we migrate to the new tag.tagging_counts method later.

## Testing

Go to these pages and check the tagging counts:

http://test.archiveofourown.org/media
http://test.archiveofourown.org/media/Anime%20*a*%20Manga/fandoms
http://test.archiveofourown.org/media/Books%20*a*%20Literature/fandoms
http://test.archiveofourown.org/collections/MerlinSexstars/fandoms

Add a work in a fandom that is referenced, and see that the count is unaffected for an hour.

## References

This should be merged before:

https://github.com/otwcode/otwarchive/pull/2743

As I would like the counts on these pages to be cached before merging that code.


app/views/fandoms/index.html.erb
```diff
@@ -1,3 +1,6 @@
+<% collection_key = @collection.nil? ? "NULL" : @collection.id %>
+<% media_key = @medium.nil? ? "NULL" : @medium.name %>
+<% cache("/v1/media_pages/#{collection_key}/#{media_key}", :expires_in => 1.hour) do %>
```
sarken: Can you use the new syntax here? Also, when you do fragment caching like this, please update the indenting of the file to make it clearer what's inside the caching -- basically, everything between the start and end should move over 1 tab (2 spaces).

PR #2750, opened 2017-02-23
AO3-4844 BOT styling on admin login

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4844

## Purpose

Changes the css so that the admin login does not take the whole width of the screen.
I have not been able to test this however it comes from sarken and It looks right to me ( that vargrant image seems to have lost all the skins :( )

## Testing

We need to remember to load the site skins in before recaching all the skins again so this will not be visible when it is first deployed.


sarken:
(I think the issue number is AO3-4844: https://otwarchive.atlassian.net/browse/AO3-4844)

It looks like a stray `?` snuck in -- maybe a copy and paste error? Also, the selector should be just `.sessions-new .new-admin` without the `form`, since `.new-admin` refers to the form.

public/stylesheets/site/2.0/16-zone-system.css
```diff
@@ -219,7 +219,7 @@ SUBSECTIONS:
   width: 48%;
 }
 
-.admin_sessions-new form {
+.sessions-new .new-admin {
```
sarken: That should be an underscore for `.new_admin`, sorry! I'll try to remember to leave a second comment rather than an edit next time.

PR #2754, opened 2017-02-25
AO3-4879 Fix error when updating an external_author fails

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4879

## Purpose

Fix an error which I don't think happens very often ( and increase the test coverage by a small amount )

## Testing

I don't think this can be tested.


spec/controllers/external_authors_controller_spec.rb
```diff
@@ -0,0 +1,118 @@
+# frozen_string_literal: true
+require 'spec_helper'
+
+describe ExternalAuthorsController do
+  include LoginMacros
+  include RedirectExpectationHelper
+  let(:user) { FactoryGirl.create(:user) }
+
+  describe "GET #get_external_author_from_invitation" do
+    it "needs a valid invitaiton" do
```
sarken: You should be describing behavior in certain contexts, not requirements. So:
```
context "with invalid invitation token"
  it "redirects with an error"
```

spec/controllers/external_authors_controller_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+      invitation = FactoryGirl.create(:invitation)
+      get :get_external_author_from_invitation, invitation_token: invitation.token
+      it_redirects_to_with_error(signup_path(invitation.token), "There are no stories to claim on this invitation. Did you want to sign up instead?")
+    end
+  end
+
+  describe "GET #complete_claim" do 
+    it "should give the user the works" do
+      external_author = FactoryGirl.create(:external_author)
+      user = FactoryGirl.create(:user)
```
sarken: You shouldn't need this since you have `let(:user) { FactoryGirl.create(:user) }` at the top.

spec/controllers/external_authors_controller_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+      fake_login_known_user(user)
+      invitation = FactoryGirl.create(:invitation, external_author: external_author)
+      get :complete_claim, invitation_token: invitation.token
+      it_redirects_to_with_notice(user_external_authors_path(user), "We have added the stories imported under #{external_author.email} to your account.")
+    end
+  end 
+
+  describe "GET #update" do
+    it "needs to be done by the right user" do
```
sarken: `it` is for describing expected behavior, often in certain circumstances (`context`). It looks like this is testing that `GET #update` redirects and errors when the user is _not_ right, so it should be more like:
```
context "when user is not right" do
  it "redirects with an error"
```
However, "not right" is not informative. What does a "not right" user look like when we're updating an external author? Code-wise, it looks like `unless (@invitation && @invitation.external_author == @external_author) || @external_author.user == current_user`, but what are we testing here in English?

spec/controllers/external_authors_controller_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+      external_author = FactoryGirl.create(:external_author)
+      wrong_external_author = FactoryGirl.create(:external_author)
+      user = FactoryGirl.create(:user)
+      fake_login_known_user(user)
+      invitation = FactoryGirl.create(:invitation, external_author: wrong_external_author)
+      get :update, invitation_token: invitation.token, id: external_author.id
+      it_redirects_to_with_error(root_path, "You don't have permission to do that.")
+    end
+
+    it "can do nothing" do
```
sarken: Above, we have 'context "user is not right"`, so down here we should have all of these inside the opposite context -- `context "user is right"`. (We need something more descriptive than "right" here, too, though.)

Also, this isn't actually that it does nothing or orphans the works. It's only testing the redirect and success message. So for the first two, we're probably looking at something like:
```
context "when user is right"
  context "when user chooses to do nothing with imported works"
    it "redirects with a success message"
  context "when user chooses to orphan imported works"
    it "redirects with a success message"
```
But `it "errors if the preferences can't be saved" do` might throw a bit of a wrench into that. I'll take a more thorough look at that later, but I wanted to get you started with a review.

spec/controllers/external_authors_controller_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+      user = FactoryGirl.create(:user)
+      user.save
+      fake_login_known_user(user)
+      get :edit, id: external_author.id, user_id: user.login
+      expect(assigns(:external_author)).to eq(external_author)
+    end
+  end
+
+  describe "GET #index" do
+    it "redirects and gives a notice when not logged in" do
```
sarken: Perhaps:
```
context "when logged out"
  it "redirects with notice"
context "when logged in as user"
  context "without archivist permissions"
    it "assigns external_authors"
    it "redirects"
  context "with archivist permissions"
    it "assigns external_authors and renders index"
```

spec/controllers/external_authors_controller_spec.rb
sarken: I _think_ this accurately describes what you want to test:
```
context "when user is right"
  context "when doing nothing with imported works"
    it "redirects with a success message"
  context "when orphaning imported works"
    it "redirects with a success message"
    context "when updating preferences"
      it "renders edit template with a success message for orphaning and an error for preferences"
```

The only thing is, line 69 doesn't specify anything that's an attribute of `external_author`. Should it maybe have `do_not_email: true` to better resemble attempting to update preferences and orphan works at the same time? (Also, I don't think it makes any practical difference, but I think it should actually be `describe "PUT #update"` and `put :update`)

You might be also able to cut back on the repetition a little by pulling out `fake_login_known_user(user)` and putting it in a `before` block.

spec/controllers/external_authors_controller_spec.rb
```diff
@@ -0,0 +1,118 @@
+# frozen_string_literal: true
+require 'spec_helper'
+
+describe ExternalAuthorsController do
+  include LoginMacros
+  include RedirectExpectationHelper
+  let(:user) { FactoryGirl.create(:user) }
```
zz9pzza: This is because of the five days between starting and finishing this pull.

sarken:
The tests for this will need to be updated now that #2755 is merged.

sarken:
The specs failed on this after the merge, so I'm restarting them to see if they pass. If they fail, I think we have a new problem /o\

spec/controllers/external_authors_controller_spec.rb
```diff
@@ -114,7 +114,7 @@
         end
 
         context "when updating preferences" do
-          xit "renders edit template with a success message for orphaning and an error for preferences" do
+          it "renders edit template with a success message for orphaning and an error for preferences" do
             allow_any_instance_of(ExternalAuthor).to receive(:update_attributes).and_return(false)
             put :update, invitation_token: invitation.token, id: external_author.id, imported_stories: "orphan", do_not_email: true
```
sarken:
```
1) ExternalAuthorsController PUT #update when the user has permission through an invitation when orphaning imported works when updating preferences renders edit template with a success message for orphaning and an error for preferences
     Failure/Error:
       params.permit(
         :id, :user_id, :utf8, :_method, :authenticity_token, :invitation_token,
         :imported_stories, :commit, :remove_pseud,
         external_author: [
           :email, :do_not_email, :do_not_import
         ]
       )
     ActionController::UnpermittedParameters:
       found unpermitted parameters: do_not_email
```

spec/controllers/external_authors_controller_spec.rb
```diff
@@ -114,9 +114,9 @@
         end
 
         context "when updating preferences" do
-          xit "renders edit template with a success message for orphaning and an error for preferences" do
+          it "renders edit template with a success message for orphaning and an error for preferences" do
             allow_any_instance_of(ExternalAuthor).to receive(:update_attributes).and_return(false)
-            put :update, invitation_token: invitation.token, id: external_author.id, imported_stories: "orphan", do_not_email: true
+            put :update, invitation_token: invitation.token, id: external_author.id, imported_stories: "orphan"
```
sarken: I think this should be fixed the _other_ way -- that is, by making sure we can pass do_not_email parameters when they're correctly-formatted. I think that would look like `external_author: { do_not_email: true }`

sarken:
Merge conflicts

app/controllers/external_authors_controller.rb
```diff
@@ -78,13 +81,9 @@ def update
...
     if @external_author.update_attributes(external_author_params[:external_author] || {})
       flash[:notice] += "Your preferences have been saved."
-      if @user
-        redirect_to user_external_authors_path(@user)
-      else
-        redirect_to root_path
-      end
+      redirect_to @user ? user_external_authors_path(@user) : root_path
     else
-      flash[:error] += "There were problems saving your preferences."
```
redsummernight: The fix is here, the rest is style.

PR #2755, opened 2017-02-25
AO3-4869 Extend tests for external_authors controller

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4869

## Purpose

Increase coverage of external_authors controller

## Testing

No testing required.


spec/controllers/external_authors_controller_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+            expect(flash[:error]).to eq "There were problems saving your preferences."
+          end
+        end
+      end
+    end
+  end
+
+  describe "GET #edit" do
+    it "assigns external_author" do
+      user.save
```
sarken: Is `user.save` necessary here? It doesn't appear to be when I try to run these.

spec/controllers/external_authors_controller_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+      context "when orphaning imported works" do
+        it "redirects with a success message" do
+          put :update, invitation_token: invitation.token, id: external_author.id, imported_stories: "orphan"
+          it_redirects_to_with_notice(root_path, "Your imported stories have been orphaned. Thank you for leaving them in the archive! Your preferences have been saved.")
+        end
+
+        context "when updating preferences" do
+          xit "renders edit template with a success message for orphaning and an error for preferences" do
+            allow_any_instance_of(ExternalAuthor).to receive(:update_attributes).and_return(false)
+            put :update, invitation_token: invitation.token, id: external_author.id, imported_stories: "orphan"
```
sarken: Should this include a `do_not_import` or `do_not_email` param to more accurately reflect the process of orphaning works and updating preferences?

spec/controllers/external_authors_controller_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+  include RedirectExpectationHelper
+  let(:user) { FactoryGirl.create(:user) }
+  let(:invitation) { FactoryGirl.create(:invitation, external_author: external_author) }
+  let(:external_author) { FactoryGirl.create(:external_author) }
+
+  before(:each) do
+    fake_login_known_user(user)
+  end
+
+  describe "GET #get_external_author_from_invitation" do
```
sarken: In the previous version, these were testing what happened while logged out, but now they're testing the behavior while logged in. I'm not sure if that's important, but it makes me wonder if we should be testing both logged in and out behavior.

spec/controllers/external_authors_controller_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+require 'spec_helper'
+
+describe ExternalAuthorsController do
+  include LoginMacros
+  include RedirectExpectationHelper
+  let(:user) { FactoryGirl.create(:user) }
+  let(:invitation) { FactoryGirl.create(:invitation, external_author: external_author) }
+  let(:external_author) { FactoryGirl.create(:external_author) }
+
+  before(:each) do
```
sarken: This works, but I think it also would have been okay to stick with repeating `fake_login_known_user(user)` where you specifically wanted to be logged in. DRY is good, but it forces you to add things like `fake_logout` and the special `before` and `after` blocks for the archivist examples, it might be _too_ DRY.

spec/controllers/external_authors_controller_spec.rb
```diff
@@ -0,0 +1,136 @@
+# frozen_string_literal: true
+require 'spec_helper'
+
+describe ExternalAuthorsController do
+  include LoginMacros
+  include RedirectExpectationHelper
+  let(:user) { FactoryGirl.create(:user) }
```
zz9pzza: I understand that this could be done however in the same way as I prefer having self around. I prefer to have the FactoryGirl around so it is clear what is being called.

spec/controllers/external_authors_controller_spec.rb
zz9pzza: I can change it back if you want ?

spec/controllers/external_authors_controller_spec.rb
zz9pzza: There is no template for the fall though option so if just fails horribly.

spec/controllers/external_authors_controller_spec.rb
zz9pzza: I don't think it does I will remove it.

spec/controllers/external_authors_controller_spec.rb
sarken: A pending test shouldn't cause failures, and `put :update, invitation_token: invitation.token, id: external_author.id, imported_stories: "orphan", do_not_email: true` works fine with the fixed code in #2754

spec/controllers/external_authors_controller_spec.rb
sarken: Like I said, I'm really not sure what the best course is -- you might want to ask for a second opinion

spec/controllers/external_authors_controller_spec.rb
```diff
@@ -0,0 +1,135 @@
...
+        get :get_external_author_from_invitation, invitation_token: no_story_invitation.token
+        it_redirects_to_with_error(signup_path(no_story_invitation.token), "There are no stories to claim on this invitation. Did you want to sign up instead?")
+      end
+    end
+  end
+
+  describe "GET #complete_claim" do
+    it "redirects with a success message" do
+      get :complete_claim, invitation_token: invitation.token
+      it_redirects_to_with_notice(user_external_authors_path(user), "We have added the stories imported under #{external_author.email} to your account.")
```
annelu: Do you think it'd be useful for us to check that `@external_author.claim!` and `@invitation.mark_as_redeemed` have been called? I'm thinking that the logic for each respective method should probably be tested in the model tests, but ensuring that `complete_claim` calls them ensures that the controller is doing what it's supposed to.

zz9pzza: @annelu This doesn't work for me.
  describe "GET #complete_claim" do
    it "redirects with a success message" do
      expect(invitation).to receive(:mark_as_redeemed)
      expect(external_author).to receive(:claim!)
      get :complete_claim, invitation_token: invitation.token
      it_redirects_to_with_notice(user_external_authors_path(user), "We have added the stories imported under #{external_author.email} to your account.")
    end
  end

sarken: I think we should probably consider this PR ready to merge and make an issue for testing this when we figure it out.

sarken: Here's the new issue: https://otwarchive.atlassian.net/browse/AO3-4916

PR #2756, opened 2017-02-26
AO3-4877 Fix gift exchange sign-ups CSV downloads

ariana-paris:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4877

## Purpose

This fixes the bug introduced in release 0.9.180 which prevents gift exchange sign-ups from being downloaded as a CSV file.

## Testing

Try to download the CSV for a gift exchange. This was consistently broken but should now work.


spec/controllers/owned_tag_sets_controller_spec.rb
```diff
@@ -81,8 +81,8 @@
         it "then sets the correct tags with the type fandom" do
```
ariana-paris: These tests started failing after I added mine because they were comparing a list of `tag_set_id` with a list of `id`. These happened to be the same as long as no other tests were creating tag sets. When mine were added, this pushed the `id` and `tag_set_ids` out of sync and revealed the bug.

annelu:
LGTM 🚢

PR #2758, opened 2017-02-26
AO3-4434 Update unicorn and raindrops

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4434
## Purpose

Updates the unicorn gem, This is the middleware between nginx and the actual rails thing

## Testing

If you can see the archive in staging then it is working I think

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -292,7 +292,7 @@ GEM
       rake (>= 0.8.7)
       rdoc (~> 3.4)
       thor (>= 0.14.6, < 2.0)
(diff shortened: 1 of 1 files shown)


sarken:
Already upgraded to even more recent versions:
- unicorn is at 5.3.0 https://github.com/otwcode/otwarchive/blob/master/Gemfile.lock#L436
- raindrops is at 0.18.0 https://github.com/otwcode/otwarchive/blob/master/Gemfile.lock#L314

PR #2759, opened 2017-02-26
AO3-4368 Update new relic gem

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4368

## Purpose

Updates the new relic gem to be up to date.

## Testing

No testing required in staging, check things look ok in beta


sarken:
Good news! This is already up to date: https://github.com/otwcode/otwarchive/blob/master/Gemfile.lock#L246

Closing both the PR and the issue.

PR #2760, opened 2017-02-26
AO3-4541 Update redis and redis-namespace gem.

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4541

## Purpose

Update redis and redis-namespace gem.

## Testing

I think this is low risk however I would check that wrangling still worked. That readings went though to the database


sarken:
Already upgraded (https://github.com/otwcode/otwarchive/blob/master/Gemfile.lock#L316), so closing!

PR #2762, opened 2017-02-27
AO3-3468 Changing User Name doesn't accurately update Search results

zz9pzza:
## Issue

## Purpose

We were not updating a users works when they changed username

## Testing

1. create an account with UserNameA
2. post works
3. Change user name to UserNameB
4. /works/search for UserNameA
5. find all posted works, though the blurb lists UserNameB


zz9pzza:
This is causing issues in the automated tests

https://app.codeship.com/projects/14476/builds/23768082

app/models/pseud.rb
```diff
@@ -127,6 +127,13 @@ def self.not_orphaned
     where("user_id != ?", User.orphan_account)
   end
 
+  def reset_author_sorting
```
sarken: This might just be me, but I kept tripping over the name for this method -- I think my first thought when I see "reset" is usually "set it back to the way it was," not "set it again, differently." They're both valid uses of the word, but maybe this could be named `update_author_sorting` instead to get rid of the ambiguity?

app/models/user.rb
```diff
@@ -183,7 +183,7 @@ def direct_filters
 
   def expire_caches
     if login_changed?
-      self.works.each{ |work| work.touch }
+      self.works.each{ |work| work.expire_caches }
```
sarken: Just checking -- will this also update the `updated_at` timestamp on the work, just like `touch` did? If so, I think this PR is :+1:, but wouldn't mind another opinion.

sarken:
Merge conflicts

redsummernight:
This issue will also be fixed by #3104 if that gets merged first.

redsummernight:
This was superseded by the Elasticsearch 6 upgrade.

sarken:
I think we still need this. Author sorting for works by pseuds isn't updating when the pseud changes, and when a work is by "pseud (username)", and the username changes, the blurbs aren't getting updated. Reopening to see if it's mergeable.

redsummernight:
Sorry @zz9pzza! It's mergeable now.

PR #2765, opened 2017-02-28
AO3-4480 Add caching to imported_from_url and update cache on deletion of work

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4480

## Purpose

It would be really easy to denial of service the database via the redirect code. I also noticed that we don't flush the cache on work deletion so I fixed that.

## Testing

http://test.archiveofourown.org/redirect?original_url=http://www.triofic.com/viewstory.php?sid=268 should redirect to http://test.archiveofourown.org/works/1034715
Import a work from a url, test that the redirect would work as above


features/importing/work_import.feature
```diff
@@ -220,3 +220,23 @@ Feature: Import Works
     When I import "http://www.intimations.org/fanfic/idol/Huddling.html"
     Then I should see "Preview"
       And I should see "English"
+
+  Scenario: Searching for a work url should be cached
+    Then there should be no work_url cached for "http://www.scarvesandcoffee.net/viewstory.php?sid=9570"
```
shalott: super minor nitpick, this should be "Given" instead of "Then" also is this necessary since you are making the same check on line 229 anyway?

app/models/work.rb
```diff
@@ -189,6 +189,7 @@ def check_for_invalid_chapters
   after_update :adjust_series_restriction
 
   after_save :expire_caches
+  after_destroy :expire_caches
```
redsummernight: There's a test for this we can now re-enable: https://github.com/otwcode/otwarchive/blob/218f15e009438e920c8eef8ae651ba51c5d05edf/spec/lib/works_owner_spec.rb#L48

features/importing/work_import.feature
zz9pzza: Well it's a test that the cache is empty at the start of the test.

ariana-paris: Personally, I would rename the step to `there is no cached work_url for "<url>"`. Then the test can go:

ariana-paris: Personally, I would rename the step to `there is no cached work_url for "<url>"`. Then the test can go:

```
Given there is no cached work_url for ...
  When I import "<url>"
  And I <post without preview>
Then there is no cached work_url for ...
When I look for a work with url ...
Then there is a cached work_url for...
```

(Ideally, we'd have perfect grammar with both a "there is " and "there should be " but this was a compromise that might avoid people misunderstanding the purpose of the test)

Not a deal breaker though - I agree that the URL search is painfully slow so looking forward to having this!

sarken: Wouldn't it make more sense to actually _empty_ the cache at the start of the test, rather than just _check_ that it's empty? So like:
```
Given /^there are no cached work_urls for "([^"]*)"$/ do |url|
  Rails.cache.delete(Work.find_by_url_cache_key(url))
end
```
And then leave all the `Then there should be (no|a) work_url cached for` bits alone.

PR #2774, opened 2017-03-05
AO3-4886 Improve coverage on ChallengeClaimsController

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4886

## Purpose

Improve coverage on ChallengeClaimsController

## Testing

No manual testing required

### DIFF
--- a/spec/controllers/challenge_claims_controller_spec.rb
+++ b/spec/controllers/challenge_claims_controller_spec.rb
@@ -1,0 +1,26 @@
+require 'spec_helper'
+
+RSpec.describe ChallengeClaimsController, type: :controller do
+  include LoginMacros
+  include RedirectExpectationHelper
+  begin
+    @user = FactoryGirl.create(:user)
+  end
+
+  describe 'destory' do
+    context 'with a claim' do
+      before do
+        @signups = create(:challenge_signup)
+        @collection = @signups.collection
+        @claim = create(:challenge_claim)
+      end
+      it 'on an exception gives an error and redirects' do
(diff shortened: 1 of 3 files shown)


spec/controllers/challenge_claims_controller_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+      second_user = create(:user)
+      fake_login_known_user(user)
+      get :index, id: claim.id, user_id: second_user.id
+      it_redirects_to_with_error(root_path, \
+                                 "You aren't allowed to see that user's claims")
+    end
+  end
+
+  describe 'show' do
+    xit 'redirects' do
```
sarken: This seems to fail because `claim.request_prompt` is `nil` because there's no prompt. It seems to work if you make a Prompt factory:
```
# frozen_string_literal: true
require 'faker'

FactoryGirl.define do
  factory :prompt do
  end
end
```
And then change the test to:
```
    it 'redirects logged in user to the prompt' do
      request_prompt = create(:prompt, collection_id: collection.id, challenge_signup_id: signups.id)
      claim_with_prompt = create(:challenge_claim, collection: collection, request_prompt_id: request_prompt.id)
      fake_login_known_user(user)
      get :show, id: claim_with_prompt.id, collection_id: collection.name
      it_redirects_to(collection_prompt_path(collection, claim_with_prompt.request_prompt))
    end
```
(If you don't sign in first, it will redirect to the login page.)

PR #2775, opened 2017-03-05
AO3-4887 Challenge Assignment controller test for no_challenge

cesy:
https://otwarchive.atlassian.net/browse/AO3-4887

Only tests two of the lines without coverage, so there's plenty more to do, but it's an improvement and it would be good to get confirmation that I'm on the right track. In parallel I'll keep working on the other lines without coverage.

### DIFF
--- a/spec/controllers/challenge_assignments_controller_spec.rb
+++ b/spec/controllers/challenge_assignments_controller_spec.rb
@@ -17,9 +17,9 @@ describe ChallengeAssignmentsController do
   context "when logged in" do
     describe 'no_challenge' do
       it 'show an error, redirect and return false' do
-        fake_login_known_user(@collection.owners.first.user)
-        get :no_challenge, collection_id: @collection.name
-        it_redirects_to_with_error(collection_path(@collection), "What challenge did you want to work with?")
+        fake_login_known_user(collection.owners.first.user)
+        get :no_challenge, collection_id: collection.name
+        it_redirects_to_with_error(collection_path(collection), "What challenge did you want to work with?")
       end
     end
   end
--- a/spec/controllers/challenge_assignments_controller_spec.rb
+++ b/spec/controllers/challenge_assignments_controller_spec.rb
@@ -3,11 +3,14 @@ require 'spec_helper'
 describe ChallengeAssignmentsController do
   include LoginMacros
   include RedirectExpectationHelper
+  
+  before(:each) do
+    let(:collection) { create(:collection) }
+  end
 
   context "when not logged in" do
     describe 'no_challenge' do
       it 'error to login page' do
-        let(:collection) { create(:collection) }
         get :no_challenge, collection_id: collection.name
         it_redirects_to_with_error(new_user_session_path, "Sorry, you don't have permission to access the page you were trying to reach. Please log in.")
       end
@@ -17,7 +20,6 @@ describe ChallengeAssignmentsController do
   context "when logged in" do
     describe 'no_challenge' do
       it 'show an error, redirect and return false' do
-        let(:collection) { create(:collection) }
         fake_login_known_user(@collection.owners.first.user)
         get :no_challenge, collection_id: @collection.name
         it_redirects_to_with_error(collection_path(@collection), "What challenge did you want to work with?")
--- a/spec/controllers/challenge_assignments_controller_spec.rb
+++ b/spec/controllers/challenge_assignments_controller_spec.rb
@@ -12,8 +12,7 @@ describe ChallengeAssignmentsController do
       
       it 'should error to login page' do
         get :no_challenge, collection_id: @collection.name
-        expect(response).to redirect_to(new_user_session_path)
-        expect(flash[:error]).to eq "Sorry, you don't have permission to access the page you were trying to reach. Please log in."
+        it_redirects_to_with_error(new_user_session_path, "Sorry, you don't have permission to access the page you were trying to reach. Please log in.")
       end
     end
   end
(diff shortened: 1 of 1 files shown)


spec/controllers/challenge_assignments_controller_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    
+    describe 'no_challenge' do
+      it 'should show an error, redirect and return false' do
+        get :no_challenge, collection_id: @collection.name
+        expect(flash[:error]).to eq "What challenge did you want to work with?"
+        expect(response).to redirect_to(collection_path(@collection))
+      end
+    end
+  end
+end
```
cesy: Fixed

spec/controllers/challenge_assignments_controller_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    before(:each) do
+      @collection = FactoryGirl.create(:collection)
+      @collection.save
+      fake_login_known_user(@collection.owners.first.user)
+    end
+    
+    describe 'no_challenge' do
+      it 'should show an error, redirect and return false' do
+        get :no_challenge, collection_id: @collection.name
+        expect(flash[:error]).to eq "What challenge did you want to work with?"
```
zz9pzza: I would use:

and this can then be:

it_redirects_to_with_error(collection_path(@collection), What challenge did you want to work with?" )

spec/controllers/challenge_assignments_controller_spec.rb
```diff
@@ -0,0 +1,36 @@
+require 'spec_helper'
+
+describe ChallengeAssignmentsController do
+  include LoginMacros
+
+  context "when not logged in" do
+    describe 'no_challenge' do
+      before(:each) do
+        @collection = FactoryGirl.create(:collection)
+        @collection.save
```
zz9pzza: I believe FactoryGirl.create has a save in it, FactoryGirl.build does not

spec/controllers/challenge_assignments_controller_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+      end
+      
+      it 'should error to login page' do
+        get :no_challenge, collection_id: @collection.name
+        expect(response).to redirect_to(new_user_session_path)
```
zz9pzza: See later comment about it_redirects_to_with_error

spec/controllers/challenge_assignments_controller_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+        expect(flash[:error]).to eq "Sorry, you don't have permission to access the page you were trying to reach. Please log in."
+      end
+    end
+  end
+
+  context "when logged in" do
+    before(:each) do
+      @collection = FactoryGirl.create(:collection)
+      @collection.save
```
zz9pzza: And earlier comment about FactoryGirl.create

cesy:
Changes all done and tests passing, ready for a fresh review.

spec/controllers/challenge_assignments_controller_spec.rb
```diff
@@ -0,0 +1,33 @@
+require 'spec_helper'
+
+describe ChallengeAssignmentsController do
+  include LoginMacros
+  include RedirectExpectationHelper
+
+  context "when not logged in" do
+    describe 'no_challenge' do
+      before(:each) do
+        @collection = FactoryGirl.create(:collection)
```
sarken: I think you'd probably be okay doing `let(:collection) { create(:collection) }` at the top instead of using before blocks to create instance variables. But either way, you can omit `FactoryGirl` for brevity. 

The "should" can also be left out of descriptions -- it's less repetitive and lets you get right to the important words.

cesy:
Merge when ready and I'll do a new issue for the rest.

spec/controllers/challenge_assignments_controller_spec.rb
cesy: Updated - yeah, I had more complexity earlier and we can simplify now.

spec/controllers/challenge_assignments_controller_spec.rb
```diff
@@ -0,0 +1,29 @@
+require 'spec_helper'
+
+describe ChallengeAssignmentsController do
+  include LoginMacros
+  include RedirectExpectationHelper
+  
+  before(:each) do
```
sarken: This shouldn't need a before block -- just `let(:collection) { create(:collection) }` and it will be executed when referenced. (In fact, if I recall correctly, `let` doesn't work in before blocks.)

cesy: It just failed testing without it.

sarken: Right -- that's because it was like this:
```
      it 'error to login page' do
         let(:collection) { create(:collection) }
         get :no_challenge, collection_id: collection.name
        end
```

It can't be in the `it`, either. For this, the `let` should be directly after `include RedirectExpectationHelper`.

spec/controllers/challenge_assignments_controller_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+        get :no_challenge, collection_id: collection.name
+        it_redirects_to_with_error(new_user_session_path, "Sorry, you don't have permission to access the page you were trying to reach. Please log in.")
+      end
+    end
+  end
+
+  context "when logged in" do
+    describe 'no_challenge' do
+      it 'show an error, redirect and return false' do
+        fake_login_known_user(@collection.owners.first.user)
```
sarken: I think this is failing because this line and line 21 need to be `collection` (which is the one you've created with the `let`) instead of `@collection` (which doesn't exist)

PR #2776, opened 2017-03-05
WIP: AO3-4888 Clean up challenge_claims_controller

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4888

## Purpose

There are a number of errors which can't be reached and a number of places where we do things in a non ruby way. These changes should only affect the ability to reach error messages rather than give 500 errors.

## Testing

This should need no manual testing.


app/controllers/challenge_claims_controller.rb
```diff
@@ -1,142 +1,99 @@
...
+  before_filter :allowed_to_destroy, only: [:destroy]
 
   # PERMISSIONS AND STATUS CHECKING
 
   def load_challenge
     if @collection
       @challenge = @collection.challenge
     elsif @challenge_claim
-      @challenge = @challenge_claim.collection.challenge
+      # What we want is the ruby &. operator
```
sarken: Maybe make this "TODO: Replace with &. operator after upgrading to Ruby 2.3" on the off chance we'll find it if we ever look at our TODO items.

spec/controllers/challenge_claims_controller_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+  include RedirectExpectationHelper
+  let(:user) { create(:user) }
+  let(:signups) { create(:challenge_signup) }
+  let(:collection) { signups.collection }
+  let(:claim) { create(:challenge_claim, collection: collection) }
+  let(:empty_claim) { create(:challenge_claim) }
+
+  describe 'index' do
+    it 'gives a notice if the collection is closed' do
+      fake_login_known_user(user)
```
annelu: Since there's a `fake_login_known_user` in every it block, it could probably stand to be moved into a `before` block for the larger context.

app/controllers/challenge_claims_controller.rb
```diff
@@ -1,142 +1,99 @@
...
-    
-  
+
   # ACTIONS
 
   def index
-    if !(@collection = Collection.find_by_name(params[:collection_id])).nil? && @collection.closed? && !@collection.user_is_maintainer?(current_user)
-      flash[:notice] = ts("This challenge is currently closed to new posts.")
-    end
+    flash[:notice] = ts("This challenge is currently closed to new posts.") if !(@collection = Collection.find_by_name(params[:collection_id])).nil? && @collection.closed? && !@collection.user_is_maintainer?(current_user)
```
annelu: Personally I think this way makes the line a bit too long

annelu:
Nice cleanup, it looks great! 🎉

sarken:
Just a heads up that this is unmergeable now that we've merged #2774!

spec/controllers/challenge_claims_controller_spec.rb
```diff
@@ -20,40 +23,39 @@
...
+    it 'needs a claim' do
+      get :show, collection_id: "none existent collection"
+      it_redirects_to_with_error(root_path, \
+                                 "What claim did you want to work on?")
     end
 
-    xit 'needs a collection' do
-      fake_login_known_user(user)
-      get :show
+    it 'needs a collection' do
```
sarken: Descriptions need to describe how the code behaves in given contexts. You test that the code does `x` when it has a valid collection parameter and `y` when it doesn't have a valid collection parameter, not that the code needs a collection.

spec/controllers/challenge_claims_controller_spec.rb
```diff
@@ -20,40 +23,39 @@
...
-      claim_with_prompt = create(:challenge_claim, collection: collection, request_prompt_id: request_prompt.id)
-      fake_login_known_user(user)
-      get :show, id: claim_with_prompt.id, collection_id: collection.name
-      it_redirects_to(collection_prompt_path(collection, claim_with_prompt.request_prompt))
+    it 'redirects' do
+      get :show, id: claim.id, collection_id: collection.name
+      it_redirects_to(root_path)
+    end
+
+    it 'needs a claim' do
```
sarken: Descriptions need to describe how the code behaves in given contexts. You test that the code does `x` when it has a valid claim parameter and `y` when it doesn't have a valid claim parameter, not that it needs a claim. 

Also, I'd suggest explicitly setting an invalid claim and a valid collection here, just to be absolutely clear what you're testing: `get :show, claim_id: 999_999, collection_id: collection`. If for some reason `collection` doesn't work there, it's okay to leave it as an invalid collection (I think the phrase you're looking for in your tests is "nonexistent [thing]," btw, not "none existent [thing].")

app/controllers/challenge_claims_controller.rb
```diff
@@ -1,142 +1,99 @@
...
   def load_challenge
     if @collection
       @challenge = @collection.challenge
     elsif @challenge_claim
-      @challenge = @challenge_claim.collection.challenge
+      # TODO: Replace with &. operator after upgrading to Ruby 2.3
+      @challenge = @challenge_claim.collection.challenge if @challenge_claim.collection
     end
-    no_challenge and return unless @challenge
+    no_challenge && return unless @challenge
```
sarken: I don't think `&& return` actually works, even though Hound says and our tests are passing.

spec/controllers/challenge_claims_controller_spec.rb
```diff
@@ -64,8 +64,17 @@
   describe 'destory' do
     context 'with a claim' do
       it 'on an exception gives an error and redirects' do
-        fake_login_known_user(@user)
+        fake_login_known_user(User.find(collection.all_owners.first.user_id))
+        collection.challenge = nil
+        collection.save
+        delete :destroy, id: claim.id, collection_id: collection.name
+        it_redirects_to_with_error(collection_path(collection), \
```
sarken: This doesn't need the `\`

spec/controllers/challenge_claims_controller_spec.rb
```diff
@@ -64,8 +64,17 @@
...
+                                   "What challenge did you want to work with?")
+      end
+
+      it 'on an exception gives an error and redirects' do
```
sarken: This has the same description as the previous test -- what's the difference between the two of them? The one above it looks like `it "gives and error and redirects if the claim's collection has no challenge"` and this one looks like it's more generic, i.e. `it "gives and error and redirects if record can't be destroyed"`.

spec/controllers/challenge_claims_controller_spec.rb
```diff
@@ -31,29 +33,27 @@
...
-    xit 'needs a collection' do
-      fake_login_known_user(user)
-      get :show
-      it_redirects_to_with_error(root_path, \
-                                 "What challenge did you want to work with?")
+    it 'redirects logged in users if no collection is given' do
+      request_prompt = create(:prompt, collection_id: collection.id, challenge_signup_id: signup.id)
+      create(:challenge_claim, collection: collection, request_prompt_id: request_prompt.id)
+      get :show, id: 999_999, collection_id: collection.name
+      it_redirects_to_with_error(collection_path(collection), \
```
sarken: This shouldn't have a `\`

spec/controllers/challenge_claims_controller_spec.rb
```diff
@@ -31,29 +33,27 @@
...
+      get :show, id: claim_with_prompt, collection_id: collection.name
       it_redirects_to(collection_prompt_path(collection, claim_with_prompt.request_prompt))
     end
 
-    xit 'needs a collection' do
-      fake_login_known_user(user)
-      get :show
-      it_redirects_to_with_error(root_path, \
-                                 "What challenge did you want to work with?")
+    it 'redirects logged in users if no collection is given' do
```
sarken: I'm not sure this is right. Contrary to what this says, it looks like a collection _is_ being specified (`get :show, id: 999_999, collection_id: collection.name`), plus the error message seems to think the _claim_ is the problem. To me, it looks like `it "redirects logged in users if the claim does not exist"`.

spec/controllers/challenge_claims_controller_spec.rb
```diff
@@ -64,8 +64,17 @@
   describe 'destory' do
```
sarken: Not part of your changes, but there's a typo up here -- it should be `destroy`

PR #2778, opened 2017-03-06
AO3-4750 Add link to most recent chapter in work blurbs

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4750

## Purpose

Where it says "Chapters: [Current #]/[Expected #]" in the blurb stats line, the [Current #] should link to that chapter – so if it's "Chapters: 5/10", the 5 will take you directly to Chapter 5 of the work (http://test.ao3.org/works/###/chapters/###) 

## Testing

Refer to JIRA issue

## References

We're doing [AO3-5905](https://otwarchive.atlassian.net/browse/AO3-5905) instead of the below.

* ~Must be merged with pull request #2779 for issue AO3-4751~
* ~Must be deployed after the announcement post~

PR #2779, opened 2017-03-06
AO3-4751 Hits to most recent chapter of a work should increase the hit count

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4751

## Purpose

Going directly to the most recent chapter of a multi-chapter work should increase the work's hits.

"Directly" here means by following a link that's _not_ on the work. Code-wise, we're defining "on the work" as "any page that has '/works/####' in the URL" (where '####' matches the id of the work the chapter belongs to).

## Testing

Refer to JIRA for manual testing instructions. 

Because of the way the hit count incrementing code relies on referrers and IP addresses, I don't _think_ there's a way to write automated tests for the more complicated parts. If there is, though, please let me know!

There's a cursory test in the work_browse feature in #2778 that will make sure following the link from the blurb increases the hit count:

```
Scenario: The recent chapter link in a work's blurb should show the adult
content notice to visitors who are not logged in

  Given I am logged in as a random user
    And a rating exists with name: "Mature", canonical: true, adult: true
    And a fandom exists with name: "Canonical Fandom", canonical: true
    And I post the 3 chapter work "WIP" with fandom "Canonical Fandom" with rating "Mature"
  When I am logged out
    And I browse the "Canonical Fandom" works
    And I follow the recent chapter link for the work "WIP"
  Then I should see "adult content"
  When I follow "Proceed"
  Then I should find myself on Chapter 3 of the work "WIP"
  When "AO3-4751" is fixed
    # And I should see "Hits: 1"
``` 

After #2778 is merged, let me know -- I'll merge master into this branch and remove the `When "AO3-4751" is fixed` step and the `#` so the test will actually run. I can't include it here off the bat because it depends on the addition of the recent chapter link, and we didn't want the code for both issues in one pull request.

## References

* Must be merged with pull request #2778 for issue AO3-4750
* Must be deployed after the announcement post


katieby:
Testing:
You can set the referrer and ip address in specs by doing
`allow_any_instance_of(ActionDispatch::Request).to receive(:remote_ip).and_return("1.1.1.1")`

app/controllers/chapters_controller.rb
```diff
@@ -70,8 +70,13 @@ def show
...
 
-      # TEMPORARY hack-like thing to fix the fact that chaptered works weren't hit-counted or added to history at all
-      if chapter_position == 0
+      # TEMPORARY hack-like thing to fix the fact that chaptered works weren't hit-counted
+      # or added to history at all. Increases hit count if accessing the first chapter or
+      # if accessing the most recent chapter when you're not already on the work (e.g. if
+      # you followed the recent chapter link or came from a subscription email).
+      if chapter_position == 0 || chapter_position == (@work.number_of_posted_chapters - 1) &&
+                                  (request.referer.nil? ||
+                                  !request.referer.match(/#{work_path(@work)}/))
```
tickinginstant: It might be good to change this regex to something like `/#{work_path(@work)}\b/` to avoid false positives (false negatives?) when one work's ID is a prefix of another's. It'd probably be pretty rare to have a work (or its comments) link to the latest chapter of a work whose ID is a prefix, but it might be worth trying to prevent errors in case it does happen.

The word boundary character should allow a regex like `/works/1\b` to match `/works/1`, `/works/1/chapter/1`, and `/works/1?view_full_work=true`, but not `/works/100`.

zz9pzza: I was trying to do that and was getting bitten by the / on the match :)

factories/works.rb
```diff
@@ -23,6 +23,22 @@
       work_skin_id 1
     end
 
+    # create(:work_with_chapters) will create a work with 2 chapters by default
+    # create(:work_with_chapters, chapters_count: 3) lets you specify how many
+    # chapters the work should have
+    factory :work_with_chapters do
```
redsummernight: Should we add `posted true` to make it posted by default?

spec/controllers/chapters_controller_spec.rb
```diff
@@ -9,7 +9,10 @@
   let(:unposted_work) { create(:work, authors: [user.pseuds.first]) }
   let(:banned_user) { create(:user, banned: true) }
   let(:banned_users_work) { create(:work, posted: true, authors: [banned_user.pseuds.first]) }
-
+  let(:chaptered_work) { create(:work_with_chapters, chapters_count: 3, posted: true) }
```
redsummernight: We can remove `posted: true` here if the factory has it by default.

spec/controllers/chapters_controller_spec.rb
```diff
@@ -264,6 +262,81 @@
         get :show, params: { work_id: work.id, id: work.chapters.first.id }
       end
     end
+
+    context "when the chapter is the first chapter" do
+      it "increases the hit count" do
+        REDIS_GENERAL.set("work_stats:#{chaptered_work.id}:last_visitor", nil)
```
redsummernight: We no longer need this line since REDIS_GENERAL gets cleared before every test.

spec/controllers/chapters_controller_spec.rb
```diff
@@ -264,6 +262,81 @@
...
+        expect {
+          get :show, params: { work_id: chaptered_work.id, id: first_chapter.id }
+        }.to change { REDIS_GENERAL.get("work_stats:#{chaptered_work.id}:hit_count").to_i }.by(1)
```
redsummernight: I'm on the fence... I think instead of looking into specific Redis keys, we should call `StatCounter.hits_to_database` then run expects on the work's `stat_counter.hit_count`, to test the whole process, but that would make the tests slower.

spec/controllers/chapters_controller_spec.rb
redsummernight: Should we move these lets inside `describe "index"` so they don't get created for the blocks that don't use them?

sarken:
Hmm, I have a bug here:
```
Scenario: The recent chapter link in a work's blurb should show the adult
content notice to visitors who are not logged in

  Given I am logged in as a random user
    And an adult canonical rating exists with name: "Mature"
    And a fandom exists with name: "Canonical Fandom", canonical: true
    And I post the 3 chapter work "WIP" with fandom "Canonical Fandom" with rating "Mature"
  When I am logged out
    And I browse the "Canonical Fandom" works
    And I follow the recent chapter link for the work "WIP"
  Then I should see "adult content"
  When I follow "Proceed"
  Then I should find myself on Chapter 3 of the work "WIP"
    And I should see "Hits: 1"
```
But the hit doesn't get counted in that case because `request.referer.match(/#{work_path(@work)}\b/)` will be true, as the adult content warning is shown on `works/1234`.

redsummernight:
> But the hit doesn't get counted in that case because `request.referer.match(/#{work_path(@work)}\b/)` will be true, as the adult content warning is shown on `works/1234`.

When on the last chapter, we can count a hit if one of these is true:

- referer is null (from elsewhere)
- referer is exactly the current chapter (`works/X/chapters/Y`) AND the current page has `params[:view_adult]`, meaning someone passes the warning on their way to this chapter
- referer does not match the current work (prev/next chapter)

redsummernight:
Also Brakeman is failing because of new code:

https://hakiri.io/github/otwcode/otwarchive/master/9d5641cc38adaf10eba316eaa146730dbfe7171f/warnings?name=Denial+of+Service

"Model attribute used in regex", but the attributes in question are the IDs of works/chapters. I think we can mark them as false positive.

sarken:
>referer is exactly the current chapter (works/X/chapters/Y) AND the current page has params[:view_adult], meaning someone passes the warning on their way to this chapter

...duh, thank you! I forgot we were looking at the new chapter links in the stats line, so I was trying to figure out how to make it work when you just clicked on the title. 🤦‍♂

app/controllers/chapters_controller.rb
```diff
@@ -267,6 +265,23 @@ def post_chapter
...
+  # -- coming without a referer, e.g. via a subscription email
+  # -- coming through the recent chapter link, but first passing through the
+  #    adult content warning
+  # -- coming from anywhere else that is not the work, e.g. social media link
+  def increase_work_hit_count?
+    chapter_position = @chapters.index(@chapter)
+    chapter_position.zero? ||
+      chapter_position == (@work.number_of_posted_chapters - 1) &&
+        (request.referer.nil? ||
+        request.referer.match(/#{chapter_path(@chapter)}/) && params[:view_adult] ||
```
redsummernight: One last change: similar to https://github.com/otwcode/otwarchive/pull/2779#discussion_r110215022, we should add `\b` to the end of the chapter path regex.

sarken:
Closing because we're going to [totally change hit counts](https://otwarchive.atlassian.net/browse/AO3-5905). 🎉

PR #2785, opened 2017-03-06
AO3-4889 Increase coverage of the series controller

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4889

## Purpose

Increase coverage of the series controller

## Testing

No manual testing required


spec/controllers/series_controller_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+
+  describe 'new' do
+    it 'assigns a series' do
+      get :new
+      expect(assigns(:series)).to be_a_new(Series)
+    end
+  end
+
+  describe 'edit' do
+    it 'redirects to orphan if there are no psueds left.' do
```
sarken: Typo -- should be "pseuds." The period can be removed as well.

spec/controllers/series_controller_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+      fake_login_known_user(user)
+      get :edit, remove: "me", id: series
+      it_redirects_to(new_orphan_path(series_id: series))
+    end
+  end
+
+  describe 'create' do
+    it 'when renders new if the series is invalid' do
```
sarken: I think the word "when" is unnecessary here.

spec/controllers/series_controller_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+    it 'gives notice and redirects on a valid series' do
+      fake_login_known_user(user)
+      post :create, series: { title: "test title" }
+      expect(flash[:notice]).to eq "Series was successfully created."
+      expect(response).to have_http_status :redirect
+    end
+  end
+
+  describe 'update' do
+    it 'stops you removing the last author of a series' do
```
sarken: "redirects and errors if removing the last author of a series" might be a bit more descriptive of the behavior than "stops you removing"

sarken:
Once Travis runs and passes, this is ready to go -- thanks! :+1:

PR #2786, opened 2017-03-07
AO3-4813 Add tests for tag_set_associations_controller, fix unreachable error

redsummernight:
Test when users visit a non-existent tag set.

## Issue

https://otwarchive.atlassian.net/browse/AO3-4813

## Purpose

Increase coverage of tag_set_associations_controller.

## Testing

Visiting the associations page of a missing tag set will now redirect to the tag sets page with a flash error "What tag set did you want to look at?", instead of returning a 404. Example:

http://test.archiveofourown.org/tag_sets/99999999/associations


features/step_definitions/tag_set_steps.rb
```diff
@@ -150,6 +150,13 @@
   visit tag_set_path(tagset)
 end
 
+When /^I view a tag set that does not exist/ do
+  id = 1
+  tagset = OwnedTagSet.find_by_id(id)
+  tagset&.destroy
```
sarken: Hound might have given you some bad advice here -- I _think_ `&.` only works in Ruby 2.3.0 and up, and we're still on 2.2.5, but if it passes on Travis, that's the important thing!

sarken:
After some discussion in Slack, we determined that this is testing owned_tag_sets_controller and not tag_set_associations_controller, so we're going to mark this "Needs coder action" for now.

features/tags_and_wrangling/tag_set.feature
```diff
@@ -55,6 +55,16 @@ Feature: creating and editing tag sets
     And I should see "Knots Landing"
     And I should see "Models Inc"
 
+  Scenario: If a tag set does not exist, no one should be able to see its associations
+  Given I am logged in as "tagsetter"
+  When I view associations for a tag set that does not exist
+  Then I should see an error flash alert saying "What tag set did you want to look at?"
```
sarken: It's probably better to keep things simple and do just `Then I should see "What tag set did you want to look at?"` here -- as you've probably noticed, we do a lot of checking error and success messages in our cukes, and for better or worse, we've always done it without checking that that it's in a flash div.

sarken:
Just noting that everything is still :+1: after the merge commit!

PR #2794, opened 2017-03-09
AO3-4908 Tag set nomination tests when the fandom nomination limit is 0.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4908

## Purpose

This adds four new cucumber tests and a few new RSpec tests to handle the case of nominating characters & relationships (and editing character & relationship nominations) when the fandom nomination limit is set to 0.

## Testing

I don't think manual testing is required -- as long as this passes automated tests, it's doing what it's supposed to.

## Credit

tickinginstant, she/her


tickinginstant:
I'm glad it looks good! :)

PR #2803, opened 2017-03-17
AO3-4919 Reduce the number of writes the new tag code makes on the database.

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4919

## Purpose

This only writes the tag cache which is used for selecting tags which are not used if is out of date by 20 uses or has less than 3 uses.

## Testing

I hope the automated tests will be sufficient as I am looking at the load on the database servers with worry.


zz9pzza:
This is not exactly what has been hotfixed on beta but I suspect it is better.

PR #2808, opened 2017-03-20
AO3-4808 Display an error if fandom is removed when editing tags

ariana-paris:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4808

## Purpose

Fixes the Broken on Test issue where no message appears when using Edit Tags to remove the fandom.

## Testing

See Jira.

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -306,7 +306,7 @@ class WorksController < ApplicationController
       render :new && return
     elsif params[:cancel_button]
       flash[:notice] = ts('New work posting canceled.')
-      redirect_to current_user && return
+      redirect_to current_user and return
     else
       # now also treating the cancel_coauthor_button case, bc it should function like a preview, really
       set_work_tag_error_messages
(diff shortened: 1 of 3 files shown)


app/controllers/works_controller.rb
```diff
@@ -455,6 +455,7 @@ def update
   end
 
   def update_tags
+    set_work_tag_error_messages
```
redsummernight: The warning will be shown if I remove all fandom tags then hit "Cancel".

sarken: Thank you -- good catch! Taking the Ready to Merge label off and putting Needs Coder Action on.

app/controllers/works_controller.rb
```diff
@@ -306,7 +306,7 @@ def create
       render :new && return
     elsif params[:cancel_button]
       flash[:notice] = ts('New work posting canceled.')
-      redirect_to current_user && return
+      redirect_to current_user and return
```
ariana-paris: No. This had actually broken that code path so my change fixes this.

redsummernight: No change needed, just noting that we use `redirect_to(current_user) && return` in some places to get around this warning.

app/controllers/works_controller.rb
```diff
@@ -455,33 +455,37 @@ def update
...
+        if params[:preview_button]
+          preview_mode(:edit_tags) do
+            render :preview_tags
+          end
+        elsif params[:save_button]
+          Work.expire_work_tag_groups_id(@work.id)
+          flash[:notice] = ts('Tags were successfully updated.')
+          redirect_to(@work)
+        else
+          if @work.has_required_tags? && @work.invalid_tags.blank?
```
redsummernight: I would use `unless @work.errors.empty?` and avoid nesting.

redsummernight:
Looks good to me.

PR #2814, opened 2017-03-22
AO3-4858 Batch tag count updates to speed up work posting

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4858

## Purpose

Removes the rails automatic caching

## Testing

Post works, I would not expect things to be quicker on test particularly as the slow down only happens when two or more people post to the same large tag.

Make sure that works turn up in tags as quickly or slowly as they did before.

I would deploy this with a bit of down time I suspect as I think I would be suspicious of the migration and two systems working in different ways.


app/models/tag.rb
```diff
@@ -23,6 +23,80 @@ class Tag < ActiveRecord::Base
...
+    REDIS_GENERAL.mset("tag_update_#{id}_value", value, "tag_update_#{id}_large_tag", large_tag)
+    value
+  end
+
+  def taggings_count=(value)
+    expiry_time = taggings_count_expiry(value)
+    # Only write to the cache if there are more than TAGGINGS_COUNT_MIN_CACHE_COUNT ( defaults to 1,000 ) uses.
+    Rails.cache.write(taggings_count_cache_key, value, race_condition_ttl: 10, expires_in: expiry_time.minutes) if value >= (ArchiveConfig.TAGGINGS_COUNT_MIN_CACHE_COUNT || 1000)
+    check_if_large_tag(expiry_time)
+    write_taggings_to_redis(value, nil)
```
tickinginstant: I believe this will overwrite the value stored in REDIS for the key `"tag_update_#{id}_large_tag"` (which may have just been set by the previous line's call to `check_if_large_tag`). From my testing, `redis-rb` treats `nil` as being equivalent to the empty string, with no special handling. It makes sense, since all REDIS keys & values are strings, and `nil.to_s` is the empty string.

Since `write_taggings_to_redis` is never called with both `value` and `large_tag` as non-nil arguments, I'd suggest separating `write_taggings_to_redis` into two functions: one for storing the taggings count, and one for storing the value of large_tag.

Alternatively, it might be more efficient to avoid storing the value of `large_tag` on REDIS, because it can be computed directly from the `taggings_count`. You could consider modifying `Tag.write_redis_to_database` to compute the `large_tag` value from the new `taggings_count` value, and write both to MySQL at the same time. (You might even be able eliminate the `large_tag` column altogether and just search for tags with a large value of `taggings_count_cache` -- there's no index either way, so I believe MySQL would have to examine all tags regardless.)

zz9pzza:
@tickinginstant Thanks for that, I have followed your advice :) 

👍

zz9pzza:
https://travis-ci.org/zz9pzza/otwarchive/builds/215332687 as travis looks if it is unhappy

features/step_definitions/work_steps.rb
```diff
@@ -74,6 +73,7 @@
     click_button("Post Without Preview")
   end
   Work.tire.index.refresh
```
ariana-paris: Something for a future refactor - move these two lines that are repeated together in all these step definitions into a "refresh all work counts and indexes" method, but let's not complicate this one.

PR #2825, opened 2017-03-28
AO3-4556 Add a link to the force logout page to go back to where you were

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4556

## Purpose

Add a link which will return you to where you were before being forced logged out

## Testing

To make the archive log you out delete the user_credentials cookie for the site.


zz9pzza:
I think this will reduce the liklihood of the force logout page getting in to the cache as well

app/controllers/application_controller.rb
```diff
@@ -144,8 +142,13 @@ def store_location
   # Redirect to the URI stored by the most recent store_location call or
   # to the passed default.
   def redirect_back_or_default(default = root_path)
+    back_url = params[:user_session][:return_to_url] if params[:user_session]
     back = session[:return_to]
     session.delete(:return_to)
+    if back_url
+      Rails.logger.debug "Returning over riding return to #{back} with #{back_url}"
```
sarken: I think "overriding" is generally one word.

app/views/home/lost_cookie.html.erb
```diff
@@ -2,4 +2,6 @@
 
 <div class="userstuff">
   <p><%= ts("You have been automatically logged out to protect your privacy.  Please make sure that your browser is set to accept cookies from archiveofourown.org and delete any existing Archive cookies.  If you continue to have problems logging in, please  <a href=\"#{new_feedback_report_url}\">contact Support</a>.".html_safe)  %></p>
+  <% page = params[:return_to_url].blank? ? root_path : params[:return_to_url] %>
+  <p><%= link_to(ts('Login returning to %{path}', path: page), new_user_session_path(return_to_url: page)) %></p>
```
sarken: "Log in and return to %{path}" would be better link text, I think (we use "log in" for a verb and "login" for a noun or adjective). But possibly a silly question: is there a reason we can't include the login form itself on this page?

zz9pzza: What I could try and do is just try and set the cookie, and redirect us back to the same location but with a counter in the url so we try and self heal things...

sarken: Do you mean make it so the user wouldn't see this error page and wouldn't need to log back in? That does sound better.

redsummernight:
This has conflicts now.

app/controllers/application_controller.rb
```diff
@@ -221,8 +226,13 @@ def store_location
   # Redirect to the URI stored by the most recent store_location call or
   # to the passed default.
   def redirect_back_or_default(default = root_path)
+    back_url = params[:user_session][:return_to_url] if params[:user_session]
```
tickinginstant, line 229: This is an unsafe redirect. Because it's in the params, it can be used for phishing (because you can send someone a link that looks like AO3 but will secretly redirect them elsewhere) or other malicious activities. When a URL is passed through params like this, it needs to be checked before it's passed to `redirect_to`.

(This also has the same issue as [AO3-5949](https://otwarchive.atlassian.net/browse/AO3-5949): there are [some](https://otwarchive.atlassian.net/browse/AO3-5953) [actions](https://otwarchive.atlassian.net/browse/AO3-5685) that really shouldn't be GETs, and I'm not sure if it's a good idea to use params to redirect before that's resolved.)

zz9pzza: Would it be suffficant to check that the url is relative ( I mean with no host portion ? , or with the host portion set to nil  ? )

tickinginstant: Host, scheme, maybe also port/username/password? It'd probably be better to pick a few whitelisted components that are allowed to have values, and reject any URLs that have any other components set.

But as I said, this has [AO3-5953](https://otwarchive.atlassian.net/browse/AO3-5953) and [AO3-5685](https://otwarchive.atlassian.net/browse/AO3-5685) as pre-requisites, so there's no rush to fix it.

PR #2826, opened 2017-03-29
AO3-4872 Redirect and return flash message when tag set or nominations can't be found

cyrilcee:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4872

## Purpose

This is a follow-up to a pull request review comment: https://github.com/otwcode/otwarchive/pull/2728/files#r102482816

In tag_set_nominations_controller.rb, there was code to return a flash message when a user tries to update or access a tag set or tag set nomination that does not exist. Those error handling paths were never reached because OwnedTagSet.find or TagSetNomination.find would immediately raise an ActiveRecord::RecordNotFound error if no record was found with the given id.

~~This code change makes that RecordNotFound error more explicit and strips out the unused error handling paths.~~
This code change makes it so that the flash messages and redirect behavior are actually reached.

## Testing

The easiest way to test this is to:

1. Try to access a non-existing nomination for an existing tag set ([example](http://test.archiveofourown.org/tag_sets/378/nominations/1000000)). You should be redirected to the tag set page w/ a flash message: "Which nominations did you want to work with?"
2. Try to access an existing nomination for a non-existing tag set ([example](http://test.archiveofourown.org/tag_sets/1000000/nominations/6624)). You should be redirected to the [Tag Sets index page](http://test.archiveofourown.org/tag_sets) w/ a flash message: "What tag set did you want to nominate for?"
3. Try to access the Review Nominations index page for a non-existing tag set ([example](http://test.archiveofourown.org/tag_sets/1000000/nominations)). You should be redirected to the [Tag Sets index page](http://test.archiveofourown.org/tag_sets) w/ a flash message: "What nominations did you want to work with?"

## Credit

*What name do you want us to use to credit your work in the [Archive of Our Own's Release Notes](http://archiveofourown.org/admin_posts?tag=1)?* cyrilcee

*What pronouns do you prefer (she/her, he/him, zie/hir etc)?* she/her

### DIFF
--- a/app/controllers/tag_set_nominations_controller.rb
+++ b/app/controllers/tag_set_nominations_controller.rb
@@ -20,7 +20,7 @@ class TagSetNominationsController < ApplicationController
   def load_tag_set
     @tag_set = OwnedTagSet.find_by_id(params[:tag_set_id])
     unless @tag_set
-      flash[:notice] = ts("What tag set did you want to nominate for?")
+      flash[:error] = ts("What tag set did you want to nominate for?")
       redirect_to tag_sets_path and return
     end
   end
@@ -28,11 +28,11 @@ class TagSetNominationsController < ApplicationController
   def load_nomination
     @tag_set_nomination = TagSetNomination.find_by_id(params[:id])
     unless @tag_set_nomination
-      flash[:notice] = ts("Which nominations did you want to work with?")
+      flash[:error] = ts("Which nominations did you want to work with?")
       redirect_to tag_set_path(@tag_set) and return
     end
     unless current_user.is_author_of?(@tag_set_nomination) || @tag_set.user_is_moderator?(current_user)
-      flash[:notice] = ts("You can only see your own nominations or nominations for a set you moderate.")
+      flash[:error] = ts("You can only see your own nominations or nominations for a set you moderate.")
       redirect_to tag_set_path(@tag_set) and return
     end
   end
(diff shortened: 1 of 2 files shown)


zz9pzza:
Hi cyrilcee, I have had a quick look at this and we are moving away from raising errors.  And we are happier with giving the users a more informative error message.

So instead of doing : 
 @tag_set = OwnedTagSet.find(params[:tag_set_id])
we have been changing it to 
 @tag_set = OwnedTagSet.find_by_id(params[:tag_set_id])
Which returns nil if the object is not found, that way we can test against nil and return a more friendly message.

cyrilcee:
@zz9pzza - Implemented the suggested change! Can you take another look?

Most of this is straightforward, but I want to flag a change I made in TagSetNominationsController.load_nomination: line 32. The previous code wanted to redirect to user_tag_set_nominations_path(@user), which doesn't seem to exist. I thought it made sense to change the redirect to tag_set_path(@tag_set) but would prefer someone more experienced with this code to give a 👍 / 👎 .

Thank you!

zz9pzza:
I think I am happy with the redirect to tag_set_path(@tag_set)

I will ask @sarken to comment on the changes from error to notice

cyrilcee:
The tests were definitely wrong originally - I wrote them a bunch of weeks ago without realizing the original code actually used flash[:notice]. But it's no problem at all to change the controller code to error, and revert the tests to match. Will ping here when done!

PR #2831, opened 2017-03-29
AO3-4892 Collection Item strong params

davidstump:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-4892

## Purpose

Implements strong parameters and ForbiddenAttributesProtection for CollectionItem

## Testing

1. Log in
2. Browse > Collections > New Collection
3. Enter a Collection name, Display title, and any other settings or information you want (e.g. make it moderated and unrevealed)
4. Submit
5. Now you'll want to fill your collection with items – that means post some works, invite some works, and add some bookmarks, preferably using a couple of different accounts (i.e. not just the collection owner's account)
6. Log back in as the collection owner
7. Browse > Collections > [Whatever you named your collection] > Manage Items
8. Where the collection items are will depend on your collection's settings (e.g. a moderated collection will have any works created by accounts who aren't the moderated under Awaiting Approval) – find them and then use the various drop down menus and sticky boxes to change the items' status
9. Submit
You should get a blue "Collection status updated!" flash notice and the items should be updated. (Note:   AO3-3908 TO DO  means you can't set a work to Rejected and check Remove for it at the same time.)
It might also be a good idea to test as the creator of some works/bookmarks:
1. Log in
2. Add some works and/or bookmarks to a variety of collections (it would also be a good idea to log in as a collection owner and invite some of the user's works to a collection – but make sure the user has "Automatically agree to your work being collected by others in the Archive" off)
3. Hi, username > My Collections > Manage Collected Works
4. Where the collection items are will depend on the collections' settings – find them and then use the various drop down menus and check boxes to change the items' status (e.g. Remove the item, accept or reject an invitation)
5. Submit
You should get a blue "Collection status updated!" flash notice and the items should be updated. (Note:   AO3-3908 TO DO  means you can't set a work to Rejected and check Remove for it at the same time.)

## References

None

## Credit

Littlelines - David Stump

he/him

### DIFF
--- a/app/controllers/collection_items_controller.rb
+++ b/app/controllers/collection_items_controller.rb
@@ -9,7 +9,7 @@ class CollectionItemsController < ApplicationController
 
   def load_item_and_collection
     if params[:collection_item]
-      @collection_item = CollectionItem.find(params[:collection_item][:id])
+      @collection_item = CollectionItem.find(collection_item_params[:id])
     else
       @collection_item = CollectionItem.find(params[:id])
     end
@@ -151,7 +151,7 @@ class CollectionItemsController < ApplicationController
   end
 
   def update_multiple
-    @collection_items = CollectionItem.update(params[:collection_items].keys, params[:collection_items].values).reject { |item| item.errors.empty? }
+    @collection_items = CollectionItem.update(collection_items_params[:collection_items].keys, collection_items_params[:collection_items].values).reject { |item| item.errors.empty? }
     if @collection_items.empty?
       flash[:notice] = ts("Collection status updated!")
       redirect_to (@user ? user_collection_items_path(@user) : collection_items_path(@collection))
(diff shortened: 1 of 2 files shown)


tickinginstant:
I'm a little confused as to why `collection_item_params` is defined with two calls to permit. Won't that just make the parameter hash empty? Or does `.permit(a).permit(b)` behave like `.permit(a, b)`?

Also, the definition of `collection_items_params` looks like it ought to work, but there seems to be a bit of a bug. When I visited the "Manage Items" page of an anonymous collection and tried to remove one of the works, I got this error:
```
RuntimeError (can't modify frozen Hash):
  app/controllers/collection_items_controller.rb:154:in `update_multiple'
```
I get the same error when I visit the "Manage Items" page of an unrevealed collection and try to remove a work, as well as when I visit my "Manage Collected Works" page and try to remove my work from any collection (unrevealed or not, anonymous or not). As far as I can tell, none of those actions result in an error with the code on the master branch. And I didn't change any fields other than the "Remove" checkbox, so it's not the result of AO3-3908.

I'm thinking it might have something to do with the order of the parameters, because the parameters `:unrevealed`, `:anonymous`, and `:user_approval_status` all occur after the `:remove` parameter in the strong params list, and the bug only seems to happen when one of those parameters is submitted (even if it hasn't been changed).

davidstump:
The first `permit` was a typo and should have been `require`. That's fixed and pushed. Also, your debugging looks correct - I moved `remove` to the end of the params and no longer see the frozen hash error. Thanks!!

tickinginstant:
No problem! Thanks for the quick response. :)

PR #2832, opened 2017-03-30
AO3-4902 AO3-4904 ChallengeSignups and Prompts strong params

WendyBeth:
This PR addresses both tickets [AO3-4904](https://otwarchive.atlassian.net/browse/AO3-4904) and [AO3-4902](https://otwarchive.atlassian.net/browse/AO3-4902), because the ChallengeSignupsController passes Prompt params through its create method.

Of note, the reason various `*_tagname` params show up twice is because lists of tagnames can show up in the parameters in different formats. When there are 0 or 1 tagnames, they show up as a string (allowed by `:tagnames`), and when there is more than one they show up as an array (allowed by `tagnames: []`), so both must be allowed by the controller to allow for full functionality.

# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-4904 and https://otwarchive.atlassian.net/browse/AO3-4902

## Purpose

What does this PR do?

Adds strong_params to ChallengeSignupsController and PromptsController (ChallengeSignupsController relies on strong_params for the Prompt model).

## Testing

From JIRA

PromptsController (4904):

Despite the fact this is named prompts, it's not just prompt memes – it's the individual requests and offers within a sign up for a gift exchange as well as the prompts (requests) within a sign up for a prompt meme.

Important! You'll want to test this on gift exchanges/prompt memes that use tag sets and allow all the things, so you can make sure every possible field works – including title, URL, description, every type of tag's "allow any" option, the "Anonymous" option on prompt memes.

Make sure you can sign up for both a gift exchange and a prompt meme challenge, and then you'll want to check that you can edit your sign-up to add more requests/offers/prompts.

When creating and editing a sign-up, make sure the sign-up has all the information you've inputted (including pseud) and that you don't receive any unexpected errors.

Make sure you can edit a prompt directly, using the Edit Prompt option that will appear when you go to the challenge's page and follow the "My Sign-up" or "My Prompts" link in the sidebar.

It wouldn't hurt to make sure you can delete requests/offers/prompts from within the sign-up as well.
To be extra thorough, run matching on your gift exchange and send assignments. Make sure you get the email.

ChallengeSignupsController (4902)

To test manually, make sure you can sign up for both a gift exchange and a prompt meme challenge, and then you'll want to check that you can edit your sign-ups.

To sign up for a gift exchange, create or go to a gift exchange that is open for sign ups and choose "Sign-up Form" from the sidebar. Fill out the form and press "Submit." To edit, choose "My Sign-up" from the sidebar and then "Edit Sign-up."

To sign up for a prompt meme, create or go to a prompt meme that is open for sign ups and choose "Prompt Form" from the sidebar. Fill out the form and press "Submit." To edit, choose "My Prompts" from the sidebar and then "Edit Sign-up."

Make sure the sign-ups are saved with all the information you've inputted (including pseud) and that you don't receive any unexpected errors.

It wouldn't hurt to make sure you can delete the sign ups as well.

If you want to be super thorough – never a bad idea with challenges – create a gift exchange with multiple sign-ups and run matching and send assignments.

## References

None

## Credit

WendyBeth

She/her

### DIFF
--- a/app/controllers/challenge_signups_controller.rb
+++ b/app/controllers/challenge_signups_controller.rb
@@ -339,6 +339,7 @@ protected
       :id,
       :collection_id,
       :title,
+      :url,
       :any_fandom,
       :any_character,
       :any_relationship,
--- a/app/controllers/challenge_signups_controller.rb
+++ b/app/controllers/challenge_signups_controller.rb
@@ -369,7 +369,7 @@ protected
         fandom_tagnames: [],
       ],
       optional_tag_set_attributes: [
-        :tagnames
+        :tagnames, tagnames: []
       ]
     ]
   end
--- a/app/controllers/challenge_signups_controller.rb
+++ b/app/controllers/challenge_signups_controller.rb
@@ -369,7 +369,7 @@ protected
         fandom_tagnames: [],
       ],
       optional_tag_set_attributes: [
-        :tagnames, tagnames: []
+        :tagnames
       ]
     ]
   end
--- a/app/controllers/challenge_signups_controller.rb
+++ b/app/controllers/challenge_signups_controller.rb
@@ -329,7 +329,7 @@ protected
   def challenge_signup_params
     params.require(:challenge_signup).permit(
       :pseud_id,
-      requests_attributes: nested_prompt_params
+      requests_attributes: nested_prompt_params,
       offers_attributes: nested_prompt_params
     )
   end
--- a/app/controllers/challenge_signups_controller.rb
+++ b/app/controllers/challenge_signups_controller.rb
@@ -196,7 +196,8 @@ class ChallengeSignupsController < ApplicationController
   end
 
   def create
-    @challenge_signup = ChallengeSignup.new(params[:challenge_signup])
+    @challenge_signup = ChallengeSignup.new(challenge_signup_params)
+
     @challenge_signup.pseud = current_user.default_pseud unless @challenge_signup.pseud
     @challenge_signup.collection = @collection
     # we check validity first to prevent saving tag sets if invalid
@@ -209,7 +210,7 @@ class ChallengeSignupsController < ApplicationController
   end
 
   def update
-    if @challenge_signup.update_attributes(params[:challenge_signup])
+    if @challenge_signup.update_attributes(challenge_signup_params)
       flash[:notice] = ts('Sign-up was successfully updated.')
       redirect_to collection_signup_path(@collection, @challenge_signup)
     else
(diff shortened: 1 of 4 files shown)


app/controllers/challenge_signups_controller.rb
```diff
@@ -322,4 +323,54 @@ def prompt_meme_to_csv
...
+      :any_relationship,
+      :any_freeform,
+      :any_category,
+      :any_rating,
+      :any_warning,
+      :anonymous,
+      :description,
+      :_destroy,
+      tag_set_attributes: [
+        :id,
```
sarken: I'm having a hard time determining why this is necessary, but the tests definitely fail without it. Someone else want to double check this and see what I'm missing? (The tests do pass if `id` is removed from `tag_set_attributes` in the prompts controller, though.)

app/controllers/prompts_controller.rb
```diff
@@ -158,4 +159,45 @@ def destroy
     end
   end
 
+  private
+
+  def prompt_params
+    params.require(:prompt).permit(
+      :id,
```
sarken: This might be unnecessary -- I couldn't find it in the relevant forms, and the tests seemed to pass without it.

app/controllers/challenge_signups_controller.rb
WendyBeth: So I've looked into this, and here are my thoughts:

The reason I included `:id` in the PromptsController (and the ChallengeSignupsController) is because it gets sent through by the params, and I whitelisted all of the params that I saw come through by using `binding.pry` to inspect them in various situations. Also, Rails passes through the primary key for nested attributes by default.

However, I've edited a prompt with existing tag sets, and while the id is present under params[:prompt][:tag_set_attributes], it is quietly excluded in the prompt_params method instead of raising a ForbiddenAttributesError. I thought that might be because TagSet doesn't have attributes protection on it, so I added that to see what would happen - and the behavior is the same. So for whatever reason, the protected attributes gem doesn't seem to care about the nested attributes' `:id`. 

So I would say that even though everything's passing without the id in the PromptsController, the fact that it does come in as a param maybe suggests that there's no reason *not* to allow it. I'm pretty stumped as to why it doesn't raise, though. :woman_shrugging:

sarken: Oh, thank you, that's all good to know! And also kind of frustrating that it behaves so unexpectedly, blargh.

Because it's not really clear what the best course of action is here, I think we can just go with the quickest -- I'm going to merge this now, and we'll just keep watch for any weirdness around the prompts controller as we test.

PR #2833, opened 2017-03-30
AO3-4921 Expand challenge assignment spec test

cesy:
https://otwarchive.atlassian.net/browse/AO3-4921

Increases coverage a little, though there's still some way to go.

### DIFF
--- a/spec/controllers/challenge_assignments_controller_spec.rb
+++ b/spec/controllers/challenge_assignments_controller_spec.rb
@@ -15,11 +15,61 @@ describe ChallengeAssignmentsController do
   end
 
   context "when logged in" do
+    let(:collection2) { create(:collection, challenge: GiftExchange.new) }
+    let(:offer) { create(:challenge_signup, collection: collection2, pseud: user2.default_pseud) }
+    let(:offer2) { create(:challenge_signup, collection: collection2, pseud: otheruser.default_pseud) }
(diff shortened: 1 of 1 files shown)


spec/controllers/challenge_assignments_controller_spec.rb
```diff
@@ -15,12 +15,62 @@
   end
 
   context "when logged in" do
+    let(:collection2) { create(:collection, challenge: GiftExchange.new) }
```
redsummernight: You should use the factory for gift_exchange added in #2756.

cesy:
Fixed, thanks for the suggestion.

spec/controllers/challenge_assignments_controller_spec.rb
```diff
@@ -15,12 +15,63 @@
   end
 
   context "when logged in" do
+    let(:gift_exchange) { create(:gift_exchange) }
+    let(:collection2) { create(:collection, challenge: gift_exchange, challenge_type: "GiftExchange") }
```
redsummernight: I think you can move some of the lets inside their specific describes, and avoid using number suffixes.

redsummernight:
LGTM. :+1:

spec/controllers/challenge_assignments_controller_spec.rb
cesy: I'll do this in the expansion I'm working on, since I'm also adding some more checks.

scottsds:
👍

PR #2836, opened 2017-03-31
AO3-4950 Migrate Readings table primary key to bigint

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4950

## Purpose

We have run out of id's on Readings we need some more

## Testing

I want this to pass travis, hot fix on test, quick check to see if test works.

hot fix on beta while we are offline.


sarken:
Quick question! The JIRA issue mentions migrating the kudos and taggings tables as well -- are you planning to add those to this pull request, or should those tables have separate JIRA issues created?

db/migrate/20170331170222_change_index_type.rb
```diff
@@ -0,0 +1,8 @@
+class ChangeIndexType < ActiveRecord::Migration
+  def up
+    change_column :readings, :id, "bigint NOT NULL AUTO_INCREMENT"
+  end
+
+  def down
```
sarken: Can you name this migration a bit more descriptively, e.g. change_readings_id_type?

PR #2837, opened 2017-04-01
AO3-4951 Update tag rake tasks to work with taggings_count_cache 

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4951

## Purpose

The rake tasks and most notably the ones that delete 0 use tags where not updated.

## Testing

Create a 0 use tag and then get the rake task run and see if the tag has been delete

### DIFF
--- a/lib/tasks/tag_tasks.rake
+++ b/lib/tasks/tag_tasks.rake
@@ -1,6 +1,6 @@
 namespace :Tag do
   desc "Reset common taggings - slow"
-  task(:reset_common => :environment) do
+  task(reset_common: :environment) do
     Work.find(:all).each do |w|
       print "." if w.id.modulo(100) == 0; STDOUT.flush
       #w.update_common_tags
@@ -9,7 +9,7 @@ namespace :Tag do
   end
 
   desc "Reset tag count"
-  task(:reset_count => :environment) do
+  task(reset_count: :environment) do
     Tag.find(:all).each do |t|
       t.taggings_count
     end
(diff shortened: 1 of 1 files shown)


lib/tasks/tag_tasks.rake
```diff
@@ -9,36 +9,36 @@ namespace :Tag do
...
   desc "Reset taggings count for obviously wrong taggings_count"
-  task(:fix_taggings_count => :environment) do
-    tag_scope = Tag.where("taggings_count < 0")
+  task(fix_taggings_count: :environment) do
+    tag_scope = Tag.where("taggings_count_cache < 0")
     tag_count = tag_scope.count
     tag_scope.each_with_index do |tag, index|
       puts "#{index} / #{tag_count}"
-      t.taggings_count
+      tag.taggings_count
```
sarken: Given the mis-named variable, was this not working before?

zz9pzza: It's not something that was run automatically, so no I don't think it was working and given the new way we do things we should never get negative counts again.

sarken: Ah, okay! That explains why we didn't notice. Is this task something we should remove, or do you want to keep it for a while just in case?

zz9pzza: So I am just running a version of this where I am redoing all the 1 use tags. I can image the very low number of usage tags might now get looked at ever so might have the wrong number of usages record. Generally I hope under the new scheme we shouldn't need it.

PR #2821, opened 2017-03-27
AO3-4939 Improve reliability of timestamp-related tests for tag comments

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4939

## Purpose

Improve the reliability of the tests 

## Testing

No manual testing required

PR #2594, opened 2016-10-24
AO3-4690 Add test for the tag_controller.

zz9pzza:
# Pull Request Checklist

* [x] Have you read [How to write the perfect pull request](https://github.com/blog/1943-how-to-write-the-perfect-pull-request)?
* [x] Have you read through the [contributor guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [x] Have you added tests for any changed functionality?
* [x] Have you added the JIRA issue number as the *first* thing in your pull request title (eg: `AO3-1234 Fix thing`)
* [x] Have you updated the JIRA issue with the information below?

## Issue

https://otwarchive.atlassian.net/browse/AO3-4690

## Purpose

Add more coverage

## Testing

No manual testing required.

PR #2839, opened 2017-04-03
AO3-4926 Extend tests for inbox_controller.rb

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4926

## Purpose

- Add specs for the controller.
- Add cukes for replying to inbox comments from the home page, both with and without JavaScript.
- Remove the route cancel_reply which has no action.

## Testing

- Check that coverage for the controller is now at 100%.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -230,7 +230,6 @@ Otwarchive::Application.routes.draw do
     resource :inbox, controller: "inbox" do
       member do
         get :reply
-        get :cancel_reply
         post :delete
       end
     end
--- a/features/comments_and_kudos/inbox.feature
+++ b/features/comments_and_kudos/inbox.feature
@@ -98,7 +98,7 @@ Feature: Get messages in the inbox
   Scenario: A user can reply to a comment from the home page without JavaScript
     Given I am logged in as "sewwiththeflo"
       And I post the work "Cat Thor's Bizarre Adventure"
-    When I am logged in as "unbeatablesg"
+      And I am logged in as "unbeatablesg"
       And I post the comment "dude this is super great!!" on the work "Cat Thor's Bizarre Adventure"
     When I am logged in as "sewwiththeflo"
       And I go to the homepage
@@ -118,7 +118,7 @@ Feature: Get messages in the inbox
   Scenario: A user can reply to a comment from the home page
     Given I am logged in as "sewwiththeflo"
       And I post the work "Cat Thor's Bizarre Adventure"
-    When I am logged in as "unbeatablesg"
+      And I am logged in as "unbeatablesg"
       And I post the comment "dude this is super great!!" on the work "Cat Thor's Bizarre Adventure"
     When I am logged in as "sewwiththeflo"
       And I go to the homepage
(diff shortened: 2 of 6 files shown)


config/routes.rb
```diff
@@ -230,7 +230,6 @@
     resource :inbox, controller: "inbox" do
       member do
         get :reply
-        get :cancel_reply
```
redsummernight: This route was introduced in d142b55ad7640fbd6d03805a7204f3123904483d. It has never had an action.

features/step_definitions/comment_steps.rb
```diff
@@ -45,7 +45,7 @@
 
 When /^I set up the comment "([^"]*)" on the work "([^"]*)"$/ do |comment_text, work|
   work = Work.find_by_title!(work)
-  visit work_url(work)
+  visit work_path(work)
```
redsummernight: In a JS-enabled test, the domain is set to `example.com` and the absolute `work_url` will include that. I saw #2594 where we got around this with `visit work_url(work).gsub("http://www.example.com","")`, but I think `work_path` is nicer.

spec/controllers/inbox_controller_spec.rb
```diff
@@ -0,0 +1,192 @@
...
+        get :show, user_id: user.login
+        expect(response).to render_template("show")
+        expect(assigns(:inbox_total)).to eq(0)
+        expect(assigns(:unread)).to eq(0)
+      end
+
+      context "with unread comments" do
+        let!(:inbox_comments) do
+          Array.new(3) do |i|
+            create(:inbox_comment, user: user, created_at: Time.now + i.days)
```
redsummernight: If I don't space them out, they can get created in the same second and break the ordering test. This is a low-budget "timecop".

jbeal-work:
Have a look at the time machine I added in https://github.com/otwcode/otwarchive/pull/2821

features/comments_and_kudos/inbox.feature
```diff
@@ -94,3 +94,44 @@ Feature: Get messages in the inbox
       And I should not see "The latest unread items from your inbox."
       And I should not see "cutman on Special Coverage"
       And I should not see "Is there anything we can do to make the fight go longer?"
+
+  Scenario: A user can reply to a comment from the home page without JavaScript
+    Given I am logged in as "sewwiththeflo"
+      And I post the work "Cat Thor's Bizarre Adventure"
+    When I am logged in as "unbeatablesg"
+      And I post the comment "dude this is super great!!" on the work "Cat Thor's Bizarre Adventure"
+    When I am logged in as "sewwiththeflo"
```
cesy: Any reason we have two Whens rather than an extra And?

features/step_definitions/comment_steps.rb
cesy: Yes, much nicer.

features/comments_and_kudos/inbox.feature
redsummernight: No reason at all, it should be an And. Fixed now, thanks.

PR #2849, opened 2017-04-08
AO3-4962 Strong parameters for BookmarkSearch.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4962

## Purpose

BookmarkSearch is a subclass of Search, and Search is a subclass of ActiveRecord::Base. So this pull request adds strong parameters for the calls to `BookmarkSearch.new` in `BookmarksController#index` and `BookmarksController#search`.

## Testing

I wrote some steps for testing in the JIRA issue.

## Credit

tickinginstant, she/her

### DIFF
--- a/app/controllers/bookmarks_controller.rb
+++ b/app/controllers/bookmarks_controller.rb
@@ -41,7 +41,7 @@ class BookmarksController < ApplicationController
 
   def search
     @languages = Language.default_order
-    options = params[:bookmark_search] || {}
+    options = params[:bookmark_search].present? ? bookmark_search_params : {}
     options.merge!(page: params[:page]) if params[:page].present?
     options[:show_private] = false
     options[:show_restricted] = current_user.present?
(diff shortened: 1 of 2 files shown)


zz9pzza:
@tickinginstant I am a bit worried about https://github.com/otwcode/otwarchive/blob/master/app/views/bookmarks/_filters.html.erb#L79 
and the five potential hidden fields there. Am I missing something, why are they not needed ?

tickinginstant:
Those are `hidden_field_tag` fields, not `f.hidden_field` tags, so they're not nested inside the `:boomark_search` params. Since I'm using `.require(:bookmark_search).permit()`, the permit statement will only affect parameters nested inside the `:bookmark_search` field. The non-nested parameters are used in [the `load_owner` function of the BookmarksController](https://github.com/otwcode/otwarchive/blob/master/app/controllers/bookmarks_controller.rb#L245) to set the variable `@owner`, which is later used to modify the BookmarkSearch if it exists.

PR #2850, opened 2017-04-09
AO3-4956 Extend tests for opendoors/tools_controller.rb and change from 404 to flash error

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4956

## Purpose

- Display an error message instead of raising 404 when a work ID cannot be found.
- Add controller spec for opendoors/tools_controller.rb.
- Add a scenario for updating a work's imported from URL.

## Testing

See https://otwarchive.atlassian.net/browse/AO3-4956.

### DIFF
--- a/features/importing/archivist.feature
+++ b/features/importing/archivist.feature
@@ -220,11 +220,18 @@ Feature: Archivist bulk imports
     When I go to the Open Doors tools page
     Then I should see "Sorry, you don't have permission to access the page you were trying to reach."
 
-  Scenario: Open Doors committee members can see their tools
-    Given I have an Open Doors committee member "OpenDoors"
+  Scenario: Open Doors committee members can update the redirect URL of a work
+    Given the work "My Immortal"
+      And I have an Open Doors committee member "OpenDoors"
       And I am logged in as "OpenDoors"
     When I go to the Open Doors tools page
     Then I should see "Update Redirect URL"
+    When I fill in "imported_from_url" with "http://example.com/my-immortal"
+      And I fill in "work_url" with the path to the "My Immortal" work page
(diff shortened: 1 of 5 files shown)


features/importing/archivist.feature
```diff
@@ -220,11 +220,18 @@ Feature: Archivist bulk imports
...
-    Given I have an Open Doors committee member "OpenDoors"
+  Scenario: Open Doors committee members can update the redirect URL of a work
+    Given the work "My Immortal"
+      And I have an Open Doors committee member "OpenDoors"
       And I am logged in as "OpenDoors"
     When I go to the Open Doors tools page
     Then I should see "Update Redirect URL"
+    When I fill in "imported_from_url" with "http://example.com/my-immortal"
+      And I fill in "work_url" with the path to the "My Immortal" work page
+      And I submit with the 2nd button
```
zz9pzza: Is there a reason that we can't use the buttons name as 2nd button feels fragile ?

redsummernight: The tools page has 2 forms: one to change the imported URL and one to block imported emails. The first form has no id/class, and both forms have the same "Submit" button.

Can we change the submit button values?

zz9pzza: I couldn't get to the right place on test or my dev. I just wanted to make sure it was like that :(

I will @sarken to comment. I think what you have done is fine. If we wanted to have two buttons called Submit which I don't like we could have an extra div but I would prefer sarken had a look ?

sarken: At some point, we should make the button text more descriptive, but I don't think we need to do anything about it now.

PR #2851, opened 2017-04-09
AO3-4963 Modify new taggings_count code to handle brand new fandoms used on External Works.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4963

## Purpose

This pull request moves the call to `Tag.update_tag_cache` from the Work's `expire_caches` function into a new callback in the Tagging class, which is triggered whenever a Tagging is created or destroyed. This ensures that the `taggings_count` will behave the same regardless of whether the new tag is used on a Work or an ExternalWork (or a Bookmark).

## Testing

There are steps for replicating the bug in the JIRA issue.

## Credit

tickinginstant, she/her

### DIFF
--- a/features/tags_and_wrangling/brand_new_fandoms.feature
+++ b/features/tags_and_wrangling/brand_new_fandoms.feature
@@ -2,8 +2,11 @@
 Feature: Brand new fandoms
 
   Background:
-    Given I have no tags
-      And basic tags
+    # The external works form will error if we don't have basic tags, and since
+    # we're trying to create a brand new tag, we just want to double-check that
+    # it doesn't exist.
+    Given basic tags
+      And the tag "My Brand New Fandom" does not exist
 
   Scenario: Brand new fandoms should be visible on the uncategorized fandoms page.
     Given I am logged in as a random user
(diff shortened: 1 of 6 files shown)


features/tags_and_wrangling/brand_new_fandoms.feature
```diff
@@ -0,0 +1,43 @@
...
+    Given basic tags
+      And the tag "My Brand New Fandom" does not exist
+
+  Scenario: Brand new fandoms should be visible on the uncategorized fandoms page.
+    Given I am logged in as a random user
+      And I post a work "My New Work" with fandom "My Brand New Fandom"
+      And The periodic tag count task is run
+    When I follow "Uncategorized Fandoms" within "#header"
+    Then I should see "My Brand New Fandom"
+
```
zz9pzza: Could we have a follow on test that checks that the count is zero if the last external work to use a tag is destroyed ?

tickinginstant: I added some tests checking whether the fandom disappears from the Uncategorized Fandoms/Fandoms by Media page when it should. Do you want me to try to add in some RSpec tests that explicitly check whether the count is zero, too?

zz9pzza: I think that the tests you have added are sufficient :)

If you feel like gilding the lily then do but don't feel obliged.

tickinginstant: I think I'll leave it as is. :)

PR #2852, opened 2017-04-12
AO3-4968 Replace some cassettes with stubbed HTTP requests

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4968

## Purpose

Stub URLs instead of keeping cassettes when only the returned status code matters, as in tests on:

- External bookmarks
- External parent works
- Importing invalid URLs

Restore tests for [AO3-1803](https://otwarchive.atlassian.net/browse/AO3-1803) and [AO3-1783](https://otwarchive.atlassian.net/browse/AO3-1783).

Remove after hook for resetting webmock, which is already the default behavior.

## Testing

Just automated tests.


redsummernight:
Removed test URLs, in case they go down in the future:

- https://ao3testing.dreamwidth.org/856.html: "Some External Work", ao3testing.

  > Fandom: Test Fandom
Rating: General Audiences
Characters and pairing: Character 1/Character 2, Character 3, Character 4
Summary: A post for testing bookmarking external works.
  > 
  > My amazing story would be here! This is for @bookmark_external_work.

- http://test.sidrasue.com/short.html: Lorem ipsum.

PR #2853, opened 2017-04-13
AO3-4967 - posting a work with ambiguous pseuds is not allowed though strong parameters

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4967

## Purpose

Adds ambiguous pseuds to the list of allowed parameters allowing you to post a work if a name is shared

## Testing

Probably not easy to test until  4966 is fixed


features/works/work_create.feature
```diff
@@ -263,3 +263,38 @@ Feature: Create Works
       And I should see "No Archive Warnings Apply"
       And I should not see "Author Chose Not To Use Archive Warnings"
       And I should see "It wasn't my fault, you know."
+
+  Scenario: Users can co-create a work with a co-creator who has multiple pseuds
+    Given basic tags
+      And I am logged in as "myself" with password "password"
+    When I go to myself's pseuds page
```
sarken: We have step definitions for `"user" creates the default pseud "pseud"` and `"user" has the pseud "pseud"` -- using those might tighten this up a bit. I think you might also be able to shorten the posting bit to:
```
I fill in the basic work information for "All Hell Breaks Loose" <-- fills in all the required fields on the form
I add the co-author "Me" <-- fills out the co-author field
I press "Post Without Preview"
etc
```

PR #2857, opened 2017-04-15
AO3-4953 - Change Support form to require email address

scottsds:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4953

## Purpose

* Makes the email field a **required** field for the Support form; and,
* removes the language in the flash notice saying that we accept anonymous feedback.

## Testing

1. Ensure that blue flash-notice on Support form says: "Our spam filter does collect IP addresses, but we never see them."
2. Email field label should now read: "Your email (required)"
3. The email field cannot be blank.
4. The email field must contain a valid email address.

### DIFF
--- a/app/views/feedbacks/new.html.erb
+++ b/app/views/feedbacks/new.html.erb
@@ -24,14 +24,14 @@
   <fieldset>
     <legend><%= ts("Contact Information") %></legend>
     <p class="notice" id="email-field-description">
-      <%= ts("We value anonymous feedback, but if you'd like a personal reply to your question or comment, we need your email address to contact you. Our spam filter does collect IP addresses, but we never see them.") %>
+      <%= ts("Our spam filter does collect IP addresses, but we never see them.") %>
     </p>
     <dl>
       <dt><%= f.label :username, ts("Your name (optional)") %></dt>
       <dd>
         <%= f.text_field :username %>
       </dd>
-      <dt><%= f.label :email, ts("Your email (optional)") %></dt>
+      <dt class="required"><%= f.label :email, ts("Your email (required)") %></dt>
       <dd>
         <%= f.text_field :email, size: 60, "aria-describedby" => "email-field-description" %>
       </dd>
--- a/app/views/feedbacks/new.html.erb
+++ b/app/views/feedbacks/new.html.erb
@@ -32,7 +32,7 @@
         <%= f.text_field :username %>
       </dd>
       <dt class="required"><%= f.label :email, ts("Your email (required)") %></dt>
-      <dd>
+      <dd class="required">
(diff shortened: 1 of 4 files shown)


app/views/feedbacks/new.html.erb
```diff
@@ -24,14 +24,14 @@
...
+      <%= ts("Our spam filter does collect IP addresses, but we never see them.") %>
     </p>
     <dl>
       <dt><%= f.label :username, ts("Your name (optional)") %></dt>
       <dd>
         <%= f.text_field :username %>
       </dd>
-      <dt><%= f.label :email, ts("Your email (optional)") %></dt>
+      <dt class="required"><%= f.label :email, ts("Your email (required)") %></dt>
       <dd>
```
sarken: Could you add the `required` class here as well?

scottsds: To the <dd> ?

sarken: Yep!

features/other_b/support.feature
```diff
@@ -21,39 +21,30 @@ Feature: Filing a support request
   When I follow "Support and Feedback"
     And I fill in "Brief summary" with "you suck"
     And I fill in "Your comment" with "blah blah blah"
-    And I fill in "Your email (optional)" with ""
+    And I fill in "Your email (required)" with ""
     And I select "Deutsch" from "feedback_language"
     And all emails have been delivered
     And I press "Send"
+  Then I should see "Email does not seem to be a valid address."
```
sarken: It looks like the original intent of this test was to fill out the form with all required information and make sure it works. I'd suggest just filling in a valid email address in this scenario right off the bat, since the whole "leave the email field blank and you'll get an error message" situation is tested in `Scenario: Not logged in, with and without email`.

scottsds:
Made the requested changes!

features/other_b/support.feature
```diff
@@ -21,39 +21,27 @@ Feature: Filing a support request
   When I follow "Support and Feedback"
     And I fill in "Brief summary" with "you suck"
     And I fill in "Your comment" with "blah blah blah"
-    And I fill in "Your email (optional)" with ""
+    And I fill in "Your email (required)" with "test@archiveofourown.org"
     And I select "Deutsch" from "feedback_language"
     And all emails have been delivered
     And I press "Send"
   Then I should see "Your message was sent to the Archive team - thank you!"
     And 1 email should be delivered
```
sarken: This needs to be updated to two now that email addresses are required, but otherwise, this looks good to me!

sarken: Just checking to see how the test fix is coming!

PR #2859, opened 2017-04-15
AO3-2556 Cascade comment destruction to parent placeholders.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-2556

## Purpose

Deleting a comment with children leaves behind a placeholder comment. Currently, when all of the comment's children are deleted, this leaves behind the placeholder. This pull request adds a new after_destroy callback to the acts_as_commentable plugin to check whether the parent comment is a placeholder that should also be deleted.

While I was working on this bug, I also discovered that the existing before_destroy callback in `lib/plugins/acts_as_commentable/comment_methods.rb` doesn't seem to get called in the test or development environments. If this is also true in production, that means that `threaded_left` and `threaded_right` are probably set wrong for a whole bunch of comment threads with deleted comments in them. This doesn't really affect anything except comment deletion, so it's fairly subtle, but it does mean that it's more difficult to fix existing comment placeholders.

To fix this problem, I modified the existing callback to ensure that it would be called. I also wrote an After task to fix `threaded_left` and `threaded_right` for all existing comments, which should be run before the After task to clean up old placeholders.

## Testing

There are two scenarios to test. The first scenario is covered in the bug report:
1. Create a comment.
2. Reply to the comment.
3. Delete the parent comment.
4. Delete the reply.
5. Check to see if there is a deleted comment placeholder.

The second scenario is slightly different -- it checks whether the existing before_destroy callback is actually being called. I haven't tested this in production, but it fails on master in the test and development environments:
1. Create a comment.
2. Reply to the comment.
3. Delete the reply.
4. Delete the parent comment.
5. Check to see if there is a deleted comment placeholder.

## Credit

tickinginstant, she/her


sarken:
Hi there! 

Barring any catastrophes during the deploy process, we're done with the Rails upgrade. Since this is your oldest reviewed pull request, we'd like to include it in our next release. Would you have time to update this in the next several days? If not, no worries! We can either save it for another release or update it ourselves -- just let us know what you prefer.

Thanks!

tickinginstant:
I tried resolving the merge conflicts -- here's hoping it'll pass the automated tests.

(Also, congrats on the Rails upgrade!)

sarken:
It passed just fine -- thank you!

redsummernight:
Enough haunting. I went to https://coveralls.io/github/otwcode/otwarchive/settings and unchecked "Leave comments?" as well as "Use status API?".

PR #2860, opened 2017-04-16
AO3-4685 Extend tests for invite_requests_controller

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4685

## Purpose

More tests.

## Testing

No pain, only tests. Check that coverage for the controller is now at 100%.

### DIFF
--- a/spec/controllers/invite_requests_controller_spec.rb
+++ b/spec/controllers/invite_requests_controller_spec.rb
@@ -1,0 +1,188 @@
+require "spec_helper"
+
+describe InviteRequestsController do
+  include LoginMacros
+  include RedirectExpectationHelper
+  let(:admin) { create(:admin) }
+  let(:user) { create(:user) }
+
+  describe "GET #index" do
+    it "renders" do
+      get :index
+      expect(response).to render_template("index")
+      expect(assigns(:invite_request)).to be_a_new(InviteRequest)
+    end
+  end
+
+  describe "GET #show" do
+    context "given invalid emails" do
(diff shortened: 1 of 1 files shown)


spec/controllers/invite_requests_controller_spec.rb
```diff
@@ -0,0 +1,188 @@
...
+      email = generate(:email)
+      post :create, invite_request: { email: email }
+      invite_request = InviteRequest.find_by_email(email)
+      it_redirects_to_with_notice(invite_requests_path, "You've been added to our queue! Yay! We estimate that you'll receive an invitation around #{invite_request.proposed_fill_date}. We strongly recommend that you add do-not-reply@archiveofourown.org to your address book to prevent the invitation email from getting blocked as spam by your email provider.")
+    end
+  end
+
+  describe "DELETE #destroy" do
+    it "blocks non-admins" do
+      delete :destroy
```
redsummernight: @zz9pzza recommended using `fake_logout` to make sure no one's logged in before the test.

spec/controllers/invite_requests_controller_spec.rb
```diff
@@ -0,0 +1,188 @@
...
+      it_redirects_to_with_notice(root_path, "I'm sorry, only an admin can look at that area")
+
+      fake_login_known_user(user)
```
redsummernight: I just noticed that I could have used `fake_login` instead, since we don't need to do anything with the user in this test.

PR #2861, opened 2017-04-16
AO3-4621 Fetch Icons via https

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4621

## Purpose

Current icons should be served by https, rather than http ( amazon icons ). New icons should be saved in the database as https...

## Testing

Check the page source


scottsds:
Look good 👍

PR #2862, opened 2017-04-16
AO3-4847 Allow admins to see ip address of commenter

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4847

## Purpose

Allows admins to see a commenter's IP address

## Testing

Ensure that a guest, an author and the commentator can't see the ip address.
Ensure that an admin can see the ip address.

### DIFF
--- a/app/views/comments/_single_comment.html.erb
+++ b/app/views/comments/_single_comment.html.erb
@@ -60,7 +60,7 @@
       <% end %><!-- end caching -->
 
       <% if logged_in_as_admin? %>
-        <p class="comment ip"><%= ts("IP Address: %{ip_address}", ip_address: single_comment.ip_address.blank? ? "No address recorded" : single_comment.ip_address) %></p>
+        <p class="ip"><%= ts("IP Address: %{ip_address}", ip_address: single_comment.ip_address.blank? ? "No address recorded" : single_comment.ip_address) %></p>
       <% end %>
       <%= render "comments/comment_actions", comment: single_comment %>
     <% end %>
--- a/app/views/comments/_single_comment.html.erb
+++ b/app/views/comments/_single_comment.html.erb
@@ -60,7 +60,9 @@
       <% end %><!-- end caching -->
 
       <%= render "comments/comment_actions", comment: single_comment %>
-
+      <% if logged_in_as_admin?  %>
+          <p><%= ts("IP Address: %{ip_address}", ip_address: single_comment.ip_address.blank? ? "None recorded" : single_comment.ip_address) %> </p>
+      <% end %>
     <% end %>
   <% end %>
 </li>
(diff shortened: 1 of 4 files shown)


app/views/comments/_single_comment.html.erb
```diff
@@ -60,7 +60,9 @@
       <% end %><!-- end caching -->
 
       <%= render "comments/comment_actions", comment: single_comment %>
-
+      <% if logged_in_as_admin?  %>
```
sarken: We use "No address recorded" for the IP address in the work meta -- we should use the same phrasing here. Also, this should come above`<%= render "comments/comment_actions", comment: single_comment %>`, and have some styling applied to better differentiate it from the comment text. I'd suggest adding the following to 24-role-admin.css, directly above `/* TABLES AND LISTS */`:

```
.comment .ip {
  margin: 1.929em 0.643em 0.643em;
  text-align: right;
}
```

(If you could remove the `#footer .beta` bit on lines 12-14 while you were in that file, I'd be much obliged. We took the beta stripe out of the footer a long time ago and I apparently forgot to update this file.)

There are also a few spacing issues: 
- There's an extra space between `logged_in_as_admin?` and ` %>`
- The `<p>` is indented two spaces too many
- There's an extra space between the text and the closing `</p>` tag

We should probably add the same IP address information in comments/_single_comment_abbreviated.html.erb as well so Abuse can see it from the /comments index.

zz9pzza:
@sarken I am not sure I am applying the style correctly as I tried turning it red in dev and it didn't work for me.

app/views/comments/_single_comment.html.erb
```diff
@@ -59,8 +59,10 @@
         <% end %>
       <% end %><!-- end caching -->
 
+      <% if logged_in_as_admin? %>
+        <p class="comment ip"><%= ts("IP Address: %{ip_address}", ip_address: single_comment.ip_address.blank? ? "No address recorded" : single_comment.ip_address) %></p>
```
sarken: These should just be `<p class="ip">` -- the `.comment` selector refers to the class on the surrounding `li` -- but that shouldn't be preventing the red from being applied. What does your non-working code look like? Is the right alignment not being applied either?

zz9pzza: The none working code is what I pushed, however I am not sure if I need to do the loading all the skins in to the database and re cache all the skins or not.

sarken: I'm not sure how skin changes work on Vagrant, so it might be worth a try. I had it working on my local dev environment with the exact code I suggested. :|

zz9pzza: Well I think I have the code you want ?

sarken: Yep, it is indeed the code I gave you, but you should probably try recaching the skins to check that it works instead of just taking my word for it

zz9pzza: I can see my issue on dev is strangeness in how nginx is serving a really old file.

PR #2863, opened 2017-04-17
AO3-4973 Increase coverage of spam_report model

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4973

## Purpose

Increase coverage of spam_report model

## Testing

No manual testing required.


spec/models/spam_report_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    third_spam.ip_address = "192.168.11.1"
+    third_spam.save
+    expect(AdminMailer).to receive(:send_spam_alert).with(
+      third_spam.pseuds.first.user_id => { score: 14, \
+                                           work_ids: [first_spam.id, second_spam.id, third_spam.id] }
+    ).and_return(double("AdminMailer", deliver: true))
+    SpamReport.run
+  end
+
+  it 'Posting in the past decreases the score' do
```
sarken: Maybe `it "decreases the spam score if the work is backdated"`? Or would this be more `it "decreases the spam score if the work was a draft for several days before posting"`? But also, if we're testing that the score _decreases_, this test should really be comparing some sort of before and after, e.g. that the score goes from X to 11.

spec/models/spam_report_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  it 'sends no email if there is no significant spam' do
+    third_spam
+    ArchiveConfig.SPAM_THRESHOLD = 70
+    expect(AdminMailer).not_to receive(:send_spam_alert)
+    SpamReport.run
+  end
+
+  it 'sends email if there is significant spam' do
```
sarken: Maybe `it "sends email if the spam score is higher than the spam threshold"`?

spec/models/spam_report_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    ArchiveConfig.SPAM_THRESHOLD = 10
+    create(:work, spam: false, posted: true)
+  end
+
+  it 'has a recent date after the new date' do
+    spam_report = SpamReport.new
+    expect(spam_report.new_date).to be > spam_report.recent_date
+  end
+
+  it 'sends no email if there is no significant spam' do
```
sarken: Maybe `it "does not send email if the spam score is lower than the spam threshold"`?

spec/models/spam_report_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+                                           work_ids: [first_spam.id, second_spam.id, third_spam.id] }
+    ).and_return(double("AdminMailer", deliver: true))
+    SpamReport.run
+  end
+
+  it 'increases the spam score if the work creator has used more than one IP address"' do
+    third_spam.ip_address = "192.168.11.1"
+    third_spam.save
+    expect(AdminMailer).to receive(:send_spam_alert).with(
+      third_spam.pseuds.first.user_id => { score: 14, \
```
sarken: This shouldn't need the `\`

spec/models/spam_report_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+                                           work_ids: [first_spam.id, second_spam.id, third_spam.id] }
+    ).and_return(double("AdminMailer", deliver: true))
+    SpamReport.run
+  end
+
+  it 'Posting works in the recent ( but not too recent ) past reduces the score' do
+    expect(AdminMailer).to receive(:send_spam_alert).with(
+      third_spam.pseuds.first.user_id => { score: 13, \
```
sarken: This shouldn't need the `\`

spec/models/spam_report_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+                                           work_ids: [first_spam.id, second_spam.id, third_spam.id] }
+    ).and_return(double("AdminMailer", deliver: true))
+    SpamReport.run
+    create(:work, spam: false, posted: true, authors: second_spam.authors, created_at: 3.days.ago)
+    expect(AdminMailer).to receive(:send_spam_alert).with(
+      third_spam.pseuds.first.user_id => { score: 11, \
```
sarken: This shouldn't need the `\`

spec/models/spam_report_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    third_spam.ip_address = "192.168.11.1"
+    third_spam.save
+    expect(AdminMailer).to receive(:send_spam_alert).with(
+      third_spam.pseuds.first.user_id => { score: 14, \
+                                           work_ids: [first_spam.id, second_spam.id, third_spam.id] }
+    ).and_return(double("AdminMailer", deliver: true))
+    SpamReport.run
+  end
+
+  it 'Posting works in the recent ( but not too recent ) past reduces the score' do
```
sarken: `it "reduces the spam score if works are posted in the recent past"` would be more rspec-y, but I'm still not sure what this means by posting works in the past. Do you mean one of these things?
- it reduces the spam score if the works were created as drafts several days before posting
- it reduces the spam score if the works are backdated to the recent past
- it reduces the spam score if the user has posted other works in the recent past

zz9pzza: "it reduces the spam score if the user has posted other works in the recent past"

Looking at the relevant code.
  # Scoring rules:
  # For every new spam work, add 4 to the score
  # For every new non-spam work, add 1
  # For every older non-spam work,
  # decrease the score by 2
    count = user.works.
                 where("works.created_at > ? AND works.created_at < ?",
                       recent_date,
                       new_date).
                 posted.not_spam.count
    score -= (count * 2)

sarken: It sounds like that (`reduces the spam score if the user has posted other works in the recent past`) would make for a good description, then! The current description (`reduces the spam score if works are posted in the recent past`) sounds like it means the _newest_ work, which is why it's confusing.

spec/models/spam_report_spec.rb
```diff
@@ -0,0 +1,56 @@
+require "spec_helper"
+
+describe SpamReport do
+  let(:first_spam) { create(:work, spam: true, posted: true) }
+  let(:second_spam) { create(:work, spam: true, posted: true, authors: first_spam.authors) }
+  let(:third_spam) { create(:work, spam: true, posted: true, authors: second_spam.authors) }
+
+  before(:example) do
+    ArchiveConfig.SPAM_THRESHOLD = 10
```
redsummernight: Will this change to `ArchiveConfig.SPAM_THRESHOLD` stick around after the test?

zz9pzza: It is only used by this model, so it is not significant if it does. 

If needed I can restore the old value ?

redsummernight: You can stub it: `allow(ArchiveConfig).to receive(:SPAM_THRESHOLD).and_return(10)`.

zz9pzza:
@sarken @redsummernight do you want to have another look ?

redsummernight:
Looks good to me.

PR #2865, opened 2017-04-18
AO3-4733 Extend tests for wrangling_guidelines_controller and correct route for reordering

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4733

## Purpose

More tests.
Change "reorder" route to use the actual action name "update_positions"

The "confirm_delete" route has no action and will be implemented in [AO3-4974](https://otwarchive.atlassian.net/browse/AO3-4974).

## Testing

No pain, only tests. Check that coverage for the controller is now at 100%.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -530,6 +530,9 @@ Otwarchive::Application.routes.draw do
     end
   end
   resources :wrangling_guidelines do
+    member do
+      get :confirm_delete
+    end
     collection do
       get :manage
       post :update_positions
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -530,15 +530,12 @@ Otwarchive::Application.routes.draw do
     end
   end
   resources :wrangling_guidelines do
-    member do
-      get :confirm_delete
-    end
     collection do
       get :manage
-      post :reorder
+      post :update_positions
     end
   end
-  
+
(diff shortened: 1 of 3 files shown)


config/routes.rb
```diff
@@ -530,15 +530,12 @@
     end
   end
   resources :wrangling_guidelines do
-    member do
-      get :confirm_delete
```
sarken: We should really implement this rather than remove it.

redsummernight: I forgot that `link_to`'s `confirm` option requires JS. I'll revert this.

redsummernight:
Some tests failed because an admin appears where a user should be.

```
      NoMethodError:
        undefined method `is_author_of?' for #<Admin:0x0000000ab38648>
```

Looks like a faked admin login isn't unfaked properly... I'm not sure why yet.

zz9pzza:
I believe it is because spec/support/login_macros.rb has two definitions of fake_logout, I did think I have a pull to get rid of it but I can't find it right now.

https://github.com/otwcode/otwarchive/blob/master/spec/support/login_macros.rb

I would try deleting the top one

redsummernight:
Looks like User.current_user is the only `cattr_accessor` in the code base right now. We shouldn't see more errors like this.

PR #2866, opened 2017-04-18
AO3-4957 Downgrade nokogumbo

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4957

## Purpose

Make it so that leaving a comment like

Hello<
world 

Does not cause the unicorn to die, giving the user a 50x error

## Testing

Leave a comment like:

Hello<
world 

And ensure that it appears

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -47,7 +47,7 @@ gem 'whenever', '~>0.6.2', :require => false
 gem 'nokogiri', '>= 1.7.1'
 gem 'mechanize'
 gem 'sanitize'
-# Until there is a working 
+# Until there is a working solution to
 # https://otwarchive.atlassian.net/browse/AO3-4957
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -47,6 +47,10 @@ gem 'whenever', '~>0.6.2', :require => false
 gem 'nokogiri', '>= 1.7.1'
 gem 'mechanize'
 gem 'sanitize'
+# Until there is a working
```
sarken: The gem change is good, but I think you forgot to complete your sentence here. Do you mean something like, 
```
Until there is a working solution to
https://otwarchive.atlassian.net/browse/AO3-4957
```

sarken:
Still looks good -- Scrutinizer is just being fail-y! (Also, just a reminder to please update the issue on JIRA when you submit a pull request.)

PR #2868, opened 2017-04-21
AO3-4965 Try to fix works showing in filters when marking canonical/synonym.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4965

## Purpose

According to the bug report, there's been trouble with making tags canonical and synning them since March 29. I have a suspicion about why that might be happening (though I haven't been able to verify it, since I think it's partly a question of scale), and I wrote this pull request to try to resolve the issue.

I'm not 100% sure why this is happening, but I believe that these two issues may be contributing to it:

1. The function `add_filter_taggings` (in the Tag model) is called asynchronously in two functions: `update_filters_for_canonical_change` and `update_filters_for_merger_change`. Both of those functions are currently called with a `before_update` hook. That means that the job to call `add_filter_taggings` is enqueued before the data is written to the database.

   The worker reloads the tag from the database when it performs its operations. So if the worker grabs the job quickly enough, it'll load the old data from the database. That means that in the first few lines of `add_filter_taggings`, `self.filter` will be nil, and the worker will bail out without adding any FilterTaggings.

   I replicated this behavior in automated tests by changing the definition of `Tag.async` to just use `Resque.enqueue` without checking the environment. All Resque jobs are inlined in the test environment regardless, so the only difference is that the old way of inlining called the function on the same Tag object (instead of loading a separate version of the tag from the database, the way that the Resque job would).

   To fix this issue, I switched the callbacks to be `after_update` callbacks instead. (And I also changed a few function calls from asynchronous to synchronous, since the tests were expecting the association data to be updated immediately, and the new `Tag.async` definition no longer does that.)

2. I suspect that the issue that I described above may contribute to [AO3-3830](https://otwarchive.atlassian.net/browse/AO3-3830), so I'm not sure this is a new issue. But apparently the big problems started on March 29, which is when the new `taggings_count` code went live (as far as I can tell), so my guess is that they might be linked.

   When you view the Edit page for a small tag (< 1000 uses), the new `taggings_count` code will automatically recompute the taggings count, set the value on REDIS, and add the tag to the `tag_update` list on REDIS -- even if the taggings count hasn't changed. This means that soon after you view the Tag Edit page, the `Tag.write_redis_to_database` worker is guaranteed to do a batch update on ~1000 tags, including the tag you're trying to edit.

   If you try to save changes to the tag during that transaction, the `before_update` calls will go off immediately, enqueuing the job to add the new FilterTaggings. However, because there's an ongoing transaction with 1000 tags, the tag update code may have to wait a little longer than usual to get a lock on the row and actually perform the update. It seems to me that that delay might increase the chances of `add_filter_taggings` being called before the data is actually saved, resulting in a lot of missing FilterTaggings. I suspect that's why the issue got worse right after the `taggings_count` update.

   To try to address this, I slightly modified the `taggings_count` code so that it only adds a tag to the `tag_update` list on REDIS if the value has actually changed. This should hopefully reduce the chances of delays introduced by conflicting locks (as well as reducing the load on both REDIS and MySQL).

## Testing

I'm not sure how to test it manually, since I think it's a problem stemming from sheer volume. I think this would have to be tested on Beta.

## Credit

tickinginstant, she/her


spec/models/tag_spec.rb
```diff
@@ -522,4 +522,107 @@
...
+          synonym.attributes = { syn_string: fandom.name }
+          synonym.save!
+
+          expect(fandom.sub_tags.reload).to contain_exactly(sub)
+          expect(fandom.meta_tags.reload).to contain_exactly(meta)
+          expect(synonym.sub_tags.reload).to contain_exactly()
+          expect(synonym.meta_tags.reload).to contain_exactly()
+        end
+
+        it "adding a synonym transfers its associations", :pending do
```
tickinginstant: I wrote this test to make sure that I didn't break `add_merger_associations` by changing it from asynchronous to synchronous, but it turns out that this also fails on the master branch as well, since the Uncategorized Fandom media is added to the synonym fandom. I don't think that's the desired behavior, but I wanted to keep the scope of this pull request fairly small and not add in loads of extra changes.

app/models/tag.rb
```diff
@@ -54,8 +53,17 @@ def taggings_count_cache_key
...
+    old_redis_value = REDIS_GENERAL.getset("tag_update_#{id}_value", value)
+
+    # If the value hasn't changed from the saved version or the REDIS version,
+    # there's no need to write an update to the database, so let's just bail
+    # out.
+    return value if value == old_redis_value && value == taggings_count_cache
+
+    # If we've reached here, then the value has changed, and we need to make
+    # sure that the new value is written to the database.
     REDIS_GENERAL.sadd("tag_update", id)
```
zz9pzza: I think the old code does this already:

https://redis.io/commands/sadd

"Add the specified members to the set stored at key. Specified members that are already a member of this set are ignored."

And treating it as a set means a maximum of one call, rather than a maximum of two.

tickinginstant: It's still a set either way, so it won't be called twice (at least until after it's been processed). But suppose that the taggings count hasn't changed, and someone views the Edit page for a small tag (or even a large one). The taggings count will be recomputed and it will be added to REDIS and the `tag_update` set for an update, even though absolutely nothing has changed since the last time the taggings count was written to the database.

zz9pzza: Ok I think I understand what you mean, however the redis database is not authoritative. We are using the set as the list of the values we are wanting to change. From what you say I would make that change in https://github.com/otwcode/otwarchive/blob/master/app/models/tag.rb#L62 however that is the only place it is being called as well and it is cleaner there.

app/models/tag.rb
```diff
@@ -29,9 +29,8 @@ def self.write_redis_to_database
       Tag.transaction do
         batch.each do |id|
           value = REDIS_GENERAL.get("tag_update_#{id}_value")
-          sql = []
-          sql.push("taggings_count_cache = #{value}") unless value.blank?
-          Tag.where(id: id).update_all(sql.join(",")) unless sql.empty?
+          next if value.blank?
+          Tag.where(id: id).update_all(taggings_count_cache: value.to_i)
```
zz9pzza: This is definitely cleaner, the sql list is a hold over from earlier code.

app/models/tag.rb
```diff
@@ -1023,7 +1033,7 @@ def remove_sub_filters(sub_tag)
   before_update :check_canonical
   def check_canonical
     if self.canonical_changed? && !self.canonical?
-      self.async(:remove_canonical_associations)
+      remove_canonical_associations
```
zz9pzza: I am a bit nervous about moving these from async to sync as it can definitely slow things down on the page that these are called from

app/models/tag.rb
```diff
@@ -735,8 +747,8 @@ def filter
     self.canonical? ? self : ((self.merger && self.merger.canonical?) ? self.merger : nil)
   end
 
-  before_update :update_filters_for_canonical_change
-  before_update :update_filters_for_merger_change
+  after_update :update_filters_for_canonical_change
```
zz9pzza: As an aside I think your right about this bit ( the important bit )

app/models/tag.rb
tickinginstant: The issue with putting it in `taggings_count=` is that there's not enough information. Suppose that you have a small tag with 10 uses. Now suppose that someone tags a work with that tag, and then immediately removes the tag. (In other words, increase the `taggings_count` by 1, and then immediately decrease the `taggings_count` by 1.) The first change would force it to be added to the `tag_update` set, but wouldn't immediately modify `taggings_count_cache`. During the second update, the code would compute the tag count, discover that the value is 10, cross-check it against the `taggings_count_cache` (which is still 10, because the new value of 11 hasn't been written to MySQL yet), and decide not to write the value to REDIS. The result would be that `taggings_count_cache` would be (incorrectly) set to 11, and the second update wouldn't go through.

The REDIS data isn't authoritative, but if the REDIS value, the `taggings_count_cache` value, and the recomputed value are all the same, then we know that we don't need to change the `taggings_count_cache` anytime soon. (Since the value is correct, and we know that there aren't any impending updates from the values on REDIS that could cause it to be incorrect.)

app/models/tag.rb
```diff
@@ -54,8 +53,17 @@ def taggings_count_cache_key
...
+
+    # If the value hasn't changed from the saved version or the REDIS version,
+    # there's no need to write an update to the database, so let's just bail
+    # out.
+    return value if value == old_redis_value && value == taggings_count_cache
+
+    # If we've reached here, then the value has changed, and we need to make
+    # sure that the new value is written to the database.
     REDIS_GENERAL.sadd("tag_update", id)
-    REDIS_GENERAL.set("tag_update_#{id}_value", value)
```
zz9pzza: I just notice why are we not setting the value here ?

app/models/tag.rb
zz9pzza: I can see what your saying, I am not convinced that dropping the write is worth the extra complexity. However I am happy enough that your have explained the why :)

app/models/tag.rb
tickinginstant: They're only changing the associations, not the FilterTaggings or the associated works. (And those changes are generally made on the tag edit pages, which already have to load all of the associations in order to display them.) But I can change them back to async and try to modify the tests instead, if that would be preferable.

app/models/tag.rb
tickinginstant: The `getset` command has already set the value: https://redis.io/commands/getset

zz9pzza: Happy with that and gets rid of my worry about adding "load" to the redis server

app/models/tag.rb
zz9pzza: I will have a look to convince myself about the order of magnitude those things will run in so let's hold off on that for a bit ( As I say I am being nervous and I will try and see if I can back it up with data ).

zz9pzza: Looking at the two, I still think remove_canonical_associations might take a while.
Why does the async make any difference to the tests as I thought all async calls were synchronous in test ?

app/models/tag.rb
tickinginstant: I think the only time it'd *really* matter is one of those 0/1 situations again. (e.g. Someone uses a tag for the very first time, and then the `taggings_count_cache` is set to 1. Then they accidentally delete the tag from their work, causing the `taggings_count` code to try to set it to 0. If they add back the tag quickly enough, the `taggings_count_cache` might stay at 0 and not show up in the wrangling interface.)

It's an extremely unlikely occurrence, but even so I figure it's good to prevent these sorts of issues from popping up. :)

zz9pzza: We could make a change to the rake that deletes 0 use tags so that any tag which was not deleted and had uses had the count set correctly.

https://github.com/otwcode/otwarchive/blob/master/lib/tasks/tag_tasks.rake#L41

if this was t.taggings_count.zero? rather than t.taggings.count.zero?

app/models/tag.rb
tickinginstant: They're synchronous, but they don't act on the same object in memory. Changes in one won't necessarily be reflected in the other. When setting `syn_string`, for instance, it updates the value of the merger on the object that you're modifying. But when `add_merger_associations` is called asynchronously in the `syn_string=` function, the new merger hasn't yet been saved to the database, so it'll load the version of the tag with no merger (and it won't bother transferring any associations).

I may be a little overzealous -- it might be okay to call `remove_canonical_associations` asynchronously (although it seems less computationally intensive than `add_merger_associations`, which does require special handling). But it is in a `before_update` callback, which makes me nervous.

app/models/tag.rb
tickinginstant: I think it'd be good to keep that as `t.taggings.count.zero?`, just in case, IDK, there's some kind of obscure caching issue that causes it to incorrectly report the taggings count as 0. I'm not even sure that could happen, but it seems to me that "better safe than sorry" might be a good philosophy to keep when deleting tags.

But forcing the rake task to recalculate the `taggings_count` for tags that are incorrectly marked with a `taggings_count_cache` of 0 seems like a good idea.

app/models/tag.rb
tickinginstant: Also, I made a mistake -- it's not just about the tests. Switching the definition of `Tag.async` causes some tests to fail, but I realize now that's actually because it's revealing more bugs in the existing code. (Stemming from the question of when the job is enqueued vs. when the required information for that job is written to the database.) I can't just change `add_merger_associations` back to async and fix up the tests.

There are a couple of more complex ways to call it asynchronously, if that's what you think is best, but it would be a more extensive change than just reverting to `async(:add_merger_associations)`, and it might take a while to iron out the bugs.

zz9pzza: I am pretty sure it would work as it is, what I might suggest is that we deploy this by itself so if it causes issues then we can roll it back without causing any other issues.

tickinginstant: I'm a little confused -- does that mean you want me to split this pull request?

zz9pzza: No its me talking about how we would deploy it. Nothing to worry about,

tickinginstant:
I restructured the code to allow remove_canonical_associations and add_merger_associations to be called asynchronously again, and switched remove_association back to async as well. (I still think that at the very least it should be okay to call remove_association synchronously, but I'm guessing it might be easier to merge this way.)

app/models/tag.rb
```diff
@@ -800,14 +812,30 @@ def filter
...
-  before_update :update_filters_for_merger_change
+  after_commit :update_associations, on: :update
+
+  # If we have a new merger, transfer our old associations to the merger.
+  # If we're newly non-canonical, clean up the associations that relied on us
+  # being canonical: mergers, child tags, meta/sub tags, favorite tags.
+  def update_associations
+    if self.saved_change_to_merger_id? && self.merger
+      async(:add_merger_associations)
+      # TODO Should this transfer favorite tags to the merger instead?
```
sarken: Just confirming that yep, that would definitely be the ideal

app/models/tag.rb
```diff
@@ -800,14 +812,30 @@ def filter
...
+      async(:remove_favorite_tags)
+    elsif self.saved_change_to_canonical? && ! self.canonical?
```
sarken: Could this be `!self.canonical?`? We usually forgo the space

app/models/tag.rb
sarken: It does sometimes incorrectly report _other_ numbers, including negatives, so it certainly seems plausible it would incorrectly report zero as well. I agree that better save that sorry is a good approach when deleting anything.

PR #2869, opened 2017-04-21
AO3-4943 Reduce the number of broken downloads

zz9pzza:
https://otwarchive.atlassian.net/browse/AO3-4943 

## Purpose

Make it more likely that download generation works by clearing up relatively common errors.

## Testing

Create broken works for example one named "-Hello world-" and see if you can download the pdf.


app/models/work.rb
```diff
@@ -797,7 +797,7 @@ def download_authors
   end
 
   def download_title
-    string = title.to_ascii.gsub(/[^[\w _-]]+/, '')
+    string = title.to_ascii.gsub(/[^[\w _]+/, '')
```
sarken: I think Hound is right here and this is causing all the tests to fail.

zz9pzza: I had seen that before, sorry.

app/views/downloads/_download_preface.html.erb
```diff
@@ -1,24 +1,25 @@
...
       <% tags = @work.tag_groups[type] %>
-      <% unless tags.blank? || (type == "Warning" && hide_warnings?(@work)) || (type == "Freeform" && 
+      <% unless tags.blank? || (type == "Warning" && hide_warnings?(@work)) || (type == "Freeform" &&
 hide_freeform?(@work)) %>
         <dt><%= type.constantize::NAME %>:</dt>
         <dd><%= tags.map {|t| link_to(t.name, tag_url(t))}.join(", ").html_safe %></dd>
       <% end %>
     <% end %>
-    <% unless @work.serial_works.blank? %>
+    <% @series_list = @work.serial_works.reject{|x| x.series.nil?} %>
```
sarken: I think `series_list` can probably be a local variable instead of an instance variable. Also, if Hound ran on views, it would say there should be spaces inside the curly braces, like so: `{ |x| x.series.nil? }` (Also, we usually use letters that relate to the thing we're handling, so maybe `sw` for serial work rather than `x`?) But this file is so messy a couple of space are definitely not the end of the world.

app/views/downloads/_download_preface.html.erb
```diff
@@ -39,29 +40,29 @@ hide_freeform?(@work)) %>
...
-    <% end %> 
+    <% end %>
   <% end %>
 
   <% unless @work.parents.blank? %>
     <dl>
       <dt><%= ts('This work was inspired by') %></dt>
       <dd>
-        <%= @work.parent_work_relationships.map{|rw| "#{link_to(rw.parent.title, 
+        <%= @work.parent_work_relationships.reject{|x| x.parent.nil?}.map{|rw| "#{link_to(rw.parent.title,
```
sarken: Just for the sake of making this a bit shorter, what if we did the same thing here that we did with the series list above? That is:
```
<% parent_work_list = @work.parent_work_relationships.reject{ |wr| wr.parent.nil? } %>
<%= parent_work_list.map{|rw| "#{link_to(rw.parent.title,
```
Just a thought, though!

PR #2871, opened 2017-04-22
AO3-4978 Remove duplicated code

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4978

## Purpose

Remove a duplicated method in rspec  and one method which appeared twice in the tags but was never used.

## Testing

I don't believe this requires any testing...


redsummernight:
:+1: for deleting `fake_logout` duplicate.

PR #2874, opened 2017-04-23
AO3-4981 test for i18n of invite email

cesy:
https://otwarchive.atlassian.net/browse/AO3-4981

Adds an automated test for the invitation email that just got internationalised.

### DIFF
--- a/spec/mailers/user_mailer_spec.rb
+++ b/spec/mailers/user_mailer_spec.rb
@@ -197,4 +197,48 @@ describe UserMailer do
       end
     end
   end
+  
+  describe "invite request" do
+    token = 'abc123'
+
+    before(:each) do
+      @invitation = FactoryGirl.create(:invitation, token: token)
+    end
+
+    let(:email) { UserMailer.invitation(@invitation.id).deliver }
+
+    # Test the headers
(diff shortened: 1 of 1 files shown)


spec/mailers/user_mailer_spec.rb
```diff
@@ -197,4 +197,52 @@
...
+      it 'should have text contents' do
+        expect(get_message_part(email, /html/)).to include("like to join us, please sign up at the following address")
+      end
+      
+      it 'should not have missing translations' do
+        expect(get_message_part(email, /html/)).not_to include("translation missing")
+      end
+      
+      it 'should not have exposed HTML' do
+        expect(get_message_part(email, /html/)).not_to include("style")
```
sarken: I think you'll need to look for something like `&lt;` or `&gt;` here instead of `style`. This seems to be looking at the email's HTML, not the text that's visible to a human reading it, so it should and does find `style`. `&lt;` or similar would indicate the email had a `<` that was visible to the user, which would generally indicate exposed HTML. (It's not foolproof for emails where we show work titles and summaries that might legitimately contain a `<`, but it should work well enough for this invitation email.)

PR #2875, opened 2017-04-23
AO3-4884 fixing exposed HTML

cesy:
https://otwarchive.atlassian.net/browse/AO3-4884

This should sort the exposed HTML.

Test run is still in progress, so it may be worth waiting for that to finish before merging.

### DIFF
--- a/app/views/user_mailer/invitation.html.erb
+++ b/app/views/user_mailer/invitation.html.erb
@@ -5,7 +5,7 @@
 
   <p>
     <% if !@user_name.blank? %>
-      <%= t ".has_invited", user_name: style_bold(@user_name) %>
+      <%= t ".has_invited", user_name: style_bold(@user_name).html_safe %>
     <% else %>
       <%= t ".been_invited" %>
     <% end %>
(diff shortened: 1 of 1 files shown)


app/views/user_mailer/invitation.html.erb
```diff
@@ -5,7 +5,7 @@
 
   <p>
     <% if !@user_name.blank? %>
-      <%= t ".has_invited", user_name: style_bold(@user_name) %>
+      <%= t ".has_invited", user_name: style_bold(@user_name).html_safe %>
```
sarken: This needs to be `<%= t(".has_invited", user_name: style_bold(@user_name)).html_safe %>` or it will still produce escaped HTML

app/views/user_mailer/invitation.html.erb
```diff
@@ -5,7 +5,7 @@
 
   <p>
     <% if !@user_name.blank? %>
-      <%= t ".has_invited", user_name: style_bold(@user_name) %>
+      <%= t(".has_invited", user_name: style_bold(@user_name).html_safe) %>
```
sarken: Sorry, this still won't work -- the final `)` needs to be after `style_bold(@user_name)` and before `.html_safe`: `<%= t(".has_invited", user_name: style_bold(@user_name)).html_safe %>`

PR #2876, opened 2017-04-23
AO3-4982 Move travis to later os

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4982

## Purpose

Moves our tests to a supported os on travis

## Testing

No manual testing required

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: ruby
+dist: trusty
 cache: bundler
 sudo: required
 env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: ruby
-dist: trusty
+dist: xenial
 cache: bundler
 sudo: required
 env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: ruby
-dist: xenial
+dist: trusty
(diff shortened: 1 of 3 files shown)


sarken:
Is there any sort of documentation about this? I mean, it's passing on Travis, so I assume all is well, but it would be good to be able to look at some instructions and say, "Yep, we've done it right."

zz9pzza:
https://blog.travis-ci.com/2017-04-17-precise-EOL is the best I have

ariana-paris:
According to that document, it sounds as if we could just sit on it and wait for the default on Travis to switch to Trusty later this year, but it's good to know the tests are still passing anyway, and I suppose it makes what OS we're expecting explicit. 

Are the database collation changes related to Trusty or just something our Travis setup happened to also need?

zz9pzza:
The changes to collation are required by the upgrade of mysql which comes with trusty,

zz9pzza:
I think it would be better to do this as soon as possible.

ariana-paris:
I don't see the urgency, but given we have to make the switch, I guess we'll have longer to see if the changes cause any issues on travis (seems unlikely since the tests pass anyway). Can you add the link to their documentation to the ticket for future reference on why we did this?

PR #2877, opened 2017-04-23
AO3-4702 Resurrect autocomplete feature tests

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4702

## Purpose

Resurrect autocomplete feature tests.

## Testing

Check that the jQuery monkey patch does not end up in a non-test environment.


features/step_definitions/autocomplete_steps.rb
```diff
@@ -41,147 +46,171 @@
     when "Characters"
       "ellen"
     when "Relationships"
-      "stiel"
-    when "Your Tags"
+      "cast"
+    when "Your tags"
```
redsummernight: `case` are indented in 2 different ways in the project: as deep as case, or 2 spaces from the start of the statement. Which style should I use?

We can get Hound to agree: http://rubocop.readthedocs.io/en/stable/cops_style/#stylecaseindentation

features/step_definitions/autocomplete_steps.rb
```diff
@@ -41,147 +46,171 @@
     when "Characters"
       "ellen"
     when "Relationships"
-      "stiel"
-    when "Your Tags"
+      "cast"
```
redsummernight: I changed "stiel" to "cast". We set up Redis to look for prefixes for each character in a ship ("c", "ca", "cas", etc), searching for a suffix doesn't work.

features/step_definitions/autocomplete_steps.rb
```diff
@@ -34,154 +39,178 @@
...
-    end
-  step %{I fill in "#{fieldname}" with "#{text}"}
+         when "Fandoms"
+           "sup"
+         when "Additional Tags"
+           "alt"
+         when "Characters"
+           "ellen"
+         when "Relationships"
+           "cast"
```
redsummernight: I changed "stiel" to "cast". We set up Redis to look for prefixes for each character in a ship ("c", "ca", "cas", etc), searching for a suffix doesn't work.

features/other_a/autocomplete.feature
```diff
@@ -1,92 +1,91 @@
...
-#     And I view the work "All The Nice Things"
-#     And I follow "Add To Collections"
-#     And I fill in "collection_names" with "Issue"
-#   Then I should see "jb_fletcher" in the autocomplete
-#     And I should see "robert_stack" in the autocomplete
+  @javascript
+  Scenario: Only matching canonical tags should appear in autocomplete, and searching for the same data twice should produce same results
+    Given I am logged in
+      And a set of tags for testing autocomplete
+       And I go to the new work page
```
sarken: Could you make this `And` line up with the one above, please? Also, if you could break the scenario descriptions onto separate, shorter lines, that would be great.

features/other_a/autocomplete.feature
```diff
@@ -1,92 +1,91 @@
...
-    And the pseud autocomplete should not contain "funny"
-    And the pseud autocomplete should not contain "different_user (funny)"
-    
-    
-  
+  @javascript
+  Scenario: For fandom-specific autocomplete, if a fandom is entered then only characters/relationships within the fandom should appear in autocomplete
+    Given I am logged in
+      And a set of tags for testing autocomplete
+       And I go to the new work page
```
sarken: Could you fix the indenting here as well?

PR #2878, opened 2017-04-23
AO3-4983 Use multiple redis servers for travis

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4983

## Purpose

Splits the redis work in to multiple servers so we are more likely to spot errors when we shard the redis functions.

## Testing

Effects travis only no manual testing


.travis.yml
```diff
@@ -32,6 +32,7 @@ script:
 before_script:
   - bash script/travis_configure.sh
   - bash script/travis_elasticsearch_upgrade.sh
+  - bash script/travis_mutliple_redis.sh
```
redsummernight: It should be "multiple".

sarken: (Both here and in the file name.)

PR #2879, opened 2017-04-24
AO3-4979 Set the admin cookie for the front end

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4979

## Purpose

Ensure that a cookie is set if an admin is logged in ( there could be a one page lag either way )

## Testing

see issue


app/controllers/application_controller.rb
```diff
@@ -31,7 +31,16 @@ def check_for_flash
   # redirect to the logout page
   before_filter :logout_if_not_user_credentials
   def logout_if_not_user_credentials
```
sarken: It seems a little suboptimal to include code for admins in a method that is specifically for users. Would it be too much extra work to make a new method for handling admin credentials?

If you can't move it to a dedicated method, it would be good to at least update the comment before this method, since it currently only refers to user_credentials.

PR #2881, opened 2017-04-25
AO3-4579 Challenge signup/creation email address warning

pupupuri:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4579

## Purpose

informs users signing up for challenges that the creator of the challenge may have access to their email address and warns challenge creators about using email addresses for anything other than running the challenge. 

i think (?) javascript would have been necessary to dynamically add a mandatory tick box to the page when someone chooses to make their collection a challenge, and to remove the tick box if/when they unselect a challenge. in which case there would have to be a noscript workaround for that too. or if you have the mandatory tick box there all the time regardless of whether or not the user is making their collection a challenge, then people might get annoyed if they ignore that part of the form and then hit submit and receive an error for not ticking a box they didn't think was relevant to them... so i thought it'd be simpler and would require less code to just add a note in the appropriate area. but if a mandatory tick box is a more solid solution i can work on that on instead

## Testing

please make sure that on challenge signup pages there's a blue warning above the "required information" note that warns the user about their email address possibly being made available

and also that in the challenge creation page there's a blue note warning the challenge creator that use of email addresses for anything other than running the challenge will lead to termination of their account


app/views/collections/_form.html.erb
```diff
@@ -164,6 +164,13 @@
         <dd>
           <%= select_tag :challenge_type, options_for_select(Collection::CHALLENGE_TYPE_OPTIONS, type) %>
         </dd>
+        <dt><strong><%= ts("Notice to Challenge Creators") %></strong></dt>
+        <dd>
+          <ul class="notes">
+            <li><%= ts("As a challenge owner, you may have access to challenge participants' email addresses.") %></li>
+            <li><%= ts("Use of those email addresses for any purpose other than running the challenge will lead to the termination of your account.") %></li>
+            </ul>
```
sarken: Could you make this line up with the opening tag?

app/views/collections/_form.html.erb
```diff
@@ -164,6 +164,13 @@
         <dd>
           <%= select_tag :challenge_type, options_for_select(Collection::CHALLENGE_TYPE_OPTIONS, type) %>
         </dd>
+        <dt><strong><%= ts("Notice to Challenge Creators") %></strong></dt>
```
sarken: Could you use sentence, rather than title, case for this?

pixelred:
Done ✌️

sarken:
No rush, but this has some conflicts that will need to be resolved before this can be merged!

pupupuri:
👍?

PR #2886, opened 2017-04-28
AO3-4993 Refactor url caching testing to use rspec rather than cucumber.

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4993

## Purpose

Move tests from cucumber to rspec

## Testing

No manual testing required.

### DIFF
--- a/spec/models/work_spec.rb
+++ b/spec/models/work_spec.rb
@@ -186,7 +186,7 @@ describe Work do
 
     it "should find works imported with irrelevant query parameters" do
       work = create(:work, imported_from_url: "http://lj-site.com/thing1?style=mine")
-      expect(Work.find_by_url("http://lj-site.com/thing1?style=other").to eq(work)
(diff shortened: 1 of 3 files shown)


spec/models/work_spec.rb
```diff
@@ -185,8 +185,18 @@
     end
 
     it "should find works imported with irrelevant query parameters" do
-      work = create(:work, imported_from_url: 'http://lj-site.com/thing1?style=mine')
-      expect(Work.find_by_url('http://lj-site.com/thing1?style=other')).to eq(work)
+      url = "http://lj-site.com/thing1?style=mine"
```
ariana-paris: I'm confused - why have you changed this test? the whole point of it is to check that the `find_by_url` method sees "lj-site.com/thing1?**style=mine**" as the same as "lj-site.com/thing1?**style=other**". Can you change it back to how it was?

PR #2887, opened 2017-04-30
AO3-4995 Start migrating from _url to _path helpers

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4995

## Purpose

The reason for this is so https doesn't start redirecting to http.

## Testing

I would browse https://test.archiveofourown.org/ and note any time we loose the s in https

I would check that the mail downloads and feeds still have http in them.

If we havn't got all the _url its not a blocker to deploying in that moving nearer a better solution doesn't need us to be all the way there.


app/controllers/challenge_assignments_controller.rb
```diff
@@ -245,7 +245,7 @@ def default
     @challenge_assignment.save
     @challenge_assignment.collection.notify_maintainers("Challenge default by #{@challenge_assignment.offer_byline}",
         "Signed-up participant #{@challenge_assignment.offer_byline} has defaulted on their assignment for #{@challenge_assignment.request_byline}. " +
-        "You may want to assign a pinch hitter on the collection assignments page: #{collection_assignments_url(@challenge_assignment.collection)}")
+        "You may want to assign a pinch hitter on the collection assignments page: #{collection_assignments_path(@challenge_assignment.collection)}")
```
redsummernight: This is for an email and should be absolute.

zz9pzza:
Thanks @redsummernight

PR #2890, opened 2017-05-02
AO3-4997 Resurrect autocomplete feature tests for tag wrangling

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4997

## Purpose

Enable pending autocomplete feature tests for tag wrangling.

Some "follow tag name link" actions become ambiguous when JavaScript is enabled because we have delete buttons with title containing tag names, so we need to switch to "Edit (tag name)".

On the tag edit page, we can refer to certain fields by their labels only when they don't already have any values pre-filled. If they do, their input fields will be preceded by a section of checkboxes to toggle existing values, and we have to refer to them by IDs. Example: "Add Fandoms" vs "tag_fandom_string_autocomplete".

Clean-up:

- Refer to the "Canonical" checkbox by its label
- Prefer the step "Given a (non)canonical (tag type)" over alternatives
- Fix a TODO in the media pages test

## Testing

Automated.

### DIFF
--- a/features/tags_and_wrangling/tag_wrangling_fandoms.feature
+++ b/features/tags_and_wrangling/tag_wrangling_fandoms.feature
@@ -2,6 +2,7 @@
 
 Feature: Tag Wrangling - Fandoms
 
+@javascript
 Scenario: fandoms wrangling - syns, mergers, autocompletes, metatags
 
   Given the following activated tag wrangler exists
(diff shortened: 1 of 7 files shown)


features/tags_and_wrangling/tag_wrangling_fandoms.feature
```diff
@@ -52,39 +53,36 @@ Scenario: fandoms wrangling - syns, mergers, autocompletes, metatags
...
-#    And I should see "Stargate SG-1" in the autocomplete
+    And I enter "Stargate" in the "Synonym of" autocomplete field
+  Then I should see "Stargate Atlantis" in the autocomplete
+    And I should see "Stargate SG-1" in the autocomplete
   When I fill in "Synonym of" with "Stargate SG-1"
-    And I fill in "tag_media_string" with "TV"
-  When "autocomplete tests with JavaScript" is fixed
-#    And I should see "TV Shows" int the autocomplete
-    And I fill in "tag_media_string" with "TV Shows"
+  When I enter "TV" in the "tag_media_string_autocomplete" autocomplete field
```
sarken: This should probably be an And, since it's right after a mother When. Also, the Then on the next line needs an indenting adjustment.

PR #2892, opened 2017-05-02
AO3-5001 Turn on strong params by default

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5001

## Purpose

While we think we have solved all the strong parameters issues. This changes the default to use strong parameters. This will help with new pulls before the rails 4.0 merge. And also reduce the risk that any issue will be found in the rails 4.0 change.

## Testing

I think we will have to live with a smoke test that things look good, its passes automated tests.


sarken:
Just noting that this has gone grey and the issue still needs updating

PR #2897, opened 2017-05-05
AO3-5004 Add tests for work URL detection in abuse reports

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5004

## Purpose

More tests for the abuse report model. Test cases from https://otwarchive.atlassian.net/browse/AO3-4257.

## Testing

Automated.

### DIFF
--- a/spec/models/abuse_report_spec.rb
+++ b/spec/models/abuse_report_spec.rb
@@ -28,7 +28,6 @@ describe AbuseReport do
   end
 
   context "invalid emails" do
-
     BAD_EMAILS.each do |email|
       let(:bad_email) {build(:abuse_report, email: email)}
       it "cannot be created if the email does not pass veracity check" do
@@ -36,7 +35,6 @@ describe AbuseReport do
         expect(bad_email.errors[:email]).to include("does not seem to be a valid address.")
       end
     end
-
   end
 
   context "invalid url" do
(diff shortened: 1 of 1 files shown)


spec/models/abuse_report_spec.rb
```diff
@@ -88,38 +86,75 @@
...
+    it_behaves_like "enough already", "http://archiveofourown.org/works/789/chapters/123"
+    it_behaves_like "enough already", "http://archiveofourown.org/works/789/chapters/123#major-character-death"
+    it_behaves_like "enough already", "http://archiveofourown.org/works/789/chapters/123?ending=1"
+    it_behaves_like "enough already", "http://archiveofourown.org/works/789/chapters/123?ending=2#major-character-death"
+
+    # not the same work
+    it_behaves_like "alright", "http://archiveofourown.org/works/9009"
+    it_behaves_like "alright", "http://archiveofourown.org/works/78"
+    it_behaves_like "alright", "http://archiveofourown.org/works/7890"
+    it_behaves_like "alright", "http://archiveofourown.org/comments/show_comments?work_id=789"
```
sarken: Should this and `it_behaves_like "alright", "http://archiveofourown.org/chapters/123"` maybe be broken out into their own section instead of placed under `# not the same work`? They are technically the same work; they're just variations we didn't think were important to cover at the time.

redsummernight: Got you covered: now in their own section.

sarken: Awesome -- thanks!

PR #2898, opened 2017-05-05
AO3-4935 change "problems when logging in" link on new password page to a path

pupupuri:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4935

## Purpose

changes "problems when logging in" link on forgotten password page to a path, to prevent https users from being redirected to the http site

(i wasn't sure if making it a path in this case just meant removing the http/host but...)

## Testing

please log out and visit https://archiveofourown.org/passwords/new
click on the 'problems when logging in' link and make sure it doesn't redirect you to the http site

### DIFF
--- a/app/views/passwords/new.html.erb
+++ b/app/views/passwords/new.html.erb
@@ -1,20 +1,19 @@
 <!--Descriptive page name, messages and instructions-->
 <h3 class="heading"><%= ts("Forgotten your password?") %></h3>
 <p>
-  <%= ts("If you've forgotten your password, we can send you a temporary one.") %>
-  <%= ts("Please tell us the user name or email address you used when you signed up for your Archive account.") %>
+  <%= ts("If you've forgotten your password, we can send you a temporary one. Please tell us the user name or email address you used when you signed up for your Archive account.") %>
 </p>
 <p>
-  <%= ts("If the temporary password doesn't work, please visit %{problems_when_logging_in} for help.", :problems_when_logging_in => link_to(ts("Problems When Logging In"), admin_posts_path("1277"))) %>
+  <%= ts("If the temporary password doesn't work, please visit %{problems_when_logging_in} for help.", problems_when_logging_in: link_to(ts("Problems When Logging In"), admin_post_path("1277"))).html_safe %>
 </p>
 <!--/descriptions-->
 
 <!--main content-->
 <%= form_tag passwords_path do %>
   <p>
-    <%= label_tag 'reset_password_for', ts("Email address <strong>or</strong> user name:".html_safe) %>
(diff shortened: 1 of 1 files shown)


app/views/passwords/new.html.erb
```diff
@@ -1,19 +1,19 @@
 <!--Descriptive page name, messages and instructions-->
 <h3 class="heading"><%= ts("Forgotten your password?") %></h3>
 <p>
-  <%= ts("If you've forgotten your password, we can send you a temporary one.") %> 
+  <%= ts("If you've forgotten your password, we can send you a temporary one.") %>
   <%= ts("Please tell us the user name or email address you used when you signed up for your Archive account.") %>
 </p>
 <p>
-  <%= ts("If the temporary password doesn't work, please visit <a href=\"http://archiveofourown.org/admin_posts/1277\">Problems When Logging In</a> for help.".html_safe) %>
+  <%= ts("If the temporary password doesn't work, please visit %{problems_when_logging_in} for help.", :problems_when_logging_in => link_to(ts("Problems When Logging In"), admin_posts_path("1277"))) %>
```
sarken: Unfortunately, this currently produces the following on the forgot password page: `If the temporary password doesn't work, please visit <a href="/admin_posts.1277">Problems When Logging In</a> for help.`

However, we can fix it! First, it needs an `.html_safe` at the end, after all the parentheses, so the HTML isn't escaped:

`<%= ts("If the temporary password doesn't work, please visit %{problems_when_logging_in} for help.", :problems_when_logging_in => link_to(ts("Problems When Logging In"), admin_posts_path("1277"))).html_safe %>`

Then the helper should be `admin_post_path("1277")` rather than `posts` -- when there's an `s`, it will generally point to an index, whereas the singular will point to an individual post (or work, or bookmark, etc):

`<%= ts("If the temporary password doesn't work, please visit %{problems_when_logging_in} for help.", :problems_when_logging_in => link_to(ts("Problems When Logging In"), admin_post_path("1277"))).html_safe %>`

And you can also use JSON style rather than the old hash rocket here and do `problems_when_logging_in: link_to` etc.

This page doesn't get touched very often, so if you're game for making a few more tweaks, it would be excellent if you had time to:
- combine the two translation strings in the first paragraph (one `ts()` per paragraph is generally good)
- change the few instances of single quotes to double
- remove the colon from the form label

If not, though, no worries!

pixelred: will do! as soon as i have a minute i'll get to work on this. i'm moving in a week but after that i should have some time.

pixelred: done ✌️

PR #2899, opened 2017-05-05
AO3-4981 removing duplicate test

cesy:
https://otwarchive.atlassian.net/browse/AO3-4981 was the original that added the duplicate - do we need a new Jira for this?


sarken:
Yes, please re: new issue. (The only time a PR doesn't need a new issue is if we're fixing something that's broken on test.)

PR #2900, opened 2017-05-05
AO3-4714 Expand serial works controller test

cesy:
https://otwarchive.atlassian.net/browse/AO3-4714

### DIFF
--- a/spec/controllers/serial_works_controller_spec.rb
+++ b/spec/controllers/serial_works_controller_spec.rb
@@ -1,0 +1,31 @@
+require 'spec_helper'
+
+describe SerialWorksController do
+  include LoginMacros
+  include RedirectExpectationHelper
+  let(:user) { create(:user) }
+  let(:series) { create(:series, pseuds: user.pseuds) }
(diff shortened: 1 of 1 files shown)


spec/controllers/serial_works_controller_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+  it "will fail if you're not the owner" do
+    fake_login
+    delete :destroy, id: first_work.id
+    it_redirects_to_with_error(series_path(series), "Sorry, you don't have permission to access the page you were trying to reach.")
+  end
+  
+  it "redirects to series when destroying one of many" do
+    fake_login_known_user(user)
+    delete :destroy, id: first_work.id
+    # it doesn't give you any success message - this might be worth improving in future
```
sarken: Can you make an issue for this so we remember to do it? If the issue included a reminder to update these tests, that would be great, and then you could make these comments into TODOs that reference the issue number.

cesy: https://otwarchive.atlassian.net/browse/AO3-5006 created

PR #2733, opened 2017-02-05
AO3-4852 Migrate to using RedirectExpectationHelper

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4852

## Purpose

Increase the number of checks that the tests do when running. There is a slight increase in the number of rspec tests...

## Testing

No manual testing required.

PR #2901, opened 2017-05-05
AO3-4947 Expanding collection item test

cesy:
https://otwarchive.atlassian.net/browse/AO3-4947

### DIFF
--- a/spec/controllers/collection_items_controller_spec.rb
+++ b/spec/controllers/collection_items_controller_spec.rb
@@ -115,7 +115,7 @@ describe CollectionItemsController do
         @approved_work_item = CollectionItem.find_by_item_id(@approved_work.id)
         fake_login_known_user(owner)
         delete :destroy, id: @approved_work_item.id
-        it_redirects_to_with_notice(collection_items_path(@collection), "Item completely removed from collection " & @collection.name)
+        it_redirects_to_with_notice(collection_items_path(@collection), "Item completely removed from collection " + @collection.name)
         expect(CollectionItem.where(item_id: @approved_work.id)).to be_empty
       end
     end
--- a/spec/controllers/collection_items_controller_spec.rb
+++ b/spec/controllers/collection_items_controller_spec.rb
@@ -36,12 +36,15 @@ describe CollectionItemsController do
         fake_login_known_user(owner)
         get :index, collection_id: @collection.name, rejected: true
         expect(response).to have_http_status(:success)
+        expect(assigns(:collection_items)).to include @rejected_work_item
         expect(response.body).to include @collection.title
         expect(response.body).to include @rejected_work.title
       end
 
       it "excludes approved and invited items" do
         get :index, collection_id: @collection.name, rejected: true
+        expect(assigns(:collection_items)).not_to include @approved_work_item
+        expect(assigns(:collection_items)).not_to include @invited_work_item
         expect(response.body).not_to include @approved_work.title
         expect(response.body).not_to include @invited_work.title
       end
--- a/spec/controllers/collection_items_controller_spec.rb
+++ b/spec/controllers/collection_items_controller_spec.rb
@@ -65,7 +65,7 @@ describe CollectionItemsController do
       end
     end
 
-    context "for collection with items in" do
+    context "for collection with items in, default approved parameter" do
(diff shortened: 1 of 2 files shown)


spec/controllers/collection_items_controller_spec.rb
```diff
@@ -15,7 +30,66 @@
       end
     end
 
-    context "create" do
+    context "rejected parameter for collection with items in" do
+      let(:owner) { @collection.owners.first.user }
+      render_views
```
sarken: Rather than testing the views, can we simply test that the value of the `@collection_items` instance variable is being correctly assigned in the controller? Like:
```
      it "includes rejected items" do
        fake_login_known_user(owner)
        get :index, collection_id: @collection.name, rejected: true
        expect(assigns(:collection_items)).to include @rejected_work_item
      end
```
Checking the page for the collection title and the work title feels a little like we're wandering away from testing the code that's actually in the controller.

spec/controllers/collection_items_controller_spec.rb
```diff
@@ -29,4 +103,26 @@
...
+    context "destroy" do
+      let(:owner) { @collection.owners.first.user }
+      render_views
+
+      it "removes things" do
+        @approvedworkitem = CollectionItem.find_by_item_id(@approvedwork.id)
+        fake_login_known_user(owner)
+        delete :destroy, id: @approvedworkitem.id
+        expect(response).to have_http_status(:redirect)
+        expect(flash[:notice]).to include "Item completely removed from collection"
```
sarken: This is checking for a flash message and a redirect, so can we use one of our redirect expectations helpers here? #2733 adds some new ones if the existing ones don't work.

cesy:
This one should also be ready for another review.

spec/controllers/collection_items_controller_spec.rb
```diff
@@ -15,7 +29,63 @@
...
+        fake_login_known_user(owner)
+        get :index, collection_id: @collection.name, rejected: true
+        expect(assigns(:collection_items)).not_to include @approved_work_item
+        expect(assigns(:collection_items)).not_to include @invited_work_item
+      end
+    end
+
+    context "invited parameter for collection with items in" do
+      let(:owner) { @collection.owners.first.user }
+
```
sarken: Not super important, but one way to reduce the repetition here might be to take these  two lines and make them into a before block:
```
fake_login_known_user(owner)
get :index, collection_id: @collection.name, invited: true
```
The same would also apply in the next context with
```
 fake_login_known_user(owner)
get :index, collection_id: @collection.name, approved: true
```
However, it's only a tiny bit of repetition.

spec/controllers/collection_items_controller_spec.rb
```diff
@@ -15,7 +29,63 @@
...
+      it "excludes approved and rejected items" do
+        fake_login_known_user(owner)
+        get :index, collection_id: @collection.name, invited: true
+        expect(assigns(:collection_items)).not_to include @approved_work_item
+        expect(assigns(:collection_items)).not_to include @rejected_work_item
+      end
+    end
+
+    context "for collection with items in" do
```
sarken: This should probably mention we're testing the approved page, since the previous context mentions we're testing the rejected page.

spec/controllers/collection_items_controller_spec.rb
```diff
@@ -29,4 +99,25 @@
...
+  describe "#destroy" do
+    before(:each) do
+      @collection = FactoryGirl.create(:collection)
+      @approved_work = FactoryGirl.create(:work)
+      @approved_work.add_to_collection(@collection) && @approved_work.save
+    end
+
+    context "destroy" do
+      let(:owner) { @collection.owners.first.user }
+      render_views
```
sarken: It's not immediately obvious to me at the moment, so I will ask: why is `render_views` necessary here?

spec/controllers/collection_items_controller_spec.rb
cesy: Fixed.

spec/controllers/collection_items_controller_spec.rb
cesy: I think it's left over from my attempts to get it working, and no longer needed - I'll test and see.

PR #2903, opened 2017-05-05
AO3-5005 clarifying duplicate test in user_mailer

cesy:
https://otwarchive.atlassian.net/browse/AO3-5005

### DIFF
--- a/spec/mailers/user_mailer_spec.rb
+++ b/spec/mailers/user_mailer_spec.rb
@@ -154,11 +154,12 @@ describe UserMailer do
     end
   end
   
-  describe "invite request" do
+  describe "invitation from a user request" do
     token = 'abc123'
 
     before(:each) do
-      @invitation = FactoryGirl.create(:invitation, token: token)
+      @user = FactoryGirl.create(:user)
+      @invitation = FactoryGirl.create(:invitation, token: token, creator_id: @user.id)
     end
 
     let(:email) { UserMailer.invitation(@invitation.id).deliver }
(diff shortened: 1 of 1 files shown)


spec/mailers/user_mailer_spec.rb
```diff
@@ -153,51 +153,7 @@
       end
     end
   end
-  
-  describe "invite request" do
```
sarken: It might be a good idea to keep these and get rid of the ones where all the examples start with `should`

spec/mailers/user_mailer_spec.rb
```diff
@@ -154,11 +154,12 @@
     end
   end
   
-  describe "invite request" do
+  describe "invitation from a user request" do
```
sarken: I just want to check -- this tests invitations sent directly by users to someone's email, while the other one tests invites sent by admins or the queue, correct?

If that's the case, rather than test for the phrase  "like to join us, please sign up at the following address," which appears in both versions of the email, we should probably test the part of the email that's different. (The version for invitations sent by users says "%{user} has invited you to join our beta!" while the other version says "You've been invited to join our beta!" I know apostrophes are a problem, and the login bit might be as well, but we can definitely test the "has invited you" and "been invited" parts of the sentences.)

cesy: Yes, that's right, good point - I'll expand it to test that.

sarken:
The test failures are unfortunately related to these tests, but I'm not sure if they've found a bug or if the tests just need a bit of tweaking.

cesy:
This should now be ready for another review.

sarken:
Link to test failures mentioned in previous comment, for reference: https://travis-ci.org/otwcode/otwarchive/jobs/232574151

PR #2904, opened 2017-05-06
AO3-5008 Validate that subscribable exists before saving

katieby:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5008

## Testing

1. Create a work.
2. Login as another user and go to the work created in (1). Keep this page open.
3. Login (in an incognito browser or on another machine) as the user from (1) and delete the work you created.
4. Go to the page from (2) and click 'Subscribe'. You should see the error "Subscribable can't be blank".

You can do this with series and users as well.

## References

This might fix (one of) the causes of https://otwarchive.atlassian.net/browse/AO3-4728 and the Subscription page 500 errors.

Note: This doesn't validate the actual type of the subscribable object, so you could still (somehow) subscribe to something that's not a work, series, or user, but not to anything that doesn't exist in the database.

## Credit

cresenne (she/her)


spec/models/subscription_spec.rb
```diff
@@ -1,57 +1,69 @@
...
       it "should be destroyed" do
-        expect { subscription.reload }.to raise_error
+        expect { subscription.reload }.to raise_error ActiveRecord::RecordNotFound
       end
     end
   end
 
+  context "when subscribable does not exist" do
+    before do
+      @work = create(:work)
```
redsummernight: You should make `work` a local variable.

sarken:
Hmm. It seems like this _will_ solve the issue of subscribing to things with invalid types -- we don't have anything in the database with the Серия type, for example, so a subscribable wouldn't be present. Could you add a test for that scenario, since it's the one we're specifically aiming to prevent?

sarken:
It looks like this pull request has merge conflicts -- would you be able to update it within the next week or so? If you can, we should be able to get it in the next release (0.9.195)

sarken:
Yay, thanks for updating!

PR #2906, opened 2017-05-06
AO3-4729 Accept 5 abuse reports per profile per month

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4729

## Purpose

Accept 5 abuse reports per profile per month, taking into account variations in the profile URL.

## Testing

See JIRA ticket.

### DIFF
--- a/app/models/abuse_report.rb
+++ b/app/models/abuse_report.rb
@@ -23,7 +23,7 @@ class AbuseReport < ActiveRecord::Base
   def clean_url
     # Work URLs: "works/123"
     # Profile URLs: "users/username"
-    if url.match(/(works\/\d+)/) || url.match(/(users\/\w+)/)
+    if url =~ /(works\/\d+)/ || url =~ /(users\/\w+)/
       uri = Addressable::URI.parse url
       uri.query = nil
       uri.fragment = nil
@@ -67,7 +67,7 @@ class AbuseReport < ActiveRecord::Base
     message = ts('URL has already been reported. To make sure the Abuse Team
                  can handle reports quickly and efficiently, we limit the number
                  of times a URL can be reported.')
-    if url.match(/\/works\/\d+/)
+    if url =~ /\/works\/\d+/
       # use "/works/123/" to avoid matching chapter or external work ids
       work_params_only = url.match(/\/works\/\d+\//).to_s
       existing_reports_total = AbuseReport.where('created_at > ? AND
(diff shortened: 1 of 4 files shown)


app/models/abuse_report.rb
```diff
@@ -6,24 +6,29 @@ class AbuseReport < ActiveRecord::Base
...
+  before_validation :clean_url, on: :create
+  def clean_url
+    # Work URLs: "works/123"
+    # Profile URLs: "users/username"
+    if url.match(/(works\/\d+)/) || url.match(/(users\/\w+)/)
+      uri = Addressable::URI.parse url
+      uri.query = nil
+      uri.fragment = nil
+      uri.path += "/" unless uri.path.end_with? "/"
+      self.url = uri.to_s
```
zz9pzza: I note there is nothing stopping the user from using one of the other hostnames to get another 5 reports in per url eg:

http://archiveofourown.org/works/832686/chapters/1584268
http://ao3.org/works/832686/chapters/1584268
http://archiveofourown.net/works/832686/chapters/1584268

Should we remove the host bit ( and the protocol bit )

redsummernight: Users cannot submit reports for ao3.org links because of `validates_format_of :url, with: app_url_regex` which uses `ArchiveConfig.APP_HOST`. The `url_is_not_over_reported` validation only looks at the "/works/123" part, so we don't need to clean up the link's protocol.

zz9pzza: Grand

redsummernight: I can add a test to show that rate limiting ignores protocol.

lib/tasks/after_tasks.rake
```diff
@@ -433,10 +433,10 @@ namespace :After do
   end
 
 
-  desc "Clean up work URLs for abuse reports from the last month"
-  task(:clean_abuse_report_work_urls => :environment) do
+  desc "Clean up URLs for abuse reports from the last month"
+  task(:clean_abuse_report_urls => :environment) do
```
zz9pzza: Is this something that should be run periodically ? or is it just something we run once on deploy ?

sarken: /me butts in

Just once when we deploy -- the new code with automatically tidy up any new reports

zz9pzza: I thought so given it's place 👍

PR #2910, opened 2017-05-06
AO3-4728 Fix subscriber page giving error 500

zz9pzza:
## Issue

## Purpose

If the subscriber page would give an error 500 then fix the subscriptions.

## Testing

We would either need to create a broken user or find one...


app/models/scheduled_database_check_job.rb
```diff
@@ -0,0 +1,18 @@
+class ScheduledDatabaseCheckJob
+  def self.perform(job_type)
+    case job_type
+    when 'clean_subscription_series'
+      Subscription.where(subscribable_type: "Series").includes(:subscribable).find_each do |s| 
+        s.delete if s.name.nil? && Series.find_by_id(s.subscribable_id).nil?
```
sarken: Have we come across any instances where the name was nil _but_ there was a series with the matching id? In cases like that, should we try to repair the subscription somehow, or maybe just delete it anyway? The name being nil is what causes the 500 error, so leaving the subscription in its present state doesn't seem like a good idea.

zz9pzza: 2.2.5 :018 > User.where("login IS ?",nil)
  User Load (0.5ms)  SELECT `users`.* FROM `users` WHERE (login IS NULL)
 => [] 
2.2.5 :019 > Series.where("title IS ?",nil)
  Series Load (0.4ms)  SELECT `series`.* FROM `series` WHERE (title IS NULL)
 => [] 
2.2.5 :020 > Work.where("title IS ?",nil)
  Work Load (0.4ms)  SELECT `works`.* FROM `works` WHERE (title IS NULL)
 => [] 

The other check is just be being paranoid when deleting data.

sarken: All righty. If we ever end up with stuff hanging around, we can reconsider our paranoia, but this should be good :+1:

elzj:
I'm really not sure we want to loop through all 36 million+ subscriptions and all of their subscribables every 24 hours. We might want to rethink this one.

sarken:
Marking deferred per Elz's comment.

sarken:
Also worth noting:
> Sarken [23:22]
> Are y'all still getting tickets from users whose Subscriptions pages are giving 500 errors? I haven't seen any come through the Database Magic board
> 
> Nary [23:43]
> Nope
> Not in a while. But also we can fix it ourselves.
> We have the troubleshoot account button
> And on the rare occasion we get one these days, that fixes the problem

zz9pzza:
```
2.3.4 :035 > Subscription.where(subscribable_type: "Work").joins("left join works on works.id = subscriptions.subscribable_id").where("works.id is NULL").count
 => 573
2.3.4 :036 > Subscription.where(subscribable_type: "User").joins("left join users on users.id = subscriptions.subscribable_id").where("users.id is NULL").count
 => 0
2.3.4 :037 > Subscription.where(subscribable_type: "Series").joins("left join series on series.id = subscriptions.subscribable_id").where("series.id is NULL").count
 => 0
```

zz9pzza:
2.6.5 :001 > Subscription.where(subscribable_type: "Work").joins("left join works on works.id = subscriptions.subscribable_id").where("works.id is NULL").count
 => 520
2.6.5 :002 > Subscription.where(subscribable_type: "User").joins("left join users on users.id = subscriptions.subscribable_id").where("users.id is NULL").count
 => 0
2.6.5 :003 > Subscription.where(subscribable_type: "Series").joins("left join series on series.id = subscriptions.subscribable_id").where("series.id is NULL").count
 => 0

2.6.5 :011 > Subscription.where(subscribable_type: "Work").joins("left join works on works.id = subscriptions.subscribable_id").where("works.id is NULL").find_each do |s|
2.6.5 :012 >      s.delete if s.name.nil? && Series.find_by_id(s.subscribable_id).nil?
2.6.5 :013?>   end
 => nil
2.6.5 :014 > Subscription.where(subscribable_type: "Work").joins("left join works on works.id = subscriptions.subscribable_id").where("works.id is NULL").count
 => 0

tickinginstant:
The issue that I see with this approach is that while the periodic task does save Support from having to run the "Troubleshoot User" option to fix user subscriptions, I don't think it prevents users from getting a 500 error if they visit their subscriptions page in between the time the subscription is broken, and the time the task to fix them is run.

To prevent 500 errors, the subscription code needs to be modified so that it doesn't error out if the subscribable is nil. Maybe by displaying "Deleted Work" or some such, so that the user can clean up their own subscriptions? Or by adding a few joins to the subscription query and computing the name from the database, so that the database can handle sorting/pagination?

(Either that, or the subscription code needs to be modified to prevent these nil subscribables. But I'm not sure there's a way to do that. Foreign key checks might work if the subscriptions table was set up differently, but as is, it's impossible. There might also be a way involving locking, but I think you'd have to lock the work every time a subscription is created, which could interfere with the work creator's ability to update popular works. So I'm not sure that's a good plan.)

PR #2913, opened 2017-05-07
AO3-5014 Use the same factories for unit and feature tests

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5014

## Purpose

Remove the factory definitions in features/support, which requires:

- Keep the language factory.
- Fix the archive_faq factory (there's no content field).
- Remove common_taggings from the character factory, because tag wrangling feature tests assume new characters are unwrangled; instead, add common_taggings only for the test that needs it.
- Fix a user step to use default pseuds instead of generated ones, so a work edit test doesn't need to reference pseuds.

## Testing

Automated.


zz9pzza:
Looks reasonable to me, but needs more eyes.

sarken:
Looks good to me too!

PR #2914, opened 2017-05-08
AO3-5015 Fix broken backup form link to tw.org

SammieLouise:
https://otwarchive.atlassian.net/browse/AO3-5015

Changed backup support form link to the correct url on tw.org (http://www.transformativeworks.org/contact_us)

There is currently a redirect in place, so there's not really anything to test. The hyperlink should be changed from http://www.transformativeworks.org/contact_us/contact-archive-support/ to http://www.transformativeworks.org/contact_us


SammieLouise:
Whoops! I thought I had, but that was just in the branch.

PR #2916, opened 2017-05-09
AO3-4275 Add task to clean up non-canonical warnings

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4275

## Purpose

Remove unused non canonical warnings and change used non canonical warnings to freeform

## Testing
See the issue. I wouldn't be too suprised if some of the tests in the issue fail without a memcached flush.


lib/tasks/tag_tasks.rake
```diff
@@ -49,6 +49,23 @@ namespace :Tag do
...
 
+  desc "Reset non canonical warnings"
+  task(reset_non_canonical_warnings: :environment) do
+    choose_not_to_warn = Warning.find_by(name: "Choose Not To Use Archive Warnings")
+    Warning.where(canonical: [false, nil]).each do |warning|
+      if warning.taggings_count.zero?
+        warning.destroy
+      else
+        warning.works.each do |work|
+          work.warnings = choose_not_to_warn if work.warnings.count <= 1
```
redsummernight: This line doesn't work, it should be:

```
work.warnings << choose_not_to_warn
```

We could use a rake spec for this.

lib/tasks/tag_tasks.rake
```diff
@@ -49,6 +49,23 @@ namespace :Tag do
...
+        end
+        # Once we get becomes!  ( rails 4.0.2 ) use it
```
redsummernight: This comment should be removed.

lib/tasks/tag_tasks.rake
```diff
@@ -49,6 +49,23 @@ namespace :Tag do
...
   end
 
+  desc "Reset non canonical warnings"
+  task(reset_non_canonical_warnings: :environment) do
+    choose_not_to_warn = Warning.find_by(name: "Choose Not To Use Archive Warnings")
+    Warning.where(canonical: [false, nil]).each do |warning|
+      if warning.taggings_count.zero?
+        warning.destroy
+      else
+        warning.works.each do |work|
```
redsummernight: Do we want to do the `print(".") && STDOUT.flush` thing for progress?

lib/tasks/tag_tasks.rake
redsummernight: `if work.warnings.count <= 1` if there's already one warning, do we need to add CNTW?

tickinginstant: At this point, the non-canonical warning hasn't been converted to a freeform yet, so if the work has one warning, it has to be the non-canonical warning (meaning it doesn't have any of the canonical warning tags).

lib/tasks/tag_tasks.rake
```diff
@@ -49,6 +49,23 @@ namespace :Tag do
     end
   end
 
+  desc "Reset non canonical warnings"
+  task(reset_non_canonical_warnings: :environment) do
+    choose_not_to_warn = Warning.find_by(name: "Choose Not To Use Archive Warnings")
+    Warning.where(canonical: [false, nil]).each do |warning|
+      if warning.taggings_count.zero?
```
tickinginstant: The tag deletion task also checks for child tags and inclusion in a tag set:
https://github.com/otwcode/otwarchive/blob/c8fe1392b0493dfd5730be590d3b60715414791e/lib/tasks/tag_tasks.rake#L40-L45
Do those need to be checked here, as well? (Or, to avoid code duplication, would it be worth getting rid of the call to `warning.destroy` here and just calling the tag deletion task afterwards?)

lib/tasks/tag_tasks.rake
redsummernight: Gotcha, thanks!

lib/tasks/tag_tasks.rake
zz9pzza: I think calling the tag clean up afterwards makes sense.

spec/lib/tasks/tag_tasks.rake_spec.rb
```diff
@@ -1,5 +1,49 @@
 require "spec_helper"
 
+describe "rake Tag:convert_non_canonical_warnings_to_freeforms" do
```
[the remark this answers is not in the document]
redsummernight: I'll add an exception for rake specs, where we need to use the task name in `describe`.

sarken:
We also need to do this for non-canonical ratings and categories. Do we want to expand this or put them in a separate issue? Or two separate issues, since Rating needs the "Not Rated" default added in the cleanup, but Categories can just be left blank.

redsummernight:
Let's have one new issue for Categories and Ratings.

PR #2917, opened 2017-05-12
AO3-4736 Rails 4.0.13 Upgrade

davidstump:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-4736

## Purpose

This PR upgrades the AO3 application to run on Rails 4.0.13. This encompasses many changes including scopes, finders, observers and many other adjustments. 

## Testing

Due to the wide scope of these changes, a full run-through of site features is required to validate this upgrade. 

## References

No explicit external references

## Credit

David Stump - Littlelines

he/him


Gemfile
```diff
@@ -54,23 +58,23 @@ gem 'nokogumbo', '1.4.9'
...
 
 gem 'cocaine'
-gem 'paperclip'
+gem 'paperclip', '4.3.6'
 
 # for looking up image dimensions quickly
 gem 'fastimage'
 
 # Gems for authentication
 gem 'devise', '~> 3.0'   # Lock on version 3 until we update to Rails 4
```
zz9pzza: Given the comment should we not lock this to version 3 now ?

app/controllers/application_controller.rb
```diff
@@ -300,6 +300,32 @@ def set_media
...
+
+  def load_locales
+    @loaded_locales ||= Locale.order(:iso)
+  end
+
+  # Sets the locale
+  def set_preferred_locale
+    # Loading the current locale
+    if session[:locale] && @loaded_locales.detect { |loc| loc.iso == session[:locale]}
+      set_locale session[:locale].to_sym
```
davidstump: That part came from my upstream merge - I don't believe it has anything to do with the actual 4.0 upgrade. Interesting that it came through as an addition? Could this have been a deletion upstream?

app/mailers/user_mailer.rb
```diff
@@ -372,8 +372,8 @@ def admin_deleted_work_notification(user, work)
 
   # Sends email to authors when a creation is hidden by an Admin
   def admin_hidden_work_notification(creation_id, user_id)
-    @user = User.find_by_id(user_id)
-    @work = Work.find_by_id(creation_id)
+    @user = User.find_by(id: user_id)
```
davidstump: `find_by_id` was used previously so I kept the same concept (return nil vs raise exception). If `@user` is required later, either would be just fine.

app/views/user_invite_requests/index.html.erb
```diff
@@ -9,7 +9,7 @@
   <!--/subnav-->
 
   <!--main content-->
-  <%= form_tag url_for(:controller => 'user_invite_requests', :action => 'update') do %>
+  <%= form_tag url_for(:controller => 'user_invite_requests', :action => 'update'), method: :put do %>
```
davidstump: I believe the error raise mentioned `put`, however either should work.

app/views/works/edit_multiple.html.erb
```diff
@@ -13,7 +13,7 @@
 
 <%= render "works/work_abbreviated_list", :works => @works %>
 
-<%= form_for :work, :url => update_multiple_user_works_path(@user), :html => { :method => :put, :class => "post" } do |form| %>
+<%= form_for :work, :url => update_multiple_user_works_path(@user), method: :put, :html => { :class => "post" } do |form| %>
```
davidstump: Routes generate both and `put` was used previously. Can certainly use either.

db/structure.sql
```diff
@@ -1985,18 +2181,12 @@ INSERT INTO schema_migrations (version) VALUES ('20151018165632');
...
 INSERT INTO schema_migrations (version) VALUES ('20160331005706');
 
-INSERT INTO schema_migrations (version) VALUES ('201604030319571');
-
-INSERT INTO schema_migrations (version) VALUES ('20160416163754');
-
 INSERT INTO schema_migrations (version) VALUES ('20160706031054');
 
-INSERT INTO schema_migrations (version) VALUES ('20160724234958');
+INSERT INTO schema_migrations (version) VALUES ('20170322092920');
```
davidstump: Ack. Should not have been pushed. Apologies.

features/step_definitions/user_steps.rb
```diff
@@ -252,13 +252,16 @@ def get_series_name(age, classname, name)
...
+  user = User.find_by(login: login)
   assert user.active?
 end
 
 Then /^I should see the current user's preferences in the console$/ do
   puts User.current_user.preference.inspect
 end
 
+When(/^I pry$/) do
+  binding.pry
```
davidstump: Ah awesome - can remove.

app/views/works/edit_multiple.html.erb
zz9pzza: It's a question David, I saw patch for the first time and read that it was the prefered way for updates however I don't know what is the right thing to do and this view is an edit.

app/mailers/user_mailer.rb
zz9pzza: Ah sorry nil, v exception ok :)

config/environments/test.rb
```diff
@@ -8,7 +8,7 @@
   config.cache_classes = true
 
   # Log error messages when you accidentally call methods on nil.
-  config.whiny_nils = true
+  # config.whiny_nils = true
```
davidstump: `whiny_nils` is deprecated as of 4.0 I believe.

app/views/works/edit_multiple.html.erb
davidstump: You are correct `patch` is the preferred method - I'll update this.

app/views/user_invite_requests/index.html.erb
davidstump: I'll update this to `patch` as well

Gemfile
davidstump: Sorry - didn't see that comment. I can bump `devise` is that is the preference.

config/environments/test.rb
zz9pzza: You are right, I thought I saw it being set in config/environments/development.rb  but I was wrong sorry :)

app/controllers/application_controller.rb
zz9pzza: I can see it on a really old branch of mine. Why it came back I don't know.

Gemfile
zz9pzza: "Devise 4.0 works with Rails 4.1 onwards. You can add it to your Gemfile with:..."

So looks like not yet :) sorry

davidstump: Oh no worries - I noticed 3.5 or so works with 4.0 so I bumped it as far as was compatible on its own branch. I'll let it run and see how the tests go. If nothing major breaks, we can do a minor bump with 4.0 if that works.

app/controllers/application_controller.rb
ariana-paris: That sounds likely as that code was removed in commit a42de43baf7337ba71b47b72516b286f39bf48e6 (PR 2834) a couple of weeks ago.

davidstump: Perfect, thanks. I'll remove the code to mesh with `master`

app/sweepers/tag_sweeper.rb
```diff
@@ -1,6 +1,6 @@
 class TagSweeper < ActionController::Caching::Sweeper
```
sarken: I noticed tag_sweeper is commented out in several places now that we've moved its contents into the model. Are we still using this file rather than the model in some places?

app/controllers/challenge/gift_exchange_controller.rb
```diff
@@ -50,7 +50,7 @@ def update
       flash[:notice] = ts('Challenge was successfully updated.')
 
       # expire the cache on the signup form
-      expire_fragment(:controller => 'challenge_signups', :action => 'new')
+      ActionController::Base.new.expire_fragment('challenge_signups/new')
```
davidstump: Upgrading to Rails 4 causes an exception: `undefined method 'expire_fragment'` that is solved by adding the `ActionController::Base.new` prefix. A number of relevant threads, but this one is concise: http://stackoverflow.com/questions/33499384/expire-fragment-caching-in-rails-4

features/step_definitions/tag_steps.rb
```diff
@@ -19,42 +19,48 @@
...
 end
 
+Given(/^an adult canonical rating exists with name: "([^"]*)"$/) do |rating|
+  Rating.find_or_create_by(name: rating, canonical: true, adult: true)
+end
+
 Given /^the basic warnings exist$/ do
-  Warning.find_or_create_by_name_and_canonical("No Archive Warnings Apply", true)
-  Warning.find_or_create_by_name_and_canonical("Choose Not To Use Archive Warnings", true)
+  Warning.find_or_create_by_name("No Archive Warnings Apply").update(canonical: true)
```
davidstump: Warning and Category inherit from `Tag` which has a custom defined `find_or_create_by_name` method. Calling `find_by(name:` directly skips this method (and some important logic). I was keeping all Tag children using the custom method. In this case, it might be simply using the find_or_create logic. Perhaps that method should be renamed to avoid confusion with Rails provided finders?

features/step_definitions/tag_steps.rb
```diff
@@ -310,35 +316,40 @@
...
   assert tag.canonical?
 end
 
 Then(/^show me what the tag "([^"]*)" is like$/) do |tagname|
-  tag = Tag.find_by_name(tagname)
+  tag = Tag.find_by(name: tagname)
   puts tag.inspect
 end
+
+Given(/^a media exists with name: "([^"]*)", canonical: true$/) do |media|
```
davidstump: No problem. Done

script/gem_security.sh
```diff
@@ -1,7 +1,8 @@
 #!/bin/sh
 TMP=/tmp/audit.$$
-bundle-audit update       
-bundle-audit check --ignore OSVDB-96425 OSVDB-131677 > $TMP
+bundle-audit update
+# Ignoring issues related to outdated gems required for 4.0 step. To be updated at earliest convinience
```
davidstump: Heh. Also fixed :)

app/decorators/homepage.rb
```diff
@@ -25,7 +25,7 @@ def admin_posts
       @admin_posts = AdminPost.non_translated.for_homepage.all
     else
       @admin_posts = Rails.cache.fetch("home/index/home_admin_posts", expires_in: 20.minutes) do
-        AdminPost.non_translated.for_homepage.all
+        AdminPost.non_translated.for_homepage
```
davidstump: Using `all` results in a deprecation warning. Without all, it returns  an ActiveRecord::Relation object. With all it is returning an array. If the array is the desired output, we can use `to_a` to get the same result without a deprecation warning. Does that work?

app/controllers/works_controller.rb
```diff
@@ -732,19 +732,19 @@ def update_multiple
     @works = Work.joins(pseuds: :user).where('users.id = ?', @user.id).where(id: params[:work_ids]).readonly(false)
     @errors = []
     # to avoid overwriting, we entirely trash any blank fields and also any unchecked checkboxes
-    work_params = params[:work].reject { |_key, value| value.blank? || value == '0' }
+    updated_work_params = work_params.reject { |_key, value| value.blank? || value == '0' }
```
davidstump: Yep.

app/models/tag.rb
```diff
@@ -589,7 +599,7 @@ def self.find_or_create_by_name(new_name)
       elsif tag
         self.find_or_create_by_name(new_name + " - " + self.to_s)
       else
-        self.create(:name => new_name)
+        self.create(name: new_name, type: self.to_s)
```
davidstump: This is due to the bug we found in 4.0 with STI. If you run `Media.create(params)` it would result in a `Fandom` object getting created. If you explicitly add in the type param, it would maintain the proper type. Ideally, this could go away at 4.2 (ish) if the bug is resolved at that point.

app/sweepers/tag_sweeper.rb
davidstump: We are not that I know of. I left the file in place as I wanted to make sure others weighed in on moving away from Sweepers slowly.

app/views/owned_tag_sets/_tag_set_form.html.erb
```diff
@@ -25,7 +25,7 @@
 
   <fieldset>
     <legend><%= ts('Management') %></legend>
-    <%= render "tag_set_form_management", :form => tag_set_form %>
+    <%= render partial: "tag_set_form_management", locals: {form: tag_set_form} %>
```
davidstump: I think it complained about it at some point - but the simpler syntax *should* work fine so I am reverting and pushing. Hopefully some other fix was the cause of the error when I was working on this one.

PR #2919, opened 2017-05-16
AO3-4736 Broken skins.

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4736

## Purpose

Skins are broken on test.

## Testing

Try and change your skin and see if it works....

We might have broken skin paths in memcached. If so it still will not work and we will need to do a memcached flush. But on beta we shouldn't need to do that as it will have never got the broken paths in there.

### DIFF
--- a/app/models/skin.rb
+++ b/app/models/skin.rb
@@ -506,7 +506,7 @@ class Skin < ActiveRecord::Base
     skin = Skin.find_or_create_by(title: "Default", css: "", public: true, role: "user")
     current_version = Skin.get_current_version
     if current_version
-      File.open(Skin.site_skins_dir + current_version + 'preview.png', 'rb') {|preview_file| skin.icon = preview_file}
+      File.open(Skin.site_skins_dir + current_version + '/preview.png', 'rb') {|preview_file| skin.icon = preview_file}
     else
       File.open(Skin.site_skins_dir + 'preview.png', 'rb') {|preview_file| skin.icon = preview_file}
     end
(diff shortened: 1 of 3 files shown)


app/models/skin.rb
```diff
@@ -386,7 +386,8 @@ def get_cached_style(roles_to_include)
   end
 
   def stylesheet_link(file, media)
-    '<link rel="stylesheet" type="text/css" media="' + media + '" href="/' + file + '" />'
+    # we want one and only one / in the url path
+    '<link rel="stylesheet" type="text/css" media="' + media + '" href="/' + file.gsub(/^\/*/,"") + '" />'
```
redsummernight: Is "file" always an absolute path? Can we remove the "/" prefix and the regex?

zz9pzza: I did try that and it didn't work :)

zz9pzza:
app/models/challenge_signup.rb:    Rails.public_path.join("static/challenge_signup_summaries").to_s
app/models/skin.rb:      File.read(Rails.public_path.join("." + filename))
app/models/skin.rb:    Rails.public_path.join(SKIN_PATH).to_s
app/models/skin.rb:    Rails.public_path.join(SITE_SKIN_PATH).to_s
app/models/work.rb:    "#{Rails.public_path}/#{self.download_folder}"
app/models/work_skin.rb:    css = File.read(File.join(Rails.public_path, "/stylesheets/work_skins/basic_formatting.css"))
app/models/work_skin.rb:    File.open(File.join(Rails.public_path, '/images/skins/previews/basic_formatting.png'), 'rb') {|preview_file| skin.icon = preview_file}

PR #2922, opened 2017-05-23
AO3-3223 When saving a coauthor for a work, save for all existing chapters as well

annelu:
## Issue

https://otwarchive.atlassian.net/browse/AO3-3223 

## Purpose

The current behavior is that when a coauthor is saved for a multichapter work, they are only saved as a coauthor for the first chapter. This change makes it so that they are saved as a coauthor for every chapter of the work. However, it does not change it so that each new chapter contains all the work's coauthors; they will only have the user who saved that chapter as an author, and all other coauthors have to be added manually.

## Testing

* Create a multichapter work with one user.
* Add a second user as a coauthor for the entire work.
* As the second user, select "Edit Chapter" for any chapter except for the first--there should be no error banner, and the user should be able to edit that chapter as normal.

## References

https://otwarchive.atlassian.net/browse/AO3-2971 (duplicate)

## Credit

potatoesque, she/her

### DIFF
--- a/features/works/work_create.feature
+++ b/features/works/work_create.feature
@@ -27,7 +27,7 @@ Feature: Create Works
 
   Scenario: Creating a new minimally valid work and posting without preview
     Given I am logged in as "newbie"
-    When I set up the draft "All Hell Breaks Loose" 
+    When I set up the draft "All Hell Breaks Loose"
       And I fill in "content" with "Bad things happen, etc."
       And I press "Post Without Preview"
     Then I should see "Work was successfully posted."
@@ -36,7 +36,7 @@ Feature: Create Works
     Then I should see "All Hell Breaks Loose"
 
   Scenario: Creating a new minimally valid work when you have more than one pseud
-    Given I am logged in as "newbie"      
+    Given I am logged in as "newbie"
       And "newbie" creates the pseud "Pointless Pseud"
     When I set up the draft "All Hell Breaks Loose"
       And I unselect "newbie" from "work_author_attributes_ids_"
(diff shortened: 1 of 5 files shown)


sarken:
The test output is a bit hard to read at the moment due to the deprecation warnings, but one of the failures looks like it might be related to the code change (the other is just a MySQL hiccup): 
```
cucumber features/other_a/orphan_work.feature:143 # Scenario: Orphaning a shared work should not affect chapters created solely by the other creator
```
I've restarted it just in case it's not, though.

I think it might also be good to have a cucumber test for this scenario that checks the bylines on individual chapters -- we've had some problems with chapter bylines not updating in the past.

features/works/work_create.feature
```diff
@@ -292,3 +292,17 @@ Feature: Create Works
       And I press "Post Without Preview"
     Then I should see "Publication date can't be in the future."
     When I jump in our Delorean and return to the present
+
+  Scenario: Adding a coauthor to a work adds the coauthor to all existing chapters.
```
sarken: Thanks for adding this! Just a quick note about formatting cukes: we try to make it so there aren't multiple steps beginning with `Given`, `When`, or `Then` in a row, and only `And` or `But` steps are indented. So something more like:

```
 Scenario: Adding a coauthor to a work adds the coauthor to all existing chapters.
    Given the following activated users exists
    | login             | password   | email                     |
    | author            | password   | author@example.com        |
    | coauthor          | password   | coauthor@example.com      |
    When I am logged in as "author"
      And I post the chaptered work "Chaptered Work"
      And I add the co-author "coauthor" to the work "Chaptered Work"
    Then I should see "author, coauthor" within ".byline"
    When I follow "Next Chapter →"
    Then I should see "Chapter 2"
      And I should see "author, coauthor" within ".byline"
```

Also, I think newer versions of FactoryGirl get rid of support for tables, so for a bit of future-proofing, it might be better to use this step instead:

```
Given the user "author" exists and is activated
  And the user "coauthor" exists and is activated
```

annelu: Cool, thanks for the info!

PR #2931, opened 2017-06-06
AO3-5032 update from hash rocket syntax to JSON style

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5032

## Purpose

Updates to hash rocket syntax using http://effectif.com/ruby/update-your-project-for-ruby-19-hash-syntax

eg ```find . -name \*.rb -exec perl -p -i -e 's/([^:]):(\w+)\s*=>/\1\2:/g' {} \;```

## Testing

Should not need any testing ( but do it after the 4.2 merge and before we do a general regression/test ).

## References

Merge after #2929


sarken:
It seems like this needs to be merged with master as well -- it's failing because it has the wrong version of the work search tests.

PR #2934, opened 2017-06-10
AO3-5035 Upgrade Ruby to 2.2.7

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5035

## Purpose

Upgrades ruby to 2.2.7

## Testing

I would put these in at the end of the rails upgrade, for example this one might go in at rails 4.2

This is really the test side of the work, there is more work on the server side.


zz9pzza:
I am going to close this as I don't think it gains enough for the amount of testing, however I think we should do 2.3 and 2.4

PR #2936, opened 2017-06-11
AO3-5039 Prevent non canonical tags from being favorited

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5039

## Purpose

Makes it so trying to favorite a non canonical tag gives an error message. Also adds tests for the favorite tags model and fixes an issue found when writing those specs.

## Testing

Refer to JIRA issue


app/models/favorite_tag.rb
```diff
@@ -11,16 +11,24 @@ class FavoriteTag < ActiveRecord::Base
...
 
   validate :within_limit, on: :create
-
-  after_save :expire_cached_home_favorite_tags
-  after_destroy :expire_cached_home_favorite_tags
-
   def within_limit
-    if user.favorite_tags(:reload).count >= ArchiveConfig.MAX_FAVORITE_TAGS
-      errors.add(:base, ts('Sorry, you can only save %{maximum} favorite tags.', maximum: ArchiveConfig.MAX_FAVORITE_TAGS))
+    if user && user.favorite_tags(:reload).count >= ArchiveConfig.MAX_FAVORITE_TAGS
```
sarken: Without `user` in here, the new spec for this failed with:
```
  2) FavoriteTag is invalid without a user_id
     Failure/Error: if user.favorite_tags(:reload).count >= ArchiveConfig.MAX_FAVORITE_TAGS
     
     NoMethodError:
       undefined method `favorite_tags' for nil:NilClass
     # ./app/models/favorite_tag.rb:15:in `within_limit'
     # ./spec/models/favorite_tag_spec.rb:25:in `block (2 levels) in <top (required)>'
```

app/models/favorite_tag.rb
```diff
@@ -11,16 +11,24 @@ class FavoriteTag < ActiveRecord::Base
...
+      errors.add(:base,
+                 ts("Sorry, you can only save %{maximum} favorite tags.",
+                    maximum: ArchiveConfig.MAX_FAVORITE_TAGS))
+    end
+  end
+
+  validate :canonical, on: :create
```
zz9pzza: What would happen if someone favorited a tag and it was changed to non canonical ?

sarken: We have code in place that destroys a favorite tag when it's de-canonized. That code shouldn't conflict with this -- it's currently tested here, and all tests are passing: https://github.com/otwcode/otwarchive/blob/master/features/tags_and_wrangling/favorite_tags.feature#L31

sarken:
This had some test failures, possibly due to changes in the tag factory since this was originally written. I've fixed them and this needs to be re-reviewed.

spec/models/favorite_tag_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+    create(:favorite_tag, tag_id: tag.id, user_id: user.id)
+    expect(build(:favorite_tag, tag_id: tag.id, user_id: user.id).save).to be_falsey
+  end
+
+  it "is invalid without a user_id" do
+    expect(no_user_favorite.save).to be_falsey
+  end
+
+  it "is invalid with a non-canonical tag" do
+    expect(noncanonical_favorite.save).to be_falsey
```
redsummernight: Some of these `save` calls can be replaced with `valid?`. We can test validations without saving to the database which makes it faster.

PR #2937, opened 2017-06-12
AO3-5036 Update to ruby 2.3

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5036

## Purpose

Update the ruby language to 2.3.

## Testing

I would check that work searching works as intended


app/controllers/works_controller.rb
```diff
@@ -18,61 +18,66 @@ class WorksController < ApplicationController
...
 
-  before_filter :clean_work_search_params, only: [:search, :index, :collected]
-
   cache_sweeper :collection_sweeper
   cache_sweeper :feed_sweeper
 
   # we want to extract the countable params from work_search and move them into their fields
   def clean_work_search_params
-    if params[:work_search].present? && params[:work_search][:query].present?
+    #https://aaronlasseigne.com/2014/07/20/know-ruby-clone-and-dup/
```
sarken: Drive-by comment! This page is informative, but I'm not sure which part(s) of it is relevant to the code. A more helpful comment might be something like "We use dup instead of clone here because we don't want to copy the frozen state of the object" or "Using dup here means the original object also gets modified when we do the gt/lt swap below" or whatever.

app/controllers/works_controller.rb
```diff
@@ -18,61 +18,66 @@ class WorksController < ApplicationController
...
+        next unless clean_params[:query].gsub!(/#{term}s?\s*(?:\_?count)?\s*:?\s*((?:<|>|=|:)\s*\d+(?:\-\d+)?)/i, '')
         # pluralize, add _count, convert to symbol
         term = term.pluralize unless term == 'word'
         term += '_count' unless term == 'hits'
         term = term.to_sym
 
         value = Regexp.last_match(1).gsub(/^(\:|\=)/, '') # get rid of : and =
         # don't overwrite if submitting from advanced search?
-        params[:work_search][term] = value unless params[:work_search][term].present?
+        clean_params[term] = value unless clean_params[term].present?
```
sarken: Perhaps a silly question! This line uses `clean_params[term]`, but on line 29, it's `clean_params[:term]`. Is that intentional? And if it's intentional, do we still need line 29? I can't see where `clean_params[:term]` is being used.

PR #2941, opened 2017-06-16
AO3-5043 Remove admin notifications feature

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5043

## Purpose

This removes the admin notifications feature and the corresponding "turn off admin emails" preference. It also removes the alphabet helper, which was used by the admin notifications feature and  views/people/show.html.erb. views/people/show.html.erb is being removed here because it's not in use and was otherwise preventing the removal of the alphabet helper.

## Testing

Refer to JIRA.


app/views/people/show.html.erb
```diff
@@ -1,23 +0,0 @@
-<!--SEARCHBROWSE Descriptive page name, messages and instructions-->
```
zz9pzza: Should this file be removed ( it doesn't look relevant to the pull ? )

sarken: From the "purpose" bit:
> It also removes the alphabet helper, which was used by the admin notifications feature and views/people/show.html.erb. views/people/show.html.erb is being removed here because it's not in use and was otherwise preventing the removal of the alphabet helper.

zz9pzza: Sorry I was thinking about index rather than show.

sarken: No worries!

PR #2950, opened 2017-06-25
AO3-5023 Fix logged in user claiming imported works

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5023

Comment: https://otwarchive.atlassian.net/browse/AO3-5023?focusedCommentId=351011&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-351011

## Purpose

Fixes a routing error that was preventing users from claiming imported works. There's a little bit of test tidying, too, because I was trying to track down several OD-related bugs, and I needed to understand what we were testing in some of these. :P Alas, this was the only one I could fix. 

## Testing

You'll need a user with two email addresses: one they use for their AO3 account and one they don't. An archivist will need to import a work and list the non-AO3 email addresses in the external author information. While logged in, the user should follow the claim link in the email and choose "Add these works to my currently-logged-in account." The works should be added to their account.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -47,7 +47,7 @@ Otwarchive::Application.routes.draw do
 
   get 'signup/:invitation_token' => 'users#new', as: 'signup'
   get 'claim/:invitation_token' => 'external_authors#claim', as: 'claim'
-  get 'complete_claim/:invitation_token' => 'external_authors#complete_claim', as: 'complete_claim'
+  post 'complete_claim/:invitation_token' => 'external_authors#complete_claim', as: 'complete_claim'
 
   #### TAGS ####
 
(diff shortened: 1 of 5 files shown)


config/routes.rb
```diff
@@ -47,7 +47,7 @@
 
   get 'signup/:invitation_token' => 'users#new', as: 'signup'
   get 'claim/:invitation_token' => 'external_authors#claim', as: 'claim'
-  get 'complete_claim/:invitation_token' => 'external_authors#complete_claim', as: 'complete_claim'
+  post 'complete_claim/:invitation_token' => 'external_authors#complete_claim', as: 'complete_claim'
```
sarken: The problem was:
```
No route matches [POST] "/complete_claim/4fbc1797f6db942c2d4ebbd8958c39fec8a1c47a" (ActionController::RoutingError)
```
Refer to "Route Matching" on https://www.sitepoint.com/get-your-app-ready-for-rails-4/ -- basically, you need to specify the correct verb now

PR #2954, opened 2017-07-03
AO3-3536 Show validation error message when parent Url unreachable

maryalice0:
## Issue

https://otwarchive.atlassian.net/browse/AO3-3536

## Purpose

Displays the first validation error message for a newly-associated external work when it cannot be saved.

## Testing

Edit an Archive work to add an associated external work with an unreachable URL.  Should see the error message "Parent work Url could not be reached. If the URL is correct and the site is currently down, please try again later." instead of "Parent work info would not save."

### DIFF
--- a/features/works/work_related.feature
+++ b/features/works/work_related.feature
@@ -284,13 +284,14 @@ Scenario: Listing external works as inspirations
   When I view my related works
   Then I should see "From N/A to English"
   # inactive URL should give a helpful message (AO3-1783)
+  # unreachable URL should give a more helpful message (A03-3536)
   When I edit the work "Followup"
     And I check "parent-options-show"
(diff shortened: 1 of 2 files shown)


features/works/work_related.feature
```diff
@@ -284,13 +284,14 @@ Scenario: Listing external works as inspirations
   When I view my related works
   Then I should see "From N/A to English"
   # inactive URL should give a helpful message (AO3-1783)
+  # unreachable URL should give a more helpful message (A03-3536)
```
ariana-paris: I don't think we need to keep either of these messages anymore - git blame will tell us which tickets this test was designed to verify

redsummernight:
The pull request title should be "AO3-3536" instead of "A03-3536".

PR #2956, opened 2017-07-06
AO3-5058 Add errors when importing for other with no author email or name

ariana-paris:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5058

## Purpose

This PR returns error messages when importing for others as an archivist without specifying an author name or email. 

It also includes a refactor of the StoryParser to reduce the 100+ messages Rubocop gives whenever I edit it which will make future changes easier (I have some elaborate changes coming - those will be easier to review if this goes in first).

For ease of review: the commit containing the actual changes is 0130c5d20139f6e7b8b0ddf6b7bfd80733f881e7

## Testing

See ticket


app/models/story_parser.rb
```diff
@@ -213,861 +210,848 @@ def parse_chapters_into_story(location, chapter_contents, options = {})
...
-    if external_author_email.present?
-      return parse_author_common(external_author_email,external_author_name)
+  # Everything below here is protected and should not be touched by outside
+  # code -- please use the above functions to parse external works.
 
-    else
+  protected
+
+  # tries to create an external author for a given url
+  def parse_author(location, external_author_name, external_author_email)
```
ariana-paris: This is one of the methods I changed to add an error message (it contains the error that appears when you import for others without an author or email, and from a site where email addresses can't be detected)

app/models/story_parser.rb
```diff
@@ -213,861 +210,848 @@ def parse_chapters_into_story(location, chapter_contents, options = {})
...
-      return parse_author_common(email, name)
     end
+    email = authornode[:href].gsub("mailto:", '')
+    name = authornode.inner_text
+
+    parse_author_common(email, name)
+  end
 
-    def parse_author_common(email, name)
+  def parse_author_common(email, name)
```
ariana-paris: This is the other method whose functionality I changed for this issue.

app/models/story_parser.rb
```diff
@@ -246,6 +225,19 @@ def parse_author(location,external_author_name,external_author_email)
...
   protected
 
+  # tries to create an external author for a given url
+  def parse_author(location, external_author_name, external_author_email)
+    if location.present? && external_author_name.blank? && external_author_email.blank?
+      source = get_source_if_known(KNOWN_AUTHOR_PARSERS, location)
+      if source.nil?
+        raise Error, "No external author name or email specified, and unable to generate email based on source location"
+      else
+        eval("parse_author_from_#{source.downcase}(location)")
```
ariana-paris: The evals will need to hang around a little longer until I extract and refactor the code for specific external sites, at which point I might be able to remove them.

app/models/story_parser.rb
```diff
@@ -213,861 +210,848 @@ def parse_chapters_into_story(location, chapter_contents, options = {})
...
-          # probably seconds since the epoch
-          date = Time.at($1.to_i)
-        end
-        date ||= Date.parse(date_string)
-        return '' if date > Date.today
-        return date
-      rescue ArgumentError, TypeError
-        return ''
+  def convert_revised_at(date_string)
+    begin
```
ariana-paris: Removing the `begin` seemed to break functionality so I left it alone for now.

app/views/works/new_import.html.erb
```diff
@@ -59,7 +59,7 @@
         </ul>
       </dd>
       <% if current_user.archivist %>
-        <dt><%= label_tag ts("Archiving for others") %></dt>
+        <dt><%= label_tag "importing_for_other", ts("Archiving for others") %></dt>
```
sarken: This doesn't seem to correspond to the id of a [labelable form element](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#Form_labelable), so I don't think it'll actually accomplish anything, but it won't hurt anything either, since right now the for attribute is just "Archiving_for_others" (which also does not correspond to anything).

app/models/story_parser.rb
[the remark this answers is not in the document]
elzj: You should be able to use send instead of eval in this case:

    send("parse_author_from_#{source.downcase}", location)

app/views/works/new_import.html.erb
ariana-paris: I don't remember now why I thought that was a good idea. I suspect RubyMine or the Safari dev tools told me that was missing the attribute - I probably saw the adjacent `importing_for_others` checkbox (which already has a label) and then typoed it to boot. I'll remove it because what's the point?

app/models/story_parser.rb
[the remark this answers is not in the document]
ariana-paris: I was also trying to minimise changes I make to facilitate reviewing, but since reviewing has happened, I'm more than happy to make more changes now!

PR #2957, opened 2017-07-08
AO3-4924 Chapter Controller Specs

katieby:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4924

## Purpose

Add specs for chapters_controller.

## Testing

2. In the "Search" field near the "Source Files on Master" heading, enter the name of the controller: controllers/chapters_controller.rb
3. Note the number in the "Coverage" column of the table

## Credit

cresenne, she/her


spec/controllers/chapters_controller_spec.rb
```diff
@@ -4,74 +4,554 @@
...
+    end
+
+    it "assigns instance variables correctly" do
+      second_chapter = create(:chapter, work: work, authors: work.authors, position: 2, posted: true)
+      third_chapter = create(:chapter, work: work, authors: work.authors, position: 3, posted: true)
+      comment = create(:comment, commentable_type: "Chapter", commentable_id: second_chapter.id)
+      kudo = create(:kudo, commentable_id: work.id, pseud: create(:pseud))
+      tag = create(:fandom)
+      expect_any_instance_of(Work).to receive(:tag_groups).and_return("Fandom" => [tag])
+      expect_any_instance_of(ChaptersController).to receive(:get_page_title).with("The 1 Fandom", user.pseuds.first.name, "My title is long enough - Chapter 2").and_return("page title")
```
sarken: Oh, I think I see why this might be failing -- this expects "The 1 Fandom," but other specs that run before this probably created fandom tags, which means the number is off:
```
expected: ("The 1 Fandom", "obonss2d142", "My title is long enough - Chapter 2")
              got: ("The 4 Fandom", "obonss2d142", "My title is long enough - Chapter 2")
```
 You might have to do something like `tag.name` here instead.

spec/controllers/chapters_controller_spec.rb
```diff
@@ -4,74 +4,554 @@
...
+        context "when the chapter and work are valid" do
+          it "posts the chapter" do
+            post :create, work_id: work.id, chapter: @chapter_attributes, post_without_preview_button: true
+            expect(assigns[:chapter].posted).to be true
+          end
+
+          it "posts the work if the work was not posted before" do
+            @unposted_work = create(:work, authors: [user.pseuds.first])
+            post :create, work_id: @unposted_work.id, chapter: @chapter_attributes, post_without_preview_button: true
+            expect(assigns[:work].posted).to be true
```
katieby: That's what the next test does.

sarken: Oh, okay! I was expecting the success message to be different here ("Work successfully posted, it may take a while to appear in listings," etc), since this test uses an unposted work and the next test uses a posted work. However, I see now that's not the current behavior.

PR #2958, opened 2017-07-13
AO3-5033 Rails 5.0 upgrade

WendyBeth:
# Pull Request Checklist

## Issue

## Purpose

What does this PR do?

Upgrades application from 4.2 to 5.0.

There are a handful of deprecations not addressed in this PR:

1. alias_method_chain - there is a file, lib/plugins/livevalidation/form_helpers.rb which will need to be restructured in order to remove alias_method_chain and replace it with Module#prepend. I've held off on this as I believe the test failures it causes in 5.1 will be useful in restructuring (instead of attempting to do so blind).

2. The dynamic :controller/:action segments in routes will not be outdated until 5.2.

Some others, left alone simply because I believe the test failures in 5.1 will make it easier to fix them.

## Testing

Due to the extensive nature of these changes, a full walkthrough in the staging environment is likely necessary to ensure everything is working as expected.

## References

## Credit

Wendy Randquist - Littlelines

she/her


app/models/creatorship.rb
```diff
@@ -4,6 +4,19 @@ class Creatorship < ActiveRecord::Base
 
   before_destroy :expire_caches
 
+  validate :unique_index
```
sarken: This code looks good, but we were wondering why it's necessary -- is there something specific that was causing creatorships to be invalid due to a non-unique index?

app/models/comment.rb
```diff
@@ -55,6 +56,232 @@ def akismet_attributes
...
+          end
+          if notify_user_by_inbox?(user)
+            add_feedback_to_inbox(user)
+          end
+        end
+      end
+    end
+  end
+
+  def notify_user_of_own_comments?(user)
```
sarken: It looks like this was a protected method when we were using an observer. Is it possible to have this (and the other formerly-protected methods) be protected here in the model? Or does moving them to the model mean we shouldn't worry about protecting them?

(Here's the full list of methods that were protected in the observer for quick reference: notify_parent_comment_owner(comment), have_different_owner?(comment, parent_comment), not_user_commenter?(parent_comment), content_too_different?(new_content, old_content), add_feedback_to_inbox(user, comment), update_feedback_in_inbox(user, comment), notify_user_by_email?(user), notify_user_by_inbox?(user), notify_user_of_own_comments?(user))

app/controllers/comments_controller.rb
```diff
@@ -511,8 +511,7 @@ def redirect_to_comment(comment, options = {})
     end
   end
 
-  def redirect_to_all_comments(commentable, options = {})
-    default_options = {anchor: "comments"}
+  def redirect_to_all_comments(commentable, options = {}) default_options = {anchor: "comments"}
```
WendyBeth: No. Sometimes my editor has its own opinion about things. >_> Will fix.

app/controllers/works_controller.rb
```diff
@@ -524,14 +524,16 @@ def import
       render(:new_import) && return
     end
 
+    importing_for_others = params[:importing_for_others] != "false" && params[:importing_for_others]
```
WendyBeth: I replied in the commit to @ariana-paris. Hopefully the logic is sound.

app/mailers/user_mailer.rb
```diff
@@ -391,7 +391,8 @@ def delete_signup_notification(user, challenge_signup)
     I18n.with_locale(Locale.find(@user.preference.preferred_locale).iso) do
       mail(
         to: user.email,
-        subject: "[#{ArchiveConfig.APP_SHORT_NAME}] Your sign-up for #{@signup.collection.title} has been deleted"
+        subject: "[#{ArchiveConfig.APP_SHORT_NAME}] Your sign-up for #{@signup.collection.title} has been deleted",
+        body: ""
```
WendyBeth: Yes - it needs a body. It looks like it's sent in the `before_destroy` callback in ChallengeSignup so I'm not sure what you mean about it not being used. ?

app/models/creatorship.rb
WendyBeth: Took me a moment to remember: I used it to make debugging easier when I was trying to figure out why creatorships were being created twice in some callbacks. This had to do with the order of the callbacks being defined that referenced associated models through join models (hence why it's in the same commit that moves the `import Taggable` line). Instead of an error message indicating that something had failed to save, I wanted an error message that told me explicitly why, and writing this is how I found it.

Since it's valid and indicates what's required explicitly I never removed it.

app/models/comment.rb
WendyBeth: Whoops! I'll get these protected again. There's nothing that should prevent them from being so.

lib/tasks/resque.rake
```diff
@@ -5,7 +5,7 @@ require 'resque/scheduler/tasks'
 namespace :resque do
   task :setup do
     require 'resque'
-    require 'resque/scheduler'
+    require 'resque/scheduler/tasks'
```
zz9pzza: I can't at this instant, I will be more free next week.

zz9pzza: However it's an easy fix, so I wouldn't treat it as blocking.

app/mailers/user_mailer.rb
sarken: I just mean I haven't gotten one of these emails when deleting challenge sign-ups in recent memory -- the last time I did it on production was April, and because this email didn't sound familiar, I tested it manually on staging shortly after this PR was submitted. It's possible Google's spam filter is eating my emails or it's been broken for a long time and we failed to notice.

PR #2962, opened 2017-07-17
AO3-5023: Restore missing order clause to wrangling query

elzj:
AO3-5023: Restore missing order clause to wrangling query

Just got dropped in the code update


sarken:
Will merge when Travis is done, thank you!

sarken:
I kicked Travis because it failed on totally unreleased orphaning tests (which we do see rather frequent failures on), but I'm merging this now anyway

PR #2963, opened 2017-07-19
AO3-5069 Prevent new works with series from 500ing when there are validation errors

katieby:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5069

## Purpose

Prevent 500 error when posting/previewing a new work in a series when other validation errors exist.
## Testing

1. Log in
2. Post > New Work
3. Fill in all required information EXCEPT a warning
4. Check "This work is part of a series"
5. Choose an existing series or fill in a name for a new one
6. Preview or Post Without Preview

You should see the edit form with a flash warning, the information you filled out in steps 4 and 5 should be persisted, and there should be no link to remove the work from the series. When you get rid of the warning the work should post as normal and be in the series.

Adding/removing existing works from series should work as before.

## Credit

cresenne, she/her

### DIFF
--- a/features/works/work_create.feature
+++ b/features/works/work_create.feature
@@ -199,8 +199,6 @@ Feature: Create Works
 
   Scenario: Creating a new work in a new series with some invalid things should return to the new work page with an error message and series information still filled in
     Given basic tags
-      And the following activated users exist
-        | login          | password    | email                   |
       And I am logged in as "thorough" with password "something"
     When I set up the draft "Bad Draft"
       And I fill in "Fandoms" with "Invalid12./"
@@ -217,9 +215,7 @@ Feature: Create Works
       And I should not see "Remove Work From Series"
 
   Scenario: Creating a new work in an existing series with some invalid things should return to the new work page with an error message and series information still filled in
-    Given basic tags
-      And the following activated users exist
-        | login          | password    | email                   |
+    Given basic tags              |
       And I am logged in as "thorough" with password "something"
       And I post the work "Work one" as part of a series "My existing series"
     When I set up the draft "Bad Draft"
--- a/features/works/work_create.feature
+++ b/features/works/work_create.feature
@@ -215,7 +215,7 @@ Feature: Create Works
       And I should not see "Remove Work From Series"
 
   Scenario: Creating a new work in an existing series with some invalid things should return to the new work page with an error message and series information still filled in
-    Given basic tags              |
+    Given basic tags
       And I am logged in as "thorough" with password "something"
       And I post the work "Work one" as part of a series "My existing series"
     When I set up the draft "Bad Draft"
--- a/features/works/work_create.feature
+++ b/features/works/work_create.feature
@@ -197,27 +197,47 @@ Feature: Create Works
       And I should see "Chapter"
       And I should see "1/?"
 
-  Scenario: Creating a new work in a series with some invalid things should not 500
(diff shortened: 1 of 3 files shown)


features/works/work_create.feature
```diff
@@ -197,6 +197,29 @@ Feature: Create Works
...
+        | login          | password    | email                   |
+        | coauthor       | something   | coauthor@example.org |
+        | badcoauthor    | something   | badcoauthor@example.org |
+      And I am logged in as "thorough" with password "something"
+      And user "badcoauthor" is banned
+    When I set up the draft "Bad Draft"
+      And I fill in "Fandoms" with "Invalid12./"
+      And I fill in "Work Title" with "/"
+      And I fill in "content" with "T"
+      And I check "chapters-options-show"
```
sarken: In the feature files, we prefer to use the user-facing label text over the field id (in step definitions, the field id is fine) whenever possible.

features/works/work_create.feature
```diff
@@ -197,6 +197,29 @@ Feature: Create Works
...
+      And I fill in "work_wip_length" with "text"
+      And I fill in "work_collection_names" with "collection1, collection2"
+      And I check "series-options-show"
+      And I fill in "work_series_attributes_title" with "My new series"
+      And I press "Preview"
+    Then I should see "Sorry! We couldn't save this work because:"
+      And I should see a collection not found message for "collection1"
+      And I should see "My new series"
```
sarken: Could you check that this text is in the desired field, i.e `And the field labeled "Or create and use a new one" should contain "My new series"`? 

Also, it would be great if you could add a scenario to test that the "Choose one of your existing series" field remembers the correct value when the form errors, too.

features/works/work_create.feature
```diff
@@ -197,6 +197,29 @@ Feature: Create Works
       And I should see "Chapter"
       And I should see "1/?"
 
+  Scenario: Creating a new work in a series with some invalid things should not 500
```
sarken: It's generally better to describe the correct behavior than the bug, so maybe something more like "should return you to the edit form, with the relevant error messages and the series information still filled in" instead of "should not 500"?

features/works/work_create.feature
```diff
@@ -197,6 +197,45 @@ Feature: Create Works
       And I should see "Chapter"
       And I should see "1/?"
 
+  Scenario: Creating a new work in a new series with some invalid things should return to the new work page with an error message and series information still filled in
+    Given basic tags
+      And the following activated users exist
```
sarken: You should be able to remove this line and the line after it now that coauthor and badcoauthor have been removed.

features/works/work_create.feature
```diff
@@ -197,6 +197,41 @@ Feature: Create Works
...
+      And I check "This work is part of a series"
+      And I fill in "Or create and use a new one:" with "My new series"
+      And I press "Preview"
+    Then I should see "Sorry! We couldn't save this work because:"
+      And I should see a collection not found message for "collection1"
+      And the field labeled "Or create and use a new one:" should contain "My new series"
+      And I should not see "Remove Work From Series"
+
+  Scenario: Creating a new work in an existing series with some invalid things should return to the new work page with an error message and series information still filled in
+    Given basic tags              |
```
sarken: Travis contradicts me /o\ There's a stray `|` at the end of this line that is causing an "undefined step" error.

PR #2964, opened 2017-07-22
AO3-5072 Fix redirect page

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5072

## Purpose

The form in http://archiveofourown.org/redirect did not work.

## Testing

Go to http://archiveofourown.org/redirect
Paste in a URL that has been imported from another archive using the import work function
Press submit
If a redirected work is found, it should take you to that work with a message to update the original link if possible.
If one is not found, it should stay on/return to the redirect search page but present you with this message "We could not find a work imported from that url in the Archive of Our Own, sorry! Try another url?"
However, entering both imported & non imported urls currently takes me to a blank page in Safari.
# The redirect functionality does seem to be working e.g.: http://archiveofourown.org/redirect?original_url=http://triofic.com/viewstory.php?sid=1

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -537,7 +537,6 @@ Otwarchive::Application.routes.draw do
   resource :redirect, controller: "redirect", only: [:show] do
     member do
       get :do_redirect
-      get :show
     end
   end
 
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -537,6 +537,7 @@ Otwarchive::Application.routes.draw do
   resource :redirect, controller: "redirect", only: [:show] do
     member do
       get :do_redirect
+      post :show
     end
   end
 
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -537,7 +537,7 @@ Otwarchive::Application.routes.draw do
   resource :redirect, controller: "redirect", only: [:show] do
     member do
       get :do_redirect
-      post :show
+      get :show
     end
   end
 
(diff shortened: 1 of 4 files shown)


config/routes.rb
```diff
@@ -537,6 +537,7 @@
   resource :redirect, controller: "redirect", only: [:show] do
     member do
       get :do_redirect
+      get :show
```
elzj: You don't actually need this line, since the route is already defined in the "only: [:show]" bit above.

features/importing/work_import.feature
```diff
@@ -220,3 +220,11 @@ Feature: Import Works
     When I import "http://www.intimations.org/fanfic/idol/Huddling.html"
     Then I should see "Preview"
       And I should see "English"
+
+  Scenario: Import and check that redirects work
+    When I import "http://www.scarvesandcoffee.net/viewstory.php?sid=9570"
+      And I press "Post"
+      And I go to the redirect page
+      And I fill in "original_url" with "http://www.scarvesandcoffee.net/viewstory.php?sid=9570"
```
sarken: Could you use the label text here instead, i.e. `And I fill in "Original URL of work" with "http://www.scarvesandcoffee.net/viewstory.php?sid=9570"`? It's nice to have feature tests correspond with what the user actually sees. (Also, a better scenario description might be "Searching for an imported work by URL will redirect you to the work," so if the test fails, we know to start by checking the search function rather than the redirect code.)

sarken:
:+1: Thanks for the test changes!

PR #2968, opened 2017-07-23
AO3-5043 Fix 500 when attempting to access removed admin notifications URL

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5043

## Purpose

Going to http://test.ao3.org/admin/users/notify while logged in as an admin was causing a 500 error because it thought "notify" was a username. Now it will redirect to http://test.ao3.org/admin/users

## Testing

Attempt to access http://test.ao3.org/admin/users/notify while logged in as an admin. Also please check that the Details page for a user shows the User History


app/controllers/admin/admin_users_controller.rb
```diff
@@ -37,7 +37,7 @@ def show
     unless @user
       redirect_to action: "index", query: params[:query], role: params[:role]
     end
-    @log_items = @user.log_items.sort_by(&:created_at).reverse
+    @log_items = @user.log_items.sort_by(&:created_at).reverse if @user
```
redsummernight: Should we add a return in the unless block instead?

PR #2972, opened 2017-07-24
AO3-5081 Fix intermittent failure in works/work_create

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5081

## Purpose

Adds a 1-second delay to the step for editing an existing work to add a co-author. Without this delay, the cache didn't have time to update, so we'd get intermittent test failures in a works/work_create (and probably a few other places I haven't noticed yet). 

The step is only used 5 times in our features, so this won't amount to a significant delay.

## Testing

No manual testing needed, but we should keep an eye on Codeship the next several times it runs to see if this helped.

### DIFF
--- a/features/other_a/orphan_work.feature
+++ b/features/other_a/orphan_work.feature
@@ -144,7 +144,6 @@ Feature: Orphan work
 
     Given I am logged in as "keeper"
       And I post the work "Half-Orphaned"
-      And I wait 1 second
       And I add the co-author "orphaneer" to the work "Half-Orphaned"
       And I post a chapter for the work "Half-Orphaned"
     # Verify that the authorship has been set up properly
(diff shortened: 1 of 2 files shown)


features/other_a/orphan_work.feature
```diff
@@ -144,7 +144,6 @@ Feature: Orphan work
 
     Given I am logged in as "keeper"
       And I post the work "Half-Orphaned"
-      And I wait 1 second
```
sarken: Unnecessary now that the 1-second wait is baked in to the `I add the co-author "name" to the work "Title"` step, whee.

zz9pzza: Do you want to remove it and I can merge.

sarken: I did remove it -- this comment was just to explain why I did.

zz9pzza: Sorry I will press the merge button

PR #2974, opened 2017-07-25
AO3 5008: Tweak subscription validation

elzj:
AO3 5008: Tweak subscription validation to prevent subscriptions to nonsubscribable models.

First, explicitly validate the types of objects you're allowed to subscribe to. Then add a condition to the subscribable validation so that you don't get a 500 error if subscribable_type is not a real model name, allowing failures to have a consistent format.

### DIFF
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -1,7 +1,7 @@
 class Subscription < ActiveRecord::Base
   include ActiveModel::ForbiddenAttributesProtection
 
-  VALID_SUBSCRIBABLES = %w(Work User Series)
+  VALID_SUBSCRIBABLES = %w(Work User Series).freeze
 
   belongs_to :user
   belongs_to :subscribable, polymorphic: true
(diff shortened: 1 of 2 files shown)


app/models/subscription.rb
```diff
@@ -1,10 +1,18 @@
 class Subscription < ActiveRecord::Base
   include ActiveModel::ForbiddenAttributesProtection
 
+  VALID_SUBSCRIBABLES = %w(Work User Series)
```
zz9pzza: I think this is fair

app/models/subscription.rb
```diff
@@ -1,10 +1,18 @@
...
-  
-  validates_presence_of :user, :subscribable
+
+  validates_presence_of :user
+
+  validates :subscribable_type, inclusion: { in: VALID_SUBSCRIBABLES }
+  # Without the condition, you get a 500 error instead of a validation error
+  # if there's an invalid subscribable type
+  validates :subscribable, presence: true,
+    if: Proc.new { |s| VALID_SUBSCRIBABLES.include?(s.subscribable_type) }
```
zz9pzza: And this too.

spec/models/subscription_spec.rb
```diff
@@ -63,18 +63,29 @@
     end
 
     it "should not save" do
-      expect(subscription.save).to be_falsey
```
zz9pzza: It now doesn't match the title ?

And the same below.

sarken:
Travis and james are both happy, so merged! Just a quick note -- could you make sure to use the issue template in the future? (It's mostly just the top part with the link that's important, since it makes it easier to go straight to JIRA and make sure the issue status and fix version are correct.)

PR #2980, opened 2017-08-02
AO3-5034 Rails 5 dot 1 upgrade

WendyBeth:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5034
## Purpose

What does this PR do?

This PR upgrades the entire application to Rails 5.1.

## Testing

Due to the fact that this is a massive set of changes, extensive testing will be needed to ensure that application behavior has not changed.

Particular attention should be paid to:

+ Making sure that all expected routes exist. Removing the whitelisted route (`get ':controller(/:action(/:id(.:format)))'`) from config/routes.rb made a significant number of routes used throughout the application invalid. These *should* all be whitelisted, but if any have been missed they will need to be added to the routes list (even/especially non-RESTful routes, such as calls to `#update` without an id that end up being read as '/resources/update' instead of '/resources/:id'.

+ Ensuring that all objects save properly (including sending out expected emails and anything else that might happen in before/after callbacks) when submitting forms since significant changes were made to callbacks with the intention of maintaining existing behavior.

## References

This is rolled in with changes from https://github.com/otwcode/otwarchive/pull/2958, so that should be closed.

## Credit

Wendy Randquist - Littlelines

she/her


spec/mailers/admin_mailer_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'spec_helper'
```
sarken: Hmm, I think this spec is for the feature we removed in #2941. I'm a little confused why it's passing.

app/models/comment.rb
```diff
@@ -263,25 +263,22 @@ def notify_parent_comment_owner
...
-            if self.edited_at_changed?
+            if self.saved_change_to_edited_at?
               update_feedback_in_inbox(parent_comment_owner)
             else
               add_feedback_to_inbox(parent_comment_owner)
             end
           end
-          if parent_comment_owner
-            return parent_comment_owner
-          end
```
elzj: Taking this out is a problem, because in after_update, we actually do this:

     if (parent_comment_owner = notify_parent_comment_owner)

So we do want it to return the value and not just nil all the time. Or we could change the other part of the code.

spec/mailers/admin_mailer_spec.rb
WendyBeth: There must have been some awkwardness in the merging. It passed because the method its referencing also got merged back in. I've removed both the file and the method.

app/models/comment.rb
WendyBeth: Got it.

WendyBeth: I'm a little wary that might break a test - I honestly can't think of why I would have removed these lines otherwise. But all of the tests I could think of to run associated with comments passed locally so perhaps it was just an odd oversight.

PR #2981, opened 2017-08-03
AO3-5089: Allow collections with bookmarks to be revealed

elzj:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5089
## Purpose

Fixes bug that prevented an unrevealed collection containing at least one bookmark from being revealed.

## Testing

Create unrevealed collection, add bookmark to it, reveal, check status after reveal process runs in resque.


app/models/collection_item.rb
```diff
@@ -187,10 +187,9 @@ def send_work_invitation
 
   after_update :notify_of_status_change
   def notify_of_status_change
-    if saved_change_to_unrevealed?
-      # making sure notify_recipients in the work model has not already notified 
-      # the user
-      if !work.new_recipients.blank?
+    if saved_change_to_unrevealed? && item.respond_to?(:new_recipients)
+      # making sure that creation_observer.rb has not already notified the user
```
sarken: I think this comment might be a little outdated -- creation_observer no longer exists.

sarken:
The only action this needs is some clarification re: testing/reproduction, but until then I'm going to mark it Needs Coder Action instead of Awaiting Review so I stop seeing it and going, "Oh, let me review that." No rush, though!

elzj:
@sarken Honestly, I'm not completely sure about reproducing this, but I stumbled on the bug in resque, when it had prevented a collection from being revealed. As long as it works, it's better to be safe than sorry?

app/models/collection_item.rb
```diff
@@ -187,10 +187,9 @@ def send_work_invitation
 
   after_update :notify_of_status_change
   def notify_of_status_change
-    if saved_change_to_unrevealed?
-      # making sure notify_recipients in the work model has not already notified 
-      # the user
-      if !work.new_recipients.blank?
+    if saved_change_to_unrevealed? && item.respond_to?(:new_recipients)
+      # making sure notify_recipients in the work model has not already notified
+      if item.new_recipients.present?
```
tickinginstant: This touches on the `item`/`work` divide on `CollectionItem`s, so I decided to double-check it.

In the old version of this code, this function doesn't do anything because `work` is fresh from the database, and so `new_recipients` is always blank (even when `item` has new recipients). In other words, this currently isn't capable of sending out any notifications under any circumstances, just because it uses `work` rather than `item`. So in theory, the switch from `work` to `item` here could result in duplicate notifications (from here and from wherever they're currently being sent out).

However, I don't think that's actually a problem here because this is an `after_update` hook, which means that with the current definition of `collection_names=`, this function can't be called while the work is being edited (which is what would cause `item.new_recipients` to be present). So at the moment, this function just plain doesn't do anything either way.

The change does prevent the error, though.

(Personally, I'd be inclined to delete the function entirely, but this also works.)

tickinginstant:
I just realized: I think this also fixes [AO3-5532](https://otwarchive.atlassian.net/browse/AO3-5532), since I think `notify_of_status_change` is the only `CollectionItem` function that assumes the item is present (instead of checking first).

PR #2985, opened 2017-08-06
AO3-5096 Fix an old FIXME by removing a duplicate method

elzj:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5096

## Purpose

Removes duplicate method in the external work model and uses alias_method instead

## Testing

No behavior should change. Bookmarks of external works should still be appropriately visible.


scottsds:
Looks good!

PR #2986, opened 2017-08-07
AO3-5097 admin users refactoring

elzj:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5097

## Purpose

* Separates the two different admin user forms into two different update actions, reducing some complexity
* Breaks the more complex actions out into a separate form management class
* Generally attempts to simplify the code and break things down into smaller and more manageable methods

## Testing

Behavior of both admin user forms should be the same as it currently is, but that should be verified, especially in regard to the page with the admin actions and the next of kin updates. Needs admin access.


app/views/admin/admin_users/show.html.erb
```diff
@@ -22,7 +22,7 @@
 
   <!--main content-->
   <h3 class="heading"><%= ts("Abuse Administration") %></h3>
-  <%= form_tag :action => "update_user", :controller => "admin/admin_users" do %>
+  <%= form_tag :action => "update_status", :controller => "admin/admin_users" do %>
```
sarken: As long as you're here, could you switch this to the JSON-style syntax like we use in app/views/admin/admin_users/_user_form.html.erb?

app/models/user_manager.rb
```diff
@@ -0,0 +1,157 @@
...
+    errors.join(" ")
+  end
+
+  private
+
+  def validate_user_and_admin
+    if user && admin
+      return true
+    else
+      errors << "Must have a valid user and admin account to proceed"
```
sarken: Should this have a period at the end, like most of the other messages?

app/models/user_manager.rb
```diff
@@ -0,0 +1,157 @@
...
+      return true
+    end
+  end
+
+  def save_next_of_kin
+    return true if user.fannish_next_of_kin.nil? && kin_name.blank?
+    if FannishNextOfKin.update_for_user(user, kin_name, kin_email)
+      successes << "Fannish next of kin was updated."
+    else
+      errors << "Fannish next of kin failed to update"
```
sarken: Should this have a period, too?

app/views/admin/admin_users/show.html.erb
elzj: Sure, done!

app/models/fannish_next_of_kin.rb
```diff
@@ -8,4 +8,26 @@ class FannishNextOfKin < ActiveRecord::Base
...
+        kin_email: kin_email
+      )
+    # second scenario: update user's FNOK
+    elsif kin_user.present? && kin_email.present?
+      current_fnok.update_attributes(
+        kin_id: kin_user.id,
+        kin_email: kin_email
+      )
+    # third scenario: delete user's FNOK
+    elsif kin_name.blank? && kin_email.blank?
```
sarken: Should this check if `current_fnok.present?` before attempting to destroy it?

app/models/user_manager.rb
```diff
@@ -0,0 +1,157 @@
...
+    if error
+      errors << error
+      false
+    else
+      true
+    end
+  end
+
+  def save_next_of_kin
+    return true if user.fannish_next_of_kin.nil? && kin_name.blank?
```
sarken: This doesn't check if `kin_email.blank?`... would that allow it to save if you entered an email but no name for the FNOK?

app/models/user_manager.rb
```diff
@@ -0,0 +1,157 @@
...
+      false
+    end
+  end
+
+  def save_admin_action
+    return true if admin_action.blank?
+    if admin_action == 'spamban'
+      ban_user
+    else
+      send("#{admin_action}_user")
```
sarken: Do we do anything to restrict the possible values of `admin_action` so we don't risk people making up their own values by editing the form? (I don't think there's much chance of that, since this is an admin thing, but...)

elzj: Yeah, I thought about adding that, but in this case, you'll just get a 500 error if you do, because the model won't respond to 'dropkick_user' or whatever you put in there.

sarken: That works for me! When you have a minute, could you answer the other two questions on this PR? If they're okay, then it's ready to merge.

- https://github.com/otwcode/otwarchive/pull/2986#discussion_r134664590
- https://github.com/otwcode/otwarchive/pull/2986#discussion_r134665631

sarken:
Just marking Needs Coder Action for the pending questions -- no rush, and otherwise 👍

app/models/user_manager.rb
elzj: Updated with changes

app/models/fannish_next_of_kin.rb
sarken: I think this is the only outstanding question/last hurdle to this getting marked ready to mere. (It's possible it was answered in chat, but August was a long time ago. Sob.)

sarken: I've pulled the code and I don't get any weird errors when the FNOK is nil, so I'll call this ready to merge.

PR #2987, opened 2017-08-07
AO3-5033 Typo in minified TinyMCE file prevented it from loading

sarken:
## Issue

## Purpose

In the Rails 5.0 pull request, the minified TinyMCE file was modified to add a newline at the end of the file. However, it also introduced a typo into the file which, which prevented TinyMCE from loading. Codeship's poltergeist settings caught this error, causing our tests to fail. This pull request restores the old, typo-free version of the JS file, meaning TinyMCE will work again and our tests will pass, so we can deploy Rails 5.0 to staging.


sarken:
Hold that thought. This might be the wrong version of the file...

sarken:
Nope, it's fine!

Here's what hung me up: the version of this file from the Rails 5.0 PR contained the following:
```
rtl",force_p_newlines:true;debugger;browser_spellcheck:true,
```

But our previous master contained:
```
rtl",browser_spellcheck:true,
```

The version that was on master (and is in this pull request) is correct.

elzj:
So we want to remove the line rather than fix the typo? Would 'force_p_newlines' add anything useful?

sarken:
Oh, sorry! The typo I'm fixing is replacing `ounction` with `function`.

The `force_p_newlines` bit was weird -- I didn't notice it until I had submitted this pull request and was looking at the diff, hence my earlier comments. There's no value to adding it, as far as I could tell. The official documentation (http://archive.tinymce.com/wiki.php/Configuration:force_p_newlines) says it relates to paragraph creation in Firefox, so I tested this branch against what's currently on production, and there's no change in behavior. 

On both this branch and production, pressing Return/Enter produces this:
```
<p>hello</p>
<p>it's me</p>
<p>your friendly paragraph</p>
```

And pressing Shift+Return/Enter produces this:
```
<p>a friendly paragraph<br />test</p>
```

elzj:
Okay, then, it looks good to me!

PR #2988, opened 2017-08-07
WIP: AO3-5098 Ensure javascript errors are noticed in testing

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5098

## Purpose

This will only effect test runs and needs no manual testing.
However if we have javascript errors in future the automated tests should be more helpful.

## Testing

No manual testing required.


sarken:
Is this still a work in progress, or is it ready for review?

redsummernight:
The JS driver has `js_errors: true` set by default, but it only seems to re-throw run-time errors. Syntax errors are silent, the broken code will only be found if we have tests checking for its results, and we don't currently have tests that use the RTE editor.

After talking with @zz9pzza in chat, there's nothing more we need to change on the JS driver configuration for now.

PR #2994, opened 2017-08-09
AO3-3704 Users can add coauthor work skins to works

cyrilcee:
## Issue

https://otwarchive.atlassian.net/browse/AO3-3704

## Purpose

Users can now add coauthors' work skins to works from the Edit Work page. Users still won't be able to add coauthors' work skins from the Edit Multiple Works page.

## Testing

1. Create a work with a work skin
2. Add a co-author
3. Post work
4. Have co-author edit work

If bugfix is working, co-author should be able to edit work without original author's work skin getting stripped. They should also be able to add or remove their coauthor's work skin at any time.

### DIFF
--- a/features/works/work_edit.feature
+++ b/features/works/work_edit.feature
@@ -144,6 +144,22 @@ Feature: Edit Works
     Then I should see "ex_friend" within ".byline"
       And I should not see "coolperson" within ".byline"
 
+  Scenario: User applies a coauthor's work skin to their work
+    Given the following activated users with private work skins
+        | login       |
+        | lead_author |
+        | coauthor    |
+        | random_user |
+      And I coauthored the work "Shared" as "lead_author" with "coauthor"
+      And I am logged in as "lead_author"
+    When I edit the work "Shared"
+    Then I should see "Lead Author's Work Skin" within "#work_work_skin_id"
+      And I should see "Coauthor's Work Skin" within "#work_work_skin_id"
+      And I should not see "Random User's Work Skin" within "#work_work_skin_id"
+    When I select "Coauthor's Work Skin" from "Select Work Skin"
+      And I press "Post Without Preview"
+    Then I should see "Work was successfully updated"
+
   Scenario: A work cannot be edited to remove its fandom
     Given basic tags
       And I am logged in as a random user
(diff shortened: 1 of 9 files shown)


features/works/work_edit.feature
```diff
@@ -144,6 +144,22 @@ Feature: Edit Works
...
+  Scenario: User applies a coauthor's work skin to their work
+    Given the following activated users with private work skins
+        | login       |
+        | lead_author |
+        | coauthor    |
+        | random_user |
+      And I coauthored the work "Shared" as "lead_author" with "coauthor"
+      And I am logged in as "lead_author"
+    When I edit the work "Shared"
+    Then I should see "Lead Author's Work Skin" within "#work_work_skin_id"
```
sarken: For Cucumber tests, we generally prefer to use the label text rather than the field id or name to better reflect what the user sees.

features/works/work_edit_multiple.feature
```diff
@@ -138,3 +138,16 @@ Feature: Edit Multiple Works
     Then I should not see "coauthor" within ".byline"
     When I view the work "Shared Work 2"
     Then I should not see "coauthor" within ".byline"
+
+  Scenario: User applies a private work skin to multiple coauthored works
+    Given the following activated users with private work skins
+      | login       |
+      | lead_author |
+      | coauthor    |
+    And I am logged in as "lead_author"
```
sarken: Could you indent the `And`s two spaces deeper than the `Given`s, `When`s, and `Then`s?

cyrilcee: Fixed.

features/works/work_edit.feature
cyrilcee: This ask would be pretty straightforward if I was looking at the actual value of the field, but testing what select options are available (not selected) with only a given label is surprisingly complicated.

I updated line 159 to use the label, since I'm just selecting in that line. But for lines 156-158, how important is it to create a new step that lets people test the presence of select options by label?

sarken: Oh, sorry about that! I thought `The field labeled "Label" should contain "option"` would work do that, but it looks like it can't handle select fields with more than one option. I think we can live with the field id here.

PR #2998, opened 2017-08-14
AO3-5033: Fix issue preventing work posting

elzj:
## Issue

## Purpose

Fixes issue where the chapter position fixing callback was trying to update all works on staging with a null work id (which is a separate problem) and then timing out, preventing any new works from being posted.

## Testing

First, we should check that works can be posted on staging. Next we should test chapter position updating as part of the regression, ie if you post a new chapter 3 for a work with 5 chapters, it should slot into the appropriate spot and all other chapters should have their positions updated appropriately.


sarken:
Closing, since we merged #3000 instea

PR #3003, opened 2017-08-23
AO3-5107 Rails BOT: Restore custom link redering for pagination

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5107

## Purpose

Makes it so we continue to use our custom HTML for pagination.

## Testing

Refer to JIRA

## References

https://github.com/mislav/will_paginate/wiki/Link-renderer

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -554,11 +554,12 @@ module ApplicationHelper
   # change the default link renderer for will_paginate
   def will_paginate(collection_or_options = nil, options = {})
     if collection_or_options.is_a? Hash
-      options, collection_or_options = collection_or_options, nil
+      options = collection_or_options
+      collection_or_options = nil
     end
     unless options[:renderer]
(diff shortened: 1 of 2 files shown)


app/helpers/application_helper.rb
```diff
@@ -550,4 +550,16 @@ def first_paragraph(full_text, placeholder_text = 'No preview available.')
       return content_tag(:p, ts(placeholder_text), class: 'placeholder')
     end
   end
+
+  # change the default link renderer for will_paginate
+  def will_paginate(collection_or_options = nil, options = {})
```
redsummernight: Does this mean we can remove the `will_paginate` method in config/initializers/gem-plugin_config/will_paginate_config.rb?

sarken: Let's find out! (It's been broken on my development environment since one of the 4.x upgrades, so I can't tell by testing there.)

PR #3005, opened 2017-08-24
AO3-5110 Rails BOT: Don't try to autoload observers that no longer exist

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5110

## Purpose

In Rails 5.0 (#2958), we moved some code out of observers and into the models they actually pertain to, but our application.rb config was still trying to autoload these observers.

## Testing

None beyond the regression test


sarken:
I found a lingering code comment that refers to an observer--I'll correct that in this pull request, too, but I need to sleep first.

PR #3013, opened 2017-08-26
AO3-5123 Refactor tag display methods

elzj:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5123

## Purpose

Adds some convenience methods for tags to allow for greater flexibility in changing names for display without needing extra logic in the views.

## Testing

Nothing should change; warning tags should still display the appropriate names in the appropriate places.


app/models/warning.rb
```diff
@@ -3,6 +3,16 @@ class Warning < Tag
   NAME = ArchiveConfig.WARNING_CATEGORY_NAME
   index_name Tag.index_name
 
+  DISPLAY_NAME_MAPPING = {
+    ArchiveConfig.WARNING_DEFAULT_TAG_NAME => ArchiveConfig.WARNING_DEFAULT_TAG_DISPLAY_NAME,
+    ArchiveConfig.WARNING_NONE_TAG_NAME => ArchiveConfig.WARNING_NONE_TAG_DISPLAY_NAME,
+    ArchiveConfig.WARNING_SOME_TAG_NAME => ArchiveConfig.WARNING_SOME_TAG_DISPLAY_NAME,
```
sarken: Do we still need this? We got rid of the "Some Archive Warnings Apply" option....idk, before I came onboard, so a long time ago. I think this is the only remaining reference to it; I don't even see anything in the config file: https://github.com/otwcode/otwarchive/blob/master/config/config.yml#L158

 If we have anything in the database still using this warning, we should probably do an announcement and clean it up.

sarken:
Don't mind me -- just another case of just swapping the labels so it's easier to see what's actually unreviewed.

app/models/warning.rb
elzj: I can get rid of it if you want, sure.

sarken: Let's do that and call this ready to merge 👍

sarken:
Unfortunately, this has gone unmergeable.

sarken:
Merge conflicts /o\

app/models/freeform.rb
```diff
@@ -2,6 +2,10 @@ class Freeform < Tag
 
   NAME = ArchiveConfig.FREEFORM_CATEGORY_NAME
 
+  def self.label_name
+    "Additional Tags"
```
redsummernight: Maybe we should have used `ArchiveConfig.FREEFORM_CATEGORY_NAME` for this. Same with `label_name` of the other tag types.

PR #3018, opened 2017-08-30
AO3-3958 Test unescaped HTML in title field when editing work/series

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-3958

## Purpose

AO3-3958 was already fixed at some point, I'm just adding tests.

## Testing

See issue.


redsummernight:
Yeah, the title is stored in the escaped form (`&lt;`) so using `Work.find_by(title: "<")` would miss it, and I don't want to make the steps more complex.

PR #3024, opened 2017-09-01
AO3-5129: Fix issue with remember_me amnesia

elzj:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5129

## Purpose

Fixes issue with remember_me setting on login not being respected

## Testing

Log in, click remember me, close browser, reopen - you should still be logged in

### DIFF
--- a/app/controllers/user_sessions_controller.rb
+++ b/app/controllers/user_sessions_controller.rb
@@ -18,10 +18,11 @@ class UserSessionsController < ApplicationController
 
   def create
     if params[:user_session]
-      @user_session = UserSession.new(
-        login: params[:user_session][:login],
-        password: params[:user_session][:password]
-      )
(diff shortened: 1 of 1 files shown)


app/controllers/user_sessions_controller.rb
```diff
@@ -34,7 +35,7 @@ def create
             if user.updated_at > 1.week.ago
               # we sent out a generated password and they're using it
               # log them in
-              @current_user = UserSession.create(user, params[:remember_me]).record
+              @current_user = UserSession.create(user, user_sesssion_params['remember_me']).record
```
sarken: Travis helpfully pointed out there's an extra `s` in the middle of `sessions` here -- can you also change to double quotes when you fix that? Then we should be ready to go 🤞

PR #3029, opened 2017-09-04
AO3-5144 Add parent thread link for comments

elzj:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5144

## Purpose

Adds a 'parent thread' link to comments.

## Testing

Reply to comments, see that the link appears and takes you to the right place.

### DIFF
--- a/app/views/comments/_comment_actions.html.erb
+++ b/app/views/comments/_comment_actions.html.erb
@@ -6,6 +6,11 @@
   <% end %>
   <% unless comment.unreviewed? %>
     <li><%= link_to ts("Thread"), comment %></li>
+    <% if comment.depth > 0 %>
+    <li>
+      <%= link_to ts("Parent Thread"), comment_url(id: comment.thread) %>
+    </li>
+    <% end %>
   <% end %>
   <% if is_author_of?(comment) && comment.count_all_comments == 0 %>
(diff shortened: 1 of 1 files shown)


app/views/comments/_comment_actions.html.erb
```diff
@@ -6,6 +6,11 @@
   <% end %>
   <% unless comment.unreviewed? %>
     <li><%= link_to ts("Thread"), comment %></li>
+    <% if comment.depth > 0 %>
+    <li>
```
sarken: Not blocking, but could you tweak the indenting on this slightly to make it clearer what the `if` encapsulates? 
```
<% if comment.depth > 0 %>
  <li>
    <%= link_to
  </li>
<% end %>
```

app/views/comments/_comment_actions.html.erb
```diff
@@ -6,6 +6,11 @@
...
+      <%= link_to ts("Parent Thread"), comment_url(id: comment.thread) %>
```
sarken: Does this need to be `comment_url` or can we use `comment_path` to avoid any https/http oddness?

elzj: Oh, sure. I don't know why I had that backwards, because it did seem very strange to me, but this makes more sense.

PR #3032, opened 2017-09-07
AO3-5148: Fix broken redirect when replying from inbox

elzj:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5148

## Purpose

Fixes issue where replying to a comment from your inbox, with filters set, would error after creating the reply and attempting to redirect you. Also fixes the same issue with the review action from the inbox, and adds a couple of checks for a potentially nil request.referer.

## Testing

Replying to and approving comments from the inbox should work with or without filters set.

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -548,6 +548,6 @@ class CommentsController < ApplicationController
   end
 
   def filter_params
-    params.require(:filters).permit!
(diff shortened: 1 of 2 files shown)


zz9pzza:
It looks reasonable to me, but I can hear the requests for tests :)

sarken:
james nailed it ;)

PR #3039, opened 2017-09-09
AO3-5153: New approach to fixing auth tokens dynamically

elzj:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5153

## Purpose

Dynamically updates form authenticity tokens for logged out users, to get around page caching bugs.

## Testing

You should be able to log in on staging.

### DIFF
--- a/public/javascripts/application.js
+++ b/public/javascripts/application.js
@@ -11,6 +11,7 @@ $j(document).ready(function() {
     attachCharacterCounters();
     setupAccordion();
     setupDropdown();
+    updateCachedTokens();
 
     // remove final comma from comma lists in older browsers
     $j('.commas li:last-child').addClass('last');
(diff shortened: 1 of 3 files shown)


zz9pzza:
Don't we still need the cross site forgery protection even for guests ?

elzj:
@zz9pzza This still uses the user's session to get their own form tokens, it doesn't actually bypass security.

zz9pzza:
Can we do the bit where the javascript is only used for guests ?

elzj:
It's only invoked if it sees the login form, which means you're logged out.

public/javascripts/application.js
```diff
@@ -643,3 +644,18 @@ function thermometer() {
...
+  // we only do full page caching when users are logged out
+  if ($j('#small_login').length > 0) {
+    $j.getJSON("/token_dispenser.json", function( data ) {
+      var token = data.token;
+      //set token on fields
+      $j('input[name=authenticity_token]').each(function(){
+        $j(this).attr('value', token);
+      });
+      $j('meta[name=csrf-token]').attr('value', token);
+      console.log("I got a token: " + token);
```
zz9pzza: Is this a good idea to deploy ?

PR #3044, opened 2017-09-10
AO3-5157: Update log levels after Rails default change

elzj:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5157

## Purpose

Makes staging and production logs less chatty

## Testing

N/A


sarken:
I see no compelling reason to wait for Travis on this

PR #3046, opened 2017-09-10
AO3-5160: Add auth error page

elzj:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5160

## Purpose

Adds a new error page to handle Invalid Authenticity Token errors, to give users some context for the problem they're running into.

## Testing

See issue.


app/views/errors/auth_error.html.erb
```diff
@@ -0,0 +1,2 @@
+<h2 class="heading">Session Expired</h2>
+<p>Your current session has expired and we can't authenticate your request. Try logging in again, refreshing the page, or clearing your cache if you continue to experience problems.</p>
```
sarken: On the maintenance page, we link to instructions for clearing your browser cache. We should probably do that here, too: `<a href="http://kb.iu.edu/data/ahic.html">clearing your cache</a>`

PR #3048, opened 2017-09-11
AO3-5163: Revert work complete-setting method to old code

elzj:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5163

## Purpose

Reverts work complete-setting method to old code to fix bug where works going from complete to incomplete didn't have their status updated.

## Testing

See issue.

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -792,10 +792,13 @@ class Work < ApplicationRecord
   end
 
   after_save :update_complete_status
+  # Note: this can mark a work complete but it can also mark a complete work
+  # as incomplete if its status has changed
   def update_complete_status
     # self.chapters.posted.count ( not self.number_of_posted_chapter , here be dragons )
-    if self.chapters.posted.count == expected_number_of_chapters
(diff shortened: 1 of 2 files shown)


app/models/work.rb
```diff
@@ -792,10 +792,13 @@ def multipart?
...
   after_save :update_complete_status
+  # Note: this can mark a work complete but it can also mark a complete work
+  # as incomplete if its status has changed
   def update_complete_status
     # self.chapters.posted.count ( not self.number_of_posted_chapter , here be dragons )
-    if self.chapters.posted.count == expected_number_of_chapters
-      Work.where("id = #{self.id}").update_all("complete = true")
+    self.complete = self.chapters.posted.count == expected_number_of_chapters
+    if self.will_save_change_to_attribute?(:complete)
+      Work.where("id = #{self.id}").update_all("complete = #{self.complete}")
```
zz9pzza: So stupid question, could we change this to a before_save and reduce the number of writes to the database ? ( and make the code look smaller as it doesn't need the will_save_change_to_attribute?

PR #3052, opened 2017-09-11
AO3-5042 Use _url for absolute links in bylines and download afterword

tuff:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5042

## Purpose

This changes byline and comment links in download files to use absolute, rather than relative URLs.

## Testing

STR are in the issue, but I found these simpler ones:

1. Log in
2. Post > New Work
3. Fill in required information
4. Post Without Preview
5. Download > HTML
6. Open the downloaded file

Expected: the byline link in the preface and the "drop by the archive and comment" link in the afterword should be absolute.
Actual: those links are relative, so they have hrefs of `file://...` and are broken when viewing the downloaded file locally.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -149,7 +149,7 @@ module ApplicationHelper
     if only_path
       link_to(pseud.byline, user_pseud_path(pseud.user, pseud), rel: "author")
     else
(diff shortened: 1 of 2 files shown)


sarken:
Oh, quick question! What name should we use when crediting you in the release notes -- tuff?

tuff:
@sarken that's fine, thanks!

PR #3054, opened 2017-09-12
AO3-5167 Improve performance of author checks

elzj:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5167

## Purpose

Switches is_author_of? method to compare based on ids rather than objects, which should save on some db calls and memory. Also preloads the set of current works/bookmarks owned by the current user on index pages so that we don't have to run separate checks over and over.

## Testing

Nothing should change: appropriate footers for work and bookmark blurbs should appear if they're yours, and anonymity should be respected.


sarken:
This has merge conflicts that need to be resolved.

elzj:
Replaced by https://github.com/otwcode/otwarchive/pull/3380

PR #3055, opened 2017-09-12
AO3-4744 Fix auto-filling work info for previously-bookmarked URLs

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4744

## Purpose

Makes it so the work info once again gets auto-filled when you're bookmarking a URL that has been bookmarked before. 

## Testing

Refer to JIRA

## References

~Merge after #2656, which is included in this PR~ Done


features/bookmarks/bookmark_external.feature
```diff
@@ -101,3 +101,45 @@ Feature: Create bookmarks of external works
     Then I should not see "Bookmark External Work"
     When I go to the bookmarks in collection "Testing BEW Collection"
     Then I should not see "Bookmark External Work"
+
+  @javascript
+  Scenario: When using the bookmarklet on a URL that has already been bookmarked, the work information will be automatically filled in
+    Given "first_user" has bookmarked an external work
+      And I am logged in as "second_user"
+    When I use the bookmarklet on a previously bookmarked URL
+      And I wait 1 second
```
redsummernight, line 110: Does this work without the wait step and with #3812?

features/step_definitions/bookmark_steps.rb
```diff
@@ -362,6 +362,11 @@
   click_button "Edit"
 end
 
+When(/^I use the bookmarklet on a previously bookmarked URL$/) do
```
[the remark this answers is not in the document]
sarken: I don't think so. This isn't going to an existing external work, but rather opening the bookmarklet with URL parameters that match a URL already on the Archive.

features/step_definitions/external_work_steps.rb
```diff
@@ -28,6 +32,25 @@
   click_button("Create")
 end
 
+Given /^"([^"]*)" has bookmarked an external work$/ do |user|
+  step %{mock websites with no content}
+  step %{basic tags}
+  step %{I am logged in as "#{user}"}
+  visit new_external_work_path
+  # We use the labels for some fields because the ids change when
+  # JavaScript is enabled
```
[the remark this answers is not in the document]
sarken: I meant it as a comment on why I was deviating for our standard preference for `ids` over labels. I'll clarify.

features/bookmarks/bookmark_external.feature
sarken: No, it turns out I added the wait to give the JavaScript a chance to populate the other fields based on the selected URL, not because of a delay with the autocomplete menu appearing. 

I've put the wait in the `When I use the bookmarklet on a previously bookmarked URL` step definition now with a comment so we don't waste time trying to fix it later.

sarken:
Closing so I can reopen #3180 and #3181.

sarken:
...I can't count. *reopens*

sarken:
Closing since this has merge conflicts. I've reopened #3657 instead.

sarken:
Reopening to run tests, since they're failing locally even though the code is working...

sarken:
It looks like the title, creator, and summary fields are getting filled in when the tests run, but not the fields for tags. :shrug:

features/bookmarks/bookmark_external.feature
```diff
@@ -101,3 +101,41 @@ Feature: Create bookmarks of external works
...
+    Given "first_user" has bookmarked an external work
+      And I am logged in as "second_user"
+    When I go to the new external work page
+      And I choose a previously bookmarked URL from the autocomplete
+      And I press "Create"
+    Then I should see "Bookmark was successfully created."
+      And the work info for my new bookmark should match the original
+
+  @javascript
+  Scenario: When getting an error (e.g. because you the title is blank) after using the autocomplete to select a previously-bookmarked URL, any changes you made to the work information are not overridden on the edit page
```
redsummernight, line 125: Extra "you" before "the title".

app/views/external_works/fetch.js.erb
```diff
@@ -2,14 +2,13 @@
...
   var categories = <%= @external_work.category_strings.to_json.html_safe %>;
-  $j('input[name="external_work[category_strings][]"]').each(function() {
-    $j(this).prop("checked", categories.includes(this.value));
-  });
-  $j('#external_work_fandom_string').val("<%= @external_work.fandom_string %>");
-  $j('#external_work_relationship_string').val("<%= @external_work.relationship_string %>");
-  $j('#external_work_character_string').val("<%= @external_work.character_string %>");
+  $j.each(categories, function(i, category) {
+     $j("input[value='" + category + "']").prop('checked', true);
+   });
```
tickinginstant, line 10: I don't know if it matters, but if the user already has some categories checked, this won't uncheck them to match the categories of the retrieved external work.

app/views/external_works/fetch.js.erb
```diff
@@ -2,14 +2,17 @@
...
+  var unwanted_categories = <%= (Category.canonical.pluck(:name) - @external_work.category_strings).to_json.html_safe %>;
+  $j.each(work_categories, function(i, category) {
+    $j("input[value='" + category + "']").prop('checked', true);
   });
-  $j('#external_work_fandom_string').val("<%= @external_work.fandom_string %>");
-  $j('#external_work_relationship_string').val("<%= @external_work.relationship_string %>");
-  $j('#external_work_character_string').val("<%= @external_work.character_string %>");
+  $j.each(unwanted_categories, function(i, category) {
+    $j("input[value='" + category + "']").prop('checked', false);
+  });
```
sarken, line 14: This is completely hideous, but it refuses to work when attempting to use a function to determine whether the `checked` property should be true or false. And I don't mean it refused to check or uncheck the box -- I mean it wouldn't fill in any of the other fields, either.

If someone else wants to try, please feel free!

redsummernight: I think the problem is that the tests are not consistently waiting for the AJAX request (fetching external work info) to finish. The failures are not because of the script to check/uncheck the categories.

We now have a step to wait for AJAX requests, which should be included in both the step to use the bookmarklet and the step to pick a URL from the autocomplete:

https://github.com/otwcode/otwarchive/blob/5c96d176853af424d981d9e6ac208cc61f3a3001/features/step_definitions/generic_steps.rb#L49-L51

app/views/external_works/fetch.js.erb
```diff
@@ -2,14 +2,17 @@
   $j('#external_work_author').val("<%= @external_work.author %>").change();
   $j('#external_work_title').val("<%= @external_work.title %>").change();
   $j('#external_work_summary').val("<%= escape_javascript(@external_work.summary&.html_safe) %>").change();
-  $j('#external_work_summary').html("<%= escape_javascript(@external_work.summary&.html_safe) %>").change();
   $j('#fetched').val("<%= @external_work.id %>");
   $j('#external_work_rating_string').val("<%= @external_work.rating_string %>");
-  var categories = <%= @external_work.category_strings.to_json.html_safe %>;
-  $j('input[name="external_work[category_strings][]"]').each(function() {
```
redsummernight, line 9: This already work, but we can make it more specific:

```js
$j('input[type="checkbox"][name="external_work[category_strings][]"]').each(function() {
```
to remove a hidden field from the matched elements.

app/views/external_works/fetch.js.erb
sarken: I can add that and try again, but I'll be shocked if it works: my various unsuccessful attempts were consistently failing both in tests _and_ in the browser (I think the time it happened worked the other day was caching), whereas this version and the previous version (without the unchecking function) consistently passed.

sarken: Yeah, no, same result.

tickinginstant: It looks like the `includes` function isn't available in older browsers (including PhantomJS, I guess?), but it might work to use `indexOf` or `$j.inArray` instead (and check for `-1`).

sarken: asdfkjlsjg

The best part is I Googled the `includes` function, but didn't check the browser compatibility. 🤦 

I will give it another whirl now that I have this information -- thanks!

app/views/external_works/fetch.js.erb
sarken: I went with a slightly shorter version -- I'm inputs with an `id` that starts with `external_work_category_strings_`, which the hidden field doesn't have.

redsummernight (approved):

PR #3056, opened 2017-09-12
AO3-5169: Improve auth error handling for kudos

elzj:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5169

## Purpose

Displays auth error message when attempting to submit kudos instead of doing nothing and only having a console error.

## Testing

See issue.


sarken:
Just doing some housekeeping and updating the issue number!

PR #3058, opened 2017-09-12
AO3-5170 Update image address in Deviant Art import test

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5170

## Purpose

Updates our DA importing test to look for the proper image address, since the image has been moved. Sticks fingers in ears and goes lalalalala about the possibility that all images imported from DA are not broken. 

## Testing

Make sure automated tests pass

### DIFF
--- a/features/importing/work_import_da.feature
+++ b/features/importing/work_import_da.feature
@@ -11,7 +11,7 @@ Feature: Import Works from deviantart
       And I fill in "urls" with "http://bingeling.deviantart.com/art/Flooded-45971613"
     When I press "Import"
     Then I should see "Preview"
-      And I should see the image "src" text "https://orig03.deviantart.net/4707/f/2007/004/a/7/flooded_by_bingeling.jpg"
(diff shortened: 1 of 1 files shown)


features/importing/work_import_da.feature
```diff
@@ -11,7 +11,7 @@ Feature: Import Works from deviantart
       And I fill in "urls" with "http://bingeling.deviantart.com/art/Flooded-45971613"
     When I press "Import"
     Then I should see "Preview"
-      And I should see the image "src" text "https://orig03.deviantart.net/4707/f/2007/004/a/7/flooded_by_bingeling.jpg"
+      And I should see the image "src" text "https://orig00.deviantart.net/4707/f/2007/004/a/7/flooded_by_bingeling.jpg"
```
redsummernight: Should we check for just the "flooded_by_bingeling.jpg" part instead of the whole link? They might move servers again.

sarken: I think we want to know when they move servers so we know when we're going to be dealing with broken images.

PR #3060, opened 2017-09-15
AO3-5172 Reduce calls to Tag.reset_filter_count in Tag.remove_filter_taggings.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5172

## Purpose

Tag.remove_filter_taggings contains too many calls to Tag.reset_filter_count -- it's called inside the self.works loop, so it'll be called once for each of the works in the tag being de-synned. Each call to reset_filter_count triggers a database query that I suspect has to iterate through all of the relevant filter-taggings, because it's a join between two tables with conditions on both sides (so it can't rely on the indices alone to count). So it's quadratic when it probably shouldn't be -- the filter counts should be reset once all of the filter-taggings have been altered.

## Testing

This just decreases the number of times that the function is being called, so it shouldn't alter the functionality much. (Except that it might be a little faster to de-syn a large tag.) I added some steps to the bug report for testing that reset_filter_count is still being called at least once.


tickinginstant:
Sorry, clicked the wrong button.

PR #3062, opened 2017-09-16
AO3-2782 Don't allow wrangling assignments for non-canonical fandoms

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-2782

## Purpose

1. Prevents anyone from assigning a non-canonical fandom to a wrangler.
2. Adds an after task to delete any existing assignments for non-canonical fandoms, since there's no way to do it through the interface. (It will also create a list of deleted fandoms and wranglers we can give to the wrangling chairs.)

## Testing

Refer to JIRA


lib/tasks/after_tasks.rake
```diff
@@ -500,6 +500,18 @@ namespace :After do
     end
   end
 
+  desc "Destroy wrangling assignments for non-canonical fandoms"
+  task(:destroy_noncanonical_wrangling_assignments => :environment) do
+    fandom_ids = Fandom.where(canonical: false).joins(:wrangling_assignments).pluck(:id)
+    assignments = WranglingAssignment.where(fandom_id: [fandom_ids])
```
sarken: I suspect there is a better, one-step way to get the assignments for non-canonical fandoms, but I couldn't come up with it.

tickinginstant: I think `WranglingAssignment.joins(:fandom).where(tags: { canonical: false })` might work here.

sarken: I think you are probably right! I was trying `WranglingAssignment.joins(:fandom).where(fandoms: { canonical: false })` and getting all sorts of fun messages about fandoms not existing, but it needing to be tags instead makes sense -- thanks!

features/tags_and_wrangling/tag_wrangling.feature
```diff
@@ -295,3 +295,13 @@ Feature: Tag wrangling
     When I view the tag "Cowboy Bebop"
     Then I follow "Reindex Tag"
       And I should see "Tag sent to be reindexed"
+
+  Scenario: A tag wrangler cannot assign a non-canonical fandom to themsleves
```
redsummernight: There's a typo for "themselves".

sarken:
Closing while I revise because I want to make this prettier as long as I'm here.

PR #3066, opened 2017-09-19
AO3-5176: Add more indexes to tag nominations

elzj:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5176

## Purpose

Add more indexes to tag nominations table for slow queries

## Testing

Things should not blow up and should ideally run a bit faster


tickinginstant:
If you're trying to speed up tag nominations with better indices, maybe it'd be worth looking at the tag_set_nominations table, too? The owned_tag_set_id isn't included in the tag_nominations table, so there are an awful lot of tag nomination lookups that join on the tag_set_nominations table to place restrictions on the owned_tag_set_id. And I don't see any indices on the tag_set_nominations table.

elzj:
There are definitely more that should be added, but I just started with the ones I could see common queries for. We should do more of an audit if we have time after this release.

PR #3068, opened 2017-09-19
AO3-4952 REDIS AUTOCOMPLETE missing lib files

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4952

## Purpose

Hopefully fix the issues building the autocomplete database.

## Testing

Look and see if the data is appearing.

### DIFF
--- a/lib/autocomplete_source.rb
+++ b/lib/autocomplete_source.rb
@@ -1,15 +1,15 @@
 module AutocompleteSource
-  AUTOCOMPLETE_DELIMITER = ": "
-  AUTOCOMPLETE_COMPLETION_KEY = "completion"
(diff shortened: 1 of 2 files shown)


lib/autocomplete_source.rb
```diff
@@ -1,31 +1,31 @@
...
   end
 
   def autocomplete_search_string_was
-    "#{name_was}"
+    name_was.to_s
   end
 
   def autocomplete_search_string_before_last_save
-    "#{name_before_last_save}"
+    name_before_last_save.TO_S
```
sarken: Looks like the caps lock here is what's making the tests bomb.

PR #3069, opened 2017-09-21
AO3-5180 Better word breaking for mobile headings

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5180

## Purpose

Modifies our narrow stylesheet so it prefers _not_ to break long words when they'll fit on a single line while still allowing long words that _need_ to be broken to break wherever. (See issue for illustration.)

## Testing

Refer to JIRA


sarken:
Fuck it, I am not untangling this mess.

sarken:
Replaced by #3563

PR #3077, opened 2017-09-24
AO3-4138 making view history more linguistically accessible

pupupuri:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4138

## Purpose

in a user's reading history, changes the user-facing text from "last viewed: (date)" and "viewed: (number)" to "last visited: (date)" and "visited: (number)"

also changes `last_viewed` and `view_count` to `last_visited` and `visit_count` in the code to avoid confusion

## Testing

please view your reading history and make sure that instead of displaying "last viewed: (date)" and "viewed: (number)" the bar under each fic says "last visited: (date)" and "visited: (number)"

## Notes

i had some trouble with my editor and am now left with a feeling i may have made a mistake in trying to change every instance of last_viewed and view_count; if the reviewer could make sure everything is still in place that would be great...


db/schema.rb
```diff
@@ -969,8 +969,8 @@
     t.integer  "user_id"
     t.integer  "work_id"
     t.datetime "created_at"
-    t.datetime "last_viewed"
-    t.integer  "view_count",         :default => 0
+    t.datetime "last_visited"
```
redsummernight: Is it OK to rename these only in the UI? If you rename the database columns you would need to setup a migration.

pupupuri: sure; it'd be much easier just to make the ui change. i was just worried about it causing confusion in the future if someone wants to make a change and there's a disconnect between them

sarken: The disconnect is fine -- column/variable/method/etc names don't need to correspond perfectly to what users see.

Also, I think you might need to update your master branch. Even though this pull request is new, it has merge conflicts, including changes to files that have been deleted.

PR #3078, opened 2017-09-25
AO3-5187 Add autocomplete Redis instance to Redis config examples

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5187

## Purpose

Updates our example Redis config files to include the new Redis instance

## Testing

Nothing to test

### DIFF
--- a/config/redis.codeship.example
+++ b/config/redis.codeship.example
@@ -8,3 +8,5 @@ redis_autocomplete:
   test: localhost:6381
 redis_rollout:
   test: localhost:6381
+redis_autocomplete:
+  test: localhost:6381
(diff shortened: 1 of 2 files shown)


config/redis.codeship.example
```diff
@@ -8,3 +8,5 @@ redis_autocomplete:
   test: localhost:6381
 redis_rollout:
   test: localhost:6381
+redis_autocomplete:
```
redsummernight: This file already has redis_autocomplete.

sarken: Okay, that's... odd. It's not showing up on my local copy, but I see it here. o_0

redsummernight:
I made this change for the Vagrant setup as well. otwcode/ao3_vagrant#1.

PR #3082, opened 2017-09-26
AO3-2312 admin post kudos

tuff:
## Issue

https://otwarchive.atlassian.net/browse/AO3-2312

## Purpose

Lets guests and users leave kudos on admin posts.

## Testing

- While logged in as a user, try to leave kudos on an admin post from that post's page. Check that the success flash message appears and your kudos is listed.
- Try it as a guest.
- Try to leave kudos multiple times. Check that the "you already left kudos" error message appears.
- Try the above steps with JS off.  

Note: I found that a kudo left while logged in as a admin counts as having been left by a guest. I assume that is expected as admins are not users. Let me know in review if that's not what's desired.

### DIFF
--- a/features/comments_and_kudos/kudos.feature
+++ b/features/comments_and_kudos/kudos.feature
@@ -167,3 +167,57 @@ Feature: Kudos
       And the email should contain "Meh Story"
       And the email should not contain "0 guests"
       And the email should not contain "translation missing"
+
+  Scenario: Leave kudos on an admin post when logged in
+
+    Given I have posted an admin post
+      And I am logged in as "happyuser"
+      And I go to the admin-posts page
+      And I follow "Default Admin Post"
+      And all emails have been delivered
+      And I press "Kudos ♥"
+    Then I should see "Thank you for leaving kudos!"
+      And I should see "happyuser left kudos on this post!"
+    When I press "Kudos ♥"
+    Then I should see "You have already left kudos here. :)"
+    When kudos are sent
+    Then 0 emails should be delivered
+
+  Scenario: Leave kudos on an admin post when logged out
+
+    Given I have posted an admin post
+      And I go to the admin-posts page
+      And I follow "Default Admin Post"
+      And I press "Kudos ♥"
+    Then I should see "Thank you for leaving kudos!"
+      And I should see "1 guest left kudos on this post!"
+    When I press "Kudos ♥"
+    Then I should see "You have already left kudos here. :)"
+    When kudos are sent
+    Then 0 emails should be delivered
+
+  @javascript
+  Scenario: Leave kudos on an admin post via JS when logged in
+
+    Given I have posted an admin post
+      And I am logged in as "happyuser"
+      And I go to the admin-posts page
+      And I follow "Default Admin Post"
+      And all emails have been delivered
+      And I press "Kudos ♥"
+    Then I should see "Thank you for leaving kudos!"
+      And I should see "happyuser left kudos on this post!"
+    When I press "Kudos ♥"
+    Then I should see "You have already left kudos here. :)"
+
+  @javascript
+  Scenario: Leave kudos on an admin post via JS when logged out
+  
+    Given I have posted an admin post
+      And I go to the admin-posts page
+      And I follow "Default Admin Post"
+      And I press "Kudos ♥"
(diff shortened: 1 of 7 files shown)


features/comments_and_kudos/kudos.feature
```diff
@@ -125,3 +125,93 @@ Feature: Leave kudos
       And the email should contain "guest"
       And the email should contain "Awesome Story"
       And the email should contain "Another Awesome Story"
+
+  Scenario: Leave kudos on an admin post
+    Given I have posted an admin post
+
+    When I am logged out as an admin
```
ariana-paris: I'd make this an `And` and remove the blank line, but it looks as if this step shouldn't be necessary anyway. Following through the complex chain of variations on `I am logged...` steps, the bottom of the pile is a step definition which calls `I am logged out` which in turn destroys any admin user sessions, so this line can probably be removed. (Same applies to subsequent tests following the same format)

features/comments_and_kudos/kudos.feature
```diff
@@ -125,3 +125,93 @@ Feature: Leave kudos
...
+      And I am logged in as "happyuser"
+      And I go to the admin-posts page
+      And I follow "Default Admin Post"
+      And all emails have been delivered
+      And I press "Kudos ♥"
+    Then I should see "Thank you for leaving kudos!"
+      And I should see "happyuser left kudos on this post!"
+    When I press "Kudos ♥"
+    Then I should see "You have already left kudos here. :)"
+    When I log out
```
ariana-paris: I would personally say that testing kudos giving while logged out is a different scenario than testing it logged in, so I would split this into a new scenario here (same applies to the other tests so you would have pairs of "Leave kudos on admin posts when logged in"/"Leave kudos on admin posts when logged out", or however you want to work them).

ariana-paris (commented):
I'd like someone from front-end to okay the class names and CSS changes, but I made some comments about the tests and aside from that, it looks good to me.

zz9pzza:
So this has lots of merge conflicts.

And now Kudo is given by a users rather than psueds.

sarken:
The kudos button should also only be available on an individual admin post's show page, like the comment form is only available there.

tuff:
@zz9pzza @sarken I've fixed this up and removed the kudos button from the admin posts index.

features/step_definitions/admin_steps.rb
```diff
@@ -101,6 +101,7 @@
 Given /^I have posted an admin post$/ do
   step(%{I am logged in as a "communications" admin})
   step("I make an admin post")
+  step("I am on the homepage")
```
tuff, line 104: Without this, the JS-enabled tests fail as we land on example.com after making the admin post. I don't know why that is...is it related to [this default URL value](https://github.com/otwcode/otwarchive/blob/77a69888b51a0403ffbf5da0b1f3da1df56e2722/config/environments/test.rb#L48)? Please suggest a better fix if you know one!

Bilka2: It's been a while, but there is a better fix now:  
In `AdminPostsController#create` change the `redirect_to(@admin_post)` to `redirect_to admin_post_path(@admin_post)`.

Explanation:
The controller calls `redirect_to` on a model object (instead of a path). When you call `redirect_to` on a model, Rails automatically tries to convert it to a full URL, instead of a path, and when the host is set to `example.com`, that means that it tries to redirect to `example.com`.

features/comments_and_kudos/kudos.feature
```diff
@@ -167,3 +167,57 @@ Feature: Kudos
...
+    Then I should see "You have already left kudos here. :)"
+    When kudos are sent
+    Then 0 emails should be delivered
+
+  Scenario: Leave kudos on an admin post when logged out
+
+    Given I have posted an admin post
+      And I go to the admin-posts page
+      And I follow "Default Admin Post"
+      And I press "Kudos ♥"
```
Bilka2, line 191: If master is merged into this branch (which would be nice), then this test needs an update:

```suggestion
      And I follow "Default Admin Post"
      And all emails have been delivered
      And I press "Kudos ♥"
```

So the email from creating the admin wont mess up the test.

Bilka2:
Adopted at #5931

PR #3091, opened 2017-10-02
AO3-3513 Fix "# recs" links in pseud blurbs

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-3513

## Purpose

Fix "# recs" links in pseud blurbs, as seen on the "All Pseuds" page of a user, or on people search results.

## Testing

See issue.


redsummernight:
After #3191, pseud search results have bookmark links instead of rec links. The only place left with rec links is a user's pseuds page, so we decided to just remove the rec links instead of fixing them.

PR #3094, opened 2017-10-02
AO3-2452 Add checks to CommonTaggings to prevent invalid tag associations.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-2452

## Purpose

Currently, it's possible for wranglers to add invalid relationships (e.g. a Character tag that's the parent of another Character tag, or a Freeform that's the parent of a Relationship tag, or a non-canonical Character that's the parent of a canonical Relationship tag, or a tag that is its own parent). This pull request does several things to fix this:

1. It adds two validation checks to the CommonTagging class: one to make sure that the parent tag is canonical, and one to make sure that the type combination of parent and child is allowed. This prevents the creation of new invalid relationships.

2. It modifies the tag associations code so that new CommonTaggings are handled more like new MetaTaggings, but with error messages passed on to the user to make it clear why the CommonTagging/MetaTagging is invalid. It also uses Tag.parent_types in order to make sure that it handles all of the different types correctly (rather than having each tag type have its own version of add_association).

3. It modifies the fandom_string/character_string/etc. code so that when you enter a tag of the wrong type, it will prevent the Tag from being saved and show an error message.

4. It hides the mass "Wrangle to Fandom(s)" option on the fandom-specific Mergers page and the fandom-specific SubTags page, since that option isn't needed (and would always cause the creation of invalid wrangling relationships in the old system).

5. It adds a task to destroy all invalid CommonTaggings, and a task to destroy all invalid MetaTaggings.

6. It adds some rspec tests to make sure that the new code will add the associations it should, and show errors for the associations it shouldn't.

## Testing

It would be good to have several Tag Wranglers run the code through its paces -- go through their standard routine, and make sure that they don't get any unexpected errors. There are a lot more validity checks now than there were before, and it'd be good to make sure that none of them breaks any useful tricks currently used by the Tag Wrangling team.

## References

Because of the way that I altered the checks for valid associations, this contributes towards [AO3-3526](https://otwarchive.atlassian.net/browse/AO3-3526), the bug where it was impossible to add non-canonical child tags from a canonical tag's edit page. However, it still doesn't include those tags in the autocomplete, and contrary to the desired behavior, if you enter a non-existent child, it will generate an error instead of failing silently.

## Credit

tickinginstant, she/her


app/models/tag.rb
```diff
@@ -1125,49 +1128,79 @@ def remove_favorite_tags
...
+
+  # Find and destroy all invalid CommonTaggings and MetaTaggings associated
+  # with this tag.
+  def destroy_invalid_associations
+    common_taggings.destroy_invalid
+    child_taggings.destroy_invalid
+    meta_taggings.destroy_invalid
+    sub_taggings.destroy_invalid
+  end
+
```
shalott: Just for anyone trying to search for where one of the *_string methods is defined: 
```suggestion
# defines fandom_string=, media_string=, character_string=, relationship_string=, freeform_string= 
```

app/models/tag.rb
```diff
@@ -1081,49 +1084,80 @@ def remove_favorite_tags
...
     names.each do |name|
-      sub = self.class.find_by_name(name)
-      if sub
-        meta_tagging = sub.meta_taggings.build(meta_tag: self, direct: true)
-        unless meta_tagging.valid? && meta_tagging.save
-          self.errors.add(:base, "You attempted to create an invalid meta tagging. :(")
-        end
-      end
+      sub = Tag.find_by_name(name)
+      meta_tagging = sub_taggings.build(sub_tag: sub, direct: true)
```
redsummernight: Very minor, I think this variable is `sub_tagging`.

tickinginstant:
The `common_taggings` table does have a unique index, but the index includes `filterable_type` (which isn't included in the CommonTagging's uniqueness validation -- hence the complaints from `consistency_fail`). But `filterable` really doesn't make use of the fact that it's technically polymorphic, so I think it would be good to drop that column and remake the unique index without it.

sarken:
Switching over to Needs Coder Action due to merge conflicts.

PR #3095, opened 2017-10-03
AO3-5193 Stop non-admins from making canonicals into synonyms

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5193

## Purpose

Prevents non-admins from using the "Synonym of" field on to make canonical tags into synonyms of other canonical tags. This takes the "hide the field" approach rather than the "disable the field" one because disabling it was going to involve fun like modifying the autocomplete to respect the `disabled` attribute on the original input, plus updating the styles for disabled text fields.

https://github.com/otwcode/otwarchive/pull/3095/commits/1fe74a78dc76074a1d7e1bf7de70a073a021379a contains lots of tidying up of the view file: get rid of the hash rockets, use just one kind of quotation mark, adjust the line length where it was easy...

## Testing

Refer to JIRA


app/views/tags/edit.html.erb
```diff
@@ -9,101 +11,128 @@
...
+        <dt><%= f.label :adult, ts("Adult") %></dt>
+        <dd>
+          <%= f.check_box("adult", disabled: !logged_in_as_admin? )%>&nbsp;
+          <%= ts("This tag indicates adult content.") %>
+        </dd>
       <% end %>
 
-      <% if Tag::USER_DEFINED.include?(@tag[:type]) %>
-        <dt><%= f.label :syn_string, ts('Synonym of') %></dt>
+      <% if Tag::USER_DEFINED.include?(@tag[:type]) && !(@tag.canonical? && !logged_in_as_admin?) %>
```
sarken: The additional logic here is the only non-tidying change to this file.

spec/models/tag_spec.rb
```diff
@@ -463,15 +463,38 @@
...
+      @noncanonical_syn_tag = create(:fandom, canonical: false)
+      @noncanonical_syn_tag.syn_string = @canonical_tag.name
+      @noncanonical_syn_tag.save
+
+      expect(@noncanonical_syn_tag.merger).to eq(@canonical_tag)
+      @canonical_tag = Tag.find(@canonical_tag.id)
+      expect(@canonical_tag.mergers).to eq([@noncanonical_syn_tag])
+    end
+
+    it "doesn't let you make a canonical tag the synonym of a canonical one" do
```
redsummernight: This test is failing:
```
  4) Tag multiple tags of the same type doesn't let you make a canonical tag the synonym of a canonical one
     Failure/Error: expect(@syn_tag.merger).to eq(nil)
     
       expected: nil
            got: #<Fandom id: 1023, name: "The 51 Fandom", canonical: true, created_at: "2019-02-03 03:49:53", updated...e, last_wrangler_id: nil, last_wrangler_type: nil, unwrangleable: false, sortable_name: "51 Fandom">
     
       (compared using ==)
     # ./spec/models/tag_spec.rb:492:in `block (3 levels) in <top (required)>'
```

sarken:
Closing this so I can push multiple times without it being super annoying.

redsummernight:
This has conflicts now.

sarken:
Conflicts begone!

sarken:
Merge conflicts I'll deal with later.

sarken:
I think I've successfully resolved the merge conflicts in the tag controller spec file.

sarken:
Merge conflicts, blast it all.

redsummernight:
There's a test failure:

```
Failures:

  1) Tag multiple tags of the same type with a synonym and a subtag and bookmarks under each finds all bookmarks that would need to be reindexed
     Failure/Error: @direct_bm = ;4mFactoryBot.create(:bookmark, tag_string: @canonical_tag.name)

     ActiveRecord::RecordInvalid:
       Validation failed: Invalid creator: 7kngufcg (1x01pxfv482) does not allow others to invite them to be a co-creator., Work must have at least one creator., Invalid chapter: Chapter must have at least one creator.
     # ./factories/bookmarks.rb:6:in `block (3 levels) in <top (required)>'
     # ./spec/models/tag_spec.rb:536:in `block (5 levels) in <top (required)>'
```

spec/models/tag_spec.rb
```diff
@@ -496,6 +519,8 @@
 
         xit "should find all works that would need to be reindexed" do
           # get all the work ids that it would queue
+          User.current_user = create(:admin)
```
redsummernight: Why do we need to be an admin here? Also this test is pending anyway.

sarken: Utterly no idea -- I just knew that we didn't need to be an admin in the scenario with the failure, so I moved the change and checked that the tests passed. I've now removed it entirely as I have no idea why it was originally there. Yay, code from two years ago.

PR #3104, opened 2017-10-13
AO3-4815 Update Elasticsearch and filter interface

WendyBeth:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-4815

## Purpose

What does this PR do?

This PR

+ Adds code for indexing and queueing Elasticsearch indices for Elasticsearch version 5.6.1 (and is compatible, to the best of my knowledge, with 6.x)
+ Adds code to use rollout to split between Elasticsearch version and code usage
+ Adds exclusion filters to bookmarks and works

## Testing

Due to the extensive nature of this change, testing will require a full run-through of all search-related functionality on the site. This includes creating, updating, and deleting indices for all four searched models (Work, Bookmark, Pseud, and Tag) and all searching and filtering.

## References

## Credit

Wendy Randquist - Littlelines

she/her


public/javascripts/application.js
```diff
@@ -644,17 +644,3 @@ function thermometer() {
     }
   });
 }
-
-function updateCachedTokens() {
```
redsummernight: Should this stay?

elzj: Yes, I'm wondering if this was the result of a merge conflict.

sarken: I think recent changes to the kudos function got overridden above (lines 457-468) as well, as did some recent CSS changes, which I'll comment on separately.

public/stylesheets/site/2.0/05-region-main.css
```diff
@@ -41,10 +41,6 @@ http://otwcode.github.com/docs/front_end_coding/patterns/supertypes
   background: url("/images/ao3_logos/logo-ruby.png") top right no-repeat;
 }
 
-#main.session {
```
sarken: I think this got removed accidentally -- it's a recent addition.

public/stylesheets/site/2.0/13-group-blurb.css
```diff
@@ -226,7 +226,6 @@ li.blurb:after, .blurb .blurb:after {
 }
 
 .claim .datetime {
-  margin-left: 65px;
```
sarken: This is a recent change that should be kept.

public/stylesheets/site/2.0/16-zone-system.css
```diff
@@ -212,8 +212,10 @@ SUBSECTIONS:
 
 /* 2. ZONE: SYSTEM: SESSIONS (login and signup)*/
 
-.session #signin {
```
sarken: This should stay as
```
.session #signin {
  margin-left: 48%;
  width: 48%;
}
```

public/stylesheets/site/2.0/26-media-narrow.css
```diff
@@ -90,10 +90,6 @@ body .narrow-shown {
   background-position: center;
 }
 
-#main.session {
```
sarken: It looks like some recent changes got clobbered in a merge conflict -- the stuff that's being removed here should stay.

app/models/search/async_indexer.rb
```diff
@@ -14,6 +14,9 @@ def self.perform(name)
     REDIS.del(name)
   end
 
+  def self.index(klass, ids, priority)
```
elzj: Looks like this is missing some things - I just submitted a PR to your branch to flesh that out, since I was doing some batch indexing to test things out locally.

app/models/bookmark_search.rb
```diff
@@ -77,6 +79,73 @@ def process_options
     self.options.delete_if { |key, value| value.blank? }
   end
 
+  # Useful for inspecting the differences between the old ES query and the new
+  # ES query, when debugging problems between ES versions
+  def search_query
```
elzj: I always meant to split this out, and I agree that it's useful, but even for temp code, can we avoid duplicating this? I think the only difference between this code and the query in search_results is the lack of pagination here - is there any way to add that back and then just call search_query within search_results?

app/models/work_search.rb
```diff
@@ -99,6 +106,79 @@ def process_options
     self.options.delete_if { |key, value| value.blank? }
   end
 
+  # Useful for inspecting the differences between the old ES query and the new
+  # ES query, when debugging problems between ES versions
+  def search_query
+    self.options ||= {}
```
elzj: Same comment as above re bookmark search.

app/models/work.rb
```diff
@@ -1421,25 +1434,35 @@ def to_indexed_json
...
+      bookmarkable: "bookmarkable"
     )
   end
 
   def pseud_ids
     creatorships.pluck :pseud_id
   end
 
+  def user_ids
+    Pseud.where(id: pseud_ids).pluck(:id)
```
elzj: I think this should be pluck(:user_id)

app/models/tag.rb
```diff
@@ -1171,9 +1186,11 @@ def indirect_bookmarks(rec=false)
   #################################
 
 
+  # ES UPGRADE TRANSITION #
+  # Remove mapping block
   mapping do
     indexes :id,           index: :not_analyzed
-    indexes :name,         analyzer: 'snowball', boost: 100
+    indexes :name#,         analyzer: 'snowball', boost: 100
```
elzj: Just checking that this is on purpose and not a typo

elzj:
Okay, merging this and we can submit new pull requests with further updates.

hatal175:
Currently, the new elasticsearch is not installed on the vagrant machine, so all tests fail on it.
I managed to run them but it required both changing the URL in the config and the following changes:

diff --git a/features/step_definitions/search_steps.rb b/features/step_definitions/search_steps.rb
index ec9698d52..0b1557c11 100644
--- a/features/step_definitions/search_steps.rb
+++ b/features/step_definitions/search_steps.rb
@@ -1,11 +1,13 @@
 require "cucumber/rspec/doubles"

 Given /^the (\w+) indexes are updated$/ do |klass|
-  es_update(klass)
+
   # ES UPGRADE TRANSITION #
   # Remove unless block
   unless !elasticsearch_enabled?($elasticsearch)
     tire_update(klass)
+  else
+    es_update(klass)
   end
 end

diff --git a/features/support/env.rb b/features/support/env.rb
index 2fd0223e4..e3bffcec7 100644
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -75,9 +75,11 @@ end
 After do
   # ES UPGRADE TRANSITION #
   # Change all instances of $new_elasticsearch to $elasticsearch
-  indices = $new_elasticsearch.indices.get_mapping.keys.select { |key| key.match("test") }
-  indices.each do |index|
-    $new_elasticsearch.indices.delete(index: index)
+  if !elasticsearch_enabled?($elasticsearch)
+    indices = $new_elasticsearch.indices.get_mapping.keys.select { |key| key.match("test") }
+    indices.each do |index|
+      $new_elasticsearch.indices.delete(index: index)
+    end
   end
 end

PR #3105, opened 2017-10-14
AO3-4625 Mass import API v2

ariana-paris:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4625

## Purpose

This rolls over the API to version 2 in order to introduce better named end points and clearer error messages.

## Testing

See Jira comment.


app/controllers/api/v2/base_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+            status = :empty_request
+            errors << "No items to import were provided."
+          elsif import_items.size >= ArchiveConfig.IMPORT_MAX_WORKS_BY_ARCHIVIST
+            status = :too_many_request
+            errors << "This request contains too many items to import. A maximum of #{ ArchiveConfig.IMPORT_MAX_WORKS_BY_ARCHIVIST } " +
+                      "items can be imported at one time by an archivist."
+          end
+        else
+          status = :forbidden
+          errors << "The 'archivist' field must specify the name of an Archive user with archivist privileges."
```
sarken: Could we use double quotes on the field name here? I noticed we're using double in various other messages and consistency is good whenever possible.

app/controllers/api/v2/bookmarks_controller.rb
```diff
@@ -0,0 +1,222 @@
...
+
+  # Create a bookmark for this archivist using the Bookmark model
+  def create_bookmark(archivist, params, archivist_bookmarks)
+    found_result = {}
+    bookmark_request = bookmark_request(archivist, params)
+    bookmark_status, bookmark_messages = bookmark_errors(bookmark_request)
+    bookmark_url = nil
+    original_url = nil
+    bookmarkable = nil
+    @some_errors = true
```
sarken: This is in the old code, so I assume it's fine, but could you explain why this is set to true?

app/controllers/api/v2/works_controller.rb
```diff
@@ -0,0 +1,222 @@
...
+    if original_urls.nil? || original_urls.blank? || original_urls.empty?
+      status = :empty_request
+      messages << "Please provide a list of URLs to find."
+    elsif original_urls.size >= ArchiveConfig.IMPORT_MAX_CHAPTERS
+      status = :too_many_request
+      messages << "Please provide no more than #{ ArchiveConfig.IMPORT_MAX_CHAPTERS } URLs to find."
+    else
+      status = :ok
+      results = find_existing_works(original_urls)
+      messages << "Successfully searched all provided urls"
```
sarken: URLs, capital?

app/controllers/api/v2/works_controller.rb
```diff
@@ -0,0 +1,222 @@
...
+    status, messages = batch_errors(archivist, external_works)
+
+    if status == :ok
+      # Process the works, updating the flags
+      works_responses = external_works.map { |external_work| import_work(archivist, external_work.merge(params.permit!)) }
+      success_works, error_works = works_responses.partition { |r| [:ok, :created, :found].include?(r[:status]) }
+      
+      # Send claim notification emails for successful works
+      if params[:send_claim_emails] && success_works.present?
+        notified_authors = send_external_invites(success_works, archivist)
```
sarken: I see how this works after scrolling down to the `send_external_invites` code, but it surprised me. I didn't expect a method named `send_external_invites` to also return a list of names.

app/controllers/api/v2/works_controller.rb
```diff
@@ -0,0 +1,222 @@
...
+  end
+
+  # Work-level error handling for requests that are incomplete or too large
+  def work_errors(work)
+    status = :bad_request
+    errors = []
+    urls = work[:chapter_urls]
+    if urls.nil? || urls.empty?
+      status = :empty_request
+      errors << "This work doesn't contain chapter_urls. Works can only be imported from publicly-accessible URLs."
```
sarken: Could this be "chapter URLs" like in the message below?

app/controllers/api/v2/works_controller.rb
```diff
@@ -0,0 +1,222 @@
...
+    elsif urls.length >= ArchiveConfig.IMPORT_MAX_CHAPTERS
+      status = :too_many_requests
+      errors << "This work contains too many chapter URLs. A maximum of #{ArchiveConfig.IMPORT_MAX_CHAPTERS} " \
```
sarken: This one is definitely consistent with our usual code style, but I noticed in several other places, there are spaces by the brackets, like `#{ ArchiveConfig.IMPORT_MAX_CHAPTERS }`. Could you fix those if you get a chance?

app/controllers/api/v2/works_controller.rb
```diff
@@ -0,0 +1,222 @@
...
+    external_works = params[:items] || params[:works]
+    works_responses = []
+
+    # check for top-level errors (not an archivist, no works...)
+    status, messages = batch_errors(archivist, external_works)
+
+    if status == :ok
+      # Process the works, updating the flags
+      works_responses = external_works.map { |external_work| import_work(archivist, external_work.merge(params.permit!)) }
+      success_works, error_works = works_responses.partition { |r| [:ok, :created, :found].include?(r[:status]) }
```
sarken: Just checking my understanding -- this is putting anything with the status `:ok`, `:created`, or `:found` into `success_works` and everything else into `error_works`?

app/models/story_parser.rb
```diff
@@ -319,11 +368,11 @@ def set_work_attributes(work, location = "", options = {})
       if chapter.content.length > ArchiveConfig.CONTENT_MAX
         # TODO: eventually: insert a new chapter
         chapter.content.truncate(ArchiveConfig.CONTENT_MAX, omission: "<strong>WARNING: import truncated automatically because chapter was too long! Please add a new chapter for remaining content.</strong>", separator: "</p>")
+      elsif chapter.content.length == 0
```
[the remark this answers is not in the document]
sarken: If `empty?` works here, I second Hound

app/models/story_parser.rb
```diff
@@ -319,11 +368,11 @@ def set_work_attributes(work, location = "", options = {})
...
+        raise Error, "Chapter #{chapter.position} of '#{work.title}' is empty."
```
sarken: Nitpick: Elsewhere we use double quotes around the title. Also, would "blank" make more sense to a layperson?

spec/api/v1/api_bookmarks_spec.rb
```diff
@@ -43,17 +42,17 @@
...
+           params: { archivist: archivist.login,
              bookmarks: [ bookmark ]
            }.to_json,
            headers: valid_headers
       assert_equal 200, response.status
     end
 
     it "should not create duplicate bookmarks for the same archivist and external URL" do
-      pseud_id = @user.default_pseud.id
+      pseud_id = archivist.default_pseud.id
```
sarken: The `default_pseud_id` method might work here as well.

spec/api/v1/api_works_spec.rb
```diff
@@ -477,15 +488,9 @@
...
-  before do
-    @work = FactoryGirl.create(:work, posted: true, imported_from_url: "foo")
-  end
-
-  after do
-    @work.destroy if @work
-  end
-
   describe "valid work URL request" do
+    work = FactoryGirl.create(:work, posted: true, imported_from_url: "foo")
```
sarken: This should be okay sans the `FactoryGirl.`

spec/api/v2/api_authorization_spec.rb
```diff
@@ -0,0 +1,33 @@
+require "spec_helper"
+require "webmock"
+require "api/api_helper"
+
+describe "API V2 Authorization" do
+  include ApiHelper
+  end_points = %w(/api/v2/works /api/v2/bookmarks)
+
+  describe "API POST with invalid request" do
+    it "should return 401 Unauthorized if no token is supplied and forgery protection is enabled" do
```
sarken: I won't pester you to change the old `v1` tests, but if the new tests could be rephrased without the `should`, that would be great!

spec/api/v2/api_authorization_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+    it "should return 401 Unauthorized if no token is supplied" do
+      end_points.each do |url|
+        post url
+        assert_equal 401, response.status
+      end
+    end
+
+    it "should return 403 Forbidden if the specified user isn't an archivist" do
+      end_points.each do |url|
+        post url, params: { archivist: "mr_nobody" }.to_json, headers: valid_headers
```
sarken: Hee @ the name

spec/api/v2/api_bookmarks_spec.rb
```diff
@@ -0,0 +1,173 @@
+require "spec_helper"
+require "api/api_helper"
+
+describe "API V2 BookmarksController" do
+  include ApiHelper
+
+  bookmark = { id: "123",
+               url: "http://foo.com",
```
sarken: Should we use `example.com`? I think that's our preferred sample domain for tests.

app/controllers/api/v2/bookmarks_controller.rb
ariana-paris: I suspect I did this because there's only one code path that indicates success (the one where `bookmark.save` succeeds), so it's simpler to only flip it in that one success case and leave it set to "yep, it didn't work" in all other cases.

spec/api/v1/api_works_spec.rb
sarken: Oh! Just thought of this: we have a [posted_work factory](https://github.com/otwcode/otwarchive/blob/bd4f323003a7f63d0e708ff334fd666e4760601a/factories/works.rb#L26) you could use here and on line 13 of spec/api/v2/api_works_search_spec.rb

sarken:
This looks good, but something is causing the specs to repeatedly and consistently time out on Travis. It seemed like it might be hanging on a specific test, so I ran the api, controllers, and models folders locally and they completed just fine (a few errors on caching tests, including `rspec ./spec/models/work_spec.rb:350` which is import-related, but these are normal when running the tests locally), so either it's another test or a case of finally having too many specs for one folder.

ariana-paris:
The problem with the specs timing out is going to be because of all the new v2 API tests. The problem will probably go away if I remove the v1 tests, but I wasn't planning to do that until I remove the v1 code. And anyway, I'm afraid it's a sign we're going to have to split up the specs the way we have split up the features.

PR #3113, opened 2017-10-22
AO3-5208: Fix filtering behavior

elzj:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5208

## Purpose

Restores previous filter functionality: filters should become more specific to match the results as you drill down. They should also include collections on the appropriate pages. And bookmark filters should not include every tag on the site.

## Testing

Filtering should work the way it does on production.


tickinginstant:
My concern with this is that as soon as you, say, exclude a particularly popular pairing using the checkboxes, the frequency of that pairing in your search results will become 0, and so if I'm not mistaken it'll drop off the list of the most popular tags in your search. This means that it won't show up in the checkboxes on the right, so if you then try to exclude a second pairing and press "Sort and Filter" again, it'll revert back to showing works/bookmarks with the first pairing.

My suggestion would be to have some extra code on the filters form to display tags that have already been excluded (with no counts, since they'd all be 0).

elzj:
@tickinginstant Good point, I'll change that.

tickinginstant:
This appears to work perfectly for the work filters, but I'm getting a bit of a glitch with excluding bookmarks. The excluded tag shows up twice: once with its original count, and once with a count of (0) afterwards. And in general, the counts don't seem to decrease as I exclude tags.

app/controllers/bookmarks_controller.rb
```diff
@@ -102,15 +102,24 @@ def index
...
-            @filtering_facets = BookmarkSearchForm.new(base_options.merge({parent: @owner})).search_results.facets
           else
             @search = BookmarkSearch.new(options.merge(faceted: true, bookmarks_parent: @owner))
-            @filtering_facets = @search.search_results.facets
           end
           @bookmarks = @search.search_results
           @facets = @bookmarks.facets
+          if @search.options[:excluded_tag_ids].present?
+            tags = Tag.where(id: @search.options[:excluded_tag_ids])
+            excluded_bookmark_tag_ids = params.dig(:exclude_bookmark_search, :tag_ids) || []
```
tickinginstant: Bit of an aside, but this line makes me think that the code above for packing all the information from :exclude_bookmark_search into :excluded_tag_ids is kind of overzealous (and I think it makes the include/exclude somewhat asymmetric). Still, I think that's probably outside the scope of this pull request. Maybe something to refactor later.

app/models/search/bookmark_query.rb
```diff
@@ -224,7 +247,12 @@ def collections_filter
 
   def tag_exclusion_filter
     if exclusion_ids.present?
-      exclusion_ids.flatten.map { |exclusion_id| term_filter(:filter_ids, exclusion_id) }
+      exclusion_ids.flatten.map { |exclusion_id|
```
tickinginstant: There's a lot of flattens here. Maybe just use a single flat_map and rely on the fact that exclusion_ids will always return a list of IDs?

app/models/search/taggable_query.rb
```diff
@@ -0,0 +1,39 @@
...
+    @filter_ids += named_tags
+    @filter_ids.uniq
+  end
+
+  def exclusion_ids
+    return @exclusion_ids if @exclusion_ids.present?
+    return if options[:excluded_tag_names].blank? && options[:excluded_tag_ids].blank?
+
+    ids = options[:excluded_tag_ids] || []
+    names = options[:excluded_tag_names]&.split(",")
```
tickinginstant: I'd be tempted to add a .map(&:squish) here (and in named_tags) just in case someone has javascript disabled and is manually typing in tag names. But again, that might be outside the scope of the pull request, particularly since that's not the way other_tag_names is handled on the live website.

lib/taggable.rb
```diff
@@ -272,7 +272,7 @@ def tag
 
   # Index all the filters for pulling works
   def filter_ids
-    filters.pluck :id
+    (tags.pluck(:id) + filters.pluck(:id)).uniq
```
tickinginstant: This slightly alters the way that searching for non-canonical tags works (it used to just fall back to searching the :tag field for matching strings), but I actually much prefer this way of handling it.

PR #3115, opened 2017-10-22
AO3-5194: Update schema and structure files

hatal175:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5194

## Purpose

Updates schema.rb and structure.sql to current DB schema and structure.

## Testing

Setup testing environment as usual. No site changes should occur.
Also in schema.rb the format changed from 'key'=> value to 'key':value. I'm not entirely sure if this affects any ruby language version compatibility.

## Credit

Tal Hayon

Please use he.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -1,4 +1,3 @@
-# encoding: UTF-8
 # This file is auto-generated from the current state of the database. Instead
 # of editing this file, please use the migrations feature of Active Record to
 # incrementally modify your database, and then regenerate this schema definition.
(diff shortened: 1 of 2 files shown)


hatal175:
To be honest, I'm not entirely sure this is still the most updated scheme.

sarken:
No, it'll definitely need to be updated before it can be merged -- we've since added some things to help us deal with spam, plus a column on the feedbacks table to help handle Elasticsearch support tickets -- but the link will still be handy when the time comes.

elzj:
I was setting up a new database, so I dumped the latest versions here: https://github.com/otwcode/otwarchive/pull/3643

PR #3117, opened 2017-10-23
AO3-5209 Don't change the appearance of non-work & bookmark filters

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5209

## Purpose

Restores the old HTML, JavaScript, and CSS for the Elasticsearch 0.90 filters. Reworks the CSS for the new filters so it doesn't negatively affect the styling of the filters on the Collections index.

## Testing

Compare the work and bookmark filters when using Elasticsearch 0.90 on staging to those on production, and compare the collections filters to those on production.


public/stylesheets/site/2.0/18-zone-searchbrowse.css
```diff
@@ -162,4 +152,66 @@ form.filters {
...
 }
 
+.filters .sort + .heading {
+  margin-top: 1.286em; /* heading top margin minus sort bottom margin */
+}
+
+/* INTERACTION: OLD FILTERS */
+/* ES UPGRADE TRANSITION: REMOVE WHEN DONE */
+
+form.old-filters {
```
sarken: `old-filters` 100% violates our class naming conventions, but this is definitely temporary, so I figure we can live with it.

public/javascripts/filters_old.min.js
```diff
@@ -0,0 +1 @@
+function showFilters(){var e=$j("dd.tags");e.each(function(e,r){var t=$j(r).find("input"),i=$j(r),n=$j("."+i.attr("id")+"_open"),s=$j("."+i.attr("id")+"_close");t.each(function(e,t){$j(t).is(":checked")&&($j(r).show(),$j(n).hide(),$j(s).show())})})}function setupNarrowScreenFilters(){var e=$j("form.old-filters"),r=$j("#outer"),t=$j("#go_to_filters"),i=$j("#leave_filters");t.click(function(t){t.preventDefault(),e.removeClass("narrow-hidden"),r.addClass("filtering"),e.find(":focusable").first().focus(),e.trap()}),i.click(function(i){i.preventDefault(),r.removeClass("filtering"),e.addClass("narrow-hidden"),t.focus()})}$j(document).ready(function(){showFilters(),setupNarrowScreenFilters()});
```
sarken: I'm only restoring the minified version of the old filter JavaScript because I couldn't think of a good reason to restore the non-minified version.

PR #3122, opened 2017-10-26
AO3-5222 handle indexing failures and ghost records

WendyBeth:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5222

## Purpose

What does this PR do?

+ Adds retry for individual failed documents that are returned by Elasticsearch batch indexing
+ Catches documents that fail more than once and stores them in memory for review - this is an idea about how to catch documents that are never going to succeed
+ Adds hook to QueryResult class that deletes indexes belonging to records that no longer exist

## Testing

+ I am not sure how to mimic a failure from Elasticsearch on our end. The best I've been able to do is get errors from undoing the fixes made to Series, Work, or External Work's `#bookmarkable_json` methods. If you add `id: "#{id}-series"` back to the attributes merged into the bookmarkable_json in Series, for example, this will make reindexing in bulk Series using BookmarkedSeriesIndexer return document failures. Using this method you can test that the queue BookmarkedSeriesIndexer:failures is created. Because the failure this creates will not be fixed with retries, you can also inspect REDIS_GENERAL for the keys BookmarkedSeriesIndexer:first_failures_store, BookmarkedSeriesIndexer:second_failures_store, and BookmarkedSeriesIndexer:permanent_failures_store. The Series IDs and error messages about their failures should eventually all end up in the permanent_failures_store. You could also disrupt your connection to Elasticsearch and get mostly the same effect.

+ For the QueryResult hook, you should be able to create a work, index all works, and then delete the work and search for it without it showing up in any results. All hints of it should be gone from the search indexes as well.

## References

## Credit

Wendy Randquist - Littlelines

she/her


app/models/search/index_sweeper.rb
```diff
@@ -0,0 +1,99 @@
...
+
+    return if permanent_store.include?(document_stamp)
+
+    if first_store.include?(document_stamp)
+      second_store << document_stamp
+      first_store.delete(document_stamp)
+      set_store "second", second_store
+      set_store "first", first_store
+      @rerun_ids << document["_id"]
+    elsif second_store .include?(document_stamp)
```
elzj: Think there's an extra space before the period here

app/models/search/index_sweeper.rb
```diff
@@ -0,0 +1,99 @@
...
+    end
+
+    return if permanent_store.include?(document_stamp)
+
+    if first_store.include?(document_stamp)
+      second_store << document_stamp
+      first_store.delete(document_stamp)
+      set_store "second", second_store
+      set_store "first", first_store
+      @rerun_ids << document["_id"]
```
tickinginstant: This looks like it would fail when used with the StatCounterIndexer. The IDs passed in as the arguments to the indexer aren't necessarily the same as the _id passed to ElasticSearch.

app/models/search/query_result.rb
```diff
@@ -20,6 +20,7 @@ def items
     if @items.nil?
       ids = hits.map { |item| item['_id'] }
       items = klass.where(:id => ids).group_by(&:id)
+      IndexSweeper.async_cleanup(klass, ids, items.values.flatten.pluck(:id))
```
elzj: You should just be able to use "items.keys" here, since the keys are the ids.

app/models/search/index_sweeper.rb
elzj: Yeah, this needs to be addressed.

elzj:
Hey Wendy - did you get a chance to look at tickinginstant's comment about the StatCounterIndexer? Also, could you try merging master in again? Hopefully that might help clean up the diff here.

app/models/search/index_sweeper.rb
WendyBeth: I am looking into this now.

WendyBeth: Hi @tickinginstant -

I've looked through this, and I don't think I understand what you're getting at. Can you explain? It looks to me like anywhere the sweeper is called, it is specifically using the Elasticsearch "_id"s and never references anything else. I see that the StatCounterIndexer uses the StatCounter's work_id as it's Elasticsearch id - is this the basis of the problem that you're referring to?

tickinginstant: Yeah, that's the basis of the issue. The IndexSweeper reruns the indexer on this line:
```
AsyncIndexer.new(@indexer, "failures").enqueue_ids(@rerun_ids) unless @rerun_ids.empty?
```
When `@indexer` is the StatCounterIndexer, it'll pass in the `work_id`s for those StatCounters, because those are the `_id`s that will be returned from the ElasticSearch bulk update and stored in `@rerun_ids`. But the StatCounterIndexer is expecting to receive StatCounter IDs, not Work IDs (since it needs to integrate with the existing IndexQueue code). So in case of a failure, I believe the IndexSweeper will rerun the StatCounterIndexer with a completely different set of StatCounters than the failure occurred with (if those StatCounters even exist).

WendyBeth: Could this be solved by using the stat_counter id in the routing info for the StatCounterIndexer (the comment doesn't explain why the work_id is being used instead of the stat counter id)? Or, as in cases with BookmarkedSeries/BookmarkedExternalWorks/BookmarkedWorks, the id could be something like "#{stat_counter.id}-stat-counter", which would be read as the stat counter's id by the where clause on line 19 but would still differentiate the stat counters in the work index from the works.

But since I'm not entirely certain why it was done that way in the first place, I'm not sure what kind of problems that might cause.

tickinginstant: The StatCounterIndexer has to use the work_id in its routing info because it's not indexing StatCounters as a separate item. Instead, it's using the info from StatCounters to update works in the work index. So there's no way that it can be changed to use the StatCounter's ID.

The only way to use the StatCounter's ID is if you want to change the WorkIndexer so that it no longer includes stats, but instead generates a separate stats object alongside the work that gets joined with a has_parent/has_child query. But that's a more complex task, and I'm not sure if it's really a good idea, given that the ES docs kind of frown on has_parent/has_child queries except in specific circumstances:

> The join field shouldn’t be used like joins in a relation database. In Elasticsearch the key to good performance is to de-normalize your data into documents. Each join field, has_child or has_parent query adds a significant tax to your query performance.
>
> The only case where the join field makes sense is if your data contains a one-to-many relationship where one entity significantly outnumbers the other entity.

I could potentially see the additional cost being worth it if the need for joint queries was very rare, but I suspect that sorting works by kudos/comments/hits is one of the more common queries.

elzj: Yes, the parent/child relationship is only barely worth it for bookmarks, and there's not much point to the stats indexing if you can't sort on the fields.

How much of a pain would it be to change the ids used in the StatCounterIndexer to work ids and to find the stat counter objects based on that? There's only one counter per work, so one should be as good as the other in theory.

tickinginstant: As I mentioned earlier, the problem with that is compatibility with IndexQueue and IndexSubqueue. If you want to have both indexers running simultaneously, you'd need to add some special-case code to IndexSubqueue to look up StatCounters based on work IDs, as well.

I feel like it'd be simpler to just have an extra class function in the Indexers, called something like `ids_from_elasticsearch`, which the IndexSweeper can pass the `@rerun_ids` through before passing them along to the AsyncIndexer. That would also have the bonus of being more flexible, so that you're not locked into always having to make the IDs match up.

elzj: That sounds like a pretty good plan to me. @WendyBeth is this something you'd have time to look at this week, or should we make a separate issue for it?

WendyBeth: That sounds doable fairly quickly, and it's clear now why the StatCounterIndexer was made that way. Thanks for the information. I'll look into getting this cleared up tomorrow morning.

PR #3124, opened 2017-10-29
AO3-5113 Change password confirmation error message

hatal175:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5113

## Purpose

This Changes the password confirmation error to requested string.

## Testing

Feature testing should be enough. If desired however, one just needs to input a different password in the confirmation field and see that the error message is "Password confirmation doesn't match new password" instead of "Password confirmation doesn't match confirmation".

## Credit

Tal Hayon

Please use he.


redsummernight:
This pull has conflicts now, and it needs to be adapted for Devise.

app/models/user.rb
```diff
@@ -172,6 +172,10 @@ def unread_inbox_comments_count
                       too_long: ts("is too long (maximum is %{max_pwd} characters)",
                                    max_pwd: ArchiveConfig.PASSWORD_LENGTH_MAX)
 
+  validates_confirmation_of :password,
```
redsummernight: Devise sets up password confirmation:

https://github.com/plataformatec/devise/blob/v4.6.1/lib/devise/models/validatable.rb#L41

which uses Rails' [`validates_confirmation_of`](https://api.rubyonrails.org/v5.1.3/classes/ActiveModel/Validations/HelperMethods.html#method-i-validates_confirmation_of), which has the default message set at:

https://github.com/otwcode/otwarchive/blob/c91671d5b586348f6c02b51566503ce71a464592/config/locales/en.yml#L15

It's simpler to modify the locale file, but that assumes we will never use the validator for anything but password fields in the future.

I went with overriding the validator in the model, but I'm on the fence about this.

sarken: I'm so confused why the error message says, "Password confirmation doesn't match confirmation." If the validation is on the `password` field, shouldn't it be saying `Password doesn't match confirmation`? 

If we used `validates_confirmation_of` on, say, an `email` field, would it do the same thing, e.g. `Email confirmation doesn't match confirmation`? 

It looks like the usual error message is `“doesn't match %{translated_attribute_name}”`, which at least makes sense (`Password confirmation doesn't match password` or `Email confirmation doesn't match email`). That seems like a fine error message... why is what's in our locale file different in the first place? :\

redsummernight: If I remove the locale file override and the validator override, going with framework defaults, the message is "Password confirmation doesn't match Password". It has weird capitalization.

sarken: You're killing me, Rails. Killing me.

I feel like that's _probably_ the best thing to do, code-wise, but ideally we could find a way to fix the capitalization.

sarken: Wow, this is ridiculous. I have the fix and will push.

config/locales/en.yml
```diff
@@ -55,6 +55,10 @@ en:
           attributes:
             pseud_id:
               taken: "is already listed as a creator."
+        user:
+          attributes:
+            password_confirmation:
```
sarken: Everything on the face of the planet, including simple logic, says this should be `password`. And yet.

redsummernight: This is much better, thanks!

sarken:
Updated post #3680!

PR #2163, opened 2015-08-27
AO3-4331 Mark works as spam or ham

shalott:
For JIRA issue AO3-4331 https://otwarchive.atlassian.net/browse/AO3-4331

Adds button to works in admin view to mark them as spam or ham, so Akismet can learn. After this is merged, AO3-4330 will also report works to Akismet as spam when they are batch-deleted.

PR #3128, opened 2017-10-30
AO3-4331 Mark works as spam or ham

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4331

## Purpose

Adds button to works in admin view to mark them as spam or ham, so Akismet can learn. After this is merged, [AO3-4330](https://otwarchive.atlassian.net/browse/AO3-4330) will also report works to Akismet as spam when they are batch-deleted.

## Testing

Refer to JIRA

## References

Update of #2163 

## Credit

This is primarily Naomi's code and she should be credited.

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -1372,15 +1372,10 @@ class Work < ApplicationRecord
   # SPAM CHECKING
   ########################################################################
 
-  def spam_checked?
-    spam_checked_at.present?
-  end
-
-  def check_for_spam
-    return unless %w(staging production).include?(Rails.env)
+  def akismet_attributes
     content = chapters_in_order.map { |c| c.content }.join
     user = users.first
-    self.spam = Akismetor.spam?(
+    {
       comment_type: 'fanwork-post',
       key: ArchiveConfig.AKISMET_KEY,
       blog: ArchiveConfig.AKISMET_NAME,
@@ -1390,7 +1385,16 @@ class Work < ApplicationRecord
       comment_author: user.login,
       comment_author_email: user.email,
       comment_content: content
-    )
+    }
+  end
+
+  def spam_checked?
+    spam_checked_at.present?
+  end
+
+  def check_for_spam
+    return unless %w(staging production).include?(Rails.env)
+    self.spam = Akismetor.spam?(akismet_attributes)
     self.spam_checked_at = Time.now
     save
   end
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -1381,7 +1381,7 @@ class Work < ApplicationRecord
     content = chapters_in_order.map { |c| c.content }.join
     user = users.first
     self.spam = Akismetor.spam?(
-      comment_type: 'Fan Fiction',
+      comment_type: 'fanwork-post',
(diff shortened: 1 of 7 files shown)


app/models/work.rb
```diff
@@ -1395,6 +1395,18 @@ def check_for_spam
     save
   end
 
+  def mark_as_spam!
+    update_attribute(:spam, true)
+    # don't submit spam reports unless in production mode
+    Rails.env.production? && Akismetor.submit_spam(akismet_attributes)
```
tickinginstant: Is akismet_attributes defined for works? I don't see it anywhere in the file.

sarken:
I've defined `akismet_attributes` for works, but the one thing I am slightly worried about is we don't send a user agent info to Akismet. It's required info (https://akismet.com/development/api/#comment-check) but we don't send it elsewhere, either. I have no idea how much this affects the accuracy of things.

tickinginstant:
It looks like it's sending a user agent both for comments and for feedback, which I think are the only other two places where Akismetor is being used. (Although the feedback sets its IP address and user agent in a very insecure way. Might need a bug report for that.)

I did a couple of experiments, and it looks like the user_agent field isn't as required as the docs say it is, since checking for spam won't return an error if the user_agent field is missing (though it will if the user_ip field is missing). But the Akismetor gem is fairly simple and doesn't actually do any error checking, so it wouldn't tell you even if there was an error. Maybe it'd be worth trying to find an akismet-related gem with better error handling? Though that's probably outside the scope of this pull request.

sarken:
...I read the code for Support tickets and comments multiple times and still somehow managed not to see `user_agent` in there. Ugh. But anyway, we already run works through Akismet and it does catch some spam, so it's definitely not something that causes an error and prevents it from working completely. I just wonder if it would be catching more if we were able to provide that.

The Akismetor gem hasn't been updated in nine years, so replacing it with a gem that _is_ actively maintained is something we'd like to do (I think the one Akismet recommends is called rakismet or something similar), but because Akismetor is currently working and we're in a bit of a hurry, that's definitely a "let's do it down the line" thing.

config/routes.rb
```diff
@@ -148,6 +148,7 @@
     resources :user_creations, only: [:destroy] do
       member do
         get :hide
+        get :set_spam
```
elzj: Can we make this a put request? If it's updating data, it really shouldn't be 'get'.

app/views/admin/_admin_options.html.erb
```diff
@@ -31,10 +31,33 @@
...
+        <%= link_to ts("Edit Tags and Language"), edit_tags_work_path(@work) %>
+      </li>
+      <% if @work.spam? %>
+        <li>
+          <%= link_to ts("Mark Not Spam"),
+                      { controller: "admin/user_creations",
+                        action: "set_spam",
+                        id: @work.id,
+                        creation_type: @work.class,
+                        spam: "false" } %>
```
tickinginstant: Since set_spam is now a put request, I think you need to add `method: :put` (not as part of the link hash, but after) to make these links work.

sarken:
Closing while I start over from the correct commit on master.

PR #3130, opened 2017-10-30
AO3-5205: Prevent sign ups with duplicate emails in the invite queue

elzj:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5205

## Purpose

Prevents users from requesting invites multiple times for the same email address. Also fixes the validator to require a full domain.

## Testing

See issue.

### DIFF
--- a/app/models/invite_request.rb
+++ b/app/models/invite_request.rb
@@ -20,7 +20,7 @@ class InviteRequest < ApplicationRecord
   def set_simplified_email
     return if email.blank?
     simplified = AdminBlacklistedEmail.canonical_email(email).split('@')
-    self.simplified_email = simplified.first.gsub(".", "") + "@#{simplified.last}"
+    self.simplified_email = simplified.first.delete(".").gsub(/\+.+$/, "") + "@#{simplified.last}"
   end
 
   # Doing this with a method so the error message makes more sense
--- a/app/models/invite_request.rb
+++ b/app/models/invite_request.rb
@@ -3,7 +3,9 @@ class InviteRequest < ApplicationRecord
   acts_as_list
   validates :email, presence: true, email_veracity: true
   validates_uniqueness_of :email, message: "is already part of our queue."
+  validate :simplified_email_uniqueness
   before_validation :compare_with_users, on: :create
+  before_validation :set_simplified_email, on: :save
 
   # Realign positions if they're incorrect
   def self.reset_order
@@ -11,6 +13,20 @@ class InviteRequest < ApplicationRecord
     unless first_request && first_request.position == 1
       requests = order(:position)
       requests.each_with_index {|request, index| request.update_attribute(:position, index + 1)}
+    end
+  end
+
+  # Borrow the blacklist cleaner but just strip out all the periods for all domains
+  def set_simplified_email
+    return if email.blank?
+    simplified = AdminBlacklistedEmail.canonical_email(email).split('@')
+    self.simplified_email = simplified.first.gsub(".", "") + "@#{simplified.last}"
+  end
+
+  # Doing this with a method so the error message makes more sense
+  def simplified_email_uniqueness
+    if InviteRequest.where(simplified_email: simplified_email).exists?
+      errors.add(:email, "is already part of our queue.")
     end
   end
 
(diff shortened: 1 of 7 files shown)


app/models/invite_request.rb
```diff
@@ -14,6 +16,20 @@ def self.reset_order
     end
   end
 
+  # Borrow the blacklist cleaner but just strip out all the periods for all domains
+  def set_simplified_email
+    return if email.blank?
+    simplified = AdminBlacklistedEmail.canonical_email(email).split('@')
+    self.simplified_email = simplified.first.gsub(".", "") + "@#{simplified.last}"
```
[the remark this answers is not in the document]
zz9pzza: We only want to do this for gmail not for anything else , jam.es_@catbus.co.uk is different to james_@catbus.co.uk

( I believe it is gmail.com and googlemail.com )

app/validators/email_veracity_validator.rb
```diff
@@ -5,23 +5,13 @@ class EmailVeracityValidator < ActiveModel::EachValidator
...
-        # treetop must respond to domain
-        # We exclude valid email values like <user@localhost.com>
-        # Hence we use m.__send__(tree).domain
-        # treetop = mail.__send__(:tree)
-
-        # A valid domain must have dot_atom_text elements size > 1
-        
-        # treetop was deprecated out of the mail gem
-        result #&&= (treetop.domain.dot_atom_text.elements.size > 1)
+        result = mail.domain && mail.domain.match('\.') && mail.address == value
```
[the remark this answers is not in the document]
zz9pzza: I think hound is right.

app/models/invite_request.rb
zz9pzza: I would also do the + thing for all domains so people can't ask for two invites at the same address in the queue at one time. They can be patient and wait

elzj: Under the circumstances, I'd rather err on the side of caution here. We're saving this as a different field so the original email is preserved and will be used to send the invite to. I can be more aggressive about stripping out the +s if you want.

spec/models/invite_request_spec.rb
```diff
@@ -1,24 +1,39 @@
...
+      end
+
+      it "should not let you sign up again with the same email" do
+        dup_request = build(:invite_request, email: @original_request.email)
+        expect(dup_request.valid?).to be_falsey
+        expect(dup_request.errors[:email]).to include("is already part of our queue.")
+      end
+
+      it "should not allow duplicates with periods and plus signs" do
+        dup_request = build(:invite_request, email: "the.god.thor+marvel@gmail.com")
```
zz9pzza: I would split the two tests out as it is not clear which bit of the code is failing it. Having two tests would be better

app/models/invite_request.rb
```diff
@@ -14,6 +16,20 @@ def self.reset_order
...
+  # Borrow the blacklist cleaner but just strip out all the periods for all domains
+  def set_simplified_email
+    return if email.blank?
+    simplified = AdminBlacklistedEmail.canonical_email(email).split('@')
+    self.simplified_email = simplified.first.delete(".").gsub(/\+.+$/, "") + "@#{simplified.last}"
+  end
+
+  # Doing this with a method so the error message makes more sense
+  def simplified_email_uniqueness
+    if InviteRequest.where(simplified_email: simplified_email).exists?
```
tickinginstant: Having the check be implemented like this means that all InviteRequests will become invalid immediately after being saved to database, which I'm pretty sure will break InviteRequest.reset_order. I'd suggest either adding `on: :create` to the validate statement above, or adding `.where.not(id: self.id)` here.

elzj: Thanks! Switched from a regular uniqueness validation and didn't notice that.

factories/invitations.rb
```diff
@@ -3,7 +3,9 @@
 FactoryGirl.define do
 
   factory :invite_request do
-    email
+    sequence :email do |n|
+      Faker::Internet.email(name="#{Faker::Name.first_name}_#{n}")
```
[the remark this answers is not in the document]
ariana-paris: Should that be `name: #{Faker::Name...`?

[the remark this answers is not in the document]
elzj: Nah, I think it's right that name doesn't need to be there at all, but that's how we had it elsewhere.

PR #3131, opened 2017-10-31
AO3-4331 Mark works as spam or ham

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4331

## Purpose

Adds button to works in admin view to mark them as spam or ham, so Akismet can learn. After this is merged, [AO3-4330](https://otwarchive.atlassian.net/browse/AO3-4330) will also report works to Akismet as spam when they are batch-deleted.

Starts from bfb847d896a8df85e30ab314d92056969f2030fa rather than current master, because we want to do this before Elasticsearch. 

## Testing

Refer to JIRA

## References

Update of #2163. Also refer to #3128.

## Credit

This is primarily Naomi's code and she should be credited.


app/controllers/admin/user_creations_controller.rb
```diff
@@ -1,52 +1,75 @@
 class Admin::UserCreationsController < ApplicationController
   
   before_action :admin_only
+  before_action :get_creation
+  before_action :can_be_marked_as_spam, only: [:set_spam]
+
+  def get_creation
+    raise "Redshirt: Attempted to constantize invalid class initialize hide #{params[:creation_type]}" unless %w(ExternalWork Bookmark Work).include?(params[:creation_type])
```
ariana-paris: Curiosity more than anything: why exactly do we have "initialise hide #" in that message? (I remember the "Redshirt" was so we could find these messages in our logs)

app/views/admin/_admin_options.html.erb
```diff
@@ -10,15 +10,16 @@
...
     <% if item.hidden_by_admin? %>
       <li>
         <%= link_to ts("Make %{item} Visible", item: item_type_name),
                     { controller: "admin/user_creations",
                       action: "hide",
                       id: item,
                       creation_type: item.class,
-                      hidden: 'false' } %>
+                      hidden: "false" },
+                    method: :put %>
```
ariana-paris: Philosophically, I'd say this was a POST (i.e. generic request to change something) rather than a PUT (usually a specific request to add a new or updated item to a collection), but I assume there was a reason for this change. (This obviously applies to the addition to `routes.rb` too)

app/controllers/admin/user_creations_controller.rb
sarken: Ugh, copy and paste error. The Redshirt bit was repeated in several different actions, but this pull request moves it to a separate method... I evidently copy and pasted the Redshirt error from the hide action. Will remove!

app/views/admin/_admin_options.html.erb
sarken: Just noting for any curious parties that we decided this was a level of philosophical debate we didn't want to get into on a Tuesday morning, so we're sticking with PUT now.

PR #3133, opened 2017-11-01
AO3-5024 Reword error message when failing to log in as admin

hatal175:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5024

## Purpose

This adds an error message on admin login failure.

## Testing

cucumber tests. I've changed the relevant test.

In order to do this, I had to override a method in Devise SessionsController and change flash[:alert] to flash[:error]. This is less invasive to the otw codebase, but perhaps not the best way. I'd appreciate a code review on this.

## Credit

Tal Hayon

Please use he.


sarken:
We're going to be adding `flash[:alert]` anyway when we move user authentication to Devise (#3378), so I think it would be fine here.

ariana-paris:
This PR has merge conflicts due to the recent upgrade the Devise, but once those are resolved, it should be good to go

sarken:
The existing "Invalid Login or password" error message works now because we've implemented `flash[:alert]`. 

We can use the new text, but we don't need or want to override the method to use `flash[:error]`, as I mentioned in https://github.com/otwcode/otwarchive/pull/3133#issuecomment-417477871

PR #3134, opened 2017-11-01
AO3-4222 Improve homepage and invite requests page when queue is off

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4222

## Purpose

* Moves the form for checking your invite request status off the invite request page and onto its own page
* Makes the invite request page more informative when the queue is off
* Adjusts the text displayed on the homepage when the invite queue is off

## Testing

Refer to JIRA


app/views/invite_requests/new.html.erb
```diff
@@ -1,27 +0,0 @@
-<!--Descriptive page name, messages and instructions-->
-<h2 class="heading">New invite_request</h2>
```
redsummernight: Would you want to disable the route `new_invite_request` here, or leave that for [AO3-4972](https://otwarchive.atlassian.net/browse/AO3-4972)?

app/views/invite_requests/status.html.erb
```diff
@@ -0,0 +1,26 @@
+<!--Descriptive page name, messages and instructions-->
+<h2 class="heading">
+  <%= ts("Invitation Request Status") %>
+</h2>
+
+<p>
+  <%= ts("There are currently %{count} people on the waiting list.", count: InviteRequest.count) %>
+  <% if @admin_settings.invite_from_queue_enabled? %>
+    <%= ts("We are sending out %{invites} invitations per day.", invites: AdminSetting.invite_from_queue_number) %>
```
redsummernight: Should we use `@admin_settings` for caching, as in [AO3-4470](https://otwarchive.atlassian.net/browse/AO3-4470)?

app/views/invite_requests/new.html.erb
sarken: Oh, I didn't even realize that issue existed! I'll just remove this change and let whoever takes that issue deal with it.

redsummernight: I was kinda hoping you'd make it a double-issue pull request. :P

sarken: If only I were slightly less sleep-deprived, I would be game! Maybe I'll make a separate PR tomorrow.

PR #3135, opened 2017-11-01
AO3-3848 Added Feature test for a locked user account

hatal175:
## Issue

https://otwarchive.atlassian.net/browse/AO3-3848

## Purpose

As I could not reproduce the issue, I added a feature test for it.

## Testing

It is a feature test.

## Credit

Tal Hayon

Please use he.

### DIFF
--- a/features/step_definitions/user_steps.rb
+++ b/features/step_definitions/user_steps.rb
@@ -158,13 +158,13 @@ When /^"([^\"]*)" creates the default pseud "([^"]*)"$/ do |username, newpseud|
   click_button "Create"
 end
 
-When /^I login username "([^"]*)" and password "([^"]*)" for ([0-9]*) times$/ do |login, password, times|
-  for i in 1..times.to_i do
+When /^I login username "([^"]*)" and password "([^"]*)" for ([0-9]*) times$/ do |login, password, n|
+  n.to_i.times {
     step(%{I am on the home page"})
     step(%{I fill in "User name" with "#{login}"})
     step(%{I fill in "Password" with "#{password}"})
     step(%{I press "Log In"})
-  end
+  }
 end
 
 When /^I fill in "([^"]*)"'s temporary password$/ do |login|
--- a/features/step_definitions/user_steps.rb
+++ b/features/step_definitions/user_steps.rb
@@ -159,12 +159,12 @@ When /^"([^\"]*)" creates the default pseud "([^"]*)"$/ do |username, newpseud|
 end
 
 When /^I login username "([^"]*)" and password "([^"]*)" for ([0-9]*) times$/ do |login, password, n|
-  n.to_i.times {
+  n.to_i.times do
     step(%{I am on the home page"})
     step(%{I fill in "User name" with "#{login}"})
     step(%{I fill in "Password" with "#{password}"})
     step(%{I press "Log In"})
-  }
+  end
 end
 
 When /^I fill in "([^"]*)"'s temporary password$/ do |login|
--- a/features/step_definitions/user_steps.rb
+++ b/features/step_definitions/user_steps.rb
@@ -158,15 +158,6 @@ When /^"([^\"]*)" creates the default pseud "([^"]*)"$/ do |username, newpseud|
   click_button "Create"
 end
 
-When /^I login username "([^"]*)" and password "([^"]*)" for ([0-9]*) times$/ do |login, password, n|
-  n.to_i.times do
-    step(%{I am on the home page"})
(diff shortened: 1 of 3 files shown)


hatal175:
It could be better if I could reduce the number of attempts for the test,
but I'm not sure how to do that.

sarken:
That's a good point. 

It would be easier to do via Cucumber if this were a setting in our config file, which it really should be, and I'd normally suggest making that change. However, we're switching our user authentication to Devise soon and won't have time to merge a code change to Authlogic before then, so it would be somewhat pointless to ask you to do that.

Instead, it might be better to rewrite this as an Rspec test and set the user's `failed_login_count` to `49`, so you only need to make one log in attempt to trigger the error. That way, we're still only changing the tests and can merge this sooner rather than later.

hatal175:
Then I'll leave this for after the auth change.

sarken:
If you could change to RSpec now rather than waiting to make the config change, that would actually be really helpful -- it would help us make sure we don't forget to implement account-locking when making the switch.

hatal175:
Ok then

features/step_definitions/user_steps.rb
```diff
@@ -158,6 +158,15 @@
   click_button "Create"
 end
 
+When /^I login username "([^"]*)" and password "([^"]*)" for ([0-9]*) times$/ do |login, password, n|
```
redsummernight: Can you undo changes to this file since the cucumber step is no longer needed?

spec/controllers/user_sessions_controller_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+require 'spec_helper'
+
+describe UserSessionsController do
+  let(:user) { create(:user, activated_at: Time.new(2007, 1, 1)) }
+
+  context "create" do
+    context "after failed to login more than consecutive login limit" do
+      before do
+        allow(UserSession).to receive(:consecutive_failed_logins_limit).and_return(2)
+        3.times do
```
redsummernight: Since you mocked `consecutive_failed_logins_limit` as 2, `2.times` here is enough for the next post to fail. Can you make "2" a local variable?

spec/controllers/user_sessions_controller_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+          post :create, params: { user_session: { login: user.login, password: "aaa", remember_me: true } }
```
redsummernight: You can leave out `remember_me: true` from all the posts, it's not needed for the test.

spec/controllers/user_sessions_controller_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+        end
+      end
+
+      context "when trying to login" do
```
redsummernight: You don't need this nested context since there's only one test, I'd move the last `post` into the `it` block.

spec/controllers/user_sessions_controller_spec.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
```
redsummernight: You should use double quotes here for consistency.

spec/controllers/user_sessions_controller_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+        end
+      end
+
+      context "when trying to login" do
+        before do
+          post :create, params: { user_session: { login: user.login, password: user.password, remember_me: true } }
+        end
+
+        it "should not succeed because user is locked" do
+          expect(flash.now[:error]).to match("Your account has been locked")
```
redsummernight: You should use `flash[:error]` to be consistent with the other tests.

spec/controllers/user_sessions_controller_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+
+describe UserSessionsController do
+  let(:user) { create(:user, activated_at: Time.new(2007, 1, 1)) }
+
+  context "create" do
+    context "after failed to login more than consecutive login limit" do
+      before do
+        max_consecutive_attempts = 2
+        allow(UserSession).to receive(:consecutive_failed_logins_limit).and_return(max_consecutive_attempts)
+        (max_consecutive_attempts + 1).times do
```
redsummernight: You don't need to add 1, this test works if you use `max_consecutive_attempts.times`.

redsummernight:
Looks good to me! 👍

PR #3138, opened 2017-11-03
AO3-4905 Removed check for recently reset password in change email sc…

hatal175:
…reen

## Issue

https://otwarchive.atlassian.net/browse/AO3-4905

## Purpose

This removes the check from email_change screen whether the user has recently requested a temporary password. This should not matter in this screen.

This could actually be used maliciously as well.

## Testing

Request temporary password. You could still login with normal password and change emails.

## Credit

Tal Hayon.
Please use he.

### DIFF
--- a/app/views/users/change_email.html.erb
+++ b/app/views/users/change_email.html.erb
@@ -19,10 +19,8 @@
     <dt><%= label_tag :email_confirmation, ts("Confirm New Email") %></dt>
     <dd><%= email_field_tag :email_confirmation %></dd>
 
-    <% unless @user.recently_reset? %>
-      <dt><%= label_tag :password_check, ts("Password") %></dt>
-      <dd><%= password_field_tag :password_check%></dd>
-    <% end %>
+    <dt><%= label_tag :password_check, ts("Password") %></dt>
+    <dd><%= password_field_tag :password_check%></dd>
 
     <dt class="landmark"><%= label_tag :submit, ts("Submit") %></dt>
     <dd class="submit actions">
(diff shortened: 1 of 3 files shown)


sarken:
Hello and thank you for this! It's going to make our Support team very happy. 

It looks good, but I have a question: can you use your temporary password to change your email, or just your regular one? And would it be possible to add a Cucumber test for this scenario?

hatal175:
I can add a cucumber test but which is the preferred way? Should you or should you not be able to change your email with the temporary password?

sarken:
I'm not sure which is preferred, which is why I was wondering what the current behavior is.

hatal175:
So currently you can't change your email with your temporary password. This makes sense as you wouldn't really have a way of getting the temporary password without your email :)

I've added feature tests to test that it fails with the temporary password and succeeds with the normal password after a password change request.

app/views/users/change_email.html.erb
```diff
@@ -19,10 +19,11 @@
     <dt><%= label_tag :email_confirmation, ts("Confirm New Email") %></dt>
     <dd><%= email_field_tag :email_confirmation %></dd>
 
-    <% unless @user.recently_reset? %>
-      <dt><%= label_tag :password_check, ts("Password") %></dt>
-      <dd><%= password_field_tag :password_check%></dd>
-    <% end %>
+    <dt><%= label_tag :password_check, ts("Password") %></dt>
+    <%= password_field_tag :password_check, nil, "aria-describedby" => "password-field-description" %>
```
sarken: The password field and the footnote still need to be wrapped in `<dd></dd>` tags to avoid invalid HTML -- sorry for being unclear!

PR #3139, opened 2017-11-04
AO3-5839 Add pseud tests

hatal175:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5839

## Purpose

UPDATE: Refer to https://github.com/otwcode/otwarchive/pull/3139#issuecomment-563439868

Even though the coverage was already at the desired level, I've added several feature tests to cover a few more code paths. I've also added a few pseud feature tests because I was in the code neighborhood.

Some of the tests are dependent on AO3-5221 - they need all the bookmarkable items to work properly.

I would like to request a review on the removal of a logic that extracts the bookmarkable item from the chapter id in bookmarks_controller. This wasn't in test coverage and as far as I can follow the code (and a few tests I did), no code path reaches it. the chapters/:chapter_id path redirects to works/:work_id/chapters/:chapter_id so it never comes into use.

## Testing

Feature tests.

## References

AO3-5221 in deferred feature tests.

## Credit

Tal Hayon

Please use he.


hatal175:
I think this reproduces on master as well

cesy:
Please note, this ticket was previously assigned to me and I was still working on it. I'll submit my PR shortly - I'd suggest creating a new ticket for your work, and then hopefully we can merge both.

features/bookmarks/bookmark_create.feature
```diff
@@ -48,7 +50,99 @@ Scenario: Create a bookmark
...
+  When I follow "Bookmark Series"
+    And I fill in "bookmark_notes" with "I liked this trilogy"
+    And I fill in "bookmark_tag_string" with "This is a series tag, and another series tag,"
+    And I check "bookmark_rec"
+    And I press "Create"
+  Then I should see "Bookmark was successfully created"
+    And I should see "My Bookmarks"
+  When I am logged in as "another_bookmark_user"
+    And I go to the bookmarks page
+    And "AO3-5221" is fixed
```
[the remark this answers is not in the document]
sarken: It turns out this isn't due to AO3-5221. It's because the index uses `Bookmark.latest` if the caching admin setting isn't enabled, and the `latest` scope only grabs works.

features/bookmarks/bookmark_create.feature
```diff
@@ -400,3 +494,22 @@ Scenario: Can use "Show Most Recent Bookmarks" from the bookmarks page
...
+@javascript
+Scenario: Can use "Show Most Recent Bookmarks" from the bookmarks page for external works
+  Given basic tags
+    And I am logged in as "bookmarker1"
+    And I bookmark the external work "Popular Work"
+    And I log out
+    And I am logged in as "bookmarker2"
+    And I bookmark the external work "Popular Work"
+  When I am on the bookmarks page
+    And "AO3-5221" is fixed
```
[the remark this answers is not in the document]
sarken: Same comment as above. This won't work on the latest bookmarks page unless test caching is enabled.

sarken:
1. This pull request previously aimed to expand coverage of the bookmarks controller. It was submitted while we were working on the Elasticsearch update and it looked like some test failures were due to an issue that was currently broken on staging; however, that turned out not to be the case. A new issue has been created to handle testing of the problematic area, and the tests are now in https://github.com/otwcode/otwarchive/pull/3708 as specs. This also allows the original issue to be reassigned to Cesy, who was its original owner and has also submitted a pull request.

2. The pull request also had some tests to expand coverage of the update action for the pseuds controller, which was outside the scope of the original issue. A new issue has been created for that and this pull request has been updated to contain only pseud tests.

features/bookmarks/bookmark_create.feature
```diff
@@ -33,7 +33,8 @@ Scenario: Create a bookmark
     When I am logged in as "first_bookmark_user"
       And I go to first_bookmark_user's user page
     Then I should not see "You don't have anything posted under this name yet"
-      And I should see "Revenge of the Sith"
+      And I should see "Revenge of the Sith"
```
redsummernight: The file can be removed from the PR.

features/step_definitions/pseud_steps.rb
```diff
@@ -19,3 +25,22 @@
...
+  click_button "Create"
+end
+
+When /^"([^"]*)" creates the pseud "([^"]*)"$/ do |username, newpseud|
+  visit new_user_pseud_path(username)
+  fill_in "Name", with: newpseud
+  click_button "Create"
+end
+
+When /^I create the pseud "([^"]*)"$/ do |newpseud|
```
redsummernight: This step is identical to:

https://github.com/otwcode/otwarchive/blob/09565b8ee6f850fe22eab08823940b7347b849bb/features/step_definitions/pseud_steps.rb#L12-L16

We can remove it and use the other one.

sarken: Oops, yeah, I updated the step that relied on this one, but forgot to delete it when I was done. 🤦‍♂

PR #3142, opened 2017-11-04
AO3-4706 Readded feature test for download unrevealed work

hatal175:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4706

## Purpose

I've tried reproducing the bug described but could not manage to. When uncommenting the relevant feature test it seems to have passed with a slight change of where you end up after it refuses to download the file.

unless anyone can manage to reproduce this, I recommend closing the bug.

## Testing

Feature tests should cover this unless someone manages to reproduce the bug.

## Credit

Tal Hayon

Please use he.

### DIFF
--- a/features/works/work_download.feature
+++ b/features/works/work_download.feature
@@ -171,6 +171,5 @@ Feature: Download a work
   When I am logged in as the owner of "Invisibility"
     And I view the work "Now You See Me"
     And I follow "HTML"
-    And "AO3-4706" is fixed
-  # Then I should see "Sorry, you can't download an unrevealed work"
-  #   And I should be on the works page
+   Then I should see "Sorry, you can't download an unrevealed work"
+     And I should be on the "Now You See Me" work page


sarken:
Unfortunately, this issue behaves a bit differently on production. The good news is it's impossible to download an unrevealed work, so that problem is still solved, but the bad news is production doesn't do what the tests (or the code) says. When you try to download an unrevealed work there, it redirects to archiveofourown.org/works and _doesn't_ show an error message. I think we still need to do some work to ensure the desired error message it show.

hatal175:
I think this has to do with the redirection mechanism. If you try to download a work several times relatively fast (on test environment), it redirects to the default works page, but still shows the error message. Best thing I can reproduce on test environment.

hatal175:
Well honestly, this should probably be merged since the feature unitest does pass. However, I suppose the bug should not be closed.
Up to you, I suppose?

sarken:
After some discussion, we've decided we'd prefer not to merge this, since it won't let us know if we've broken or changed the behavior on production, which is what we're most interested in. Thank you, though!

PR #3143, opened 2017-11-05
AO3-5040 Anonymous works and series will now send subscription emails

hatal175:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5040

## Purpose

This fixes the issue where subscriptions to anonymous works and series do not send notification mails.

## Testing

I've added feature tests for anonymous works and anonymous series + a test to make sure that subscription to a user will send mails about his anonymous works.
Testing is described in the bug report.

I'd like a review if possible for some of the changes, especially the addition of the creators function to creatable and inheriting classes.

## Credit

Tal Hayon

Please use he.


features/other_b/subscriptions.feature
```diff
@@ -172,6 +172,63 @@
   When I press "Subscribe"
   Then the page title should include "Chapter 2"
 
+  Scenario: Subscription to a work in an anonymous collection
```
sarken: I think this test might be a bit more complicated than strictly necessary (a lot of the existing tests in here appear to be /o\). This might work:
```
Scenario: When a chapter is added to an anonymous work, subscription emails are sent
to users who have subscribed to the work

Given the anonymous collection "test_collection"
  And I am logged in as "first_user"
  And I post the work "Multi Chapter Work" to the collection "test_collection"
When I am logged in as "second_user"
  And I view the work "Multi Chapter Work"
  And I press "Subscribe"
When I am logged in as "first_user"
  And a chapter is added to "Multi Chapter Work"
  And subscription notifications are sent
Then 1 email should be delivered to "second_user@foo.com"
  And the email should contain "Multi Chapter Work"
  And the email should contain "Anonymous"
  And the email should not contain "first_user"
```

The other new tests might benefit from being similarly tweaked in places, including making the descriptions a bit more detailed.

features/other_b/subscriptions.feature
```diff
@@ -172,6 +172,63 @@
...
+  When I am logged in as "first_user"
+    And a chapter is added to "Multi Chapter Work"
+    Then 0 emails should be delivered
+  When subscription notifications are sent
+    Then 1 email should be delivered to "second_user@foo.com"
+    And the email should contain "Multi Chapter Work"
+    And the email should contain "Anonymous"
+    And the email should not contain "first_user"
+
+    Scenario: Subscription to a user should not inform you about his anonymous works
```
sarken: It might be a good idea to extend this test to cover not just new works posted to anonymous collections, but also new chapters added to existing anonymous works.

features/other_b/subscriptions.feature
```diff
@@ -172,6 +172,63 @@
...
+    When I am logged in as "first_user"
+      And I set up the collection "test_collection" with name "test1"
+      And I check "collection_collection_preference_attributes_anonymous"
+      And I submit
+      And I post the work "Multi Chapter Work" to the collection "test_collection"
+    Then 0 emails should be delivered
+    When subscription notifications are sent
+    Then 0 emails should be delivered
+
+    Scenario: Subscription to a series in an anonymous collection
```
sarken: It looks like this is testing that a subscription email is sent when a chapter is added to an anonymous work in an anonymous series. Should we also test that an email is sent when a new anonymous work is added to the series?

app/views/user_mailer/batch_subscription_notification.text.erb
```diff
@@ -3,11 +3,12 @@
 <% @seen_summary = {} %>
 <% @creations.each_with_index do |creation, index| %>
 <% this_work = creation.respond_to?(:work) ? creation.work : creation %>
-<%= this_work.pseuds.map{|p| p.byline}.to_sentence.html_safe %> posted a <% if creation.is_a?(Work) then %><%= this_work.backdate ? "backdated" : "new" %> work:<% else %>new chapter of <%= this_work.title %> (<%= this_work.word_count %> words):<% chapter_title = (creation.chapter_header + (creation.title.blank? ? "" : ": #{creation.title}")).html_safe %><% end %>
+<% creators = creator_links(this_work) %>
```
sarken: I think what we're aiming for up here is just a list of bylines, e.g. `user1, User 2 (user2), and user3 posted a new work:`, and below it should be the user's byline followed by the pseud's URL in parentheses, e.g. `by user1 (http://archiveofourown.org/users/user1/pseuds/user1), User 2 (user2)  (http://archiveofourown.org/users/user2/pseuds/User%202), and user3 (http://archiveofourown.org/users/user3/pseuds/user3)`.

However, it looks like `creator_links` uses the `style_pseud_link` helper, which involves a lot of HTML and CSS we probably don't want in plain text emails. Is it getting stripped out somewhere?

sarken:
This has some merge conflicts now, but no rush!

app/views/user_mailer/batch_subscription_notification.html.erb
```diff
@@ -5,13 +5,15 @@
 
     <% this_work = creation.respond_to?(:work) ? creation.work : creation %>
     <% work_link = style_link(this_work.title.html_safe, work_url(this_work)) %>
+    <% creators = creator_links(this_work) %>
```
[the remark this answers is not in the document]
sarken: That wasn't intentional. The cache block was implemented after this PR was opened, and I wasn't using my head when I resolved the conflicts this morning.

app/models/subscription.rb
```diff
@@ -38,11 +38,18 @@ def name
   end
 
   def subject_text(creation)
-    authors = creation.pseuds.map{ |p| p.byline }.to_sentence
+    authors = if creation.is_a?(Work) && creation.anonymous? || creation.is_a?(Chapter) && creation.work.anonymous?
```
redsummernight: You can use `anonymous_creation?` here.

sarken: This was annoying the daylights out of me. I tried `anonymous_creation?(creation)` and `self.anonymous_creation?(creation)` and couldn't get it to work even though I _knew_ there had to be something. But now I see the problem! 🤦‍♂

redsummernight: I'd prefer `self.class.anonymous_creation?(creation)`. `self.class` allows subclassing and less work when renaming the class (though neither are likely in this case).

PR #3157, opened 2017-11-12
AO3-4928 extend gifts_controller coverage

hatal175:
Fixed return problem in access_denied error path.

## Issue

https://otwarchive.atlassian.net/browse/AO3-4928

## Purpose

I extended test coverage for gifts_controller.

## Testing

Run tests.

## Credit

Tal Hayon

Please use he.


factories/gift.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+require 'faker'
+
+FactoryGirl.define do
+  factory :gift do
+    user { create(:user) }
+    work { create(:work) }
+    recipient "recipient"
```
redsummernight: Since you already fake the user recipient, you can leave out this field, which is for non-user recipients.

factories/gift.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+require 'faker'
```
redsummernight: Can you change these to double quotes?

spec/controllers/gifts_controller_spec.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+describe GiftsController do
+  include LoginMacros
+  include RedirectExpectationHelper
+
+  describe "toggle_rejected" do
+    let (:gift) { create(:gift) }
```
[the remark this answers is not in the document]
sarken: Hound is right -- we do `let(:gift)` without the space.

spec/controllers/gifts_controller_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    it "errors and redirects to login page if no user is logged on" do
+      pending("AO3-5445 fix")
```
sarken: We use `xit` for [pending specs ](https://github.com/otwcode/otwarchive/wiki/Automated-Testing#Pending_Tests)

factories/gift.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+require "faker"
+
+FactoryGirl.define do
```
sarken: Since this is a new factory, it should probably have a [model test](https://github.com/otwcode/otwarchive/blob/master/spec/models/gift_spec.rb) along the lines of

```
  it "has a valid factory" do
    expect(build(:gift)).to be_valid
  end
```

PR #3165, opened 2017-11-17
AO3-4862 Extending work controller test coverage

hatal175:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4862

## Purpose

This PR extends the test coverage for works controller.

I've also made a few changes for this purpose and I would really appreciate a review on them:
1. I've added a route for show_multiple under pseuds so you can see and edit only the works under that pseud.
2. I've deleted a code path in preview_mode that could not really be reached in current code base.
3. Deleted unused function tag_list
4. various bug fixes in previously uncovered code.

## Testing

Run tests and see coverage improvement.

## Credit

Tal Hayon

Please use he.


sarken:
Thanks for the pull request! However, this appears to have gone a bit beyond the scope of the issue, which just calls for the test coverage to be extended. We really do appreciate the bug fixes, but for numerous reasons, we prefer small pull requests that match the issue as closely as possible. 

Before we can review this, we'd appreciate it if you could reduce its scope a bit.

>I've added a route for show_multiple under pseuds so you can see and edit only the works under that pseud.

This isn't something we've had any requests for and aren't currently interested in implementing. 

>I've deleted a code path in preview_mode that could not really be reached in current code base.

Hound had an additional comment on this regarding a lingering unused argument. Ultimately, this seems like more of a refactor of the method than a simple deletion of unused code, so we'll make an issue for that once we know what to test to make sure the method still works.

>Deleted unused function tag_list

We've made a separate issue for this: AO3-5457

>various bug fixes in previously uncovered code.

We'll need a detailed list of the bugs and descriptions of how to manually reproduce them (whenever possible) to make issues as appropriate. 

You can either give us the info here or email otw-coders@transformativeworks.org with the details -- whichever is easier. 

Thanks again!

hatal175:
Well, I removed the changes I made beyond adding tests and pending tests. I've sent a mail about the things I found.
I'd still like to reexamine the preview mode lines and another line dealing with collections before I consider this ready for review

PR #3166, opened 2017-11-17
AO3-4818 Complete coverage of prompt meme controller

hatal175:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4818

## Purpose

Complete coverage of prompt meme controller

## Testing

Run tests.

## Credit

Tal Hayon

Please use he.

### DIFF
--- a/spec/controllers/prompt_meme_controller_spec.rb
+++ b/spec/controllers/prompt_meme_controller_spec.rb
@@ -3,6 +3,46 @@ require 'spec_helper'
 describe Challenge::PromptMemeController do
   include LoginMacros
 
+  describe "new" do
+    context "when the collection already has a challenge" do
+      before do
+        @collection = FactoryGirl.create(:collection, challenge: PromptMeme.new)
+        fake_login_known_user(@collection.owners.first.user)
+        post :new, params: { collection_id: @collection.name }
+      end
+
+      it "should set flash notice" do
+        expect(flash[:notice]).to eq("There is already a challenge set up for this collection.")
+      end
+
+      it "should redirect to edit meme collection path" do
+        expect(response).to redirect_to(edit_collection_prompt_meme_path(@collection))
+      end
+    end
+  end
+
+  describe "update" do
+    context "when it fails to udpate parameters" do
+      before do
+        challenge = PromptMeme.new
+        @collection = FactoryGirl.create(:collection, challenge: challenge)
+        fake_login_known_user(@collection.owners.first.user)
+        allow_any_instance_of(PromptMeme).to receive(:update_attributes).and_return(false)
+        allow(controller).to receive(:prompt_meme_params).and_return({})
+        post :update, params: { collection_id: @collection.name, propmt_meme: {} }
+      end
+
+      it "should render edit page" do
+        expect(response).to render_template "edit"
+      end
+
+      after do
+        allow(controller).to receive(:prompt_meme_params).and_call_original
+        allow_any_instance_of(PromptMeme).to receive(:update_attributes).and_call_original
+      end
+    end
+  end
+
   describe "destroy" do
     before(:each) do
       @collection = FactoryGirl.create(:collection, challenge: PromptMeme.new)


spec/controllers/prompt_meme_controller_spec.rb
```diff
@@ -3,6 +3,46 @@
...
 
+  describe "new" do
+    context "when the collection already has a challenge" do
+      before do
+        @collection = FactoryGirl.create(:collection, challenge: PromptMeme.new)
+        fake_login_known_user(@collection.owners.first.user)
+        post :new, params: { collection_id: @collection.name }
+      end
+
+      it "should set flash notice" do
```
sarken: Because it's something we test a lot, we've set up a bunch of [redirect expectation helpers](https://github.com/otwcode/otwarchive/blob/master/spec/support/redirect_expectation_helper.rb) that we prefer to use for testing redirects with notice or error messages. For example, we might condense this down to:

```
context "when the collection already has a challenge" do
  [...]

  it "redirects to challenge edit page with notice"
    it_redirects_to_with_notice(edit_collection_prompt_meme_path(@collection), "There is already a challenge set up for this collection.")
  end
```

Generally, we prefer to leave off the "should" in our descriptions, per [these guidelines](http://www.betterspecs.org/#should).

spec/controllers/prompt_meme_controller_spec.rb
```diff
@@ -3,6 +3,46 @@
...
+      it "should redirect to edit meme collection path" do
+        expect(response).to redirect_to(edit_collection_prompt_meme_path(@collection))
+      end
+    end
+  end
+
+  describe "update" do
+    context "when it fails to udpate parameters" do
+      before do
+        challenge = PromptMeme.new
```
sarken: It looks like the only reference to the `challenge` variable here is in `@collection = FactoryGirl.create(:collection, challenge: challenge)`. Would it be possible to avoid creating the variable and do `FactoryGirl.create(:collection, challenge: PromptMeme.new)` here like in the test above? 

You can leave off `FactoryGirl` for brevity if you like and it should still work.

spec/controllers/prompt_meme_controller_spec.rb
```diff
@@ -3,6 +3,46 @@
...
+      end
+
+      it "should redirect to edit meme collection path" do
+        expect(response).to redirect_to(edit_collection_prompt_meme_path(@collection))
+      end
+    end
+  end
+
+  describe "update" do
+    context "when it fails to udpate parameters" do
```
sarken: Little typo here -- should be "update."

spec/controllers/prompt_meme_controller_spec.rb
```diff
@@ -2,6 +2,42 @@
 
 describe Challenge::PromptMemeController do
   include LoginMacros
+  include RedirectExpectationHelper
+
+  describe "new" do
+    context "when the collection already has a challenge" do
+      before do
+        @collection = FactoryGirl.create(:collection, challenge: PromptMeme.new)
```
redsummernight: We should use let to avoid sharing state between tests in instance variables. You don't need to update the existing instance variables though.

spec/controllers/prompt_meme_controller_spec.rb
```diff
@@ -2,6 +2,42 @@
...
+  end
+
+  describe "update" do
+    context "when it fails to update parameters" do
+      before do
+        @collection = FactoryGirl.create(:collection, challenge: PromptMeme.new)
+        fake_login_known_user(@collection.owners.first.user)
+        allow_any_instance_of(PromptMeme).to receive(:update_attributes).and_return(false)
+        allow(controller).to receive(:prompt_meme_params).and_return({})
+        post :update, params: { collection_id: @collection.name, propmt_meme: {} }
```
redsummernight: Should be `prompt_meme`.

spec/controllers/prompt_meme_controller_spec.rb
```diff
@@ -2,6 +2,42 @@
...
+      it "redirects to edit meme collection path with notice" do
+        it_redirects_to_with_notice(edit_collection_prompt_meme_path(@collection), "There is already a challenge set up for this collection.")
+      end
+    end
+  end
+
+  describe "update" do
+    context "when it fails to update parameters" do
+      before do
+        @collection = FactoryGirl.create(:collection, challenge: PromptMeme.new)
```
redsummernight: Just `create(:collection, challenge: PromptMeme.new)` should work here and below. We prefer to [leave out `FactoryGirl` in the specs](https://github.com/otwcode/otwarchive/wiki/Automated-Testing#Our_RSpec_Standards).

PR #3167, opened 2017-11-17
AO3-5019 Fix text overlap in related works section on downloads

tuff:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5019

## Purpose

Fix overlapping text in related works section on downloads.

## Testing

See issue.  The new related works section should look like this:
image

### DIFF
--- a/app/views/downloads/_download_preface.html.erb
+++ b/app/views/downloads/_download_preface.html.erb
@@ -54,16 +54,17 @@ hide_freeform?(@work)) %>
     <% end %>
   <% end %>
 
-  <% unless @work.parents.blank? %>
-    <dl>
-      <dt><%= ts('This work was inspired by') %></dt>
-      <dd>
-        <% parent_work_list = @work.parent_work_relationships.reject{ |wr| wr.parent.nil? } %>
-        <%= parent_work_list.map{ |rw| "#{link_to(rw.parent.title,
-               url_for(action: :show, controller: rw.parent_type.underscore.pluralize, id: rw.parent_id,
-                       only_path: false)) } #{ts('by')} #{byline(rw.parent, only_path: false)}" }.join(", ").html_safe %>
-      </dd>
-    </dl>
+  <% related_works = @work.parent_work_relationships.reject{ |wr| !wr.parent } %>
+  <% if related_works.length > 0 %>
+    <ul>
+    <% related_works.each do |work| %>
+      <li>
+        <%= ts(work.translation ? 'A translation of' : 'Inspired by') %>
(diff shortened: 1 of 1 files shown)


app/views/downloads/_download_preface.html.erb
```diff
@@ -54,16 +54,17 @@ hide_freeform?(@work)) %>
...
-      </dd>
-    </dl>
+  <% related_works = @work.parent_work_relationships.reject{ |wr| !wr.parent } %>
+  <% if related_works.length > 0 %>
+    <ul>
+    <% related_works.each do |work| %>
+      <li>
+        <%= ts(work.translation ? 'A translation of' : 'Inspired by') %>
+        <%= link_to work.parent.title.html_safe, work_url(work.parent) %>
+        <%= ts("by") %> <%= byline(work.parent) %>.
```
redsummernight: Previously `byline` was called with the option `only_path: false`, is it why in the screenshot "testy" are not pseud links?

tuff: I noticed that they were not links on the page either, and assumed that was normal!

I think, on my local instance, the works were not being picked up as AO3 works and so were external...and it's probably only normal for external works.  I'll restore `only_path`, which I definitely removed in error.

redsummernight:
Looks good, thanks! One last thing, I think it would be easier for future translators if your `ts` contains the entire context. So instead of having 2 `ts` pieces with the work link in between, we should do something like:

    if work.translation
        string = "A translation of #{work} by #{author}"
    else
        string = "Inspired by #{work} by #{author}"
    end

I'd wait for someone more familiar with Rails i18n to weigh in before changing things, though.

sarken:
Just confirming that red's suggestion is a good one!

app/views/downloads/_download_preface.html.erb
```diff
@@ -54,16 +54,22 @@ hide_freeform?(@work)) %>
...
+  <% related_works = @work.parent_work_relationships.reject{ |wr| !wr.parent } %>
+  <% if related_works.length > 0 %>
+    <ul>
+    <% related_works.each do |work| %>
+      <li>
+        <% relation_text = work.translation ?
+          "A translation of %{work_link} by %{author_link}" :
+          "Inspired by %{work_link} by %{author_link}"
+        %>
+        <% work_link = link_to work.parent.title.html_safe, work_url(work.parent) %>
```
redsummernight: The related work can be an ExternalWork so we can't use `work_url` here, it should be a `url_for` like before.

redsummernight: I like it a lot more than `url_for(action: :show, ...)` :+1:

PR #3153, opened 2017-11-11
AO3-4605 Add language field to import from url page

hatal175:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4605

## Purpose

Adds a language field to the "Import from URL" option.

## Testing

I've added some checks to the relevant feature tests but this may be tested by importing from a url and setting a language other than "English".

## Credit

Tal Hayon

Please use he.

PR #3169, opened 2017-11-18
AO3-4205 Add IP address when importing works without preview

hatal175:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4205

## Purpose

Adds IP address of importer when posting without preview.

## Testing

I added feature tests for these cases but the described procedure in the bug should do.

## Credit

Tal Hayon

Please use he.


features/importing/work_import.feature
```diff
@@ -110,6 +110,32 @@ Feature: Import Works
...
+  Scenario: Admins see IP address on multi-chapter works imported without preview
+    Given I import the urls with mock websites as chapters without preview
+      """
+      http://import-site-without-tags
+      http://second-import-site-without-tags
+      """
+    When I am logged in as an admin
+      And I go to the "Untitled Imported Work" work page
+    Then I should see "Chapters:2/2"
+    Then I should see "IP Address: 127.0.0.1"
```
sarken: Can this second "Then" be an "And", indented two more spaces?

redsummernight: :+1: I will also wait for #3153 to be merged since it hits the same import tests.

PR #3175, opened 2017-11-20
AO3-4987 Add POST pagination support for admin email bulk search

hatal175:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4987

## Purpose

I've changed the form method of the bulk email search to get instead of post. The paginate works now. I would like a code review on that though. Not entirely sure why it was post in the first place - they went to the same function anyway.

## Testing

Recreation steps in the bug report.

## Credit

Tal Hayon

Please use he.

### DIFF
--- a/app/renderers/post_link_renderer.rb
+++ b/app/renderers/post_link_renderer.rb
@@ -3,7 +3,7 @@
 class PostLinkRenderer < WillPaginate::ActionView::LinkRenderer
   def previous_or_next_page(page, text, classname)
     if page
-      submit(text, page, classname,class: classname)
+      submit(text, page, classname, class: classname)
     else
       tag(:span, text, class: classname + ' disabled')
     end
@@ -18,7 +18,7 @@ class PostLinkRenderer < WillPaginate::ActionView::LinkRenderer
   end
 
   def submit(text, target, target_name, attributes = {})
-    string_attributes = attributes.inject(''.dup) do |attrs, pair|
+    string_attributes = attributes.inject(+'') do |attrs, pair|
       unless pair.last.nil?
         attrs << %( #{pair.first}="#{CGI.escapeHTML(pair.last.to_s)}")
       end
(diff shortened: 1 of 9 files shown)


hatal175:
Let me consider this for a bit.

hatal175:
I think I managed to create a decently designed post paginator. Mind taking a look?

ariana-paris:
I patched up my local Archive installation and I can confirm that your link renderer works correctly and allows the form to handle hundreds of emails while also correctly paginating the results. \o/

Just one thing, we already have a `will_paginate` renderer called **PaginationListLinkRenderer**, which lives in `lib` and is hooked in by a `will_paginate()` override in **ApplicationHelper**. It makes some changes to how the links are rendered and it would be nice if the new POST pagination was the same (I think the main visual difference was just the current page link being in an `em` in your code, and a `span` in the one currently in use)

Also, I think it would make sense for both renderers to live in the same folder distinct from `lib`, so if you could also move the existing one to join yours in `app/renderers`, that would be great! (we're still bogged down in putting out fires, so there's time before this will be merged)

app/renderers/post_link_renderer.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+class PostLinkRenderer < WillPaginate::ActionView::LinkRenderer
+  def previous_or_next_page(page, text, classname)
+    if page
+      submit(text, page, classname, class: classname)
+    else
+      tag(:span, text, class: classname + ' disabled')
```
ariana-paris: Can you use double quotes throughout as that is our standard?

app/renderers/post_link_renderer.rb
```diff
@@ -0,0 +1,35 @@
...
+    end
+  end
+
+  def page_number(page)
+    if page == current_page
+      tag(:em, page, class: 'current')
```
ariana-paris: This is one of the differences with our current **PaginationListLinkRenderer**: the existing one renders `span`s instead of `em`s, so it would be nice if this one did too so that the post pagination looks the same as the pagination we have elsewhere on the site (but maybe have a look at our existing renderer and see if there are any other differences too)

PR #3178, opened 2017-11-22
AO3-4989 Returning to referer if deleting bookmark

hatal175:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4989

## Purpose

This makes it so you return to referer if you delete a bookmark unless you came from a page specifically editing or deleting it. I saw the pattern used in the comment controller but I'm not sure it it's the best way of implementing this.

## Testing

Delete bookmark from all possible places. If page still exists (user bookmarks second page, recent bookmarks, etc.) then you should still be at that page.

## Credit

Tal Hayon

Please use he.

### DIFF
--- a/app/controllers/bookmarks_controller.rb
+++ b/app/controllers/bookmarks_controller.rb
@@ -213,7 +213,12 @@ class BookmarksController < ApplicationController
   def destroy
     @bookmark.destroy
     flash[:notice] = ts("Bookmark was successfully deleted.")
-    redirect_to user_bookmarks_path(current_user)
+
+    if request.referer&.match(/confirm_delete|edit/)
+      redirect_to(user_bookmarks_path(current_user))
+    else
+      redirect_to(request.referer || user_bookmarks_path(current_user))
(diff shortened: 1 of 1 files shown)


sarken:
Can keep it like it originally was and just add the page parameter to the path (i.e. `redirect_to user_bookmarks_path(current_user, page: params[:page])`, or does that lead to an error?

hatal175:
You could delete a bookmark from the browse bookmarks pages and end up in
your bookmarks page. I thought that was also counter-intuitive so i added
the referer which solved this issue as well You think I overdid it?

sarken:
Yeah, I think we can consider that behavior separately and focus on the problem described in the issue.

hatal175:
Looking at the code again, I really don't think the optimal solution would be to add the page somehow... the deletion link is deep inside the rendering parts and has no idea where it is rendered, nor I think it should know.

If this is not acceptable, I recommend closing this pull request and extending the requirements of this feature.

PR #3190, opened 2017-11-30
AO3-5229: Simple fix for ancient date errors

elzj:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5229

## Purpose

Fix error encountered when a user enters a very large date in date search. Since we're really just concerned with things posted by contemporaneous people, a simple fix should suffice for now.

## Testing

Entering a large value in date search should no longer error.


sarken:
redsummernight gave the thumbs-up in chat, so merging!

PR #3191, opened 2017-12-01
AO3-5226: People Search

elzj:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5226
https://otwarchive.atlassian.net/browse/AO3-5151

## Purpose

Updates people search to add an autocomplete field and a fandom search. Also simplifies some unnecessary people pages that we weren't really using.

## Testing

See issue.

### DIFF
--- a/app/decorators/pseud_decorator.rb
+++ b/app/decorators/pseud_decorator.rb
@@ -1,0 +1,111 @@
+class PseudDecorator < SimpleDelegator
+
+  attr_reader :data
+
+  # TODO: pull this out into a reusable module
+  def self.decorate_from_search(results, search_hits)
+    search_data = search_hits.group_by { |doc| doc["_id"] }
+    results.map do |result|
+      data = search_data[result.id.to_s].first&.dig('_source') || {}
+      new_with_data(result, data)
+    end
+  end
+
+  def self.decorate(pseuds)
+    users = User.where(id: pseuds.map(&:user_id)).group_by(&:id)
+    work_counts
+    bookmark_counts
+    work_key = User.current_user.present? ? :general_works_count : :public_works_count
+    pseuds.map do |pseud|
+      data = {
+        user_login: users[user_id].login,
+        public_bookmarks_count: bookmark_counts[id],
+        work_key => work_counts[id]
+      }
+      new_with_data(pseud, data)
+    end
+  end
+
+  def self.new_with_data(pseud, data)
+    new(pseud).tap do |decorator|
+      decorator.data = data
+    end
+  end
+
+  def data=(info)
+    @data = HashWithIndifferentAccess.new(info)
+  end
+
+  def works_count
+    User.current_user.present? ? data[:general_works_count] : data[:public_works_count]
+  end
+
+  def bookmarks_count
+    data[:public_bookmarks_count]
+  end
+
+  def byline
+    data[:byline] ||= constructed_byline
+  end
+
+  def user_login
+    data[:user_login]
+  end
+
+  def pseud_path
+    "/users/#{user_login}/pseuds/#{name}"
+  end
+
+  def works_path
+    "#{pseud_path}/works"
+  end
+
+  def works_link
+    return unless works_count > 0
+    text = ActionController::Base.helpers.pluralize(works_count, "works")
+    "<a href='#{works_path}'>#{text}</a>"
+  end
+
+  def bookmarks_path
+    "#{pseud_path}/bookmarks"
+  end
+
+  def bookmarks_link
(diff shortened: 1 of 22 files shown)


app/models/search/pseud_query.rb
```diff
@@ -28,12 +28,29 @@ def collection_filter
...
   ###########
   # QUERIES
   ###########
 
   def general_query
-    { query_string: { query: options[:query] } } if options[:query]
+    {
+      simple_query_string:{
+        query: escape_reserved_characters(options[:query]),
+        fields: ["name^5", "user_login^2", "description"],
```
redsummernight: 👍 

For my own reference, this means name is 5 times as important as description. https://www.elastic.co/guide/en/elasticsearch/reference/6.0/query-dsl-multi-match-query.html#_literal_fields_literal_and_per_field_boosting

app/decorators/pseud_decorator.rb
```diff
@@ -0,0 +1,111 @@
...
+      data = search_data[result.id.to_s].first&.dig('_source') || {}
+      new_with_data(result, data)
+    end
+  end
+
+  def self.decorate(pseuds)
+    users = User.where(id: pseuds.map(&:user_id)).group_by(&:id)
+    work_counts
+    bookmark_counts
+    work_key = User.current_user.present? ? :general_works_count : :public_works_count
```
sarken: I can never remember -- will `User.current_user.present?` be true if you're logged in as an admin? If not, should this be updated so admins will see the same work counts as users? (I know this isn't intended as an admin tool, but I have a feeling it will be useful at times.)

app/models/bookmark.rb
```diff
@@ -106,6 +107,13 @@ def invalidate_bookmark_count
     end
   end
 
+  # We index the bookmark count, so if it should change, update the pseud
+  def update_pseud_index
+    return unless $rollout.active?(:start_new_indexing)
+    return unless destroyed? || saved_change_to_id? || saved_change_to_private?
```
sarken: What case is `saved_change_to_id?` covering? Making a new bookmark?

app/models/search/pseud_indexer.rb
```diff
@@ -25,7 +36,58 @@ def document(object)
     object.as_json(
       root: false,
       only: [:id, :user_id, :name, :description, :created_at],
-      methods: [:user_login]
-    )
+      methods: [
+        :user_login,
+        :byline,
+        :collection_ids
```
sarken: `collection_ids` is for people search inside collections?

elzj: Yes, it might not be very useful, but it still exists, so I thought we should keep it in the index.

app/models/bookmark.rb
elzj: Yep. I don't think new_record? works after you save it.

app/decorators/pseud_decorator.rb
elzj: Yep: https://github.com/otwcode/otwarchive/blob/master/app/controllers/application_controller.rb#L295

app/views/people/_new_search_form.html.erb
```diff
@@ -0,0 +1,27 @@
+<%= form_for @search, as: :people_search, :url => search_people_path, :html => {:class => 'search', :method => :get} do |f| %>
+  <fieldset>
+    <legend>Search people</legend>
```
sarken: Could you make this translatable and remove the colons from the field labels? Also, if you could convert it from hash rockets, that would be great

app/views/people/index.html.erb
```diff
@@ -1,49 +1,25 @@
 <!--SEARCHBROWSE Descriptive page name, messages and instructions-->
 
 <div class="pseud">
```
sarken: Could you remove this surrounding div? There's no godly reason for it to be here -- I wouldn't be surprised if it were left over from layout 1.0

app/views/people/_new_search_form.html.erb
```diff
@@ -0,0 +1,27 @@
...
+        <%= f.text_field :name, autocomplete_options("pseud") %>
+      </dd>
+      <dt>
+        <%= f.label :fandom, ts("Fandom:") %>
+      </dt>
+      <dd>
+        <%= f.text_field :fandom, autocomplete_options("fandom") %>
+      </dd>
+    </dl>
+    <p class="submit actions"><%= f.submit ts('Search people'), :name => nil %></p>
```
sarken: Could this be made title case (Search People) to go with our other buttons?

PR #3192, opened 2017-12-01
AO3-5100 Rework Open Challenges pages

hatal175:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5100

## Purpose

This does most of what is described in the bug report.

1. It correctly sorts the 'all open challenge page' - not just soonest to close on top but also merges both types of challenges which previously was not done.
2. In each individual challenge, it adds a partial filter sidebar (no sorting or choosing challenge type).

I would like to request an opinion on whether the changes are what the author had in mind for this feature. Also an opinion for the changes themselves, maybe with regards to the filtering and their efficiency, would be appreciated.  

## Testing

I've added some basic tests but the bug report describes the feature.

## Credit

Tal.
Please use he.


sarken:
Closing -- we'll come back to this after https://github.com/otwcode/otwarchive/pull/3912

PR #3193, opened 2017-12-02
Experiments in filtering tweaks

sarken:
* Reworked the filter JavaScript to 
  * use different classes and variable names (`tags` no longer made sense) (views updated to use the new JavaScript class names)
  * make sure the word count and date filters would auto-open when filled in
  * make sure crossover and status filters won't auto-open when the default value is chosen
* Changed the section heading from "Date" to "Date Updated" because I wasn't sure which it was searching by, and figured others might wonder, too
* Removed colons and underlines on the "To" and "From" labels for word count and date ranges
* Reworded label from "Show only crossovers" to "Only crossovers," although maybe we want to say "Include only" or "Only include" 

One bit of weirdness that still exists is the help link for the crossover section of filters also serves to toggle the crossover section -- that is, if you click it, the pop up will open and the section will expand/collapse from its current state. That happens because the headings of expandable sections are turned into full-width buttons when JavaScript is enabled (this is to provide a larger clickable/tappable area for ease of use), so the help pop up must either be a) included in the button or b) pushed on to the next line. The alternative is to shrink the clickable area so it only includes the word,  but I think the convenience of the larger touch area trumps the minor oddness of tapping on the crossover section's help link expanding the crossover section.


sarken:
God damn it, this was supposed to be pull request on elz's branch

PR #3198, opened 2017-12-04
AO3-5268: Fix for bookmarks on tag pages

elzj:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5268

## Purpose

Fixes issue where adding a tag to a bookmark did not make it show up on that tag's bookmark listings.

This is branched off the fix for 5273 (https://github.com/otwcode/otwarchive/pull/3197) to avoid raging code conflicts.

## Testing

See issue.


tickinginstant:
I think there's a fundamental issue with this approach.

Suppose that I'm looking for the tags "A" and "B" jointly. If bookmarkable filter_ids are interchangeable with bookmark tag_ids, then there are four ways that a bookmark could match:

1. The bookmark has both tags.
2. The work has both tags.
3. The bookmark has tag "A", the work has tag "B."
4. The work has tag "A", the bookmark has tag "B."

The approach that you've taken works fine for looking up bookmarks. The trouble starts once you start trying to do the bookmarkable query for the filters in the sidebar. If a work has two bookmarks, one with tag "A" and one with tag "B", then that's not a match -- the work's tags shouldn't show up in the sidebar. So you can't just say (work tagged A OR bookmark tagged A) AND (work tagged B OR bookmark tagged B), because when you're doing a bookmarkable query, you need to specify that it's the same bookmark for both.

Fixing that would require rewriting the query structure you're using here. You'd need to transform it into something more like (bookmark tagged A AND B) OR (work tagged A AND B) OR (work tagged A AND bookmark tagged B) OR (bookmark tagged B AND work tagged A).

Effectively, you have a boolean formula in conjunctive normal form, but you need it in disjunctive normal form (or something very close to it). Which is going to be an exponential blow-up in the number of factors.

I think if you want to support queries like this, you'd need to do one of the following:
1. Limit the number of tags that users can add to their filter.
2. Do some tricks with score (and even then, you'd be bounded by the fact that it's a single-precision float, so it would be impossible to have the sidebar queries look right once someone has applied enough tags).
3. Use nested has_parent queries inside has_child queries (if that's even possible -- I'm guessing it's not since you went to a lot of trouble to avoid them, but I haven't checked).
4. Go back to storing bookmarkable filter_ids with their bookmarks.

elzj:
@tickinginstant Yeah, I initially said it wouldn't work and then I threw a bunch of scenarios at the wall and got myself turned around. But looking at it with a clearer head, I think you're right and that there's no way to make this work without denormalizing the bookmark data again.

PR #3199, opened 2017-12-05
AO3-5269 Fix regex for getting sortable work authors

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5269

## Purpose

The hyphen should be escaped, otherwise we're removing the first character if it's anything in the range from "+" to "=".

## Testing

See issue.

### DIFF
--- a/spec/models/work_search_form_spec.rb
+++ b/spec/models/work_search_form_spec.rb
@@ -1,50 +1,40 @@
 require 'spec_helper'
 
 describe WorkSearchForm do
-  before(:each) do
-    work.save
-    # This doesn't work properly in the factory.
-    second_work.collection_ids = [collection.id]
-    second_work.save
+  describe "searching" do
+    let!(:collection) do
+      FactoryGirl.create(:collection, id: 1)
+    end
 
-    update_and_refresh_indexes('work')
-  end
+    let!(:work) do
+      FactoryGirl.create(:work,
+                         title: "There and back again",
+                         authors: [Pseud.find_by(name: "JRR Tolkien") || FactoryGirl.create(:pseud, name: "JRR Tolkien")],
+                         summary: "An unexpected journey",
+                         fandom_string: "The Hobbit",
+                         character_string: "Bilbo Baggins",
+                         posted: true,
+                         expected_number_of_chapters: 3,
+                         complete: false,
+                         language_id: 1)
+    end
 
-  after(:each) do
-    Work.destroy_all
-    delete_index 'work'
-  end
+    let!(:second_work) do
+      FactoryGirl.create(:work,
+                         title: "Harry Potter and the Sorcerer's Stone",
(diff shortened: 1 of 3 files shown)


spec/models/work_search_form_spec.rb
```diff
@@ -243,4 +232,29 @@
       end
     end
   end
+
+  describe "sorting results" do
+    describe "by authors" do
+      before do
+        %w(-zzz _yasuho 21st_wombat 007aardvark 6tasmanian_devil).each do |pseud_name|
```
elzj: My only thought here is that since it's the work method that was broken, maybe we should test the behavior of it primarily in a work spec unit test. Then when we know it's saving the values correctly, we might not need to create as many works here and we can save some runtime that way.

spec/models/work_spec.rb
```diff
@@ -102,6 +102,26 @@
     end
   end
 
+  describe "authors to sort on" do
```
elzj: So this is totally fine, but while I'm nitpicking, I'd say you want the spec names to be a little more readable, and, whenever possible, to not touch the database if you can avoid it. I would make it something like this:

```
describe "#set_author_sorting" do
    let(:work) { build(:work) }

    context "when the pseuds start with special characters" do
      it "should remove those characters" do
        work.authors = [Pseud.new(name: "-jolyne")]
        work.set_author_sorting
        expect(work.authors_to_sort_on).to eq "jolyne"

        work.authors = [Pseud.new(name: "_hermes")]
        work.set_author_sorting
        expect(work.authors_to_sort_on).to eq "hermes"
      end
    end

    context "when the pseuds start with numbers" do
      it "should not remove them" do
        work.authors = [Pseud.new(name: "007james")]
        work.set_author_sorting
        expect(work.authors_to_sort_on).to eq "007james"        
      end
    end

    context "when the work is anonymous" do
      it "should set the author sorting to Anonymous" do
        work.in_anon_collection = true
        work.authors = [Pseud.new(name: "stealthy")]
        work.set_author_sorting
        expect(work.authors_to_sort_on).to eq "Anonymous"
      end
    end

    context "when the work has multiple pseuds" do
      it "should combine them with commas" do
        work.authors = [Pseud.new(name: "diavolo"), Pseud.new(name: "doppio")]
        work.set_author_sorting
        expect(work.authors_to_sort_on).to eq "diavolo,  doppio"
      end
    end
  end
```

There, you can add as many examples as you like, and it's not going to add significantly to the test runtime because nothing gets saved. And then the integration test in the feature will confirm that the whole thing works together - that the field gets saved and that we can sort on it.

redsummernight: Thanks! It didn't occur to me to just call `work.set_author_sorting`, I'll update the specs.

Should I remove or keep the new tests in `spec/models/work_search_form_spec.rb`?

PR #3205, opened 2017-12-10
AO3-5284 - Fix searching tags by type. [Old indexing code.]

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5284

## Purpose

The purpose of this pull request is to modify the fields that are indexed for Tags to ensure that it's possible to search by type. Currently, the tags on the archive can be divided into three categories:

1. **Tags manually indexed with reindex_document, or indexed before the IndexSubqueue code was written.** These tags have their _type set to, e.g. "fandom" or "character," and since the search code currently only checks the _type field, these are the only tags that will be returned in the search.

2. **Tags indexed with the IndexSubqueue code.** These tags have their _type set to "tag" (meaning they won't be returned by the current search code) and the JSON documents that are sent to ES contain no fields relating to their type. In order to be searched by type, all of these tags would have to be reindexed.

3. **Tags indexed both ways.** They show up in the search results twice. One of the results should probably be deleted.

It's very difficult to delete tags when their _type is set to "fandom" or "character" because as soon as a tag is deleted, the type association disappears, and so the delayed IndexSubqueue code has no idea what to use as the _type to delete the tag from the index. So I've chosen to modify the code so that all newly-reindexed tags have the _type "tag," regardless of the method of indexing, and store the tag type (e.g. "Fandom" or "Character") in a field of the JSON called tag_type (modeled after the new indexing code).

This pull request also sets up the tag search so that it will return results with the proper type value in *either* the _type field or the tag_type field. That way, even if no reindexing is performed, all tags indexed the old way will still appear. But newly created/modified tags will start to show up in the search, and search will work for all tags if a full reindexing is performed. (Though there may still be duplicates unless all entries are deleted before the reindexing happens.)

## Testing

See bug report for testing steps.

### DIFF
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -35,6 +35,20 @@ class Tag < ApplicationRecord
     end
   end
 
+  # ES UPGRADE TRANSITION #
+  # Delete this function, since it's unnecessary.
+  def self.document_type
+    "tag"
+  end
+
+  delegate :document_type, to: :class
+
+  # ES UPGRADE TRANSITION #
+  # Delete this function, since it's unnecessary.
+  def to_indexed_json
+    as_json.merge(tag_type: type).to_json
+  end
+
   def document_json
     TagIndexer.new({}).document(self)
   end
@@ -189,12 +203,6 @@ class Tag < ApplicationRecord
     if unwrangleable? && is_a?(UnsortedTag)
       self.errors.add(:unwrangleable, "can't be set on an unsorted tag.")
     end
-  end
-
-  before_update :remove_index_for_type_change, if: :type_changed?
-  def remove_index_for_type_change
-    @destroyed = true
-    reindex_document
   end
 
   before_validation :check_synonym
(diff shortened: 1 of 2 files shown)


app/models/tag.rb
```diff
@@ -1212,8 +1220,18 @@ def self.search(options={})
...
+          if options[:type].present?
+            # To support the tags indexed prior to IndexSubqueue, we want to
+            # find the type either in the tag_type field or the _type field:
+            should { term '_type', options[:type].downcase }
+            should { term :tag_type, options[:type].downcase }
+
+            # The tire gem doesn't natively support :minimum_should_match,
+            # but elasticsearch 0.90 does, so we hack it in.
+            @value[:minimum_should_match] = 1
+          end
```
elzj: There's no good reason for this whole method to be duplicated, but we can make a note to remove it when we're whacking all the old Tire code.

PR #3206, opened 2017-12-11
AO3-5286 - Use ScheduledReindexJob in cucumber tests.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5286

## Purpose

This PR modifies the Searchable module so that it no longer automatically calls reindex_document in the test environment, and changes the cucumber tests to use the new "all indexing jobs have been run" step, which uses the ScheduledReindexJob object to more closely match the indexing behavior outside of the test environment.

## Testing

This basically only changes tests (plus a single line of code that only runs in the test environment), so no manual testing should be required.

## References

There are 11 automated test failures, as a result of [AO3-5281](https://otwarchive.atlassian.net/browse/AO3-5281) and [AO3-5284](https://otwarchive.atlassian.net/browse/AO3-5284). These can be fixed by merging in #3203 and #3205. I left them separate so that it's easier to see which changes belong to which pull request, but I can merge them if that would be preferable.

### DIFF
--- a/features/admins/admin_works.feature
+++ b/features/admins/admin_works.feature
@@ -29,11 +29,11 @@ Feature: Admin Actions for Works and Bookmarks
     When I am logged in as an admin
       And I view the work "ToS Violation"
       And I follow "Hide Work"
-      And the work indexes are updated
+      And all indexing jobs have been run
     Then I should see "Item has been hidden."
       And all emails have been delivered
     When I follow "Make Work Visible"
-      And the work indexes are updated
+      And all indexing jobs have been run
     Then I should see "Item is no longer hidden."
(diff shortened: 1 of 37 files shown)


elzj:
Just re-running the tests on Travis now that the other branches are merged.

tickinginstant:
It looks like it's passing the tests that it failed before, but is now failing a gift exchange test relating to time zones. The same error seems to be occurring on a couple of other pull requests, though, so I think it's unrelated?

elzj:
Yep, that one's because of operating system upgrades, so I'll merge this and we'll figure out what to do with that.

PR #3207, opened 2017-12-14
AO3-5278: Fix pseud search error when logged out

elzj:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5278

## Purpose

Fixes 500 error when searching people, which happened because values were set to nil for users with no public works.

## Testing

See issue.


redsummernight:
The failed test is unrelated and will be addressed in [AO3-5289](https://otwarchive.atlassian.net/browse/AO3-5289).

PR #3208, opened 2017-12-14
AO3-5288 Update script to generate multiple admins at a time

ariana-paris:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5288

## Purpose

Updates the admin creation script to make it slightly easier to create more than one at a time.

## Testing

This is a backend task that can only be run by someone with ssh access to the AO3 servers. (It has already been run multiple times on production as this is the script I've used the last few times I created admins there)

### DIFF
--- a/script/create_admin.rb
+++ b/script/create_admin.rb
@@ -1,20 +1,52 @@
 #!script/rails runner
-# usage:
-# RAILS_ENV=production script/create_admin.rb
+# 
+# 1. Run this script on production as follows:
+#      RAILS_ENV=production bundle exec rails r script/create_admin.rb
+# 2. When prompted, type or paste in the admins, one per line, in the format USERNAME, EMAIL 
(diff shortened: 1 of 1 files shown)


script/create_admin.rb
```diff
@@ -1,20 +1,52 @@
...
+
+  if a.save
+    puts
+    puts "username: admin-#{name}"
+    puts "password: #{password}"
+    puts "http://archiveofourown.org/admin/login"
+    puts
+    admins << a
+  else
+    y a.errors.full_messages
```
redsummernight: `y` is [not around](https://stackoverflow.com/q/11571801) in our current Ruby/Rails setup, can you change to `puts`?

ariana-paris: Annoyingly, I'd already fixed that (since this obviously didn't work last time I ran it on production) but apparently forgot to commit 🤦‍♀️ Now fixed

PR #3213, opened 2017-12-31
AO3-5276 Add total number of works to results when searching for fandom

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5276

## Purpose

If you search people by fandom(s), the results should now display as "50 works, 33 bookmarks | 14 works in Thor (Movies), 1 work in Twilight (Movies)"

## Testing

Refer to JIRA

### DIFF
--- a/app/decorators/pseud_decorator.rb
+++ b/app/decorators/pseud_decorator.rb
@@ -96,7 +96,7 @@ class PseudDecorator < SimpleDelegator
       fandom_links = [options[:fandom_id]].flatten.map do |fandom_id|
         fandom_link(fandom_id)
       end
-      fandom_links.compact.join(", ") + " | " + general_links
+      general_links  +  " | " + fandom_links.compact.join(", ")
     else
       general_links
     end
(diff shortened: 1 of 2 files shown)


app/decorators/pseud_decorator.rb
```diff
@@ -90,16 +90,16 @@ def fandom_link(fandom_id)
...
   def authored_items_links(options = {})
+    general_links = [works_link, bookmarks_link].compact.join(", ")
     if options[:fandom_id].present?
       # This can potentially be an array
-      links = [options[:fandom_id]].flatten.map do |fandom_id|
+      fandom_links = [options[:fandom_id]].flatten.map do |fandom_id|
         fandom_link(fandom_id)
       end
-      links << bookmarks_link
+      general_links + " | " + fandom_links.compact.join(", ")
```
redsummernight: Do we need to handle cases where either `general_links` or `fandom_links` is empty? This doesn't affect the old code since we only did a join in the end.

sarken: I don't think so. I didn't run into any errors when search for pseuds who had no works and/or no bookmarks, and if `fandom_links` is blank, the pseud won't be included in the results for a fandom search.

PR #3217, opened 2018-01-05
AO3-5291 Don't delete false bookmark search terms.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5291

## Purpose

This fixes another place in the new bookmark search code that disregards values that have been deliberately set to false.

## Testing

Detailed in the bug report.


redsummernight:
Super minor: can you rename the test file to `bookmark_search_form_spec.rb`?

PR #3219, opened 2018-01-06
AO3-5309 Remove bookmarkable suffix in IndexSweeper.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5309

## Purpose

This PR is intended to make the IndexSweeper's retries work properly with the three BookmarkableIndexer subclasses: BookmarkedWorkIndexer, BookmarkedSeriesIndexer, and BookmarkedExternalWorkIndexer.

Previously, each retry in IndexSweeper would append an extra suffix, resulting in incorrect failure handling. This pull request overrides BookmarkableIndexer.find_elasticsearch_ids to make sure that the suffix is stripped in indexSweeper before the ID is passed along to a BookmarkableIndexer and the suffix is re-added.

## Testing

This one is very difficult to test manually. I think you'd have to deliberately modify the code to produce an error, and then watch the failure stores. One way to consistently produce errors is to modify the line `@batch << document(object)` in Indexer.batch to read `@batch << document(object).merge("_id" => "asdf")`, since Elasticsearch doesn't like it when you try to embed the "_id" in your document.

### DIFF
--- a/spec/models/async_indexer_spec.rb
+++ b/spec/models/async_indexer_spec.rb
@@ -42,4 +42,44 @@ describe AsyncIndexer do
     AsyncIndexer.perform("WorkIndexer:34:#{Time.now.to_i}")
   end
 
+  context "when persistent failures occur" do
+    before do
+      # Make elasticsearch always fail.
+      allow($new_elasticsearch).to receive(:bulk) do |options|
+        {
+          "errors" => true,
+          "items" => options[:body].map do |line|
+            action = line.keys.first
+            next unless (id = line[action]["_id"])
+
+            {
+              action.to_s => {
+                "_id" => id,
+                "error" => { "an error" => "with a message" }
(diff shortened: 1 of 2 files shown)


spec/models/async_indexer_spec.rb
```diff
@@ -42,4 +42,44 @@
     AsyncIndexer.perform("WorkIndexer:34:#{Time.now.to_i}")
   end
 
+  context "when persistent failures occur" do
+    before do
+      # Make elasticsearch always fail.
+      allow($new_elasticsearch).to receive(:bulk) do |options|
+        {
+          "errors" => true,
```
redsummernight: 👍

For the record, this is the error output to fake: https://www.elastic.co/guide/en/elasticsearch/reference/6.0/docs-bulk.html

PR #3220, opened 2018-01-06
AO3-5295: Never count a work with one fandom as a crossover

elzj:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5295

## Purpose

Tweaks crossover index logic so that a work with one fandom is never counted as a crossover, as that's proved to be technically possible via tag wrangling.

## Testing

See issue.

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -1576,7 +1576,7 @@ class Work < ApplicationRecord
 
   # Does this work have only one relationship tag?
   def otp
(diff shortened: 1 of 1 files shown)


sarken:
Marking this Needs Coder Action because, per conversation in Slack, we want to fix the OTP filter as well.

ariana-paris:
If fixing the OTP filter is holding this up, should we just open a new ticket and do that separately so this can be merged? This ticket is only about the fandom anyway.

PR #3221, opened 2018-01-09
AO3-5311 Add user's rollout info to support tickets

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5311

## Purpose

Adds a method for passing rollout information to Support (it's kind of unnecessarily complex because we might use rollout for two features, even though we do not currently). Uses said method to pass Elasticsearch version info to both the ticket tracker and mailing list.

## Testing

Refer to Jira.

### DIFF
--- a/app/models/feedback.rb
+++ b/app/models/feedback.rb
@@ -55,7 +55,7 @@ class Feedback < ApplicationRecord
     string = ""
     # ES UPGRADE TRANSITION #
     # Remove ES version logic, but leave this method for future rollout use
-    string << if use_new_search?
+    string << if Feedback.use_new_search?
                 "ES 6.0"
               else
                 "ES 0.90"
--- a/app/models/feedback.rb
+++ b/app/models/feedback.rb
@@ -55,12 +55,11 @@ class Feedback < ApplicationRecord
     string = ""
     # ES UPGRADE TRANSITION #
     # Remove ES version logic, but leave this method for future rollout use
-    if $rollout.active?(:use_new_search) || User.current_user.present? && $rollout.active?(:use_new_search, User.current_user)
-      string << "ES 6.0"
-    else
-      string << "ES 0.90"
-    end
-    string
+    string << if $rollout.active?(:use_new_search) || User.current_user.present? && $rollout.active?(:use_new_search, User.current_user)
+                "ES 6.0"
+              else
+                "ES 0.90"
+              end
   end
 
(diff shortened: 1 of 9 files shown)


app/views/admin_mailer/feedback.html.erb
```diff
@@ -1,21 +1,27 @@
...
+  <% end %>
+  <% unless @feedback.user_agent.blank? %>
+    <p>User agent: <%= raw @feedback.user_agent %></p>
+  <% end %>
 
 
   Feedback:
-  <p><b><%= (raw @summary) %></b></p>
-  <p><%= (raw @comment) %></p>
+  <b><%= (raw @feedback.summary) %></b>
```
sarken: The summary and comment already have their own paragraph tags because they go through the parser/sanitizer, so we don't need the ones in the template.

app/models/feedback.rb
```diff
@@ -51,6 +51,17 @@ def email_and_send
     send_report
   end
 
+  def rollout_string
+    string = ""
+    # ES UPGRADE TRANSITION #
+    # Remove ES version logic, but leave this method for future rollout use
+    string << if $rollout.active?(:use_new_search) || User.current_user.present? && $rollout.active?(:use_new_search, User.current_user)
```
redsummernight: This can be `if use_new_search?`, inherited from `app/models/application_record.rb`.

sarken: Sadly, it wasn't able to access the method when I tried that.

sarken: I edited the file here so the tests can run, in case it was just my webdev being funny, which is always a possibility.

app/views/feedbacks/new.html.erb
```diff
@@ -74,6 +74,7 @@
     <legend><%= ts("Send Your Feedback") %></legend>
     <%= f.hidden_field :user_agent, value: request.env["HTTP_USER_AGENT"] %>
     <%= f.hidden_field :ip_address, value: request.remote_ip %>
+    <%= f.hidden_field :rollout, value: @feedback.rollout_string %>
```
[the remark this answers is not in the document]
sarken: Yeah, I figure we don't care so much because like you said, it's already possible to do that with the IP and user agent. If someone changes one of those, all they accomplish is possibly getting themselves an unhelpful answer that doesn't pertain to them.

PR #3224, opened 2018-01-17
AO3-5265 Change tag search back to querystring and ensure exact results are near top

ariana-paris:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5265

## Purpose

This PR amends the tag search and query code such that the following examples are true:

* performs a case-insensitive search (AbC matches abc)
* performs a query string search (ab or cd matches ab cd)
* lists exact matches (without punctuation) at the top of the results (abc result lists abc or abc+ first)
* matches every token (d abc matches abc d and abc-d, but not abc or abc+)
* performs a wildcard search at the end of a term (abc* matches abcd and abcde)
* performs a wildcard search in the middle of a term (a*d matches abcd)
* performs a wildcard search at the beginning of a term (*cd matches abcd)
* preserves plus (+) character (abc+ matches abc+ and abc, but not abccc)
* preserves minus (-) character (abc-d matches abc-d, abc -d, abc d but not abc or abcd)
* preserves minus (-) preceded by a space (abc -d matches abc -d, abc d and abc-d, but not abc)
* preserves slashes without quotes (ab/cd should match ab/cd and ab cd)
* matches tags with canonical punctuation (yuri!!! on ice matches Yuri!!! On Ice)
* matches tags without canonical punctuation (yuri on ice matches Yuri!!! On Ice)

## Testing

See above and on Jira ticket.

### DIFF
--- a/app/models/search/indexer.rb
+++ b/app/models/search/indexer.rb
@@ -44,7 +44,7 @@ class Indexer
           index: {
             number_of_shards: 5,
           }
-        },
+        }.merge(settings),
         mappings: mapping,
       }
     )
@@ -61,14 +61,24 @@ class Indexer
 
   def self.mapping
     {
-      document_type => {
+      document_type: {
         properties: {
-          #add properties in subclasses
+          # add properties in subclasses
         }
       }
     }
   end
 
+  def self.settings
+    {
+      analyzer: {
+        custom_analyzer: {
+          # add properties in subclasses
+        }
+      }
+    }
+  end
+  
   def self.index_all(options={})
     unless options[:skip_delete]
       delete_index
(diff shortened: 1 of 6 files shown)


sarken:
Question! Do you know how this feels about tags like `Dave ♦ Sawbuck`?

ariana-paris:
I added a spec test for this and also double-checked manually on my local environment, and that tag is handled fine too, and returned for searches on `Dave ♦ Sawbuck`, `dave sawbuck`, `sawbuck` and `dave` - though sadly not `♦` which ES doesn't seem to index (maybe a ticket for another time)

sarken:
Yay, sounds good! "Returns more things than it should" is what the wranglers are used to on 0.90.7; the original 6.0 upgrade made that search return nothing, which was going to be a problem for TW. (https://otwarchive.atlassian.net/browse/AO3-3762)

PR #3225, opened 2018-01-17
AO3-5303 People search bookmark count should respect bookmarkable privacy

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5303

## Purpose

Currently, people search results are telling _everyone_, logged in or out, how many public bookmarks a user has of _both_ public and restricted items (works, series, external works). This makes it so logged out users see a total that only includes public bookmarks of public items, while logged in users continue to see a total that includes public bookmarks of both public and restricted items.

## Testing

Refer to Jira

## References

Builds off #3223 -- you could honestly just merge this and close that, as long as you update both issues.

### DIFF
--- a/app/decorators/pseud_decorator.rb
+++ b/app/decorators/pseud_decorator.rb
@@ -16,10 +16,11 @@ class PseudDecorator < SimpleDelegator
     work_counts
     bookmark_counts
     work_key = User.current_user.present? ? :general_works_count : :public_works_count
+    bookmark_key = User.current_user.present? ? :general_bookmarks_count : :public_bookmarks_count
     pseuds.map do |pseud|
       data = {
         user_login: users[user_id].login,
-        public_bookmarks_count: bookmark_counts[id],
+        bookmark_key => bookmark_counts[id],
         work_key => work_counts[id]
       }
       new_with_data(pseud, data)
@@ -42,7 +43,7 @@ class PseudDecorator < SimpleDelegator
   end
 
   def bookmarks_count
-    data[:public_bookmarks_count]
+    User.current_user.present? ? data[:general_bookmarks_count] : data[:public_bookmarks_count]
   end
 
   def byline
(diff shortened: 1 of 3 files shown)


app/decorators/pseud_decorator.rb
```diff
@@ -42,7 +43,7 @@ def works_count
   end
 
   def bookmarks_count
-    data[:public_bookmarks_count]
+    User.current_user.present? ? data[:general_bookmarks_count] : data[:public_bookmarks_count]
```
redsummernight: Should we do `count || 0` like works_count in case of nil?

sarken: I'm not really sure!

@elzj was there a reason the `|| 0` was necessary for works but not bookmarks in the initial code? Or was it just an oversight?

redsummernight: That was #3207. I don't know if it's different for bookmarks.

sarken: It looks like if a user has no bookmarks, both the general and public bookmark count are properly set to 0 instead of nil and no 500 error occurs:

```
"_index": "ao3_development_pseuds",
        "_type": "pseud",
        "_id": "318811",
        "_score": 1,
        "_source": {
          "id": 318811,
          "user_id": 293707,
          "name": "testy4",
          "description": null,
          "created_at": "2014-02-25T05:10:19.000-05:00",
          "user_login": "testy4",
          "byline": "testy4",
          "collection_ids": [],
          "sortable_name": "testy4",
          "fandoms": [
            {
              "id": 6670,
              "name": "Testing",
              "count": 1
            }
          ],
          "general_bookmarks_count": 0,
          "public_bookmarks_count": 0,
          "general_works_count": 1,
          "public_works_count": 1
        }
```

redsummernight: I think I see the difference:

- for work counts, we do `pseud.works.where(countable_works_conditions).group(:restricted).count`, get a hash, access the 'false' key for public works, which can be nil
- for bookmark counts, it's always count queries without grouping so we don't have that problem

PR #3228, opened 2018-01-20
AO3-5301 Search form should search bookmarkable tags and bookmark tags

ariana-paris:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5301

## Purpose

Makes the bookmark search form look for bookmarks whose bookmarkable has a given tag (current behaviour on production), rather than the bookmarks.

## Testing

See Jira.


app/models/search/bookmark_query.rb
```diff
@@ -40,12 +40,49 @@ def exclusion_filters
     @exclusion_filters ||= tag_exclusion_filter
   end
 
-  # Instead of doing a standard query, which would only match bookmark fields
-  # we'll make this a should query that will try to match either the bookmark or its parent
-  def should_query
```
redsummernight: https://github.com/otwcode/otwarchive/blob/f274643b0dcca0f4a21f00d23c0f59b2b8096ffd/app/models/search/query.rb#L85-L88

The method `should_query` is still around in the parent class, and used by bookmarkable_query. Should it be removed in those classes?

ariana-paris: Thanks for taking a look. I was wary of just ripping it out from the parent class and bookmarkable in case they still served a purpose there - though since the one in bookmarkable will never be used because the one in bookmark is gone, and that's the only one remaining, I guess that kind of answers the question 🤔 I'll have to take another look in the morning to make sure I'm reading this right before ripping it out.

ariana-paris: I've pushed a new version without that code and we'll at least see if all the tests pass (the bookmark specs do locally, at least)

ariana-paris:
Superseded by a change of plan

PR #3230, opened 2018-01-28
AO3-5325 Reindex pseuds when bookmarks are hidden

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5325

## Purpose

Reindexes pseuds if an admin hides (or unhides) a bookmark.

## Testing

Refer to JIRA

### DIFF
--- a/app/models/bookmark.rb
+++ b/app/models/bookmark.rb
@@ -110,7 +110,7 @@ class Bookmark < ApplicationRecord
   # We index the bookmark count, so if it should change, update the pseud
   def update_pseud_index
     return unless $rollout.active?(:start_new_indexing)
-    return unless destroyed? || saved_change_to_id? || saved_change_to_private?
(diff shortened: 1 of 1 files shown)


app/models/bookmark.rb
```diff
@@ -110,7 +110,7 @@ def invalidate_bookmark_count
   # We index the bookmark count, so if it should change, update the pseud
   def update_pseud_index
     return unless $rollout.active?(:start_new_indexing)
-    return unless destroyed? || saved_change_to_id? || saved_change_to_private?
+    return unless destroyed? || saved_change_to_id? || saved_change_to_private? || saved_change_to_hidden_by_admin?
```
redsummernight: This covers fields similar to the work counterpart:

https://github.com/otwcode/otwarchive/blob/abab68178842662391095acdc206f1b114ab5649/app/models/work.rb#L271-L275

so I don't think anything else is missing.

PR #3231, opened 2018-01-29
AO3-5324 Re-allow hiding of series by admin

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5324

## Purpose

- Adds Series to the whitelist to prevent the 500 error that was stopping us from hiding them
- Makes it so the button to hide a series correctly says Hide Series (instead of Hide Work)
- Add a red lock to the series show page to make it clearer to the admin and the series owners that the series has been hidden

## Testing

Log in as an admin, press the "Hide Series" button on a series, and confirm that a success message and a red lock icon appear. Use the "Make Series Visible" icon and confirm that a success message appears and the red lock icon goes away.

### DIFF
--- a/app/controllers/admin/user_creations_controller.rb
+++ b/app/controllers/admin/user_creations_controller.rb
@@ -31,17 +31,6 @@ class Admin::UserCreationsController < ApplicationController
     elsif @creation_class == ExternalWork || @creation_class == Bookmark
       redirect_to(request.env["HTTP_REFERER"] || root_path)
     else
-      unless action == "unhide" || @creation_class != Work
-        # Email users so they're aware of Abuse action
-        # Emails for works are handled in the work class
-        orphan_account = User.orphan_account
-        users = @creation.pseuds.map(&:user).uniq
-        users.each do |user|
-          unless user == orphan_account
-            UserMailer.admin_hidden_work_notification(@creation.id, user.id).deliver
-          end
-        end
-      end
       redirect_to(@creation)
     end
   end  
--- a/app/controllers/admin/user_creations_controller.rb
+++ b/app/controllers/admin/user_creations_controller.rb
@@ -5,7 +5,7 @@ class Admin::UserCreationsController < ApplicationController
   before_action :can_be_marked_as_spam, only: [:set_spam]
 
   def get_creation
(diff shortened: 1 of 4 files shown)


app/controllers/admin/user_creations_controller.rb
```diff
@@ -31,17 +31,6 @@ def hide
     elsif @creation_class == ExternalWork || @creation_class == Bookmark
       redirect_to(request.env["HTTP_REFERER"] || root_path)
     else
-      unless action == "unhide" || @creation_class == Work
```
sarken: Although the comment above the method says "creations," `UserMailer.admin_hidden_work_notification` is _only_ for works: https://github.com/otwcode/otwarchive/blob/7570b001476c05306fe9b47172fdfd6432dc02ee/app/mailers/user_mailer.rb#L359 

Since sending that email is handled here: https://github.com/otwcode/otwarchive/blob/58c1dd964fc6b7a2b6ae19adabc9ea59a00da965/app/models/work.rb#L1454

There's no reason to leave this hanging around, unless we enjoy errors when hiding anything that isn't a Work, Bookmark, or External Work.

tickinginstant: Is it worth it to leave parts of this code here (maybe commented out, or in an unreachable block) to make it easier for someone to eventually fix [AO3-4314](https://otwarchive.atlassian.net/browse/AO3-4314)? Or is it better to keep the code clean and have whoever ends up tackling that bug rewrite it from scratch?

sarken:
I'm more inclined to let them start from scratch -- better logic, different mailer -- assuming we decide to add an email. (Abuse doesn't generally need to hide series, hence not even realizing we'd broken this some time ago. And when they do hide or delete things, they always contact users directly and don't just rely on the automated email.)

app/controllers/admin/user_creations_controller.rb
sarken: Grumble, grumble, apparently replying by email doesn't put replies in threads. Deleting and pasting here:

I'm more inclined to let them start from scratch -- better logic, different mailer -- assuming we decide to add an email. (Abuse doesn't generally need to hide series, hence not even realizing we'd broken this some time ago. And when they do hide or delete things, they always contact users directly and don't just rely on the automated email.)

PR #3232, opened 2018-01-29
AO3-4923 Expand bookmark controller test

cesy:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4923

## Purpose
Adds more tests. See also #3139. I'm also working on further improvements to test fetching recent bookmarks with JS.

### DIFF
--- a/factories/chapters.rb
+++ b/factories/chapters.rb
@@ -3,7 +3,7 @@ require 'faker'
 FactoryBot.define do
   factory :chapter do
     content { "Awesome content!" }
-	title { "Cool chapter title" }
+    title { "Cool chapter title" }
     work
 
     transient do
--- a/factories/chapters.rb
+++ b/factories/chapters.rb
@@ -3,7 +3,6 @@ require 'faker'
 FactoryBot.define do
   factory :chapter do
     content { "Awesome content!" }
-    title { "Cool chapter title" }
     work
 
     transient do
--- a/factories/chapters.rb
+++ b/factories/chapters.rb
@@ -3,6 +3,7 @@ require 'faker'
 FactoryBot.define do
   factory :chapter do
     content { "Awesome content!" }
+	title { "Cool chapter title" }
     work
 
     transient do
--- a/spec/controllers/bookmarks_controller_spec.rb
+++ b/spec/controllers/bookmarks_controller_spec.rb
@@ -28,7 +28,7 @@ describe BookmarksController do
       it "with chapter" do
         fake_login
         get :new, params: { chapter_id: chapter2.id }
-        expect(response).to render_template('new')
+        expect(response).to render_template("new")
         expect(assigns(:bookmarkable)).to eq(chaptered_work)
       end
     end
@@ -217,7 +217,7 @@ describe BookmarksController do
     let(:bookmark) { create(:bookmark, bookmarkable_id: chaptered_work.id) }
     # render_views check why I needed this
 
-    it "should find a work from a bookmark" do
+    it "finds a work from a bookmark" do
       fake_login_known_user(bookmark.pseud.user)
       get :show, params: { id: bookmark }
       expect(response).to have_http_status(:success)
--- a/spec/controllers/bookmarks_controller_spec.rb
+++ b/spec/controllers/bookmarks_controller_spec.rb
@@ -212,17 +212,15 @@ describe BookmarksController do
   end
 
   describe "show" do
-    let(:chaptered_work) { create(:work) }
+    let(:chaptered_work) { create(:work, title: "Cool title") }
     let(:chapter2) { create(:chapter, work: chaptered_work, position: 2, posted: true, title: "Second title") }
     let(:bookmark) { create(:bookmark, bookmarkable_id: chaptered_work.id) }
-    render_views # to see the title appear in the right place
 
-    it "finds a work from a bookmark" do
+    it "shows a bookmark" do
       fake_login_known_user(bookmark.pseud.user)
       get :show, params: { id: bookmark }
       expect(response).to have_http_status(:success)
       expect(assigns(:bookmark)).to eq(bookmark)
-      expect(response.body).to include(chaptered_work.title)
     end
   end
 end
--- a/spec/controllers/bookmarks_controller_spec.rb
+++ b/spec/controllers/bookmarks_controller_spec.rb
@@ -48,7 +48,7 @@ describe BookmarksController do
         fake_login
         get :new, params: { chapter_id: chapter2.id, format: :js }, xhr: true
         expect(response).to render_template("bookmark_form_dynamic")
-        expect(assigns(:bookmarkable)).to eq(chapteredwork)
+        expect(assigns(:bookmarkable)).to eq(chaptered_work)
       end
     end
   end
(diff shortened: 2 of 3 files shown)


spec/controllers/bookmarks_controller_spec.rb
```diff
@@ -10,6 +11,10 @@ def it_redirects_to_user_login
 
   describe 'new' do
     context 'without javascript' do
+      let(:chapteredwork) { create(:work) }
```
sarken: Could you add an underscore between words to bring this more in line with our variable naming conventions?

spec/controllers/bookmarks_controller_spec.rb
```diff
@@ -42,4 +65,19 @@ def it_redirects_to_user_login
       end
     end
   end
+
+  describe "show" do
+    let(:chapteredwork) { create(:work) }
+    let(:chapter2) { create(:chapter, work: chapteredwork) }
+    let(:bookmark) { create(:bookmark, bookmarkable_id: chapteredwork.id) }
+    render_views
```
sarken: We generally avoid `render_views` in our specs -- if something needs the view, we tend to prefer to use features. Is there a specific untested line in the controller you're trying to hit here?

factories/chapters.rb
```diff
@@ -3,6 +3,7 @@
 FactoryGirl.define do
   factory :chapter do
     content "Awesome content!"
+    title "Cool chapter title"
```
sarken: Is it necessary to have a title for all chapters we test, or can we just set the title in the specific tests that need it?

spec/controllers/bookmarks_controller_spec.rb
```diff
@@ -20,20 +25,38 @@ def it_redirects_to_user_login
         get :new
         expect(response).to render_template('new')
       end
+
+      it 'with chapter' do
+        fake_login
+        get :new, params: { chapter_id: chapter2.id}
```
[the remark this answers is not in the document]
sarken: I'm drawing a blank here... what would this URL look like? Where do we use it?

factories/chapters.rb
cesy: I tried setting it just in those tests, and couldn't get it to work that way.

spec/controllers/bookmarks_controller_spec.rb
cesy: Will do.

spec/controllers/bookmarks_controller_spec.rb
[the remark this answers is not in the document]
cesy: As far as I can tell, if you're on a chapter other than the first chapter of a work, and you create a bookmark. However, I can't actually reproduce it by testing - I haven't managed to get to the bookmarks/new URL at all, as it's all done within the bookmark form in a page.

sarken:
Just noting that the Travis failure appears to be an actual test failure, hence the Needs Coder Action label remaining despite the coder actioning the previous review.

factories/chapters.rb
```diff
@@ -3,6 +3,7 @@
 FactoryBot.define do
   factory :chapter do
     content { "Awesome content!" }
+	title { "Cool chapter title" }
```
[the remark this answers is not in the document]
cesy: Fixed

spec/controllers/bookmarks_controller_spec.rb
```diff
@@ -20,20 +24,38 @@ def it_redirects_to_user_login
         get :new
         expect(response).to render_template("new")
       end
+
+      it "with chapter" do
+        fake_login
+        get :new, params: { chapter_id: chapter2.id }
+        expect(response).to render_template('new')
```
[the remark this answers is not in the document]
cesy: Fixed

spec/controllers/bookmarks_controller_spec.rb
```diff
@@ -188,4 +210,19 @@ def it_redirects_to_user_login
...
     end
   end
+
+  describe "show" do
+    let(:chaptered_work) { create(:work) }
+    let(:chapter2) { create(:chapter, work: chaptered_work) }
+    let(:bookmark) { create(:bookmark, bookmarkable_id: chaptered_work.id) }
+    # render_views check why I needed this
+
+    it "should find a work from a bookmark" do
```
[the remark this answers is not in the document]
cesy: Fixed

factories/chapters.rb
cesy: Will fix.

spec/controllers/bookmarks_controller_spec.rb
cesy: Will fix.

spec/controllers/bookmarks_controller_spec.rb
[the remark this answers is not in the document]
cesy: Removed this and found another way to do it.

spec/controllers/bookmarks_controller_spec.rb
```diff
@@ -10,7 +10,11 @@ def it_redirects_to_user_login
 
   describe "new" do
     context "without javascript" do
-      it "redirects anyone not logged in" do
+      let(:chaptered_work) { create(:work) }
+      let(:chapter2) { create(:chapter, work: chaptered_work) }
+      let(:bookmark) { create(:bookmark, bookmarkable_id: chaptered_work.id) }
+
+      it "redirects and should not return the form for anyone not logged in" do
```
sarken, line 17: I think we can just use `it "redirects logged out users" do` here, since we're not explicitly checking the form isn't rendered. If we wanted more specificity, we could say `it "redirects logged out users to the user login page" do`

spec/controllers/bookmarks_controller_spec.rb
```diff
@@ -20,20 +24,38 @@ def it_redirects_to_user_login
         get :new
         expect(response).to render_template("new")
       end
+
+      it "with chapter" do
```
sarken, line 28: This looks like a `context` description, not a description of the expected behavior. Line 47 also has the same issue.

spec/controllers/bookmarks_controller_spec.rb
```diff
@@ -188,4 +210,17 @@ def it_redirects_to_user_login
       end
     end
   end
+
+  describe "show" do
+    let(:chaptered_work) { create(:work, title: "Cool title") }
+    let(:chapter2) { create(:chapter, work: chaptered_work, position: 2, posted: true, title: "Second title") }
+    let(:bookmark) { create(:bookmark, bookmarkable_id: chaptered_work.id) }
+
+    it "shows a bookmark" do
```
sarken, line 219: How about `it "returns bookmark"` to go with the phrasing used elsewhere in this file? It would probably also make sense to give this some context descriptions to make it easier for people to slot in additional tests in the future, e.g.:
```
context "when logged in as owner of bookmarker" do
  context "when bookmark is public" do
    context "when bookmarkable is chaptered public work" do
      it "returns bookmark" do
        etc
      end
    end
    
    context "when bookmarkable is work hidden by admin" do
      it "does something" do
        etc
      end
    end
  end

  context "when bookmark is private" do
    context "when bookmarkable is a restricted series" do
      it "does something" do
        etc
      end
    end
  end
end

context "when logged in as admin" do
  context "when bookmark is private" do
    context "when bookmarkable is a restricted series" do
      it "does something" do
        etc
      end
    end
  end
end
```

spec/controllers/bookmarks_controller_spec.rb
cesy: Agreed

spec/controllers/bookmarks_controller_spec.rb
cesy: Fixed

spec/controllers/bookmarks_controller_spec.rb
cesy: Good idea, and I might add some more of those later.

PR #3120, opened 2017-10-25
AO3-5190 Abuse and support form language on error

Meepu:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5190

## Purpose

When submitting an abuse or support form with an error (such as an invalid email address), chosen language should persist.

## Testing

Follow the "Report Abuse" link in the footer
Select a language other than English
Fill in required information, but in the URL field, enter a URL that will make the form error (e.g. enter http://test.ao3.org)
Submit

PR #3235, opened 2018-01-31
AO3-5190 Abuse and support form language on error

Meepu:
This PR replaces https://github.com/otwcode/otwarchive/pull/3120, which got really complicated because of merge conflicts.

## Issue

https://otwarchive.atlassian.net/browse/AO3-5190

## Purpose

When submitting an abuse or support form with an error (such as an invalid email address), chosen language should persist.

## Testing

Follow the "Report Abuse" or "Technical Support and Feedback" link in the footer
Select a language other than English
Fill in required information, but in the URL field, enter a URL that will make the form error (e.g. enter http://test.ao3.org) or make another mistake such as entering an invalid email address
Submit

Includes two Cucumber tests

### DIFF
--- a/features/other_a/abuse_report.feature
+++ b/features/other_a/abuse_report.feature
@@ -56,7 +56,7 @@ Feature: Filing an abuse report
     And I select "Deutsch" from "abuse_report_language"
     And I press "Submit"
     And I should see "Email does not seem to be a valid address."
-    And the field labeled "abuse_report_language" should contain "Deutsch"
+    And "Deutsch" should be selected within "Select language (required)"
   Then I fill in "Your email" with "valid@archiveofourown.org"
     And I press "Submit"
     And I should see "Your abuse report was sent to the Abuse team."
--- a/features/other_a/abuse_report.feature
+++ b/features/other_a/abuse_report.feature
@@ -53,10 +53,11 @@ Feature: Filing an abuse report
     And I fill in "Your comment (required)" with "This is wrong"
     And I fill in "Link to the page you are reporting" with "http://www.archiveofourown.org/works"
     And I fill in "Your email (required)" with ""
+    And I select "Deutsch" from "abuse_report_language"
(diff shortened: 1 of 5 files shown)


features/other_a/abuse_report.feature
```diff
@@ -53,10 +53,11 @@ Feature: Filing an abuse report
     And I fill in "Your comment (required)" with "This is wrong"
     And I fill in "Link to the page you are reporting" with "http://www.archiveofourown.org/works"
     And I fill in "Your email (required)" with ""
+    And I select "Deutsch" from "abuse_report_language"
     And I press "Submit"
     And I should see "Email does not seem to be a valid address."
+    And the field labeled "abuse_report_language" should contain "Deutsch"
```
redsummernight: This will pass without your fix because it's checking if "Deutsch" appears in the select, not if it's selected. You should use the step:

```
And "Deutsch" should be selected within "Select language (required)"
```

features/other_b/support.feature
```diff
@@ -44,3 +44,16 @@ Feature: Filing a support request
...
+  When I am on the home page
+    And basic languages
+    And I follow "Support and Feedback"
+  When I select "Deutsch" from "feedback_language"
+    And I fill in "Brief summary" with "Just a brief note"
+    And I fill in "Your comment" with "Men have their old boys' network, but we have the OTW. You guys rock!"
+    And I fill in "Your email (required)" with "a"
+    And I press "Send"
+  Then I should see "Sorry, your message could not be saved"
+    And the field labeled "feedback_language" should contain "Deutsch"
```
redsummernight: Instead of adding a new scenario you could also build off the existing one ("Scenario: Not logged in, with and without email", which has an error on submission part), the way you added to the abuse report scenario.

Meepu:
Thanks for the review, redsummernight! I made the fixes you requested.

PR #3236, opened 2018-02-03
AO3-5326 Reindex bookmarker pseuds when bookmarkables change restricted status

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5326

## Purpose

When a work or a series becomes (un)restricted, we need to reindex its bookmarkers so they get the correct public/general bookmark counts.

External works can't be restricted, so they're fine.

## Testing

See JIRA issue.

## References

The new specs run the indexing jobs (instead of totally refreshing ES), similar to the feature tests in #3206.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -172,6 +172,14 @@ def refresh_index_without_updating(klass_name)
   $new_elasticsearch.indices.refresh(index: "ao3_test_#{klass_name}s")
 end
 
+def run_all_indexing_jobs
+  %w[main background stats].each do |reindex_type|
+    ScheduledReindexJob.perform reindex_type
+  end
+  # Specs are too fast! Wait for ES...
+  sleep 1
+end
+
 def delete_index(index)
   # ES UPGRADE TRANSITION #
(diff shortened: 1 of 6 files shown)


spec/spec_helper.rb
```diff
@@ -172,6 +172,14 @@ def refresh_index_without_updating(klass_name)
   $new_elasticsearch.indices.refresh(index: "ao3_test_#{klass_name}s")
 end
 
+def run_all_indexing_jobs
+  %w[main background stats].each do |reindex_type|
+    ScheduledReindexJob.perform reindex_type
+  end
+  # Specs are too fast! Wait for ES...
+  sleep 1
```
tickinginstant: Instead of introducing a delay, would it work to refresh the indices using `refresh_index_without_updating`?

redsummernight: Thank you! I was about to up the delay 😂

tickinginstant: Glad to help! Fingers crossed that it works. The PR looks good otherwise.

PR #3234, opened 2018-01-29
AO3-5307 update series bookmarks when work is added to series

WendyBeth:
# Pull Request Checklist

* [x] Have you read [How to write the perfect pull request](https://github.com/blog/1943-how-to-write-the-perfect-pull-request)?
* [x] Have you read through the [contributor guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [x] Have you added tests for any changed functionality?
* [x] Have you added the JIRA issue number as the *first* thing in your pull request title (eg: `AO3-1234 Fix thing`)
* [x] Have you updated the JIRA issue with the information below?

## Issue

https://otwarchive.atlassian.net/browse/AO3-5307

## Purpose

What does this PR do?

This PR fixes a bug where, if a series has works all with fandom A and is bookmarked, and then a work is added to that series with fandom B, then the series bookmark should show up for bookmarks tagged with Fandom B. It also ensures that the updated_at order is preserved.

## Testing

How can the Archive's QA team verify that this is working as you intended? (If you have access, please copy this into the JIRA ticket for them!)

Create a series with a work with any Fandom A, the bookmark the work. Go the Fandom A bookmarks to ensure the series bookmark is listed there. Add a work with Fandom B to the series, and then go to the Fandom B bookmarks and ensure that the series bookmark is listed there as well. Add other bookmarks, and then go through the same process again and ensure that bookmarks for series which have just had new works appended to them show up first when sorting the bookmarks by 'Updated Date'.

## References

Are there any other relevant issues / PRs / mailing lists discussions related to this?

## Credit

*What name do you want us to use to credit your work in the [Archive of Our Own's Release Notes](http://archiveofourown.org/admin_posts?tag=1)?*

Wendy Randquist - Littlelines

*What pronouns do you prefer (she/her, he/him, zie/hir etc)?*

she/her

(if you do not fill this in, we will use your GitHub account name and will use "they" to avoid making assumptions about your gender)

PR #3239, opened 2018-02-05
AO3-5327 Fix the test that hides a series being bookmarked

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5327

## Purpose

~Without the reload call, update_attribute fails, with bookmarkable.errors containing an error with no messages. I don't know why this happens. Years from now, if you figure this out, please let me know.~

When a series and its work are first created, the series loads an empty collection of bookmarks, which stays unupdated when we pluck the bookmark IDs to reindex bookmarker pseuds, so no pseuds get reindexed. We need to reload the series.

## Testing

See JIRA issue.

## References

The test was introduced in #3238.

### DIFF
--- a/spec/models/pseud_search_form_spec.rb
+++ b/spec/models/pseud_search_form_spec.rb
@@ -138,8 +138,10 @@ describe PseudSearchForm do
         User.current_user = nil
         expect(result.bookmarks_count).to eq 1
 
+        # This reload is here because the following update_attribute fails otherwise
+        # I don't know why :(
+        bookmarkable.reload
         bookmarkable.update_attribute(:hidden_by_admin, true)
-        expect(bookmarkable.hidden_by_admin).to be_truthy
         run_all_indexing_jobs
         result = PseudSearchForm.new(name: bookmark.pseud.name).search_results.first
         expect(result).to eq bookmark.pseud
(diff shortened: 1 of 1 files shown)


tickinginstant:
As far as I can tell, the issue is that the code from #3234 loads all of the series' bookmarks on saving new SerialWorks. This means that when the series and its serial work are first created, the series loads an empty collection of bookmarks. In one of the Rails updates (not sure which), pluck was revised so that it no longer queries the database if the records have been loaded. So when the series is updated to set hidden_by_admin to true, it plucks the pseud_ids from the (already loaded) empty collection of bookmarks, and decides not to update any pseuds, even though a bookmark (with a pseud) has been added in the meantime.

I don't think it's an error with update_attribute itself.

redsummernight:
Thanks @tickinginstant!

The `pluck` change happens [in Rails 5](https://blog.bigbinary.com/2016/02/17/active-support-improvements-in-Rails-5.html#pluck). Do we need to somehow reload bookmarks when reindexing bookmarkable's bookmarkers? Or is this fix fine, because the issue only happens in this specific test setup? (In a real environment we would be hiding series like the feature tests in #3231).

tickinginstant:
Fixing just the test seems reasonable to me.

PR #3240, opened 2018-02-05
AO3-5330 Add an error page for blocked IP's

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5330

## Purpose

Adds a page that nginx can use to inform people that their access has been blocked.

## Testing

I think the only useful test is https://test.archiveofourown.org/403.html looks ok.

### DIFF
--- a/public/403.html
+++ b/public/403.html
@@ -1,0 +1,150 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <meta http-equiv="x-ua-compatible" content="ie=edge" />
+    <meta name="keywords" content="fanfiction, transformative works, otw, fair use, archive" />
+    <meta name="language" content="en-US" />
+    <meta name="subject" content="fandom" />
+    <meta name="description" content="An Archive of Our Own, a project of the
+    Organization for Transformative Works" />
+    <meta name="distribution" content="GLOBAL" />
+    <meta name="classification" content="transformative works" />
+    <meta name="author" content="Organization for Transformative Works" />
+    <title>
+      Exceeded maximum posting rate.
+    </title>
+    <link rel="stylesheet" type="text/css" media="screen" href="/stylesheets/skins/skin_873_archive_2_0/1_site_screen_.css">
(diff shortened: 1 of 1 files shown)


public/403.html
```diff
@@ -0,0 +1,150 @@
...
+    <meta name="keywords" content="fanfiction, transformative works, otw, fair use, archive" />
+    <meta name="language" content="en-US" />
+    <meta name="subject" content="fandom" />
+    <meta name="description" content="An Archive of Our Own, a project of the
+    Organization for Transformative Works" />
+    <meta name="distribution" content="GLOBAL" />
+    <meta name="classification" content="transformative works" />
+    <meta name="author" content="Organization for Transformative Works" />
+    <title>
+      Exceeded maximum posting rate.
```
sarken: Per chat, this should be: `Access Blocked | Archive of Our Own`

public/403.html
```diff
@@ -0,0 +1,150 @@
...
+            </form>
+          </li>
+        </ul>
+        <div class="clear"></div>
+      </div>
+      <!-- END header -->
+      <div id="inner" class="wrapper">
+        <div id="main" class="system errors error-403 region">
+          <h2 class="heading">Error 403</h2>
+          <h3 class="heading">Access Blocked.</h3>
```
sarken: Should be "blocked," lowercase b.

public/403.html
```diff
@@ -0,0 +1,150 @@
...
+          <p>We have blocked access to the Archive, these blocks are generally long term if you believe this is in error then <a href="http://www.transformativeworks.org/contact_us">you can contact us</a>.</p>
```
sarken: `Your IP address has been blocked from accessing the Archive of Our Own. These blocks are generally long-term. If you believe you have been blocked by mistake, please <a href="http://www.transformativeworks.org/contact_us">contact us</a>.`

redsummernight:
The menu items should be also fixed like [AO3-5335](https://otwarchive.atlassian.net/browse/AO3-5335). Since the 403 page is new, it's not covered by that issue's pull request (#3253). This may be out of scope though.

PR #3245, opened 2018-02-16
AO3-4310 Translate invite request decline email

cesy:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4310

## Purpose

Makes the invite request declined email translatable

## Testing

Request an invite, get an admin to decline it, get an email notification, it shows in English with no errors.

## Credit

Cesy, she/her

### DIFF
--- a/app/views/user_mailer/invite_request_declined.text.erb
+++ b/app/views/user_mailer/invite_request_declined.text.erb
@@ -4,6 +4,6 @@
 <%= t('.reason') %>
 <%= to_plain_text(raw @reason) %>
 
-<%= t '.signoff' %>
+<%= t('.signoff') %>
 <%= ArchiveConfig.APP_SHORT_NAME %>
 <% end %>
--- a/app/views/user_mailer/invite_request_declined.text.erb
+++ b/app/views/user_mailer/invite_request_declined.text.erb
@@ -1,9 +1,9 @@
 <% content_for :message do %>
-<%= ts("Dear #{@user.login}, ")%>
-<%= ts("We regret to inform you that your request for #{@total} new ") %><%= @total == 1 ? ts("invitation") : ts("invitations") %><%= ts(" cannot be fulfilled at this time.") %>
-<%= ts("Your request was: ") %>
+<%= t('.greeting', user: @user.login) %>
+<%= t('.main', count: @total) %>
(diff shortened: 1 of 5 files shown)


cesy:
Looking at the codeclimate failures, I think those are all things that we don't want to change at this stage.

app/views/user_mailer/invite_request_declined.text.erb
```diff
@@ -1,9 +1,9 @@
...
-<%= ts("We regret to inform you that your request for #{@total} new ") %><%= @total == 1 ? ts("invitation") : ts("invitations") %><%= ts(" cannot be fulfilled at this time.") %>
-<%= ts("Your request was: ") %>
+<%= t('.greeting', user: @user.login) %>
+<%= t('.main', count: @total) %>
+<%= t('.reason') %>
 <%= to_plain_text(raw @reason) %>
 
-Regards,
-AO3
+<%= t '.signoff' %>
```
sarken: Can you add the parentheses to this for consistency? Or does it need to be done without parentheses?

spec/mailers/user_mailer_spec.rb
```diff
@@ -311,4 +311,52 @@
...
+      text = "[#{ArchiveConfig.APP_SHORT_NAME}] Additional Invite Code Request Declined"
+      expect(email.subject).to eq(text)
+    end
+
+    # Test both body contents
+    it_behaves_like "multipart email"
+
+    describe 'HTML version' do
+      it 'has text contents' do
+        expect(get_message_part(email, /html/)).to include("We regret to inform you")
```
sarken: It looks like the tests are set up to simulate the user requesting two invitations. Instead of checking for the text "We regret to inform you," which is the same no matter how many emails are sent, it might be a good idea to check for "new invitations cannot be fulfilled" instead, just to make sure we're getting the correct text for this scenario.

app/views/user_mailer/invite_request_declined.text.erb
cesy: Yes, fine.

spec/mailers/user_mailer_spec.rb
cesy: Yes, good point.

PR #3251, opened 2018-02-28
AO3-5340 Show bookmarks of deleted items.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5340

## Purpose

The purpose of this pull request is to make sure that bookmarks of deleted items (works, series, or external works) don't completely disappear. In order to accomplish this, I made the following changes:

1. Modify the bookmarkable visibility filters in BookmarkQuery so that instead of requiring the bookmarkable be unhidden, unrestricted, and posted, it requires *not* having a hidden, restricted, or unposted bookmarkable. A non-existent bookmarkable can't have hidden_by_admin set to false, but it won't have hidden_by_admin set to true, so this reversal results in bookmarks of deleted works being included in searches. (This is inspired by Elz's plans several months ago.)

2. Modify the bookmark counts in the pseud search to include bookmarks of deleted works. In order to accomplish this, I slightly refactored the visible_to_all and visible_to_restricted_user scopes in the Bookmark class. This made it possible to construct a Bookmark relation that included bookmarks with a visible bookmarkable *or* with a missing bookmarkable.

3. Add a call to update_bookmarker_pseuds_index when a bookmarkable is destroyed.

## Testing

The bug report has steps for reproducing the bug.

### DIFF
--- a/app/models/search/bookmark_query.rb
+++ b/app/models/search/bookmark_query.rb
@@ -37,7 +37,10 @@ class BookmarkQuery < Query
   end
 
   def exclusion_filters
-    @exclusion_filters ||= tag_exclusion_filter
+    @exclusion_filters ||= [
+      tag_exclusion_filter,
+      bookmarkable_visibility_filters
+    ].flatten.compact
   end
 
   # Instead of doing a standard query, which would only match bookmark fields
@@ -140,10 +143,19 @@ class BookmarkQuery < Query
   def visibility_filters
     [
       privacy_filter,
-      posted_filter,
-      hidden_filter,
+      hidden_filter
+    ]
+  end
+
+  # Filters dealing with the visibility of the parent.
+  # Note that in order to include bookmarks of deleted works/series/external
+  # works in some search results, we set these up as *exclusion* filters, so
+  # the meanings of the filters are flipped.
+  def bookmarkable_visibility_filters
+    [
+      unposted_parent_filter,
       hidden_parent_filter,
-      restricted_filter
+      restricted_parent_filter
     ]
   end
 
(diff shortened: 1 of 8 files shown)


app/models/search/bookmark_query.rb
```diff
@@ -140,10 +143,19 @@ def aggregations
   def visibility_filters
```
sarken: Just checking: we don't need the `posted_filter` here because this is used for _bookmarks_, and only _bookmarkables_ can be unposted?

lib/bookmarkable.rb
```diff
@@ -6,6 +6,7 @@ def self.included(bookmarkable)
       has_many :user_tags, through: :bookmarks, source: :tags
       after_update :update_bookmarks_index
       after_update :update_bookmarker_pseuds_index
+      after_destroy :update_bookmarker_pseuds_index
```
sarken: Just thinking out loud here for the sake of recording it for posterity, but at some point, we might need/want to do `after_destroy :update_bookmarks_index` here. A recent conversation revealed that if I bookmark a work tagged "The X-Files" and the work is deleted, the bookmark will continue to appear on the bookmark page for "The X-Files"... _but_ if "The X-Files" later gets synned to "The X-Files (TV 1993)", my bookmark will (understandably) disappear.

app/models/search/bookmark_query.rb
tickinginstant: Roughly, yeah. The `posted_filter` in the `BookmarkQuery` class has always been a parent filter (constructed with the `parent_term_filter` function), meaning it applies only to the bookmarkable. The old `posted_filter` was included in the AND/`must` portion of the query, meaning that there was a constraint in the query saying, "in order to return this bookmark, it must have a bookmarkable that has `posted` set to `true`." However, the "it must have a bookmarkable" part of that constraint meant that that query could never return any bookmarks with a deleted bookmarkable.

The idea behind this pull request is to take that constraint and take the double negative: "in order to return this bookmark, it must *NOT* have a bookmarkable that has `posted` set to `false`." That way it will correctly exclude bookmarks on draft works, correctly allow bookmarks on posted works, and correctly allow bookmarks on deleted works.

To implement that modification to the constraint, I moved `posted_filter` from `visibility_filters` (which are included in the AND/`must` portion of the query) to `bookmarkable_visibility_filters` (which are included in the NOT/`must_not` portion of the query). I also flipped the boolean from "true" to "false" to get the proper double-negative.

(I also renamed `posted_filter` to `unposted_parent_filter`, on the theory that the name should reflect its role better. But I think that may have been more confusing in the short term.)

lib/bookmarkable.rb
tickinginstant: In theory, the separation of the bookmarks index into bookmark documents and bookmarkable documents should make that issue disappear once the new search code is deployed.

Because the work's filters are no longer being stored in the bookmark's document, the relationship between the bookmark and the tag "The X-Files" ought to disappear as soon as the work disappears from the bookmarkable index. (And because the Work class includes `lib/searchable.rb`, it has an `after_destroy` hook that queues it up for reindexing.)

app/models/search/bookmark_query.rb
sarken: `unposted_parent_filter` totally made sense to me, fwiw -- it was kind of my "ah-ha, I see" moment, in fact. I think it's really easy to understand "we're just excluding things that are marked unposted, so everything else gets included."

tickinginstant:
It occurred to me that flipping the basic parent filters means that it's possible to have bookmarkable documents returned in a BookmarkQuery. I believe it can only happen when `show_private` is set to true and both `pseud_id` and `user_id` are null, which should be impossible to achieve outside of tests/the console, but I figure it's worth preventing future confusion if that ever changes.

PR #3252, opened 2018-03-07
AO3-5305 Overhaul bookmark index pages to arrange by bookmarkable.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5305

## Purpose

It's very difficult to compute accurate counts on bookmarkable filters in an efficient way, so instead this pull request is overhauling the bookmark listings so that they show bookmarkable items, with a small number of bookmarks per bookmarkable.

In order to accomplish this, I refactored the BookmarkQuery/BookmarkableQuery classes so that they're each responsible for generating their section of the query (bookmark filters in BookmarkQuery, bookmarkable filters in BookmarkableQuery), and call each other to generate the rest of the query. I also added child aggregations in the BookmarkableQuery class (since it's now being used to return actual search results), and some sorting options (ditto). In addition, I added the ability to return the 5 most recent matching bookmarks. (The number is customizable.)

In order to return the bookmarkable search results using the standard QueryResult class, I wrote a special BookmarkableDecorator class that can read in the 5 extra bookmarks associated with each bookmarkable item. I also restructured the QueryResult.items function so that it's more customizable, in the event of other classes that require special handling like BookmarkableDecorator.

To display the search results, I modified the controller class to use bookmarkable queries when the new search is enabled and there is no `@user` specified, and I added a new view to display a BookmarkableDecorator object and the 5 most recent matching bookmarks.

Finally, I re-enabled the ES=2 tests and modified the test harness so that it's possible to have cucumber tests run only with ES=1 enabled or only with ES=2 enabled. This allows for more complete testing of the new bookmark code.

## Testing

There's a lot of changes here, unfortunately, which means it's probably going to be a bit painful to test. I think the critical things to test would be:

1. View the Bookmarks page for a tag with bookmarks. Try filtering in various ways and seeing if the results look reasonable -- the counts in the sidebars, the display of the bookmarkables, the display of the bookmarks, which items show up and which don't, what happens when you click the buttons, what the user interface looks like, etc.

2. View the Bookmarks page for a collection with bookmarks, and do the same.


.travis.yml
```diff
@@ -5,39 +5,39 @@ sudo: required
...
-  # - ES=2 TEST_GROUP="cucumber -r features features/other_b            -f Ao3Cucumber::Formatter --color"
-  # - ES=2 TEST_GROUP="cucumber -r features features/prompt_memes_a     -f Ao3Cucumber::Formatter --color"
-  # - ES=2 TEST_GROUP="cucumber -r features features/prompt_memes_b     -f Ao3Cucumber::Formatter --color"
-  # - ES=2 TEST_GROUP="cucumber -r features features/prompt_memes_c     -f Ao3Cucumber::Formatter --color"
-  # - ES=2 TEST_GROUP="cucumber -r features features/search             -f Ao3Cucumber::Formatter --color"
-  # - ES=2 TEST_GROUP="cucumber -r features features/tag_sets           -f Ao3Cucumber::Formatter --color"
-  # - ES=2 TEST_GROUP="cucumber -r features features/tags_and_wrangling -f Ao3Cucumber::Formatter --color"
-  # - ES=2 TEST_GROUP="cucumber -r features features/users              -f Ao3Cucumber::Formatter --color"
-  # - ES=2 TEST_GROUP="cucumber -r features features/works              -f Ao3Cucumber::Formatter --color"
+  - ES=1 TEST_GROUP="cucumber --tag ~@new-search -r features features/admins             -f Ao3Cucumber::Formatter --color"
```
tickinginstant: The `~` character indicates negation. The ES=1 tests exclude any cucumber tests tagged `@new_search`, while the ES=2 tests exclude any cucumber tests tagged `@old_search`.

app/models/search/bookmark_query.rb
```diff
@@ -72,11 +80,16 @@ def add_owner
...
-    [
-      general_query,
-      parent_query
-    ]
+  # Instead of doing a standard query, which would only match bookmark fields
+  # we'll make this a should query that will try to match either the bookmark
+  # or its parent
+  # TODO This isn't right, since it requires all of the fields to be on one or
+  # the other (and, in particular, can't handle tags on the parent with a
+  # specified bookmarker on the child).
```
tickinginstant: This is AO3-5301/AO3-5320. Rather than make this PR even more complex, I left the bug in.

features/bookmarks/bookmark_privacy.feature
```diff
@@ -101,32 +102,33 @@ Feature: Private bookmarks
...
-      And I should not see "bookmarker"
+      And I should not see "avid_bookmarker"
 
-      # Private bookmarks should not show on tag's page
+    # Private bookmarks should not show on tag's page
     When I go to the bookmarks tagged "Stargate SG-1"
     Then I should not see "Secret Masterpiece"
       And I should see "Public Masterpiece"
-      And I should not see "bookmarker"
+      And I should not see "avid_bookmarker"
```
tickinginstant: Because of the separation of the filters/queries in the sidebar, the lowercase string "bookmarker" now appears on all bookmark listings when the new search is enabled. So I renamed the user in this test to make it easier to check whether their username is associated with any publicly visible bookmarks.

app/decorators/pseud_decorator.rb
```diff
@@ -11,6 +19,7 @@ def self.decorate_from_search(results, search_hits)
     end
   end
 
+  # TODO Either eliminate this function or make it work properly.
```
sarken: Just wondering about this comment! Did something new get broken in pseud search?

spec/models/bookmark_query_spec.rb
```diff
@@ -5,7 +5,22 @@
   it "should allow you to perform a simple search" do
```
ariana-paris: As we have quite a lot of search-related specs now, I'm wondering if we should put them into the `search` subfolder so they're not lost in the pile in `models`. But not something that urgently needs to be done right now.

spec/models/bookmark_query_spec.rb
```diff
@@ -36,24 +51,28 @@
 
   it "should not return bookmarks of hidden objects" do
     q = BookmarkQuery.new
-    expect(q.exclusion_filters).to include(has_parent: { parent_type: 'bookmarkable', query: { term: { hidden_by_admin: 'true' } } })
+    parent_filter = q.exclusion_filters.first { |f| f.key? :has_parent }
```
ariana-paris: It looks as if there's some repeated code that could possibly be extracted into a `before :each` - unless that creates more problems than it's worth

spec/models/bookmark_search_form_spec.rb
```diff
@@ -10,4 +10,84 @@
       expect(bsf.options).to include(show_private: false)
     end
   end
+
+  describe "bookmarkable_search_results" do
+    describe "sorting" do
+      let(:tag) { create(:canonical_fandom) }
+
+      let!(:work1) do
+        Delorean.time_travel_to 40.minutes.ago do
```
ariana-paris: Nice use of `time_travel_to` to create items at specific relative dates rather than our usual approach of hard-coding dates that come back to bite us. *takes notes*

tickinginstant:
Thanks for the review! I made most of the suggested fixes, except for moving the search specs into a separate folder, and moving the load_from_elasticsearch function (since there's no clear answer as to where it belongs).

app/models/search/bookmark_query.rb
```diff
@@ -13,44 +15,50 @@ def document_type
...
+  def initialize(options = {})
+    @options = HashWithIndifferentAccess.new(options)
+    add_owner
+    self.bookmarkable_query = BookmarkableQuery.new(self)
+  end
+
   # After the initial search, run an additional query to get work/series tag filters
   # Elasticsearch doesn't support parent aggregations, and doing the main query on the parents
   # limits searching and sorting on the bookmarks themselves
   # Hopefully someday they'll fix this and we can get the data from a single query
```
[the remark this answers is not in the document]
tickinginstant: I'm hoping some of the similarities between nested and joined documents mean that Elasticsearch will eventually add parent join aggregations, as well. But I'm not sure what the innards of ES look like. I'm a little worried that the fact that they haven't already added it means that it's intractable for some reason.

app/decorators/pseud_decorator.rb
tickinginstant: Nothing new got broken. I think the function just never got finished, and since it never gets called, no one noticed that it always errors. (`NameError: undefined local variable or method 'work_counts' for PseudDecorator:Class`)

I added more details about how to fix it in the comment -- I know TODOs have a tendency to get lost in the shuffle if there's no associated bug report, but I wanted to mark that function so that anyone who tries to use it in the future knows that it's not done.

PR #3257, opened 2018-03-10
AO3-5359 Show more precise results at top in people search

ariana-paris:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5359

## Purpose

This will allow users to search for a pseud that includes numbers (e.g.: **testy2**) and see that result at the top, while also still seeing results for e.g.: **testy3** or **testy**.

## Testing

See Jira ticket.


app/models/search/pseud_indexer.rb
```diff
@@ -18,7 +18,7 @@ def self.mapping
           },
           byline: {
             type: "text",
-            analyzer: "simple"
+            analyzer: "standard"
```
redsummernight: For [reference](https://www.elastic.co/guide/en/elasticsearch/reference/6.0/analysis-analyzers.html):

> The standard analyzer divides text into terms on word boundaries, as defined by the Unicode Text Segmentation algorithm. It removes most punctuation, lowercases terms, and supports removing stop words. 

and

> The simple analyzer divides text into terms whenever it encounters a character which is not a letter. It lowercases all terms.

PR #3258, opened 2018-03-12
AO3-5355 Add filtering bookmarks by bookmarker's tag names

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5355

## Purpose

These new fields now work:

- Other bookmarker's tags to include
- Other bookmarker's tags to exclude

Previously we have these fields on the BookmarkSearchForm options:

- tag_ids: IDs of bookmarker tags, handled by BookmarkQuery, search on bookmarks
- other_tag_names: handled by TaggableQuery, search on bookmarkables, but not bookmarks (causing [AO3-5317](https://otwarchive.atlassian.net/browse/AO3-5317))
- excluded_tag_ids/excluded_tag_names: handled by both BookmarkQuery and BookmarkableQuery through TaggableQuery, search on bookmarkables and bookmarks

I refactored so the new fields are as follows:

- <tag_type>_ids/other_tag_names: for bookmarkables (similar to work filters, this is the shared, unchanged logic of TaggableQuery)
- excluded_tag_ids/excluded_tag_names: for bookmarkables (same fields, still handled by TaggableQuery)
- tag_ids/other_bookmark_tag_names: for bookmarks; the former is already there, the latter is new
- excluded_bookmark_tag_ids/excluded_bookmark_tag_names: for bookmarks; the former is now separated from excluded_tag_ids, the latter is new

Now, **only** the "Bookmarker's Tags" checkboxes and the 2 "Other bookmarker's tags to include/exclude" fields will be used to search tags on bookmarks, everything else will apply to bookmarkables.

All bookmark-specific fields will be processed in BookmarkQuery, the rest in BookmarkableQuery/TaggableQuery. BookmarkQuery no longer needs to include TaggableQuery.

There's another new field, "Search within bookmarker's tags and notes", which I'm leaving out of this pull request.

## Testing

- Try filtering using all 4 "other" tag fields in the bookmark filters
- Try filtering using the "Bookmarker's Tags" checkboxes, include/exclude
- See [AO3-5317](https://otwarchive.atlassian.net/browse/AO3-5317)


app/models/search/bookmark_query.rb
```diff
@@ -322,4 +319,24 @@ def user_ids
...
+    @excluded_bookmark_tag_ids = bookmark_tag_ids(:excluded_bookmark_tag_ids, :excluded_bookmark_tag_names)
+  end
+
+  def bookmark_tag_ids(ids_field, names_field)
+    return if options[ids_field].blank? && options[names_field].blank?
+
+    ids = options[ids_field] || []
+    names = options[names_field]&.split(",")
+    ids += Tag.where(name: names).pluck(:id) if names
+    ids += Tag.where(id: ids).pluck(:merger_id)
```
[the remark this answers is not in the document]
redsummernight: I changed the function to only add merger IDs for exclusions. There's a test for excluding bookmarks by synonyms:

https://github.com/otwcode/otwarchive/blob/235d5773d24752c4db54f404381141a1880d178d/spec/models/bookmark_search_form_exclusion_filters_spec.rb#L120

features/search/filters.feature
```diff
@@ -133,6 +133,69 @@ Feature: Filters
...
+    # Use a second include checkbox for bookmarks with both tags
+    When I check "been here (1)" within "#include_tag_tags"
+      And I press "Sort and Filter"
+    Then I should see "1 Bookmark by recengine"
+      And I should see "Bilbo Does the Thing"
+
+    # Use an exclude checkbox
+    When I go to my bookmarks page
+      And I press "Bookmarker's Tags" within "dd.exclude"
+    Then the "to read (2)" checkbox within "#exclude_tag_tags" should not be checked
```
redsummernight: These checkbox steps don't actually do anything:

https://github.com/otwcode/otwarchive/blob/235d5773d24752c4db54f404381141a1880d178d/features/step_definitions/web_steps.rb#L268-L278

However, adding assertions to them reveal failures elsewhere, so I'm leaving that for later. This particular test does pass if the checkbox steps have assertions.

app/helpers/advanced_search_helper.rb
```diff
@@ -20,7 +20,11 @@ def filter_boolean_value(filter_action, tag_type, tag_id)
       @search.send("#{tag_type}_ids").present? &&
         @search.send("#{tag_type}_ids").include?(tag_id)
     else
-      @search.excluded_tag_ids.present? && @search.excluded_tag_ids.include?(tag_id)
+      value = @search.excluded_tag_ids.present? && @search.excluded_tag_ids.include?(tag_id)
+      if @search.respond_to?(:excluded_bookmark_tag_ids)
```
redsummernight: This helper is also used on WorkSearchForm which doesn't have the `excluded_bookmark_tag_ids` accessor.

redsummernight:
Thanks @tickinginstant!

app/models/search/bookmark_query.rb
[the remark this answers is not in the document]
redsummernight: @sarken recommended keeping the behavior for inclusion and exclusion consistent, so I'm removing merger IDs from exclusions too: #3261.

features/search/filters.feature
redsummernight: [AO3-5380](https://otwarchive.atlassian.net/browse/AO3-5380).

PR #3260, opened 2018-03-14
AO3-5361 Link text and alignment changes for bookmarks

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5361

## Purpose

On a tag's bookmark page, change the link text from "View All" to "All Bookmarks." On that page and on an item's bookmark page, moves icon and "Bookmarked by" info a smidge to the left.

## Testing

Refer to Jira

### DIFF
--- a/public/stylesheets/site/2.0/13-group-blurb.css
+++ b/public/stylesheets/site/2.0/13-group-blurb.css
@@ -368,10 +368,14 @@ blurbs on the manage collection items pages, mostly reseting styles inherited fr
 }
 
 .bookmark .short .status {
-	left: 30px;
+	left: 0;
 	margin-top: 0;
 	top: 0;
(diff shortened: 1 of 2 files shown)


public/stylesheets/site/2.0/13-group-blurb.css
```diff
@@ -368,12 +368,16 @@ blurbs on the manage collection items pages, mostly reseting styles inherited fr
...
 
 .bookmark .short .status {
-	left: 30px;
+	left: 0;
 	margin-top: 0;
 	top: 0;
 	width: 25px;
 }
 
+.bookmark .short .header h5 {
```
sarken: (This probably would've worked without .header, but we don't want to accidentally affect users' h5s in bookmark notes.)

PR #3261, opened 2018-03-16
AO3-5355 Stop adding merger IDs in bookmarker tag exclusions

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5355

## Purpose

We [don't add merger IDs for bookmarker tag inclusions](https://github.com/otwcode/otwarchive/pull/3258#discussion_r174285725). Bookmarker tag exclusions don't have the same problem, but it's better to be consistent.

## Testing

The filter under Exclude > Bookmarker's Tags should now exclude only the exact tag in the checkbox, but not its canonical synonym.


spec/models/bookmark_search_form_exclusion_filters_spec.rb
```diff
@@ -117,19 +117,19 @@
         expect(search.search_results).not_to include(excluded_bookmark)
       end
 
-      it "should exclude bookmarks tagged with a canonical tag given that tag's synonym" do
-        excluded_bookmark.update(tag_string: "Complete")
-        update_and_refresh_indexes("bookmark")
+      # it "should exclude bookmarks tagged with a canonical tag given that tag's synonym" do
```
sarken: I think we might want to either a) remove this or b) add a note about why it's commented out, just so no one comes across this later and thinks "Oh, a failing test to fix!" or "Oh, behavior I should implement!"

redsummernight: I can remove it. Should I remove the other commented out tests?

sarken: If they're tests for things we won't want to implement, yep, please do!

PR #3262, opened 2018-03-17
AO3-5274 In filters, text of selected tag not bolded on old browsers

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5274

## Purpose

Changes the selector we use to bold the name of checked tags in the filters because Safari 8 and 9 don't support the one we use. This might not work.

## Testing

Use the filter checkboxes in older versions of Safari.

### DIFF
--- a/public/stylesheets/site/2.0/18-zone-searchbrowse.css
+++ b/public/stylesheets/site/2.0/18-zone-searchbrowse.css
@@ -187,7 +187,7 @@ form.filters {
   border-color: #aaa;
 }
 
-.filters input:checked + .indicator + span {
+.filters input:checked ~ .label {
   font-weight: 700;
 }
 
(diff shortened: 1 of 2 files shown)


public/stylesheets/site/2.0/18-zone-searchbrowse.css
```diff
@@ -187,7 +187,7 @@ form.filters {
   border-color: #aaa;
 }
 
-.filters input:checked + .indicator + span {
+.filters input:checked ~ .label {
```
redsummernight: Do we need to add a label class?

sarken:
That... is probably the change that was in the application helper file I didn't commit. (I thought that file was just part of the changes to make ES work on webdev.) I'll update this with it tomorrow--hopefully I didn't check it out.

PR #3263, opened 2018-03-17
AO3-5364 Turn collection bookmark counts into bookmarked item counts for new search.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5364

## Purpose

This PR makes the bookmark counts in the sidebar of collections into *bookmarked item* counts, instead. It also changes the count in blurbs on the Browse > Collections page to match the sidebar counts, as well. (Since that blurb is cached, this necessitated some changes to the cache key to account for the difference in count and label when logged in/logged out or using new search/old search. Which in turn required some changes to the sweepers.)

I also noticed while testing that bookmark search and bookmark listings don't show bookmarks of restricted works to admins, so I fixed that to make sure that the numbers are consistent even when logged in as admin.

## Testing

1. While logged in as an admin, check the bookmark count in the collection's sidebar to make sure it matches the number in the header of the collection's /bookmarks page. Also check the number in the collection's blurb on a /collections listing (perhaps on a parent collection's "Subcollections" page).
2. While logged in as a regular user, check the same three numbers.
3. While logged out, check the same three numbers.

### DIFF
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -343,8 +343,8 @@ class Collection < ApplicationRecord
     # Note that "approved_by_collection" forces the bookmarks to be approved
     # both by the collection AND by the user.
     bookmarks = Bookmark.is_public.joins(:collection_items).
-      merge(CollectionItem.approved_by_collection).
-      where(collection_items: { collection_id: children.ids + [id] })
+                merge(CollectionItem.approved_by_collection).
+                where(collection_items: { collection_id: children.ids + [id] })
 
     logged_in = User.current_user.present?
 
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -335,6 +335,29 @@ class Collection < ApplicationRecord
     count
   end
 
+  # Return the count of all bookmarkable items (works, series, external works)
+  # that are in this collection (or any of its children) and visible to
+  # the current user. Excludes bookmarks of deleted works/series.
+  def all_bookmarked_items_count
+    # The set of all bookmarks in this collection and its children.
+    # Note that "approved_by_collection" forces the bookmarks to be approved
+    # both by the collection AND by the user.
+    bookmarks = Bookmark.is_public.joins(:collection_items).
+      merge(CollectionItem.approved_by_collection).
+      where(collection_items: { collection_id: children.ids + [id] })
+
+    logged_in = User.current_user.present?
+
+    [
+      logged_in ? Work.visible_to_registered_user : Work.visible_to_all,
+      logged_in ? Series.visible_to_registered_user : Series.visible_to_all,
+      ExternalWork.visible_to_all
+    ].map do |relation|
+      relation.joins(:bookmarks).merge(bookmarks).distinct.
+        count("bookmarks.bookmarkable_id")
+    end.sum
+  end
+
   def all_fandoms
     # We want filterable fandoms, but not inherited metatags:
     Fandom.for_collections([self] + children).
--- a/app/sweepers/collection_sweeper.rb
+++ b/app/sweepers/collection_sweeper.rb
@@ -46,12 +46,20 @@ class CollectionSweeper < ActionController::Caching::Sweeper
   def expire_collection_cache_for(record)
     collections = get_collections_from_record(record)
     collections.each do |collection|
-
-      # expire the collection blurb and dashboard and profile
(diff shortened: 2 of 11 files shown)


spec/models/collection_spec.rb
```diff
@@ -72,4 +72,44 @@
...
+
+  describe "all_bookmarked_items_count" do
+    let(:collection) { create(:collection) }
+
+    it "does not include bookmarks of deleted works" do
+      work = create(:posted_work)
+      create(:bookmark, collections: [collection], bookmarkable: work)
+      expect do
+        work.destroy
+      end.to change { collection.all_bookmarked_items_count }.from(1).to(0)
```
redsummernight: This expect is cool, I'm gonna start using it.

app/models/collection.rb
```diff
@@ -335,6 +335,29 @@ def all_approved_bookmarks_count
...
 
+  # Return the count of all bookmarkable items (works, series, external works)
+  # that are in this collection (or any of its children) and visible to
+  # the current user. Excludes bookmarks of deleted works/series.
+  def all_bookmarked_items_count
+    # The set of all bookmarks in this collection and its children.
+    # Note that "approved_by_collection" forces the bookmarks to be approved
+    # both by the collection AND by the user.
+    bookmarks = Bookmark.is_public.joins(:collection_items).
+      merge(CollectionItem.approved_by_collection).
```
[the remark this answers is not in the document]
ariana-paris: Since Ruby inexplicably doesn't approve of start of line dots to indicate that the next line is a continuation of the previous one (sorry, Java habits coming to the fore), I second at least aligning the chained methods with the start of the chain to make it clearer this is still the chain and not a separate statement with accidental indentation.

app/models/collection.rb
```diff
@@ -335,6 +335,29 @@ def all_approved_bookmarks_count
...
+      where(collection_items: { collection_id: children.ids + [id] })
+
+    logged_in = User.current_user.present?
```
ariana-paris: Are we not concerned about admins here?

app/sweepers/collection_sweeper.rb
```diff
@@ -46,12 +46,20 @@ def get_collections_from_record(record)
...
-      ActionController::Base.new.expire_fragment("collection-blurb-#{collection.id}-v3")
-      ActionController::Base.new.expire_fragment("collection-profile-#{collection.id}")
-
+      CollectionSweeper.expire_collection_blurb_and_profile(collection)
     end
   end
 
+  # Expire the collection blurb and profile
+  def self.expire_collection_blurb_and_profile(collection)
+    # Expire all versions of the blurb: whether the new search is enabled or
```
ariana-paris: Do we want a TODO for removing the old-search/new-search consideration when ES is full switched over?

spec/models/collection_spec.rb
ariana-paris: Looks useful. I might adopt that too.

app/models/collection.rb
tickinginstant: The `User.current_user` variable contains either a `User`, an `Admin`, or `nil`, so this does actually handle admins. (It's the `ApplicationController.current_user` variable that only contains `User`s or `nil`.)

app/sweepers/collection_sweeper.rb
tickinginstant: Currently, it also changes the view based on whether `@admin_settings.disable_filtering?` is on or off, since disabling filtering also disables the ES bookmarkable-based view. I'm not sure whether it's worth keeping separate caches just for that, so I added a TODO to consider it later.

app/models/collection.rb
ariana-paris: That's not confusing... but it's a pre-existing behaviour so let's ignore it and move swiftly on 👍

app/sweepers/collection_sweeper.rb
ariana-paris: I live in hope that we might round these up and action them some day so thanks for adding the explanation for future reference.

app/models/collection.rb
tickinginstant: I've been thinking about trying to change it for months. I hate how easy it is to mix it up. (But I agree that now is not the time.)

app/sweepers/collection_sweeper.rb
sarken: iirc, we've been muttering to ourselves about getting rid of that admin setting

PR #3266, opened 2018-03-19
AO3-5363 Update bookmarkable blurb

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5363

## Purpose

* Adds some classes so we can fix Low Vision Default
* Removes a TODO we won't do
* Moves the logic for displaying the count to a helper so it isn't repeated in the bookmarkable and bookmark blurbs

## Testing

Refer to JIRA

## References

Rest of fixes for Low Vision Default will be in the JIRA issue

### DIFF
--- a/app/helpers/bookmarks_helper.rb
+++ b/app/helpers/bookmarks_helper.rb
@@ -108,4 +108,11 @@ module BookmarksHelper
     end
   end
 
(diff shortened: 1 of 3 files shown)


sarken:
Closing because I have the issue number wrong AGAIN.

PR #3277, opened 2018-04-05
AO3-5376 Add complete field to series and work bookmarkables

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5376

## Purpose

Index series and work bookmarkable documents on whether or not they are complete so it is once again possible to search for `complete: true` on bookmarked items.

## Testing

Refer to Jira

### DIFF
--- a/app/models/series.rb
+++ b/app/models/series.rb
@@ -227,7 +227,8 @@ class Series < ApplicationRecord
   def bookmarkable_json
     as_json(
       root: false,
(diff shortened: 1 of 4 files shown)


redsummernight:
Should we do anything for external works? Are they considered neither complete nor incomplete by default?

redsummernight:
Never mind, I see your test. :+1:

PR #3282, opened 2018-04-12
AO3-5332 Remove creator from work when they remove themselves from last chapter they created

DreamSea:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5332

## Purpose

When a co-creator removes themselves from the last chapter for which they were credited, it should also remove them as a co-creator of the work and remove all associated edit permissions etc.

## Testing

- Add at least two chapters as co-creator. 
- Using chapter edit and/or manage, remove yourself from one of the chapters - verify still have edit access to work.
- Remove yourself from from other (last credited) chapter - verify redirected to user screen, verify no longer have edit access to work.


DreamSea:
Was hoping to get some clarification on the difference between 'co-creator for chapter' and 'co-creator for work'. 

Poking around in the app, it seemed like actions 'add/remove from work' were mainly just shortcuts for 'add/remove from all chapters', without any special distinction other than 'a co-creator on any chapter is considered a co-creator on the overall work' as far as the bylines go. 

Adding co-creator during 'create work' just seemed to be a shortcut of 'add co-creator to chapter 1'.

~

If the above is true, is the goal of the issue to add and logic off some indicator of 'co-created added through work'?

wip implementation assumes only one type of co-creator, and removes from work if removed from last authored chapter.

~

and then misc question (new issue?): if the removal from work is triggered, should it follow removal from work flow - 'alert user that they are removed from work', 'redirect_to current_user', warn user that they'll lose co-creator access, etc.

DreamSea:
Sounds good, I'll bump the request off WIP once author work workflow commit(s) are in. o7

DreamSea:
~~build checks struggle~~

Not familiar enough to know what logic is better for where :thinking: figured want the co-creator removal logic at least at the controller level so that it can't be bypassed by blind GETs at urls (though probably unlikely that it would happen).

Copied controller forwarding strats from places in project (ex. series -> orphans), open to suggestion.

---
codeclimate: `File chapters_controller.rb has 251 lines of code (exceeds 250 allowed). Consider refactoring.`

I can probably cheese one line off the controller by bundling
```ruby
@chapter.authors_to_remove = current_user.pseuds
@chapter.save
```
into something like `@chapter.remove_author(current_user)` at the model similar to the existing `@work.remove_author(current_user)`.

---
travis:
```
(::) failed steps (::)
Unable to find select box "Rating" (Capybara::ElementNotFound)
./features/step_definitions/web_steps.rb:93:in `block (2 levels) in <top (required)>'
./features/step_definitions/web_steps.rb:14:in `with_scope'
./features/step_definitions/web_steps.rb:92:in `/^(?:|I )select "([^"]*)" from "([^"]*)"(?: within "([^"]*)")?$/'
features/tags_and_wrangling/tag_wrangling_relationships.feature:70:in `And I select "Not Rated" from "Rating"'
Failing Scenarios:
cucumber features/tags_and_wrangling/tag_wrangling_relationships.feature:6 # Scenario: relationship wrangling - syns, mergers, characters, autocompletes
```

didn't seem like it in the travis configuration, but would the build screenshots happen to be uploaded anywhere? in local run, it passed `feature:70` successfully before hitting TimeoutError at `feature:77` ~~cus i have potato laptop~~
```
HTML screenshot: /home/travis/build/otwcode/otwarchive/tmp/capybara/screenshot_2018-04-15-02-13-49.403.html
Image screenshot: /home/travis/build/otwcode/otwarchive/tmp/capybara/screenshot_2018-04-15-02-13-49.403.png
```

redsummernight:
> I can probably cheese one line off the controller

We have codeclimate around as a guide, not a strict check like Travis. We can refactor to make things more readable but I wouldn't worry about meeting the line limit.

> would the build screenshots happen to be uploaded anywhere

I restarted the tags_and_wrangling Travis job for this PR and the test failure went away. The screenshots are not uploaded anywhere. If you're having trouble running things locally, you can also [set up Travis for your own fork](https://github.com/otwcode/otwarchive/wiki/Travis-CI#travis-cigithub-link-setup), so you don't need to open/update a PR to run tests.

app/controllers/chapters_controller.rb
```diff
@@ -93,8 +93,12 @@ def edit
     if params["remove"] == "me"
       @chapter.authors_to_remove = current_user.pseuds
       @chapter.save
-      flash[:notice] = ts("You have been removed as a creator from the chapter")
-      redirect_to @work
+      if @work.chapters.any? { |c| current_user.is_author_of?(c) }
+        flash[:notice] = ts("You have been removed as a creator from the chapter")
+        redirect_to @work
+      else # remove from work if no longer co-creator on any chapter
+        redirect_to controller: "works", action: "edit", id: @work.id, remove: "me"
```
sarken: Is it possible to do this as `redirect_to edit_work_path(@work, remove: "me")` instead? (I haven't actually done a full review, so feel free to let this wait in case something else comes up.)

DreamSea: I'll give it a try :+1: I've been meaning to look into how rails ~~magics~~ routes things like 'edit_work_path'

spec/controllers/chapters_controller_spec.rb
```diff
@@ -366,6 +358,28 @@
         it_redirects_to_with_error(work_path(work), "Sorry, you don't have permission to access the page you were trying to reach.")
       end
     end
+
+    context "when co-creator is logged in" do
+      before do
+        @co_creator = create(:user)
```
redsummernight: We should use let to avoid sharing state between tests in instance variables. You don't need to update the existing instance variables though.

redsummernight:
This looks good!

Also since [AO3-5440](https://otwarchive.atlassian.net/browse/AO3-5440), the HTML screenshots after feature test failures are printed in Travis logs if you need them.

DreamSea:
haha yea, when i saw AO3-5440 merged in :tada: :confetti_ball:

spec/controllers/chapters_controller_spec.rb
DreamSea: will do :+1:  mini-commits on branch shouldn't be too noisy since the commits get squashed before merge?

redsummernight: Yeah more commits is fine, we always use squash and merge.

DreamSea:
I no longer have an environment set up for running/testing otwarchive :cloud_with_rain:. Y'all are welcome to take from this pull request and change it as needed. Alternatively (depending on the changes from https://otwarchive.atlassian.net/browse/AO3-5658), it might be cleaner to close this request and redo https://otwarchive.atlassian.net/browse/AO3-5332 from the ground up (though the scenario/feature test might be at a high enough level to reuse).

Cheers :coffee:

sarken:
Thank you for letting us know!

features/works/chapter_edit.feature
```diff
@@ -366,9 +367,31 @@ Feature: Edit chapters
       And I should see "Chapter by originalposter"
 
 
+  Scenario: Removing yourself as a co-creator from last remaining co-created chapter
```
redsummernight: I think this scenario and "Removing yourself as a co-creator from the chapter edit page" can be combined.

spec/controllers/chapters_controller_spec.rb
```diff
@@ -361,6 +361,29 @@
...
+        get :edit, params: { work_id: work.id, id: co_second_chapter.id, remove: "me" }
+        expect(assigns[:chapter].pseuds).to eq [user.pseuds.first]
+        it_redirects_to_with_notice(work_path(work), "You have been removed as a creator from the chapter")
+      end
+
+      it "delegates to work removal of user after user removed from last co-created chapter" do
+        get :edit, params: { work_id: work.id, id: co_second_chapter.id, remove: "me" }
+        expect(assigns[:work].pseuds).to eq [user.pseuds.first, co_creator.pseuds.first]
+        get :edit, params: { work_id: work.id, id: co_third_chapter.id, remove: "me" }
+        it_redirects_to edit_work_path(work, remove: "me")
```
redsummernight: Before checking the redirect, can we check if `co_second_chapter` and `co_third_chapter` have the correct pseuds (just `user.pseuds.first`)?

spec/controllers/chapters_controller_spec.rb
```diff
@@ -361,6 +361,33 @@
...
+
+      it "delegates to work removal of user after user removed from last co-created chapter" do
+        get :edit, params: { work_id: work.id, id: co_second_chapter.id, remove: "me" }
+        expect(assigns[:work].pseuds).to eq [user.pseuds.first, co_creator.pseuds.first]
+        get :edit, params: { work_id: work.id, id: co_third_chapter.id, remove: "me" }
+
+        expect(co_second_chapter.reload.pseuds).to eq [user.pseuds.first]
+        expect(co_third_chapter.reload.pseuds).to eq [user.pseuds.first]
+
+        it_redirects_to_with_notice(edit_work_path(work, remove: "me"), "You have been removed as a creator from the chapter.")
```
redsummernight: This check looks odd: reading the code, we should be redirected without any notice. Looks like the notice is left over from the first `get :edit`.

https://stackoverflow.com/questions/7850650/making-two-requests-to-the-same-controller-in-rails-integrations-specs

So I think we need to keep to one `get` per spec and do all the setup in `before` blocks:
- Set up 2 chapters.
- In context 1, no extra setup, test removing co-creator from last chapter, removing them from the work.
- In context 2, add a third co-created chapter, test removing co-creator from just one chapter, leaving them on the work.

PR #3289, opened 2018-04-21
AO3-3468 Change authors_to_sort_on from a column to a method

redsummernight:
## Issue

## Purpose

When a work's author changes their username:

- authors_to_sort_on is a column: the work needs to be resaved, then reindexed, and we're doing so synchronously.
- authors_to_sort_on is a method: its value is up-to-date when called, the work doesn't need to be resaved, just reindexed, and we can use the asynchronous indexing queues.

We can drop the database column authors_to_sort_on whenever.

Also add a hook so when a pseud gets renamed, its works and bookmarks get reindexed.

## Testing

See issue.

### DIFF
--- a/app/models/pseud.rb
+++ b/app/models/pseud.rb
@@ -73,7 +73,7 @@ class Pseud < ApplicationRecord
 
   after_update :check_default_pseud
   after_update :expire_caches
-  after_update :reindex_works
+  after_update :reindex_creations
 
   scope :on_works, lambda {|owned_works|
     select("DISTINCT pseuds.*").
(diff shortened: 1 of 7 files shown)


app/models/pseud.rb
```diff
@@ -497,4 +491,15 @@ def self.search(options={})
...
+  def should_reindex_works?
+    pertinent_attributes = %w[id name]
+    destroyed? || (saved_changes.keys & pertinent_attributes).present?
+  end
+
+  # If the pseud gets renamed, anything with the old name needs to be reindexed.
+  def reindex_works
+    return unless should_reindex_works?
+    IndexQueue.enqueue_ids(Work, works.pluck(:id), :main)
+    IndexQueue.enqueue_ids(Bookmark, bookmarks.pluck(:id), :main)
```
sarken: I'm having a brain issue -- is this reindexing the pseud's bookmarks, or bookmarks of works (and series?) by the pseud? Either way, if it's indexing more than just works, maybe a better name would be `reindex_creations`?

redsummernight: It's indexing works by the pseuds and bookmarks by the pseuds. I can rename the method.

This hook is not indexing bookmarks of works (and series) by the pseud (or bookmarkables by the pseud). Previously this wasn't a problem because when we updated the work (with `set_author_sorting`), the bookmarkable callbacks will run, but they no longer do. It's a new bug. :<

When user name changes, I'll add tests to check that these get reindexed:

- works by affected pseuds
- bookmarkables by affected pseuds (only works and series; external works have no pseuds)
- bookmarks by affected pseuds (since pseud names are indexed as bookmarkers)

redsummernight: When we reindex a work, its bookmarkable index gets updated as well, so I didn't add any bug there.

I did miss series reindexing though.

sarken: I love this new world order \o/

This looks good to go now 👍

sarken:
Belated thought: did we need to do this in the user model, too?

PR #3290, opened 2018-04-23
AO3-5391 Remove unused taggable methods

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5391

## Purpose

Removes unused code! :tada: (Note that it also removes something from the config file.)

ETA: One method -- `fandom_tags` -- turned out to be used in showing fandoms on external work blurbs, so I replaced it with a different way of doing that (the same way we use on work blurbs).

ETA 2: Two methods, `freeform_tags` and `warning_tags`, turned out to be used in showing tags on series blurbs when the user has tag-hiding preferences enabled. I switched them to the usual `freeforms` and `warnings` methods and updated the tests that didn't quite catch this issue.

## Testing

* Make sure external works don't 500.
* Set your preferences to hide warnings and additional tags, and then make sure you can reveal warnings and freeforms on series blurbs (that is, the blurb on /users/USERNAME/series -- not the work blurbs on the /series/1234 pages.)


tickinginstant:
FYI, `warning_tags` and `freeform_tags` are used in `TagsController#show_hidden`:
https://github.com/otwcode/otwarchive/blob/80c950525269715a85ebe63a58c234a2822fdac9/app/controllers/tags_controller.rb#L145-L149
But it's probably pretty easy to change those to `warnings` and `freeforms`.

sarken:
Closing while I rewrite the tests that should've caught this /o\

sarken:
Hookay, I think it's fixed and the tests are updated.

sarken:
Marked deferred because this is going in with the [Warning class rename](https://otwarchive.atlassian.net/browse/AO3-5613).

sarken:
This test failure is due to a change in AO3-5613 -- the test that's failing is new as of this PR, which is why we didn't catch it.

I'm going to fix it, but I need to think about how best to do it. I know the _easy_ way, but I want something that reduces the code duplication that led to the problem in the first place.

PR #3293, opened 2018-04-24
AO3-5253 Reindex external works when wrangling changes

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5253

## Purpose

Reindex external works when their tags' metatags or canonical status change. 

Includes PR #3292 and #3291 

## Testing

Refer to Jira


app/models/tag.rb
```diff
@@ -688,10 +689,12 @@ def recategorize(new_type)
   def reindex_taggables
     work_ids = all_filtered_work_ids
     series_ids = all_filtered_series_ids
+    external_work_ids = all_filtered_external_work_ids
```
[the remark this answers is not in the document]
ariana-paris: It's right, you know, `work_ids` and `series_ids` are used further down in the `reindex_all_*` methods, but the call to `reindex_all_external_works` is re-running the code in the `all_filtered_external_work_ids` method.

features/other_b/bookmark_indexing.feature
```diff
@@ -60,48 +60,59 @@ Feature: Bookmark Indexing
...
+  make the bookmarks appear in the metatag's bookmark listings; de-subbing
+  should remove them
     Given I am logged in as "author"
-      And a canonical freeform "Alternate Universe"
-      And a canonical freeform "Alternate Universe - High School"
-      And I post the work "The Story Telling: Beginnings" with freeform "Alternate Universe - High School" as part of a series "Telling Stories"
+      And a canonical character "Laura"
+      And a canonical character "Laura Roslin"
+      And I post the work "The Story Telling: Beginnings" with character "Laura Roslin" as part of a series "Telling Stories"
       And I am logged in as "bookmarker"
```
ariana-paris: This should probably be a `When` as we're switching from posting the work to bookmarking it

features/other_b/bookmark_indexing.feature
```diff
@@ -60,48 +60,59 @@ Feature: Bookmark Indexing
     Then I should not see "Series"
 
   @new-search
-  Scenario: Synning a canonical tag used on a bookmarked series should move the 
-  bookmark to the new canonical's bookmark listings; de-synning should remove it
+  Scenario: Synning a canonical tag used on bookmarked series and external works
+  should move the bookmarks to the new canonical's bookmark listings; de-synning
+  should remove them
     Given I am logged in as "author"
       And a canonical fandom "Veronica Mars"
```
ariana-paris: Is see these step names are a bit light on readability - this reads `Given I am logged in as "author" and a canonical character "Veronica Mars"` - probably lacking an "exists" in the step definition. But that's a pre-existing problem and it isn't as if we have a "business" that actually reads these scenarios!

app/models/tag.rb
```diff
@@ -688,10 +689,12 @@ def recategorize(new_type)
...
     bookmark_ids = all_bookmark_ids
     yield if block_given?
     reindex_all_works(work_ids)
     reindex_all_series(series_ids)
+    reindex_all_external_works(all_filtered_external_work_ids)
```
ariana-paris: This should use the `external_work_ids` instead of calling the `all_filtered_external_work_ids` method a second time.

features/other_b/bookmark_indexing.feature
sarken: I'd consider the bookmark creation to be a set up step, since we're not interested in the creation of the bookmark, but the wrangling changes (we don't have "Given a bookmark of the external work/series" steps). So what I'll actually do is get rid of the `When I go to the bookmarks tagged...`bit -- I have no idea why I included that -- and make the switch to `When` when we get to the tag wrangler bit.

ariana-paris: All sounds good, but I wouldn't worry about it on my account since I only commenting on the `And`/`When` which sounded odd.

sarken:
There were test failures, so I updated the tests.

features/step_definitions/external_work_steps.rb
```diff
@@ -20,14 +20,11 @@
   fill_in("bookmark_tag_string", with: DEFAULT_BOOKMARK_TAGS)
 end
 
-Given /^I bookmark the external work "([^\"]*)"$/ do |title|
-  visit new_external_work_path
-  fill_in("bookmark_external_url", with: DEFAULT_EXTERNAL_URL)
-  fill_in("bookmark_external_author", with: DEFAULT_EXTERNAL_CREATOR)
+Given /^I bookmark the external work "([^\"]*)"(?: with fandom "([^"]*)")?(?: with character "([^"]*)")?$/ do |title, fandom, character|
```
redsummernight: This new step ends up unused after you switched to factories, should we keep it?

app/models/tag.rb
```diff
@@ -925,15 +947,20 @@ def remove_filter_taggings(old_filter_id=nil)
...
     return unless meta_tag.present?
-    self.filtered_works.each do |work|
-      unless work.filters.include?(meta_tag)
-        work.filter_taggings.create!(inherited: true, filter_id: meta_tag.id)
-        RedisSearchIndexQueue.reindex(work, priority: :low)
-        IndexQueue.enqueue_ids(Series, work.series.pluck(:id), :background) if $rollout.active?(:start_new_indexing)
+    filtered_items = self.filtered_works + self.filtered_external_works
+    filtered_items.each do |item|
+      unless item.filters.include?(meta_tag)
+        item.filter_taggings.create!(inherited: true, filter_id: meta_tag.id)
```
tickinginstant: When initially creating an ExternalWork (or when editing an ExternalWork to add a new tag) no FilterTaggings are added for any meta-tags. So adding the meta-tags here means that ExternalWorks are going to have inconsistent filters: the ExternalWorks added before the meta-tag was added will have the meta-tag as a filter, but the ExternalWorks added after the meta-tag was added won't have the meta-tag as a filter.

features/other_b/bookmark_indexing.feature
```diff
@@ -60,48 +60,45 @@ Feature: Bookmark Indexing
...
+      And I should see "Series to Bookmark"
     When I de-syn the tag "Veronica Mars" from "Veronica Mars (TV)"
       And the tag "Veronica Mars" is canonized
       And I go to the bookmarks tagged "Veronica Mars (TV)"
-    Then I should not see "Telling Stories"
+    Then I should not see "External Work to Bookmark"
+      And I should not see "Series to Bookmark"
     When I go to the bookmarks tagged "Veronica Mars"
-    Then the 1st bookmark result should contain "Telling Stories"
+    Then I should see "External Work to Bookmark"
```
tickinginstant: The fact that this step works is somewhat deceptive. At this point, the ExternalWork in the Veronica Mars tag doesn't have any FilterTaggings related to the Veronica Mars tag, the way that it ought to. (This is because `add_filter_taggings`, which is the only thing that creates FilterTaggings when a tag goes from non-canonical to canonical, currently doesn't process ExternalWorks.) The only reason that this ExternalWork shows up in the Veronica Mars tag is because it's directly tagged with "Veronica Mars," and the filter_ids field for all Taggable items is defined to be `(tags.pluck(:id) + filters.pluck(:id)).uniq`.

If you attempted to syn Veronica Mars to "Veronica Mars (TV)" again after this step, the ExternalWork would not show up in the "Veronica Mars (TV)" tag, because `add_filter_taggings` wouldn't add the new filter, and `move_filter_taggings_to_merger` would have no FilterTaggings to transfer.

sarken: Yeah. The issue with `add_filter_taggings` had a separate Jira issue and should be fixed by the next PR, #3294.

app/models/tag.rb
redsummernight: Sarken addressed both of your comments on a new pull request, #3326.

PR #3296, opened 2018-04-25
AO3-3468 Reindex user's works and series after username change

redsummernight:
## Issue

## Purpose

Works and series are indexed with the author's byline, which has the old username.

## Testing

See issue.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -513,9 +513,11 @@
 
   def reindex_user_works
     # reindex the user's works to make sure they show up on the user's works page
-    works.each do |work|
-      IndexQueue.enqueue(work, :main)
-    end
+    IndexQueue.enqueue_ids(Work, works.pluck(:id), :main)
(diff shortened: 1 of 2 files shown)


app/models/user.rb
```diff
@@ -558,21 +560,26 @@ def update_pseud_name
...
+          old_pseud.save!(validate: false)
+        else
+          # shouldn't be able to get here, but just in case
+          Pseud.create!(name: login, user_id: id)
+        end
       end
     end
+
+    # Works, series, and pseuds are indexed with the user's byline,
+    # which has the old username, so they all need to be reindexed.
```
sarken: I was going to ask about bookmarks, but honestly, the issue just says "works" so we can fix bookmarks in 0.9.212 >.>

redsummernight: Once more with feelings (and bookmarks)! #3300.

PR #3298, opened 2018-04-26
AO3-5397 Don't automatically check "Remember me" on /login

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5397

## Purpose

Makes it so "Remember me" on the non-JavaScript form on /login isn't pre-checked when the page loads.

## Testing

Go to the page, look at the thing.


redsummernight:
This now has conflicts with the Devise changes.

PR #3299, opened 2018-04-26
AO3-4377 Change field ids on header login form

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4377

## Purpose

Makes it so the log in box in the header uses different ids than the login box in the body of the /login page. This makes sure the HTML is valid and plays nice with the accessibility feature in most browsers where clicking on the label places focus on the field.

## Testing

Refer to Jira


redsummernight:
This now has conflicts with the Devise changes.

PR #3301, opened 2018-04-30
AO3-3726 Update series date indexing when adding and removing chapters

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-3726

## Purpose

Updates the Date Updated for series bookmarks when a work in the series has chapters added or removed.

## Testing

Refer to Jira

### DIFF
--- a/app/models/chapter.rb
+++ b/app/models/chapter.rb
@@ -92,7 +92,7 @@ class Chapter < ApplicationRecord
 
   after_commit :update_series_index
   def update_series_index
-    return unless work.series.present? && should_reindex_series?
+    return unless work&.series.present? && should_reindex_series?
     work.series.each do |series|
       # ES UPGRADE TRANSITION #
       # Remove logic pertaining to which indexing we're using
--- a/app/models/chapter.rb
+++ b/app/models/chapter.rb
@@ -93,10 +93,7 @@ class Chapter < ApplicationRecord
   after_commit :update_series_index
   def update_series_index
     return unless work&.series.present? && should_reindex_series?
-    IndexQueue.enqueue_ids(Series, work.series.pluck(:id), :main)
-    work.series.each do |series|
-      IndexQueue.enqueue_ids(Bookmark, series.bookmarks.pluck(:id), :background)
-    end
+    work.serial_works.each(&:update_series_index)
   end
 
   def should_reindex_series?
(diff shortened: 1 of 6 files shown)


app/models/chapter.rb
```diff
@@ -90,6 +90,29 @@ def fix_positions_after_destroy
     end
   end
 
+  after_commit :update_series_index
+  def update_series_index
+    return unless work&.series.present? && should_reindex_series?
+    work.series.each do |series|
+      # ES UPGRADE TRANSITION #
```
redsummernight: IndexQueue checks rollout to see which ES version to use, so you don't need to check that here, you can just do:

```rb
IndexQueue.enqueue_ids(Series, work.series.pluck(:id), :main)
work.series.each do |series|
  IndexQueue.enqueue_ids(Bookmark, series.bookmarks.pluck(:id), :background)
end
```

sarken: Oh, nice, thanks!

It's _technically_ outside the scope of this, but I think I'll change it in https://github.com/otwcode/otwarchive/blob/8a28b63145973fa40701f027931872896765641a/app/models/serial_work.rb#L28 as well, since that's how I ended up with this >.>

app/models/serial_work.rb
```diff
@@ -26,9 +26,9 @@ def delete_empty_series
 
   # Ensure series bookmarks are reindexed when a new work is added to a series
   def update_series_index
+    series.enqueue_to_index
     if $rollout.active?(:start_new_indexing)
-      series.enqueue_to_index
-      series.bookmarks.each(&:enqueue_to_index)
+      IndexQueue.enqueue_ids(Bookmark, series.bookmarks.pluck(:id), :main)
```
redsummernight: IndexQueue will index for both versions of ES (if we have `start_new_indexing` but not `stop_old_indexing`), so `update_series_index` will end up indexing the bookmarks 4 times: once per ES version, and once per queue (main and background). I don't think we can do background for 0.90, main for 6 here.

If we want to be extra careful, we could use background, and add an "upgrade transition" comment for switching to main later.

sarken: Hmm. That would be a problem with the code that's currently on staging, too, wouldn't it?

redsummernight: It is. :<

sarken: I'm afraid to ask if there's similar code in other places as well.

redsummernight: This is the only place I can find that uses `stop_old_indexing` like this.

PR #3303, opened 2018-05-02
AO3-5384 Update sanitize gem

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5384

## Purpose

Updates the sanitize gem and its dependencies.

## Testing

Refer to Jira


DreamSea:
Was there also a change to `Gemfile` for this (if that's where `Gemfile.lock` changes came from)?

`script/gem_security.sh` also seems to reference the CVE, but not sure if that's something that gets updated as a group once all the gem updates are in.

DreamSea:
> conflicts vs one go

ahh, got it

> gemfile

when i originally made the comment, id completely forgotten/overlooked that just calling update on the gem would have made the same changes since it wasnt bounded by a ~> declaration (and assuming other dependencies checked out i guess). whoops :sweat_smile:

PR #3305, opened 2018-05-02
AO3-5384 Update sanitize to 4.6.3 or later

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5384

## Purpose

Update the library we use to check html.

## Testing

- As an admin, check that you can add a screencast URL to an FAQ question and ensure it is not enclosed in paragraph tags
- Check that you can add a video embed in a work, but not in a work summary
- Check that you can use CSS classes in a work content, but not a work summary
- Make sure you can import a work and it looks as expected
- Make sure that if you post a work using the HTML editor without using paragraph tags, they get added


sarken:
Done in #3343!

PR #3306, opened 2018-05-06
AO3-5329 work search summary

DreamSea:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5329

## Purpose

Puts work status and crossover selection info into new search's summary

## Testing

See Jira @ What should happen

## References

Issue mentions possible rename from "Only incomplete works" to "Works in Progress" [AO3-5263](https://otwarchive.atlassian.net/browse/AO3-5263?focusedCommentId=353440&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-353440), which seems to be the intended direction going forward. Wasn't sure if that also applied to what is shown in the "You searched for:" summary.

Kept `%w(matches).include?` existing style, but guessing with new search can move to specific/only checks for T/F?

re: houndci-bot
I'll make an update following hound recommendations sometime tomorrow with assumption it can be trusted (not familiar enough with ruby yet to be able to call out false alarms)


features/search/works_info.feature
```diff
@@ -56,6 +56,30 @@ Feature: Search works by work info
       And I should see "fifth"
       And I should see "I am <strong>er Than Yesterday & Other Lies"
 
+  @new-search
+  Scenario: Search by status
+    Given I have loaded the fixtures
+    When I am on the search works page
+      And I choose "Only complete works"
+      And I press "Search" within "#new_work_search"
+    Then I should see "You searched for: Complete"
```
redsummernight: The new tests look good!

Can you add a few checks that certain works show up for each new search (like `And I should see "3 Found"` from the other tests)? It's fine if you don't have time though, since it's not strictly related to the current issue.

features/search/works_stats.feature
```diff
@@ -85,7 +85,6 @@ Feature: Search works by stats
 
   # This is basically the same scenario as above, but the new search has
   # changed the "Complete" checkbox into a "Only complete works" radio button,
```
redsummernight: Very minor: this should end in a period now.

features/search/works_info.feature
DreamSea: Sounds good. Is there a preference on using one of the work sets (Battle Set, etc) vs plain ol' fixtures? Seems to work for another test later down (maybe the one you're referencing):
```
  Scenario: Search and sort by title
    Given I have loaded the fixtures
...
      And I should see "3 Found"
```

features/search/works_stats.feature
DreamSea: :+1: 

re: #3281 
haha, yea, it's probably because that one was still vaguely in memory that I took notice of the `* Or "Works in Progress"` comment.

To confirm - so original "Only incomplete works" was represented as "Incomplete" in the summary. New version "Works in progress only" should be represented as "Works in Progress" in the summary?

Since the naming relies on the other pull, I guess for this pull I'll keep the radio button selecting steps the same (`And I choose "Only incomplete works"`, etc.), and then later on depending on whichever pull makes it in first, I/whoever can go back and fix what will become a failing step.

features/search/works_info.feature
redsummernight: "Battle Set" has more steps to set up a challenge and takes longer so let's go with "loaded the fixtures" if possible.

features/search/works_stats.feature
redsummernight: Someone else will confirm the new wording, but waiting for the other pull sounds good.

app/models/search/work_search_form.rb
```diff
@@ -126,8 +126,15 @@ def summary
     unless tags.empty?
       summary << "Tags: #{tags.uniq.join(", ")}"
     end
-    if %w(1 true).include?(self.complete.to_s)
```
redsummernight: For the new options, just checking for T/F is fine, but "Complete" is there in old search so we still need to check for `1 true`.

You can add a comment:

```
# ES UPGRADE TRANSITION #
# Remove checks for 1 and true
```
so we can deal with it later.

features/search/works_info.feature
redsummernight: You don't need to, I think this test is fine.

app/models/search/work_search_form.rb
DreamSea: Just to confirm, old search would still use `models/search/work_search_form.rb`? I had thought that old search stuff picked `models/work_search.rb` as the model. :thinking:

redsummernight: Oh you're right! This is good, I got it backwards. ^^

DreamSea: Haha, all good. Keep me honest :coffee:

PR #3308, opened 2018-05-15
AO3-5411 Reduce REDIS calls in IndexSweeper.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5411

## Purpose

This PR revises how IndexSweeper stores errors in order to reduce the load on REDIS when there are a large number of errors.

Currently, all errors are stored in one of three different strings on REDIS (one for errors that have occurred once, one for errors that have occurred twice, and one for errors that have occurred three or more times).

This pull request stores all errors in a single REDIS hash. Errors are arranged in the hash by the ID of the document that experienced the error. The value associated with each document ID is a JSON encoding of an array containing information about all errors (including a separate count for each error, which increments every time we see the error).

By separating the errors and hashing them with their ID as a key, it's much easier to delete the errors associated with a document when the document is successfully indexed. (We just have to delete the key for that document.) In addition, it's much easier to figure out whether a particular error has occurred before -- instead of looking through all three error stores for the correct combination of ID and error, we know exactly where the error is stored.

By using `REDIS.hmset` and `REDIS.hmget` and `REDIS.hdel` (which can take multiple arguments), it's possible to ensure that each batch will only send three commands to REDIS. In addition, this way of storing the error data means that the size of the data transmitted to/received from REDIS is proportional to the number of items in the batch, and doesn't have any dependence on the total number of previous errors. This means that in situations where every single Elasticsearch query is triggering an error, the amount of data received by REDIS should be linear in the number of errors, instead of quadratic.

## Testing

This is, unfortunately, very difficult to test on staging, since it requires generating Elasticsearch errors. I think the only way to do that at the moment is to modify the code to deliberately introduce an error.

### DIFF
--- a/app/models/search/index_sweeper.rb
+++ b/app/models/search/index_sweeper.rb
@@ -15,14 +15,16 @@ class IndexSweeper
     @indexer = indexer
     @success_ids = []
     @rerun_ids = []
-
-    ensure_failure_stores
   end
 
   def process_batch
+    load_errors
+
     @batch["items"].each do |item|
       process_document(item)
     end
+
+    save_errors
 
     if @success_ids.present? && @indexer.respond_to?(:handle_success)
       @indexer.handle_success(@success_ids)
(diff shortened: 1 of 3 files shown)


app/models/search/index_sweeper.rb
```diff
@@ -35,74 +37,74 @@ def process_batch
...
+    @batch_ids ||= @batch["items"].map { |item| item.values.first["_id"].to_s }
   end
 
-  def all_stores_empty?
-    get_store_as_json("first").empty? &&
-      get_store_as_json("second").empty? &&
-      get_store_as_json("permanent").empty?
+  # Load information about previous errors for all the items in this batch.
+  def load_errors
+    errors = REDIS.hmget("#{@indexer}:failures", batch_ids).
```
redsummernight: If you use `mapped_hmget` and `mapped_hmset` you won't have to do the array-hash conversions yourself.

tickinginstant: Thanks for the suggestion! I changed the two calls in IndexSweeper.

PR #3316, opened 2018-05-23
AO3-5416 Prompt people to agree to new TOS version

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5416

## Purpose

If JavaScript is enabled, 1.5 seconds after the page loads, a screen will fade in asking people to agree to the new TOS. When they check the box to indicate they have read the TOS, the accept button will be activated. When they press the button, the screen will fade out. 

* If the person is a guest, it will set a 20-year cookie called accept_tos noting the version of the TOS the user has agreed to (the version number is the date the TOS version was approved, e.g. 20180523). 
* If the person is a user, it will record the version number in the database. 
**Also, if someone accepts the TOS while creating an account, the version will be recorded in the database.

If the cookie or database value matches the current TOS version set in the application controller, the screen will not be shown on subsequent pages.

The screen will never appear in automated tests (because we do not have time to fix every test that uses JavaScript) or on the DMCA, TOS, Abuse, or Support pages. 

## Testing

Refer to Jira

## References

When the new TOS is uploaded, it will display its date of approval. It will also contain a friendly code comment telling people to update the current TOS version in the application controller so the "please accept the TOS" screen gets shown again.


sarken:
Ugh, hold that thought. I found a mistake :P

sarken:
Okay, I think I fixed it.

redsummernight:
@bingeling noted that the cookie should be applied site-wide instead of just on the current page.

PR #3318, opened 2018-05-23
AO3-5420 Increase line-height for stats links in blurbs

cruxicheiros:
## Issue
https://otwarchive.atlassian.net/browse/AO3-5420

## Purpose
Currently, in some browsers, the underlines of stats links touch the line of text below them. This can be seen in screenshots on the issue page. Setting `line-height` to `1.4` within `.blurb dl.stats` ensures that the link underlines don't touch the line below.

## Testing
Issue is currently present when viewing blurbs in Firefox 60.0.1 on 64-bit Windows 10. If you compare the version without `line-height: 1.4;` vs the version with it, you will see that gaps are added between the lines of text in the stats blurbs.

## References
Was previously discussed in Slack

## Credit
Credit as Hazel Seanor


public/stylesheets/site/2.0/13-group-blurb.css
```diff
@@ -73,6 +73,7 @@ li.blurb:after, .blurb .blurb:after {
 
 .blurb dl.stats {
   float: right;
+  line-height: 1.4;
```
sarken: Could you make this 1.429 to fit in a bit better with our [em scale](http://otwcode.github.io/docs/front_end_coding/em-scale.html)?

sarken: And approved -- thank you!

PR #3323, opened 2018-05-28
AO3-5428 Specify number of shards

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5428
## Purpose

Sets the number of shards on the elastic search cluster.

## Testing

Get an admin to use curl  for example:

```james_@elastic04:~$ curl -X GET "localhost:8201/_cat/indices?v"
health status index                    uuid                   pri rep docs.count docs.deleted store.size pri.store.size
green  open   ao3_production_tags      EOfrTO0AS-yvltks0tsclg   5   1    6700754      1963506      2.2gb            1gb
green  open   ao3_production_bookmarks PuHTs8c3Q06Tz2HGx7GMzg   5   1   63794528            0     36.7gb         18.5gb
green  open   ao3_production_pseuds    mOT6HShOSOav7jf5D75xSQ   5   1    1571111       212730        1gb        519.9mb
green  open   ao3_production_works     5096D7qBTyG1HzC4T7nnnA   5   1    3889071       936128     20.2gb          9.8gb```
Each index has 5  primary shards

### DIFF
--- a/app/models/search/work_indexer.rb
+++ b/app/models/search/work_indexer.rb
@@ -4,6 +4,15 @@ class WorkIndexer < Indexer
     "Work"
   end
 
+  def self.index_all(options={})
+    unless options[:skip_delete]
+      delete_index
+      create_index(shards: 9)
+    end
+    options[:skip_delete] = true
+    super(options)
+  end
+
   def self.mapping
     {
       "work" => {
--- a/app/models/search/work_indexer.rb
+++ b/app/models/search/work_indexer.rb
@@ -4,10 +4,10 @@ class WorkIndexer < Indexer
     "Work"
   end
 
-  def self.index_all(options={})
+  def self.index_all(options = {})
     unless options[:skip_delete]
       delete_index
-      create_index(shards: 9)
+      create_index(shards: 12)
     end
     options[:skip_delete] = true
     super(options)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: app/models/search/work_indexer.rb [hunk 1]
```diff
@@ -4,6 +4,15 @@ def self.klass
     "Work"
   end
 
+  def self.index_all(options={})
```
redsummernight: Instead of overriding `create_index` you could override `settings`.

In the parent class:

https://github.com/otwcode/otwarchive/blob/a5487041e1e064521262ba34ef14d814b67ad8cc/app/models/search/indexer.rb#L40-L50

So in this subclass, settings can return `index: { number_of_shards: 9 }`, and we switch `merge` to `deep_merge` in the parent class.

zz9pzza:
`james_@elastic04:~$ curl -X GET "localhost:8201/_cat/indices?v"
health status index                    uuid                   pri rep docs.count docs.deleted store.size pri.store.size
green  open   ao3_production_tags      EOfrTO0AS-yvltks0tsclg   5   1    6722783      2030148      2.2gb          1.1gb
green  open   ao3_production_bookmarks PuHTs8c3Q06Tz2HGx7GMzg   5   1   81668359     15543265     52.3gb           26gb
green  open   ao3_production_pseuds    mOT6HShOSOav7jf5D75xSQ   5   1    1574976       280266        1gb        555.9mb
green  open   ao3_production_works     5096D7qBTyG1HzC4T7nnnA   5   1    3900985      2539889       31gb         14.8gb`

app/models/search/work_indexer.rb
redsummernight: On second thoughts, doing so will break the shards override for `create_index`.

PR #3325, opened 2018-05-29
AO3-5429 Add tasks for reindexing in new world

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5429

## Purpose

Add some task to catch up on broken indexes

## Testing

Needs to be have a shell account

### DIFF
--- a/lib/tasks/search.rake
+++ b/lib/tasks/search.rake
@@ -1,4 +1,7 @@
 namespace :search do
+
+  BATCH_SIZE = 1000
+
   desc "Reindex tags"
   task(:index_tags => :environment) do
     TagIndexer.index_all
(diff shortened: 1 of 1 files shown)


lib/tasks/search.rake
```diff
@@ -1,18 +1,59 @@
...
+  task(index_bookmarks: :environment) do
     BookmarkIndexer.index_all
   end
+  desc 'Reindex all'
+  task timed_all: %i[timed_works timed_tags timed_pseud timed_bookmarks] do
+  end
+  desc 'Reindex bookmarks'
+  task timed_bookmarks: :environment do
+    time = ENV['TIME_PERIOD'] || 'NOW() - INTERVAL 1 DAY'
+    ExternalWork.where("external_works.updated_at >  #{time}").find_in_batches(batch_size: BATCH_SIZE) do |group|
```
redsummernight: You could use `.select(:id).find_in_batches` if you only need the IDs.

PR #3326, opened 2018-06-02
AO3-5253 & AO3-4093 Make non-work bookmarks obey wrangling changes

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5253, replacing #3293
https://otwarchive.atlassian.net/browse/AO3-4093, replacing #3294 

## Purpose

Slices, dices, juliennes itself into a floor wax and a dessert topping. Behold, it is a kitchen sink pull request! :sob:

- Updates `check_filter_taggings` and `add_filter_taggings` in the ExternalWork model so it adds filter taggings for meta tags when the external work is created or updated
- Updates `add_filter_taggings`, `remove_filter_taggings`, `inherit_meta_filters`, and `remove_meta_filters` in the Tag model to make sure external works get updated
  - The `add_filter_taggings` change resolves AO3-4093
- And then adds external works to the list of things that should be reindexed when `reindex_taggables` is called in the Tag model
- Reworks `all_filtered_series_ids` to make sure series get reindexed when a synonymous tag is moved from one merger (canonical tag) to another

## Testing

Create new (sadly, this is not a retroactive fix) bookmarks of external works and series and make sure those bookmarks appear on the bookmark listings for any meta tags or mergers that already exist _or_ that are added after the bookmark is created. Also check that the bookmarks _don't_ appear on the listings for any meta tags or mergers you remove after the bookmarks are created. Note that it may take several minutes for things to be reindexed and caches to expire.

One lingering issue: this does _not_ fix the problem where decanonizing a subtag fails to remove bookmarks from its former meta tag's bookmark listing. You still have to remove the meta tag first, then decanonize. (There is no issue for this problem specifically; I'll make one.)

## References

See above for pull requests replaced by this one; they have some comments that might be worth reading.


features/other_b/bookmark_indexing.feature
```diff
@@ -60,48 +60,71 @@ Feature: Bookmark Indexing
...
-    Then I should not see "Telling Stories"
+    Then I should not see "BookmarkedExternalWork"
+      And I should not see "BookmarkedSeries"
     When I go to the bookmarks tagged "Veronica Mars"
-    Then the 1st bookmark result should contain "Telling Stories"
+    Then I should see "BookmarkedExternalWork"
+      And I should see "BookmarkedSeries"
+    When I syn the tag "Veronica Mars" to "Veronica Mars (TV)"
+      And I go to the bookmarks tagged "Veronica Mars (TV)"
+    Then I should see "BookmarkedSeries"
```
redsummernight: Should we see "BookmarkedExternalWork" here along with "BookmarkedSeries"?

features/other_b/bookmark_indexing.feature
```diff
@@ -60,48 +60,71 @@ Feature: Bookmark Indexing
...
-    Then the 1st bookmark result should contain "Telling Stories"
+      And I subtag the tag "Laura Roslin" to "Laura"
+      And I go to the bookmarks tagged "Laura"
+    Then I should see "BookmarkedExternalWork"
+      And I should see "BookmarkedSeries"
+    When I remove the metatag "Laura" from "Laura Roslin"
+      And I go to the bookmarks tagged "Laura"
+    Then I should not see "BookmarkedExternalWork"
+      And I should not see "BookmarkedSeries"
+    When I go to the bookmarks tagged "Laura Roslin"
```
redsummernight: This has no followup then steps.

tickinginstant:
This seems to fix the issues that I had!

Unfortunately, while I was checking an edge case, I noticed another issue with editing tags on External Works. It's not new in this pull request, but it may affect testing.

Any time a tag is removed from a Taggable, `remove_filter_tagging` is called. If the tag that was removed has an associated filter, then the ExternalWork's `tags` association will be loaded by `ExternalWork.remove_filter_tagging`.

If a new tag is added at the same time the old tag is removed, then the `tags` association won't be updated with the new tag (because it was loaded and cached right after the old tag was deleted, and before the new tag was added). Thus, in `check_filter_taggings`, the filters will be calculated based on outdated information, and the filter for the new tag will be missing.

This is a scenario that causes the failure on my machine:
```
  Scenario: Replacing one tag with another.
    Given basic tags
      And a canonical fandom "Alice Smith"
      And a synonym "Alice" of the tag "Alice Smith"
      And a canonical fandom "Beth Jones"
      And a synonym "Beth" of the tag "Beth Jones"
      And I am logged in
      And I bookmark the external work "Tricky" with fandom "Alice"

    When I go to the bookmarks tagged "Alice Smith"
    Then I should see "Tricky"

    When I am logged in as an admin
      And I view the external work "Tricky"
      And I follow "Edit External Work"
      And I fill in "Fandoms" with "Beth"
      And I press "Update External work"
    Then I should see "External work was successfully updated."
      And I should see "Beth"
      And I should not see "Alice"

    When I go to the bookmarks tagged "Alice Smith"
    Then I should not see "Tricky"

    When I go to the bookmarks tagged "Beth Jones"
    Then I should see "Tricky"
```

(Works avoid this problem because the `before_destroy` hook in the Tagging class runs `remove_filter_tagging` an additional time for each deleted tag. Because the `taggable` association in Taggings isn't set up as the inverse of the `taggings` association in Works, the call from the Tagging class loads a fresh copy of the work for every tagging deleted or added. So the first time `remove_filter_tagging` is run -- from the Tagging class -- the `tags` association is loaded fresh, resulting in more SQL queries and processing time, but also the correct outcome. The second time `remove_filter_tagging` is run -- from the Taggable class -- it's a no-op. And there's no equivalent of `check_filter_taggings`.)

PR #3327, opened 2018-06-02
AO3-5432 Remove metatag filter taggings when a subtag is decanonized

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5432

## Purpose

If a tag that is being decanonized has metatags, make sure to remove the filter taggings for both the tag _and_ its metatags. 

## Testing

Refer to Jira


sarken:
Oops, wait, I also need to test the works page. Grr.

PR #3330, opened 2018-06-03
AO3-5432 Remove metatag filter taggings when a subtag is decanonized

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5432

## Purpose

If you decanonize a tag that has metatags, the filter taggings for both the tag and its metatags will be removed from works/external works/bookmarks/series. (Before it was just the filter tagging for the tag itself.)

## Testing

Refer to Jira


tickinginstant:
I doubt this will show up in automated tests, but I think this change may introduce new timing-based issues when de-canonizing a tag. Specifically, in certain circumstances, I believe it may fail to delete the filter-taggings from synonyms of the tag being de-canonized. So the de-canonized tag may still show up in the sidebar sometimes, in listings that include a lot of the former synonyms.

Pressing the "Make tag non-canonical and unhook all associations" button asynchronously calls two methods from `before_update` hooks: `remove_filter_taggings` and `remove_canonical_associations`. The call to `remove_canonical_associations` updates all of the tag's synonyms to set their `merger_id` to `nil`. Those synonyms won't have their value of `canonical` updated, just the `merger_id`. So the only `before_update` hook they'll trigger is `update_filters_for_merger_change`, which will only call `remove_filter_taggings` for works tagged with that synonym if the merger is still canonical. The merger which is in the process of becoming non-canonical.

It's basically a race: if the transaction updating the original tag completes (and the changes propagate to the read-only DB servers) before the Resque workers run `remove_canonical_associations`, then the only call to `remove_filter_taggings` is for the original tag. Which is guaranteed to always remove all filter-taggings for all synonyms on master, but not with the changes made in this PR.

sarken:
Ugh, I forgot that option still existed (we removed it from the regular wrangling interface). I'm going to close this and release the issue back into the wild with a link to your comment.

PR #3336, opened 2018-06-08
AO3-4194 Associate fields/labels of request invites form

JenRodge:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4194

## Purpose

Associates the fields and labels of the Request Invites form so that when the labels are clicked, the proper field receives focus.

## Testing

1. Enable invite requests
2. Log in
3. Go to /user_invite_requests/new
4. Click on either form label.  The associated field should now have the focus.

## Credit

Jen Rogers (she/her)


sarken:
Hi, Jen!

Thank you so much for submitting this pull request -- it looks great. Would you be able to update [these two lines in our tests](https://github.com/otwcode/otwarchive/blob/5c610798f66ed4466f6a7eccca8d6205dc2d0d98/features/other_a/invite_request.feature#L10-L11) to use the label text instead of the input ids? (We generally prefer the label text, which better illustrates actual user experience, but were forced to use ids in the tests due to this issue.)

It looks like you've made an account on our issue tracker under the name Jen Rogers, so I've assigned the issue to you. In case you feel like working on anything else, I've also given you permissions to assign issues, add comments, create attachments, and transition issues. (But no pressure!)

Since we're wrapping up a major upgrade to our searching and filtering code, it may be a little while before we can merge this, but we're very grateful for your contribution and will merge it as soon as we can.

Thanks again!

JenRodge:
Done, with a caveat: Elasticsearch isn't playing well with my Vagrant instance (and it doesn't seem worth recovering it mid-transition), so I am currently unable to run all Cucumber tests.  However, this does seem to pass all tests pertinent to the modified features.

Also, at the risk of straying a bit from the reported bug, the 1fd8e3ee9873268321ef53062ecd49d9a5d8db05 commit (with an unfortunate typo in the commit message) also addresses this same form labeling issue on the New Work and New Chapter pages.

sarken:
The test fixes look good, thank you! However, we generally prefer pull requests stay within the scope of the issue whenever possible -- it makes our reviewing and manual testing processes quicker and smoother if we keep changes as isolated as possible. I think we can make an exception this time and update the issue to cover the changes to the "Chapter X of" and "Chapter Title" labels (in fact, I'll do that now), but the fix for "Work Text" and "Chapter Text" may need a bit more consideration. While it does properly associate the labels with the field for the HTML editor, that field is hidden when a user toggles to the Rich Text editor. Your fix is _probably_ enough of an improvement there wouldn't be any harm in deploying it, but whenever we're uncertain about something, we prefer to be cautious and keep it separate from any other changes.

sarken:
https://github.com/otwcode/otwarchive/pull/3607 resolves merge conflicts and removes unrequested changes; credit should still go to Jen Rogers.

PR #3337, opened 2018-06-08
AO3-3547 Add option to add notes to user

hatal175:
## Issue

https://otwarchive.atlassian.net/browse/AO3-3547

## Purpose

This adds a 'note' option to the admin abuse administration page.

## Testing

I hope that this was the intention in the bug request. I've added a feature test for the new feature.

## Credit
Tal.

Please use he.


sarken:
This looks like exactly what we want -- thanks! We do have a refactor of this code coming up, though, so while the general concept and interface are spot on, the code will need to be revised once #2986 is merged. I've labeled this with "Deferred or Superseded" until then.

sarken:
#2986 has been merged

PR #3338, opened 2018-06-09
AO3-5443 Make recent bookmarks clear bookmarkable stats and Save button

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5443

## Purpose

On new Elasticsearch's tag bookmark pages, the list of a bookmarkable item's recent bookmarks was overlapping the bookmarkable's stats and Save/Saved button. This makes the recent bookmarks clear those things.

## Testing

Refer to Jira

### DIFF
--- a/public/stylesheets/site/2.0/13-group-blurb.css
+++ b/public/stylesheets/site/2.0/13-group-blurb.css
@@ -335,9 +335,12 @@ blurbs on the manage collection items pages, mostly reseting styles inherited fr
 	top: 28px;
 }
 
+.bookmark div.user, .bookmark div.recent {
+  clear: right;
+}
+
 .bookmark .user {
   border: 1px solid #ccc;
-  clear: right;
   margin-top: 0.643em;
   overflow: hidden;
 }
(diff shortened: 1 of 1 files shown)


public/stylesheets/site/2.0/13-group-blurb.css
```diff
@@ -335,9 +335,12 @@ blurbs on the manage collection items pages, mostly reseting styles inherited fr
 	top: 28px;
 }
 
+.bookmark div.user, .bookmark div.recent {
```
sarken: We don't need it on `li.user`, which is was also getting applied to without the div.

public/stylesheets/site/2.0/13-group-blurb.css
```diff
@@ -383,7 +386,8 @@ blurbs on the manage collection items pages, mostly reseting styles inherited fr
 }
 
 .bookmark .recent .index {
-    width: 100%;
+  float: none;
```
redsummernight: `float:none` is also inherited from:

https://github.com/otwcode/otwarchive/blob/5c610798f66ed4466f6a7eccca8d6205dc2d0d98/public/stylesheets/site/2.0/26-media-narrow.css#L109-L115

unless you want to explicitly have it here.

sarken: Yeah, we need it in all widths, or else `<ul class="bookmark index group">` ends up escaping `<div class="recent module group">` (so if you, for example, put a blue background on the div, you won't see it because the div's not actually holding the ul)

PR #3339, opened 2018-06-09
AO3-5428 Fix method call for create_index shard parameter

ariana-paris:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5428

## Purpose

The previous version using `shards:` didn't work because the `shards` parameter is a single value, not a hash.


zz9pzza:
I have woken up and  reviewed it and I would prefer it to be merged :)

PR #3345, opened 2018-06-15
AO3-5447 Give 404 instead of 500 for works pages for nonexistent users

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5447

## Purpose

* Gives a 404 error (by way of `ActiveRecord::RecordNotFound`) if you try to access the works page for a user that doesn't exist, regardless of whether you're aiming for the user's works page or a pseud's
* Finishes testing the various `load_owner` possibilities, since that's the method I was editing 
* Cleans up tests in controllers/works/default_rails_actions_spec.rb

## Testing

Refer to Jira


app/controllers/works_controller.rb
```diff
@@ -797,7 +797,9 @@ def mark_as_read
   def load_owner
     if params[:user_id].present?
       @user = User.find_by(login: params[:user_id])
-      if params[:pseud_id].present?
+      if @user.blank?
+        raise ActiveRecord::RecordNotFound, "Couldn't find user named '#{params[:user_id]}'"
```
redsummernight: Instead of manually raising the exception, we could use `find_by!`.

spec/controllers/works/default_rails_actions_spec.rb
```diff
@@ -301,18 +301,106 @@ def call_with_params(params)
...
             get :index, params: { tag_id: @fandom.name }
             expect(assigns(:works).items).to include(@work)
             expect(assigns(:works).items).not_to include(@work2)
           end
 
         end
 
+        context "when tag is a synonym" do
+          before do
+            @fandom_synonym = create(:fandom, merger: @fandom)
```
redsummernight: We should use `let` to avoid leaking state between tests in instance variables. You don't need to change all the existing instance variables though.

sarken:
This needs updating

sarken:
I took a stab at the merge conflicts. 🤞 I didn't make a mess. I hate past me.

PR #3346, opened 2018-06-16
AO3-5375 bookmark search filter languages

DreamSea:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5375

## Purpose

Adds dropdown for (elasticsearch 6) searching/filtering bookmarks by work language

## Testing

see Jira

--- 

slight WIP while trying to debug consistently failing travis test that appears to pass locally :thinking: but opening for review with a few questions~
- clarification re: `on search page - it can be before or after the field for selecting the type` https://test.archiveofourown.org/bookmarks/search, is it referring to Work/Series type or Bookmark type? I guessed Work/Series type since 'work language' rather than 'bookmark language'.
- any link_to_help wanted for this? A guess copying https://test.archiveofourown.org/help/work-search-language-help.html would change the title to `Bookmark Search: Work Language` while leaving the content the same.
- for bookmark-search-language feature i just tacked a language onto an existing setup step, is it preferable to make a new setup/given instead? something like `Given I have bookmarks of works in various languages to search`


sarken:
You guessed right about the position -- sorry for not being clearer! (You're always welcome to comment on issues or send us an email at otw-coders@transformativeworks.org if you have questions you'd like answered before you get to work.)  

I think we can live without help text here. If people are frequently confused by it, we can add some later. 

Making a new `Given I have bookmarks of works in various languages to search` setup step might be a better approach because we really only need two works/bookmarks here: one in the language you're searching for, and one in another language. It would shave a little bit of time off the tests since it wouldn't be creating a bunch of data we don't need.

Is it possible the test is failing because English has a different id on Travis than on your local test database? Something like `language_id: Language.find_by(name: "English").id` would help determine if that was the case.

DreamSea:
Ahh, I forgot that commenting on the issue itself was a thing. That would be a reasonable place~ (no worries on clarity/etc, half the time I need to experiment to make a guess on if I'd actually be able to figure out and fix an issue :sweat_smile: )

Didn't end up digging into what was causing issue, though the language_id seems to have been a likely cause. Went with loading language fixtures and grabbing by name as suggested since it seems like a more preferable way to identify languages anyways. :v:

app/views/bookmarks/_search_form.html.erb
```diff
@@ -26,6 +26,12 @@
       <%= f.select :bookmarkable_type,
                    options_for_select(["", "Work", "Series", "External Work"]) %>
     </dd>
+    <dt>
+      <%= f.label :language_id, ts('Work language') %>
```
sarken: Could you use double quotes here?

features/bookmarks/bookmark_filter.feature
```diff
@@ -0,0 +1,20 @@
+@bookmarks
+Feature: Filter bookmarks
+  In order to search an archive full of bookmarks
+  As a humble user
+  I want to filter some bookmarks
+
+  @new-search
+  Scenario: Filter a user's bookmarks by work language
+    Given I have loaded the fixtures
```
sarken: This might be slightly faster and simpler if you used `I have bookmarks of works in various languages to search` here instead. It should be possible to modify the `I have bookmarks of works in various languages to search` step definition to assign the bookmarks to a specific user.

spec/models/bookmark_search_form_spec.rb
```diff
@@ -91,6 +91,24 @@
...
+  describe "searching" do
+    let(:work1) { create(:posted_work, language_id: 1) }
+    let(:work2) { create(:posted_work, language_id: 2) }
+
+    let!(:bookmark1) { create(:bookmark, bookmarkable: work1) }
+    let!(:bookmark2) { create(:bookmark, bookmarkable: work2) }
+
+    before { run_all_indexing_jobs }
+
+    context "by work language" do
```
sarken: Would it make sense to move this `context` block into the previous `describe "bookmarkable_search_results"` block, since it's also testing `bookmarkable_search_results`? 

The organization in this whole file is a little iffy, though, so it's not super important.

features/bookmarks/bookmark_filter.feature
DreamSea: as far as connection between feature and steps, thoughts between
```gherkin
# secret knowledge that Given step uses specific user
Given I have bookmarks of works in various languages to search
And I am logged in as "recengine"
```
vs
```gherkin
# with user name param "(.*?)"
Given "recengine" has bookmarks of works in various languages to search
And I am logged in as "recengine"
```
or maybe something else entirely?

sarken: The version wit the user name param sounds good! Doing things that way is more transparent and more reusable -- I know I have frequently been tripped up or inconvenienced by some of the collection steps that secretly use `mod1` as the user.

DreamSea:
I think that covers the comments
- prefer doublequote
- load more specific bookmark setup
  - in hindsight (username aside), theres already secret knowledge built in with language and work titles :zzz: 
- lifted describe searching under describe bookmarkable_search_results
  - same level as describe sorting, unless meant to put it under sorting and reuse timetravel works

DreamSea:
made a blind guess at the merge, I'll try to manual confirmation tomorrow that the layout still looks as expected, and will add another comment after once verified. :tea:

sarken:
Thank you!

sarken:
I did a quick check on my dev environment and this looks fine to me, so I'm going to merge it. Thank you again!

DreamSea:
Awesome, thanks for checking / verifying.

Was getting up to speed again trying to remember the local modifications needed to get pass server500s (missing index error - think I was hitting wrong elasticsearch instance locally, I probably needed to redo the steps from https://github.com/otwcode/otwarchive/wiki/Development-Environment#copy-vagrant-specific-config-files). :turtle: 

Y'all good for taking in low/misc priority pull requests again? (or perhaps still in a busy-ish time?)

sarken:
We're still pretty busy -- we're just doing a quick release of small, easy-to-test fixes while our Systems team prepares the servers for our next big change -- but it's fine if you'd like to submit one or two pull requests.

If you still need help with the missing index error, please feel free to email otw-coders@transformativeworks.org! I don't use Vagrant, but I know some other coders do and might be able to help.

DreamSea:
Sounds good. I'll keep a max of two ongoing requests and let merge speed determine the rate :coffee: 

re: index, I'll give the vagrant image(?) and ao3 repo a reinstall since I probably had some local changes that might not play nicely with current elasticsearch settings. If issues still persist after, I'll send an email over.

PR #3350, opened 2018-06-18
AO3-5313 Set Auth Error page title correctly

hatal175:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5313

## Purpose

Fixes Auth Error page title.

## Testing

Open auth error page and check title.

## Credit

Tal.
Please use he.

### DIFF
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -10,6 +10,7 @@ class ErrorsController < ApplicationController
   end
 
   def auth_error
+    @page_title = "Auth Error"
   end
   
 end


app/controllers/errors_controller.rb
```diff
@@ -10,6 +10,7 @@ class ErrorsController < ApplicationController
   end
 
   def auth_error
+    @page_title = "Auth Error"
```
sarken: I think this needs to be `@page_subtitle` so we don't lose the ` | Archive of Our Own` that appears after `Auth Error`.

PR #3351, opened 2018-06-19
AO3-4080 Fix css regex to support shadow box fully

hatal175:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4080

## Purpose

This fixes the css regex so it doesn't fail on development.

## Testing

I've removed the 'pending' state from the spec test for this bug. I will say this though, I've no idea why this works on production. This regex makes me cry tears of blood.

## Credit

Tal.
Please use he.

### DIFF
--- a/lib/css_cleaner.rb
+++ b/lib/css_cleaner.rb
@@ -202,7 +202,7 @@ module CssCleaner
     value_stripped = value.downcase.gsub(/(!important)/, '').strip
 
     # if it's a comma-separated set of valid values it's fine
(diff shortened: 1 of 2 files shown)


sarken:
I think we need Tal's fix, and we should remove `box-shadow` and `opacity` from the list of shorthand properties and put them in the regular list of properties on both staging and production. They're not shorthand properties in either the sense we use it or the commonly-understood sense, but I think shorthand properties _are_ treated differently in our code, which probably explains why it was working.

sarken:
I'm taking off Has Config Changes -- the relevant changes were already made, which is why we need to merge this.

PR #3359, opened 2018-06-26
AO3-5415 Remove the order by on large amounts of kudos

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5415
## Purpose

If a work has more than the maximum number of kudos we show a set of them will be displayed in a random order.Works with less than this should be unaffected.

## Testing

Look at works with more and less than the maximum number of kudos we display and check they have the correct kudos

### DIFF
--- a/app/views/kudos/_kudos.html.erb
+++ b/app/views/kudos/_kudos.html.erb
@@ -1,10 +1,10 @@
 <% # expects local variables kudos, guest_kudos_count, commentable %>
 <div id="kudos">
-  <% if kudos.length > 0 || guest_kudos_count > 0 %>
-    <% cache "#{commentable.cache_key}/kudos-v1", skip_digest: true do %>
+  <% cache "#{commentable.cache_key}/kudos-v1", skip_digest: true do %>
+    <% if kudos.length> 0 || guest_kudos_count > 0 %>
       <p class="kudos">
         <% if kudos.length <= ArchiveConfig.MAX_KUDOS_TO_SHOW %>
-          <%= kudos.map {|kudo| link_to kudo.pseud.byline, kudo.pseud.user}.
+          <%= kudos.order("created_at DESC").map {|kudo| link_to kudo.pseud.byline, kudo.pseud.user}.
               to_sentence.
               html_safe %>
         <% else %>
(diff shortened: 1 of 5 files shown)


app/views/kudos/_kudos.html.erb
```diff
@@ -1,29 +1,29 @@
...
             <% else %>
               <%= ts(" and %{collapsed_count} more users", collapsed_count: collapsed_count) %>
             <% end %>
           <% end %>
           <span class="kudos_expanded hidden">
-            <%= kudos[ArchiveConfig.MAX_KUDOS_TO_SHOW..-1].map {|kudo| link_to kudo.pseud.byline, kudo.pseud.user}.
-                to_sentence(two_words_connector: ts(" and "), last_word_connector: ts(" and ")).
-                html_safe %>
-          </span>
+                <%= kudos[ArchiveConfig.MAX_KUDOS_TO_SHOW..-1].map {|kudo| link_to kudo.pseud.byline, kudo.pseud.user}.
```
sarken: I think the indenting got a little messed up here. Should be:
```
          <span class="kudos_expanded hidden">
            <%= kudos[ArchiveConfig.MAX_KUDOS_TO_SHOW..-1].map {|kudo| link_to kudo.pseud.byline, kudo.pseud.user}.
                to_sentence(two_words_connector: ts(" and "), last_word_connector: ts(" and ")).
                html_safe %>
          </span>
        <% end %>
```

app/views/kudos/_kudos.html.erb
```diff
@@ -33,8 +33,8 @@
         <% end %>
         <%= ts(" left kudos on this work!") %>
         <span class="kudos_expanded hidden">
-          <%= link_to ts("(collapse)"), "#", id: "kudos_collapser" %>
-        </span>
+              <%= link_to ts("(collapse)"), "#", id: "kudos_collapser" %>
```
sarken: And this should be indented like:
```
        <span class="kudos_expanded hidden">
          <%= link_to ts("(collapse)"), "#", id: "kudos_collapser" %>
        </span>
      </p>
```

sarken:
This is a good start and we'll probably build off this, but closing it for now because we're going to need a new issue and a work around for the recent kudos-giver's name issue.

zz9pzza:
I think we may have to do it..

PR #3362, opened 2018-07-03
AO3-5456 Remove tests for Elasticsearch 0.90

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5456

## Purpose

Start only Elasticsearch 6.0 in test builds.

Update test for [AO3-5331](https://otwarchive.atlassian.net/browse/AO3-5331): searching by additional tags for a synonym of a metatag returns only works using the exact tag.

## Testing

Builds should succeed on Travis and Codeship.

On both environments this will appear in the logs:

```
Skipping index creation, cannot connect to Elasticsearch
(The original exception was: #<Errno::ECONNREFUSED: Failed to open TCP connection to 127.0.0.1:9500 (Connection refused - connect(2) for "127.0.0.1" port 9500)>)
```

This is because we don't start ES 0.90, but still try to connect to it:

https://github.com/otwcode/otwarchive/blob/279e1ef0341285b53168b7d3e715ade3d6652943/config/initializers/gem-plugin_config/elasticsearch.rb#L4

The error is harmless, as none of the tests will actually use old ES. We can only remove this bit of config once everything else related to old ES has been removed.


redsummernight:
I missed a thing, this step should be removed:

https://github.com/otwcode/otwarchive/blob/1deb4086f99b72b5fe6b42f26f8ca486ef5b899e/features/step_definitions/search_steps.rb#L59-L62

sarken:
Barring merge conflicts (existing ones and ones we're likely to introduce), this looks good!

PR #3363, opened 2018-07-03
AO3-5461 Revise the crossover check to look for a single fandom whose top-level meta-tags are equal to the set of all top-level meta-tags.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5461

## Purpose

Currently, to check whether a work is a crossover, we compute the total number of top-level meta tags (top-level meaning "doesn't have any meta tags of its own"). This causes a problem when a single fandom has two top-level meta tags -- it counts as being unrelated to all other tags, including its own synonyms.

This PR revises the crossover code to check whether there exists at least one fandom on the work that has all of the top-level meta tags involved (thus meaning that all of the fandoms on the work are related to that one tag).

(Thanks to @redsummernight for figuring out this definition of a crossover!)

## Testing

See the detailed steps on the JIRA issue.

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -1568,24 +1568,28 @@ class Work < ApplicationRecord
   # A work with multiple fandoms which are not related
   # to one another can be considered a crossover
   def crossover
-    # If the filter_taggings table is always correct, we only need one line:
-    # fandoms.count > 1 && filters.by_type('Fandom').first_class.count > 1
-
+    # Short-circuit the check if there's only one fandom tag:
     return false if fandoms.count == 1
 
     # Replace fandoms with their mergers if possible,
     # as synonyms should have no meta tags themselves
-    unrelated_fandoms = fandoms.map { |f| f.merger ? f.merger : f }.uniq
+    all_without_syns = fandoms.map { |f| f.merger || f }.uniq
 
-    # Replace each fandom with the top tags of the meta trees it belongs to
-    loop do
-      n = unrelated_fandoms.map { |f| f.meta_tags.any? ? f.meta_tags : f }.flatten.uniq
-      break if n == unrelated_fandoms
-      unrelated_fandoms = n
+    # For each fandom, find the set of top-level meta tags (i.e. meta-tags that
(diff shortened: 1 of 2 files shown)


app/models/work.rb
```diff
@@ -1568,24 +1568,28 @@ def creators
...
     end
 
-    # These fandoms have no meta tags, and they cannot be related
-    unrelated_fandoms.count > 1
+    # Find the biggest group of top-level meta tags.
+    biggest_group_size = top_meta_groups.map(&:size).max
+
+    # If the biggest group is the same size as the total number in all groups,
+    # that means that all top-level meta-tags in all groups also occur in the
+    # biggest group, so we don't don't have any fandoms that are unrelated to
```
redsummernight: There's an extra "don't" here.

spec/models/work_spec.rb
```diff
@@ -161,6 +176,54 @@
...
+    context "when one tagged fandom has two unrelated meta tags" do
+      let(:meta1) { create(:canonical_fandom) }
+      let(:meta2) { create(:canonical_fandom) }
+      let(:fandom) { create(:canonical_fandom) }
+
+      before do
+        fandom.update_attribute(:meta_tag_string, "#{meta1.name},#{meta2.name}")
+      end
+
+      it "is not a crossover with the fandom's synonym" do
```
redsummernight: This is similar to https://github.com/otwcode/otwarchive/blob/0a980356c82bad64cc248e7b5501106163a27864/spec/models/work_spec.rb#L112-L117

tickinginstant: Should I delete the existing one? I figured it would be good to keep around both, to make it easier to pinpoint the source of the error if this code ever needs to be modified again. (e.g. Is the issue synonyms alone, or synonyms of fandoms with two top-level meta tags?)

redsummernight: Good point, the tests cover different cases. Let's keep both.

app/models/work.rb
tickinginstant: Oops! Thanks for pointing it out.

PR #3367, opened 2018-07-05
AO3-5463 Show anonymous & unrevealed works in user's Works in Collections

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5463

## Purpose

This is missing in new search.

Unrevealed works should appear to all, while anonymous works should only appear if the logged in user is the author.

## Testing

See JIRA issue.

### DIFF
--- a/app/models/search/work_query.rb
+++ b/app/models/search/work_query.rb
@@ -234,7 +234,7 @@ class WorkQuery < Query
     [:title, :creators].each do |field|
       search_text << split_query_text_words(field, options[field])
     end
-    if self.options[:collection_ids].blank? && options[:collected]
+    if options[:collection_ids].blank? && collected?
       search_text << " collection_ids:*"
     end
     escape_slashes(search_text.strip)
--- a/app/models/search/work_query.rb
+++ b/app/models/search/work_query.rb
@@ -151,7 +151,7 @@ class WorkQuery < Query
   end
 
   def user_filter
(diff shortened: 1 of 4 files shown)


app/models/search/work_query.rb
```diff
@@ -275,20 +275,29 @@ def facet_tags?
...
     options[:collected]
   end
 
   def include_restricted?
     User.current_user.present? || options[:show_restricted]
   end
 
+  # Include unrevealed works only if we're on a collection page
+  # OR the collected works page of a user
   def include_unrevealed?
```
redsummernight: From old search, for comparison:

https://github.com/otwcode/otwarchive/blob/0a980356c82bad64cc248e7b5501106163a27864/app/models/work_search.rb#L312-L322

app/models/search/work_query.rb
```diff
@@ -275,20 +275,29 @@ def facet_tags?
...
-    options[:collection_ids].present?
+    options[:collection_ids].present? || collected?
   end
 
+  # Include anonymous works unless we're on a user or pseud page
+  # OR unless the user is viewing their own collected works
```
sarken: I think this part of the comment is backwards -- we're including anonymous works _if_ the user is viewing their own collected works page, not _unless_. It'd be "OR unless we're viewing someone else's collected works"

PR #3368, opened 2018-07-05
AO3-4842 Stop stripping carriage returns from Support and Abuse request descriptions

scottsds:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4842

## Purpose

Formatting was previously being stripped before being sent to the Bugtracker. This was because of changed that occured in https://otwarchive.atlassian.net/browse/AO3-4216 . This was no longer desired behavior, so I have removed the format stripping from the 'description' field only. 

## Testing

Submit an Abuse and Support request and make sure that formatting (italics, bold, newline, etc.) correctly show up on the Bug/Abuse trackers. 

## References

Original issue that called for removing formatting: https://otwarchive.atlassian.net/browse/AO3-4216

### DIFF
--- a/app/models/feedback_reporters/feedback_reporter.rb
+++ b/app/models/feedback_reporters/feedback_reporter.rb
@@ -20,10 +20,6 @@ class FeedbackReporter
     strip_html_breaks_simple(@title)
   end
 
-  def description
-    @description
-  end
-
   def send_report!
     HTTParty.post("#{ArchiveConfig.NEW_BUGS_SITE}#{project_path}",
                   body: "&xml=#{xml.to_str}")
--- a/app/models/feedback_reporters/feedback_reporter.rb
+++ b/app/models/feedback_reporters/feedback_reporter.rb
@@ -21,16 +21,12 @@ class FeedbackReporter
   end
 
   def description
-    strip_html_breaks_simple(@description)
+    @description
   end
 
   def send_report!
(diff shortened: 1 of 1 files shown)


app/models/feedback_reporters/feedback_reporter.rb
```diff
@@ -21,16 +21,12 @@ def title
   end
 
   def description
```
sarken: Did you try removing this completely? I think it was only necessary because we were essentially saying "instead of using the description, use a modified version of the description that has the HTML breaks stripped."

app/models/feedback_reporters/feedback_reporter.rb
```diff
@@ -21,16 +21,12 @@ def title
...
-    strip_html_breaks_simple(@description)
+    @description
   end
 
   def send_report!
-    # We're sending the XML data via a URL to our Support ticket service. The
-    # URL needs to be Percent-encoded so that everything shows up correctly on
-    # the other end. (https://en.wikipedia.org/wiki/Percent-encoding)
-    encoded_xml = CGI.escape(xml.to_str)
```
redsummernight: Why is percent encoding no longer necessary?

scottsds:
@sarken I meant to check removing the description method today, but then got sidetracked. Just checked it, you were correct, there was no need for it to be there. The @description variable was accessible as normal without it.

PR #3218, opened 2018-01-05
AO3-2839 Migration to Devise

WendyBeth:
# Pull Request Checklist

* [x] Have you read [How to write the perfect pull request](https://github.com/blog/1943-how-to-write-the-perfect-pull-request)?
* [x] Have you read through the [contributor guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [x] Have you added tests for any changed functionality?
* [x] Have you added the JIRA issue number as the *first* thing in your pull request title (eg: `AO3-1234 Fix thing`)
* [x] Have you updated the JIRA issue with the information below?

## Issue

https://otwarchive.atlassian.net/browse/AO3-2839

## Purpose

What does this PR do?

This PR migrates all authentication to Devise

Additional notes: I don't fully understand what's going on with the Elasticsearch commits between October 20th and the 25th - I'm assuming this is strangeness from all the trouble surrounding the spam problems. I've looked at the diff and the only changes that are present there are the changes that are related to Devise and the changes that should be made with this PR.

## Testing

How can the Archive's QA team verify that this is working as you intended? (If you have access, please copy this into the JIRA ticket for them!)

Make sure that you can sign up as a regular user, sign in as both a regular user and an admin, and sign out.

## References

Are there any other relevant issues / PRs / mailing lists discussions related to this?

## Credit

*What name do you want us to use to credit your work in the [Archive of Our Own's Release Notes](http://archiveofourown.org/admin_posts?tag=1)?*

Wendy Randquist - Littlelines

*What pronouns do you prefer (she/her, he/him, zie/hir etc)?*

she/her

(if you do not fill this in, we will use your GitHub account name and will use "they" to avoid making assumptions about your gender)

PR #3378, opened 2018-07-10
AO3-2839 Migration to Devise

elzj:
## Issue

## Purpose

Updates previous pull request (#3218) with fixes for merge conflicts.

## Testing

See issue.


app/controllers/users/sessions_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+
+  layout "session"
+  before_action :admin_logout_required
+  skip_before_action :store_location
+
+  def create
+    if warden.authenticate(auth_options)
+      super do |resource|
+        if resource.remember_me
+          flash[:notice] += ts(" <strong>You'll stay logged in for %{number} weeks even if you close your browser, so make sure to log out if you're using a public or shared computer.</strong>", number: ArchiveConfig.DEFAULT_SESSION_LENGTH_IN_WEEKS).html_safe
```
sarken: I think this is where we noticed some escaped HTML when testing on dev.

app/views/layouts/session.html.erb
```diff
@@ -38,7 +38,7 @@
       <div id="inner" class="wrapper">
         <!-- BEGIN main -->
         <div id="main" class="<%= controller.controller_name + '-' + controller.action_name %> system session region" role="main">
-          <%= flash_div :error, :caution, :notice %>
+          <%= flash_div :error, :caution, :notice, :alert %>
```
sarken: I think we need to add `.alert` to all the places `.error` appears in 22-system-messages.css, except `span.error`. Otherwise it'll be unstyled, as we saw with the "You need to sign in or sign up before continuing" message on dev.

Gemfile.lock
```diff
@@ -203,9 +198,6 @@ GEM
       multipart-post (>= 1.2, < 3)
     fastimage (2.1.0)
     ffi (1.9.18)
```
redsummernight: Not a complete review, I just found that ffi and scrypt are only dependencies of authlogic and should also be removed from `Gemfile.lock`. A `bundle install` should do it.

redsummernight:
test/fixtures should be updated to match the new migrations, and we should also leave out changes to the schema dumps to be handled in [AO3-5194](https://otwarchive.atlassian.net/browse/AO3-5194).

app/views/users/sessions/_passwd_small.html.erb
```diff
@@ -1,22 +1,21 @@
...
-  </p>    
+  </p>
 <% end %>
 
 <ul class="footnote actions">
   <li><%= link_to ts("Forgot password?"), new_password_path %></li>
   <% if @admin_settings.account_creation_enabled? && !@admin_settings.creation_requires_invite? %>
     <li>
-      <%= link_to ts("Create an Account"), new_user_path %>
+      <%= link_to ts("Create an Account"), signup_path%>
```
sarken: There's a space missing here.

Gemfile.lock
redsummernight: Once we removed ffi and pulled from master, we can stop ignoring the vulnerability CVE-2018-1000201 (#3425).

features/step_definitions/user_steps.rb
```diff
@@ -18,6 +18,51 @@
...
     user.pseuds.first.add_to_autocomplete
+    step %{confirmation emails have been delivered}
+  end
+end
+
+Given /the following users exist with BCrypt encrypted passwords/ do |table|
+  table.hashes.each do |hash|
+    user = FactoryGirl.create(:user, hash)
+    user.activate
+    user.pseuds.first.add_to_autocomplete
```
sarken: Hmmm... is this necessary? I'm not seeing autocomplete-related tests that use "the following users exist with BCrypt encrypted passwords." Ditto "the following users exist with SHA-512 encrypted passwords," actually.

features/step_definitions/web_steps.rb
```diff
@@ -183,6 +183,11 @@ def with_scope(locator)
   end
 end
 
+Then(/^I should see "([^"]*)" or "([^"]*)"$/) do |text1, text2|
```
sarken: I might be missing it -- I don't entirely trust Ctrl+F on pull requests -- but I don't think this is used.

features/users/authenticate_users.feature
```diff
@@ -25,48 +25,119 @@ Feature: User Authentication
...
+
+  Scenario: With expired password token
+    Given I have no users
+      And the following activated user exists
+        | login | password |
+        | sam   | password |
+      And all emails have been delivered
+    When I am on the login page
+      And I follow "Reset password"
+      And I fill in "reset_password_for" with "sam"
```
sarken: If we could use the label text ("Email address or user name") instead, that would be great. (If it doesn't work, we might have an HTML booboo.)

features/users/authenticate_users.feature
```diff
@@ -25,48 +25,119 @@ Feature: User Authentication
...
+
+  Scenario: Forgot password, logging in with email address
+    Given I have no users
+      And the following activated user exists
+        | login | email           | password |
+        | sam   | sam@example.com | password |
+      And all emails have been delivered
+    When I am on the login page
+      And I follow "Reset password"
+      And I fill in "reset_password_for" with "sam@example.com"
```
sarken: Could we use the label text ("Email address or user name") here?

features/users/authenticate_users.feature
```diff
@@ -130,6 +201,40 @@ Feature: User Authentication
...
+      And I fill in "User name" with "themaduser"
+      And I fill in "Password" with "password1"
+      And I press "Log In"
+    Then I should see "Successfully logged in."
+      And I should see "Hi, TheMadUser!"
+
+  Scenario: Login with email or login
+    Given the following activated user exists
+      | login      | email                          | password |
+      | TheMadUser | themaduser@archiveofourown.org | password |
```
sarken: We should use `@example.com` email addresses when adding new tests (although this one's not too bad since we own the domain and know no one is using it).

app/controllers/application_controller.rb
```diff
@@ -202,7 +190,14 @@ def redirect_back_or_default(default = root_path)
...
 
   def after_sign_in_path_for(resource)
-    admin_users_path
+    if resource.is_a?(Admin)
+      admin_users_path
+    else
+      back = session[:return_to]
+      session.delete(:return_to)
+
+      back || root_path
```
tickinginstant: I may be mistaken, but doesn't the existing authlogic code redirect to the user's dashboard on successful login (at least, if it can't find a value in `session[:return_to]`)? Is redirecting to the homepage here a deliberate change?

features/step_definitions/generic_steps.rb
```diff
@@ -2,6 +2,10 @@
   unselect(item, from: selector)
 end
 
+When /^(\d) minutes have passed$/ do |number|
```
redsummernight: We have cucumber-timecop, so we can use the step `When it is currently x minutes from now`, without defining this one.

features/step_definitions/user_steps.rb
redsummernight: It's not but I'm kinda on the fence... is it worth being consistent? In the unlikely event that we (somehow) use this step to set up a test that involves autocomplete we're not left head-scratching why autocomplete doesn't work. That happened to me on https://github.com/otwcode/otwarchive/pull/2966/files#diff-904a4107d89f81a525345d37804626a9.

features/step_definitions/web_steps.rb
redsummernight: It's definitely unused: https://github.com/elzj/otwarchive/commit/75b037d6b4222ee5d30e456d4ecb6a1aceabec6a#diff-0fa40971eef4a01211c40236d846f289. It was added to deal with the PDT/PST time zone name issue in that one challenge test.

.ruby-version
```diff
@@ -0,0 +1 @@
+2.3.4
```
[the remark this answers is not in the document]
sarken: I'm thinking it might not be something we want committed? It appears to have been added in a commit that was just merging with master: c0811eee68216a5251fed8bac73a45472244af05

When I pull the branch, I get the message:
```
error: The following untracked working tree files would be overwritten by merge:
	.ruby-version
Please move or remove them before you merge.
```

app/views/layouts/session.html.erb
sarken: I'm pretty sure we don't need in in `div.error` either so I'm not going to add it at the moment. If we run into problems, 🐟 slap me later.

features/step_definitions/user_steps.rb
sarken: Laziness wins: we'll keep it.

sarken:
Updated: #3443

test/fixtures/users.yml
```diff
@@ -2,11 +2,11 @@
 user_00025: 
   salt: 2aa64c10d3ea67188845a59d3d9d411d52301838
   created_at: 2009-12-12 21:07:52 Z
-  activated_at: 2009-12-12 21:08:07 Z
+  confirmed_at: 2009-12-12 21:08:07 Z
   recently_reset: false
   crypted_password: acfecdcd9a91ee01aa7da07d29d3ffb6036714fa
```
tickinginstant: I believe this (and all similar lines in this file) needs to be changed to `encrypted_password` instead of `crypted_password`.

test/fixtures/users.yml
```diff
@@ -2,11 +2,11 @@
 user_00025: 
   salt: 2aa64c10d3ea67188845a59d3d9d411d52301838
```
tickinginstant: I believe this (and all similar lines in this file) needs to be changed to `password_salt` instead of `salt`.

spec/controllers/users/registrations_controller_spec.rb
```diff
@@ -1,6 +1,6 @@
 require 'spec_helper'
 
-describe UsersController do
+describe Users::RegistrationsController do
```
tickinginstant: Since I wrote these tests, I figured I should make sure that they're compatible with this PR. When I ran them, the error message said that this block of tests needs to have this code inserted at (or near) the beginning:
```
  before do
    @request.env["devise.mapping"] = Devise.mappings[:user]
  end
```
In addition, all of the parameters use `user_registration` instead of `user` now.

app/controllers/users/sessions_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+  before_action :admin_logout_required
+  skip_before_action :store_location
+
+  def create
+    if warden.authenticate(auth_options)
+      super do |resource|
+        if resource.remember_me
+          message = ts(" <strong>You'll stay logged in for %{number} weeks even if you close your browser, so make sure to log out if you're using a public or shared computer.</strong>", number: ArchiveConfig.DEFAULT_SESSION_LENGTH_IN_WEEKS)
+        end
+        flash[:notice] += message
```
sarken:
```(::) failed steps (::)
no implicit conversion of nil into String (TypeError)
./app/controllers/users/sessions_controller.rb:13:in `+'
./app/controllers/users/sessions_controller.rb:13:in `block in create'
./app/controllers/users/sessions_controller.rb:9:in `create'
./features/step_definitions/web_steps.rb:50:in `block (2 levels) in <top (required)>'
./features/step_definitions/web_steps.rb:14:in `with_scope'
./features/step_definitions/web_steps.rb:49:in `/^(?:|I )press "([^"]*)"(?: within "([^"]*)")?$/'
features/tags_and_wrangling/tag_comment.feature:138:in `And I press "Log In"'
Failing Scenarios:
cucumber features/tags_and_wrangling/tag_comment.feature:93 # Scenario: Issue 2185: email notifications for tag commenting; TO DO: replies to comments
```

PR #3379, opened 2018-07-11
AO3-5477 AO3-5489 AO3-5490 Remove old search, fix best match sorting and creator/title boosting in work search

elzj:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5477
https://otwarchive.atlassian.net/browse/AO3-5489
https://otwarchive.atlassian.net/browse/AO3-5490

## Purpose

- Now that we've migrated to the new search system, this cleans up all the remnants of the old one, from the Tire gem to the model code to the old search classes and the legacy CSS. Branches off from #3362.
- When searching works or bookmarks, the "Best Match" option is the default and should return results sorted by relevance.
- Elasticsearch allows us to add weight to certain fields when searching, so we want to emphasize titles and creator names so that searching for "Jane" would bring up a work titled "Jane Eyre" first, then works posted by Jane, and then every Spider-Man/Thor/Jane the Virgin/Jane Eyre/etc. fandom work.

## Testing

See issues.


app/models/work_search_cleanser.rb
```diff
@@ -0,0 +1,129 @@
...
+    set_language!
+    clean_up_angle_brackets
+
+    # Clean up blank options from forms
+    options.delete_if { |key, value| value.blank? }
+  end
+
+  private
+
+  def set_parent_fields!
```
redsummernight: We have similar code in WorkQuery:

https://github.com/otwcode/otwarchive/blob/f91b3606fd270add0933d16049ef49ebe244551c/app/models/search/work_query.rb#L42-L57

The WorkQuery version adds only the user ID if the parent is a user, instead of all the pseud IDs of that user, which should make loading orphan_account's works faster ([AO3-4871](https://otwarchive.atlassian.net/browse/AO3-4871)).

Can we remove `set_parent_fields!`?

elzj: Ideally, we'd get rid of the whole class, but I thought we might want to take things step by step. I can take a look and see if that causes any problems.

app/controllers/bookmarks_controller.rb
```diff
@@ -103,25 +97,18 @@ def index
...
-          if use_new_search?
-            @search = BookmarkSearchForm.new(options.merge(faceted: true, parent: @owner))
-          else
-            @search = BookmarkSearch.new(options.merge(faceted: true, bookmarks_parent: @owner))
-          end
+          @search = BookmarkSearchForm.new(options.merge(faceted: true, parent: @owner))
 
-          if use_new_search? && @user.blank?
+          if @user.blank?
             # We're using the new search, but it's not a particular user's
```
redsummernight: There's a reference to new search that should be removed.

app/models/search_range.rb
```diff
@@ -1,49 +1,46 @@
-# ES UPGRADE TRANSITION #
-# Remove class
-class Search < ApplicationRecord
```
redsummernight: Should we have a migration to drop the "searches" table?

app/models/search/work_query.rb
```diff
@@ -252,6 +261,11 @@ def sort
     sort_hash
   end
 
+  # When searching outside of filters, use relevance instead of date
+  def default_sort
```
redsummernight: We should mention this behaviour change in the issue so we can check for it.

redsummernight: Also I think `facet_tags?` is for the filters so the conditional should be the other way around.

elzj: Nope, it's "if we're using the filters, default to date, otherwise, default to score." (That was the behavior of the old system.)

app/models/search_range.rb
elzj: We can make a separate issue for that, I'd say

app/models/bookmark_search.rb
```diff
@@ -1,379 +0,0 @@
...
-  def self.count_for_pseuds(pseuds)
-    terms = [
-      { term: { hidden_by_admin: 'F' } },
-      { terms: { pseud_id: pseuds.map(&:id) } }
-    ]
-    unless pseuds.map(&:user).uniq == [User.current_user]
-      terms << { term: { private: 'F' } }
-    end
-    query = { bool: { must: terms } }
-    response = ElasticsearchSimpleClient.perform_count(Bookmark.index_name, 'bookmark', query)
```
redsummernight: Should we remove ElasticsearchSimpleClient?

tickinginstant:
It's not marked with the `# ES UPGRADE TRANSITION #` comment, but I think `app/models/search_result.rb` can also be removed.

PR #3380, opened 2018-07-11
AO3-5167: Improve performance of ownership checks

elzj:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5167

## Purpose

Switches is_author_of? method to compare based on ids rather than objects, which should save on some db calls and memory. Also preloads the set of current works/bookmarks owned by the current user on index pages so that we don't have to run separate checks over and over.

Replaces https://github.com/otwcode/otwarchive/pull/3054 which was approved but then became hopelessly conflicted.

## Testing

Nothing should change: appropriate footers for work and bookmark blurbs should appear if they're yours, and anonymity should be respected.

### DIFF
--- a/app/controllers/bookmarks_controller.rb
+++ b/app/controllers/bookmarks_controller.rb
@@ -59,6 +59,7 @@ class BookmarksController < ApplicationController
       end
       @bookmarks = @search.search_results
       flash_max_search_results_notice(@bookmarks)
+      set_own_bookmarks
       render 'search_results'
     end
   end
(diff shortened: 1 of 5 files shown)


sarken:
Test failure was `features/other_a/invite_request.feature:75:in 'Then 1 email should be delivered to test@archiveofourown.org'` but I've restarted it and hopefully it will pass -- it really shouldn't be affected by this...

sarken:
Whoops, I didn't remember that change at all. Thanks for the catch! This really should be good to go then :P

PR #3381, opened 2018-07-11
AO3-5479 Smear the tag counts a bit

zz9pzza:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5479

## Purpose

Spread out the load for large tags.

## Testing

Really little. I would see if the count changes after 80 minutes on a large tag.


app/models/tag.rb
```diff
@@ -71,7 +71,7 @@ def self.taggings_count_expiry(count)
     # would be TAGGINGS_COUNT_MIN_TIME ( defaults to 3 minutes ) and the maximum amount of time would be
     # TAGGINGS_COUNT_MAX_TIME ( defaulting to an hour ).
     expiry_time = count / (ArchiveConfig.TAGGINGS_COUNT_CACHE_DIVISOR || 1500)
-    [[expiry_time, (ArchiveConfig.TAGGINGS_COUNT_MIN_TIME || 3)].max, (ArchiveConfig.TAGGINGS_COUNT_MAX_TIME || 60)].min
+    [[expiry_time, (ArchiveConfig.TAGGINGS_COUNT_MIN_TIME || 3)].max, (ArchiveConfig.TAGGINGS_COUNT_MAX_TIME || 50) + count % 20 ].min
```
tickinginstant: It's worth noting that all tags with an expiry time > 40 minutes will have their cached count recalculated (and the expiration time reset) by the add_counts_to_queue task:
https://github.com/otwcode/otwarchive/blob/000f559dac47f200b36b7ece415a27a6f2d65210/app/models/scheduled_tag_job.rb#L4-L7
That task is supposed to be run every half hour, before the expiration time has a chance to occur:
https://github.com/otwcode/otwarchive/blob/000f559dac47f200b36b7ece415a27a6f2d65210/config/resque_schedule.yml#L15-L20
Since this PR is only affecting the expiry time for tags with an expiry time of 50+ minutes, I'm not sure if this will affect when those tags have their taggings_count recalculated.

PR #3383, opened 2018-07-13
AO3-4842 Add <br /> between paragraphs for Support and PAC tickets

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4842

## Purpose

Adds a `<br />` between the paragraphs in Support and PAC tickets because the ticket tracker's paragraph styling doesn't have margins, so the paragraphs looked more like:

Paragraph 1
Paragraph 2

than like

Paragraph 1

Paragraph 2

even though they were correctly HTMLified as `<p>Paragraph 1</p><p>Paragraph 2</p>`

They are now rather horrifically done as  `<p>Paragraph 1</p><br /><p>Paragraph 2</p>` because a single empty `<p></p>` did not have sufficient line height to resemble a blank line, and `&nbsp;` or `&#160;` caused the tickets not to make it to the tracker, while an actual typed out space would get stripped and go right back to the line height not resembling a blank line.

## Testing

Refer to Jira

### DIFF
--- a/app/models/feedback.rb
+++ b/app/models/feedback.rb
@@ -63,7 +63,7 @@ class Feedback < ApplicationRecord
   end
 
   def send_report
-    # return unless %w(staging production).include?(Rails.env)
+    return unless %w(staging production).include?(Rails.env)
     reporter = SupportReporter.new(
       title: summary,
       description: comment,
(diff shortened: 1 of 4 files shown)


scottsds:
Looks good! 👍 Thank you for doing this!

PR #3385, opened 2018-07-14
AO3-5465 Task for reindexing mistaken crossovers from AO3-5461.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5465

## Purpose

Thanks to AO3-5461, some works in fandoms with multiple first-class metatags are incorrectly classified as crossovers. This adds a task that reindexes all filtered works in those fandoms, so that their crossover field will be recalculated.

## Testing

The bug report contains steps for testing.

### DIFF
--- a/lib/tasks/after_tasks.rake
+++ b/lib/tasks/after_tasks.rake
@@ -581,6 +581,28 @@ namespace :After do
       end
     end
   end
+
+  desc "Fix crossover status for fandoms with two first-class meta-tags."
+  task(crossover_reindex_two_first_class_meta_tags: :environment) do
+    # Find all fandoms with two first-class meta tags:
+    Fandom.joins(:meta_taggings).
+           joins("LEFT JOIN meta_taggings as first_class_check " \
+                 "ON first_class_check.sub_tag_id = meta_taggings.meta_tag_id").
+           where(first_class_check: { id: nil }).
+           group("tags.id").
+           having("COUNT(meta_taggings.id) > 1").
+           find_each do |fandom|
+      print "Reindexing all filtered works for #{fandom.name}" and STDOUT.flush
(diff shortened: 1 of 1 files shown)


lib/tasks/after_tasks.rake
```diff
@@ -581,6 +581,28 @@ namespace :After do
...
+    Fandom.joins(:meta_taggings).
+      joins("LEFT JOIN meta_taggings as first_class_check " \
+            "ON first_class_check.sub_tag_id = meta_taggings.meta_tag_id").
+      where(first_class_check: { id: nil }).
+      group("tags.id").
+      having("COUNT(meta_taggings.id) > 1").
+      find_each do |fandom|
+      print("Reindexing all filtered works for #{fandom.name}") && STDOUT.flush
+      # For each fandom, reindex all of its filtered works:
+      fandom.filter_taggings.
```
redsummernight: I'm not sure if we want to rely on the filter_taggings table, which can be incorrect ([AO3-5430](https://otwarchive.atlassian.net/browse/AO3-5430)). There are works that should have certain filters but don't.

tickinginstant: This will work even when the inherited filter taggings are incorrect, which is the issue with the example in the link. (In fact, I should maybe use `fandom.direct_filter_taggings` here to avoid duplication, if I don't modify this to avoid the filter_taggings table altogether.)

But I'm guessing there are probably also a few works missing direct filter taggings, too, as a result of #2868 and similar -- is there a lot of overlap between those works and the ones affected by this issue? It's not too hard to add more joins, but I'm not sure what the effect on the runtime would be.

tickinginstant:
Closing this until I have the chance to update it to work with #3387.

PR #3395, opened 2018-07-22
AO3-5493 Add wrangling data to tag index

elzj:
## Issue

## Purpose

Adds indexing for tag wrangling associations in the tag search index and then uses that as the basis for finding and counting unwrangled tags on wrangling pages. Index should update when tags gain/lose parents/children and when a tag is used early in its lifespan so that we can gather contextual data. Also adds sorting and additional filters to tag queries.

Should also fix https://otwarchive.atlassian.net/browse/AO3-2270

## Testing

Tag wrangling pages should have the correct data, and the data should sort correctly. There's a little bit removed from the wrangle page that didn't seem to be working on production, but it would be good to get a tag wrangler to verify that on staging.

And pages hopefully shouldn't take ten minutes to load anymore.


app/models/tagging.rb
```diff
@@ -43,4 +44,11 @@ def self.find_by_tag(taggable, tag)
   def update_taggings_count
     tagger.update_tag_cache unless tagger.blank? || tagger.destroyed?
   end
+
+  def update_search
+    reindex_boundary = 100
```
sarken: This part has me a little confused. Will things just not get reindexed if there are more than 100 taggings? Could that cause any sort of issue?

app/views/tags/edit.html.erb
```diff
@@ -95,16 +95,13 @@
           <% if tag_type == 'Fandom' && !@suggested_fandoms.blank? %>
             <dt><%= ts("Suggested Fandoms") %>:</dt>
             <dd>
-            <% @suggested_fandoms[0..19].in_groups(2, false) do |fandom_list| %>
+            <% @suggested_fandoms.in_groups(2, false) do |fandom_list| %>
```
sarken: Not particularly important since it's exisiting code, but I'm curious: what's the benefit of using `in_groups` here?

app/models/search/tag_query.rb
```diff
@@ -25,6 +34,28 @@ def per_page
...
+    when "created_at"
+      column = "created_at"
+      direction ||= "desc"
+    else
+      column = "sortable_name"
+      direction ||= "asc"
+    end
+    sort_hash = { column => { order: direction } }
+
+    if column == 'created_at'
```
sarken: Could you use double quotes here for consistency?

app/views/tags/edit.html.erb
elzj: Honestly not sure there is any, but I didn't want to get into changing the view.

app/models/tagging.rb
sarken: That makes sense to me, but I think we might want to run it by Tag Wrangling, since they were able to see all suggested fandoms on unwrangled tags before. If they're okay with the limitation, we can add a code comment that is basically what you said here.

app/models/search/tag_indexer.rb
```diff
@@ -53,7 +53,28 @@ def document(object)
     object.as_json(
       root: false,
       only: [:id, :name, :merger_id, :canonical, :created_at]
-    ).merge(tag_type: object.type)
+    ).merge(
+      tag_type: object.type,
+      uses: object.taggings_count_cache,
+      sortable_name: object.name.downcase
```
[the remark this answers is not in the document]
elzj: I think that was just a typing failure, yeah. Will fix.

sarken:
Unfortunately, this has conflicts, so I'm switching the label to Needs Coder Action.

sarken:
I was in the neighborhood so I took care of the aforementioned merge conflicts.

app/models/tag.rb
```diff
@@ -1214,6 +1211,53 @@ def indirect_bookmarks(rec=false)
...
   end
 
+  #################################
+  ## SEARCH #######################
+  #################################
+
+  def unwrangled_query(tag_type, options = {})
+    TagQuery.new(options.merge(
+      type: tag_type,
+      unwrangleable: false,
```
tickinginstant: I think this part of the query might be broken. The `TagIndexer` uses `as_json(only: [])`, but doesn't include `unwrangleable` as one of the fields. And the `TagQuery` function `unwrangleable_filter` will only add the term query if `options[:unwrangleable].present?`, but `false.present?` is false.

app/models/search/tag_indexer.rb
```diff
@@ -52,8 +52,28 @@ def self.settings
...
+  # Index parent data for tag wrangling searches
+  def parent_data(tag)
+    data = {}
+    %w(Media Fandom Character).each do |parent_type|
+      if tag.parent_types.include?(parent_type)
+        key = "#{parent_type.downcase}_ids"
+        ids = tag.parents.by_type(parent_type).pluck(:id)
+        # add a dummy value so we can find the unwrangled tags more easily
+        # since you can't search for an empty array
+        data[key] = ids.empty? ? [0] : ids
```
tickinginstant: FYI, it looks like there is a query that lets you search for an empty array:
https://www.elastic.co/guide/en/elasticsearch/reference/6.4/query-dsl-exists-query.html

sarken:
I fixed the broken bit around unwrangleable tags but didn't make other suggested changes.

app/models/search/tag_indexer.rb
sarken: Kibana tells me this field is a number, so I _think_ we'd have to set it to `0` anyway: https://www.elastic.co/guide/en/elasticsearch/reference/current/number.html

It seems like this method is ultimately a less verbose way of doing the same thing

tickinginstant: I meant the exists query, not the null_value mapping. I suspect something like this should work, though I haven't finished [running all of the tests](https://travis-ci.org/tickinginstant/otwarchive/builds/459989946):

https://github.com/tickinginstant/otwarchive/commit/499256ade703fce97aff6102319074111349802a

app/models/tag.rb
```diff
@@ -1241,6 +1238,48 @@ def indirect_bookmarks(rec=false)
...
+    ))
+  end
+
+  def unwrangled_tags(tag_type, options = {})
+    unwrangled_query(tag_type, options).search_results
+  end
+
+  def unwrangled_tag_count(tag_type)
+    key = "unwrangled_#{tag_type}_#{self.id}_#{self.updated_at}"
+    Rails.cache.fetch(key, expires_in: 4.hours) do
```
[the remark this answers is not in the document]
zz9pzza: Archiveconfig with default is my preference :)
( I know we go against the having a default value )

app/models/search/tag_indexer.rb
redsummernight: I prefer using 0 for empty (basically ad-hoc null_value mapping), since it means not having to flip the logic in my head for an exclusion filter. If the code looks kinda mysterious, we can do something like:

```rb
TagQuery.new(options.merge(
  no_fandom: true,
))
```
Then in the query class:
```rb
def no_fandom_filter
  terms_filter(:fandom_ids, [0]) if options[:no_fandom]
end
```
then add `no_fandom_filter` to `filters` instead of `exclusion_filters`. Add some comments in (plus a test in tag_query_spec.rb) and it should be pretty readable.

sarken: I'd also call it `unwrangled_filter` rather than no fandom, to avoid any confusion between unwrangled tags and no fandom tags (which actually have a fandom called "No Fandom").

tickinginstant: I disagree about using 0 for empty. The exists query accomplishes the same thing without having to modify the data (which means that there is less danger of running into future issues where the 0 is accidentally treated as real data -- e.g. in aggregations, or classes like PseudDecorator -- and also means that related techniques can be used in other query classes without having to reindex everything to add in the null value).

If it absolutely has to be in the list of positive filters, it can be wrapped in another bool query as a `must_not`. But I think it's cleaner to just have it be an exclusion filter -- probably with a better name, since that seems to be an issue.

app/models/tag.rb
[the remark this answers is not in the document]
sarken: Just to answer my own question, this isn't something we usually configure.

PR #3396, opened 2018-07-24
AO3-5494 Rescue search errors and respond with error message

elzj:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5494

## Purpose

We try to rescue syntax errors in searches, but when we fail, users should see an error message and not a 500 page. 

## Testing

Currently, exclamation points and other punctuation characters (like tildes) cause errors in main work and bookmark search, so we can try those and verify that the new message comes up.


sarken:
This has gone grey, hence the switch from Ready to Merge to Needs Coder Action.

sarken:
:+1:

PR #3397, opened 2018-07-25
AO3-4037: Automatically retry failing emails

elzj:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4037

## Purpose

Rescues and retries resque email failures that occur when redis is ahead of mysql, which should hopefully keep those out of the failed jobs list, making it easier to track. Also cleans up a bit of duplication in the mailers.

## Testing

It could be hard to simulate this on staging, so we may need a database admin to create some fake failures.

### DIFF
--- a/config/initializers/gem-plugin_config/resque_mailer.rb
+++ b/config/initializers/gem-plugin_config/resque_mailer.rb
@@ -1,1 +1,10 @@
 Resque::Mailer.excluded_environments = [:test, :cucumber]
+Resque::Mailer.error_handler = lambda { |mailer, message, error, action, args|
+  if error.is_a?(Resque::TermException)
+    Resque.enqueue(mailer, action, *args)
+  elsif error.is_a?(ActiveRecord::RecordNotFound) && !args.include?(:retried)
+    Resque.enqueue_in(5.minutes, mailer, action, :retried, *args)
+  else
+    raise error
+  end
+}
--- a/app/mailers/admin_mailer.rb
+++ b/app/mailers/admin_mailer.rb
@@ -1,10 +1,4 @@
-class AdminMailer < ActionMailer::Base
-  include Resque::Mailer # see README in this directory
-
-  layout 'mailer'
-  helper :mailer
-  default from: "Archive of Our Own " + "<#{ArchiveConfig.RETURN_ADDRESS}>"
-
+class AdminMailer < ArchiveMailer
   def abuse_report(abuse_report_id)
     abuse_report = AbuseReport.find(abuse_report_id)
     @email = abuse_report.email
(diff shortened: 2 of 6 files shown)


redsummernight:
I think we should send emails from `after_commit` instead of `after_update` (in addition to this retry fix), so we don't have Redis ahead of MySQL in the first place.

config/initializers/gem-plugin_config/resque_mailer.rb
```diff
@@ -1 +1,10 @@
 Resque::Mailer.excluded_environments = [:test, :cucumber]
+Resque::Mailer.error_handler = lambda { |mailer, message, error, action, args|
+  if error.is_a?(Resque::TermException)
+    Resque.enqueue(mailer, action, *args)
+  elsif error.is_a?(ActiveRecord::RecordNotFound) && !args.include?(:retried)
+    Resque.enqueue_in(5.minutes, mailer, action, :retried, *args)
```
tickinginstant: This looks like it prepends the symbol `:retried` to the arguments passed to the mailer. Doesn't that mean all the arguments will be shifted over by 1, and the first argument (which is frequently supposed to be an ID of some sort) will be `:retried`? Or is there some code that strips the `:retried` symbol?

Also, I poked around the mailer code and I think `Resque::Mailer.perform` expects the arguments to be serialized:
https://github.com/zapnap/resque_mailer/blob/35c2a7c7ff9038b9bff95c50190d162f95afcd5f/lib/resque_mailer.rb#L52-L54

tickinginstant:
After some more poking around in the Resque Mailer code, I don't think it's going to be possible to use the built-in error handling to accomplish this, at least not without modifying some of the mailer code. For any mailers that take advantage of `Resque::Mailer::Serializers::ActiveRecordSerializer` by passing ActiveRecords as arguments, the `ActiveRecord::RecordNotFound` exception occurs in the serializer, before it's finished deserializing the arguments. But the error handler passes only the deserialized arguments (which in this case would be nil, because the error was thrown before they could be set). So it's impossible to resend the email from the error handler, because it doesn't have all the information it needs.

I'd suggest either modifying `config/initializers/resque_mailer.rb` to stop using `Resque::Mailer::Serializers::ActiveRecordSerializer` (with all of the changes that would entail), or overriding `self.perform` in the `ArchiveMailer` class to add a custom error handler (with a call to super).

(Alternatively, it might be possible to report this bug and get the Resque Mailer to handle errors in the serializer more sensibly.)

sarken:
Just doing some housekeeping! We're going to close this since it doesn't sound like the approach will work. The issue will still be open and assigned to you if you want to take another crack at it, though! <3

PR #3399, opened 2018-07-29
AO3-5496: Update transaction gems

elzj:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5496

## Purpose

Updates transaction isolation gem to fix confusing error messages when db transactions fail.

## Testing

Make sure automated tests pass and the staging site doesn't blow up.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -27,7 +27,7 @@ gem 'mysql2', '0.3.20'
 #https://github.com/qertoip/transaction_retry
 # We don't use the isolation gem directly, but it needs to be
 # at the latest version to avoid errors
-# gem 'transaction_isolation', '1.0.5'
+gem 'transaction_isolation', '1.0.5'
 gem 'transaction_retry'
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -25,6 +25,9 @@ gem 'rails-controller-testing'
 gem 'mysql2', '0.3.20'
 
 #https://github.com/qertoip/transaction_retry
+# We don't use the isolation gem directly, but it needs to be
+# at the latest version to avoid errors
+# gem 'transaction_isolation', '1.0.5'
```
redsummernight: Should we keep this if it's all commented out, and "latest version" will eventually drift? I'm fine with just the lock file change.

sarken: The first two lines of the new comment might not be a bad idea, but I agree the third line is probably unnecessary.

elzj: Okay, uncommented it so it's clearer.

PR #3403, opened 2018-08-03
AO3-5486 Make user data output less confusing

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5486

## Purpose

* Writes out names of favorite tags
* Adjusts phrasing for collection roles
* Adds relevant IPs from audit table to IP list
* Adds lists of previous emails and usernames, if the user has changed their email or username

### DIFF
--- a/script/get_user_data.rb
+++ b/script/get_user_data.rb
@@ -55,11 +55,7 @@ end
 collection_roles = []
 u.pseuds.each do |pseud|
   pseud.collection_participants.pluck(:participant_role, :collection_id)&.map do |role, collection_id|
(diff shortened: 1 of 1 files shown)


script/get_user_data.rb
```diff
@@ -15,6 +15,12 @@
 # Find the user
 u = User.find_by(login: login)
 
+# Names of user's favorite tags
+favorite_tags = []
```
redsummernight: You can do `favorite_tags = u.favorite_tags.joins(:tag).pluck(:name)`.

PR #3410, opened 2018-08-16
AO3-5510: Make tag update transaction size configurable

elzj:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5510

## Purpose

Makes the batch transaction size for tag count updates configurable so that we can tinker with it and see if that's the cause of problematic long-running transactions.

## Testing

Nothing should be visibly different, and tag counts should still eventually update.

### DIFF
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -50,7 +50,8 @@ class Tag < ApplicationRecord
   end
 
   def self.write_redis_to_database
-    REDIS_GENERAL.smembers("tag_update").each_slice(1000) do |batch|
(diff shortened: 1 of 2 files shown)


app/models/tag.rb
```diff
@@ -50,7 +50,8 @@ def document_json
   end
 
   def self.write_redis_to_database
-    REDIS_GENERAL.smembers("tag_update").each_slice(1000) do |batch|
+    batch_size = ArchiveConfig.TAG_UPDATE_BATCH_SIZE
```
zz9pzza: Can we set a default of say 100

PR #3420, opened 2018-09-08
AO3-4638 Use MySQL 5.7.22 in CI

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4638

## Purpose

Have Travis use 5.7 latest (it has 5.7.23 and I can't get 5.7.22 from the APT add-on). Have Codeship use 5.7.22. In both cases we need to exclude ONLY_FULL_GROUP_BY from the SQL mode, or a lot of queries using GROUP BY would break.

## Testing

None, automated.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,13 @@ env:
   - TEST_GROUP="cucumber -r features features/tags_and_wrangling -f Ao3Cucumber::Formatter --color"
   - TEST_GROUP="cucumber -r features features/users              -f Ao3Cucumber::Formatter --color"
   - TEST_GROUP="cucumber -r features features/works              -f Ao3Cucumber::Formatter --color"
+addons:
+  apt:
+    sources:
+      - mysql-5.7-trusty
+    packages:
+      - mysql-server
+      - mysql-client
 rvm:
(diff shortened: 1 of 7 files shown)


redsummernight:
Example of a failure if we don't exclude ONLY_FULL_GROUP_BY:

```rb
An error occurred while loading ./spec/controllers/api/v2/api_works_search_spec.rb.
Failure/Error:
  Subscription.for_work(work).each do |subscription|
    RedisMailQueue.queue_subscription(subscription, self)
  end
ActiveRecord::StatementInvalid:
  Mysql2::Error: Expression #1 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'otwarchive_test.subscriptions.id' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by: SELECT `subscriptions`.* FROM `subscriptions` WHERE ((subscribable_id = 2 AND subscribable_type = 'Work')
              OR (subscribable_id IN (2) AND subscribable_type = 'User')
              OR (subscribable_id IN (NULL) AND subscribable_type = 'Series')) GROUP BY `subscriptions`.`user_id`
```

PR #3422, opened 2018-09-16
AO3-5382 Store the signup summary in Memcached.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5382

## Purpose

This PR does the following:

1. It modifies the signup summary code so that it uses the Rails cache instead of the file system in order to generate the summary in background. This should ensure that the cached version is generated and saved in a place that's accessible to all servers, so that it can be displayed properly.

2. It moves all of the signup summary code into the `ChallengeSignupSummary` class.

3. In order to render the signup summary, it now uses [`ChallengeSignupsController.render`](https://api.rubyonrails.org/classes/ActionController/Renderer.html).

## Testing

Find or create a gift exchange with enough sign-ups to display a summary (and too many sign-ups to have the summary generated live), and make sure that the summary displays.

### DIFF
--- a/app/controllers/challenge_signups_controller.rb
+++ b/app/controllers/challenge_signups_controller.rb
@@ -127,25 +127,32 @@ class ChallengeSignupsController < ApplicationController
   end
 
   def summary
+    @summary = ChallengeSignupSummary.new(@collection)
+
     if @collection.signups.count < (ArchiveConfig.ANONYMOUS_THRESHOLD_COUNT/2)
       flash.now[:notice] = ts("Summary does not appear until at least %{count} sign-ups have been made!", count: ((ArchiveConfig.ANONYMOUS_THRESHOLD_COUNT/2)))
     elsif @collection.signups.count > ArchiveConfig.MAX_SIGNUPS_FOR_LIVE_SUMMARY
       # too many signups in this collection to show the summary page "live"
-      if !File.exists?(ChallengeSignup.summary_file(@collection)) ||
-          (@collection.challenge.signup_open? && File.mtime(ChallengeSignup.summary_file(@collection)) < 1.hour.ago)
-        # either the file is missing, or signup is open and the last regeneration was more than an hour ago.
+      modification_time = @summary.cached_time
 
-        # touch the file so we don't generate a second request
-        summary_dir = ChallengeSignup.summary_dir
-        FileUtils.mkdir_p(summary_dir) unless File.directory?(summary_dir)
-        FileUtils.touch(ChallengeSignup.summary_file(@collection))
+      # The time is always written alongside the cache, so if the time is
+      # missing, then the cache must be missing as well -- and we want to
+      # generate it. We also want to generate it if signups are open and it was
+      # last generated more than an hour ago.
+      if modification_time.nil? ||
+         (@collection.challenge.signup_open? && modification_time < 1.hour.ago)
 
-        # generate the page
-        ChallengeSignup.generate_summary(@collection)
+        # Touch the cache so that we don't try to generate the summary a second
+        # time on subsequent page loads.
+        @summary.touch_cache
+
+        # Generate the cache of the summary in the background.
+        @summary.enqueue_for_generation
       end
     else
       # generate it on the fly
-      @tag_type, @summary_tags = ChallengeSignup.generate_summary_tags(@collection)
+      @tag_type = @summary.tag_type
+      @summary_tags = @summary.summary
       @generated_live = true
     end
   end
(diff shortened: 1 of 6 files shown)


features/gift_exchanges/signup_summary.feature
```diff
@@ -0,0 +1,41 @@
+Feature: Gift Exchange Signup Summary
```
sarken: This is definitely an improvement over the current lack of tests, but it might be good to check that the number of Offers and Requests for the fandoms are correct as well -- if those are wrong, the summary isn't as useful as it could be. If that's too big of a pain in the butt, though, this is fine as it is.

tickinginstant: There are a few pre-existing tests for the sign-up summary in `challenge_yuletide.feature` (which involves more sign-ups and checks the offer/request numbers). I specifically wanted to write two smaller tests just to check for the difference in behavior between the live version and the delayed version, but it's easy enough to check the numbers here, too.

sarken: Thanks!

app/controllers/challenge_signups_controller.rb
```diff
@@ -127,25 +127,32 @@ def index
...
+      if modification_time.nil? ||
+         (@collection.challenge.signup_open? && modification_time < 1.hour.ago)
 
-        # touch the file so we don't generate a second request
-        summary_dir = ChallengeSignup.summary_dir
-        FileUtils.mkdir_p(summary_dir) unless File.directory?(summary_dir)
-        FileUtils.touch(ChallengeSignup.summary_file(@collection))
+        # Touch the cache so that we don't try to generate the summary a second
+        # time on subsequent page loads.
+        @summary.touch_cache
```
zz9pzza: I can't see how this works when the code can be run on lost of different servers ( I can see there is no migrations so the shared state is not in mysql ).

I can imagine writing a please wait message in to memcached which would achive this ?

app/models/challenge_signup_summary.rb
```diff
@@ -38,6 +51,85 @@ def tag_summary(tag)
...
+  # GENERATING THE SUMMARY IN RESQUE
+  ######################################################################
+
+  @queue = :collection
+
+  # The action to be performed when this class is enqueued using Resque.
+  def self.perform(collection_id)
+    collection = Collection.find(collection_id)
+    summary = ChallengeSignupSummary.new(collection)
+    summary.generate_in_background
```
zz9pzza: Is it worth scheduling another run in say 50 minutes if the challenge is still open ?

app/controllers/challenge_signups_controller.rb
zz9pzza: Ariana pointed out its a methed call.... sorry

app/models/challenge_signup_summary.rb
```diff
@@ -38,6 +51,85 @@ def tag_summary(tag)
...
+  def cached_time
+    cached_info[:time]
+  end
+
+  # The equivalent of touching a file in the file system. Update the
+  # modification time.
+  def touch_cache
+    data = cached_info.dup
+    data[:time] = Time.now
+    Rails.cache.write(cache_key, data)
```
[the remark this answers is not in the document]
tickinginstant: I tried to mimic the old behavior as much as possible. While the old code did set up the summary to be regenerated after an hour, it also deliberately kept the old summary around after that hour had elapsed, so that the old summary could be displayed even while the new summary was being generated in the background. (And it also kept the old sign-up summary indefinitely once sign-ups were closed.) The new code is intended to work in much the same way.

The disadvantage of setting `expires_in` is that it deletes the old sign-up summary, resulting in a period where the old sign-up summary is missing, but the new sign-up summary hasn't been generated yet. That's not necessarily a problem, particularly if generating the summary is snappy and the Resque queue is short, but it's also not ideal. IMHO, if `expires_in` is set at all, it should be set to something fairly long.

(If the summary should be regenerated when the challenge is closed -- which is one of the things that setting `expires_in` would accomplish -- it would probably be simplest to add another clause to the timeout check in `ChallengeSignupsController` instead: e.g. `modification_time < 24.hours.ago`. That way, there'll always be a sign-up summary to display, even if it's a little out of date. Well, unless Memcached runs out of room and starts deleting old keys.)

app/models/challenge_signup_summary.rb
tickinginstant: I'm not sure I understand what you're suggesting. Do you want me to make sure that it's consistently regenerated every 50 minutes until sign-ups close, even if no one is looking at it?

PR #3424, opened 2018-09-17
AO3-5310 Track and display IP address of invitation requests 

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5310

## Purpose

Lets admins see the IP address of those who have requested invitations.

## Testing

Make sure the IP address is displayed as described in Jira

## Note

When this is merged, add IP address of invitation requests to our document listing all of the PII data we collect.


spec/controllers/invite_requests_controller_spec.rb
```diff
@@ -60,11 +60,24 @@
...
+      let(:ip) { "867.53.09" }
+
+      before do
+        allow_any_instance_of(ActionDispatch::Request).to receive(:remote_ip).and_return(ip)
+      end
+
+      it "redirects to index with error given valid emails" do
+        email = generate(:email)
+        post :create, params: { invite_request: { email: email } }
+        invite_request = InviteRequest.find_by_email(email)
```
redsummernight: This is meant to test if the invite request is actually created, so we should use `find_by!` to fail the test if the invite is not found.

spec/controllers/invite_requests_controller_spec.rb
```diff
@@ -60,11 +60,24 @@
...
+      let(:ip) { "867.53.09" }
+
+      before do
+        allow_any_instance_of(ActionDispatch::Request).to receive(:remote_ip).and_return(ip)
+      end
+
+      it "redirects to index with notice" do
+        email = generate(:email)
+        post :create, params: { invite_request: { email: email } }
+        invite_request = InviteRequest.find_by!(email)
```
redsummernight: This will pass the email as a WHERE clause. You should do `InviteRequest.find_by!(email: email)`.

sarken: Here is the truly embarrassing thing: I hallucinated that the `email:` was there. When I was changing it to `find_by!` I stopped and asked myself, "Huh, why is `InviteRequest.find_by_email(email: email)` even working to begin with?" 🤦‍♂️

PR #3425, opened 2018-09-17
Ignore vulnerability in ruby-ffi 1.9.18

redsummernight:
## Issue

None.

## Purpose

This only affects Windows, and the gem will be removed when we switch from Authlogic to Devise.

## Testing

None, automated.

## References

See also #3378.


redsummernight:
I added a note on the Devise pull request.

PR #3426, opened 2018-09-24
AO3-5529 Order collected works by date updated by default

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5529

## Purpose

We accidentally started ordering users' collected works pages by score, not date updated. This puts it back.

## Testing

Refer to Jira.

## Credit

@redsummernight did the app/models/search/work_search_form.rb changes.


scottsds:
Looks good 👍

PR #3427, opened 2018-09-24
AO3-5531 Don't use hidden field for abuse report IP, plus fix live validation

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5531

## Purpose

* Gets the IP address for abuse reports in the controller and not via a hidden field, which was subject to manipulation
* Fixes the LiveValidation on the "Brief summary of Terms of Service violation" field
* Updates the quotation marks
* Writers out `root_path` for the redirect instead of using `''` because 😭 old code

I can move any of that to separate, lower priority pull requests as long as we don't mind increasing the pile.

## Testing

Refer to Jira


redsummernight:
We should remove `ip_address` from `abuse_report_params`.

PR #3432, opened 2018-10-13
AO3-2046 Fix word count for Chinese, Japanese and Thai

yfjia:
## Issue

https://otwarchive.atlassian.net/browse/AO3-2046

## Purpose

The current word count doesn't work for languages that don't use spaces for word separation. This new word count method changed the regex when scanning words and count characters in Chinese, Japanese and Thai as words. The new word count is the count expected by users in these languages.

When theses languages are mixed with other languages, words in the three languages above would be counted based on characters while other languages are counted as usual. 

Chinese, Japanese and Thai are the only languages that don't use space as separators and are supported by AO3 that I am aware of. Other such languages can be added to `languages_without_spaces` to support this special character counting.

## Testing

Added two spec tests for text with Chinese/Japanese words and Chinese/English words. The code passed the old and new tests.

Tested `WordCounter` locally with English and Chinese text corpus, English word count is the same as before, and Chinese word count is in the right range.

## References

Word count has always been a big problem for writers and readers using Chinese and other languages. Some writers resort to tags like "AO3's chinese word count times ten and then you get the correct number" to tell readers the actual word count. 

Even though the JIRA issue is listed as "In Progress," the issue has opened for 8 years and hasn't been updated in 10 months. It seems that the assignee chose not to move because counting based on characters might be confusing. However, at least in Chinese, it is generally accepted that "word count" refers to the number of characters, and using character count as word count for Chinese is consistent with the behavior of Microsoft Word's word count. 

## Credit

Tatayafinwe, she/her


lib/word_counter.rb
```diff
@@ -14,14 +14,22 @@ def initialize(text)
...
   def count
     count = 0
     # avoid blank? so we don't need to load Rails for tests
     return count if @text.nil? || @text.empty?
     body = Nokogiri::HTML(@text).xpath('//body').first
     body.traverse do |node|
       if node.is_a? Nokogiri::XML::Text
-        count += node.inner_text.gsub(/--/, "—").gsub(/['’‘-]/, "").scan(/[[:word:]]+/).size
+        # includes Chinese, Japanese and Thai
+        languages_without_spaces = "\\p{Han}|\\p{Katakana}|\\p{Hiragana}|\\p{Thai}"
```
redsummernight: We should move this to the configuration file, so we can add more scripts without deploying code changes. In config/config.yml, if you add:

```yml
# List of Unicode scripts where spaces are not used as separators, and
# character counts should be used for word counts.
# Check if a script is supported by the version of Ruby we use:
# https://ruby-doc.org/core/Regexp.html#class-Regexp-label-Character+Properties
CHARACTER_COUNT_SCRIPTS: ["Han", "Hiragana", "Katakana", "Thai"]
```

then you can get the list here by using `ArchiveConfig.CHARACTER_COUNT_SCRIPTS`.

yfjia: Yes, I'm yj77. I would really appreciate it if you give me the permission on JIRA! And sorry for not reading the testing heading carefully. There are no testing instructions I wish to add.

I've changed the mode of the file back and moved the language list to config. Thank you for the review!

redsummernight:
Sorry, I'm really off my game today... it should be `require "spec_helper"` at the top of `word_counter_spec.rb`, not `word_counter.rb`.

redsummernight:
Thanks!

yfjia:
I have thought about whether to include punctuation for Chinese word count as well. But I think the only way is to pass the language into word counter and handle Chinese separately, and I was new to the codebase and didn't want to make too much change.

Also, there are word processors like Apple's Pages that doesn't count punctuation marks for Chinese, so I'm not sure if adding it is necessary. But it certainly could be part of future work if most Chinese users prefer including punctuation (I am a native Chinese speaker and personally I don't have a preference).

sarken:
Thanks for the extra information! It's good to know there's some variation in how other software treats punctuation and that there's no universal preference. We'll just keep an eye out for any feedback.
 
Thanks again for coding this!

PR #3433, opened 2018-10-13
AO3-5530 Test posting works/comments with 4-byte UTF-8 characters

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5530

## Purpose

Once more with feeling (emoji).

Follow instructions from "[Converting a Rails database to utf8mb4 without downtime or data loss](https://railsmachine.com/articles/2017/05/19/converting-a-rails-database-to-utf8mb4.html)" and have the databases on Travis/Codeship use utf8mb4.

Also have the Travis database use the Barracuda file format.

Update the example local database configuration file.

## Testing

None.


config/database.codeship.yml
```diff
@@ -1,7 +1,7 @@
 development:
   adapter: mysql2
   host: 127.0.0.1
-  encoding: utf8
+  encoding: utf8mb4
```
zz9pzza: I would add  collation: utf8mb4_unicode_ci as well to these bits as we have that on staging

PR #3434, opened 2018-10-13
AO3-5152 use pluck where possible

meara:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5152

## Purpose

Takes advantage of `pluck` in queries where possible to avoid unnecessarily loading objects into memory.

## Testing

No manual testing required.

## Credit

@meara, she/her

### DIFF
--- a/app/views/abuse_reports/new.html.erb
+++ b/app/views/abuse_reports/new.html.erb
@@ -28,7 +28,7 @@
             (link_to ts("DMCA Policy"), dmca_path)).html_safe %></li>
   </ul>
   <p><%= ts("<strong>We can answer Abuse reports in %{list}.</strong> Please allow for an additional delay for responses
-            in any language other than English.", list: @abuse_languages.map(&:name).to_sentence).html_safe %></p>
+            in any language other than English.", list: @abuse_languages.pluck(:name).to_sentence).html_safe %></p>
 </div>
 <%= form_for @abuse_report, class: "post" do |f| %>
   <fieldset>
--- a/app/views/feedbacks/new.html.erb
+++ b/app/views/feedbacks/new.html.erb
@@ -23,7 +23,7 @@
   <div class="userstuff">
     <p><%= ts("If your question or problem concerns potential violations of the Terms of Service, control of Orphaned Works, or issues requiring identity verification such as regaining access to a lost account, please %{contact_abuse_link} instead.", contact_abuse_link: link_to(ts("contact our Policy and Abuse team"), new_abuse_report_path)).html_safe %></p>
     <p><%= ts("For current updates on Archive performance, please check out our Twitter feed: <a href=\"http://twitter.com/ao3_status\">@AO3_status</a>.").html_safe %></p>
-    <p><%= ts("<strong>We can answer Support inquiries in %{list}.</strong> Please allow for up to one week delay for responses in any language other than English.", list: @support_languages.map(&:name).to_sentence).html_safe %></p>
+    <p><%= ts("<strong>We can answer Support inquiries in %{list}.</strong> Please allow for up to one week delay for responses in any language other than English.", list: @support_languages.pluck(:name).to_sentence).html_safe %></p>
   </div>
   <%= form_for(@feedback, html: {class: "post feedback"}) do |f| %>
     <fieldset>
--- a/app/controllers/admin_posts_controller.rb
+++ b/app/controllers/admin_posts_controller.rb
@@ -84,7 +84,7 @@ class AdminPostsController < ApplicationController
   protected
 
   def load_languages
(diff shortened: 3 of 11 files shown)


meara:
Not sure how to update the JIRA issue--I don't appear to have access to change anything, and it doesn't seem like there's an integration picking up that the PR was created.

zz9pzza:
The test failure in travis is due to an issue with nokogiri which we need to deal with separately.

( Thanks for the pull request )

tickinginstant:
I've been meaning to mention this for a while: the issue of pluck vs. map is more complex than the bug report might suggest. In my experiments, pluck actually behaves much worse than map in certain contexts. In particular:

1. When called on an array of ActiveRecords (e.g. in `app/models/tag.rb`), pluck can be up to 3x slower than map.
2. When called on an ActiveRecord relation/collection that has already been loaded (e.g. on line 313 of `app/models/work.rb`), pluck can also be up to 3x slower than map.
3. When called on a collection of ActiveRecords that has not yet been loaded, but will end up being loaded later, switching from map to pluck results in a redundant database query.

In those circumstances, the switch from map to pluck is just a downgrade, with no upside.

There are also some circumstances where it's a mixed bag. For example, in `app/models/search/bookmark_indexer.rb`, switching from map to pluck would mean that the bookmark's tags never have to be loaded (which reduces the memory requirements), but it would also result in a second database query (which may make the whole thing a fair amount slower).

Basically, map vs. pluck is very context-dependent. There are, of course, several places where pluck would be a clear improvement, but it may require some careful examination to figure out which ones.

meara:
Thanks for all the comments folks! I’ll take another look at this on Sunday probably.

meara:
Notes--I took into account all the comments I think, and made a few judgment calls in cases where the best path is ambiguous (for example `bookmark_indexer.rb`). Without an actual DB snapshot with some real data in it, it's hard for me to benchmark actual performance, so I took my best guess.  If anyone does want to clock the alternatives and discovers that any of my changes are less performant, I'm happy to revert--or if it's easy to provide a DB dump I can check myself, although it might be a minute before I get around to it.

Thanks folks!

sarken:
It would be tricky to provide a database dump, but if there's benchmarking you'd like done, we'd be happy to run it on production!

app/views/abuse_reports/new.html.erb
```diff
@@ -28,7 +28,7 @@
             (link_to ts("DMCA Policy"), dmca_path)).html_safe %></li>
   </ul>
   <p><%= ts("<strong>We can answer Abuse reports in %{list}.</strong> Please allow for an additional delay for responses
-            in any language other than English.", list: @abuse_languages.map(&:name).to_sentence).html_safe %></p>
+            in any language other than English.", list: @abuse_languages.pluck(:name).to_sentence).html_safe %></p>
```
tickinginstant: I'm not 100% certain, but I think that the call to `f.collection_select` on line 50 below will force `@abuse_languages` to be loaded:
https://github.com/otwcode/otwarchive/blob/df6ce85db5cb1a857186215e43cd0cafba3234a2/app/views/abuse_reports/new.html.erb#L50
I think this is a case where the switch to pluck just results in an extra database query.

app/views/feedbacks/new.html.erb
```diff
@@ -23,7 +23,7 @@
   <div class="userstuff">
     <p><%= ts("If your question or problem concerns potential violations of the Terms of Service, control of Orphaned Works, or issues requiring identity verification such as regaining access to a lost account, please %{contact_abuse_link} instead.", contact_abuse_link: link_to(ts("contact our Policy and Abuse team"), new_abuse_report_path)).html_safe %></p>
     <p><%= ts("For current updates on Archive performance, please check out our Twitter feed: <a href=\"http://twitter.com/ao3_status\">@AO3_status</a>.").html_safe %></p>
-    <p><%= ts("<strong>We can answer Support inquiries in %{list}.</strong> Please allow for up to one week delay for responses in any language other than English.", list: @support_languages.map(&:name).to_sentence).html_safe %></p>
+    <p><%= ts("<strong>We can answer Support inquiries in %{list}.</strong> Please allow for up to one week delay for responses in any language other than English.", list: @support_languages.pluck(:name).to_sentence).html_safe %></p>
```
tickinginstant: Similarly here. There's a call to `f.collection_select` on line 53: https://github.com/otwcode/otwarchive/blob/df6ce85db5cb1a857186215e43cd0cafba3234a2/app/views/feedbacks/new.html.erb#L53

sarken:
Thank you for your work on this, everyone! However, given the discussion on this and the lack of clarity around what will actually be a helpful change, we're going to close this pull request and the associated issue.

PR #3438, opened 2018-10-15
AO3-3507 exclude RSS Feed button from filtered results

niconicosette:
## Issue

https://otwarchive.atlassian.net/browse/AO3-3507

## Purpose

removes RSS feed button from filtered pages, but not from admin pages or unfiltered tags

## Testing

confirm that RSS feed button has been removed from filtered work result pages, but not from admin pages or unfiltered work results

### DIFF
--- a/app/views/works/index.html.erb
+++ b/app/views/works/index.html.erb
@@ -33,7 +33,7 @@
         <% if @tag && logged_in? && !@collection %>
           <li><%= render 'favorite_tags/form', current_user: @current_user, favorite_tag: @favorite_tag %></li>
         <% end %>
-        <% if @tag && !@collection && (%w(Fandom Character Relationship).include?(@tag.type.to_s) || @tag.name == "F/F") %>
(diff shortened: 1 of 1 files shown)


app/views/works/index.html.erb
```diff
@@ -33,7 +33,7 @@
         <% if @tag && logged_in? && !@collection %>
           <li><%= render 'favorite_tags/form', current_user: @current_user, favorite_tag: @favorite_tag %></li>
         <% end %>
-        <% if @tag && !@collection && (%w(Fandom Character Relationship).include?(@tag.type.to_s) || @tag.name == "F/F") %>
+        <% if @tag && !@collection && !params[:work_search] && (%w(Fandom Character Relationship).include?(@tag.type.to_s) || @tag.name == "F/F") %>
```
redsummernight: After ES6, we also need to check for `params[:include_work_search]` and `params[:exclude_work_search]`, similar to #3550.

sarken:
Just checking if you'll be able to update this pull request, or if someone else should grab it! We are, in a shocking turn of events, running low on quick and easy things to merge /o\

sarken:
I spoke with the coder and they won't have time to work on this soon, so I'm going to close it for now.

PR #3444, opened 2018-10-15
AO3-5527 Reset invitations for purged accounts.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5527

## Purpose

My fix for the issue of invitations used on deleted accounts made it so that users who sign up, fail to confirm their account within the 2 week deadline, and subsequently have their account purged will not be able to reuse their invitations. But we want them to be able to reuse their invitations, so this PR adds a few lines to the account-purging rake task to reset the invitations for any purged accounts.

## Testing

This one's probably a bit tricky to test manually unless you have a two-week old unconfirmed account lying around, or you're willing to wait two weeks for QA. You might consider creating an unconfirmed account, then using the Rails console to update its `created_at` date so that it's caught in the next purge.

### DIFF
--- a/spec/controllers/users_controller_spec.rb
+++ b/spec/controllers/users_controller_spec.rb
@@ -1,4 +1,5 @@
 require 'spec_helper'
+require 'rake'
 
 describe UsersController do
   include RedirectExpectationHelper
@@ -62,7 +63,7 @@ describe UsersController do
       end
 
       context "signing up with a valid invitation" do
-        it "succeeeds in creating the account" do
+        it "succeeds in creating the account" do
           post :create, params: { user: valid_user_attributes,
                                   invitation_token: invitation.token }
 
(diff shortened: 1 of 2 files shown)


spec/controllers/users_controller_spec.rb
```diff
@@ -106,6 +107,31 @@ def valid_user_attributes
             )
           end
         end
+
+        context "when the previous user's account was purged" do
```
redsummernight: Could you move this test to `admin_tasks.rake_spec`?

The official reason is to match the file of the code under test.
The real reason is this will introduce conflicts with #3378.

tickinginstant: I can test for the properties of the invitation there, but checking the controller's response there would require a lot of code duplication (and would introduce compatibility issues with #3378 anyway, since I believe it changes both the parameter name and the path).

Are you okay with not checking the controller?

sarken: Could we punt the tests to another issue and add them sometime after we've finished Devise? We can test manually for now. 

(I foolishly didn't think about potential conflicts yesterday morning before approving this issue for a quick single-issue deploy while we updated the Devise branch /o\)

tickinginstant: I reverted the test changes in `spec/controllers/users_controller_spec.rb` and added a couple of tests for the purge account task (including one that checks the fields on the invitation). It doesn't check the controller's response, but it should be compatible with #3378 without requiring any extra code changes.

PR #3448, opened 2018-10-22
AO3-5547 Upgrade Elasticsearch to 6.4.3 for CI

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5547

## Purpose

Upgrades Elasticsearch! Which actually just involves upgrading tests (no code changes this time!) while james_ does work on the servers.

## Testing

We'll want to give searching and filtering a quick poke once james_ has upgraded staging just to make sure there's nothing broken that our tests didn't catch.

## Credit

james_ (he/him) and Sarken (she/her)


script/travis_elasticsearch_upgrade.sh
```diff
@@ -1,7 +1,9 @@
 #!/bin/bash
 cd /tmp
-wget https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-6.2.4.tar.gz
-tar xvfz /tmp/elasticsearch-6.2.4.tar.gz
```
redsummernight: Can you move the version to a variable `ES_VERSION`, like the Codeship script?

sarken: Indeed I can! (Fair warning, this is going to be a force push so this comment will probably disappear.)

sarken: ...false alarm 🎉

PR #3449, opened 2018-10-24
AO3-5537 Spam check abuse reports

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5537

## Purpose

Runs abuse reports through the spam checker and gives the user an error if it gets flagged as spam.

## Testing

Refer to Jira

### DIFF
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -8,6 +8,7 @@
 require 'simplecov'
 require 'cucumber/timecop'
 require 'capybara/poltergeist'
+require 'cucumber/rspec/doubles'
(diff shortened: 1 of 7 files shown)


features/support/env.rb
```diff
@@ -8,6 +8,7 @@
 require 'simplecov'
 require 'cucumber/timecop'
 require 'capybara/poltergeist'
+require 'cucumber/rspec/doubles'
```
redsummernight: I knew I've seen this somewhere. Can you remove the same require from here?

https://github.com/otwcode/otwarchive/blob/df6ce85db5cb1a857186215e43cd0cafba3234a2/features/step_definitions/search_steps.rb#L1

PR #3454, opened 2018-10-29
AO3-3953 Remove collection-scoped tag links

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-3953

## Purpose

Links to tag pages (showing meta tags, sub tags, etc.) should never be collection-scoped.

Links to tag work listings should be collection-scoped only in two places:

- collections/COLLECTIONNAME/tags
- collections/COLLECTIONNAME/fandoms

## Testing Instructions

See issue.

## References

Updated version of #1678.


tickinginstant:
Thank you for submitting this PR! It looks great, and I'm so happy that we're one step closer to never encountering this bug ever again.

sarken:
Damn, just occurred to me... should we have expired the work blurb cache key for this to make sure any old collection scoped links get cleared out?

redsummernight:
Yeah we should, #3583.

PR #3463, opened 2018-11-12
AO3-5559 Remove class attributes on all but some fields

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5559
https://otwarchive.atlassian.net/browse/AO3-5238 (because you can no longer use classes there)

## Purpose

Better enforcement of the config setting that only allows class attributes on specific fields.

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -1042,7 +1042,7 @@ class WorksController < ApplicationController
       relationship: params[:work][:relationship_string],
       category: params[:work][:category_string],
       freeform: params[:work][:freeform_string],
-      notes: params[:notes],
+      notes: params[:work_notes],
       encoding: params[:encoding],
       external_author_name: params[:external_author_name],
       external_author_email: params[:external_author_email],
--- a/app/models/story_parser.rb
+++ b/app/models/story_parser.rb
@@ -699,7 +699,7 @@ class StoryParser
     # cleanup the notes
     notes.gsub!(%r{<br\s*\/?>}i, "\n") # replace the breaks with newlines
     notes = clean_storytext(notes)
-    work_params[:notes] = notes
+    work_params[:work_notes] = notes
 
     work_params.merge!(scan_text_for_meta(notes, detect_tags))
     work_params[:title] = title
--- a/app/views/works/new_import.html.erb
+++ b/app/views/works/new_import.html.erb
@@ -123,11 +123,11 @@
   <%= render :partial => 'work_form_tags', locals: { include_blank: true, import: true } %>
 
   <fieldset>
-    <legend><%= label_tag :notes, ts("Notes") %></legend>
+    <legend><%= label_tag :work_notes, ts("Notes") %></legend>
     <dl>
-      <dt><%= label_tag :notes, ts("Notes at the beginning") %></dt>
+      <dt><%= label_tag :work_notes, ts("Notes at the beginning") %></dt>
       <dd>
-        <%= text_area_tag :notes, nil, class: "observe_textlength" %>
+        <%= text_area_tag :work_notes, nil, class: "observe_textlength" %>
         <%= generate_countdown_html(".notes", ArchiveConfig.NOTES_MAX) %>
       </dd>
     </dl>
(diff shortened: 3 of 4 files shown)


config/initializers/gem-plugin_config/sanitizer_config.rb
```diff
@@ -38,17 +39,26 @@ module Config
         'img' => {'src' => ['http', 'https', :relative]},
         'q' => {'cite' => ['http', 'https', :relative]}
       }
-    }
+    )
+
+    CLASS_ATTRIBUTE = freeze_config(
+      # see in the Transformers section for what classes we strip
+      attributes: {
+        all: ['align', 'title', 'dir', 'class'],
```
redsummernight: You can do:
```ruby
all: ARCHIVE[:attributes][:all] + ['class']
```
and we don't need the "MUST be duplicated" comment.

config/initializers/gem-plugin_config/sanitizer_config.rb
```diff
@@ -38,17 +39,26 @@ module Config
...
+      }
+    )
+
+    CSS_ALLOWED = merge(ARCHIVE, CLASS_ATTRIBUTE)
```
redsummernight: If `CLASS_ATTRIBUTE` is not used elsewhere you can inline it here. `CSS_ALLOWED` needs a `freeze_config` on it.

lib/html_cleaner.rb
```diff
@@ -158,10 +158,14 @@ def sanitize_value(field, value)
       if ArchiveConfig.FIELDS_ALLOWING_CSS.include?(field.to_s)
         transformers << Sanitize::Transformers::ALLOW_USER_CLASSES
       end
-      # the screencast field shouldn't be wrapped in <p> tags
-      unless field.to_s == "screencast"
+      # Now that we know what transformers we need, let's santize the unfrozen value
```
redsummernight: Should be "sanitize".

lib/html_cleaner.rb
```diff
@@ -158,10 +158,14 @@ def sanitize_value(field, value)
...
+      if ArchiveConfig.FIELDS_ALLOWING_CSS.include?(field.to_s)
+        unfrozen_value = add_paragraphs_to_text(Sanitize.clean(fix_bad_characters(unfrozen_value),
+                               Sanitize::Config::CSS_ALLOWED.merge(transformers: transformers)))
```
redsummernight: Notes for myself: `Sanitize::Config::CSS_ALLOWED` is to let class attributes exist in the first place, while `Sanitize::Transformers::ALLOW_USER_CLASSES` is to make sure those class attributes don't have weird values.

app/views/works/_notes_form.html.erb
```diff
@@ -15,25 +17,27 @@
...
         <legend><%= f.label :notes, ts("Notes") %></legend>
         <% if @posting_claim && @posting_claim.prompt_description.present? %>
-          <%= f.text_area :notes, :class => "observe_textlength", :value => "<strong>Prompt:</strong> #{@posting_claim.prompt_description}" %>
-          <%= generate_countdown_html("#{type}_notes", ArchiveConfig.NOTES_MAX) %>
+          <%= f.text_area notes_object, :class => "observe_textlength", :value => "<strong>Prompt:</strong> #{@posting_claim.prompt_description}", :id => notes_id %>
+          <%= generate_countdown_html(notes_id, ArchiveConfig.NOTES_MAX) %>
         <% else %>
-          <%= f.text_area :notes, :class => "observe_textlength" %>
-          <%= generate_countdown_html("#{type}_notes", ArchiveConfig.NOTES_MAX) %>
+          <%= f.text_area notes_object, :class => "observe_textlength", id: notes_id %>
```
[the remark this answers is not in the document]
redsummernight: Nah you're fine. If you do want to update the syntax, you could do it only for the lines you've already edited.

redsummernight:
Adding to @tickinginstant's point about fields that we miss, is it worth (I'm sorry) blocking off the old getters and setters, like this?

```
  alias_attribute :work_notes, :notes
  def notes
    raise "WTF"
  end
  def notes=(value)
    raise "WTF"
  end
```

Now there's no chance we use the wrong one. >.<

PR #3465, opened 2018-11-13
AO3-5560 Comment permissions on hidden works.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5560

## Purpose

Checks whether the work is hidden by an admin when viewing/modifying comments.

## Testing Instructions

See the bug report.


app/views/comments/_comment_actions.html.erb
```diff
@@ -1,7 +1,7 @@
 <h5 class="landmark heading"><%= ts("Comment Actions") %></h5>
 
 <ul class="actions" role="menu" id="navigation_for_comment_<%= comment.id %>">
-  <% unless comment.unreviewed? || no_anon_reply(comment) %>
+  <% if can_reply_to_comment?(comment) %>
```
sarken: Much better 😍

tickinginstant: I know whoever ends up tackling 5561 will probably have to add another condition here, so I figured I'd move the logic into the helpers before it gets more complex.

PR #3467, opened 2018-11-13
AO3-5559 Rename bookmark notes, comment content, and series notes columns

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5559

## Purpose

Renames some columns (and updates code accordingly) so we can apply a different sanitizer to them.

Quick note on why I chose `bookmarker_notes` rather than `bookmark_notes`: to make sure bookmark notes are displayed correctly, we need the parser to add HTML to them. However, we previously ran into an issue with search where calling the field `notes` meant HTML would be added to it, so [we did this](https://github.com/otwcode/otwarchive/blob/bd4f323003a7f63d0e708ff334fd666e4760601a/app/models/search/bookmark_search_form.rb#L72) to work around it. If I had gone with `bookmark_notes` for the new column name, we would've needed to change that as well. It just seemed easier to stick the `er` on the column name than to make more code changes. :P 

## Testing Instructions

Hahaha sob. I'll update the issue.

### DIFF
--- a/config/config.yml
+++ b/config/config.yml
@@ -263,8 +263,8 @@ FIELDS_ALLOWING_LESS_THAN: ["query", "words", "kudos", "hits", "date", "bookmark
 #
 # This will ensure that the field has been sanitized with the latest version of the sanitizer.
 #
-FIELDS_ALLOWING_HTML: ["about_me", "comment", "content", "description", "endnotes", "faq", "intro", "note", "notes",
-  "rules", "signup_instructions_general", "signup_instructions_offers", "signup_instructions_requests", "summary",
+FIELDS_ALLOWING_HTML: ["about_me", "bookmark_notes", "comment", "comment_content", "content", "description", "endnotes", "faq", "intro", "note", "notes",
+  "rules", "series_notes", "signup_instructions_general", "signup_instructions_offers", "signup_instructions_requests", "summary",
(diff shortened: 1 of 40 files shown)


config/config.yml
```diff
@@ -263,9 +263,8 @@ FIELDS_ALLOWING_LESS_THAN: ["query", "words", "kudos", "hits", "date", "bookmark
 #
 # This will ensure that the field has been sanitized with the latest version of the sanitizer.
 #
-FIELDS_ALLOWING_HTML: ["about_me", "comment", "content", "description", "endnotes", "faq", "intro", "note", "notes",
-  "rules", "signup_instructions_general", "signup_instructions_offers", "signup_instructions_requests", "summary",
-  "screencast", "disabled_support_form_text"]
+FIELDS_ALLOWING_HTML: ["about_me", "bookmark_notes", "comment", "comment_content", "content", "description", "disabled_support_form_text", "endnotes", "faq", "intro", "note", "notes",
```
tickinginstant: Should this be `bookmarker_notes`? (I'm not sure I follow the `bookmarker_notes` vs. `bookmark_notes` distinction yet, so feel free to ignore me if this is a silly question.)

sarken: It should! I forgot to change it after my last minute change to the column name. I will fix that and add a test.

PR #3473, opened 2018-11-23
AO3-2378 Add task to reset word count for works of a language

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-2378

## Purpose

Add a new task for fixing word counts.
Add new test type for Rake tasks.

## Testing Instructions

See issue.

### DIFF
--- a/lib/tasks/work_tasks.rake
+++ b/lib/tasks/work_tasks.rake
@@ -5,7 +5,7 @@ namespace :work do
     puts "Unposted works (#{count}) created more than one month ago have been purged"
   end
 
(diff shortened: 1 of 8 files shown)


lib/tasks/work_tasks.rake
```diff
@@ -15,4 +15,25 @@ namespace :work do
     end
   end
 
+  # Usage: rake work:reset_word_counts[en]
```
sarken: Would this be better in the after tasks rake file? That's where we usually put one-time use clean-up type things.

PR #3477, opened 2018-11-30
AO3-2839 Fix Devise settings, remove unused code in users_controller

redsummernight:
## Issue

## Purpose

- Remove unused routes users#new and users#create, which have been moved to registrations_controller.
- Set password reset tokens to expire in 48h.
- Set remember me sessions to expire in 3 months.

### DIFF
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -195,7 +195,7 @@ Devise.setup do |config|
   # Time interval you can reset your password with a reset password key.
   # Don't put a too small interval or your users won't have the time to
   # change their passwords.
-  config.reset_password_within = 6.hours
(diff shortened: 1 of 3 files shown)


config/initializers/devise.rb
```diff
@@ -195,7 +195,7 @@
   # Time interval you can reset your password with a reset password key.
   # Don't put a too small interval or your users won't have the time to
   # change their passwords.
-  config.reset_password_within = 6.hours
+  config.reset_password_within = 48.hours
```
redsummernight: This doesn't actually do anything right now. We'll change that later in [AO3-5576](https://otwarchive.atlassian.net/browse/AO3-5576).

PR #3478, opened 2018-12-01
AO3-5533 & AO3-5544 Don't use after_commit for update_anon_unrevealed.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5533
https://otwarchive.atlassian.net/browse/AO3-5544

## Purpose

Using `after_commit` for the anonymous/unrevealed updates means that any changes to the `CollectionItem` that were just written to the database might not be available to read (thanks to various delays in replication). So `after_save`/`after_destroy` is preferred. Unfortunately, that change requires a number of other alterations to avoid causing errors and/or introducing undesirable behavior.

In particular:
1. The `update_work` callback should not run when the work is first being created. (The work already has `set_anon_unrevealed` as a `before_save` hook, and if you're adding multiple collection items in the very first save, then the first collection item's `after_save` hook will run before the second collection item is written to the database. If the second collection item is the one that's supposed to hide the work, this could result in the work being revealed during the posting process.)

   However, the order of saving for brand new records is: the record is saved first, and then any of its associations. So when the collection item's `after_save` callback is run, the work is already saved to the database and is no longer a `new_record?`. So we add an extra check for `saved_change_to_id?`

2. The `update_work` callback should not run when the work is being destroyed. However, the order for destroying records is: any associations set up with `dependent: :destroy` are destroyed, and then the work itself. The work is not marked as `destroyed?` until after the collection items have already had their `after_destroy` callback, which means it still counts as `persisted?` even though it's in the process of being destroyed. So we remove the `dependent: :destroy` option and set up a manual callback to destroy the collection items instead.

3. Currently, there's a bit of broken behavior that occurs when you try to remove a collection from a work, and hit "Preview" instead of "Post Without Preview." Instead of succeeding, it generates an error (which results in a blank white page on staging).

   This occurs because when in preview mode, all of the `#{tag_type}_string` fields are initially set to blank, and they're only filled in as the attributes are assigned. The `collections` attribute is assigned before `warning_strings`, so at the time that the collection is removed from the work, the work is "invalid" as a result of having no warnings (even though it does have warnings in the database). But removing a collection calls `update_visibility`, which saves the work, which runs the validations, which complain because the work is "missing a warning" (even though it's not).

   Note that this doesn't occur when adding a collection because adding a collection relies on the `CollectionItem`'s callback (which calls `update_anon_unrevealed!` on a different copy of the work that doesn't have `preview_mode` set). So in order to fix this bug, I decided to remove all of the calls to `save!` from `lib/collectible.rb`, and move them into the `CollectionItem` callback. (This also required adding a `before_remove` callback to ensure that the collection item's destroy callbacks would fire.)

4. This meant that I could merge some of the duplicate callbacks, so I did.

5. However, it also meant that the `after_commit` callback `notify_of_association` was being called whenever a collection item was destroyed. So I changed it into an `after_create_commit` callback (which I think matches the desired behavior: https://github.com/otwcode/otwarchive/commit/5876949be309058f79c01aa2399f6a572faf1190).

## Testing Instructions

This one is very difficult to reproduce, so pretty much all QA can do is put the CollectionItem code through its paces to make sure that these changes didn't break any of the existing behavior.

## References

One of the changes that I made was actually specifically designed to fix this issue:
https://otwarchive.atlassian.net/browse/AO3-5206

I also wrote tests that are suitable for this issue:
https://otwarchive.atlassian.net/browse/AO3-5569

And for this issue, which I think has been fixed and should be tests only at this point:
https://otwarchive.atlassian.net/browse/AO3-4613


features/collections/collection_anonymity.feature
```diff
@@ -375,3 +375,81 @@ Feature: Collection
...
+
+    When I view the work "Secret Work"
+    Then I should see "This work is part of an ongoing challenge and will be revealed soon!"
+
+    When I am logged in as the owner of "Hidden Collection"
+      And I go to "Hidden Collection" collection edit page
+      And I follow "Delete Collection"
+      And I press "Yes, Delete Collection"
+      And I view the work "Secret Work"
+    Then I should not see "This work is part of an ongoing challenge and will be revealed soon!"
```
redsummernight: We could use `Then I should see "Secret Work"` so the check will work even if we change the unrevealed work notice.

tickinginstant: I'll have to add an `And I am logged out` step for that, since in the tests I wrote the viewer is either the work's creator or the collection's owner, both of whom can always see the work's title even when it's hidden. But that's a good idea, thanks!

PR #3482, opened 2018-12-06
AO3-2839 Handle lost cookies for users

sarken:
## Issue

## Purpose

Sign users out and redirect if they've lost their user_credentials cookie to prevent logged out users from seeing cached logged in pages. (We'll need this for admins, too, but that's a known issue, whereas this is something that was previously fixed.)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -80,17 +80,20 @@ class ApplicationController < ActionController::Base
     end
   end
 
-  # The user_credentials cookie is used by nginx to figure out whether or not
-  # to cache the page, so we want to make sure that it's set when the user is
-  # logged in, and cleared when the user is logged out.
-  before_action :ensure_user_credentials
-  def ensure_user_credentials
-    if logged_in?
-      cookies[:user_credentials] = 1 unless cookies[:user_credentials]
-    else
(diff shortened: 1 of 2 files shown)


tickinginstant:
I'm not sure, but I think the callback order here means that the code to fix cookies could prevent the forced logout code from firing when it needs to. Maybe it'd work to swap the order that the `before_action` callbacks are set? Or use `prepend_before_action`/`prepend: true`? Or bail out of the cookie-fixing code unless `controller_name == "sessions"`, so that the two functions can't fire on the same page?

sarken:
Reordering seems to be enough to make it work. If I delete the cookie, I am redirected to the lost cookie page, whereas before the cookie was just magically restored.

tickinginstant:
[It looks like it might work to change to `after_action` for the `ensure_*_credentials` functions.](https://travis-ci.org/tickinginstant/otwarchive/builds/465147534) (Which would also have fixed the earlier ordering issue, too ... I wish I'd thought to suggest it then!)

redsummernight:
Thanks @tickinginstant! I'm gonna add your commit https://github.com/tickinginstant/otwarchive/commit/4fd6c9cfa8588c6c2a73c25da7a0cc5b451755fa to this pull request in 30m.

PR #3485, opened 2018-12-16
AO3-2839 Keep invitation token when sign-up form is submitted

redsummernight:
## Issue

## Purpose

- Sign up with an invitation token, mess up on the form, get errors
- Fix the errors, click "Create Account"
- Get redirected to /invite_requests, no accounts were made, since the invitation token got lost when the form errors. This pull fixes that.

## Testing Instructions

See issue.

### DIFF
--- a/app/controllers/users/registrations_controller.rb
+++ b/app/controllers/users/registrations_controller.rb
@@ -4,7 +4,11 @@ class Users::RegistrationsController < Devise::RegistrationsController
 
   def new
     super do |resource|
-      set_invitation(resource, true)
+      if params[:invitation_token]
+        @invitation = Invitation.find_by(token: params[:invitation_token])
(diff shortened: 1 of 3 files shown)


app/controllers/users/registrations_controller.rb
```diff
@@ -24,6 +20,7 @@ def create
       resource.transaction do
         # skip sending the Devise confirmation notification
         resource.skip_confirmation_notification!
+        set_invitation(resource)
```
tickinginstant: Using `set_invitation` here seems like overkill, since the User already loads the invitation token when saving. Is there a reason you prefer it over the pre-Devise way of setting it?
https://github.com/otwcode/otwarchive/blob/317ebf14ab1234cd9861e8f12214c593a0e44dfe/app/controllers/users_controller.rb#L145

PR #3486, opened 2018-12-16
AO3-5584 CollectionItemsController permissions.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5584

## Purpose

This PR modifies `CollectionItemsController#update_multiple` in several ways:
1. It checks whether the user has permission to perform the desired action.
2. It checks whether the IDs submitted to be updated are from assignments that belong to the correct user/collection, and silently excludes the ID if it doesn't belong.
3. It restricts the allowed parameters so that collection owners can't update `user_approval_status`, and the work's owner can't update `collection_approval_status`, `unrevealed`, or `anonymous`. (And neither one can update `collection_id`.)

## Testing Instructions

Available on Slack, or the Wiki.

### DIFF
--- a/spec/controllers/collection_items_controller_spec.rb
+++ b/spec/controllers/collection_items_controller_spec.rb
@@ -170,11 +170,11 @@ describe CollectionItemsController do
         before { fake_login_known_user(work_owner) }
 
         context "setting user_approval_status" do
-          let(:attributes) { { user_approval_status: -1 } }
+          let(:attributes) { { user_approval_status: CollectionItem::REJECTED } }
 
           it "updates the collection item and redirects" do
             patch :update_multiple, params: params
-            expect(item.reload.user_approval_status).to eq(-1)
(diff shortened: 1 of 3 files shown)


spec/controllers/collection_items_controller_spec.rb
```diff
@@ -119,4 +119,211 @@
...
+
+    describe "on the collection items page for a different collection" do
+      let(:other_collection) { create(:collection) }
+      before { fake_login_known_user(other_collection.owners.first.user) }
+
+      let(:params) do
+        {
+          collection_id: other_collection.name,
+          collection_items: {
+            item.id => { collection_approval_status: -1 }
```
redsummernight: Should we use `CollectionItem::REJECTED` instead of -1?

tickinginstant: Good idea, thanks! Will do.

PR #3489, opened 2018-12-22
AO3-5576 Use Devise::Recoverable.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5576
https://otwarchive.atlassian.net/browse/AO3-5577

## Purpose

Adds Recoverable to the list of Devise modules for Users, and tries to move over the existing messages/emails to the Devise system.

## Testing Instructions

Reset password, follow instructions, make sure that the user experience is what you expect (error messages, good instructions and phrasing, and so on).


app/views/users/passwords/new.html.erb
```diff
@@ -0,0 +1,18 @@
+<!--Descriptive page name, messages and instructions-->
+<h3 class="heading"><%= ts("Forgotten your password?") %></h3>
+<p>
+  <%= ts("If you've forgotten your password, we can send instructions that will allow you to reset it.") %>
```
redsummernight: There's another instance of this text in the form for changing user names that should be updated to match:

https://github.com/otwcode/otwarchive/blob/5cf864c5d48637c8bea9dc9223ebceb086ca44ff/app/views/users/change_username.html.erb#L44-L52

Maybe we should link to the new password reset page instead of having a mini-form on the change username page?

config/initializers/devise.rb
```diff
@@ -190,12 +190,12 @@
...
   #
   # Defines which key will be used when recovering the password for an account
-  # config.reset_password_keys = [:email]
+  config.reset_password_keys = [:login]
 
   # Time interval you can reset your password with a reset password key.
   # Don't put a too small interval or your users won't have the time to
   # change their passwords.
-  config.reset_password_within = 48.hours
+  config.reset_password_within = 7.days
```
redsummernight: We should mention in the comment that the duration is currently hard-coded in the password reset email templates.

Let's also move this to ArchiveConfig.

app/views/users/passwords/new.html.erb
tickinginstant: Whoops, sorry! I missed that form. (The dangers of unpolished code!)

Is that block of code reachable? I'm having trouble getting it to display.

redsummernight:
For my own notes, we now have 2 forms to change passwords: one linked from preferences, requiring the old password, the other one from Devise, reachable by email, requiring an auto-generated token.

PR #3493, opened 2018-12-24
AO3-5587 Add unique index to user emails

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5587

## Purpose

Enforce email uniqueness at database level, on just at Rails level.

## Testing Instructions

The migration should run OK.


sarken:
I just realized this won't work on staging. We've set all users' email addresses to the same placeholder to avoid any issues with leaking personal info or accidentally sending emails.

PR #3501, opened 2019-01-10
AO3-4949 Reveal and de-anon all collection items regardless of approval status

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4949

## Purpose

Previously, revealing a collection only revealed CollectionItems that had been approved by both the collection moderator and work creator. This was bad: for example, if a moderator rejected a work from the collection before revealing, that work could stay unrevealed until the creator noticed and removed it from the collection. (For bonus fun, revealing a collection takes away the "Unrevealed" checkbox on the Manage Items pages, so even if the moderator noticed the problem, they would have to know to set the collection _back_ to unrevealed so they could update the rejected work's status.) 

Now revealing a collection sets _all_ of its CollectionItems' `unrevealed` columns to false. 

This also adjusts which works gift, prompt fill, and related work notifications are sent for when a collection is revealed. 

Previously, revealing a collection would send notifications for approved items -- that made a certain amount of sense, given those were the only works being revealed. Now that we are revealing _all_ works, however, we should only send notifications for works that were _actually_ unrevealed: that means works where the CollectionItem's `user_approval_status` was approved. (If a user adds a work to an unrevealed moderated collection, the work will be unrevealed regardless of its `collection_approval_status`. A collection moderator cannot make a work unrevealed unless the user has approved its inclusion in the collection.)

I have also added a check to the prompt fill and gift notification mailers that will make sure the collection name is _not_ included in these notifications _unless_ both the user and the moderator have approved the CollectionItem. We don't want the gift notification for a work that has been rejected from a collection to say "A gift work has been posted for you in the Awesome Works collection" or whatever.

There are a couple of issues that might come up in testing that I want to note, but they are existing bugs:
- If a work is in both Regular Collection and Unrevealed Collection, the notification emails that are sent when Unrevealed Collection is revealed _might_ say "A gift work has been posted for you in the Regular Collection collection." That's because the emails only list the work's first collection.
- If the moderator reveals the collection, sets it back to unrevealed, and then reveals it again, a second set of notifications will be sent.


app/mailers/user_mailer.rb
```diff
@@ -269,10 +269,11 @@ def related_work_notification(user_id, related_work_id)
   end
 
   # Emails a recipient to say that a gift has been posted for them
-  def recipient_notification(user_id, work_id, collection_id=nil)
+  def recipient_notification(user_id, work_id, collection_id = nil)
     @user = User.find(user_id)
     @work = Work.find(work_id)
-    @collection = Collection.find(collection_id) if collection_id
+    # If we've supplied a collection_id, make sure the collection mods and work creators have approved the work's inclusion in the collection before adding the collection name to the email
+    @collection = Collection.find(collection_id) if collection_id && @work.collection_items.where(collection_id: collection_id, collection_approval_status: CollectionItem::APPROVED, user_approval_status: CollectionItem::APPROVED).present?
```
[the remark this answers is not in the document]
redsummernight: > there is a minor performance issue with the definition of `approved_collections`

Do we have a ticket for that yet?

app/models/collection.rb
```diff
@@ -459,20 +460,34 @@ def reveal_authors!
...
-  end
-
+    collection_items.each { |collection_item| collection_item.update_attribute(:anonymous, false) }
+  end
+
+  # A work is not marked unrevealed until the user has approved its inclusion
+  # in a collection, so we don't want to send notifications unless 
+  # user_approval_status is approved.
+  #
+  # TODO: When AO3-2240 is fixed, we should ideally only send notifictations
```
redsummernight: "notifications"

app/models/collection.rb
```diff
@@ -459,20 +460,34 @@ def reveal_authors!
...
+  # A work is not marked unrevealed until the user has approved its inclusion
+  # in a collection, so we don't want to send notifications unless 
+  # user_approval_status is approved.
+  #
+  # TODO: When AO3-2240 is fixed, we should ideally only send notifictations 
+  # for items that had their unrevealed status changed from true to false when 
+  # the collection's unrevealed setting was changed. With the current code,
+  # it's possible to send notifications twice for the same work if the 
+  # moderator reveals the collection, sets it back to unrevealed (which does 
+  # not and should not set exisiting collection items back to unrevaeld), and
```
redsummernight: "existing", "unrevealed"

app/mailers/user_mailer.rb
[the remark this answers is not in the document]
tickinginstant: No. I originally ran into the issue when I was working on the n+1 work code (since it goes from performance issue to actual bug when you try to call `Work.includes(:approved_collections)`), so I included a fix for it in the PR that I submitted for that, but then I had to close the PR because it was incompatible with the search upgrade.

It's probably worth a separate ticket, because my WIP n+1 work code has become a lot more complex since then, and the fix for `approved_collections` is only a few lines. Easier to merge in smaller increments.

features/collections/collection_reveals_notifications.feature
```diff
@@ -0,0 +1,296 @@
+@collections @promptmemes @giftexchanges @challenges @works @gifts
+
+Feature: Notification emails for newly revealed collections
+  When a moderator reveals or de-anons an entire collection, the relevant notification emails should be sent.
+
+  #########################################
+  #
+  # WORKS POSTED DIRECTLY TO THE COLLECTION
```
redsummernight: For works posted to a moderated+unrevealed collection, tests cover: gift/related/prompt fill x unapproved/approved/rejected = 9. Very thorough!

features/collections/collection_reveals_notifications.feature
```diff
@@ -0,0 +1,296 @@
...
+  Given the user "recip" exists and is activated
+    And the user "creator" exists and is activated
+    And "subscriber" subscribes to author "creator"
+    And I have the collection "Future Anon Unrevealed Collection"
+    And I am logged in as "inspiration"
+    And I post the work "Inspirational Work"
+    And I am logged in as "creator"
+  When I set up the draft "Invited Work" as a gift for "recip"
+    And I list the work "Inspirational Work" as inspiration
+    # HACK: AO3-2373 means related work emails don't always go out when posting without preview
```
redsummernight: Can you mark it TODO?

features/collections/collection_reveals_notifications.feature
sarken: I did notice some duplication between these tests and the ones in collection_anonymity.feature, but I'm going to fix that in the PR for AO3-2240, which should be coming up later tonight.

redsummernight: Do we want this one merged first? So you don't have to fight cascading conflicts.

app/models/collection.rb
```diff
@@ -459,20 +460,34 @@ def reveal_authors!
...
+  # A work is not marked unrevealed until the user has approved its inclusion
+  # in a collection, so we don't want to send notifications unless 
+  # user_approval_status is approved.
+  #
+  # TODO: When AO3-2240 is fixed, we should ideally only send notifications for
+  # items that had their unrevealed status changed from true to false when the 
+  # collection's unrevealed setting was changed. With the current code, it's
+  # possible to send notifications twice for the same work if the moderator
+  # reveals the collection, sets it back to unrevealed (which does not and
+  # should not set existing collection items back to unreveald), and then
```
redsummernight: "unrevealed"

app/mailers/user_mailer.rb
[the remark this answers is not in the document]
redsummernight: [AO3-5594](https://otwarchive.atlassian.net/browse/AO3-5594) is all yours.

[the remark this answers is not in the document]
redsummernight: Dang I didn't see AO3-5593, I'll close mine as duplicate.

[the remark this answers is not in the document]
redsummernight: No worries! Yours had a better write-up.

features/collections/collection_reveals_notifications.feature
sarken: That would be great, yeah.

app/models/collection.rb
sarken: I'll get this one in the next PR.

sarken: And when I say I'll get it, I mean the fix for AO3-2240 is to remove this method entirely and delete a couple of lines from `notify_of_status_change` in the collection item model.

sarken: Just noting for myself that's not PRed because it'll need to be updated once this is merged. It's here, though: https://github.com/otwcode/otwarchive/compare/master...sarken:AO3-2240_final?expand=1

sarken:
No merge conflicts, but I need to update the tests to work with the new text on the posting buttons.

app/models/collection.rb
tickinginstant: I'm not sure where to put this comment, because the PR hasn't been submitted yet, but since this and AO3-2240 seem to be treated as two halves of a whole, I figured I'd mention it here:

In the code for AO3-2240 that you've linked to, the recipient notification email will probably be very susceptible to stale data issues. Because `UserMailer#recipient_notification` is checking whether the work is unrevealed, and the recipient notification email is queued up right after the work is changed from unrevealed to revealed, any kind of delay in reads will cause the "A Gift Work For You" email to go missing.

sarken: Hmm, that's what I was trying to avoid by having it check the collection item rather than the work. How else might I get around it?

tickinginstant: The collection item and the work are updated in the same transaction, so checking one vs. the other won't make a difference in production. (It does make a difference in the test environment, because delayed jobs are performed synchronously,`notify_of_status_change` is an `after_update` hook, and `update_work` is an `after_save` hook. But it won't help with stale data issues.)

I'm not really sure what the best approach is.

You could make `notify_of_status_change` an `after_update_commit` hook, which would probably handle most of the intermittent email failures. But I'm not sure it'd fix them all. Additionally, `after_commit` hooks with `saved_change_to_attribute?` checks make me nervous, because if an object is saved multiple times in the same transaction, you may not get consistent results.

You could also try to move the check out of `UserMailer#recipient_notification` and into one of the functions that calls it (as long as it's within an `after_update` or `after_save` hook). Though that could run into problems if the commit is rolled back after the notification is enqueued, because then recipients would be notified about gifts that they can't see. Maybe you could use the `after_commit_everywhere` gem for that?

Alternatively, you could try to add a significant delay to the emails, like subscription emails. The subscription code can afford to check whether the works are anonymous because subscription emails are only sent once an hour.

(Also, I noticed another potential issue: It looks like it might send notifications even if the `CollectionItem` isn't `approved_by_user`.)

sarken: Opened a PR here for discussion: https://github.com/otwcode/otwarchive/pull/3672

Would the delay need to be an hour? I think a five or ten minute delay on all gift notifications would be fine, even beneficial -- it would give the creator time to correct or remove the recipient info -- but an hour seems less than ideal.

>Also, I noticed another potential issue: It looks like it might send notifications even if the CollectionItem isn't approved_by_user.

The check in the mailer should handle that, I think? (The code is most of a year old so I'm a little fuzzy on it.) `approved_collection_items` should be checking both collection and user approval status.

tickinginstant: A five or ten minute delay would probably be fine.

> The check in the mailer should handle that, I think? (The code is most of a year old so I'm a little fuzzy on it.) `approved_collection_items` should be checking both collection and user approval status.

The check in the mailer ensures that you'll only get notifications if the work is visible. But if someone has a work that's already visible, with an unrevealed CollectionItem that's not `approved_by_user`, the recipient will get a notification when that CollectionItem is revealed, even though they (probably) already received a notification when the work first became visible.

sarken:
Closing this so I can pull request a fix for something blocking it.

sarken:
Never mind, closed a different one.

PR #3504, opened 2019-01-12
AO3-5493 Allow searching for tags used only in drafts

redsummernight:
## Issue

## Purpose

This just sets up indexing/querying, without actually using it in the wrangling UI yet.

Move `TAGGINGS_COUNT_MIN_CACHE_COUNT` to config.yml, so we don't need to specify a default every time. Use this property for the tagging reindex limit, instead of hard-coding 100.

Tags are reindexed to get updated draft-only status in two ways:

- When taggings are created/destroyed; this currently only applies to characters, relationships, and freeforms, so we need to add fandoms.
- When works become posted; we need to add a reindex callback.

## Testing Instructions

None(?), we have no UI for this yet.

### DIFF
--- a/app/models/tagging.rb
+++ b/app/models/tagging.rb
@@ -48,9 +48,10 @@ class Tagging < ApplicationRecord
   def update_search
     return unless tagger && Tag::USER_DEFINED.include?(tagger.type)
 
-    # Reindex the tag for updated suggested tags, which give you an idea of
-    # how unwrangled tags are used, but only if it has less than a number of uses.
-    # For comprehensive data on really popular tags, we still have work search.
+    # Reindex the tag for updated suggested tags.
+    # Suggested tags help wranglers figure out where to wrangle new tags
+    # and if it's necessary to disambiguate existing canonical/unfilterable tags
+    # in multiple fandoms.
     tagger.enqueue_to_index if tagger.taggings_count < ArchiveConfig.TAGGINGS_COUNT_REINDEX_LIMIT
   end
 end
--- a/app/models/tagging.rb
+++ b/app/models/tagging.rb
@@ -46,11 +46,11 @@ class Tagging < ApplicationRecord
   end
 
   def update_search
-    # When a tag's count is more than this threshold, we start caching it,
-    # so taggings_count becomes less accurate. That's our cue to stop
-    # eagerly reindexing counts as well.
     return unless Tag::USER_DEFINED.include?(tagger.type)
-    reindex_boundary = ArchiveConfig.TAGGINGS_COUNT_MIN_CACHE_COUNT
-    tagger.enqueue_to_index if tagger.taggings_count < reindex_boundary
+
+    # Reindex the tag for updated suggested tags, which give you an idea of
+    # how unwrangled tags are used, but only if it has less than a number of uses.
+    # For comprehensive data on really popular tags, we still have work search.
+    tagger.enqueue_to_index if tagger.taggings_count < ArchiveConfig.TAGGINGS_COUNT_REINDEX_LIMIT
   end
 end
--- a/app/models/tagging.rb
+++ b/app/models/tagging.rb
@@ -46,9 +46,11 @@ class Tagging < ApplicationRecord
   end
 
   def update_search
-    reindex_boundary = 100
-    if %w(Character Relationship Freeform).include?(tagger.type)
-      tagger.enqueue_to_index if tagger.taggings_count < reindex_boundary
-    end
+    # When a tag's count is more than this threshold, we start caching it,
+    # so taggings_count becomes less accurate. That's our cue to stop
+    # eagerly reindexing counts as well.
+    return unless %w[Fandom Character Relationship Freeform].include?(tagger.type)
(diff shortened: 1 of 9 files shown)


app/models/tagging.rb
```diff
@@ -46,9 +46,11 @@ def update_taggings_count
...
 
   def update_search
-    reindex_boundary = 100
-    if %w(Character Relationship Freeform).include?(tagger.type)
-      tagger.enqueue_to_index if tagger.taggings_count < reindex_boundary
-    end
+    # When a tag's count is more than this threshold, we start caching it,
+    # so taggings_count becomes less accurate. That's our cue to stop
+    # eagerly reindexing counts as well.
+    return unless %w[Fandom Character Relationship Freeform].include?(tagger.type)
```
tickinginstant: Could use Tag::USER_DEFINED here, now that the check isn't omitting Fandom anymore.

app/models/tagging.rb
```diff
@@ -46,9 +46,11 @@ def update_taggings_count
...
+    reindex_boundary = ArchiveConfig.TAGGINGS_COUNT_MIN_CACHE_COUNT
```
tickinginstant: Should this use a different config variable than the taggings count code? It seems like it might be useful to have them separate so that one can be adjusted without the other.

redsummernight: Yeah, that makes sense. I added a new config variable, and the comment should hopefully make it clear that the limit only applies to tagging changes. Changes to the tag itself should reindex no matter what.

app/models/tagging.rb
```diff
@@ -46,9 +46,11 @@ def update_taggings_count
   end
 
   def update_search
-    reindex_boundary = 100
-    if %w(Character Relationship Freeform).include?(tagger.type)
-      tagger.enqueue_to_index if tagger.taggings_count < reindex_boundary
-    end
+    return unless tagger && Tag::USER_DEFINED.include?(tagger.type)
+
+    # Reindex the tag for updated suggested tags, which give you an idea of
```
redsummernight: Comment adapted from https://github.com/otwcode/otwarchive/pull/3395/files#r204898786.

config/config.yml
```diff
@@ -167,6 +167,12 @@ TAGS_PER_SEARCH_PAGE: 50
 # When updating tag counts, how many to do in one transaction
 TAG_UPDATE_BATCH_SIZE: 100
 
+# We only start caching tag counts for tags used more than a certain number of times
+TAGGINGS_COUNT_MIN_CACHE_COUNT: 1000
+
+# For tagging changes, we only reindex tags used less than a certain number of times
+TAGGINGS_COUNT_REINDEX_LIMIT: 1000
```
tickinginstant: I think this default can be a fair amount lower, since `suggested_parent_tags` only checks the first 50 works by default, and I suspect that there aren't all that many tags with 50+ drafts that don't have any posted works. But it's just a default so it's easy enough to change later, too.

app/models/tagging.rb
sarken: This comment (and the one it was adapted from) might be slightly misleading. From Tag Wrangling chairs:

> but sometimes when reviewing things that have a lot of use, people look over the suggested fandoms
> because they'll still appear in addition to the attached fandom for canonical/unfilterable tags whether a tag used in multiple fandoms needs to be disambiguated.

redsummernight: How's this?
> Reindex the tag for updated suggested tags.
> Suggested tags help wranglers figure out new tag and disambiguate existing canonical/unfilterable tags in multiple fandoms.

tickinginstant: Hmm, that comment from the Tag Wrangling chairs makes it sound like the change to suggested fandoms is going to be a problem. (`suggested_parent_tags`, which is used to generate `@suggested_fandoms`, only looks at tags on the first 50 works in the tag. I think it picks them in increasing order by ID, so suggested fandoms wouldn't give any idea of what the recent usage looks like, and whether a pre-existing canonical character has the same name as a new character in another fandom.)

sarken: I think that's a good comment, @redsummernight 

@tickinginstant, it sounds like the suggested fandom code will be okay: one of the chairs has now indicated wranglers are actually _not_ supposed to use suggested tags to check for ambiguity, but rather the works page -- most tags with more than one suggested fandom come from crossovers, not ambiguous usage.

sarken: @redsummernight Oops, actually, I think a couple words went missing in your potential comment. Something like "figure out where to wrangle new tags and if it's necessary to disambiguate..."

redsummernight: Previously:

    has_many :suggested_fandoms, -> { distinct }, through: :works, source: :fandoms

So it can be slow.

I'm wondering if we can use `WorkQuery` results/facets for the new `suggested_parent_tags` instead of that first-50 SQL query.

redsummernight: Never mind, the list of suggested fandoms doesn't need to be exhaustive.

@sarken I edited the previous comment.

tickinginstant: Yeah, if the 50-work limit wasn't okay, I think the only reasonable approach would have been to go through Elasticsearch and aggregate the fandom tags. I'm relieved that that's unnecessary!

PR #3509, opened 2019-01-18
AO3-5493 Fixes for the Elasticsearch-ification of unwrangled tags

sarken:
## Issue

## Purpose

* Fix alphabetical sorting of unwrangled tags
* Make wrangled tags appear in tag search results again
* Reindex a tag's suggested (and actual) children when the tag's merger_id changes
** Put it in a background queue since it could be 100s or 1000s of tags to reindex; Tag Wrangling says they'd be fine even if this took hours
* Show unwrangled relationships for character tags again

## Testing Instructions

Let the wranglers have at it


app/models/indexing/scheduled_reindex_job.rb
```diff
@@ -5,7 +5,7 @@ def self.perform(reindex_type)
               when 'main'
                 %w(Pseud Tag Work Bookmark Series ExternalWork)
               when 'background'
-                %w(Work Bookmark Pseud Series ExternalWork)
```
redsummernight: Do we ever want to have a class in one but not both of "main" and "background", or can we do `when "main", "background"`?

app/models/tag.rb
```diff
@@ -1238,7 +1238,7 @@ def unwrangled_query(tag_type, options = {})
       type: tag_type,
       unwrangleable: false,
       wrangled: false,
-      pre_fandom_ids: [self.id]
+      "pre_#{self.type.downcase}_ids": [self.id]
     ))
```
redsummernight: We should add `per_page: Tag.per_page` so the limit matches the SQL queries for the other bins.

features/other_a/preferences_edit.feature
```diff
@@ -374,3 +374,43 @@ Feature: Edit preferences
     And I should not see "Scary tag"
     And I should not see "Scarier"
     And I should see "Show additional tags"
+
+  @javascript
```
redsummernight: Are these 2 tests just moved over from tag_wrangling_more.feature without changes?

spec/models/search/tag_query_spec.rb
```diff
@@ -156,4 +156,15 @@
     results = tag_query.search_results
     expect(results.size).to eq 5
   end
+
+  before do
+    @wrangled_tag = create(:freeform, name: "wrangled ff")
```
redsummernight: We should use let to avoid sharing state between tests in instance variables.

app/models/indexing/scheduled_reindex_job.rb
sarken: Not that I can find, so I'll do that.

spec/models/search/tag_query_spec.rb
```diff
@@ -158,13 +160,12 @@
   end
 
   before do
-    @wrangled_tag = create(:freeform, name: "wrangled ff")
-    create(:common_tagging, common_tag_id: @wrangled_tag.id)
+    create(:common_tagging, common_tag_id: wrangled_tag.id)
     update_and_refresh_indexes('tag', 1)
```
redsummernight: `update_and_refresh_indexes` is only necessary if we're updating the shard count, for most tests it's overkill. But I can move things around later in [AO3-5581](https://otwarchive.atlassian.net/browse/AO3-5581).

app/models/tag.rb
```diff
@@ -1276,6 +1277,23 @@ def suggested_parent_ids(parent_type)
...
+      "pre_#{self.type.downcase}_ids": [self.id]
+    ))
+  end
+
+  def suggested_child_tags(options = {})
+    suggested_child_tags_query(options).search_results
+  end
+
+  def suggested_child_ids
+    suggested_child_tags.pluck(:id)
```
tickinginstant: It's probably faster not to go through the database for this, because ES will return the IDs already. I think you might be able to extract them with some code like
```
suggested_child_tags_query(options).search.dig("hits", "hits").map { |hit| hit["id"] }
```
But it can also be left as-is if you don't mind having this step load the tags. (Note that it loads the complete tags, not just the IDs -- `pluck` is basically just acting like `map` here because `load_from_elasticsearch` converts the results to an array.)

spec/models/search/tag_query_spec.rb
```diff
@@ -156,4 +158,14 @@
     results = tag_query.search_results
     expect(results.size).to eq 5
   end
+
+  before do
```
tickinginstant: FYI, this `before` isn't scoped, so it'll be called before all of the tests in this file, not just the single test that needs it.

app/models/tag.rb
```diff
@@ -1311,6 +1328,7 @@ def after_update
       if tag.merger_id.present?
         tag.merger.update_works_index_timestamp!
       end
+      IndexQueue.enqueue_ids(Tag, tag.suggested_child_ids, :background)
```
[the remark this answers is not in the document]
sarken: Damn. We do need to reindex all of the suggested child tags. We're trying to make it so even if I de-syn a big tag like Hockey RPF from the less frequently used canonical tag Men's Hockey RPF, all of the tags that had Hockey RPF as suggested parent will leave the unwrangled bins for Men's Hockey RPF and appear in the bins for Hockey RPF instead.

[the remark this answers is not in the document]
sarken: I stuck this in a `queue_child_tags_for_reindex` method and then made it so the `after_update` calls `async(:queue_child_tags_for_reindex)` but after I did that, I realized I wasn't 100% sure that was what you meant. Do I need to make `suggested_child_ids` itself asynchronous? 

To get around the first page issue, I set the `per_page` on that query to `ArchiveConfig.MAX_SEARCH_RESULTS` results, but now I'm wondering if we should be going through the database instead to avoid having a limit. otoh, if there's ever a tag with over 100,000 suggested children, we probably don't want to be getting that from the database, either.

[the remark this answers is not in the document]
tickinginstant: Yes, that's what I meant in terms of making the computation asynchronous. Just calling `async(:queue_child_tags_for_reindex)` is fine.

I'm not sure whether it's reasonable to try to make ES return 100,000 records in one go, even if it is in a Resque job. It might be good to try a [scrolled search](https://www.elastic.co/guide/en/elasticsearch/reference/6.5/search-request-scroll.html), since it gets around the search results limit and loads results page by page:
```
query = suggested_child_tags_query

parameters = {
  index: query.index_name,
  type: query.document_type,
  body: { query: query.filtered_query, sort: "_doc", size: 1000, _source: false },
  scroll: "1m"
}

response = $elasticsearch.search(parameters)
while response.dig("hits", "hits").present?
  tag_ids = response.dig("hits", "hits").map { |hit| hit["_id"] }
  IndexQueue.enqueue_ids(Tag, tag_ids, :background)
  response = $elasticsearch.scroll(body: { scroll_id: response["_scroll_id"] }, scroll: "1m")
end

$elasticsearch.clear_scroll(scroll_id: response["_scroll_id"])
```
But this is complex enough that it might be worth rewriting as a general-purpose `scroll` function in the Query class that yields hits a little at a time. I don't like having this quantity of ES-related code outside of an ES-specific class.

Alternatively, if you'd prefer to work with the database instead of ES, you could load batches of works and queue up all the tags for each batch:
```
all_with_child_type = Tag.where(type: child_types & Tag::USER_DEFINED)
works.select(:id).find_in_batches do |batch|
  relevant_taggings = Tagging.where(taggable: batch)
  tag_ids = all_with_child_type.joins(:taggings).merge(relevant_taggings).distinct.pluck(:id)
  IndexQueue.enqueue_ids(Tag, tag_ids, :background)
end
```
I'm not sure which would be better/faster.

app/models/tag.rb
redsummernight: It might be worth adding:
```
# Returns the IDs without loading all items
def ids
    hits.map { |item| item['_id'] }
end
```
to query_result.rb, so we can do:
```
suggested_child_tags_query.search_results.ids
```
and keep the ES-specific structures nicely wrapped.

redsummernight: We're getting child tags from the database now, this section has been removed.

app/models/tag.rb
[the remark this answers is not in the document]
redsummernight: We went with a database query, which is simpler for now.

spec/models/search/tag_query_spec.rb
redsummernight: I can clean this up while on [AO3-5581](https://otwarchive.atlassian.net/browse/AO3-5581).

spec/models/search/tag_query_spec.rb
redsummernight: Fixed in #3512.

spec/models/search/tag_query_spec.rb
redsummernight: Fixed in #3512.

PR #3513, opened 2019-02-02
AO3-3880 Work download tweaks

sarken:
## Issue

## Purpose

* Fix the font size on PDFs by preventing shrinking
* Stop logging to allow large PDFs to be generated
* Adjust margins on EPUB and MOBI chapter content

## Testing Instructions

Look with your special eyes.

### DIFF
--- a/script/codeship_ebook_converters.sh
+++ b/script/codeship_ebook_converters.sh
@@ -2,11 +2,9 @@
 set -e
 
 # PDF
-if ! [ -x "$(command -v wkhtmltopdf)" ]; then
-  wget https://github.com/wkhtmltopdf/wkhtmltopdf/releases/download/0.12.4/wkhtmltox-0.12.4_linux-generic-amd64.tar.xz
-  tar --directory="${HOME}" -xvf wkhtmltox-0.12.4_linux-generic-amd64.tar.xz
-  ln -s "${HOME}/wkhtmltox/bin/wkhtmltopdf" "${HOME}/bin/wkhtmltopdf"
-fi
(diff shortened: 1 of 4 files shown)


redsummernight:
```
Unknown long argument --log-level
Name:
  wkhtmltopdf 0.12.4 (with patched qt)
```
The option `--log-level` is only available for 0.12.5 and later, so we need to update wkhtmltopdf.

Alternatively, we can use `--quiet` (same as `--log-level none`).

script/codeship_ebook_converters.sh
```diff
@@ -2,11 +2,9 @@
 set -e
 
 # PDF
-if ! [ -x "$(command -v wkhtmltopdf)" ]; then
-  wget https://github.com/wkhtmltopdf/wkhtmltopdf/releases/download/0.12.4/wkhtmltox-0.12.4_linux-generic-amd64.tar.xz
-  tar --directory="${HOME}" -xvf wkhtmltox-0.12.4_linux-generic-amd64.tar.xz
-  ln -s "${HOME}/wkhtmltox/bin/wkhtmltopdf" "${HOME}/bin/wkhtmltopdf"
-fi
+wget https://downloads.wkhtmltopdf.org/0.12/0.12.5/wkhtmltox_0.12.5-1.trusty_amd64.deb
+sudo dpkg -i ./wkhtmltox_0.12.5-1.trusty_amd64.deb
```
redsummernight: Codeship seems to be using an older version of apt so I can't do `apt install <file>` like on Travis.

https://unix.stackexchange.com/a/159114

PR #3516, opened 2019-02-07
AO3-3880 Fix download file name transliteration

redsummernight:
## Issue

## Purpose

Fix download file name transliteration, nothing groundbreaking, just putting things back the way they were.

https://github.com/otwcode/otwarchive/blob/5c29fc05e5f56019855970763b50f2efff928f4c/app/models/work.rb#L883-L887

## Testing Instructions

Post works with Russian/Farsi/Chinese/Japanese/Hebrew titles and see what the downloaded file names are.

### DIFF
--- a/app/models/download.rb
+++ b/app/models/download.rb
@@ -64,7 +64,7 @@ class Download
   def file_name
     name = clean(work.title)
     name += " Work #{work.id}" if name.length < 3
-    name
+    name.strip
   end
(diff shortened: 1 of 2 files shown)


redsummernight:
Possibly to reduce dependencies and use Rails instead of `to_ascii` which comes from a gem, but I'm not sure.

PR #3517, opened 2019-02-08
AO3-5603 Revise WorksController to eliminate set_instance_variables.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5603

## Purpose

1. I overhauled the WorksController and the ChaptersController to try to eliminate the function `set_instance_variables`.
    - I moved the code to build a blank item into `#new`, the code to build an item from params into `#create`, and the code to modify an existing item into `#update`.
    - I moved the instance variable assignment into `load_pseuds` (for the ChaptersController) and `set_work_form_fields` (for the WorksController).
    - I did a similar thing for `WorksController#set_instance_variables_tags`.
2. In the process, I cleaned up the before_action filters a little for both controllers.
3. I also refactored several of the functions that I was modifying, to make it easier to follow the flow of control.
    - I moved the check for `params[:cancel_button]` to the very beginning of the function, before assigning to attributes. Because some attribute assignment is written to the database immediately (e.g. tags), I figured that that was contrary to the purpose of the button.
    - To further simplify the WorksController code, I also fixed up the work function `set_challenge_claim_info`. This means that it can be used alongside `set_challenge_info` (which works only for the work's assignments).
    - I also changed `Work#check_for_invalid_tags` into a validation. (It used to be a before_save callback that aborted in case of error, which meant that it would only display an error to the user if `Work#save` was called. Hence the complicated code in the `WorksController#preview_mode` function to try to catch invalid tags on preview.)
4. I added a new function `WorksController#work_tag_params` to make sure that it only allows the tags (and the language) to be changed, and not any of the other fields that are available in `WorksController#work_params`.
5. I also rearranged a few functions in the WorksController to make sure that more of the utility functions were protected, instead of mixed in alongside the public actions.

## Testing

See Slack. In addition, it would be good to check that the refactored code behaves as expected.

### DIFF
--- a/app/controllers/chapters_controller.rb
+++ b/app/controllers/chapters_controller.rb
@@ -114,7 +114,7 @@ class ChaptersController < ApplicationController
     @work.wip_length = params[:chapter][:wip_length]
     load_pseuds
 
-    if params[:edit_button] || chapter_is_invalid?
+    if params[:edit_button] || chapter_cannot_be_saved?
       render :new
     elsif chapter_has_pseuds_to_fix?
       render :_choose_coauthor
@@ -149,7 +149,7 @@ class ChaptersController < ApplicationController
     @work.wip_length = params[:chapter][:wip_length]
     load_pseuds
 
-    if params[:edit_button] || chapter_is_invalid?
+    if params[:edit_button] || chapter_cannot_be_saved?
       render :edit
     elsif chapter_has_pseuds_to_fix?
       render :_choose_coauthor
@@ -243,7 +243,7 @@ class ChaptersController < ApplicationController
 
   # Check whether we should display :new or :edit instead of previewing or
   # saving the user's changes.
-  def chapter_is_invalid?
+  def chapter_cannot_be_saved?
     # make sure at least one of the pseuds is actually owned by this user
     if @chapter.authors.present? && (@chapter.authors & current_user.pseuds).empty?
       flash.now[:error] = ts("You're not allowed to use that pseud.")
--- a/app/controllers/chapters_controller.rb
+++ b/app/controllers/chapters_controller.rb
@@ -1,12 +1,13 @@
 class ChaptersController < ApplicationController
   # only registered users and NOT admin should be able to create new chapters
   before_action :users_only, except: [ :index, :show, :destroy, :confirm_delete ]
-  before_action :load_work, except: [:index, :auto_complete_for_pseud_name, :update_positions]
-  # only authors of a work should be able to edit its chapters
-  before_action :check_ownership, only: [ :new, :create, :edit, :update, :manage, :preview, :destroy, :confirm_delete ]
-  before_action :set_instance_variables, only: [ :new, :create, :edit, :update, :preview, :post, :confirm_delete ]
-  before_action :check_visibility, only: [ :show]
   before_action :check_user_status, only: [:new, :create, :edit, :update]
+  before_action :load_work
+  # only authors of a work should be able to edit its chapters
+  before_action :check_ownership, except: [:index, :show]
+  before_action :check_visibility, only: [:show]
+  before_action :load_chapter, only: [:show, :edit, :update, :preview, :post, :confirm_delete, :destroy]
+  before_action :set_author_attributes, only: [:create, :update]
 
   cache_sweeper :feed_sweeper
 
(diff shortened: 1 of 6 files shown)


app/controllers/chapters_controller.rb
```diff
@@ -105,15 +105,19 @@ def draft_flash_message(work)
   # POST /work/:work_id/chapters
   # POST /work/:work_id/chapters.xml
   def create
-  	@work.wip_length = params[:chapter][:wip_length]
+    if params[:cancel_button]
+      redirect_back_or_default('/')
```
sarken: Could we make this `root_path` instead of `'/'`? Ditto line 144.

app/controllers/chapters_controller.rb
```diff
@@ -236,12 +241,33 @@ def destroy
 
   private
 
+  # Check whether we should display :new or :edit instead of previewing or
+  # saving the user's changes.
+  def chapter_is_invalid?
+    # make sure at least one of the pseuds is actually owned by this user
+    if @chapter.authors.present? && (@chapter.authors & current_user.pseuds).empty?
```
sarken: Would it be worthwhile to break this out into its own method, like `user_can_use_pseud` (which is a terrible name)? I don't feel super strongly about it, and it's definitely not worth it if it's a pain in the neck, but checking the user's ownership of a pseud just feels a little out of place in a method called `chapter_is_invalid?`

app/controllers/works_controller.rb
```diff
@@ -361,107 +347,79 @@ def edit_tags
 
   # PUT /works/1
   def update
-    set_edit_form_fields
+    if params[:cancel_button]
+      return cancel_posting_and_redirect
+    end
 
+    @work.preview_mode = !!(params[:preview_button] || params[:edit_button] ||
```
sarken: The double `!!` is throwing me off here -- could you translate this from code to English for me?

tickinginstant: It's basically a type-conversion to ensure that `@work.preview_mode` is true or false instead of `"Preview"`, `"Edit"`, `"Cancel"`, or `nil`. A single `!` acts as negation and forces the results to be either true or false, so a double-negation cancels out the negation and just converts to a boolean. I'm pretty sure it would work without the double negation -- I only added it because at one point I tried to print out `@work.preview_mode` and found the fact that it held a string confusing.

app/controllers/chapters_controller.rb
tickinginstant: At this point, `@chapter.authors` has been set. If the chapter were to be saved at this point, it would remove the current editor as one of the chapter's creators. So it's sort of invalid in the sense that the chapter shouldn't be saved with its current attributes, even though `@chapter.valid?` would return true.

I could rename the whole function to something else, like `chapter_cannot_be_saved?` or something. Just to differentiate between things that should prevent the chapter from being saved and things that make the chapter invalid.

sarken: Renaming sounds good!

app/controllers/works_controller.rb
```diff
@@ -239,108 +238,95 @@ def navigate
...
+      render :new
+    elsif work_has_pseuds_to_fix?
+      render :_choose_coauthor
+    else
+      @work.posted = @chapter.posted = true if params[:post_button]
+      @work.set_revised_at_by_chapter(@chapter)
+
+      if @work.save
+        # HACK: for empty chapter authors in cucumber series tests
+        @chapter.pseuds = @work.pseuds if @chapter.pseuds.blank?
```
redsummernight: I removed this hack and series.feature still passed. Let's remove it entirely.

app/models/work.rb
```diff
@@ -474,13 +476,18 @@ def set_challenge_info
...
   end
 
+  # If this is fulfilling a challenge claim, add the collection.
+  #
+  # Unlike set_challenge_info, we don't automatically add the prompter as a
+  # recipient, because (a) some prompters are anonymous, so there has to be a
+  # prompter notification (separate from the recipient notification) ensuring
+  # that anonymous prompters are notified, and (b) if the prompter is not
+  # anonymous, they'll receive two notifications with roughly the same info
+  # (gift notification + prompter notification).
```
redsummernight: This method was never called before, so having it now would likely fix [AO3-3556](https://otwarchive.atlassian.net/browse/AO3-3556) as well, we'll see.

----

Notes for me: @sarken said we handle prompter notification at:

https://github.com/otwcode/otwarchive/blob/bd4f323003a7f63d0e708ff334fd666e4760601a/lib/creatable.rb#L100-L109

and also (for unrevealed works specifically):

https://github.com/otwcode/otwarchive/blob/59544920efb52c5ffedb6fb062c7d2388264feac/app/models/collection_item.rb#L310-L313

PR #3522, opened 2019-02-13
AO3-3880 Allow users to download restricted works

sarken:
## Issue

## Purpose

Whoops -- we introduced a bug where, if a logged in user attempted to download a restricted work, they would receive a red flash error that said, "You are already signed in."

## Testing Instructions

Log in and attempt to download a restricted work. It should work.

Also, make sure there is no Download option available on works that are drafts, unrevealed, or hidden by an admin. If you construct a download URL for one such work, you should be given a flash error telling you you cannot download a draft/unrevealed work/work that's been hidden by an admin.

## References

Some discussion:

> bingeling [17:06]
> I think I may have found the line that is causing this
> 
> Sarken [17:07]
> \o/
> 
> bingeling [17:07]
> https://github.com/otwcode/otwarchive/blob/master/app/controllers/application_controller.rb#L428
> 
> Sarken [17:07]
> Sadly, I must go afk
> 
> bingeling [17:08]
> I’m pretty sure our current setup redirects to the `new_user_session_path`, which then tells us that one already exists
> 
> bingeling [17:14]
> We have `before_action :check_visibility, only: :show` in the downloads controller, which calls the code linked above. We can’t remove that, but we could adjust `@check_visibility_of` to maybe lead to a different if-branch in `check_visibility`? (edited) 

And later... 

> Sarken [18:39]
> I will be honest and say I do not have even the slightest idea how to approach this. My best guess is it's going to involve removing `before_action :check_visibility, only: :show` from the downloads controller and replacing it with a whole other approach.
> 
> bingeling [18:45]
> I think you’re right
> Maybe if we’re lucky it’s as simple as replacing it with `before_action :authenticate_user!`
> (which I stole from the readme https://github.com/plataformatec/devise where it says that that adds authentication)

## Credit

Please credit @bingeling, not me!

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -424,7 +424,7 @@ public
   # Make sure the user is allowed to see a specific page
   # includes a special case for restricted works and series, since we want to encourage people to sign up to read them
   def check_visibility
-    if @check_visibility_of.respond_to?(:restricted) && @check_visibility_of.restricted && User.current_user.nil?
+    if @check_visibility_of.respond_to?(:restricted) && @check_visibility_of.restricted && !(current_user || current_admin)
       redirect_to new_user_session_path(restricted: true)
     elsif @check_visibility_of.is_a? Skin
       access_denied unless logged_in_as_admin? || current_user_owns?(@check_visibility_of) || @check_visibility_of.official?
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -424,7 +424,7 @@ public
   # Make sure the user is allowed to see a specific page
   # includes a special case for restricted works and series, since we want to encourage people to sign up to read them
   def check_visibility
-    if @check_visibility_of.respond_to?(:restricted) && @check_visibility_of.restricted && !(current_user || current_admin)
+    if @check_visibility_of.respond_to?(:restricted) && @check_visibility_of.restricted && User.current_user.nil?
       redirect_to new_user_session_path(restricted: true)
     elsif @check_visibility_of.is_a? Skin
       access_denied unless logged_in_as_admin? || current_user_owns?(@check_visibility_of) || @check_visibility_of.official?
--- a/app/controllers/downloads_controller.rb
+++ b/app/controllers/downloads_controller.rb
@@ -2,7 +2,7 @@ class DownloadsController < ApplicationController
 
   skip_before_action :store_location, only: :show
   before_action :load_work, only: :show
-  before_action :check_visibility, only: :show
+  before_action :authenticate_user!, only: :show
   after_action :remove_downloads, only: :show
 
   def show
--- a/app/controllers/downloads_controller.rb
+++ b/app/controllers/downloads_controller.rb
@@ -2,7 +2,7 @@ class DownloadsController < ApplicationController
 
   skip_before_action :store_location, only: :show
   before_action :load_work, only: :show
-  before_action :check_visibility, only: :show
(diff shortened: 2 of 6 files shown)


app/controllers/downloads_controller.rb
```diff
@@ -2,7 +2,7 @@ class DownloadsController < ApplicationController
 
   skip_before_action :store_location, only: :show
   before_action :load_work, only: :show
-  before_action :check_visibility, only: :show
+  before_action :authenticate_user!, only: :show
```
redsummernight: `check_visibility` handles a few more things that `authenticate_user!` doesn't:

- if the work is a draft
- if the work is hidden by admin

This check also seemingly means that you cannot download any work at all unless logged in. Logged out users should be able to get at unrestricted works.

sarken: Okay, back to the other plan of adding another branch to check_visibility, then!

app/controllers/application_controller.rb
```diff
@@ -424,7 +424,7 @@ def check_ownership_or_admin
   # Make sure the user is allowed to see a specific page
   # includes a special case for restricted works and series, since we want to encourage people to sign up to read them
   def check_visibility
-    if @check_visibility_of.respond_to?(:restricted) && @check_visibility_of.restricted && User.current_user.nil?
+    if @check_visibility_of.respond_to?(:restricted) && @check_visibility_of.restricted && !(current_user || current_admin)
```
tickinginstant: `User.current_user` is assigned here:
https://github.com/otwcode/otwarchive/blob/0da96bee3ca1587403ec5b73e8d3c2d192e7b0c6/app/controllers/application_controller.rb#L301-L303
It's equivalent to the code you've replaced it with, as far as I can tell.

nginx is configured to hide the user's cookies for downloads, I believe, so all visitors are being treated as logged out by the DownloadsController. Which means that the only way to make restricted works visible to logged-in users is to make them visible to everyone, which isn't quite compatible with the current definition of check_visibility. (The old code called `check_visibility` too, but never ran into this issue because `@check_visibility_of` was set only after `check_visibility` ran, making it a no-op.)

If you want to keep the `before_action :check_visibility` filter, you could try modifying the behavior when the `controller_name` is "downloads." Though it's probably cleaner to just write a custom visibility check for the DownloadsController.

(The other alternative involves playing around with nginx.)

sarken: It is indeed -- that's actually where I got it from. It seemed like `User.current_user` wasn't set, so I thought maybe I could get around it by checking `current_user` and `current_admin` directly.

I'm going to ping you elsewhere re: the nginx option, if you don't mind.

sarken:
I'm not convinced this will fix it, either, tbh. `visible` also relies on `User.current_user`.

tickinginstant:
Yeah, `@work.visible` probably won't work. Ditto anything that checks whether the user is logged in, or who they're logged in as (since the information needed for that check is stored in the session, and the session is stored in cookies, which are inaccessible to the DownloadsController on staging and production).

It's probably better just to manually check `posted` and `hidden_by_admin` and bail out if either has the wrong value, since that's really the only information available. It's kind of barebones, but without cookies, there's not much else that can be done.

app/controllers/downloads_controller.rb
```diff
@@ -45,14 +45,24 @@ def load_work
       redirect_to work_path(@work)
       return
```
redsummernight: Should we move this to `check_download_visibility` too?

PR #3528, opened 2019-02-24
AO3-5630 Always delete /tmp files after download generation

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5630

## Purpose

Use [around_action with begin/ensure](https://guides.rubyonrails.org/action_controller_overview.html#after-filters-and-around-filters) to run cleanup even when the action has errors.

Use Download.remove because if download generation fails, `@download` will be nil.

## Testing Instructions

Confirm downloads still work on staging.

Locally, I tested this by removing the zip command (`sudo apt remove zip`), forcing epub generation to crash every time, and checking that the `/tmp/<work ID>` directory gets created then deleted properly.

### DIFF
--- a/app/controllers/downloads_controller.rb
+++ b/app/controllers/downloads_controller.rb
@@ -4,7 +4,7 @@ class DownloadsController < ApplicationController
   before_action :load_work, only: :show
   before_action :check_download_posted_status, only: :show
   before_action :check_download_visibility, only: :show
-  after_action :remove_downloads, only: :show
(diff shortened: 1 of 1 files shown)


sarken:
Just to make absolutely sure, you also tested that the opposite was true, i.e. with the old code, /tmp/directory_name wasn't getting destroyed when the zip command was removed?

redsummernight:
Yeah, the old code leaves the `/tmp/<work ID>` directory around if the generation code errors.

PR #3532, opened 2019-02-28
AO3-5631 Make /tmp download directory unique per generation

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5631

## Purpose

Add the file type and current timestamp to the directory name. This means each generation has to stay with the same Download instance for proper cleanup, and the Download class methods become obsolete.

Having timestamped directory names means we need to freeze time so the tests can find the generated files.

## Testing Instructions

Confirm downloads still work on staging.

Locally, I repeated the testing I did for #3528 as well.

### DIFF
--- a/app/controllers/downloads_controller.rb
+++ b/app/controllers/downloads_controller.rb
@@ -48,7 +48,7 @@ protected
   def remove_downloads
     yield
   ensure
-    @download.remove unless Rails.env.test?
+    @download.remove
   end
 
   # We can't use check_visibility because this controller doesn't have access to
--- a/app/controllers/downloads_controller.rb
+++ b/app/controllers/downloads_controller.rb
@@ -8,10 +8,11 @@ class DownloadsController < ApplicationController
 
   def show
     respond_to :html, :pdf, :mobi, :epub, :azw3
-    @download = Download.generate(@work, mime_type: request.format)
+    @download = Download.new(@work, mime_type: request.format)
+    @download.generate
 
     # Make sure we were able to generate the download.
-    unless @download.present? && @download.exists?
+    unless @download.exists?
       flash[:error] = ts("We were not able to render this work. Please try again in a little while or try another format.")
       redirect_to work_path(@work)
       return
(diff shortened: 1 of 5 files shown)


app/models/download.rb
```diff
@@ -1,22 +1,13 @@
...
-  attr_reader :work, :file_type, :mime_type
+  attr_reader :work, :file_type, :mime_type, :timestamp
 
   def initialize(work, options = {})
     @work = work
     @file_type = set_file_type(options.slice(:mime_type, :format))
     # TODO: Our current version of the mime-types gem doesn't include azw3, but
     # the gem cannot be updated without updating rest-client
     @mime_type = @file_type == "azw3" ? "application/x-mobi8-ebook" : MIME::Types.type_for(@file_type).first
+    @timestamp = Time.now.to_i
```
redsummernight: Is this unique enough or should I go for [fractions of seconds](https://ruby-doc.org/core-2.3.4/Time.html#method-i-to_f) as well?
```
(Time.now.to_f * 1000).to_i
```

tickinginstant: If you're concerned about timing issues, what about using `Dir.mktmpdir`? It uses random numbers and checks for `EEXIST` on `mkdir`.

redsummernight: Nice, thanks! I'll use that. I need to rewrite the tests to not rely on file location checks since we won't know the directory name.

app/models/work.rb
```diff
@@ -846,11 +846,6 @@ def set_word_count(preview = false)
     end
   end
 
-  after_update :remove_outdated_downloads
```
zz9pzza: Does this mean that more files will be left around ?

redsummernight: No, because we delete them after we generate and send them off.

app/controllers/downloads_controller.rb
```diff
@@ -47,7 +48,7 @@ def load_work
   def remove_downloads
     yield
   ensure
-    Download.remove(@work) unless Rails.env.test?
+    @download.remove
```
zz9pzza: Do we not want to delete the directory and things like the zip file ?

redsummernight: We do, `@download.remove` does that.

zz9pzza: FileUtils.rm_rf(dir) is not subtle

PR #3547, opened 2019-03-24
AO3-5492 Index work tags with numbers

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5492

## Purpose

Use the default analyzer which doesn't discard numbers, unlike [the simple analyzer](https://www.elastic.co/guide/en/elasticsearch/reference/6.6/analysis-simple-analyzer.html).

## Testing Instructions

See issue. The work index will need to be rebuilt (there's a mapping change).


sarken:

PR #3548, opened 2019-03-24
AO3-5648 Keep anchor links intact in imported works

ariana-paris:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5648

## Purpose

Preserves internal anchor links in imported documents

## Testing Instructions

When importing a work which contains links to anchors within the document (eg: `#notes`), the links should no longer be turned into absolute links to the external site url.


zz9pzza:
Given its the story parser can we have a test ?

PR #3551, opened 2019-03-29
AO3-5245 Update Open Doors rake task to use AO3 ids to find works

ariana-paris:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5245

## Purpose

Fix a bug that caused the rake task to be unable to update the URL on imported works.

## Testing Instructions

See ticket.

### DIFF
--- a/spec/lib/tasks/opendoors.rake_spec.rb
+++ b/spec/lib/tasks/opendoors.rake_spec.rb
@@ -1,0 +1,68 @@
+require 'spec_helper'
+require 'rake'
+
+describe 'Update imported urls' do
+  before do
+    @rake = Rake.application
+    begin
+      @rake.init
+    rescue SystemExit
+    end
+    @rake.load_rakefile
+  end
+
+  context "Rake task" do
+    it 'runs and takes a parameter' do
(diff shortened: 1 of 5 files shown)


spec/lib/tasks/opendoors.rake_spec.rb
```diff
@@ -0,0 +1,68 @@
+require 'spec_helper'
+require 'rake'
+
+describe 'Update imported urls' do
```
redsummernight: As of #3473 there's an easier way to set up specs for a rake.

I think it's better if you have the rake run against an actual CSV fixture, and write the expects on the work objects (e.g. their updated times / import URLs) in addition to checking opendoors_result.txt.

ariana-paris: Good point. I've added a test to check that the CSV file is actually processed from the file system (and indeed it turned out it ... wasn't). Got into a bit of a tangle with having to create the CSV programmatically as it needs the work ids to be correct, but you can't use let with a before all and you can't use create outside a block and creating a Work from scratch is a pain and anyway, I ended up with instance variables and it now works, even if there's probably a better way to organise this.

PR #3552, opened 2019-03-29
AO3-5572 Make API messages an array rather than a string

ariana-paris:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5572

## Purpose

Messages returned by the mass import API are sometimes strings and sometimes an array. They should be an array.

## Testing Instructions

See ticket


app/controllers/api/v2/works_controller.rb
```diff
@@ -3,9 +3,11 @@ class Api::V2::WorksController < Api::V2::BaseController
 
   # POST - search for works based on imported url
   def search
+    Rails.logger.debug("------- API v2: search for works ----------")
```
redsummernight: Are you keeping this for staging?

app/models/story_parser.rb
```diff
@@ -794,17 +799,20 @@ def download_with_timeout(location, limit = 10)
...
         case response
-        when Net::HTTPSuccess
-          story = response.body
-        when Net::HTTPRedirection
-          if limit > 0
-            story = download_with_timeout(response['location'], limit - 1)
-          end
-        else
-          nil
+          when Net::HTTPSuccess
```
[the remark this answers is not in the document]
redsummernight: I think this is worth fixing.

spec/controllers/api/v2/api_works_spec.rb
```diff
@@ -11,12 +11,12 @@
     before :all do
       mock_external
     end
-    
+
     before :each do
       Rails.cache.clear
```
redsummernight: Should we remove this step? We do that at:

https://github.com/otwcode/otwarchive/blob/c8fe1392b0493dfd5730be590d3b60715414791e/spec/spec_helper.rb#L96

PR #3555, opened 2019-04-01
AO3-5649 Sort authors_to_sort_on to match byline

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5649

## Purpose

When a work has multiple co-creators, `authors_to_sort_on` should sort them similar to how the byline does it:

https://github.com/otwcode/otwarchive/blob/d7c8504d24eb9c1d2a8a69ffe538c5781a75e5dd/app/helpers/application_helper.rb#L117-L124

## Testing Instructions

## References

See issue. Works need to be reindexed.

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -1170,14 +1170,13 @@ class Work < ApplicationRecord
 
   SORTED_AUTHOR_REGEX = %r{^[\+\-=_\?!'"\.\/]}
 
-  # TODO drop unused database column authors_to_sort_on
   def authors_to_sort_on
     if self.anonymous?
       "Anonymous"
     elsif self.authors.present?
-      self.authors.map(&:name).join(",  ").downcase.gsub(SORTED_AUTHOR_REGEX, '')
+      self.authors.map(&:name).sort.join(",  ").downcase.gsub(SORTED_AUTHOR_REGEX, '')
     else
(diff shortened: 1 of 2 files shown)


tickinginstant:
Capitalization interacts with this in interesting ways. Pseuds have a custom comparison function that downcases:

https://github.com/otwcode/otwarchive/blob/d7c8504d24eb9c1d2a8a69ffe538c5781a75e5dd/app/models/pseud.rb#L133-L135

So the byline helper, which sorts the pseuds before retrieving each pseud's byline (and therefore uses this comparison function), should place "alice" before "Zeke," because when they're downcased, `"alice" < "zeke"`. But this PR sorts after retrieving the names but before downcasing, and `"alice" > "Zeke"`, so I believe this would sort a work co-created by "alice" and "Zeke" among the creators starting with Z.

I tried to run a test to verify, and I think this would be the order for these ten bylines:
1. by alice
2. by alice, zach
3. by Anna
4. by alice, Anna
5. by Anna, zach
6. by Anna, Zeke
7. by zach
8. by Zeke
9. by alice, Zeke
10. by zach, Zeke

redsummernight:
Thanks! I moved the sort ahead of getting the names.

sarken:

PR #3557, opened 2019-04-02
AO3-5598 Set race_condition_ttl on cached AdminSetting

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5598

## Purpose

> Setting `:race_condition_ttl` is very useful in situations where a cache entry is used very frequently and is under heavy load. If a cache expires and due to heavy load several different processes will try to read data natively and then they all will try to write to cache.

https://api.rubyonrails.org/v5.1.6.2/classes/ActiveSupport/Cache/Store.html#method-i-fetch

## Testing Instructions

See issue.

### DIFF
--- a/app/models/admin_setting.rb
+++ b/app/models/admin_setting.rb
@@ -30,7 +30,7 @@ class AdminSetting < ApplicationRecord
   }.freeze
 
   def self.current
-    Rails.cache.fetch("admin_settings") { AdminSetting.first } || OpenStruct.new(DEFAULT_SETTINGS)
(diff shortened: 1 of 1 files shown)


redsummernight:
https://github.com/rails/rails/blob/v5.1.6.2/activesupport/lib/active_support/cache.rb#L565

We can specify seconds (making it an ActiveSupport::Duration) or just a number, but in the end Rails will call `to_i` and normalize the parameter to an integer. I went with seconds, following the docs, so we don't need to look up the unit.

sarken:
So potentially slightly more convenient for us if we add it elsewhere, but by no means necessary 👍

PR #3558, opened 2019-04-02
AO3-5653 Fix intermittent AbuseReport alternate URL test failures

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5653

## Purpose

These tests need to be in the "when report is not spam" block which stubs Akismetor.

## Testing Instructions

Travis should be good.


sarken:
BLESS YOU

PR #3562, opened 2019-04-06
AO3-5624 Reorder buttons on homepage unread messages

NerineLuna:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5624

## Purpose

Rearranges the Reply, Mark Read, and Delete bottons on messages in the Unread Message section of the homepage from Delete, Mark Read, Reply, to Reply, Mark Read, Delete, to more closely mirror the order of these buttons on the comments section of works.

## Testing Instructions

Go to https://test.archiveofourown.org/ and log in to an account. On the homepage, scroll to the Unread Messages section. Confirm that the buttons on the message(s) are Reply, Mark Read, and Delete, in that order, and that they work as expected. 

## Credit
Nerine Luna Cyran, she/her

### DIFF
--- a/app/views/home/_inbox_module.html.erb
+++ b/app/views/home/_inbox_module.html.erb
@@ -19,13 +19,13 @@
             <li>
               <%= render 'inbox/reply_button', feedback_comment: inbox_comment.feedback_comment %>
             </li>
+          <% end %>
           <li id="read_comment_form_<%= inbox_comment.feedback_comment.id %>">
             <%= render 'inbox/read_form', inbox_comment: inbox_comment, current_user: current_user %>
           </li>
           <li>
             <%= render 'inbox/delete_form', inbox_comment: inbox_comment, current_user: current_user %>
           </li>
-          <% end %>
         </ul>
       </li>
     <% end %>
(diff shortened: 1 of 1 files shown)


app/views/home/_inbox_module.html.erb
```diff
@@ -25,10 +19,16 @@
             <li>
               <%= render 'inbox/reply_button', feedback_comment: inbox_comment.feedback_comment %>
             </li>
+          <li id="read_comment_form_<%= inbox_comment.feedback_comment.id %>">
+            <%= render 'inbox/read_form', inbox_comment: inbox_comment, current_user: current_user %>
+          </li>
+          <li>
+            <%= render 'inbox/delete_form', inbox_comment: inbox_comment, current_user: current_user %>
+          </li>
```
tickinginstant: It looks like these two buttons used to appear regardless of whether the comment was unreviewed, and now will only appear if the comment isn't unreviewed. Is that an intentional change?

sarken: Drat. No, it's not! Thanks for catching that. I had to read it three times even after your comment to see where that was happening. /o\ 

So what we want is:
1. Approve, which should only show when the comment is unreviewed
2. Reply, which should only show when the comment is not unreviewed
3. Mark Read, which should appear regardless of reviewed status
4. Delete, which should also appear regardless of reviewed status

PR #3568, opened 2019-04-21
AO3-5233 Case insensitive admin bulk email search

ariana-paris:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5233

## Purpose

This improves the bulk email search in the admin section so it looks for case insensitive emails and reports on the number of duplicates to make it easier to know how many matches there were.

## Testing Instructions

This can be tested through the Admin interface. See Jira for the steps to reproduce the original bug.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -299,10 +299,10 @@ class User < ApplicationRecord
     users = User.where(email: unique_emails)
     found_emails = users.map(&:email).map(&:downcase)
     # Remove found users from the total list of unique emails and count duplicates
-    not_found = unique_emails - found_emails
+    not_found_emails = unique_emails - found_emails
     num_duplicates = emails.size - unique_emails.size
-    
-    [users, not_found, num_duplicates]
+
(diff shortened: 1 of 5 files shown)


app/models/user.rb
```diff
@@ -292,10 +292,17 @@ def self.filter_by_name_or_email(users, query, exact)
...
+    all_emails = emails.map(&:downcase)
+    unique_emails = all_emails.uniq
+    # Find users and their email addresses
+    users = User.where(email: unique_emails)
+    found_emails = users.map(&:email).map(&:downcase)
+    # Remove found users from the total list of unique emails and count duplicates
+    not_found = unique_emails - found_emails
+    num_duplicates = emails.size - unique_emails.size
+    
+    [users, not_found, num_duplicates]
```
redsummernight: For consistency, can we name the second value `not_found_emails`?

CristinaRO:
Updated (with Ariana's permission)!

PR #3570, opened 2019-04-27
AO3-5671 Update Elasticsearch to 7.12.0

ariana-paris:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5671

## Purpose

Upgrades the Elasticsearch gem to 7.12.0 and removes types from mapping and routing info as these are no longer used.

## Testing Instructions

See issue.


Gemfile
```diff
@@ -140,7 +140,7 @@ group :test do
   gem 'pickle'
   gem 'shoulda'
   gem 'capybara', '~> 2.16.1'
-  gem 'database_cleaner', '1.5.2'
+  gem 'database_cleaner', '1.7.0'
```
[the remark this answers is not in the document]
ariana-paris: That was to stop Rails whingeing about a deprecated feature used by older versions of databade_cleaner when running the specs - but yes, it isn't related to ES so I've reverted the version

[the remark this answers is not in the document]
redsummernight: It's a test-only gem so I think having it would be fine. What were the Rails warnings it fixed?

sarken:
Taking off the deferred label because we've merged 6.8 and the needs reindex label because we've checked now and it doesn't.

app/models/search/query_result.rb
```diff
@@ -104,7 +104,7 @@ def total_entries
   end
 
   def unlimited_total_entries
-    response.dig('hits', 'total') || 0
+    response.dig('hits', 'total', 'value') || 0
```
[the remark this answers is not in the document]
sarken: Could we save ourselves some effort by having downtime to deploy, so we're not looking at a situation where we'd be running mismatched versions of the gem and code?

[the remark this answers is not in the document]
redsummernight: We could, yeah. If we're fine with downtime, we can skip the compatibility questions.

That leaves the question of what to do with [`track_total_hits` defaulting to 10,000](https://www.elastic.co/guide/en/elasticsearch/reference/7.0/breaking-changes-7.0.html#track-total-hits-10000-default) for performance reasons. In the short term, we can force ES to [always return the accurate total](https://www.elastic.co/guide/en/elasticsearch/reference/7.11/search-your-data.html#track-total-hits), but we might eventually need to change our pagination approach.

script/travis/elasticsearch_upgrade.sh
```diff
@@ -1,9 +1,9 @@
 #!/bin/bash
-ES_VERSION="6.8.8"
+ES_VERSION="7.6.2"
```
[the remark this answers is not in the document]
redsummernight: With recent changes, instead of Travis we'll need to update GitHub Actions and docker-compose.yml.

app/models/search/query_result.rb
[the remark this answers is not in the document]
sarken: Let's go with the short term solution and deal with the bigger issue at some other point.

sarken:
We should bump the version on this as well. 7.12.0 is the most recent version as of this comment.

(Ariana, if you're low on time or energy or just sick of dealing with this PR, I'm happy to take over.)

app/models/search/work_creator_indexer.rb
```diff
@@ -12,14 +12,13 @@ def self.mapping
...
       "_index" => index_name,
-      "_type" => document_type,
       "_id" => document_id(id),
       "routing" => parent_id(id, nil)
     }
   end
 
   def document_id(id)
-    "#{id}-creator"
+    id.to_s.ends_with?("creator") ? id : "#{id}-creator"
```
tickinginstant, line 21: Maybe it'd be good to add a [find_elasticsearch_ids](https://github.com/otwcode/otwarchive/blob/b63fce67a3975ec0614bb535c75d76c8dafbda71/app/models/search/bookmarkable_indexer.rb#L15-L18) method here, instead? If there's already something with an id of `9-creator-creator-creator` knocking around in the system, this would let it slide, instead of fixing it and making it `9-creator`.

ariana-paris: I'll try out a find_elasticsearch_ids and see how it goes 👍 . Before my change, the [tests on this branch were failing](https://github.com/otwcode/otwarchive/runs/2375780750?check_suite_focus=true#step:13:4522) with the new and exciting `Validation Failed: 1: id [57-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator-creator] is too long, must be no longer than 512 bytes but was: 514;2:` so it should be easy to see if it's working.

tickinginstant: Hmm, I believe that error only shows up if there are persistent indexing failures unrelated to the `-creator` suffix issue (since with this bug, each `-creator` suffix added represents one pass through the failures queue, so that particular ID represents the end of a long sequence of indexing errors). Did you track down the underlying indexing error? If not, it may still be failing silently.

ariana-paris: Actually, you're right, the error that was making it do that became apparent when I tweaked the id: there was a nested array in the query that ES didn't like which I fixed with the [other change in this commit](https://github.com/otwcode/otwarchive/pull/3570/commits/899b416a98c684fb253111d02af98f40b16686b7#diff-e018acad1b34c8885b0293476d3257a17cf82055a34e1e69a1374986b4a3ac56R21)  (I was able to troubleshoot locally so both changes ended up in the same commit)

tickinginstant: Was that really the cause? Does it work if you disable the `-creator` fix? Because that looks like it'd only make the query fail, not the indexing. (And it looks like it's for a different index, too.)

tickinginstant: I just looked at the test runs for more of the commits, and it looks like it may have been fixed in the commit before the one you linked: https://github.com/otwcode/otwarchive/pull/3570/commits/b6ffa99f8f496bce148b4cdc20ea6e2f8b2abac1. Not sure why that would be marked as an error instead of a warning, but it's related to indexing and it does involve the right index, so I think it's a more likely candidate than an issue with the pseud query.

(I double-checked locally, and it looks like with `_type` included in the routing info, the indexing error is `{"type"=>"illegal_argument_exception", "reason"=>"can't merge a non object mapping [creator_join] with an object mapping"}`. I'm guessing that maybe specifying a type caused something to go wrong with the lookup for the mapping? IDK.)

ariana-paris: Well, I've now also removed the find_elasticsearch_ids method you'd suggested the other day and the tests are still passing without the `creator-creator-creator` ids being generated, so whatever it was, it appears to be fixed now without any unnecessary changes to this file. 🎉

tickinginstant: The `find_elasticsearch_ids` method is still a good bug fix to have (either here or in a different PR). The `creator-creator-creator` issue is a bug in the error-handling system; removing the deprecated `_type` prevents the error that caused the error-handling system to run (and thus prevents the bug in the error-handling system from showing up in tests), but the bug still does need to be addressed.

features/bookmarks/bookmark_search.feature
```diff
@@ -241,4 +241,4 @@ Feature: Search Bookmarks
   Scenario: Inputting bad queries
   When I fill in "Any field on work" with "bad~query~~!!!"
     And I press "Search Bookmarks"
-  Then I should see "Your search failed because of a syntax error"
+  Then I should see "No results found. You may want to edit your search to make it less specific."
```
redsummernight, line 244: This test is meant to check the "Your search failed because of a syntax error" rescue clause, so I don't think it should be updated this way. If ES 7 can handle this query string, let's change to this one used in a work search test which still fails as expected:

https://github.com/otwcode/otwarchive/blob/1f7610f25b28388bd6b2beb4715f2a80ece78ef7/features/search/works_info.feature#L6-L10

app/models/search/query_result.rb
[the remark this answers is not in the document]
redsummernight: Can you set [`track_total_hits`](https://rubydoc.info/gems/elasticsearch-api/7.12.0/Elasticsearch%2FAPI%2FActions:search) on every request in the base query class?

Gemfile.lock
```diff
@@ -79,578 +79,677 @@ GEM
       activerecord (>= 3.0)
     addressable (2.7.0)
       public_suffix (>= 2.0.2, < 5.0)
-    after_commit_everywhere (0.1.5)
+    after_commit_everywhere (1.0.0)
```
redsummernight, line 82: There are a lot of gem updates in here, are they all related to ES 7? Did you have to explicitly manually update any other gems for version compatibility issues?

features/bookmarks/bookmark_search.feature
ariana-paris: Thanks for the suggestion. It turns out the underlying query is different on that one and runs against the works index which is why it still fails even though the bookmark one is inexplicably successful (the number of tildes in the query string doesn't make a difference). 

I've done some testing comparing the test index on ES6 (400 as expected) vs the test index on ES7 (200 OK - not expected) vs my local development index on ES7 (400 as expected too). I've summarised the tests in chat so I guess we can discuss more in depth there.

Gemfile.lock
ariana-paris: No, those all came along for the ride when I did `bundle update` - I've now merged with master so the extra changes should be gone now (in particular all the ones for newfangled AWS services!)

features/bookmarks/bookmark_search.feature
ariana-paris: (FWIW, it turns out the solution was to populate the index in the test)

redsummernight (approved):

redsummernight:
I'll leave the green label for a second reviewer.

lib/css_cleaner.rb
```diff
@@ -104,7 +104,7 @@ def sanitize_css_property(property)
   #   empty property returned.
   def sanitize_css_declaration_value(property, value)
     clean = ""
-    property.downcase!
+    property = property.downcase
```
sarken, line 107: Is this still relevant now that we're no longer updating the CSS parser gem?

ariana-paris: Evidently not since the tests all passed when I reverted the change 👍

sarken (approved):

ThePadawan:
Small aside: 

https://otwarchive.atlassian.net/browse/AO3-5984 currently occurs due to https://github.com/elastic/elasticsearch/issues/57917, which has been fixed in https://github.com/elastic/elasticsearch/pull/57987 .

It is worth investigating of AO3-5984 is obsolete (fixed for free) after this ES upgrade.

redsummernight:
> 
> 
> Small aside:
> 
> https://otwarchive.atlassian.net/browse/AO3-5984 currently occurs due to [elastic/elasticsearch#57917](https://github.com/elastic/elasticsearch/issues/57917), which has been fixed in [elastic/elasticsearch#57987](https://github.com/elastic/elasticsearch/pull/57987) .
> 
> It is worth investigating of AO3-5984 is obsolete (fixed for free) after this ES upgrade.

Nice catch, thanks! I'll make a note on AO3-5984 to retest it with ES7.

PR #3526, opened 2019-02-15
AO3-5613 Renames Warning class to ArchiveWarning

mattsears:
# Pull Request Checklist

* [X] Have you read [How to write the perfect pull request](https://github.com/blog/1943-how-to-write-the-perfect-pull-request)?
* [X] Have you read through the [contributor guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [X] Have you added tests for any changed functionality?
* [X] Have you added the [JIRA](https://otwarchive.atlassian.net) issue number as the *first* thing in your pull request title (eg: `AO3-1234 Fix thing`)
* [X] Have you updated or commented on the JIRA issue with the information below?

## Issue

https://otwarchive.atlassian.net/browse/AO3-5613 (please fill in issue number and remove this comment)

## Purpose

This pull requests renames Warning class to ArchiveWarning to prep for upcoming Ruby upgrades. In addition, this pull requests repairs broken tests as a result of the name change. Given there is quite a bit of meta programming around tags and warnings and the fact that "ArchiveWarning" introduced two words instead of one, required some regex updates and other updates to accommodate the new change. 

It should be noted that every was change was made fo "Warning" -> "ArchiveWarming" and warning  -> archive_warning, etc. In order to avoid confusion with Ruby's warning.

## Testing Instructions

How can the Archive's QA team verify that this is working as you intended? (If you have access, please copy this into the JIRA ticket for them!)

I've verified that all tests pass on TravisCI.

## References

Are there any other relevant issues / PRs / mailing lists discussions related to this?

Not to my knowledge

## Credit

*What name do you want us to use to credit your work in the [Archive of Our Own's Release Notes](https://archiveofourown.org/admin_posts?tag=1)?*

(If you have an account on JIRA, please include the same name in the "Full name" field on your JIRA profile,so we can find you and assign you the issues you're working on.)

*What pronouns do you prefer (she/her, he/him, zie/hir etc)?*

he/him

(If you do not fill this in, we will use your GitHub account name and will use "they" to avoid making assumptions about your gender.)

PR #3575, opened 2019-05-15
AO3-5678 Fix Brakeman warnings for SQL injection

zz9pzza:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5678

## Purpose

Fixes Brakeman warnings for SQL injection.

## Testing Instructions

See jira ticket

### DIFF
--- a/app/controllers/archive_faqs_controller.rb
+++ b/app/controllers/archive_faqs_controller.rb
@@ -104,7 +104,7 @@ class ArchiveFaqsController < ApplicationController
   # reorder FAQs
   def update_positions
     if params[:archive_faqs]
-      @archive_faqs = ArchiveFaq.reorder(params[:archive_faqs])
+      @archive_faqs = ArchiveFaq.reorder(params[:archive_faqs].map { |s| s.to_i.to_s })
       flash[:notice] = ts("Archive FAQs order was successfully updated.")
     elsif params[:archive_faq]
       params[:archive_faq].each_with_index do |id, position|
--- a/app/controllers/archive_faqs_controller.rb
+++ b/app/controllers/archive_faqs_controller.rb
@@ -104,7 +104,7 @@ class ArchiveFaqsController < ApplicationController
   # reorder FAQs
   def update_positions
     if params[:archive_faqs]
-      @archive_faqs = ArchiveFaq.reorder(params[:archive_faqs].map { |s| s.to_i.to_s })
+      @archive_faqs = ArchiveFaq.reorder_list(params[:archive_faqs])
       flash[:notice] = ts("Archive FAQs order was successfully updated.")
     elsif params[:archive_faq]
       params[:archive_faq].each_with_index do |id, position|
--- a/app/controllers/challenge_requests_controller.rb
+++ b/app/controllers/challenge_requests_controller.rb
@@ -21,7 +21,7 @@ class ChallengeRequestsController < ApplicationController
     set_sort_order
     
     # actual content, do the efficient method unless we need the full query
-    
+
     if @sort_column == "fandom"
       query = "SELECT prompts.*, GROUP_CONCAT(tags.name) AS tagnames FROM prompts INNER JOIN set_taggings ON prompts.tag_set_id = set_taggings.tag_set_id 
       INNER JOIN tags ON tags.id = set_taggings.tag_id 
@@ -29,9 +29,9 @@ class ChallengeRequestsController < ApplicationController
       @requests = Prompt.paginate_by_sql(query, page: params[:page], per_page: ArchiveConfig.ITEMS_PER_PAGE)
     elsif @sort_column == "prompter" && !@collection.prompts.where(anonymous: true).exists?
       @requests = @collection.prompts.where("type = 'Request'").
-                              joins(challenge_signup: :pseud).
-                              order("pseuds.name #{@sort_direction}").
-                              paginate(page: params[:page])
+        joins(challenge_signup: :pseud).
+        order(["pseuds.name ?", @sort_direction]).
+        paginate(page: params[:page])
     else
       @requests = @collection.prompts.where("type = 'Request'").order(@sort_order).paginate(page: params[:page], per_page: ArchiveConfig.ITEMS_PER_PAGE)
     end
--- a/app/controllers/challenge_requests_controller.rb
+++ b/app/controllers/challenge_requests_controller.rb
@@ -16,7 +16,7 @@ class ChallengeRequestsController < ApplicationController
 
   def index
     @show_request_fandom_tags = (@collection.challenge.request_restriction.allowed("fandom") > 0 || (!@collection.challenge.prompt_restriction.nil? && @collection.challenge.prompt_restriction.allowed("fandom") > 0))
-    
+    direction = (@sort_direction == "ASC" ? "ASC" : "DESC")
     # sorting
     set_sort_order
     
@@ -30,7 +30,7 @@ class ChallengeRequestsController < ApplicationController
     elsif @sort_column == "prompter" && !@collection.prompts.where(anonymous: true).exists?
       @requests = @collection.prompts.where("type = 'Request'").
         joins(challenge_signup: :pseud).
-        order(["pseuds.name ?", @sort_direction]).
+        order("pseuds.name #{direction}").
         paginate(page: params[:page])
     else
       @requests = @collection.prompts.where("type = 'Request'").order(@sort_order).paginate(page: params[:page], per_page: ArchiveConfig.ITEMS_PER_PAGE)
--- a/app/controllers/challenge_requests_controller.rb
+++ b/app/controllers/challenge_requests_controller.rb
@@ -16,7 +16,6 @@ class ChallengeRequestsController < ApplicationController
 
   def index
     @show_request_fandom_tags = (@collection.challenge.request_restriction.allowed("fandom") > 0 || (!@collection.challenge.prompt_restriction.nil? && @collection.challenge.prompt_restriction.allowed("fandom") > 0))
-    direction = (@sort_direction == "ASC" ? "ASC" : "DESC")
     # sorting
     set_sort_order
     
@@ -30,7 +29,7 @@ class ChallengeRequestsController < ApplicationController
     elsif @sort_column == "prompter" && !@collection.prompts.where(anonymous: true).exists?
       @requests = @collection.prompts.where("type = 'Request'").
         joins(challenge_signup: :pseud).
-        order("pseuds.name #{direction}").
+        order("pseuds.name #{@sort_direction}").
         paginate(page: params[:page])
     else
       @requests = @collection.prompts.where("type = 'Request'").order(@sort_order).paginate(page: params[:page], per_page: ArchiveConfig.ITEMS_PER_PAGE)
--- a/app/controllers/challenge_requests_controller.rb
+++ b/app/controllers/challenge_requests_controller.rb
@@ -16,9 +16,11 @@ class ChallengeRequestsController < ApplicationController
 
   def index
     @show_request_fandom_tags = (@collection.challenge.request_restriction.allowed("fandom") > 0 || (!@collection.challenge.prompt_restriction.nil? && @collection.challenge.prompt_restriction.allowed("fandom") > 0))
+
     # sorting
     set_sort_order
-    
+    direction = (@sort_direction == "ASC" ? "ASC" : "DESC")
+
     # actual content, do the efficient method unless we need the full query
 
     if @sort_column == "fandom"
@@ -29,7 +31,7 @@ class ChallengeRequestsController < ApplicationController
     elsif @sort_column == "prompter" && !@collection.prompts.where(anonymous: true).exists?
       @requests = @collection.prompts.where("type = 'Request'").
         joins(challenge_signup: :pseud).
-        order("pseuds.name #{@sort_direction}").
+        order("pseuds.name #{direction}").
         paginate(page: params[:page])
     else
       @requests = @collection.prompts.where("type = 'Request'").order(@sort_order).paginate(page: params[:page], per_page: ArchiveConfig.ITEMS_PER_PAGE)
--- a/app/controllers/challenge_requests_controller.rb
+++ b/app/controllers/challenge_requests_controller.rb
@@ -19,7 +19,7 @@ class ChallengeRequestsController < ApplicationController
 
     # sorting
     set_sort_order
-    direction = (@sort_direction == "ASC" ? "ASC" : "DESC")
+    direction = (@sort_direction.casecmp("ASC").zero? ? "ASC" : "DESC")
 
     # actual content, do the efficient method unless we need the full query
 
--- a/app/controllers/chapters_controller.rb
+++ b/app/controllers/chapters_controller.rb
@@ -177,7 +177,7 @@ class ChaptersController < ApplicationController
 
   def update_positions
     if params[:chapters]
-      @work.reorder(params[:chapters])
+      @work.reorder(params[:chapters].map { |c| c.to_i.to_s })
       flash[:notice] = ts("Chapter order has been successfully updated.")
     elsif params[:chapter]
       params[:chapter].each_with_index do |id, position|
--- a/app/controllers/chapters_controller.rb
+++ b/app/controllers/chapters_controller.rb
@@ -177,7 +177,7 @@ class ChaptersController < ApplicationController
 
   def update_positions
     if params[:chapters]
-      @work.reorder(params[:chapters])
+      @work.reorder_list(params[:chapters])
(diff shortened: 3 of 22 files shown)


app/controllers/archive_faqs_controller.rb
```diff
@@ -104,7 +104,7 @@ def update
   # reorder FAQs
   def update_positions
     if params[:archive_faqs]
-      @archive_faqs = ArchiveFaq.reorder(params[:archive_faqs])
+      @archive_faqs = ArchiveFaq.reorder(params[:archive_faqs].map { |s| s.to_i.to_s })
```
tickinginstant: I'm fairly certain that this is a false positive. It looks like Brakeman thinks that this is calling the built-in Rails `reorder` function, which can be vulnerable to SQL injection if you pass arbitrary user input. But the `ArchiveFaq` model overrides the definition of `reorder` to do something completely different:
https://github.com/otwcode/otwarchive/blob/3862fc69353d3c9c6e06550ac860b6aa067170e5/app/models/archive_faq.rb#L30-L32
And I don't believe that definition of `reorder` is vulnerable to SQL injection, since it calls `collect!(&:to_i)` on its input:
https://github.com/otwcode/otwarchive/blob/3862fc69353d3c9c6e06550ac860b6aa067170e5/lib/sortable_list.rb#L2-L13
It seems to me that it might be cleaner to just rename the `reorder` function to `reorder_questions` to avoid confusion (both for Brakeman and for anyone viewing the code), instead of calling `to_i.to_s` everywhere.

app/controllers/challenge_requests_controller.rb
```diff
@@ -21,17 +21,17 @@ def index
...
       INNER JOIN tags ON tags.id = set_taggings.tag_id 
       WHERE prompts.type = 'Request' AND tags.type = 'Fandom' AND prompts.collection_id = " + @collection.id.to_s + " GROUP BY prompts.id ORDER BY tagnames " + @sort_direction
       @requests = Prompt.paginate_by_sql(query, page: params[:page], per_page: ArchiveConfig.ITEMS_PER_PAGE)
     elsif @sort_column == "prompter" && !@collection.prompts.where(anonymous: true).exists?
       @requests = @collection.prompts.where("type = 'Request'").
-                              joins(challenge_signup: :pseud).
-                              order("pseuds.name #{@sort_direction}").
-                              paginate(page: params[:page])
+        joins(challenge_signup: :pseud).
+        order(["pseuds.name ?", @sort_direction]).
```
tickinginstant: The automated tests don't catch it because this line isn't covered, but I believe this doesn't work. I believe it encloses the sort direction in quotes, which isn't allowed.

app/controllers/chapters_controller.rb
```diff
@@ -177,7 +177,7 @@ def update
 
   def update_positions
     if params[:chapters]
-      @work.reorder(params[:chapters])
+      @work.reorder(params[:chapters].map { |c| c.to_i.to_s })
```
tickinginstant: I'm pretty sure this is the same `reorder` issue as above, and could be addressed by renaming the function to `reorder_chapters` instead.

app/controllers/questions_controller.rb
```diff
@@ -20,7 +20,7 @@ def load_archive_faq
   def update_positions
     if params[:questions]
       @archive_faq = ArchiveFaq.find_by_slug(params[:archive_faq_id])
-      @archive_faq.reorder(params[:questions])
+      @archive_faq.reorder(params[:questions].map { |s| s.to_i.to_s })
```
tickinginstant: Another `reorder` false positive.

app/controllers/series_controller.rb
```diff
@@ -136,7 +136,7 @@ def update
   def update_positions
     if params[:serial_works]
       @series = Series.find(params[:id])
-      @series.reorder(params[:serial_works])
+      @series.reorder(params[:serial_works].map { |s| s.to_i.to_s })
```
tickinginstant: Another call to `reorder` which is confusing Brakeman, and could be renamed to `reorder_works` instead.

app/controllers/wrangling_guidelines_controller.rb
```diff
@@ -53,7 +53,7 @@ def update
   # reorder FAQs
   def update_positions
     if params[:wrangling_guidelines]
-      @wrangling_guidelines = WranglingGuideline.reorder(params[:wrangling_guidelines])
+      @wrangling_guidelines = WranglingGuideline.reorder(params[:wrangling_guidelines].map { |s| s.to_i.to_s })
```
tickinginstant: `reorder` again.

app/models/challenge_claim.rb
```diff
@@ -48,11 +48,11 @@ def ==(other)
   scope :order_by_date, -> { order("created_at ASC") }
 
   def self.order_by_requesting_pseud(dir="ASC")
-    joins(REQUESTING_PSEUD_JOIN).order("pseuds.name #{dir}")
+    joins(REQUESTING_PSEUD_JOIN).order(["pseuds.name ?", dir])
```
tickinginstant: I believe this also doesn't work, for the same reason as the query in `ChallengeRequestsController`.

app/models/challenge_claim.rb
```diff
@@ -48,11 +48,11 @@ def ==(other)
...
   end
 
   def self.order_by_offering_pseud(dir="ASC")
-    joins(CLAIMING_PSEUD_JOIN).order("pseuds.name #{dir}")
+    joins(CLAIMING_PSEUD_JOIN).order(["pseuds.name ?", dir])
```
tickinginstant: Again, I believe this escapes the sort direction.

app/models/inbox_comment.rb
```diff
@@ -16,17 +16,21 @@ class InboxComment < ApplicationRecord
       when 'true' then true
       when 'false' then false
       else [true, false]
-    end
+                 end
```
tickinginstant: The indentation here doesn't match the rest of the block above.

app/models/inbox_comment.rb
```diff
@@ -16,17 +16,21 @@ class InboxComment < ApplicationRecord
...
+    direction = case filters[:date]&.upcase
```
tickinginstant: This only has two possible outcomes, so using a `case` statement is overkill. It'd be cleaner to write it as
```
direction = (filters[:date]&.upcase == "ASC" ? "created_at ASC" : "created_at DESC")
```
Or, since `created_at` is shared by both, you could even shorten it to:
```
direction = (filters[:date]&.upcase == "ASC" ? "ASC" : "DESC")
```
And then call `order(created_at: direction)` below.

app/controllers/wrangling_guidelines_controller.rb
sarken: Should we maybe use `g` as our variable here, as in `guideline`? I'm not sure what the `s` stands for.

app/models/potential_match.rb
```diff
@@ -167,7 +167,15 @@ def self.matching_signup_ids(collection, signup, collection_tag_sets, required_t
...
-      condition = "any_#{required_types.first.downcase} = 1"
+      condition = {
+        fandom: "any_fandom = 1",
+        character: "any_character = 1",
+        rating: "any_tating = 1",
+        relationship: "any_relationship = 1",
+        category: "any_category = 1",
+        warning: "any_warning = 1",
+        freeform: "any_freeform = 1"
+      }[ required_types.first.downcase.to_sym ]
```
sarken: I'm not sure why Hound isn't saying anything, but I don't think we want spaces inside the brackets here.

app/views/invite_requests/_invite_request.html.erb
```diff
@@ -6,7 +6,7 @@
 
 <!--main content-->
 <p>
-  <%= ts("You are currently number <strong>%{position}</strong> on our waiting list!", position: InviteRequest.where("position <= #{invite_request.position}").count).html_safe %>
+  <%= ts("You are currently number <strong>%{position}</strong> on our waiting list!", position: InviteRequest.where(["position <= ?", invite_request.position]).count).html_safe %>
```
sarken: If at all possible, I'd _love_ to take the opportunity to move `InviteRequest.where(["position <= ?", invite_request.position]).count` out of the view and put it in the controller instead, maybe as `@position_in_queue` or something.

app/models/work.rb
```diff
@@ -820,7 +820,7 @@ def update_complete_status
     # self.chapters.posted.count ( not self.number_of_posted_chapter , here be dragons )
     self.complete = self.chapters.posted.count == expected_number_of_chapters
     if self.will_save_change_to_attribute?(:complete)
-      Work.where("id = #{self.id}").update_all("complete = #{self.complete}")
+      Work.where(id: id).update_all(['complete = ?', complete])
```
sarken: Is there a reason for the switch from double quotes to single quotes here? (Just asking because the other ones appear to have stayed double.)

app/controllers/wrangling_guidelines_controller.rb
zz9pzza: s is for string :)

zz9pzza: Will be g in the next push

sarken: I think this can just have the method renamed in the model, like ticking suggested, so a variable rename won't be necessary

https://github.com/otwcode/otwarchive/blob/3862fc69353d3c9c6e06550ac860b6aa067170e5/app/models/wrangling_guideline.rb#L9-L11

app/controllers/archive_faqs_controller.rb
zz9pzza: I am not disagreeing that it is probably a false positive I am trying to minimise the  number of issues brakeman raises so that we don't just list all the things as false positives and we are less likely to make a mistake if we do that.  So I would also be happy with "rename the reorder function to reorder_questions". I have been trying to make the smallest changes I could and I had the sortable list code to be imported from elsewhere.

app/models/challenge_claim.rb
zz9pzza: `  def self.order_by_offering_pseud(dir="ASC")
    if dir == "ASC"
      joins(CLAIMING_PSEUD_JOIN).order("pseuds.name ASC")
    else
      joins(CLAIMING_PSEUD_JOIN).order("pseuds.name DESC")
    end
  end`

@tickinginstant that will work, happy with that approach ?

tickinginstant: Does `dir` need to be upcased for the comparison, or is it upcased before it gets passed in? Other than that, it looks like it should work.

(It might be nice to reduce the code duplication between the two branches of the if, but for code this short it's probably okay to leave it.)

app/models/challenge_claim.rb
```diff
@@ -48,11 +48,19 @@ def ==(other)
   scope :order_by_date, -> { order("created_at ASC") }
 
   def self.order_by_requesting_pseud(dir="ASC")
-    joins(REQUESTING_PSEUD_JOIN).order("pseuds.name #{dir}")
+    if dir.casecmp("ASC")
```
tickinginstant: `casecmp` returns `-1`/`0`/`1`, all of which count as `true` for the purposes of an `if` statement. So I believe this will always take the first branch.

app/models/potential_match.rb
```diff
@@ -167,7 +167,15 @@ def self.matching_signup_ids(collection, signup, collection_tag_sets, required_t
                             pluck(:challenge_signup_id).compact
 
       # now add on "any" matches for the required types
-      condition = "any_#{required_types.first.downcase} = 1"
+      condition = {
+        fandom: "any_fandom = 1",
+        character: "any_character = 1",
+        rating: "any_tating = 1",
```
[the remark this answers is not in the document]
zz9pzza: I read up around this and people seem to prefer hashes over case where you don't need the range that case can give you.

app/controllers/challenge_requests_controller.rb
```diff
@@ -16,22 +16,22 @@ def check_visibility
 
   def index
     @show_request_fandom_tags = (@collection.challenge.request_restriction.allowed("fandom") > 0 || (!@collection.challenge.prompt_restriction.nil? && @collection.challenge.prompt_restriction.allowed("fandom") > 0))
-    
+    direction = (@sort_direction == "ASC" ? "ASC" : "DESC")
```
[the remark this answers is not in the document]
zz9pzza: Could you look at this now ?

app/controllers/challenge_requests_controller.rb
```diff
@@ -16,22 +16,23 @@ def check_visibility
 
   def index
     @show_request_fandom_tags = (@collection.challenge.request_restriction.allowed("fandom") > 0 || (!@collection.challenge.prompt_restriction.nil? && @collection.challenge.prompt_restriction.allowed("fandom") > 0))
-    
+
     # sorting
     set_sort_order
-    
+    direction = (@sort_direction == "ASC" ? "ASC" : "DESC")
```
tickinginstant: Does this need case-insensitive comparison?

app/controllers/invite_requests_controller.rb
```diff
@@ -49,7 +50,7 @@ def manage
     end
   end
 
-  def reorder
+  def reorder_list
```
tickinginstant: Why was this marked as resolved?

app/models/challenge_claim.rb
```diff
@@ -48,11 +48,19 @@ def ==(other)
...
+    if dir.casecmp("ASC").zero?
+      joins(REQUESTING_PSEUD_JOIN).order("pseuds.name ASC")
+    else
+      joins(REQUESTING_PSEUD_JOIN).order("pseuds.name DESC")
+    end
   end
 
   def self.order_by_offering_pseud(dir="ASC")
-    joins(CLAIMING_PSEUD_JOIN).order("pseuds.name #{dir}")
+    if dir.casecmp("ASC")
```
tickinginstant: This needs the same treatment as the above.

app/models/potential_match.rb
```diff
@@ -167,7 +167,15 @@ def self.matching_signup_ids(collection, signup, collection_tag_sets, required_t
...
-      condition = "any_#{required_types.first.downcase} = 1"
+      condition = {
+        fandom: "any_fandom = 1",
+        character: "any_character = 1",
+        rating: "any_rating = 1",
+        relationship: "any_relationship = 1",
+        category: "any_category = 1",
+        warning: "any_warning = 1",
+        freeform: "any_freeform = 1"
+      }[required_types.first.downcase.to_sym]
```
tickinginstant: After poking around brakeman some more, I think you can avoid enumerating all tag types in this case by explicitly quoting the column name, like in the following:
```
      column_name = "any_#{required_types.first.underscore}"
      condition = "#{connection.quote_column_name(column_name)} = 1"
```

app/controllers/invite_requests_controller.rb
zz9pzza: ( As when I changed it all I ended up going though a loop of getting the brakeman errors back ) . I will do the change as an isolated commit and see if that is still the case.

app/models/potential_match.rb
zz9pzza: However I think it is more complex and I think the existing code is easier to understand ?

app/controllers/invite_requests_controller.rb
zz9pzza: Done

app/models/potential_match.rb
tickinginstant: The reason I suggested the `quote_column_name` approach in this particular case is because hard-coding all possibilities (whether in a hash or a case statement) makes it more annoying to fix up PRs like #3526, since it adds more places that need to be updated to match the new type name.

(I do prefer the case statement over the hash, but I'd like it better if neither was necessary.)

sarken:
It looks like everything's been addressed, but I'll leave it up to @tickinginstant to confirm.

tickinginstant:
I did have [one more comment that may have slipped through the cracks](https://github.com/otwcode/otwarchive/pull/3575#discussion_r287573535), but it's somewhat subjective. The PR can be merged as is. (Though if you don't get rid of the case statement, either this or #3526 is going to have to be updated, depending on the order that PRs are merged.)

sarken:
> I did have [one more comment that may have slipped through the cracks](https://github.com/otwcode/otwarchive/pull/3575#discussion_r287573535), but it's somewhat subjective.

Yeah, we decided we preferred the approach that requires either the case or the hash. james prefers to have the values hard coded in case there's ever a tag type we don't want to allow in there.

>Though if you don't get rid of the case statement, either this or #3526 is going to have to be updated, depending on the order that PRs are merged.

I'm hoping this will be after 3526 so we don't need to make more changes to 3526, but we'll see how things work out.

Thanks!

PR #3576, opened 2019-05-15
AO3-5682 Refactor code that uses send

zz9pzza:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5682

## Purpose

Refactor to remove brakeman warnings.

## Testing Instructions

Check that the manage page for invites work, with the filters ( eg unsent )
In tag wrangling show hidden: check tag type: warnings and freeform
And wrangling still works

### DIFF
--- a/app/controllers/invitations_controller.rb
+++ b/app/controllers/invitations_controller.rb
@@ -22,11 +22,10 @@ class InvitationsController < ApplicationController
   end
 
   def manage
-    if params[:status].blank? || !['unsent', 'unredeemed', 'redeemed'].include?(params[:status])
-      @invitations = @user.invitations
-    else
-      @invitations = @user.invitations.send(params[:status])
-    end
+    @invitations = @user.invitations
+    @invitations = @invitations.unsent if params[:status] == "unsent"
+    @invitations = @invitations.unredeemed if params[:status] == "unredeemed"
+    @invitations = @invitations.redeemed if params[:status] == "redeemed"
   end
 
   def show
--- a/app/controllers/invitations_controller.rb
+++ b/app/controllers/invitations_controller.rb
@@ -22,10 +22,16 @@ class InvitationsController < ApplicationController
   end
 
   def manage
-    @invitations = @user.invitations
-    @invitations = @invitations.unsent if params[:status] == "unsent"
-    @invitations = @invitations.unredeemed if params[:status] == "unredeemed"
-    @invitations = @invitations.redeemed if params[:status] == "redeemed"
+    @invitations = case params[:status]
+                   when "unsent"
+                     @user.invitations.unsent
+                   when "unredeemed"
+                     @user.invitations.unredeemed
+                   when "redeemed"
+                     @user.invitations.redeemed
+                   else
+                     @user.invitations
+                   end
   end
 
   def show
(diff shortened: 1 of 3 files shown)


app/controllers/tags_controller.rb
```diff
@@ -134,10 +134,13 @@ def feed
   def show_hidden
     unless params[:creation_id].blank? || params[:creation_type].blank? || params[:tag_type].blank?
       raise "Redshirt: Attempted to constantize invalid class initialize show_hidden #{params[:creation_type].classify}" unless %w(Series Work Chapter).include?(params[:creation_type].classify)
-      model = begin
-                params[:creation_type].classify.constantize
-              rescue
-                nil
+      model = case params[:creation_type].downcase
```
ariana-paris: Definitely an improvement over `rescue nil` 😖 
I assume this will set the model to nil anyway if the creation_type didn't match.

app/controllers/invitations_controller.rb
```diff
@@ -22,11 +22,10 @@ def index
   end
 
   def manage
-    if params[:status].blank? || !['unsent', 'unredeemed', 'redeemed'].include?(params[:status])
-      @invitations = @user.invitations
-    else
-      @invitations = @user.invitations.send(params[:status])
-    end
+    @invitations = @user.invitations
```
ariana-paris: Maybe in line with what you've done elsewhere, this might work as a case too?
```
@invitations = 
  case params[:status]
  when "unsent"
    @user.invitations.unsent
  when "unredeemed"
    @user.invitations.unredeemed
  when "redeemed"
    @user.invitations.redeemed
  else
    @user.invitations
  end
```

tickinginstant:
I think there may be a simpler way to handle some of these false positives. When calling `send` in a controller, Brakeman understands that if you have something of the form:
```
if %w(a b c).include?(y)
  x.send(y)
end
```
Then you've whitelisted the allowed actions, and therefore the call to `send` is safe even if `y` is extracted from `params`. The `TagsController` already has that structure to check all of the calls to `send`, except that it uses `params[:y]` instead of a variable `y`, which means that Brakeman doesn't recognize that the parameter has been checked against a whitelist.

But if you set `show = params[:show]`, `status = params[:status]`, and `tag_type = params[:tag_type]`, and then replace all parameter lookups with the corresponding variables throughout, then the pre-existing checks for those parameters should be enough to satisfy Brakeman, allowing you to revert back to the original code calling `send` and keep the changes to the code minimal.

(You may be able to do a similar thing with the `InvitationsController`, but the structure there is different enough that you'd probably have to refactor to make it work.)

app/controllers/tags_controller.rb
zz9pzza: yes

zz9pzza:
@tickinginstant would you be happier with https://github.com/zz9pzza/otwarchive/pull/29

tickinginstant:
I like that the changes are much shorter! I think you can also get rid of the `if` statement you added on line 318:
```
if %w(fandoms characters relationships freeforms sub_tags mergers).include?(show)
```
Since it's identical to the enclosing `if` statement on line 300.

sarken:
Should this be closed? It looks like #3578 replaces it.

PR #3577, opened 2019-05-15
AO3-5689 Refactor code to avoid use of eval

zz9pzza:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5689

## Purpose

Refactors the issues brakeman points out for eval

## Testing Instructions

Editing the settings on a prompt meme and gift exchange when there were none originally set
Editing the settings on a collections when you are the owner and edit the settings from the sidebar.
Deleting a collections( prompt and gift ) form the collection profile.
And the same from the challenge deletion form
Check that the user stats are right, check all the things stat wise “"date,"hits"  "kudos": work.kudos.count,      "comment_thread_count":  "bookmarks count":   "subscriptions.count": work.subscriptions.count,      "word_count"”
creating and editing a collection


app/controllers/stats_controller.rb
```diff
@@ -110,4 +105,15 @@ def index
...
+  private
+
+  def stat_element(work, element)
+    { "date": work.date,
+      "hits": work.hits,
+      "kudos.count": work.kudos.count,
+      "comment_thread_count": work.comment_thread_count,
+      "bookmarks.count": work.bookmarks.count,
+      "subscriptions.count": work.subscriptions.count,
+      "word_count": work.word_count }[element.to_sym]
```
tickinginstant: I think you should use a case statement here. Using a hash results in the kudos count, comment thread count, bookmarks count, and subscription count all being computed each time this function is called, even though at most one of those values is accessed and returned.

(I think some of the queries may be cached with low-level query caching, so this might not result in a 4x slowdown. But it's probably not a good idea to count on query caching to make up for this inefficiency.)

app/views/challenge/shared/_challenge_form_delete.html.erb
```diff
@@ -1,7 +1,12 @@
 <ul class="navigation actions" role="menu">
   <% if @collection.challenge %>
-    <li><%= link_to ts("Delete Challenge"), eval("collection_#{challenge_class_name(@collection)}_path(@collection)"),
-      data: {confirm: ts('Are you certain you want to delete the challenge from this collection? All signups, assignments, and settings will be lost. (Works will remain in the collection.)')},
-      :method => :delete %></li>
+    <li><%= link_to ts("Delete Challenge"),
+                    if  @collection.prompt_meme? then
```
sarken: Is explicitly writing the `then` necessary?

app/views/collections/_sidebar.html.erb
```diff
@@ -48,8 +48,13 @@
       <li><%= span_if_current(ts('Manage Items', key: 'dashboard'), collection_items_path(@collection)) %></li>
       <% if @collection.user_is_owner?(current_user) %>
         <% if @collection.challenge %>
-          <li><%= span_if_current ts('Challenge Settings', key: 'dashboard'), eval("edit_collection_#{challenge_class_name(@collection)}_path(@collection)") %></li>
-      	<% end %>
+          <% if @collection.prompt_meme? %>
+            <li><%= span_if_current ts('Challenge Settings', key: 'dashboard'), edit_collection_prompt_meme_path(@collection) %></li>
```
sarken: Why this approach instead of

if @collection.prompt_meme?
  yada
elsif @collection.gift_exchange?
  yada
end

app/views/potential_matches/_no_match_required.html.erb
```diff
@@ -3,7 +3,11 @@
   will be purely random. 
   (You can adjust the assignments by hand afterwards, though!)
   If you want participants matched up on the tags they signed up with, 
-  please update your ") %> 
-  <strong><%= link_to ts("challenge match settings"), 
-    eval("edit_collection_#{challenge_class_name(@collection)}_path(@collection, :anchor => 'match_settings')") %></strong>. 
+  please update your ") %>
+  <strong><%= link_to ts("challenge match settings"),
```
sarken: Wait, why did this even _have_ the eval? This page is shown only for gift exchanges, not for prompt memes (which don't have matching). You can get rid of the if/else here and just do `edit_collection_gift_exchange_path(@collection, :anchor => 'match_settings')`

app/views/collections/_sidebar.html.erb
zz9pzza: When I parse elsif I always have to look up what it does, while a set of ifs where the conditions are exclusive is clear to me.

sarken: Ah, okay. Please change to elsif, then!

app/views/challenge/shared/_challenge_form_delete.html.erb
```diff
@@ -1,7 +1,12 @@
 <ul class="navigation actions" role="menu">
   <% if @collection.challenge %>
-    <li><%= link_to ts("Delete Challenge"), eval("collection_#{challenge_class_name(@collection)}_path(@collection)"),
-      data: {confirm: ts('Are you certain you want to delete the challenge from this collection? All signups, assignments, and settings will be lost. (Works will remain in the collection.)')},
-      :method => :delete %></li>
+    <li><%= link_to ts("Delete Challenge"),
+                    if  @collection.prompt_meme?
```
sarken: It looks like there's two spaces after `if` instead of one.

app/views/collection_profile/show.html.erb
```diff
@@ -98,7 +98,11 @@
 
         <% if @collection.challenge %>
           <li><%= link_to ts('Delete Challenge'),
-                          eval("collection_#{challenge_class_name(@collection)}_path(@collection)"),
+                          if  @collection.prompt_meme? then
```
sarken: This has the same extra space and unnecessary `then` problem as the previous view file, btw.

app/controllers/collections_controller.rb
```diff
@@ -109,9 +109,12 @@ def create
     if @collection.save
       flash[:notice] = ts('Collection was successfully created.')
       unless params[:challenge_type].blank?
-        # This is a challenge collection
-        # TODO: remove unsafe usage of eval, this is vulnerable and a security risk
-        redirect_to eval("new_collection_#{params[:challenge_type].demodulize.tableize.singularize}_path(@collection)") and return
+        if params[:challenge_type] == "PromptMeme"
```
sarken: I think I mentioned it in chat the other day, but because we're dealing with two mutually exclusive conditions, this should get the same if/elsif treatment the sidebar view got, i.e.

```
if params[:challenge_type] == "PromptMeme"
  redirect_to new_collection_prompt_meme_path(@collection) and return
elsif params[:challenge_type] == "GiftExchange"
  redirect_to new_collection_gift_exchange_path(@collection) and return
end
```

sarken: Just clarifying that this applies throughout the entire file, not just on lines 112-117.

redsummernight:
This has test failures.

sarken:
Just needs the brand new merge conflicts resolved

PR #3580, opened 2019-05-30
AO3-5694 Don't load fandoms list on orphan_account user dashboard

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5694

## Purpose

Makes `@fandoms = []` for the orphan_account so we can see if loading that list is the reason the orphan_account's dashboard has stopped loading. 

This is a temporary change (famous last words!) to help us diagnose the actual problem. If the orphan_account dashboard starts loading once this is deployed, we should come up with a new way of counting fandoms. If it doesn't start loading, we should remove this code and keep looking for the problem.

## Testing Instructions

Refer to Jira.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -285,18 +285,18 @@ class UsersController < ApplicationController
                  []
                else
                  Fandom.select('tags.*, count(tags.id) as work_count').
-                        joins(:direct_filter_taggings).
-                        joins("INNER JOIN works
-                               ON filter_taggings.filterable_id = works.id
-                               AND filter_taggings.filterable_type = 'Work'").
-                        group('tags.id').
-                        merge(Work.send(visible_method).revealed.non_anon).
-                        merge(Work.joins("INNER JOIN creatorships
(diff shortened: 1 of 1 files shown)


app/controllers/users_controller.rb
```diff
@@ -281,14 +281,23 @@ def visible_items(current_user)
...
-                     .joins(:direct_filter_taggings)
-                     .joins("INNER JOIN works ON filter_taggings.filterable_id = works.id AND filter_taggings.filterable_type = 'Work'")
-                     .group('tags.id')
-                     .merge(Work.send(visible_method).revealed.non_anon)
-                     .merge(Work.joins("INNER JOIN creatorships ON creatorships.creation_id = works.id AND creatorships.creation_type = 'Work'
-  INNER JOIN pseuds ON creatorships.pseud_id = pseuds.id
-  INNER JOIN users ON pseuds.user_id = users.id").where('users.id = ?', @user.id))
+    @fandoms = if @user == User.orphan_account
+                 []
+               else
```
sarken: Lines 287-299 are just spacing and line length changes.

PR #3582, opened 2019-06-03
AO3-5699 Pass works instead of whole response to claim notification method

ariana-paris:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5699

## Purpose

Fix a blocking bug in the mass import API

## Testing Instructions

See Jira

### DIFF
--- a/app/controllers/api/v2/works_controller.rb
+++ b/app/controllers/api/v2/works_controller.rb
@@ -34,17 +34,17 @@ class Api::V2::WorksController < Api::V2::BaseController
     if status == :ok
       # Process the works, updating the flags
       works_responses = external_works.map { |external_work| import_work(archivist, external_work.merge(params.permit!)) }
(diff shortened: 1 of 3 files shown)


app/controllers/api/v2/works_controller.rb
```diff
@@ -34,17 +34,17 @@ def create
     if status == :ok
       # Process the works, updating the flags
       works_responses = external_works.map { |external_work| import_work(archivist, external_work.merge(params.permit!)) }
-      success_works, error_works = works_responses.partition { |r| [:ok, :created, :found].include?(r[:status]) }
+      success_reponses, error_responses = works_responses.partition { |r| [:ok, :created, :found].include?(r[:status]) }
```
redsummernight: This should be "success_responses".

ariana-paris: So it should. It wasn't the cause of the failing test, but I've sorted that out too, so this is ready to go!

PR #3583, opened 2019-06-07
AO3-3953 Expire caches of work blurbs, metas, and links on tag pages

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-3953

## Purpose

Expire caches that can unwanted contain collection-scoped tag links (#3454).

## Testing Instructions

See issue.

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -270,7 +270,7 @@ class Work < ApplicationRecord
   end
 
   def self.work_blurb_tag_cache_key(id)
-    "/v1/work_blurb_tag_cache_key/#{id}"
+    "/v2/work_blurb_tag_cache_key/#{id}"
   end
 
   def self.work_blurb_tag_cache(id)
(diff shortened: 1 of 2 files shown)


app/models/work.rb
```diff
@@ -270,7 +270,7 @@ def update_tag_index
   end
 
   def self.work_blurb_tag_cache_key(id)
-    "/v1/work_blurb_tag_cache_key/#{id}"
+    "/v2/work_blurb_tag_cache_key/#{id}"
```
sarken: Should we be expiring the view for the blurb and meta, too? Or will this force them to update? 

https://github.com/otwcode/otwarchive/blob/bd4f323003a7f63d0e708ff334fd666e4760601a/app/views/works/_work_module.html.erb#L27

https://github.com/otwcode/otwarchive/blob/bd4f323003a7f63d0e708ff334fd666e4760601a/app/views/works/_meta.html.erb#L1

redsummernight: Those will be updated, as they use `Work.work_blurb_tag_cache`:

https://github.com/otwcode/otwarchive/blob/0ec885aac78cfa16027d0e672649300c34b55e06/app/models/work.rb#L276-L278

which itself is a cached value, whose key we just changed.

If we need to expire just the blurbs, or just the metas, but not both, then we would change the keys in the views.

sarken: If I had actually _read_ the lines in the views beyond the word "cache", I would have seen that. 🤦‍♂

PR #3584, opened 2019-06-07
AO3-5652 Index works with their series titles

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5652

## Purpose

If a work belongs to series, include the series names while indexing the work.

## Testing Instructions

See issue.

### DIFF
--- a/app/models/serial_work.rb
+++ b/app/models/serial_work.rb
@@ -7,8 +7,8 @@ class SerialWork < ApplicationRecord
   after_create :adjust_series_visibility
   after_destroy :adjust_series_visibility
   after_destroy :delete_empty_series
-  after_create :update_series_index
-  after_destroy :update_series_index
+  after_create :update_series_index, :update_work_index
+  after_destroy :update_series_index, :update_work_index
 
   scope :in_order, -> { order(:position) }
 
@@ -22,6 +22,11 @@ class SerialWork < ApplicationRecord
     if self.series.present? && self.series.serial_works.blank?
       self.series.destroy
     end
+  end
+
+  # Reindex works added to or removed from a series
+  def update_work_index
+    work&.enqueue_to_index
   end
 
   # Ensure series bookmarks are reindexed when a new work is added to a series
--- a/app/models/series.rb
+++ b/app/models/series.rb
@@ -29,6 +29,7 @@ class Series < ApplicationRecord
 
   after_create :notify_after_creation
   before_update :notify_before_update
+  after_update :update_work_index
 
   # return title.html_safe to overcome escaping done by sanitiser
   def title
@@ -243,6 +244,10 @@ class Series < ApplicationRecord
       bookmarkable_type: 'Series',
       bookmarkable_join: { name: "bookmarkable" }
     )
+  end
+
+  def update_work_index
+    self.works.each(&:enqueue_to_index) if saved_change_to_title?
   end
 
   def word_count
--- a/spec/models/search/work_search_form_spec.rb
+++ b/spec/models/search/work_search_form_spec.rb
@@ -145,6 +145,47 @@ describe WorkSearchForm do
       end
     end
 
+    describe "when searching by series title" do
+      before do
+        work.series_attributes = { title: "Megami Tensei" }
+        work.save!
+
+        second_work.series_attributes = { title: "Persona" }
+        second_work.save!
+
+        run_all_indexing_jobs
+      end
+
+      it "returns only works in series matching the title" do
+        results = WorkSearchForm.new(series_titles: "megami").search_results
+        expect(results).to include(work)
+        expect(results).not_to include(second_work)
+        results = WorkSearchForm.new(query: "series_titles: *").search_results
+        expect(results).to include(work, second_work)
+
+        # Rename a series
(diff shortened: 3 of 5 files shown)


redsummernight:

app/models/serial_work.rb
```diff
@@ -24,6 +24,11 @@ def delete_empty_series
     end
   end
 
+  # Reindex works added to or removed from a series
+  def update_work_index
+    work&.enqueue_to_index
```
CristinaRO: When is the work potentially nil? Does this differ whether the work is being deleted and the serial work follows, as opposed to the work being removed from a series (but not deleted)?

spec/models/search/work_search_form_spec.rb
```diff
@@ -145,6 +145,61 @@
...
     end
 
+    describe "when searching by series title" do
+      let!(:main_series) { create(:series, title: "Persona: Dancing in Starlight", works: [work]) }
+      let!(:spinoff_series) { create(:series, title: "Persona 5", works: [second_work]) }
+
+      it "returns only works in matching series" do
+        run_all_indexing_jobs
+
+        results = WorkSearchForm.new(series_titles: "dancing").search_results
```
CristinaRO: I would find the spec easier to read if each set of expectations had its own context, to make it clear what aspect of it we're testing, such as "when using a query", "when using a wildcard" etc. It also makes it easier to track where it breaks, if we ever get spec failures.

spec/models/search/work_search_form_spec.rb
```diff
@@ -145,6 +145,61 @@
...
+      it "returns only works in matching series after a series is renamed" do
+        main_series.update!(title: "Megami Tensei")
+        run_all_indexing_jobs
+
+        results = WorkSearchForm.new(series_titles: "megami").search_results
+        expect(results).to include(work)
+        expect(results).not_to include(second_work)
+      end
+
+      it "returns only works in matching series after a works is removed from a series" do
```
CristinaRO: Typo: "a works"

app/models/series.rb
```diff
@@ -39,6 +39,7 @@ def title
     too_long: ts("must be less than %{max} letters long.", max: ArchiveConfig.NOTES_MAX)
 
   after_save :adjust_restricted
+  after_update_commit :update_work_index
```
CristinaRO: Today I Learned about `after_update_commit` 🤓💥

app/models/series.rb
```diff
@@ -215,6 +216,10 @@ def bookmarkable_json
     )
   end
 
+  def update_work_index
+    self.works.each(&:enqueue_to_index) if saved_change_to_title?
```
CristinaRO: Today I Learned about `saved_change_to_attribute?` 😲

app/models/serial_work.rb
redsummernight: When the work is destroyed, the associated serial works should be destroyed first, so work should be not nil in that case.

https://github.com/otwcode/otwarchive/blob/7c6681c62bab9bbfac6fc1a0ca3867e4247309b6/app/models/work.rb#L25

I'm actually not sure when or how work can be nil, but we now test that serial works can be destroyed properly even when its work is nil (#3483), so we need this nil check to avoid breaking the test.

spec/models/search/work_search_form_spec.rb
redsummernight: I split the specs, the new output is:
```
WorkSearchForm
  searching
    when searching by series title
      using the "series_titles" field
        returns only works in matching series
        returns only works in matching series with numbers in titles
        returns all works in series for wildcard queries
      using the "query" field
        returns only works in matching series
        returns only works in matching series with numbers in titles
        returns all works in series for wildcard queries
      after a series is renamed
        returns only works in matching series
      after a work is removed from a series
        returns only works in matching series
      after a series is deleted
        returns only works in matching series
```

app/models/series.rb
redsummernight: We found that we should be using `after_commit` for ES callbacks (like #3300). Ideally. Eventually.

app/models/serial_work.rb
tickinginstant: It can be caused by a race condition. You can simulate it in the dev console by doing the following:

1. Create a work to experiment with:
```
FactoryGirl.create(:work)
```
2. Start a transaction to add the work to a brand new series, but use `sleep` to make sure that the transaction doesn't finish quickly
```
Work.transaction do
  User.current_user = Work.last.users.first
  Work.last.update(series_attributes: { title: "Brand New Series" })
  sleep 30
end
```
3. Before that transaction finishes, open a second window with a second instance of the dev console, and destroy the work:
```
Work.last.destroy
```
4. Wait for the both transactions to finish.
5. Check `SerialWork.last.work` and note that it's `nil`.

This isn't a perfect reflection of the steps that would trigger this in practice, but it's close enough to give an idea of how it could happen: If a work is added to a series and deleted at almost exactly the same time, there's a (very low) chance that the SQL queries could be interleaved to create a `SerialWork` with a `nil` work.

CristinaRO: Awesome, thank you! Race conditions are some of the trickiest to wrap my head around.

PR #3587, opened 2019-06-26
AO3-5184 Make Advanced section of skins toggleable again in Safari

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5184

## Purpose

In desktop Safari and on iOS browsers, you can't open the Advanced section of the skins edit page. This fixes that.

## Testing Instructions

Refer to Jira.

### DIFF
--- a/public/stylesheets/site/2.0/08-actions.css
+++ b/public/stylesheets/site/2.0/08-actions.css
@@ -150,6 +150,10 @@ ul#skiplinks, .landmark, .landmark a, .index .heading.landmark {
 
 /* CONTEXTS */
 
+legend .action:link {
+  position: static;
+}
+
 .heading .actions, .heading .action, .heading span.actions {
   height: auto;
   font: 100 75%/1.286 'Lucida Grande', 'Lucida Sans Unicode', 'GNU Unifont', Verdana, Helvetica, sans-serif;


niconicosette:
this looks good! 👍

public/stylesheets/site/2.0/08-actions.css
```diff
@@ -150,6 +150,10 @@ ul#skiplinks, .landmark, .landmark a, .index .heading.landmark {
 
 /* CONTEXTS */
 
+legend .action:link {
```
redsummernight: Why do we target unvisited links specifically? Is this a documented workaround?

sarken: No, there's no documented fix -- this is an us problem, not a widespread issue with links in legends. (If you disabled CSS in your browser, you can see the "Show" and "Hide" links appear just fine.) The selector had to include `:link` because otherwise it gets overridden by the `.action:link` style higher up in the cascade. I was thinking the link would never be `:visited`, but you're right -- I should include the other pseudo-classes, too, in case there was some JavaScript malfunction that means `:visited` applies. (What I actually _want_ to do at some point is change it from a link to a button, but that seemed outside the scope of the issue.)

sarken: Actually! I forced the pseudo classes with my browser's inspector to double check and `legend .action:link` appears to be the only one we need, because link pseudo classes are weird. `:visited` styles can only differ from `:link` styles in a very limited set of ways, and `position` is not one of them. ([More info](https://stackoverflow.com/a/10320628).)

PR #3588, opened 2019-06-29
AO3-5704 Add script which dumps stats

zz9pzza:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5704

## Purpose

This would allow us to deliver stats files that academics and fan statisticians information to think over. 

## Testing Instructions

Get systems to do the work so the files become visible.

### DIFF
--- a/script/dump_stats.rb
+++ b/script/dump_stats.rb
@@ -2,115 +2,103 @@
 #      RAILS_ENV=production bundle exec rails r script/dump_stats.rb
 # frozen_string_literal: true
 
-work_output = File.open(ENV['WORKS_FILENAME'] || '/tmp/works.csv', 'w')
-work_output.puts "Created_at,language,restricted,complete,word_count,tags\n"
-if ENV['QUICK'].nil?
-  stats_output = File.open(ENV['STATS_FILENAME'] || '/tmp/stats.csv', 'w')
-  stats_output.puts "count_of_days,to,from,day_of_week,tag_type,tag_ids,count"
-end
-first_time = Work.first.created_at
-days = 1
-end_date = DateTime.now.beginning_of_day
-start_date = end_date - 1.day
-loop do
-  w = Work.where(posted: true, created_at: start_date..end_date)
-  fandoms = {}
-  fandoms.default = 0
-  categories = {}
-  categories.default = 0
-  ratings = {}
-  ratings.default = 0
-  warnings = {}
-  warnings.default = 0
-  characters = {}
-  characters.default = 0
-  relationships = {}
-  relationships.default = 0
-  freeforms = {}
-  freeforms.default = 0
-  warnings_multi = {}
-  warnings_multi.default = 0
-  categories_multi = {}
-  categories_multi.default = 0
-  w.each do |work|
-    work_output.puts "#{work.created_at.strftime('%F')},#{work&.language&.short},#{work&.restricted},#{work&.complete},#{work&.word_count},#{work&.tags&.pluck(:id)&.join('+')}\n"
-    next unless ENV['QUICK'].nil?
+CSV.open(ENV['WORKS_FILENAME'] || '/tmp/works.csv', "wb", :write_headers => true, \
+         :headers => ["created_at", "language", "restricted", "complete", "word_count", "tags"]) do |work_csv|
+  CSV.open(ENV['STATS_FILENAME'] || '/tmp/stats.csv', "wb", :write_headers => true, \
+         :headers => ["count_of_days", "to,from", "day_of_week", "tag_type", "tag_ids", "count"]) do |stats_csv|
+    first_time = Work.first.created_at
+    days = 1
+    end_date = DateTime.now.beginning_of_day
+    start_date = end_date - 1.day
+    loop do
+      w = Work.where(posted: true, hidden_by_admin: false, in_unrevealed_collection: false, created_at: start_date..end_date)
+      fandoms = Hash.new(0)
+      categories = Hash.new(0)
+      ratings = Hash.new(0)
+      warnings = Hash.new(0)
+      characters = Hash.new(0)
+      relationships = Hash.new(0)
+      freeforms = Hash.new(0)
+      warnings_multi = Hash.new(0)
+      categories_multi = Hash.new(0)
+      puts "#{days},#{end_date},#{start_date},#{w.count}"
+      w.each do |work|
+        work_csv << [days,start_date.strftime('%F'), work&.language&.short, work&.restricted, work&.complete, \
+                work&.word_count, work&.tags&.pluck(:id)&.join('+')]
+        next unless ENV['QUICK'].nil?
 
-    warnings_multi[work&.warnings&.pluck(:id)&.sort&.join('+')] += 1
-    categories_multi[work&.category_ids&.sort&.join('+')] += 1
-    work&.fandoms&.pluck(:id)&.each do |id|
-      fandoms [id] += 1
-    end
-    work&.ratings&.pluck(:id)&.each do |id|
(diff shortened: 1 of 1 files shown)


script/dump_stats.rb
```diff
@@ -0,0 +1,107 @@
...
+
+work_output = File.open(ENV['WORKS_FILENAME'] || '/tmp/works.csv', 'w')
+work_output.puts "Created_at,language,restricted,complete,word_count,tags\n"
+if ENV['QUICK'].nil?
+  stats_output = File.open(ENV['STATS_FILENAME'] || '/tmp/stats.csv', 'w')
+  stats_output.puts "count_of_days,to,from,day_of_week,tag_type,tag_ids,count"
+end
+first_time = Work.first.created_at
+days = 1
+to = DateTime.now.beginning_of_day
```
sarken: The `to` and `from` keep tripping me up... would `end_date` and `start_date` be accurate variable names?

script/dump_stats.rb
```diff
@@ -0,0 +1,107 @@
...
+
+  days += 1
+  to = from
+  from -= 1.day
+  break if to < first_time
+end
+work_output.close
+stats_output.close unless ENV['QUICK'].nil?
+tag_output = File.open(ENV['TAGS_FILENAME'] || '/tmp/tags.csv', 'w')
+tag_output.puts "id,type,name,canonical,cached_count,adult,merger_id,unwrangleable\n"
```
sarken: `adult` seems unnecessary -- it only applies to the rating tags that make the adult content notice show up on works.

script/dump_stats.rb
```diff
@@ -0,0 +1,116 @@
...
+
+  days += 1
+  end_date = start_date
+  start_date -= 1.day
+  break if end_date < first_time
+end
+work_output.close
+stats_output.close unless ENV['QUICK'].nil?
+tag_output = File.open(ENV['TAGS_FILENAME'] || '/tmp/tags.csv', 'w')
+tag_output.puts "id,type,name,canonical,cached_count,merger_id,unwrangleable\n"
```
sarken: TW would prefer not to include `unwrangleable`, since that info is just for wranglers.

script/dump_stats.rb
```diff
@@ -0,0 +1,116 @@
+# 1. Run this script on production as follows:
+#      RAILS_ENV=production bundle exec rails r script/dump_stats.rb
+# frozen_string_literal: true
+
+work_output = File.open(ENV['WORKS_FILENAME'] || '/tmp/works.csv', 'w')
+work_output.puts "Created_at,language,restricted,complete,word_count,tags\n"
```
tickinginstant: Why is `Created_at` the only capitalized column name?

script/dump_stats.rb
```diff
@@ -0,0 +1,116 @@
...
+if ENV['QUICK'].nil?
+  stats_output = File.open(ENV['STATS_FILENAME'] || '/tmp/stats.csv', 'w')
+  stats_output.puts "count_of_days,to,from,day_of_week,tag_type,tag_ids,count"
+end
+first_time = Work.first.created_at
+days = 1
+end_date = DateTime.now.beginning_of_day
+start_date = end_date - 1.day
+loop do
+  w = Work.where(posted: true, created_at: start_date..end_date)
```
tickinginstant: Is it okay that this is going to include unrevealed works and works hidden by an admin?

script/dump_stats.rb
```diff
@@ -0,0 +1,116 @@
...
+  fandoms = {}
+  fandoms.default = 0
```
tickinginstant: If you want to make this a little more concise, I believe you should be able to replace the two lines initializing `fandoms` with a single line setting `fandoms = Hash.new(0)`. (And ditto with all the other hashes.)

script/dump_stats.rb
```diff
@@ -0,0 +1,116 @@
...
+  relationships = {}
+  relationships.default = 0
+  freeforms = {}
+  freeforms.default = 0
+  warnings_multi = {}
+  warnings_multi.default = 0
+  categories_multi = {}
+  categories_multi.default = 0
+  w.each do |work|
+    work_output.puts "#{work.created_at.strftime('%F')},#{work&.language&.short},#{work&.restricted},#{work&.complete},#{work&.word_count},#{work&.tags&.pluck(:id)&.join('+')}\n"
```
tickinginstant: I don't think you need safe navigation on the work here. You should be able to replace `work&.` with `work.` throughout the loop. (In fact, I think the only place you need safe navigation in this loop is for `work.language&.short`.)

script/dump_stats.rb
```diff
@@ -0,0 +1,116 @@
...
+    work&.warnings&.pluck(:id)&.each do |id|
+      warnings [id] += 1
+    end
+    work&.characters&.pluck(:id)&.each do |id|
+      characters [id] += 1
+    end
+    work&.freeforms&.pluck(:id)&.each do |id|
+      freeforms [id] += 1
+    end
+    work&.relationship_ids&.each do |id|
```
tickinginstant: [This will return filter IDs, not tag IDs.](https://github.com/otwcode/otwarchive/blob/dd6ca1180f875ae8def2b21b7b78adc68c805c1f/lib/taggable.rb#L295-L297)

(Ditto with any calls to `work.category_ids`, but because all of the categories are supposed to be canonical, the tag IDs ought to be the same as the filter IDs in most cases. Though you might want to change them anyway, just for consistency.)

script/dump_stats.rb
```diff
@@ -0,0 +1,116 @@
...
+      relationships [id] += 1
+    end
+    work&.category_ids&.each do |id|
+      categories [id] += 1
+    end
+  end
+  next unless ENV['QUICK'].nil?
```
tickinginstant: Doesn't this mean that if the `QUICK` setting is enabled, the `break if end_date < first_time` is unreachable, and the loop will never terminate?

script/dump_stats.rb
```diff
@@ -0,0 +1,116 @@
...
+  end
+  stats_output.flush
+
+  days += 1
+  end_date = start_date
+  start_date -= 1.day
+  break if end_date < first_time
+end
+work_output.close
+stats_output.close unless ENV['QUICK'].nil?
```
tickinginstant: I think the logic here is backwards, since `stats_output` is only initialized if `ENV['QUICK'].nil?`.

script/dump_stats.rb
```diff
@@ -0,0 +1,116 @@
...
+tag_output = File.open(ENV['TAGS_FILENAME'] || '/tmp/tags.csv', 'w')
+tag_output.puts "id,type,name,canonical,cached_count,merger_id\n"
+Tag.find_in_batches do |batch|
+  batch.each do |tag|
+    tag_output.puts "#{tag.id},#{tag.type}," \
+    "#{
+    if tag.taggings_count_cache <= (ENV['TAGS_REDACTED_COUNT'] || 5) && !tag.canonical
```
tickinginstant: I agree with Hound here. Having this chunk of code inside the string interpolation makes it harder to read. Maybe you can assign the results of the `if` statement to a variable outside the string interpolation, and use the variable inside the string interpolation?

script/dump_stats.rb
```diff
@@ -0,0 +1,116 @@
...
+      "Redacted"
+    else
+      tag.name
```
tickinginstant: Since tag names can include quotes, and quotes are a special character in CSV files, I think you either need to escape the tag name in some way, or [use a library designed for handling CSV](https://ruby-doc.org/stdlib-2.6.1/libdoc/csv/rdoc/CSV.html).

PR #3590, opened 2019-07-06
AO3-4571 Fix challenge assignment equality check.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4571

## Purpose

When the work form is trying to figure out which assignment checkboxes ought to be checked by default, it checks whether each of the user's assignments is included in the work's assignments. This inclusion check calls the `ChallengeAssignment` class's equality check.

Because the `ChallengeAssignment` class includes `Comparable` and defines a comparison function `<=>`, the comparison function overrides the default rails equality check.  But the `<=>` function only checks the offering/recipients pseuds, not the assignment ID or the challenge name. This means that if a user has the same assigned recipient in multiple challenges, all of the assignments are treated as "equivalent" by the work form.

This PR removes the unnecessary comparison function, so that the work form will use rails' built-in equality check (which relies on IDs) to determine which assignments to check by default.

## Testing Instructions

See JIRA.

### DIFF
--- a/app/models/challenge_assignment.rb
+++ b/app/models/challenge_assignment.rb
@@ -166,7 +166,6 @@ class ChallengeAssignment < ApplicationRecord
   end
   alias_method :defaulted?, :defaulted
 
-  include Comparable
   # sort in order that puts assignments with no request ahead of assignments with no offer,
   # ahead of assignments with both request and offer, and within each group sorts by
   # request byline and then offer byline.
--- a/app/models/challenge_assignment.rb
+++ b/app/models/challenge_assignment.rb
@@ -166,19 +166,6 @@ class ChallengeAssignment < ApplicationRecord
   end
   alias_method :defaulted?, :defaulted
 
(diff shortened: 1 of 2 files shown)


redsummernight:
Note that test failures are for dA imports, this PR is fine.

CristinaRO:
Silly question: how are we sure the comparison method is unnecessary? It seems like a big thing to implement without using it anywhere. :O

tickinginstant:
I believe it was useful at one point, since a little poking around shows that assignments used to be loaded from the database in an arbitrary order, and then ordered in Ruby with a call to `sort`. But now all of the ordering takes place in the SQL query.

(To double-check, I separated the change into two commits so that I could check codecov after removing `include Comparable`. Any call to `sort` involving multiple assignments would have triggered a call to `<=>` even without `Comparable`. But as soon as I removed `Comparable`, the function stopped being covered. The only place it was being used was for the equality check.)

PR #3591, opened 2019-07-06
AO3-5549 Handle deleted users in spam alert.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5549

## Purpose

This PR updates the spam report email so that it won't generate an error when a user's account is deleted after the spam report is generated, but before the email with that report is sent. Instead, it silently omits any missing users (and aborts the entire email if all listed users have been deleted).

## Testing Instructions

See JIRA.

### DIFF
--- a/app/mailers/admin_mailer.rb
+++ b/app/mailers/admin_mailer.rb
@@ -1,4 +1,4 @@
-class AdminMailer < BulletproofMailer::Base
+class AdminMailer < ActionMailer::Base
   include Resque::Mailer # see README in this directory
 
   layout 'mailer'
@@ -48,7 +48,7 @@ class AdminMailer < BulletproofMailer::Base
   # Sends a spam report
   def send_spam_alert(spam)
     @users = User.where(id: spam.keys).to_a
-    abort_delivery if @users.empty?
+    return if @users.empty?
 
     # Make sure that the keys of the spam array are integers, so that we can do
     # an easy look-up with user IDs.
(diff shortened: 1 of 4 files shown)


app/mailers/admin_mailer.rb
```diff
@@ -47,7 +47,17 @@ def edited_comment_notification(comment_id)
 
   # Sends a spam report
   def send_spam_alert(spam)
-    @spam = spam
+    @users = User.where(id: spam.keys).to_a
+    abort_delivery if @users.empty?
```
redsummernight: In Rails 5, can we call `mail` conditionally instead of using `BulletproofMailer`? https://stackoverflow.com/a/13276740

tickinginstant: Ooh, good point! Looks like the `BulletproofMailer` workaround is no longer necessary.

redsummernight: Can you add a TODO remove at the top of `BulletproofMailer`?

tickinginstant: It's still used for `UserMailer#batch_subscription_notification`, so it can't be removed immediately. Can I maybe add a bug report for removing it and updating the subscription mailer, instead?

redsummernight: Please and thank you, that's better than a TODO.

tickinginstant: I added the issue: https://otwarchive.atlassian.net/browse/AO3-5708

redsummernight: Thanks!

PR #3595, opened 2019-07-22
AO3-5686 Fix Brakeman warnings for file access

zz9pzza:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5686

## Purpose

What does this PR do?

## Testing Instructions

Make a change to a skin, cache it an check that the changes are made.

Caching skins needs to be done on the front end machine(s).

So needs console access.

### DIFF
--- a/app/models/skin.rb
+++ b/app/models/skin.rb
@@ -2,6 +2,7 @@ require 'fileutils'
 include HtmlCleaner
 include CssCleaner
 include SkinCacheHelper
+include SkinFileHelper
 include SkinWizard
 
 class Skin < ApplicationRecord
@@ -22,9 +23,6 @@ class Skin < ApplicationRecord
   DEFAULT_ROLE = "user"
   DEFAULT_ROLES_TO_INCLUDE = %w(user override site)
   DEFAULT_MEDIA = ["all"]
-
-  SKIN_PATH = 'stylesheets/skins/'
-  SITE_SKIN_PATH = 'stylesheets/site/'
 
   belongs_to :author, class_name: 'User'
   has_many :preferences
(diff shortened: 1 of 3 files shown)


redsummernight (commented):
Do the methods in the new helper class do things differently from before? If not, should they trip the Brakeman checks?

tickinginstant:
What about this file access warning? Is it supposed to be included or are you deliberately only doing the skin-related ones?

https://github.com/otwcode/otwarchive/blob/ebf2c63c33b03877b287dc64ecd3db1401ca9b50/config/brakeman.ignore#L482-L501

zz9pzza:
I looked at it and I was convinced that it was not an issue. And that the users would get grumpy if we set the filepath to be random.  I would prefer we used a random filename and sent a user path back to the web server but I thought it wasn't necessary on this iteration. If there is an obvious solution to you I would be happy to hear.

zz9pzza:
> Do the methods in the new helper class do things differently from before? If not, should they trip the Brakeman checks?

They have the same functionality but do not annoy brakeman.

redsummernight:
> They have the same functionality but do not annoy brakeman.

Since we're not really changing anything, do we need to? We could just leave the warnings in brakeman.ignore.

zz9pzza:
> > They have the same functionality but do not annoy brakeman.
> 
> Since we're not really changing anything, do we need to? We could just leave the warnings in brakeman.ignore.

The whole point of this series of pulls is not to change functionality. Just to reduce the incidents of brakeman warnings as much as we can.

redsummernight:
With the previous PRs for example, `eval` is unsafe, so you changed things to stop using them, it's a clear necessary change.

With this one, you're not changing or avoiding file access. You're just moving code to a different place to avoid Brakeman's complaints. I don't think that's different from having them in the ignore file.

PR #3596, opened 2019-07-26
AO3-5658 Add creator invitations, and a preference option to disable them.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5658

This also includes changes that may or may not fix the following issues:
https://otwarchive.atlassian.net/browse/AO3-4998
https://otwarchive.atlassian.net/browse/AO3-5061
https://otwarchive.atlassian.net/browse/AO3-5261
https://otwarchive.atlassian.net/browse/AO3-5475
https://otwarchive.atlassian.net/browse/AO3-5606
https://otwarchive.atlassian.net/browse/AO3-5625
https://otwarchive.atlassian.net/browse/AO3-5659
https://otwarchive.atlassian.net/browse/AO3-5662
https://otwarchive.atlassian.net/browse/AO3-5663
https://otwarchive.atlassian.net/browse/AO3-5692

## Purpose

See JIRA.

## Testing Instructions

See JIRA.


redsummernight:
I have a fix for [AO3-5691](https://otwarchive.atlassian.net/browse/AO3-5691), changing:

https://github.com/otwcode/otwarchive/blob/854a4926c67c0419b10b1a4fe2165d8ccb8e9d87/app/controllers/works_controller.rb#L384

into

```
@chapter.errors.full_messages.each { |err| @work.errors.add(:base, err) }
```

will get rid of the "translation missing" error and also make testing [AO3-5625](https://otwarchive.atlassian.net/browse/AO3-5625) easier. I can pull that separately.

----

What a behemoth of a PR, good work from all involved. :D

GitHub's "[mark files as viewed](https://github.blog/2019-07-01-mark-files-as-viewed/)" got here just in time.

tickinginstant:
Actually, now that I look at it, I was mistaken -- I didn't fix AO3-5691. I just assumed I did because the behavior on AO3-5625 doesn't result in a translation missing error anymore, and I knew that the two were linked. (I'll edit the list of issues to remove that one.)

I don't think including AO3-5691 will affect testing AO3-5625, but if you want to toss it in it probably won't result in conflicts.

app/controllers/pseuds_controller.rb
```diff
@@ -36,31 +36,24 @@ def show
 
     # very similar to show under users - if you change something here, change it there too
     if !(logged_in? || logged_in_as_admin?)
-      # hahaha omg so ugly BUT IT WORKS :P
```
redsummernight: Nice!

app/models/chapter.rb
```diff
@@ -80,11 +79,9 @@ def fix_positions
   after_save :invalidate_chapter_count,
     if: Proc.new { |chapter| chapter.saved_change_to_posted? }
 
-  after_save :expire_cache_on_coauthor_removal
-
   before_destroy :fix_positions_after_destroy, :invalidate_chapter_count
   def fix_positions_after_destroy
```
redsummernight: Should we rename to `fix_positions_before_destroy`?

app/models/chapter.rb
```diff
@@ -80,11 +79,9 @@ def fix_positions
...
-    if work && position
+    if work&.persisted? && position
```
redsummernight: Is this because work now has `has_many :chapters, inverse_of: :work, autosave: true`?

features/step_definitions/user_steps.rb
```diff
@@ -88,6 +88,20 @@
   step %{confirmation emails have been delivered}
 end
 
+Given(/^the user "([^"]*)" disallows co-creators$/) do |login|
```
sarken: Should these be in https://github.com/otwcode/otwarchive/blob/master/features/step_definitions/preferences_steps.rb with the other steps for setting preferences instead?

app/models/tagging.rb
```diff
@@ -3,7 +3,6 @@ class Tagging < ApplicationRecord
   belongs_to :taggable, polymorphic: true, touch: true
 
   validates_presence_of :tagger
-  validates_associated :taggable
```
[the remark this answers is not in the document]
tickinginstant: This code didn't originally validate `taggable`. In the Rails 4.0 update, this changed from
```
validates_presence_of :tagger, :taggable
```
to
```
validates_presence_of :tagger
validates_associated :taggable
```

I'm guessing the reason the `validates_presence_of :taggable` check was disabled is because the inverses aren't set up properly on taggings, which meant that the validation always failed on new works. Though I'm not sure why the check was changed to `validates_associated` (instead of trying to fix the inverses or deleting it entirely).

The issue with `validates_associated` is that it will load a fresh copy of the taggable object from the database (because broken inverses) and try to validate it. But that fresh copy will be missing any info that isn't in the database when the `Tagging` is being saved. When saving the work for the first time, this includes almost all associations, because `Taggable` is the first module included in works, and I'm pretty sure associations are saved in the order that they're declared. In particular, the version of the work that it validates during work creation will always be missing all creatorship information, making it look like the work has no creators.

(It works on master because any validity checks involving the presence of associations are coded up as `before_save` hooks, instead of `validate` methods, so that they're not called when `validates_associated` tries to check it. But that causes its own issue: it won't display the errors until the user actually tries to save the work. So if you want those issues to be flagged when the user is previewing, you have to add code to the `WorksController` for each validation that's implemented as a `before_save` hook. For example, `WorksController#set_work_tag_error_messages` vs. `Taggable#validate_tags`. This results in the `WorksController` being more complex than it ought to be.)

The obvious solution to this is to fix up the inverses. Unfortunately, fixing the inverses is more complicated than it sounds because of [things like this](https://github.com/otwcode/otwarchive/pull/3326#issuecomment-394096066):
>Works avoid this problem because the before_destroy hook in the Tagging class runs remove_filter_tagging an additional time for each deleted tag. **Because the taggable association in Taggings isn't set up as the inverse of the taggings association in Works,** the call from the Tagging class loads a fresh copy of the work for every tagging deleted or added. So the first time remove_filter_tagging is run -- from the Tagging class -- **the tags association is loaded fresh, resulting in more SQL queries and processing time, but also the correct outcome.** The second time remove_filter_tagging is run -- from the Taggable class -- it's a no-op. And there's no equivalent of check_filter_taggings.

Basically: Using `validates_associated` while the inverses are broken means that validity of `taggable` may not be the same as the validity of the work whose tags are being edited. So I don't think it's a very useful check at the moment. But fixing the inverses will break the calculation of filters, and including the fix for that would make this PR even more unmanageable. In my opinion, the cleanest way to handle this is to delete the `validates_associated`.

features/importing/archivist.feature
```diff
@@ -81,8 +81,7 @@ Feature: Archivist bulk imports
       | user1 | a@ao3.org |
     When I go to the import page
       And I import the work "http://ao3testing.dreamwidth.org/593.html" by "name1" with email "a@ao3.org" and by "name2" with email "b@ao3.org"
-    When the system processes jobs
```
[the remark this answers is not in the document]
tickinginstant: Yeah, I don't see any reason it needs to stay.

lib/pseudable.rb
```diff
@@ -1,55 +1,18 @@
 # For models which have pseuds (authors)
 module Pseudable
```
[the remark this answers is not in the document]
tickinginstant: My inclination is to move it to HasCreatorships, since it's basically an alternate version of `new_bylines=`. (If I do move it there, do you think I should still cite which module the `new_bylines=` function comes from? Since it'd be in the same module then.)

lib/creatable.rb
```diff
@@ -1,15 +1,13 @@
 module Creatable
```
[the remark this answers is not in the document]
tickinginstant: Ooh, good idea.

lib/tasks/creatorships.rake
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+namespace :creatorships do
+  desc "Clean up creatorships for deleted chapters"
```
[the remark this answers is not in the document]
tickinginstant: I added clean-up tasks for series and chapters because they currently don't have `dependent: :destroy` set on their creatorships, while works do. So while I'm positive that there are plenty of lingering creatorships for deleted chapters and deleted series, I suspect that there are relatively few lingering creatorships for deleted works.

But I guess I'll add the task regardless. It doesn't hurt to add it, and it might be helpful if there are more issues with works than I suspected.

lib/tasks/creatorships.rake
```diff
@@ -0,0 +1,35 @@
...
+  task(remove_deleted_series_creatorships: :environment) do
+    Creatorship.joins("LEFT JOIN series ON " \
+                      "creatorships.creation_id = series.id AND " \
+                      "creatorships.creation_type = \"Series\"").
+      where(series: { id: nil }).
+      in_batches.delete_all
+  end
+
+  desc "Add missing series creatorships"
+  task(add_missing_series_creatorships: :environment) do
```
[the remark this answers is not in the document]
tickinginstant: It doesn't matter which one runs first. The `add_missing_series_creatorships` task can only affect the creatorships of series that have at least one `SerialWork`. Whereas `remove_orphaned_empty_series` does a left-join to ensure that it only affects series that have no `SerialWork`s. There should be no overlap in the series that they affect.

spec/miscellaneous/lib/works_owner_spec.rb
```diff
@@ -81,6 +81,8 @@
       end
 
       it "should change after a work is orphaned" do
+        # Ensure that the orphan account exists:
+        build(:user, login: "orphan_account").save
```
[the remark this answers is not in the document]
tickinginstant: Unfortunately, no. It does try to create it:
https://github.com/otwcode/otwarchive/blob/854a4926c67c0419b10b1a4fe2165d8ccb8e9d87/app/models/user.rb#L547-L553
But because the record that it builds is missing a lot of basic info (like email and password), the new user doesn't get saved.

features/support/hooks.rb
```diff
@@ -1,4 +1,7 @@
 Before do
+  # Reset the current user:
+  User.current_user = nil
```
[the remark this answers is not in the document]
tickinginstant: The issue that I ran into was creating works/series/chapters with `FactoryGirl`. It would treat all of the creatorships as if `User.current_user` was inviting new co-creators, and as a result either (a) the creatorship wouldn't be saved because it was "inviting" a user who had co-creation disabled, or (b) the creatorship would be treated as an invitation. In either case, the work/series/chapters would have no approved creators, and would fail to save.

There might be a way to handle the issue more consistently by temporarily overriding `User.current_user` in the factory, but I'm not sure how to do that.

app/models/chapter.rb
tickinginstant: Sort of. I added it because at the time I was worried that the inverses would break deleting works, but after some poking, I'm not sure it's necessary. According to the Rails docs, the inverse name can be inferred if it matches the name of the class that it's defined in, so the inverse is already working, and `inverse_of: :work` is pretty much just a formality. The only benefit of checking `persisted?` at this point is that it decreases the number of SQL queries required during work deletion.

I can revert the change, if you'd prefer.

app/models/tagging.rb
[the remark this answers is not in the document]
redsummernight: Thanks for explaining this! Leaving out this validation sounds good.

lib/pseudable.rb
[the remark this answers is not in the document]
redsummernight: If it ends up in the same place as `new_bylines=` you don't have to mention the module.

spec/miscellaneous/lib/works_owner_spec.rb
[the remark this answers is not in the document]
redsummernight: I see how it works in feature tests now:

https://github.com/otwcode/otwarchive/blob/854a4926c67c0419b10b1a4fe2165d8ccb8e9d87/features/step_definitions/user_steps.rb#L11-L14

Not ideal, but we don't have to deal with this now.

features/support/hooks.rb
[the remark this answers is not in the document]
redsummernight: Setting it to nil in a before hook is fine, we already do the same for specs. We should have done this anyway, even without the new creatorship specs breaking.

app/models/chapter.rb
redsummernight: Fewer SQL queries sound good, let's keep it.

features/importing/archivist.feature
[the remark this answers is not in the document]
redsummernight: I made [AO3-5722](https://otwarchive.atlassian.net/browse/AO3-5722).

lib/tasks/creatorships.rake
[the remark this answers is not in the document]
redsummernight: I just checked:
```
2.3.4 :003 > Creatorship.joins("LEFT JOIN works ON creatorships.creation_id = works.id AND creatorships.creation_type = \"Work\"").where(works: { id: nil }).count
 => 18482298
```
We'll need the task.

app/models/pseud.rb
```diff
@@ -259,28 +249,37 @@ def self.parse_byline(byline, options = {})
...
+        next
       end
-      if pseuds.length == 1
-        valid_pseuds << pseuds.first
-      elsif pseuds.length > 1
+      if pseuds.size > 1
         ambiguous_pseuds[pseuds.first.name] = pseuds
-      else
-        failures << byline.strip
+        remove_pseuds = options[:remove_ambiguous]
```
redsummernight: Where is this option used?

app/models/creatorship.rb
```diff
@@ -1,47 +1,276 @@
...
+
+  after_commit :update_indices
+
+  after_create_commit :notify_creator, if: :enable_notifications
+
+  # If a pseud is listed as a work co-creator (not invited, actually listed),
+  # they should also be listed on all of the work's series. Similarly, if a
+  # pseud is listed as a chapter co-creator, they should also be listed on the
+  # work.
+  def add_to_parents
```
redsummernight: We don't remove from parents right? e.g. if a co-creator is removed from all chapters / works in a series, they should still be a co-creator in the work / series?

app/models/work.rb
```diff
@@ -1,35 +1,26 @@
...
   ########################################################################
   # ASSOCIATIONS
   ########################################################################
 
-  # creatorships can't have dependent => destroy because we email the
-  # user in a before_destroy callback
-  has_many :creatorships, as: :creation
-  has_many :pseuds,
-           through: :creatorships,
-           after_remove: [:expire_pseud, :reindex_changed_pseud]
```
redsummernight: The method `reindex_changed_pseud` should be removed.

lib/creation_notifier.rb
```diff
@@ -1,15 +1,13 @@
-module Creatable
+module CreationNotifier
   def notify_after_creation # after_create
-    notify_co_authors
-    return unless !self.is_a?(Series) && self.posted?
```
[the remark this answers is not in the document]
tickinginstant: Yep, I removed the include from Series a while back. (It doesn't look like it, because it still says `include Creatable`. But that's because I renamed this module to `CreationNotifier` and renamed the `HasCreatorships` module to `Creatable`.)

factories/works.rb
```diff
@@ -9,16 +9,20 @@
...
-    after(:build) do |work|
-      work.authors = [FactoryGirl.build(:pseud)] if work.authors.blank?
+    transient do
+      authors { [FactoryGirl.build(:pseud)] }
     end
 
-    factory :no_authors do
-      after(:build) do |work|
-        work.authors = []
+    after(:build) do |work, evaluator|
```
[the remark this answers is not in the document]
tickinginstant: Yeah, I needed a way to assign pseuds in factories that was simple, but didn't use the `pseuds` association.

app/models/creatorship.rb
```diff
@@ -1,47 +1,276 @@
...
-  def expire_caches
-    if creation_type == 'Work' && self.pseud.present?
-      CacheMaster.record(creation_id, 'pseud', self.pseud_id)
-      CacheMaster.record(creation_id, 'user', self.pseud.user_id)
-    end
+  # Calculate whether this creatorship should count as approved, or whether
+  # it's just a creatorship invitation.
+  def should_automatically_approve?
+    # Approve if the current user has special permissions:
+    return true if User.current_user.nil? ||
```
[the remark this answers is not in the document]
tickinginstant: The works API requires an API key, but doesn't check whether the user is logged in:
https://github.com/otwcode/otwarchive/blob/1c43c2c1d7341986b5a1fc811112e4437d6f836c/app/controllers/api/v2/base_controller.rb#L17-L22

spec/controllers/creatorships_controller_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+
+  let(:pending) do
+    Creatorship.new(pseud: user.default_pseud, creation: pending_work,
+                    approved: false)
+  end
+
+  before do
+    # Make sure that the invitation is saved without altering the current value
+    # of approved:
+    pending.save(validate: false)
```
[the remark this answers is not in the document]
tickinginstant: Yes, that's right.

features/other_b/skin_public.feature
```diff
@@ -111,11 +111,11 @@ Feature: Public skins
 
   Scenario: A user can preview a cached public site skin, and it will take the 
   user to the works page for a canonical tag with between 10 and 20 works
-  Given the approved public skin "Usable Skin"
-    And the skin "Usable Skin" is cached
-    And the canonical fandom "Dallas" with 2 works
+  Given the canonical fandom "Dallas" with 2 works
     And the canonical fandom "Major Crimes" with 11 works
     And the canonical fandom "Rizzoli and Isles" with 21 works
+    And the approved public skin "Usable Skin"
```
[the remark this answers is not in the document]
tickinginstant: The `canonical fandom X with Y works` steps use `FactoryGirl` to create the works, which means that they need `User.current_user` to be nil in order to work properly. But the `approved_public_skin` step involves logging in, which sets `User.current_user`.

(If you'd prefer, I can try adding `And I am logged out` instead of reordering the steps. I haven't tested it, but it seems like it ought to work if you think it'd be cleaner.)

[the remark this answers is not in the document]
redsummernight: Yeah, I think we should leave these steps alone and add the log out step here instead:

https://github.com/otwcode/otwarchive/blob/fe59be31fb7ddd7c1e71742ea2f26e9bdb6781dd/features/step_definitions/skin_steps.rb#L75-L79

Some of the other skin steps have a similar problem but let's leave them be for now.

app/models/creatorship.rb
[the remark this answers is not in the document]
redsummernight: Can you add to the comment, "Approve if we're using an API key, or if the current user has special permissions:"?

app/models/creatorship.rb
tickinginstant: Right.

app/models/creatorship.rb
```diff
@@ -1,47 +1,276 @@
...
+    # If this is being deleted and it's a work, then its chapters are also going
+    # to be deleted (which will cause their creatorships to be deleted as
+    # well). If this is being deleted and it's a series, then we shouldn't
+    # delete the work creatorships.
+    return if creation.nil? || creation.destroyed?
+
+    children = if creation.is_a?(Work)
+                 creation.chapters.to_a
+               elsif creation.is_a?(Series)
+                 creation.works.to_a
```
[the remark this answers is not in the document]
tickinginstant: The comment was referring only to the line directly below it, as a reminder to myself about why it's okay to skip `remove_from_children` if the creation is in the process of being deleted. ("this" was supposed to refer to the creation, not the creatorship.) I've rephrased the comment to try to make it clearer.

app/models/work.rb
tickinginstant: Good catch! I also removed `expire_pseud`, since it's in the same boat.

features/other_a/pseud_delete.feature
```diff
@@ -90,3 +90,42 @@ Feature: Delete pseud.
     When I view the collection "PromptsGalore"
       And I follow "My Prompts"
     Then I should see "Antidisestablishmentarianism."
+
+  Scenario: Deleting a pseud should preserve creatorships and creator invites.
```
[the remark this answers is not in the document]
tickinginstant: This test was actually two cases in one (one for each of the works in the test):
- Making sure that approved creatorships are preserved if the default pseud has an invitation on the same work that the deleted pseud had an approved creatorship on. This makes it look like the invitation is being deleted, but it's effectively being accepted.
- Making sure that if a user has a creatorship invitation on a deleted pseud, it's transferred to the default pseud.

I've split it into two test cases to try to make it clearer.

features/other_a/pseud_delete.feature
```diff
@@ -90,3 +90,42 @@ Feature: Delete pseud.
...
+    Then I should see "Work was successfully posted."
+
+    When I set up the draft "Other Invited"
+      And I try to invite the co-author "other_pseud"
+      And I press "Post"
+    Then I should see "Work was successfully posted."
+
+    When I am logged in as "original_pseud"
+      And I go to my creator invitations page
+      And I check the 2nd checkbox with id matching "selected"
```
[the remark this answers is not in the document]
redsummernight: With the split tests, there should be only 2 here. Resolving.

app/models/pseud.rb
tickinginstant: It's not used, but I'm not certain I should remove it. The changes in `Pseud.parse_bylines` don't really fix anything (except for a few bugs involving banned pseuds). The bug that `parse_bylines` sort of addresses is the issue with adding ambiguous names as owners of a TagSet ([AO3-4867](https://otwarchive.atlassian.net/browse/AO3-4867)). But it doesn't actually fix AO3-4867, it just causes a different bad behavior: instead of being unable to add ambiguous pseuds, it'll add every user who has a pseud with that name.

In general, I think it's important to be able to exclude ambiguous pseuds from the list of valid pseuds. IMHO, they should always be removed, and the OwnedTagSet bug should be resolved by calling `parse_bylines` with the `assume_matching_login` option. (Or by writing more complex code to allow users to resolve the ambiguity.)

But IIRC, the change to `parse_bylines` was one of the things that @zz9pzza wanted me to include, so I'm not the right person to ask.

spec/miscellaneous/lib/works_owner_spec.rb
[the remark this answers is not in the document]
redsummernight: I made [AO3-5725](https://otwarchive.atlassian.net/browse/AO3-5725).

app/models/creatorship.rb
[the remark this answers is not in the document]
redsummernight: Gotcha. So if we're removing a co-creator from a series, then they should be removed from the series' works as well, but if we're deleting the series, then we don't touch the series' works' co-creators?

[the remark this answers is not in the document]
tickinginstant: Yeah. Every co-creator on the work is supposed to be listed as a co-creator on the series as well. [So if a co-creator is removed from the series, they should be removed from all works in the series.](https://github.com/otwcode/otwarchive/blob/c125f2eff02e6f271709a54665c3af0ca0b0c350/app/helpers/series_helper.rb#L79) But if deleting a series also removed all series co-creators from the work, we'd just end up trying to delete all of the work's creatorships. It's best to just leave the creatorships alone in that case, which matches the previous behavior.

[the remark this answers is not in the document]
redsummernight: :+1:

app/models/creatorship.rb
```diff
@@ -1,47 +1,278 @@
...
   end
 
+  ########################################
+  # OTHER METHODS
+  ########################################
+
+  attr_reader :ambiguous_pseuds
+
+  # We define a virtual "byline" attribute to make it easier to handle
+  # ambigous/missing pseuds. By storing the desired name in the @byline
```
redsummernight: Should be "ambiguous".

app/models/series.rb
```diff
@@ -156,37 +146,12 @@ def allpseuds
...
   # Remove a user as an author of this series
   def remove_author(author_to_remove)
-    pseuds_with_author_removed = self.pseuds - author_to_remove.pseuds
+    pseuds_with_author_removed = pseuds.where.not(user_id: author_to_remove.id)
     raise Exception.new("Sorry, we can't remove all authors of a series.") if pseuds_with_author_removed.empty?
-    Series.transaction do
-      self.pseuds = pseuds_with_author_removed
+    transaction do
+      creatorships.where(pseud: author_to_remove.pseuds).destroy_all
       authored_works_in_series = (author_to_remove.works & self.works)
```
redsummernight: Does this overlap with the `remove_from_children` callback in Creatorship? Is it because we want to use a transaction in this case?

I have a similar question for `Work.remove_author`.

app/models/pseud.rb
redsummernight: > instead of being unable to add ambiguous pseuds, it'll add every user who has a pseud with that name.

The new bad behavior sounds worse than the old one.

sarken: **Infinitely** worse and something we absolutely cannot deploy. The old bad behavior was annoying, but we could tell folks to work around it by entering `pseud (username)`. The new bad behavior runs the risk of the wrong person getting access to a tag set and deleting it.

tickinginstant: I'll set it to always remove ambiguous pseuds.

features/other_b/series.feature
```diff
@@ -179,18 +179,37 @@ Feature: Create and Edit Series
...
+    Then I should see "Series was successfully updated."
+      But I should not see "moon"
+    When the user "moon" accepts all co-creator invites
+    Then "moon" should be a creator of the series "Ponies"
     When I view the series "Ponies"
-      And I wait 1 second
       And I follow "Remove Me As Author"
+    Then I should see "Sorry, we can't remove all creators of a work."
+
+  Scenario: Removing self as co-author from co-authored series
```
sarken: Should be "co-creator"

features/other_b/series.feature
```diff
@@ -227,3 +246,65 @@ Feature: Create and Edit Series
...
       And I follow "Edit Series"
     Then I should see "What a title! :< :& :>" in the "Series Title" input
+
+  Scenario: You can edit a series to add someone as a co-creator if their preferences are set to permit it.
+    Given I am logged in as "foobar"
+      And the user "barbaz" exists and is activated
+      And the user "barbaz" allows co-creators
+      And I post the work "Behind her back she’s Gentleman Jack" as part of a series "Gentleman Jack"
+    When I view the series "Gentleman Jack"
+    And I follow "Edit Series"
```
sarken: Should be indented two more spaces.

features/users/user_delete.feature
```diff
@@ -94,3 +94,26 @@ Scenario: Delete a user who has coauthored a work
   When I go to the works page
   Then I should see "otheruser"
     And I should not see "testuser"
+
+  Scenario: Can delete a user who has an empty series
+    Given I am logged in as "testuser"
+    When I set up a draft "Invalid" with fandom "%" as part of a series "Empty"
+      And I press "Post"
+    Then I should see "Sorry! We couldn't save this work because:"
```
sarken: It took me a sec to realize these steps were here because trying and failing to create a work in a new series would result in an empty series being created. Could you put a comment here explaining that's what's going on?

features/works/chapter_edit.feature
```diff
@@ -274,45 +274,51 @@ Feature: Edit chapters
...
+    When the user "amy" accepts all co-creator invites
+      And I view the work "Forever Friends"
+    Then I should see "amy, karma"
+      And I should see "Chapter by karma"
+    When I follow "Next Chapter"
+    Then I should not see "Chapter by"
 
 
-  Scenario: You should be able to edit a chapter to add a co-creator who is
+  Scenario: You can edit a chapter to add (not invite) a co-creator who is
```
sarken: This sends a creatorship_notification email now, right? We should probably check for it in this test, if so.

app/models/series.rb
tickinginstant: There's one case that the creatorship callbacks in this PR don't handle at the moment: when there's a co-created work that has a chapter with only one creator, and that creator tries to remove themselves from the work (or from one of the work's series). The way that it works on Beta, the work doesn't complain about an attempt to remove chapter's last creator, but instead sets the chapter's creators equal to the work's (minus the user removing themselves).

That's what the two `remove_author` functions are for: `Work#remove_author` adds the other co-creators to the sole-authored chapters, and `Series#remove_author` calls `Work#remove_author` to handle any works in the series that have tricky chapter creatorships.

I added comments on the two functions to make it clear, refactored the code a little, and added some more tests to make sure that the code is covered.

(If you think it'd be cleaner, I can try changing the `check_not_last` callback to behave differently on chapters -- specifically, copy the work's creatorships at the time the last chapter creatorship is deleted. The downside would be that in some cases where a user has multiple pseuds on the same work, it'd end up saving creatorships and then immediately destroying them, which is kind of wasteful.)

app/models/pseud.rb
redsummernight: Looks good, thanks!

app/models/series.rb
redsummernight: Thanks for the explanation! Using `#remove_author` for this case sounds good to me, you don't have to modify `check_not_last`.

We just need to remember to always use `#remove_author` when we need to edit works or series.

For chapters, we don't need `#remove_authors`, is it because removing chapter co-creators doesn't affect works or series, and `check_no_creators` blocks if there are no creators left?

app/controllers/users_controller.rb
```diff
@@ -280,27 +266,19 @@ def visible_items(current_user)
...
-                     AND filter_taggings.filterable_type = 'Work'").
-                   group('tags.id').
-                   merge(Work.send(visible_method).revealed.non_anon).
-                   merge(Work.joins("INNER JOIN creatorships
-                     ON creatorships.creation_id = works.id
-                     AND creatorships.creation_type = 'Work'
-                     INNER JOIN pseuds ON creatorships.pseud_id = pseuds.id
-                     INNER JOIN users ON pseuds.user_id = users.id").
-                   where('users.id = ?', @user.id))
+                 Fandom.select("tags.*, count(DISTINCT works.id) as work_count").
```
redsummernight: This creates invalid SQL for me. As soon as we use `joins(:filtered_works)` or `.merge(visible_works)`, Rails adds `DISTINCT works.*` to the select.

tickinginstant: Are you running it on master, or in the creatorships branch? I cleaned up the definition of the user `works` association as part of this PR.

redsummernight: Ha, good catch. I ran it on master, without any of the other fixes.

app/models/concerns/creatable.rb
```diff
@@ -0,0 +1,216 @@
...
+  # Excludes creatorships with a missing pseud, because those orphaned
+  # creatorships can break various bits of code if they're considered valid.
+  def creatorships_after_saving
+    creatorships.select(&:valid?).reject(&:marked_for_destruction?).
+      reject { |creatorship| creatorship.pseud.nil? }
+  end
+
+  # Calculate what the pseuds on this work will be after saving, taking into
+  # account validity, approval, and @current_user_pseuds.
+  def pseuds_after_saving
```
sarken: Just going to throw this out for discussion: `creatorships_after_saving` is looking at all creatorships, regardless of approval status, but this method is only looking at pseuds from approved creatorships. Should it maybe be called `approved_pseuds_after_saving` for clarity? (Also kind of future-proofing, in case we ever need/want a method that gets all pseuds after saving regardless of approval status.)

app/models/creatorship.rb
```diff
@@ -1,47 +1,278 @@
...
+    end
+  end
+
+  # Make sure that if this is an invitation, we're not inviting someone who has
+  # disabled invitations.
+  def check_disallowed
+    return if approved? || pseud.nil?
+    return if pseud&.user&.preference&.allow_cocreator
+
+    errors.add(:base, ts("%{name} does not allow others to add them as a co-creator.",
```
sarken: Since we're only ever dealing with invitations here, maybe this error should be `%{name} does not allow others to invite them to be a co-creator`.

app/models/pseud.rb
```diff
@@ -335,51 +328,80 @@ def remove_stale_from_autocomplete_before_save
...
+        existing.update(approved: existing.approved || creatorship.approved)
+      else
+        creatorship.update(pseud: replacement)
+      end
+    end
+
+    # Update the pseud ID for all comments. Also updates the timestamp, so that
+    # the cache is invalidated and the pseud change will be visible.
+    Comment.where(pseud_id: self.id).update_all(pseud_id: replacement.id,
+                                                updated_at: Time.now)
```
sarken: 🎉 💃

app/models/series.rb
tickinginstant: Right, the error message is handled elsewhere and it doesn't trigger any creatorship changes for other objects.

app/models/concerns/creatable.rb
tickinginstant: The `pseuds` association only includes pseuds from approved creatorships. So I picked the name `pseuds_after_saving` because the function calculates which pseuds will be listed in the `pseuds` association after the creation is saved. (Which is similar to the behavior of `creatorships_after_saving`.)

sarken: Makes sense! 👍

PR #3602, opened 2019-08-02
AO3-5338 Work factories should create posted works by default

dlhardin:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5338 (please fill in issue number and remove this comment)

## Purpose

This changes the default output of the works factory to create posted works by default, rather than defaulting to drafts.

## Testing Instructions

Existing tests work, factory unchanged except for default.

## Credit

dlhardin she/her


dlhardin:
Thanks for letting me learn on your project. I've dug into this more, and I see where I need to replace the FactoryGirl calls to :posted_work with :work. But I see another potential problem, as there are a number of existing tests that call :work, with the old default behavior of creating draft works. 

Do these existing tests need to be altered to explicitly call :draft_work? It doesn't look like the new default behavior will break the tests, but if you're expecting a draft to result there's a problem.

redsummernight:
Thanks for taking this on!

Most of the existing tests that call `:work` should be using posted works. We do want the majority of the tests to use posted works by default. But yeah, there's a small number of them that actually need to use drafts.

Some of them are already using the draft factory:

https://github.com/otwcode/otwarchive/blob/1e369cdfb6c49e80b7290a3f1090ced7ecf14f2d/spec/miscellaneous/lib/tasks/work_tasks.rake_spec.rb#L6-L8

and some explicitly set posted to false (can be changed to `:draft` to be more concise):

https://github.com/otwcode/otwarchive/blob/1e369cdfb6c49e80b7290a3f1090ced7ecf14f2d/spec/controllers/works/drafts_spec.rb#L15

There's no way to tell for sure if we're always using the draft factory where the tests expect drafts, but we can check to some extent by text-searching for "draft" and "unposted" in the spec/ and features/ directories. For example, I see this test which needs to updated to use drafts:

https://github.com/otwcode/otwarchive/blob/1e369cdfb6c49e80b7290a3f1090ced7ecf14f2d/spec/controllers/chapters_controller_spec.rb#L9

spec/controllers/works/default_rails_actions_spec.rb
```diff
@@ -184,7 +184,7 @@ def call_with_params(params)
 
     it "doesn't allow a user to create a work in a series that they don't own" do
       @series = create(:series)
-      work_attributes = attributes_for(:work)
+      work_attributes = attributes_for(:work).except(:posted)
```
redsummernight, line 187: What happens if we don't exclude `:posted` btw?

sarken: The tests fail because `posted` isn't one of the parameters allowed in the controller.

```
WorksController
  create
    doesn't allow a user to create a work in a series that they don't own (FAILED - 1)

Failures:

  1) WorksController create doesn't allow a user to create a work in a series that they don't own
     Failure/Error:
       params.require(:work).permit(
         :rating_string, :fandom_string, :relationship_string, :character_string,
         :archive_warning_string, :category_string, :expected_number_of_chapters, :revised_at,
         :freeform_string, :summary, :notes, :endnotes, :collection_names, :recipients, :wip_length,
         :backdate, :language_id, :work_skin_id, :restricted, :anon_commenting_disabled,
         :moderated_commenting_enabled, :title, :pseuds_to_add, :collections_to_add,
         :unrestricted,
         current_user_pseud_ids: [],
         collections_to_remove: [],
         challenge_assignment_ids: [],
     
     ActionController::UnpermittedParameters:
       found unpermitted parameter: :posted
     # ./app/controllers/works_controller.rb:902:in `work_params'
     # ./app/controllers/works_controller.rb:282:in `create'
     # ./spec/controllers/works/default_rails_actions_spec.rb:190:in `block (4 levels) in <top (required)>'
     # ./spec/controllers/works/default_rails_actions_spec.rb:189:in `block (3 levels) in <top (required)>'
```

PR #3604, opened 2019-08-03
AO3-4592 Fix task to reveal stuck anonymous/unrevealed works

zz9pzza:
# Pull Request Checklist

* [ ] Have you added the [JIRA](https://otwarchive.atlassian.net) issue number as the *first* thing in your pull request title (eg: `AO3-1234 Fix thing`)

## Issue

https://otwarchive.atlassian.net/browse/AO3-4592


spec/miscellaneous/lib/tasks/after_tasks.rake_spec.rb
```diff
@@ -46,3 +46,73 @@
...
   end
 end
+
+
+
+
+describe "rake After:unhide_invited_works" do
+
+
+  let(:anonymous_collection) { create (:anonymous_collection) }
```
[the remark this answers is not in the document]
sarken: Hound's not quite right -- we do want the parentheses, but it should be `create(:anonymous_collection)`, without the space.

spec/miscellaneous/lib/tasks/after_tasks.rake_spec.rb
```diff
@@ -46,3 +46,67 @@
...
+    expect(anonymous_work.unrevealed?).to be(false)
+    expect(anonymous_work.anonymous?).to be(true)
+    expect(unrevealed_work.unrevealed?).to be(true)
+    expect(unrevealed_work.anonymous?).to be(false)
+    expect(anonymous_unrevealed_work.anonymous?).to be(true)
+    expect(anonymous_unrevealed_work.unrevealed?).to be(true)
+  end
+
+  context "When works get stuck" do
+    let(:chapter) { create(:chapter, work: en_work, posted: true, position: 2, authors: en_work.authors, content: "A few more words never hurt.") }
```
sarken: I think this just got dragged along while copy/pasting and can be removed.

spec/miscellaneous/lib/tasks/after_tasks.rake_spec.rb
```diff
@@ -46,3 +46,67 @@
...
+    expect(normal_work.anonymous?).to be(false)
+    expect(anonymous_work.unrevealed?).to be(false)
+    expect(anonymous_work.anonymous?).to be(true)
+    expect(unrevealed_work.unrevealed?).to be(true)
+    expect(unrevealed_work.anonymous?).to be(false)
+    expect(anonymous_unrevealed_work.anonymous?).to be(true)
+    expect(anonymous_unrevealed_work.unrevealed?).to be(true)
+  end
+
+  context "When works get stuck" do
```
sarken: Something more descriptive would be better, like `when invited works are incorrectly anonymous or unrevealed`

spec/miscellaneous/lib/tasks/after_tasks.rake_spec.rb
```diff
@@ -46,3 +46,67 @@
...
+  let(:anonymous_collection) { create (:anonymous_collection) }
+  let(:unrevealed_collection) { create (:unrevealed_collection) }
+  let(:other_collection) { create (:collection) }
+  let(:anonymous_unrevealed_collection) { create (:anonymous_unrevealed_collection)}
+  let(:anonymous_work) { create(:posted_work, collections: [anonymous_collection]) }
+  let(:unrevealed_work) { create(:posted_work, collections: [unrevealed_collection]) }
+  let(:normal_work) { create(:posted_work, collections: [other_collection]) }
+  let(:anonymous_unrevealed_work) { create(:posted_work, collections: [anonymous_unrevealed_collection]) }
+
+  it "works start as expected" do
```
sarken: While this can be useful while debugging tests, it's probably unnecessary to leave here longterm. (The test above, for example, doesn't check the initial state of the word counts.)

spec/miscellaneous/lib/tasks/after_tasks.rake_spec.rb
```diff
@@ -46,3 +46,67 @@
...
+    expect(unrevealed_work.anonymous?).to be(false)
+    expect(anonymous_unrevealed_work.anonymous?).to be(true)
+    expect(anonymous_unrevealed_work.unrevealed?).to be(true)
+  end
+
+  context "When works get stuck" do
+    let(:chapter) { create(:chapter, work: en_work, posted: true, position: 2, authors: en_work.authors, content: "A few more words never hurt.") }
+
+    before do
+      # Screw up collections
```
sarken: `Screw up collection items` would be more accurate.

spec/miscellaneous/lib/tasks/after_tasks.rake_spec.rb
```diff
@@ -46,3 +46,67 @@
...
+      anonymous_work.collection_items.first.update_columns(user_approval_status: 0)
+      unrevealed_work.collection_items.first.update_columns(user_approval_status: 0)
+      anonymous_unrevealed_work.collection_items.first.update_columns(user_approval_status: 0)
+    end
+
+    it "works can get stuck" do
```
sarken: Again, useful for debugging, but probably not necessary to keep around.

spec/miscellaneous/lib/tasks/after_tasks.rake_spec.rb
```diff
@@ -46,3 +46,67 @@
...
+      expect(normal_work.anonymous?).to be(false)
+      expect(anonymous_work.unrevealed?).to be(false)
+      expect(anonymous_work.anonymous?).to be(true)
+      expect(unrevealed_work.unrevealed?).to be(true)
+      expect(unrevealed_work.anonymous?).to be(false)
+      expect(anonymous_unrevealed_work.anonymous?).to be(true)
+      expect(anonymous_unrevealed_work.unrevealed?).to be(true)
+    end
+
+    it "will be fixed by running the rake task" do
```
sarken: `it "updates the anonymous and unrevealed status of invited works"`

spec/miscellaneous/lib/tasks/after_tasks.rake_spec.rb
```diff
@@ -46,3 +46,67 @@
...
+
+describe "rake After:unhide_invited_works" do
+  let(:anonymous_collection) { create (:anonymous_collection) }
+  let(:unrevealed_collection) { create (:unrevealed_collection) }
+  let(:other_collection) { create (:collection) }
+  let(:anonymous_unrevealed_collection) { create (:anonymous_unrevealed_collection)}
+  let(:anonymous_work) { create(:posted_work, collections: [anonymous_collection]) }
+  let(:unrevealed_work) { create(:posted_work, collections: [unrevealed_collection]) }
+  let(:normal_work) { create(:posted_work, collections: [other_collection]) }
+  let(:anonymous_unrevealed_work) { create(:posted_work, collections: [anonymous_unrevealed_collection]) }
```
sarken: It would be a good idea to have some works that are actually supposed to be anonymous or unrevealed as well, just to make sure we don't accidentally de-anon or reveal them. So there could be `anonymous_work`, `anonymous_invited_work`, `unrevealed_work`, `unrevealed_invited_work`, etc.

spec/miscellaneous/lib/tasks/after_tasks.rake_spec.rb
zz9pzza: It was indeed initial debuging happy to take out

spec/miscellaneous/lib/tasks/after_tasks.rake_spec.rb
```diff
@@ -46,3 +46,66 @@
     end
   end
 end
+
+describe "rake After:unhide_invited_works" do
+  let(:anonymous_collection) { create(:anonymous_collection) }
+  let(:unrevealed_collection) { create(:unrevealed_collection) }
+  let(:other_collection) { create(:collection) }
+  let(:anonymous_unrevealed_collection) { create (:anonymous_unrevealed_collection)}
```
sarken: There's an extra space here: `create (:anonymous_unrevealed_collection)`

spec/miscellaneous/lib/tasks/after_tasks.rake_spec.rb
zz9pzza: ( it used to be accurate, which was why I couldn't get it to fail :), changing )

spec/miscellaneous/lib/tasks/after_tasks.rake_spec.rb
zz9pzza: Sure :)

spec/miscellaneous/lib/tasks/after_tasks.rake_spec.rb
```diff
@@ -46,3 +46,49 @@
...
+
+  context "When invited works are incorrectly anonymous or unrevealed" do
+    before do
+      # Screw up collection items
+      invited_anonymous_work.collection_items.first.update_columns(user_approval_status: 0)
+      invited_unrevealed_work.collection_items.first.update_columns(user_approval_status: 0)
+      invited_anonymous_unrevealed_work.collection_items.first.update_columns(user_approval_status: 0)
+    end
+
+    it "updates the anonymous and unrevealed status of invited work" do
```
sarken: Should be `works`.

Some whitespace like the other test has might help with readability here -- maybe a couple comments, too:

```
    it "updates the anonymous and unrevealed status of invited work" do
      subject.invoke

      normal_work.reload
      invited_anonymous_work.reload
      invited_unrevealed_work.reload
      invited_anonymous_unrevealed_work.reload
      unrevealed_work.reload
      anonymous_work.reload

      # Accepted works should be unchanged
      expect(normal_work.unrevealed?).to be(false)
      expect(normal_work.anonymous?).to be(false)
      expect(anonymous_work.unrevealed?).to be(false)
      expect(anonymous_work.anonymous?).to be(true)
      expect(unrevealed_work.unrevealed?).to be(true)
      expect(unrevealed_work.anonymous?).to be(false)

      # Invited works should no longer be anonymous or unrevealed
      expect(invited_anonymous_work.unrevealed?).to be(false)
      expect(invited_anonymous_work.anonymous?).to be(false)
      expect(invited_unrevealed_work.unrevealed?).to be(false)
      expect(invited_unrevealed_work.anonymous?).to be(false)
      expect(invited_anonymous_unrevealed_work.anonymous?).to be(false)
      expect(invited_anonymous_unrevealed_work.unrevealed?).to be(false)
    end
```

Probably in an ideal world, these specs would be broken down into something like: 
```
context "when collection is anonymous" do
  it "updates the anonymous status of invited works" do
  end
end

context "when collection is unrevealed" do
  it "updates the unrevealed status of invited works" do
  end
end

context "when collection is unrevealed and anonymous" do
  it "updates the anonymous and unrevealed status of invited works" do
  end
end
```
But I think what you have works well enough for our purposes -- we're not exactly at risk of breaking this task in future code changes and needing to know exactly which scenario is broken.

lib/tasks/after_tasks.rake
```diff
@@ -618,7 +618,7 @@ namespace :After do
   end
   
   desc "Reveal works and creators hidden upon invitation to unrevealed or anonymous collection"
-  task(:unhide_invited_works) do
+  task(unhide_invited_works: :environment) do
     works = Work.where("in_anon_collection IS true OR in_unrevealed_collection IS true")
     works.find_in_batches.each do |batch|
```
redsummernight: The `.each` here can be left out.

spec/miscellaneous/lib/tasks/after_tasks.rake_spec.rb
```diff
@@ -46,3 +46,49 @@
...
+  let(:invited_unrevealed_work) { create(:posted_work, collections: [invited_unrevealed_collection]) }
+  let(:normal_work) { create(:posted_work, collections: [other_collection]) }
+  let(:invited_anonymous_unrevealed_work) { create(:posted_work, collections: [invited_anonymous_unrevealed_collection]) }
+  let(:anonymous_work) { create(:posted_work, collections: [anonymous_collection]) }
+  let(:unrevealed_work) { create(:posted_work, collections: [unrevealed_collection]) }
+
+  context "When invited works are incorrectly anonymous or unrevealed" do
+    before do
+      # Screw up collection items
+      invited_anonymous_work.collection_items.first.update_columns(user_approval_status: 0)
```
redsummernight: We should use `CollectionItem::NEUTRAL` instead of 0.

lib/tasks/after_tasks.rake
```diff
@@ -618,7 +618,7 @@ namespace :After do
...
       batch.each do |work|
```
redsummernight: Should we print the total `works` at the start, and `print(".") && STDOUT.flush` after each batch, similar to the two most recent tasks?

spec/miscellaneous/lib/tasks/after_tasks.rake_spec.rb
```diff
@@ -46,3 +46,49 @@
     end
   end
 end
+
+describe "rake After:unhide_invited_works" do
+  let(:invited_anonymous_collection) { create(:anonymous_collection) }
+  let(:invited_unrevealed_collection) { create(:unrevealed_collection) }
+  let(:anonymous_collection) { create(:anonymous_collection) }
+  let(:unrevealed_collection) { create(:unrevealed_collection) }
+  let(:other_collection) { create(:collection) }
```
redsummernight: This can be `collection` instead of `other_collection`.

spec/miscellaneous/lib/tasks/after_tasks.rake_spec.rb
```diff
@@ -46,3 +46,49 @@
...
+  let(:invited_anonymous_unrevealed_collection) { create(:anonymous_unrevealed_collection) }
+  let(:invited_anonymous_work) { create(:posted_work, collections: [invited_anonymous_collection]) }
+  let(:invited_unrevealed_work) { create(:posted_work, collections: [invited_unrevealed_collection]) }
+  let(:normal_work) { create(:posted_work, collections: [other_collection]) }
```
redsummernight: This can be `work` instead of `normal_work`.

lib/tasks/after_tasks.rake
zz9pzza: Loading production environment (Rails 5.1.6.2)
2.3.4 :001 > Work.where("in_anon_collection IS true OR in_unrevealed_collection IS true").count
 => 15279

spec/miscellaneous/lib/tasks/after_tasks.rake_spec.rb
```diff
@@ -46,3 +46,49 @@
...
 end
+
+describe "rake After:unhide_invited_works" do
+  let(:invited_anonymous_collection) { create(:anonymous_collection) }
+  let(:invited_unrevealed_collection) { create(:unrevealed_collection) }
+  let(:anonymous_collection) { create(:anonymous_collection) }
+  let(:unrevealed_collection) { create(:unrevealed_collection) }
+  let(:other_collection) { create(:collection) }
+  let(:invited_anonymous_unrevealed_collection) { create(:anonymous_unrevealed_collection) }
+  let(:invited_anonymous_work) { create(:posted_work, collections: [invited_anonymous_collection]) }
```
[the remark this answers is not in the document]
zz9pzza: I think its simpler and clearer to have one work per collection as it makes the breaking the collection item deterministic and not dependent of the order of the items in the collection.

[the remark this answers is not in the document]
sarken: It wouldn't depend on the order of the works in the collection. You're finding the `collection_item` based on the work: `invited_anonymous_work.collection_items.first.update_columns(user_approval_status: CollectionItem::NEUTRAL)`.  Each work will still only have one `collection_item` because it will only be in one collection.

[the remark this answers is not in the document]
zz9pzza: Your right.

PR #3605, opened 2019-08-09
AO3-5480 Switch to Zoho Desk REST API for Support and Abuse forms

CristinaRO:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5480

## Purpose

What does this PR do?

* Implements two mini-clients for the Zoho Desk REST APIs, one for the authorisation server (used to obtain the access token) and one for the resource server, used to search for an existing contact, create a contact, and create the ticket.

* The experience of the user who is sending a ticket should not change.

* Underneath the app will make at least three external requests, as we need to
- obtain a fresh access token (OAuth access tokens are short-lived)
- use the email to look for an existing `contactId`
- use the email (and the username if provided) to create a new contact in Zoho, in case one didn't exist
- use this `contactId` to create the ticket

## Assumptions:
- We are content with getting the latest contact that matches the given email. We won't be creating duplicates going forward, and Zoho have said that they're preventing duplication, but in the unlikely event that two contacts with the same email exist, we will only return one, and it will be the most recently modified.
- There are a whole bunch of new ArchiveConfig variables needing configging, and someone other than me knows how to do this.

## Feedback

I welcome any ideas on how to do things safer and more efficiently, in that order.

Are the specs sufficient? Any ideas on how to fill in the pending spec?

Should more things be env vars / settings rather than hardcoded constants?

## Testing Instructions

## References

## Credit

Enigel

she/her


app/models/feedback_reporters/feedback_reporter.rb
```diff
@@ -25,17 +28,36 @@ def description
...
 
   def send_report!
-    HTTParty.post("#{ArchiveConfig.NEW_BUGS_SITE}#{project_path}",
-                  body: "&xml=#{URI.encode_www_form_component(xml.to_str)}")
+    zoho_resource_client.create_ticket(
+      ticket_attributes: report_attributes,
+    )
+  end
+
+  def report_attributes
```
sarken: It looks like both kinds of reports have `departmentId` in the `report_attributes`, too. Would it be possible to borrow the approach Elz used for setting the `project_path` and use it for `departmentId`, e.g. add this method here:

```
def department_id
  self.class::DEPARTMENT_ID
end
```

And then set `DEPARTMENT_ID = ArchiveConfig.ABUSE_ZOHO_DEPARTMENT_ID` and `DEPARTMENT_ID = ArchiveConfig.SUPPORT_ZOHO_DEPARTMENT_ID` in the appropriate classes?

lib/zoho_auth_client.rb
```diff
@@ -0,0 +1,40 @@
+class ZohoAuthClient
+  ACCESS_TOKEN_REQUEST_ENDPOINT = 'https://accounts.zoho.com/oauth/v2/token'.freeze
+
+  CLIENT_ID = ArchiveConfig.ZOHO_CLIENT_ID
+  CLIENT_SECRET = ArchiveConfig.ZOHO_CLIENT_SECRET
+  ZOHO_REFRESH_TOKEN = ArchiveConfig.ZOHO_REFRESH_TOKEN
+
+  # TODO: Can we get rid of this? We don't redirect, but it has to match
+  # whatever was filled in when creating the original grant_code
```
sarken: And we have to have it, maybe we should change it to the homepage, just because it's weird to have it be `/support` when we're using it for both the Support and PAC forms. (Not that it makes a big difference considering it's not being used for anything, but...)

lib/zoho_resource_client.rb
```diff
@@ -0,0 +1,73 @@
...
+  def create_contact
+    response_raw = HTTParty.post(
+      CONTACT_CREATE_ENDPOINT,
+      headers: headers,
+      body: contact_body.to_json
+    )
+    JSON.parse(response_raw.body)
+  end
+
+  def search_params
```
sarken: Should we have a comment that says despite us explicitly limiting it to 1 result, there should only ever be one client per email address because Zoho requires emails to be unique? Just because it was a pain in the neck to find that information (assuming it's accurate, lolsob).

Also, is there any reason we might want to call this `contact_search_params`? Or could the same method potentially hold search parameters for other things as well?

lib/html_cleaner.rb
```diff
@@ -405,6 +405,6 @@ def strip_html_breaks_simple(value)
 
   def add_break_between_paragraphs(value)
     return "" if value.blank?
-    value.gsub(/\s*<\/p>\s*<p>s*/, "</p><br /><p>")
+    value.gsub(/\s*<\/p>\s*<p>\s*/, "</p><br /><p>")
```
sarken: lol, this was me. At least we had a good laugh? :snake:

lib/zoho_resource_client.rb
```diff
@@ -0,0 +1,73 @@
...
+    {
+      "Content-Type" => "application/json",
+      "orgId" => ZOHO_ORG_ID,
+      "Authorization" => "Zoho-oauthtoken #{@access_token}"
+    }
+  end
+
+  def contact_body
+    {
+      "lastName" => @username.presence || @email,
```
sarken: I can't remember... we're using `lastName` like this because it's required? Whether it's that or something else, it might be a good idea to add a comment explaining.

app/models/feedback_reporters/feedback_reporter.rb
```diff
@@ -25,17 +28,36 @@ def description
...
+    {
+      "email" => email,
+      "contactId" => zoho_contact_id,
+      "cf" => {
+        "Language" => language.presence || "English",
```
sarken: Do we want to hard code English here or get the language from `Locale.default.language`?

lib/zoho_auth_client.rb
```diff
@@ -0,0 +1,40 @@
...
+  REDIRECT_URI = 'https://archiveofourown.org/support'.freeze
+
+  # TODO: Can we reduce this?
+  SCOPE = 'Desk.tickets.ALL,Desk.contacts.READ,Desk.contacts.WRITE,Desk.contacts.CREATE,Desk.basic.READ,Desk.search.READ'.freeze
+
+  def initialize
+    @access_token = access_token
```
tickinginstant: What's the purpose of assigning this to `@access_token`? The variable doesn't appear to be accessed anywhere.

(Also, doesn't this mean that calling `ZohoAuthClient.new.access_token`, as you do above, would cause the token to be retrieved/refreshed twice? Once when `access_token` is called in the constructor, and again when `access_token` is called on the resulting object?)

lib/zoho_auth_client.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  def access_token
+    response_raw = HTTParty.post(ACCESS_TOKEN_REQUEST_ENDPOINT + '?' + query_string(access_token_params))
+    JSON.parse(response_raw.body)['access_token']
```
tickinginstant: The access token typically lasts an hour, according to the docs. I'm guessing there's usually more than one support request/abuse report per hour, so to reduce the overhead, maybe this should use memcached or redis to cache the token, and refresh only when necessary? Or is there a particular reason you want to refresh each time?

The response is supposed to return the number of seconds that the token will be good for, so something like this might work if you want to cache with memcached:
```
return cached_token if (cached_token = Rails.cache.read("zoho_access_token")).present?
response_raw = HTTParty.post(ACCESS_TOKEN_REQUEST_ENDPOINT + '?' + query_string(access_token_params))
response_json = JSON.parse(response_raw.body)
# Subtract extra time to make sure that the token doesn't expire at an inconvenient time:
cache_time = response_json["expires_in_sec"] - 1.minute
Rails.cache.write("zoho_access_token", response_json["access_token"], expires_in: cache_time)
```

lib/zoho_resource_client.rb
```diff
@@ -0,0 +1,73 @@
...
+  TICKET_CREATE_ENDPOINT = 'https://desk.zoho.com/api/v1/tickets'.freeze
+
+  def initialize(access_token:, email:, username: nil)
+    @access_token = access_token
+    @email = email
+    @username = username
+  end
+
+  def retrieve_contact_id
+    (find_contact || create_contact).fetch('id')
```
tickinginstant: The Zoho docs have a warning about searching for brand new records:
> New resources that you add to your help desk might take some time to be indexed for search. Therefore, if the resources do not appear in search results, wait for a few minutes and try again.

Have you tried submitting two reports at almost exactly the same time with the same (never-before-used) email address? Does it create duplicate contact records? Does it generate an error when trying to create the second contact? If it does generate an error, should that error be handled in some way? Or is the indexing too fast to run into that sort of issue at the moment?

lib/zoho_resource_client.rb
```diff
@@ -0,0 +1,73 @@
...
+
+  def contact_body
+    {
+      "lastName" => @username.presence || @email,
+      "email" => @email,
+    }
+  end
+
+  def query_string(params)
+    params.map {|k, v| "#{k}=#{v}" }.join('&')
```
tickinginstant: Do these query parameters need to be encoded, since they include the user-provided email (which can include `=`, `+`, and `&`)?

Also, emails can contain `*`, which is the character that Zoho uses for wildcard searches. So that might need to be escaped separately in order to ensure that you retrieve the right contact with your search. (Assuming Zoho lets you escape wildcards. It stands to reason that they ought to, though I couldn't find a mention of it when I skimmed the search docs.)

(Side note: If Zoho does let you escape `*` using `\*`, it's worth noting that you can use the `\` character in email addresses if it's somewhere inside a double-quoted section. So you might or might not need to escape that, too.)

lib/zoho_resource_client.rb
tickinginstant: Correct me if I'm wrong, but it looks like `@username` is only used here, which means that it's recorded when the user submits their first ticket, but never updated in subsequent tickets. So if a user changes their name between tickets (maybe they're logged-in and changed their username? maybe they made a typo the first time?), the second ticket will not contain the new name.

Maybe the `lastName` in the contact should be updated if it doesn't match the name the user submitted, so that Support/Abuse always knows the current preferred name of the person they're interacting with? Or maybe the `@username` should be added as a custom field, so that each ticket has the name that the user wanted to be associated with that ticket? (In that case, it might make sense to always set `lastName` equal to the email, so that it never needs to be updated.)

sarken: I like the idea of always using the email as the lastName and making username a custom field.

app/models/feedback_reporters/feedback_reporter.rb
```diff
@@ -25,17 +28,36 @@ def description
...
+    }
+  end
+
+  private
+
+  def zoho_contact_id
+    zoho_resource_client.retrieve_contact_id
   end
 
-  def xml
```
sarken: Do we still need the XML files or should they be deleted?

lib/zoho_auth_client.rb
```diff
@@ -0,0 +1,40 @@
+class ZohoAuthClient
+  ACCESS_TOKEN_REQUEST_ENDPOINT = 'https://accounts.zoho.com/oauth/v2/token'.freeze
+
+  CLIENT_ID = ArchiveConfig.ZOHO_CLIENT_ID
+  CLIENT_SECRET = ArchiveConfig.ZOHO_CLIENT_SECRET
+  ZOHO_REFRESH_TOKEN = ArchiveConfig.ZOHO_REFRESH_TOKEN
```
sarken: Can you make sure `config.yml` has placeholder values for all of these settings? They should go around here: https://github.com/otwcode/otwarchive/blob/1f182644505830de9eb9548c97846456be59e66f/config/config.yml#L347-L353

I think this is all of them:
ArchiveConfig.ABUSE_ZOHO_DEPARTMENT_ID
ArchiveConfig.SUPPORT_ZOHO_DEPARTMENT_ID
ArchiveConfig.ZOHO_CLIENT_ID
ArchiveConfig.ZOHO_CLIENT_SECRET
ArchiveConfig.ZOHO_REFRESH_TOKEN
ArchiveConfig.ZOHO_ORG_ID

lib/zoho_auth_client.rb
sarken: Actually... this should probably go in the config file as `ZOHO_REDIRECT_URI`.

lib/zoho_auth_client.rb
redsummernight: We should probably set [`race_condition_ttl`](https://api.rubyonrails.org/v5.1.6.2/classes/ActiveSupport/Cache/Store.html#method-i-fetch) on the cache block as well.

lib/zoho_auth_client.rb
redsummernight: The `ArchiveConfig` values that are no longer used (e.g. `ABUSE_AUTH`) should be removed from this file.

lib/zoho_auth_client.rb
tickinginstant: Doesn't that only work when calling `fetch`?

(I didn't write this as a `fetch` block because `expires_in` depends on the value that would be calculated inside the cache block. If there were a cache block, that is.)

redsummernight: The option works is there for `write` too, `fetch` just passes it along.

Agree on not having cache blocks though.

tickinginstant: But `race_condition_ttl` can only be applied properly when reading from the cache (and discovering that it's expired), so setting it with `write` doesn't actually do anything except allow it to be used within `fetch`. (Since it's not stored alongside the cache entry.)

The only place in the rails code that it's actually applied is in `handle_expired_entry`, which is only called by `fetch`.

redsummernight: :+1: yeah, we don't need it. Thanks for catching that.

app/models/feedback_reporters/feedback_reporter.rb
redsummernight: I'm going with `Language.default.name` since we use Language objects for the dropdowns.

app/models/feedback_reporters/feedback_reporter.rb
redsummernight: Currently just stubbing `ArchiveConfig` makes the tests work. If we do this, we'll also need to stub `<class>::DEPARTMENT_ID`s as well.

sarken: Let's file this under "not worth creating extra work at the moment," then.

lib/zoho_auth_client.rb
CristinaRO: Doh! Good catch.

lib/zoho_resource_client.rb
CristinaRO: No idea what happens, thanks for catching that! Ideas welcome 😬

app/models/feedback_reporters/feedback_reporter.rb
```diff
@@ -1,6 +1,11 @@
+ # frozen_string_literal: true
```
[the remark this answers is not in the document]
CristinaRO: AAAAAAAAAARGH

lib/zoho_resource_client.rb
CristinaRO: We're not searching for anything else yet, was my main reason. It could be reused to search for anything that holds an email field, yes.

lib/zoho_resource_client.rb
CristinaRO: Yeah, `lastName` is the only required field for a contact.

CristinaRO: Do you want to liaise with Support to decide what they want the custom field to be named? `username` is a bit misleading, given that it's not pulled from the Archive username.

sarken: Can we just go with "Name" for the custom field (the forms both say "Your name" so it would be most logical)? If not, then yeah, I'll talk to Support and PAC.

CristinaRO: "Name" it is!

PR #3609, opened 2019-08-14
AO3-5709 Use pseud byline for download authors

CristinaRO:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5709

## Purpose

What does this PR do?

The pseud byline includes the user name when it's different from the pseud name, so we use the byline instead of the name for generating the `author_names`.

I've also added a bunch of specs for previously unspecced methods, so we can see what knock-on effects changing the way we generate `author_names` will have.

In addition to the metadata generator, `author_names` is also used to generate the `page_title` as `file_authors`, so it needs to be "filesystem safe" and on the short side.

Getting truncated at 24 characters could make one creator with a parenthetical name push all the others out of the listing. OTOH this can already happen with a pseud that's the same as the user name, so I guess we're not making it much worse?

## Testing Instructions

Already listed in JIRA.

## References

## Credit

Enigel

she/her

### DIFF
--- a/app/models/download.rb
+++ b/app/models/download.rb
@@ -102,7 +102,7 @@ class Download
   end
 
   def author_names
-    work.anonymous? ? ["Anonymous"] : work.pseuds.sort.map(&:name)
+    work.anonymous? ? ["Anonymous"] : work.pseuds.sort.map(&:byline)
   end
 
   # need the next two to be filesystem safe and not overly long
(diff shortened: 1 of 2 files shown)


sarken:
We talked about this a bit privately, but I think we should make another issue to change `page_title` to match the default way we display the title on work pages, minus any "Archive of Our Own" or chapter bits. (`page_title`, as far as I can tell, is only used to generate the `<title>` of an HTML download.) 

https://github.com/otwcode/otwarchive/blob/3463d1a8f9b349bd3f827098d43c10c786777f9b/app/controllers/application_controller.rb#L333-L353

sarken:
Approved and made https://otwarchive.atlassian.net/browse/AO3-5734 for the page title

PR #3610, opened 2019-08-16
AO3-2669 Wranglers can only add and remove themselves from fandoms

CristinaRO:
*NOTE*: The text below is all from Tal's original PR, as is the code. All I've done is fix the conflicts in `tag_steps.rb`, `tag_wrangling_more.feature`, and `factories/users.rb`.

## Issue

https://otwarchive.atlassian.net/browse/AO3-2669

## Purpose

Changes view and controller so wranglers can only add or remove themselves from fandoms.

## Testing

Added feature and spec tests that assert the requirements.

## Credit

Tal.
Please use he.


CristinaRO:
It looks like I was able to force push to the original after all! Closing this one (have already reopened the original) 🙈

PR #3614, opened 2019-08-29
AO3-5702 Require language when posting a new work

CristinaRO:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5702

## Purpose

What does this PR do?

* Makes `language` a required field when posting (not importing) a new work.
* Adds a custom validation rather than using the Rails presence validator for two reasons:
  * Validating that the language itself exists, rather than the `language_id` being populated
  * Controlling where in the form the error message gets displayed: we wanted it at the top with any other errors, rather than crammed next to the language dropdown
* Adds specific automated tests for the new functionality
* Updates all the automated tests that were broken due to the new validation

## Testing Instructions

* Attempt to post a new work without choosing the language, observe that the work is not posted, and the error message appears at the top.
* Post a valid work with a language from the dropdown, and observe that it posts correctly.
* For extra credit, attempt to modify the form with a browser developer tool or similar, and replace the valid `language_id` with a made-up id. Observe that the work does not get posted and the error message appears at the top.
* Edit an existing work, deselect the language, and attempt to update the work. Observe that the work is not updated and the error message appears at the top.

## References

## Credit

Enigel

she/her


features/step_definitions/challenge_promptmeme_steps.rb
```diff
@@ -386,6 +387,7 @@
   step %{I fill in "Work Title" with "Second Story"}
     step %{I select "Not Rated" from "Rating"}
     step %{I check "No Archive Warnings Apply"}
+    step %{I select "English" from "Choose a language"}
```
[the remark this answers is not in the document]
CristinaRO: It's a stylistic choice, Hound.

app/views/works/_work_form_associations_language.html.erb
```diff
@@ -1,11 +1,12 @@
 <% # expects local variable f %>
-<dt class="language">
+<dt class="language required">
   <%= f.label :language_id, ts('Choose a language') %>
   <%= link_to_help "languages-help" %>
 </dt>
 <dd class="language">
```
sarken: Can you add `class="required"` here as well?

app/views/works/_work_form_associations_language.html.erb
```diff
@@ -1,11 +1,12 @@
...
   <select id="work_language_id" name="work[language_id]">
+    <option value=''>Please select a language</option>
```
sarken: Could you change this to double quotes on the `value` attribute?

We don't have any hard rules we follow when it comes to default values, but I think we could go with an empty option here. The "Choose a language" label and "Please select a language" placeholder seem a little redundant. 

Opinions, anyone else?

If we keep text here, can it be done inside `<%= ts("") %>`? (Legit question; I'm not sure off the top of my head.)

spec/models/work_spec.rb
```diff
@@ -76,14 +76,41 @@
...
+
+    it "is not valid with a language we don't support" do
+      work = build(:work, language_id: deleted_language_id)
+
+      expect(work).not_to be_valid
+      expect(work.errors.messages[:base]).to include("Language cannot be blank.")
+    end
+
+    it "is not valid without a language" do
+      work = build(:work, language_id: '')
```
sarken: Can you use double quotes here for consistency?

app/views/works/_work_form_associations_language.html.erb
CristinaRO: Changed to empty text for now, while awaiting opinions. ;)

ariana-paris: Since we're gathering opinions, mine is that explicit placeholder text on the actual dropdown like `-- Please select a language --` (particularly with extra punctuations like that) more clearly indicates that you have to do _something_ with this dropdown box, independently of what's also on the label. The main reason being that IIRC our labels aren't always particularly close to the corresponding UI component in our layout, making them easy to ignore. But that's just my opinion; it's fine if we don't want to mess about changing it back and as Sarken said, we don't use placeholder text on the other dropdowns in that view anyway.

I assume it should be translatable since this an erb file that will be pre-processed before it's turned into HTML and dumped in the user's browser.

sarken: Screen readers read extra punctuation, so we should really avoid that even if we do go with placeholder text. However, because I wanted to bring up the punctuation/screen reader point, I had to Google if screen readers read such text -- "some do, some don't" seems to be the current case -- and I ended up on the [Nielsen Group's page about placeholders](https://www.nngroup.com/articles/form-design-placeholders/). 

It seems like current advice is to avoid using it because it can cause a number of usability issues. (While I do think Ariana has a good point about our labels occasionally being far from our fields, I don't think that is necessarily true with this field -- they're pretty close, at least compared to the checkboxes above them, which are a mile and a half from their labels.) A fair number of the problems listed don't apply to us because we use labels and text-based instructions, but I think the one that might apply is users thinking the field is filled in when they see text already selected... especially since users are currently used to something being selected in that field.

I'm not sure if it's still true, but [this June 2018 Smashing Magazine post](https://www.smashingmagazine.com/2018/06/placeholder-attribute/) also says translation tools will skip over placeholder text, leaving it in the original language. Since we're making this change to help people who don't post in English and maybe don't have great English skills, that seems rather unfortunate.

sarken: Since this has merge conflicts and needs to be updated anyway, let's add the "Please select a language" placeholder (sans extra punctuation) back in and see what happens when we run it through a translation tool and a screen reader.

zz9pzza:
Merge conflicts :(

CristinaRO:
Solved conflicts.

sarken:
Continues to look good!

app/views/works/_work_form_associations_language.html.erb
```diff
@@ -1,11 +1,12 @@
 <% # expects local variable f %>
-<dt class="language">
-  <%= f.label :language_id, ts('Choose a language') %>
+<dt class="language required">
+  <%= f.label :language_id, ts("Choose a language") %>
```
sarken: I just pulled this to check something for Translation and realized this field doesn't have an asterisk to mark it required. Can you add one? The `*` should be outside the `ts` like `<%= form.label :title, ts("Category name") + "*" %>`  so we don't have to worry about losing it in Translation.

PR #3625, opened 2019-09-06
AO3-4353 AO3-5786 Cache tag sets with tag types, fix 500 when listing tags in tag sets

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4353
https://otwarchive.atlassian.net/browse/AO3-5786

## Purpose

The tag set pages that are linked from challenge sign up forms would sometimes show the wrong type of tags. This updates the cache key to prevent that from happening.

I was going to sit on this until the mountain got smaller, but challenge season will soon be upon us.

ETA: I just noticed that the issue also mentions a different problem ("Even if the correct type of tags shows, the tags in it may be wrong (especially if wrangling, not edits to the tag set, has changed the tags)"), which this ~does not fix. That's a slightly different kettle of fish -- cache expiration -- that needs a different set of changes and could probably have its own issue.~ should fix now!

ETA 2: The 500 error is tag_type instead of @tag_type on current line 3:
```
/home/ao3app/app/releases/20191009182403/app/views/owned_tag_sets/show_options.html.erb:3:in `block in app_views_owned_tag_sets_show_options_html_erb__3214890490530188503_46983882745600'
```
which is now line 12 and uses @tag_type.


app/views/owned_tag_sets/show_options.html.erb
```diff
@@ -1,19 +1,19 @@
-<% cache("tagsets_#{@tag_set_ids.join('-')}_#{@tag_sets.collect(&:updated_at).sort.last}", skip_digest: true) do %>
+<% cache("tagsets_#{@tag_set_ids.join('-')}_#{@tag_sets.collect(&:updated_at).sort.last}_#{@tag_type}", skip_digest: true) do %>
```
sarken: This uses the most recent `updated_at` of the tag sets to expire the cache... could we collect the _tags_ in the tag sets and use the newest of _their_ `updated_at` dates to handle the cache expiration?

sarken: My other idea is every time a tag is updated, we check which tag sets it is in, and if there are any, we expire their cache. That doesn't seem great, either, though, because I'm sure most tags are not in tag sets.

tickinginstant: If you generate the cache key for a collection of ActiveRecords, the max `updated_at` is one of the fields used in the cache key:
```
2.3.4 :012 > Character.joins(:set_taggings).where(set_taggings: { tag_set_id: [1, 2, 3, 4]}).cache_key
   (0.9ms)  SELECT COUNT(*) AS `size`, MAX(`tags`.`updated_at`) AS timestamp FROM `tags` INNER JOIN `set_taggings` ON `set_taggings`.`tag_id` = `tags`.`id` WHERE `tags`.`type` IN ('Character') AND `set_taggings`.`tag_set_id` IN (1, 2, 3, 4)
  ↳ (irb):12:in `irb_binding'
 => "characters/query-3ab0f549a9f08bbfbe3afd3eef0d90cb-2-20190406194349000000" 
```
It also includes the number of items in the collection, which helps it expire the cache in some cases where a record has been added or deleted. Although if you want to handle the case where a the same number of things has been added *and* deleted, I think you might want to check `MAX(set_taggings.updated_at)`, too. Or make `SetTagging`s have `touch: true` for their `tag_set` association, and include `MAX(tag_sets.updated_at)`.

(You might be able to use the max `updated_at` for the tag sets regardless, but I just did some experiments on my own machine, and it looked like adding a tag to a tag set didn't actually update any timestamps aside from the one on the brand new `SetTagging`. So I figured I'd suggest checking the `SetTagging` timestamps just to be sure.)

sarken: I want to make sure I understand before I push anything...

To account for as many scenarios as possible, I should:
* add `touch: true` to the `tag_set` association in the `SetTagging model`
* make the cache key in the view file include:
  * the ids of the tag sets
  * the `maximum(:updated_at)` of the tag sets
  * the cache key for the collection of ActiveRecords

tickinginstant: You don't have to use the cache key for the ActiveRecord collection -- I was just trying to point out that Rails' does use the max `updated_at` value by default, so your idea about using it seemed reasonable. (Though using the built-in `cache_key` method might be convenient, depending on how you want to handle it.)

Also, after looking at the tag set code a bit more, I now doubt that `touch: true` would work, because removing tags from a tag set uses `delete_all`, which skips callbacks:
https://github.com/otwcode/otwarchive/blob/3f8ad8109e6bf3d111fd51ab3ca5a1dd5c8cb385/app/models/tagset_models/tag_set.rb#L117-L121

I'm not sure of the best way to resolve it. Maybe adding a call to `touch` in `remove_from_set` and `add_to_set`? But I haven't checked if there are other parts of the tag set code that can add/remove tags.

sarken: From a user standpoint, there are three ways to add tags, afaik:
- moderator adds them on the tag set edit page
- moderator approves nominated tags
- moderator batch adds tags

And those all ultimately use `add_to_set`, as far as I can tell. I don't think there's really any way for a tag to get removed beyond the moderator editing the tag set. (There's the tag being deleted from the database, of course, but I don't think that's a case we generally need to worry about.) 

So it does seem like adding a call to touch in those two methods might be a good way to fix this is all but the oddest cases. (Tag sets don't usually get updated after sign-ups are open, anyway, so Support is less concerned about getting the listed tags updated than making sure the right type of tag is displayed for the right link.)

app/views/owned_tag_sets/show_options.html.erb
```diff
@@ -1,18 +1,28 @@
...
-      <% end %>
-    </ul>
+<% # ids of tag sets whose tags are being shown %>
+<% ids = @tag_set_ids.join("-") %>
+
+<% # the show tag sets' most recent updated_at %>
+<% newest_updated_at = @tag_sets.collect(&:updated_at).sort.last %>
+
+<% # cache key for all of the sets' tags, which expires when the number of tags changes or when the most recent updated_at changes %>
+<% tags_cache_key = Tag.tag_sets_cache_key(@tag_type, @tag_set_ids) %>
```
tickinginstant: Would it work to call `@tags.cache_key` here? I think it's defined pretty similarly to `tag_sets_cache_key`, with the exception of the call to `by_name_without_articles`:
https://github.com/otwcode/otwarchive/blob/67b7b149d41ef626aca3596f2b1448b34d0cdc05/app/controllers/owned_tag_sets_controller.rb#L64

app/models/tagset_models/tag_set.rb
```diff
@@ -118,6 +118,7 @@ def remove_from_set(tags_to_remove)
     return unless tags_to_remove.present?
     self.set_taggings.where(tag_id: tags_to_remove.map(&:id)).delete_all
     remove_tags_from_autocomplete(tags_to_remove)
+    owned_tag_set.touch
```
tickinginstant: Whoops, spoke too soon! Looks like this needs to be `&.touch` here and below.

app/views/owned_tag_sets/show_options.html.erb
sarken: Ohh, yeah, much better than adding a new method -- good catch!

PR #3628, opened 2019-09-10
AO3-5703 Search for works and bookmarks by language short name

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5703

## Purpose

Maintain backward compatibility for old work/bookmark searches/filters using language IDs.

## Testing Instructions

See issue.


app/models/series.rb
```diff
@@ -207,13 +207,18 @@ def revised_at
   def bookmarkable_json
     as_json(
       root: false,
-      only: [:title, :summary, :hidden_by_admin, :restricted, :created_at,
-        :complete],
-      methods: [:revised_at, :posted, :tag, :filter_ids, :rating_ids,
+      only: [
```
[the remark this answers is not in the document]
redsummernight: Should I fix this for all 3 models (work, series, external work)?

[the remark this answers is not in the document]
CristinaRO: Personal opinion: I disagree with this rule to the point where I'd be in favour of turning off this particular "cop".

I think knowing instantly whether an entity is a symbol or string is useful, especially since coders everywhere still get occasionally bitten by hashes having keys as symbols instead of strings and vice versa. I tend to not notice the significance of whatever is used to frame the array.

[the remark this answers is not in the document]
sarken: I don't mind -- what is it %w? for strings because it spares the quotation marks, but I really don't see much value in %i here. I'd leave the code as is and be fine turning off the cop, which I'm pretty sure we've always ignored anyway.

app/models/search/work_search_form.rb
```diff
@@ -98,6 +99,25 @@ def standardize_creator_queries
...
+    if @options[:language_id].present? && @options[:language_id].to_i != 0
+      language = Language.find_by(id: options[:language_id])
+      options[:language_id] = language.short if language.present?
+    end
+
+    # - Using language IDs in "Any field" (search) or "Search within results" (filters)
+    if @options[:query].present?
+      @options[:query] = @options[:query].gsub(/\blanguage_id\s*:\s*(\d+)/) do |m|
+        lang = Language.find_by(id: "#{$1}")
+        lang = Language.default if lang.blank?
```
CristinaRO: Could we remove the language search param if the language can't be found? Allow more works to be returned rather than restrict to the Archive default?

redsummernight: If people use `language_id: 1` as a clause in an OR/AND search, removing it will cause a query syntax error (similar to [AO3-5104](https://otwarchive.atlassian.net/browse/AO3-5104?focusedCommentId=355062&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-355062)).

features/bookmarks/bookmark_search.feature
```diff
@@ -230,13 +230,13 @@ Feature: Search Bookmarks
     Given "someuser" has bookmarks of works in various languages
     # reload search page to bring new language-id mappings for dropdown
     When I reload the page
-      And I select "English" from "Work language"
+      And I select "Deutsch" from "Work language"
```
sarken: Potentially stupid question: why'd we switch which language we're testing with?

redsummernight: I wanted to know if things work for languages other than the Archive's default.

sarken: Makes sense to me! :+1:

sarken:
Oh! She's on holiday (or trying to be), so let's merge this.

PR #3630, opened 2019-09-10
AO3-5688 Change "co-author" to "co-creator" on delete preview page

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5688

## Purpose

* Changes remaining "co-author" to "co-creator."
* Improves indenting/spacing.
* Stops using hash rockets.
* Uses symbols rather than strings where possible.
* Uses double quotes for strings.

## References

If you're feeling adventurous, I also have a version that makes it mostly translatable: https://github.com/otwcode/otwarchive/compare/master...sarken:AO3-5688_BOT_i18n?expand=1 ("Mostly" because I didn't change the way it handles the list of pseuds.)

### DIFF
--- a/app/views/users/delete_preview.html.erb
+++ b/app/views/users/delete_preview.html.erb
@@ -1,6 +1,6 @@
 <h2 class="heading"><%=h 'What do you want to do with your works?' %></h2>
 
-<%= form_tag(@user, method: 'delete') do %>
+<%= form_tag(@user, method: :delete) do %>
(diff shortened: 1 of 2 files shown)


CristinaRO:
Ariana reviewed, I merge! (I've also reviewed, and <3 to sarken for doing the changes commit by commit so the diff didn't get overwhelming! This is now the standard to which everyone will be held. ;) )

PR #3631, opened 2019-09-10
AO3-5707: Fix for legacy warning search

elzj:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5707

## Purpose

Branches off AO3-5613 (Warning class rename) to enable support for browser-bookmarked searches including warning tag ids. 

## Testing Instructions

* Copy the link for a search that includes a warning filter.
* Change archive_warning_ids to warning_ids.
* Go to that url and see if you get the same results.

## References

Should obviously be merged after #3526


app/controllers/bookmarks_controller.rb
```diff
@@ -385,6 +385,7 @@ def bookmark_search_params
       :other_bookmark_tag_names,
       :excluded_bookmark_tag_names,
       rating_ids: [],
+      warning_ids: [],
```
CristinaRO: Do we need something similar in the works controller, to permit `warning_ids` in `work_search_params`?

app/models/search/bookmark_search_form.rb
```diff
@@ -68,20 +68,10 @@ def self.count_for_pseuds(pseuds)
...
-        @options[countable].gsub!("&gt;", ">")
-        @options[countable].gsub!("&lt;", "<")
-      end
-    end
-
-    # If we call the form field 'notes', the parser adds html to it
-    @options[:notes] = @options[:bookmark_notes]
-
+    @options = processed_options(options)
     # We need to respect some options that are deliberately set to false, and
```
CristinaRO: I would like the comment moved in the method as well, above the `nil?` check.

app/controllers/bookmarks_controller.rb
elzj: Yes, good catch, that one escaped.

elzj:
@CristinaRO @sarken I could fix the merge conflicts here, but that wouldn't fix them in the original pull request for #3526 so the question is whether this one should become the sole PR and that one should be closed. Otherwise, I'm probably just creating a future conflict for myself by fixing this one.

sarken:
@elzj  I fixed the merge conflicts in #3526 a little bit ago so you could probably pull those changes?

app/models/search/bookmark_search_form.rb
```diff
@@ -181,4 +169,26 @@ def sort_direction(sort_column)
     'desc'
   end
 
+  private
+
+  def processed_options(opts={})
```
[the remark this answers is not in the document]
sarken: A fair point from Hound.

spec/models/search/bookmark_search_form_spec.rb
```diff
@@ -166,4 +166,34 @@
...
   end
+
+  describe "#processed_options" do
+    it "removes blank options" do
+      options = { foo: nil, bar: false }
+      searcher = BookmarkSearchForm.new(options)
+      expect(searcher.options).to have_key(:bar)
+      expect(searcher.options).not_to have_key(:foo)
+    end
+    it "renames the notes field" do
```
sarken: I think we usually do blank lines between our `it` bocks.

app/models/search/bookmark_search_form.rb
```diff
@@ -181,4 +169,26 @@ def sort_direction(sort_column)
...
+    [:date, :bookmarkable_date].each do |countable|
+      if opts[countable].present?
```
sarken: Was there a reason we originally wrote this the way we did instead of chaining the `gsub` methods? I have this vague feeling there might have been, but can't quite recall.

(Doesn't need any action, Elz, just kind of making a note in case we notice something hinky after this is deployed... or in case someone has a better memory.)

elzj: I doubt it, and if we're trying to freeze strings for future Ruby compatibility, we should avoid methods like gsub! that change them in place.

app/models/search/work_indexer.rb
```diff
@@ -1,4 +1,5 @@
 class WorkIndexer < Indexer
```
redsummernight: This file can be left out, it no longer has code changes.

app/controllers/bookmarks_controller.rb
redsummernight: Should we comment here that this is for backwards compatibility?

app/controllers/works_controller.rb
```diff
@@ -957,6 +957,7 @@ def work_search_params
       :words_to,
 
       archive_warning_ids: [],
+      warning_ids: [],
```
redsummernight: Should we comment here that this is for backwards compatibility?

sarken:
There's a couple of merge conflicts now in addition to red's last few comments.

PR #3632, opened 2019-09-11
AO3-5688 Use double quotes and translatable text on delete user preview page

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5688

## Purpose

* Use a consistent translation method on the users/delete_preview view 
* Use double quotes in the option `Change my pseud to "orphan" and attach to the orphan account`
* Update the `When I choose...` step definition to work when the radio button label text has double quotes in it

## Testing Instructions

* Make sure there are double quotes in that option
* Make sure it doesn't say "translation missing" anywhere on the page


features/step_definitions/web_steps.rb
```diff
@@ -106,7 +106,7 @@ def with_scope(locator)
   end
 end
 
-When /^(?:|I )choose "([^"]*)"(?: within "([^"]*)")?$/ do |field, selector|
+When /^(?:|I )choose "(.*)"(?: within "([^"]*)")?$/ do |field, selector|
```
redsummernight: This works because we never use the optional "within" clause, a greedy match would have eaten up any second parameter.

sarken: Do you think it would be better to remove the unused clause or to set up a step definition like `When I opt to transfer my works/collections to the orphan account without my pseud attached`?

redsummernight: Let's remove the unused clause for now.

Whenever we update to Cucumber 3.x, we can use [Cucumber expressions](https://cucumber.io/docs/cucumber/cucumber-expressions/), which allows this exact case:

https://github.com/cucumber/cucumber-expressions-ruby/blob/1c5871a622739e3424640515eab950f8cc3c1cec/spec/cucumber/cucumber_expressions/cucumber_expression_spec.rb#L42-L44

config/locales/en.yml
```diff
@@ -6,6 +6,28 @@ en:
...
+  users:
+    delete_preview:
+      heading: "What do you want to do with your works?"
+      orphan: "orphan"
+      orphaning: "orphaning"
+      submit: "Save"
+      cancel: "Cancel"
+      confirm: "Are you sure? This can't be undone!"
+      options:
+        keep_pseud: "Leave my pseud but attach to the orphan account"
```
CristinaRO: I have to say (I know this is existing copy) that "keep pseud" and "leave my pseud" being used for the same action is twisting my mind into pretzels. I would be as confused as our users if I had to choose between these. 😬

sarken: Yeah. I do not love the phrasing on this or the next one. Reworking the text on the various orphaning pages might be nice someday (although at least "Leave my pseud" is standard between this and the work page).

features/step_definitions/web_steps.rb
```diff
@@ -106,7 +106,7 @@ def with_scope(locator)
   end
 end
 
-When /^(?:|I )choose "([^"]*)"(?: within "([^"]*)")?$/ do |field, selector|
```
redsummernight: I think we need to remove the selector argument too.

```
Your block takes 2 arguments, but the Regexp matched 1 argument. (Cucumber::ArityMismatchError)
features/step_definitions/web_steps.rb:109:in `/^(?:|I )choose "(.*)"$/'
/home/travis/.rvm/rubies/ruby-2.3.4/bin/bundle:30:in `block in <main>'
/home/travis/.rvm/rubies/ruby-2.3.4/bin/bundle:22:in `<main>'
features/admins/users/admin_abuse_users.feature:102:in `And I choose "Lift permanent suspension, effective immediately."'
```

sarken: Fuck me in the eye. Will do!

redsummernight: No, we need your eye(s). Should be good to merge when tests pass.

PR #3633, opened 2019-09-11
AO3-5639 Don't count hits on unrevealed and hidden works

CristinaRO:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5639

## Purpose

What does this PR do?

Previously, visiting a work that was part of an unrevealed collection was increasing the hit count, even though all a regular (non-admin) visitor could access was a message about the work being unrevealed, and not the work itself. This was upsetting or spooking out the creators, because "unrevealed" suggests no one can access the work other than its creator(s).

This PR adds a check that the work is not part of an unrevealed collection.

Previously, visiting a work that was hidden by an admin (which all admins can do) was increasing the hit count. This was upsetting to the creators, because "hidden" suggests no one can access the work other than its creator(s).

This PR adds a check that the work is not hidden by admin.

I've also consolidated the checks that prevent a hit count from being recorded into one method, to make the code more readable.

We have briefly discussed disregarding *all* admins visits from hit counts, but decided this was out of scope for this issue, and we do not feel strongly enough about this to allow the scope creep.

## Testing Instructions

In JIRA, thanks sarken.

## References

## Credit

Enigel

she/her

### DIFF
--- a/lib/work_stats.rb
+++ b/lib/work_stats.rb
@@ -47,10 +47,7 @@ module WorkStats
   end
 
   def increment_hit_count(visitor)
-    # skip if this is the same visitor as before or if the current user is the author of this work
-    # Or the request comes from a unicorn which sets the REQUEST_FROM_BOT enviroment variable
-    # We will set this in the unicorn which servers bots, which is chosen by nginx.
-    unless ENV['REQUEST_FROM_BOT'] || self.last_visitor == visitor || (User.current_user.is_a?(User) && User.current_user.is_author_of?(self))
+    unless disregard_visit?(visitor)
       set_last_visitor(visitor)
       add_to_hit_count(1)
       REDIS_GENERAL.sadd(WORKS_TO_UPDATE_KEY, get_work_id)
@@ -72,6 +69,21 @@ module WorkStats
       comments_count: self.count_visible_comments,
       bookmarks_count: self.bookmarks.where(private: false).count
     )
+  end
+
+  # Do not count the visit if:
+  # - This is the same visitor as before
+  # - The current user is a creator of this work
+  # - The request comes from a unicorn which sets the REQUEST_FROM_BOT enviroment variable
+  # (We will set this in the unicorn which serves bots, which is chosen by nginx.)
+  # - The work is hidden by an admin (admins can visit it but it shouldn't count as a hit)
+  # - The work is part of an unrevealed collection
+  private def disregard_visit?(visitor)
+    ENV['REQUEST_FROM_BOT'] ||
+    self.last_visitor == visitor ||
+    User.current_user.is_a?(User) && User.current_user.is_author_of?(self) ||
+    self.hidden_by_admin? ||
+    self.in_unrevealed_collection?
   end
 
   protected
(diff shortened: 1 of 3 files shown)


lib/work_stats.rb
```diff
@@ -74,6 +71,21 @@ def update_stat_counter
...
   end
 
+  # Do not count the visit if:
+  # - This is the same visitor as before
+  # - The current user is a creator of this work
+  # - The request comes from a unicorn which sets the REQUEST_FROM_BOT enviroment variable
+  # (We will set this in the unicorn which serves bots, which is chosen by nginx.)
+  # - The work is hidden by an admin (admins can visit it but it shouldn't count as a hit)
+  # - The work is part of an unrevealed collection
+  private def disregard_visit?(visitor)
```
sarken: I think we usually put `private` above the method rather than next to it. Is that possible here?

lib/work_stats.rb
```diff
@@ -47,10 +47,7 @@ def downloads
   end
 
   def increment_hit_count(visitor)
-    # skip if this is the same visitor as before or if the current user is the author of this work
-    # Or the request comes from a unicorn which sets the REQUEST_FROM_BOT enviroment variable
-    # We will set this in the unicorn which servers bots, which is chosen by nginx.
-    unless ENV['REQUEST_FROM_BOT'] || self.last_visitor == visitor || (User.current_user.is_a?(User) && User.current_user.is_author_of?(self))
+    unless disregard_visit?(visitor)
```
sarken: 😻

lib/work_stats.rb
CristinaRO: Sure, the `protected` below will switch modes. (I got this habit from work, where we prefer it in order to not accidentally `private`ise a whole section of file. 😅 )

CristinaRO: However, indenting stuff under a modifier is somewhat unusual, and I'd rather not do it. 😬  (I'll ignore the indented `protected` section because it's already there, but I'd rather not repeat the pattern.)

sarken: That's fine! I don't think we usually do it, tbh.

(I actually really like the explicitness of putting `private` on the same line as the method name, but probably best to stick with the style we already use.)

lib/work_stats.rb
```diff
@@ -74,6 +71,23 @@ def update_stat_counter
...
+  # Do not count the visit if:
+  # - This is the same visitor as before
+  # - The current user is a creator of this work
+  # - The request comes from a unicorn which sets the REQUEST_FROM_BOT enviroment variable
+  # (We will set this in the unicorn which serves bots, which is chosen by nginx.)
+  # - The work is hidden by an admin (admins can visit it but it shouldn't count as a hit)
+  # - The work is part of an unrevealed collection
+  def disregard_visit?(visitor)
+    ENV['REQUEST_FROM_BOT'] ||
+    self.last_visitor == visitor ||
```
tickinginstant: Why are you disregarding the Hound here? I'd argue that it's more confusing to eschew indentation here, given that it's a single expression spread across five lines. With no indentation on the last four lines, you have to look back at the previous line in order to see that it's a continuation of the previous line, which impacts the readability.

PR #3634, opened 2019-09-13
AO3-5740 Fix adding work to the series on work preview

lethnie:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5740

## Purpose

Adds work to the series without saving changes to the database when user is previewing work.
Saves series to the database only when user updates the work.

## Testing Instructions

Steps for testing are already described in JIRA ticket.

## Credit

lethnie, she/her


app/controllers/works_controller.rb
```diff
@@ -358,17 +368,17 @@ def update
     @work.set_challenge_claim_info
     set_work_form_fields
 
-    if params[:edit_button] || work_cannot_be_saved?
-      set_work_tag_error_messages
-      render :edit
-    elsif params[:preview_button]
+    if params[:preview_button]
```
tickinginstant: I don't think you can reorder the conditions here, at least not without more extensive changes. If someone enters invalid inputs and presses preview, they're not supposed to see a preview with those invalid inputs. Instead, they should be bounced back to the edit form with errors showing what they need to fix.

app/models/work.rb
```diff
@@ -608,6 +608,9 @@ def default_date
 
   # Virtual attribute for series
   def series_attributes=(attributes)
+    if self.preview_mode
```
tickinginstant: I think that the series isn't supposed to be saved immediately in two cases: (1) if the user presses "Preview", or (2) if the user has entered invalid data elsewhere on the form and pressed "Post."

In other words, if the user is bounced back to the edit form with a message saying, "Sorry! We couldn't save this work," then I don't think the changes made to the work's series should be saved. But that wasn't mentioned in the description of the issue, so maybe I'm misremembering the desired behavior?

If that is the desired behavior, though, this approach won't work, and you might want to play around with calling `serial_works.build(series: old_series)` instead of `self.series << old_series`.

app/controllers/works_controller.rb
lethnie: Got it, changed conditions order back.

app/models/work.rb
lethnie: Didn't think about saving errors case. Changed implementation so that work-series relationship is not being saved in both cases. Thanks for the tip!

app/controllers/works_controller.rb
```diff
@@ -283,6 +284,14 @@ def create
     @work = Work.new(work_params)
     @chapter = @work.first_chapter
     @chapter.attributes = work_params[:chapter_attributes] if work_params[:chapter_attributes]
+    if work_params[:series_attributes]
```
redsummernight, line 287: Is it possible to move most of this parsing logic into the work model, instead of having them in the work controller?

I think the setup can be similar to how we handle related works, with a method for holding the temporary state, both for the preview mode and for preserving info in case of form errors:

https://github.com/otwcode/otwarchive/blob/bedbbef8c43585d97b5a56062355921e4c8f5fa7/app/models/work.rb#L934-L935

and a method for actually saving the state:

https://github.com/otwcode/otwarchive/blob/bedbbef8c43585d97b5a56062355921e4c8f5fa7/app/models/work.rb#L973-L974

(It may not be 100% the case though, since related works don't show up in preview mode right now.)

app/views/works/_hidden_fields.html.erb
```diff
@@ -33,6 +33,10 @@
   <%= c.hidden_field :title, :value => "#{@chapter.title}" %>
   <%= c.hidden_field :published_at, :value => "#{@chapter.published_at}" %>
 <% end %>
+<%= fields_for "work[series_attributes]", @serial do |s| %>
```
redsummernight, line 36: When the form errors out, can we retrieve the unsaved series info from `work[series_attributes]`, instead of adding new hidden fields? Similar to:

https://github.com/otwcode/otwarchive/blob/bedbbef8c43585d97b5a56062355921e4c8f5fa7/app/helpers/works_helper.rb#L58-L63

redsummernight (commented):
Hi @lethnie! Sorry it has taken us a while to review this.

This bug is unfortunately not unique to series (for example, tags, which can similarly be created anew as part of updating a work, have the same issue - [AO3-3455](https://otwarchive.atlassian.net/browse/AO3-3455)), and for a long time we've been unsure of the right approach for a fix. I've left comments, but please do let me know if there are details I've overlooked.

app/controllers/works_controller.rb
lethnie: Hi! I'll take a look at it in a week or two if that's okay

redsummernight: That's totally fine, no rush!

lethnie: About similarity with related works behavior: related works are still saved if we set it up in editor, then open work's preview and cancel the changes, so I guess there is the same problem that I'm trying to fix for series. But I removed parsing logic from controller and changed it a bit.

There are some other things I think worth noting.
It took some time for me to update the system (vagrant etc; I'm working on Windows) with the last version; don't know if it's only my problem, but just in case - I couldn't load the site from the host (got ERR_CONNECTION_REFUSED error) until added this setting to Vagrantfile:
config.vm.network "forwarded_port", guest: 7000, host: 7000, host_ip: "127.0.0.1"

Also, while working on this problem, I discovered another one; not sure if it isn't already in Jira, but anyway here are the steps to reproduce it:
1. Check "This work is part of a series" and set a new series title in editor
2. Go to Preview
3. On Preview page, click "Edit" button
4. Open Preview again
Actual result: new series with set title has been added to the database several times. You can see it in a dropdown list back on Edit page.
I wasn't sure about the correct behavior here: can there be different series with the same title? If not, then it is possible to just check series by title before saving like it works with id now. Or maybe series title is unique for one user.

app/views/works/_hidden_fields.html.erb
lethnie: I've changed the code to get series id/title from work[series_attributes], but hidden fields are still necessary to save series when we save changes from Preview page, otherwise series_attributes data is not sent to the server

app/controllers/works_controller.rb
lethnie: Sorry, apparently the last changes don't seem to work properly, and this problem with series is the result (like several other failed test cases). So for now I can't see how to manage passing the new series id other than the way it was done before, using @serial.

lethnie: I've added series id update for new series on saving - not sure if this way is better than using `@serial`; but there are still some tests failing:
features/other_b/series.feature:76
features/other_b/series.feature:225
They fail locally, too, but when I try to reproduce the scenario in an app, everything seems to work just fine.
I tried my best to investigate, but at this point I have no idea what can go wrong there; series list in the footer displays all the series and the one at the top ignores one, even though they use the same data, from what I can see.

tickinginstant: I think one of the issues here is that the new series shouldn't be saved until the work is saved. [If you delete the last work from a series, it deletes the series](https://github.com/otwcode/otwarchive/blob/d2568009c73f19c24248a07c9bc27717460766fa/app/models/serial_work.rb#L20-L25), so I think creating empty series in the first place could be considered a bug. The code surrounding parameters should get a bit simpler if you delete this line:
https://github.com/otwcode/otwarchive/blob/c08b2b6159e415ac0df7519bc33875ef1a0e8bdb/app/models/work.rb#L621

([There are some tests which rely on the existing behavior](https://github.com/otwcode/otwarchive/blob/1f7610f25b28388bd6b2beb4715f2a80ece78ef7/features/users/user_delete.feature#L98-L109), but I don't think it's a good plan to keep the buggy behavior around just for testing purposes. It'd probably be better to modify that scenario by creating a `{user} has an empty series` step that uses a factory to create the empty series.)

As for the preview issue, that looks like it's caused by a caching issue related to [AO3-3455](https://otwarchive.atlassian.net/browse/AO3-3455) and [AO3-5041](https://otwarchive.atlassian.net/browse/AO3-5041). Caching is disabled in the development environment, but not in tests. As long as red approves, it makes sense to me to add the `@disable_caching` tag to those tests and add this to AO3-5041 (not AO3-3455, that's a slightly more complex issue) once this PR gets merged.

app/helpers/series_helper.rb
```diff
@@ -7,7 +7,11 @@ def show_series_data(work)
 
   # this should only show prev and next works visible to the current user
   def series_data_for_work(work)
-    series = work.series.select(&:visible?)
+    series = if work.preview_mode
+               work.series
```
tickinginstant, line 11: The `save_series` function seems to be designed to make this line of code work, but it might be simpler to change this line of code:
```suggestion
               work.serial_works.map(&:series).compact
```
The `serial_works` list is already being updated by the `series_attributes=` function, so we don't need extra code to update `work.series` if we read series info from `work.serial_works` instead.

(Alternatively, we could just add `work.new_series` to the list of series to be displayed.)

app/helpers/works_helper.rb
```diff
@@ -62,15 +62,15 @@ def work_parent_value(field)
...
-  # and added to the work even before the work is saved. The only time that the
-  # series isn't added is when the work is a new record, and therefore the
-  # SerialWork can't be created.
-  def work_series_id(work)
-    if work.new_record? && (series = work.series.first)
-      series.id
-    end
+  # Determines whether or not "manage series" dropdown should appear
+  def check_series_box(work)
+    work.series.present? || (params[:work] && !(work_series_value(:id).blank? && work_series_value(:title).blank?))
```
tickinginstant, line 67: Why check `params[:work]` here? Isn't that covered by `work_series_value`?

app/helpers/works_helper.rb
```diff
@@ -62,15 +62,15 @@ def work_parent_value(field)
...
+  end
+
+  # Passes value of fields for work series back to form when an error occurs on posting
+  def work_series_value(field)
+    return unless params[:work] && params[:work][:series_attributes]
+    params[:work][:series_attributes][field]
```
tickinginstant, line 73: I think [dig](https://api.rubyonrails.org/classes/ActionController/Parameters.html#method-i-dig) would be useful here.

app/models/work.rb
```diff
@@ -605,7 +606,8 @@ def series_attributes=(attributes)
         errors.add(:base, ts("You can't add a work to that series."))
         return
       end
-      self.series << old_series unless (old_series.blank? || self.series.include?(old_series))
+      self.new_series = old_series unless (old_series.blank? || self.series.include?(old_series))
+      self.serial_works.build(series: old_series) unless (old_series.blank? || self.series.include?(old_series))
```
tickinginstant, line 610: Why repeat this condition instead of enclosing both lines in a block?

app/controllers/works_controller.rb
lethnie: I checked locally - adding `@disable_caching` tag really helped with those tests!

redsummernight: Yeah, it should be fine to use `@disable_caching` and add a `TODO: AO3-5041 etc.` to eventually remove the tag.

app/views/works/_hidden_fields.html.erb
redsummernight: Makes sense, thanks!

app/models/work.rb
```diff
@@ -617,7 +621,8 @@ def series_attributes=(attributes)
         new_series.creatorships.build(pseud: pseud)
       end
       new_series.save
```
redsummernight, line 623: Can we remove this line, following @tickinginstant's comment [in the other thread](https://github.com/otwcode/otwarchive/pull/3634#discussion_r553664121), to avoid creating the new series if we can't save the serial work?

redsummernight:
There's [an unrelated security warning](https://otwarchive.atlassian.net/browse/AO3-6107) from Hakiri which you can fix by pulling from master.

app/models/work.rb
```diff
@@ -628,6 +632,13 @@ def adjust_series_restriction
     end
   end
 
+  # Save relationship to series if applicable
+  def save_series
```
redsummernight, line 636: `series_attributes` already sets up a series and associating it with the work, based on info extracted from `params`. I think you should be able to remove `save_series` completely and everything would still work.

features/step_definitions/user_steps.rb
```diff
@@ -166,6 +166,16 @@
   work.creatorships.unapproved.each(&:accept!)
 end
 
+Given /^"(.*?)" has an empty series "(.*?)"$/ do |login, title|
+  new_series = Series.new
```
redsummernight, line 170: You don't need to add all pseuds to the series, just one will do. You can write this method shorter like this:
```rb
series = Series.new(title: title)
series.creatorships.build(pseud: User.find_by(login: login).default_pseud)
series.save
```

redsummernight (commented):
Just two things, otherwise this looks good to me! (unless @tickinginstant have other comments)

redsummernight (approved):
Thank you!

A note for when we merge: there remains an issue with previewing series info on drafts, similar to [AO3-5041](https://otwarchive.atlassian.net/browse/AO3-5041). See scenarios "A work can be in two series" and "A work's series information is visible and up to date when previewing the work while posting or editing".

app/models/work.rb
```diff
@@ -70,6 +70,7 @@ def create_stat_counter
   # Can't write to work.pseuds until the work has an id
   attr_accessor :new_parent, :url_for_parent
   attr_accessor :new_recipients
+  attr_accessor :new_series
```
tickinginstant, line 73: Is this used anywhere? It's set in the `series_attributes=` function, but I don't see any references to it elsewhere.

app/models/work.rb
```diff
@@ -605,7 +606,10 @@ def series_attributes=(attributes)
         errors.add(:base, ts("You can't add a work to that series."))
         return
       end
-      self.series << old_series unless (old_series.blank? || self.series.include?(old_series))
+      unless old_series.blank? || self.series.include?(old_series)
+        self.new_series = old_series 
+        self.serial_works.build(series: old_series)
+      end
       self.adjust_series_restriction
```
tickinginstant, line 613: I think this line can be removed, since the series' restricted status shouldn't change until the new `SerialWork` is saved to the database.

tickinginstant (approved):
There are two things that I think could be removed to make the code a little shorter, but it looks good otherwise!

PR #3635, opened 2019-09-14
AO3-5652 Index more series data

elzj:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5652

## Purpose

Slightly changes and expands the series data being indexed with works. The behavior should be similar (with a small change from "series_titles:" to "series.title:") but the extra data should give us more to work with later on. General searches should also include the series title field, though at a lower priority than the work title.

## Testing Instructions

See instructions in issue.

## Credit

elz and redsummernight

### DIFF
--- a/app/models/search/work_indexer.rb
+++ b/app/models/search/work_indexer.rb
@@ -26,8 +26,8 @@ class WorkIndexer < Indexer
           tag: {
             type: "text"
           },
-          series_titles: {
-            type: "text"
+          series: {
+            type: "object"
           },
           authors_to_sort_on: {
             type: "keyword"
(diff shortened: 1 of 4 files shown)


app/models/serial_work.rb
```diff
@@ -12,6 +12,11 @@ class SerialWork < ApplicationRecord
 
   scope :in_order, -> { order(:position) }
 
+  def self.with_title
```
redsummernight: This is unused.

app/models/search/work_indexer.rb
```diff
@@ -26,8 +26,8 @@ def self.mapping
           tag: {
             type: "text"
           },
-          series_titles: {
-            type: "text"
+          series: {
+            type: "object"
```
redsummernight: For the record, [arrays of inner object fields are flattened](https://www.elastic.co/guide/en/elasticsearch/reference/6.0/nested.html), meaning the association between id, title, and position is lost when a work is indexed. However, we're not querying for combinations e.g. series title AND series position, we're only querying for series title, and the other 2 fields are there for eventual use in work blurbs ([AO3-5760](https://otwarchive.atlassian.net/browse/AO3-5760)). So we don't need to use the nested data type.

PR #3636, opened 2019-09-14
AO3-5613 Fix revealing hidden warning tags in work meta

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5613?focusedCommentId=355413&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-355413

## Purpose

If you enable the “Hide warnings” preference, the “Show warnings” link in the work meta should once again show the warning tags.

* Grabs the tests from #3290, which originally revealed the problem.
* Refactors `blurb_tag_block(item, tag_groups=nil)` so the code branch responsible for handling the "Show warnings..." is a separate method (`show_hidden_tag_link_list_item(item, category, options = {})`).
* Updates views/works/_meta to use the new method.

There were a few variable names I wanted to change and some methods that could have had better names, but I wanted to keep it clear for the reviewer how the new code relates to the old. 

## Testing Instructions

* Enable the preference for hiding warning tags. 
* Go to a work and make sure you can still show the warnings.
* Make sure there are no changes to how tags on work blurbs display/function normally.

## References

I was originally going to put this in #3290, which can't be merged without this fix, but decided it was better to have it as a separate pull request.

### DIFF
--- a/app/helpers/tags_helper.rb
+++ b/app/helpers/tags_helper.rb
@@ -187,9 +187,7 @@ module TagsHelper
     sub_ul.html_safe
   end
 
-
   def blurb_tag_block(item, tag_groups=nil)
-    item_class = item.class.to_s.underscore
     tag_groups ||= item.tag_groups
     categories = ['ArchiveWarning', 'Relationship', 'Character', 'Freeform']
     tag_block = ""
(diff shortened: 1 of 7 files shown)


app/helpers/tags_helper.rb
```diff
@@ -218,6 +213,30 @@ def blurb_tag_block(item, tag_groups=nil)
     tag_block.html_safe
   end
 
+  # Takes a tag category class name, e.g. Relationship, and returns a string.
+  # The returned string is plural and used for things other than the HTML class
```
redsummernight: I see the result of `tag_block_class_name` being used as a class in `blurb_tag_block`. Should this comment be changed?

sarken: Sure! I meant it was being used as not just an HTML class, but didn't phrase it clearly.

app/views/works/_meta.html.erb
```diff
@@ -22,8 +22,7 @@
           <dd class="<%= tag_type_css_class(type) %> tags">
             <ul class="commas">
               <% if (type == "ArchiveWarning" && hide_warnings?(work)) || (type == "Freeform" && hide_freeform?(work)) %>
-
-                <li id="work_<%= work.id %>_category_<%= type.tableize %>"><%= show_hidden_tags_link(work, type.tableize) %> </li>
```
tickinginstant: The old version of this code doesn't wrap the hidden tags link in `<strong>` tags, but I believe the new version does. So this will change the formatting slightly. Does the version number for the surrounding cache block need to be incremented, or is it a minor enough change that it's okay to have inconsistent formatting for a while?

sarken: I've bumped it -- thanks, I didn't realize the toggle was inexplicably not bold on the meta before.

PR #3637, opened 2019-09-14
AO3-5693 Support HTML5 media tags

elzj:
## Issue

## Purpose

Adds support for audio and video tags in the same fields that allow video embeds and sets up a blacklist option in the config. Also extracts the existing sanitize transformers out into separate files for improved comprehensibility.

## Testing Instructions

See issue for which attributes are allowed and which should be stripped out by the sanitizer.


lib/otw_sanitize/multimedia.rb
```diff
@@ -0,0 +1,73 @@
+module OTWSanitize
+  module Multimedia
+    # Attribute whitelists
+    AUDIO_ATTRIBUTES = %w[
+      class controls crossorigin dir
+      loop muted preload src title
+    ].freeze
+    VIDEO_ATTRIBUTES = %w[
+      class controls crossorigin dir height
+      loop muted poster preload src title width
```
redsummernight: Should "playsinline" be here too, if only for consistency?

config/config.yml
```diff
@@ -288,6 +288,8 @@ FIELDS_ALLOWING_VIDEO_EMBEDS: ["content"]
 
 FIELDS_ALLOWING_CSS: ["content", "endnotes", "notes"]
 
+BLACKLISTED_SRCS: []
```
redsummernight: Should this config have a more specific name, like `BLACKLISTED_MEDIA_SRCS` or `BLACKLISTED_HTML_MEDIA_HOSTS`?

It should have a comment describing what it does as well:
> The domain blacklist for src URLs of video, audio, source, and track elements.
> This will allow us to block embeds from certain domains in case of abuse.

sarken: +1 to `BLACKLISTED_MEDIA_SRCS`

sarken: Or, actually, what about `BLACKLISTED_MULTIMEDIA_SRCS` since it's used by `Multimedia`?

lib/otw_sanitize/media_sanitizer.rb
```diff
@@ -0,0 +1,99 @@
...
+        'video' => {
+          'controls'    => 'controls',
+          'playsinline' => 'playsinline',
+          'crossorigin' => 'anonymous',
+          'preload'     => 'metadata'
+        }
+      },
+      protocols: {
+        'video' => {
+          'poster' => %w[http https]
```
sarken: Why do we need to specify the allowed protocols for `poster` but not `src`?

spec/miscellaneous/lib/html_cleaner_spec.rb
```diff
@@ -198,6 +210,11 @@
           result = sanitize_value(field, value)
           expect(result).to eq("")
         end
+        it "should strip video" do
```
sarken: Could you add a blank line above this and just make it `it "strips video" do` without the `should`?

spec/miscellaneous/lib/media_sanitizer_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+        html = "
+          <video controls width='250'>
+            <source src='example.com/flower.webm' type='video/webm'>
+            <source src='example.com/flower.mp4' type='video/mp4'>
+            Sorry, your browser doesn't support embedded videos.
+          </video>"
+        content = Sanitize.fragment(html, config)
+        expect(content).to match("flower.webm")
+      end
+      it "removes unwhitelisted attributes" do
```
sarken: Can you add a blank line above this?

lib/otw_sanitize/embed_sanitizer.rb
```diff
@@ -0,0 +1,190 @@
...
+    # disable script access and networking
+    def disable_scripts(embed_node)
+      embed_node['allowscriptaccess'] = 'never'
+      embed_node['allownetworking'] = 'internal'
+
+      embed_node.search("param").each do |param_node|
+        if param_node[:name].downcase == "allowscriptaccess"
+          param_node.unlink
+        end
+        if param_node[:name].downcase == "allownetworking"
```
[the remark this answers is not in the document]
redsummernight: Should we do this?

sarken:
Just noting that I don't have any additional comments on `EmbedSanitizer`

sarken:
Just switching labels since this has been reviewed and needs a few tweaks.

lib/otw_sanitize/media_sanitizer.rb
elzj: It prevents you from putting scripts in, which was part of the spec, but I can add it to source too.

PR #3638, opened 2019-09-19
AO3-5755 Prevent inbox reply form from jumping in Chrome

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5755

## Purpose

Stops the inbox reply form from jumping around when it has focus in Chrome.

## Testing Instructions

Refer to Jira.


sarken:
Chrome fixed it as of Version 77.0.3865.90 (Official Build) (64-bit).

PR #3639, opened 2019-09-20
AO3-5613 Prevent error when challenge assignment email contains warning tags

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5613?focusedCommentId=355425&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-355425

## Purpose

Prevents error when sending challenge assignment emails containing warning tags. Also ensures the Warning or Warnings label shows up.

## Testing Instructions

* Create a gift exchange that allows users to including warnings in their requests and offers
* Have folks sign up for the exchange with warnings in requests and offers
* Run matching and send assignments

The participants should receive emails containing their assignment, including warning tags.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -106,6 +106,10 @@ en:
         other: "Warnings"
       work: "Work"
     attributes:
+      archive_warning:
+        name_with_colon:
+          one: "Warning:"
+          other: "Warnings:"
       category:
         name_with_colon:
           one: "Category:"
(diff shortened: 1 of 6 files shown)


config/locales/en.yml
```diff
@@ -106,6 +106,10 @@ en:
         other: "Warnings"
       work: "Work"
     attributes:
+      archive_warning:
```
sarken: Should I also update lines 104-106? I have no idea if it's used anywhere, but the model is no longer named Warning, so presumably something somewhere might be broken...

elzj: I'd say go for it. There's no warning model, so at best it will help, and at worst, it can't hurt.

redsummernight:
@elzj would you like to re-review?

PR #3640, opened 2019-09-20
AO3-5762 Only do search setup in rspec when needed

elzj:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5762

## Purpose

Instead of deleting and recreating all search indexes for every rspec test, only clear the ones needed for a specific spec file to cut down on run time.

Also updated the pseud query spec to update the syntax and also to change some specs that relied on an order that was not necessarily reliable.

## Testing Instructions

Only affects automated tests.


app/models/search/indexer.rb
```diff
@@ -51,6 +63,16 @@ def self.create_index(shards = 5)
     )
   end
 
+  def self.prepare_for_testing
+    raise "Wrong environment for test prep!" unless Rails.env.test?
+    delete_index
+    create_index
+  end
+
+  def self.refresh_index
```
redsummernight: We only do this for tests, should we add a guard clause like `prepare_for_testing`?

spec/spec_helper.rb
```diff
@@ -174,4 +139,4 @@ def get_message_part (mail, content_type)
     expect(email.body.parts.length).to eq(2)
     expect(email.body.parts.collect(&:content_type)).to eq(["text/plain; charset=UTF-8", "text/html; charset=UTF-8"])
   end
-end
+end
```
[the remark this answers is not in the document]
redsummernight: We should follow this.

spec/models/search/tag_query_wrangling_spec.rb
```diff
@@ -20,7 +20,11 @@
              freeform_string: "action,horror")
     end
 
-    before { run_all_indexing_jobs }
+    before(:each) {
```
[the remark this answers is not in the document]
redsummernight: We can leave out `:each` since it's the default, and use `do ... end` for this cop.

spec/spec_helper.rb
```diff
@@ -54,12 +53,26 @@
...
     DatabaseCleaner.clean
-    delete_test_indices
   end
 
   config.after :suite do
     DatabaseCleaner.clean_with :truncation
-    delete_test_indices
+  end
+
+  config.before :each, bookmark_search: true do
```
redsummernight: Should we have an after each counterpart for deleting the indices, in case tests not marked as using search try to use search?

spec/spec_helper.rb
```diff
@@ -108,61 +121,13 @@ def clean_the_database
   REDIS_RESQUE.flushall
   REDIS_ROLLOUT.flushall
   REDIS_AUTOCOMPLETE.flushall
-
-  ['work', 'bookmark', 'pseud', 'tag'].each do |index|
-    update_and_refresh_indexes index
-  end
-end
-
-def update_and_refresh_indexes(klass_name, shards = 5)
```
redsummernight: Thank you for removing this. :D

spec/models/search/pseud_query_spec.rb
```diff
@@ -22,61 +22,55 @@
       pseud_bar_2: create(:pseud, user: users[:user_bar], name: "foo"),
       pseud_aisha: create(:pseud, user: users[:user_aisha], name: "عيشة")
     }
-    update_and_refresh_indexes("pseud", 1)
```
redsummernight: We set the number of shards to 1 to get stable order of results for tag search (#3224). The new hooks don't do this (default multiple shards), so the order varies. Can we bring this back?

app/models/search/indexer.rb
```diff
@@ -51,6 +63,16 @@ def self.create_index(shards = 5)
...
+    $elasticsearch.indices.refresh(index: index_name)
```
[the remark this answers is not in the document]
redsummernight: I'll tweak this cop in [AO3-5758](https://otwarchive.atlassian.net/browse/AO3-5758) to allow "$elasticsearch".

spec/models/search/pseud_query_spec.rb
elzj: But all of our indexes have multiple shards on production. Shouldn't we want the tests to mimic the deployed behavior?

redsummernight: Generally yes, though in this case, we want to test that pseud searches favor exact matches more:

https://github.com/otwcode/otwarchive/blob/67b7b149d41ef626aca3596f2b1448b34d0cdc05/app/models/search/tag_query.rb#L122

It's hard to test TagQuery's behavior on multiple indices, so we check that its behavior on one index hasn't changed and assume that its behavior on multiple indices hasn't changed, unit-test-like.

ariana-paris: The problem wasn't the number of shards per se but how ES distributes and then gathers up data for sorting in the presence of tiny amounts of data (official explanations here: https://www.elastic.co/guide/en/elasticsearch/reference/current/consistent-scoring.html and here: https://www.elastic.co/guide/en/elasticsearch/guide/2.x/relevance-is-broken.html). The original discussion leading to the change I made in #3224 is here in one of our private channels: https://transformativeworks.slack.com/archives/G2CRYKP8W/p1520496182000310.

If we want the tests to reflect production and use multiple shards, we would need to load a lot more data in order to get a consistent relevance score that could be used for an ordering test, so our options were to load more data or use one shard. Or I guess, stop testing ordering 😬

spec/spec_helper.rb
elzj: It shouldn't matter unless we're testing search-specific behavior, and if we are, we should be using one of the search tags. Plus, not much is going to happen without the reindexer task being invoked. I'd kind of like to err on the side of reducing redundant calls and see if we have any issues with that going forward.

tickinginstant: I suspect the issues would arise from people new to the codebase, who might be unfamiliar with which functions require ES. They might find it easier to debug their new tests (and realize that they're unintentionally using ES) if they get an error about the indices not existing, instead of blank or incorrect results.

redsummernight: +1 to not leaking state between tests.

PR #3641, opened 2019-09-21
AO3-5758 Add Hound rules for bundler, layout, and style

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5758

## Purpose

Turn off some checks that we always ignore.

## Testing Instructions

None, it's just Hound.


config/.rubocop.yml
```diff
@@ -3,6 +3,15 @@
 AllCops:
   TargetRubyVersion: 2.3
 
+Bundler/OrderedGems:
+  Enabled: false
+
+Layout/DotPosition:
+  EnforcedStyle: leading
+
+Layout/MultilineMethodCallIndentation:
```
redsummernight: With these 2 layout rules, Hound will ask us to rewrite, for example:

https://github.com/otwcode/otwarchive/blob/67b7b149d41ef626aca3596f2b1448b34d0cdc05/app/controllers/api/v2/bookmarks_controller.rb#L64-L66

into:
```ruby
archivist_bookmarks = archivist_bookmarks
  .select { |b| b&.bookmarkable.is_a?(ExternalWork) ? b&.bookmarkable&.url == current_bookmark_url : false }
  .map    { |b| [b, b.bookmarkable] }
````

File: config/.rubocop.yml [hunk 2]
```diff
@@ -3,6 +3,18 @@
...
   TargetRubyVersion: 2.3
 
+Bundler/OrderedGems:
+  Enabled: false
+
+Layout/DotPosition:
+  EnforcedStyle: leading
+
+Layout/IndentArray:
+  EnforcedStyle: consistent
```
redsummernight:
```ruby
# consistent
array = [
  :value
]
and_in_a_method_call([
  :no_difference
])
```

config/.rubocop.yml
sarken: Should we do this? I think we use the first style more than the second.

redsummernight: I can switch to `indented_relative_to_receiver`.

sarken: I mean, I really don't know what the thing to do here is! I do see a benefit to using what you've put here -- less indenting means more room for your line of code, which is nice. However, I was under the impression we've mainly used the first approach over the years, which would mean getting a lot more Hound complaints as we edit stuff. 

So I really am asking, does my perception of what we've been doing align with others', and if so, should we make this change anyway?

elzj: I like the leading indentation, but I've also always been a trailing dot person - when we started the project, we were still on Ruby 1.8, which didn't even allow the leading dot syntax. So there's got to be a lot of that in the codebase.

sarken: We were definitely still using the trailing dot due to necessity as late as 2015 -- I vividly remember the experience of finding out leading dots didn't always work right while working on the homepage code.

redsummernight: MultilineMethodCallIndentation: we have been inconsistent over time. Here are the Rubocop violation counts in the repo (app, lib) for each style:

- indented: 172
- indented_relative_to_receiver: 183
- aligned: 157

Hound's default is "aligned".

---

DotPosition:

- leading: 188
- trailing: 40

I like the leading style better (https://github.com/otwcode/otwarchive/pull/3471#discussion_r234980705) but I can leave it alone.

tickinginstant: There are also some advantages to trailing: (a) it makes it easier to copy/paste code into the console to test it, and (b) it makes it easier to see that the first line is incomplete, without having to look at the next line.

There's some detailed discussion of pros and cons here: https://github.com/rubocop-hq/ruby-style-guide/pull/176

sarken: Easier copying and pasting to the console is a good point and makes me lean even more strongly toward sticking with trailing, as does the note about comments in https://github.com/rubocop-hq/ruby-style-guide/pull/176#issuecomment-20421523.

sarken: ...however, upon further thought, if we switch our indenting to plain old `indented`, I will be 100% in favor of leading-dot. It does make a huge difference in being able to understand wtf is going on in that case. 
```
archivist_bookmarks = archivist_bookmarks
  .select { |b| b&.bookmarkable.is_a?(ExternalWork) ? b&.bookmarkable&.url == current_bookmark_url : false }
  .map    { |b| [b, b.bookmarkable] }
```

vs
```
archivist_bookmarks = archivist_bookmarks.
  select { |b| b&.bookmarkable.is_a?(ExternalWork) ? b&.bookmarkable&.url == current_bookmark_url : false }.
  map    { |b| [b, b.bookmarkable] }
```

But if we use `aligned` or `indented_relative_to_receiver`, I think we should keep our dots `trailing`. The deep indenting is a pretty clear signal that the second line is a chained method call.

redsummernight: So we have 2 options:
1. `indented` + `leading`
2. `aligned`/`indented_relative_to_receiver` + `trailing`

I like `indented` because the later lines won't have to be re-indented any time the first line changes, less diffs. I'd prefer 1.

CristinaRO: I prefer 1 as well.

sarken: Let's 🐹 this at 1, then!

PR #3642, opened 2019-09-23
AO3-5765 Rails 5.2

elzj:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5765

## Purpose

Upgrade to Rails 5.2, and upgrade gems to match.

Changes:
- Uses Ruby 2.6.3, but it should also work on 2.5 if we want to downgrade that. Does not work with 2.3.4
- Switched back to the main globalize repo
- Upgraded testing gems, which required some updates to cucumber tests. Notably, cucumber now gives you regex-captured numbers as integers, not strings, and returns zero instead of nil when a value isn't found.
- Changed collection item removal process, which had previously involved a destroy inside a save callback, which was no longer working
- Changed search header construction so it doesn't involve string mutation
- Replaced deprecated Fixnum class with Integer
- Fixed a collection cleanup callback that gets invoked when a challenge is destroyed
- Fixed a comment update callback that was also trying to call save within a save
- Made a central 'random_order' scope so we only needed to escape the sql in one place. Also escaped it in a language sorting scope - the order method will no longer accept unescaped sql in Rails 6.
- Added a hook to clean up page caching when testing and uncommented an old test that had previously been failing

## Testing Instructions

Test all the things.


sarken:
Just checking so I update Jira correctly... does this contain fixes for the issues under the epic (AO3-5766 through AO3-5768)?

elzj:
> Just checking so I update Jira correctly... does this contain fixes for the issues under the epic (AO3-5766 through AO3-5768)?

It does, yes.

app/models/comment.rb
```diff
@@ -65,19 +63,19 @@ def akismet_attributes
...
+  def moderate_update
+    return if !moderated_commenting_enabled? || is_creator_comment?
+    self.unreviewed = true if significant_content_change?
+  end
+
   def after_update
     users = []
     admins = []
 
-    if self.saved_change_to_edited_at? && self.saved_change_to_comment_content? && self.moderated_commenting_enabled? && !self.is_creator_comment?
```
tickinginstant: Why did you stop checking whether `edited_at` has changed?

(While I do think the rewritten code is cleaner, it makes it harder to see how the code has changed, and what the issue was: `self.comment_content_was` vs `self.comment_content_before_last_save`.)

elzj: I couldn't think of a scenario where that would matter, but now I'm pretty sure it's so we don't catch comments marked deleted (which have their content changed). I'll put it back and make a note to make that a little more explicit.

tickinginstant: I think it might also matter if the sanitizer makes significant changes. (Say, if you disallow an HTML tag that used to be permitted.)

tickinginstant:
> Fixed a comment update callback that was also trying to call save within a save

Would you mind changing this description? It might confuse people, since Rails still does allow a save within a save:

https://github.com/otwcode/otwarchive/blob/b2ab5e7c722f8ee99dcd7c5ef33f2fe93c22b7a6/app/models/user.rb#L35

https://github.com/otwcode/otwarchive/blob/b2ab5e7c722f8ee99dcd7c5ef33f2fe93c22b7a6/app/models/user.rb#L461-L469

The issue with the comment update callback was this deprecation warning, which has been kicking around in the automated test suite for a year or two:
>DEPRECATION WARNING: The behavior of `attribute_was` inside of after callbacks will be changing in the next version of Rails. The new return value will reflect the behavior of calling the method after `save` returned (e.g. the opposite of what it returns now). To maintain the current behavior, use `attribute_before_last_save` instead. (called from after_update at /home/travis/build/tickinginstant/otwarchive/app/models/comment.rb:75)

Gemfile
```diff
@@ -108,7 +105,7 @@ gem 'google_visualr', git: 'https://github.com/winston/google_visualr'
 
 # Globalize for translations
 # Must use master branch and activemodel-serializers-xml for Rails 5 upgrade
-gem 'globalize', git: 'https://github.com/panorama-berlin/globalize'
+gem 'globalize', git: 'https://github.com/globalize/globalize'
```
sarken: Should the comment on the line above this be removed?

features/other_b/subscriptions_fandoms.feature
```diff
@@ -10,10 +10,9 @@ Feature: Subscriptions
   When I am logged in as "reader"
     And I view the "F/F" works index
   Then I should see "RSS Feed"
-  # TODO: If you uncomment the next line, it fails horribly. Is this a genuine bug, or a cucumber error?
-  # When I follow "RSS Feed"
-  # Then I should not see "My Work Title"
-  #   And I should not see "Stargate SG-1"
+  When I follow "RSS Feed"
+    Then I should not see "My Work Title"
```
sarken: The `Then` is indented two spaces too many here.

Gemfile
```diff
@@ -1,12 +1,12 @@
...
 
-ruby '2.3.4'
+ruby '2.6.3'
 
 gem 'test-unit', '~> 3.2'
 
 gem 'bundler'
 
-gem 'rails', '~> 5.1.6.2'
+gem 'rails', '~> 5.2.3'
```
redsummernight: We need to bump this to the latest patch.

sarken:
Updated in #3862

sarken:
I've labeled this Deferred or Superseded since Enigel is working on an updated version. We can close this once we feel we no longer need it for reference.

redsummernight:
We've merged the updated version, #3862.

PR #3643, opened 2019-09-23
AO3-5194 Update schema and structure files

elzj:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5194

## Purpose

Updates our database files to reflect current configuration. This is an automatic dump from a clean database from a clean codebase, so I'm not sure we want to make any manual edits to it.

## Testing Instructions

No testing required.

## References

Supersedes https://github.com/otwcode/otwarchive/pull/3115 which is out of date.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -1,4 +1,3 @@
-# encoding: UTF-8
 # This file is auto-generated from the current state of the database. Instead
 # of editing this file, please use the migrations feature of Active Record to
 # incrementally modify your database, and then regenerate this schema definition.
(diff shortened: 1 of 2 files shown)


db/structure.sql
```diff
@@ -1,86 +1,121 @@
...
+/*!40111 SET @OLD_SQL_NOTES=@@SQL_NOTES, SQL_NOTES=0 */;
+DROP TABLE IF EXISTS `abuse_reports`;
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+/*!40101 SET character_set_client = utf8 */;
 CREATE TABLE `abuse_reports` (
-  `id` int(11) NOT NULL AUTO_INCREMENT,
-  `email` varchar(255) DEFAULT NULL,
-  `url` varchar(255) NOT NULL,
-  `comment` text NOT NULL,
+  `id` bigint(20) NOT NULL AUTO_INCREMENT,
```
redsummernight: We should generate these files from production. Most primary keys there are not migrated to bigint yet.

db/structure.sql
```diff
@@ -1,86 +1,121 @@
...
+  `summary` varchar(255) COLLATE utf8mb4_unicode_ci DEFAULT NULL,
+  `summary_sanitizer_version` varchar(255) COLLATE utf8mb4_unicode_ci DEFAULT NULL,
+  `language` varchar(255) COLLATE utf8mb4_unicode_ci DEFAULT NULL,
+  `username` varchar(255) COLLATE utf8mb4_unicode_ci DEFAULT NULL,
   PRIMARY KEY (`id`)
-) ENGINE=InnoDB DEFAULT CHARSET=utf8;
-
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
+/*!40101 SET character_set_client = @saved_cs_client */;
+DROP TABLE IF EXISTS `admin_activities`;
```
redsummernight: The last time we updated this file, we removed the DROP TABLE statements.

db/schema.rb
```diff
@@ -9,1324 +8,1301 @@
 # from scratch. The latter is a flawed and unsustainable approach (the more migrations
 # you'll amass, the slower it'll run and the greater likelihood for issues).
 #
-# It's strongly recommended to check this file into your version control system.
+# It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(:version => 20160706031054) do
+ActiveRecord::Schema.define(version: 20190611212339) do
```
redsummernight: If we bump the version here, we need to check the existing PRs for migrations and bump their timestamps to later.

sarken: I don't think that's a problem -- we originally created the Has Migrations label to remind us to check the migration date and potentially update it following a schema change.

sarken: Left a comment on the two PRs that will need updating.

db/structure.sql
redsummernight: Or maybe the dumpers didn't use to add them, I'm not sure. It's fine to leave these be.

PR #3644, opened 2019-09-27
AO3-5037 Update Ruby to 2.4.5

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5037

## Purpose

Updates Ruby to 2.4.5. cherry-picks from and replaces https://github.com/otwcode/otwarchive/pull/3535, which I can't seem to push to for some reason.

## Testing Instructions

Refer to Jira.

## Credit

Matt Sears of Littlelines, he/him

PR #3645, opened 2019-09-29
AO3-5613 Avoid breaking work warnings for mass import API

redsummernight:
## Issue

## Purpose

The mass import API used to expect a request like this:

```
POST https://test.archiveofourown.org/api/v2/works/
Accept: application/json
Content-Type: application/json

{
  "archivist": "0efnaqgw1",
  "works": [
    {
      "id": "123",
      "title": "API Title",
      "warnings": "Major Character Death",
      "chapter_urls": [
        "http://example.com"
      ]
    }
  ]
}
```

After #3526, we expect "archive_warnings" instead of "warnings" from the temporary import sites, but we can revert that and keep import sites working.

## Testing Instructions

(Open Doors) Import a work with different warnings through the API.


redsummernight:
As a precaution, I'm also rebuilding #3644 with changes from this PR: https://travis-ci.org/redsummernight/otwarchive/builds/590995121.

PR #3646, opened 2019-10-04
AO3-5761 Add Rack Attack for basic rate limiting

elzj:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5761

## Purpose

Adds Rack Attack middleware with basic configuration for rate limiting.

## Testing Instructions

Generate a huge number of requests and see what happens. Otherwise, it's a fairly isolated change.


sarken:
This has merge conflicts.

config/config.yml
```diff
@@ -160,6 +160,12 @@ MAX_OPTIONS_TO_SHOW: 20
 MAX_KUDOS_TO_SHOW: 50
 MAX_FAVORITE_TAGS: 20
 
+# Used by Rack Attack, periods are measured in seconds
+RATE_LIMIT_NUMBER: 300
+RATE_LIMIT_PERIOD: 300
+RATE_LIMIT_LOGIN_ATTEMPTS: 5
```
sarken: Don't we have a limit on login attempts through Devise?

sarken: Oh, wait, is the difference that this will prevent them from submitting the form again, while Devise doesn't?

zz9pzza: So one comment is the processing in rack attack is done at a "lower" level and is faster.

elzj: It was part of the default config, and I figured it couldn't hurt. We don't need people hammering the login form even if we have extra layers of protection.

PR #3647, opened 2019-10-05
AO3-3847 Expire cached work blurbs and metas after renaming series

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-3847

## Purpose

Update the work blurbs and meta headers after renaming a series.
Also remove unused module CacheHelper.

## Testing Instructions

See issue.


redsummernight:
It just occurred to me that we should expire the cached work metas.

PR #3649, opened 2019-10-06
AO3-5632 Set Referrer-Policy header

elzj:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5632

## Purpose

Adds a referrer policy to our headers.

## Testing Instructions

Check the response headers on staging when you request a page. See issue for details.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -80,7 +80,12 @@ module Otwarchive
 
     # Only send referrer information to ourselves
     config.action_dispatch.default_headers = {
-      'Referrer-Policy' => 'same-origin'
+      "Referrer-Policy" => "same-origin",
+      "X-Frame-Options" => "SAMEORIGIN",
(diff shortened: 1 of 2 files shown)


config/application.rb
```diff
@@ -77,5 +77,10 @@ class Application < Rails::Application
 
     # Bring the log under control
     config.lograge.enabled = true
+
+    # Only send referrer information to ourselves
+    config.action_dispatch.default_headers = {
```
bingeling: Of course I only ever have questions after approving... Sorry, sigh. Does this overwrite the default header – which contains a bunch of things according to the guide (https://guides.rubyonrails.org/configuring.html#configuring-action-dispatch) – with only this? And is that because we don't want the other things?

elzj: You know, I assumed that it would merge, because that's what Rails usually does, but I think you're right that we need to specify all of them.

PR #3656, opened 2019-10-17
AO3-4139 Reversi: Make notes on required fields easier to read

niconicosette:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4139

## Purpose

the notes on required fields are hard to read - this PR changes the text colour to white so that it can be read more easily

## Testing Instructions

set your site skin to reversi, then try creating a new work or a new collection, and check that the notes on required fields have white text

### DIFF
--- a/public/stylesheets/masters/reversi/reversi_site_screen_.css
+++ b/public/stylesheets/masters/reversi/reversi_site_screen_.css
@@ -329,6 +329,10 @@ fieldset fieldset.listbox {
   outline: none;
 }
 
+fieldset .footnote, .character_counter, #comment-field-description {
+  color:#FFF;
+}
+
 .mce-container input:focus {
   background: #F3EFEC;
 }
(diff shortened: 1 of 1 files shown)


public/stylesheets/masters/reversi/reversi_site_screen_.css
```diff
@@ -329,6 +329,10 @@ fieldset fieldset.listbox {
   outline: none;
 }
 
+fieldset .footnote, .character_counter, #comment-field-description {
```
sarken: I think the best way to fix this is actually doing to be 

```
form dd.required {
  background: #eee;
}
```

As the issue points out, the problem is only with required fields... and that's because the default stylesheet uses that selector to set the text to our default text color of #2a2a2a (otherwise, in some cases, it would show up in the red we use for required fields). 

So here we need to apply Reversi's default text color of #eee to the text of required fields.

PR #3658, opened 2019-10-18
AO3-5773 Avoid re-escaping ampersands in titles of external work bookmarks

TurkeyFried:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5773

## Purpose

External Works have some characters such as ampersands encoded when posting, even when attempting to edit as an Admin. This PR makes sure the special characters are saved to the DB instead of their HTML encoded versions.

## Testing Instructions

Just need to repeat the Jira ticket's steps on master vs on this branch.

## Credit

bird

I like they :)

### DIFF
--- a/app/models/external_work.rb
+++ b/app/models/external_work.rb
@@ -45,9 +45,8 @@ class ExternalWork < ApplicationRecord
   end
 
   # Allow encoded characters to display correctly in titles
-  before_save :decode_title
-  def decode_title
-    self.title = HTMLEntities.new.decode(self.title)
+  def title
+    read_attribute(:title).try(:html_safe)
   end
 
   ########################################################################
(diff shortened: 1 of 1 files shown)


app/models/external_work.rb
```diff
@@ -44,6 +44,12 @@ def set_url_status
     self.update_attribute(:dead, true) unless url_active?(self.url)
   end
 
+  # Allow encoded characters to display correctly in titles
```
sarken: I think we'd prefer to take the same approach we're using to resolve this problem on works:

https://github.com/otwcode/otwarchive/blob/2ff9cebf7a92f442e98e8c9874e8b50ca2c7f448/app/models/work.rb#L76-L78

That will let us keep the titles encoded in the database and just change how they're displayed, which means we won't have to worry about cleaning up existing titles in the database.

TurkeyFried: That makes sense! Testing the update first, then will push to this branch.

PR #3659, opened 2019-10-19
AO3-5794 Update Whenever schedule file

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5794

## Purpose

Add cron jobs for reindexing. These are currently used in production.

## Testing Instructions

See issue.

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -73,6 +73,16 @@ every 10.minutes do
   rake "readings:to_database"
 end
 
+# Reindex works
+every 7.minutes do
+  rake "search:reindex_works"
+end
+
+# Reindex bookmarks
+every 13.minutes do
+  rake "search:reindex_bookmarks"
+end
+
 # Rerun redis jobs
 every 10.minutes do
   rake "resque:run_failures"


config/schedule.rb
```diff
@@ -73,6 +73,16 @@
   rake "readings:to_database"
 end
 
+# Reindex works
+every 7.minutes do
+  rake "search:reindex_works"
```
tickinginstant: Where are these rake tasks defined?

redsummernight: https://github.com/otwcode/otwarchive/blob/2ff9cebf7a92f442e98e8c9874e8b50ca2c7f448/lib/tasks/search.rake#L14-L21

Good call... these aren't the kind of tasks to be run on the regular. These are defined in the schedule.rb files on beta, but I checked again and they're not actually in `crontab -l`. I'll ask the sysadmins. Hopefully we won't need this PR.

tickinginstant: But those are `index_works` and `index_bookmarks`, not `reindex_works` and `reindex_bookmarks`. Is `reindex` a typo?

redsummernight: I should slow down. Yeah, I can't find those tasks.
```
# bundle exec rake -T | grep search
rake search:index_bookmarks                                 # Reindex bookmarks
rake search:index_pseuds                                    # Reindex pseuds
rake search:index_tags                                      # Reindex tags
rake search:index_works                                     # Reindex works
rake search:timed_all                                       # Reindex all
rake search:timed_bookmarks                                 # Reindex bookmarks
rake search:timed_pseud                                     # Reindex psueds
rake search:timed_tags                                      # Reindex tags
rake search:timed_works                                     # Reindex works
```
We don't need this PR.

PR #3661, opened 2019-10-19
[WIP] AO3-5793 Docker support

tlee911:
# Pull Request Checklist

* [ ✔️] Have you read [How to write the perfect pull request](https://github.com/blog/1943-how-to-write-the-perfect-pull-request)?
* [✔️] Have you read through the [contributor guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [✔️] Have you added tests for any changed functionality?
* [✔️] Have you added the [JIRA](https://otwarchive.atlassian.net) issue number as the *first* thing in your pull request title (eg: `AO3-1234 Fix thing`)
* [❌ ] Have you updated or commented on the JIRA issue with the information below? **No comment permissions**

## Issue

https://otwarchive.atlassian.net/browse/AO3-5793

## Purpose

Easier setup for local development using docker.

Side-effects:
- Had to upgrade gem mysql2 to 0.5.2 (current latest) otherwise app was crashing on startup
- Noticed that config/environments/development.rb was forcing localhost for memcache rather than reading from config

## Testing Instructions

Note that this simulates a new setup (empty DB, no dependencies available)
1.  Install Docker (https://www.docker.com/get-started)
2. Create config/database.yml with defaults:host:db
3. Create config/local.yml with 
> ES_URL: es:9200
> MEMCACHED_SERVERS: mc:11211
4. Run `docker-init.sh` (or do the commands within manually, but this seems easier)
5. Run `docker-compose up`
6. Browse to http://localhost:3000

## References

## Credit


sarken:
> No comment permissions

Sorry about that -- I've given them to you now!

redsummernight:
> Create config/database.yml with defaults:host:db
> Create config/local.yml with

I think you can automate this part as well, by having `.docker.example` versions of them that you copy to the right place in either Dockerfile or the setup script, like https://github.com/otwcode/otwarchive/blob/2ff9cebf7a92f442e98e8c9874e8b50ca2c7f448/script/travis_configure.sh.

gbhsrspm:
Neat! :) And yeah, another vote for having `docker.example` config included. :) I tried this quickly yesterday; unfortunately the rake tasks in `docker-init.sh` failed with a mysql connection error. I think I was missing some config in my `database.yml`, since mysql was running in the containers when I checked interactively. (I'm using an oldish system and also already had mysql running on 3306, so had to change the port config and restart some services, so it might be some quirk of my setup... but sample config would make it easier to check.)

I agree on the Jira note about splitting out the mysql2 upgrade into its own PR; sounds like a change that could potentially have knock-on effects in unexpected places.

Misc. notes for future docs:
* If the user changes `docker-compose.yml`, they need to run  `docker-compose down` before rerunning `docker-init.sh` for the changes to be picked up. (I forgot about this and got very confused)
* Docker eats harddrive space and doesn't clean up after itself. Good to keep an eye on disk space and run manual cleaning commands every so often (eg: `docker $whatever prune`, `docker rmi $whatever`. (especially relevant on this project since people are more likely to be using personal devices).

Thanks!

tlee911:
> I think you can automate this part as well, by having `.docker.example` versions of them that you copy to the right place in either Dockerfile or the setup script

Good point.  I will provide the example files.  I'm a bit wary of having them copied automatically as someone might accidentally overwrite their own local config just by running the init script.

> since mysql was running in the containers when I checked interactively. 

Ah yes, I used default ports everywhere without thinking about it, but there will be port binding problems if you have any of those services running locally already.  I'm inclined to leave the defaults and note in the documentation that ports can be changed if need.  And there's no reason someone couldn't use docker for everything except the DB if they wanted to do that.  They'd just have to adjust their config to use the correct hostname/IP.

> Misc. notes for future docs:

Thanks!  I have a bunch of notes too, which I will consolidate into docs/wiki.

And finally, there probably should be a warning that `docker-compose down` is potentially destructive.  Your (docker) DB will be wiped, so tread lightly before putting in a bunch of data you want to keep.

gbhsrspm:
> I'm inclined to leave the defaults and note in the documentation that ports can be changed if need

+1. I mostly noted it because if I just said 'mysql connection error' it might look like a code issue rather than an issue with my dev env being inspired by Frankenstein's monster, and I didn't want anyone to get sucked into debugging. :D 

(Yeah, all the cleanup commands need warnings, too.)

gbhsrspm:
Hi! I got a chance to try this again. It didn't work for me locally at first; it's been a couple of years since I used docker so it's possible I'm doing something ridiculous, but I thiiiink that when started from scratch, these instructions currently build the services, but don't start the containers. Then they try to run the rake tasks, which fail because the db container hasn't got up and running. This then means rails fails to start because it has no db to work with. 

I hacked things into working locally by moving the rake commands from the shellscript into the 'web' 'command' section of the docker-compose.yml (also put a sleep statement in there at the start so mysql had time to get ready, though not 100% sure if that's needed, and feels like a horrible hack anyway!), and just running `docker-compose up`. There is probably a much cleaner way to do this, and it's possible I've done it a really convoluted way because I missed an obvious step that wasn't written out. But I figure even if I'm 9000% wrong I'll just provide comedy for anyone reading this, so hey. :) Hope it helps!

gbhsrspm:
Oh also elasticsearch needs extra virtual memory to work; on Ubuntu I had to do:
`sudo sysctl -w vm.max_map_count=262144`
(https://www.elastic.co/guide/en/elasticsearch/reference/6.0/vm-max-map-count.html). (The error shows up in the logs for the es container, and is easy enough to search, but those logs can get obscured by the logs for the db transitions and so on, so probably still worth documenting.)

gbhsrspm:
Hey! I got some time to clean up my local branch a bit. I don't intend to send it as a PR anytime soon because the changes are very hacky (and I put the ascii unicorns back, on a whim), but it works for me for local dev, and it's probably easier to look at the code there than my review comments that try to describe the same. It's over at: https://github.com/gbhsrspm/otwarchive/commits/ao3-docker . I intend to keep it updated for my own use, though no timescale for when I'll try to make it less hacky. But it may be useful in a pinch for anyone else wanting to use docker. :whale: 

(Yeah this whole thing was an excuse to use that emoji.)

curtisxk38:
I've been playing around with this. I think I have a slightly less hacky way to wait for mysql to be up. You can ping it with netcat and see if mysql is ready for connections.
see [here](https://github.com/curtisxk38/otwarchive/blob/e2974fc6434dbac67e89b0829d3ddc8789d5970f/script/docker_web_rails.sh#L21-L49) 

Also you can run elasticsearch in dev mode, so that you `vm.max_map_count` error just becomes a warning instead of stopping execution.
see [here](https://github.com/curtisxk38/otwarchive/blob/e2974fc6434dbac67e89b0829d3ddc8789d5970f/docker-compose.yml#L20)

About the tests, they fail since `redis_test_setup.rb` tries to run `redis-server` which isn't installed on the container running rails. I think the fix for this would be:

1. Stop `redis_test_setup.rb` from trying to run `redis-server`. maybe by checking for a `DOCKER` environment variable

2. Have our redis container run `config/redis-cucumber.conf.example`. I think this would require making our own docker image from `redis:5` then `COPY`ing in the `.conf` and passing it to the `redis-server` command.

3. maybe something else? never used cucumber before

curtisxk38:
Got the rspec tests to work. Well, they're still running, but I think it works. 

Didn't end up doing anything with `redis-cucumber.conf.example`, since it doesn't seem like this config does anything special. I just added a check to stop `start_redis!` from running.

One thing you need to do is to configure mysql so it doesn't have `ONLY_FULL_GROUP_BY`, see [here](https://github.com/curtisxk38/otwarchive/commit/23d33509c314e5a13da29dea37cd4600a218ed5f)

gbhsrspm:
> I've been playing around with this. I think I have a slightly less hacky way to wait for mysql to be up. You can ping it with netcat and see if mysql is ready for connections.
> see [here](https://github.com/curtisxk38/otwarchive/blob/e2974fc6434dbac67e89b0829d3ddc8789d5970f/script/docker_web_rails.sh#L21-L49)

Ah, cool! I'll try this out sometime. :)

> Also you can run elasticsearch in dev mode, so that you `vm.max_map_count` error just becomes a warning instead of stopping execution.
> see [here](https://github.com/curtisxk38/otwarchive/blob/e2974fc6434dbac67e89b0829d3ddc8789d5970f/docker-compose.yml#L20)
Ah, nice, yeah, it makes much more sense to run it in dev mode than to change host config each time (which also required setting it again every time the host was rebooted).

Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM ruby:2.6.5
+RUN apt-get update && apt-get install -y default-mysql-client
+
+WORKDIR /otwa
+COPY Gemfile .
+COPY Gemfile.lock .
+RUN gem install bundler -v 1.17.3 && bundle install
+
+COPY . .
```
elebow: I don't believe we need to copy the app into the image because we're mounting it as a volume at runtime.

The rake tasks in `docker-init.sh` don't run that way, but I think that's just because the volume isn't mounted yet. That can probably be avoided by starting the web service (thereby mounting the volume) before running the rake tasks.

tommilligan:
I'm interested in helping get this over the line into master. From reading everyone's different branches and suggestions, I'm at the point where _almost_ all of the `rspec` tests pass for me locally in Docker, by applying [this commit](https://github.com/tommilligan/otwarchive/commit/441d03d6dd21614bf1c3e01148932b66cfbbe487) on top of this PR.

The ones that don't seem to (almost?) all be caching-related failures:
```log
$ docker-compose run web bundle exec rspec spec

...snip...

Failed examples:

rspec ./spec/controllers/bookmarks_controller_spec.rb:176 # BookmarksController index with caching returns the same result the second time when a new bookmark is created within the expiration time
rspec ./spec/controllers/works/default_rails_actions_spec.rb:316 # WorksController index with caching with NO owner tag returns the same result the second time when a new work is created within the expiration time
rspec ./spec/helpers/works_helper_spec.rb:88 # WorksHelper#sorted_languages returns all languages sorted alphabetically
rspec ./spec/lib/tasks/after_tasks.rake_spec.rb:17 # rake After:reset_word_counts when there are multiple languages updates only works in the specified language
rspec ./spec/lib/zoho_auth_client_spec.rb:41 # ZohoAuthClient#access_token caches the access token from the response
rspec ./spec/lib/zoho_auth_client_spec.rb:48 # ZohoAuthClient#access_token returns the cached token if it's available
rspec ./spec/models/tag_spec.rb:32 # Tag checking count caching without updating taggings_count_cache will start caching a when tag when that tag is used significantly
rspec ./spec/models/tag_spec.rb:56 # Tag checking count caching updating taggings_count_cache will start caching a when tag when that tag is used significantly
rspec ./spec/models/work_spec.rb:449 # Work#find_by_url gets the work from cache when searching for an imported work by URL
```

The full tracebacks are [on pastebin here](https://pastebin.com/0H9fF7HM).

If someone with more prior knowledge can spot what links these failures, I'm happy to try and fix them, I'm just not sure where to start.

zz9pzza:
I am willing to have a look at this as wsl2 has made docker a lot more useable.

tommilligan:
@zz9pzza that sounds great. Let me know anything I can do to help.

FWIW I've now got the `cucumber` tests working with [this patch](https://github.com/otwcode/otwarchive/commit/bfc1a02767a9fc42d6beacd0aeb69eaca2c51639) and I'm getting similar errors involving caching:
```log
  Scenario: Admins should be able to edit but not delete public approved skins                # features/admins/admin_skins.feature:87
    ...snip...
    # the test below failed
    Then the cache of the skin on "public skin" should expire after I save the skin           # features/step_definitions/skin_steps.rb:190
      Cache key 0 matches 0. (Minitest::Assertion)
      ./features/step_definitions/skin_steps.rb:196:in `/^the cache of the skin on "([^\"]*)" should expire after I save the skin$/'
      features/admins/admin_skins.feature:101:in `Then the cache of the skin on "public skin" should expire after I save the skin'
    HTML screenshot: /otwa/tmp/capybara/screenshot_2020-07-10-17-33-02.860.html
```

ariana-paris:
I can't test it right now but it seems to me that we might need the same [Memcached config change from development.rb](https://github.com/otwcode/otwarchive/blob/a31968db6eb369a51293158313f79f2b51a15d17/config/environments/development.rb#L9) to be in the [equivalent place in test.rb](https://github.com/otwcode/otwarchive/blob/a31968db6eb369a51293158313f79f2b51a15d17/config/environments/test.rb#L18) as well?

zz9pzza:
This branch passes rspec and looks good for cucumber so far

https://github.com/otwcode/otwarchive/compare/master...zz9pzza:docker-compose-james?expand=1

Reading the comments on the thread would have got me there faster.

tommilligan:
@ariana-paris thanks, you're entirely right, I just didn't have time to look at it this weekend.

@zz9pzza your branch looks good to me, I'm going to run it locally & on Travis now. I have a couple of comments on your branch, when you're ready to PR it.

zz9pzza:
@tommilligan Would you prefer to do the pull request ? I am happy enought to do ?

I think all the contributors on this will be credited on the release notes however we actually get something merged.

( I would prefer for example to have all the docker related files in a docker directory to avoid clogging up the application root )

tommilligan:
@zz9pzza I'm happy to PR your branch and do some tidying up as you suggest. Will work on it this evening.

redsummernight:
The updated PR (#3863) has been merged. Everyone on this PR will be credited in our [next release notes](https://archiveofourown.org/admin_posts?tag=1):

- Coders: @tlee911, Tom Milligan (@tommilligan), @zz9pzza
- Reviewers: Z (@gbhsrspm), @curtisxk38, @elebow, @ariana-paris

Thank you all for sharing notes and contributing to the project's Docker support!

PR #3664, opened 2019-10-24
AO3-5796 Add prefers-color-scheme media query options for skins

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5796

## Purpose

Adds two new media query options for site skins: prefers-color-scheme: light and prefers-color-scheme: dark. These will let users create skins that automatically apply when their OS color scheme is set a certain way.  

## Testing Instructions

Refer to Jira.

### DIFF
--- a/app/models/skin.rb
+++ b/app/models/skin.rb
@@ -13,7 +13,7 @@ class Skin < ApplicationRecord
                  ]
 
   # any media types that are not a single alphanumeric word have to be specially handled in get_media_for_filename/parse_media_from_filename
-  MEDIA = %w(all screen handheld speech print braille embossed projection tty tv) + ['only screen and (max-width: 42em)'] + ['only screen and (max-width: 62em)']
+  MEDIA = %w(all screen handheld speech print braille embossed projection tty tv) + ['only screen and (max-width: 42em)', 'only screen and (max-width: 62em)', '(prefers-color-scheme: dark)', '(prefers-color-scheme: light)']
   IE_CONDITIONS = %w(IE IE5 IE6 IE7 IE8 IE9 IE8_or_lower)
   ROLES = %w(user override)
   ROLE_NAMES = {"user" => "add on to archive skin", "override" => "replace archive skin entirely"}
@@ -257,6 +257,10 @@ class Skin < ApplicationRecord
         "narrow"
       when m.match(/max-width: 62em/)
         "midsize"
+      when m.match(/prefers-color-scheme: dark/)
+        "dark"
+      when m.match(/prefers-color-scheme: light/)
+        "light"
       else
         m
       end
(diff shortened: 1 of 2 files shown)


app/models/skin.rb
```diff
@@ -13,7 +13,7 @@ class Skin < ApplicationRecord
                  ]
 
   # any media types that are not a single alphanumeric word have to be specially handled in get_media_for_filename/parse_media_from_filename
-  MEDIA = %w(all screen handheld speech print braille embossed projection tty tv) + ['only screen and (max-width: 42em)'] + ['only screen and (max-width: 62em)']
+  MEDIA = %w(all screen handheld speech print braille embossed projection tty tv) + ['only screen and (max-width: 42em)', 'only screen and (max-width: 62em)', '(prefers-color-scheme: dark)', '(prefers-color-scheme: light)']
```
redsummernight: Can you break this into multiple lines?

sarken: How do we want that indented?

```
  MEDIA = %w(all screen handheld speech print braille embossed projection tty tv) +
    ['only screen and (max-width: 42em)', 'only screen and (max-width: 62em)',
    '(prefers-color-scheme: dark)', '(prefers-color-scheme: light)']
```

or something else?

redsummernight:
```ruby
  MEDIA = %w[all screen handheld speech print braille embossed projection tty tv] + [
    'only screen and (max-width: 42em)',
    'only screen and (max-width: 62em)',
    '(prefers-color-scheme: dark)',
    '(prefers-color-scheme: light)'
  ]
```
I checked, Hound wouldn't complain about this one.

spec/models/skin_spec.rb
```diff
@@ -164,13 +164,42 @@
...
+        "allows braille" => "braille",
+        "allows embossed" => "embossed",
+        "allows projection" => "projection",
+        "allows tty" => "tty",
+        "allows tv" => "tv",
+        "allows only screen and (max-width: 42em)" => "only screen and (max-width: 42em)",
+        "allows only screen and (max-width: 62em)" => "only screen and (max-width: 62em)",
+        "allows (prefers-color-scheme: dark)" => "(prefers-color-scheme: dark)",
+        "allows (prefers-color-scheme: light)" => "(prefers-color-scheme: light)"
+      }.each_pair do |description, media_query|
```
redsummernight: I like that you're listing them explicitly instead of using `Skin::MEDIA`, so we don't accidentally break both code and test at once.

You can do this for less duplication:
````rb
      [
        "all",
        "(prefers-color-scheme: light)"
      ].each do |media_query|
        it "allows #{media_query}" do
````

sarken: 🤦‍♂ 

As I was writing this, I was like, "I know I'm doing something inefficiently here, but I don't know what. Hopefully someone can tell me."

redsummernight:
Reminder to delete all the AO3-5758 lines from the squashed commit when merging.

sarken:
I'll clean those up -- I can merge with master now without my dev environment exploding.

sarken:
Et voilà!

PR #3672, opened 2019-10-27
AO3-2240 Gift notifications not sent when works are revealed individually

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-2240

## Purpose

Gift notification emails were only sent out when the collection moderator revealed the entire collection. This attempts to send them when works are revealed individually, although sometimes they might fail due to stale data issues.

## Testing Instructions

Refer to Jira.

## References

Includes https://github.com/otwcode/otwarchive/pull/3501

Discussion of stale data issues starts here https://github.com/otwcode/otwarchive/pull/3501#discussion_r329504805


app/models/collection_item.rb
```diff
@@ -191,12 +191,15 @@ def send_work_invitation
...
-        notify_of_reveal
-      end
-    end
+    unrevealed_collection_ids = item
+      .collection_items
+      .unrevealed
+      .approved_by_user
+      .pluck(:collection_id)
+    # If the item is also in other unrevealed collections, skip notifications.
+    return if (unrevealed_collection_ids - [collection_id]).any?
```
tickinginstant: FYI, the extra complexity here is unnecessary. You don't need to worry about whether the current `CollectionItem` is set wrong, because `after_update` hooks are run inside the same transaction that set that value, which means that you're guaranteed to see the updates from that transaction. You should just be able to copy the check from `lib/collectible.rb`:
https://github.com/otwcode/otwarchive/blob/c364d7f1a76af0d73a260158bd824e1c52962637/lib/collectible.rb#L118

(You can't use the new value of `item.unrevealed` or `work.unrevealed` here because `in_unrevealed_collection` is updated in the `after_save` hook `update_work`, and `after_save` hooks are always run after `after_update` hooks. But the `CollectionItem`s can't run into stale data issues here.)

Also, as I mentioned, this approach can still result in seeing notifications for unrevealed gifts if the transaction is rolled back (hence my suggestion about `after_commit_everywhere`). But I believe that's a pre-existing bug, so I guess it's okay to leave it in.

tickinginstant:
I think [this issue I mentioned](https://github.com/otwcode/otwarchive/pull/3501#discussion_r339395458) may need to be addressed in this PR:
> But if someone has a work that's already visible, with an unrevealed CollectionItem that's not approved_by_user, the recipient will get a notification when that CollectionItem is revealed, even though they (probably) already received a notification when the work first became visible.

With the current code in this PR, the notifications will be sent as long as (a) the CollectionItem's status has just changed from unrevealed to revealed, and (b) the work is visible afterwards. But if the CollectionItem isn't approved by the user, it can't affect the visibility of the work. So if the work is visible afterwards, it was visible before -- no need to send notifications.

(You can't invite someone's work to an unrevealed collection, but you can set the collection to unrevealed after you've invited the work. And then you can set the CollectionItem to unrevealed under the Manage Items tag. Alternatively, I think you could start with an unrevealed CollectionItem approved by both user and collection, then set the user_approval_status to rejected.)

sarken:
Damn, that raises another issue: changing the item's status to unrevealed will also send notifications because `saved_change_to_unrevealed?` will be `true`.

app/models/collection_item.rb
sarken: I was going to update this now that we've added the gem, but it looks like #3818 will take care of using `after_commit_everywhere` for `notify_on_reveal`, correct?

tickinginstant: Yep, that should take care of it.

PR #3673, opened 2019-10-27
AO3-5754 Fix min-length warning overlapping comment buttons

luis-pabon-tf:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5754

## Purpose

Prevent the minimum length warning overlapping buttons on the home page. The reporter's style suggestion does not affect the inbox version of the notice, so it might be a good way to handle the problem.

## Testing Instructions

In the JIRA ticket.

## Credit

bird -  (they)


public/stylesheets/site/2.0/07-interactions.css
```diff
@@ -179,6 +179,7 @@ We only use error messages for LiveValidation. Style spoofs the system error mes
 .LV_validation_message {
   font-weight: 900;
   position: absolute;
+  margin-right: 15em;
```
sarken: I think we also need to adjust the right margin on `.LV_invalid` below, or else this will get overridden when displaying an error message.

sarken (changes requested):

public/stylesheets/site/2.0/07-interactions.css
TurkeyFried: Haven't forgotten about this, just haven't had a chance to do this + test visually. Will do though.

sarken: No worries! Thanks for letting us know.

TurkeyFried:
Very, very late, but I applied the suggested tweaks and it looks better now.

public/stylesheets/site/2.0/07-interactions.css
TurkeyFried: I've updated the css so all the changes are on `.LV_invalid` only.

sarken (approved):
Looks good to me now, thank you!

PR #3674, opened 2019-10-28
AO3-4539 Remove spam verification for logged-in users on the Support, Abuse forms

TurkeyFried:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4539

## Purpose

Before running spam verification, checks if the user is logged in, and the email they submit matches the one on their account.

## Testing Instructions

- Create User with any information and submit both types of forms with generic data without editing the pre-filled fields. Both forms should submit successfully.
- With the same user, attempt to submit the same forms, but this time using the email address "akismet-guaranteed-spam@example.com". This should trigger a spam warning on submission.

- Repeat last 2 steps without logging in. Fill in the empty fields with standard, valid-looking data. Both steps should yield the same results as a logged in user.

- Create another User, with "akismet-guaranteed-spam@example.com" as its email address. Repeat the first step, but this time the forms should accept the "spam" email address.

## Credit

bird

they


TurkeyFried:
> Yay, thank you for this! It will make a lot of people quite happy. However, right now, it appears there's a test failure related to these changes. It's in the specs: spec/models/abuse_report_spec.rb:258

Thanks for the heads up! I'll get that sorted out as soon as possible.

TurkeyFried:
The issue has been solved. I haven't written tests for the new behavior yet, but will try to get that done soon.

app/models/feedback.rb
```diff
@@ -14,7 +14,15 @@ class Feedback < ApplicationRecord
...
   def check_for_spam
-    errors.add(:base, ts("^This comment looks like spam to our system, sorry! Please try again, or create an account to comment.")) unless check_for_spam?
+    errors.add(:base, ts("This report looks like spam to our system!")) unless check_for_spam?
+  end
+
+  def check_for_spam?
+    self.approved = logged_with_matching_email? || !Akismetor.spam?(akismet_attributes)
+  end
+
+  def logged_with_matching_email?
```
redsummernight: Should this be `logged_in_with_matching_email`?

app/models/feedback.rb
```diff
@@ -14,7 +14,15 @@ class Feedback < ApplicationRecord
 
   validate :check_for_spam
   def check_for_spam
-    errors.add(:base, ts("^This comment looks like spam to our system, sorry! Please try again, or create an account to comment.")) unless check_for_spam?
+    errors.add(:base, ts("This report looks like spam to our system!")) unless check_for_spam?
+  end
+
+  def check_for_spam?
```
redsummernight: Can you combine `check_for_spam?` and `check_for_spam`? We can avoid having 2 very similarly named methods.

app/models/feedback.rb
TurkeyFried: true, renaming in the next commit

app/models/feedback.rb
TurkeyFried: making an attempt. I'm still fairly new to Ruby so fingers crossed my solution attempt is good!

sarken:
The changes look good! I won't switch the label to "Reviewed: Ready To Merge" yet because you mentioned you were working on some tests.

TurkeyFried:
Pushed the new tests. Will check the Hound violations soon.

redsummernight:
Fixed conflicts with #3758 which was recently merged.

PR #3676, opened 2019-10-31
AO3-5693 Supply values for whitelisted boolean attributes

sarken:
## Issue

## Purpose

If you use the boolean attributes `loop`, `muted`, or `default` in your media, the attribute value should be the same as the attribute name, e.g. `loop="loop"`, to be consistent with the way we handle our automatically-added boolean attributes, e.g. `controls`.

## Testing Instructions

Use the aforementioned attributes when embedding audio or video, but do not include a correct value (or any value) -- for example, you might do `<video loop="potato" muted>`. After posting, edit your work to check that the HTML is updated to include a value that matches the attribute name (so for our example, the desired outcome would be `<video loop="loop" muted="muted">`).

### DIFF
--- a/lib/otw_sanitize/media_sanitizer.rb
+++ b/lib/otw_sanitize/media_sanitizer.rb
@@ -72,6 +72,7 @@ module OTWSanitize
 
       config = Sanitize::Config.merge(Sanitize::Config::ARCHIVE, WHITELIST_CONFIG)
       Sanitize.clean_node!(node, config)
+      tidy_boolean_attributes(node)
       { node_whitelist: [node] }
     end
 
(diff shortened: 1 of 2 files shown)


redsummernight:
Looks good, but let's wait and see if the Sanitize update ([AO3-5801](https://otwarchive.atlassian.net/browse/AO3-5801)) changes the current behavior.

sarken:
👍 I had similar thoughts. It would be nice if this were unnecessary!

sarken:
This isn't fixed by updating Sanitize. I updated it locally and ran this test:

```
      it "allow valueless boolean attributes" do
        html = "
          <video muted>
          </video>"
        content = Sanitize.fragment(html, config)
        expect(content).not_to match('muted=""')
      end
```

It failed:

```
1) OTWSanitize::MediaSanitizer.transformer when sanitizing allow valueless boolean attributes
     Failure/Error: expect(content).not_to match('muted=""')
     
       expected "\n          <video muted=\"\" controls=\"controls\" playsinline=\"playsinline\" crossorigin=\"anonymous\" preload=\"metadata\">\n          </video>" not to match "muted=\"\""
       Diff:
       @@ -1,2 +1,4 @@
       -muted=""
       +
       +          <video muted="" controls="controls" playsinline="playsinline" crossorigin="anonymous" preload="metadata">
       +          </video>
```

PR #3678, opened 2019-11-03
AO3-5801 Update Sanitize gem to remove unwanted closing tags in HTML5 media

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5801

## Purpose

Updates the Sanitize gem (and its dependency, Nokogumbo) so we stop adding `</track>` and `</source>` tags in video and audio embeds. It also adds tests to demonstrate that it does in fact do those things.

There is, however, one catch: tests indicate it doesn't prevent unwanted closing tags when there is an HTML fallback included in the embed. I've left the test failing so everyone can see it and we can decide what to do; we can switch it to pending or whatever following review. If anyone has a fix for that issue, please feel free to push to my branch or create another PR.

## Testing Instructions

Refer to Jira.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -57,11 +57,8 @@ gem 'htmlentities'
 gem 'whenever', '~>0.6.2', require: false
 gem 'nokogiri', '>= 1.8.5'
 gem 'mechanize'
-gem 'sanitize', '>= 4.6.5'
-# Until there is a working solution to
-# https://otwarchive.atlassian.net/browse/AO3-4957
-# https://github.com/rubys/nokogumbo/issues/50
-gem 'nokogumbo', '1.4.9'
+gem 'sanitize', '>= 5.0.1'
+gem 'nokogumbo', '~> 2.0'
 gem 'rest-client', '~> 1.8.0', require: 'rest_client'
 gem 'resque', '>=1.14.0'
 gem 'resque_mailer'
--- a/spec/miscellaneous/lib/html_cleaner_spec.rb
+++ b/spec/miscellaneous/lib/html_cleaner_spec.rb
@@ -219,7 +219,7 @@ describe HtmlCleaner do
         context "class has one value" do
           it "keeps values containing only letters, numbers, and hyphens" do
             result = sanitize_value(field, '<p class="f-5">foobar</p>')
-            doc = Nokogiri::HTML.fragment(result)
+            doc = Nokogiri::HTML5.fragment(result)
             expect(doc.xpath("./p[@class='f-5']").children.to_s.strip).to eq("foobar")
           end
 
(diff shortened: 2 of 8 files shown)


Gemfile
```diff
@@ -57,11 +57,8 @@ gem 'htmlentities'
 gem 'whenever', '~>0.6.2', require: false
 gem 'nokogiri', '>= 1.8.5'
 gem 'mechanize'
-gem 'sanitize', '>= 4.6.5'
-# Until there is a working solution to
```
sarken: AO3-4957 was deployed, so I've removed these comments.

spec/miscellaneous/lib/html_cleaner_spec.rb
```diff
@@ -584,7 +584,7 @@
 
     it "leaves audio tags alone" do
       original = "<audio controls=\"controls\" crossorigin=\"anonymous\" preload=\"metadata\">
-  <source src=\"http://example.com/podfic.mp3\" type=\"audio/mpeg\"></source>
+  <source src=\"http://example.com/podfic.mp3\" type=\"audio/mpeg\">
```
sarken: This is where the test failure is. A `</source>` is getting added right before the `</audio>`.

redsummernight: This is tricky. Sanitize 5 (and our MediaSanitizer, by extension) does not add closed tags for `<source>`, because it uses Nokogumbo 2.0's HTML5 serializer (`Nokogiri::HTML5.fragment`, rubys/nokogumbo#14). If you move this spec to `media_sanitizer_spec`, it will pass.

Our HTMLCleaner uses Nokogiri's old HTML serializer (`Nokogiri::HTML5.fragment`) which does add closed tags for `<source>`.

https://github.com/otwcode/otwarchive/blob/c364d7f1a76af0d73a260158bd824e1c52962637/lib/html_cleaner.rb#L172-L174

Other things using the old HTML serializer: the word counter, the importing story parser. That's a lot of stuff to update.

sarken: Ugh, okay, here we go...

> If you move this spec to `media_sanitizer_spec`, it will pass. 

Which position reflects the _actual_ behavior we'd get? 

"Sarken, you should be able to answer that by trying it on your development environment," you say. Which is a good point, but when I tried the embeds on my dev environment, it continued to insert the `</source>` and `</track>` tags. Because @elzj had mentioned this resolved the issue, I assumed my dev environment was just being cranky and not updating, so I shrugged and wrote automated tests to prove it was fixed instead. 

Now that you've said this, though, I have the sinking feeling the behavior I was seeing is correct and indicates updating Sanitize does not by itself fix the unwanted tags.

redsummernight: I think `html_cleaner_spec` is closer to the actual behavior. We don't use MediaSanitizer on its own, but as part of HtmlCleaner.

sarken: Yeeeeah, I figured. idk about you, but even though I don't love releasing this with invalid HTML, I'm even less keen on shoving all of those changes into this release, especially considering this release has taken -- let me check my watch -- almost a month, and we've been struggling to find folks to test the changes to the sanitizer.

sarken:
We're going to attempt a hack-y fix for now and handle this update later.

redsummernight:
Said hack-y fix: #3679.

PR #3679, opened 2019-11-04
AO3-5693 Remove invalid closing tags on track and source elements

redsummernight:
## Issue

## Purpose

This is a hack until we switch to an HTML5 serializer.

## Testing Instructions

See issue.


lib/html_cleaner.rb
```diff
@@ -172,6 +172,11 @@ def sanitize_value(field, value)
       doc = Nokogiri::HTML::Document.new
       doc.encoding = "UTF-8"
       unfrozen_value = doc.fragment(unfrozen_value).to_xhtml
+
+      # Hack! the herald angels sing
```
sarken: 😂

PR #3680, opened 2019-11-05
AO3-5806 Reset and reorganize locale files and reconfigure Phrase

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5806

## Purpose

This looks involved af, but it's mostly a lot of deleting and moving.

Commit-by-commit:
* Deletes all of the locale files except for en.yml to wipe out all of Translation's previous work.
* Updates application.rb so it can handle locale files being placed in subfolders of /config/locales.
* Grabs the locale files for Translation's desired locales from rails-i18n and places them in config/locales/defaults, because we'd like to keep the defaults separate from our work going forward.
    * rails-i18n doesn't have a pt-PT.yml for Portugal's Portuguese, but Translation was chill with using pt.yml and renaming it pt-PT.yml.
* Takes the translations for our mailers out of config/locales/en.yml and moves them into config/locales/mailer/en.yml.
* Takes the translations for our views out of config/locales/en.yml and moves them into config/locales/views/en.yml.
* Just deletes all of the translations that can also be found in rails-i18n's en.yml file from our /locales/en.yml file, and then adds rails-i18n's en.yml to config/locales/defaults. Basically, we hit the reset button for English, too. There were a few differences that I will note here, though:
    * Our en.yml file had `confirmation: doesn't match confirmation`, but the new one from rails-i18n has `confirmation: doesn't match %{attribute}`. This actually improves the troublesome error message in [AO3-5113](https://otwarchive.atlassian.net/browse/AO3-5113). 🎉  I have updated the issue with the information.
    * Our en.yml file had... something going on in the `errors` section with `messages: &errors_messages` ([line 33](https://github.com/otwcode/otwarchive/blob/b8d40d3d70bfdd7fadccc6a40f59ca66079ce4f1/config/locales/en.yml#L33)) and `template: &errors_template` ([line 61](https://github.com/otwcode/otwarchive/blob/b8d40d3d70bfdd7fadccc6a40f59ca66079ce4f1/config/locales/en.yml#L61)). It was... [referenced lower down](https://github.com/otwcode/otwarchive/blob/b8d40d3d70bfdd7fadccc6a40f59ca66079ce4f1/config/locales/en.yml#L67-L72), in the ActiveRecord errors section. I have no idea what it's for, and it never existed in rails-i18n's version of the file. Google results tend to point me to it being used back in 2011, so my assumption is it's not relevant in the current version of Rails. Anyway, it's gone now and the tests are passing, but here's a note saying that's a thing that happened.
    * Our en.yml was also missing a bunch of translations rails-i18n added, so yay for that. I'm not documenting them all. :P 
* Takes the translations for our validators out of config/locales/en.yml and moves them into config/locales/validators/en.yml.
* Takes the translations for our models out of config/locales/en.yml and moves them into config/locales/models/en.yml, thereby emptying out config/locales/models/en.yml and allowing me to delete it. 
* Moves the Devise-related translation file out of config/locales and into config/locales/devise. Renames it en.yml. Yay, everything is nestled nice and comfy in subdirectories now.
* Resolves two test failures related to aforementioned `confirmation` message change.
* Updates the .phraseapp.yml config file to use the new locale folder structure and new project ID. I removed the access token from it before committing, since presumably that is meant to be private, but we'll just have to see if it works.
* Fixed typo in Phrase config file.
* Updated Phrase config file to [keep files separate](https://help.phrase.com/en/articles/2136270).
* Update list of `available_locales` in config/application.rb to match the list of locales the Translators will be working with.
* Style fixes for above change.
* Add a README saying what version the locale files are and how to start over in case of 🔥.

## Testing Instructions

Refer to Jira.


.phraseapp.yml
```diff
@@ -1,13 +1,12 @@
...
   push:
     sources:
-    - file: ./config/locales/<locale_name>.yml
+    - file: ./config/locales/**/<locale_name>.yml
       params:
         file_format: yml
   pull:
     targets:
-    - file: ./config/locales/<locale_name>.yml
+    - file: ./confit/locales/**/<locale_name>.yml
```
bingeling: typo, should be `config`

sarken:
Funny you should mention that! I was wondering if I should update that list to contain the new list from Translation.

bingeling:
Ha! I was in the process of ammending my comment to add that it was missing locales, when your reply popped up. My vote would be to change the list, since presumably if Translations requested them, they are confident they can fully translate these locales.

sarken:
I tried, but Hound had complaints about pt-BR without quotation marks, so I put them back.

sarken:
https://github.com/svenfuchs/rails-i18n/tree/master/rails/locale

It's version 6.0, but since we're using the locale files and not the gem, it seemed prudent to grab the latest versions available. (The English file, for example, apparently has a fix for a pluralization issue with the time unit "second" that apparently resulted from a copy & paste error 10 years ago, sob. Of course, we've been operating with an en.yml that didn't even have that information anyway...) 

I can grab the 5.0 versions instead if we're worried about things that didn't turn up in tests.

redsummernight:
Latest version sounds fine. I think we should add a `config/locales/README.md` file saying where to get the files (https://github.com/svenfuchs/rails-i18n#manual-installation) and what version.

config/locales/views/en.yml
```diff
@@ -0,0 +1,28 @@
+en:
```
bingeling: I assume strings we currently have in the controllers will also go in here? Since presumably they will end up in one view or another?

bingeling:
Several of these locales are missing some ActiveRecord error message entries, they are missing in rails-i18n already, so nothing to do about it. But something to keep an eye on when testing.

config/locales/views/en.yml
sarken: We could do that or put them in their own controllers folder if there are a lot of them.

sarken:
Thank you both for reviewing!

PR #3682, opened 2019-11-07
AO3-5809 Add paste to RTE context menu and toolbar, and and hr to the context menu

sarken:
##  Issue

https://otwarchive.atlassian.net/browse/AO3-5809

## Purpose

Adds paste and hr options to the RTE's context menu.

## Testing Instructions

On desktop, go to the RTE. Right-click. Confirm options are there and do relatively reasonable things.

## References

https://www.tiny.cloud/docs/configure/editor-appearance/#contextmenu


public/javascripts/mce_editor.js
```diff
@@ -67,25 +67,28 @@ tinyMCE.init({
   // Override the list of targets provided in the link plugin. We do not allow the target attribute, so we want an empty list.
   target_list: [
     { title: 'None', value: '' }  
-  ]
+  ],
+
+  // Options to display in the context (right-click) menu.
+  contextmenu: 'paste | link | hr'
```
redsummernight: The default configuration is `link image imagetools table spellchecker`. We should keep `image` (it only shows up when you right click on an image).

sarken: Ah, I thought it was lying because I wasn't seeing that. I will add it!

PR #3685, opened 2019-11-09
AO3-5772 Exclude third party or minified JS files from checks

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5772

## Purpose

- Stop checking (Hound, Code Climate) JS files that are third-party (e.g. TinyMCE, dewplayer) or minified. There are some libraries I'm not excluding because we may or may not have patched them (e.g. jquery.tokeninput.js, which we definitely patched).
- Use JSHint default configuration, instead of Hound's stricter legacy configuration. Full context in the relevant commit message.

## Testing Instructions

As long as Hound and Code Climate succeed on this PR, we're good.

Locally, I tried:

```sh
# npm install -g jshint@2.9.5
jshint . # on the whole repo
```


redsummernight:
> codeclimate — 306 issues to fix

Those are all issues on files not excluded. It's working! :) :(

sarken:
Merging so we don't have to hear Hound's opinions if we need another RTE fix

PR #3686, opened 2019-11-10
AO3-5802 Rewrite assignment generation code.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5802

## Purpose

This PR has two purposes:

1. The rewritten assignment generation code relies on the database much less than the pre-existing assignment generation code does. It does so by loading all of the potential matches in one go, and then not touching the database until it's already calculated all of the assignments.

   This means that there's no danger of reading out-of-date values for `assigned_as_request` / `assigned_as_offer` -- the kind of issue that likely produced the duplicate assignments seen on the test servers. It also helps with Resque errors on lines like this one: https://github.com/otwcode/otwarchive/blob/47becf0041a7750a81951617111da409895cf098/app/models/challenge_assignment.rb#L407-L408

   If the new value of `assigned_as_request` is available, but the corresponding assignment is not, that can result in a `NoMethodError` from calling `offer_signup` on a `nil` assignment. By keeping all of the information about assignments in memory, the new code avoids these issues.

2. The rewritten assignment generation code also uses [the Hopcroft-Karp matching algorithm](https://en.wikipedia.org/wiki/Hopcroft-Karp_algorithm) to try to maximize the number of complete assignments/minimize the number of assignments with a missing recipient or giver.

## Testing Instructions

See JIRA.


app/models/assignment_generator.rb
```diff
@@ -0,0 +1,308 @@
+# A challenge assignment generator that uses Hopcroft-Karp to try to get a
+# maximum matching (thus maximizing the number of complete assignments, and
+# minimizing the number of assignments with no recipient or no giver). Sorts
+# the potential matches (using the default sort, which relies on the fields
```
redsummernight: Can you add a comment in PotentialMatch's `<=>` that the assignment generator depends on its logic?

app/models/assignment_generator.rb
```diff
@@ -0,0 +1,308 @@
...
+    @recipients_by_priority = @potential_givers.keys.sort_by do |key|
+      [@potential_givers[key].size, rand]
+    end
+  end
+
+  # Set up the hashtables we use to store our current assignments.
+  def set_up_assignments
+    # This table will contain a mapping from each request to the offer
+    # that they've been (temporarily) assigned to.
+    @assignment_as_recipient = {}
```
redsummernight: I feel like "as" makes it hard to parse if recipient is the key or the value. `assignment_by_recipient` maybe? I could be convinced otherwise.

Ditto `assignment_by_giver`.

app/models/assignment_generator.rb
```diff
@@ -0,0 +1,308 @@
...
+
+        ChallengeAssignment.create(collection_id: @collection.id,
+                                   request_signup_id: recipient,
+                                   offer_signup_id: giver)
+      end
+    end
+
+    # Update the challenge_signups to indicate whether they've been assigned.
+    #
+    # TODO: These fields are no longer used for anything. Can the columns be
```
redsummernight: Are they only used as part of the old assignment generation? Let's drop them in a following release.

app/models/assignment_generator.rb
```diff
@@ -0,0 +1,308 @@
...
+    # dropped from the ChallengeSignup table?
+    ChallengeSignup.where(id: @assignment_as_recipient.keys)
+      .update_all(assigned_as_request: true)
+    ChallengeSignup.where(id: @assignment_as_recipient.values)
+      .update_all(assigned_as_offer: true)
+  end
+
+  # Add blank assignments for any users that couldn't be assigned.
+  def save_placeholder_assignments
```
redsummernight: Why are placeholder assignments saved in a separate transaction? (I don't have a reason for or against, though.)

app/models/assignment_generator.rb
```diff
@@ -0,0 +1,308 @@
...
+  end
+
+  # Do breadth-first search in the augmenting path graph. This is the equivalent
+  # of the BFS() function in Wikipedia's pseudo-code for the Hopcroft-Karp
+  # algorithm.
+  #
+  # Returns nil if there are no augmenting paths.
+  #
+  # Otherwise, returns a map from offers to layer number/distance.
+  def augmenting_path_layers
```
redsummernight: I think this is the first time I'm reviewing BFS and DFS in the wild, instead of in a coding interview. :D

script/gift_exchange/load_json_challenge.rb
```diff
@@ -157,5 +158,5 @@ def make_prompt(challenge_signup, info, prompt_type)
     make_prompt(challenge_signup, info, 'offer')
   end
 
-  challenge_signup.save!
+  challenge_signup.save!(validate: false)
```
redsummernight: This is a test script but I'm curious why validation is skipped here.

app/models/assignment_generator.rb
tickinginstant: Hmm. I'm not sure if I like "by" here. I feel like if `giver` was assigned to `recip`, I'd expect `assignment_by_recipient[recip]` and `assignment_by_giver[giver]` to return the same object (with information about both the recipient and the giver).

What about `assignment_for_recipient` and `assignment_for_giver`?

app/models/assignment_generator.rb
tickinginstant: No particular reason. I think I originally split them into different functions to keep the code more organized, and as a side effect split the transaction as well. But I can move the transaction block into `generate` to make it a single transaction.

script/gift_exchange/load_json_challenge.rb
tickinginstant: Speed. I created two gift exchanges with 2000 participants on my local machine, and got rather impatient waiting for the script to finish. Almost all of the changes I made in the `script` directory were intended to make that process faster.

I can change it back, if you'd like. I think skipping validation only saves ~30 seconds, so I don't have strong feelings either way.

app/models/assignment_generator.rb
redsummernight: `_for_` sounds good to me.

script/gift_exchange/load_json_challenge.rb
redsummernight: Let's keep it this way. Can you add a comment that we skip validation for speed?

app/models/assignment_generator.rb
redsummernight: You have it in one transaction now, still looks fine to me. How long did the single transaction take in your 2k participants test?

tickinginstant: On my machine, the transaction takes ~6 seconds (out of ~9 seconds total for regenerating assignments). I'm not sure how much time that would translate into in production, but hopefully it'll be fast enough to prevent most locking issues.

tickinginstant:
Closing to make room for more PRs.

PR #3687, opened 2019-11-11
AO3-5815 Add shared examples for mailer specs

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5815

## Purpose

There was a ton of repetition in user_mailer_spec.rb because we check the sender and look for exposed HTML or missing translations on each email... and the amount of repetition will only increase as we translate more emails. So I went and added shared examples to reduce the repetition. 

And because I had the wind beneath my wings, I didn't stop there. I've traded in the before blocks and instance variables for `let` and rearranged the tests so they're grouped by method (and, more importantly, _named_ by method, making it clear instantly clear what the devil they're testing).

I changed the work titles used in some tests because, as noted in the commit message for that change

>The work titles in the test were previously set up in such a way that even if the work with the first title wasn't being listed, the tests would still pass (because the second title contained the full text of the first). Further, the text of the first title was also in the email body.

(I considered changing the work factory to use Faker for work titles and chapter content instead of the fixed strings that are in there, but that would've involved updating a couple of specs that look for that text, as opposed to `work.title` or whatever, so I decided that would be better in another issue if we thought it might be worthwhile.)

I also updated the tests to use the helpers from [email-spec](https://github.com/email-spec/email-spec) that we were using elsewhere.

I applied the same changes to the specs for the comment and collection mailers too because they were tiny.

## Testing Instructions

None, automated.


spec/support/shared_examples/mailer_shared_examples.rb
```diff
@@ -0,0 +1,28 @@
+shared_examples_for "multipart email" do
```
redsummernight: While you're here, "a multipart email"?

spec/support/shared_examples/mailer_shared_examples.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+end
+
+shared_examples_for "an email with a valid sender" do
+  it "is delivered from Archive of Our Own <do-not-reply@example.org>" do
+    expect(email).to deliver_from("Archive of Our Own <#{ArchiveConfig.RETURN_ADDRESS}>")
+  end
+end
+
+shared_examples_for "a well formed HTML email" do
```
redsummernight: Should we combine this with "multipart email"? If an email has HTML, it should be well-formed as well.

spec/spec_helper.rb
```diff
@@ -168,10 +168,3 @@ def delete_test_indices
 def get_message_part (mail, content_type)
```
redsummernight: This can be removed entirely, in favor of `email.html_part` or `email.text_part`.

spec/support/shared_examples/mailer_shared_examples.rb
sarken: GLADLY.

spec/support/shared_examples/mailer_shared_examples.rb
sarken: I'm game. I left it separate in case we ever wanted to test that an email was multipart _and_ capable of handing a < or > sign in the text, but that seems somewhat unlikely... and if it ever becomes likely, well, we can just not use the "a multipart email" shared example in that particular test.

PR #3695, opened 2019-11-26
AO3-5825 Cache work and bookmark counts on dashboard sidebar

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5825

## Purpose

This cuts down on the number of Elasticsearch count queries.

Unlike [cached search results in WorksController](https://github.com/otwcode/otwarchive/blob/cccba97ecd7b4ab69ff22a00f3df9cddf7f22860/app/controllers/works_controller.rb#L103), cached keys for the counts are not timestamped and cannot be invalidated by e.g. [WorksOwner#update_works_index_timestamp](https://github.com/otwcode/otwarchive/blob/cccba97ecd7b4ab69ff22a00f3df9cddf7f22860/lib/works_owner.rb#L37-L43). The counts will lag
behind the actual search results.

## Testing Instructions

See issue.

### DIFF
--- a/app/models/search/bookmark_search_form.rb
+++ b/app/models/search/bookmark_search_form.rb
@@ -48,21 +48,6 @@ class BookmarkSearchForm
 
   attr_accessor :options
 
-  def self.count_for_user(user)
-    BookmarkQuery.new(
-      user_ids: [user.id],
-      show_private: User.current_user.is_a?(Admin) || user == User.current_user
-    ).count
-  end
-
-  def self.count_for_pseuds(pseuds)
-    BookmarkQuery.new(
-      pseud_ids: pseuds.map(&:id),
-      show_private: User.current_user.is_a?(Admin) ||
-                    pseuds.map(&:user).uniq == [User.current_user]
-    ).count
-  end
-
   ATTRIBUTES.each do |filterable|
     define_method(filterable) { options[filterable] }
   end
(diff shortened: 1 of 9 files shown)


features/users/user_dashboard.feature
```diff
@@ -65,7 +66,10 @@ Feature: User dashboard
     And I go to meatloaf's user page
   Then I should see "Newest Work"
     And I should not see "Oldest Work"
-    And I should see "Works (6)" within "#dashboard"
+    And I should see "Works (5)" within "#dashboard"
+  When I wait 11 seconds
```
redsummernight: Waiting can be imprecise... if I set the expiration time too low, the tests might actually take longer, and the stale check (`see "Works (5)"`) will fail.

Ideally I'd like to use Timecop to control time precisely and instantly (`And it is currently 25 minutes from now`), but Timecop doesn't affect memcached because memcached is a separate process, but we can switch to [an in-memory, same-process cache store](https://guides.rubyonrails.org/caching_with_rails.html#activesupport-cache-memorystore) which can time-travel with Timecop, except we cannot switch to the in-memory cache store because it uses [Ruby's built-in serialization that cannot handle IO objects](https://ruby-doc.org/core-2.6.5/Marshal.html#method-c-dump).

And we do cache IO objects: [paperclip attachments](https://github.com/thoughtbot/paperclip/issues/1114), on the [default Skin object](https://github.com/otwcode/otwarchive/blob/cccba97ecd7b4ab69ff22a00f3df9cddf7f22860/app/models/skin.rb#L504-L508).

So `Kernel::sleep` for now.

app/models/search/work_search_form.rb
```diff
@@ -247,4 +221,28 @@ def default_sort_direction
...
+
+  def self.count_for_pseud(pseud)
+    Rails.cache.fetch(count_cache_key(pseud), count_cache_options) do
+      WorkQuery.new(pseud_ids: [pseud.id]).count
+    end
+  end
+
+  def self.count_cache_key(owner)
+    status = User.current_user ? 'logged_in' : 'logged_out'
+    "work_count_#{owner.class.name.underscore}_#{owner.id}_#{status}"
```
redsummernight: To have the old cache invalidation behavior we just use `#{owner.works_index_cache_key}` instead of `#{owner.class.name.underscore}_#{owner.id}`.

app/models/search/bookmark_search_form.rb
```diff
@@ -169,6 +154,36 @@ def sort_direction(sort_column)
...
+  end
+
+  def self.count_cache_key(owner, show_private)
+    status = User.current_user ? "logged_in" : "logged_out"
+    status << "_private" if show_private
+    "bookmark_count_#{owner.class.name.underscore}_#{owner.id}_#{status}"
+  end
+
+  def self.count_cache_options
+    { expires_in: ArchiveConfig.SECONDS_UNTIL_DASHBOARD_COUNTS_EXPIRE.seconds }
```
sarken: Should we have a `race_condition_ttl` in our options?

app/models/search/work_search_form.rb
sarken: 1. Should we add a comment saying that's an option available to us, in case we decide to revisit this?
2. Just checking my understanding -- `owner.class.name.underscore` the `underscore` is just a precaution and right now it could only be `user` or `pseud`?

redsummernight: 1. I added a comment.
2. For `owner.class.name.underscore`, I just followed `lib/works_owner.rb`'s example. For this method, it will only be `user` or `pseud` (a "work owner" could be `tag` or `collection` but we don't pass them in here).

PR #3698, opened 2019-11-29
AO3-3306 AO3-5831 AO3-5833 Improve autocomplete compatibility with screen readers

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-3306
https://otwarchive.atlassian.net/browse/AO3-5831
https://otwarchive.atlassian.net/browse/AO3-5833

## Purpose

* Uses HTML5 `data-*` attributes for the autocomplete's custom attributes. This means we don't get as many validation errors on our HTML now. 🎉 
* Updates the JavaScript responsible for the autocomplete so screen readers can interact with the suggestions. I've based this on [the first example on the W3's combobox with listbox examples page](https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html). 
  * I've moved `role="combobox"` and `aria-expanded` onto the parent of the `<input>`. Having them directly _on_ the `<input>` meant that iOS VoiceOver didn't recognize it as a box you can type in.
    * I've also made it so `aria-expanded` is set appropriately depending on whether or not the dropdown containing suggestions or tips is present.
  * I've added the recommended `aria-owns` and `aria-haspopup` attributes and values to the combobox, the `aria-controls` attribute to the text field, and `aria-labelledby` to the listbox. tbh, I didn't see that they made any difference, but they're there now.
  * When you use the arrow keys to navigate the options, an `aria-activedescendant` attribute referencing the selected option will be added to the `<input>`. The selected option will also get the attribute `aria-selected="true"`.
    * Note: if you're using a mouse and hover in certain places, it's possible for the `<input>` to acquire the `aria-selected` and everything else added by the `select_dropdown_item` function. I suspect this has to do with the somewhat peculiar HTML structure that sticks the text field in a `<li>`, but I'm not sure. You can see it on production, too -- a tiny bit of visible red to the left side of the field if you wave the mouse around just so. I'm just... not going to touch that right now.
* Removes the `title` attributes on the work form tag section's autocomplete fields. (AO3-5831)
* Makes sure that if the original `<input>` has an `aria-describedby` attribute, it is placed on the new `<input>` as well.
* Increases the timeout for searching when entering text. This makes autocomplete work better on iPhones, which also makes this easier to test on such devices. (AO3-5833)

## Testing Instructions

We'll want to see how autocomplete fields work with a variety of screen readers.


app/helpers/application_helper.rb
```diff
@@ -298,11 +298,12 @@ def expand_contract_shuffle(list_id, shuffle=true)
   def autocomplete_options(method, options={})
     {
       class: "autocomplete",
-      autocomplete_method: (method.is_a?(Array) ? method.to_json : "/autocomplete/#{method}"),
-      autocomplete_hint_text: ts("Start typing for suggestions!"),
-      autocomplete_no_results_text: ts("(No suggestions found)"),
-      autocomplete_min_chars: 1,
-      autocomplete_searching_text: ts("Searching...")
+      # We don't use data: {} to ensure values will be correctly overridden
```
redsummernight: I think we can use `data: {}`, as long as we also use it in all the overriding places, and we switch `merge` later in this method to [`deep_merge`](https://api.rubyonrails.org/classes/Hash.html#method-i-deep_merge) which merges recursively.

public/javascripts/jquery.tokeninput.js
```diff
@@ -788,7 +802,11 @@ $.TokenList = function (input, url_or_data, settings) {
                 .hide();
 
             $.each(results, function(index, value) {
-                var this_li = $("<li role=\"option\">" + highlight_term(escapeHTML(value.name), query) + "</li>") // was role=\"menuitem\"
+                var this_li = $("<li>" + highlight_term(escapeHTML(value.name), query) + "</li>") // was role=\"menuitem\"
+                                  .attr({
+                                    "id": hidden_input_id + "_autocomplete_suggestion_" + index,
+                                    "role": "option"
```
redsummernight: We should comment `was "menuitem"` here and remove the comment on `this_li`.

app/helpers/application_helper.rb
sarken: I will give `deep_merge` a shot! The first two things alone weren't enough.

sarken:
This started off so neat and tidy and then went to hell on me. 😭

sarken:
`aria-describedby` can reference more than one `id` , so I'm going to tweak it to add the autocomplete hint's `id` to any existing `aria-describedby` values

sarken:
Fun fact: that `aria-describedby` is not currently working on production. *cracks knuckles*

sarken:
I reworked this a bit, which eliminated the need for adding a new `aria-describedby` for the instructions -- now assistive teach should tell the user quite clearly that they are in a text field and should type. However, I've made sure we're keeping any `aria-describedby` attributes added in the view files.

public/javascripts/jquery.tokeninput.js
```diff
@@ -776,7 +805,11 @@ $.TokenList = function (input, url_or_data, settings) {
     function populate_dropdown (query, results) {
         if(results && results.length) {
             dropdown.empty();
-            var dropdown_ul = $("<ul role=\"listbox\" aria-activedescendant=\"ui-active-menuitem\">")
+            var dropdown_ul = $("<ul>")
+                .attr({
+                  "aria-labelledby": hidden_input_id + '_autocomplete_label',
```
redsummernight: Can we use `hidden_input_label.attr("id")`?

public/javascripts/jquery.tokeninput.js
```diff
@@ -729,11 +753,14 @@ $.TokenList = function (input, url_or_data, settings) {
...
             $(selected_dropdown_item).removeClass(settings.classes.selectedToken);
         }
+        input_token.attr("aria-expanded", "false");
+        input_box.removeAttr("aria-activedescendant");
         selected_dropdown_item = null;
     }
 
     function show_dropdown() {
         dropdown
+            .attr("id", hidden_input_id + "_autocomplete_dropdown")
```
redsummernight: We should use a var at the start of TokenList instead of doing ` + "_autocomplete_dropdown"` in 3 places.

public/javascripts/jquery.tokeninput.js
```diff
@@ -207,19 +207,35 @@ $.TokenList = function (input, url_or_data, settings) {
     var hidden_input_id = $(input)
                               .attr('id');
 
-    // Keep a reference to the label whose for attribute that matches the original input box's id        
+    // Keep a reference to the label whose for attribute matches the original input box's id
```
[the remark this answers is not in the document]
redsummernight: Ignoring the length warning, but can you remove the trailing spaces of this line?

sarken:
Resolved merge conflicts.

PR #3708, opened 2019-12-09
AO3-5838 Expand test coverage of bookmarks controller index action

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5838

## Purpose

Makes sure the bookmarks controller index action is covered a bit better by automated tests, particularly when the experimental caching setting is enabled. It also allows us to actually enable that setting in tests 😭 

## Testing Instructions

None, automated.

### DIFF
--- a/spec/controllers/bookmarks_controller_spec.rb
+++ b/spec/controllers/bookmarks_controller_spec.rb
@@ -43,7 +43,7 @@ describe BookmarksController do
     end
   end
 
-  describe "index", bookmark_search: true do
+  describe "index" do
     let!(:external_work_bookmark) { create(:external_work_bookmark) }
     let!(:series_bookmark) { create(:series_bookmark) }
     let!(:work_bookmark) { create(:bookmark) }
@@ -152,7 +152,7 @@ describe BookmarksController do
       end
     end
 
-    context "with caching" do
+    context "with caching", bookmark_search: true do
       before do
         AdminSetting.first.update_attribute(:enable_test_caching, true)
         run_all_indexing_jobs
--- a/spec/controllers/bookmarks_controller_spec.rb
+++ b/spec/controllers/bookmarks_controller_spec.rb
@@ -54,6 +54,78 @@ describe BookmarksController, bookmark_search: true do
       expect(assigns(:bookmarks)).to include(work_bookmark)
     end
 
+    context "when given chapter_id parameters" do
+      it "sets the work as the bookmarkable" do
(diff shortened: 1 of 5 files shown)


spec/models/bookmark_spec.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
```
redsummernight: We should use double quotes.

spec/controllers/bookmarks_controller_spec.rb
```diff
@@ -1,6 +1,6 @@
 require 'spec_helper'
 
-describe BookmarksController do
+describe BookmarksController, bookmark_search: true do
```
redsummernight: You can add metadata to any describe/context block, not just the top-level one. In this case you should add `bookmark_search: true` to `context "with caching"`, which are the only tests needing ES.

spec/controllers/bookmarks_controller_spec.rb
```diff
@@ -42,4 +42,156 @@ def it_redirects_to_user_login
...
     end
   end
+
+  describe "index" do
+    let!(:external_work_bookmark) { create(:external_work_bookmark) }
+    let!(:series_bookmark) { create(:series_bookmark) }
+    let!(:work_bookmark) { create(:bookmark) }
+
+    it "returns search results when given bookmark_search parameters" do
+      params = { :bookmark_search => { bookmarkable_query: "restricted: false" } }
```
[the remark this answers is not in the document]
redsummernight: We should fix this.

spec/controllers/bookmarks_controller_spec.rb
```diff
@@ -8,37 +8,189 @@ def it_redirects_to_user_login
...
+      end
+    end
+  end
+
+  describe "index" do
+    let!(:external_work_bookmark) { create(:external_work_bookmark) }
+    let!(:series_bookmark) { create(:series_bookmark) }
+    let!(:work_bookmark) { create(:bookmark) }
+
+    it "returns search results when given bookmark_search parameters" do
```
redsummernight: Rereading this again, I don't think this is testing ES. The bookmark shows up because it's in `Bookmark.latest`.

sarken: Yeah, you're right -- I switched it to checking for the series bookmark (which wouldn't by in Bookmark.latest) and that didn't turn up.

PR #3709, opened 2019-12-10
AO3-5813 Remove HTML for mailers from locale file

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5813

## Purpose

The follow-up for the Phrase pull request (#3680)! 

The general mailer layout (specifically, the footer that appears on emails), kudos mailer, and user invite increase mailer all have HTML in the locale file. This banishes back to the view files. 

It also makes a few other tweaks so we can all have a more pleasant time working with the translated emails: 

47a27cd The sent_by key is only used by a mailer that is no longer used. Ergo the mailer and its code should go bye-bye too. Admin notification feature was removed in AO3-5043.

e409667 We reuse the italic bold link style a lot for titles in emails. Let's make it a helper.

35f1631 Improve handling of different numbers in invite_increase_notification. Also capitalize "Archive" because it made Priscilla and me twitch.

## Testing Instructions

Refer to Jira.


spec/miscellaneous/helpers/mailer_helper_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'spec_helper'
+
+describe MailerHelper do
+  describe "style_creation_link" do
+    it "nests red link inside bold inside italics" do
+      work = create(:posted_work)
```
redsummernight: If we can get away with not saving the model (`build(:posted_work)`), we should do that instead.

spec/miscellaneous/helpers/mailer_helper_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'spec_helper'
```
redsummernight: Should we turn on Hound's double quote nagging?

app/views/layouts/mailer.text.erb
```diff
@@ -11,14 +11,11 @@ Archive of Our Own
 <%= yield(:footer_note) %>
 
 <% end %>
-<%= t 'mailer.general.footer.general.text', support_url: root_url+"support", donate_url: donate_url %>
+<%= t 'mailer.general.footer.general.text.part1', support_url: root_url+"support" %>
```
redsummernight: We can use `new_feedback_report_url`.

config/locales/mailers/en.yml
```diff
@@ -3,48 +3,38 @@ en:
...
-          friend at %{invitation_page}.
-          Cheers,
-          %{archive_name}
+      greeting: "Hi %{login}!"
+      body: 
+        one: "We just wanted to let you know that you have %{count} new invitation, which can be used to create a new account at the Archive. You can invite a friend at %{invitation_page}."
+        other: "We just wanted to let you know that you have %{count} new invitations, which can be used to create new accounts at the Archive. You can invite a friend at %{invitation_page}."
+      invitation_page_link_text: your invitations page
+      closing: "Cheers,"
+      signature: "%{app_short_name}"
```
redsummernight: This is unused. The templates hardcode `ArchiveConfig.APP_SHORT_NAME`. Do we want it translatable?

spec/miscellaneous/mailers/user_mailer_spec.rb
```diff
@@ -285,4 +285,68 @@
...
+      end
+
+      # Test both body contents
+      it_behaves_like "a multipart email"
+
+      it_behaves_like "a translated email"
+
+      describe "HTML version" do
+        it "has the correct content" do
+          expect(email).to have_html_part_content("you have #{count} new invitation")
```
redsummernight: You should match on `invitation,` since `invitation` can match `invitations`.

spec/miscellaneous/mailers/user_mailer_spec.rb
```diff
@@ -285,4 +285,68 @@
...
+
+      # Test both body contents
+      it_behaves_like "a multipart email"
+
+      it_behaves_like "a translated email"
+
+      describe "HTML version" do
+         it "has the correct content" do
+          expect(email).to have_html_part_content("you have #{count} new invitations")
+        end
```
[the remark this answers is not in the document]
redsummernight: We should fix this.

spec/miscellaneous/helpers/mailer_helper_spec.rb
sarken: Nah, it gets super annoying -- that's why it's not currently on. Now that I know these aren't always single quotes a A Reason, I won't make this mistake again.

config/locales/mailers/en.yml
sarken: I'll ask!

spec/miscellaneous/mailers/user_mailer_spec.rb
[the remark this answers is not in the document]
sarken: I'm not sure why this didn't get dismissed... it definitely got fixed when I dealt with `it "has the correct content" do` being misaligned (it was indented 3 spaces instead of 2). Here are the lines copied with their leading spaces:

```
        it "has the correct content" do
        end
```

spec/miscellaneous/helpers/mailer_helper_spec.rb
sarken: We need it saved for the route, unfortunately, so it fails like this if I use `build`:

```
  1) MailerHelper style_creation_link nests red link inside bold inside italics
     Failure/Error: expect(style_creation_link(work.title, work_url(work))).to eq("<i><b><a style=\"color:#990000\" href=\"#{work_url(work)}\">#{work.title}</a></b></i>")
     
     ActionController::UrlGenerationError:
       No route matches {:action=>"show", :controller=>"works", :id=>nil}, missing required keys: [:id]
     # ./spec/miscellaneous/helpers/mailer_helper_spec.rb:7:in `block (3 levels) in <top (required)>'

Finished in 1.47 seconds (files took 6.14 seconds to load)
1 example, 1 failure
```

redsummernight: I forgot it needs the ID. This is good as is.

config/locales/mailers/en.yml
redsummernight: If translatable, it may be worth having app_short_name itself as a key (instead of signature), in a more general place (instead of mailer specific).

sarken: Yeah, I agree. It seems rather inefficient to have to translate it every time it's used.

spec/miscellaneous/mailers/user_mailer_spec.rb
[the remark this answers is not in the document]
redsummernight: Looks good, we'll just manually dismiss.

PR #3710, opened 2019-12-11
AO3-5759 Add Lockup to password protect staging

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5759

## Purpose

Adds the [Lockup gem](https://lockupgem.com) for staging and allows us to set a password and password hint in config.yml. 

## Testing Instructions

Refer to Jira.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -180,6 +180,10 @@ group :test, :development, :staging do
   gem 'bullet', '>= 5.7.3'
 end
 
+group :staging do
+  gem 'lockup'
+end
+
 # Deploy with Capistrano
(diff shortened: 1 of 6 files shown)


zz9pzza:
The cookie set by Lockup defaults to 5 years.  ?

Is this too long ?

sarken:
The cookie will expire whenever we change the code word, so I think five years is fine.

derenrich:
you do realize this leaks the password in the source? I assume this is temporary or something?

sarken:
Yes, we put temporary values in the repository and then change them in Ansible.

PR #3720, opened 2020-01-11
AO3-5667 Rewrite the code that generates FilterTaggings from Taggings

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5783

## Purpose

I wrote up a fairly detailed description of this on JIRA, but the short explanation is that it's a rewrite of the code for generating `FilterTagging`s from `Tagging`s.

## Testing Instructions

See JIRA.


app/models/redis_search_index_queue.rb
```diff
@@ -1,74 +0,0 @@
-# Queue the ids of works and bookmarks to be reindexed in resque
-# Usage:
-# RedisSearchIndexQueue.queue_work(work) or queue_bookmark(bookmark)
-# RedisSearchIndexQueue.queue_works(work_ids)
-# RedisSearchIndexQueue.queue_bookmarks(bookmark_ids)
-class RedisSearchIndexQueue
```
redsummernight: I've been wanting this deletion since 2018 :D Thanks!

app/models/work.rb
```diff
@@ -819,65 +806,38 @@ def has_required_tags?
...
+
+    # Reindex any series associated with works whose filters have changed.
+    series_ids = SerialWork.where(work_id: changed_ids).pluck(:series_id)
+    IndexQueue.enqueue_ids(Series, series_ids, queue)
 
-  # Must be called before save
-  def visibility_changed?
-    self.posted_changed? || self.restricted_changed? || self.hidden_by_admin_changed?
+    # Reindex any pseuds associated with works whose filters have changed.
+    pseud_ids = Creatorship.where(creation_id: changed_ids,
```
sarken: Should this be limited to approved creatorships to avoid unnecessary reindexing? (I don't imagine we'll pick up many extras even if we don't limit it to approved, though.)

app/models/concerns/taggable.rb
```diff
@@ -1,63 +1,23 @@
 module Taggable
-
```
redsummernight: Do we ever want to use Taggable without Filterable? Worth a comment here in that case.

I like keeping them as 2 separate modules even if Taggable is never used alone though.

spec/controllers/troubleshooting_controller_spec.rb
```diff
@@ -0,0 +1,253 @@
...
+    end
+
+    context "when logged in as a tag wrangler" do
+      before { fake_login_known_user(tag_wrangler) }
+
+      it "shows a form with the allowed actions for tags" do
+        get :show, params: { tag_id: tag.to_param }
+        expect(response).to render_template(:show)
+        expect(assigns[:allowed_actions]).to \
+          contain_exactly(*%w[fix_associations fix_counts fix_meta_tags])
```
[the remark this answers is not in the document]
redsummernight: I think you can get around this warning by using `eq` the array.

app/models/tag.rb
```diff
@@ -674,302 +654,99 @@ def recategorize(new_type)
...
+    return unless name_changed? || type_changed?
+
+    reindex_pseuds = (type == "Fandom") || (type_was == "Fandom")
+    async_after_commit(:reindex_all, reindex_pseuds)
+  end
+
+  # Reindex anything even remotely related to this tag. This is overkill in
+  # most cases, but necessary when something fundamental like the name or type
+  # of a tag has changed.
+  def reindex_all(reindex_pseuds = false)
```
redsummernight: I find this method name confusing, because it's for reindexing everything related to this tags, not for reindexing all tags (like Searchable's `reindex_all`). Can we rename it and `reindex_all_for_name_or_type_change`?

`reindex_anything_related` but I feel like there's better noun.

spec/controllers/troubleshooting_controller_spec.rb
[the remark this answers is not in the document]
tickinginstant: Hmm, I wanted to use `contain_exactly` so that the tests won't fail if someone decides to reorder the allowed actions (since the order affects how they're displayed). I guess I'll avoid `%w` to try to make Hound happy here.

app/models/concerns/filterable.rb
```diff
@@ -0,0 +1,111 @@
...
+        # Allow for progress messages in long-running updates.
+        yield if block_given?
+      end
+    end
+
+    # This is the callback that gets called when FilterUpdater is done with a
+    # single batch of Filterables of this type. Designed to reindex all of the
+    # Filterables whose filters changed. It's called after_commit, to minimize
+    # issues with stale data.
+    def reindex_for_filter_changes(ids, filter_taggings, queue)
```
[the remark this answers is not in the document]
tickinginstant: Do you mind if I change it to `_ids` instead, or do you prefer the explicit override?

app/models/concerns/taggable.rb
```diff
@@ -1,63 +1,23 @@
...
+    has_many :taggings, as: :taggable, inverse_of: :taggable, dependent: :destroy
+    has_many :tags, through: :taggings, source: :tagger, source_type: 'Tag'
+
+    Tag::VISIBLE.each do |type|
+      has_many type.underscore.pluralize.to_sym,
+               -> { where(tags: { type: type }) },
+               through: :taggings,
+               source: :tagger,
+               source_type: 'Tag',
+               before_remove: :destroy_tagging
```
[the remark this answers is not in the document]
tickinginstant: They're automatically deleted when the tag is removed from the work. But the automatic deletion uses `delete` instead of `destroy` on the taggings, so it was impossible to use callbacks on the taggings before.

features/support/paths.rb
```diff
@@ -10,6 +10,8 @@ def path_to(page_name)
 
     when /the home\s?page/
       '/'
+    when /the media page/
```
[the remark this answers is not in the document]
tickinginstant: It's actually because there's a tricky issue with two of the scenarios in `features/other_a/media.feature`. If you omit this, those scenarios run into an error when calling `path_to_pickle` to get the media page:
```
Could not figure out a path for ["the media"] {} (RuntimeError)
./features/support/paths.rb:37:in `path_to'
./features/step_definitions/web_steps.rb:45:in `/^(?:|I )go to (.+)$/'
[...]/.rvm/rubies/ruby-2.6.5/bin/bundle:23:in `load'
[...]/.rvm/rubies/ruby-2.6.5/bin/bundle:23:in `<main>'
features/other_a/media.feature:42:in `When I go to the media page'
```
It's related to introducing the `:media` factory and removing the explicit `a media exists with name: "NAME", canonical: true` step (which conflicts with the step introduced by adding the factory), but I've never figured it out beyond that.

(Oddly enough, only the scenarios that *don't* use that step that seem to trigger the error. I considered adding an extra `a media exists` step to the two buggy scenarios instead of adding a new path case, but that wouldn't help anyone who wants to add future tests involving the media page.)

app/controllers/troubleshooting_controller.rb
```diff
@@ -0,0 +1,173 @@
...
+      polymorphic_path([@item, :troubleshooting])
+    end
+  end
+
+  # Load the @item based on params. Results in a 404 error if the item in
+  # question can't be found, and a 500 error if there's an unknown type. Also
+  # sets the variable @allowed_actions.
+  def load_item
+    if params[:tag_id]
+      @item = Tag.find_by_name(params[:tag_id])
```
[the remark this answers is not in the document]
tickinginstant: Using `find_by!` instead of `find_by_name` will break the page for all tags with special characters:
https://github.com/otwcode/otwarchive/blob/ff332552fadbec7161f5e0bd5a4c904f333ba88e/app/models/tag.rb#L600-L611

app/models/concerns/async_with_resque.rb
```diff
@@ -0,0 +1,51 @@
...
+    # The class version of the async function. Uses base_class so that we
+    # handle subclasses properly.
+    def async(method, *args)
+      Resque.enqueue(base_class, method, *args)
+    end
+
+    # Actually perform the delayed action.
+    def perform(method, *args)
+      if method.is_a?(Integer)
+        # For backwards compatibility, if the "method" is an integer, we treat
```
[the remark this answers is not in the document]
tickinginstant: Yes, that should be possible.

app/models/concerns/taggable.rb
tickinginstant: Bookmarks have tags without filters, so in theory that would be a class that's Taggable but not Filterable. But bookmarks don't divide up their tags by type, which means that pretty much all of the code in `taggable.rb` is useless for bookmarks. So the Bookmark class doesn't `include Taggable`.

app/models/tag.rb
tickinginstant: Just brainstorming here:
- `reindex_related_records`
- `reindex_related_documents`
- `reindex_related`
- `reindex_associated`
- `reindex_references`

Any of those sound good?

app/controllers/troubleshooting_controller.rb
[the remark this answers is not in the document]
redsummernight: I forgot we did this. Wish we used percent-encoding instead of a custom scheme.

app/models/tag.rb
redsummernight: `reindex_associated` sounds good, matches Rails' terminology too.

https://github.com/rails/rails/commit/4e076b03b6312b928a845f9a31288d10d7243338

app/models/concerns/filterable.rb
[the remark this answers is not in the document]
redsummernight: `_ids` is fine. Can we still name it `ids` in Work since it's used there?

app/models/concerns/taggable.rb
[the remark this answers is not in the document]
redsummernight: :+1:, thanks!

app/models/concerns/filterable.rb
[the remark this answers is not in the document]
tickinginstant: Yep, that was my plan.

app/models/concerns/async_with_resque.rb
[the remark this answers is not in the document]
redsummernight: Is it worth a TODO? Maybe not an issue though.

app/controllers/troubleshooting_controller.rb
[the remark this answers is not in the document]
tickinginstant: Maybe it deserves an issue? Unless there's a specific reason to want to use those custom escape characters. (I guess they're easier to understand than hex?)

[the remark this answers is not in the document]
redsummernight: I don't know of a specific reason for the scheme, but other reviewers might.

I think we're unlikely to change it at this point because of backwards compatibility, though.

app/models/filter_tagging.rb
```diff
@@ -1,89 +1,21 @@
...
-    IndexQueue.enqueue_ids(Pseud, filterable.pseud_ids, :background)
-  end
-
-  def self.find(*args)
-    raise "id is not guaranteed to be unique. please install composite_primary_keys gem and set the primary key to id,filter_id"
-  end
-  def self.find_by(id: id_arg)
-    raise "id is not guaranteed to be unique. please install composite_primary_keys gem and set the primary key to id,filter_id"
-  end
+  validates_uniqueness_of :filter_id, scope: [:filterable_type, :filterable_id]
```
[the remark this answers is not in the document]
tickinginstant: I agree that it would be good to have a DB constraint for `filter_taggings`.

The `meta_taggings` constraint was added in #3585, so that's handled. The `common_taggings` constraint is [sort of enforced already, but not exactly](https://otwarchive.atlassian.net/browse/AO3-5700).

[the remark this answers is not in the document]
redsummernight: Nice, thanks! I was looking at a very outdated output of consistency_fail.

spec/support/matchers/add_to_reindex_queue.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+require 'rspec/expectations'
+
+# A matcher for checking whether a given item has been added to a particular
+# indexing queue.
+#
+# Clears the queue in question before running, so it doesn't work properly in
+# conjunction with run_all_indexing_jobs.
```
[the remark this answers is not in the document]
tickinginstant: I don't want to be picky about whether `enqueue_id` or `enqueue_ids` gets called, and I can't get the `.or()` operator for RSpec matchers to work when using `receive` or `have_received`.

I suppose I could edit the `IndexQueue` to make `enqueue_id` call `enqueue_ids`, but the thing which most frequently calls `enqueue_id` is `enqueue`, which passes `klass` as a string. Whereas most of the calls to `enqueue_ids` involve passing `klass` as a class. So the argument matcher would need to be able to handle that. Maybe a custom argument matcher?

I could alternatively try stubbing out `REDIS_GENERAL.sadd`? Though the ID would sometimes be passed as an item in an array vs. by itself. Custom argument matcher again? Or I could modify `IndexQueue#add_id` to call `sadd` with an array.

(Though if I take that approach, I think I might want to put that into a custom matcher anyway. Something that low-level, used in multiple locations throughout the tests, would make it very annoying to update if we ever want to change how the index queue is handled. It does have the advantage of not wiping the queue, though.)

app/controllers/troubleshooting_controller.rb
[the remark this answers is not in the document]
sarken: I'm pretty sure it was just to make it easier for folks to read/memorize, yeah. I wouldn't be opposed to having a discussion about changing it.

spec/support/matchers/add_to_reindex_queue.rb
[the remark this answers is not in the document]
redsummernight: Stubbing `REDIS_GENERAL` is too low-level, yeah. We should clean up the API of IndexQueue (e.g. passing `klass` consistently as one type), making it easier to be stubbed in the process. Using these matchers currently, it's easy to miss the interaction with `run_all_indexing_jobs`.

We can do this later though.

config/locales/views/en.yml
```diff
@@ -26,3 +26,32 @@ en:
...
+      update_tag_filters:
+        title: "Update Tag Filters"
+        description: "Recalculate filters for all works directly tagged with this tag or one of its synonyms. Does not include subtags. Use this option if wrangling seems to have gone wrong. (e.g. If the tag works list is missing works tagged with a synonym, or if the tags listed in the sidebar are extremely wrong. Keep in mind that the tags listed in the sidebar will never be 100% correct, but if there's a non-canonical tag listed, or if the count is off by an order of magnitude, you should probably use this option.)"
+      reindex_tag:
+        title: "Reindex Tag"
+        description: "Reindex this tag and all related works, bookmarks, series, pseuds, and external works."
+      fix_counts:
+        title: "Fix Tag Counts"
+        description: "Try to recalculate some counts associated with this tag (both filter counts and taggings counts). Keep in mind that the tag counts listed in the fandom lists include unrevealed works, so if the count is only a little off it may just be unrevealed drafts."
```
sarken: >Keep in mind that the tag counts listed in the fandom lists include unrevealed works, so if the count is only a little off it may just be unrevealed drafts.

What is this referring to, https://otwarchive.atlassian.net/browse/AO3-4083? Or is it referring to the count on the tag wrangling sidebar?

tickinginstant: Not the count on the tag wrangling sidebar, and it's not exactly AO3-4083, either (since that involves works being double-counted, and I think it's caused by duplicate FilterTaggings). It's because the FilterCount code includes unrevealed works:
https://github.com/otwcode/otwarchive/blob/f531c2b79b44121e63a9d411e35396619ec39eba/app/models/filter_count.rb#L137
(It does exclude drafts, so I should probably change the wording on the description.)

app/models/tag.rb
```diff
@@ -674,302 +654,99 @@ def recategorize(new_type)
...
+    if reindex_pseuds
+      Pseud.joins(works: :filter_taggings)
+        .merge(self.direct_filter_taggings).reindex_all
     end
-    RedisSearchIndexQueue.queue_bookmarks(bookmark_ids, priority: :low)
-  end
-
-  # We call this to get the ids of all the bookmarks that are tagged by this tag or its subtags
-  # We use ids rather than actual bookmark objects to avoid passing around a lot of instantiated AR objects around
-  # Per discussion with TW chair Emilie, I'm limiting depth of the recursion to 10 here so we don't get stuck in some endlessly deep loop
```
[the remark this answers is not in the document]
tickinginstant: I'm pretty sure we could have stopped then, yeah.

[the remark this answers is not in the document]
redsummernight: Better late than never!

app/views/troubleshooting/show.html.erb
```diff
@@ -0,0 +1,31 @@
...
+
+<% if @item.is_a?(Tag) %>
+  <p><%= link_to @item.name, tag_path(@item) %> (<%= @item.type %>)</p>
+<% elsif @item.is_a?(Work) %>
+  <ol class="work index group">
+    <%= render "works/work_blurb", work: @item %>
+  </ol>
+<% end %>
+
+<%= form_tag troubleshooting_path, method: "patch" do %>
```
sarken: A question from Support, whose committee will be using these pages: Would it cause problems if someone selected all the checkboxes on this page at once?

config/locales/views/en.yml
sarken: Interesting. That doesn't seem like desirable behavior for https://ao3.org/media and https://archiveofourown.org/media/Movies/fandoms, since we no longer display blurbs for unrevealed works on tags' works pages. (It would, however, be desirable for unrevealed collections' fandom pages, e.g. https://ao3.org/collections/smut4smut/fandoms.)

config/locales/views/en.yml
```diff
@@ -26,3 +26,32 @@ en:
...
+      reindex_work:
+        title: "Reindex Work"
+        description: "Send in this work to be reindexed. Use this option if the work isn't appearing in searches, or is appearing in searches that it shouldn't. (e.g. If someone has orphaned their work and it can still be found with the old name, or if it's not behaving as expected when you filter crossovers, completion status, word count, etc.)"
+      update_tag_filters:
+        title: "Update Tag Filters"
+        description: "Recalculate filters for all works directly tagged with this tag or one of its synonyms. Does not include subtags. Use this option if wrangling seems to have gone wrong. (e.g. If the tag works list is missing works tagged with a synonym, or if the tags listed in the sidebar are extremely wrong. Keep in mind that the tags listed in the sidebar will never be 100% correct, but if there's a non-canonical tag listed, or if the count is off by an order of magnitude, you should probably use this option.)"
```
sarken: What does "tags listed in the sidebar are extremely wrong" refer to here? It sounds like the tags listed in the filters, but I'm not sure.

config/locales/views/en.yml
```diff
@@ -26,3 +26,32 @@ en:
...
+      reindex_tag:
+        title: "Reindex Tag"
+        description: "Reindex this tag and all related works, bookmarks, series, pseuds, and external works."
+      fix_counts:
+        title: "Fix Tag Counts"
+        description: "Try to recalculate some counts associated with this tag (both filter counts and taggings counts). Keep in mind that the tag counts listed in the fandom lists include unrevealed works, so if the count is only a little off it may just be unrevealed works."
```
sarken: Are there any differences between this and the current "Fix Taggings Count" checkbox?

config/locales/views/en.yml
```diff
@@ -26,3 +26,32 @@ en:
...
+      fix_associations:
+        title: "Fix Tag Associations"
+        description: "Try to delete invalid associations involving this tag. (e.g. Parents of the wrong type, children of the wrong type, duplicate parents/children/meta-tags/sub-tags, etc.)"
+      fix_meta_tags:
+        title: "Fix Meta Tags"
+        description: "Try to recalculate inherited meta tags. Use this option if this tag has grandparent meta tags (i.e. a meta tag with its own meta tag), and the works aren't showing up in the grandparent's listing. You might also try this if you've removed a meta tag, but the works are still showing up in the former meta tag's listing."
```
sarken: So this one would be used to fix existing instances of https://otwarchive.atlassian.net/browse/AO3-5507 and https://otwarchive.atlassian.net/browse/AO3-5508?

config/locales/views/en.yml
```diff
@@ -26,3 +26,32 @@ en:
...
+      reindex_work:
+        title: "Reindex Work"
+        description: "Send in this work to be reindexed. Use this option if the work isn't appearing in searches, or is appearing in searches that it shouldn't. (e.g. If someone has orphaned their work and it can still be found with the old name, or if it's not behaving as expected when you filter crossovers, completion status, word count, etc.)"
+      update_tag_filters:
+        title: "Update Tag Filters"
+        description: "Recalculate filters for all works directly tagged with this tag or one of its synonyms. Does not include subtags. Use this option if wrangling seems to have gone wrong. (e.g. If the tag works list is missing works tagged with a synonym, or if the tags listed in the sidebar are extremely wrong. Keep in mind that the tags listed in the sidebar will never be 100% correct, but if there's a non-canonical tag listed, or if the count is off by an order of magnitude, you should probably use this option.)"
+      reindex_tag:
+        title: "Reindex Tag"
+        description: "Reindex this tag and all related works, bookmarks, series, pseuds, and external works."
```
sarken: What sort of scenarios would this be useful in?

app/models/concerns/filterable.rb
```diff
@@ -0,0 +1,113 @@
...
+        end
+
+        # Allow for progress messages in long-running updates.
+        yield if block_given?
+      end
+    end
+
+    # This is the callback that gets called when FilterUpdater is done with a
+    # single batch of Filterables of this type. Designed to reindex all of the
+    # Filterables whose filters changed. It's called in after_commit, to
```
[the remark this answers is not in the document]
tickinginstant: Because removing `taggings` triggers the removal of the corresponding `filter_taggings`, and `taggings` are deleted when the work is deleted, the `after_destroy :clean_up_filter_taggings` callback isn't strictly necessary. I think I may have written it during a period where I was trying to handle work deletion differently, and never realized that I could delete it once I abandoned the changes that I was working on.

app/models/tag.rb
```diff
@@ -674,302 +654,99 @@ def recategorize(new_type)
...
+  # When canonical or merger_id changes, only the items directly tagged with
+  # this tag need their filters updated, so we queue up a call to
+  # update_filters_for_taggables after commit.
+  #
+  # Note that when a tag becomes non-canonical, all of its filter-taggings need
+  # to be deleted. But when a tag becomes non-canonical, all of its mergers and
+  # sub-tags will be deleted, which will result in the necessary items having
+  # their filters fixed.
+  after_update :update_filters_for_canonical_or_merger_change
+  def update_filters_for_canonical_or_merger_change(*)
```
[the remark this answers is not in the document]
tickinginstant: I think this is another case of code left over from some experiments. I'll delete it.

app/models/tag.rb
```diff
@@ -674,302 +654,99 @@ def recategorize(new_type)
...
-            tags_that_need_filter_count_reset << m
-          end
-        end
+  # When canonical or merger_id changes, only the items directly tagged with
+  # this tag need their filters updated, so we queue up a call to
+  # update_filters_for_taggables after commit.
+  #
+  # Note that when a tag becomes non-canonical, all of its filter-taggings need
+  # to be deleted. But when a tag becomes non-canonical, all of its mergers and
+  # sub-tags will be deleted, which will result in the necessary items having
```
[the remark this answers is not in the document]
tickinginstant: Yeah. Calling `update_filters_for_filterables` on the main tag and `update_filters_for_taggables` on all of its synonyms would cause all of the works tagged with one of the synonyms to be processed twice.

app/models/tag.rb
```diff
@@ -1049,65 +826,87 @@ def remove_association(tag_id)
...
+    end
+
+    self.direct_sub_tags.find_each do |tag|
+      self.merger.sub_taggings.create(sub_tag: tag)
+    end
   end
 
-  def remove_favorite_tags
-    favorite_tags.destroy_all
+  # If this tag has a canonical merger, move all favorite tags to the merger.
```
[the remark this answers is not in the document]
tickinginstant: It's mostly for inconsistencies. When a tag becomes non-canonical, there should be a job enqueued to remove all of its synonyms, but if that doesn't run (or if it takes a long time to run), then the non-canonical merger will exist in the meantime.

app/models/filter_updater.rb
```diff
@@ -0,0 +1,210 @@
...
+      inherited = filter_ids.flat_map { |filter_id| @meta_tags[filter_id] }
+      @inherited_filters[item_id] = (inherited - filter_ids).uniq
+    end
+  end
+
+  # Given a list of pairs of IDs, treat each pair as a (key, value) pair, and
+  # return a hash that associates each key with a list of values. Sets the
+  # default value of the hash to an empty frozen list.
+  def hash_from_pairs(pairs)
+    hash = Hash.new([].freeze)
```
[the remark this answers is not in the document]
tickinginstant: I froze the default value because having a non-frozen default value can cause confusion if you ever try to modify the default value. All of the other lists in the hash shouldn't have that issue, but I can still freeze them if you think it'd be useful.

app/views/troubleshooting/show.html.erb
tickinginstant: It should be fine. Some of the checkboxes have overlapping functionality (e.g. fixing filters always triggers a reindex), but it shouldn't cause problems aside from generating extra work for the servers.

app/models/filter_updater.rb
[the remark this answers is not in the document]
redsummernight: I think we should. Freezing will make it clear we don't expect to modify the lists afterwards.

config/locales/views/en.yml
tickinginstant: Yeah, it's about the tags listed in the filters. Like what happened with the BTVS tags and the Hockey RPF tags a while back.

config/locales/views/en.yml
tickinginstant: This option does what the current "Fix Taggings Count" checkbox does, but also recalculates the count on the media/fandom pages (the filter count).

config/locales/views/en.yml
tickinginstant: The "Fix Meta Tags" option would be for AO3-5508. AO3-5507 doesn't affect the inherited meta tags (those are calculated correctly), so you'd need the "Update Tag Filters" option to fix that one.

config/locales/views/en.yml
tickinginstant: This runs the code that's meant to be run whenever a tag is renamed, so it could be useful if there was a tag renamed before this code is deployed and the search results are off. (Or if it was renamed after the deploy, and the job failed and/or ran into stale data issues.)

It's meant to replace the current "Reindex" button, though it works slightly differently.

spec/support/matchers/add_to_reindex_queue.rb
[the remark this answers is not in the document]
redsummernight: :+1: for the updated matchers that don't interfere with `run_all_indexing_jobs`.

spec/models/tag_wrangling_spec.rb
```diff
@@ -0,0 +1,536 @@
...
+
+    it "deletes meta filters when a syn and a meta tag are deleted" do
+      canonical = create(:canonical_fandom)
+      meta = create(:canonical_fandom)
+      canonical.meta_tags << meta
+      synonym = create(:fandom, merger: canonical)
+
+      work = create(:work, fandom_string: synonym.name)
+
+      suspend_resque_workers do
```
redsummernight: Is it possible for this block to miss any `Resque.enqueue` calls before it? Should all the set up steps be inside the block too?

app/controllers/troubleshooting_controller.rb
```diff
@@ -0,0 +1,181 @@
...
+  def fix_meta_tags
+    modified = MetaTagging.transaction do
+      InheritedMetaTagUpdater.new(@item).update
+    end
+
+    if modified
+      # Fixing the meta taggings is all well and good, but unless the filters
+      # are adjusted too, this will have no immediate effects.
+      @item.async(:update_filters_for_filterables)
+      flash[:notice] << ts("Inherited meta tags recalculated. This tag has " \
```
sarken: Could you change this, and the instances on lines 171 and 172, to the official spelling of "metatags"?

app/models/concerns/filterable.rb
```diff
@@ -0,0 +1,104 @@
...
+    (tags.pluck(:id) + filters.pluck(:id)).uniq
+  end
+
+  # Index only direct filters (non meta-tags) for facets
+  def filters_for_facets
+    @filters_for_facets ||= direct_filters.to_a
+  end
+
+  def rating_ids
+    filters_for_facets.select { |t| t.type == 'Rating' }.map(&:id)
```
sarken: Do we need to use single quotes here, or will double work?

app/models/concerns/taggable.rb
```diff
@@ -1,63 +1,23 @@
...
+
+    has_many :taggings, as: :taggable, inverse_of: :taggable, dependent: :destroy
+    has_many :tags, through: :taggings, source: :tagger, source_type: 'Tag'
+
+    Tag::VISIBLE.each do |type|
+      has_many type.underscore.pluralize.to_sym,
+               -> { where(tags: { type: type }) },
+               through: :taggings,
+               source: :tagger,
+               source_type: 'Tag',
```
sarken: Same question, can we change to double quotes here?

app/models/filter_tagging.rb
```diff
@@ -1,89 +1,21 @@
 # This is essentially a mirror of the taggings table as applied to works (right now)
 # except with all works connected to canonical tags instead of their synonyms for
 # browsing and filtering purposes. Filter = tag, filterable = thing that's been tagged.
+#
+# Note that this doesn't have very many validations or callbacks -- this is
+# because the ONLY class that should be adding/updating FilterTaggings is the
+# FilterUpdater, which handles those itself.
 class FilterTagging < ApplicationRecord
-  self.primary_key = 'id'
+  belongs_to :filter, class_name: 'Tag', inverse_of: :filter_taggings
```
sarken: Double quotes here and line 18?

app/models/meta_tagging.rb
```diff
@@ -25,42 +25,14 @@ def meta_tag_validation
       if self.meta_tag == self.sub_tag
         self.errors.add(:base, "A tag can't be its own meta tag.")
       end
-      if self.meta_tag.meta_tags.include?(self.sub_tag)
-        self.errors.add(:base, "A meta tag can't be its own grandpa.")
+      if self.meta_tag.meta_taggings.where(meta_tag: self.sub_tag).exists?
+        self.errors.add(:base, "A meta tag can't be its own grandparent.")
```
sarken: Can we change this and line 26 to "metatag"?

spec/models/tag_wrangling_spec.rb
tickinginstant: It is guaranteed to miss any `Resque.enqueue` calls before it, but that's by design -- the point of the test is to have everything working properly before the wrangling queue is delayed and breaks (or preferably doesn't break) everything.

app/controllers/troubleshooting_controller.rb
tickinginstant: The header on the tag page currently spells it as "Meta tags."
https://github.com/otwcode/otwarchive/blob/78bdfeefc003aa3e904f6ca66a9ad00f4557886b/app/views/tags/show.html.erb#L74
Are there some contexts where "meta tags" is preferred over "metatags," or should I change the header, too? Or should I restrict myself to fixing the spelling only on the lines that I've already touched? (Does that include test names and comments, or is it only the error messages that matter?) Is it okay to have an error message that spells it once as "meta tags" and once as "metatags"? That is, do these lines also need changing?
https://github.com/otwcode/otwarchive/blob/78bdfeefc003aa3e904f6ca66a9ad00f4557886b/app/models/tag.rb#L1182
https://github.com/otwcode/otwarchive/blob/78bdfeefc003aa3e904f6ca66a9ad00f4557886b/app/models/meta_tagging.rb#L10
(I think I'd also need to change it in the locales file, too, because the attribute name is used in that second error message, and it defaults to turning the underscore in `meta_tag_id` into a space.)

sarken: "Metatags" is what's consistently used in documentation, so in an ideal world, it should be that everywhere. But unless you're itching to fix it everywhere, I'm happy to just stick with the lines you've touched, and really only worrying about user-facing stuff (e.g. error messages, not code comments or test names). If fixing all instances of "meta tags" in certain error messages you've touched will be a pain, we can live with a mix of "meta tags" and "metatags."

tickinginstant: I think I'll take a crack at changing it in more places. Should "sub tag" become "subtag" as well?

sarken: Sounds good! And yeah, "subtag" is also preferred. Thank you so much!

PR #3721, opened 2020-01-12
AO3-5759  adding Lockup to remove password protection on staging

zz9pzza:
Reverts otwcode/otwarchive#3717


sarken:
Exact same code we've done before, merging.

PR #3723, opened 2020-01-16
AO3-5844 Migration to add user_id column to kudos table

chrisrlc:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5844

## Purpose

This PR adds a new (unused) user_id column and (not yet unique) index to the kudos table.

For staging + beta, use pt-online-schema-change --alter args:

up:
```
ADD COLUMN user_id INT 
ADD INDEX index_kudos_on_user_id (user_id)
```

down:
```
DROP COLUMN user_id 
DROP INDEX index_kudos_on_user_id (user_id)
```

Side note: I've seen migrations (outside of otw) that require a ptosc in specific environments wrap the migration code in a conditional a la:
```
def up
    if Rails.env.production? || Rails.env.staging?
    =begin
        pt-online-schema-change command with ADD COLUMN and ADD INDEX in block comment
    =end
    else
        # migration code
    end
end

def down
    # Same as above but reversed
end
```

But I didn't see this done in any of our existing migrations. How do we normally prevent migrations from being run in beta and staging?

## References

Epic: https://otwarchive.atlassian.net/browse/AO3-5597

## Credit

cyrilcee

she/her

### DIFF
--- a/db/migrate/20200115232918_add_user_id_to_kudos.rb
+++ b/db/migrate/20200115232918_add_user_id_to_kudos.rb
@@ -1,0 +1,6 @@
+class AddUserIdToKudos < ActiveRecord::Migration[5.1]
+  def change
+    add_column :kudos, :user_id, :int
+    add_index :kudos, :user_id
+  end
+end
--- a/db/migrate/20200115232918_add_user_id_to_kudos.rb
+++ b/db/migrate/20200115232918_add_user_id_to_kudos.rb
@@ -1,6 +1,34 @@
 class AddUserIdToKudos < ActiveRecord::Migration[5.1]
-  def change
-    add_column :kudos, :user_id, :int
-    add_index :kudos, :user_id
+  def up
+    if Rails.env.staging? || Rails.env.production?
+=begin
+      pt-online-schema-change -uroot --alter \
+      "ADD COLUMN user_id INT, ADD INDEX index_kudos_on_user_id (user_id)" \
+      D=$DATABASE,t=kudos --chunk-size=10k --max-flow-ctl 0 \
+      --pause-file /tmp/pauseme --max-load Threads_running=25 \
+      --critical-load Threads_running=400 --set-vars innodb_lock_wait_timeout=2 \
+      --alter-foreign-keys-method=auto --execute
+      insert into schema_migrations (version) values ("20200115232918")
+=end
+    else
+      add_column :kudos, :user_id, :int
+      add_index :kudos, :user_id
+    end
+  end
+
+  def down
+    if Rails.env.staging? || Rails.env.production?
+=begin
+      pt-online-schema-change -uroot --alter \
+      "DROP COLUMN user_id, DROP INDEX index_kudos_on_user_id (user_id)" \
+      D=$DATABASE,t=kudos --chunk-size=10k --max-flow-ctl 0 \
+      --pause-file /tmp/pauseme --max-load Threads_running=25 \
+      --critical-load Threads_running=400 --set-vars innodb_lock_wait_timeout=2 \
+      --alter-foreign-keys-method=auto --execute
+=end
+    else
+      remove_index :kudos, name: 'index_kudos_on_user_id'
+      remove_column :kudos, :user_id
+    end
   end
 end
--- a/db/migrate/20200115232918_add_user_id_to_kudos.rb
+++ b/db/migrate/20200115232918_add_user_id_to_kudos.rb
@@ -19,15 +19,13 @@ class AddUserIdToKudos < ActiveRecord::Migration[5.1]
   def down
     if Rails.env.staging? || Rails.env.production?
 =begin
-      pt-online-schema-change -uroot --alter \
-      "DROP COLUMN user_id, DROP INDEX index_kudos_on_user_id (user_id)" \
+      pt-online-schema-change -uroot --alter "DROP COLUMN user_id" \
       D=$DATABASE,t=kudos --chunk-size=10k --max-flow-ctl 0 \
       --pause-file /tmp/pauseme --max-load Threads_running=25 \
       --critical-load Threads_running=400 --set-vars innodb_lock_wait_timeout=2 \
       --alter-foreign-keys-method=auto --execute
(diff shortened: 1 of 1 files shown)


redsummernight:
> But I didn't see this done in any of our existing migrations. How do we normally prevent migrations from being run in beta and staging?

We don't do anything special other than not running `rake db:migrate` for such migrations. @zz9pzza would if necessary mark a migration as done by manually running this directly in mysql:
```
INSERT INTO schema_migrations (version) VALUES ("<version>");
```
----
Can you write down the `--alter` args for a rollback as well?

chrisrlc:
> Can you write down the `--alter` args for a rollback as well?

Updated.

redsummernight:
Thanks!

> Side note: I've seen migrations (outside of otw) that require a ptosc in specific environments wrap the migration code in a conditional a la:

We should do that. Can you have the migration use pt-osc for all environments except development?

Taking inspiration from https://github.com/departurerb/departure, we should also allow specifying pt-osc arguments, like:

```
PERCONA_ARGS='--chunk-size=10k' bundle exec rake db:migrate:up VERSION=xxx
```

chrisrlc:
@redsummernight - 3 questions:

1. Do we want the migration to actually run the pt-osc? Or should we just add the pt-osc command as commented out code, so that we can run it manually on the appropriate servers? For example, I could write the following:

```
class AddUserIdToKudos < ActiveRecord::Migration[5.1]
  def up
    if Rails.env.development?
      add_column :kudos, :user_id, :int
      add_index :kudos, :user_id
    else
      =begin
      pt-online-schema-change --alter \
      "ADD COLUMN user_id INT, ADD INDEX index_kudos_on_user_id (user_id)" \
      D=TODO,t=kudos --chunk-size=10k
      insert into schema_migrations (version) values ("20200115232918")
      =end
    end
  end

  def down
    if Rails.env.development?
      drop_column :kudos, :user_id
      drop_index :kudos, :user_id
    else
      =begin
      pt-online-schema-change --alter \
      "DROP COLUMN user_id, DROP INDEX index_kudos_on_user_id (user_id)" \
      D=TODO,t=kudos --chunk-size=10k
      =end
    end
  end
end
```

This way would mean we'd still be able to run the pt-osc manually, we wouldn't accidentally run this migration if we ran bundle exec rake db:migrate in non-development environments, and we'd have the pt-osc command where folks can see it.

If instead we wanted to have the migration run the pt-osc automatically, I can put the pt-osc command in a ruby exec(), but this gives us less control over when we run it. And it seems like we'd need to find a gem that would give us the functionality to specify the percona args while we run `bundle exec rake db:migrate` (does one exist?), or write it ourselves (possibly overkill?).

It seems it would be an improvement just adding in the environment condition and the commented out pt-osc command, but what do you think?

2. Do we want the pt-osc to run for all non-development environments, even test?

3. I didn't know what the full pt-osc command should be (incl. percona args and db name), so I made it up for the example above. If we do want the pt-osc command in this migration (either as a comment or exec() argument), what's the full command?

redsummernight:
> It seems it would be an improvement just adding in the environment condition and the commented out pt-osc command, but what do you think?

That sounds good. I misread your original suggestion (thinking the comment is pseudo-code). If you add the environment condition where staging/production is the pt-osc comment, that covers everything (having the pt-osc command visible somewhere, not accidentally running Rails migration on the wrong environments, and also not having to worry about issues with exec and flexible pt-osc's arguments).

chrisrlc:
Updated PR with prod/staging environment condition and pt-osc command as comment (using prior pt-osc args and $DATABASE var).

db/migrate/20200115232918_add_user_id_to_kudos.rb
```diff
@@ -0,0 +1,34 @@
...
+      add_column :kudos, :user_id, :int
+      add_index :kudos, :user_id
+    end
+  end
+
+  def down
+    if Rails.env.staging? || Rails.env.production?
+=begin
+      pt-online-schema-change -uroot --alter \
+      "DROP COLUMN user_id, DROP INDEX index_kudos_on_user_id (user_id)" \
```
tickinginstant: This is unlikely to ever come up, but: I don't think the syntax on this `DROP INDEX` statement is right. You shouldn't include the parenthesized column name the way that you did for `ADD INDEX`.

(Also, I'm not sure you need the call to `DROP INDEX`, because the only column that index includes is `user_id`, which is being dropped. So you might be able to get away with removing the `DROP INDEX` statement entirely.)

chrisrlc:
Thanks @tickinginstant and @redsummernight - I went ahead and removed the unnecessary (and incorrect) DROP INDEX from the pt-osc comment and the unnecessary remove_index from the rollback code.

db/migrate/20200115232918_add_user_id_to_kudos.rb
```diff
@@ -0,0 +1,32 @@
+class AddUserIdToKudos < ActiveRecord::Migration[5.1]
+  def up
+    if Rails.env.staging? || Rails.env.production?
+=begin
+      pt-online-schema-change -uroot --alter \
+      "ADD COLUMN user_id INT, ADD INDEX index_kudos_on_user_id (user_id)" \
+      D=$DATABASE,t=kudos --chunk-size=10k --max-flow-ctl 0 \
+      --pause-file /tmp/pauseme --max-load Threads_running=25 \
+      --critical-load Threads_running=400 --set-vars innodb_lock_wait_timeout=2 \
+      --alter-foreign-keys-method=auto --execute
```
zz9pzza: missing a \ at the end

chrisrlc: @zz9pzza - These should be 2 separate commands that I just provided for copy-paste convenience: the multi-line pt-osc command that's run on the command line ("pt-online-schema-change...") and a sql command that's run on the mysql server ("insert into..."). Or at least, I'm assuming you usually run those commands separately like that (this is the first migration I've written for OTW)?

I can add extra documentation in this comment to make that clearer? Or is there something else that seems broken?

PR #3735, opened 2020-01-23
AO3-5708 Remove BulletproofMailer.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5708

## Purpose

Remove the BulletproofMailer class, which is no longer necessary.

## Testing Instructions

See JIRA.


app/mailers/user_mailer.rb
```diff
@@ -116,7 +116,7 @@ def batch_subscription_notification(subscription_id, entries)
 
     # die if we haven't got any creations to notify about
```
redsummernight: Can you remove these comments?

PR #3736, opened 2020-01-24
AO3-5844 Remove unnecessary schema_migrations insert comment from migration

chrisrlc:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5844

## Purpose

According to comments in the [JIRA issue](https://otwarchive.atlassian.net/browse/AO3-5844), we no longer need to run "insert into schema_migrations..." now that rake db:migrate can run in all environments. So this PR removes this unnecessary comment.

Sorry about all the back and forth on this straightforward JIRA ticket, everyone! But since this is the first time we're adding the migration commands for staging and prod in the migration file as comments, hopefully this can serve as an example for future migrations to mimic.

## References

Previous PR, already merged into otwcode/otwarchive:master: https://github.com/otwcode/otwarchive/pull/3723


sarken:
No need to apologize! Like you said, this is the first time we're doing this, so it takes a while to get things figured out. Thank you for your patience on this!

PR #3738, opened 2020-01-26
AO3-5028 Notify creators when a collection maintainer makes their work unrevealed and/or anonymous.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5028

## Purpose

This PR adds a new callback to the CollectionItem class that checks whether the maintainer is making the item unrevealed/anonymous when the work isn't already unrevealed/anonymous. If so, the callback sends a notification email to all of the work's creators.

## Testing Instructions

See JIRA.

### DIFF
--- a/app/models/collection_item.rb
+++ b/app/models/collection_item.rb
@@ -341,8 +341,6 @@ class CollectionItem < ApplicationRecord
     return if item.users.include?(User.current_user)
 
     item.users.each do |user|
-      next if user.preference.collection_emails_off
-
       UserMailer.anonymous_or_unrevealed_notification(
         user.id, item.id, collection.id,
         anonymous: newly_anonymous, unrevealed: newly_unrevealed
--- a/app/models/collection_item.rb
+++ b/app/models/collection_item.rb
@@ -320,4 +320,33 @@ class CollectionItem < ApplicationRecord
       end
     end
   end
+
+  after_update :notify_of_unrevealed_or_anonymous
+  def notify_of_unrevealed_or_anonymous
+    # This CollectionItem's anonymous/unrevealed status can only affect the
+    # item's status if (a) the CollectionItem is approved by the user and (b)
+    # the item is a work. (Bookmarks can't be anonymous/unrevealed at the
+    # moment.)
+    return unless approved_by_user? && item.is_a?(Work)
+
+    # Check whether anonymous/unrevealed is becoming true, when the work
+    # currently has it set to false:
+    newly_anonymous = (saved_change_to_anonymous?(to: true) && !item.anonymous?)
+    newly_unrevealed = (saved_change_to_unrevealed?(to: true) && !item.unrevealed?)
+
+    return unless newly_unrevealed || newly_anonymous
+
+    # Don't notify if it's one of the work creators who is changing the work's
+    # status.
+    return if item.users.include?(User.current_user)
+
+    item.users.each do |user|
+      next if user.preference.collection_emails_off
+
+      UserMailer.anonymous_or_unrevealed_notification(
+        user.id, item.id, collection.id,
+        anonymous: newly_anonymous, unrevealed: newly_unrevealed
+      ).deliver
+    end
+  end
 end
--- a/app/views/user_mailer/anonymous_or_unrevealed_notification.html.erb
+++ b/app/views/user_mailer/anonymous_or_unrevealed_notification.html.erb
@@ -1,0 +1,26 @@
+<% content_for :message do %>
+  <p><%= t(".greeting", user: style_bold(@user.default_pseud.byline)).html_safe %></p>
+
+  <p><%= t(".changed_status",
+           collection: style_link(@collection.title, collection_url(@collection)),
+           work: style_link(@work.title.html_safe, work_url(@work)),
+           status: @status).html_safe %></p>
+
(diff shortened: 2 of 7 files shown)


app/views/user_mailer/anonymous_or_unrevealed_notification.html.erb
```diff
@@ -0,0 +1,26 @@
+<% content_for :message do %>
+  <p><%= t(".greeting", user: style_bold(@user.default_pseud.byline)).html_safe %></p>
+
+  <p><%= t(".changed_status",
+           collection: style_link(@collection.title, collection_url(@collection)),
+           work: style_link(@work.title.html_safe, work_url(@work)),
+           status: @status).html_safe %></p>
```
sarken: I'm checking with Translation on this. I know when it comes to numbers, they want us to remove interpolation and instead create a separate key depending on the possible value, but I'm not sure if they'll have issues with adjectives as well. If that's a problem for them, we'll need to change it to something more like:
```
<p><%= t(".changed_status.#{@status}",
            collection: style_link(@collection.title, collection_url(@collection)),
            work: style_link(@work.title.html_safe, work_url(@work)).html_safe %></p>

changed_status: 
  unrevealed: "The collection maintainers of %{collection} have changed the status of your work %{work} to unrevealed."
  anonymous: "The collection maintainers of %{collection} have changed the status of your work %{work} to anonymous."
  anonymous_unrevealed: "The collection maintainers of %{collection} have changed the status of your work %{work} to anonymous and unrevealed."
```

app/views/user_mailer/anonymous_or_unrevealed_notification.html.erb
```diff
@@ -0,0 +1,26 @@
+<% content_for :message do %>
+  <p><%= t(".greeting", user: style_bold(@user.default_pseud.byline)).html_safe %></p>
+
+  <p><%= t(".changed_status",
+           collection: style_link(@collection.title, collection_url(@collection)),
+           work: style_link(@work.title.html_safe, work_url(@work)),
```
sarken: I think we should use `style_creation_link` here, now that we have it:  https://github.com/otwcode/otwarchive/blob/b3c9f726a879b754c84a960d1611f0037dcef0f4/app/helpers/mailer_helper.rb#L12-L16

app/views/user_mailer/anonymous_or_unrevealed_notification.html.erb
tickinginstant: Have you heard back from Translation about this?

sarken: Not yet, but I've poked them again and will let you know when I hear back!

sarken: The verdict is it's probably not _necessary_ to change it, but they'd find it neater, so if you could change it, that would be great!

app/models/collection_item.rb
```diff
@@ -320,4 +320,33 @@ def notify_of_reveal
...
+    newly_unrevealed = (saved_change_to_unrevealed?(to: true) && !item.unrevealed?)
+
+    return unless newly_unrevealed || newly_anonymous
+
+    # Don't notify if it's one of the work creators who is changing the work's
+    # status.
+    return if item.users.include?(User.current_user)
+
+    item.users.each do |user|
+      next if user.preference.collection_emails_off
```
sarken: I have a hunch we might want to ignore the preference for something as important as this, but I'm going to check.

sarken: Yup, Support would rather deal with users grumpy about an email than grumpy about their work disappearing. If you could remove this check and update the text of the help file with the new version Docs provided, this should be ready to merge. 🎉 

The old help text line: https://github.com/otwcode/otwarchive/blob/57e7cab589abc7cc666ddf4d3abe768d801b3493/public/help/collection-preferences.html#L9

The new version:
```html
  <dd>Enable this option if you would prefer not to receive email alerts from collections, such as creator and work reveals. We’ll still email you if your username or work is hidden after being added to a collection. Notifications will still be delivered to your Archive inbox unless you have chosen to disable them.</dd>
```

tickinginstant: About the wording: Users currently aren't notified when their works are revealed. (a) Should there be an issue to add those notifications? (b) Is it okay that the help text is more about what it's intended to control than about what it currently controls? (Collection invitations, mostly.)

sarken: I'll be honest: I don't think we put that much thought into the original version of the help text or even verified that it does as it says. There's no paper trail beyond a decision to add some hypothetical text in a November 2011 meeting, the issue itself ([AO3-2704](https://otwarchive.atlassian.net/browse/AO3-2704)), and Ariana's pull request in 2012 (#662). So I wouldn't really be inclined to just make an issue about that.

Let me look into what the preference _actually_ does and then ping Docs to see if we can get wording that reflects the present reality.

sarken: All righty, here we go:

>Enable this option if you would prefer not to receive email alerts when your works are added or invited to collections.

tickinginstant: Thanks for checking with Docs to figure out what the wording should be!

PR #3739, opened 2020-01-26
AO3-5850 Update user IDs when logged in kudos are changed

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5850

## Purpose

Begin saving to the new user ID column on the kudos table.

## Testing Instructions

See issue.

### DIFF
--- a/spec/controllers/kudos_controller_spec.rb
+++ b/spec/controllers/kudos_controller_spec.rb
@@ -6,27 +6,35 @@ describe KudosController do
   include LoginMacros
   include RedirectExpectationHelper
 
-  describe "create" do
+  describe "POST #create" do
     context "when work is public" do
       let(:work) { create(:posted_work) }
+      let(:referer) { work_path(work) }
+      before { request.headers["HTTP_REFERER"] = referer }
 
-      context "when kudos are given from work" do
-        it "redirects to referer with a notice" do
-          referer = work_path(work)
-          request.headers["HTTP_REFERER"] = referer
-          post :create, params: { kudo: { commentable_id: work.id, commentable_type: "Work" } }
-          it_redirects_to_with_comment_notice(referer, "Thank you for leaving kudos!")
-          expect(assigns(:kudo).user).to be_nil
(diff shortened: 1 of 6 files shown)


spec/controllers/kudos_controller_spec.rb
```diff
@@ -16,6 +16,7 @@
           request.headers["HTTP_REFERER"] = referer
           post :create, params: { kudo: { commentable_id: work.id, commentable_type: "Work" } }
           it_redirects_to_with_comment_notice(referer, "Thank you for leaving kudos!")
+          expect(assigns(:kudo).user).to be_nil
```
sarken: I'm going to be a pain in the ass and suggest

a) we should wrap these tests in a `context "when kudos giver is a guest"`
b) we update the descriptions in all the new/modified specs to mention we're checking that `user` is/is not assigned correctly, not just checking the the `referer` and notice. (If I wanted to be an even bigger pain in the ass, I'd say we probably shouldn't be checking two things at once like that, but I don't care that strongly.)

spec/models/user_spec.rb
```diff
@@ -1,119 +1,133 @@
-require 'spec_helper'
-
-describe User, :ready do
+require "spec_helper"
```
sarken: ty for the tidying!

spec/controllers/kudos_controller_spec.rb
```diff
@@ -6,38 +6,66 @@
   include LoginMacros
   include RedirectExpectationHelper
 
-  describe "create" do
+  describe "POST #create" do
     context "when work is public" do
       let(:work) { create(:posted_work) }
+      let(:referer) { work_path(work) }
+      before { request.headers["HTTP_REFERER"] = referer }
```
redsummernight: Previously we had `referer = work_chapter_path(work, work.first_chapter)` in one case, but for the purpose of this test it doesn't matter what the referer is, only that it is present. So I pulled it out here.

PR #3743, opened 2020-01-30
AO3-5735 Fix admin post direction attributes

palindromordnilap:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5735

## Purpose

Fixes the previous PR for this issue, which didn't work. This should add the `dir="rtl"` attribute to the header and userstuff divs of admin posts with a language set to Arabic or Hebrew.

## Testing Instructions

See the Jira ticket and the previous PR.

## References

Previous, broken pull request: https://github.com/otwcode/otwarchive/pull/3722

## Credit

Alix R, she/her


palindromordnilap:
Actually, I just realized I could use a helper method instead of directly writing in the code. Hold on.

app/helpers/language_helper.rb
```diff
@@ -7,6 +7,10 @@ def rtl?
     %w(ar he).include?(Globalize.locale.to_s)
   end
 
+  def post_rtl?(post)
```
redsummernight: Should we have the language instead of the post as the parameter in case we need to use the helper elsewhere?

sarken: Yeah, I agree it would be good to make it more reusable. We could potentially use it for works, too.

app/views/admin_posts/_admin_post.html.erb
```diff
@@ -1,7 +1,7 @@
 <% # expects "admin_post" %>
-<div <% if rtl? %>dir="rtl"<% end %> class="header">
+<div <% if post_rtl?(admin_post) %>dir="rtl"<% end %> class="header">
   <h3 class="heading">
-    <%= link_to admin_post.title.html_safe, admin_post %>
+   <%= link_to admin_post.title.html_safe, admin_post %>
```
redsummernight: Minor thing: can you indent this one more space?

palindromordnilap: Ah, yes, I think my IDE's autoformatter got confused.

app/helpers/language_helper.rb
palindromordnilap: Done! I thought it worked the same way for works, but this is clearer that way regardless.

app/helpers/language_helper.rb
```diff
@@ -7,6 +7,10 @@ def rtl?
     %w(ar he).include?(Globalize.locale.to_s)
   end
 
+  def is_rtl?(lang)
```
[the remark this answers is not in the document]
palindromordnilap: Yeah, I saw that one coming. Should we merge the two helpers somehow?

app/helpers/language_helper.rb
```diff
@@ -7,6 +7,10 @@ def rtl?
     %w(ar he).include?(Globalize.locale.to_s)
   end
 
+  def rtl_lang?(lang)
```
sarken: Could we tweak this slightly, both for clarity and so there's a little less repetition in the view? 

```
def rtl_language?(language)
  %w(ar he).include?(language.short)
end
```

That way, the views can do `rtl_language?(admin_post.language)`

PR #3745, opened 2020-02-03
AO3-5863 Add new Devise info to GDPR script

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5863

## Purpose

When we updated to Devise, we started auditing IP addresses for some new things, or the names of the things we were already auditing changed. This updates the GDPR user data script in light of that.

- Add encrypted_password case to handle IPs recorded when someone changes their password.
- Add failed_attempts case to handle recorded IPs from failed login attempts, which are currently only logged in certain cases due to the way Devise updates the count of failed attempts.
- Add locked_at case to handle recorded IPs from the last failed login attempt before an account is locked due to too many login attempts.
- Clarify that the recently_reset case is from before the Devise change, and add a reset_password_sent_at case to handle IPs recorded for password reset requests now that we're using Devise.
- Add sign_in_count case to handle IPs recorded when someone successfully signs in to an account.

## Testing Instructions

Refer to Jira.

## References

Includes #3744

### DIFF
--- a/script/get_user_data.rb
+++ b/script/get_user_data.rb
@@ -66,9 +66,11 @@ u.works.pluck(:ip_address)&.map { |ip| ips << ip unless ip.blank? }
 user_agents = []
 u.comments.pluck(:user_agent)&.map { |ua| user_agents << ua unless ua.blank? }
 
+
 # Lists of IP addresses and previous email addresses and usernames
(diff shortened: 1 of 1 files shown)


redsummernight:
Reminder to merge #3744 first.

PR #3746, opened 2020-02-03
AO3-5851 Add rake task to fill in user_id for existing kudos

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5851

## Purpose

What it says on the tin. 

## Testing Instructions

Tests should pass, and a database admin should run this and make sure the output looks good and the kudos are indeed updated.

### DIFF
--- a/lib/tasks/after_tasks.rake
+++ b/lib/tasks/after_tasks.rake
@@ -631,6 +631,24 @@ namespace :After do
     end
     puts && STDOUT.flush
   end
+
+  desc "Add user_id to existing logged in kudos"
+  task(add_user_id_to_kudos: :environment) do
+    kudos = Kudo.where("pseud_id IS NOT NULL")
+    kudos_count = kudos.count
+    estimated_batches = kudos_count / 1000
+    puts "Total number of kudos to check: #{kudos_count}"
+
+    kudos.find_in_batches.with_index do |batch, index|
+      batch_number = index + 1
+      batch.each do |kudos|
+        kudos.user_id = kudos.pseud&.user_id
+        kudos.save if kudos.changed?
+      end
+      print("Batch #{batch_number} of about #{estimated_batches} complete\n") && STDOUT.flush
+    end
+    puts && STDOUT.flush
+  end
 end # this is the end that you have to put new tasks above
 
 ##################
--- a/lib/tasks/after_tasks.rake
+++ b/lib/tasks/after_tasks.rake
@@ -638,12 +638,11 @@ namespace :After do
     total_batches = (total_users + 999) / 1000
     puts "Updating #{total_users} users' kudos in #{total_batches} batches"
 
-    User.find_in_batches.with_index do |batch, index|
+    User.includes(:pseuds).find_in_batches.with_index do |batch, index|
       batch_number = index + 1
       progress_msg = "Batch #{batch_number} of #{total_batches} complete"
       batch.each do |user|
-        Kudo.where(pseud_id: user.pseud_ids)
-          .where(user_id: nil)
+        Kudo.where(pseud_id: user.pseud_ids, user_id: nil)
           .update_all(user_id: user.id)
       end
       puts(progress_msg) && STDOUT.flush
--- a/lib/tasks/after_tasks.rake
+++ b/lib/tasks/after_tasks.rake
@@ -632,22 +632,19 @@ namespace :After do
     puts && STDOUT.flush
   end
 
-  desc "Add user_id to existing logged in kudos without user_ids"
+  desc "Update each user's kudos with the user's id"
   task(add_user_id_to_kudos: :environment) do
-    kudos_to_update = Kudo.where("pseud_id IS NOT NULL AND user_id IS NULL")
-    kudos_count = kudos_to_update.count
-    total_batches = kudos_count / 1000
-    puts "Total number of kudos to check: #{kudos_count}"
+    total_users = User.all.size
+    total_batches = (total_users + 999) / 1000
+    puts "Updating #{total_users} users' kudos in #{total_batches} batches"
 
-    kudos_to_update.find_in_batches.with_index do |batch, index|
+    User.find_in_batches.with_index do |batch, index|
       batch_number = index + 1
-      progress_msg = "Batch #{batch_number} of about #{total_batches} complete"
-      batch.each do |kudos|
-        # Skip unless the pseud and user exist.
-        next unless kudos&.pseud&.user
-
-        kudos.user_id = kudos.pseud.user_id
-        kudos.save if kudos.changed?
+      progress_msg = "Batch #{batch_number} of #{total_batches} complete"
+      batch.each do |user|
+        Kudo.where(pseud_id: user.pseud_ids)
+          .where(user_id: nil)
+          .update_all(user_id: user.id)
       end
       puts(progress_msg) && STDOUT.flush
     end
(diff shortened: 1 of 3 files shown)


spec/miscellaneous/lib/tasks/after_tasks.rake_spec.rb
```diff
@@ -97,3 +97,61 @@
     end
   end
 end
+
+describe "rake After:add_user_id_to_kudos" do
+  context "when the kudos has no pseud_id" do
+    it "doesn't add user_id if the kudos has an ip_address" do
+      guest_kudos = create(:kudo, ip_address: "0.00.0.000")
```
zz9pzza: If we are making changes could we change the ip address to 127.0.0.2  or at least something valid, its not important but its going to annoy me to have this string.

lib/tasks/after_tasks.rake
```diff
@@ -631,6 +631,28 @@ namespace :After do
     end
     puts && STDOUT.flush
   end
+
+  desc "Add user_id to existing logged in kudos without user_ids"
+  task(add_user_id_to_kudos: :environment) do
+    kudos_to_update = Kudo.where("pseud_id IS NOT NULL AND user_id IS NULL")
+    kudos_count = kudos_to_update.count
+    total_batches = kudos_count / 1000
```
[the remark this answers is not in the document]
sarken: It does, yeah. The math to correct it just seemed too hacky to be worth satisfying my own annoyance with it, so I went with saying "about X batches" instead. Since it's annoying other people, though, I can change it!

lib/tasks/after_tasks.rake
```diff
@@ -631,6 +631,28 @@ namespace :After do
...
+    puts "Total number of kudos to check: #{kudos_count}"
+
+    kudos_to_update.find_in_batches.with_index do |batch, index|
```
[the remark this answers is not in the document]
sarken: lol, that was my initial thought in the wee hours of Sunday morning:

> sarken  05:45
> I was thinking about https://otwarchive.atlassian.net/browse/AO3-5851, and since we only need/want to update the user_id column for users and pseuds that exist... would it make more sense to loop through all of our _users_, get their kudos (user.kudos does the trick), and then update -- no, never mind, as I'm saying this I'm realizing that would just mean going through the entire kudos table once per user

But kudos are indexed on pseuds, so maybe that would be better?

spec/miscellaneous/lib/tasks/after_tasks.rake_spec.rb
sarken: We don't validate IP addresses, so it shouldn't matter, but sure! I'll fix the others, too.

lib/tasks/after_tasks.rake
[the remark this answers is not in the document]
tickinginstant: You could do `Kudo.where(pseud_id: user.pseud_ids).update_all(user_id: user.id)` or some such.

(I think the only user where that query might be unreasonable for is `orphan_account`, but I don't think the `orphan_account` can even leave kudos, so you can probably skip it if that's a concern.)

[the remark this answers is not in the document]
sarken: I've checked and there are no kudos from `orphan_account`, so that shouldn't be a problem.

lib/tasks/after_tasks.rake
```diff
@@ -631,6 +631,28 @@ namespace :After do
...
+      batch_number = index + 1
+      progress_msg = "Batch #{batch_number} of about #{total_batches} complete"
+      batch.each do |kudos|
+        # Skip unless the pseud and user exist.
+        next unless kudos&.pseud&.user
+
+        kudos.user_id = kudos.pseud.user_id
+        kudos.save if kudos.changed?
```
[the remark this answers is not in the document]
sarken: That's a good point about co-creators. 😬 It would almost be nice to clean those up at the end of this as well, since we're already telling users they may lose kudos as a result of these changes, but there's nothing to stop it from happening in the future, so we'll just 🙈

lib/tasks/after_tasks.rake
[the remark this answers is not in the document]
sarken: Hmm. The code for adding the `user_id` when new kudos are created will be deployed before this. Should we be selecting _all_ of the user's kudos, or just the ones that don't have a `user_id`? I feel like skipping the ones with the `user_id` is probably not worth the trouble, considering the new code will only have been live for a couple weeks when this is run.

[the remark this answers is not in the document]
tickinginstant: It's not too hard to add `.where(user_id: nil)` to whatever the kudos query is. But I'm guessing it wouldn't affect the speed too much either way, so maybe it's just down to preference? IDK.

spec/miscellaneous/lib/tasks/after_tasks.rake_spec.rb
```diff
@@ -97,3 +97,81 @@
     end
   end
 end
+
+describe "rake After:add_user_id_to_kudos" do
```
sarken: I hate these tests. 😭 The old ones were neater, but didn't make much sense with the new approach. I will gladly accept advice on how to rewrite these from the ground up.

redsummernight: How so? The old tests should still pass, however the rake chooses to do things.

sarken: They did still pass, but since we're going through the users table rather directly working on the kudos table now, it didn't make sense to me to check that guest kudos were going untouched... but it did seem to be important to test a case where the user had multiple pseuds with kudos, which wasn't a consideration before.

lib/tasks/after_tasks.rake
```diff
@@ -631,6 +631,25 @@ namespace :After do
...
     puts && STDOUT.flush
   end
+
+  desc "Update each user's kudos with the user's id"
+  task(add_user_id_to_kudos: :environment) do
+    total_users = User.all.size
+    total_batches = (total_users + 999) / 1000
+    puts "Updating #{total_users} users' kudos in #{total_batches} batches"
+
+    User.find_in_batches.with_index do |batch, index|
```
redsummernight: You can do `User.includes(:pseuds).find_in_batches` to batch pseud lookups and avoid n+1 queries.

lib/tasks/after_tasks.rake
```diff
@@ -631,6 +631,25 @@ namespace :After do
...
+      batch_number = index + 1
+      progress_msg = "Batch #{batch_number} of #{total_batches} complete"
+      batch.each do |user|
+        Kudo.where(pseud_id: user.pseud_ids)
```
redsummernight: You can specify both conditions in one where: `.where(pseud_id: user.pseud_ids, user_id: nil)`. Same SQL queries but shorter code.

spec/miscellaneous/lib/tasks/after_tasks.rake_spec.rb
```diff
@@ -97,3 +97,81 @@
...
+
+      pseudless_user.default_pseud.delete
+      pseudless_user.reload
+      deleted_pseud_kudos.reload
+
+      subject.invoke
+
+      deleted_pseud_kudos.reload
+      expect(deleted_pseud_kudos.user_id).to be(nil)
+      expect(deleted_pseud_kudos.pseud_id).not_to be(nil)
```
redsummernight: We should save the pseud ID before deleting the pseud, then compare to the saved ID here.

spec/miscellaneous/lib/tasks/after_tasks.rake_spec.rb
redsummernight: :+1:

PR #3747, opened 2020-02-08

adt-chair:


sarken:
This is just in case I break something -- I'll close when I'm done.

sarken:
Translation staffer has confirmed existing translations are still present, so I'm closing this.

PR #3748, opened 2020-02-09
AO3-5859 Change all kudos displays to use users instead of pseuds

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5859

## Purpose

- Update the list of kudos on a work, either the JS-loaded fragment or the full page. Expire caches for both.
- Update the names used in the kudos mailer.
- Keep saving/deleting pseud IDs on kudos for now (however, stop editing pseud IDs on kudos when changing default pseuds).

## Testing Instructions

See issue.


app/controllers/chapters_controller.rb
```diff
@@ -57,7 +57,7 @@ def show
           @work.anonymous? ? ts("Anonymous") : @work.pseuds.sort.collect(&:byline).join(', '),
           @work.title + " - Chapter " + @chapter.position.to_s)
 
-      @kudos = @work.kudos.with_pseud.includes(pseud: :user).order("created_at DESC")
+      @kudos = @work.kudos.with_user.includes(:user).order("created_at DESC")
```
redsummernight: @tickinginstant suggested `order by id DESC` may be better.

sarken:
Just checking -- is this still a WIP?

redsummernight:
It's a WIP, yeah. It's blocked by [AO3-5868](https://otwarchive.atlassian.net/browse/AO3-5868) and #3750.

redsummernight:
This can be reviewed now.

app/controllers/chapters_controller.rb
redsummernight: We'll leave the order clause as is for now.

tickinginstant:
I think this may also need to change the keys used to cache `guest_kudos_count` and `all_kudos_count`, to make sure that the cached counts are flushed:
https://github.com/otwcode/otwarchive/blob/fc785a103b3db76c1fdc18166ca50ef1b132d9cf/app/models/work.rb#L943-L953
(This change can't be performed in the same deploy as the migration itself, because the cached counts need to be cleared after the migration has been run. So I think this is the right place for it.)

zz9pzza:
I am planning on "just" flushing the cache when we are done.

redsummernight:
I can change the keys still.

tickinginstant:
If the plan is to flush the entire cache after the migration is run, changing the keys should be unnecessary. But I suppose it can't hurt, and it might help if we end up taking a different approach to handling the cache.

app/models/kudo.rb
```diff
@@ -53,22 +55,21 @@ def dup?
...
-        errors.add(:cannot_be_author,
-                   ts("^You can't leave kudos on your own work."))
-      end
+    return unless user
+
+    commentable = nil
+    if commentable_type == "Work"
+      commentable = Work.find_by(id: commentable_id)
+    elsif commentable_type == "Chapter"
+      commentable = Chapter.find_by(id: commentable_id).work
```
tickinginstant: Are there any kudos on chapters?

(I don't think this is in the scope of this PR, because the goal as I see it is to keep the changes minimal aside from the switch from pseuds to users. But it might be nice to have an issue to simplify this code in the future.)

app/views/kudos/show.html.erb
```diff
@@ -2,4 +2,7 @@
   <%= ts("Kudos on") %> <%= link_to @kudo.commentable.title, @referrer %>
 </h3>
 
-<%= render "kudos/kudos", :kudos => @kudo.commentable.kudos.with_pseud.order("created_at DESC"), :guest_kudos_count => @kudo.commentable.guest_kudos_count, :commentable => @kudo.commentable %>
+<%= render "kudos/kudos",
+  kudos: @kudo.commentable.kudos.with_user.includes(:user).by_date,
+  guest_kudos_count: @kudo.commentable.guest_kudos_count,
+  commentable: @kudo.commentable %>
```
tickinginstant: Also not relevant to this PR, but ... is this view file actually used anywhere? The `KudosController` doesn't seem to have a `show` action defined, and it doesn't seem to be included in `config/routes.rb`, either.

zz9pzza:
Changing the key is the correct thing to do :)

app/models/kudo.rb
redsummernight: We have a small number of kudos (100+) on chapters from a bygone era, so this code will have to stay for now.

We can reopen/update https://otwarchive.atlassian.net/browse/AO3-2197 for cleaning up chapter kudos (adding a rake task) and removing chapter-related validations. We should also restrict what types can be used in the polymorphic association, like we did for subscriptions:

https://github.com/otwcode/otwarchive/blob/6f987a7665d309c6f4e749a34787bcab632d864b/app/models/subscription.rb#L11-L15

What do you think?

app/views/kudos/show.html.erb
redsummernight: Looks unused, I'm removing it here and now.

app/models/kudo.rb
tickinginstant: Those all sound like good things to do! Would the plan be to delete the chapter kudos, or transfer them to the works?

redsummernight: I think we can try transfering to works, and if duplicate errors happen, deleting them.

PR #3749, opened 2020-02-09
AO3-5316 Fix flash messages in UsersController#activate.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5316

## Purpose

- Change the message when activating a user account from "Signup complete! Please log in." to "Account activation complete! Please log in."
- Change the error message when activating an account that has already been activated to use `flash` instead of `flash.now`, so that the error message will show up after redirect.
- Change the other two errors in the `UsersController#activate` to redirect to `root_path` instead of a blank string. This doesn't actually change where it redirects to, but it makes the code a little clearer.
- Add a few new specs to increase coverage of the `activate` action.

## Testing Instructions

See JIRA.

(Though I'm not sure whether the nginx changes have been applied to staging. If they haven't, that may need to be addressed before any of these flash messages can appear.)

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -110,7 +110,7 @@ class UsersController < ApplicationController
   def activate
     if params[:id].blank?
       flash[:error] = ts('Your activation key is missing.')
-      redirect_to ''
+      redirect_to root_path
 
       return
     end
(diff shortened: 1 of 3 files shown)


sarken:
For future reference, I've put "Priority: Critical - Hot Fixed" on this because it's probably the closest thing we have to describe the current state of this (the display of the initial confirmation message is fixed on production but not staging, although it was an nginx change, not a code change).

PR #3752, opened 2020-02-13
AO3-4139 fix hard to read text on required fields in reversi

niconicosette:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4139

## Purpose

the notes on required fields are hard to read - this PR changes the text colour to white so that it can be read more easily

## Testing Instructions

set your site skin to reversi, then try creating a new work or a new collection, and check that the notes on required fields have white text


sarken:
We don't need Travis to run on this, so I'm merging. Thanks!

PR #3760, opened 2020-02-29
AO3-5868 Change kudo uniqueness validation to use users instead of pseuds

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5868

## Purpose

When we add the uniqueness constraint at the database level (#3750), we'll have the error handling ready at the application level (people will see "left kudos here :)" as usual).

For the rare cases where users can get past the Rails validation and run into the (future) database constraint, we should catch `ActiveRecord::RecordNotUnique` exceptions, similar to what we do on user creations:

https://github.com/otwcode/otwarchive/blob/57e7cab589abc7cc666ddf4d3abe768d801b3493/app/controllers/users_controller.rb#L9-L24

## Testing Instructions

See issue.

### DIFF
--- a/app/controllers/kudos_controller.rb
+++ b/app/controllers/kudos_controller.rb
@@ -63,8 +63,20 @@ class KudosController < ApplicationController
     # by database unique indices, use the usual duplicate error message.
     #
     # https://api.rubyonrails.org/v5.1/classes/ActiveRecord/Validations/ClassMethods.html#method-i-validates_uniqueness_of-label-Concurrency+and+integrity
-    flash[:comment_error] = ts("You have already left kudos here. :)")
-    redirect_to request.referer
+    respond_to do |format|
+      format.html do
+        flash[:comment_error] = ts("You have already left kudos here. :)")
+        redirect_to request.referer
+      end
+
+      format.js do
+        # TODO: AO3-5635 Clean up kudos error handling in JS.
+        # The JS error handler only checks for the existence of keys,
+        # e.g. "ip_address" will show the "already left kudos" message.
+        errors = { ip_address: "ERROR" }
+        render json: { errors: errors }, status: :unprocessable_entity
+      end
+    end
   end
 
   private
--- a/app/controllers/kudos_controller.rb
+++ b/app/controllers/kudos_controller.rb
@@ -70,7 +70,6 @@ class KudosController < ApplicationController
       end
 
       format.js do
-        # TODO: AO3-5635 Clean up kudos error handling in JS.
         # The JS error handler only checks for the existence of keys,
         # e.g. "ip_address" will show the "already left kudos" message.
         errors = { ip_address: "ERROR" }
--- a/app/controllers/kudos_controller.rb
+++ b/app/controllers/kudos_controller.rb
@@ -57,6 +57,20 @@ class KudosController < ApplicationController
         end
       end
     end
+  rescue ActiveRecord::RecordNotUnique => e
+    # Uniqueness checks at application level (Rails validations) are inherently
+    # prone to race conditions. If we pass Rails validations but get rejected
+    # by database unique indices, use the usual duplicate error message.
+    #
+    # https://api.rubyonrails.org/v5.1/classes/ActiveRecord/Validations/ClassMethods.html#method-i-validates_uniqueness_of-label-Concurrency+and+integrity
+    if e.message =~ /Mysql2?::Error: Duplicate entry/i
+      flash[:comment_error] = ts("You have already left kudos here. :)")
+      redirect_to request.referer
+      return
+    end
+
+    # If we get unknown errors, let them propagate normally.
+    raise
   end
 
   private
--- a/app/controllers/kudos_controller.rb
+++ b/app/controllers/kudos_controller.rb
@@ -57,20 +57,14 @@ class KudosController < ApplicationController
         end
       end
     end
-  rescue ActiveRecord::RecordNotUnique => e
+  rescue ActiveRecord::RecordNotUnique
     # Uniqueness checks at application level (Rails validations) are inherently
     # prone to race conditions. If we pass Rails validations but get rejected
     # by database unique indices, use the usual duplicate error message.
     #
     # https://api.rubyonrails.org/v5.1/classes/ActiveRecord/Validations/ClassMethods.html#method-i-validates_uniqueness_of-label-Concurrency+and+integrity
-    if e.message =~ /Mysql2?::Error: Duplicate entry/i
-      flash[:comment_error] = ts("You have already left kudos here. :)")
-      redirect_to request.referer
-      return
-    end
-
-    # If we get unknown errors, let them propagate normally.
-    raise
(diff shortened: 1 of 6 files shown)


app/controllers/kudos_controller.rb
```diff
@@ -57,6 +57,14 @@ def create
         end
       end
     end
+  rescue ActiveRecord::RecordNotUnique
```
tickinginstant: I like keeping the `rescue` here rather than having it apply to all actions like it does in the `UsersController`. 👍

app/controllers/kudos_controller.rb
```diff
@@ -57,6 +57,14 @@ def create
...
+    # Uniqueness checks at application level (Rails validations) are inherently
+    # prone to race conditions. If we pass Rails validations but get rejected
+    # by database unique indices, use the usual duplicate error message.
+    #
+    # https://api.rubyonrails.org/v5.1/classes/ActiveRecord/Validations/ClassMethods.html#method-i-validates_uniqueness_of-label-Concurrency+and+integrity
+    flash[:comment_error] = ts("You have already left kudos here. :)")
+    redirect_to request.referer
```
tickinginstant: I'm not 100% certain, but I think not using a `format` block with separate code for `js` means that both the error and the redirect will get eaten when the user has javascript enabled (which is probably most of the time).

On the other hand, I think this error is most likely to occur if the user double-submits the form, in which case the javascript response returned by the first form submission would update the page, and the second error message would be unnecessary. So it might be good that the error and redirect will be eaten? IDK.

redsummernight: Thanks! The redirect will be lost, I forgot to check for `format: :js`. This should be lawful-good-fixed now.

app/controllers/kudos_controller.rb
redsummernight: Yeah same. Now I'll look for a reason to edit the user controller and smuggle in a similar change.

redsummernight:
I think 6 new TODOs in a PR is a personal best.

app/models/kudo.rb
```diff
@@ -9,15 +9,15 @@ class Kudo < ApplicationRecord
...
-    scope: [:commentable_id, :commentable_type],
-    message: ts("^You have already left kudos here. :)"),
-    if: Proc.new { |kudo| !kudo.pseud.nil? }
-
   validates_uniqueness_of :ip_address,
-    scope: [:commentable_id, :commentable_type],
-    message: ts("^You have already left kudos here. :)"),
-    if: Proc.new { |kudo| !kudo.ip_address.blank? }
+                          scope: [:commentable_id, :commentable_type],
+                          message: ts("^You have already left kudos here. :)"),
```
redsummernight: The carets did work once upon a time:

https://github.com/otwcode/otwarchive/blob/2352531b7e024ac77ecd29ceb832620752fc3646/app/controllers/kudos_controller.rb#L12

To be cleaned up in [AO3-5635](https://otwarchive.atlassian.net/browse/AO3-5635).

app/controllers/kudos_controller.rb
tickinginstant: [AO3-5677](https://otwarchive.atlassian.net/browse/AO3-5677) seems like a good choice for that, if you're looking for an excuse to fix it.

app/controllers/kudos_controller.rb
```diff
@@ -57,6 +57,26 @@ def create
...
+    #
+    # https://api.rubyonrails.org/v5.1/classes/ActiveRecord/Validations/ClassMethods.html#method-i-validates_uniqueness_of-label-Concurrency+and+integrity
+    respond_to do |format|
+      format.html do
+        flash[:comment_error] = ts("You have already left kudos here. :)")
+        redirect_to request.referer
+      end
+
+      format.js do
+        # TODO: AO3-5635 Clean up kudos error handling in JS.
```
tickinginstant: It looks like [AO3-5635](https://otwarchive.atlassian.net/browse/AO3-5635) is specifically for the error handling when javascript is disabled, but this is the code for when javascript is enabled. Using the keys of the hash and hard-coding the messages in JS isn't very i18n-compatible, but other than that I think the error messages have the correct behavior. Should there be a new issue for kudos error handling when JS is enabled?

public/javascripts/application.js
```diff
@@ -455,7 +455,8 @@ $j(document).ready(function() {
         var msg = 'Sorry, we were unable to save your kudos';
         var data = $j.parseJSON(jqXHR.responseText);
 
-        if (data.errors && (data.errors.pseud_id || data.errors.ip_address)) {
+        // TODO: AO3-5859 Remove data.errors.pseud_id.
+        if (data.errors && (data.errors.pseud_id || data.errors.ip_address || data.errors.user_id)) {
```
tickinginstant: Does the check for `data.errors.pseud_id` have to be kept until [AO3-5859](https://otwarchive.atlassian.net/browse/AO3-5859)? I think this is the PR that changes the error keys.

app/controllers/kudos_controller.rb
redsummernight: If we fix 5635 using [Rails 6's `i18n_customize_full_message`](https://blog.bigbinary.com/2019/04/22/rails-6-allows-to-override-the-activemodel-errors-full_message-format-at-the-model-level-and-at-the-attribute-level.html) we can remove all the hard-coded error messages in this controller and in JS.

I'll make a new issue for the refactoring if someone fixes 5635 before Rails 6 though.

public/javascripts/application.js
redsummernight: Yeah, I keep it because while deploying we'll have a mix of old and new versions.

app/controllers/kudos_controller.rb
tickinginstant: Maybe the issue for 5635 should be rewritten to reflect the fact that it should also change the javascript code, as well? Or this comment should be removed?

redsummernight: I can remove the 5635 comments.

public/javascripts/application.js
tickinginstant: Isn't caching going to break the error messages during deploy regardless? People who've already opened the page before the deploy and press the kudos button after the deploy will get a nondescript "Sorry, we were unable to save your kudos" message instead of the specific "You have already left kudos here! :)" message.

redsummernight: Good point, it still breaks in that case. I was thinking of the case where people load the JS file of the new version, but their kudos request gets routed to a server running the old version.

tickinginstant: I'm just not sure it's worth the effort of yet more incremental changes when it doesn't break anything important. It seems cleaner just to make the change and accept that some of the error messages will be less specific during the deploy, rather than adding yet more things to clean up later.

redsummernight: Yeah, I can remove it.

redsummernight:
Thanks!

sarken:
Marked approved based on ticking's review!

PR #3764, opened 2020-03-15
Ao3 5888

zz9pzza:
Shifting the sha1 of head.
This can be reverted in the next release.


sarken:
I've updated the description and title of this pull request, but please be more mindful of getting the issue ID correct and including all the details. It's always important, but especially in emergency situations where you're self-merging: the rest of us need to be able to see what was done and why.

PR #3775, opened 2020-03-26
AO3-4874 Allow HTML subscription emails to include HTML in summaries

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4874

## Purpose

~Guess who is currently subscribed to two works with paragraphs in the summary and currently being driven 'round the bend.~

Allows the HTML subscription email to include HTML in work and chapter summaries. The issue just says `<br>` and `<p>`, but I looked at [comment emails](https://github.com/otwcode/otwarchive/blob/51c8fae1a12677b7a2433bc9346611404458d9b6/app/views/comment_mailer/comment_notification.html.erb#L18), which also have user-entered text inside a `<blockquote>` and they aren't picky about tags. I can't imagine why we'd want to be more picky for work or chapter summaries. 

I've removed the paragraph tags around the summary block quotes here because that was inconsistent with the comment emails, and it doesn't make sense to be nesting paragraphs inside paragraphs (maybe that's why we stripped all the HTML? who knows!). Then I put them around the "Summary:" and "Chapter Summary:" text so it's not just chilling outside a block element. It does not have negative effects on the appearance.

HTML email showing work and chapter summaries with HTML.

## Testing Instructions

Refer to Jira.


sarken:
Oops, gotta also do it for chapters. Hold, please.

features/other_b/subscriptions.feature
```diff
@@ -259,3 +259,26 @@
     And I go to my subscriptions page
     And I press "Unsubscribe from I am <strong>er Than Yesterday & Other Lies"
   Then I should see "You have successfully unsubscribed from I am <strong>er Than Yesterday & Other Lies"
+
+  Scenario: Subscription emails include HTML in work and chapter summaries in the HTML email and convert it for plain text emails.
```
redsummernight: I'd prefer this in user_mailer_spec.rb. We don't need to add new cuke steps in that case.

sarken: Unfortunately, I did it this way because I could not figure out how to get subscription emails working in specs. /o\

sarken:
Closing so I can get under 5 pull requests and open a new one for an issue that's blocking two of my existing PRs.

spec/mailers/user_mailer_spec.rb
```diff
@@ -381,4 +381,55 @@
       end
     end
   end
+
+  describe "batch_subscription_notification" do
+    let(:work) { create(:posted_work, summary: "<p>Paragraph <u>one</u>.</p><p>Paragraph 2.</p>") }
```
redsummernight, line 386: Tests are failing because of the renamed factory:
```
     # --- Caused by: ---
     # KeyError:
     #   key not found: "posted_work"
     #   ./spec/mailers/user_mailer_spec.rb:386:in `block (3 levels) in <top (required)>'
```

sarken: Thanks!

Looking at the next line in this test, I wonder if we should do the same thing with chapters -- that is, make them posted by default.

redsummernight: We should ([AO3-5968](https://otwarchive.atlassian.net/browse/AO3-5968)).

sarken: ❤️ ❤️ ❤️

PR #3777, opened 2020-03-27
AO3-5757 Update Cucumber to 3.1 and retry feature tests

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5757

## Purpose

Update Cucumber to 3.1, which gives us 2 nice things:

- [Cucumber expressions](https://cucumber.io/docs/cucumber/cucumber-expressions/), which is more readable than regular expressions.
- The ability to retry failed tests, using `--retry 1`. Retried tests are marked "flaky", and flaky tests will fail the build in strict mode (`--strict`), which is why we also need to add `--no-strict-flaky`. **This is the real draw of 3.1**. Note that in development environments, we use `--retry 0` so tests will still fail right away.

Here's [what a passing flaky test looks like on Travis](https://travis-ci.org/github/redsummernight/otwarchive/jobs/667528403): 
```
(::) failed steps (::)

Expected false to be truthy. (Minitest::Assertion)
./features/step_definitions/comment_steps.rb:180:in `/^the comment on "([^\"]*?)" should be marked as unreviewed/'
features/comments_and_kudos/comment_moderation.feature:267:in `Then the comment on "Moderation" should be marked as unreviewed'

72 scenarios (1 flaky, 71 passed)
955 steps (1 failed, 954 passed)
9m16.617s
The command "bundle exec "$TEST_GROUP"" exited with 0.
```

Changes for the upgrade:

- Use tag expressions for tag filtering (`not @wip` instead of `~@wip`).

- Fix 2 prompt cukes. Starting with Cucumber 3.0.0, capture groups with `(\d+)` are automatically converted to integers. This break assertions that compare strings. A simple `.to_s` would have worked, but I rewrote the affected steps to use [matchers that wait for AJAX](https://thoughtbot.com/blog/write-reliable-asynchronous-integration-tests-with-capybara#checking-a-field39s-value) and Cucumber Expressions.

- Remove the auto-generated notice in web_steps.rb, since cucumber-rails [stopped generating it in 2.0.0](https://github.com/cucumber/cucumber-rails/blob/f2b77091a614874bea21fc2a0a29913f7882bc96/CHANGELOG.md#v200-2019-09-013).

- Reset the files auto-generated by Cucumber's setup task ([`rails generate cucumber:install`](https://github.com/cucumber/cucumber-rails#upgrading-from-a-previous-version)). Affected files include cucumber.rake, cucumber.yml, and env.rb.

- Update env.rb: we modified this file a lot over time, so when it's reset, I sorted through what we need to keep: Before/After hooks moved to hooks.rb, Capybara settings moved to capybara.rb. The coverage setup stays in env.rb, because it needs to happen before everything else.

- Simplify DatabaseCleaner settings (a.k.a stick with the defaults in env.rb).
    - Non-JS tests use `:transaction`, this hasn't changed.
    - JS tests (`Cucumber::Rails::Database.javascript_strategy`) previously used `:transaction` for no documented reason. Switch to `:truncation` as recommended by cucumber-rails.
    - Remove `@no-txn` tags from cukes: this tag doesn't have any effect, even before the migration.

## Testing Instructions

Travis/Codeship should pass.

Before merging, [Codeship "Test Pipelines"](https://github.com/otwcode/otwarchive/wiki/Codeship) should be updated from:
```sh
export CUCUMBER_FORMAT=Ao3Cucumber::Formatter
```
to:
```sh
export CUCUMBER_FORMAT=Ao3Cucumber::Formatter
export CUCUMBER_RETRY=1
```
This is not as needed for Codeship as it is for Travis, because Codeship already has a retry mechanism (`./script/codeship/try.sh`). We can still save some time with the new option though: `./script/codeship/try.sh` will retry an entire set of tests, but `CUCUMBER_RETRY=1` will retry only the specific tests that failed.


sarken:
Wiki and Codeship updated

PR #3778, opened 2020-03-30
AO3-5162 Stop preloading name and email on guest comment forms

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5162

## Purpose

So we can cache individual work and chapter pages for logged-out users.

We still need jquery.cookie for the TOS prompts, however.

## Testing Instructions

See issue.


redsummernight:
[The Travis build for this PR](https://travis-ci.org/github/otwcode/otwarchive/builds/669002948), which is not linked here for whatever reason.

PR #3779, opened 2020-03-31
AO3-5497 Don't show TOS prompt if URL has ?tos=yes params

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5497

## Purpose

Doesn't add the TOS prompt if the URL has ?tos=yes at the end. (We don't add these parameters to the URL automatically after you press "I agree/consent to its terms" because we don't want people accidentally sharing links with them. We want it to be something a user has to manually add so we can be relatively sure they actually agreed to the TOS and didn't just follow a link that happened to have the ?tos=yes params.)

Needs nginx changes to remember the parameters.

## Testing Instructions

Refer to Jira.

### DIFF
--- a/app/views/layouts/_javascripts.html.erb
+++ b/app/views/layouts/_javascripts.html.erb
@@ -50,8 +50,7 @@
     params[:controller] == "home" && params[:action] == "tos_faq" ||
     params[:controller] == "home" && params[:action] == "dmca" ||
     params[:controller] == "abuse_reports" && params[:action] == "new" ||
-    params[:controller] == "feedbacks" && params[:action] == "new" ||
-    params[:tos] == "yes" %>
+    params[:controller] == "feedbacks" && params[:action] == "new" %>
   <%= javascript_tag do %>
     $j(document).ready(function() {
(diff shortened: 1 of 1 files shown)


sarken:
Hmm. I'm having a second thought about this now. This is the app code saying don't include the JavaScript to add the TOS prompt if the parameters are present... should it be the JavaScript saying don't add the prompt if these parameters are present? Or are we cool serving a different version of the page depending on the parameters?

sarken:
red confirmed we do indeed need to use JavaScript. 

[This approach](https://stackoverflow.com/questions/19491336/get-url-parameter-jquery-or-how-to-get-query-string-values-in-js) is ugly, but it's compatible with more browsers than newer approaches, so here we are. 

I'm not checking the params for logged in users because if the TOS prompt is still showing up despite their preference being set, something is very wrong.

zz9pzza:
"Or are we cool serving a different version of the page depending on the parameters?" 

From a caching point of view only I am happy.

sarken:
Closing because we want to take a whole different approach!

PR #3781, opened 2020-04-06
AO3-5497 Let guests bypass the TOS prompt (GET param & localStorage)

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5497

## Purpose

With JS enabled:

- Logged in users must accept the prompt and cannot use "?tos=yes".
- Guests can accept the prompt and localStorage will remember their choice.
- Guests can use "?tos=yes" to accept the prompt without seeing it.

"?tos=yes" should work fine with nginx caching, since we include the query string in the cache key.

I think we could have had the nginx cache be [cookie aware](https://dev.to/ale_ukr/how-to-make-nginx-cache-cookie-aware-2ffl) and stayed with the old cookie approach, but since we're not showing the prompt to no-JS guests anyway, using localStorage will save nginx from having to cache the same requests with and without the TOS cookie.

## Testing Instructions

See issue.


redsummernight:
PR Travis build succeeded: https://travis-ci.org/github/otwcode/otwarchive/builds/671444211.

PR #3782, opened 2020-04-06
AO3-5906 Move adult content warning acceptance to separate cookie

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5906

## Purpose

Creates a new session cookie for users who have accepted the adult content warning. We can use this cookie to determine which cached page nginx should give a user.

Note: We had some trouble with the session cookie not being sticky enough for logins, which is why even a non-"Remember me" log in lasts two weeks. We might have to set the expiry on this to 2 weeks as well.

## Testing

Refer to Jira.


redsummernight:
This test should be updated:

https://github.com/otwcode/otwarchive/blob/eea20003da26c7cae66c2377a58816e20645f010/spec/controllers/chapters_controller_spec.rb#L103-L106

PR #3785, opened 2020-04-06
AO3-5878 Add error to invalid prompt update

teyla86:
## Issue
https://otwarchive.atlassian.net/browse/AO3-5878

## Purpose
After signing up for a challenge or exchange, users can edit their requests and offers in the collection by clicking on "My Sign-up"/"My Prompts" in the sidebar. Each request/offer/prompt has an "Edit Prompt"-button, which leads to a form where the specific prompt can be edited.

Currently, if the user changes the prompt so it becomes invalid (i.e. doesn't add the required number of character tags, or similar), the page simply reloads without returning an error message or applying the changes.

The committed change adds a detailed error message to a failed prompt update in the edit prompt view using the validation helper. It also adds a check for stray "^" signs into the error formatting of the validation helper to prevent prompt errors being rendered with a "^"  at the beginning.

## Testing Instructions

**Bug fix test:**

1. Create challenge: Gift Exchange or Prompt Meme
2. Define required challenge attributes: e.g. 2 required characters, or 2 required fandoms ...
3. Sign up for the challenge using correct attributes
4. Visit the collection home page ( .../collections/[your test collection]) and click on "My Sign-up" (Gift Exchange) or "My Prompts" (Prompt Meme) in the sidebar
5. Edit any prompt by using the "Edit Prompt" button
6. Change attributes so they don't comply with the challenge requirements anymore and click "Update"

Expected outcome: page reloads showing a detailed, correctly formatted error message telling you which prompt attributes don't comply with the challenge settings, similar to what happens when you try to edit a prompt on the "Edit Sign-Up" page.

**Regression tests for the validation helper:**

These should not be necessary in all places it's being used (61 views), since the change only adds a search for "^" and removes all found instances. Instances should only be found if the "^" that marks the beginning of the error message has not already been removed (as was the case with the rendering of the error message in the prompt edit view), in which case removal is desired.

For a selective regression test, proceed as described above. In Step 5, use the "Edit Sign-up" instead of the "Edit Prompt" button. Expected outcome: page reloads showing detailed and correctly formatted error messages.

outdated:

> The committed change adds an error flash to the prompts controller to be shown if a user tries to update invalid prompt changes.
> 
> The Jira issue asks for a similarly detailed message as is shown when a user tries to submit an invalid change with the sign-up form ("Edit Sign-up"-button on the "My Sign-up" page). Since the sign-up form uses a different controller than the prompts form, though, this would be quite complicated to implement. I decided to go with a simple error flash which mentions that more details can be gained from using the "Edit Sign-up" function:
> 
> image
> 
> I also added the ts()-element to the successful update prompt for consistency.
> 
> ## Testing Instructions
> 
> 1. Create challenge: Gift Exchange or Prompt Meme
> 2. Define required challenge attributes: e.g. 2 required characters, or 2 required fandoms ...
> 3. Sign up for the challenge using correct attributes
> 4. Visit the collection home page ( .../collections/[your test collection]) and click on "My Sign-up" (Gift Exchange) or "My Prompts" (Prompt Meme) in the sidebar
> 5. Edit any prompt by using the "Edit prompt" button
> 6. Change attributes so they don't comply with the challenge requirements anymore and click "Update"
> 
> Expected outcome: error message (see above) is shown.

## References

n/a

## Credit

teyla

she or they


bingeling:
Hey teyla, thank you for the pull request! We'll get back to you soon on that. :)

app/controllers/prompts_controller.rb
```diff
@@ -145,9 +145,10 @@ def create
 
   def update
     if @prompt.update_attributes(prompt_params)
-      flash[:notice] = 'Prompt was successfully updated.'
+      flash[:notice] = ts("Prompt was successfully updated.")
       redirect_to collection_signup_path(@collection, @challenge_signup)
     else
+      flash[:error] = ts("This change does not comply with sign-up rules, sorry. For more details, try using \"Edit Sign-up\".")
```
redsummernight, line 151: Instead of adding this error message, I think you should be able to display the detailed errors on the prompt view, using this helper:

https://github.com/otwcode/otwarchive/blob/030fa4607d3c59fba2cacbe4e9c99b3c49fbc1cb/app/helpers/validation_helper.rb#L32-L42

For example, in the Edit Sign-up form:

https://github.com/otwcode/otwarchive/blob/030fa4607d3c59fba2cacbe4e9c99b3c49fbc1cb/app/views/challenge_signups/_signup_form.html.erb#L35

teyla86: Thanks for the pointer! I'm new to Rails, so I'm slowly finding my way around. I changed things accordingly and updated the description & testing instructions above.

teyla86:
I can't seem to change the label on this, but it should be "Coder Has Actioned Review" .

app/helpers/validation_helper.rb
```diff
@@ -43,7 +43,7 @@ def error_messages_for(object)
   
   def error_messages_formatted(errors, intro = "")
     return unless errors && !errors.empty?
-    error_messages = errors.map {|msg| content_tag(:li, msg.gsub(/^(.+)\^/, '').html_safe)}.join("\n").html_safe
+    error_messages = errors.map { |msg| content_tag(:li, msg.gsub(/^(.+)\^/, '').gsub(/\^/, '').html_safe) }.join("\n").html_safe
```
redsummernight, line 46: Instead of `msg.gsub(/^(.+)\^/, '').gsub(/\^/, '')`, would `msg.gsub(/^(.*)\^/, '')` work?

----

You can also remove `.gsub(/\^/, '')` from here:

https://github.com/otwcode/otwarchive/blob/030fa4607d3c59fba2cacbe4e9c99b3c49fbc1cb/app/views/challenge_signups/_signup_form.html.erb#L53

app/views/prompts/edit.html.erb
```diff
@@ -9,6 +9,15 @@
 <!--/subnav-->
 
 <!--main content-->
+
+<% unless @prompt.errors.empty? %>
```
redsummernight, line 13: You can leave out this block, since `error_messages_for` already says "We couldn't save this X because".

app/views/prompts/edit.html.erb
```diff
@@ -9,6 +9,15 @@
...
+    <div class="error">
+      <h4 class="heading"><%= ts("There were some problems with this submission. Please correct the mistakes below.") %></h4>
+    </div>
+  <% end %>
+
+<%= error_messages_for @prompt %>
```
redsummernight, line 19: Can you add a test for this? You can create a valid prompt meme sign-up similar to:

https://github.com/otwcode/otwarchive/blob/030fa4607d3c59fba2cacbe4e9c99b3c49fbc1cb/features/prompt_memes_a/challenge_promptmeme_setup.feature#L111-L129

Then have the test edit the prompt, delete some info, and check that the errors show up.

teyla86: Thanks for all your feedback! One question: the challenge_promptmeme_setup feature seems to test for a variety of user actions already, among them "Edit individual prompt via signup show". Is it okay to add an "Update individual prompt via signup" scenario here, or should it be in a separate feature?

app/helpers/validation_helper.rb
teyla86: Done and done, thank you!

app/views/prompts/edit.html.erb
redsummernight: Nice find! I think adding to the existing scenario "Edit individual prompt via signup show" sounds good.

app/views/prompts/edit.html.erb
teyla86: I removed the heading. I also added the "Required information" element to the prompts form to let users know which fields are required, and moved the "Required information" element and the error messages into the actual form tag to get the correct styling.

app/views/prompts/edit.html.erb
teyla86: Done, and I ran it successfully on local.

features/prompt_memes_a/challenge_promptmeme_setup.feature
```diff
@@ -172,7 +172,7 @@ Feature: Prompt Meme Challenge
   Then I should see "Edit"
     And I should see "Delete"
   
-  Scenario: Edit individual prompt via signup show
+  Scenario: Edit individual prompt and get invalid update error
```
redsummernight, line 175: We can name this scenario "Edit individual prompt".

features/prompt_memes_a/challenge_promptmeme_setup.feature
```diff
@@ -181,6 +181,8 @@ Feature: Prompt Meme Challenge
   When I follow "Edit Prompt"
   Then I should see single prompt editing
   And I should see "Edit Sign-up"
+  When I change my Battle 12 signup to invalid combination A2
+  Then I should see errors
```
redsummernight, line 185: I think it's better to be specific what changes we make and what exact errors we get as a result. So instead of adding a new step, you could write:

```gherkin
When I uncheck "Stargate Atlantis"
  And I press "Submit"
Then I should see "<the specific error message>"
```

app/views/prompts/edit.html.erb
redsummernight: Code Climate complained about similarities between 2 controllers that can both edit sign-ups. That's not something we need to deal with here, so I've set it to passing.

features/prompt_memes_a/challenge_promptmeme_setup.feature
teyla86: Done!

features/prompt_memes_a/challenge_promptmeme_setup.feature
teyla86: And done. Thank you for talking me through this.

redsummernight: No problem! Looks good, just one last change, can you add to the check step?
```gherkin
Then I should see "Sorry! We couldn't save this request because:"
  And I should see "Your Request must include exactly 1 fandom tags"
```

teyla86: Done!

PR #3791, opened 2020-04-09
AO3-5915 Write TOS acceptance to cookie when localStorage not available

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5915

## Purpose

Some browsers that support localStorage do weird things like stop you from putting anything in it in private mode. This means visitors using those browsers can't get past the GDPR TOS prompt.

So what we're doing here is testing setting something in localStorage and then, if that doesn't work, repurposing it to create (or update or remove) a cookie.

## Testing Instructions

Refer to Jira.

## References

Refer to code comments.

### DIFF
--- a/app/views/layouts/_javascripts.html.erb
+++ b/app/views/layouts/_javascripts.html.erb
@@ -52,14 +52,14 @@
     params[:controller] == "abuse_reports" && params[:action] == "new" ||
     params[:controller] == "feedbacks" && params[:action] == "new" %>
   <%= javascript_tag do %>
-    <% # localStorage polyfill we only needed for guests. %>
-    <% if !current_user %>
+    <% # localStorage polyfill we only need for guests. %>
+    <% unless current_user %>
       // We can't rely on !window.localStorage to test localStorage support in
       // browsers like Safari 9, which technically support it, but which have a
       // storage length of 0 in private mode.
       // Credit: https://github.com/getgrav/grav-plugin-admin/commit/cfe2188f10c4ca604e03c96f3e21537fda1cdf9a
       function isSupported() {
-          var item = "localStoragePollyfill";
+          var item = "localStoragePolyfill";
           try {
               localStorage.setItem(item, item);
               localStorage.removeItem(item);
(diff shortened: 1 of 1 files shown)


app/views/layouts/_javascripts.html.erb
```diff
@@ -52,6 +52,63 @@
     params[:controller] == "abuse_reports" && params[:action] == "new" ||
     params[:controller] == "feedbacks" && params[:action] == "new" %>
   <%= javascript_tag do %>
+    <% # localStorage polyfill we only needed for guests. %>
+    <% if !current_user %>
```
redsummernight, line 56: Can you use `unless current_user`?

app/views/layouts/_javascripts.html.erb
```diff
@@ -52,6 +52,63 @@
     params[:controller] == "abuse_reports" && params[:action] == "new" ||
     params[:controller] == "feedbacks" && params[:action] == "new" %>
   <%= javascript_tag do %>
+    <% # localStorage polyfill we only needed for guests. %>
```
redsummernight, line 55: "need" over "needed"?

app/views/layouts/_javascripts.html.erb
```diff
@@ -52,6 +52,63 @@
...
+    <% if !current_user %>
+      // We can't rely on !window.localStorage to test localStorage support in
+      // browsers like Safari 9, which technically support it, but which have a
+      // storage length of 0 in private mode.
+      // Credit: https://github.com/getgrav/grav-plugin-admin/commit/cfe2188f10c4ca604e03c96f3e21537fda1cdf9a
+      function isSupported() {
+          var item = "localStoragePollyfill";
```
redsummernight, line 62: It's their code, but this should be "Polyfill".

app/views/layouts/_javascripts.html.erb
```diff
@@ -52,6 +52,63 @@
...
+              return false;
+          }
+      }
+
+      // For browsers that don't support localStorage according to our check,
+      // make its functions handle cookies instead.
+      // Credit: https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Local_storage
+      if (!isSupported()) {
+        // 20 years is the Rails maximum and what we previous used.  
+        var expiry_date = new Date;
```
redsummernight, line 79: Can you use `expiryDate` to match the surrounding JS style?

redsummernight:
I will merge this with `[skip codeship tests]`.

PR #3793, opened 2020-04-09
AO3-5820 Add "Select" label to checkboxes for inbox comments

melinath:
# Pull Request Checklist

^ had some problems with this - see comments.
^ I don't seem to have access to do this, sorry! I must have missed a step somewhere...

## Issue

https://otwarchive.atlassian.net/browse/AO3-5820

## Purpose

This PR adds a "Select" label to the checkbox for bulk-selecting comments in the Inbox, to increase the clickable area and improve accessibility.

## Testing Instructions

1. Log in
2. Find a work and leave a comment on it (or post a work)
3. Log out
4. Reply to the comment you left (or comment on your work)
5. Log in
6. Hi, username! > My Dashboard > Inbox
7. Click on "Select" and verify that the checkbox is selected.

## Credit

Stephen Burrows

he/him

### DIFF
--- a/app/views/inbox/show.html.erb
+++ b/app/views/inbox/show.html.erb
@@ -64,8 +64,11 @@
                 </li>
               <% end %>
 
-              <li class="action">
-                <%= check_box_tag "inbox_comments[]", inbox_comment.id, false, :title => ts("select this comment"), :id => "inbox_comments_#{inbox_comment.id}" %>
+              <li>
+                <%= label_tag "inbox_comments_#{inbox_comment.id}" do %>
+                  <%= check_box_tag "inbox_comments[]", inbox_comment.id, false, :id => "inbox_comments_#{inbox_comment.id}" %>
+                  <%= ts("Select") %>
+                <% end %>
               </li>
             </ul>
           </li>
--- a/app/views/inbox/show.html.erb
+++ b/app/views/inbox/show.html.erb
@@ -66,7 +66,7 @@
 
               <li>
                 <%= label_tag "inbox_comments_#{inbox_comment.id}" do %>
-                  <%= check_box_tag "inbox_comments[]", inbox_comment.id, false, :id => "inbox_comments_#{inbox_comment.id}" %>
+                  <%= check_box_tag "inbox_comments[]", inbox_comment.id, false, id: "inbox_comments_#{inbox_comment.id}" %>
                   <%= ts("Select") %>
                 <% end %>
               </li>
--- a/features/comments_and_kudos/inbox.feature
+++ b/features/comments_and_kudos/inbox.feature
@@ -49,6 +49,21 @@ Feature: Get messages in the inbox
       And I press "Filter"
     Then I should not see "guest on Down for the Count"
 
+  Scenario: A user can select a comment by clicking on 'Select' on the inbox page
+    Given I am logged in as "boxer" with password "10987tko"
+      And I post the work "The Fight"
+      And I set my preferences to turn on messages to my inbox about comments
+    When I am logged in as "cutman"
+      And I post the comment "You should receive this in your inbox." on the work "The Fight"
+    When I am logged in as "boxer" with password "10987tko"
+      And I go to my inbox page
+    Then I should see "cutman on The Fight"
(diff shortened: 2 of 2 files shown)


melinath:
One thing I'm not quite sure about is that the text for the label doesn't seem to align nicely with the checkbox.

**Here's the expected end state from the ticket:**

inbox_select

**Here's the actual end state:**

inbox_select

I'm not sure if the expected end state from the ticket was supposed to be matched exactly or if there's a separate bug impacting the display of this kind of checkbox label. I'm pretty sure I'm matching the way that this kind of checkbox label is handled in other places in the codebase.

melinath:
The cucumber test that I'd tried to add was failing. I looked into the available cucumber steps and it looks like the closest correct test I would be able to write while maintaining cucumber code style would be something like (I know this isn't actual cucumber syntax):

1. Verify "Select" checkbox is unchecked
2. Click "Select" label
3. Verify "Select" checkbox is checked

But that feels very tautological so I'm not sure it's useful? I'm happy to add something back in if that would be preferred or if there's another test you'd like to see!

melinath:
The test failure for `TEST_GROUP="cucumber features/tag_sets"` seems unrelated but I don't have access to re-run it to see if it's intermittent.

sarken:
Thanks for the pull request! You're right that the failure is unrelated -- I'll retry it until it passes. Someone will be along to give a proper review in a few days.

melinath:
@redsummernight thanks for having a look! Re: the test, sounds good! I've restored the test I removed and cleaned it up a bit to meet that specification.

melinath:
I don't seem to have access to "remove the Reviewed: Action Needed label and apply the Coder Has Actioned Review label" as described in the wiki so I'm not sure what the next step is - I've gone with re-requesting review – sorry if that's incorrect!

app/views/inbox/show.html.erb
```diff
@@ -64,8 +64,11 @@
                 </li>
               <% end %>
 
-              <li class="action">
-                <%= check_box_tag "inbox_comments[]", inbox_comment.id, false, :title => ts("select this comment"), :id => "inbox_comments_#{inbox_comment.id}" %>
+              <li>
+                <%= label_tag "inbox_comments_#{inbox_comment.id}" do %>
+                  <%= check_box_tag "inbox_comments[]", inbox_comment.id, false, :id => "inbox_comments_#{inbox_comment.id}" %>
```
redsummernight, line 69: Minor style: can you use `id: ` instead of `:id => `?

features/comments_and_kudos/inbox.feature
```diff
@@ -49,6 +49,25 @@ Feature: Get messages in the inbox
       And I press "Filter"
     Then I should not see "guest on Down for the Count"
 
+  Scenario: I can bulk edit comments in my inbox by clicking 'Select'
+    Given I am logged in as "boxer" with password "10987tko"
```
redsummernight, line 53: You can use `Given I am logged in as "boxer"` since we don't care what the passwords are in this test. You don't have to update the existing tests though.

features/comments_and_kudos/inbox.feature
```diff
@@ -49,6 +49,25 @@ Feature: Get messages in the inbox
...
+      And I post the work "The Fight"
+      And I set my preferences to turn on messages to my inbox about comments
```
redsummernight, line 55: You can skip this step, as the preference is on my default.

sarken:
Let's skip fixing the checkbox label alignment on this one, since it's an existing issue that happens in other places as well.

app/views/inbox/show.html.erb
melinath: Absolutely! Resolved in 8f66cefa8.

features/comments_and_kudos/inbox.feature
melinath: Resolved in ae6fb4b98

features/comments_and_kudos/inbox.feature
melinath: Resolved in 168a07ecc

PR #3794, opened 2020-04-10
AO3-5917 Set TOS acceptance cookie in browsers that don't handle localStorage

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5917

## Purpose

Checks if we can write TOS acceptance to local storage in a user's browser and, if not, creates a cookie with that info instead. This will help with private windows in older versions of desktop and iOS Safari. 

We used to use [jquery-cookie](https://github.com/carhartl/jquery-cookie) for managing cookies with JavaScript, but that was removed in a #3781, when we thought we would no longer need cookies for this (ah, how naive we were 😭),  and is no longer maintained. I've switched us to [js-cookie](https://github.com/js-cookie/js-cookie) per jquery-cookie's README. 

## Credit

teyla, she/her or they/them
Sarken, she/her


app/views/layouts/_javascripts.html.erb
```diff
@@ -52,7 +52,7 @@
     params[:controller] == "abuse_reports" && params[:action] == "new" ||
     params[:controller] == "feedbacks" && params[:action] == "new" %>
   <%= javascript_tag do %>
-    <% # localStorage polyfill we only need for guests. %>
+    <% # localStorage support check we only need for guests. %>
```
redsummernight, line 55: Can we remove the lines that try `sessionStorage`?

app/views/layouts/_javascripts.html.erb
```diff
@@ -71,41 +71,13 @@
...
-          hasOwnProperty: function (sKey) {
-            return (new RegExp("(?:^|;\\s*)" + escape(sKey).replace(/[\-\.\+\*]/g, "\\$&") + "\\s*\\=")).test(document.cookie);
-          }
-        };
-        window.localStorage.length = (document.cookie.match(/\=/g) || window.localStorage).length;
+      function acceptTOS() {
+        if (isSupported()) {
+          localStorage.setItem("accepted_tos", "<%= @current_tos_version %>");
+        } else {
+          // 20 years is the Rails maximum and what we previously used.
```
redsummernight, line 78: "and what we previously used" is not necessary, we can just say it's the same as Rails' permanent cookies.

PR #3795, opened 2020-04-10
AO3-5771 Correct URLs in tag lists generated from "show hidden" links

melinath:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5771 (please fill in issue number and remove this comment)

## Purpose

Users can decide to hide archive warnings and/or freeform tags in their preferences. If they do, the tag lists they see on work list and detail pages will have "show warnings" / "show additional tags" links instead of the respective tag lists. Clicking on one of those links causes an AJAX request to `/tags/show_hidden` which returns a javascript file that replaces the link with custom-generated tag content.

Previously, the view for /tags/show_hidden generated links to the tag detail pages; however, the single location that uses the show_hidden functionality explicitly links to tag works, as can be seen here: https://github.com/otwcode/otwarchive/blob/030fa46/app/views/works/_meta.html.erb#L27\

This PR corrects the mismatch by making the response from /tags/show_hidden use tag works urls instead of tag detail urls.

## Testing Instructions

1. View a work and click on a warning / freeform tag
2. The link will go to the /tags/TAG/works page for that tag
3. View a work list page (like Browse > works) and click on a warning / freeform tag
4. The link will go to the /tags/TAG/works page for that tag
5. Go to preferences and turn on hide warnings/hide additional tags
6. View the work and show the hidden tags and click on one
7. The link should go to the /tags/TAG/works page for that tag
8. View a work list page (like Browse > works) and show the hidden tags and click on one
9. The link should go to the /tags/TAG/works page for that tag

## Credit

Stephen Burrows

he/him

### DIFF
--- a/factories/tags.rb
+++ b/factories/tags.rb
@@ -68,6 +68,10 @@ FactoryBot.define do
     sequence(:name) { |n| "Unsorted Tag #{n}" }
   end
 
+  factory :archive_warning do
+    sequence(:name) { |n| "The #{n} Archive Warning"}
+  end
+
   factory :fandom do
     sequence(:name) { |n| "The #{n} Fandom" }
 
--- a/factories/tags.rb
+++ b/factories/tags.rb
@@ -69,7 +69,7 @@ FactoryBot.define do
   end
 
   factory :archive_warning do
-    sequence(:name) { |n| "The #{n} Archive Warning"}
+    sequence(:name) { |n| "The #{n} Archive Warning" }
   end
 
   factory :fandom do
--- a/factories/tags.rb
+++ b/factories/tags.rb
@@ -68,10 +68,6 @@ FactoryBot.define do
     sequence(:name) { |n| "Unsorted Tag #{n}" }
   end
 
-  factory :archive_warning do
-    sequence(:name) { |n| "The #{n} Archive Warning" }
-  end
-
   factory :fandom do
     sequence(:name) { |n| "The #{n} Fandom" }
 
--- a/features/other_a/preferences_edit.feature
+++ b/features/other_a/preferences_edit.feature
@@ -104,6 +104,7 @@ Feature: Edit preferences
   by-case basis.
 
   Given I limit myself to the Archive
+    And a canonical freeform "Scary tag"
     And I am logged in as "someone_else"
     And I post the work "Someone Else's Work" as part of a series "A Series"
     And I am logged in as "tester"
--- a/features/other_a/preferences_edit.feature
+++ b/features/other_a/preferences_edit.feature
@@ -123,6 +123,8 @@ Feature: Edit preferences
     And I should not see "Show additional tags"
   When I follow "Show warnings"
   Then I should see "No Archive Warnings Apply" within "dl.work.meta"
+  When I follow "No Archive Warnings Apply" within "dl.work.meta"
+  Then I should be on the works tagged "No Archive Warnings Apply"
   When I view the work "My Work"
   Then I should see "No Archive Warnings Apply" within "dl.work.meta"
     And I should not see "Show warnings"
@@ -138,6 +140,8 @@ Feature: Edit preferences
     And I should not see "Show additional tags"
   When I follow "Show warnings"
   Then I should see "No Archive Warnings Apply" within "li.warnings"
+  When I follow "No Archive Warnings Apply" within "li.warnings"
+  Then I should be on the works tagged "No Archive Warnings Apply"
   When I go to my works page
   Then I should see "My Work"
     And I should see "No Archive Warnings Apply" within "li.warnings"
@@ -154,6 +158,8 @@ Feature: Edit preferences
     And I should not see "Show additional tags"
   When I follow "Show warnings"
   Then I should see "No Archive Warnings Apply" within "li.warnings"
+  When I follow "No Archive Warnings Apply" within "li.warnings"
+  Then I should be on the works tagged "No Archive Warnings Apply"
 
   # Warnings are hidden in bookmark blurbs.
   # This is slightly excessive -- bookmarks use the work blurb -- but we'll
@@ -166,6 +172,8 @@ Feature: Edit preferences
     And I should not see "Show additional tags"
   When I follow "Show warnings"
   Then I should see "No Archive Warnings Apply" within "li.warnings"
+  When I follow "No Archive Warnings Apply" within "li.warnings"
+  Then I should be on the works tagged "No Archive Warnings Apply"
 
   # Change tester's preferences to hide freeforms as well as warnings.
   When I go to my preferences page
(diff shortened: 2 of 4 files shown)


factories/tags.rb
```diff
@@ -68,6 +68,10 @@
     sequence(:name) { |n| "Unsorted Tag #{n}" }
   end
 
+  factory :archive_warning do
```
redsummernight, line 71: We don't add a factory for warnings because there is a set number of them and you can refer to them by names, e.g. `ArchiveConfig.WARNING_DEFAULT_TAG_NAME`.

spec/controllers/tags_controller_spec.rb
```diff
@@ -173,6 +173,66 @@
     end
   end
 
+  describe "show_hidden" do
+    render_views
```
redsummernight, line 177: We generally avoid `render_views`. If we need to inspect the final HTML, we prefer Cucumber tests, which run a headless browser, even more so when the output involves JavaScript.

There's a Cucumber test that checks if the "Show warnings" and "Show additional tags" buttons work correctly:
https://github.com/otwcode/otwarchive/blob/3b3262673b355ff1980e57b21d300de4b576235b/features/other_a/preferences_edit.feature#L117-L130

You can extend the test to follow the links and check if you end up on the right works paths (`Then I should be on the works tagged "???"`).

The only RSpec test we should keep is the one for "redirects to referer with an error for HTML format".

spec/controllers/tags_controller_spec.rb
```diff
@@ -173,6 +173,66 @@
...
+
+    before do
+      @fandom1 = FactoryBot.create(:fandom, canonical: true)
```
redsummernight, line 180: - You should use `let` instead of instance variables: http://www.betterspecs.org/#let.
- House style: we usually leave out `FactoryBot.` in RSpec tests.

We don't do either consistently yet, but we prefer them for new tests. You don't have to change existing ones, though.

factories/tags.rb
melinath: Sounds good! I wasn't sure since there weren't other tests using archive warnings. But it makes sense, since the tag will be auto-created as part of work creation if it doesn't already exist. I'll go ahead and remove this!

spec/controllers/tags_controller_spec.rb
melinath: Okay, can do! I didn't realize there was already a test for this! Thanks.

spec/controllers/tags_controller_spec.rb
melinath: Thanks, will do!

melinath: Ah crud, I missed this in the test style docs - sorry, I'm just learning Ruby so I didn't have the context for that the first time I saw it. I'll have a look through those.

melinath: Resolved in 40e540a5ab909648b65b259ee99ec8c10de550b6

factories/tags.rb
melinath: Resolved in da1fb7273

spec/controllers/tags_controller_spec.rb
melinath: Removed render_views in 0b9be2643d228c57cf4e0960c899ea173adfbc3f. Extended preferences_edit.feature cucumber tests in 74fbc36c03483f2ad945b34e137e9b325b33c066 – but in local testing, I'm getting a failure because `Scary tag` is (sometimes?) not common enough to support a works page.

melinath:
Just to make it clearer what's going on I'm moving this comment to here: The test failure is real, but it's happening because the "Scary tag" works page redirects to the tag detail page, which has the following note:

> This tag has not been marked common and can't be filtered on (yet).

I'm at a bit of a loss for what to do here; I can do some digging but if someone happens to be familiar enough with the fixture data and archive in general to offer advice off the top of their head I would be grateful.

spec/controllers/tags_controller_spec.rb
melinath: Left a longer version of the above comment below - resolving this conversation for now.

features/other_a/preferences_edit.feature
```diff
@@ -250,6 +258,8 @@ Feature: Edit preferences
     And I should not see "Scary tag" within "dl.work.meta"
   When I follow "Show additional tags"
   Then I should see "Scary tag" within "dl.work.meta"
+  When I follow "Scary tag" within "dl.work.meta"
+  Then I should be on the works tagged "Scary tag"
```
sarken, line 262: You can change this to `Then I should be on the "Scary tag" tag page`. What's happening is `Scary tag` isn't a canonical (common) tag, so it doesn't have a works page.  Being taken to the tag page instead is the expected fallback behavior.

melinath: ah, "common" means the same as canonical? Great! I do need to have a test for ending up on the works page (since the bug is ending up on the detail page) but maybe I can find a different tag to work with!

melinath: Okay, I think I've resolved this in 84827e9 by marking Scary tag as canonical, but if that's not okay for some reason I'm happy to do something different.

melinath:
👆 this test failure seems to be unrelated to my changes - here's the corresponding test run for my fork, which has features/works passing: https://travis-ci.org/github/melinath/otwarchive/builds/674608155

redsummernight:
It's one of our [known flaky tests](https://github.com/otwcode/otwarchive/wiki/Automated-Testing#Known_Test_Failures).

spec/controllers/tags_controller_spec.rb
```diff
@@ -173,6 +173,45 @@
     end
   end
 
+  describe "show_hidden" do
+    let!(:fandom1) { create(:fandom, canonical: true) }
+    let!(:freeform1) { create(:freeform, canonical: false) }
+    let!(:character1) { create(:character, canonical: false) }
+    let!(:character2) { create(:character, canonical: false) }
+    let!(:work) do
```
redsummernight, line 181: You can remove the 2 ajax success tests, which we already cover in Cucumber.

For the remaining 2 tests, we're just looking for the error message "you need to have JavaScript". We don't actually need to set up the tags, so the test setup can just be:

```rb
describe "show_hidden" do
  let(:work) { create(:work, posted: true) }
```

It's fine to use `let` over `let!` here and have the work be lazily created when `work.id` is first used.

spec/controllers/tags_controller_spec.rb
```diff
@@ -173,6 +173,45 @@
...
+      create(:work,
+             posted: true,
+             fandom_string: fandom1.name,
+             character_string: "#{character1.name},#{character2.name}",
+             freeform_string: freeform1.name,
+             archive_warning_string: ArchiveConfig.WARNING_DEFAULT_TAG_NAME)
+    end
+
+    it "redirects to referer with an error for non-ajax warnings requests" do
+      referer = root_path
```
redsummernight, line 191: The referer defaults to `root_path` if unset:

https://github.com/otwcode/otwarchive/blob/030fa4607d3c59fba2cacbe4e9c99b3c49fbc1cb/app/controllers/tags_controller.rb#L175

so it's better to use a different path to test it, e.g. `tags_path`.

melinath: yep, sounds good! Resolved in 20ee3b511

spec/controllers/tags_controller_spec.rb
melinath: Sure - resolved in 20ee3b511.

PR #3798, opened 2020-04-14
AO3-3582 Update jquery.livequery.min.js to v1.3.6

melinath:
# Pull Request Checklist

  ^ autocomplete already seems to be thoroughly tested in the cucumber tests; the show / hide links are also tested in cucumber tests.

## Issue

https://otwarchive.atlassian.net/browse/AO3-3582

## Purpose

Updates the version of the jquery livequery plugin to 1.3.6 (from 1.1.1) to be slightly more up-to-date.

## Testing Instructions

Because this is a version upgrade, I believe it can only be tested by verifying that the relevant functionality is still working. As far as I know this package is now only used for autocomplete, so to test, you can go to a page that uses an autocomplete widget and walk through a workflow that uses autocomplete. For example:

1. Go to Search > Edit Your Search
2. Fill out a fandom in the search field
3. Submit the search form.

Livequery was also being used to make scrolling work correctly for comment pagination and for the comments link at the top of a work.

1. Go to a work with more than one page of comments.
2. Click on the Comments link at the top of the page. You should be scrolled down to the comments section and comments should display.
3. Click on a pagination link at the bottom of the comments list. You should be scrolled back up to the top of the comments list.

## Additional information

The version of 1.3.6 from https://plugins.jquery.com/livequery/ wasn't working right for me so I got the version from https://github.com/hazzik/livequery/tree/1.3.6 (the repository that was linked as the homepage for the project in the plugin registry.)

I had to make a couple of updates to application.js because there was lingering usage of livequery to capture live events on elements. This usage was removed from livequery in v1.2 because it was supplanted by jquery.live, which was removed in jquery 1.9 (the current version in use by AO3) in favor of [event delegation with `.on()`](https://api.jquery.com/on/)

I'm not sure if there's an appropriate way to test that the scrolling functionality the changed callbacks represent is working properly in cucumber, but there are already tests for the other behavior associated with these links.

## Credit

Stephen Burrows

he/him


public/javascripts/jquery.livequery.min.js
```diff
@@ -1,9 +1,8 @@
-/* Copyright (c) 2010 Brandon Aaron (http://brandonaaron.net)
```
redsummernight, line 1: Since we're using jQuery 1.9, can we switch to `.on` and outright remove this plugin instead of updating it?

There's another instance of livequery use we can replace here:

https://github.com/otwcode/otwarchive/blob/58ac2bec9cf942e60911a54ff96f951d0fe1de76/public/javascripts/application.js#L53-L56

I don't think there's an easy way to test scrolling functionality in Cucumber, we'll just manually test that.

melinath: I would love to help remove this plugin! Unfortunately I don't _think_ it's easy to replace that particular usage of livequery with `.on`.

My understanding is that [`on`](https://api.jquery.com/on/) only handles attaching event handlers to a set of elements; that behavior was duplicated in the older version of the plugin but [doesn't exist in the newer version](https://github.com/hazzik/livequery/blob/master/README.md).

The new version of the plugin still exists to support the syntax currently being used for autocomplete, which allows specifying callbacks for when a matching element is added to or removed from the DOM. From what I can tell, there is a Mutation Observer API that could be used to monitor those changes instead, [as long as you don't need to support IE < 11](https://caniuse.com/#feat=mutationobserver), but [it's not using events](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) so I'm pretty sure it would not be compatible with `.on()`. [Here's a short blog post on how to use MutationObserver.](https://gabrieleromanato.name/jquery-detecting-new-elements-with-the-mutationobserver-object)

If you'd like, I would be happy to rewrite the autocomplete initializer to use MutationObserver or some other methodology so that the plugin can be removed fully?

redsummernight: `.on` could handle future elements if you use delegated event handlers. We do that with modals for example:

https://github.com/otwcode/otwarchive/blob/63adfaf4641a6fc9c2e24911d86a563fe8867508/public/javascripts/ao3modal.js#L312-L313

melinath: @redsummernight sorry, just saw your reply. Yes, `.on()` allows you to attach i.e. click events to elements that are added to the DOM later.

What livequery allows is attaching a callback to the "event" of a new element being added or removed (which can't be handled with `.on()` because it's not a standard javascript event.) There are alternative implementations of this in modern browsers, as I've described above.

In the case of autocomplete, livequery is still being used to say that whenever an autocomplete input is added to the DOM, it should be initialized. I could replace that with MutationObserver if the browser support is sufficient.

However, I'm pretty sure I can't replace it with `.on()` – is there a specific event that you think I should be capturing that way?

redsummernight: Thanks for looking into this. You're right that `.on()` needs specific events, and moving away from the plugin is more complicated than I thought. For now, let's stick with updating livequery.

melinath: sounds good, thanks!

PR #3811, opened 2020-04-21
AO3-5266 Enforce HTTPS for ning.com and vidders.net embeds

janakuz:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5266 

## Purpose

Ensures embeds from vidders.net use https.

## Testing Instructions

Post a work and include the following embed code:
\<embed wmode="opaque" src="http://static.ning.com/socialnetworkmain/widgets/video/flvplayer/flvplayer.swf?xn_version=3704000907" flashvars="config=http%3A%2F%2Fvidders.net%2Fvideo%2Fvideo%2FshowPlayerConfig%3Fid%3D2820784%253AVideo%253A172097%26ck%3D-%26theme982Version%3D39&amp;video_smoothing=on&amp;autoplay=off&amp;hideShareLink=1&amp;isEmbedCode=1" width="456" height="197" allowfullscreen="true" type="application/x-shockwave-flash" allowscriptaccess="never" allownetworking="internal"></embed><br /> 

The values in src and flashvars should be converted to https.

## References

Issue https://otwarchive.atlassian.net/browse/AO3-5192 is linked in the issue text and does the same thing for other sources.

## Credit

J


lib/otw_sanitize/embed_sanitizer.rb
```diff
@@ -116,6 +116,10 @@ def standardize_url(url)
     def ensure_https
       return unless supports_https? && node['src'].present?
       node['src'] = node['src'].gsub("http:", "https:")
+      if allows_flashvars? && node['flashvars'].present?
```
redsummernight, line 119: Can you add a test where HTTP links in flashvars get converted to HTTPS?

lib/tasks/after_tasks.rake
```diff
@@ -582,6 +582,23 @@ namespace :After do
     end
   end
 
+  desc "Enforce HTTPS where available for embedded media from ning.com and vidders.net"
+  task(enforce_https: :environment) do
+    Chapter.find_each do |chapter|
+      if chapter.id % 1000 == 0
```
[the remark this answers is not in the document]
janakuz: Should I do anything about the style violations in this file? Since I just copy-pasted the similar task and modified it. And if so, should I also change the other task?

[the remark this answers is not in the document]
redsummernight: There's a lot of code around with style violations, since we weren't always consistent in using style checkers. In general, we fix violations on lines added/modified in the current pull request, and leave the existing ones alone.

If you'd like to check for violations locally, you can [run `rubocop`](https://github.com/otwcode/otwarchive/wiki/Coding-standards#Previewing_Hound_comments). It's not a perfect solution since it reports everything in a file, not just the changed lines, but it can save you from the back-and-forth with the Hound bot.

lib/tasks/after_tasks.rake
```diff
@@ -582,6 +582,21 @@ namespace :After do
     end
   end
 
+  desc "Enforce HTTPS where available for embedded media from ning.com and vidders.net"
+  task(enforce_https: :environment) do
```
redsummernight, line 586: Can you give this task a different name from the existing HTTPS task?

lib/tasks/after_tasks.rake
```diff
@@ -582,6 +582,21 @@ namespace :After do
...
+    Chapter.find_each do |chapter|
+      puts chapter.id if (chapter.id % 1000).zero?
+      if chapter.content.match /<(embed|iframe).*(ning.com|vidders.net)/
```
redsummernight, line 589: You should escape the dots in `ning.com|vidders.net` or they'll match any character.

Even with that fix, there can still be false matches (e.g. works can link to `vidders.net` outside the embed tags), but I think this is a good enough heuristic.

janakuz: The <(embed|iframe) is still in my expression, and my goal was to match <(embed or iframe)(any characters)(ning.com or vidders.net). I checked it with a regex tester (https://rubular.com/) and it doesn't match if I replace embed with something else (other than iframe, which maybe isn't necessary here, but I figured it wouldn't do any harm). It does match embed tags that only contain ning.com or vidders.net and not both, but I don't know how likely it is for those to appear, and if they should be ignored if they do.

But yeah, I did escape the dots.

PR #3813, opened 2020-04-21
AO3-5901 Limit number of pseuds on profile and sidebar

CristinaRO:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5901

## Purpose

What does this PR do?

Limit the number of pseuds shown on the user's profile and on the user sidebar to the configured value in `ArchiveConfig.ITEMS_PER_PAGE`.

I have chosen to always include the default pseud as the first one, and then the rest of the pseuds alphabetically.

I've also repackaged the profile into a presenter, to keep the logic of arranging of data pulled from three different models (user, profile, preference) away from the view.

## Testing Instructions

Instructions are on the JIRA ticket. In order to test _before_ the nginx block is removed, one of the testy users might have to temporarily gain more than 25 pseuds...

## References

Yes, AO3-5889 - paginate pseuds, which I'm also working on.

## Credit

Enigel

she/her


app/views/profile/show.html.erb
```diff
@@ -2,35 +2,35 @@
   <%= render 'users/header' %>
 
   <!--main content-->
-  <% unless @user.profile.title.blank? %>
-    <h3 class="heading"><%=h @user.profile.title %></h3>
+  <% if @profile.title.present? %>
```
sarken, line 5: 😍

app/models/pseud.rb
```diff
@@ -112,6 +112,7 @@ class Pseud < ApplicationRecord
 
   scope :alphabetical, -> { order(:name) }
   scope :starting_with, lambda {|letter| where('SUBSTR(name,1,1) = ?', letter)}
+  scope :for_profile, -> { order(is_default: :desc).merge(alphabetical).limit(ArchiveConfig.ITEMS_PER_PAGE) }
```
sarken, line 115: I'm a little iffy on the name of the scope, since it's not _just_ for the profile, but also the dashboard. I'm having trouble coming up with something better, though. `short_list` was my initial thought, but if we ever increase `ITEMS_PER_PAGE` to something like `100`, is that really short? I guess maybe, compared to the full list. `first_page` to go with the whole per page thing? 

Anyone else have better suggestions?

features/other_a/profile_edit.feature
```diff
@@ -99,6 +103,10 @@ Scenario: Entering date of birth and displaying
   When I change my preferences to display my date of birth
   Then I should see "My birthday: 1980-11-30"
     And 0 emails should be delivered
+  When I log out
+    And I go to editname's user page
+    And I follow "Profile"
+    Then I should see "My birthday: 1980-11-30"
```
sarken, line 109: Could you outdent the `Then` line so it lines up with the `When`?

features/other_a/preferences_edit.feature
```diff
@@ -50,10 +50,8 @@ Feature: Edit preferences
   Then I should not see "My email address"
     And I should not see "My birthday"
   When I am logged in as "editname" with password "password"
-  Then I should see "Hi, editname!"
-    And I should see "Log Out"
-  When I post the work "This has two chapters"
-  And I follow "Add Chapter"
+    And I post the work "This has two chapters"
```
sarken, line 53: Oof, this scenario. Testing way too many preferences at once. 😭 

I think we can take "personal details" out of the scenario description since we've stopped testing the birthdate and email address parts in this scenario (thank you for that!).

And I think we can streamline and update the work/chapter parts using some of the new steps that have been defined in the time since these tests were original written:

* `And I post the 2 chapter work "This has two chapters"` instead of manually going through and adding the second chapter here
* `Then I should be on the 2nd chapter of the work "This has two chapters"` and `Then I should  be on the 1st chapter of the work "This has two chapters"` instead of checking the chapter title on lines 58 and 60 to determine we're seeing the proper chapter
* And I _think_ we can do `Then I should be on the work "This has two chapters" in full mode` for the final check on line 85

spec/controllers/profile_controller_spec.rb
```diff
@@ -16,5 +16,15 @@
       get :show, params: { user_id: @user }
       expect(@user.profile).to be
     end
+
+    it 'should use the profile presenter for the profile' do
```
sarken, line 20: Could you write this as `uses the profile presenter for the profile`? I know you were just copying the `should`s already in here since you didn't  use them in your next spec file, but I think I'd rather cope with the inconsistency within the file than add another unnecessary `should` 😉

spec/models/profile_presenter_spec.rb
```diff
@@ -0,0 +1,71 @@
+require 'spec_helper'
+
+describe ProfilePresenter do
+  let(:user) { create(:user, email: "example@example.com") }
```
sarken, line 4: Should we use `Faker::Internet.email` here and do `expect(subject.email).to eq(user.email)` on line 29? (I can never decide what the best thing to do is in cases like this, so I'm genuinely asking.)

spec/requests/profile_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+
+    before do
+      FactoryBot.create(:pseud, user: user, name: "Slartibartfast")
+      FactoryBot.create(:pseud, user: user, name: "Agrajag")
+      FactoryBot.create(:pseud, user: user, name: "Betelgeuse")
+      allow(ArchiveConfig).to receive(:ITEMS_PER_PAGE).and_return(3)
+    end
+
+    it "shows only the configured maximum number of pseuds" do
+      visit "/users/Zaphod/profile"
```
sarken, line 17: We don't usually test views in specs... idk if that's something we want to start doing here?

app/models/pseud.rb
redsummernight: `:alphabetical_preview`?

tickinginstant, line 115: Probably cleaner to call `.alphabetical` instead of `.merge(alphabetical)`.

app/views/profile/show.html.erb
```diff
@@ -2,35 +2,35 @@
...
+    <h3 class="heading"><%=h @profile.title %></h3>
   <% end %>
 
   <div class="wrapper">
     <dl class="meta">
       <dt class="pseuds"><%= link_to ts("My pseuds:"), user_pseuds_path(@user) %></dt>
-      <dd class="pseuds"><%= print_pseud_list(@user.pseuds) %></dd>
+      <dd class="pseuds"><%= print_pseud_list(@profile.pseuds) %></dd>
```
tickinginstant, line 12: It seems like there ought to be some kind of text to indicate when the user has more than just these pseuds, e.g. "and X more" or something that links to the pseuds page.

app/views/users/_sidebar.html.erb
```diff
@@ -8,7 +8,7 @@
       <% pseud_link_text = current_page?(@user) ? ts("Pseuds") : (@author ? @author.name : @user.login) %>
       <a href="#" title="<%= ts("Pseud Switcher") %>"><%= pseud_link_text %></a>
       <ul class="expandable secondary">
-        <%= print_pseud_selector(@user.pseuds) %>
+        <%= print_pseud_selector(@user.pseuds.for_profile) %>
```
tickinginstant, line 11: The pseud selector is set up as a sequence of `span_if_current` links. This means that if the user is viewing a particular pseud's dashboard and they view the selector in the sidebar, the current pseud is highlighted in pale grey. But always displaying the first 20 pseuds means that when the user is viewing a particular pseud, they're not guaranteed to see it in the sidebar. Is that okay?

(Should the current pseud be added to the list at the top? Should the list of pseuds be paginated and show the page that the pseud would be on?)

spec/models/profile_presenter_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+  let(:profile) { user.create_profile }
+  let(:preference) { user.create_preference }
```
tickinginstant, line 6: Why `create_profile`/`create_preference` here, instead of `profile` and `preference`? They should have been created by the callbacks in the user class:
https://github.com/otwcode/otwarchive/blob/86d9cfcb409ff316ef2ffa9ad2188241e151abf9/app/models/user.rb#L284-L288

app/views/profile/show.html.erb
sarken: It could be the same "All Pseuds (#)" text we use on the sidebar.

app/views/users/_sidebar.html.erb
sarken: Putting the currently-selected pseud at the top sounds good.

features/other_a/preferences_edit.feature
CristinaRO: We couldn't do that for the final line, because it was expecting the query param in the URL, hence now checking for the lack of a "Next Chapter" text. Yes, negatives can be deceitful if we ever change the next chapter text, but we'd have to break both preferences and the text for that to bite us, and I think (hope) preferences are tested elsewhere as well, or will be by the time we break them...

spec/models/profile_presenter_spec.rb
CristinaRO: I was taught that sometimes it's best / most readable to check for something we know we put there, and can see in the spec. There are different mini-schools of thought, and so far it hasn't convinced me that one is The Best in any significant way. 🤷‍♀️

spec/requests/profile_spec.rb
CristinaRO: I definitely wanted to test the view in this case, since it didn't seem to be covered enough by unit tests, and this was the easiest way to have an integration-like test that wouldn't make me want to yeet the laptop into the sun. (The way anything cucumber does. 😭 )

spec/models/pseud_spec.rb
```diff
@@ -48,4 +48,28 @@
...
+    it "returns a maximum of ITEMS_PER_PAGE pseuds" do
+      expect(subject.count).to eq(3)
+    end
+
+    it "always contains the default pseud first" do
+      expect(subject.first.name).to eq("Zaphod")
+    end
+
+    it "is in alphabetical order after the default pseud" do
+      expect(subject.map(&:name)).to eq(["Zaphod", "Agrajag", "Betelgeuse"])
```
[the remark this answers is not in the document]
CristinaRO: I disagree very much with this "style" rec. I hate it, it makes it harder to tell at a glance what the thing is, and no one could convince me that it's a genuine benefit to code readability or maintainability.

app/models/pseud.rb
CristinaRO: This is the last outstanding thing, I'll let everyone percolate some last minute ideas on what we should name the bike shed while I sleep. ;-D

CristinaRO: @sarken's `abbreviated_list` won!

app/views/profile/show.html.erb
```diff
@@ -2,35 +2,37 @@
...
+  <% if @profile.title.present? %>
+    <h3 class="heading"><%=h @profile.title %></h3>
   <% end %>
 
   <div class="wrapper">
     <dl class="meta">
-      <dt class="pseuds"><%= link_to ts("My pseuds:"), user_pseuds_path(@user) %></dt>
-      <dd class="pseuds"><%= print_pseud_list(@user.pseuds) %></dd>
+      <dt class="pseuds">
+        <%= link_to ts("All my pseuds (%{pseud_number})", :pseud_number => @user.pseuds.count), user_pseuds_path(@user) %>
```
tickinginstant, line 12: I think the phrasing here is confusing. By having `All my pseuds (#)` as the `dt` and a list of pseuds as the `dd`, it seems to be implying that all of the pseuds are included in the list.

That's why I suggested adding `and # more pseuds` (or some similar text) at the end of the pseud list. It's hard to quickly count how many pseuds are displayed. If someone has 15 or 25 pseuds, it's going to be difficult to know immediately whether the list includes all of them, or whether the viewer need to click the link to see the rest.

app/views/users/_sidebar.html.erb
```diff
@@ -5,10 +5,10 @@
   <li><%= span_if_current ts("Profile"), user_profile_path(@user) %></li>
   <% if @user.pseuds.size > 1 %>
     <li class="pseud" aria-haspopup="true">
-      <% pseud_link_text = current_page?(@user) ? ts("Pseuds") : (@author ? @author.name : @user.login) %>
+      <% pseud_link_text = current_page?(@user) ? ts("Pseuds") : (@pseud ? @pseud.name : @user.login) %>
```
tickinginstant, line 8: Hmm. I like the idea of having the current pseud name as the head of the pseud switcher (particularly since that seems to have been the original intention), but it does mean that on the page for a pseud, the username isn't displayed anywhere prominent.

For example, [testy's orphantest1 pseud](https://archiveofourown.org/users/testy/pseuds/orphantest1) has no works and no bookmarks, so the only place testy's name appears at the moment is at the top of the pseud switcher. You can see also see the testy pseud in the switcher, but only if you click on it. For pseuds with works and/or bookmarks, like [testy's orphantest2 pseud](https://archiveofourown.org/users/testy/pseuds/orphantest2), you can also see the username in the byline, but that's a lot easier to miss.

Of course, the current location of the username isn't as prominent as the pseud name, and I suspect that's the cause of a number of accidental subscriptions to orphan_account. But I think removing it from its current location (without finding somewhere else to put it) might make those types of mistakes more frequent.

sarken: Hmm, interesting. I actually thought the current behavior was a bug -- I expected it to either show "Pseuds" all the time or the currently-selected pseud and couldn't fathom why it was showing something else.

I think if we're worried about confusion regarding which user's pseud we're looking at, it makes more sense to change the page headings to display "pseud (username)" -- since that's the format we prefer in most other places on the Archive -- rather than relying on someone figuring it out from the pseud switcher.

spec/requests/profile_spec.rb
sarken: Would you like me to write the cukes, then?

app/views/profile/show.html.erb
sarken: Yeah, I meant put the "All Pseuds (#)" text at the end rather than changing the text in the `<dt>`, but you make a good point about the math. 

However, I'd expect something like "and X more" to expand the list to show the remaining pseuds (like the "x more users" link in kudos or the "See all" link in the "Signed up" section of [challenge profiles](https://archiveofourown.org/collections/yuletide2015/profile)), not take me to a separate page. 

Maybe something more like "(All pseuds, including X more)"?

app/views/users/_sidebar.html.erb
sarken: Made an issue for fixing the dashboard's heading: https://otwarchive.atlassian.net/browse/AO3-6006

spec/requests/profile_spec.rb
cesy: https://github.com/CristinaRO/otwarchive/pull/1 has an equivalent, the only thing it doesn't check is that it's within dl.meta.

app/views/profile/show.html.erb
CristinaRO: I would prefer to have a clear decision handed to me, my decision-making cognitive tank is depleted.

app/views/users/_sidebar.html.erb
CristinaRO: Same: please let me know if this thread still needs resolving, and how.

CristinaRO:
This Hakiri warning is about our version of `rack`, not in the scope of this PR.

app/views/users/_sidebar.html.erb
sarken: It can stay how you have it. We have a new issue for displaying the username more prominently.

app/views/profile/show.html.erb
sarken: You can just use my previous suggestion since I am also similarly exhausted.

spec/models/pseud_spec.rb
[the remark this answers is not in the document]
CristinaRO: ...but then I gave up and applied it anyway.

spec/models/pseud_spec.rb
```diff
@@ -10,7 +10,7 @@
   end
 
   it "is invalid if there are special characters" do
-      expect(build(:pseud, name: '*pseud*')).to be_invalid
+    expect(build(:pseud, name: '*pseud*')).to be_invalid
```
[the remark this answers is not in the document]
CristinaRO: Not my change.

spec/models/profile_presenter_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+
+describe ProfilePresenter do
+  let(:user) { create(:user, email: "example@example.com") }
+  let(:profile) { user.profile }
+  let(:preference) { user.preference }
+  let(:subject) { ProfilePresenter.new(profile) }
+
+  describe "pseuds" do
+    it "uses the profile scope" do
+      allow(user).to receive_message_chain(:pseuds, :abbreviated_list).and_return("dummy")
```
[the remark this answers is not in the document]
CristinaRO: Nah.

spec/controllers/profile_controller_spec.rb
```diff
@@ -16,5 +16,15 @@
       get :show, params: { user_id: @user }
       expect(@user.profile).to be
     end
+
+    it 'uses the profile presenter for the profile' do
+      @user = FactoryBot.create(:user)
+      profile_presenter = double(:profile_presenter)
```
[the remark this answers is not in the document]
CristinaRO: I don't understand what this mean. 😬 Help?

spec/controllers/profile_controller_spec.rb
```diff
@@ -16,5 +16,15 @@
...
+      allow(ProfilePresenter).to receive(:new).and_return(profile_presenter)
+
+      get :show, params: { user_id: @user }
```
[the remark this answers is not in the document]
CristinaRO: True! And possible, in this case.

spec/controllers/profile_controller_spec.rb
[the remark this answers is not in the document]
CristinaRO: Just to learn, mind you, I am absolutely done with the Hound for this PR. 😩

CristinaRO:
The UI isn't letting me re-request a review from @tickinginstant , but consider yourself invited. ;)

@sarken This still has the "Coder has actioned review" label, hence not being considered "Reviewed: Ready to merge". It definitely doesn't feel right for me to change the labels...

app/views/profile/show.html.erb
```diff
@@ -2,35 +2,37 @@
...
+  <% if @profile.title.present? %>
+    <h3 class="heading"><%=h @profile.title %></h3>
   <% end %>
 
   <div class="wrapper">
     <dl class="meta">
-      <dt class="pseuds"><%= link_to ts("My pseuds:"), user_pseuds_path(@user) %></dt>
-      <dd class="pseuds"><%= print_pseud_list(@user.pseuds) %></dd>
+      <dt class="pseuds">
+        <%= link_to all_pseuds_link(@profile), user_pseuds_path(@user) %>
```
tickinginstant, line 12: I still think that the link should come inside the `dd`, not the `dt`. (Or in both, if you're set on having it in the `dt`.) Having the text `including X more` here, instead of after the list of pseuds, seems like it would be confusing. It also means that someone might think that the `dd` already includes the X more.

app/views/profile/show.html.erb
```diff
@@ -2,35 +2,37 @@
...
+      </dt>
+      <dd class="pseuds"><%= print_pseud_list(@profile.pseuds) %></dd>
       <dt><%= ts("I joined on:") %></dt>
-      <dd><%= ts("%{date}", :date => l(@user.created_at.to_date)) %></dd>
-      <% if @user.profile.location? %>
+      <dd><%= ts("%{date}", :date => l(@profile.created_at)) %></dd>
```
tickinginstant, line 16: `l()` should be sufficient to localize the date, shouldn't it? Is the `ts("%{date}")` part necessary?

(Not really in scope, but I noticed it while I was scrolling through the code.)

app/views/profile/show.html.erb
CristinaRO: @tickinginstant I am not set on anything! I am just tired and confused. Would you mind making a PR to my fork (or directly on my PR if technically possible!) with the changes you'd like to see regarding the `dt`/`dd`?

tickinginstant: The simple way would just be to have it look something like this:
```
<dt class="pseuds"><%= ts("My pseuds:") %></dt>
<dd class="pseuds">
  <p><%= print_pseud_list(@profile.pseuds) %></p>
  <p>(<%= link_to all_pseuds_link(@profile), user_pseuds_path(@user) %>)</p>
</dd>
```
Though I think my true preference would be for [something like this](https://github.com/tickinginstant/otwarchive/commit/9e68b3e825354c98e407428e6216ffd12ed6c576).

spec/models/pseud_spec.rb
```diff
@@ -48,4 +48,28 @@
...
+
+    it "returns a maximum of ITEMS_PER_PAGE pseuds" do
+      expect(subject.count).to eq(3)
+    end
+
+    it "always contains the default pseud first" do
+      expect(subject.first.name).to eq("Zaphod")
+    end
+
+    it "is in alphabetical order after the default pseud" do
```
redsummernight, line 71: The last assertion covers the previous two, I think we can remove those and update this description.

spec/models/profile_presenter_spec.rb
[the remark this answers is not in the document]
redsummernight: I'd rather not have tests using [`receive_message_chain`](https://github.com/rspec/rspec-mocks/issues/921#issuecomment-91418492) considering rspec-mocks devs aren't particularly fond of it either. We're adequately covered by other tests in this PR.

app/views/profile/show.html.erb
redsummernight: +1 to @tickinginstant's simple way.

redsummernight (changes requested):

spec/models/profile_presenter_spec.rb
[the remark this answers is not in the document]
cesy: Removed this test in my PR update

app/views/profile/show.html.erb
cesy: Fixed in my PR to Cristina's

app/views/profile/show.html.erb
cesy: Fixed in my PR to Cristina

features/other_a/profile_edit.feature
```diff
@@ -61,6 +61,10 @@ Scenario: Changing email address and viewing
     And the email should not contain "translation missing"
   When I change my preferences to display my email address
   Then I should see "My email address: valid2@archiveofourown.org"
+  When I log out
+    And I go to editname's user page
+    And I follow "Profile"
+  Then I should see "My email address"
```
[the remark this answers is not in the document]
cesy: Fixed in my PR to Cristina

spec/models/pseud_spec.rb
cesy: Fixed in my PR to Cristina, though I personally prefer the more granular tests for helping investigate failures

sarken:
Updated in https://github.com/otwcode/otwarchive/pull/4554

PR #3814, opened 2020-04-22
AO3-5889 Paginate pseuds using items per page limit

CristinaRO:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5889

## Purpose

What does this PR do?

Paginate pseuds on the user's pseuds page using items per page limit. Pagination elements both at the top and bottom of pseuds list.

## Testing Instructions

## References

Yes, https://otwarchive.atlassian.net/browse/AO3-5901

## Credit

### DIFF
--- a/app/controllers/pseuds_controller.rb
+++ b/app/controllers/pseuds_controller.rb
@@ -14,7 +14,7 @@ class PseudsController < ApplicationController
   # GET /pseuds.xml
   def index
     if @user
-      @pseuds = @user.pseuds
+      @pseuds = @user.pseuds.paginate(page: params[:page])
       @rec_counts = Pseud.rec_counts_for_pseuds(@pseuds)
       @work_counts = Pseud.work_counts_for_pseuds(@pseuds)
       @page_subtitle = @user.login
--- a/spec/requests/pseuds_spec.rb
+++ b/spec/requests/pseuds_spec.rb
@@ -1,0 +1,40 @@
+require "spec_helper"
+
+describe "Pseuds" do
+  subject { page }
+
+  describe "show" do
+    let(:user) { create(:user, login: "Agrajag") }
+
+    before do
+      create(:pseud, user: user, name: "Zaphod")
+      create(:pseud, user: user, name: "Slartibartfast")
+      create(:pseud, user: user, name: "Betelgeuse")
+    end
+
+    it "shows only the configured maximum number of pseuds" do
+      allow(Pseud).to receive(:per_page).and_return(3)
+
+      visit "/users/Agrajag/pseuds"
+
+      within("ol.pagination", match: :first) do
+        is_expected.to have_content("Next")
+        is_expected.to have_content("1")
+        is_expected.to have_link("2")
+      end
+
+      within("ul.pseud.index.group") do
+        is_expected.to have_content("Agrajag")
+        is_expected.to have_content("Betelgeuse")
+        is_expected.to have_content("Slartibartfast")
+        is_expected.not_to have_content("Zaphod")
+      end
+    end
+
+    it "doesn't show pagination UI when there are fewer pseuds than the limit" do
+      visit "/users/Agrajag/pseuds"
+
+      is_expected.to_not have_css("ol.pagination")
+    end
+  end
+end
--- a/app/views/pseuds/index.html.erb
+++ b/app/views/pseuds/index.html.erb
@@ -11,6 +11,10 @@
 <% end %>
 <!--/subnav-->
 
+<!--subnav-->
+<%= will_paginate @pseuds %>
+<!--/subnav-->
+
 <!--main content-->
 <h3 class="landmark heading"><%= ts("List of Pseuds") %></h3>
 <ul class="pseud index group">
@@ -20,4 +24,6 @@
 </ul>
 <!--/content-->
 
-
+<!--subnav-->
+<%= will_paginate @pseuds %>
+<!--/subnav-->


spec/requests/pseuds_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+      FactoryBot.create(:pseud, user: user, name: "Slartibartfast")
+      FactoryBot.create(:pseud, user: user, name: "Betelgeuse")
+    end
+
+    it "shows only the configured maximum number of pseuds" do
+      allow(ArchiveConfig).to receive(:ITEMS_PER_PAGE).and_return(3)
+
+      visit "/users/Agrajag/pseuds"
+
+      within("ol.pagination") do
```
redsummernight, line 20: This test is failing because of the 2 subnav sections:
```
Capybara::Ambiguous:
  Ambiguous match, found 2 elements matching visible css "ol.pagination"
```

spec/requests/pseuds_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    before do
+      FactoryBot.create(:pseud, user: user, name: "Zaphod")
+      FactoryBot.create(:pseud, user: user, name: "Slartibartfast")
+      FactoryBot.create(:pseud, user: user, name: "Betelgeuse")
+    end
+
+    it "shows only the configured maximum number of pseuds" do
+      allow(ArchiveConfig).to receive(:ITEMS_PER_PAGE).and_return(3)
+
+      visit "/users/Agrajag/pseuds"
```
redsummernight, line 18: Similar to the thread at https://github.com/otwcode/otwarchive/pull/3813#discussion_r414959949, should we start using request specs here?

spec/requests/pseuds_spec.rb
```diff
@@ -0,0 +1,42 @@
+require 'spec_helper'
```
redsummernight, line 1: Can you use double quotes here?

app/controllers/pseuds_controller.rb
```diff
@@ -14,7 +14,7 @@ def load_user
   # GET /pseuds.xml
   def index
     if @user
-      @pseuds = @user.pseuds
+      @pseuds = @user.pseuds.paginate(page: params[:page], per_page: ArchiveConfig.ITEMS_PER_PAGE)
```
redsummernight, line 17: Can we leave out the `per_page` argument, since `ArchiveConfig.ITEMS_PER_PAGE` is the default value?

spec/requests/pseuds_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+
+describe "Pseuds", type: :request do
+  subject { page }
+
+  describe "show" do
+    let(:user) { FactoryBot.create(:user, login: "Agrajag") }
+
+    before do
+      FactoryBot.create(:pseud, user: user, name: "Zaphod")
```
redsummernight, line 10: Can you [leave out `FactoryBot.`](https://github.com/otwcode/otwarchive/wiki/Automated-Testing#Our_RSpec_Standards)?

spec/requests/pseuds_spec.rb
```diff
@@ -0,0 +1,42 @@
+require 'spec_helper'
+
+describe "Pseuds", type: :request do
```
redsummernight, line 3: Can you [leave out `type: :request`](https://github.com/otwcode/otwarchive/blob/f5e775b6e87e822e03baa67abd9e9e516e4ab7f3/spec/spec_helper.rb#L120)?

app/controllers/pseuds_controller.rb
CristinaRO: Oh, I suppose we can! From a consistency POV, we're already using it like this (almost) everywhere, but I guess there's no point introducing redundancy when writing _new_ code.

(I had to dig to find out where the defaults are actually set, and found them in `config/initializers/gem-plugin_config/will_paginate_config.rb`, for anyone wondering the same now.)

spec/requests/pseuds_spec.rb
CristinaRO: Is the question about request versus view specs versus other kind of RSpec integration test, or is it about writing cucumber features instead?

I can't remember why I didn't make them controller specs - I think I tried and couldn't surmount some kind of obstacle, possibly with checking existence of things on the generated page.

I know why I didn't want to make them cucumber specs: they're hard to write, hard to think about, hard to reuse steps for, and hard to check if a scenario isn't already tested somewhere. I think cucumber let us, or even encouraged us, to write a very unwieldy test suite, slow and full of redundancy. :/ I still have cucumber burnout, and every time I have to interact with them it takes me a month to be able to come back to the project. 🤷‍♀️

redsummernight: I think we should wait for #3813 to be completed, then update this PR to add new tests to the "Many pseuds" Cucumber scenario, while removing `spec/requests/pseuds_spec.rb`:

- Generally we don't test pagination behavior (except for comments where we add a lot of custom behavior). We trust that `will_paginate` works, and instead focus on checking the order of list of things (which you've done in #3813).
- We resolved the other thread by using controller/model specs + cucumber tests, so let's not add a new type of spec here. Request specs shouldn't have access to Capybara matchers in the first place, that's something we add to every type of spec and I plan to remove (in #3936):

  https://github.com/otwcode/otwarchive/blob/bedbbef8c43585d97b5a56062355921e4c8f5fa7/spec/spec_helper.rb#L40

sarken:
We merged an updated version of this, so closing.

PR #3815, opened 2020-04-23
AO3-5907 Move share work modal loading to a controller action

melinath:
# Pull Request Checklist

  ^ Technically the functionality is the same as before but there wasn't a test for anonymous usage of this feature so I added one.

## Issue

https://otwarchive.atlassian.net/browse/AO3-5907

## Purpose

This PR moves loading of "share work" modal into a separate action on the WorksController so that rendering of the share options can happen when users press the share button rather than when they view a work. Part of the motivation for this change is to limit the exposure of user behavior to external sites like twitter and tumblr by not sending http requests to them on every work page load.

Note: In order to accomplish the privacy goal, I moved loading of the twitter widgets javascript into the new share action's view as well. This means that the js would load every time a user shares rather than once per page, but it's the simplest way to ensure that it's _not_ loaded every time the page renders (and my guess would be that people usually only share things once per page load?)

Potential future work could convert bookmark sharing to use a similar asynchronous pattern.

## Testing Instructions

1. View a work that can be shared as a logged-in user with javascript enabled and verify that the "Share" button works.
2. View a work that can be shared as a logged-out user with javascript enabled and verify that the "Share" button works.
3. View a work that can be shared with javascript disabled and verify that the "Share" button is not displayed.
4. Verify in Developer Tools > Network Traffic that loading a work detail page does not cause any http requests to twitter or tumblr.
5. Verify that visiting the share URL (/works/:id/share) for a work in an unrevealed collection does not give away any information about the work.
6. View a work that can be shared as a logged-out user, open the share modal, close the share modal, and log in. Verify that after login you are viewing the work.
7. View a multi-chapter work and verify that the "share" button works

## Credit

Stephen Burrows

he/him


sarken:
It looks like there are two test failures here that need to be addressed:
* features/works/work_delete.feature:91:in `And I should see "Characters: Sam Winchester, Dean Winchester"`
* features/works/work_delete.feature:91:in `And I should see "Characters: Sam Winchester, Dean Winchester"`

They were passing before because the share box actually has the work information out in that format. They're failing now because the meta at the top of the work doesn't have a nice, comma-separated list of characters: the commas are just inserted with CSS, so the list is actually like "Characters: Sam WinchesterDean Winchester."

tickinginstant:
The failures in `features/collections/collection_anonymity.feature` are similar -- the works have the proper anonymous/non-anonymous flags, but when viewing a work, the text "by Anonymous" or "by [username]" isn't visible anywhere except the share box. You probably need to try changing several parts of these two step definitions, which were written to rely on that text from the share box:
https://github.com/otwcode/otwarchive/blob/86d9cfcb409ff316ef2ffa9ad2188241e151abf9/features/step_definitions/collection_steps.rb#L215-L235
(Note that when viewing a work blurb, the text "by Anonymous" or "by [username]" is visible. So some of those steps can stay as they are, but some of them have to be updated.)

melinath:
Thank you so much! I hadn't gotten around to looking into the failures yet and that context helps a lot. :-)

melinath:
Okay, there are still a couple of test failures left due to a mismatch in server hosts – the expected value is using example.com but the output value is using the capybara server's address. I'm going to try to figure out a good solution but am definitely open to suggestions.

tickinginstant:
I don't think you should be checking the share box in those steps. If you want to add a separate test to check that the text of the share box doesn't include the author name when the work is anonymous, that would be useful, but I think making that step depend on the share box is overcomplicating it. The only reason those steps checked the share box in the first place was by accident, I'm pretty sure.

melinath:
@tickinginstant Sure, I'd be happy to split that apart. Just to make sure I'm on the same page – if I'm splitting off the test for the share button from the test for the collection page, I would be inclined to also split off the test for the work detail page, and either group it with the test for the share button or have three separate tests. For example, if I split that step into three steps:

```
  Scenario: The moderator can reveal all the authors in an anonymous collection
    Given I have the anonymous collection "Anonymous Hugs"
      And "second_user" subscribes to author "first_user"
      And the user "third_user" exists and is activated
    When I am logged in as "first_user"
      And I post the work "Old Snippet" to the collection "Anonymous Hugs" as a gift for "third_user"
    Then "third_user" should be emailed
      And the email to "third_user" should not contain "first_user"
      And the email to "third_user" should contain "Anonymous"
    When subscription notifications are sent
    Then "second_user" should not be emailed
    When all emails have been delivered
      And I am logged in as "moderator"
      And I reveal authors for "Anonymous Hugs"
    Then the author of "Old Snippet" should be publicly visible on the work detail page
      And the author of "Old Snippet" should be publicly visible on the work detail share modal
      And the author of "Old Snippet" should be publicly visible on the collection page
    When subscription notifications are sent
    Then "second_user" should be emailed
      And the email to "second_user" should contain "first_user"
      And "third_user" should not be emailed
```

Is that sort of what you had in mind?

tickinginstant:
No, I meant a separate scenario, not a separate step.

tickinginstant:
I don't think you need copies of each of these scenarios to check the share box, particularly since nothing in there is cached. A single scenario (separate because javascript) checking what the share box looks like for an anonymous work should be fine, and then you could just revert all of the other changes to `collection_anonymity.feature`.

melinath:
@tickinginstant That makes sense, thanks! I've moved the test for the share modal into its own scenario in c263008. I'm happy to leave the existing test cases alone.

features/step_definitions/collection_steps.rb
```diff
@@ -212,21 +212,26 @@
   page.should have_content("Anonymous [#{authors}]")
 end
 
+# Scenarios using this step need the @javascript tag
```
tickinginstant, line 215: This is no longer true, right?

features/step_definitions/collection_steps.rb
```diff
@@ -212,21 +212,26 @@
...
+  step %{I should see "#{byline}" within "title"}
+  step %{I should see "#{byline}" within ".byline"}
   if work.collections.first
     visit collection_path(work.collections.first)
-    page.should have_content("#{title} by #{work.users.first.pseuds.first.byline}")
+    page.should have_content("#{title} by #{byline}")
   end
 end
 
+# Scenarios using this step need the @javascript tag
```
tickinginstant, line 228: And here, too?

features/step_definitions/collection_steps.rb
melinath: correct, thanks! resolved in 186364e72d5a1338b46a584378e481f0b2142afb

features/step_definitions/collection_steps.rb
melinath: yep, resolved in 186364e72d5a1338b46a584378e481f0b2142afb

melinath:
Sorry, I had some technical difficulties today so wasn't able to make as much progress as I was hoping.

Re: blocking the share URL for unrevealed works, that makes sense to me! I'll get to that shortly.

Re: the delay, I'm not sure there's anything I can do about that and still fulfill the aspect of this ticket related to not making requests to twitter until a user clicks the share button. (Unless I misunderstood the goal / parameters of the ticket.) But I'm open to suggestions!

melinath:
re: the tweet button size, I've been thinking about it and one solution would be to put something in there and force it to be the right size, so that the reflow is less obvious? Alternately I could probably hook into the modal loading javascript to allow defining a javascript file to load after rendering the new content but before displaying it to the user - but that feels like a pretty complicated thing to support if this is the only use case.

melinath:
^ Gonna have a fast-follow for that commit because the share button does get displayed on unrevealed works for admins and authors; my thought is that probably unrevealed works should never be shared, so it would be better to have the share action not return any information and to not display the button at all.

sarken:
I'm not seeing a noticeable delay with the loading of the Twitter button, but I agree that using a placeholder sounds like the better of the two options. I also agree that there's no reason to show the Share option to anyone on an unrevealed work.

Two other things I did notice:

1. If I'm looking at a multichapter work in Chapter by Chapter mode, and I press "Share," the spinner starts turning, but the modal never opens. When I look at the console, there is a 404 on the share resource. This doesn't happen if I am looking at the same work in Entire Work mode. It doesn't appear to matter if I'm logged in or out.
2. If I'm logged out, and I have gone to a work, opened the share modal, and then I close it and log in, I get redirected to /works/###/share upon signing in. The same thing happens if I log out immediately after looking at a work's share modal. I don't really think we want users going to that page at all, given it is layout-less, but getting redirected there is particularly likely to generate Support tickets.

melinath:
Thanks for noticing those issues, @sarken! I've addressed 1 in c445eda  and 2 in 71fc5ba.

Regarding the question of allowing users to access this view –  I think it would be hard for someone to end up there at this point, but I could do something similar to TagsController#show_hidden and return a redirect with a flash error for non-js requests?

Regarding the placeholder for the tweet button - I did some an experiment to set the width and height of the current placeholder to the correct dimensions (by using the css related to [AO3-5845](https://otwarchive.atlassian.net/browse/AO3-5845) in sandbox.css and adding `display: inline-block`) but it's actually worse, because you get two size changes instead of just one: a size change when the placeholder is removed, and a second one when the actual button is added.

However, I can avoid a change in modal size if I put a new class on the enclosing `<li>` tag and modify its size instead. I've taken a stab at this in 8aef760 - lmk if I should be putting that CSS somewhere else! My only concern is that any solution that hardcodes the expected size of the twitter widget is potentially fragile since they could change that at any time – but OTOH a lot of people are probably expecting its size to remain constant.

app/views/share/_share.html.erb
```diff
@@ -10,7 +10,7 @@
   <ul>
     
     <% # Twitter share %>
-    <li>
+    <li class="twitter_share_li">
```
sarken, line 13: Can we just call this class `twitter`? We prefer single-word names for our classes -- the `twitter-share-button` class down there is an outlier, required by Twitter itself to make the button work.

melinath: sure, made that change in e9ae4465782a845fe82682c0d5749cfacaecda15

sarken:
Hey, Stephen, just checking in! Are you working on adding the code to redirect non-JS requests to /share?

melinath:
Thank you for the ping! I lost track of the fact that I had something left to do here :-) I'll have a look in the morning.

melinath:
I made that change but now I'm again running into the issue that I was running into before (and had forgotten about): a mismatch in server hosts, where the expected hostname generated by the feature is different than the hostname on the running server that capybara is interacting with.

https://github.com/otwcode/otwarchive/pull/3815#issuecomment-620283620

Previously I "resolved" the issue by tweaking what was being tested; ;I'll see if I can find a real solution.

melinath:
I made a new work step for cucumber tests that allows dynamically inserting the current capybara server domain into the expected result text - let me know if y'all would prefer a different resolution :-)

melinath:
The test failure seems to be because the iframe isn't loading (or not loading fast enough?) The test is passing locally, but I'd rather have it passing on travis as well. Definitely open to suggestions; otherwise, I'll have a look tomorrow.

melinath:
The previous test for the tweet share button was flakey locally as well; this version seems to pass consistently.

melinath:
@sarken This should be ready for review again! Now with flash errors on non-js access!

features/step_definitions/work_steps.rb
```diff
@@ -693,3 +693,25 @@
 Then /^the Remove Me As Chapter Co-Creator option should not be on the ([\d]+)(?:st|nd|rd|th) chapter$/ do |chapter_number|
   step %{I should not see "Remove Me As Chapter Co-Creator" within "ul#sortable_chapter_list > li:nth-of-type(#{chapter_number})"}
 end
+
+Then /^the share modal should contain ("|')(.*?)\1$/ do |quote, text|
```
sarken, line 697: Instead of doing this, it might be good to set the test environment's `action_controller.default_url_options ` to `{ host: "http://www.example.com", port: nil }`, much like we do for Action Mailer: https://github.com/otwcode/otwarchive/blob/a18a4302b6f9e0a95f53dd0c2e7a54f05f4acedc/config/environments/test.rb#L55

We can probably even make things a bit more DRY by doing this:
```
  config.action_controller.default_url_options = { host: "http://www.example.com", port: nil }
  config.action_mailer.default_url_options = config.action_controller.default_url_options
```

(Thanks for the suggestion, redsummernight!)

sarken: We might need to use `http://test.host` instead of `http://www.example.com` since Travis seems rather attached to the former.

app/controllers/works_controller.rb
```diff
@@ -223,6 +225,26 @@ def show
...
+  def share
+    if request.xhr?
+      if @work.unrevealed?
+        render template: "errors/404", status: :not_found
+      else
+        render layout: false
+      end
+    else
+      # Avoid getting an unstyled page if JavaScript is disabled
+      flash[:error] = ts('Sorry, you need to have JavaScript enabled for this.')
```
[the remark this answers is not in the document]
melinath: Resolved in 822cde7509801076aa025a4449dbf48b46aa5cef.

app/controllers/works_controller.rb
```diff
@@ -223,6 +225,26 @@ def show
...
+      if request.env['HTTP_REFERER']
+        redirect_to(request.env['HTTP_REFERER'] || root_path)
+      else
+        # else branch needed to deal with bots, which don't have a referer
+        redirect_to '/'
```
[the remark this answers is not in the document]
melinath: Resolved in e2d2fd3983873eb6740404a483256150640bd9b8.

features/step_definitions/work_steps.rb
melinath: Thank you! ✨ 🧙 ✨  that is the magical incantation I was looking for. I've taken a stab with example.com in b9bbcd4610e2c724b0cc138bb41392f1832d0f11. If that doesn't work I'll try test.host.

melinath:
hm, neither example.com nor test.host Just Works (tm) - I will try to find some time to look into this further but am definitely open to suggestions if there's an obvious solution to people more familiar with ruby than I am.

sarken:
I think I've figured it out!

I saw there were a lot fewer failures with the `www.example.com` approach, so I switched back to that and worked from there.

To deal with the failures in the controller specs, I set the `@request.host` as described in [this StackOverflow thread](https://stackoverflow.com/questions/598933/how-do-i-change-the-default-www-example-com-domain-for-testing-in-rails). 

That just left the feature tests. Most of the failures were because we weren't using `_path` for the redirects in the controllers. The one exception was actually the test reflecting the existence of [AO3-5877](https://otwarchive.atlassian.net/browse/AO3-5877). (I have no earthly idea why it wasn't failing before.)

[My branch is here](https://github.com/otwcode/otwarchive/compare/master...sarken:AO3-5907?expand=1) if you'd like to merge these fixes into your branch. (I can push to your branch, but I didn't want to presume.) The last commit might need a few tweaks -- while I ran the failing tests locally and they were passing, I went to bed before the full build completed on Travis.

melinath:
Okay, I think that should do it - it looks like the tests are passing on your [last commit](https://github.com/sarken/otwarchive/commit/488b860a666aac631e99c8de3d089e8ce0816d20) and the only change I made after that was to fix a houndci issue.

Thanks, @sarken! I don't know when or if I would've been able to figure that all out at this stage in my ruby learning. :-)

melinath:
Tests are officially passing on travis-ci!

sarken:
You're welcome! And yay, they pass!

I'm going to leave this set to "Coder Has Actioned Review" because I'd like someone else who didn't write the tests to check them, but I'm pretty sure it's good to go. Thanks so much!

PR #3817, opened 2020-04-26
AO3-5936 Enable experimental caching in feature tests

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5936

## Purpose

- Add AdminSetting.default for use in both initializers and testing setups.
- Fix reading history tests, because with experimental caching, timestamps of readings are not relative.
- Simplify step definitions for setting up AdminSetting.first, remove step for turning on wrangling (it's on by default).
- Always clear cached results of "Latest Works" and "Latest Bookmarks" before checking them, as these caches never get expired by any user action.

## Testing Instructions

None.

### DIFF
--- a/features/other_a/reading.feature
+++ b/features/other_a/reading.feature
@@ -1,74 +1,75 @@
 @users
 Feature: Reading count
 
-  Scenario: only see own reading history
-    Given the following activated user exists
-    | login          | password   |
-    | first_reader        | password   |
+  Scenario: A user can only see their own reading history
+
+  Given the following activated user exists
+    | login        |
+    | first_reader |
   When I am logged in as "second_reader"
     And I go to first_reader's reading page
-    Then I should see "Sorry, you don't have permission"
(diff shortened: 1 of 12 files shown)


features/other_a/reading.feature
```diff
@@ -78,7 +79,7 @@ Feature: Reading count
       And I follow "First work"
       And I am on testuser's works page
       And I follow "second work"
-      And I am on testuser2's works page
+      And I am on testuser2 works page
```
sarken, line 82: Why remove the possessive?

redsummernight: The uneven number of single quotes in this file is screwing up the syntax highlighting. Currently after this line, it's all blue like a quoted string that never ends:

https://github.com/otwcode/otwarchive/blob/17307a8a6294d2af7fab882ea5aa14eb786d51af/features/other_a/reading.feature#L81

sarken: @ god why

PR #3819, opened 2020-04-27
AO3-5667 Update wording of some troubleshooting options, and change permissions on "Reindex Work" option.

tickinginstant:
## Issue

## Purpose

- Rename the troubleshooting option from "Fix Meta Tags" to "Fix Metatags."
- Change "meta-tags/sub-tags" to "metatags/subtags" in the "Fix Tag Associations" description.
- Change the "Reindex Work" option so that it's only available to admins.

## Testing Instructions

Make sure that the troubleshooting page for tags looks right.

### DIFF
--- a/config/locales/views/en.yml
+++ b/config/locales/views/en.yml
@@ -53,5 +53,5 @@ en:
         title: "Fix Tag Associations"
         description: "Try to delete invalid associations involving this tag. (e.g. Parents of the wrong type, children of the wrong type, duplicate parents/children/meta-tags/sub-tags, etc.)"
       fix_meta_tags:
-        title: "Fix Meta Tags"
+        title: "Fix Metatags"
         description: "Try to recalculate inherited metatags. Use this option if this tag has grandparent metatags (i.e. a metatag with its own metatag), and the works aren't showing up in the grandparent's listing. You might also try this if you've removed a metatag, but the works are still showing up in the former metatag's listing."
--- a/config/locales/views/en.yml
+++ b/config/locales/views/en.yml
@@ -51,7 +51,7 @@ en:
         description: "Try to recalculate some counts associated with this tag (both filter counts and taggings counts). Keep in mind that the tag counts listed in the fandom lists include unrevealed works, so if the count is only a little off it may just be unrevealed works."
       fix_associations:
         title: "Fix Tag Associations"
-        description: "Try to delete invalid associations involving this tag. (e.g. Parents of the wrong type, children of the wrong type, duplicate parents/children/meta-tags/sub-tags, etc.)"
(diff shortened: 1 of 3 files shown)


sarken:
Hookay, got  the new troubleshoot option wording worked out with Tag Wrangling! If you could make these changes, that would be great:

```
Fix Tag Associations
Try to delete invalid associations involving this tag (e.g., parents of the wrong type, children of the wrong type, duplicate parents/children/metatags/subtags, etc.). Most of these impossible relationships appear on tag landing pages.

Fix Tag Counts
Try to recalculate some counts associated with this tag (both filter counts and taggings counts). Keep in mind that the tag counts listed in the fandom lists include unrevealed works and drafts, so if the count is only a little off it may just be unrevealed works or drafts.

Fix Metatags
Try to recalculate inherited metatags. Use this option if this tag has grandparent metatags (i.e., a metatag with its own metatag), and the works aren't showing up in the grandparent's listing. You might also try this if you've removed a metatag, but the works are still showing up in the former metatag's listing. If the works still don't go away, ask a staffer to run "Update Tag  Filters" on this tag.

Update Tag Filters
Recalculate filters for all works directly tagged with this tag or one of its synonyms. Does not include subtags. Use this option if the filters include old canonical tags that have been renamed or old parent tags that works aren't tagged with, or if the tag counts are wildly off. Also use this option if the works from a synonym are not showing on the works page of the canonical within a half hour of the tag being synned. Keep in mind that the tags listed in the filters will never be 100% correct.

Reindex Tag
Reindex this tag and all related works, bookmarks, series, pseuds, and external works. Use this option if none of the others have worked. It will not fix autocomplete issues.
```

config/locales/views/en.yml
```diff
@@ -42,16 +42,16 @@ en:
...
+        description: "Try to delete invalid associations involving this tag (e.g., parents of the wrong type, children of the wrong type, duplicate parents/children/metatags/subtags, etc.). Most of these impossible relationships appear on tag landing pages."
       fix_meta_tags:
-        title: "Fix Meta Tags"
-        description: "Try to recalculate inherited metatags. Use this option if this tag has grandparent metatags (i.e. a metatag with its own metatag), and the works aren't showing up in the grandparent's listing. You might also try this if you've removed a metatag, but the works are still showing up in the former metatag's listing."
+        title: "Fix Metatags"
+        description: "Try to recalculate inherited metatags. Use this option if this tag has grandparent metatags (i.e., a metatag with its own metatag), and the works aren't showing up in the grandparent's listing. You might also try this if you've removed a metatag, but the works are still showing up in the former metatag's listing. If the works still don't go away, ask a staffer to run \"Update Tag  Filters\" on this tag."
```
sarken, line 57: There's an extra space in `Update Tag  Filters`, which is definitely my fault.

PR #3820, opened 2020-04-27
AO3-5721 Delete challenge without JavaScript

teyla86:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5721

## Purpose

A challenge mod can delete a challenge either on the Challenge Settings page, or on the profile of the collection that the challenge belongs to. Both buttons ("Delete Challenge") trigger a JavaScript function that shows a warning prompt and, after user confirmation, deletes the challenge. Users who have JavaScript disabled, however, receive a 404 error. They can't delete their challenge.

The suggested change adds a fallback option for users who have JavaScript disabled. These users get routed to a page where they can confirm that they want to delete the challenge by submitting a form. Users with JavaScript enabled can still delete via JavaScript prompt.

This pull request further contains added steps in the Gift Exchange and the Prompt Meme Features to test this new/ changed functionality.

I forgot to add the ticket number to two of my commit messages. :( Sorry!

## Testing Instructions

1. Turn off JavaScript in your browser
2. Create a collection with a Prompt Meme challenge
3. Go to your new collection's profile page and press "Delete Challenge"

Expectation: You should be routed to a page whose URL ends with /confirm_delete where you're asked if you are sure that you want to delete your challenge.

4. Press "Yes, Delete Challenge".

Expectation: You should now be routed to your collection page with a notice at the top saying "Challenge settings were deleted".

5. In your collection, create another Prompt Meme challenge
6. After saving your challenge settings, return to your Challenge Settings page
7. In the top right corner, press "Delete Challenge"

Expectation: see above after step 3

8. Continue with step 4 until you see "Challenge settings were deleted."

The same should be tested for Gift Exchange challenges.

Finally, challenge deletion with JavaScript enabled should be re-tested. Do the following for both challenge types, Prompt Memes and Gift Exchanges. For each type of challenge, test both the button on the collection profile page and the button on the Challenge Settings page:

1. Make sure JavaScript is enabled in your browser
2. Create a collection and a challenge, Prompt Meme or Gift Exchange
3. On the collection's profile page or on the Challenge Settings page, press "Delete Challenge"

Expectation: You should receive a prompt/ pop-up asking you if you want to delete your challenge.

4. Press "Okay"

Expectation: You should be routed to the collection page and see a notice "Challenge settings were deleted"

## Credit

teyla

she or they


app/views/challenge/shared/_challenge_form_delete.html.erb
```diff
@@ -1,12 +1,15 @@
...
-                      collection_prompt_meme_path(@collection)
-                    else
-                      collection_gift_exchange_path(@collection)
-                    end,
-                    data: {confirm: ts('Are you certain you want to delete the challenge from this collection? All signups, assignments, and settings will be lost. (Works will remain in the collection.)')},
-                    :method => :delete %></li>
-  <% end %>
-</ul>
+<% 
+  # This is a list element that should always be rendered within an ul with classes "navigation actions" and role "menu".
```
sarken, line 2: The `menu` role is actually a relic that should be removed -- the way we use it is not correct at all. Could you remove them from the `ul` elements and update this note?

And, actually, it would probably be better to move the `li` tags out of this file in case we ever want to use the link outside a `ul`. That's what we do with [the subscribe button](https://github.com/otwcode/otwarchive/blob/master/app/views/subscriptions/_form.html.erb#L1), for example.

features/gift_exchanges/challenge_giftexchange.feature
```diff
@@ -532,11 +532,13 @@ Feature: Gift Exchange Challenge
         And the email should not contain "Warnings:"
         And the email should not contain "Optional Tags:"
 
-  Scenario: A mod can delete a gift exchange and all the assignments and
+  Scenario: A mod can delete a gift exchange without needing Javascript and all the assignments and
   sign-ups will be deleted with it, but the collection will remain
     Given everyone has their assignments for "Bad Gift Exchange"
       And I am logged in as "mod1"
     When I delete the challenge "Bad Gift Exchange"
+	Then I should see "Are you certain you want to delete the challenge from this collection? All sign-ups, assignments, and settings will be lost. (Works will remain in the collection)."
```
sarken, line 540: Could you fix the indenting on these lines? They should line up with the "When" and "Then" directly above and below them.

app/views/challenge/prompt_meme/confirm_delete.html.erb
```diff
@@ -0,0 +1,6 @@
+<!--Descriptive page name, messages and instructions-->
+<h2 class="heading"><%= ts("Delete Prompt Meme") %></h2>
+
+<!--main content-->
+<%= render "challenge/shared/challenge_form_confirm_delete" %>
+<!--/content-->
```
sarken, line 6: Could you add a final newline to this file, `app/views/challenge/shared/_challenge_form_confirm_delete.html.erb`, and `app/views/challenge/shared/_challenge_form_delete.html.erb`? (Hound would ordinarily comment asking for that, but it doesn't check views.)

app/views/challenge/shared/_challenge_form_confirm_delete.html.erb
```diff
@@ -0,0 +1,14 @@
+<%= form_for(@collection, :url => 
+  if @collection.prompt_meme?
+    collection_prompt_meme_path(@collection)
+  else
+    collection_gift_exchange_path(@collection)
+  end,
+  :html => {:method => :delete}) do |f| %>
+  <p class="caution notice">
+    <%= ts("Are you certain you want to <strong><em>delete</em></strong> the challenge from this collection? All sign-ups, assignments, and settings will be lost. (Works will remain in the collection).").html_safe %>
```
sarken, line 9: Hmm. This text is different from what's in the issue, but I see why: you went with text that matches the existing confirmation text. That makes sense! But it could use a few tweaks. 

- We're currently inconsistent about this throughout the code, but let's go with "Are you sure" over "certain," if only because we use "sure" in [the fallback text](https://github.com/otwcode/otwarchive/blob/17307a8a6294d2af7fab882ea5aa14eb786d51af/public/javascripts/application.js#L421) as well.
- Instead of "This collection," we should still reference the collection name. We generally try to do that when you're taken to a separate deletion page, just to make it a bit clearer what you're deleting.
- It should say "Works and bookmarks will remain in the collection," and the period should be inside the parentheses.

app/views/challenge/shared/_challenge_form_confirm_delete.html.erb
```diff
@@ -0,0 +1,14 @@
+<%= form_for(@collection, :url => 
+  if @collection.prompt_meme?
+    collection_prompt_meme_path(@collection)
+  else
+    collection_gift_exchange_path(@collection)
+  end,
+  :html => {:method => :delete}) do |f| %>
```
sarken, line 7: Could you add spaces inside the braces here, e.g. `:html => { :method => :delete }`?

app/views/challenge/shared/_challenge_form_delete.html.erb
```diff
@@ -1,12 +1,15 @@
...
+%>
+
+<li>
+  <%= link_to ts("Delete Challenge"),
+    if @collection.prompt_meme?
+      confirm_delete_collection_prompt_meme_path(@collection)
+    else
+      confirm_delete_collection_gift_exchange_path(@collection)
+    end,
+    data: {confirm: ts("Are you certain you want to delete the challenge from this collection? All sign-ups, assignments, and settings will be lost. (Works will remain in the collection).")}
```
sarken, line 13: Could you add spaces inside the braces here, switch from "certain" to "sure," and update the end of the message to say, "(Works and bookmarks will remain in the collection.)"?

config/routes.rb
```diff
@@ -453,8 +453,16 @@
     end
     resources :requests, controller: "challenge_requests"
     # challenge types
-    resource :gift_exchange, controller: 'challenge/gift_exchange'
-    resource :prompt_meme, controller: 'challenge/prompt_meme'
+    resource :gift_exchange, controller: 'challenge/gift_exchange' do
```
sarken, line 456: Could you switch to double quotes  as long as you're here? (Just on these lines; no need to worry about the rest of the file.)

app/views/challenge/shared/_challenge_form_delete.html.erb
teyla86: As per the commit messages, I did the following:

- remove role attributes from ul
- remove li from shared form and add in the views them where necessary
- add final newlines
- change confirmation text for confirm_delete page, Javascript prompt
- fix indent on steps in gift exchange feature
- change confirmation text in features for all challenge types
- add spaces inside braces on forms for _delete and _confirm_delete
- switch to double quotes in routes.rb

I hope I got them all!

app/views/challenge/shared/_challenge_form_confirm_delete.html.erb
```diff
@@ -0,0 +1,14 @@
+<%= form_for(@collection, :url => 
+  if @collection.prompt_meme?
+    collection_prompt_meme_path(@collection)
+  else
+    collection_gift_exchange_path(@collection)
+  end,
+  :html => { :method => :delete }) do |f| %>
+  <p class="caution notice">
+    <%= ts("Are you sure you want to <strong><em>delete</em></strong> the challenge from the collection %{title}? All sign-ups, assignments, and settings will be lost. (Works and bookmarks will remain in the collection.)", :title => @collection.title).html_safe %>
```
sarken, line 9: Could you update the syntax at the end to `title: @collection.title`?

app/views/collection_profile/show.html.erb
```diff
@@ -96,16 +96,10 @@
...
-          <li><%= link_to ts('Delete Challenge'),
-                          if @collection.prompt_meme?
-                            collection_prompt_meme_path(@collection)
-                          else
-                            collection_gift_exchange_path(@collection)
-                          end,
-                          data: {confirm: ts('Are you certain you want to delete the challenge from this collection? All sign-ups, assignments, and settings will be lost. (Works will remain in the collection.)')},                          :method => :delete %>
-          </li>
-        <% end %>
+		<% if @collection.challenge %>
```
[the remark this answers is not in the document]
teyla86: Fixed! Sorry, my editor likes to put in random tabs ... I've updated my preferences now. Fixed the syntax in the other document, too.

PR #3821, opened 2020-04-29
AO3-5518 Create roles for admins

AncyentMariner:
# Pull Request Checklist

* [ ] Have you added the [JIRA](https://otwarchive.atlassian.net) issue number as the *first* thing in your pull request title (eg: `AO3-1234 Fix thing`)

## Issue

https://otwarchive.atlassian.net/browse/AO3-5518

## Purpose
Adds roles for admins, and adjusts some functionality of the site to respond to those admin roles.

## Testing Instructions
Check the pages referenced in the tickets to see if the admin's functionality is limited in the ways the ticket suggests. Add/ remove permissions to admin users to test these checks are being enforced.

## References
https://otwarchive.atlassian.net/browse/AO3-5518

## Credit

(If you have an account on JIRA, please include the same name in the "Full name" field on your JIRA profile,
so we can find you and assign you the issues you're working on.)

(If you do not fill this in, we will use your GitHub account name and will use "they" to avoid making assumptions about your gender.)


AncyentMariner:
@redsummernight This branch is reverted and ready for review, whenever somebody has the chance. :)

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20190611212339) do
+ActiveRecord::Schema.define(version: 20200423205608) do
```
sarken, line 13: We prefer not to update the schema file in the same pull as migrations -- instead, we'll run the migration on production and get a schema dump from there.

spec/requests/admin_settings_request_spec.rb
```diff
@@ -0,0 +1,25 @@
+require "spec_helper"
+
+describe "Admin settings", type: :request do
+  before(:all) do
```
redsummernight, line 4: Can you use a let definition instead of an instance variable?

spec/requests/admin_settings_request_spec.rb
```diff
@@ -0,0 +1,25 @@
+require "spec_helper"
+
+describe "Admin settings", type: :request do
```
redsummernight, line 3: You can omit `type: :request`, since the test is in `spec/requests` and we use `infer_spec_type_from_file_location!`.

app/policies/application_policy.rb
```diff
@@ -0,0 +1,53 @@
+class ApplicationPolicy
+  attr_reader :user, :record
+
+  def initialize(user, record)
+    @user = user
+    @record = record
+  end
+
+  def index?
```
redsummernight, line 9: Should we define all the controller actions upfront? Not every controller will have all 7 of these actions. I'd prefer we define only the ones we need in the policy subclasses, but I can be convinced otherwise.

app/policies/admin_user_policy.rb
```diff
@@ -0,0 +1,53 @@
+class AdminUserPolicy < ApplicationPolicy
+  USER_SEARCH_ROLES = %w(superadmin policy_and_abuse open_doors support tag_wrangling).freeze
+  USER_ACTION_ROLES = %w(superadmin policy_and_abuse).freeze
+
+  def index?
+    can_search_users?
```
redsummernight, line 6: Should we prefer defining methods that call `can_search_users?`, or creating aliases for `can_search_users?` like in `AdminPostPolicy`?

app/policies/admin_setting_policy.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  def update?
+    user_has_roles?(SETTINGS_ROLES)
+  end
+
+  def permitted_attributes
+    if user.roles.include?('superadmin')
+      [
+        :account_creation_enabled, :invite_from_queue_enabled, :invite_from_queue_number,
+        :invite_from_queue_frequency, :days_to_purge_unactivated, :last_updated_by,
```
redsummernight, line 16: This is an existing issue, but we could fix it here: can you remove `last_updated_by` from the list of permitted attributes? As a result, you'll have to remove the hidden field `last_updated_by` from the settings form, then set `last_updated_by` to `current_admin` in the settings controller.

app/policies/admin_user_policy.rb
```diff
@@ -0,0 +1,53 @@
...
+  def permitted_user_params
+    if user_has_roles?(USER_ACTION_ROLES)
+      %w(roles email)
+    else
+      %w(roles)
+    end
+  end
+
+  def permitted_management_params
+    basics = %w(user_login next_of_kin_name next_of_kin_email admin_note)
```
redsummernight, line 46: Can you rename `basics` to `params`?

app/models/admin.rb
```diff
@@ -15,4 +18,21 @@ class Admin < ApplicationRecord
...
+  def allowed_roles
+    if roles && (roles - VALID_ROLES).present?
+      errors.add(:roles, :invalid)
+    end
+  end
+
+  ### PERMISSIONS ###
+
+  def admin_post_access?
+    AdminPostPolicy.can_post?(self)
```
redsummernight, line 32: Does `policy(AdminPost).can_post?` work here? If so, we can remove `AdminPostPolicy.can_post?`.

spec/requests/admin_settings_request_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+  before(:all) do
+    @admin = create(:admin)
+  end
+  describe "#index" do
+    it "denies random access" do
+      get admin_settings_path
+      expect(response).to redirect_to root_url
+    end
+    it "denies random admin access" do
```
redsummernight, line 12: Can you add an empty line between `it` blocks?

features/step_definitions/admin_steps.rb
```diff
@@ -52,6 +59,17 @@ def update_settings(settings)
   step("I should see \"Successfully logged in\"")
 end
 
+Given /^I am logged in as superadmin$/ do
+  step("I have an AdminSetting")
+  step("I am logged out")
+  admin = Admin.find_by(login: "superadmin") || FactoryBot.create(:superadmin)
```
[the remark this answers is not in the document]
redsummernight: Can you remove the assignment?

app/models/admin.rb
```diff
@@ -1,5 +1,8 @@
 class Admin < ApplicationRecord
   include ActiveModel::ForbiddenAttributesProtection
+  VALID_ROLES = %w(superadmin communications translation tag_wrangling docs support policy_and_abuse open_doors)
```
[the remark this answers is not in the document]
redsummernight: I accidentally resolved this comment, can you freeze the list here?

app/policies/application_policy.rb
sarken: I lean slightly toward doing it this way because it's the default for pundit, but much like you, I can be swayed. I think I know why you'd rather define the necessary ones in policy subclasses, but just in case I'm wrong, could you elaborate?

redsummernight: Not every controller will have all 7 of these actions, so some of these will be unused. Even in controllers with most of these actions, these will not necessarily be useful defaults.

For example: AdminPostsController has an index action, but `AdminPostPolicy#index?` is not used, because we don't lock down anything there. For the 5 actions that do use `authorize`, we end up overriding the ApplicationPolicy methods in AdminPostPolicy anyway.

That said, I'm also ok with having superclass defaults. The defaults block access to everything, so it's not a big deal if we miss overriding one of them.

app/models/admin.rb
AncyentMariner: It appears that the "policy" method is only available in views and controllers, so we can't call that here.

app/policies/application_policy.rb
AncyentMariner: So just to be clear, are you asking for any changes here?

sarken: I'm checking right now to see if anyone else on the committee has strong feelings about this and I'll let you know where we end up.

sarken: I think we'll leave this as it is. Thanks!

db/structure.sql
```diff
@@ -2,7 +2,7 @@
 /*!40101 SET @OLD_CHARACTER_SET_CLIENT=@@CHARACTER_SET_CLIENT */;
 /*!40101 SET @OLD_CHARACTER_SET_RESULTS=@@CHARACTER_SET_RESULTS */;
 /*!40101 SET @OLD_COLLATION_CONNECTION=@@COLLATION_CONNECTION */;
-/*!40101 SET NAMES utf8mb4 */;
+/*!50503 SET NAMES utf8mb4 */;
```
sarken, line 5: Can you revert this file?

app/controllers/application_controller.rb
```diff
@@ -264,6 +270,8 @@ def access_denied(options ={})
       destination = options[:redirect].blank? ? user_path(current_user) : options[:redirect]
       flash[:error] = ts "Sorry, you don't have permission to access the page you were trying to reach."
       redirect_to destination
+    elsif logged_in_as_admin? && options[:admin_access_denied]
```
redsummernight, line 273: Can you remove this `elsif` now that `rescue_from Pundit::NotAuthorizedError` no longer calls `access_denied`? Same for the options parameter of `admin_only_access_denied`.

app/models/admin.rb
redsummernight: Gotcha. In that case, can we use `policy` in the views, and remove the `### PERMISSIONS ###` methods here?

app/controllers/application_controller.rb
```diff
@@ -272,8 +278,8 @@ def access_denied(options ={})
     false
   end
 
-  def admin_only_access_denied
-    flash[:error] = ts("I'm sorry, only an admin can look at that area.")
+  def admin_only_access_denied()
```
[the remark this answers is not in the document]
redsummernight: Can you remove the parentheses?

app/views/works/show.html.erb
```diff
@@ -24,19 +24,16 @@
 <!--/descriptions-->
 
 <!--subnav-->
-<% if logged_in_as_admin? %>
-  <%= render :partial => 'admin/admin_options', :locals => {:item => @work} %>
+<% if AdminModerationPolicy.can_edit_works?(current_admin) %>
```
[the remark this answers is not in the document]
redsummernight: **1. Instance vs class**

I think the semantically correct way of using Pundit is, test against a specific instance if you have it, otherwise test against the class. We test against the class anyway in some cases, as a shortcut, because we know admins have the same permissions for all instances of that class (we should be consistent and catch such cases in code review, though).

**2. `policy` vs static methods taking `current_admin`**

The reason we're using `AdminModerationPolicy#can_edit_works?` is that we cannot use `policy(@work)`. Frustratingly, `policy` in views relies on name matching to figure out which policy class to pick, and unlike `authorize`, you don't have the option to specify a policy class. We also cannot [declare the policy class](https://github.com/varvet/pundit/#manually-specifying-policy-classes) in AdminModerationPolicy, because it was set up to cover both works and comments.

I stopped here when I last reviewed, but you have a good point about consistency. Maybe we should break AdminModerationPolicy into separate policies for works and comments.

features/step_definitions/admin_steps.rb
```diff
@@ -38,6 +38,13 @@ def update_settings(settings)
   end
 end
 
+Given /^I am logged in as admin with role "([^\"]*)"$/ do |role|
+  step("I am logged in as an admin")
+  admin = Admin.find_by(login: "testadmin")
+  admin.roles << role
```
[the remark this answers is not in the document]
redsummernight: Yeah, since the database gets truncated after every test.

app/views/admin_posts/index.html.erb
```diff
@@ -1,5 +1,5 @@
 <div class="news admin home">
-  <% if logged_in_as_admin? %>
+  <% if policy(@admin_posts).can_post? %>
```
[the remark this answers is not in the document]
redsummernight: It is. `policy` uses the argument to figure out which policy class to use (Pundit [can auto-detect classes from relation objects](https://github.com/varvet/pundit/blob/df96d2ae6bcf28991c1501d5ff0bde4c42aa4acd/lib/pundit/policy_finder.rb#L62-L63)) then send the argument to the policy class (AdminPostPolicy, which doesn't care if it gets an object or a collection of objects because we set it up that way).

app/views/admin/_header.html.erb
```diff
@@ -19,7 +19,9 @@
     <%= link_to ts("Admin Posts", key: "header"), admin_posts_path %>
     <ul class="menu" role="menu">
       <li><%= span_if_current ts("AO3 News", key: "header"), admin_posts_path %></li>
-      <li><%= span_if_current ts("Post AO3 News", key: "header"), new_admin_post_path %></li>
+      <% if AdminPostPolicy.can_post?(current_admin) %>
```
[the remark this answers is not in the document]
redsummernight: Can you use `policy(AdminPost).can_post?` and remove `AdminPostPolicy#can_post?`?

features/step_definitions/admin_steps.rb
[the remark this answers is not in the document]
redsummernight: I should specify, we have database_cleaner use truncation for JS tests, transaction for non-JS tests, but in both cases the roles get reset.

app/views/works/show.html.erb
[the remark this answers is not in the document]
CristinaRO: Ahh, thank you, I get it now (I think).

app/policies/admin_moderation_policy.rb
```diff
@@ -0,0 +1,32 @@
...
+  def set_spam?
+    can_moderate_record?
+  end
+
+  def destroy?
+    can_moderate_record?
+  end
+
+  def can_moderate_record?
+    record.is_a?(Comment) ? can_moderate_comment? : can_moderate_content?
```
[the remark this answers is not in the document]
sarken: I'm actually going to ask a really silly question: what is `can_moderate_comments?` for? I assumed it was for the ability to delete comments or mark them as spam, but that seems to be covered by `CommentPolicy` in #3836.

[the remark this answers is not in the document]
AncyentMariner: @sarken  Yeah, good call, again. These are never used on comments, just on user_creations and chapters. It looks like this is a relic from the initial work on the roles tasks. Thanks for catching that; I overlooked it.

features/step_definitions/admin_steps.rb
```diff
@@ -78,23 +106,23 @@
...
+  step(%{I am logged in as admin with role "communications"})
   step("I make an admin post")
   step("I am logged out as an admin")
 end
 
 Given /^the fannish next of kin "([^\"]*)" for the user "([^\"]*)"$/ do |kin, user|
   step %{the user "#{kin}" exists and is activated}
   step %{the user "#{user}" exists and is activated}
-  step %{I am logged in as an admin}
+  step %{I am logged in as superadmin}
```
sarken, line 125: Could we use the `policy_and_abuse` role here and on lines 134 and 144?

app/policies/admin_setting_policy.rb
```diff
@@ -0,0 +1,26 @@
...
+  def index?
+    update?
+  end
+
+  def update?
+    user_has_roles?(SETTINGS_ROLES)
+  end
+
+  def permitted_attributes
+    if user.roles.include?('superadmin')
```
sarken, line 13: Could we use double quotes here and on line 22?

features/step_definitions/admin_steps.rb
```diff
@@ -41,30 +61,38 @@
...
+  step("I am logged in as superadmin")
+  visit(admin_settings_path)
+  step(%{I uncheck "Turn off tag wrangling for non-admins"})
+  step(%{I press "Update"})
   step("I am logged out as an admin")
 end
 
 Given /^the support form is disabled and its text field set to "Please don't contact us"$/ do
-  step("I am logged in as an admin")
+  step("I am logged in as superadmin")
```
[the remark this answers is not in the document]
sarken: Ah, I see the issue: support isn't given this ability until #3836

PR #3824, opened 2020-05-02
AO3-5667 Allow metatags to be upgraded from inherited to direct, and make removing associations synchronous.

tickinginstant:
## Issue

## Purpose

- If there's a metatag tree where A is the metatag of B is the metatag of C, and you decide that you want C to be the direct subtag of A instead of the direct subtag of B, the current way to do that involves removing B as C's metatag, waiting for the inherited association with C to be cleaned up, and then adding C as a direct metatag. This PR makes it possible to upgrade inherited metatags to direct metatags, to make the process easier for tag wranglers.
- In the current code, when you check the box to remove a metatag, it doesn't immediately remove the metatag. Instead, it queues up a job to remove the metatag. This means that the metatag remains on the edit page until the user refreshes (or, if the utilities queue is long, after the the queue has finished running and the user refreshes). This PR makes the action synchronous, and modifies the `remove_association` function to ensure that removing the association is a fast operation.

## Testing Instructions

1. Pick three canonical tags of the same type, with works under each.
2. Wrangle them into a chain, so that A is the metatag of B, and B is the metatag of C.
3. Wait until the works of C show up under both B and A to make sure that this worked.
4. Edit tag C.
5. Check the box to remove B as a metatag.
6. Fill in the name of A in the "Add MetaTag" box.
7. Press "Save changes."
8. Look at the MetaTags section on the edit page to make sure the changes went through.
9. View the tag, and check the "Metatags" section to make sure the changes are reflected there.
10. Wait about 10-20 minutes.
11. Check the works pages for each of the tags to make sure that A has works from all three tags, but B and C each have works only from their own tag.

### DIFF
--- a/app/models/meta_tagging.rb
+++ b/app/models/meta_tagging.rb
@@ -7,7 +7,7 @@ class MetaTagging < ApplicationRecord
   validates_presence_of :meta_tag, :sub_tag, message: "does not exist."
   validates_uniqueness_of :meta_tag_id,
                           scope: :sub_tag_id,
-                          message: "has already been added."
+                          message: "has already been added (possibly as an indirect metatag)."
 
   after_create :expire_caching
   after_destroy :expire_caching
--- a/app/models/meta_tagging.rb
+++ b/app/models/meta_tagging.rb
@@ -7,7 +7,7 @@ class MetaTagging < ApplicationRecord
   validates_presence_of :meta_tag, :sub_tag, message: "does not exist."
   validates_uniqueness_of :meta_tag_id,
                           scope: :sub_tag_id,
-                          message: "has already been added (possibly as an indirect metatag)."
+                          message: "has already been added."
 
   after_create :expire_caching
   after_destroy :expire_caching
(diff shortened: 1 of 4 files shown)


tickinginstant:
Yes, this will only make inherited metatags into direct meta tags when the wrangler adds that metatag through the wrangling interface. (Or when synning a canonical that has direct metatags/subtags, which also must have been added through the wrangling interface.)

Inherited metatags will still be recalculated asynchronously, since that callback is unchanged:
https://github.com/otwcode/otwarchive/blob/173138f65a04dfe0aa098754d764f43a92c7b30a/app/models/meta_tagging.rb#L34-L36
(I could try to make them be recalculated synchronously, to resolve one of the remaining issues where you might have to wait to add a metatag: When you have a pre-existing chain A -> B -> C, and you want to remove B -> C and/or A -> B and connect C -> A instead. You'd get the "A metatag can't be its own grandparent" message until the asynchronous process finished. But I don't know how long the inherited calculation might take on a tag like "Alternate Universe," with hundreds or thousands of subtags.)

redsummernight:
> But I don't know how long the inherited calculation might take on a tag like "Alternate Universe," with hundreds or thousands of subtags.

Yeah let's not do that. I was more checking my understanding of the code than actually requesting changes. This looks good to me.

PR #3827, opened 2020-05-09
AO3-5938 Links in bookmark blurb stats have their bottom borders cut off (on Windows)

niconicosette:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5938

## Purpose

makes bottom borders visible again under stats (follow issue link for pictures)

## Testing Instructions

on a windows device (or emulator) go to https://test.archiveofourown.org/users/testy/bookmarks and check that all stats links for each work (chapter count, bookmark count, collection count, comment count) have borders underneath them. try making your browser window smaller and check that borders under these links don't disappear when they're pushed down to a second line.

### DIFF
--- a/public/stylesheets/site/2.0/13-group-blurb.css
+++ b/public/stylesheets/site/2.0/13-group-blurb.css
@@ -357,7 +357,7 @@ blurbs on the manage collection items pages, mostly reseting styles inherited fr
   clear: both;
 }
 
-.bookmark .blurb dl.stats {
+.bookmark dl.stats {
   margin-bottom: 0.643em;
 }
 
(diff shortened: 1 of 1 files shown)


public/stylesheets/site/2.0/13-group-blurb.css
```diff
@@ -357,6 +357,10 @@ blurbs on the manage collection items pages, mostly reseting styles inherited fr
   clear: both;
 }
 
+.bookmark .blurb dl.stats {
```
sarken, line 360: Does just  `.bookmark dl.stats` happen to work, just to bring it more in line with the other bookmark selectors?

PR #3828, opened 2020-05-09
AO3-5822 Reverse margins for lists written in RTL languages

niconicosette:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5822

## Purpose

adds a `1.75` margin to the right of lists written in RTL languages and changes the left margin to `0`

## Testing Instructions

go to https://archiveofourown.org/admin_posts/13294 and scroll down to the lists in the post. check that there is a margin to the right of each bullet. check that there isn't any left margin by making your browser window tiny and zooming in to create two lines of text per bullet. you can compare how the reversed margins look to the [english version of the same post](https://archiveofourown.org/admin_posts/12035).

### DIFF
--- a/public/stylesheets/site/2.0/21-userstuff.css
+++ b/public/stylesheets/site/2.0/21-userstuff.css
@@ -31,7 +31,6 @@
   display: block;
   line-height: 1.5;
   margin-block-start: 1.25em;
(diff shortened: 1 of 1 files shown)


sarken (approved):
Looks good, thanks!

PR #3831, opened 2020-05-12
AO3-5948 Reduce some cookies' expiration date to one year

palindromordnilap:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5948

## Purpose

Three cookies were set as "permanent", that being an expiration date twenty years in the future. The European ePrivacy directive recommends that cookies expire at most a year after being set -- while not strictly *required*, it's definitely a more reasonable delay.

## Testing Instructions

Use your browser's developer tools to check the expiration date of the `accepted_tos`, `user_credentials` and `admin_credentials` cookies after having, respectively, accepted the ToS while logged out, logged in as a regular user, and logged in as an admin.

## Credit

Alix R, she/her

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -83,7 +83,7 @@ class ApplicationController < ActionController::Base
     if logged_in_as_admin?
       # if we are logged in as an admin and we don't have the admin_credentials
       # set then set that cookie
-      cookies.permanent[:admin_credentials] = 1 unless cookies[:admin_credentials]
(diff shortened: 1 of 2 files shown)


app/controllers/application_controller.rb
```diff
@@ -465,11 +465,11 @@ def convert_date(hash, date_symbol_or_string)
   def valid_sort_column(param, model='work')
     allowed = []
     if model.to_s.downcase == 'work'
-      allowed = ['author', 'title', 'date', 'created_at', 'word_count', 'hit_count']
+      allowed = %w(author title date created_at word_count hit_count)
```
sarken, line 468: Thank you!

PR #3834, opened 2020-05-16
AO3-5685 Use DELETE requests for logging out

palindromordnilap:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5685

## Purpose

Replaces the GET request for logging out with a DELETE request. Works the exact same way as before for users with JavaScript, but due to technical limitations, users with JS disabled clicking the "Log Out" link are brought to a logout confirmation form. (Thank redsummernight for that fix!)

## Testing Instructions

See Jira ticket.

## References

This issue blocks [AO3-5949](https://otwarchive.atlassian.net/browse/AO3-5949), and is related to [AO3-5953](https://otwarchive.atlassian.net/browse/AO3-5953).

## Credit

Alix R, she/her


config/routes.rb
```diff
@@ -470,8 +475,8 @@
 
   namespace :api do
     namespace :v2 do
-      resources :bookmarks, only: [:create], defaults: { format: :json }
-      resources :works, only: [:create], defaults: { format: :json }
+      resources :bookmarks, only: [:create], defaults: {format: :json}
+      resources :works, only: [:create], defaults: {format: :json}
```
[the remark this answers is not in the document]
palindromordnilap: Looks like every problem there is my IDE's formatter disagreeing with Hound.

palindromordnilap:
I take it we're _not_ using the driver I thought we were using. Hold on.

features/support/http.rb
```diff
@@ -0,0 +1,9 @@
+module HttpHelpers
+  def http_delete(path)
```
redsummernight, line 2: Switching the driver [will not retain session or cookie information from previous requests](https://stackoverflow.com/questions/9228831/how-to-make-capybara-do-a-delete-request-in-a-cucumber-feature#comment60760638_19244681), so you're not logging out as the user you want logged out.

An alternative we could try is to [keep the JS driver and make the DELETE request using jQuery](https://stackoverflow.com/a/36580783). We already have a function for the waiting part:

https://github.com/otwcode/otwarchive/blob/df7cac7f4103b8fa1a42cfe509097ad4cb71e9e2/features/support/wait_for_ajax.rb#L3-L7

Can you add to the existing file `features/support/capybara.rb` instead of adding `features/support/http.rb`?

app/controllers/users/sessions_controller.rb
```diff
@@ -4,6 +4,7 @@ class Users::SessionsController < Devise::SessionsController
   before_action :admin_logout_required
   skip_before_action :store_location
 
+  # POST /users/login
```
redsummernight, line 7: You don't have to add these comments, btw. They're just holdovers from early auto-generated Rails scaffolding, and we don't use them consistently.

app/views/admin/sessions/confirm_logout.html.erb
```diff
@@ -0,0 +1,10 @@
+<h3 class="heading"><%= ts('Log Out as Admin') %></h3>
+
+<div id='logoutconfirm' class=>
```
redsummernight, line 3: Can you use double quotes in views, and remove empty "class" attributes?

app/controllers/users/sessions_controller.rb
palindromordnilap: I know, I just thought it was nice to have as documentation for the next time we'd have to make changes to this controller.

features/support/http.rb
palindromordnilap: This causes the following error in tests:

    undefined method `zero?' for nil:NilClass (NoMethodError)

    ./features/support/wait_for_ajax.rb:10:in `finished_all_ajax_requests?'

I'm not sure if this is an issue with my Vagrant install (PhantomJS initially didn't work at all, I had to run `export QT_QPA_PLATFORM=offscreen` before anything else), or if it's something that actually broke.

palindromordnilap:
I'm running into another weird issue with the Cucumber tests. The step "I log out", even with the rack_test driver (which shouldn't support JS), fails because the link works right away, without the confirmation prompt. I manually confirmed that it works as intended, but the automated tests still fail.

palindromordnilap:
Looks like this [is in fact](https://stackoverflow.com/a/43242438) a rack_test feature, if the `respect_data_method` option is set to `true`.

I think I'll just get rid of the button press step, but should I make the option `false` instead?

features/support/http.rb
palindromordnilap: Done, but the Ajax request fails and throws this error now:

    ReferenceError: Can't find variable: jQuery

palindromordnilap: Alright, turns out this was because the method didn't check if a page was loaded before calling jQuery functions. I just fixed that.

features/step_definitions/user_steps.rb
```diff
@@ -139,8 +139,8 @@
 end
 
 Given /^I am logged out$/ do
-  visit destroy_user_session_path
-  visit destroy_admin_session_path
+  http_delete destroy_user_session_path
+  http_delete destroy_admin_session_path
 end
 
 Given /^I log out$/ do
```
redsummernight, line 146: Does this step need to be updated to also click the confirm log out button, if the test is not using JS?

features/step_definitions/user_steps.rb
```diff
@@ -139,8 +139,8 @@
 end
 
 Given /^I am logged out$/ do
-  visit destroy_user_session_path
-  visit destroy_admin_session_path
+  http_delete destroy_user_session_path
```
redsummernight, line 142: Per @tickinginstant's comment elsewhere:

> I think it might be better to just replace the "I am logged out" step with something that resets cookies. Any of the `has_selector?` methods are going to cause delays one way or another.

We can have "I am logged out" use `page.driver.clear_cookies`.

app/views/admin/sessions/confirm_logout.html.erb
```diff
@@ -0,0 +1,10 @@
+<h3 class="heading"><%= ts("Log Out as Admin") %></h3>
+
+<div id="logoutconfirm">
```
redsummernight, line 3: Do we need wrapper elements with IDs around the confirm forms?

palindromordnilap: I'm fairly sure we don't, this seems to be a leftover from something I was trying to debug earlier that I then forgot to get rid of. Thanks for pointing it out!

features/step_definitions/user_steps.rb
palindromordnilap: It actually doesn't, this sends a DELETE request, which is what the confirm button does rather than the logout link. I'll be replacing this step with something that uses cookies anyway, as per your comment a bit lower.

features/step_definitions/user_steps.rb
```diff
@@ -191,13 +191,13 @@
...
-  step (%{I follow "Profile"})
-  step (%{I follow "Delete My Account"})
+  step(%{I go to #{login}\'s user page})
+  step(%{I follow "Profile"})
+  step(%{I follow "Delete My Account"})
 end
 
 When /^I try to delete my account$/ do
-  step (%{I try to delete my account as #{DEFAULT_USER}})
+  step(%{I try to delete my account as #{DEFAULT_USER}})
```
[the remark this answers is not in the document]
palindromordnilap: ... Well that's gonna take a while.

palindromordnilap:
From tickinginstant:
> I don't think it's a good idea to delete the `user_credentials`/`admin_credentials` cookies. They only exist for caching reasons, and deleting them will result in a redirect to the `/lost_cookie` page (which will indirectly cause a logout, but only in a roundabout way). The cookies actually used for authentication are either signed or encrypted, and stored in the session cookie (which makes them difficult to delete in isolation).

palindromordnilap:
Due to RL stuff, I unfortunately won't be able to work on this anymore. Closing the PR and unassigning the Jira issue.

sarken:
Thank you for letting us know!

PR #3846, opened 2020-06-19
AO3-5007 Only allow taggable.rb to create new noncanonicals for user-defined tags

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5007

## Purpose

taggable.rb currently finds or creates any tags it receives. This is fine for user defined tags, but not so much for other kinds. This will silently fail (that is, the work will post successfully, just without the tag) unless the tag is the only one of a required type.

## Testing Instructions

Refer to Jira. 

## References

#2916 cleans up some of the existing tags created this way.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -43,15 +43,15 @@ RSpec.configure do |config|
     DatabaseCleaner.strategy = :transaction
     DatabaseCleaner.clean
     Indexer.all.map(&:prepare_for_testing)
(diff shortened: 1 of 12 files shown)


spec/spec_helper.rb
```diff
@@ -43,6 +43,15 @@
     DatabaseCleaner.strategy = :transaction
     DatabaseCleaner.clean
     Indexer.all.map(&:prepare_for_testing)
+    ArchiveWarning.find_or_create_by_name(ArchiveConfig.WARNING_CHAN_TAG_NAME).update(canonical: true)
```
sarken, line 46: In an ideal world, we'd just create all the default tags in an initializer. This is not an ideal world.

PR #3854, opened 2020-06-29
AO3-5973 Make the WorksController forward-compatible.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5973

## Purpose

The view `works/standard_form` in #3852 has radio buttons for the name `work[comment_permissions]`. But at the moment, the `WorksController` doesn't include `comment_permissions` in the permitted `work_params`. So if someone accesses the form from a worker running the new code, then submits the form to a worker that's still running the old code, the user's choice of `comment_permissions` would disappear and be reverted to the default.

The deploy involves restarting all the workers over the course of an hour, so we need to plan around the new form being submitted to the old controller, and vice versa. So this PR adds `comment_permissions` to the list of permitted `work_params`.

In addition, the new form in #3852 uses `enable_all`, `disable_anon`, and `disable_all` as the form values, relying on the enum to convert those values. So we need to back-port the enum from #3852.

## Testing Instructions

See the issue description.

## References

This modifies the code introduced in #3851 so that it's forward-compatible with #3852.

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -147,6 +147,12 @@ class Work < ApplicationRecord
   # Run Taggable#check_for_invalid_tags as a validation.
   validate :check_for_invalid_tags
 
+  enum comment_permissions: {
+    enable_all: 0,
+    disable_anon: 1,
+    disable_all: 2
+  }, _suffix: :comments
+
   ########################################################################
   # HOOKS
   # These are methods that run before/after saves and updates to ensure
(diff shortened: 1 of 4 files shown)


app/models/work.rb
```diff
@@ -922,14 +928,25 @@ def update_stat_counter
...
+    return unless has_attribute?(:comment_permissions)
 
-    write_attribute(:anon_commenting_disabled, value)
+    write_attribute(:comment_permissions, value)
+
+    # Map the special value back to an integer, and write it to the
+    # anon_commenting_disabled column so that if we do have to revert, we can
+    # go back to using the anon_commenting_disabled column without data loss.
+    write_attribute(:anon_commenting_disabled,
+                    Work.comment_permissions[comment_permissions])
```
redsummernight, line 939: Nice, any non-zero value becomes true.

PR #3858, opened 2020-07-11
AO3-5885 Rework restrictions on modifying user creations + grant support admins edit access

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5885

## Purpose

* Renames the policy class used for controlling which admin roles can do things to user creations.
* Allow for more granular control of actions performed on user creations, e.g. in case PAC wants to allow new committee members to hide user creations but not delete them, and update the admin options partial to reflect this.
* Grant admins with the support role permission to edit user creations.
* Update admin options partial to use path helpers.
* Internationalize admin options partial.

## Testing Instructions

Refer to Jira.


app/policies/user_creation_policy.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  def self.can_mark_creations_spam?(user)
+    self.new(user, nil).can_mark_creations_spam?
+  end
+
+  def can_destroy_creations?
+    user_has_roles?(DESTROY_ROLES)
+  end
+
+  # Currently applies to editing ExternalWorks and the tags or laguage of Works.
```
CristinaRO, line 30: Minor typo: "laguage"

PR #3860, opened 2020-07-13
AO3-5883 Show restricted admin settings as disabled fields

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5883

## Purpose

Admins with any of the settings roles (superadmin, support, policy_and_abuse, or tag_wrangling) should be able to see all settings, even if they can update some of them.

We can disable the inputs of the settings they cannot update, instead of hiding them entirely.

## Testing Instructions

See issue.


app/controllers/admin/settings_controller.rb
```diff
@@ -1,39 +1,31 @@
...
 
   # PUT /admin_settings/1
-  # PUT /admin_settings/1.xml
   def update
     authorize @admin_setting
-    authorized_params = policy(@admin_setting).verify_permitted_params(admin_setting_params) 
 
-    if authorized_params == true && @admin_setting.update(admin_setting_params)
-      flash[:notice] = ts("Archive settings were successfully updated.")
+    if policy(@admin_setting).unpermitted_attributes(params).present?
```
redsummernight, line 12: `unpermitted_attributes` exists only so we can have this error message. Since we keep restricted settings disabled, the only time admins can easily update settings they don't have permissions to, is when they have the settings form open, but their roles have been changed in the database.

I'm not attached to this error message. I can remove it if we think it's OK to silently ignore settings not allowed by the current admin's roles.

sarken: Yeah, let's get rid of it.

app/views/admin/settings/index.html.erb
```diff
@@ -1,103 +1,97 @@
...
-<h2 class="heading"><%= ts('Archive Settings') %></h2>
+<h2 class="heading"><%= t(".heading") %></h2>
 <!--/descriptions-->
 
 <!--main content-->
-<%= form_for @admin_setting, :url => admin_setting_path(@admin_setting), html: { method: "put", class: "manage" } do |f| %>
+<%= form_for @admin_setting, html: { class: "manage" } do |f| %>
   <%= error_messages_for @admin_setting %>
+  <fieldset>
+    <legend><%= t(".landmark.account_and_invitations") %></legend>
```
sarken, line 9: Could we use `legend` instead of `landmark` in the key? 

Landmark headings are only necessary on forms where the legend is hidden -- screen readers won't read the legend text, but they _will_ read the heading. (At least, this was the case 10 years ago; it may have changed.) Our [rule](https://github.com/otwcode/otwarchive/blob/156720eb90c391d183f598edf531f940b0fd5fe3/public/stylesheets/site/2.0/07-interactions.css#L554) is that if there are more than 3 field sets on a form, the `verbose` class should be added to the form, which will make the legend visible (and available to screen readers). 

This form has more than 3 field sets, so we will one day add the `verbose` class to the form and yanking out the unnecessary landmark headings. When that day comes, it makes more sense to have the key use `legend` in its name than the name of the thing we removed.

features/step_definitions/admin_steps.rb
```diff
@@ -138,15 +138,15 @@
 Given /^the support form is disabled and its text field set to "Please don't contact us"$/ do
   step("I am logged in as support_admin")
   visit(admin_settings_path)
-  check("Turn off support form")
+  check("Turn off Support form")
```
sarken, line 141: Hmm, should it be? 🤔 I think we're using "support' in the generic sense here, i.e. it's the support and feedback form, as opposed to the Support Committee form.

PR #3861, opened 2020-07-13
AO3-5886 Separate permissions in CommentPolicy and add notice to comments marked as spam

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5886

## Purpose

* Divide CommentPolicy so we can have one set of roles with deletion permissions and another set with spam marking permissions.
* Add a "This comment has been marked as spam" banner to comments marked as spam, since the only way to tell used to be if the "Not Spam" button appeared in place of the "Spam" button. This wasn't ideal and got even less ideal when we started hiding the spam buttons from admins without marking as spam permissions.

Note: The tests for the "marked as spam" banner might fail intermittently. I had it fail once, but a whole bunch of retries passed. 🤷 

## Testing Instructions

Refer to Jira.


app/policies/comment_policy.rb
```diff
@@ -1,15 +1,24 @@
 class CommentPolicy < ApplicationPolicy
-  DELETE_COMMENT_ROLES = %w(superadmin policy_and_abuse communications support).freeze
+  DESTROY_ROLES = %w(superadmin policy_and_abuse communications support).freeze
+  SPAM_ROLES = %w(superadmin policy_and_abuse communications support).freeze
 
-  def self.can_delete_comment?(user)
-    self.new(user, nil).can_delete_comment?
+  def self.can_destroy_comment?(user)
```
redsummernight, line 5: The 2 static methods here are unused. I think we only need the static methods in the case of UserCreationPolicy (which handles multiple models and cannot be looked up by the `policy` view helper).

Everything else looks good.

app/policies/comment_policy.rb
```diff
@@ -1,15 +1,16 @@
...
   end
-    
-  def can_delete_comment?
-    user_has_roles?(DELETE_COMMENT_ROLES)
+
+  def can_mark_comment_spam?
+    user_has_roles?(SPAM_ROLES)
   end
 
   alias destroy? can_delete_comment?
```
redsummernight, line 13: I saw some failed tests, this should be `can_destroy_comment?`.

PR #3863, opened 2020-07-14
AO3-5793 Docker support

tommilligan:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5793

## Purpose

This PR adds support for development/testing in Docker.

## Testing Instructions

I'll update the [wiki page](https://github.com/otwcode/otwarchive/wiki/Docker) once this PR is approved. But in essence, to test this you should:

- change the default password from `change_me` to a new secret in `docker-compose.yml` and `config/database.yml`
- run `./docker/init.sh` to build and setup docker fixtures (images, database migrations, etc.)
- run `docker-compose up web` to run the Archive locally

## References

See issue #3661 for previous discussion and contributions.

## Credit

Tom Milligan

they/them


app/controllers/bookmarks_controller.rb
```diff
@@ -138,7 +138,7 @@ def index
           search = BookmarkSearchForm.new(show_private: false, show_restricted: false, sort_column: 'created_at')
           results = search.search_results
           flash_search_warnings(results)
-          @bookmarks = results.to_a
+          results.to_a
```
tommilligan, line 141: I don't understand why this was changed, however if the tests continue to pass I'm happy with it.

docker/init.cmd
```diff
@@ -0,0 +1,13 @@
+:: Change directory to root of the repo
+cd /D "%~dp0\.."
```
tommilligan, line 2: I haven't tested this line as I don't have a windows setup. If someone could test this on Windows I'd be grateful.

app/controllers/bookmarks_controller.rb
zz9pzza: I kept on looking at this bit of code when I was wondering why the memcached tests were failing and the assignment at the end of the cache block is just wrong 

The old code is

>        @bookmarks = Rails.cache.fetch("bookmarks/index/latest/v2_true", expires_in: ArchiveConfig.SECONDS_UNTIL_BOOKMARK_INDEX_EXPIRE.seconds) do
>           search = BookmarkSearchForm.new(show_private: false, show_restricted: false, sort_column: 'created_at')
>           results = search.search_results
>           flash_search_warnings(results)
>            @bookmarks =results.to_a
>         end

Rails.cache.fetch evaluates the block on a cache miss, at the end of this block we assign @bookmarks  ( which is a side effect that will not happen on a cache hit ), the result of the Rails.cache.fetch is the last evaulation, which is the result of the assignment which is the value being assigned. That value is then assigned to @bookmarks again.

Basically once I understood the code it made me cry and I couldn't help but want to make it simpler. Its not related to this and I should have done it in another merge request but they are identical in execution but this makes me want to cry less.

docker/Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM ruby:2.6.5
+
+# Install additional packages
+RUN apt-get update             && \
+    apt-get install -y            \
+        calibre                   \
+        default-mysql-client      \
+        phantomjs
```
zz9pzza, line 8: I missed out zip wkhtmltopdf which are needed for downloads sorry

docker/init.cmd
zz9pzza: will do

zz9pzza:
If the machine running the code is not exposing ports than changing the password is not needed ?

zz9pzza:
Is this missing docker-compose.yml ?

docker/Dockerfile
tommilligan: No worries, added and squashed in

app/controllers/bookmarks_controller.rb
tommilligan: :joy: Thanks for explaining! That makes sense. Happy to include it in this PR if you are, I'll just leave it as a separate commit so it's clear it's not related to the other changes we're making

tommilligan:
@zz9pzza thanks, ready for review again. To address your other points:

- db password: this is something that was already in the wiki. If you think it's unnecessary, I'm happy to remove it, however I think this guards against people running these services without a firewall setup (in which case, they will definitely be exposing their ports on the local network). FWIW I port scanned my machine from another device on my network while running the stack, and got:
- yes, the `docker-compose.yml` was unstaged while I was rebasing - now added.

docker/init.cmd
zz9pzza: > PS C:\Users\james\RubymineProjects\otwarchive\docker> .\init.cmd
> 
> C:\Users\james\RubymineProjects\otwarchive\docker>cd /D "C:\Users\james\RubymineProjects\otwarchive\docker\\.."
> 
> C:\Users\james\RubymineProjects\otwarchive>copy docker\database.yml config\database.yml
>         1 file(s) copied.
> 
> C:\Users\james\RubymineProjects\otwarchive>copy docker\redis.yml config\redis.yml
>         1 file(s) copied.
> 
> C:\Users\james\RubymineProjects\otwarchive>copy docker\local.yml config\local.yml
>         1 file(s) copied.
> 
> C:\Users\james\RubymineProjects\otwarchive>docker-compose up -d
> Creating network "otwarchive_default" with the default driver
> Creating otwa_mc    ... done
> Creating otwa_es    ... done
> Creating otwa_redis ... done
> Creating otwa_db    ... done
> Creating otwa_web   ... done
> 
> C:\Users\james\RubymineProjects\otwarchive>timeout 60
> 
> Waiting for  0 seconds, press a key to continue ...
> Starting otwa_redis ... done
> Starting otwa_mc    ... done
> Starting otwa_es    ... done
> Dropped database 'otwarchive_development'
> Dropped database 'otwarchive_test'
> Created database 'otwarchive_development'
> Created database 'otwarchive_test'
> -- create_table("abuse_reports", {:id=>:integer, :force=>:cascade, :options=>"ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci ROW_FORMAT=DYNAMIC"})

tommilligan: That looks good. Yeah I just wasn't sure the working directory would change correctly so the file copies would fail, but LGTM there

docker/redis.yml
```diff
@@ -0,0 +1,15 @@
...
+  development: redis:6379
+redis_general:
+  test: redis:6379
+  development: redis:6379
+redis_rollout:
+  test: redis:6379
+  development: redis:6379
+redis_autocomplete:
+  test: redis:6379
+  development: redis:6379
```
zz9pzza, line 15: Sorry this needs:

> 
> redis_hits:
>   test: redis:6379
>   development: redis:6379

docker-compose.yml
```diff
@@ -0,0 +1,66 @@
...
+    volumes:
+      - esdata1:/usr/share/elasticsearch/data:rw
+    environment:
+      - transport.host=localhost
+      - bootstrap.memory_lock=false
+      - "ES_JAVA_OPTS=-Xms1500m -Xmx1500m"
+      - discovery.type=single-node
+  mc:
+    image: memcached:1.5
+    container_name: otwa_mc
```
zz9pzza, line 46: Is there an advantage to having the container name, I just tried to bring two instances up so I could run cucumber tests and rspec at the same time and failed due to having a duplicate container name ?

tommilligan: No, I don't think there is any advantage here. I don't use `container_name` in any of my other work. Will remove.

docker-compose.yml
```diff
@@ -0,0 +1,66 @@
...
+    ports:
+      - "11211:11211"
+  web:
+    build:
+      context: .
+      dockerfile: ./docker/Dockerfile
+    container_name: otwa_web
+    environment:
+      - DOCKER=1
```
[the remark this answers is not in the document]
tommilligan: As you say, I believe it's only used in this test fixture. I'm happy to rename this variable to something like `NO_START_REDIS=1` to make it's purpose clearer.

docker/database.yml
```diff
@@ -0,0 +1,22 @@
+defaults: &defaults
```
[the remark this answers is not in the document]
tommilligan: Yes, that's fine. The only file that can't be moved is `docker-compose.yml` as it is expected in the project root.

If you want, I've worked around this before by moving the file and then making a symlink in the repo root, but I think this is probably more trouble than it's worth

docker/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+RUN rm -rf /otwa && mkdir -p /otwa
+WORKDIR /otwa
+VOLUME /otwa
+
+# Install ruby packages
+COPY Gemfile .
+COPY Gemfile.lock .
+RUN gem install bundler -v 1.17.3 && bundle install
+
+# If no other command, serve the rails application
```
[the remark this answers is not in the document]
tommilligan: This refers to the default command used when you call `docker run`. For instance;
```bash
# will run `bundle exec rails s -p 3000` in the container (default)
docker-compose run web

# will run `bundle exec rspec spec --fail-fast ./spec/controllers/bookmarks_controller_spec.rb` in the container
docker-compose run web bundle exec rspec spec --fail-fast ./spec/controllers/bookmarks_controller_spec.rb
```

I'll reword to say the 'default command to run in a new container`

config/environments/development.rb
```diff
@@ -6,7 +6,7 @@
   # since you don't have to restart the webserver when you make code changes.
   config.cache_classes = false
   config.eager_load = false
-  config.cache_store = :dalli_store, '127.0.0.1:11211',
+  config.cache_store = :dalli_store, YAML.load_file("#{Rails.root}/config/local.yml")["MEMCACHED_SERVERS"],
```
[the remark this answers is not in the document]
tommilligan: Thanks, updated

docker/database.yml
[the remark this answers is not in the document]
redsummernight: This is fine, I like docker-compose.yml at root, which makes it easy for people to notice Docker support.

Can you comment at the top of docker-compose.yml with something like "you can run docker-compose or you can use these init scripts..."?

docker-compose.yml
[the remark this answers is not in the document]
redsummernight: Thanks! I have plans to remove RedisTestSetup entirely, I'll remove this variable when I do that.

docker/database.yml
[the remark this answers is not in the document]
tommilligan: Done

tommilligan:
@sarken Thanks for your comments, updated. Once we're happy with the end state I'll do some feedback squashing.

I'd like to point out I still get a couple of minor failing rspec tests (currently retesting locally):
```bash
$ docker-compose run -e RAILS_ENV=test web bundle exec rspec spec --fail-fast ./spec/controllers/bookmarks_controller_spec.rb

...snip...

Failures:

  1) WorksHelper#sorted_languages returns all languages sorted alphabetically
     Failure/Error: expect(Language.count).to eq(1)

       expected: 1
            got: 2

       (compared using ==)
     # ./spec/helpers/works_helper_spec.rb:90:in `block (3 levels) in <top (required)>'

Finished in 17 minutes 30 seconds (files took 11.42 seconds to load)
1207 examples, 1 failure, 9 pending

Failed examples:

rspec ./spec/helpers/works_helper_spec.rb:88 # WorksHelper#sorted_languages returns all languages sorted alphabetically
```

However, this could well be a problem with my local setup. I'm happy to get the PR as is into master, as it does allow you to run the Archive locally and pass the _vast majority_ of tests. If there are specific failures, I think iterating on them separately is valid.

redsummernight:
Yeah, I agree if the remaining tests keep being a problem, we can get this change into master and deal with the failures later. Though just in case they're easy to deal with:

**Extra languages errors**
```
rspec ./spec/helpers/works_helper_spec.rb:88 # WorksHelper#sorted_languages returns all languages sorted alphabetically
rspec ./spec/lib/tasks/after_tasks.rake_spec.rb:17 # rake After:reset_word_counts when there are multiple languages updates only works in the specified language
```
We're probably no cleaning up the test databases completely. You could try replacing this line:

https://github.com/otwcode/otwarchive/blob/3b2cf664ea303b1066f7372b6930c9b51cef64b0/spec/spec_helper.rb#L45

with `DatabaseCleaner.clean_with :truncation`. We still want to use `:transaction` as the default strategy, which is faster, but we should truncate everything before the entire test suite for a fresh start.

**External work errors**
```
rspec ./spec/models/collection_spec.rb:107 # Collection all_bookmarked_items_count counts bookmarks of all types
rspec './spec/models/concerns/filterable_spec.rb[2:1:1:1:2]' # ExternalWork behaves like a filterable adding a tag when the tag is canonical adds the filterable to the main reindexing queue
rspec './spec/models/concerns/filterable_spec.rb[2:1:2:1:5:1]' # ExternalWork behaves like a filterable removing a tag when the tag is canonical when the filterable is also tagged with a synonym doesn't delete the filter
```
These 3 errors look similar, they all involve tests that create external works.

https://github.com/otwcode/otwarchive/blob/3b2cf664ea303b1066f7372b6930c9b51cef64b0/factories/works.rb#L37-L40

When external works are validated, we check [whether their URLs are reachable](https://github.com/otwcode/otwarchive/blob/3b2cf664ea303b1066f7372b6930c9b51cef64b0/app/validators/url_active_validator.rb#L11). Can you ping example.org from inside the container?

tommilligan:
@redsummernight thanks for your suggestions

#### Extra languages errors

I tried your suggestion of cleaning up the tables with `truncation` instread of `transaction`, but this caused another test to fail. Will leave this alone for now (it sounds like a pretty minor failure in the grand scheme of things).
```log
$ docker-compose run -e RAILS_ENV=test web bundle exec rspec spec --fail-fast ./spec/controllers/bookmarks_controller_spec.rb
...snip...
Admin::AdminUsersController
  ...snip...
  PUT #update
    ...snip...
    when admin has correct authorization
      updates user roles (FAILED - 1)

Failures:

  1) Admin::AdminUsersController PUT #update when admin has correct authorization updates user roles
     Failure/Error: self.roles = Role.find(role_list)

     ActiveRecord::RecordNotFound:
       Couldn't find Role with 'id'=1
     # /usr/local/bundle/gems/activerecord-5.1.6.2/lib/active_record/relation/finder_methods.rb:343:in `raise_record_not_found_exception!'
     ...snip...
     # ./spec/controllers/admin/admin_users_controller_spec.rb:102:in `block (4 levels) in <top (required)>'
```

#### External work errors

I verified that I can `ping example.org` and get a response inside the container - this is almost certainly just a local setup issue for me (my internet was being really flaky yesterday for no apparent reason). Kind of suprised we rely on an external resource to do this though, I would have expected you could test this with something like `http://localhost:3000` which would also be available offline.

tlee911:
Super cool that this got finished and made it's way in.  I know I opened that other PR months ago and then abandoned it (because real life got in the way).  Thanks everyone!

PR #3864, opened 2020-07-15
AO3-5884 Remove unused class AdminUserPolicy

redsummernight:
## Issue

## Purpose

- Omit specifying UserPolicy when authorizing User objects.
- Disable "Abuse Administration" form for admins without access.
- Authorize before the "update_status" action, which means we don't need to check for valid admin roles inside UserManager, and the authorization error message is the same as everywhere else.
- Update UserManager specs to stop checking admin roles.

Extra:

- Authorize the external work, not the admin, when using UserCreationPolicy.
- Stop passing current_admin as an argument, call the helper directly.
- Remove unused static methods in policy classes.
- Update Rubocop rules RSpec/DescribeClass and Style/PercentLiteralDelimiters.

## Testing Instructions

See issue.


app/policies/user_policy.rb
```diff
@@ -1,27 +1,34 @@
...
+  TROUBLESHOOT_ROLES = %w(superadmin policy_and_abuse open_doors support tag_wrangling).freeze
 
-  def can_suspend_users?
-    user_has_roles?(POLICY_ROLES)
-  end
+  # Roles that allow:
+  # - updating a user's Fannish Next of Kin
+  # - suspending and banning
+  # - deleting all of a spammer's creations
+  JUDGE_ROLES = %w(superadmin policy_and_abuse).freeze
```
sarken, line 12: Good name!

app/policies/user_policy.rb
```diff
@@ -1,27 +1,34 @@
 class UserPolicy < ApplicationPolicy
-  USER_MANAGEMENT_ROLES = %w(superadmin policy_and_abuse open_doors support tag_wrangling).freeze
-  POLICY_ROLES = %w(superadmin policy_and_abuse).freeze
+  # Roles that allow:
+  # - troubleshooting for a user
+  # - managing invitations
+  # - updating a user's email and roles (e.g. wranglers, archivists, not admin roles)
+  TROUBLESHOOT_ROLES = %w(superadmin policy_and_abuse open_doors support tag_wrangling).freeze
```
sarken, line 6: idk, do we want to call this just `MANAGE_ROLES` and `can_manage_users`? I get what it's going for and didn't really bat an eye at it, but when I got down to the `<% if policy(User).can_troubleshoot_users? %>` surrounding  the "Manage Users" menu, I was like, "Ehhh, that's actually a bit confusing. It seems like that should just be on the Troubleshoot buttons."

app/policies/user_policy.rb
```diff
@@ -1,27 +1,34 @@
 class UserPolicy < ApplicationPolicy
-  USER_MANAGEMENT_ROLES = %w(superadmin policy_and_abuse open_doors support tag_wrangling).freeze
-  POLICY_ROLES = %w(superadmin policy_and_abuse).freeze
+  # Roles that allow:
+  # - troubleshooting for a user
+  # - managing invitations
```
sarken, line 4: "managing a user's invitations" or "managing invitations granted to a user," maybe? We're going to end up with a whole other policy class for managing who can send invitations to email addresses and such via the admin/invitations page, and I keep getting confused why `UserPolicy` affects invitations at all.

PR #3866, opened 2020-07-17
AO3-5623 AO3-5518 Fix admin creation script, clean up admin-related feature steps

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5623
https://otwarchive.atlassian.net/browse/AO3-5980
https://otwarchive.atlassian.net/browse/AO3-5518

## Purpose

AO3-5623 AO3-5980 Fix minor issues with admin creation script

- Remove stray quote.
- Print filename so we can identify error messages.
- Sort wiki output by admins' created timestamps.

AO3-5518 Clean up feature steps to create admins

- Have one step for all types of admins.
- Use the admin role in the admin's username so we can reuse the admin later when we need the same role.
- Use a more specific role than "superadmin" whenever we can.
- Use the generic no-role admin if the feature under test is not yet locked behind any admin roles.
- Remove tests where admins reply to news post comments, as we won't do AO3-3685.

## Testing Instructions

Run the script:

<details>

    Pandemonica,monika2@example.com,support,tag_wrangling
    Malina
    Beelzebub,,
    Hell taker,,support
    Justice,prosecutest@example.com,translation
    Lucifer,lucifer@example.com,policy_and_abuse

    For new admins, copy and paste each section into a separate file and upload to the admin's Vault:
    ==== admin-Pandemonica.txt

    ==== admin-Malina.txt
    Email can't be blank

    ==== admin-Beelzebub.txt
    Email can't be blank

    ==== admin-Helltaker.txt
    Email can't be blank

    ==== admin-Justice.txt

    ==== admin-Lucifer.txt
    username: admin-Lucifer
    password: aiT0iewo
    https://127.0.0.1/admin/login

</details>

### DIFF
--- a/features/admins/admin_invitations.feature
+++ b/features/admins/admin_invitations.feature
@@ -255,7 +255,7 @@ Feature: Admin Actions to Manage Invitations
       And I should not see "Joining the Archive currently requires an invitation; however, we are not accepting new invitation requests at this time."
 
   Scenario: An admin can send an invitation to a user via email
-    Given I am logged in as a "support" admin
+    Given I am logged in as an admin
       And all emails have been delivered
     When I follow "Invite New Users"
       And I fill in "invitation[invitee_email]" with "fred@bedrock.com"
@@ -264,7 +264,7 @@ Feature: Admin Actions to Manage Invitations
       And 1 email should be delivered
 
   Scenario: An admin can't create an invite without an email address.
-    Given I am logged in as a "support" admin
+    Given I am logged in as an admin
       And all emails have been delivered
     When I follow "Invite New Users"
       And I press "Invite user"
@@ -278,7 +278,7 @@ Feature: Admin Actions to Manage Invitations
       | odo   | mybucket9  |
       And "dax" has "0" invitations
       And "odo" has "3" invitations
-      And I am logged in as a "support" admin
+      And I am logged in as an admin
     When I follow "Invite New Users"
       And I fill in "Number of invitations" with "2"
       And I select "All" from "Users"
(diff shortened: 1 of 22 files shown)


features/admins/admin_invitations.feature
```diff
@@ -256,7 +255,7 @@ Feature: Admin Actions to Manage Invitations
       And I should not see "Joining the Archive currently requires an invitation; however, we are not accepting new invitation requests at this time."
 
   Scenario: An admin can send an invitation to a user via email
-    Given I am logged in as superadmin
+    Given I am logged in as a "support" admin
```
sarken, line 258: I think all the scenarios from here down are actually things that haven't been locked down yet. These things are done via admin/invitations, /invite_requests, and /invite_requests/manage, but we've only locked down pages like users/username/invitations. 

Ergo, per your description of the PR, we might want to use generic admin accounts here.

features/admins/admin_works.feature
```diff
@@ -6,7 +6,7 @@ Feature: Admin Actions for Works, Comments, Series, Bookmarks
   Scenario: Can troubleshoot works
     Given I am logged in as "regular_user"
       And I post the work "Just a work you know"
-    When I am logged in as superadmin
+    When I am logged in as a "policy_and_abuse" admin
```
sarken, line 9: fwiw, Support or Tag Wrangling usually does this one.

redsummernight: I'll give this scenario to Support.

features/admins/admin_invitations.feature
redsummernight: Thanks, I'll switch to no-role admins. The only exception is

https://github.com/otwcode/otwarchive/blob/3b2cf664ea303b1066f7372b6930c9b51cef64b0/features/admins/admin_invitations.feature#L338

which does use /invitations in part, so I'll keep `as a "support" admin` there.

PR #3868, opened 2020-07-19
AO3-5983 Add Hound rules for Rails and ERB template linting

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5983

## Purpose

Update Hound preferences:

- Add Rubocop extension for Rails best practices and coding conventions (see [the style guide](https://rails.rubystyle.guide/) it enforces).
- Add linting rules on ERB or HTML files ([erb-lint](https://github.com/Shopify/erb-lint)), which covers Ruby/Rails code in the templates. erb-lint can also nag about CSS classes we no longer want to keep using.

## Testing Instructions

None.


.erb-lint.yml
```diff
@@ -0,0 +1,40 @@
+---
+EnableDefaultLinters: true
+linters:
+  DeprecatedClasses:
+    enabled: true
+    addendum: "Reference: https://sarken.github.io/docs/front_end_coding/classes.html."
```
redsummernight, line 6: Should we link to this yet?

.erb-lint.yml
```diff
@@ -0,0 +1,40 @@
...
+    rule_set:
+      - deprecated: ["bookmarks", "collections", "readings", "works"]
+        suggestion: "Avoid the plural form of these classes."
+  Rubocop:
+    enabled: true
+    rubocop_config:
+      inherit_from:
+        - .rubocop.yml
+      Layout/ArgumentAlignment:
+        EnforcedStyle: with_fixed_indentation
```
redsummernight, line 16: Which style do we like more for view files? `with_fixed_indentation` is like this:

https://github.com/otwcode/otwarchive/blob/057b4ff672ef4ef480b9b2fdddb710e1b052cba7/app/views/abuse_reports/new.html.erb#L65-L66

while the default (`with_first_argument`) is like this:

https://github.com/otwcode/otwarchive/blob/057b4ff672ef4ef480b9b2fdddb710e1b052cba7/app/views/abuse_reports/new.html.erb#L58-L59

Gemfile.linters
```diff
@@ -0,0 +1,14 @@
+# All gems required to run linters:
```
redsummernight, line 1: This file is here because juggling the compatibility of these gems can be difficult (for running linters locally), but I don't want to bring all of their dependencies into the main Gemfile and make it harder for automatic updates.

.erb-lint.yml
```diff
@@ -0,0 +1,40 @@
+---
+EnableDefaultLinters: true
+linters:
+  DeprecatedClasses:
```
redsummernight, line 4: I tried adding a check for hyphens/underscores in class names, but got too many false positives (e.g. Rails code for creating classes), so we'll use this only for exact matches.

.erb-lint.yml
sarken: Let's go with fixed, just because it gives you more room on any given line.

.erb-lint.yml
sarken: No, that's just a temporary home for it while it's in progress -- the goal is to eventually put it under otwcode.

.erb-lint.yml
sarken: /me shakes fist at Rails

.erb-lint.yml
redsummernight: I'll remove it for now.

spec/controllers/admin_posts_controller_spec.rb
```diff
@@ -66,7 +66,7 @@
       end
     end
 
-    context "when admin does not have correct authorization" do
+    context "when admin has correct roles" do
```
redsummernight, line 69: Duplicate context descriptions is [actually a violation](https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#rspecrepeateddescription), which is nice!

Gemfile.linters
sarken: Can you explain in small words how this is used? Do you run the commands on lines 3-5 to install the gems for the linters, or...?

redsummernight: I got the idea from ["Is it possible to override gemfile for local development?"](https://stackoverflow.com/questions/48863711/is-it-possible-to-override-gemfile-for-local-development).

- Run `BUNDLE_GEMFILE=Gemfile.linters bundle install` once to install the linter gems + their dependencies + our existing dependencies. Re-run it occasionally to update (bundler will prompt us).
- Run `BUNDLE_GEMFILE=Gemfile.linters bundle exec rubocop <file>` and `BUNDLE_GEMFILE=Gemfile.linters bundle exec erblint <file>` as regularly as we want, to see Rubocop and ERB Lint violations.

sarken: Makes sense! Should we document that on the [coding standards wiki page](https://github.com/otwcode/otwarchive/wiki/Coding-standards) for folks who want to check that their changes meet our style guidelines before submitting PRs?

redsummernight: Yep, replacing this section specifically: https://github.com/otwcode/otwarchive/wiki/Coding-standards#Previewing_Hound_comments.

The downside of running these commands is that people will see all existing violations in a file instead of just the ones on the lines they added/edited (like how Hound works on GitHub). I think https://github.com/AtakamaLLC/lint-diffs looks promising, but that's for later.

redsummernight: Wiki updated.

PR #3869, opened 2020-07-20
AO3-5884 Restrict which admins can update user roles and which can update user emails

sarken:
#  Issue

## Purpose

* Limits updating user roles to super admins, Tag Wrangling, and Open Doors.
* Limits updating user emails to super admins, Policy & Abuse, and Support.
* Adds a hidden roles field with an array containing an empty string to the form for updating user roles to ensure it's possible for Tag Wrangling and Open Doors admins  to remove all roles. Previously, pressing "Update" would submit something like `user => { email => string, roles => [] }` to remove the last remaining role. However, because we've disabled the email field for Tag Wrangling and Open Doors admins, it no longer submits the email string... and because the roles array is empty, it no longer submits _that_, and _then_ it gets the bright idea to just submit nothing instead of an empty hash, prompting [this issue](https://travis-ci.com/github/sarken/otwarchive/jobs/362485192#L1716). (The update to the model is to make sure it doesn't error when it's given this non-existent role to add/remove.)

## Testing Instructions

Refer to Jira.

## References

*  https://github.com/rails/rails/issues/26569 was a bit informative


spec/controllers/admin/admin_users_controller_spec.rb
```diff
@@ -92,16 +92,61 @@
...
 
-        expect(user.roles.include?(role)).to be_truthy
+      ["support", "policy_and_abuse"].each do |admin_role|
+        context "when admin has #{admin_role} role" do
+          before { admin.update(roles: [admin_role]) }
+          it "prevents admins with #{admin_role} role from updating roles" do
+            expect do
+              put :update, params: { id: user.login, user: { roles: [role.id.to_s] } }
+            end.to raise_exception(ActionController::UnpermittedParameters)
+            expect(user.reload.roles.include?(role)).to be_falsey
```
redsummernight, line 141: Same here.

app/views/admin/admin_users/_user_form.html.erb
```diff
@@ -4,11 +4,15 @@
   <%= hidden_field_tag :query, params[:query] %>
   <%= hidden_field_tag :role, params[:role] %>
   <%= hidden_field_tag "id", user.login %>
+  <% # HACK: We need the user param to be present so we can remove all roles.
```
redsummernight, line 7: Can you remove the trailing space here?

app/views/admin/admin_users/_user_form.html.erb
```diff
@@ -4,11 +4,15 @@
   <%= hidden_field_tag :query, params[:query] %>
   <%= hidden_field_tag :role, params[:role] %>
   <%= hidden_field_tag "id", user.login %>
+  <% # HACK: We need the user param to be present so we can remove all roles. 
+     # However, if we simply submit the form with all roles unchecked, the user
+     # param is stripped. Therefore, we need a placeholder. %>
+  <%= hidden_field_tag "user[roles][]", [""], disabled: !admin_can_update_user_roles? %>
```
redsummernight, line 10: `user[roles][]` is the part specifying the array, you can use `""` instead of `[""]` for the value.

sarken: Ah, okay! I thought that wouldn't work because it left the value attribute totally blank in the HTML (like not even `value=""` -- just `value`), but it does indeed seem to work.

spec/controllers/admin/admin_users_controller_spec.rb
```diff
@@ -92,16 +92,63 @@
...
+              email: "updated@example.com",
+              roles: [role.id.to_s]
+            }
+          }
+
+          it_redirects_to_with_notice(root_path, "User was successfully updated.")
+        end
+      end
+
+      %w(open_doors tag_wrangling).each do |admin_role|
```
redsummernight, line 116: I'm having second thoughts about this style... I think because it's an array, it looks better with `[]`.

sarken: I agree -- the brackets are much more intuitive for an array.

redsummernight: I'll update .rubocop.yml and clean up the `%w` brackets in #3868 (at least the ones that got added this release).

PR #3870, opened 2020-07-20
AO3-5999 Make comment factories look more like real comments, and reorganize to use traits.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5999

## Purpose

Refactor comment factories so that they don't include all three of `pseud`, `email`, and `name`, and so that they default to having their commentable set to a chapter instead of a work.

Also modifies the structure of comment factories to use traits, so that it's easier to generate comments with different sets of properties. This reduces the amount of repetition on the factory definitions.

## Testing Instructions

This is tests-only, so just check whether the automated tests pass.

### DIFF
--- a/features/step_definitions/work_search_steps.rb
+++ b/features/step_definitions/work_search_steps.rb
@@ -134,16 +134,21 @@ end
 Given /^a set of works with comments for searching$/ do
   step %{basic tags}
 
-  # Comments created with factories are not added to a work's stat totals
-  # even after running the rake task, so we're doing it through steps that add
-  # comments through the interface
-  step %{I have a work "Work 1"}
-  step %{the work "Work 2" with 1 comments setup}
-  step %{the work "Work 3" with 1 comments setup}
-  step %{the work "Work 4" with 1 comments setup}
-  step %{the work "Work 5" with 3 comments setup}
-  step %{the work "Work 6" with 3 comments setup}
-  step %{the work "Work 7" with 10 comments setup}
+  counts = {
+    "Work 1" => 0,
+    "Work 2" => 1,
+    "Work 3" => 1,
+    "Work 4" => 1,
+    "Work 5" => 3,
+    "Work 6" => 3,
+    "Work 7" => 10
+  }
+
+  counts.each_pair do |title, comment_count|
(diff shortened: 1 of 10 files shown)


features/step_definitions/work_search_steps.rb
```diff
@@ -134,16 +134,21 @@
 Given /^a set of works with comments for searching$/ do
   step %{basic tags}
 
-  # Comments created with factories are not added to a work's stat totals
-  # even after running the rake task, so we're doing it through steps that add
-  # comments through the interface
-  step %{I have a work "Work 1"}
-  step %{the work "Work 2" with 1 comments setup}
```
redsummernight, line 141: Do you think we should update this step as well? Using factories will probably be faster looping through the UI:

https://github.com/otwcode/otwarchive/blob/37fa9b799f39d76fe3cde07d920ddf3abf451af5/features/step_definitions/work_steps.rb#L163-L173

spec/requests/comments_spec.rb
```diff
@@ -14,7 +14,7 @@ def comment_attributes_user
       @work2 = create(:work, fandom_string: "Merlin (TV)", title: "My title is long enough", restricted: false)
       @work2.reindex_document
       @comment2 = create(:comment)
-      @work2.first_chapter.comments << @comment2
+      @work2.last_posted_chapter.comments << @comment2
```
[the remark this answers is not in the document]
redsummernight: You can ignore these. I'll rewrite the file in [AO3-5958](https://otwarchive.atlassian.net/browse/AO3-5958).

features/step_definitions/work_search_steps.rb
tickinginstant: Sure, that seems reasonable. Do you want me to create the work using `FactoryBot`, too? Or just the comments?

And should I do it for this step, too? https://github.com/otwcode/otwarchive/blob/a7bb25ac11754fd8079b4ac9d6c877f541735842/features/step_definitions/work_steps.rb#L125-L146

redsummernight: Please and thanks! Factories for work + comments, in both steps.

PR #3871, opened 2020-07-21
AO3-5883 Let policy_and_abuse admins turn queue on/off and specify how many invites to send per day

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5883

## Purpose

What it says on the tin.

## Testing Instructions

Refer to Jira.

### DIFF
--- a/spec/controllers/admin/settings_controller_spec.rb
+++ b/spec/controllers/admin/settings_controller_spec.rb
@@ -88,7 +88,7 @@ describe Admin::SettingsController do
         {
           hide_spam: true,
           invite_from_queue_enabled: false,
-          invite_from_queue_number: 10,
+          invite_from_queue_number: 10
         }.each_pair do |field, value|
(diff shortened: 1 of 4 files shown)


spec/controllers/admin/settings_controller_spec.rb
```diff
@@ -85,10 +85,16 @@
...
         end
 
-        it "allows admins with policy_and_abuse role to update spam setting" do
-          put :update, params: { id: setting.id, admin_setting: { hide_spam: "1" } }
-          expect(setting.reload.hide_spam?).to be_truthy
-          it_redirects_to_with_notice(admin_settings_path, "Archive settings were successfully updated.")
+        {
+          hide_spam: true,
+          invite_from_queue_enabled: false,
+          invite_from_queue_number: 10
```
redsummernight, line 91: 10 is the default (`INVITE_FROM_QUEUE_NUMBER`), you should have this value be something different.

PR #3873, opened 2020-07-24
AO3-5884 Fix deleting roles when admins can't update roles

redsummernight:
## Issue

## Purpose

Previously, if an admin does not have permissions to update roles, and we don't include "roles" in the permitted attributes, the user's roles will be set to an empty array. Now we skip setting roles entirely in that case.

Also stop skipping validations when updating users, and show validation errors to admins, so admins can't accidentally break users by saving blank emails, etc.

## Testing Instructions

See issue. Specifically:

- PAC admins clicking "Update" should not clear all roles of a user.
- PAC/Support admins should be prevented from entering invalid emails for a user (emails already in use, or blank strings).

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -143,7 +143,7 @@ RSpec.configure do |config|
   config.file_fixture_path = "spec/support/fixtures"
 end
 
-RSpec::Matchers.define_negated_matcher :not_change, :change
+RSpec::Matchers.define_negated_matcher :avoid_changing, :change
 
 def clean_the_database
   # Now clear memcached
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -143,6 +143,8 @@ RSpec.configure do |config|
   config.file_fixture_path = "spec/support/fixtures"
 end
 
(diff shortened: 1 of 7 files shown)


spec/spec_helper.rb
```diff
@@ -143,6 +143,8 @@
   config.file_fixture_path = "spec/support/fixtures"
 end
 
+RSpec::Matchers.define_negated_matcher :not_change, :change
```
sarken, line 146: The [RSpec docs](https://www.rubydoc.info/gems/rspec-expectations/3.5.0/RSpec%2FMatchers.define_negated_matcher) say this will lead to slightly confusing error messages:

>Note: While the most obvious negated form may be to add a not_ prefix, the failure messages you get with that form can be confusing (e.g. "expected [actual] to not [verb], but did not"). We've found it works best to find a more positive name for the negated form, such as avoid_changing rather than not_change.

I can live with that, but thought I'd mention it in case you wanted to change it.

redsummernight: Good idea, I'll change it.

PR #3874, opened 2020-07-25
AO3-5987 Defer generating share options for bookmarks

tommilligan:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5987

## Purpose

Moves the sharing options off of the bookmark pages and into their own controller, to be loaded dynamically when the user presses the "Share" button.

## Testing Instructions

Make sure that you can still use the "Share" button on bookmarks - behaviour should remain unchanged.

Some edge cases that should be tested:

- JavaScript disabled
- Bookmark is private
- Work is unrevealed/part of an unrevealed collection
- Logged out user should not have option to share bookmark

## References

There has already been very similar work done for the `works` page, which I used as a reference while writing this PR.
- Issue: https://otwarchive.atlassian.net/browse/AO3-5907
- PR: https://github.com/otwcode/otwarchive/pull/3815

## Credit

Tom Milligan

they/them

### DIFF
--- a/app/controllers/bookmarks_controller.rb
+++ b/app/controllers/bookmarks_controller.rb
@@ -1,14 +1,16 @@
 class BookmarksController < ApplicationController
   before_action :load_collection
   before_action :load_owner, only: [ :index ]
-  before_action :load_bookmarkable, only: [ :index, :new, :create, :fetch_recent, :hide_recent ]
+  before_action :load_bookmarkable, only: [ :index, :new, :create, :fetch_recent, :hide_recent, :share ]
   before_action :users_only, only: [:new, :create, :edit, :update]
   before_action :check_user_status, only: [:new, :create, :edit, :update]
-  before_action :load_bookmark, only: [ :show, :edit, :update, :destroy, :fetch_recent, :hide_recent, :confirm_delete ]
-  before_action :check_visibility, only: [ :show ]
+  before_action :load_bookmark, only: [ :show, :edit, :update, :destroy, :fetch_recent, :hide_recent, :confirm_delete, :share ]
+  before_action :check_visibility, only: [ :show, :share ]
   before_action :check_ownership, only: [ :edit, :update, :destroy, :confirm_delete ]
 
   before_action :check_pseud_ownership, only: [:create, :update]
+
+  skip_before_action :store_location, only: [:share]
 
   def check_pseud_ownership
     if params[:bookmark][:pseud_id]
@@ -260,6 +262,26 @@ class BookmarksController < ApplicationController
       @bookmark.update_attributes(bookmark_params)
       @bookmarkable = @bookmark.bookmarkable
       render :edit and return
+    end
+  end
+
+  # GET /bookmarks/1/share
+  def share
+    if request.xhr?
+      if @bookmark.bookmarkable.unrevealed?
+        render template: "errors/404", status: :not_found
+      else
+        render layout: false
+      end
+    else
+      # Avoid getting an unstyled page if JavaScript is disabled
+      flash[:error] = ts("Sorry, you need to have JavaScript enabled for this.")
+      if request.env["HTTP_REFERER"]
+        redirect_to(request.env["HTTP_REFERER"] || root_path)
+      else
+        # else branch needed to deal with bots, which don't have a referer
+        redirect_to root_path
+      end
     end
   end
 
--- a/app/controllers/bookmarks_controller.rb
+++ b/app/controllers/bookmarks_controller.rb
@@ -276,7 +276,7 @@ class BookmarksController < ApplicationController
     else
       # Avoid getting an unstyled page if JavaScript is disabled
       flash[:error] = ts("Sorry, you need to have JavaScript enabled for this.")
-      redirect_back_or_default(root_path)
+      redirect_back(fallback_location: root_path)
     end
   end
 
--- a/app/controllers/bookmarks_controller.rb
+++ b/app/controllers/bookmarks_controller.rb
@@ -1,12 +1,12 @@
 class BookmarksController < ApplicationController
   before_action :load_collection
   before_action :load_owner, only: [ :index ]
-  before_action :load_bookmarkable, only: [ :index, :new, :create, :fetch_recent, :hide_recent, :share ]
+  before_action :load_bookmarkable, only: [ :index, :new, :create, :fetch_recent, :hide_recent ]
   before_action :users_only, only: [:new, :create, :edit, :update]
   before_action :check_user_status, only: [:new, :create, :edit, :update]
   before_action :load_bookmark, only: [ :show, :edit, :update, :destroy, :fetch_recent, :hide_recent, :confirm_delete, :share ]
   before_action :check_visibility, only: [ :show, :share ]
-  before_action :check_ownership, only: [ :edit, :update, :destroy, :confirm_delete ]
+  before_action :check_ownership, only: [ :edit, :update, :destroy, :confirm_delete, :share ]
 
   before_action :check_pseud_ownership, only: [:create, :update]
 
@@ -268,7 +268,7 @@ class BookmarksController < ApplicationController
   # GET /bookmarks/1/share
   def share
     if request.xhr?
-      if @bookmark.bookmarkable.unrevealed?
+      if @bookmark.bookmarkable.is_a?(Work) && @bookmark.bookmarkable.unrevealed?
         render template: "errors/404", status: :not_found
       else
         render layout: false
@@ -276,12 +276,7 @@ class BookmarksController < ApplicationController
     else
       # Avoid getting an unstyled page if JavaScript is disabled
       flash[:error] = ts("Sorry, you need to have JavaScript enabled for this.")
-      if request.env["HTTP_REFERER"]
-        redirect_to(request.env["HTTP_REFERER"] || root_path)
-      else
-        # else branch needed to deal with bots, which don't have a referer
-        redirect_to root_path
-      end
+      redirect_back_or_default(root_path)
     end
   end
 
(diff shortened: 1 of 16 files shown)


app/views/bookmarks/share.html.erb
```diff
@@ -0,0 +1,4 @@
+<%= render 'share/share', :shareable => @bookmark, :bookmark => @bookmark %>
+
+<% # FRONT END NOTE: Twitter widget JavaScript included here for share/_share to ensure it is only loaded once per page %>
```
tickinginstant, line 3: This comment doesn't seem right anymore. Now that the `<script>` element is here instead of in `bookmarks/show` and `bookmarks/index`, it seems like it'll be reloaded every time someone clicks the Share button, even if was already loaded. That's not really a huge issue for works, which only ever have one Share button per page, but bookmark listings (e.g. `/users/id/bookmarks`) can include multiple Share buttons.

Is it worth adapting [the script here](https://developer.twitter.com/en/docs/twitter-for-websites/javascript-api/guides/set-up-twitter-for-websites) to avoid loading it multiple times?

app/controllers/bookmarks_controller.rb
```diff
@@ -1,15 +1,17 @@
 class BookmarksController < ApplicationController
   before_action :load_collection
   before_action :load_owner, only: [ :index ]
-  before_action :load_bookmarkable, only: [ :index, :new, :create, :fetch_recent, :hide_recent ]
+  before_action :load_bookmarkable, only: [ :index, :new, :create, :fetch_recent, :hide_recent, :share ]
```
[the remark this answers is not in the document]
tommilligan: Thanks for explaining this - makes sense. Updated.

app/controllers/bookmarks_controller.rb
```diff
@@ -1,15 +1,17 @@
...
   before_action :users_only, only: [:new, :create, :edit, :update]
   before_action :check_user_status, only: [:new, :create, :edit, :update]
-  before_action :load_bookmark, only: [ :show, :edit, :update, :destroy, :fetch_recent, :hide_recent, :confirm_delete ]
-  before_action :check_visibility, only: [ :show ]
+  before_action :load_bookmark, only: [ :show, :edit, :update, :destroy, :fetch_recent, :hide_recent, :confirm_delete, :share ]
+  before_action :check_visibility, only: [ :show, :share ]
   before_action :check_ownership, only: [ :edit, :update, :destroy, :confirm_delete ]
```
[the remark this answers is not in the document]
tommilligan: Updated.

app/controllers/bookmarks_controller.rb
```diff
@@ -263,6 +265,26 @@ def update
     end
   end
 
+  # GET /bookmarks/1/share
+  def share
+    if request.xhr?
+      if @bookmark.bookmarkable.unrevealed?
```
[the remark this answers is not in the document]
tommilligan: Updated.

app/controllers/bookmarks_controller.rb
```diff
@@ -263,6 +265,26 @@ def update
...
+      end
+    else
+      # Avoid getting an unstyled page if JavaScript is disabled
+      flash[:error] = ts("Sorry, you need to have JavaScript enabled for this.")
+      if request.env["HTTP_REFERER"]
+        redirect_to(request.env["HTTP_REFERER"] || root_path)
+      else
+        # else branch needed to deal with bots, which don't have a referer
+        redirect_to root_path
+      end
```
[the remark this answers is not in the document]
tommilligan: Updated to the commonly used `redirect_back_or_default(root_path)` and updated matching test

app/views/bookmarks/share.html.erb
tommilligan: Updated the comment. I played around with only loading the twitter javascript once as per the link you suggested, but this broke the rendering of the twitter button on subsequent loads of the share modal;

- First load
image
- Subsequent loads
image

I think this is a pretty well scoped piece of work though - could be worth opening a small ticket just for this?

app/controllers/bookmarks_controller.rb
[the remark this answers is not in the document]
tickinginstant: Please don't use `redirect_back_or_default`. [It's buggy and there are plans to remove it.](https://otwarchive.atlassian.net/browse/AO3-5949)

app/views/bookmarks/share.html.erb
tickinginstant: [There are instructions for initializing widgets after the page loads.](https://developer.twitter.com/en/docs/twitter-for-websites/javascript-api/guides/scripting-loading-and-initialization) Did you experiment at all with that?

app/controllers/bookmarks_controller.rb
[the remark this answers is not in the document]
tommilligan: Updated, thanks

app/views/bookmarks/share.html.erb
tommilligan: Implemented in https://github.com/otwcode/otwarchive/pull/3874/commits/217e8dacec79b06f356047b42fb983598cdb7585, thanks for telling me to RTFM :wink: that was actually a lot less painless than I was expecting

tickinginstant: Thanks for making the JS load only once!

app/views/share/_share.html.erb
```diff
@@ -27,4 +27,18 @@
...
+  # On subsequent loads, we call `load()` explicitly to trigger rendering widgets.
+%>
+<script>
+  // If we've already loaded the twitter widgets bundle
+  if(window.twttr){
+    // Trigger rendering of widget elements in the div `share` defined above
+    window.twttr.widgets.load(document.getElementById("share"));
+  } else {
+    // Otherwise, load the widgets bundle
+    window.twttr=(function(d,s,id){ var js,fjs=d.getElementsByTagName(s)[0],t=window.twttr||{};if(d.getElementById(id))return t;js=d.createElement(s);js.id=id;js.src="https://platform.twitter.com/widgets.js";fjs.parentNode.insertBefore(js,fjs);t._e=[];t.ready=function(f){t._e.push(f);};return t;}(document,"script","twitter-wjs"));
```
tickinginstant, line 42: I know this is copied from Twitter's docs, but formatted like this makes it difficult to parse. Would it be possible to add some newlines and whitespace so that it's easier on the eyes? I believe that the nginx frontend reformats in-page JS, so the extra whitespace shouldn't be a problem from a bandwidth perspective.

Also, it'd be good to include a link to the page this was copied from in the comment above it, both as attribution and as a reference in case anyone needs to change or fix this code in the future (say if Twitter changes their API).

tommilligan: Updated to make the whole thing more readable, and adds doc references for the added JS code.

tommilligan:
@redsummernight thanks for the feedback. Added another two commits fixing the minor Ruby issues, and a third commit which removes the twitter JS entirely in favour of [sharingbuttons.io](https://sharingbuttons.io/) as you suggested.

I have manually QA'd that the generated links still behave as expected on Twitter/Tumblr, but another pair of eyes on that would be great.

Changing the button implementation causes the following visual changes:

#### This PR

image


#### master

image

features/works/work_share.feature
```diff
@@ -4,7 +4,8 @@ Feature: Share Works
 
   @javascript
   Scenario: Share a work
-    Given I have a work "Blabla"
+    Given I am logged in as "testuser1"
+      And I post the work "Blabla"
```
tommilligan, line 8: Added this as the test asserts we generate an aurhor link for `testuser1`, but the fixture actually creates the work under a random user. Not sure why this was passing in CI up to now, but I got some flaky failures with it locally.

features/step_definitions/share_steps.rb
```diff
@@ -0,0 +1,8 @@
+Then /^the share modal should contain a Twitter share button$/ do
+  with_scope("#share") do
+    iframe = find("li.twitter #twitter-widget-0")
```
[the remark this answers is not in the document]
tommilligan: Updated, and factored out checking both social buttons to a single step.

features/works/work_share.feature
tickinginstant: The reason it passes Travis CI is because it's not exactly a random user:
https://github.com/otwcode/otwarchive/blob/f0a35415338b6499caf5f71d777cc7c867c3d073/features/step_definitions/user_steps.rb#L124-L128

As long as the database is empty (which is always true for Travis CI, because it creates the database from scratch, and then cleans the database after each test), it'll always use `testuser1`.

(Is the database you're using for your local tests empty? If not, the javascript tests are supposed to wipe all of the data, because [they're supposed to use truncation for the database cleaning between tests](https://github.com/otwcode/otwarchive/blob/f0a35415338b6499caf5f71d777cc7c867c3d073/features/support/env.rb#L67-L70).)

redsummernight: :point_up: It's worth checking if your test database is being cleaned up correctly to avoid issues like this in the future. I won't block the PR on this, though.

tommilligan: Thanks for explaining, I only ran into this the once and haven't since, but I'll keep an eye out for future flakiness.

FWIW I think specifying the user to log in as explicitly is much better when it's later required in an assertion, and to only use a random user when we don't care about the user later in the test.

tommilligan:
@redsummernight is there anything else I need to do before merge?

redsummernight:
No, everything's good! It'll be just a while before we merge ready PRs. Our [next planned release](https://otwarchive.atlassian.net/issues/?jql=project%20%3D%20AO3%20AND%20fixVersion%20%3D%200.9.294%20ORDER%20BY%20updated%20DESC) includes the Rails 5.2 upgrade and some involved regression testing.

PR #3877, opened 2020-08-03
AO3-5975 Display warning on proxies

sarken:
#  Issue

https://otwarchive.atlassian.net/browse/AO3-5975

## Purpose

Puts a giant pink warning about security on any page that is not being access from a host that matches ArchiveConfig.APP_HOST. The warning has a "Dismiss Notice" button, and pressing the button will set a session cookie to hide the banner.

I've added previews to Jira.

## Testing Instructions

Maybe we can temporarily change ArchiveConfig.APP_HOST on staging and make sure the banner appears and can be dismissed?


app/views/layouts/_proxy_notice.html.erb
```diff
@@ -0,0 +1,18 @@
+<div id="proxy-notice">
+  <div class="content">
+    <p class="english">Notice: (1) You are using a proxy site that is not affiliated with AO3. (2) The entity that set up the proxy site can see whatever data you submit to the AO3 through the proxy site, including your IP address and the content of comments.</p>
```
sarken, line 3: I wasn't sure if we should make the banner text translatable. Putting the English in the en.yml locale file made sense, but not so much putting the Chinese there.

app/views/layouts/application.html.erb
```diff
@@ -38,6 +38,15 @@
   <body class="<%= logged_in? ? 'logged-in' : 'logged-out' %>" <%= PhraseApp::InContextEditor.enabled? ? "data-no-turbolink" : "" %>>
     <div id="outer" class="wrapper">
       <ul id="skiplinks"><li><a href="#main"><%= ts('Main Content') %></a></li></ul>
+      <%= content_for :footer_js do %>
```
sarken, line 41: I put the code here because it's where the JavaScript will insert the banner, but it could've gone anywhere because it's really just inserting JavaScript in the footer.

app/views/layouts/_proxy_notice.html.erb
redsummernight: We can put both in locale files, because it's possible to force a specific language for one `I18n.t` call.

https://guides.rubyonrails.org/i18n.html#setting-and-passing-a-locale

app/views/layouts/application.html.erb
redsummernight: In that case, can we skip `content_for` and put the block below `<%= render :partial => 'layouts/javascripts' %>`?

app/views/layouts/_proxy_notice.html.erb
```diff
@@ -0,0 +1,18 @@
...
+    <p class="chinese"> 重要提示：<br/>
```
redsummernight, line 4: Do we plan to selectively show the notice in different languages, hence the "english" and "chinese" classes?

sarken: No, I just thought it might be nice to have the class names there.

app/views/layouts/application.html.erb
sarken: Sure, I can put a comment here instead.

app/views/layouts/_proxy_notice.html.erb
```diff
@@ -0,0 +1,17 @@
+<div id="proxy-notice">
+  <div>
+    <% # TODO: When the interface is localized, this should be revised to only show the notice in the user's selected language. %>
+    <p><%= t(".text_html") %></p>
+    <p><%= t(".text_html", locale: :"zh-CN") %></p>
```
sarken, line 5: This isn't currently working -- it just gives the English. I have tried the following:
```
    <%= t("layouts.proxy_notice.text_html", locale: :"zh-CN") %>
    <%= I18n.t("layouts.proxy_notice.text_html", locale: :"zh-CN") %>
    <%= I18n.t :"layouts.proxy_notice.text_html", locale: :"zh-CN" %>
    <%= I18n.t(".text_html", locale: :"zh-CN") %>
    <%= I18n.t ".text_html", locale: :"zh-CN" %>
    <%= I18n.t :text_html, locale: :"zh-CN" %>

    <%= t("layouts.proxy_notice.text_html", locale: "zh-CN") %>
    <%= I18n.t("layouts.proxy_notice.text_html", locale: "zh-CN") %>
    <%= I18n.t :"layouts.proxy_notice.text_html", locale: "zh-CN" %>
    <%= I18n.t(".text_html", locale: "zh-CN") %>
    <%= I18n.t ".text_html", locale: "zh-CN" %>
    <%= I18n.t :text_html, locale: "zh-CN" %>
```
I also tried putting it in `de.yml` just to see if the hyphenated name was the problem, but no luck.

sarken: To clarify, some of those result in English, while others result in `translation missing: zh-CN.text_html translation`

app/views/layouts/_proxy_notice.html.erb
```diff
@@ -0,0 +1,22 @@
+<div id="proxy-notice">
+  <div>
+    <% # TODO: When the interface is localized, this should be revised to only show the notice in the user's selected language. %>
+    <p><%= t(".text_html") %></p>
+    <p>
+      <% # Doing it as a block saves us from needing to specify the scope. %>
+      <% I18n.with_locale("zh-CN") do %>
```
redsummernight, line 7:
```
<p><%= t(".text_html", locale: :"zh-CN") %></p>
```
This works for me. You may have to [restart Rails](https://stackoverflow.com/questions/20555995/automatically-reload-rails-yml-files-in-config-locales) after making changes to the locale yml files.

app/views/layouts/_javascripts.html.erb
```diff
@@ -106,4 +106,15 @@
   <% end %>
 <% end %>
 
+<% # Renders layouts/proxy_notice below skip links in layouts/application. %>
+<% if Rails.env.production? %>
```
redsummernight, line 110: We should allow this on staging for testing.

app/views/layouts/_proxy_notice.html.erb
sarken: Yeah, sorry, I did get it working after the third restart -- I didn't realize it hadn't dismissed the comment saying it wasn't working.

sarken: Oh, do you mean it works for you without specifying the scope? 

I had it working with `<%= I18n.t :text_html, scope: [:layouts, :proxy_notice], locale: :"zh-CN" %>`, then tried switching back to `<%= t(".text_html", locale: :"zh-CN") %>`, but no luck. I didn't make any changes to the locale file during that process -- only to this file -- so I don't think that was the issue.

sarken: I have no idea if it's working or not, but I've pushed your version. (It _seems_ to be working, but I don't understand why, especially since it was inexplicably insistent about it needing `html_safe` to handle the line breaks this morning despite using `_html` in the key, and now it's quite happy without that. My best guess at this point is browser caching.)

redsummernight: It might actually be browser caching. You should check the Cache-Control header on responses from the Rails dev server, and see if it's [enough to get Safari to stop caching](https://stackoverflow.com/a/2068407).

PR #3880, opened 2020-08-08
AO3-5861 Fix 500 on /users/{invalid}/works/collected

tlee911:
# Pull Request Checklist

✓ Have you read ["How to write the perfect pull request"](https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/)?
✓ Have you read the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
✓ Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)?
✓ Have you added the [Jira](https://otwarchive.atlassian.net) issue number as the *first* thing in your pull request title (e.g. `AO3-1234 Fix thing`)
✓ Do you have fewer than 5 pull requests already open? If not, please wait until they are reviewed and merged before creating new pull requests.

## Issue

https://otwarchive.atlassian.net/browse/AO3-5861

## Purpose

Fixes 500 error when accessing URLs like https://archiveofourown.org/users/fakeuser/works/collected where the user is invalid.  Now returns 404 instead.

## Testing Instructions

Should give 404 (not 500): https://archiveofourown.org/users/{invalid_user}/works/collected
Should still work: https://archiveofourown.org/users/{valid_user}/works/collected

Unit tests have been updated.  Note that I'm not entirely convinced these tests belong in the "with restricted works" context since they actually don't depend on the restricted works at all, but that's where the existing tests were.  I'm open to moving them up the tree a bit to the general "collected" context.

## References

N/A

## Credit

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -138,7 +138,7 @@ class WorksController < ApplicationController
     options[:page] = params[:page] || 1
     options[:show_restricted] = current_user.present? || logged_in_as_admin?
 
-    @user = User.find_by(login: params[:user_id])
+    @user = User.find_by!(login: params[:user_id])
 
     return unless @user.present?
 
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -139,9 +139,6 @@ class WorksController < ApplicationController
     options[:show_restricted] = current_user.present? || logged_in_as_admin?
 
     @user = User.find_by!(login: params[:user_id])
-
-    return unless @user.present?
-
     @search = WorkSearchForm.new(options.merge(works_parent: @user, collected: true))
     @works = @search.search_results
     flash_search_warnings(@works)
--- a/spec/controllers/works/default_rails_actions_spec.rb
+++ b/spec/controllers/works/default_rails_actions_spec.rb
@@ -505,6 +505,18 @@ describe WorksController, work_search: true do
     let(:collection) { create(:collection) }
     let(:collected_user) { create(:user) }
 
+    it "returns not found error if user does not exist" do
+      expect do
+        get :collected, params: { user_id: "dummyuser" }
+      end.to raise_error(ActiveRecord::RecordNotFound)
+    end
+
+    it "returns not found error if no user is set" do
+      expect do
+        get :collected
+      end.to raise_error(ActiveRecord::RecordNotFound)
+    end
+
     context "with anonymous works" do
       let(:anonymous_collection) { create(:anonymous_collection) }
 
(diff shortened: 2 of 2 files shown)


spec/controllers/works/default_rails_actions_spec.rb
```diff
@@ -577,14 +577,21 @@ def call_with_params(params)
...
-          get :collected
-          expect(assigns(:works)).to be_nil
+        it "returns not found error if user does not exist" do
+          expect {
+            get :collected, params: { user_id: "dummyuser" }
+          }.to raise_error(ActiveRecord::RecordNotFound)
+        end
+
+        it "returns not found error if no user is set" do
+          expect {
```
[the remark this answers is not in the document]
redsummernight: You can avoid these warnings by using `do ...  end`, like this:

https://github.com/otwcode/otwarchive/blob/f0a35415338b6499caf5f71d777cc7c867c3d073/spec/controllers/admin/admin_users_controller_spec.rb#L136-L138

app/controllers/works_controller.rb
```diff
@@ -138,7 +138,7 @@ def collected
     options[:page] = params[:page] || 1
     options[:show_restricted] = current_user.present? || logged_in_as_admin?
 
-    @user = User.find_by(login: params[:user_id])
+    @user = User.find_by!(login: params[:user_id])
 
     return unless @user.present?
```
tickinginstant, line 143: I believe the `find_by!` in line 141 means that `@user.present?` is guaranteed to be true at this point. So you should be able to delete this line now.

spec/controllers/works/default_rails_actions_spec.rb
```diff
@@ -577,14 +577,21 @@ def call_with_params(params)
...
         end
 
-        it "does NOT return any works if no user is set" do
-          get :collected
-          expect(assigns(:works)).to be_nil
+        it "returns not found error if user does not exist" do
+          expect {
+            get :collected, params: { user_id: "dummyuser" }
+          }.to raise_error(ActiveRecord::RecordNotFound)
+        end
```
tickinginstant, line 589: This test and the one below it are nested inside of the `with restricted works` context, but there's nothing about these tests that is specific to users with restricted works. (I think they should apply regardless of which properties the works have.) It might be good to move them out of this context block, possibly into their own context block(s).

tlee911: Yeah, that's what I was thinking too.  I'll just put them top-level under the describe `collected` block.

PR #3881, opened 2020-08-09
AO3-5924 Adjust donate page title

tlee911:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5924

## Purpose

Changes the page title for https://archiveofourown.org/donate to "Donate or Volunteer | Archive of Our Own" as suggested.

## Testing Instructions

Verify page title on donate page.

The scenario in `feature/other_a/homepage.feature` has been extended to check this.

## References

N/A

## Credit

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -45,7 +45,7 @@ class HomeController < ApplicationController
   
   # donate
   def donate
-    @page_subtitle = "Donate or Volunteer"
+    @page_subtitle = t("layouts.home.donate.page_title")
     render action: "donate", layout: "application"
   end
   
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -45,7 +45,7 @@ class HomeController < ApplicationController
   
   # donate
   def donate
-    @page_subtitle = t("layouts.home.donate.page_title")
+    @page_subtitle = t(".page_title")
     render action: "donate", layout: "application"
   end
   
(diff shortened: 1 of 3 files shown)


app/controllers/home_controller.rb
```diff
@@ -45,6 +45,7 @@ def site_map
   
   # donate
   def donate
+    @page_subtitle = "Donate or Volunteer"
```
sarken, line 48: Could you make this a translatable string with the key `page_title`?

tlee911: I put this under `layouts.home.donate.page_title`, but let me know if that's too deeply nested or there's some other convention we should be following.

config/locales/views/en.yml
```diff
@@ -70,6 +70,9 @@ en:
     proxy_notice:
       text_html: "Notice: (1) You are using a proxy site that is not affiliated with AO3. (2) The entity that set up the proxy site can see whatever data you submit to the AO3 through the proxy site, including your IP address and the content of comments."
       button: "Dismiss Notice"
+    home:
```
sarken, line 73: Even though this is technically used by a file in `layouts`, I think we can move it out from under there and treat it like [the subtitle for the troubleshoot controller's show page](https://github.com/otwcode/otwarchive/blob/a3a88a4ed830db48a5fe35f590318792ac7805b8/app/controllers/troubleshooting_controller.rb#L10-L14).

tlee911: I made the suggested change and it works fine.

Nitpicking my own code:
It looks odd to have `@page_subtitle = t(".page_title")`, where key is named "title", but it only has one use, which is to be a subtitle.  The example you gave does the same thing though.

sarken: Yeah, we've got  some renaming to do around `@page_title` and `@page_subtitle`, but for now, this is fine!

PR #3885, opened 2020-08-15
AO3-6016 Remove anon_commenting_disabled column, and associated code.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6016

## Purpose

- Add a migration to remove the `anon_commenting_disabled` column.
- Clean up all code related to the `anon_commenting_disabled` column.
- Add `anon_commenting_disabled` to the list of ignored columns for the `Work` class.
- Continue working on the Edit Multiple Works TODO introduced in #3852.


db/migrate/20200814233538_remove_anon_commenting_disabled_from_works.rb
```diff
@@ -0,0 +1,49 @@
...
+    else
+      remove_column :works, :anon_commenting_disabled
+    end
+  end
+
+  def down
+    if Rails.env.staging? || Rails.env.production?
+      database = Work.connection.current_database
+
+      puts <<~PTOSC
```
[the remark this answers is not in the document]
tickinginstant: I disagree in this case. The logger is designed to make the location that you're printing to flexible, but the whole point of this migration is to print the command to console. It'd be unhelpful to make people dig through log files to find the `pt-osc` command that needs to be run.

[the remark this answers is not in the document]
redsummernight: Makes sense. If you'd like, in this PR you can use `Exclude` to make Rubocop ignore this warning in `'db/**/*.rb'`.

lib/tasks/after_tasks.rake
```diff
@@ -694,16 +694,6 @@ namespace :After do
       REDIS_GENERAL.del(key)
     end
   end
-
-  desc "Copy anon_commenting_disabled to comment_permissions."
```
redsummernight, line 698: We keep `add_user_id_to_kudos` around in case people need to update their development database, although in that case, without running the task before the pseud_id removal migration, you would lose all logged-in user kudos data. This task is not as necessary.

I'm okay with removing this task, but I could be convinced otherwise.

tickinginstant: I don't have strong feelings either way. Maybe I should just add it back?

redsummernight: Yeah, let's add it back for now.

PR #3884, opened 2020-08-14
AO3-6006 Show "pseud (username)" in dashboard heading

tlee911:
# Pull Request Checklist

* [x] Have you read ["How to write the perfect pull request"](https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/)?
* [x] Have you read the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [x] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)?
* [x] Have you added the [Jira](https://otwarchive.atlassian.net) issue number
  as the *first* thing in your pull request title (e.g. `AO3-1234 Fix thing`)
* [x] Do you have fewer than 5 pull requests already open? If not, please wait
  until they are reviewed and merged before creating new pull requests.

## Issue

https://otwarchive.atlassian.net/browse/AO3-6006

## Purpose

Make it clear on user dashboard that a pseud is being displayed by showing the username as well if a non-default pseud is being displayed.  Note that I have removed the links from the heading here.  I felt they weren't useful since you were already viewing this particular user's page.

## Testing Instructions

1. Go to user page of user with extra pseuds
2. Verify that if default pseud is selected only username is displayed in the heading
3. Verify that if a different pseud is selected, the byline format "pseud (username)" is displayed in the heading

New test added in `user_dashboard.feature` to account for this, note that I added examples to also check works and bookmarks pages for the same thing (didn't see tests for this elsewhere).  Also, this can easily be expanded with one more example for the series page (AO3-6007).

## References

See also AO3-6007 and #3887.  Same scenario outline is used for testing, since the expected outcome is similar.  There may be conflicts with that one depending on the merge order.

## Credit

tlee911 (he/him)

PR #3887, opened 2020-08-18
AO3-6007 Show series counts on pseud dashboard page

tlee911:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6007

## Purpose

Changes the heading for a pseud's series page to be "X Series by pseud (username) to match the same format as works or bookmarks pages.  Was previously just "pseud's series".

## Testing Instructions

1. Have a user with additional pseuds
2. View the series page for the user, the heading should be "X Series by username"
3. View the series page for a non-default pseud, the heading should be "X Series by pseud (username)"

The scenario outline in `user_dashboard.feature` from #3884 was leveraged to test this the same as other pseud pages.  A few other test case expectations were also modified to handle the counts in the heading now.

## References

See also AO3-6006 and #3884.  Same scenario outline is used for testing, since the expected outcome is similar.  There may be conflicts with that one depending on the merge order.

## Credit

tlee911 (he/him)


app/views/series/index.html.erb
```diff
@@ -1,10 +1,11 @@
 <!--descriptive page name and system messages, descriptions, and instructions.-->
-<% if @user %>
-	<h2 class="heading"><%=h t('.owned_series', :default => "%{user_name}'s Series", :user_name => @user.login) %></h2>
+<h2 class="heading">
+  <% if @user %>
+    <%= search_header @series, nil, "Series", @pseud ? @pseud: @user %>
```
redsummernight, line 4: Can you add a space between `@pseud` and `:`?

features/support/paths.rb
```diff
@@ -125,11 +125,16 @@ def path_to(page_name)
       user_path(id: $1)
     when /^(.*?)(?:'s)? "(.*)" pseud page$/i
       user_pseud_path(user_id: $1, id: $2)
+    when /^the (user|dashboard) page for user "(.*)" with pseud "(.*)"$/i
```
redsummernight, line 128: Can you use this existing path instead?

https://github.com/otwcode/otwarchive/blob/a3a88a4ed830db48a5fe35f590318792ac7805b8/features/support/paths.rb#L126-L127

features/support/paths.rb
```diff
@@ -125,11 +125,16 @@ def path_to(page_name)
...
+      user_pseud_path(user_id: $2, id: $3)
     when /^(.*?)(?:'s)? user url$/i
       user_url(id: $1)
     when /^([^ ]*?)(?:'s)? works page$/i
       step %{all indexing jobs have been run}
       user_works_path(user_id: $1)
+    when /^the works page for user "(.*)" with pseud "(.*)"$/i
+      step %{all indexing jobs have been run}
+      user_pseud_works_path(user_id: $1, pseud_id: $2)
```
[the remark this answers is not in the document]
redsummernight: Can you replace the backrefs with [`Regexp.last_match(n)`](https://rubystyle.guide/#no-perl-regexp-last-matchers)? You can ignore the existing ones, though.

features/support/paths.rb
```diff
@@ -125,11 +125,16 @@ def path_to(page_name)
...
     when /^(.*?)(?:'s)? "(.*)" pseud page$/i
       user_pseud_path(user_id: $1, id: $2)
+    when /^the (user|dashboard) page for user "(.*)" with pseud "(.*)"$/i
+      user_pseud_path(user_id: $2, id: $3)
     when /^(.*?)(?:'s)? user url$/i
       user_url(id: $1)
     when /^([^ ]*?)(?:'s)? works page$/i
       step %{all indexing jobs have been run}
       user_works_path(user_id: $1)
+    when /^the works page for user "(.*)" with pseud "(.*)"$/i
```
redsummernight, line 135: Can you use [the non-greedy version `(.*?)` instead of `(.*)`](https://github.com/otwcode/otwarchive/wiki/Automated-Testing#our-cucumber-standards)?

features/support/paths.rb
tlee911: I guess technically this particular case doesn't belong in this PR since the test using this is commented out.  However, the intent was to collapse all the individual scenarios for testing dashboard/works/bookmarks/series into examples so I could use one pattern in the steps (https://github.com/otwcode/otwarchive/pull/3887/files#diff-bb02ae9a598f27c708bad226e9f77559R124).  I'm basically aliasing the step to avoid making a separate scenario because of the different naming convention.  What's preferable in this case?  Is there another option I'm not thinking of?

redsummernight: Oh good idea. In that case we should deprecate the old path. Can you add a comment like this on the old path?

https://github.com/otwcode/otwarchive/blob/a3a88a4ed830db48a5fe35f590318792ac7805b8/features/support/paths.rb#L133-L135

sarken (changes requested):
It looks like something went wrong when this was merged with master -- we have a lot of test failures now.

tlee911:
> It looks like something went wrong when this was merged with master -- we have a lot of test failures now.

I think I see the problem.  In search_helper.rb, the rails upgrade (https://github.com/otwcode/otwarchive/pull/3862) changed header from string to array and I blindly did a "use mine" conflict resolution.

sarken (approved):

PR #3888, opened 2020-08-19
AO3-4862 Extending work controller test coverage

cesy:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4862

## Purpose

Updated version of https://github.com/otwcode/otwarchive/pull/3165

Please make sure Tal Hayon gets correct credit.


spec/controllers/works/default_rails_actions_spec.rb
```diff
@@ -325,43 +409,61 @@ def call_with_params(params)
...
+          it "shows results when filters are disabled" do
+            get :index, params: { tag_id: fandom.name }
+            expect(assigns(:works)).to include(work)
+          end
         end
 
         context "with restricted works" do
           before do
-            @work2 = create(:work, fandom_string: @fandom.name, restricted: true)
+            work2 = create(:work, fandom_string: @fandom.name, restricted: true)
```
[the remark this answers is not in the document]
cesy: It's not useless, we're checking it doesn't come up in the search results.

spec/controllers/works/default_rails_actions_spec.rb
```diff
@@ -325,43 +409,61 @@ def call_with_params(params)
 
       context "with a valid owner tag" do
         before do
-          @fandom2 = create(:canonical_fandom)
-          @work2 = create(:work, fandom_string: @fandom2.name)
+          fandom2 = create(:canonical_fandom)
+          work2 = create(:work, fandom_string: fandom2.name)
```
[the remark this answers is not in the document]
cesy: Not useless, we're checking it's not in the results.

features/works/work_edit_multiple.feature
```diff
@@ -28,22 +28,54 @@ Feature: Edit Multiple Works
...
+    And I should see "All"
+    And I should see "None"
+  When I select "Glorious" for editing
+    And I select "Excellent" for editing
+    And I press "Edit"
+  Then I should see "Your edits will be applied to all of the following works"
+    And I should see "Glorious"
+    And I should see "Excellent"
+  When I set the fandom to "Random"
+   And I press "Update All Works"
```
sarken, line 46: I think this line needs to be indented one more space.

features/works/work_edit_multiple.feature
```diff
@@ -28,22 +28,54 @@ Feature: Edit Multiple Works
       And I should see "Lovely"
 
   Scenario: I can edit multiple works at once
-    Given I am logged in as "author"
-      And I post the work "Glorious" with fandom "SGA"
-      And I post the work "Excellent" with fandom "Star Trek"
-      And I go to my works page
+  Given I am logged in as "author"
```
sarken, line 31: Should we stick to the original indenting here, since the other tests in this file appear to use it?

spec/controllers/works/default_rails_actions_spec.rb
```diff
@@ -176,10 +176,43 @@ def call_with_params(params)
     end
   end
 
+  describe "edit" do
+    let(:user) { create(:user) }
+    let(:work) { create(:work, authors: [user.default_pseud], posted: true) }
```
sarken, line 181: The work factory creates posted works by default as of #3602, so `posted: true` shouldn't be necessary in most of these tests.

spec/controllers/works/default_rails_actions_spec.rb
```diff
@@ -176,10 +176,43 @@ def call_with_params(params)
...
+      fake_login_known_user(user)
+    end
+
+    it "redirects to orphan work page if only author is being removed" do
+      get :edit, params: { id: work.id, remove: "me" }
+      expect(response).to redirect_to controller: "orphans", action: "new", work_id: work.id
+    end
+  end
+
+  context "destroy" do
```
sarken, line 193: This should be `describe`

spec/controllers/works/default_rails_actions_spec.rb
```diff
@@ -208,19 +241,36 @@ def call_with_params(params)
...
         include "Invalid creator: Could not find a pseud *impossible*."
     end
 
+    xit "renders new if edit is pressed" do
+      work_attributes = attributes_for(:work)
+      post :create, params: { work: work_attributes, edit_button: true }
+      expect(response).to render_template("new")
+    end
+
+    context "cancel button is pressed" do
```
sarken, line 258: How about `context "with cancel_button params"` or "context "when cancel_button params are present"` or similar?

spec/controllers/works/default_rails_actions_spec.rb
```diff
@@ -277,30 +327,55 @@ def call_with_params(params)
...
     it "sets the fandom when given a fandom id" do
-      params = { fandom_id: @fandom.id }
+      params = { fandom_id: fandom.id }
       get :index, params: params
-      expect(assigns(:fandom)).to eq(@fandom)
+      expect(assigns(:fandom)).to eq(fandom)
+    end
+
+    it "redirects to tag page for noncanonical tags" do
+      unsorted_tag = create(:unsorted_tag)
```
sarken, line 345: Could we use a different kind of tag here? When you use a tag on a work in reality, it will always be a rating, category, fandom, warning, freeform, character, or relationship. Unsorted tags only exist on bookmarks -- they're tags that haven't already been classified as fandom, freeform, etc.

spec/controllers/works/miscellaneous_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+require "spec_helper"
+
+describe WorksController do
+  include LoginMacros
+  include RedirectExpectationHelper
+
+  let(:user) { create(:user) }
+  let!(:work) { create(:work, authors: [user.default_pseud]) }
+
+  context "preview_tags" do
```
sarken, line 12: This should be `describe`

spec/controllers/works/miscellaneous_spec.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+describe WorksController do
+  include LoginMacros
+  include RedirectExpectationHelper
+
+  let(:user) { create(:user) }
+  let!(:work) { create(:work, authors: [user.default_pseud]) }
```
sarken, line 10: Does this definitely need to be `let!`? (I only ask because I know `let` works with the regular `edit` action, so I thought it might here as well.)

spec/controllers/works/default_rails_actions_spec.rb
```diff
@@ -325,43 +404,61 @@ def call_with_params(params)
...
+          expect(assigns(:works).items).not_to include(work2)
         end
 
         it "shows different results on second page" do
           get :index, params: { tag_id: @fandom.name, page: 2 }
-          expect(assigns(:works).items).not_to include(@work)
+          expect(assigns(:works).items).not_to include(work)
+        end
+
+        context "when disabling filtering" do
```
sarken, line 423: I don't think this accurately describes what's being tested. It looks like this is testing when the `suspend_filter_counts` setting is enabled -- disabling filters was removed in #3546.

spec/controllers/works/default_rails_actions_spec.rb
```diff
@@ -540,6 +637,24 @@ def call_with_params(params)
         get :collected, params: { user_id: collected_user.login }
         expect(assigns(:works)).to include(work, anonymous_work)
       end
+
+      context "when filtering is disabled" do
```
sarken, line 641: We removed this admin setting in #3546, so this can probably go.

sarken (changes requested):
Thanks for updating this! I've left some comments, and I think this might need to be updated with master before Travis will successfully run. No rush, though!

features/works/work_edit_multiple.feature
```diff
@@ -28,22 +28,54 @@ Feature: Edit Multiple Works
...
+    And I should not see "SGA"
+    And I should not see "Star Trek"
+  When I view the work "Glorious"
+  Then I should see "Random"
+    And I should not see "SGA"
+  When I view the work "Excellent"
+  Then I should see "Random"
+    And I should not see "Star Trek"
+
+  Scenario: I can edit multiple works at once using a pseud
```
tickinginstant, line 58: I don't think this scenario should be included. [The code that this is intended to test](https://github.com/otwcode/otwarchive/blob/b5bf2e7a5cd362d1e350f2be087e8049721a2484/app/controllers/works_controller.rb#L647-L648) can't be accessed through the normal interface, IIRC -- the scenario was only added because the original PR included code to make it accessible. But [the plan is to delete the code](https://otwarchive.atlassian.net/browse/AO3-5476), not make it accessible.

spec/controllers/works/default_rails_actions_spec.rb
```diff
@@ -176,10 +176,43 @@ def call_with_params(params)
...
+      fake_login_known_user(user)
+    end
+
+    it "sets flash message in case of error" do
+      allow_any_instance_of(Work).to receive(:destroy).and_raise("Cannot save")
+
+      delete :destroy, params: { id: work }
+      expect(flash[:error]).to eq("We couldn't delete that right now, sorry! Please try again later.")
+
+      allow_any_instance_of(Work).to receive(:destroy).and_call_original
```
tickinginstant, line 207: I feel like RSpec ought to be able to clean up its own mocks. Have you checked whether this line of code is necessary?

spec/controllers/works/default_rails_actions_spec.rb
```diff
@@ -208,19 +241,36 @@ def call_with_params(params)
...
-      work_attributes[:author_attributes] = { ids: @user.pseud_ids,
+      work_attributes[:author_attributes] = { ids: user.pseud_ids,
                                               byline: "*impossible*" }
       post :create, params: { work: work_attributes }
       expect(response).to render_template("new")
       expect(assigns[:work].errors.full_messages).to \
         include "Invalid creator: Could not find a pseud *impossible*."
     end
 
+    xit "renders new if edit is pressed" do
```
tickinginstant, line 252: Why is this test disabled? Is the test broken, or is there something wrong with the code it's checking?

spec/controllers/works/default_rails_actions_spec.rb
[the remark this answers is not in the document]
tickinginstant: The results run in a different scope, so these variables are inaccessible in the results. (Hence Hound's complaint.) I think `let!` would be good here.

spec/controllers/works/multiple_actions_spec.rb
```diff
@@ -124,6 +124,19 @@
...
     end
 
+    context "when work parameters are invalid" do
+      let(:work_params) { { work: { summary: "a" * (ArchiveConfig.SUMMARY_MAX + 1) } } }
+
+      before do
+        put :update_multiple, params: params
+      end
+
+      xit "redirects to the user multiple work path with an error" do
```
tickinginstant, line 134: Why is this test disabled?

spec/spec_helper.rb
```diff
@@ -193,3 +193,9 @@ def suspend_resque_workers
   # Resume the original Resque.enqueue_to behavior.
   allow(Resque).to receive(:enqueue_to).and_call_original
 end
+
+def create_archivist
+  user = create(:user)
+  user.roles << Role.create(name: "archivist")
+  user
+end
```
tickinginstant, line 201: This doesn't appear to be used in this PR.

features/works/work_edit_multiple.feature
cesy: Fixed

features/works/work_edit_multiple.feature
cesy: Fixed

spec/controllers/works/default_rails_actions_spec.rb
cesy: Fixed

spec/controllers/works/default_rails_actions_spec.rb
cesy: Fixed

spec/controllers/works/default_rails_actions_spec.rb
cesy: Fixed

spec/controllers/works/default_rails_actions_spec.rb
cesy: Fixed

spec/controllers/works/miscellaneous_spec.rb
cesy: Fixed

spec/controllers/works/default_rails_actions_spec.rb
cesy: Fixed

spec/controllers/works/default_rails_actions_spec.rb
cesy: Fixed

features/works/work_edit_multiple.feature
cesy: Removed

spec/spec_helper.rb
cesy: Yeah, that was left over from another PR, looks like, thanks for catching

spec/controllers/works/default_rails_actions_spec.rb
cesy: Looks like it works fine without

spec/controllers/works/default_rails_actions_spec.rb
cesy: I think it was broken in an earlier version but is now working

spec/controllers/works/multiple_actions_spec.rb
cesy: Also doesn't seem to have a reason so I've enabled it

spec/controllers/works/miscellaneous_spec.rb
cesy: Yeah, should be fine without !, fixed

cesy:
I've fixed all the previous comments, can't currently figure out if the failing test is related to the changes.

spec/controllers/works/default_rails_actions_spec.rb
```diff
@@ -277,30 +326,55 @@ def call_with_params(params)
...
+      expect(assigns(:fandom)).to eq(fandom)
+    end
+
+    it "redirects to tag page for noncanonical tags" do
+      noncanonical_tag = create(:character)
+      get :index, params: { id: work, tag_id: noncanonical_tag.name }
+      expect(response).to redirect_to(tag_path(noncanonical_tag))
+    end
+
+    it "redirects to tags works page for noncanonical merged tags page" do
```
sarken, line 349: Should this example and the one on line 357 be tested under the `without caching` section, since we already have the same tests in the _with_ caching section?

https://github.com/otwcode/otwarchive/blob/584c4daabf3caf755692f9202256bd9af4a4ec78/spec/controllers/works/default_rails_actions_spec.rb#L456-L474

It would also makes sense to use the same descriptions for the new tests that we used for the existing tests. That will make it clearer the examples just cover the same thing in different contexts. (If you could change the existing `describe "with caching"` and `describe "without caching"` to _context_ instead, that would be great.)

spec/controllers/works/default_rails_actions_spec.rb
```diff
@@ -324,44 +402,64 @@ def call_with_params(params)
...
+            AdminSetting.first.update_attribute(:suspend_filter_counts, true)
+            admin_settings = AdminSetting.first
+            controller.instance_variable_set("@admin_settings", admin_settings)
+          end
+
+          after do
+            allow(controller).to receive(:fetch_admin_settings).and_call_original
+          end
+
+          it "shows results when filters are disabled" do
```
sarken, line 435: Let's describe this as `it "returns the work" do`. That will

* match the description of the related example on line 332
* get rid of the incorrect part about filters being disabled (we're testing when filter counts are suspended, which is made clear by the fact this is inside `context "when suspend_filter_counts is on" do`)

sarken (changes requested):

spec/controllers/works/default_rails_actions_spec.rb
```diff
@@ -208,19 +239,37 @@ def call_with_params(params)
...
-      work_attributes[:author_attributes] = { ids: @user.pseud_ids,
+      work_attributes[:author_attributes] = { ids: user.pseud_ids,
                                               byline: "*impossible*" }
       post :create, params: { work: work_attributes }
       expect(response).to render_template("new")
       expect(assigns[:work].errors.full_messages).to \
         include "Invalid creator: Could not find a pseud *impossible*."
     end
 
+    it "renders new if edit is pressed" do
```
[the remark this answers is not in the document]
cesy: Fixed

spec/controllers/works/default_rails_actions_spec.rb
```diff
@@ -208,19 +239,37 @@ def call_with_params(params)
...
+      work_attributes = attributes_for(:work).except(:posted)
+      # returns a failure unpermitted param :posted, so removed it following same method as above
```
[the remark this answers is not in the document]
cesy: Fixed

spec/controllers/works/default_rails_actions_spec.rb
```diff
@@ -434,30 +532,30 @@ def call_with_params(params)
...
       end
     end
   end
 
   describe "update" do
     let(:update_user) { create(:user) }
-    let(:update_work) {
-      work = create(:work, authors: [update_user.default_pseud])
+    let!(:update_work) do
+      work = create(:work, authors: [update_user.default_pseud], posted: true)
```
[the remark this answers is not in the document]
cesy: Fixed

spec/controllers/works/default_rails_actions_spec.rb
```diff
@@ -277,30 +326,55 @@ def call_with_params(params)
...
+    end
+
+    it "redirects to tag page for noncanonical tags" do
+      noncanonical_tag = create(:character)
+      get :index, params: { id: work, tag_id: noncanonical_tag.name }
+      expect(response).to redirect_to(tag_path(noncanonical_tag))
+    end
+
+    it "redirects to tags works page for noncanonical merged tags page" do
+      noncanonical_fandom = create(:fandom, canonical: false)
```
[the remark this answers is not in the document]
cesy: Fixed

spec/controllers/works/miscellaneous_spec.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
```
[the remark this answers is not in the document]
cesy: Fixed

spec/controllers/works/default_rails_actions_spec.rb
cesy: I've tweaked slightly further to "shows the work in the index", and also clarified the description a couple of lines down.

spec/controllers/works/default_rails_actions_spec.rb
cesy: Yes, fixed

spec/controllers/works/multiple_actions_spec.rb
```diff
@@ -124,6 +124,19 @@
       end
     end
 
+    context "when work parameters are invalid" do
```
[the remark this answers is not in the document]
cesy: I'm not sure what the original developer was trying to test, but checking both of those sounds like a good idea to me. I'll have a go.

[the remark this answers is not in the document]
cesy: It looks like an unpermitted work parameter actually works fine, which is a minor loophole if people want to be editing their form submissions manually, but not a big security risk since it still goes through the normal work controller checks. And a permitted parameter with an invalid value gives an error in the controller rather than a friendly error page, but that's okay because the form doesn't let you do that unless you edit the form submission manually anyway. So I'm going to scrap this test as it requires coding skills to do either rather than being something a user normally could do.

spec/controllers/works/multiple_actions_spec.rb
cesy: Yes, oops, should now be properly enabled, and I've fixed the name

cesy:
When I come back to this, I have one cucumber test still to fix, the rest of the comments have been addressed.

sarken:
Hi, Cesy! Someone pointed out this has been in draft for two years, so I wanted to check if you'd be picking this back up soon. If not, no worries, we can close it until you're ready.

cesy:
I'd still be up for it if you think it would be useful? I imagine better automated tests are still a good thing.

sarken:
Yeah, more test coverage is always good!

Bilka2:
Adopted at #5959 based on the [policy for stalled pull requests](https://github.com/otwcode/otwarchive/wiki/Commit-Policy#user-content-Stalled_Pull_Requests_and_Jira_Issues)

PR #3891, opened 2020-08-19
AO3-5615 Fandom validation rules for unwrangleable tags

tlee911:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5615

## Purpose

Tags which are fandoms or tags which have no fandoms can no longer be marked unwrangleable. 

Note that I did not use the suggested error message text from the JIRA issue, but instead matched the format of previously existing error messages. 

## Testing Instructions

For this specific feature:
- Log in as tag wrangler
- Find a tag which is a fandom
- Attempt to mark it unwrangleable (note error message informing user fandoms cannot be unwrangleable)
- Find a tag with no fandoms
- Attempt to mark it unwrangleable (note error message informing user that tags with no fandoms cannot be unwrangleable)

However, part of this PR includes a substantial refactor of what was previously a single monster test case:
https://github.com/otwcode/otwarchive/blob/master/features/tags_and_wrangling/tag_wrangling.feature#L26-L256

I've broken it down to be more atomic.  Please double-check that no scenarios have gone missing or just become incorrect.  I've never done tag wrangling so I'm just going by what I see in code base as expected behaviour.

## References

N/A

## Credit

tlee911 (he/him)


features/tags_and_wrangling/tag_wrangling.feature
```diff
@@ -23,225 +23,217 @@ Feature: Tag wrangling
...
+    Given the tag wrangling setup
+      And I am logged in as a tag wrangler
+    When I go to the "Stargate SG-1" tag page
+      And I follow "Edit" within ".header"
     Then I should see "Edit Stargate SG-1 Tag"
+      And I should see "Name"
+      And I should see "Category"
+      And I should see "Canonical"
+      And I should see "Synonym"
+      And I should see "Unwrangleable"
```
tickinginstant, line 55: From the issue description:
>If the tag you're editing is a fandom, the Unwrangleable section of the form should just... not be there, because fandoms are never supposed to be marked unwrangleable.

"Stargate SG-1" is a fandom tag, so I believe the "Unwrangleable" section shouldn't be visible here.

tlee911: Good catch.  I forgot to implement that part.

This test case is meant to validate the form, so I'll change this section to used a tag that's not a fandom, then add separate steps to check this behaviour for fandom tags.

tlee911:
There seems to be a disconnect with Travis on this one.  If I click through to the build it was successful days ago, but status isn't being updated here.

redsummernight:
Yeah no worries about the checks. As long as a PR has the "awaiting review" label someone will (usually) come back to it.

features/step_definitions/tag_steps.rb
```diff
@@ -79,15 +79,9 @@
   rel.add_association(fand)
 end
 
-Given /^a canonical (\w+) "([^\"]*)"$/ do |tag_type, tagname|
+Given /^a (non-?c|c)anonical (\w+) "([^\"]*)"$/ do |canonical, tag_type, tagname|
```
sarken, line 82: Thanks for combining the two similar step definitions! However, we had a bit of trouble parsing this. Could we rework the variable names to be a little clearer on the first read? 

```
Given /^a (noncanonical|canonical) (\w+) "(.*?)"$/ do |canonical_status, tag_type, tag_name|
  tag = tag_type.classify.constantize.find_or_create_by_name(tag_name)
  tag.canonical = canonical_status == "canonical"
  tag.save
end
```

(I have no idea why we used `tagname` as the variable name elsewhere in this file, since we generally prefer _not_ to squish words together and instead use underscores, but things happen!)

sarken (changes requested):
This looks good, thanks! Just a minor suggestion to make one of the step definitions a bit clearer.

features/step_definitions/tag_steps.rb
tlee911: The variable naming suggestions make total sense.

I was trying to be clever with the first regex and allow use of any among "canonical", "noncanonical", or "non-canonical" (with optional hyphen).  Was that too much?  The hyphenated version wasn't previously supported, and to be fair I probably only put that in because it got tripped up by it.

sarken: No, the hyphen part was fine! The confusing bit was just the part where it didn't use the whole word to determine the canonical status. If it'll work and you want to do `(non-?canonical|canonical)`, you're more than welcome to go for it!

sarken (approved):
Thank you!

PR #3892, opened 2020-08-19
AO3-6033 Update codecov gem to 0.2.0

cesy:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6033

## Purpose

Update to newer version, shouldn't make any difference to users, only to coders

## Testing Instructions

Check the Archive homepage loads

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -159,7 +159,7 @@ group :test do
   gem 'cucumber-timecop', require: false
   # Code coverage
   gem 'simplecov', '~> 0.14.0'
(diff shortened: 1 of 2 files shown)


sarken:
I'm not sure what's up with Travis: it's showing one pull request build passed and another queued, but the queued one is actually finished and passing as well. 🤷 

I'm going to assume it just got confused by the force push and merge this now. We'll know soon enough if there's an actual issue.

PR #3893, opened 2020-08-25
AO3-4709 Allow comments to be frozen

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4709

## Purpose

Lets work creators and select admins freeze comments so people can't reply to them, yay.

~This is a mess. 🙃 It has been a mess for about two years now, so I come begging for help, or this will never be finished.~

Specific questions to get us started:

1. What do we want to call the database column that says the comment is frozen? `frozen` is not actually a possibility, so I temporarily used `on_ice`, which I'm sure will not make every developer who reads this code want to kill me.
2. ~SURELY there is a better way to handle the admin permissions? The gist is wrangling admins need to be able to freeze comments on tags, PAC needs to be able to freeze comments on works, and everyone should be able to freeze comments on news posts because it takes The Village People (or something to that effect).~
3. I don't _hate_ using `current` as the class name on the "Frozen" indicator, since it does indicate the current state, but I don't love it, either. Should I change it to, idk, `indicator`? (I'm somewhat inclined to keep the `current` style, though.) 

I'm also quite willing to hand this over to someone else's tender mercies. 

## Testing Instructions

Refer to Jira.


app/controllers/comments_controller.rb
```diff
@@ -381,6 +412,18 @@ def reject
     redirect_to_all_comments(@comment.ultimate_parent, show_comments: true)
   end
 
+  # PUT /comments/1/freeze
+  def freeze
+    if !@comment.on_ice?
+      @comment.full_set.each { |c| c.mark_frozen! }
```
tickinginstant, line 418: FYI, I believe that [AO3-5939](https://otwarchive.atlassian.net/browse/AO3-5939) means we don't know whether `full_set` will return the right set of comments. It should in most cases, because AO3-5939 is a concurrency issue, but there may be cases where it doesn't.

(If AO3-5939 does happen, it'll affect random comments, but permanently, not intermittently. Because if the race condition causes the `threaded_left` and `threaded_right` values to be incorrectly computed while the new comment is being inserted, those incorrect values will be saved to the database. And that means that strange things will always happen when trying to freeze/unfreeze those comments, and likely other comments nearby.)

sarken: I wonder if that explains the issue I ran into with [this test](https://github.com/otwcode/otwarchive/pull/3893/files#diff-c37d231b9eefa56b4c32761eb6bf136bR710), where the comments had to be created in a specific order in the specs, which would presumably be creating the comments so quickly it might as well be simultaneous.

I guess the way to avoid using `full_set` would be to:

1. Get the tag, admin post, or chapter of the comment we are freezing.
2. Get all the comments on that tag, admin post, or chapter.
3. Look through that set of comments to find any comments whose commendable is the comment we are freezing.
3. Then look through our set of comments _again_ to find any with _those_ as commentables.
4. And so on, until we don't find any more comments.
5. Finally, freeze all the comments we've found.

...however, I just looked, and `find_all_comments` relies on `all_children`, which also relies on `threaded_left` and `threaded_right`. 🤦  But would that perhaps still work with AO3-5939? It doesn't sound like the comments become totally detached and unfindable -- just ordered incorrectly.

tickinginstant: I wouldn't be surprised if it came up during tests. It should be possible to simulate the issue by creating a collection of nested comments, and deliberately not reloading the parent comment before adding its children. Which is precisely the sort of situation that you could end up with when crafting comments by hand for a test.

Unfortunately, it's not exclusively an issue of order. Consider a comment tree like the following:
```
Comment A (threaded_left: 1, threaded_right: 6)
- Comment B (threaded_left: 2, threaded_right: 3)
- Comment C (threaded_left: 4, threaded_right: 5)
```
Now suppose that two comments are added in quick succession: Comment D, a reply to Comment B; and Comment E, a reply to Comment C. Comment D is inserted at index 3 (the `threaded_right` value of Comment B), so it'll increment all higher `threaded_left`/`threaded_right` values by 2, so that the tree of comments becomes:
```
Comment A (threaded_left: 1, threaded_right: 8)
- Comment B (threaded_left: 2, threaded_right: 5)
- - Comment D (threaded_left: 3, threaded_right: 4)
- Comment C (threaded_left: 6, threaded_right: 7)
```
So far, everything is correct. Now Comment E is inserted. Because it's lagging a little behind the database, it thinks it's being inserted at index 5, the `threaded_right` value of Comment C before Comment D was inserted and changed everything. So it'll bump up any `threaded_left`/`threaded_right` values that are at least 5, resulting in the following set of comments:
```
Comment A (threaded_left: 1, threaded_right: 10)
- Comment B (threaded_left: 2, threaded_right: 7)
- - Comment D (threaded_left: 3, threaded_right: 4)
- Comment C (threaded_left: 8, threaded_right: 9)
- - Comment E (threaded_left: 5, threaded_right: 6)
```
Comment E is nested under Comment C according to its `commentable_id`/`commentable_type` values, but its `threaded_left`/`threaded_right` values are more appropriate for a comment nested under Comment B. And it'll be included in Comment B's `full_set` and `all_children`, not in Comment C's. The same will be true of any descendants of Comment E that might be added in the future.

(I think it may also simultaneously mangle the `threaded_right` of Comment C, setting it to 7, leaving it with a `threaded_left` larger than its `threaded_right`. But I haven't actually checked.)

sarken: Thanks for that explanation! I'll link to this comment on AO3-5939, since it may be helpful there.

>Comment E is nested under Comment C according to its commentable_id/commentable_type values, but its threaded_left/threaded_right values are more appropriate for a comment nested under Comment B. And it'll be included in Comment B's full_set and all_children, not in Comment C's. The same will be true of any descendants of Comment E that might be added in the future.

That sounds fine for my purposes. I'm only interested in the commentable columns and the ultimate parent (or pseudo ultimate parent, if I can swing it: getting all the comments for a chapter would be preferable to getting all the comments for the work as a whole). As long as Comment E shows up threaded under _some_ comment on the ultimate parent, it should be fine. 

(And indeed, the test scenario that failed with the `full_set` approach works with this new one.)

tickinginstant: As long as you don't use anything that filters based on `threaded_left`/`threaded_right` values, I believe you should be okay.

Alternatively, I could just try to fix AO3-5939.

sarken: Excellent. 

I'm happy to do it either way, so if you get a chance to fix AO3-5939, that would be great! (No rush, though, since we're doing Rails next, and that will take a bit of testing time.) If not, I'll just leave a comment in my code saying it can be changed once AO3-5939 is fixed.

app/helpers/comments_helper.rb
```diff
@@ -269,6 +284,19 @@ def do_cancel_delete_comment_link(comment)
     end
   end
 
+  def freeze_comment_button(comment)
+    if comment.on_ice?
+      button_to ts("Unfreeze"), freeze_comment_path(comment), method: :put
```
tickinginstant, line 289: Would it make sense to have Freeze/Unfreeze as two separate actions? As is, if two people are trying to moderate at the same time (e.g. admins on an admin post), one of them might click the "Freeze" button on a comment that the other has already frozen, and end up accidentally unfreezing the comment instead.

tickinginstant:
>SURELY there is a better way to handle the admin permissions? The gist is wrangling admins need to be able to freeze comments on tags, PAC needs to be able to freeze comments on works, and everyone should be able to freeze comments on news posts because it takes The Village People (or something to that effect).

IDK if this is the cleanest way to handle it, but maybe something along the lines of:
```
authorize([:commentable_policies, @comment.ultimate_parent])
```
Or:
```
policy([:commentable_policies, @comment.ultimate_parent]).freeze?
```
That way, you could define separate policy classes for each type of commentable: `CommentablePolicies::WorkPolicy`, `CommentablePolicies::TagPolicy`, and `CommentablePolicies::AdminPostPolicy`. (You might need some extra code to handle tags, because of the subclassing, but it looks like Pundit does allow you to define `Tag.policy_class` and override the defaults.)

Alternatively, all policies do have access to the `record`, so you could write code that checks `record.parent_type` or `record.ultimate_parent.class`.

db/migrate/20200817045853_add_freezing_to_comments.rb
```diff
@@ -0,0 +1,49 @@
...
+    else
+      add_column :comments, :on_ice, :boolean, default: 0, null: false
+    end
+  end
+
+  def down
+    if Rails.env.staging? || Rails.env.production?
+      database = Comment.connection.current_database
+
+      puts <<~PTOSC
```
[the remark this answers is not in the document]
redsummernight: We'll exclude this rule for migrations (like in #3885).

sarken:
Thanks for your comments, all! I will close this while I work on them.

sarken:
I'm baaaack.

- I've divided freezing and unfreezing into separate actions, but I need to reduce the resultant duplication in `check_permission_to_unfreeze` and `check_permission_to_unfreeze`.
- I've stopped using `full_set`, which is broken per [this discussion](https://github.com/otwcode/otwarchive/pull/3893#discussion_r476769093).
- I have not changed the way the admin permission checks work because the alternatives didn't seem that much better.
- I have not yet addressed Hound comments.
- I have inexplicably broken comment creation? I have no idea what I did, only that the tests started failing after 733eaa6.

```Failures:

  1) CommentsController POST #create when replying from the inbox creates the reply and redirects to user inbox path
     Failure/Error:
       def new
         if @commentable.nil?
           flash[:error] = ts("What did you want to comment on?")
           redirect_back_or_default(root_path)
         else
           @comment = Comment.new
           @controller_name = params[:controller_name] if params[:controller_name]
           @name =
             case @commentable.class.name
             when /Work/
     
     ArgumentError:
       wrong number of arguments (given 1, expected 0)
     # ./app/controllers/comments_controller.rb:249:in `new'
     # ./app/controllers/comments_controller.rb:290:in `block in create'
     # ./app/controllers/comments_controller.rb:289:in `create'
     # ./spec/controllers/comments_controller_spec.rb:202:in `block (4 levels) in <top (required)>'

  2) CommentsController POST #create when the commentable is a valid tag when logged in as an admin posts the comment and shows it in context
     Failure/Error:
       def new
         if @commentable.nil?
           flash[:error] = ts("What did you want to comment on?")
           redirect_back_or_default(root_path)
         else
           @comment = Comment.new
           @controller_name = params[:controller_name] if params[:controller_name]
           @name =
             case @commentable.class.name
             when /Work/
     
     ArgumentError:
       wrong number of arguments (given 1, expected 0)
     # ./app/controllers/comments_controller.rb:249:in `new'
     # ./app/controllers/comments_controller.rb:290:in `block in create'
     # ./app/controllers/comments_controller.rb:289:in `create'
     # ./spec/controllers/comments_controller_spec.rb:215:in `block (5 levels) in <top (required)>'

  3) CommentsController POST #create when the commentable is a valid tag when logged in as a tag wrangler posts the comment and shows it in context
     Failure/Error:
       def new
         if @commentable.nil?
           flash[:error] = ts("What did you want to comment on?")
           redirect_back_or_default(root_path)
         else
           @comment = Comment.new
           @controller_name = params[:controller_name] if params[:controller_name]
           @name =
             case @commentable.class.name
             when /Work/
     
     ArgumentError:
       wrong number of arguments (given 1, expected 0)
     # ./app/controllers/comments_controller.rb:249:in `new'
     # ./app/controllers/comments_controller.rb:290:in `block in create'
     # ./app/controllers/comments_controller.rb:289:in `create'
     # ./spec/controllers/comments_controller_spec.rb:234:in `block (5 levels) in <top (required)>'

  4) CommentsController POST #create when the commentable is a comment when the commentable is spam creates comment, gives success notice, and marks previous comment as approved
     Failure/Error:
       def new
         if @commentable.nil?
           flash[:error] = ts("What did you want to comment on?")
           redirect_back_or_default(root_path)
         else
           @comment = Comment.new
           @controller_name = params[:controller_name] if params[:controller_name]
           @name =
             case @commentable.class.name
             when /Work/
     
     ArgumentError:
       wrong number of arguments (given 1, expected 0)
     # ./app/controllers/comments_controller.rb:249:in `new'
     # ./app/controllers/comments_controller.rb:290:in `block in create'
     # ./app/controllers/comments_controller.rb:289:in `create'
     # ./spec/controllers/comments_controller_spec.rb:341:in `block (5 levels) in <top (required)>'

Finished in 1 minute 44.51 seconds (files took 10.96 seconds to load)
311 examples, 4 failures

Failed examples:

rspec ./spec/controllers/comments_controller_spec.rb:197 # CommentsController POST #create when replying from the inbox creates the reply and redirects to user inbox path
rspec ./spec/controllers/comments_controller_spec.rb:214 # CommentsController POST #create when the commentable is a valid tag when logged in as an admin posts the comment and shows it in context
rspec ./spec/controllers/comments_controller_spec.rb:233 # CommentsController POST #create when the commentable is a valid tag when logged in as a tag wrangler posts the comment and shows it in context
rspec ./spec/controllers/comments_controller_spec.rb:340 # CommentsController POST #create when the commentable is a comment when the commentable is spam creates comment, gives success notice, and marks previous comment as approved
```

app/controllers/comments_controller.rb
```diff
@@ -381,6 +424,32 @@ def reject
     redirect_to_all_comments(@comment.ultimate_parent, show_comments: true)
   end
 
+  # PUT /comments/1/freeze
+  def freeze
+    # TODO: When AO3-5939 is fixed, we can use
+    # @comment.full_set.each { |c| c.mark_frozen! }
+    if !@comment.on_ice? && @comment.save
+      set_to_freeze_or_unfreeze.each { |c| c.mark_frozen! }
```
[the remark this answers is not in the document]
sarken: `set_to_freeze_or_unfreeze.each(&:mark_frozen!)` does not actually appear to work.

app/models/comment.rb
```diff
@@ -395,6 +395,16 @@ def mark_as_ham!
     Rails.env.production? && Akismetor.submit_ham(akismet_attributes)
   end
 
+  # Freeze single comment. Use on full_set to freeze comment and replies.
```
sarken, line 398: TODO: Update these comments when I do the Hound fixes.

lib/plugins/acts_as_commentable/comment_methods.rb
```diff
@@ -54,7 +54,7 @@ def children_count
     # Returns comment itself if unthreaded
     def full_set
       if self.threaded_left
-        Comment.include(:pseud).where("threaded_left BETWEEN (?) and (?) AND thread = (?)",
+        Comment.includes(:pseud).where("threaded_left BETWEEN (?) and (?) AND thread = (?)",
```
sarken, line 57: This doesn't technically need to be in here since I stopped using `full_set`, but it seemed better to keep these fixes than not.

sarken:
Okay, I ended up tidying the permission checks after all, so now we're just left with questions 1 & 3 from the "Purpose" section.

app/controllers/comments_controller.rb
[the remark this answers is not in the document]
tickinginstant: What error did you get? It seems to work on my machine.

app/policies/comment_policy.rb
```diff
@@ -1,11 +1,23 @@
...
   end
 
+  def can_freeze_comment?
+    if record.ultimate_parent.class == Work
+      user_has_roles?(FREEZE_WORK_COMMENT_ROLES)
+    elsif Tag::TYPES.include?(record.ultimate_parent.class.to_s)
+      user_has_roles?(FREEZE_TAG_COMMENT_ROLES)
+    elsif record.ultimate_parent.class == AdminPost
+      user&.is_a?(Admin)
+    end
```
tickinginstant, line 18: You may be able to rewrite this as:
```
case record.ultimate_parent
when Work
when Tag
when AdminPost
end
```
That could simplify the case for tags, if it works.

app/controllers/comments_controller.rb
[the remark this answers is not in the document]
sarken: I don't remember if it was an error or it just totally failing to have an observable effect when I pressed the button, but I do remember it wasn't in the tests, but rather when I actually tested the button. It seems to be working now, though, so I'll make the change and assume it was just gremlins.

app/helpers/comments_helper.rb
```diff
@@ -269,6 +279,19 @@ def do_cancel_delete_comment_link(comment)
     end
   end
 
+  def freeze_comment_button(comment)
+    if comment.on_ice?
+      button_to ts("Unfreeze"), unfreeze_comment_path(comment), method: :put
```
redsummernight, line 284: These buttons won't work without JS, but we can live with that.

app/controllers/comments_controller.rb
```diff
@@ -555,6 +619,38 @@ def redirect_to_all_comments(commentable, options = {})
...
+  def permission_to_modify_frozen_status
+    parent = find_parent
+    return true if policy(@comment).can_freeze_comment?
+    return true if parent.is_a?(Work) && current_user_owns?(parent)
+
+    false
+  end
+
+  # TODO: Remove when AO3-5939 is fixed.
+  def set_to_freeze_or_unfreeze
```
redsummernight, line 631: Can you move this into `CommentMethods`?

redsummernight (commented):
Jira doesn't mention preventing editing comments, but that seems expected.

Re: Purpose:

1\. Even shorter, `iced`. If you do keep `on_ice`, I'll see if I can work in a Yuri on Ice reference in tests at some point.
3\. `frozen_status`/`<whatever>_status` depending on the column name? Though `current` is fine.

I'll actually look at the specs later. 😎

app/helpers/comments_helper.rb
sarken: I think that's only true if it's `link_to`, where we rely on JavaScript to make a link behave like a button. This seems to work fine without JavaScript, since it's a form with a submit button:

```
<li>
  <form class="button_to" method="post" action="/comments/123/unfreeze">
    <input type="hidden" name="_method" value="put">
    <input type="submit" value="Unfreeze">
    <input type="hidden" name="authenticity_token" value="hulk-SMASH">
  </form>
</li>
```

app/controllers/comments_controller.rb
```diff
@@ -108,6 +111,13 @@ def check_unreviewed
     redirect_to logged_in? ? root_path : new_user_session_path
   end
 
+  def check_if_commentable_frozen
+    return unless @commentable&.respond_to?(:iced?) && @commentable&.iced?
```
tickinginstant, line 115: FYI, `nil` has a `respond_to?` function, so `&.` is unnecessary here.

app/controllers/comments_controller.rb
```diff
@@ -108,6 +111,13 @@ def check_unreviewed
...
+
+    flash[:error] = ts("Sorry, you cannot reply to a frozen comment.")
+    redirect_to(request.env["HTTP_REFERER"] || root_path)
```
tickinginstant, line 118: Would it be possible to use `redirect_back(fallback_location: root_path)` here instead?

app/controllers/comments_controller.rb
```diff
@@ -142,11 +152,37 @@ def check_permission_to_delete
...
+  end
+
+  # Comments on works can be frozen by admins with proper authorization or the
+  # work creator.
+  # Comments on tags can be frozen by admins with proper authorization.
+  # Comments on admin posts can be frozen by any admin.
+  def check_permission_to_freeze
+    return if permission_to_modify_frozen_status
+
+    flash[:error] = ts("Sorry, you don't have permission to freeze that comment thread.")
```
tickinginstant, line 173: I feel like it'd be cleaner to use i18n here, rather than having two separate functions that do almost the same thing apart from the text being displayed. Maybe just `flash[:error] = t(".not_permitted")` or some such?

(That would also mean you could move the contents of `permission_to_modify_frozen_status` back into here, so that the logic and the error message aren't so far apart.)

app/controllers/comments_controller.rb
```diff
@@ -142,11 +152,37 @@ def check_permission_to_delete
...
+    redirect_to(request.env["HTTP_REFERER"] || root_path) && return
```
tickinginstant, line 174: You don't need `&& return` here, since it's the end of the function. (Ditto with all of the places it's used below, as well.)

app/controllers/comments_controller.rb
sarken: It was more of a philosophical decision than one related to the message. I originally had them combined only to [change them when I split freeze and unfreeze into separate actions](https://github.com/otwcode/otwarchive/pull/3893/commits/d401a73f382b148e3e4ebda02bb26a5d8944baf3#diff-34e52388abd1ae0658edd56e12cd6d935fac98673361084ba07a99ca4bb836e7L168). 

If we think it's better to keep them together, though, I can change it back.

spec/controllers/comments_controller_spec.rb
```diff
@@ -1030,6 +2304,48 @@
         post :create, params: { work_id: work.id, comment: anon_comment_attributes }
         it_redirects_to_with_error(work_path(work), edit_error_message)
       end
+
+      context "PUT #freeze" do
+        # TODO: Should this be more like check_permission_to_delete and use access_denied?
```
redsummernight, line 2309: I think this way is fine. `access_denied` currently doesn't let us use `redirect_back`.

redsummernight (approved):

redsummernight:
I'll merge this if it looks fine to you @tickinginstant.

spec/controllers/comments_controller_spec.rb
redsummernight: Feel free to remove these TODOs unless you'd like a consistency cleanup later.

tickinginstant:
LGTM! I like all of the translatable error messages.

sarken:
I'll take the TODOs out real quick before we merge

PR #3899, opened 2020-09-11
AO3-6051 AO3-4551 AO3-4829 Use MemCacheStore and overhaul skin caching

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6051
https://otwarchive.atlassian.net/browse/AO3-4551
https://otwarchive.atlassian.net/browse/AO3-4829

## Purpose

- Switch to using `MemCacheStore` instead of `DalliStore`. They both use `Dalli::Client` under the hood, but `MemCacheStore` makes `race_condition_ttl` work properly.
- Overhaul skin caching. The switch to `MemCacheStore` means that `Marshal.dump` is called outside of [this block in Dalli::Server](https://github.com/petergoldstein/dalli/blob/ca8b286de230546c3bfe2789e9b6bc5a2c619570/lib/dalli/server.rb#L71-L75), which means that marshaling errors will cause an error instead of silently failing and writing an error to the log. So I rewrote the caching code to cache only `skin.get_style`, and not the skin itself.

## Testing Instructions

1. As a user, make a custom skin that overrides the site skin, and start using it. Make sure that the skin is included in subsequent page loads. Then try editing the skin, and make sure that the changes are immediately reflected on the next page you visit.
2. As a user, make a custom skin that adds onto the site skin, and start using it. Make sure that the skin is included in subsequent page loads. Then try editing the skin, and make sure that the changes are immediately reflected on the next page you visit.
3. As an admin, go to Skins > Approved Skins and pick a new skin to use as the default. Make sure that the new skin is included in subsequent page loads, both for logged out users, and for users with a custom skin adding onto the default.
4. ~~As an admin, go to Skins > Approved Skins, click the link for the current default skin, and then edit it to add some CSS. Make sure that the changes are immediately reflected on the next page you visit, both for logged-out users, and for users with a custom skin adding onto the default.~~ **EDIT:** This won't necessarily work on staging/production because the filesystem isn't shared between different servers, so if the default skin is cached to the filesystem, the changes will only show up on the server that handled the `SkinsController#update` request, not any other servers.
5. As an admin, go to Skins > Approved Skins, check "Chooser" for one of the listed skins, and press "Update." Then check the footer to make sure that the skin you selected is listed under "Choose Skin."
6. As an admin, go back to the Approved Skins page, check "Not In Chooser" for the same skin, and press "Update." Then check the footer to make sure that the skin is no longer listed under "Choose Skin."


app/models/skin.rb
```diff
@@ -315,14 +325,16 @@ def get_all_parents
...
-        style += AdminSetting.default_skin != Skin.default ? '' : (Skin.get_current_site_skin ? Skin.get_current_site_skin.get_style(roles_to_include) : '')
-      end
-      style += self.get_style_block(roles_to_include)
-      style.html_safe
+    style = ""
+
+    if self.get_role != "override" && self.get_role != "site" &&
+       self.id != AdminSetting.default_skin_id &&
+       AdminSetting.default_skin.is_a?(Skin)
+      style += AdminSetting.default_skin.get_style(roles_to_include)
```
tickinginstant, line 333: I'm not sure if using `AdminSetting.default_skin` here instead of `Skin.get_current_site_skin` is the right idea. Currently, `AdminSetting.default_skin = Skin.default`, which is a blank skin with the `user` role, so using `Skin.get_current_site_skin` is the only way to make the skins work. But if we run the following command before the deploy:
```
Skin.default.update(role: "site", skin_parents_attributes: [{ position: 1, parent_skin_title: "Archive 2.0" }])
```
Then the "Default" skin will be a site skin that has "Archive 2.0" listed as a parent, so the contents of the "Archive 2.0" skin will be included whenever the "Default" skin is used.

This has a couple of advantages: (1) the relationship between the "Archive 2.0" skin and the "Default" skin is explicit, which should make it easier to expire caches when `Skin.get_current_site_skin` changes versions or adds new files, and (2) it means that we don't have to check the file system to look up the most recent site skin version number every time a skin's cache expires.

I'm still not certain this is the right decision, though -- input welcome!

app/helpers/skin_cache_helper.rb
```diff
@@ -1,42 +1,28 @@
...
-  def skin_memcache_cache_key(skin)
-    'skins_generation/' + (skin.type.nil? ? ("site_skin_#{skin.id.to_s}") : "work_skin_#{skin.id.to_s}")
+  def skin_cache_version_update(skin_id)
+    Rails.cache.write(skin_cache_version_key(skin_id), cache_timestamp)
   end
 
-  def skin_cache(skin)
-    Rails.cache.increment(skin_memcache_cache_key(skin))
-    Rails.cache.increment('skins_generation/site_skin')  # This is the general key used for the footer
+  def skin_chooser_key
```
sarken, line 21: For this, should we also test that the list of skins in the footer gets updated when a skin is added to/removed from that list?

tickinginstant: Huh, I thought there was a test for that already, but I guess not! I added some extra steps to the admin test for adding/removing from the chooser.

sarken: Oh, thanks! I was thinking about manual testing when I asked, but automated tests are good too!

tickinginstant: Oh, right, manual tests! I added two lines to the Testing Instructions section.

(I'll copy the instructions over to JIRA once this PR is further along. Some of the deploy/testing instructions will depend on how `Skin.get_current_site_skin` is handled, so I don't want to update the issue until that's decided.)

app/models/skin.rb
sarken: The only potential problem I can think of is the Default skin not properly updating when Archive 2.0 changes. 

Right now, if I am using a skin that has parent skins, and I make a change to one of the parents, then I don't see those changes until I edit the one I'm actually using. 

_However_, it's possible the re-caching james_ does on deploy will mean that's a non-issue.

zz9pzza: The recaching I do on deploy is writing the files out to disk, and is nothing to do with memcached.

tickinginstant: Ah, I think that's because editing a skin calls `cache!` to reconstruct the file system cache on edit, but doesn't do the same for any child skins: https://github.com/otwcode/otwarchive/blob/54c7496ea8f717b0434ce4e550a5c0211649c985/app/controllers/skins_controller.rb#L93
I can fix that.

redsummernight: That's [AO3-4551](https://otwarchive.atlassian.net/browse/AO3-4551), thanks for fixing it!

redsummernight: > - As an admin, go to Skins > Approved Skins and pick a new skin to use as the default. Make sure that the new skin is included in subsequent page loads, both for logged out users, and for users with a custom skin adding onto the default.
> - As an admin, go to Skins > Approved Skins, click the link for the current default skin, and then edit it to add some CSS. Make sure that the changes are immediately reflected on the next page you visit, both for logged-out users, and for users with a custom skin adding onto the default.

And that's [AO3-4829](https://otwarchive.atlassian.net/browse/AO3-4829).

redsummernight: > Then the "Default" skin will be a site skin that has "Archive 2.0" listed as a parent, so the contents of the "Archive 2.0" skin will be included whenever the "Default" skin is used.

With the old approach we need to test 2 cases:

- Admin default skin is `Skin.default`, include `Skin.get_current_site_skin`;
- Admin default skin is something else, include its `get_style` instead.

With the new approach we have just 1 test case:

- Admin default skin can be whatever, include its `get_style`.

I'm good with the new approach. Can you include the command to undo this if necessary?
```rb
Skin.default.update(role: "site", skin_parents_attributes: [{ position: 1, parent_skin_title: "Archive 2.0" }])
```

redsummernight (commented):
Never mind the Hakiri warning, we'll bump Rails in another PR.

app/models/skin.rb
tickinginstant: The command to undo it is more complicated, but I believe this should do it:
```
Skin.transaction do
  skin = Skin.default
  skin.parent_skins = []
  skin.update(role: "user")
end
```

redsummernight (approved):
Thanks!

(This is like your third PR within a year with BFS. :sunglasses:)

redsummernight:
Testing instructions:

Moved to [AO3-6051](https://otwarchive.atlassian.net/browse/AO3-6051):

> - As a user, make a custom skin that overrides the site skin, and start using it. Make sure that the skin is included in subsequent page loads. Then try editing the skin, and make sure that the changes are immediately reflected on the next page you visit.
> - As a user, make a custom skin that adds onto the site skin, and start using it. Make sure that the skin is included in subsequent page loads. Then try editing the skin, and make sure that the changes are immediately reflected on the next page you visit.
> - As an admin, go to Skins > Approved Skins, check "Chooser" for one of the listed skins, and press "Update." Then check the footer to make sure that the skin you selected is listed under "Choose Skin."
> - As an admin, go back to the Approved Skins page, check "Not In Chooser" for the same skin, and press "Update." Then check the footer to make sure that the skin is no longer listed under "Choose Skin."

Moved to [AO3-4829](https://otwarchive.atlassian.net/browse/AO3-4829):

> - As an admin, go to Skins > Approved Skins and pick a new skin to use as the default. Make sure that the new skin is included in subsequent page loads, both for logged out users, and for users with a custom skin adding onto the default.

Left out:

> - ~~As an admin, go to Skins > Approved Skins, click the link for the current default skin, and then edit it to add some CSS. Make sure that the changes are immediately reflected on the next page you visit, both for logged-out users, and for users with a custom skin adding onto the default.~~ **EDIT:** This won't necessarily work on staging/production because the filesystem isn't shared between different servers, so if the default skin is cached to the filesystem, the changes will only show up on the server that handled the `SkinsController#update` request, not any other servers.

PR #3903, opened 2020-10-03
AO3-5357 Address issues with the related works page & anonymous/unrevealed works.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5357

## Purpose

Modify the related works page so that it no longer displays anonymous/unrevealed remixes on the remixer's related works page, and no longer displays remixes inspired by an anonymous/unrelated work on the original creator's related works page. (The same also holds for translations.)

Since I was in the area, I also made a few other changes to the `related_works/index` template:
- Omitted `html_safe` on all of the titles, since that's unnecessary for both [works](https://github.com/otwcode/otwarchive/blob/5333e396ca4be8238d40369d8533789bda6843dd/app/models/work.rb#L74-L77) and [external works](https://github.com/otwcode/otwarchive/blob/5333e396ca4be8238d40369d8533789bda6843dd/app/models/external_work.rb#L49-L52).
- Changed the phrasing from "A work in an unrevealed challenge" to "A work in an unrevealed collection," as suggested in [AO3-3929](https://otwarchive.atlassian.net/browse/AO3-3929). (Though that's the only part of that issue I tackled.)
- Changed two links with `method: :delete` to use `button_to`, to support browsers with JS disabled. [The redirect after clicking the button is still broken, but that's a separate issue.](https://otwarchive.atlassian.net/browse/AO3-2151)

### DIFF
--- a/app/views/related_works/index.html.erb
+++ b/app/views/related_works/index.html.erb
@@ -144,7 +144,7 @@
             <%= ts("inspired") %>
             <%= link_to related_work.work.title, related_work.work %>
             <% if current_user == @user %>
-              <span class="actions" role="button"><%= button_to ts("Remove"), related_work_path(related_work), data: { confirm: ts("Are you sure?") }, method: :delete %></span>
+              <span class="actions"><%= button_to ts("Remove"), related_work_path(related_work), data: { confirm: ts("Are you sure?") }, method: :delete %></span>
             <% end %>
           </dd>
         <% end %>
--- a/app/views/related_works/index.html.erb
+++ b/app/views/related_works/index.html.erb
@@ -36,10 +36,10 @@
       <% if related_work.parent && related_work.work %>
       <tr>
         <% if related_work.work.unrevealed? %>
-          <td><%= ts("A work in an unrevealed challenge") %></td>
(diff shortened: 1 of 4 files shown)


app/views/related_works/index.html.erb
```diff
@@ -132,23 +134,17 @@
...
-              <%= ts("A work in an unrevealed or anonymous challenge") %>
-            <% else %>
-              <%= link_to related_work.work.title, related_work.work %>
-            <% end %>
+          <dd class="child">
+            <%= ts("inspired") %>
+            <%= link_to related_work.work.title, related_work.work %>
             <% if current_user == @user %>
-              <span class="actions" role="button"><%= link_to ts("Remove"), {:controller => :related_works, :action => :destroy, :id => related_work.id},  data: {confirm: ts('Are you sure?')}, :method => :delete %></span>
+              <span class="actions" role="button"><%= button_to ts("Remove"), related_work_path(related_work), data: { confirm: ts("Are you sure?") }, method: :delete %></span>
```
sarken, line 147: Could you remove the `role="button"` part? I'm not sure it ever should have been used here, but it definitely is unnecessary now that we have an actual button.

sarken (approved):

PR #3907, opened 2020-10-12
AO3-5685 Use DELETE method when logging out.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5685

## Purpose

Change user/admin logout link to use `method: :delete`. In order to handle browsers with JS disabled, the PR also includes two new routes to confirm whether the user/admin wants to logout.

## References

This is an updated version of #3834. The only things I changed were to try to make the automated tests pass.

## Credit

Credit should go to Alix R.


tickinginstant:
No problem!

My original plan was to replace the "I am logged out" step with the "I log out" step, since I think the "I log out" phrasing makes it clearer that the step involves clicking the log out link. (To me, the passiveness of "I am logged out" makes it sound a little fuzzier about how the logging out is occurring.) But that would require changing all of the existing uses of the step (making the PR harder to review, and potentially causing conflicts with other PRs), so I left in both steps. Not sure whether or not that's the right decision.

features/step_definitions/user_steps.rb
```diff
@@ -138,9 +138,12 @@
   user.save
 end
 
+Given /^I start a new session$/ do
+  page.driver.reset!
+end
+
 Given /^I am logged out$/ do
```
redsummernight, line 145: Keep both logging out steps for now is fine. Can you add a TODO on one of them saying it should be eventually removed?

features/step_definitions/user_steps.rb
```diff
@@ -230,8 +232,8 @@
 end
 
 Then /^a user account should exist for "(.*?)"$/ do |login|
-   user = User.find_by(login: login)
-   assert !user.blank?
+  user = User.find_by(login: login)
+  assert user.present?
```
redsummernight, line 236: Can you switch this to an RSpec expect instead? (We should remove minitest assertions at some point.)

redsummernight (commented):
Looks good, thanks! Just two minor comments.

redsummernight (approved):

PR #3908, opened 2020-10-13
AO3-5950 Set "do not track" parameter and update homepage Twitter widget

palindromordnilap:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5950

## Purpose

The Twitter timeline widget on the home page currently doesn't have the "Do Not Track" parameter set, and takes up way too much space on the page due to a missing `data-height` attribute. This PR fixes both issues by adding the `data-dnt="true"` attribute and setting the widget's height to the previous default of 600 pixels.

## Testing Instructions

Check if the widget is displaying normally again (with a reasonable height). As mentioned on the Jira issue, there's not much of a way to see if the DNT parameter works as intended, but we can at least check if the cookies set by Twitter are different.

## Credit

Alix R, she/her

### DIFF
--- a/app/views/home/index.html.erb
+++ b/app/views/home/index.html.erb
@@ -50,7 +50,7 @@
 
   <div class="latest tweets module">
     <h3 class="heading"><%= ts('Tweets') %></h3>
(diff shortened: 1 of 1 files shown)


sarken (approved):
Yay, thank you!

PR #3911, opened 2020-10-25
AO3-6058 Hide byline of unrevealed work in notes of child related work

zz9pzza:
# Pull Request Checklist

## Issue

## Purpose

What does this PR do?

## Testing Instructions

See issue.

### DIFF
--- a/app/views/downloads/_download_preface.html.erb
+++ b/app/views/downloads/_download_preface.html.erb
@@ -66,13 +66,16 @@
       <ul>
         <% related_works.each do |work| %>
           <li>
-            <% if work.parent.unrevealed? then
-                 relation_text = "a work in an unrevealed collection."
-               else
-                 relation_text = work.translation ?
-                                     "A translation of %{work_link} by %{author_link}" :
-                                     "Inspired by %{work_link} by %{author_link}"
-               end %>
+            <%  work_info = if work.parent.unrevealed?
+                              "a work in an unrevealed collection"
+                            else
+                              "%{work_link} by %{author_link}"
+                            end
+                relation_text = if work.translation
+                                  "A translation of %{work_info}."
+                                else
+                                  "Inspired by %{work_info}."
+                                end %>
             <% work_link = link_to work.parent.title.html_safe, url_for(work.parent) %>
             <% author_link = byline(work.parent, visibility: 'public', only_path: false) %>
 
--- a/app/views/downloads/_download_preface.html.erb
+++ b/app/views/downloads/_download_preface.html.erb
@@ -79,7 +79,7 @@
             <% work_link = link_to work.parent.title.html_safe, url_for(work.parent) %>
             <% author_link = byline(work.parent, visibility: 'public', only_path: false) %>
 
-            <%= ts(relation_text, work_link: work_link, author_link: author_link).html_safe %>
+            <%= ts(relation_text, work_link: work_link, author_link: author_link, work_info: work_info).html_safe %>
           </li>
         <% end %>
       </ul>
--- a/app/views/downloads/_download_preface.html.erb
+++ b/app/views/downloads/_download_preface.html.erb
@@ -66,10 +66,13 @@
       <ul>
         <% related_works.each do |work| %>
           <li>
-            <% relation_text = work.translation ?
-              "A translation of %{work_link} by %{author_link}" :
-              "Inspired by %{work_link} by %{author_link}"
-            %>
+            <% if work.parent.unrevealed? then
+                 relation_text = "a work in an unrevealed collection."
+               else
+                 relation_text = work.translation ?
+                                     "A translation of %{work_link} by %{author_link}" :
+                                     "Inspired by %{work_link} by %{author_link}"
+               end %>
             <% work_link = link_to work.parent.title.html_safe, url_for(work.parent) %>
             <% author_link = byline(work.parent, visibility: 'public', only_path: false) %>
 
--- a/app/views/works/_work_header_notes.html.erb
+++ b/app/views/works/_work_header_notes.html.erb
@@ -38,10 +38,10 @@
 
           <% if related_work.parent.is_a?(ExternalWork) %>
             <%= link_to related_work.parent.title.html_safe, related_work.parent %> by <%= byline(related_work.parent) %>.
+          <% elsif related_work.parent.unrevealed? %>
+            <%= ts("a work in an unrevealed collection.") %>
           <% elsif related_work.parent.restricted? && !logged_in? %>
             <%= ts("[Restricted Work] by") %> <%= byline(related_work.parent) %>. <%= ts("Log in to view.") %>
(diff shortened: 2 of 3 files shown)


app/views/works/_work_header_notes.html.erb
```diff
@@ -40,6 +40,8 @@
             <%= link_to related_work.parent.title.html_safe, related_work.parent %> by <%= byline(related_work.parent) %>.
           <% elsif related_work.parent.restricted? && !logged_in? %>
             <%= ts("[Restricted Work] by") %> <%= byline(related_work.parent) %>. <%= ts("Log in to view.") %>
+          <% elsif (related_work.parent.unrevealed? ||related_work.parent.anonymous?) %>
```
sarken, line 43: There's a space missing after the `||`

app/views/works/_work_header_notes.html.erb
```diff
@@ -40,6 +40,8 @@
...
+            <%= ts("Inspired by a work in an unrevealed collection.") %>
```
sarken, line 44: Because the words "Inspired by" are already included on line 36, I think this is going to result in text that says, "Inspired by Inspired by a work in an unrevealed collection." 

This line probably just needs to say "a work in an unrevealed collection," much like the other lines just include the work titles.

Also, I think we might need to move this up. If the unrevealed work is restricted and you're logged out, I'm pretty sure the text is going to say, "Inspired by [Restricted Work] by CREATOR.  Log in to view." The fact that it's unrevealed should trump the fact that it's restricted.

features/works/work_related.feature
```diff
@@ -583,3 +583,14 @@ Scenario: When a user is notified that a co-authored work has been inspired by a
       And I should not see "A work in an unrevealed collection"
       And I should not see "Worldbuilding Translated by translator"
       And I should not see "From English to Deutsch"
+
+  Scenario: A work in a hidden collection should not be revealed by posting a related work
```
sarken, line 587: Could we rephrase this for clarity? Right now, it sounds like we're checking that the work itself isn't getting revealed. "Scenario: A work that is inspired by a work in an unrevealed collection does not list the title or creator of the unrevealed work in its notes" or something similar might be clearer.

sarken (changes requested):
One of these days, I will not press the "Single Comment" button when I mean to use "Start Review." 😩

tickinginstant:
The title is also exposed [in downloads](https://github.com/otwcode/otwarchive/blob/a69c666cbe31ea9cae8087eb09fc64124320af11/app/views/downloads/_download_preface.html.erb#L69-L76), as I mentioned [on Jira](https://otwarchive.atlassian.net/browse/AO3-6058?focusedCommentId=356869).

sarken (commented):
I think what's here is good, but it needs the downloads fix ticking mentioned as well.

app/views/downloads/_download_preface.html.erb
```diff
@@ -66,10 +66,13 @@
       <ul>
         <% related_works.each do |work| %>
           <li>
-            <% relation_text = work.translation ?
-              "A translation of %{work_link} by %{author_link}" :
-              "Inspired by %{work_link} by %{author_link}"
-            %>
+            <% if work.parent.unrevealed? then
```
sarken, line 69: This won't quite work. The downloads currently show either:

* `A translation of %{work_link} by %{author_link}`
* `Inspired by %{work_link} by %{author_link}`

What we're trying to do is -- for unrevealed works -- replace _just_ the `%{work_link} by %{author_link}` portion with `a work in an unrevealed collection`. 

But what you have will write `a work in an unrevealed collection` without the words `A translation of` or `Inspired by` in front of it.

(Also, minor code style note: we typically don't use `then`.)

So basically we'd need to divide it up more like: 

```
work_info = if work.parent.unrevealed?
              "a work in an unrevealed collection"
            else
              "%{work_link} by %{author_link}"
            end

relation_text = if work.translation
                  "A translation of %{work_info}."
                else
                  "Inspired by %{work_info}."
                end
```

sarken:
The tests are failing, but that's understandable... that wasn't meant to be functioning code to copy and paste, just an illustration of how to think about it -- sorry for being unclear!

app/views/downloads/_download_preface.html.erb
```diff
@@ -66,14 +66,20 @@
...
+                relation_text = if work.translation
+                                  "A translation of %{work_info}."
+                                else
+                                  "Inspired by %{work_info}."
+                                end %>
             <% work_link = link_to work.parent.title.html_safe, url_for(work.parent) %>
             <% author_link = byline(work.parent, visibility: 'public', only_path: false) %>
 
-            <%= ts(relation_text, work_link: work_link, author_link: author_link).html_safe %>
+            <%= ts(relation_text, work_link: work_link, author_link: author_link, work_info: work_info).html_safe %>
```
tickinginstant, line 82: Does this work when the parent isn't unrevealed? I would have thought I18n would stop after the first substitution.

tickinginstant:
There's still an issue here with the expiration of downloads. IIRC, downloads are cached for a year, and they're only expired when the work is touched (i.e. when `updated_at` changes). If you want the child work's downloads to be updated reasonably promptly when the parent is revealed (or, perhaps more critically, when the parent becomes unrevealed), there needs to be some mechanism for expiring them.

(Alternatively, if it's okay that the child work's downloads won't get updated for a year, then it might be good to note that in the issue, which currently says, "Please make sure the work title eventually appears once the work is revealed!")

zz9pzza:
Given that I think this is really to stop people creating works only to get the other work information, I am not too worried.  However.

app/views/downloads/_download_preface.html.erb
sarken: I've spoke with Translation, and if it turns out it doesn't work, it would actually be better to have this i18ned like:

```
<% relation = work.translation ? "translation_of" : "inspired_by" %>

<% if work.parent.unrevealed? %>
  <%= t(".#{relation}.unrevealed") %>
<% else %>
  <%= t(".#{relation}.revealed", work_link: work_link, creator_link: author_link) %>
<% end %>
```

Giving Translation this:
```
  inspired_by:
    revealed: Inspired by %{work_link} by %{creator_link}.
    unrevealed: Inspired by a work in an unrevealed collection.
  translation_of:
    revealed: Translation of %{work_link} by %{creator_link}.
    unrevealed: Translation of a work in an unrevealed collection.
```

sarken:
I don't think it would be the end of the world to wait a year if we were only going from unrevealed -> revealed, but like ticking said, we're also looking at revealed -> unrevealed. Those are situations where the creator has, for some reason, decided to make their work no longer available to the public... so yeah, I think it's going to be important to expire the download.

sarken:
It looks like we need to handle download expiration and fix the translations, so I'm flipping this back to Needs Coder Action.

PR #3825, opened 2020-05-09
AO3-5828 Add IP whitelist for Rack Attack

zz9pzza:
# Pull Request Checklist

* [x] Have you read [How to write the perfect pull request](https://github.com/blog/1943-how-to-write-the-perfect-pull-request)?
* [x] Have you read through the [contributor guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [ ] Have you added tests for any changed functionality?
* [x] Have you added the [JIRA](https://otwarchive.atlassian.net) issue number as the *first* thing in your pull request title (eg: `AO3-1234 Fix thing`)
* [x] Have you updated or commented on the JIRA issue with the information below?

## Issue

https://otwarchive.atlassian.net/browse/AO3-5828

## Purpose

Allows us to white list specific ip address 

## Testing Instructions

Have a ip address range of the tester added to the config and show that the rate limit does not apply.

PR #3912, opened 2020-10-29
AO3-6026 Move collections to Elasticsearch

AncyentMariner:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6026

## Purpose

What does this PR do?
Creates models for creating and querying elasticsearch indexes for collections.
Modifies collections indexes to use ES to query collections, instead of sql

## Testing Instructions

Run the command to reindex all collection `CollectionIndexer.index_all` to see these results in the app.
Also, run tests for collection search models and collections controller

(If you have a Jira account with access, please update or comment on the issue
with any new or missing testing instructions.)

## References

## Credit

(If you have a Jira account, please include the same name in the "Full name"
field on your Jira profile, so we can assign you the issues you're working on.)

(If you do not fill in this section, we will use your GitHub account name and
they/them pronouns to avoid making assumptions about your gender.)


app/controllers/collections_controller.rb
```diff
@@ -173,6 +170,30 @@ def destroy
...
+      end
+      
+      new_hash
+    end 
+
+    parsed_filters.symbolize_keys
+  end
+
+  def collection_filter_params
+    white_list = %w(title fandom challenge_type moderated closed sort_column sort_direction)
```
sarken, line 190: We're trying to avoid whitelist/blacklist terminology going forward... can we use `safelist` instead? That's what we went with in #3825

app/decorators/collection_decorator.rb
```diff
@@ -0,0 +1,43 @@
+class CollectionDecorator < SimpleDelegator
+
+  attr_reader :data
+
+  # Pseuds need to be decorated with various stats from the "_source" when
```
sarken, line 5: Should this be "Collections need to be..."?

app/models/search/collection_indexer.rb
```diff
@@ -0,0 +1,115 @@
...
+    object.as_json(
+      root: false,
+      only: [
+        :id, :name, :title, :description, :parent_id, :collection_type, :created_at
+      ]
+    ).merge(
+      closed: object.closed?,
+      unrevealed: object.unrevealed?,
+      anonymous: object.anonymous?,
+      owner_ids: object.all_owners.pluck(:user_id),
```
sarken, line 78: Just noting that we originally wanted to use pseud ids here and in `moderator_ids`, but I can't think of a reason that's actually necessary -- users' collections pages are indeed for the user, not the pseud. 

However, it might be worth double checking how this interacts with orphaning?

app/models/search/collection_query.rb
```diff
@@ -0,0 +1,103 @@
...
+
+  def parent_filter
+    match_filter(:parent_id, options[:parent_id]) if options[:parent_id].present?
+  end
+
+  ####################
+  # QUERIES
+  ####################
+
+  # Search for a tag by name
```
sarken, line 75: "Search for a collection by name or title," maybe? Also, I don't think the example of `complete:true` two lines down is relevant.

app/models/search/indexer.rb
```diff
@@ -7,7 +7,8 @@ class Indexer
     "Tag" => %w(TagIndexer),
     "Pseud" => %w(PseudIndexer),
     "Series" => %w(BookmarkedSeriesIndexer),
-    "ExternalWork" => %w(BookmarkedExternalWorkIndexer)
+    "ExternalWork" => %w(BookmarkedExternalWorkIndexer),
+    "Collection" => %w(CollectionIndexer),
```
sarken, line 11: Should we skip the comma on this last line? I think that's our usual style.

app/models/search/collection_search_form.rb
```diff
@@ -0,0 +1,114 @@
...
+  def search_results
+    @searcher.search_results
+  end
+
+  ###############
+  # SORTING
+  ###############
+
+  SORT_OPTIONS = [
+    ['Date Posted', 'created_at'],
```
sarken, line 78: This should probably be "Date Created" -- "posted" is more specific to works. (+1 to Hound's comment about the quotes.)

sarken (commented):
Thank you!

I just did a quick read through and left a few very minor comments -- I'll try to take a more thorough look before the weekend is over.

app/controllers/application_controller.rb
```diff
@@ -488,7 +488,7 @@ def valid_sort_column(param, model='work')
     elsif model.to_s.downcase == 'tag'
       allowed = %w(name created_at taggings_count_cache)
     elsif model.to_s.downcase == 'collection'
```
redsummernight, line 490: Is the "collection" case still used or can we remove it?

app/models/collection.rb
```diff
@@ -490,4 +491,10 @@ def clear_icon
     self.icon = nil if delete_icon? && !icon.dirty?
   end
 
+  after_save :reindex_collection
```
redsummernight, line 494: Should we remove these callbacks? The Searchable module set up similar callbacks:

https://github.com/otwcode/otwarchive/blob/63adfaf4641a6fc9c2e24911d86a563fe8867508/lib/searchable.rb#L5-L6

app/models/challenge_models/gift_exchange.rb
```diff
@@ -80,4 +80,21 @@ def user_allowed_to_see_requests_summary?(user)
   def user_allowed_to_see_prompt?(user, prompt)
     self.collection.user_is_maintainer?(user) || prompt.pseud.user == user
   end
+
+  # reindex collection after creation, deletion, and certain attribute updates
```
redsummernight, line 84: Can you move this to ChallengeCore, since it's shared by both types of challenges?

app/models/challenge_models/gift_exchange.rb
```diff
@@ -80,4 +80,21 @@ def user_allowed_to_see_requests_summary?(user)
...
+  after_create :reindex_collection
+  after_destroy :reindex_collection
+  after_update :reindex_collection, if: Proc.new { |c| c.saved_change_to_signup_open? ||
```
redsummernight, line 87: You can write the `saved_change_` checks like this, where you provide a list of attributes (`id` replaces `after_create`):

https://github.com/otwcode/otwarchive/blob/f057b6b9f85ccdd26dcff2aca2efa17cf7f38e26/app/models/pseud.rb#L478-L486

redsummernight, line 87: You should use `after_commit` for the new reindexing callbacks if possible. We plan to update the existing ones as well ([AO3-5944](https://otwarchive.atlassian.net/browse/AO3-5944)).

app/models/collection_item.rb
```diff
@@ -342,4 +342,13 @@ def notify_of_unrevealed_or_anonymous
       ).deliver_after_commit
     end
   end
+
+  # reindex collection after creation, deletion, and approval_status update
+  after_create :reindex_collection
+  after_destroy :reindex_collection
+  after_update :reindex_collection, if: Proc.new { |c| c.saved_change_to_user_approval_status? || c.saved_change_to_collection_approval_status? }
```
redsummernight, line 349: Similar to the challenge classes, you can provide a list of attributes to check (`saved_changes.keys & pertinent_attributes`), and you should use an `after_commit` callback.

app/views/collections/_filters.html.erb
```diff
@@ -27,44 +27,44 @@
...
       <dt class="autocomplete search">
-        <%= label_tag "collection_filters_title",
-            ts("Filter by title or name") %>
+        <%= label_tag "title",
+            ts("Filter by title") %>
       </dt>
       <dd class="autocomplete search">
-        <%= text_field_tag "collection_filters[title]",
-            params[:collection_filters][:title],
+        <%= text_field_tag "title",
```
redsummernight, line 34: Instead of using `_tag` helpers, you can rewrite this view to use `form_for` with the new CollectionSearchForm class. You can reference [the work filters view](https://github.com/otwcode/otwarchive/blob/f057b6b9f85ccdd26dcff2aca2efa17cf7f38e26/app/views/works/_filters.html.erb#L1-L7) for example.

features/collections/collection_browse.feature
```diff
@@ -58,7 +58,7 @@ Feature: Collection
 
   Given I have loaded the fixtures
   When I go to the collections page
-    And I choose "collection_filters_closed_true"
+    And I choose "closed_true"
```
redsummernight, line 61: In .feature files, can you use the field's label text instead of the ID?

app/decorators/collection_decorator.rb
```diff
@@ -0,0 +1,43 @@
...
+  # viewing search results, so we first load the pseuds with the base search
+  # class, and then decorate them with the data.
+  def self.load_from_elasticsearch(hits)
+    items = Collection.load_from_elasticsearch(hits)
+    decorate_from_search(items, hits)
+  end
+
+  def self.decorate_from_search(results, search_hits)
```
redsummernight, line 13: Can you keep the comment `# TODO: pull this out into a reusable module` on this method, similar to the pseud version?

app/controllers/collections_controller.rb
```diff
@@ -27,42 +27,39 @@ def index
...
       @collections = @work.approved_collections.by_title.includes(:parent, :moderators, :children, :collection_preference, owners: [:user]).paginate(page: params[:page])
     elsif params[:collection_id] && (@collection = Collection.find_by(name: params[:collection_id]))
-      @collections = @collection.children.by_title.includes(:parent, :moderators, :children, :collection_preference, owners: [:user]).paginate(page: params[:page])
+      @collections = CollectionSearchForm.new({ parent_id: @collection.id }).search_results.to_a.paginate(page: params[:page])
     elsif params[:user_id] && (@user = User.find_by(login: params[:user_id]))
-      @collections = @user.maintained_collections.by_title.includes(:parent, :moderators, :children, :collection_preference, owners: [:user]).paginate(page: params[:page])
+      @collections = CollectionSearchForm.new(moderator_ids: [@user.id]).search_results.to_a.paginate(page: params[:page])
       @page_subtitle = ts("%{username} - Collections", username: @user.login)
     else
       if params[:user_id]
```
redsummernight, line 35: Let's remove all these `flash.now[:error]` checks. Instead can you change the 3 `find_by`s into `find_by!`s to raise 404 errors?

app/controllers/collections_controller.rb
```diff
@@ -27,42 +27,39 @@ def index
...
     end
   end
 
   # display challenges that are currently taking signups
   def list_challenges
     @page_subtitle = "Open Challenges"
     @hide_dashboard = true
-    @challenge_collections = (Collection.signup_open("GiftExchange").limit(15) + Collection.signup_open("PromptMeme").limit(15))
+
+    @challenge_collections = (CollectionSearchForm.new(challenge_type: 'GiftExchange').search_results.take(15) +
```
redsummernight, line 51: Can you also use Query's pagination here instead of `take`s?

redsummernight (commented):
Thanks for working on this!

app/models/challenge_models/gift_exchange.rb
AncyentMariner: Is there a trick to working with this shared module. I couldn't get it to recognize the activerecord methods necessary for this. I kept getting: undefined method `saved_changes' for #<Class:0x000000000804d960>`

app/controllers/application_controller.rb
AncyentMariner: This is still used

features/collections/collection_browse.feature
AncyentMariner: For those radio buttons, the text of the label and the id are the same.

app/models/search/collection_indexer.rb
```diff
@@ -0,0 +1,115 @@
...
+      challenge_type: object.challenge_type,
+      signup_open: object.challenge&.signup_open,
+      signups_open_at: object.challenge&.signups_open_at,
+      signups_close_at: object.challenge&.signups_close_at,
+      assignments_due_at: object.challenge&.assignments_due_at,
+      works_reveal_at: object.challenge&.works_reveal_at,
+      authors_reveal_at: object.challenge&.authors_reveal_at,
+      general_fandom_ids: get_fandom_ids(object),
+      public_fandom_ids: get_fandom_ids(object, true),
+      general_fandoms_count: get_fandom_ids(object).count,
```
sarken, line 90: I don't think using `get_fandom_ids(object).count` will work here, since [the issue](https://otwarchive.atlassian.net/browse/AO3-6026)/SOW notes:

>general_fandoms_count is the total number of fandoms on all public and restricted works approved for inclusion in the collection or any of its subcollections. Unlike the general_fandom_ids, this field counts only direct filter_taggings. It will be used in blurbs to display the fandom count to logged-in users.

For example...

Let's say we have the canonical fandom tag "Marvel Cinematic Universe" and it's the metatag of the canonical fandom "Captain America Movies."

If I post a work with the "Captain America Movies" fandom tag, it will have a direct [filter_tagging](https://github.com/otwcode/otwarchive/wiki/Guide:-Tag-Code#filtertaggings) for that, but _also_ an inherited filter_tagging for "Marvel Cinematic Universe." That way, if someone goes to the work page for the "Marvel Cinematic Universe" tag, they will see my work about Captain America listed there.

However, if I make the collection "Best Captain America Works Ever" and add my work to it, when I go to my collection's Fandoms page (located at a path like `/collections/best_captain_america_works_ever/fandoms`), the only fandom listed there will be "Captain America Movies" -- the [list on that page is drawn from _direct_ filter taggings](https://github.com/otwcode/otwarchive/blob/cbfa1db7e6dd16d9ca2807d63bed7d736aa88f5e/app/controllers/fandoms_controller.rb#L12-L13).

So the idea is general_fandoms_count and public_fandoms_count should align with the number of fandoms you'll see when you go to the collection's Fandoms page. Otherwise, we'll end up in a situation where someone sees "Fandoms: 2" on the blurb for the "Best Captain America Works Ever" collection, but when they go to /collections/best_captain_america_works_ever/fandoms, they'll only see one fandom, leading them to believe there's a) content they can't see or b) a bug.

AncyentMariner: Thanks for this feedback. I was worried that this data wouldn't be quite correct. I had a bit of a hard time wrapping my head around the structure of the data, by just looking at the methods on the collection model.
Do you have any tips for how to compile the data correctly for these methods (and the comment above)?

sarken: Before I try to be helpful, two notes:
1. Someone else might come along and say doing it this way will make the servers cry and there's a better way to do it (I'm guessing a complex join), in which case you should listen to them.
2. Someone else might come along and say this is a terrible abuse of  these methods, in which case you should also listen to them.

That said, [Filterable](https://github.com/otwcode/otwarchive/blob/master/app/models/concerns/filterable.rb) has some methods that might be useful. You can use them on both works and bookmarkables.

Here's a messy example I knocked together on the console for calculating a collection's `general_fandom_ids`, which includes both inherited and direct Fandom filter taggings:
```ruby
bookmarkable_fandom_ids = []
work_fandom_ids = []

approved_bookmarks.map do |bookmark|
  bookmarkable_fandom_ids << bookmark.bookmarkable.filters.where(type: "Fandom").pluck(:filter_id)
end

approved_works.map do |work|
  work_fandom_ids << work.filters.where(type: "Fandom").pluck(:filter_id)
end

general_fandom_ids = [work_fandom_ids + bookmarkable_fandom_ids].flatten.uniq
```

For the `general_fandoms_count`, which only needs direct filter taggings, we can replace `filters` with [`direct_filters`](https://github.com/otwcode/otwarchive/blob/3785907b55d3d844e4d73deb54f0a8f8c3d9a44b/app/models/concerns/filterable.rb#L17-L19) and then count the results.... or [`fandom_ids`](https://github.com/otwcode/otwarchive/blob/3785907b55d3d844e4d73deb54f0a8f8c3d9a44b/app/models/concerns/filterable.rb#L91-L93) should also work and be much shorter, e.g. `work.fandom_ids`.

It's probably also worth noting that just [`collection.fandoms`](https://github.com/otwcode/otwarchive/blob/3785907b55d3d844e4d73deb54f0a8f8c3d9a44b/app/models/collection.rb#L64) will get you the direct filter tagging Fandoms for all approved works in the collection.

And one last note before you get too involved in working on the counts: I just noticed something in our SOW/issue _might_ be wrong. Currently, we say that the counts should include fandoms listed on bookmarkables. However, right now, the Fandoms page for a collection only includes fandoms used on _works_. I'm going to check  with everyone else, but I'm thinking we might want to stick with the works-only approach for counts, or else we're going to have people telling us, "The blurb said the collection had 10 fandoms, but the sidebar only says 2 and there are only 2 on the Fandoms page."

AncyentMariner: This is all really helpful. So it looks like this _could_ be the answer, but I should hold off on making changes until you check in with the other devs?

sarken: Yeah, I think we might end up scrapping this fandoms thing, so please hold off!

app/controllers/collections_controller.rb
```diff
@@ -24,45 +24,37 @@ def load_collection_from_id
...
+    if params[:work_id] && (@work = Work.find_by!(id: params[:work_id]))
       @collections = @work.approved_collections.by_title.includes(:parent, :moderators, :children, :collection_preference, owners: [:user]).paginate(page: params[:page])
-    elsif params[:collection_id] && (@collection = Collection.find_by(name: params[:collection_id]))
-      @collections = @collection.children.by_title.includes(:parent, :moderators, :children, :collection_preference, owners: [:user]).paginate(page: params[:page])
-    elsif params[:user_id] && (@user = User.find_by(login: params[:user_id]))
-      @collections = @user.maintained_collections.by_title.includes(:parent, :moderators, :children, :collection_preference, owners: [:user]).paginate(page: params[:page])
+    elsif params[:collection_id] && (@collection = Collection.find_by!(name: params[:collection_id]))
+      @collections = CollectionSearchForm.new({ parent_id: @collection.id }).search_results
+    elsif params[:user_id] && (@user = User.find_by!(login: params[:user_id]))
+      @collections = CollectionSearchForm.new(moderator_ids: [@user.id]).search_results
```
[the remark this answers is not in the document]
sarken: If I'm the owner or moderator of Parent Collection, but I set my role to something else (e.g. Member or None) in Subcollection, it's true that Subcollection isn't currently listed on my collections page. However, I do maintain owner or moderator powers in Subcollection. 

So I'm going to say it's fine to include the Subcollection on my collections page... and we might want to, at some future point, prevent someone who owns/moderates Parent Collection from downgrading their role in Subcollection.

app/controllers/collections_controller.rb
```diff
@@ -24,45 +24,37 @@ def load_collection_from_id
...
     end
   end
 
   # display challenges that are currently taking signups
   def list_challenges
     @page_subtitle = "Open Challenges"
     @hide_dashboard = true
-    @challenge_collections = (Collection.signup_open("GiftExchange").limit(15) + Collection.signup_open("PromptMeme").limit(15))
+
+    @challenge_collections = (CollectionSearchForm.new(challenge_type: 'GiftExchange', page: 1, per_page: 15).search_results.to_a +
```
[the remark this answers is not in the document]
sarken: I think we want to keep the current behavior for now. If that's difficult to do, any changes should move us toward the behavior described in [AO3-5100](https://otwarchive.atlassian.net/browse/AO3-5100).

app/models/search/collection_indexer.rb
sarken: Okay, hi, I have returned! 

We had a talk about just how much reindexing we'd have to do, and we've decided we no longer want to deal with `general_fandom_ids`, `public_fandom_ids`, `general_fandoms_count`, or `public_fandoms_count`. 

What we'd like to do instead is make the collections themselves taggable and filterable. We're still working on exactly what that will entail, but once we figure that out, we'll give you a detailed rundown and you can give us an updated estimate and let us know when you'll have time to work on it. (Assuming you want to, of course!) 

I think we can have that info for you within the next 2.5 weeks (I was going to say 2, but then I realized that would land us right on the U.S. Thanksgiving holiday).

For now, though, we're definitely sure you can remove 
* these fields, as well as the `get_fandom_ids` method and any code for reindexing based on changes to the collection's works' or bookmarks' fandoms
* the corresponding "Filter by fandom" section of `app/views/collections/_filters.html.erb`
* [the fandom count on `app/views/collections/_collection_blurb.html.erb`](https://github.com/otwcode/otwarchive/blob/b5bf2e7a5cd362d1e350f2be087e8049721a2484/app/views/collections/_collection_blurb.html.erb#L46-L49)

app/models/search/query_result.rb
```diff
@@ -27,8 +27,13 @@ def items
 
   # Laying some groundwork for making better use of search results
   def decorate_items(items)
```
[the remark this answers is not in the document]
AncyentMariner: Gotcha. The SOW specifically asked us to set up a decorator, in the style of that one, but if I should add them to the CollectionQuery class, I can. That definitely makes sense.

app/controllers/collections_controller.rb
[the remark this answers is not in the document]
AncyentMariner: Just clarifying, is there something that should be changed here? Should it check for owners and moderators in that check? Or is the code fine as is?

app/controllers/collections_controller.rb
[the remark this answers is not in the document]
AncyentMariner: Just following up on this, as well. By "current behavior" do you mean what's in production, meaning I should change the code we added? Or are is this code going to work?

features/collections/collection_browse.feature
```diff
@@ -18,7 +18,8 @@ Feature: Collection
   Scenario: Filter collections index to only show prompt memes
 
   Given I have loaded the fixtures
-  When I go to the collections page
+  When all indexing jobs have been run
+    And I go to the collections page
```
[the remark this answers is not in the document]
AncyentMariner: That's probably a question for @sarken. What do you think?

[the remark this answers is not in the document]
sarken: Changing it like ticking suggests sounds great to me!

app/controllers/collections_controller.rb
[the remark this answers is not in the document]
sarken: Oops, thanks for checking -- I meant the code should be changed to match the behavior on production.

app/controllers/collections_controller.rb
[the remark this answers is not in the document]
sarken: I think it needs to be changed to address ticking's first point:

>The original code looks for any collections owned by or moderated by the given user, but if I'm reading the CollectionSearchForm/CollectionQuery/CollectionIndexer code correctly, filtering by moderator_ids will only return collections where the user has the role Moderator, not collections where the user has the role Owner. (The two roles are mutually exclusive, IIRC.)

That is, we need to include collections where `@user.id` is in _either_ the `moderator_ids` _or_ the `owner_ids`.

app/models/search/query_result.rb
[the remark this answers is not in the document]
AncyentMariner: @sarken Any thoughts on this?

AncyentMariner:
@sarken I have one outstanding question about the collection decorator, but I believe I've resolved all of the other comments. Is there anything else I should do for the PR to get it ready/ approved?

app/models/search/query_result.rb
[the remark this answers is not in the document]
sarken: I've checked with ticking to clarify what needed to be done here, and she says:

>The `PseudDecorator` is used for pseud searches because it's referenced in the `PseudQuery` file; to make the `CollectionDecorator` be used for collection searches, it needs to be referenced in the `CollectionQuery` file. 
>
>(We probably should remove the `decorate_items` method, to prevent this sort of confusion.)

So if you can make those changes, that would be great!

features/users/user_delete.feature
```diff
@@ -60,6 +60,7 @@ Scenario: Delete a user with a collection
   When I am logged in as "moderator" with password "password"
     And all emails have been delivered
     And I create the collection "fake"
+    And all indexing jobs have been run
```
sarken, line 63: This can be removed now that `I go to the collections page` runs the indexing jobs.

features/users/user_delete.feature
```diff
@@ -71,7 +72,8 @@ Scenario: Delete a user with a collection
     And 0 emails should be delivered
     And I should be logged out
     And a user account should not exist for "moderator"
-  When I go to the collections page
+  When all indexing jobs have been run
```
sarken, line 75: This can be removed now that `I go to the collections page` runs the indexing jobs.

features/prompt_memes_c/challenge_promptmeme_deletion.feature
```diff
@@ -126,7 +126,8 @@ Feature: Prompt Meme Challenge
     And I should not see the prompt meme dashboard for "Battle 12"
     And no one should have a claim in "Battle 12"
     And no one should be signed up for "Battle 12"
-  When I go to the collections page
+  When all indexing jobs have been run
```
sarken, line 129: This can be removed now that `I go to the collections page` runs the indexing jobs.

features/prompt_memes_c/challenge_promptmeme_deletion.feature
```diff
@@ -187,7 +188,8 @@ Feature: Prompt Meme Challenge
   Then I should see "Collection was successfully deleted."
     And no one should have a claim in "Battle 12"
     And no one should be signed up for "Battle 12"
-  When I go to the collections page
+  When all indexing jobs have been run
```
sarken, line 191: This can be removed now that `I go to the collections page` runs the indexing jobs.

features/prompt_memes_a/challenge_promptmeme_setup.feature
```diff
@@ -160,7 +163,8 @@ Feature: Prompt Meme Challenge
   Given I have Battle 12 prompt meme fully set up
     And I am logged in as "myname1"
   When I sign up for Battle 12 with combination A
-  When I go to the collections page
+  When all indexing jobs have been run
```
sarken, line 166: This can be removed now that `I go to the collections page` runs the indexing jobs.

features/gift_exchanges/challenge_yuletide.feature
```diff
@@ -611,6 +624,7 @@ Feature: Collection
 
   # mod reveals authors on Jan 1st
   When I am logged in as "mod1"
+    And all indexing jobs have been run
```
sarken, line 627: This can be removed now that `I go to the collections page` runs the indexing jobs.

features/prompt_memes_a/challenge_promptmeme_setup.feature
```diff
@@ -26,21 +26,24 @@ Feature: Prompt Meme Challenge
   
   Given I have Battle 12 prompt meme fully set up
     And I am logged in as a random user
-  When I go to the collections page
+  When all indexing jobs have been run
```
sarken, line 29: This can be removed now that `I go to the collections page` runs the indexing jobs.

features/gift_exchanges/challenge_yuletide.feature
```diff
@@ -448,7 +459,8 @@ Feature: Collection
   When I follow "Gifts"
   Then I should not see "Stargate Atlantis"
     And I should not see "myname" within "ul.gift"
-  When I go to the collections page
+  When all indexing jobs have been run
```
sarken, line 462: This can be removed now that `I go to the collections page` runs the indexing jobs.

features/gift_exchanges/challenge_yuletide.feature
```diff
@@ -521,6 +533,7 @@ Feature: Collection
 
   # Mod checks for unfulfilled assignments, and gets pinch-hitters to do them.
   When I am logged in as "mod1"
+    And all indexing jobs have been run
```
sarken, line 536: This can be removed now that `I go to the collections page` runs the indexing jobs.

features/gift_exchanges/challenge_yuletide.feature
```diff
@@ -327,6 +337,7 @@ Feature: Collection
   # mod can view signups
   When I log out
     And I am logged in as "mod1"
+    And all indexing jobs have been run
```
sarken, line 340: This can be removed now that `I go to the collections page` runs the indexing jobs.

features/gift_exchanges/challenge_yuletide.feature
```diff
@@ -315,7 +324,8 @@ Feature: Collection
     And I fill in "challenge_signup_offers_attributes_1_tag_set_attributes_character_tagnames" with "Teyla Emmagan, John Sheppard"
     And I press "Submit"
   Then I should see "Sign-up was successfully created"
-  When I go to the collections page
+  When all indexing jobs have been run
```
sarken, line 327: This can be removed now that `I go to the collections page` runs the indexing jobs.

features/gift_exchanges/challenge_yuletide.feature
```diff
@@ -304,7 +312,8 @@ Feature: Collection
   #    offers: Tiny (Foo, Obscure), SGA (JS, TE)
   When I log out
     And I am logged in as "myname6"
-  When I go to the collections page
+  When all indexing jobs have been run
```
sarken, line 315: This can be removed now that `I go to the collections page` runs the indexing jobs.

features/gift_exchanges/challenge_yuletide.feature
```diff
@@ -288,7 +295,8 @@ Feature: Collection
   Then I should see "Sign-up was successfully created"
 
   # ordinary users can't see signups but can see summary
-  When I go to the collections page
+  When all indexing jobs have been run
```
sarken, line 298: This can be removed now that `I go to the collections page` runs the indexing jobs.

features/gift_exchanges/challenge_yuletide.feature
```diff
@@ -275,7 +281,8 @@ Feature: Collection
   #   offers Tiny (Foo, Obscure), SGA (JS, TE)
   When I log out
     And I am logged in as "myname5"
-  When I go to the collections page
+  When all indexing jobs have been run
```
sarken, line 284: This can be removed now that `I go to the collections page` runs the indexing jobs.

features/gift_exchanges/challenge_yuletide.feature
```diff
@@ -262,7 +267,8 @@ Feature: Collection
   Then I should see "Sign-up was successfully created"
 
   # ordinary users can't see signups until 5 people have signed up
-  When I go to the collections page
+  When all indexing jobs have been run
```
sarken, line 270: This can be removed now that `I go to the collections page` runs the indexing jobs.

features/collections/collection_create.feature
```diff
@@ -153,7 +154,8 @@ Given I have the collection "Scotts Collection" with name "scotts_collection"
     And I fill in "Collection name" with "child_collection"
     And I press "Submit"
   Then I should see "Collection was successfully created"
-  When I go to the collections page
+  When all indexing jobs have been run
```
sarken, line 157: This can be removed now that `I go to the collections page` runs the indexing jobs.

features/collections/collection_create.feature
```diff
@@ -130,7 +130,8 @@ Given I have the collection "Scotts Collection" with name "scotts_collection"
     And I fill in "Collection name" with "temporary_subcollection"
     And I press "Submit"
   Then I should see "Collection was successfully created"
-  When I follow "Collection Settings"
+  When all indexing jobs have been run
```
sarken, line 133: Do we need the indexing jobs to run here, since we're not looking at an index?

features/gift_exchanges/challenge_giftexchange.feature
```diff
@@ -121,6 +123,7 @@ Feature: Gift Exchange Challenge
     Given the gift exchange "Awesome Gift Exchange" is ready for signups
       And I am logged in as "myname1"
     When I sign up for "Awesome Gift Exchange" with combination A
+      And all indexing jobs have been run
```
sarken, line 126: This can be removed now that `I go to the collections page` runs the indexing jobs.

features/gift_exchanges/challenge_giftexchange.feature
```diff
@@ -543,7 +546,8 @@ Feature: Gift Exchange Challenge
       And I should not see the gift exchange dashboard for "Bad Gift Exchange"
       And no one should have an assignment for "Bad Gift Exchange"
       And no one should be signed up for "Bad Gift Exchange"
-    When I am on the collections page
+    When all indexing jobs have been run
```
sarken, line 549: This can be removed now that `I go to the collections page` runs the indexing jobs. (And if you can change the next line to "I go to the collections page" instead of "I am on," that would be great.)

features/gift_exchanges/challenge_yuletide.feature
```diff
@@ -32,7 +32,8 @@ Feature: Collection
     And I add the fandom "Starsky & Hutch" to the character "Foo The Wonder Goat"
     And a character exists with name: "Obscure person", canonical: true
     And I add the fandom "Tiny fandom" to the character "Obscure person"
-  When I go to the collections page
+  When all indexing jobs have been run
```
sarken, line 35: This can be removed now that `I go to the collections page` runs the indexing jobs.

features/gift_exchanges/challenge_yuletide.feature
```diff
@@ -87,7 +88,8 @@ Feature: Collection
   Then I should see "Challenge was successfully created"
   When I log out
     And I am logged in as "myname1"
-  When I go to the collections page
+  When all indexing jobs have been run
```
sarken, line 91: This can be removed now that `I go to the collections page` runs the indexing jobs.

features/gift_exchanges/challenge_yuletide.feature
```diff
@@ -194,7 +196,8 @@ Feature: Collection
   # another person signs up
   When I log out
     And I am logged in as "myname2"
-  When I go to the collections page
+  When all indexing jobs have been run
```
sarken, line 199: This can be removed now that `I go to the collections page` runs the indexing jobs.

features/gift_exchanges/challenge_yuletide.feature
```diff
@@ -224,7 +227,8 @@ Feature: Collection
   #           offers: SGA (JS, TE), S&H (JS, TE, Foo)
   When I log out
     And I am logged in as "myname3"
-  When I go to the collections page
+  When all indexing jobs have been run
```
sarken, line 230: This can be removed now that `I go to the collections page` runs the indexing jobs.

features/gift_exchanges/challenge_yuletide.feature
```diff
@@ -248,7 +252,8 @@ Feature: Collection
   #     offers Tiny (Obscure, JS), S&H (Foo, JS)
   When I log out
     And I am logged in as "myname4"
-  When I go to the collections page
+  When all indexing jobs have been run
```
sarken, line 255: This can be removed now that `I go to the collections page` runs the indexing jobs.

features/gift_exchanges/challenge_giftexchange.feature
```diff
@@ -35,7 +35,8 @@ Feature: Gift Exchange Challenge
       And I am on "My Gift Exchange" gift exchange edit page
     When I check "Sign-up open?"
       And I submit
-    When I view open challenges
+    When all indexing jobs have been run
+      And I view open challenges
```
sarken, line 39: If possible, it probably makes sense to give `I view open challenges` the same treatment we gave `I go to the collections page` -- we're pretty much always going to want to run indexing jobs before we go there so we're getting the latest info. If that works, then you should be able  to remove all of the `all indexing  jobs have been run` steps you had to add before `I view open challenges`.

app/models/search/collection_search_form.rb
```diff
@@ -0,0 +1,113 @@
...
+  def set_sorting
+    @options[:sort_column] ||= default_sort_column
+    @options[:sort_direction] ||= default_sort_direction
+  end
+
+  def persisted?
+    false
+  end
+
+  def summary
```
sarken, line 54: Hmm... is this actually used?

app/models/search/collection_query.rb
```diff
@@ -0,0 +1,98 @@
...
+
+  def owner_filter
+    options[:owner_ids].flatten.uniq.map { |owner_id| term_filter(:owner_ids, owner_id) } if options[:owner_ids].present?
+  end
+
+  def moderator_filter
+    options[:moderator_ids].flatten.uniq.map { |moderator_id| term_filter(:moderator_ids, moderator_id) } if options[:moderator_ids].present?
+  end
+
+  def moderator_or_owner_filter
```
sarken, line 51: I think calling this `maintainer_filter` might be more consistent with how the other filters are named.

sarken:
@AncyentMariner I've left a few minor comments, plus it looks like there are a couple of comments from @redsummernight that still need to be addressed:
* [this comment concerning a callback](https://github.com/otwcode/otwarchive/pull/3912/files#r515496548)
* [this comment on the filter view](https://github.com/otwcode/otwarchive/pull/3912/files#r515497326)

app/models/search/query_result.rb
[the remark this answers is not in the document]
sarken: I noticed something odd when I tried this branch out on my dev environment, and I suspect it might be related to the decorator, but I'm not sure. When I look at the blurbs on the collection page, I don't see the count of works, but rather the path to the collection's works page:

Collection blurbs. Each blurb displays a partial URL rather than a count for the Works stat.

This is the line of code:
https://github.com/littlelines/otwarchive/blob/4c6acff02cb452884b1549f574a5dea8875d9ca8/app/views/collections/_collection_blurb.html.erb#L51

app/decorators/collection_decorator.rb
```diff
@@ -0,0 +1,47 @@
...
+  def works_count
+    count = User.current_user.present? ? data[:general_works_count] : data[:public_works_count]
+    count || 0
+  end
+
+  def bookmarks_count
+    User.current_user.present? ? data[:general_bookmarked_items_count] : data[:public_bookmarked_items_count]
+  end
+
+  def fandoms_count
```
sarken, line 44: I don't think this is still used, now that we've taken the fandoms count out of the collection blurb.

app/decorators/collection_decorator.rb
```diff
@@ -0,0 +1,47 @@
...
+  # class, and then decorate them with the data.
+  def self.load_from_elasticsearch(hits)
+    items = Collection.load_from_elasticsearch(hits)
+    decorate_from_search(items, hits)
+  end
+
+   # TODO: Either eliminate this function or add definitions for work_counts and
+  # bookmark_counts (and possibly fandom information, as well?). The NameError
+  # that this causes isn't a problem at the moment because the function isn't
+  # being called from anywhere, but it needs to be fixed before it can be used.
```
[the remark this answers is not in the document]
AncyentMariner: Someone above asked me to add this back in when I has left it out before. And there's also some confusion about whether or not to include this at all. @sarken, thoughts on this?

app/models/search/collection_search_form.rb
AncyentMariner: Just tested to find out it's not. I'll remove it in the upcoming changes.

app/models/search/collection_search_form.rb
```diff
@@ -0,0 +1,113 @@
...
+  def sort_columns
+    options[:sort_column] || default_sort_column
+  end
+
+  def sort_direction
+    options[:sort_direction] || default_sort_direction
+  end
+
+  def sort_options
+    options[:faceted] || options[:collected] ? SORT_OPTIONS[1..-1] : SORT_OPTIONS
```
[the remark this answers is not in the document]
AncyentMariner: You're correct. Just using SORT_OPTIONS now.

features/users/user_delete.feature
AncyentMariner: @sarken I'm not sure if it's an issue with how I set it up or something else, but I get test failures when I remove that indexing line anywhere but the "collection_browse" tests. It looks like, at least in this file, the "I go to the collections page" step is triggering this more generic step: https://github.com/otwcode/otwarchive/blob/b5bf2e7a5cd362d1e350f2be087e8049721a2484/features/step_definitions/web_steps.rb#L15.
Any ideas on what to tweak to fix this?

app/models/search/collection_indexer.rb
```diff
@@ -0,0 +1,104 @@
...
+      public_bookmarked_items_count: get_bookmarked_items_count(object, true),
+    )
+  end
+
+  def get_bookmarked_items_count(collection, is_public = false)
+    bookmarks = Bookmark.is_public.joins(:collection_items)
+                .merge(CollectionItem.approved_by_collection)
+                .where(collection_items: { collection_id: collection.children.ids + [collection.id] })
+
+    bookmarks = bookmarks.select{ |b| b.bookmarkable.restricted == false } if is_public == true
```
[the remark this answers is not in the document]
AncyentMariner: I'm a little unclear with exactly what you're asking here. Could you explain how this should be changed a little further?

app/models/collection_participant.rb
```diff
@@ -57,4 +59,9 @@ def user_allowed_to_promote?(user, role)
     (role == MEMBER || role == NONE) ? self.collection.user_is_maintainer?(user) : self.collection.user_is_owner?(user)
   end
 
+  def reindex_collection
+    return unless MAINTAINER_ROLES.include?(participant_role)
```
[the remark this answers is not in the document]
AncyentMariner: Would you be able to explain this a little bit more? I'm not certain exactly what the change you're asking for looks like.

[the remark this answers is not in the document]
sarken: I think ticking means we need to reindex the collection if
* The role changes from member/invited/none to either owner or moderator. In this case, reindexing will add the user's id to the maintainer_ids array and and either the owner_ids or moderator_ids array.
* The role changes from moderator/owner to member, invited, or none. In this case, reindexing will remove the user's id from the maintainer_ids array and either the owner_ids or moderator_ids array.
* The role changes from moderator to owner or vice versa. In this case, reindexing will move the user's id out of moderator_ids and into owner_ids (or vice versa).

app/models/search/collection_indexer.rb
```diff
@@ -0,0 +1,104 @@
...
+      maintainer_ids: object.maintainers.pluck(:user_id),
+      challenge_type: object.challenge_type,
+      signup_open: object.challenge&.signup_open,
+      signups_open_at: object.challenge&.signups_open_at,
+      signups_close_at: object.challenge&.signups_close_at,
+      assignments_due_at: object.challenge&.assignments_due_at,
+      works_reveal_at: object.challenge&.works_reveal_at,
+      authors_reveal_at: object.challenge&.authors_reveal_at,
+      general_works_count: object.all_approved_works.count,
+      public_works_count: object.all_approved_works.where(restricted: false).count,
```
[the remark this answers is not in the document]
sarken: I saw you raised the question of `hidden_by_admin` for bookmark counts below, but does it apply here as well? I can't see anything preventing works hidden by an admin from being counted toward the total.

Possibly I'm entirely offbase here, but maybe we need to reindex a work's collections (and parent collections) anytime the work's `restricted`, `hidden_by_admin`, or `posted` changes?

app/models/search/collection_indexer.rb
```diff
@@ -0,0 +1,115 @@
...
+      works_reveal_at: object.challenge&.works_reveal_at,
+      authors_reveal_at: object.challenge&.authors_reveal_at,
+      general_fandom_ids: get_fandom_ids(object),
+      public_fandom_ids: get_fandom_ids(object, true),
+      general_fandoms_count: get_fandom_ids(object).count,
+      public_fandoms_count: get_fandom_ids(object, true).count,
+      general_works_count: object.all_approved_works.count,
+      public_works_count: object.all_approved_works.where(restricted: false).count,
+      general_bookmarked_items_count: get_bookmarked_items_count(object), 
+      public_bookmarked_items_count: get_bookmarked_items_count(object, true),
```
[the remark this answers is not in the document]
sarken: Hound is right; this comma should go.

app/models/search/collection_indexer.rb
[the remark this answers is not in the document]
sarken: I think ticking means that in order to match the count we get when we use [the `all_bookmarked_items_count` method](https://github.com/otwcode/otwarchive/blob/b5bf2e7a5cd362d1e350f2be087e8049721a2484/app/models/collection.rb#L313-L334), it's not enough to check if `b.bookmarkable.restricted == false`. 

You also have to make sure:
* the bookmarkable is not `hidden_by_admin`
* if the bookmarkable is a work, it's `posted`
 
(The `visible_to_registered_user` and `visible_to_all` scopes used in the linked `all_bookmarked_items_count` method handle checking `restricted`, `hidden_by_admin`, and `posted` all at once.)

app/models/search/collection_query.rb
```diff
@@ -0,0 +1,101 @@
...
+    input = (options[:query] || options[:title] || "").dup
+    query = escape_reserved_characters(input)
+
+    return {
+      query_string: {
+        query: query,
+        fields: ["title^5", "name"],
+        default_operator: "AND"
+      }
+    } unless query.blank?
```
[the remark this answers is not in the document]
sarken: This seems like a good suggestion from Hound.

app/models/search/indexer.rb
```diff
@@ -28,7 +29,8 @@ def self.all
       BookmarkIndexer,
       PseudIndexer,
       TagIndexer,
-      WorkIndexer
+      WorkIndexer,
+      CollectionIndexer
```
sarken, line 33: It looks like the existing list is aiming for alphabetical order. Can we scoot this one up between `BookmarkIndexer` and `PseudIndexer`?

features/users/user_delete.feature
sarken: I think the problem is in features/support/paths.rb:

https://github.com/littlelines/otwarchive/blob/b0f34ddd59d77fcbc02ada8906e693173bddb270/features/support/paths.rb#L24-L26

It should just be `when /^the collections page$/i`, not `when /^go to the collections page$/i`.

app/controllers/collections_controller.rb
```diff
@@ -11,7 +11,7 @@ class CollectionsController < ApplicationController
   # In both CollectionsController#create and CollectionsController#update there are a vulnerable usages of eval
   # For now just make sure the values passed to it are safe
   def validate_challenge_type
-    if params[:challenge_type] and not ["", "GiftExchange", "PromptMeme"].include?(params[:challenge_type])
+    if params[:challenge_type] and not ["", "GiftExchange", "PromptMeme", "no_challenge"].include?(params[:challenge_type])
```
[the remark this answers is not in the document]
sarken: I like the idea of the `form_for` approach, fwiw.

app/models/collection_participant.rb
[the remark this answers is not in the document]
tickinginstant: Yep. The code here covers the first and third cases (because in both of those cases, the participant role after the update is included in `MAINTAINER_ROLES`), but not the second (because in that case, the participant role is included in `MAINTAINER_ROLES` only before the update, not after, so it won't make it past this guard statement).

app/models/search/collection_indexer.rb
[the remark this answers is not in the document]
tickinginstant: The issue here is that `Collection#all_approved_works` and `Collection#all_approved_works_count` have pre-existing bugs. Neither one excludes hidden works. (And the latter is displayed in collection blurbs and sidebars, which isn't great.) But I wasn't sure if that bug was in scope for this PR or not.

Obviously, if it is in scope, it'd be better to exclude hidden works in both the collection class and the indexer. But if it's not in scope, I'm not sure if it'd be better to have all counts be consistent but incorrect, or to have the indexed count be correct but inconsistent.

app/models/search/collection_indexer.rb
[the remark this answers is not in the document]
tickinginstant: Yeah, that's what I meant.

Maybe it'd make sense to try to modify `Collection#all_bookmarked_items_count` to take `logged_in` as an optional argument with an appropriate default value? That might save some work replicating the logic, and also reduce code duplication.

app/models/search/collection_indexer.rb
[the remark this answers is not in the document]
sarken: I think it's fair to consider it in scope. 💸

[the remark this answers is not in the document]
AncyentMariner: Ok, just to be sure I'm on the same page, the request here is to change the `Collection#all_approved_works` and `Collection#all_approved_works_count` methods to exclude hidden works? (Which, if I'm reading the code correctly would mean adding the `works#hidden` scope to the queries of works in these methods?)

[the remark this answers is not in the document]
sarken: > Ok, just to be sure I'm on the same page, the request here is to change the `Collection#all_approved_works` and `Collection#all_approved_works_count` methods to exclude hidden works? 

Right. Also, to make sure that changing a work's visibility causes any collections (and parent collections) the work is in to be reindexed.

> (Which, if I'm reading the code correctly would mean adding the `works#hidden` scope to the queries of works in these methods?)

I think the `unhidden` scope is the one that excludes works hidden by admin, whereas `hidden` _includes_ works hidden by admin.

AncyentMariner:
@sarken I just pushed up some changes, and I THINK I've addressed all of the comments. Would you and the other folks keep me honest and double check that I've addressed all of them?
And if all of those are taken care of, is there anything else to do before this task can be accepted? Just trying to wrap this up before the holidays. Thanks in advance. :)

app/models/work.rb
```diff
@@ -233,15 +233,17 @@ def expire_caches
     Work.expire_work_tag_groups_id(self.id)
   end
 
-  def update_pseud_index
-    return unless should_reindex_pseuds?
+  def update_pseud_and_collection_index
+    return unless should_reindex_pseuds_and_collections?
+
     IndexQueue.enqueue_ids(Pseud, pseud_ids, :background)
+    IndexQueue.enqueue_ids(Collection, collection_ids, :background)
```
sarken, line 240: I think we also need to reindex any parent collection here as well, since the work count for parents includes the works for subcollections. 

(h/t ticking for [mentioning a similar issue on the CollectionItems model](https://github.com/otwcode/otwarchive/pull/3912#discussion_r528185995), which is now resolved.)

app/models/bookmark.rb
```diff
@@ -126,9 +126,10 @@ def invalidate_bookmark_count
   end
 
   # We index the bookmark count, so if it should change, update the pseud
-  def update_pseud_index
+  def update_pseud_and_collection_index
     return unless destroyed? || saved_change_to_id? || saved_change_to_private? || saved_change_to_hidden_by_admin?
     IndexQueue.enqueue_id(Pseud, pseud_id, :background)
+    IndexQueue.enqueue_ids(Collection, collection_ids, :background)
```
sarken, line 132: Do we also need to reindex any parent collections here? The bookmarked item count for parent collections includes the works for subcollections. It's true the bookmarked item count won't _always_ change in these scenarios -- e.g. if two people have bookmarked Work A, and only one of those bookmarks is hidden by an admin -- but I think it will sometimes (e.g. if only one person has bookmarked Work A and then sets the bookmark to private).

(h/t ticking for [mentioning a similar issue on the CollectionItems model](https://github.com/otwcode/otwarchive/pull/3912#discussion_r528185995), which is now resolved.)

app/models/search/collection_indexer.rb
```diff
@@ -0,0 +1,102 @@
...
+      signup_open: object.challenge&.signup_open,
+      signups_open_at: object.challenge&.signups_open_at,
+      signups_close_at: object.challenge&.signups_close_at,
+      assignments_due_at: object.challenge&.assignments_due_at,
+      works_reveal_at: object.challenge&.works_reveal_at,
+      authors_reveal_at: object.challenge&.authors_reveal_at,
+      general_works_count: object.all_approved_works.count,
+      public_works_count: object.all_approved_works.where(restricted: false).count,
+      general_bookmarked_items_count: get_bookmarked_items_count(object), 
+      public_bookmarked_items_count: get_bookmarked_items_count(object, true)
```
sarken, line 90: [ticking had a comment here](https://github.com/otwcode/otwarchive/pull/3912#discussion_r528200411) that got buried:

>Does this update properly when a work or series with a bookmark in the collection changes from restricted to unrestricted? What about when a bookmark changes visibility? (Either private or hidden_by_admin.) What about when the bookmark is in a subcollection, for either case?

I think changes to _bookmark_ visibility are covered in the bookmark model, although as I noted in a comment on `Bookmark#update_pseud_and_collection_index`, it might need a tweak to handle when the bookmark is in a subcollection.

However, I don't see where reindexing is being handled for _bookmarkables_ (works, series, external works) that change visibility. I'm pretty sure works, series, and external works can all have their `hidden_by_admin` value changed, and works or series can have their `restricted` value changed.

sarken (commented):
Okay, I've taken another look! I think I've found some things from ticking, red, and me that are still unaddressed, and of course there are some outstanding comments from Hound. 

I've marked anything that definitely doesn't need attention as resolved, so hopefully it will be easier to find  the stuff that does need action.

app/views/collections/_filters.html.erb
```diff
@@ -100,36 +74,26 @@
       <dd>
         <ul>
           <li>
-            <%= label_tag "collection_filters_challenge_type_gift_exchange" do %>
-              <%= radio_button_tag "collection_filters[challenge_type]",
-                  "gift_exchange",
-                  params[:collection_filters][:challenge_type] == "gift_exchange"
-              %>
+            <%= f.label "challenge_type_gift_exchange" do %>
+              <%= f.radio_button "challenge_type", "GiftExchange" %>
```
tickinginstant, line 78: Changing from `gift_exchange` to `GiftExchange` as the value appears to have broken this radio button. I think the label needs to change to `challenge_type_giftexchange`, and similarly with the label for the prompt meme radio button below.

app/models/bookmark.rb
sarken: Excellent, thanks!

app/models/search/collection_indexer.rb
```diff
@@ -0,0 +1,102 @@
...
+    )
+  end
+
+  def get_bookmarked_items_count(collection, is_public = false)
```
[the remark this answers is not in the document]
sarken: Oh, duh. I was so busy being excited by the scopes that I forgot to check the end result 🤦

app/models/collection_participant.rb
```diff
@@ -57,4 +59,14 @@ def user_allowed_to_promote?(user, role)
...
   end
 
+  def reindex_collection
+    is_assigned_as_maintainer = MAINTAINER_ROLES.include?(participant_role)
+    participant_change = previous_changes[:participant_role]
+    is_removed_as_maintainer = participant_change.present? && MAINTAINER_ROLES.include?(participant_change.first) && MAINTAINER_ROLES.exclude?(participant_change.second)
+    return unless is_assigned_as_maintainer || is_removed_as_maintainer
+
+    ids = [collection_id]
+    ids.push(collection.children.pluck(:id)) if collection.children.any?
```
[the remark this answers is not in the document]
AncyentMariner: Oops, yeah, that makes sense. Should I flatten it out?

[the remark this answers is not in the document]
tickinginstant: Maybe just set `ids = [collection_id] + collection.children.pluck(:id)`? That's similar to the way child IDs are handled in `Collection#all_items`:
https://github.com/otwcode/otwarchive/blob/b5bf2e7a5cd362d1e350f2be087e8049721a2484/app/models/collection.rb#L281-L283
Alternatively, maybe `+=` or `concat` instead of `push`? `flatten` would work as well, but it does involve an extra step.

app/models/search/collection_indexer.rb
```diff
@@ -0,0 +1,102 @@
...
+        }
+      }
+    }
+  end
+
+  def document(object)
+    object.as_json(
+      root: false,
+      only: [
+        :id, :name, :title, :description, :parent_id, :collection_type, :created_at
```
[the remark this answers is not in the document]
AncyentMariner: Facepalm. That should be 'challenge_type'

app/views/collections/_filters.html.erb
AncyentMariner: I'm a bit confused. It seems to be working for me. It's passing the value to the controller and filtering the results correctly. It's also highlighting the correct option on reload. Is there something I'm overlooking?

tickinginstant: In my browser, clicking on the label won't select the radio button. Even clicking on the radio button doesn't seem to work. I think it's because the real radio button is 1px, while the thing that looks like a radio button is actually a `<span>` styled to look like a radio button, and it can't control the radio button unless the label is correct.

(You need skins enabled to encounter this bug.)

AncyentMariner: Ah, gotcha. I didn't have skins enabled. Thanks for the explanation. Will be changed soon. :)

app/models/search/collection_indexer.rb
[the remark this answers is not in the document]
AncyentMariner: I'm not quite sure how to count the bookmarkables. Would you mind providing just a bit more context here?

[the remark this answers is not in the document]
tickinginstant: A work can have multiple bookmarks associated with it, and a collection can contain multiple of those bookmarks. So counting the number of bookmarks associated with a collection won't necessarily match the number of bookmarked works associated with the same collection. The same is true of series and external works.

To count the bookmarkables associated with a collection, you need to count the number of distinct works, external works, and series that have at least one bookmark in the collection (or in one of the collection's children). You also need to filter by work/external work/series/bookmark visibility, of course.

(`Collection#all_bookmarked_items_count` counts bookmarkables. What makes it unsuitable for calling here is that it calculates `logged_in` based on the value of `User.current_user`, instead of taking it as a parameter. But if `logged_in` were an optional parameter whose default value was set to `User.current_user.present?`, it'd work both for pre-existing calls, and for calculating `public_bookmarked_items_count` and `general_bookmarked_items_count`.)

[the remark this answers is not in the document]
AncyentMariner: That makes sense. Are you asking me to modify the method in the way you're suggesting? I'm always a bit hesitant to modify existing code that may be used elsewhere, to avoid side affects, but with the reasonable default for the existing functionality here, this seems like it shouldn't be an issue, right?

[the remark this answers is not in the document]
tickinginstant: I don't think it'll cause any issues with the existing code, and I'd prefer not to have two different functions doing almost the same thing, so changing it seems like the best option to me.

lib/bookmarkable.rb
```diff
@@ -25,4 +25,12 @@ def update_bookmarker_pseuds_index
     return unless should_reindex_pseuds?
     IndexQueue.enqueue_ids(Pseud, bookmarks.pluck(:pseud_id), :background)
   end
+
+  def update_bookmarker_collections_index
+    return unless respond_to?(:should_reindex_collections?)
+    return unless should_reindex_collections?
+
+    collection_ids = bookmarks.map(&:collection_ids).flatten
```
tickinginstant, line 33: The potential runtime of this line worries me a little. This line of code generates N+1 database queries, and there are a few works with 10,000+ bookmarks, so N is potentially very large here. It might be good to reduce this to a single database query.

AncyentMariner: I believe I replicated the logic with a single query. Does it look correct to you?

sarken (changes requested):
Just a few notes for now -- nothing resembling a through review, so you're welcome to either tackle them now or wait until there's more.

app/views/collections/_collection_blurb.html.erb
```diff
@@ -37,22 +37,27 @@
       (<%= collection.closed? ? ts("Closed") : ts("Open") %>, <%= collection.moderated? ? ts("Moderated") : ts("Unmoderated") %><%= collection.unrevealed? ? ts(", Unrevealed") : "" %><%= collection.anonymous? ? ts(", Anonymous") : "" %><%= collection.gift_exchange? ? ts(", Gift Exchange Challenge") : "" %><%= collection.prompt_meme? ? ts(", Prompt Meme Challenge") : "" %>)
     </p>
 
+    <% unless collection.tag_string.blank? %>
+      <h6 class="meta heading"><%= ts('Collection\'s Tags:') %></h6>
```
sarken, line 41: Can this say "Collection Tags" to match the mockup?

app/views/collections/_collection_blurb.html.erb
```diff
@@ -37,22 +37,27 @@
...
+      <ul class="meta tags commas">
+        <% collection.tags.each do |tag| %>
+          <li><%= link_to(tag.name, link_to_tag_bookmarks(tag), :class => 'tag') %></li>
```
sarken, line 44: This should point to the tag's work listings, not the bookmarks.

app/models/collection_participant.rb
```diff
@@ -57,4 +59,11 @@ def user_allowed_to_promote?(user, role)
     (role == MEMBER || role == NONE) ? self.collection.user_is_maintainer?(user) : self.collection.user_is_owner?(user)
   end
 
+  def reindex_collection
```
sarken, line 62: It looks like in some places, the method in charge of the reindexing is called `reindex_collection`, but in others, it's `update_collection_index` (or similar). I don't immediately see a reason for this distinction... is there one? If not, could we call it `update_collection_index` everywhere?

app/models/bookmark.rb
```diff
@@ -124,10 +124,11 @@ def invalidate_bookmark_count
     end
   end
 
-  # We index the bookmark count, so if it should change, update the pseud
-  def update_pseud_index
+  # We index the bookmark count for pseuds and collections, so update those if they should change
+  def update_pseud_and_collection_index
```
sarken, line 128: Question for AD&T folks: are we good with combining these two concepts here, or do we want to keep them separate like in the Work model?

sarken (commented):
(Still not a final review -- feel free to wait and tackle everything at once.)

app/models/bookmark.rb
redsummernight: I think we could combine the callbacks in Work, since the indexing triggers are exactly the same.

app/models/collection_participant.rb
redsummernight: I don't think there's a reason, we were just inconsistent (CommonTagging adds `update_search` into the mix). But using `update_*_index` for the new callbacks sounds good.

spec/models/search/collection_decorator_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+        "signup_open" => nil,
+        "signups_open_at" => nil,
+        "signups_close_at" => nil,
+        "assignments_due_at" => nil,
+        "works_reveal_at" => nil,
+        "authors_reveal_at" => nil,
+        "general_fandom_ids" => [],
+        "public_fandom_ids" => [],
+        "general_fandoms_count" => 5,
+        "public_fandoms_count" => 10,
```
sarken, line 31: Are we still indexing these? I think we might've stopped, at least with the counts.

app/models/concerns/fandom_crossover.rb
```diff
@@ -0,0 +1,41 @@
...
+    # Replace fandoms with their mergers if possible,
+    # as synonyms should have no meta tags themselves
+    all_without_syns = fandoms.map { |f| f.merger || f }.uniq
+
+    # For each fandom, find the set of all meta tags for that fandom (including
+    # the fandom itself).
+    meta_tag_groups = all_without_syns.map do |f|
+      # TODO: This is more complicated than it has to be. Once the
+      # meta_taggings table is fixed so that the inherited meta-tags are
+      # correctly calculated, this can be simplified.
```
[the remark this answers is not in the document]
tickinginstant: Yep, that should be fixed now.

app/controllers/collections_controller.rb
```diff
@@ -24,45 +24,40 @@ def load_collection_from_id
...
-      elsif params[:work_id]
-        flash.now[:error] = ts("We couldn't find that work.")
-      end
       @sort_and_filter = true
-      params[:collection_filters] ||= {}
-      params[:sort_column] = "collections.created_at" if !valid_sort_column(params[:sort_column], 'collection')
-      params[:sort_direction] = 'DESC' if !valid_sort_direction(params[:sort_direction])
-      sort = params[:sort_column] + " " + params[:sort_direction]
-      @collections = Collection.sorted_and_filtered(sort, params[:collection_filters], params[:page])
+      @search = CollectionSearchForm.new(collection_filter_params.merge(page: params[:page]))
```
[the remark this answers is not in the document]
tickinginstant: The nice thing about listing child collections is that it sort of resolves https://otwarchive.atlassian.net/browse/AO3-3748. It also handles challenge type mismatches between parent and child. (The "Gift Exchange Challenge" radio button on the collections page actually excludes some recurring gift exchanges like Yuletide, because only the child collections are marked as a gift exchange, and the child collections aren't listed in production.)

If the search only returns parent collections, I think the solution to AO3-3748 will have to involve adding a field with all child collection names to the index (and maybe all child challenge types, as well). Should that be part of this PR, to save having to reindex again later?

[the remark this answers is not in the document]
sarken: Yeah, sorry, I wasn't clear -- I just meant ao3.org/collections in its initial state. I don't particularly mind the change for results.

However, I can see how it would be confusing to have the page do something different when it was filtered than when it wasn't, _and_ it is clearer and more convenient to show the subcollection that matches the search/filter criteria than to show its parent. So maybe it's not a bad change, even if it makes the unfiltered list a bit more cluttered/overwhelming.

If we keep the new behavior, we just need to make sure to change the buttons on the Open Challenges pages that say "Top-Level Collections."

PR #3929, opened 2020-12-01
AO3-6085 AO3-6002 Upgrade to Rails 6

elzj:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6085
https://otwarchive.atlassian.net/browse/AO3-6002

## Purpose

Upgrades to Rails 6.0.3.4. Also upgrades rspec-rails and addresses deprecation warnings (update_attributes is now just update, and uniqueness validations need to specify whether or not they're case sensitive).

## Testing Instructions

Notes in issue.


config/application.rb
```diff
@@ -12,8 +12,11 @@ class Application < Rails::Application
     # Application configuration should go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded.
 
+    config.load_defaults "6.0"
```
tickinginstant, line 15: It looks like a bunch of the test failures are because this sets `cache_versioning` to true, which means that any calls to `cache_key` will no longer include `updated_at`. So caches like these will never be expired:
https://github.com/otwcode/otwarchive/blob/59fa3bb71e4376dd1f61a521e338681f7d6a0716/app/helpers/application_helper.rb#L118
https://github.com/otwcode/otwarchive/blob/59fa3bb71e4376dd1f61a521e338681f7d6a0716/app/views/chapters/_chapter.html.erb#L20
https://github.com/otwcode/otwarchive/blob/59fa3bb71e4376dd1f61a521e338681f7d6a0716/app/controllers/tags_controller.rb#L81
I think you can either disable `cache_versioning`, replace all calls to `cache_key` with `cache_key_with_version`, or start trying to pass a separate `version` parameter to all of the caches.

elzj: Yeah, I missed that one because it was a 5.2 change. I'd say it makes the most sense to set the config value since we relied on the old behavior in a lot of places.

redsummernight:
I added "Has Production Config Changes" because we will need to add `PERMITTED_HOSTS` to staging/production local.yml before the deploy can go through.

redsummernight (commented):
Should FAQ slugs be case-insensitive?

https://github.com/otwcode/otwarchive/blob/1f7610f25b28388bd6b2beb4715f2a80ece78ef7/app/models/archive_faq.rb#L10

app/models/tag.rb
```diff
@@ -1194,4 +1194,15 @@ def update_tag_nominations(tag, deleted=false)
...
+  def only_case_changed?
+    new_normalized_name = normalized_without_unicode(self.name)
+    old_normalized_name = normalized_without_unicode(self.name_was)
+    (self.name.downcase == self.name_was.downcase) ||
+      (new_normalized_name == old_normalized_name)
+  end
+
+  # Produce an ASCII, downcased version of a string
+  def normalized_without_unicode(str)
+    str.mb_chars.unicode_normalize(:nfkd).gsub(/[^\x00-\x7F]/u,'').downcase.to_s
```
redsummernight, line 1206: Because the normalization form was changed, can you add tests for the new behavior, as well as notes for what QA should check?

elzj: The behavior should be the same, and I believe the form is "normalize form kd" in both cases. It doesn't change any values, it just normalizes the strings for comparison, and we test the desired behavior in the tag model spec.

redsummernight: Got it, I missed that `normalize` was deprecated in favor of `String#unicode_normalize`.

tickinginstant:
Hmm, that bug's tricky.

It looks like Rails 6 modifies the way that databases are checked for being "up-to-date." In addition to checking the list of migration timestamps against the table stored in the database, it hashes the schema file whenever you call `db:schema:load` or `db:structure:load`, and stores the hash in the database so that it can check whether the file has changed later.

However, what appears to be happening in this PR is that the first hash (the one stored in the database when calling `rake db:schema:load`) is based on `schema.rb`, while the second one (the one computed to check whether the database is up-to-date) is based on `structure.sql`. This is because in Rails 6.0, calling `db:schema:load` (as opposed to `db:structure:load`) always uses the ruby schema, whereas the later computation is based on the `schema_format` config value set in `config/application.rb` (currently `:sql`).

I think that either the tests should be set to use `db:structure:load` to load the schema, or the config value should be set to `:ruby` (or deleted, because `:ruby` is the default). I lean towards the latter because we use `db:schema:load` in multiple places, and because `db:structure:load` is going to be deprecated in 6.1.

config/environments/production.rb
```diff
@@ -1,6 +1,10 @@
 Otwarchive::Application.configure do
   # Settings specified here will take precedence over those in config/environment.rb
 
+  local_config = YAML.load_file(Rails.root.join("config/local.yml"))
+
+  config.hosts = local_config["PERMITTED_HOSTS"]
```
sarken, line 6: Should this be `ArchiveConfig.PERMITTED_HOSTS` now? I wasn't sure, so I left it alone.

sarken:
Well, that certainly didn't work out like I'd hoped. /o\ 

I'm not sure if it's the same error Elz had been getting, though -- when I tried to check the test logs to see what error switching to `:ruby` for loading the schema was supposed to resolve, they were no longer available.

sarken:
Just to record it for posterity, lest the logs disappear, the error now is:

```
/home/runner/work/otwarchive/otwarchive/vendor/bundle/ruby/2.7.0/gems/nokogumbo-1.4.9/lib/nokogumbo.rb:24: [BUG] Segmentation fault at 0x0000000200000001
[18](https://github.com/otwcode/otwarchive/runs/5087178857?check_suite_focus=true#step:14:18)
ruby 2.7.3p183 (2021-04-05 revision 6847ee089d) [x86_64-linux]
```

tickinginstant:
The error caused by using the SQL schema format said something about migrations, IIRC. Not a segfault.

The most recent version of Nokogumbo is 2.0.5 -- maybe it needs to be bumped even further? Alternatively, Nokogumbo is apparently [at the end of its life and being merged into Nokogiri](https://github.com/rubys/nokogumbo#notice-end-of-life), so you could try updating Nokogiri instead.

app/models/download.rb
```diff
@@ -1,4 +1,5 @@
 class Download
+  require 'mimemagic'
```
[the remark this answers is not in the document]
sarken: So it looks like [Rails ditched MimeMagic for Marcel](https://www.theregister.com/2021/03/29/rails_mime_fix/), and I'm not really sure how to get Marcel to do what we do with MimeMagic in the download model. I took the lazy path and just added the gem back in.

tickinginstant (commented):
You got it to pass its tests! 🎊 

1. I think red had an unaddressed comment about whether the slug for the `ArchiveFaq` class should be case-sensitive.
2. The issue has a line saying that the PR "Update[s] the ActionMailer delivery job default to resolve deprecation warning." But there's still a deprecation warning for `ActionMailer::DeliveryJob` in the logs. There are other deprecation warnings that have to be tackled before the 6.1 update, so I don't much care either way, but I think we should either remove that line from the issue, or update `ApplicationMailerJob` to subclass `ActionMailer::MailDeliveryJob` instead.

sarken:
(I'm not done with the mailer or FAQ, but I wanted to see the tests and my diffs, 'cause I think I goofed and I'm not sure where.)

sarken:
>I added "Has Production Config Changes" because we will need to add PERMITTED_HOSTS to staging/production local.yml before the deploy can go through.

We have PERMITTED_HOSTS in config.yml already and that's what we're using now, so I think we're okay?

zz9pzza:
Just a \o/ for working on this, thank you

sarken:
As far as I can tell, the deprecation errors have been silenced without adding `config.action_mailer.delivery_job = "ActionMailer::MailDeliveryJob"` to `application.rb`. I'm not sure if there's a reason to add it there as well. 

(The test failure is unrelated -- `Validation failed: Url could not be reached.` for an external work -- but I'm going to rerun them because I'd like to see a green checkmark.)

tickinginstant:
Looks good to me! 👍 

>We have PERMITTED_HOSTS in config.yml already and that's what we're using now, so I think we're okay?

Yeah, it looks like the switch to using `ArchiveConfig.PERMITTED_HOSTS` means that the hosts don't have to be added to `local.yml` anymore.

> As far as I can tell, the deprecation errors have been silenced without adding config.action_mailer.delivery_job = "ActionMailer::MailDeliveryJob" to application.rb. I'm not sure if there's a reason to add it there as well.

I think we can count on Rails to have reasonable defaults for the mail delivery job, so I wouldn't worry about configuring it. The important part is to remove any explicit uses of `ActionMailer::DeliveryJob`, which you've done.

redsummernight:
> We have `PERMITTED_HOSTS` in config.yml already and that's what we're using now, so I think we're okay?

`ArchiveConfig.PERMITTED_HOSTS` will be overridden by local.yml in production and staging, which contains 2 extra entries:

```diff
--- a/config/config.yml
+++ b/config/config.yml
@@ -467,6 +467,8 @@ PERMITTED_HOSTS: [
   "download.archiveofourown.org",
   "insecure.archiveofourown.org",
   "secure.archiveofourown.org",
+  "www.archiveofourown.com",
+  "www.archiveofourown.net",
   "www.archiveofourown.org",
   # Staging
   "insecure-test.archiveofourown.org",
```
We can add them in config.yml to match, though that's optional.

sarken:
@redsummernight I'll do that, since I've got to bump Rails again due to a security alert. (We're on 6.0.4.4 in this branch, and it needs to be 6.0.4.6, I believe).

.github/workflows/automated-tests.yml
```diff
@@ -104,6 +104,7 @@ jobs:
         uses: ruby/setup-ruby@v1
         with:
           bundler-cache: true
+          cache-version: 1
```
redsummernight, line 107: This will [ignore contents of the cache and get and build all the gems anew](https://github.com/ruby/setup-ruby#dealing-with-a-corrupted-cache), and it was added while testing (https://github.com/elzj/otwarchive/commit/64e847e0a05f485a7e2ac6d3632325dbc23d9db9). Unless it's strictly necessary, let's remove it.

Gemfile
```diff
@@ -126,9 +123,10 @@ gem 'connection_pool'
 gem 'dalli'
 gem 'kgio', '2.10.0'
 
+gem "mimemagic", "0.3.10"
```
redsummernight, line 126: Previously we kept mimemagic, because it was required by paperclip even though it was no longer required by Active Storage (#3988). mimemagic is no longer a dependency of either, so let's eventually get rid of it. Can you add a TODO here?

sarken: I'm not entirely sure we can get rid of it because of how we're using it for downloads:  https://github.com/otwcode/otwarchive/pull/3929#discussion_r800303934

redsummernight: I think we can, Marcel has similar lookup methods we can use as replacements. We don't need to do that now though, hence a TODO.

redsummernight (approved):
Thank you!

sarken:
Switching back to Coder Actioned etc etc because I resolved the merge conflict in Gemfile.lock

redsummernight:
Still good.

app/models/skin.rb
```diff
@@ -111,7 +111,7 @@ def valid_public_preview
   end
 
   validates_presence_of :title
-  validates_uniqueness_of :title, message: ts('must be unique')
+  validates_uniqueness_of :title, message: ts('must be unique'), case_sensitive: false
```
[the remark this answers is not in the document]
sarken: ticking pointed out this will fix [AO3-5918](https://otwarchive.atlassian.net/browse/AO3-5918), but it'll leave us with skins with invalid titles. Do we want to change this to `case_sensitive: true` until we can deal with the existing skins, or do we have faith in our users' ability to change the names of their skins if they encounter a validation error?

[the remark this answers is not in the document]
redsummernight: Let's set it to true and keep 5918 out for now.

PR #3934, opened 2020-12-11
AO3-6086 Make collections taggable behind the scenes

AncyentMariner:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6086

## Purpose
Sets up code to prepare for the larger feature that migrates Collections search features to Elasticsearch

## Testing Instructions

(If you have a Jira account with access, please update or comment on the issue
with any new or missing testing instructions.)

## References

## Credit

(If you have a Jira account, please include the same name in the "Full name"
field on your Jira profile, so we can assign you the issues you're working on.)

(If you do not fill in this section, we will use your GitHub account name and
they/them pronouns to avoid making assumptions about your gender.)

### DIFF
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -145,10 +145,6 @@ class Collection < ApplicationRecord
     errors.add(:title, ts("^Sorry, the ',' character cannot be in a collection Display Title.")) if
       title.match(/\,/)
   end
-  validate :validate_tag_limit
-  def validate_tags
-    errors.add(:taggings, "You have reached the tag limit of %{max}") if taggings.count > 
-  end
 
   # return title.html_safe to overcome escaping done by sanitiser
   def title
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -1,5 +1,6 @@
 class Collection < ApplicationRecord
   include ActiveModel::ForbiddenAttributesProtection
+  include Filterable
   include UrlHelpers
   include WorksOwner
 
@@ -60,9 +61,6 @@ class Collection < ApplicationRecord
 
   has_many :bookmarks, through: :collection_items, source: :item, source_type: 'Bookmark'
   has_many :approved_bookmarks, -> { where('collection_items.user_approval_status = ? AND collection_items.collection_approval_status = ?', CollectionItem::APPROVED, CollectionItem::APPROVED) }, through: :collection_items, source: :item, source_type: 'Bookmark'
-
-  has_many :fandoms, -> { distinct }, through: :approved_works
-  has_many :filters, -> { distinct }, through: :approved_works
 
   has_many :collection_participants, dependent: :destroy
   accepts_nested_attributes_for :collection_participants, allow_destroy: true
@@ -146,6 +144,10 @@ class Collection < ApplicationRecord
   def no_reserved_strings
     errors.add(:title, ts("^Sorry, the ',' character cannot be in a collection Display Title.")) if
       title.match(/\,/)
+  end
+  validate :validate_tag_limit
+  def validate_tags
+    errors.add(:taggings, "You have reached the tag limit of %{max}") if taggings.count > 
   end
 
   # return title.html_safe to overcome escaping done by sanitiser
--- a/app/models/concerns/taggable.rb
+++ b/app/models/concerns/taggable.rb
@@ -77,6 +77,34 @@ module Taggable
     parse_tags(Freeform, tag_string)
   end
 
+  def tag_string
+    tags.map{|tag| tag.name}.join(ArchiveConfig.DELIMITER_FOR_OUTPUT)
+  end
+
+  def tag_string=(tag_string)
+    # Make sure that we trigger the callback for our taggings.
+    self.taggings.destroy_all
+
+    self.tags = []
+
+    # Replace unicode full-width commas
+    tag_string.gsub!(/\uff0c|\u3001/, ',')
+    tag_array = tag_string.split(ArchiveConfig.DELIMITER_FOR_INPUT)
+
+    tag_array.each do |string|
+      string.strip!
+      unless string.blank?
+        tag = Tag.find_by_name(string)
+        if tag
+          self.tags << tag
+        else
+          self.tags << UnsortedTag.create(name: string)
+        end
+      end
+    end
+    return self.tags
+  end
+
   alias category_strings= category_string=
   alias archive_warning_strings= archive_warning_string=
 
--- a/app/models/concerns/taggable.rb
+++ b/app/models/concerns/taggable.rb
@@ -49,6 +49,7 @@ module Taggable
   def freeform_string
     tag_category_string(:freeforms)
   end
+
   # Returns a string of tag names of all types
   def tag_string
     tags.map(&:name).join(ArchiveConfig.DELIMITER_FOR_OUTPUT)
@@ -80,6 +81,7 @@ module Taggable
   def freeform_string=(tag_string)
     parse_tags(Freeform, tag_string)
   end
+
   # Process a string of tags from any tag class. Unlike #parse_tags, it allows the creation of UnsortedTag instances.
   # Used in bookmarks and collections.
   def tag_string=(tag_string)
@@ -94,9 +96,9 @@ module Taggable
 
     tag_array.each do |string|
       string.strip!
-      next unless string.present?
+      next if string.blank?
       
-      if tag = Tag.find_by_name(string)
+      if (tag = Tag.find_by_name(string))
         self.tags << tag
       else
         self.tags << UnsortedTag.create(name: string)
--- a/app/models/concerns/taggable.rb
+++ b/app/models/concerns/taggable.rb
@@ -49,6 +49,10 @@ module Taggable
   def freeform_string
     tag_category_string(:freeforms)
   end
+  # Returns a string of tag names of all types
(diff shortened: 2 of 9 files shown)


tickinginstant:
I think we need to add collections to these two functions as well, to make sure that their filters are updated properly when tags are wrangled:

https://github.com/otwcode/otwarchive/blob/6cd05daeb4c3fb34a890e26a5e7f6e86b99ed68e/app/models/tag.rb#L695-L707

app/models/tagging.rb
```diff
@@ -3,6 +3,7 @@ class Tagging < ApplicationRecord
   belongs_to :taggable, polymorphic: true, touch: true, inverse_of: :taggings
 
   validates_presence_of :tagger, :taggable
+  validate :collection_tag_limit
```
redsummernight, line 6: We only need to enforce this `on: :create`.

app/models/tagging.rb
```diff
@@ -42,4 +43,10 @@ def update_search
     # in multiple fandoms.
     tagger.enqueue_to_index if tagger.taggings_count < ArchiveConfig.TAGGINGS_COUNT_REINDEX_LIMIT
   end
+
+  def collection_tag_limit
+    return unless taggable.class == Collection && taggable.tags.count >= ArchiveConfig.COLLECTION_TAG_MAX
+
+    errors.add(:tags, "You have reached the tag limit")
```
redsummernight, line 50: Can you make this a translatable string and include the limit? Similar to:

https://github.com/otwcode/otwarchive/blob/bedbbef8c43585d97b5a56062355921e4c8f5fa7/app/models/favorite_tag.rb#L14-L20

app/models/collection.rb
```diff
@@ -1,5 +1,6 @@
 class Collection < ApplicationRecord
   include ActiveModel::ForbiddenAttributesProtection
+  include Filterable
```
redsummernight, line 3: Can you add a test for Collection being filterable?

https://github.com/otwcode/otwarchive/blob/bedbbef8c43585d97b5a56062355921e4c8f5fa7/spec/models/concerns/filterable_spec.rb#L197-L205

app/models/concerns/taggable.rb
```diff
@@ -77,6 +77,34 @@ def freeform_string=(tag_string)
     parse_tags(Freeform, tag_string)
   end
 
+  def tag_string
```
redsummernight, line 80: Can you move this method next to the other `_string` getters? It should also have a comment:

```
# Returns a string of tag names of all types
```

app/models/concerns/taggable.rb
```diff
@@ -77,6 +77,34 @@ def freeform_string=(tag_string)
...
+    tags.map{|tag| tag.name}.join(ArchiveConfig.DELIMITER_FOR_OUTPUT)
+  end
+
+  def tag_string=(tag_string)
```
redsummernight, line 84: Can you add a comment for this method?

```
# Process a string of tags from any tag class. Unlike #parse_tags, it allows the creation of UnsortedTag instances.
# Used in bookmarks and collections.
```

(@sarken, from the changes we plan for tag wrangling pages I assume that collection owners can make unsorted tags, like bookmarkers and unlike work creators.)

app/models/concerns/taggable.rb
```diff
@@ -77,6 +77,34 @@ def freeform_string=(tag_string)
...
+    self.tags = []
+
+    # Replace unicode full-width commas
+    tag_string.gsub!(/\uff0c|\u3001/, ',')
+    tag_array = tag_string.split(ArchiveConfig.DELIMITER_FOR_INPUT)
+
+    tag_array.each do |string|
+      string.strip!
+      unless string.blank?
+        tag = Tag.find_by_name(string)
```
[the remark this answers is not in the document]
redsummernight: Except for the false positive `Rails/DynamicFindBy`, can you fix the other Rubocop warnings in this class?

redsummernight (changes requested):

redsummernight:
> I think we need to add collections to these two functions as well, to make sure that their filters are updated properly when tags are wrangled:
> 
>     https://github.com/otwcode/otwarchive/blob/6cd05daeb4c3fb34a890e26a5e7f6e86b99ed68e/app/models/tag.rb#L695-L707

@tickinginstant, should we also reindex collections here?

https://github.com/otwcode/otwarchive/blob/bedbbef8c43585d97b5a56062355921e4c8f5fa7/app/models/tag.rb#L664-L673

tickinginstant:
Collections aren't indexed at the moment, so we can't reindex them in this PR. But you're right, once both this and #3912 are in, we'll need to add collections and filtered collections to the `reindex_associated` function.

app/models/tag.rb
```diff
@@ -697,13 +698,15 @@ def filter
...
     external_works.update_filters
+    collections.update_filters
   end
 
   # Update filters for all works and external works that already have this tag
   # as one of their filters.
   def update_filters_for_filterables
     filtered_works.update_filters
     filtered_external_works.update_filters
+    collections.update_filters
```
tickinginstant, line 709: This needs to be `filtered_collections`, not `collections`.

app/models/concerns/taggable.rb
```diff
@@ -77,6 +82,32 @@ def freeform_string=(tag_string)
...
+
+    # Replace unicode full-width commas
+    tag_string.gsub!(/\uff0c|\u3001/, ",")
+    tag_array = tag_string.split(ArchiveConfig.DELIMITER_FOR_INPUT)
+
+    tag_array.each do |string|
+      string.strip!
+      next if string.blank?
+      
+      if (tag = Tag.find_by_name(string))
```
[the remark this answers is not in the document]
redsummernight: I think the [Style/ConditionalAssignment](https://docs.rubocop.org/rubocop/cops_style.html#styleconditionalassignment) warning means we should do something like:
```rb
self.tags << if (tag = Tag.find_by_name(string))
```

app/models/tagging.rb
```diff
@@ -42,4 +43,10 @@ def update_search
     # in multiple fandoms.
     tagger.enqueue_to_index if tagger.taggings_count < ArchiveConfig.TAGGINGS_COUNT_REINDEX_LIMIT
   end
+
+  def collection_tag_limit
+    return unless taggable.class == Collection && taggable.tags.count >= ArchiveConfig.COLLECTION_TAGS_MAX
+
+    errors.add(:tags, ts("Sorry, you can only save %{maximum} favorite tags.", maximum: ArchiveConfig.COLLECTION_TAGS_MAX))
```
redsummernight, line 50: This should say "collection tags".

app/models/concerns/taggable.rb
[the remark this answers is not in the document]
AncyentMariner: What would you do with the else statement, if the "if" is in line like that? I think that warning is for a different scenario, where you're assigning a variable to a conditional, to use later in the code. In this case, we need the variable inside the conditional, so assigning it from the entire statement isn't appropriate.

Since it's a pretty simple condition, it might be cleanest to convert it to a ternary, something like this: 
```
tag = Tag.find_by_name(string)
tag.present? ? self.tags << tag : self.tags << UnsortedTag.create(name: string)
```
Does that sound alright?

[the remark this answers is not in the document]
tickinginstant: If you run `rubocop -a` on the code that you have now, it becomes:
```
      self.tags << if (tag = Tag.find_by_name(string))
                     tag
                   else
                     UnsortedTag.create(name: string)
                   end
```
Which I think is what red is suggesting (and Hound is suggesting).

If instead you run `rubocop -a` on the original code (with `tag = Tag.find_by_name(string)` as a separate line, instead of as the condition of the `if`), it becomes:
```
      tag = Tag.find_by_name(string)
      self.tags << (tag || UnsortedTag.create(name: string))
```
Though of course, the absolute simplest way to rewrite it is:
```
self.tags << (Tag.find_by_name(string) || UnsortedTag.create(name: string))
```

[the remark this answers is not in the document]
AncyentMariner: Awesome. Thanks for the explanation. :)

app/models/concerns/taggable.rb
sarken: Right, any new tags created by the collection owners should be unsorted tags.

app/models/tagging.rb
```diff
@@ -42,4 +43,10 @@ def update_search
     # in multiple fandoms.
     tagger.enqueue_to_index if tagger.taggings_count < ArchiveConfig.TAGGINGS_COUNT_REINDEX_LIMIT
   end
+
+  def collection_tag_limit
+    return unless taggable.class == Collection && taggable.tags.count >= ArchiveConfig.COLLECTION_TAGS_MAX
+
+    errors.add(:tags, ts("Sorry, you can only save %{maximum} collections tags.", maximum: ArchiveConfig.COLLECTION_TAGS_MAX))
```
sarken, line 50: We had a quick discussion about this error message and decided it could maybe be worded a bit differently. Could you change it to: `Sorry, a collection can only have %{maximum} tags.`? Thank you!

AncyentMariner: Sure thing!

redsummernight (approved):
Looks good, thanks!

PR #3935, opened 2020-12-11
AO3-6066 Avoid errors when indexing batch is empty.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6066

## Purpose

Add a bunch of checks for blank/nil batches, so that we can avoid producing `BadRequest` errors whenever we try to index an empty batch.

### DIFF
--- a/app/models/search/indexer.rb
+++ b/app/models/search/indexer.rb
@@ -175,6 +175,8 @@ class Indexer
   end
 
   def batch
+    return @batch if @batch
+
     @batch = []
     ids.each do |id|
       object = objects[id.to_i]
@@ -189,7 +191,7 @@ class Indexer
   end
 
   def index_documents
-    return if (batch = self.batch).empty?
+    return if batch.empty?
 
     $elasticsearch.bulk(body: batch)
   end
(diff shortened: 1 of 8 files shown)


app/models/search/indexer.rb
```diff
@@ -189,6 +189,8 @@ def batch
   end
 
   def index_documents
+    return if (batch = self.batch).empty?
```
redsummernight, line 192: Is this to avoid calling the `batch` method twice?

tickinginstant: Yeah, I didn't want to generate each document twice.

redsummernight: Do we ever want `batch` to re-generate its documents? Or can we do something like this in `batch`:

https://github.com/otwcode/otwarchive/blob/b5bf2e7a5cd362d1e350f2be087e8049721a2484/app/models/search/taggable_query.rb#L5

tickinginstant: I don't think we ever need it to regenerate the batch. I'll make the change and simplify this line.

redsummernight (approved):

PR #3936, opened 2020-12-13
AO3-5958 Use Devise helpers for authentication in specs

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5958

## Purpose

- Use Devise helpers for `LoginMacros`.
- Create activated users by default or they can't be logged in with Devise.
- Use `controller.current_user` to access the logged in user in controller tests, only use `User.current_user` in model tests where the controller helper is not available.
- Remove duplicated request specs and stop including Capybara matchers in every spec type.

## Testing Instructions

None, automated.


factories/users.rb
```diff
@@ -16,22 +16,31 @@
...
-    password_confirmation { |u| u.password }
+    age_over_13 { "1" }
+    terms_of_service { "1" }
+    password_confirmation(&:password)
     email { generate(:email) }
 
-    factory :invited_user do
-      login { generate(:login) }
-      invitation_token { nil }
+    # By default, create activated users who can logged in, since we use
```
sarken, line 24: >who can logged in

Should be "log in," I think.

spec/controllers/comments_controller_spec.rb
```diff
@@ -226,6 +220,22 @@
                                      "access the page you were trying to " \
                                      "reach. Please log in.")
         end
+
+        context "when the commentable is a restricted work" do
+          let(:work) { comment.ultimate_parent }
+
+          before { work.update!(restricted: true) }
+
+          it "(on a comment) redirects to the login page" do
```
sarken, line 229: Can I be annoying for a sec? If I'm reading this correctly, I don't really love this new test being under `context "when the commentable is a restricted work"`. It seems like the commentable is actually a _comment_, while the restricted work is the ultimate parent.

So maybe it's more like: 
```
when the ultimate parent is a restricted work
  when the commentable is also the work
    it redirects to the login page
  when the commentable is a comment
    it redirects to the login page
```
Or:
```
when the commentable is a restricted work
  it redirects to the login page
when the commentable is a comment on a restricted work
  it redirects to the login page
```

sarken (changes requested):
Ooh, this is so much nicer! Just two little nitpicks.

sarken (approved):
Merging, since it's only tests! 🎉

PR #3938, opened 2020-12-13
AO3-4313 Display first posted chapter when Chapter 1 is a draft

elzj:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4313

## Purpose

Fixes issue with works that had a draft chapter set to Chapter 1.

## Testing Instructions

Described in issue: post a work, then add a draft chapter and set that to Chapter 1 of 2. When you access the work as a guest, you should see the existing posted chapter rather than an access denied message (or the draft).

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -780,7 +780,11 @@ class WorksController < ApplicationController
   end
 
   def load_first_chapter
-    @chapter = @work.first_chapter
+    if current_user_owns?(@work)
+      @chapter = @work.first_chapter
+    else
+      @chapter = @work.chapters.in_order.posted.first
+    end
   end
 
   # Check whether we should display :new or :edit instead of previewing or
--- a/spec/requests/works_show_request_spec.rb
+++ b/spec/requests/works_show_request_spec.rb
@@ -1,0 +1,20 @@
+require "spec_helper"
+
+describe "Works#show", :type => :request do
+  let(:work) { create(:work) }
+  let(:chapter) { work.chapters.first }
+
+  context "when the first chapter of a work is unposted" do
+    before do
+      work.chapters.create(position: 1, posted: false, content: "Draft content")
+      chapter.update(position: 2)
+    end
+
+    it "displays the first posted chapter" do
+      get "/works/#{work.id}"
+      expect(response).to render_template(:show)
+      expect(response.body).to include(chapter.content)
+      expect(response.body).not_to include("Draft content")
+    end
+  end
+end


sarken:
I think the test failures here might be real -- they keep happening every time I rerun the tests -- so I'm marking this Action Needed for now.

tickinginstant:
The issue appears to be that [admins should be able to view draft works](https://otwarchive.atlassian.net/browse/AO3-5751), while this change allows only the owner to view draft works.

Also, it'd be good to change the code to account for people who have been invited to be a co-creator on the work. (Since they need to be able to see what they're agreeing to be a co-creator of, even if it's a draft.)

sarken:
Hi, @elzj! Are you be available to update this, or should someone else take it over?

britnesissom:
@sarken if they don't respond, I can look at this one since my PR relies on it :)

sarken:
@britnesissom Terrific, thank you! I'll ping you if we don't hear from Elz in a few more days.

sarken:
@britnesissom The issue is all yours if you'd like!

britnesissom:
Maybe I'm testing this incorrectly but I can't reproduce this issue. I post the work, create a draft 2nd chapter, then go to edit chapter for the draft and change it to chapter 1 of 2 instead of 2 of 2. When I click save as draft from the edit chapter screen, it just posts both chapters (which seems wrong but I don't think that's what this issue is about). both titles say chapter 2 but the chapter content is in the original ch1, ch2 order, not swapped like it should be (which seems wrong too but also isn't what this issue is about). I can access both chapters as a guest.

@sarken are you able to reproduce what's actually in the testing steps?? If there's already a posted chapter, and I try to change a draft chapter's number to the posted chapter, should the two chapters just be swapped e.g. ch1 -> ch5's place, ch5 -> ch1's place? I don't know what the intended behavior should actually be in that scenario.

sarken:
@britnesissom I encountered the same behavior you described and was thoroughly confused... but it turns out the issue's instructions are a bit unclear.

I'm not sure if it was ever possible to reproduce the bug using the "Chapter X of Y" fields, but there are currently two ways (that I could find) to reproduce this bug. 

The first is by posting a work, adding a draft chapter, and then updating the draft chapter's position via the Rails console

```
draft_chapter = Chapter.find(123)
draft_chapter.position = 1
draft_chapter.save
```

But the actual, real-world way someone might cause this issue on AO3 is by following the steps in [AO3-5056](https://otwarchive.atlassian.net/browse/AO3-5065).

I hope that helps! I'll update AO3-4313 to make it a bit clearer.

sarken:
@britnesissom Actually, wait! I'm going to go a step further. I will close AO3-4313, which is no longer happening, and which this pull request does not address (I see Elz even refers to an access denied message, not a 500 error, in the description). Then I will update the info on the pull request to say it's for AO3-5065.

sarken:
@britnesissom ...wrong again! 5056 is _also_ no longer accurate. 

We're going to have to take a better look at the current behavior and figure out what actually needs fixing. I will let you know once we do! I'm sorry for all the confusion.

sarken:
@britnesissom Okay, we've done some poking around and I can now say with certainty that currently, the _only_ way to create a posted work with a draft first chapter is to modify the first chapter from the Rails console:

```
draft_chapter = Chapter.find(123)
draft_chapter.position = 1
draft_chapter.save
```

I've updated issue AO3-4313 to reflect this.

However, once #3978 is merged, it **will** once again be possible to create a work with a draft first chapter by following the steps in [AO3-5627](https://otwarchive.atlassian.net/browse/AO3-5627):

1. Post a work
2. Add a draft chapter
3. Delete the first posted chapter

So this does still need to be fixed at the same time as #3978, even though the only way to currently reproduce is from the console.

I hope that helps! If you need more clarification, let me know.

britnesissom:
@sarken thanks so much for looking into this! I'll see what I can do to fix it.

britnesissom:
@sarken I think something might be wrong with the failing tests themselves since the code changes don't have anything to do with the failures as far as I can tell. I've gotten them to pass sometimes and not others and I haven't changed any code. I just switched between the master and 4313 branches and it failed on both sporadically. Basically, there's no discernible pattern for when they fail or not.

I tested the scenarios for what this PR is actually fixing and everything seemed to work fine; it's just the automated tests.

tickinginstant:
@britnesissom Did you fix the permissions for admins & invited co-creators? If so, you might want to submit a new PR so that reviewers can see the updated code & the test failures. (And maybe add the failures to the list of [known intermittent failures](https://github.com/otwcode/otwarchive/wiki/Automated-Testing#Known_Test_Failures).)

britnesissom:
That was already working with the other person's code. I didn't change anything anywhere

sarken:
While I can see the draft chapters on a posted work if I'm logged in as admin/an invited co-creator, I can't see actual draft works. If I create a draft as UserA and then log in as an admin and go to `localhost:3000/users/UserA/works/drafts`, when I follow the link to the draft work, I get this error:

```
NoMethodError in Works#show
Showing /otwarchive/app/views/works/show.html.erb where line #47 raised:

undefined method `cache_key' for nil:NilClass

Extracted source (around line #47):
45 <% else %> 
46 <h3 class="landmark heading" id="work"><%= ts("Work Text:") %></h3>
47  <% cache("#{@chapter.cache_key}-show-content", skip_digest: true) do %>

Rails.root: /Users/sarken/Documents/OTW/otwarchive

Application Trace | Framework Trace | Full Trace
/otwarchive/app/views/works/show.html.erb:47:in `_app_views_works_show_html_erb__1612961005387163918_70340256523820'
/otwarchive/app/controllers/works_controller.rb:221:in `show'
```

Is that the error you're getting, @tickinginstant?

tickinginstant:
As far as I can tell, the PR hasn't been updated since I posted this comment:

> The issue appears to be that [admins should be able to view draft works](https://otwarchive.atlassian.net/browse/AO3-5751), while this change allows only the owner to view draft works.
> 
> Also, it'd be good to change the code to account for people who have been invited to be a co-creator on the work. (Since they need to be able to see what they're agreeing to be a co-creator of, even if it's a draft.)

So the code does need updating. I haven't tried running it, so I'm not actually getting any errors, but that one sounds related.

britnesissom:
@sarken that's interesting because I'm not getting that but I am getting an infinite redirect loop. I'm not sure why we'd be getting different results. I tried on firefox and edge and the result was the same.

britnesissom:
@sarken @tickinginstant is there a function for determining coauthor? this PR uses current_user_owns but that doesn't include current/invited co-creators. I can't find one but I could just be missing it.

britnesissom:
@sarken @tickinginstant I've created a new PR for this: #3985

sarken:
Replaced by #3985

PR #3943, opened 2020-12-22
AO3-3765 Fix links in tag comment notifications.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-3765

## Purpose

Changes the links in tag comment notifications so that they point to the individual comment, and not the page for all comments on that tag.


sarken (commented):
Looks good! My only question: should we also be testing the `Go to the thread to which this comment belongs` link somewhere? It looks like there's only tests for `the thread starting from this comment`.

tickinginstant:
I added some more tests for the comment's thread link. It only shows up on reply comments (comments where `thread != id`), so I added a few more tests involving replies, as well.

sarken (approved):
LGTM! I'll merge when the tests finish rerunning.

PR #3946, opened 2020-12-25
AO3-6096 Replace Dewplayer embeds with audio tags when sanitizing HTML

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6096

## Purpose

- Update the embed sanitizer to replace Dewplayer embeds with HTML5 audio tags.
- Add a rake task to re-sanitize existing works using Dewplayer embeds.

## Testing Instructions

- Post a work using the old embed code, then after the staging deploy, run the rake and make sure the work is updated.
- Post a work using the old embed code after the staging deploy. The work should automatically be updated as well.

### DIFF
--- a/lib/otw_sanitize/embed_sanitizer.rb
+++ b/lib/otw_sanitize/embed_sanitizer.rb
@@ -145,8 +145,10 @@ module OTWSanitize
       mp3_urls = CGI.parse(flashvars)["mp3"]
       return if mp3_urls.blank?
 
-      audio_fragment = mp3_urls.map { |url| "<audio src='#{url}'></audio>" }.join
-      node.replace(audio_fragment)
+      # Dewplayer allows specifying multiple sources.
+      mp3_urls = mp3_urls.map { |url| url.split("|") }.flatten
+      audio_fragment = mp3_urls.map { |url| "<audio src='#{url}'></audio>" }.join("<br>")
+      node.replace("<p>#{audio_fragment}</p>")
     end
 
     # We're now certain that this is an embed from a trusted source, but we
--- a/lib/otw_sanitize/embed_sanitizer.rb
+++ b/lib/otw_sanitize/embed_sanitizer.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
+
 require "addressable/uri"
+require "cgi"
 
 module OTWSanitize
   # Creates a Sanitize transformer to sanitize embedded media
@@ -35,6 +37,9 @@ module OTWSanitize
     # Creates a callable transformer for the sanitizer to use
     def self.transformer
       lambda do |env|
+        # Don't continue if this node is already safelisted.
+        return if env[:is_whitelisted]
+
         new(env[:node]).sanitized_node
       end
     end
@@ -51,6 +56,11 @@ module OTWSanitize
       return unless source_url && source
 
       ensure_https
+
+      # If a Dewplayer embed has been replaced with <audio> tags, return
+      # without safelisting them so the Sanitize transformer for <audio>
+      # tags can modify them later.
+      return if replace_dewplayer
 
       if parent_name == 'object'
         sanitize_object
@@ -122,6 +132,23 @@ module OTWSanitize
       end
     end
 
+    # If the embed is hosted on the Archive, it's Dewplayer and can be replaced
+    # with <audio> tag(s).
+    #
+    # Refer to https://archiveofourown.org/admin_posts/250.
+    def replace_dewplayer
+      return unless source == :ao3 && node_name == "embed" && source_url.downcase.include?("dewplayer")
+
+      flashvars = node["flashvars"]
+      return if flashvars.blank?
+
+      mp3_urls = CGI.parse(flashvars)["mp3"]
+      return if mp3_urls.blank?
+
+      audio_fragment = mp3_urls.map { |url| "<audio src='#{url}'></audio>" }.join
+      node.replace(audio_fragment)
+    end
+
     # We're now certain that this is an embed from a trusted source, but we
     # still need to run it through a special Sanitize step to ensure
     # that no unwanted elements or attributes that don't belong in
@@ -161,7 +188,7 @@ module OTWSanitize
         disable_scripts(node)
         node['flashvars'] = "" unless allows_flashvars?
       end
-      { node_whitelist: [node, parent] }
+      { node_whitelist: [node] }
     end
 
     # disable script access and networking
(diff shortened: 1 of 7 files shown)


lib/otw_sanitize/embed_sanitizer.rb
```diff
@@ -122,6 +132,23 @@ def ensure_https
       end
     end
 
+    # If the embed is hosted on the Archive, it's Dewplayer and can be replaced
+    # with <audio> tag(s).
+    #
+    # Refer to https://archiveofourown.org/admin_posts/250.
+    def replace_dewplayer
```
redsummernight, line 139: This will convert embed examples from [the old news post](https://archiveofourown.org/admin_posts/250). Open questions:

- It's possible to embed .swf files using `<object>` tags, but that seems to be broken on beta at the moment (???). Cursory debugging shows that this line doesn't work:
  https://github.com/otwcode/otwarchive/blob/0622729c9d4fca1ec7b611235bb5e318256ebd70/lib/otw_sanitize/embed_sanitizer.rb#L77
  because Sanitize currently doesn't return the `<object>` parent of the `<param>` tag, but instead a non-element.
- It's possible to embed the multi-variant of the player (https://archiveofourown.org/system/dewplayer/dewplayer-multi.swf) and specify multiple audio URLs separated by pipe characters (`|`). Should we try to handle this case? We didn't document this in the old news post.

redsummernight: To answer my second question, yeah there are fics using the multi version, so let's handle that.

sarken: Soooo, wait. For the .swf files embedded with `<object>`, were we theoretically allowing them to be embedded from anywhere on the web, or just certain safelisted hosts? If it was anywhere, yikes. If it was safelisted hosts, that makes sense -- a lot of places used `object` at one time, but judging from [the examples in this issue)(https://otwarchive.atlassian.net/browse/AO3-5192), only 8tracks still used it, which would explain why we had no idea it was broken (8tracks was mostly dead for a while and still isn't in great shape).

redsummernight: Sorta... we allow embeds only from safelisted hosts, and for safelisted hosts, embeds can be any of `<object>`, `<embed>`, or `<iframe>`. There's no restriction that each safelisted host must go with a specific embed type. That we use `<embed>` for Archive-hosted players, or `<object>` for 8tracks, is just because we follow documented examples, but we can certainly [rewrite any `<embed>` as `<object>` or vice versa](https://web.archive.org/web/20201113034702/https://perishablepress.com/embed-flash-or-die-trying/).

sarken: Ahh, good. As long as it's limited to safelisted hosts, that doesn't sound too terrible, even if it wasn't entirely intentional. 

Anyway, what exactly was the question about this part?

redsummernight: Are we fine with ignoring potential `<object>` Dewplayer embeds, since no new `<object>` embeds can be created at the moment anyway?

sarken: Yeah, I'd say if you varied that greatly from the code we gave you, it's up to you to fix it manually.

redsummernight: Thanks, no more questions from me on this thread.

lib/otw_sanitize/embed_sanitizer.rb
```diff
@@ -35,6 +37,9 @@ class EmbedSanitizer
     # Creates a callable transformer for the sanitizer to use
     def self.transformer
       lambda do |env|
+        # Don't continue if this node is already safelisted.
+        return if env[:is_whitelisted]
```
sarken, line 41: Could you explain how something would already be safelisted? (For me, not necessarily in the comment.)

lib/otw_sanitize/embed_sanitizer.rb
```diff
@@ -161,7 +190,7 @@ class type width
         disable_scripts(node)
         node['flashvars'] = "" unless allows_flashvars?
       end
-      { node_whitelist: [node, parent] }
+      { node_whitelist: [node] }
```
sarken, line 193: Could you explain this change? Or rather, why it was necessary as part of this fix? (It's fine if it wasn't; I'm just utterly lost.)

lib/otw_sanitize/embed_sanitizer.rb
redsummernight: Transformers can safelist particular nodes, so they don't get cleaned up by the overall Sanitize config (e.g. we don't allow `<embed>` in `Sanitize::Config::ARCHIVE`, so `EmbedSanitizer` has to safelist such nodes).

If a node has been safelisted by a previous transformer, subsequent transformers ([`MediaSanitizer` in our case](https://github.com/otwcode/otwarchive/blob/0622729c9d4fca1ec7b611235bb5e318256ebd70/lib/html_cleaner.rb#L157-L158)) can know about it by checking [`:is_whitelisted`](https://github.com/rgrove/sanitize/tree/v4.6.6#transformers).

It's not required to honor `:is_whitelisted` (and indeed we were fine not doing so until now), but since `EmbedSanitizer` can now generate nodes that should be sanitized by `MediaSanitizer` (`<audio>`), I'd like to make it clear what is/isn't safelisted and follow that in all of our sanitizers.

lib/otw_sanitize/embed_sanitizer.rb
redsummernight: https://github.com/otwcode/otwarchive/blob/0622729c9d4fca1ec7b611235bb5e318256ebd70/lib/otw_sanitize/embed_sanitizer.rb#L74-L79

This transformer looks for either:

- `<param>` tags with `<object>` parents, in which case we sanitize and safelist both the current node and the parent node.
- `<embed>`/`<iframe>` tags, in which case we sanitize just the current node and don't look at the parent. So we shouldn't safelist the parent in these cases.

lib/otw_sanitize/embed_sanitizer.rb
sarken: Thank you! I couldn't find that `is_whitelisted` documentation when I was Googling, and I, uh. Just realized that's because I had the name of the gem wrong. (I was looking for "sanitizer" with an "r".) 🤦

sarken (approved):

lib/otw_sanitize/embed_sanitizer.rb
redsummernight: I also had to link to the README of the 4.x version we're using, because the option was renamed to `:is_allowlisted` in Sanitize 5.x.

PR #3949, opened 2020-12-28
AO3-6096 Handle encoded characters in Dewplayer audio URLs

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6096

## Purpose

URL-encoded spaces in audio URLs of Dewplayer embeds get decoded when we [parse flashvars](https://helpx.adobe.com/flash/kb/pass-variables-swfs-flashvars.html), then the URLs with unencoded spaces get rejected as invalid by `URI` when we try to check their hosts in `MediaSanitizer`. We'll instead use `Addressable::URI` to parse URLs, which is more forgiving.

Also print out errored chapter IDs at the end of the rake task.

## Testing Instructions

See issue.

### DIFF
--- a/spec/lib/html_cleaner_spec.rb
+++ b/spec/lib/html_cleaner_spec.rb
@@ -179,8 +179,18 @@ describe HtmlCleaner do
         end
 
         it "converts an Archive-hosted Dewplayer embed into an audio tag" do
-          html = '<embed type="application/x-shockwave-flash" flashvars="mp3=http://example.com/next%20color%20planet%27.mp3?dl=0" src="https://archiveofourown.org/system/dewplayer/dewplayer.swf" width="200" height="27" allowscriptaccess="never" allownetworking="internal"></embed>'
-          expect(sanitize_value(field, html)).to include('<audio src="https://example.com/next%20color%20planet%27.mp3?dl=0" controls="controls" crossorigin="anonymous" preload="metadata"></audio>')
+          html = '<embed type="application/x-shockwave-flash" flashvars="mp3=http://example.com/next%20color%20planet.mp3?dl=0" src="https://archiveofourown.org/system/dewplayer/dewplayer.swf" width="200" height="27" allowscriptaccess="never" allownetworking="internal"></embed>'
+          expect(sanitize_value(field, html)).to include('<audio src="https://example.com/next%20color%20planet.mp3?dl=0" controls="controls" crossorigin="anonymous" preload="metadata"></audio>')
+        end
+
+        it "converts an Archive-hosted Dewplayer embed with single quotes in the source URL into an audio tag" do
+          html = '<embed type="application/x-shockwave-flash" flashvars="mp3=http://example.com/\'quote%27.mp3" src="https://archiveofourown.org/system/dewplayer/dewplayer.swf" width="200" height="27" allowscriptaccess="never" allownetworking="internal"></embed>'
+          expect(sanitize_value(field, html)).to include('<audio src="https://example.com/\'quote%27.mp3" controls="controls" crossorigin="anonymous" preload="metadata"></audio>')
+        end
+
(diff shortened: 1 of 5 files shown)


tickinginstant:
If the process of parsing the `flashvars` param decodes `%20`, wouldn't it also decode `%27` (encoded single quote)? Couldn't that cause problems with the replacement `<audio>` tag?
https://github.com/otwcode/otwarchive/blob/27d089d5189069609eeea552e6a7d2ba52c7917e/lib/otw_sanitize/embed_sanitizer.rb#L150

redsummernight:
Yeah it'd be a problem, thanks for catching that! I think I'll parse flashvars manually and avoid modifying the source URLs in any way, instead of using `CGI.parse` and having to re-escape parts of the source URLs.

redsummernight:
> EDIT: Come to think of it, is there anything preventing flashvars from containing an unencoded single quote?

Users shouldn't be able to write that in the first place, as the HTML cleaner will cut off the attribute after the first unencoded single quote.

tickinginstant:
> Users shouldn't be able to write that in the first place, as the HTML cleaner will cut off the attribute after the first unencoded single quote.

Are you sure? I get a white screen when I try to do that, but when I reload the changes have gone through and there is an unescaped single quote in `flashvars`. (Testing locally on a previous version, because the code that's on staging doesn't demonstrate what sorts of embeds could have been created before the dewplayer changes went live.)

I don't think it's a security risk, but it may lose part of the URL the user wanted.

redsummernight:
> Are you sure? I get a white screen when I try to do that, but when I reload the changes have gone through and there is an unescaped single quote in `flashvars`.

You're right, single quotes can go through when the attribute is surrounded by double quotes. I saw the cut-off behavior when the attribute is also wrapped by single quotes.

spec/lib/html_cleaner_spec.rb
```diff
@@ -179,15 +179,29 @@
...
+          expect(sanitize_value(field, html)).to include('<audio src="https://example.com/\'quote%27.mp3" controls="controls" crossorigin="anonymous" preload="metadata"></audio>')
+        end
+
+        it "converts an Archive-hosted Dewplayer embed with double quotes in the source URL into an audio tag" do
+          html = '<embed type="application/x-shockwave-flash" flashvars=\'mp3=http://example.com/"quote%22.mp3\' src="https://archiveofourown.org/system/dewplayer/dewplayer.swf" width="200" height="27" allowscriptaccess="never" allownetworking="internal"></embed>'
+          expect(sanitize_value(field, html)).to include('<audio src="https://example.com/%22quote%22.mp3" controls="controls" crossorigin="anonymous" preload="metadata"></audio>')
+        end
+
+        it "converts an Archive-hosted Dewplayer embed with an encoded source URL into an audio tag" do
+          html = '<embed type="application/x-shockwave-flash" flashvars="param=val&amp;mp3=http%3A%2F%2Fexample.com%2Fnext%2520color%2520planet%2527.mp3%3Fdl%3D0" src="https://archiveofourown.org/system/dewplayer/dewplayer.swf" width="200" height="27" allowscriptaccess="never" allownetworking="internal"></embed>'
```
sarken, line 197: Do you know when we'd have a case like this? Or are we just covering our bases in the unlikely event it happens?

redsummernight: It's a valid way to set flashvars (like [the vidders.net example from AO3-5192](https://otwarchive.atlassian.net/browse/AO3-5192)). Probably not common since we didn't document it in the news post, but we're covering our bases.

It wouldn't be the first time we do this for flashvars. There's an existing example in the same sanitizer:
https://github.com/otwcode/otwarchive/blob/27d089d5189069609eeea552e6a7d2ba52c7917e/lib/otw_sanitize/embed_sanitizer.rb#L129-L132

sarken (approved):

PR #3952, opened 2020-12-31

otwtranslation:


sarken:
Let's summarize the text failure!

> **sarken**  21:35
> Ohh, this is fun. We have a test failure due to the text in the file Phrase submitted not matching the intended text -- ie the English text appears to have gotten changed?
> 
> **sarken**  21:39
> I think what happened is we updated the English text, but didn't change the key, so what Phrase had never updated
> 21:39
> Which means the pull request is now trying to return the test text to its old version (edited) 
> 21:39
> It's the change in https://otwarchive.atlassian.net/browse/AO3-5723
> 
> **sarken**  21:40
> That is bad
> 
> 21:40
> Extremely bad
> 
> **sarken**  21:41
> Because we don't know how many places it's doing that. We don't test every line of text... so we're just lucky it caught this instance of it

sarken (changes requested):
Two unwanted changes; there are probably more.

sarken:
What I think needs to happen is we need to merge #3960 first. If we do that, we won't be looking at endless reordering and reformatting of the existing translations in addition to all the actual new/updated translations -- we'll just see the new/updated translations. That will allow us to more easily identify where things have gone wrong.

sarken:
We'll make a new version once https://otwarchive.atlassian.net/browse/AO3-6109 is in place

PR #3957, opened 2021-01-09
AO3-5585 Modify ChallengeAssignmentsController.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5585

## Purpose

- Remove unused and test-only routes.

- Move the `ChallengeAssignmentsController#default` route so that it's nested under the collection, so that we can handle the `show`/`default` actions more consistently.

- Load the challenge assignment for `show`/`default` in a way that prevents accessing other collections' assignments.

- Rewrite the permissions for `ChallengeAssignmentsController#default`, so that we actually check that the assignment belongs to the logged-in user, and not the user specified by `user_id`.

- Remove `collection_id` (and `id`) from the list of permitted parameters for `ChallengeAssignmentsController#set`.

- Modify both `ChallengeAssignmentsController#update_multiple` and `ChallengeAssignmentsController#set` so that they don't allow assignments from other collections to be edited.

- Add many more RSpec tests for `ChallengeAssignmentsController`.

## Testing Instructions

See Jira.


app/controllers/challenge_assignments_controller.rb
```diff
@@ -170,7 +139,16 @@ def send_out
   def set
     # update all the assignments
     # see http://asciicasts.com/episodes/198-edit-multiple-individually
```
redsummernight, line 141: Can you remove this link?

app/controllers/challenge_assignments_controller.rb
```diff
@@ -170,7 +139,16 @@ def send_out
...
-    @assignments = ChallengeAssignment.update(challenge_assignment_params[:challenge_assignments].keys, challenge_assignment_params[:challenge_assignments].values).reject {|a| a.errors.empty?}
+
+    all_assignment_params = challenge_assignment_params.transform_keys(&:to_i)
+
+    @assignments = []
+
+    @collection.assignments.where(id: all_assignment_params.keys).each do |assignment|
+      assignment_params = all_assignment_params[assignment.id]
+      @assignments << assignment unless assignment.update(assignment_params)
```
redsummernight, line 149: Do you use this instead of Persistence::ClassMethods#update to tell which assignments are actually updated?

tickinginstant: It's to avoid 404 errors when someone malicious enters an invalid assignment ID. ([That version of `update` uses `find` internally](https://github.com/rails/rails/blob/5f3ff60084ab5d5921ca3499814e4697f8350ee7/activerecord/lib/active_record/persistence.rb#L289), which would throw a `RecordNotFound` error when trying to edit someone else's assignment.) I wasn't sure if failing silently was the desired behavior, but IIRC that was the desired behavior for a similar issue with `CollectionItem`s.

redsummernight (approved):

PR #3958, opened 2021-01-13
AO3-6105 Prevent mismatches in OwnedTagSets for TagSetNominations.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6105

## Purpose

The `TagSetNominations` controller allows you to view pages like `/tag_sets/:owned_tag_set_id/nominations/:id` where the tag set nomination's `owned_tag_set_id` doesn't match the `owned_tag_set_id` specified in the URL. This modifies the `before_action` hooks to make sure that the only tag set nominations accessible for a given `OwnedTagSet` are nominations for that tag set.

### DIFF
--- a/app/controllers/tag_set_nominations_controller.rb
+++ b/app/controllers/tag_set_nominations_controller.rb
@@ -26,7 +26,7 @@ class TagSetNominationsController < ApplicationController
   end
 
   def load_nomination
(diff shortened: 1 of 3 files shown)


sarken (approved):
Thank you!

PR #3960, opened 2021-01-15
Reorganized translations from Phrase

sarken:
These aren't actually updated translations -- rather, they're simply Phrase's reformatted versions of the existing files. This will help us find the English text that got wrongly reverted to previous versions in #3952, since it's very hard to find those changes in amongst all the reordering/reformatting changes.


config/locales/mailers/en.yml
```diff
@@ -1,24 +1,27 @@
...
+            support_link_text: contact Support
           text:
-            part1: "If you've received this message in error, please contact Support at %{support_url}."
-            part2: "The Archive of Our Own is a fan-run and fan-supported archive that relies on your donations: %{donate_url}."
-        sent_at: "Sent at %{sent_at}."
+            part1: If you've received this message in error, please contact Support
+              at %{support_url}.
+            part2: 'The Archive of Our Own is a fan-run and fan-supported archive
+              that relies on your donations: %{donate_url}.'
+        sent_at: Sent at %{sent_at}.
```
sarken, line 18: In #3952, there's some sent_by keys that get added here.

config/locales/mailers/en.yml
```diff
@@ -27,224 +30,331 @@ en:
         left_kudos:
           one: "%{givers_list} left kudos on %{commentable_link}."
           other: "%{givers_list} left kudos on %{commentable_link}."
+      you_have: "[%{app_name}] You've got kudos!"
```
sarken, line 33: The other PR has some HTML tags that get added in the html left_kudos keys. https://github.com/otwcode/otwarchive/pull/3952/files#r550414843

redsummernight (commented):
You should update the [README](https://github.com/otwcode/otwarchive/blob/e3635a9f648924df8adf82058ae76804076b8ee9/config/locales/README.md), since grabbing the copies from rails-i18n is no longer a straightforward process now that we've let Phrase reorder everything.

sarken:
Oh, lord. 

Sent from my iPhone

sarken:
We'll probably want the non-default en files when we do https://otwarchive.atlassian.net/browse/AO3-6109, but the rest are unnecessary, so I'll close this.

PR #3961, opened 2021-01-16
AO3-6108 Cache comment counts.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6108

## Purpose

Adds caching for various comment counts on works & admin posts. The cache is automatically expired when a comment is added, removed, or changes visibility, so it shouldn't affect the user experience, but it might reduce the server load a little.

### DIFF
--- a/features/comments_and_kudos/comments_delete.feature
+++ b/features/comments_and_kudos/comments_delete.feature
@@ -13,7 +13,9 @@ Feature: Delete a comment
       And I post the comment "Fail comment" on the work "Awesome story"
       And I delete the comment
     Then I should see "Comment deleted."
-    
+      And I should not see "Comments:1"
+      And I should not see a link "Hide Comments (1)"
+
   Scenario: User deletes a comment they added to a work and which is the parent of another comment
     When I am logged in as "author"
       And I post the work "Awesome story"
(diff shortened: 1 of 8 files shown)


lib/plugins/acts_as_commentable/commentable_entity.rb
```diff
@@ -23,18 +23,42 @@ def count_all_comments
...
     ).count
+  end
+
+  # The total number of visible comments on this commentable. Cached to reduce
+  # computation. The cache is manually expired whenever a comment is added,
+  # removed, or changes visibility, but the cache also expires after an hour in
+  # case of issues with the cache (e.g. stale data when calculating the count).
+  def count_visible_comments
+    @count_visible_comments ||=
+      Rails.cache.fetch(count_visible_comments_key, expires_in: 1.hour) do
```
redsummernight, line 48: - Should we use `race_condition_ttl` here?
- Can you make this a config option?

features/comments_and_kudos/comments_delete.feature
```diff
@@ -34,6 +38,8 @@ Feature: Delete a comment
       And I view the work "Awesome story" with comments
       And I delete the comment
     Then I should see "Comment deleted."
+      And I should not see "Comments:1"
```
redsummernight, line 41: Can we use `I should not see "Comments:"`, since this stat element shouldn't show up at all?

lib/plugins/acts_as_commentable/commentable_entity.rb
tickinginstant: 👍 

I set `race_condition_ttl` to 10 seconds, which seems to be the most common value for `race_condition_ttl` elsewhere in the codebase.

redsummernight (approved):
Thanks!

PR #3963, opened 2021-01-18
AO3-4709 Comment freezing fixes

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4709

## Purpose

* Stops checking comments for spam whenever they are saved. This will allow spam comments to be frozen/unfrozen and prevent errors when you freeze/unfreeze the same comments repeatedly for testing purposes. It will also prevent comments you manually marked as spam from getting reverted by the spam checker.
* Removes the `cursor: pointer` CSS style that was being applied to all manner of things. Now we defer to whatever cursor the browser chooses for specific elements. (This will stop the pointer from showing up on "Frozen" and other things with the `.current` style, incorrectly making it look clickable. It also means some clickable things like buttons will now have the default arrow cursor, because that's what browsers do for those.)
* Update the redirect when freezing and unfreezing comments. Simply redirecting back would result in the comment section collapsing if you'd loaded comments with AJAX. So now we redirect to all comments, expanded, like when you mark something as ham/spam. AO3-6036 is the existing issue for it not redirecting you to the proper page.

## Testing Instructions

* Mark a comment as spam and make sure it doesn't get unmarked as spam when you freeze or unfreeze it.
* Make sure you don't get an error when repeatedly freezing and unfreezing the same comment.
* Make sure the "Frozen" indicator doesn't have the same cursor you see when hovering over a link.
* Make sure you get redirected to the comment's ultimate parent with comments shown when freezing/unfreezing.

### DIFF
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -20,7 +20,7 @@ class Comment < ApplicationRecord
     maximum: ArchiveConfig.COMMENT_MAX,
     too_long: ts("must be less than %{count} characters long.", count: ArchiveConfig.COMMENT_MAX)
 
-  validate :check_for_spam
+  validate :check_for_spam, on: :create
 
   def check_for_spam
     errors.add(:base, ts("This comment looks like spam to our system, sorry! Please try again, or create an account to comment.")) unless check_for_spam?
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -419,11 +419,11 @@ class CommentsController < ApplicationController
     # @comment.full_set.each(&:mark_frozen!)
     if !@comment.iced? && @comment.save
       @comment.set_to_freeze_or_unfreeze.each(&:mark_frozen!)
-      flash[:notice] = t(".success")
+      flash[:comment_notice] = t(".success")
     else
-      flash[:error] = t(".error")
+      flash[:comment_error] = t(".error")
     end
-    redirect_back(fallback_location: root_path)
+    redirect_to_all_comments(@comment.ultimate_parent, show_comments: true)
   end
 
   # PUT /comments/1/unfreeze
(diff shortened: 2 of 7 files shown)


app/models/comment.rb
```diff
@@ -20,7 +20,7 @@ class Comment < ApplicationRecord
     maximum: ArchiveConfig.COMMENT_MAX,
     too_long: ts("must be less than %{count} characters long.", count: ArchiveConfig.COMMENT_MAX)
 
-  validate :check_for_spam
+  validate :check_for_spam, on: :create
```
redsummernight, line 23: Should we include [AO3-6038](https://otwarchive.atlassian.net/browse/AO3-6038) in this release, since it's the same fix?

sarken: I think it's only half a fix for AO3-6038 -- you can still reply to the comment, just not un-spam it in the process -- so it's better to leave it out. Maybe it would be better to update the issue.

redsummernight (approved):

bexsentance:
I tested everything except the spam one, as I don't have an admin account on Test which I think would be needed to freeze a comment marked as spam (I did have one previously, but the login credentials don't seem to work any more). Here are my observations (tested on Firefox for Windows 10).

**Make sure you don't get an error when repeatedly freezing and unfreezing the same comment.**
This was fine; I tested repeatedly freezing and unfreezing the comment several times in a row with no errors.

**Make sure the "Frozen" indicator doesn't have the same cursor you see when hovering over a link.**
This works. My browser now shows a different cursor for the "Freeze Thread" and "Unfreeze Thread" button as for the other clickable buttons; I get an arrow cursor for those buttons and a little hand for the others. I assume this is just a side-effect of the change and can't be helped, but thought I'd mention it. For the "Frozen" indicator I get a carrot so it doesn't look clickable.

**Make sure you get redirected to the comment's ultimate parent with comments shown when freezing/unfreezing.**
This worked for me on single-chapter works - but on works with multiple chapters comments were not shown after freezing or unfreezing a thread. I tested this on various multi-chapter works with the same result (it took me a second to figure out what the variable was, but it happens on all the multi-chapter works I've tested it on).

PR #3964, opened 2021-01-19
AO3-6111 Paginate kudos.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6111

## Purpose

This PR makes the "X more users" link a remote link that loads another page of kudos from the server, instead of a link that uses CSS to unhide the (already loaded) list of all kudos. It also gets rid of the "(collapse)" link.

I modified the `/works/:id/kudos` page so that it returns paginated results, as well. However, I'm not really sure about the phrasing/formatting on this page. It's only displaying user kudos (and therefore the search header will display the number of user kudos), but I didn't want users to get confused about where all their guest kudos went. So I changed the phrasing in the header to `1 - 50 of 101 User Kudos` and added a subheading with the number of guest kudos. I'm not sure if that was the right solution, though.

(I didn't want to repeat the guest kudos count in the paginated section, because then the user might think that they should have X * Y guest kudos, where X is the number of guest kudos and Y is the number of pages. Also, it's easier to miss there.)


app/controllers/chapters_controller.rb
```diff
@@ -57,7 +57,7 @@ def show
           @work.anonymous? ? ts("Anonymous") : @work.pseuds.sort.collect(&:byline).join(', '),
           @work.title + " - Chapter " + @chapter.position.to_s)
 
-      @kudos = @work.kudos.with_user.includes(:user).by_date
```
redsummernight, line 60: Can the `by_date` scope be removed?

tickinginstant: Yeah, it's not used anymore. I removed it.

redsummernight (approved):

PR #3965, opened 2021-01-22
AO3-6111 Reverse kudos order on /works/:id/kudos page.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6111

## Purpose

Make the `/works/:id/kudos` page display kudos in decreasing order by ID.


sarken (approved):
yay, thank you!

PR #3967, opened 2021-01-27
Replace HTML cleaner/parser with more robust one

momijizukamori:
- Upgrade nokogumbo and sanitize
- Add fixup processor based on nokogumbo HTML5 processor
- Replace home-rolled renderer with one based on the redcarpet markdown renderer
- Tweak a handful of unit tests to reflect fixed output
- Add new unit tests for bugfixes

Fixes the following issues:
- AO3-4599 - Parser: PRE tags insert extra spaces
- AO3-3769 - HTML editor's parser won't give an open-paragraph tag to the paragraph after text enclosed in "center" tags, causing weird line spacing
- AO3-3234 - using blockquote tags with paragraphs tags inside results in strange output
- AO3-3565 - HTML parser will not allow formatting tags (italics, bold, emphasis, etc.) to span a line break tag
- AO3-3313 - Using an image as a link causes the HTML parser to move the "close anchor" tag to eliminate that or another link in the same paragraph under certain circumstances
- AO3-3464 - HTML parser eats text after a missing quote upon preview or posting (partial)
- AO3-3916 - Carriage returns won't create line or paragraph breaks in the HTML editor if text has already been through the parser
- AO3-3710 - Blockquote tags develop extra blank paragraphs before them if a work is edited

# Pull Request Checklist

  as the *first* thing in your pull request title (e.g. `AO3-1234 Fix thing`) - Fixes multiple issues

## Issue

https://otwarchive.atlassian.net/browse/AO3-4599 
https://otwarchive.atlassian.net/browse/AO3-3234  
https://otwarchive.atlassian.net/browse/AO3-3464
https://otwarchive.atlassian.net/browse/AO3-3916

## Purpose
This is more or less a ground-up rewrite of the formatter/cleaner code, which uses the combination of the nokogumbo HTML5 parser and the fast, lightweight redcarpet Markdown processor. The HTML5 parser is uses to correct what errors we can, and then the document is processed into Markdown compatible format, markdown special characters are escaped so the processor doesn't convert them, and then the processor is used to convert linebreaks to HTML paragraph and linebreak elements. It may seem a little silly to use a markdown processor while ignoring most of markdown syntax, but doing it this way gives a significant boost to reliability. Markdown is widely used and has a well-defined syntax. This means output  is reliable, and that the libraries generating it are extensively tested, both for output and for security issues.

There are three kind of messy hacks in this, of varying severity. The first is that the Redcarpet library has some incorrect categorization of element types - namely, it thinks `center` is an inline element, and `ins` and `del` are block level ones. I've opened an MR with the project ( https://github.com/vmg/redcarpet/pull/702 ) and have a workaround in place for testing, but if the redcarpet devs are slow about merging I think it would be better to vendorize the fixed version of the library instead of letting the workaround go to prod.

The second is the HTML fixer, which is fully half of this code. I put it in place because the existing version had it, and if we were going to do it, we should do it better, but trying to repair HTML programmatically is very fragile and error prone. The cleaner solution would be to return an error to the user, along with details about that error (nokogumbo's errors are largely fairly informative) so that they can make the fix themselves. I'd be happy to give that work a go, but it's out of scope for this, and ties into the last issue.

The last issue is the anti-pattern of sanitizing and formatting fields before they're saved to database. This results in errors introduced by buggy versions of the sanitzer or formatter being unfixable, because we can't reconstruct the original input, causes issues with editing due to multiple roundtrips through the sanitizer/formatter, results in text being improperly encoded for the context (such as HTML entities in email headers), and in combination with the auto-correcting attempts means we end up losing data sometimes. Standard best pratices are to save the raw user input, and parse/sanitize it as necessary for output, which may vary based on where it's being used - the edit field would show the raw input, with only enough escaping to keep it from breaking the field, text used in email subjects would be escaped to plaintext, and text in an HTML page would be fully HTML-sanitized and encoded. We would want to implement to bad HTML fix first, so that we're only putting well-formed user input into the db. 

I realize this is a much bigger production, both in terms of making sure encoding/sanitizing is performance-optimized, and in terms of making sure we handle both old pre-sanitized data and new raw data properly. I've included an `unparse` function that replaces HTML linebreaks/paragraphs with raw linebreak characters, so that multiple edit/save cycles doesn't generate infinitely nested linebreak elements, but it can't do anything about other HTML changes we make, and it will process any user-input linebreak elements the same way, as it can't tell what the user added versus what the parser added.

## Testing Instructions
I've done testing with a bunch of fics I have sitting around, along with some text files from Project Gutenberg. What this really needs more testing with is stuff with 'weird' mark-up - extensive use of divs/classes, tables, lots of embedded images, etc. 

## Credit

momijizukamori & they/them is fine :)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -51,11 +51,9 @@ gem 'htmlentities'
 gem 'whenever', '~>0.6.2', require: false
 gem 'nokogiri', '>= 1.8.5'
 gem 'mechanize'
-gem 'sanitize', '>= 4.6.5'
-# Until there is a working solution to
-# https://otwarchive.atlassian.net/browse/AO3-4957
-# https://github.com/rubys/nokogumbo/issues/50
-gem 'nokogumbo', '1.4.9'
+gem 'sanitize', '>= 5.0.0'
+gem 'nokogumbo', '> 2.0.0'
+gem 'redcarpet'
 gem 'rest-client', '~> 1.8.0', require: 'rest_client'
 gem 'resque', '>=1.14.0'
 gem 'resque-scheduler'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1041,6 +1041,8 @@ GEM
       simplecov
     coderay (1.1.3)
     colorize (0.8.1)
+    commonmarker (0.21.1)
+      ruby-enum (~> 0.5)
     concurrent-ruby (1.1.7)
     connection_pool (2.2.3)
     crack (0.4.3)
@@ -1185,11 +1187,11 @@ GEM
     netrc (0.11.0)
     newrelic_rpm (6.11.0.365)
     nio4r (2.5.4)
-    nokogiri (1.11.0)
+    nokogiri (1.11.1)
       mini_portile2 (~> 2.5.0)
       racc (~> 1.4)
-    nokogumbo (1.4.9)
-      nokogiri
+    nokogumbo (2.0.4)
+      nokogiri (~> 1.8, >= 1.8.4)
     ntlm-http (0.1.1)
     orm_adapter (0.5.0)
     paperclip (6.1.0)
@@ -1263,6 +1265,7 @@ GEM
       thor (>= 0.19.0, < 2.0)
     raindrops (0.19.1)
     rake (13.0.1)
+    redcarpet (3.5.1)
     redis (3.3.5)
     redis-namespace (1.7.0)
       redis (>= 3.0.4)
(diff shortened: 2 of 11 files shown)


momijizukamori:
Oh, as a quick note, I also ran some benchmarking between the old and new versions of the code, and they come out around the same  - differences were between 2x as fast and 2x as slow. The old parser is faster at very long plaintext files (stuff around the per-chapter character max), the new one is faster at very long html-formatted files. If the recursive nokogumbo formatting step is removed, performance is the same or better as the old code.

sarken:
Hi, momijizukamori!

Thanks for taking a shot at fixing these issues! A rewrite of the parser is something that we've been exploring off and on as time permits, but we're still in the process of figuring out exactly what sort of functionality we want to offer (e.g. Markdown support, options regarding how whitespace is interpreted) and how best to implement it (e.g. by allowing both parsers to coexist for a time). Once we decide how to tackle a full rewrite, we'll have an issue -- probably several -- specifically for that. 

However, for the time being, we'd prefer a piecemeal approach to fixing bugs to a total rewrite. If you'd like to take a shot at fixing any of the issues while keeping the current setup, or to work on something else from the issue tracker, you're welcome to create a Jira account and we'll give you permissions to claim and transition issues. (If you do that, we ask that the Full Name on the Jira account either match or include the name you wish to be credited by -- it helps us keep things organized.)

potpotkettle:
FWIW, a fairly large, different rewrite was done in 2022: https://github.com/otwcode/otwarchive/pull/4391

PR #3968, opened 2021-02-01
AO3-6122 Allow fandom tags and tags without fandoms to be unwrangleable

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6122

## Purpose

Removes changes that would prevent the use of certain tags or works or render works uneditable:

* unwrangleable validation check for fandom tags and tags that have no fandom
* visibility change on the unwrangleable checkbox for fandoms (arguably, this one could stay, though it may give us a false sense of security)

I didn't do a straight revert of #3891 because it would be nice to keep the test and syntax tweaks.  

## Testing Instructions

Refer to Jira.


sarken:
I keep getting all sorts of unrelated intermittent failures on this, so I'm just going to leave it with the failing check instead of rerunning it yet again.

redsummernight (approved):

PR #3971, opened 2021-02-07
AO3-6115 Fix rate limiting on user login attempts

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6115

## Purpose

Fix Rack::Attack configurations (URLs and parameters to check).

## Testing Instructions

Use cURL to POST to /users/login? idk.

### DIFF
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -62,28 +62,23 @@ class Rack::Attack
   login_limit = ArchiveConfig.RATE_LIMIT_LOGIN_ATTEMPTS
   login_period = ArchiveConfig.RATE_LIMIT_LOGIN_PERIOD
 
-  # Throttle POST requests to /login by IP address
+  # Throttle POST requests to /users/login by IP address
   #
   # Key: "rack::attack:#{Time.now.to_i/:period}:logins/ip:#{req.ip}"
-  throttle('logins/ip', limit: login_limit, period: login_period) do |req|
-    if req.path == '/login' && req.post?
-      req.ip
-    end
+  throttle("logins/ip", limit: login_limit, period: login_period) do |req|
+    req.ip if req.path == "/users/login" && req.post?
   end
 
-  # Throttle POST requests to /login by email param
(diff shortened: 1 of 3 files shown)


config/initializers/rack_attack.rb
```diff
@@ -62,28 +62,23 @@ class Rack::Attack
...
-    end
+  throttle("logins/ip", limit: login_limit, period: login_period) do |req|
+    req.ip if req.path == "/users/login" && req.post?
   end
 
-  # Throttle POST requests to /login by email param
+  # Throttle POST requests to /users/login by login param (user name or email)
   #
-  # Key: "rack::attack:#{Time.now.to_i/:period}:logins/email:#{req.email}"
+  # Key: "rack::attack:#{Time.now.to_i/:period}:logins/email:#{login}"
```
sarken, line 74: Question! Why is it `#{login}` in the key instead of `#{req.login}`?

redsummernight: `req.login` is not actually field. Neither was `req.email`, which was copied from Rack::Attack's examples. The value should be `req.params.dig("user", "login").presence` but I didn't want to repeat that in the comment.

spec/requests/rack_attack_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+      expect(response).to have_http_status(:too_many_requests)
+    end
+
+    it "does not throttle the next attempt from the same IP after some time" do
+      travel ArchiveConfig.RATE_LIMIT_LOGIN_PERIOD.seconds
+      post user_session_path, params: unique_user_params.to_query, env: { "REMOTE_ADDR" => ip }
+      expect(response).to have_http_status(:ok)
+    end
+
+    it "does not throttle an attempt from a different IP" do
```
sarken, line 41: Could I be annoying and suggest flipping the order, so `it "does not throttle an attempt from a different IP"` is above `it "does not throttle the next attempt from the same IP after some time"`? I think that way it might be clearer that the attempt from a different IP is coming before the rate limiting period is up.

sarken (approved):

PR #3972, opened 2021-02-08
AO3-6124 Keep admin post translation comment permissions in sync with original

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6124

## Purpose

Makes it so when you post a translation of an admin post, the translation inherits its comment permissions from the original admin post.

Also, if you update the comment permissions of an admin post that has translations, it updates the comment permissions of the translation.

## Testing Instructions

Refer to Jira

### DIFF
--- a/app/models/admin_post.rb
+++ b/app/models/admin_post.rb
@@ -70,6 +70,14 @@ class AdminPost < ApplicationRecord
     end
   end
 
+  private
+
+  def expire_cached_home_admin_posts
+    unless Rails.env.development?
+      Rails.cache.delete("home/index/home_admin_posts")
+    end
+  end
+
   def inherit_translated_post_comment_permissions
     return unless translated_post_id.present? && AdminPost.find_by(id: translated_post_id)
 
@@ -77,21 +85,13 @@ class AdminPost < ApplicationRecord
   end
 
   def update_translation_comment_permissions
-    return unless translations.present?
+    return if translations.blank?
 
     transaction do
-      AdminPost.where(translated_post_id: id).each do |post|
+      AdminPost.where(translated_post_id: id).find_each do |post|
         post.comment_permissions = self.comment_permissions
         post.save
       end
-    end
-  end
-
-  private
-
-  def expire_cached_home_admin_posts
-    unless Rails.env.development?
-      Rails.cache.delete("home/index/home_admin_posts")
     end
   end
 end
--- a/app/models/admin_post.rb
+++ b/app/models/admin_post.rb
@@ -38,7 +38,8 @@ class AdminPost < ApplicationRecord
 
   scope :for_homepage, -> { order("created_at DESC").limit(ArchiveConfig.NUMBER_OF_ITEMS_VISIBLE_ON_HOMEPAGE) }
 
-  after_save :expire_cached_home_admin_posts
+  before_validation :inherit_translated_post_comment_permissions
+  after_save :expire_cached_home_admin_posts, :update_translation_comment_permissions
   after_destroy :expire_cached_home_admin_posts
 
   # Return the name to link comments to for this object
@@ -66,6 +67,23 @@ class AdminPost < ApplicationRecord
   def translated_post_must_exist
     if translated_post_id.present? && AdminPost.find_by(id: translated_post_id).nil?
       errors.add(:translated_post_id, 'does not exist')
+    end
+  end
+
+  def inherit_translated_post_comment_permissions
+    return unless translated_post_id.present? && AdminPost.find_by(id: translated_post_id)
+
+    self.comment_permissions = translated_post.comment_permissions
+  end
+
+  def update_translation_comment_permissions
+    return unless translations.present?
+
+    transaction do
+      AdminPost.where(translated_post_id: id).each do |post|
+        post.comment_permissions = self.comment_permissions
+        post.save
+      end
     end
   end
 
--- a/app/models/admin_post.rb
+++ b/app/models/admin_post.rb
@@ -38,8 +38,7 @@ class AdminPost < ApplicationRecord
 
   scope :for_homepage, -> { order("created_at DESC").limit(ArchiveConfig.NUMBER_OF_ITEMS_VISIBLE_ON_HOMEPAGE) }
 
-  before_validation :inherit_translated_post_comment_permissions
-  after_save :expire_cached_home_admin_posts, :update_translation_comment_permissions
+  after_save :expire_cached_home_admin_posts, :inherit_translated_post_comment_permissions, :update_translation_comment_permissions
   after_destroy :expire_cached_home_admin_posts
 
   # Return the name to link comments to for this object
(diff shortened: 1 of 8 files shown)


features/comments_and_kudos/comments_adminposts.feature
```diff
@@ -112,3 +112,15 @@ Feature: Commenting on admin posts
       And I press "Comment"
     Then I should see "Comment created!"
       And I should see "zug zug"
+
+  Scenario: Translation of admin post with comments disabled
```
sarken, line 116: Whoops, need a test for changing the original's permissions.

spec/controllers/admin_posts_controller_spec.rb
```diff
@@ -56,33 +68,41 @@
...
+          context "with valid translated_post_id" do
+            let!(:translation) { create(:admin_post, translated_post_id: post.id) }
+
+            context "with valid comment_permissions" do
+              it "does not change comment_permissions and redirects with notice" do
+                admin.update(roles: [admin_role])
+                fake_login_admin(admin)
+                expect {
+                  put :update,
+                  params: {
```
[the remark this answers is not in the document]
sarken: Could someone tell me where the alignment issue is? I can't see it.

app/models/admin_post.rb
```diff
@@ -38,7 +38,8 @@ class AdminPost < ApplicationRecord
 
   scope :for_homepage, -> { order("created_at DESC").limit(ArchiveConfig.NUMBER_OF_ITEMS_VISIBLE_ON_HOMEPAGE) }
 
-  after_save :expire_cached_home_admin_posts
+  before_validation :inherit_translated_post_comment_permissions
```
redsummernight, line 41: With this callback, calling `.valid?` on a translation post could change its comment permissions, which feels kinda unexpected. Is there any reason we use `before_validation` instead of `before_save`?

app/models/admin_post.rb
```diff
@@ -76,4 +77,21 @@ def expire_cached_home_admin_posts
       Rails.cache.delete("home/index/home_admin_posts")
     end
   end
+
+  def inherit_translated_post_comment_permissions
+    return unless translated_post_id.present? && AdminPost.find_by(id: translated_post_id)
```
redsummernight, line 82: Can this be `return if translated_post.blank?`?

app/models/admin_post.rb
```diff
@@ -76,4 +77,21 @@ def expire_cached_home_admin_posts
...
+
+    self.comment_permissions = translated_post.comment_permissions
+  end
+
+  def update_translation_comment_permissions
+    return if translations.blank?
+
+    transaction do
+      AdminPost.where(translated_post_id: id).find_each do |post|
```
redsummernight, line 91: Can this be `translations.find_each`?

spec/controllers/admin_posts_controller_spec.rb
```diff
@@ -44,9 +44,21 @@
         end
       end
 
-      context "when admin has correct authorization" do
-        %w[superadmin board communications support translation].each do |admin_role|
-          context "with #{admin_role} role" do
+      context "with invalid translated_post_id" do
```
redsummernight, line 47: Updating a post would fail without correct admin roles, and the params of the update make no difference, so I don't think this test is necessary. I think we can have just one test under "when admin does not have correct authorization", and remove both nested contexts "with valid title" and "with invalid translated_post_id".

spec/controllers/admin_posts_controller_spec.rb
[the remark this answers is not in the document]
redsummernight: `params: {` needs to be indented 2 spaces (along with everything in it). Or better still, you can just move it to the previous line (`put :update, params: {`).

features/comments_and_kudos/comments_adminposts.feature
```diff
@@ -112,3 +112,28 @@ Feature: Commenting on admin posts
...
+    Given I have posted an admin post with comments disabled
+      And basic languages
+      And I am logged in as a "translation" admin
+    When I make a translation of an admin post
+    Then I should see "Sorry, this news post doesn't allow comments."
+    When I follow "Edit Post"
+    Then I should see "No one can comment"
+      And I should not see "Registered users and guests can comment"
```
redsummernight, line 137: I don't like these negative assertions because they won't fail if we change the text of the options. Can we do without them?

(A much better test would be to check for a selected disabled radio button "No one can comment" option, but as discussed elsewhere we're not doing that this issue.)

app/models/admin_post.rb
sarken: There's not any reason we need to, no -- I'd just recently read a suggestion that said anytime you change a value with a callback, you should use `before_validation` rather than `before_save` to make sure you weren't saving something with an invalid value. But we don't validate `comment_permissions` afaik (which is probably an issue of its own), so it doesn't really matter here.

features/comments_and_kudos/comments_adminposts.feature
sarken: Hmm. I wanted something to test that the section with the radio buttons was missing when you went to edit the post, since otherwise that's totally uncovered. Maybe I can check any make sure the page has no radio buttons? I don't think there are any others on that page.

spec/controllers/admin_posts_controller_spec.rb
sarken: Can do! I was just shuffling what already existed so I didn't have to add a third "when admin has correct authorization" section, but I'm game for deleting existing tests.

spec/controllers/admin_posts_controller_spec.rb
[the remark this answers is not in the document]
sarken: ty!

features/comments_and_kudos/comments_adminposts.feature
redsummernight: We'll need a new test step for that as well. I can live with a TODO.

app/models/admin_post.rb
sarken: Yeah, I was clearly overthinking here -- I wanted to make sure things wouldn't explode if we were calling this on a post with a translated post that didn't exist in the database. But the translated_post_id validation runs before we save, so we never actually get to the point of invoking this on a post without a valid translated_post_id.

spec/controllers/admin_posts_controller_spec.rb
```diff
@@ -56,33 +68,41 @@
...
           end
+
+          context "with valid translated_post_id" do
+            let!(:translation) { create(:admin_post, translated_post_id: post.id) }
+
+            context "with valid comment_permissions" do
+              it "does not change comment_permissions and redirects with notice" do
+                admin.update(roles: [admin_role])
+                fake_login_admin(admin)
+                expect {
```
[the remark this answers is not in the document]
redsummernight: Can you use `do...end` instead?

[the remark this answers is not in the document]
sarken: Can do! I was going to argue we normally ignore that for specs, but I searched before opening my big mouth and I see we do have some like that.

redsummernight (approved):

PR #3974, opened 2021-02-21
AO3-6135 Move server lists out of deploy scripts

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6135

## Purpose

They are now in servers.yml, which can be more easily replaced by automation.

Remove unused stage i18n.

## Testing Instructions

Staging and production deploys should work normally. I also checked that Capistrano still gets the same set of servers and roles by adding this task to `namespace :deploy` in `config/deploy.rb`:

```rb
  desc "Print servers"
  task :debug do
    puts "All servers: #{find_servers.size}"
    puts "All app servers: #{find_servers(roles: :app).size}"
    puts "All db servers: #{find_servers(roles: :db).join(', ')}"
    puts "All schedulers servers: #{find_servers(roles: :schedulers).join(', ')}"
    puts "All workers servers: #{find_servers(roles: :workers).join(', ')}"
    puts "All front-ends: #{find_servers(roles: :web).join(', ')}"
    find_servers(only: { primary: true }).each do |server|
      puts server.inspect
    end
  end
```

Then I ran:
```sh
bundle exec cap production deploy:debug
bundle exec cap staging deploy:debug
```
on both the main branch and this PR branch, making sure the output is the same.

We're still using the old 2.x version of Capistrano, and the docs for that is at https://github.com/capistrano/capistrano-2.x-docs.


config/deploy/production.rb
```diff
@@ -1,34 +1,3 @@
-#
-# For a detailed explanation of roles (ie :web, :app, etc) see:
-# http://stackoverflow.com/questions/1155218/what-exactly-is-a-role-in-capistrano
-#
-# The :primary attribute is used for tasks we only want to run on one machine
-# 
-require 'capistrano/gitflow_version'
```
zz9pzza, line 7: Why are we loosing this ?

redsummernight: We're not, I moved the comments to the top, the `require` is on line 10.

PR #3977, opened 2021-02-23
AO3-5525 Block password resets for certain users

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5525

## Purpose

Allows us to provide a list of user IDs in the config file that we do not want to send password reset emails to. Replaces an old (April 2018) hot fix that never made it into the repo.

This is one possible way to approach this. Two other options:

1. Place the check in the [Devise mailer's reset_password_instructions method](https://github.com/otwcode/otwarchive/blob/1f7610f25b28388bd6b2beb4715f2a80ece78ef7/app/mailers/archive_devise_mailer.rb#L11) _and_ include error message handling here in the controller. This would be slightly neater, since the error message could be inside the `super do |user|` block. However, that just makes it harder to send the email a reset email from the console in case of an emergency, or to one day allow Support/PAC to send such emails from admin.
2. In the User model, override Devise's `send_reset_password_instructions` method as described in [this StackOverflow answer](https://stackoverflow.com/a/62299668). This is ugly and has the same drawback as the first one.

## Testing Instructions

Refer to Jira

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -201,3 +201,6 @@ Style/SymbolArray:
 
 Style/TernaryParentheses:
   EnforcedStyle: require_parentheses_when_complex
+
+Style/AndOr:
+  EnforcedStyle: conditionals
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -160,6 +160,9 @@ RSpec/NestedGroups:
 RSpec/PredicateMatcher:
   Enabled: false
 
+Style/AndOr:
+  EnforcedStyle: conditionals
+
 Style/ClassAndModuleChildren:
   Enabled: false
 
@@ -201,6 +204,3 @@ Style/SymbolArray:
 
 Style/TernaryParentheses:
   EnforcedStyle: require_parentheses_when_complex
-
-Style/AndOr:
-  EnforcedStyle: conditionals
--- a/app/controllers/users/passwords_controller.rb
+++ b/app/controllers/users/passwords_controller.rb
@@ -6,7 +6,7 @@ class Users::PasswordsController < Devise::PasswordsController
   layout "session"
 
   def create
-    user_id = User.find_for_authentication(login: params[:user][:login])&.id
+    user_id = User.find_for_authentication(login: resource_params[:login])&.id
     if ArchiveConfig.PROTECTED_USER_IDS.include?(user_id)
       flash[:error] = t(".reset_blocked", contact_abuse_link: view_context.link_to(t(".contact_abuse"), new_abuse_report_path)).html_safe
       redirect_to :root
--- a/app/controllers/users/passwords_controller.rb
+++ b/app/controllers/users/passwords_controller.rb
@@ -6,8 +6,7 @@ class Users::PasswordsController < Devise::PasswordsController
   layout "session"
 
   def create
-    user = User.find_for_authentication(login: resource_params[:login])
-    if user.protected_user
+    if User.find_for_authentication(login: resource_params[:login])&.protected_user
       flash[:error] = t(".reset_blocked", contact_abuse_link: view_context.link_to(t(".contact_abuse"), new_abuse_report_path)).html_safe
       redirect_to :root
     else
--- a/app/controllers/users/passwords_controller.rb
+++ b/app/controllers/users/passwords_controller.rb
@@ -8,7 +8,7 @@ class Users::PasswordsController < Devise::PasswordsController
   def create
     if User.find_for_authentication(resource_params.permit(:login))&.protected_user
       flash[:error] = t(".reset_blocked", contact_abuse_link: view_context.link_to(t(".contact_abuse"), new_abuse_report_path)).html_safe
-      redirect_to root_path and return
+      redirect_to root_path && return
     end
     super do |user|
       flash.now[:notice] = ts("We couldn't find an account with that email address or username. Please try again?") if user.nil? || user.new_record?
--- a/app/controllers/users/passwords_controller.rb
+++ b/app/controllers/users/passwords_controller.rb
@@ -6,15 +6,12 @@ class Users::PasswordsController < Devise::PasswordsController
   layout "session"
 
   def create
-    if User.find_for_authentication(login: resource_params[:login])&.protected_user
+    if User.find_for_authentication(resource_params.permit(:login))&.protected_user
       flash[:error] = t(".reset_blocked", contact_abuse_link: view_context.link_to(t(".contact_abuse"), new_abuse_report_path)).html_safe
-      redirect_to :root
-    else
-      super do |user|
-        if user.nil? || user.new_record?
-          flash.now[:notice] = ts("We couldn't find an account with that email address or username. Please try again?")
-        end
-      end
+      redirect_to root_path and return
+    end
+    super do |user|
+      flash.now[:notice] = ts("We couldn't find an account with that email address or username. Please try again?") if user.nil? || user.new_record?
     end
   end
 end
--- a/app/controllers/users/passwords_controller.rb
+++ b/app/controllers/users/passwords_controller.rb
@@ -7,7 +7,7 @@ class Users::PasswordsController < Devise::PasswordsController
 
   def create
     user_id = User.find_for_authentication(login: params[:user][:login])&.id
-    if ArchiveConfig.TARGETED_USER_IDS.include?(user_id)
+    if ArchiveConfig.PROTECTED_USER_IDS.include?(user_id)
       flash[:error] = t(".reset_blocked", contact_abuse_link: view_context.link_to(t(".contact_abuse"), new_abuse_report_path)).html_safe
       redirect_to :root
     else
--- a/app/controllers/users/passwords_controller.rb
+++ b/app/controllers/users/passwords_controller.rb
@@ -6,9 +6,15 @@ class Users::PasswordsController < Devise::PasswordsController
   layout "session"
 
   def create
-    super do |user|
-      if user.nil? || user.new_record?
-        flash.now[:notice] = ts("We couldn't find an account with that email address or username. Please try again?")
+    user_id = User.find_for_authentication(login: params[:user][:login])&.id
+    if ArchiveConfig.TARGETED_USER_IDS.include?(user_id)
+      flash[:error] = t(".reset_blocked", contact_abuse_link: view_context.link_to(t(".contact_abuse"), new_abuse_report_path)).html_safe
+      redirect_to :root
+    else
+      super do |user|
+        if user.nil? || user.new_record?
+          flash.now[:notice] = ts("We couldn't find an account with that email address or username. Please try again?")
+        end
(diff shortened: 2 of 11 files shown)


app/controllers/users/passwords_controller.rb
```diff
@@ -6,9 +6,15 @@ class Users::PasswordsController < Devise::PasswordsController
   layout "session"
 
   def create
-    super do |user|
-      if user.nil? || user.new_record?
-        flash.now[:notice] = ts("We couldn't find an account with that email address or username. Please try again?")
+    user_id = User.find_for_authentication(login: params[:user][:login])&.id
```
redsummernight, line 9: Can you use `resource_params` as the argument?

- https://github.com/heartcombo/devise/blob/45b831c4ea5a35914037bd27fe88b76d7b3683a4/app/controllers/devise_controller.rb#L207-L209
- https://github.com/heartcombo/devise/blob/45b831c4ea5a35914037bd27fe88b76d7b3683a4/test/controllers/internal_helpers_test.rb#L38-L51

features/step_definitions/user_steps.rb
```diff
@@ -166,6 +166,11 @@
   work.creatorships.unapproved.each(&:accept!)
 end
 
+Given /^the user "(.*?)" is blocked from resetting their password$/ do |login|
```
redsummernight, line 169: Can you use Cucumber expressions here?

features/users/authenticate_users.feature
```diff
@@ -226,22 +226,21 @@ Feature: User Authentication
...
-      | login | user  | action                   |
-      | sam   | sam   | not see "Log In"         |
-      | sam   | sam   | see "Log Out"            |
-      | sam   | sam   | see "Preferences" within "#dashboard"    |
-      | sam   | dean  | see "Log Out"            |
-      | sam   | dean  | not see "Preferences" within "#dashboard" |
-      | sam   | dean  | not see "Log In"         |
+  Scenario: Passwords cannot be reset for users who are on the config list of users being trolled or harassed.
+    Given the following activated user exists
+      | login  | email            | password |
```
redsummernight, line 231: Can you omit the password column?

app/controllers/users/passwords_controller.rb
```diff
@@ -6,9 +6,15 @@ class Users::PasswordsController < Devise::PasswordsController
   layout "session"
 
   def create
-    super do |user|
-      if user.nil? || user.new_record?
-        flash.now[:notice] = ts("We couldn't find an account with that email address or username. Please try again?")
+    user_id = User.find_for_authentication(login: resource_params[:login])&.id
```
redsummernight, line 9: You can do `User.find_for_authentication(resource_params)`.

sarken: Hmm. I'll try it again, but it didn't seem to ever find the user when I did that -- the test would fail here:

```    And I fill in "Email address or user name" with "target"                                                   # features/step_definitions/web_steps.rb:52
    And I press "Reset Password"                                                                               # features/step_definitions/web_steps.rb:40
    Then I should be on the home page                                                                          # features/step_definitions/web_steps.rb:268
      expected: "/"
           got: "/users/login" (using ==) (RSpec::Expectations::ExpectationNotMetError)
      ./features/step_definitions/web_steps.rb:271:in `/^(?:|I )should be on (.+)$/'
      features/users/authenticate_users.feature:238:in `Then I should be on the home page'
      ```

File: features/users/authenticate_users.feature (hunk 3 above)
Comment (sarken) [reply features/users/authenticate_users.feature]: askjaglks I spent 15 minutes trying to figure out why omitting the password column was making everything get all wonky... and it's because I was leaving off the damn `|` following the email column.

File: app/controllers/users/passwords_controller.rb (hunk 4 above)
Comment (sarken) [reply app/controllers/users/passwords_controller.rb:9]: Now that I have straightened out the password column issue, ahem, I'm getting a more helpful error, which happens in every scenario we press the Reset Password button:

```
      unable to convert unpermitted parameters to hash (ActionController::UnfilteredParameters)
      ./app/models/user.rb:203:in `find_first_by_auth_conditions'
      ./app/controllers/users/passwords_controller.rb:9:in `create'
      ./features/step_definitions/web_steps.rb:42:in `block (2 levels) in <top (required)>'
      ./features/step_definitions/web_steps.rb:6:in `with_scope'
      ./features/step_definitions/web_steps.rb:41:in `/^(?:|I )press "([^"]*)"(?: within "([^"]*)")?$/'
      features/users/authenticate_users.feature:237:in `And I press "Reset Password"'
```

The referenced line is 203: https://github.com/otwcode/otwarchive/blob/c2f27c95463958b1fd184670616364ba9592029c/app/models/user.rb#L202-L203

But the `resource_params` are definitely `{"login"=>"sarken"}`

But testing manually with the code I have (`User.find_for_authentication(login: resource_params[:login])&.id`) is also erroring now, even though the tests are passing. :thinking:

Comment (sarken) [reply app/controllers/users/passwords_controller.rb:9]: Okay, I _did_ figure out why my manual tests were erroring: I needed to restart after changing the name of the config  variable. That's not embarrassing at all.

But what I can't figure out is how to sweet talk Devise into permitting these parameters. The [documentation](https://github.com/heartcombo/devise#strong-parameters) made it seem like you could only permit additional parameters for three particular actions from the Registrations or Sessions controllers, but the [source](https://github.com/heartcombo/devise/blob/45b831c4ea5a35914037bd27fe88b76d7b3683a4/lib/devise/parameter_sanitizer.rb#L82-L87) made it seem like I could specify any action for any controller... but no luck with either of these as a `before_action` in the Passwords controller:

```
  def configure_permitted_parameters
    devise_parameter_sanitizer.permit(:create) do |user|
      user.permit(:login)
    end
  end

  def configure_permitted_parameters
    devise_parameter_sanitizer.permit(:create, keys: [:login])
  end
  ```

I even tried using sign_up, sign_in, and account_update just for the hell of it, in case they were the only allowed options.

features/users/authenticate_users.feature
```diff
@@ -226,22 +226,21 @@ Feature: User Authentication
...
-
-    Examples:
-      | login | user  | action                   |
-      | sam   | sam   | not see "Log In"         |
-      | sam   | sam   | see "Log Out"            |
-      | sam   | sam   | see "Preferences" within "#dashboard"    |
-      | sam   | dean  | see "Log Out"            |
-      | sam   | dean  | not see "Preferences" within "#dashboard" |
-      | sam   | dean  | not see "Log In"         |
+  Scenario: Passwords cannot be reset for users who have been given the protected role due to trolling or harassment.
```
sarken, line 229: There's already a test for actually assigning roles in admin_manage_users.feature, but I can go through the whole process here as well if we'd rather.

https://github.com/otwcode/otwarchive/blob/398a1f8c20238fa1a7ca0c12d3b1367c974d84da/features/admins/users/admin_manage_users.feature#L28-L43

app/controllers/users/passwords_controller.rb
redsummernight: For the record (1):
```rb
before_action :configure_permitted_parameters, only: [:create]
def configure_permitted_parameters
  devise_parameter_sanitizer.permit(:create, keys: [:login])
end
```
This would work if you also call:
```rb
User.find_for_authentication(devise_parameter_sanitizer.sanitize(:create))
```

For the record (2): with password resets, Devise ends up permitting all parameters, and does [extra filtering to get the `:login` key](https://github.com/heartcombo/devise/blob/45b831c4ea5a35914037bd27fe88b76d7b3683a4/lib/devise/models/authenticatable.rb#L290-L297). So no different from `login: resource_params[:login]`.

That said, there's a nicer way we can write it:
```rb
User.find_for_authentication(resource_params.permit(:login))
```

features/users/authenticate_users.feature
redsummernight: No, I think it's fine to leave testing role assignment to that other file.

app/controllers/users/passwords_controller.rb
```diff
@@ -6,9 +6,14 @@ class Users::PasswordsController < Devise::PasswordsController
   layout "session"
 
   def create
-    super do |user|
-      if user.nil? || user.new_record?
-        flash.now[:notice] = ts("We couldn't find an account with that email address or username. Please try again?")
+    if User.find_for_authentication(login: resource_params[:login])&.protected_user
+      flash[:error] = t(".reset_blocked", contact_abuse_link: view_context.link_to(t(".contact_abuse"), new_abuse_report_path)).html_safe
+      redirect_to :root
```
redsummernight, line 11: Can you return here instead of having an else block?

sarken: Good idea! I'm also changing `:root` to `root_path` -- while a quick Google leaves me fairly certain symbols are treated as `_path` rather than `_url`, I'd rather be explicit so the next person to look at the code can be sure without Googling.

app/controllers/users/passwords_controller.rb
```diff
@@ -6,10 +6,12 @@ class Users::PasswordsController < Devise::PasswordsController
   layout "session"
 
   def create
+    if User.find_for_authentication(resource_params.permit(:login))&.protected_user
+      flash[:error] = t(".reset_blocked", contact_abuse_link: view_context.link_to(t(".contact_abuse"), new_abuse_report_path)).html_safe
+      redirect_to root_path and return
```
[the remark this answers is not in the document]
sarken: No. Bad dog.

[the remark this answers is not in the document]
redsummernight: This would work if you use `redirect_to(root_path) && return`.

app/controllers/users/passwords_controller.rb
sarken: Ah, thank you! I think that makes sense to me now. For some reason, I was thinking `configure_permitted_parameters` alone would magically alter how `find_for_authentication` treated `resource_params`. But that was silly -- it's not like defining the permitted parameters works that way in any other controller; we define the parameters and then reference that method when passing it along to the model.

app/controllers/users/passwords_controller.rb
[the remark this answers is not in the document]
sarken: Do we prefer one over the other? (And if so... maybe we should document it.)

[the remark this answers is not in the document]
redsummernight: We use both styles, so let's relax the check. You can add this to stop the complaints:
```yml
Style/AndOr:
  EnforcedStyle: conditionals
```

.rubocop.yml
```diff
@@ -201,3 +201,6 @@ Style/SymbolArray:
 
 Style/TernaryParentheses:
   EnforcedStyle: require_parentheses_when_complex
+
+Style/AndOr:
```
redsummernight, line 205: Can you have this sorted?

redsummernight (approved):

PR #3978, opened 2021-02-23
AO3-3498 Display word count for drafts

britnesissom:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-3498

## Purpose

When a draft or multiple drafts exist for an unposted work, display the word count as the sum of all drafts. For a posted work with drafts, display the word count as the sum of posted chapters only.

## Testing Instructions

Scenario 1
1. Create work with whichever fields you desire filled in
2. Click “preview” 
3. On the preview screen, click “save as draft”. 
4. go to your user’s drafts page and click “post draft”
5. word count will now have the correct number instead of being blank

Scenario 2
1. Create work with whichever fields you desire filled in
2. Click “preview” 
3. On the preview screen, click “save as draft”. 
4. Add a chapter to the work and save it as a draft too
5. word count will be the sum of these two chapters
6. Post the first chapter
7. word count will now be the count for the first chapter only

## Credit

agenderdanvers

they/them


sarken (changes requested):
Yay, thank you for working on this! There are some style comments, but the fix looks good.

Unfortunately, one of our coders noted that the changes to `post_first_chapter` mean that saving a previously-posted work with a draft first chapter won't cause the first chapter to be posted. That's ultimately the desired behavior, but we have two issues that make it undesirable at the present time: [AO3-4313](https://otwarchive.atlassian.net/browse/AO3-4313) & [AO3-5627](https://otwarchive.atlassian.net/browse/AO3-5627)

Because of that, we definitely don't want to merge this until pull request #3938 (for AO3-4313) is updated. Ideally, we'd also fix AO3-5627 in the same release, although I don't think that's quite as important as getting AO3-4313 fixed.

app/models/work.rb
```diff
@@ -770,7 +775,12 @@ def set_word_count(preview = false)
         self.word_count += chapter.set_word_count
       end
     else
-      self.word_count = Chapter.select("SUM(word_count) AS work_word_count").where(work_id: self.id, posted: true).first.work_word_count
+      # AO3-3498: work created but not posted = draft so get word count for draft
+      if !self.posted
```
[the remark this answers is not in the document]
tickinginstant: If you want to simplify it further, I think you should also be able to compute the total word count with:
```
if self.posted
  chapters.posted.sum(:word_count)
else
  chapters.sum(:word_count)
end
```
That would also avoid Hound's complaint about `where(...).first`. (But that's a flaw in the existing code, so I don't think it's necessary to fix it.)

britnesissom:
@sarken @tickinginstant thanks for the coding suggestions! this is my first time using ruby/ror and cucumber so I did what I thought was best based on previous code ha. I will make these changes

sarken (commented):
The code looks good! I just had a few more comments about code comment style and how the last two tests can be condensed.

One thing I did notice was that, even when I pulled both this **and** #3938 into a single branch, I got some odd behavior when deleting a work's only posted chapter as described in [AO3-5627](https://otwarchive.atlassian.net/browse/AO3-5627).

Regardless of whether I was logged in as an admin, a logged out user, or a logged in user who isn't a creator or invited co-creator on the work, I got the following error:

>Safari Can't Open the Page
>
>Too many redirects occurred trying to open "localhost:3000/works/1677723." This might occur if you open a page that is redirected to open another page which then is redirected to open the original page.

I don't think that needs to block us from merging this, since it's still arguably better than the current behavior (i.e. unexpectedly posting the chapter) I noted in AO3-5627's comments, but we'll definitely need to update the issue once this is merged.

features/works/work_drafts.feature
```diff
@@ -138,3 +138,50 @@ Feature: Work Drafts
...
+        And I go to my drafts page
+        And I follow "Post Draft"
+      Then I should be on the work "Unicorns are everywhere"
+        And I should see "Words:5"
+
+    Scenario: Word count should equal all draft chapters' word counts if work isn't posted
+      Given I am logged in as "test_user"
+        And I set up the draft "Unicorns are everywhere"
+        And I fill in "content" with "Help there are unicorns everywhere"
+      When I press "Preview"
```
sarken, line 157: This should probably be an "And", since it's still part of the initial set up.

features/works/work_drafts.feature
```diff
@@ -138,3 +138,50 @@ Feature: Work Drafts
...
+      When a chapter is set up for "Unicorns are everywhere"
+        And I press "Preview"
+      When I press "Save As Draft"
+      Then I should see "Words:27"
+
+      Scenario: When posting chapter(s) in unpublished multichapter work, word count should equal posted chapter(s) word count
+      Given I am logged in as "test_user"
+        And I set up the draft "Unicorns are everywhere"
+        And I fill in "content" with "Help there are unicorns everywhere"
+      When I press "Preview"
```
sarken, line 172: This should also probably be an "And", since it's part of the initial set up of the draft (the "Given" steps).

features/works/work_drafts.feature
```diff
@@ -138,3 +138,50 @@ Feature: Work Drafts
...
+      Then I should see "Words:16"
+      When a chapter is set up for "Unicorns are everywhere"
+        And I press "Preview"
+        And I press "Save As Draft"
+      Then I should see "Words:27"
+      When I view the work "Unicorns are everywhere"
+        And I press "Post Chapter"
+      Then I should see "Words:5"
+      When I follow "Next Chapter"
+      When I press "Post Chapter"
```
sarken, line 186: This should be "And" as well, since the step before it also begins with "When"

sarken (commented):
Three more tiny tweaks to the tests and this should be ready to go -- thank you!

britnesissom:
@sarken sorry for all the updates. My tests will be better next time

sarken (approved):
Looks good, thank you! And no need to apologize -- it always takes a while to get used to the style on a new project. That's what code reviewers are here to help with!

PR #3979, opened 2021-02-25
AO3-6137 Log admin changes to user roles

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6137

## Purpose

Create a log item when an admin changes a user's roles.

## Testing Instructions

Refer to Jira.

### DIFF
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -3,7 +3,7 @@
 # "moderator" for an instance of a model (i.e., an object), a model class,
 # or without any specification at all.
 class Role < ApplicationRecord
-  has_many :roles_users, dependent: :destroy
+  has_many :roles_users
   has_many :users, through: :roles_users
   belongs_to :authorizable, polymorphic: true
 
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -3,7 +3,7 @@
 # "moderator" for an instance of a model (i.e., an object), a model class,
 # or without any specification at all.
 class Role < ApplicationRecord
-  has_many :roles_users
+  has_many :roles_users, dependent: :destroy
   has_many :users, through: :roles_users
   belongs_to :authorizable, polymorphic: true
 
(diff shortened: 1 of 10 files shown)


app/models/roles_user.rb
```diff
@@ -1,4 +1,26 @@
 class RolesUser < ApplicationRecord
   belongs_to :user
   belongs_to :role
+
+  after_create :log_role_addition
+  def log_role_addition
+    admin = User.current_user
+    note = "Change made by #{admin.login}"
+    user = User.find_by(id: user_id)
+    user.create_log_item( options = { admin_id: admin.id,
```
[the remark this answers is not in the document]
sarken: I _could_ be nice and fix this in every instance of create_log_item.

app/models/role.rb
```diff
@@ -3,7 +3,7 @@
 # "moderator" for an instance of a model (i.e., an object), a model class,
 # or without any specification at all.
 class Role < ApplicationRecord
-  has_many :roles_users
+  has_many :roles_users, dependent: :destroy
```
redsummernight, line 6: What requires this change btw?

app/models/roles_user.rb
```diff
@@ -1,4 +1,26 @@
 class RolesUser < ApplicationRecord
   belongs_to :user
   belongs_to :role
+
+  after_create :log_role_addition
+  def log_role_addition
+    admin = User.current_user
+    note = "Change made by #{admin&.login}"
+    user = User.find_by(id: user_id)
```
redsummernight, line 9: The model has `belongs_to :user` so you can skip the lookup here and simply refer to `user`.

app/models/role.rb
sarken: Nothing specific -- I thought it would be better to have it than not, since a) a missing `dependent: :destroy` on the user model side of roles_users was causing callback issues, and b) we were talking about possibly removing the translation_admin role, but I can definitely leave it for that instead.

redsummernight (approved):

PR #3980, opened 2021-03-02
AO3-5525 Add environment dependency to defaults:create_roles task

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5525

## Purpose

Fixes  the rake task 😬

### DIFF
--- a/lib/tasks/defaults.rake
+++ b/lib/tasks/defaults.rake
@@ -1,6 +1,6 @@
 namespace :defaults do
   desc "Create default roles by name"
-  task(:create_roles) do
+  task(create_roles: :environment) do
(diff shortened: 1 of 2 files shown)


redsummernight (commented):
Let's add a test for this. Can you add:
```rb
    it "depends on the :environment task" do
      expect(subject.all_prerequisite_tasks.map(&:name)).to include("environment")
    end
```
to the `included` block of [the example group](https://github.com/otwcode/otwarchive/blob/28782355f4693c15e8262b209ff4b574252aaed0/spec/support/task_example_group.rb#L23)?

redsummernight (approved):

PR #3982, opened 2021-03-06
AO3-3585 Reenable submit buttons after live validation errors are addressed

britnesissom:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-3585

## Purpose

Buttons that are disabled after submitting a form like creating a work or adding a comment are never re-enabled. This PR re-enables the button after 2 seconds with the original text.

## Testing Instructions

1. Create a work with no title and click "post" or "preview"
2. The post button now displays as "please wait..."
3. After 2 seconds, the button is clickable again with the text "post"

I also added a cucumber test but for some reason, it always says the xpath can't be found even though I tested the xpath in the firefox console and it worked. Maybe someone can help with that. Testing manually works though.

## Credit

agenderdanvers

they/them


sarken (changes requested):
Part of the reason the tests are failing might be that with our tests, you need to explicitly enable JavaScript using the [`@javascript` before hook](https://github.com/otwcode/otwarchive/blob/40fb60907f998e86424c097b4a4edfca1ca3e0f5/features/support/hooks.rb#L47-L50), like we do in [this bookmark test](https://github.com/otwcode/otwarchive/blob/40fb60907f998e86424c097b4a4edfca1ca3e0f5/features/bookmarks/bookmark_create.feature#L389).

That said, I don't think this is the correct approach to fixing this particular issue. I think we need to make sure the LiveValidation error is resolved before we reenable the submit button. If we don't, it's just going to do the "Please wait..." thing all over again without the user having any idea what's going on. (And if for some reason the button _isn't_ stuck in that state due to a LiveValidation error, but rather because the site or the user's connection is horrendously slow, we definitely don't want to reenable the button and let the user resubmit the form.)

I think the best thing to do is going to be scroll the user to the first field with the error, make sure it and all the other errors are resolved, and then reenable the button.

britnesissom:
@sarken that makes sense. Could you update the jira ticket story then or if you need me to I can? The reason I did it like this was because the ticket really only talked about how the submit button never changed from "Please wait" after clicking.

sarken:
@britnesissom Issue updated! (Unfortunately, older issues tend to be lacking specifics about the desired solutions, so it's something you may run into again.)

public/javascripts/application.js
```diff
@@ -362,6 +362,22 @@ jQuery.fn.preventDoubleSubmit = function() {
   });
 };
 
+function scrollToErrorIfFound() {
```
sarken, line 365: Since these two new functions are only used in the LiveValidation JavaScript file, can/should they be moved there?

public/javascripts/application.js
```diff
@@ -362,6 +362,22 @@ jQuery.fn.preventDoubleSubmit = function() {
...
+  var errorField = $j(".LV_invalid_field").first();
+  if (errorField.length !== 0) {
+    $j('html, body').animate({
```
sarken, line 368: Could you use double quotes here and on line 377? (I know this file and many others are super inconsistent about the quotation marks, but we've finally decided we like double. 😆 )

public/javascripts/livevalidation_standalone.js
```diff
@@ -314,17 +314,18 @@ LiveValidation.prototype = {
...
+        var isValid = this.doValidations();
+        if(isValid){
+          this.onValid();
+          enableSubmit();
+          return true;
+        } else {
+          this.onInvalid();
+          return false;
+        }
+      }else{
```
sarken, line 326: While you're here, can you fix the spacing on this line as well?

public/javascripts/livevalidation_standalone.js
```diff
@@ -314,17 +314,18 @@ LiveValidation.prototype = {
...
-			return true;
-		}else {
-			this.onInvalid();
-			return false;
-		}
-	  }else{
-      return true;
-    }
+        var isValid = this.doValidations();
+        if(isValid){
```
sarken, line 318: As long as this line is being edited anyway, can you update the spacing on this line to `if (isValid) {`?

sarken (commented):
Thank you so much for your patience! This looks good to me -- I tried it on my dev environment and it was so lovely. I just had a couple of minor style comments/suggestions.

public/javascripts/application.js
britnesissom: yeah the reason I didn't put them there was because the LiveValidation file is just classes/prototypes so it seemed weird to move them. I can if you think that's better though.

sarken: Eh, let's move them, then. I think it's better to have as much of the LiveValidation code as possible contained in the LiveValidation file, and it wasn't really our decision to just have classes and prototypes in that file -- that's just the way it came, and we're modifying it anyway.

sarken (approved):
Yay, thank you so much!

PR #3987, opened 2021-03-24
AO3-6150 Move user_ids & pseud_ids into child document for work search.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6150

## Purpose

Move a copy of `user_ids` and `pseud_ids` to a child document, and make the usual `user_ids` and `pseud_ids` nil when the work is unrevealed and/or anonymous. Also modifies bookmarkables to make the `pseud_ids` nil when the bookmarkable is anonymous/unrevealed.

Also adds two tasks to reindex all works/bookmarkables without recreating the index (so that most searches will continue to work during the deployment).

### DIFF
--- a/app/models/search/work_creator_indexer.rb
+++ b/app/models/search/work_creator_indexer.rb
@@ -1,0 +1,39 @@
+# A class for reindexing private work creator info (info that should not be
+# available during normal searches).
+class WorkCreatorIndexer < Indexer
+  def self.klass
+    "Work"
+  end
+
+  def self.mapping
+    WorkIndexer.mapping
+  end
+
+  def routing_info(id)
+    {
+      "_index" => index_name,
+      "_type" => document_type,
+      "_id" => document_id(id),
+      "routing" => parent_id(id, nil)
+    }
+  end
+
+  def document_id(id)
+    "#{id}-creator"
+  end
+
+  def parent_id(id, _object)
+    id
+  end
+
+  def document(object)
+    {
+      private_user_ids: object.user_ids,
+      private_pseud_ids: object.pseud_ids,
+      creator_join: {
+        name: :creator,
+        parent: object.id
+      }
+    }
+  end
+end
--- a/app/models/search/work_indexer.rb
+++ b/app/models/search/work_indexer.rb
@@ -103,7 +103,7 @@ class WorkIndexer < Indexer
     else
       {
         user_ids: work.user_ids,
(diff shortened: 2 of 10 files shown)


app/models/search/work_indexer.rb
```diff
@@ -94,8 +92,20 @@ def document(object)
...
     ).merge(
       language_id: object.language&.short,
-      series: series_data(object)
-    )
+      series: series_data(object),
+      creator_join: { name: :work }
+    ).merge(creator_document(object))
+  end
+
+  def creator_document(work)
```
redsummernight, line 100: Can you rename this to `creator_data` to match the series method?

app/models/search/work_creator_indexer.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  def document_id(id)
+    "#{id}-creator"
+  end
+
+  def parent_id(id, _object)
+    id
+  end
+
+  def document(object)
```
redsummernight, line 29: To check my understanding, every work should have this child document indexed, but the child document is only used when `viewing_own_collected_works_page?` / on works that belong to at least one collection?

lib/tasks/search.rake
```diff
@@ -3,55 +3,80 @@
...
+
+  desc "Reindex all works without recreating the index"
+  task(reindex_works: :environment) do
+    WorkIndexer.index_from_db
+    WorkCreatorIndexer.index_from_db
+  end
+
+  desc "Reindex all bookmarkables without recreating the index"
+  task(reindex_bookmarkables: :environment) do
+    BookmarkedExternalWorkIndexer.index_from_db
```
redsummernight, line 35: Should we add a task specifically for this issue, which requires reindexing works and series but not external works?

app/models/search/work_creator_indexer.rb
tickinginstant: Yeah, that's the idea.

lib/tasks/search.rake
tickinginstant: I can if you think it'd be a good idea, though I expect that the extra cost of reindexing all external works is probably pretty small relative to the cost of reindexing all works.

redsummernight: Makes sense, yeah. Let's just use this task.

redsummernight (approved):
Thank you!

PR #3990, opened 2021-03-31
AO3-5565 Prevent users from bookmarking work multiple times

korrien:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5565

## Purpose

~Code prevents users from bookmarking the same work for one pseud more than once.~
I am not sure that's the right approach though.

From the code design I assumed that different pseuds of the same user should still be able to bookmark same work.
However it is possible that bookmarks were designed to be connected to pseuds rather than to users for other reasons.

`index_bookmarkable_pseud` is also not made unique, I am unsure if it has to be, but rubocop throws `Rails/UniqueValidationWithoutIndex: Uniqueness validation should be with a unique index.`

Please let me know if this is wrong and the constraint somehow should be placed on user instead of the pseud, or if something must be done with the index.

-----
**Update**: 
Code prevents users from bookmarking the same work twice, regardless of used pseud. 
Database Index is not made unique.

## Testing Instructions

Before applying the code do:
1. Go to [root]/works/[id]/bookmarks/new
2. Create bookmark with pseudA - it should be created successuflly
3. Go to [root]/works/[id]/bookmarks/new again
4. Create second bookmark with pseudA or pseudB - it would be created successfully

After applying the code changes do:
1. Go to [root]/works/[id]/bookmarks/new
2. Create bookmark with pseudA or pseudB - error message should be visible
3. Update bookmarks created in the "Before" step - user should be able to do so, as well as reassign bookmarks from pseudB to pseudA and back.

## Credit

korrien, she/her


sarken:
Thanks for taking a look at this!

>Please let me know if this is wrong and the constraint somehow should be placed on user instead of the pseud, or if something must be done with the index.

The constraint should be on the user, yeah. The site isn't really set up in a way that encourages a user to bookmark a work with multiple pseuds, e.g. if you go to a work you've already bookmarked, you don't have an option to create a new bookmark, only to edit the existing one. My apologies for not making that clear in the issue!

I don't think we want to do anything with the index because, as far as I know, that would render existing duplicate bookmarks invalid.

korrien:
May I please request removal of the "Action Needed" label as I believe the code was updated? 😅 😇

sarken:
Of course! Sorry about that -- the email about the recent commits got buried in my inbox.

spec/controllers/bookmarks_controller_spec.rb
```diff
@@ -55,6 +55,92 @@ def it_redirects_to_user_login
...
+    context "when user is logged in" do
+      let(:user) { create(:user) }
+      let(:pseud) { user.default_pseud }
+
+      before do
+        fake_login_known_user(user)
+      end
+
+      context "when pseud doesn't belong to the user" do
+        let(:otheruser) { create(:user) }
```
sarken, line 77: Could you use an underscore to separate the words in variable names, e.g. `other_user`?

app/models/bookmark.rb
```diff
@@ -12,6 +12,17 @@ class Bookmark < ApplicationRecord
   validates_length_of :bookmarker_notes,
     maximum: ArchiveConfig.NOTES_MAX, too_long: ts("must be less than %{max} letters long.", max: ArchiveConfig.NOTES_MAX)
 
+  validate :not_already_bookmarked_by_user, on: :create
+  def not_already_bookmarked_by_user
+    return unless self.pseud && self.bookmarkable
+
+    bookmarks = Bookmark.join_bookmarkable.where(bookmarkable_id: self.bookmarkable_id)
```
sarken, line 19: I think there might be an issue here. Bookmarkables can be of different types of objects (Work, Series, ExternalWork), which means we could have as many as three bookmarkables with the same id. That could give us false positives, e.g. if the user has bookmarked a Series with the id 123 and is now trying to bookmark a Work with the id 123, they might be told they have already bookmarked the item.

I'm also not sure if it's necessary to look at all of the bookmarks for a given bookmarkable. Would it instead make sense to do something like `pseud.user.bookmarks.where(bookmarkable: bookmarkable)` to look through the existing bookmarks for the user who owns the pseud? Or is the join just more efficient?

korrien: I have no idea how to check efficiency of that code, but I highly doubt that join is more efficient in that case.
From the search size perspective it should make no difference if we go users-who-bookmarked-the-work or bookmarks-which-belong-to-user.
Changed.

-----------
This comment seems like a good occasion to ask for a bit of help :)

I struggled to find a way how to reach bookmarks-made-by-user. In the end I found similar code in [Gift](https://github.com/otwcode/otwarchive/blob/master/app/models/gift.rb#L31)  and copied it. Apparently it was not the best decision 😄 
I also was sure that as Bookmarkable is the parent "import" class (can't find what's is called properly in Ruby) of bookmarkable objects, the ids must be unique.

So I clearly misunderstand a lot :)

My problem is that I do not know how to get the full picture of classes, attributes, methods and relationships in the model:
- Visualizing the schema with rails-erd wasn't much help as result is either too simple or too convoluted
- I can't find any auto-docs
- Searching schema.rb clearly didn't help me see that there is a bookmarks attribute found on User :D 
- Simply reading through model classes doesn't help me much because of implicit attributes and relationships
- Ruby extension for vscode doesn't seem to be of help for that problem   

So I wanted to ask if there is any obvious way to create a full picture of the model or if it just must come from experience? :)

sarken: >In the end I found similar code in Gift and copied it. Apparently it was not the best decision

That's a great way to approach a problem if you don't know what else to do, imo! It won't always give you the best solution, but it will at least give you a starting point, and then someone can help you refine it during code review.

>So I wanted to ask if there is any obvious way to create a full picture of the model or if it just must come from experience? :)

I will poke some folks and see if they have practical advice for you, since I do a lot of relying on familiarity with the code and site functionality. (For questions like "How do I get all of a user's bookmarks?", it might be a good idea to read over the [Active Record Associations docs](https://guides.rubyonrails.org/association_basics.html) if you haven't already, particularly the `has_one` and `has_many` sections.)

sarken (changes requested):
Just two little things (minor improvements, not actual problems) and then I think this is ready!

app/controllers/bookmarks_controller.rb
```diff
@@ -208,7 +208,7 @@ def update
     new_collections = []
     unapproved_collections = []
     errors = []
-    bookmark_params[:collection_names].split(',').map {|name| name.strip}.uniq.each do |collection_name|
+    bookmark_params[:collection_names] && bookmark_params[:collection_names].split(",").map(&:strip).uniq.each do |collection_name|
```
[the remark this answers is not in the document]
korrien: Huh. Thanks for letting me know about the concept.
Looks like lazy evaluation is not the preferable Ruby way to deal with that situation, as I can't find anything similar.
I can't say that I find four "&" more readable, but at least this is used in many other places in the project :D
Using "unless" at any point seems even less readable to me, so changed.

sarken (commented):
Yay, thank you!

The only reason I'm not marking this approved/ready to merge is there's a merge conflict that has developed. You can resolve that whenever you have time, or if you want, we can deal with it when we're actually ready to merge (I don't think it'll be in this release, but maybe the next one).

sarken (approved):

app/controllers/bookmarks_controller.rb
```diff
@@ -253,7 +253,7 @@ def update
 
     if @bookmark.update(bookmark_params) && errors.empty?
       flash[:notice] ||= ""
-      flash[:notice] = ts(" Bookmark was successfully updated. ").html_safe + flash[:notice]
+      flash[:notice] = ts("Bookmark was successfully updated.").html_safe + flash[:notice]
```
tickinginstant, line 256: Is stripping the trailing space here going to make the notice run into the collection notice? It looks like the `unapproved_collections` notice has a space at the front, but the `new_collections` notice doesn't.

sarken: I was going to say it's okay, but I just double checked and the test I looked at is actually for using the "Add to Collections" button on the bookmark (which leaves off the "Bookmark was successfully updated" part), not the "Edit" button, so we should probably put it back.

PR #3993, opened 2021-04-14
AO3-5964 Replace excess spaces in some tags

CristinaRO:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5964

## Purpose

What does this PR do?

This is a one-time After task to:
- Identify tags that have extraneous spaces in their name
- Replace each space in their name with an underscore (`_`)
- Output the list of tags with their ID, old name, and new name

We have to use `update_column` because only admins are allowed to rename tags the regular way.

The way these tags have sneaked through our validation process is most likely the fact that `squish` used to use a more lax regular expression to compress blank spaces (`\s`) than the one it currently uses (`[[:space]]`). (Thank you @tickinginstant for explaining the issue!)

## Testing Instructions

## References

This issue is part of the prerequisites for AO3-6153.

## Credit

Enigel

she/her

### DIFF
--- a/lib/tasks/after_tasks.rake
+++ b/lib/tasks/after_tasks.rake
@@ -740,6 +740,34 @@ namespace :After do
     WorkIndexer.create_mapping
   end
 
+  desc "Fix tags with extra spaces"
+  task(fix_tags_with_extra_spaces: :environment) do
+    total_tags = Tag.count
+    total_batches = (total_tags + 999) / 1000
+    puts "Inspecting #{total_tags} tags in #{total_batches} batches"
+
+    report_string = "Tag ID,Old tag name,New tag name\n"
+    Tag.find_in_batches.with_index do |batch, index|
+      batch_number = index + 1
+      progress_msg = "Batch #{batch_number} of #{total_batches} complete"
+
+      batch.each do |tag|
+        next unless tag.name != tag.name.squish
+
+        old_tag_name = tag.name
+        new_tag_name = old_tag_name.gsub(/[[:space:]]/, "_")
+
+        tag.update_column(:name, new_tag_name)
+
+        report_row = [tag.id, old_tag_name, new_tag_name].join(",") + "\n"
+        report_string += report_row
+      end
+
+      puts(progress_msg) && STDOUT.flush
+    end
+    puts(report_string) && STDOUT.flush
+  end
+
   # This is the end that you have to put new tasks above.
 end
 
--- a/spec/lib/tasks/after_tasks.rake_spec.rb
+++ b/spec/lib/tasks/after_tasks.rake_spec.rb
@@ -153,3 +153,15 @@ describe "rake After:replace_dewplayer_embeds" do
     end.to output("Couldn't convert 1 chapter(s): #{dewplayer_work.first_chapter.id}\nConverted 0 chapter(s).\n").to_stdout
   end
 end
+
+describe "rake After:fix_tags_with_extra_spaces" do
+  let(:borked_tag) { Freeform.create(name: "whatever") }
+
+  it "replaces the spaces with the same number of underscores" do
+    borked_tag.update_column(:name, "\u00A0\u2002\u2003\u202F\u205FBorked\u00A0\u2002\u2003\u202Ftag\u00A0\u2002\u2003\u202F\u205F")
+    subject.invoke
+
+    borked_tag.reload
+    expect(borked_tag.name).to eql("_____Borked____tag_____")
+  end
+end


lib/tasks/after_tasks.rake
```diff
@@ -740,6 +740,39 @@ namespace :After do
...
+    report_string = "Tag ID,Old tag name,New tag name\n"
+    Tag.find_in_batches.with_index do |batch, index|
+      batch_number = index + 1
+      progress_msg = "Batch #{batch_number} of #{total_batches} complete"
+
+      batch.each do |tag|
+        if tag.name != tag.name.squish
+          old_tag_name = tag.name
+
+          leading_spaces = old_tag_name.match(/^(\s+)/)
```
sarken, line 758: Since the issue mentioned looking at how [`squish`](https://apidock.com/rails/v5.2.3/String/squish%21) detects spaces... is `\s` exactly the same as `[[:space:]]`?

I honestly do not know the answer and I found the use of "etc" in the [Regexp class documentation](https://ruby-doc.org/core-3.0.1/Regexp.html#class-Regexp-label-Character+Classes) entirely unhelpful. I'm guessing yes based on [this page on POSIX bracket expressions](https://www.regular-expressions.info/posixbrackets.html), but I'd rather ask a silly question that assume wrong.

CristinaRO: I was wondering *why* the issue mentioned that. (I still don't know.)

The issue also only mentions "leading and trailing" spaces, but not any accidentally duplicated spaces in the middle of the tag, which `squish` also compresses.

Do we in fact want to squish the middle spaces as well? Or is using squish just a heuristic to narrow down which tags might have blanks in the front and in the back, even if it might catch tags that only have extra spaces in the middle, but that we don't want to mess with?

tickinginstant: They are not the same:
```
2.6.5 :038 > "\u00A0\u2002\u2003\u202F\u205F".gsub(/\s/, "_")
 => "     " 
2.6.5 :039 > "\u00A0\u2002\u2003\u202F\u205F".gsub(/[[:space:]]/, "_")
 => "_____" 
```
The difference in their behavior for various unicode space characters is what caused the issue in the first place, I believe. The tag code has been [squishing tag names before validation since 2008](https://github.com/otwcode/otwarchive/blob/b63fce67a3975ec0614bb535c75d76c8dafbda71/app/models/tag.rb#L210-L213), but with the update to Rails 4, the behavior of `squish` changed (because it switched from `\s` to `[[:space:]]`). That means that there may be some legacy tags from before the Rails upgrade that have `tag.name != tag.name.squish`, since various space characters (like the non-breaking space) didn't used to be treated as space characters when calling `squish`.

And yeah, it should also squish sequences of `[[:space:]]` characters in the middle of the name (or replace them with underscores as needed). All tags need to have `tag.name == tag.name.squish` to prevent issues in the future.

sarken (commented):
This looks good to me, but I'll leave final approval for @tickinginstant, who knows more about `[[:space:]]` 🪐  than me.

lib/tasks/after_tasks.rake
```diff
@@ -740,6 +740,34 @@ namespace :After do
...
+      batch_number = index + 1
+      progress_msg = "Batch #{batch_number} of #{total_batches} complete"
+
+      batch.each do |tag|
+        next unless tag.name != tag.name.squish
+
+        old_tag_name = tag.name
+        new_tag_name = old_tag_name.gsub(/[[:space:]]/, "_")
+
+        tag.update_column(:name, new_tag_name)
```
tickinginstant, line 760: What happens if there's already a version of the tag with the underscores in it?

Also, `update_column` skips callbacks like reindexing and cache expiration. Maybe it'd be good to use `update_attribute` instead? I think that skips validations while still performing callbacks.

lib/tasks/after_tasks.rake
```diff
@@ -740,6 +740,34 @@ namespace :After do
...
+
+        report_row = [tag.id, old_tag_name, new_tag_name].join(",") + "\n"
```
tickinginstant, line 762: Does this work for tags with quotes in them? I know quotation marks are a special character for some CSV implementations.

tickinginstant (commented):
I have some concerns about other parts of the task, but the handling of `[[:space:]]` looks good now!

lib/tasks/after_tasks.rake
tlee911: This will escape double quotes as per ruby standard by prefixing with a backslash: `\"`

CSV handling may require double double-quotes (depending on the program), but is that something we want to do in the task here, or just leave it up to find and replace for whoever ends up needing to open the CSV?

lib/tasks/after_tasks.rake
tlee911: The task will fail if there is an existing tag already with the underscores in it the name clashes.  Rails is skipping validations but there is still a unique constraint in the DB table and it will fail to save the tag.

Not sure what's the desired behaviour in this case, abort?  Skip and log?  Keep adding underscores (or some other prefix/suffix) to make it unique?

tlee911:
I have a branch where I've taken the changes from @CristinaRO, pulled latest master, and added a couple test cases to cover the questions brought up by @tickinginstant:
https://github.com/otwcode/otwarchive/compare/master...tlee911:AO3-5964/strip-tags

lib/tasks/after_tasks.rake
sarken: Let's keep adding underscores, yeah.

lib/tasks/after_tasks.rake
sarken: Since it's program-dependent, I think we can leave it for whoever needs to open the file.

tickinginstant: I just tested it, and it doesn't escape quotes.

`to_csv` does the right thing and would save the people downstream some trouble.

tlee911: You're right of course.  I was doing string matching with stdout and didn't double escape.

sarken:
The work continues in #4020!

PR #3998, opened 2021-04-25
AO3-6151 Use aws-sdk-s3 instead of aws-sdk

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6151

## Purpose

We only use S3 (via [kt-paperclip](https://github.com/thoughtbot/paperclip/commit/572bba0ec2a8b3b9f56d845a2cebecff34b5a11b) for hosting icons), we don't need the rest of the AWS services.

## Testing Instructions

See issue.


sarken (approved):
It's so beautiful 😭 I'd like to put this in this release, but we should probably wait to merge until we test that the kt-paperclip change works, just so we don't have to guess which of the two changes caused any issues we might encounter.

PR #4000, opened 2021-04-28
AO3-5910 Update schema and structure files

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5910

## Purpose

Generate new versions of schema dumps from production.

For Ruby dump, run on an app server:
```sh
bundle exec rake db:schema:dump
```
then manually change the version in `ActiveRecord::Schema.define` to the latest migration timestamp.

For SQL dump, run on a database server:
```sh
mysqldump --no-data --routines --skip-comments --result-file=structure.sql otwarchive_production
```
```
Warning: A partial dump from a server that has GTIDs will by default include the GTIDs of all transactions, even those that changed suppressed parts of the database. If you don't want to restore GTIDs, pass --set-gtid-purged=OFF. To make a complete dump, pass --all-databases --triggers --routines --events.
```
then manually:
- remove [AUTO_INCREMENT values](https://stackoverflow.com/questions/2210719/out-of-sync-auto-increment-values-in-development-structure-sql-from-rails-mysql);
- append a dump of the schema_migrations table, generated from `bundle exec rails r "puts ActiveRecord::Base.connection.dump_schema_information"`.

The resulting output will be similar to that of `bundle exec rake db:structure:dump` (which we can't use because we've started provisioning app and database servers differently, and app servers [no longer have mysqldump available](https://otwarchive.atlassian.net/browse/AO3-5928)).

## Testing Instructions

CI build should pass.

After this is merged:
- Check the existing pull requests for migrations and [bump their timestamps to later](https://stackoverflow.com/a/4047884).
- Update https://github.com/otwcode/otwarchive/wiki/Migrations#updating-schema-dumps.


db/structure.sql
```diff
@@ -2,13 +2,29 @@
 /*!40101 SET @OLD_CHARACTER_SET_CLIENT=@@CHARACTER_SET_CLIENT */;
 /*!40101 SET @OLD_CHARACTER_SET_RESULTS=@@CHARACTER_SET_RESULTS */;
 /*!40101 SET @OLD_COLLATION_CONNECTION=@@COLLATION_CONNECTION */;
-/*!40101 SET NAMES utf8mb4 */;
+/*!40101 SET NAMES utf8 */;
```
sarken, line 5: This is from production so it must be right, but... is this right?

redsummernight: I think this is just the overall default. Most tables have explicit overrides (`CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci`) which haven't changed.

sarken (approved):

sarken:
For the wiki update, do you want me to just replace the entire "Updating schema dumps" section with what you've written under "Purpose," or are there bits on the wiki we should keep?

redsummernight:
Yeah, and I've updated the wiki.

PR #4001, opened 2021-05-07
AO3-6065 Add task to set rating to "Not Rated" if missing

tlee911:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6065

## Purpose

Adds a rake task `rake After:add_default_rating_to_works` which will find works missing a rating and set them to the canonical "Not Rated" instead.

## Testing Instructions

- Find some works which have no rating (or delete the ratings yourself with DB access)
- `rake After:add_default_rating_to_works`
- Verify the works in question are now tagged "Not Rated"
- Verify works which had valid ratings were not changed

Automated tests have also been added in `after_tasks.rake_spec.rb`


lib/tasks/work_tasks.rake
```diff
@@ -25,4 +25,14 @@ namespace :work do
     end
   end
 
+  # AO3-6065
+  desc "Add 'Not Rated' to works missing a rating"
+  task(:not_rated => :environment) do
+    Work.find_each do |work|
+      if work.rating_string.blank?
```
sarken, line 32: We could probably avoid tucking everything inside an `if...end` by doing `next unless work.rating_string.blank?` or `next if work.rating_string.present?` instead.

lib/tasks/work_tasks.rake
```diff
@@ -25,4 +25,14 @@ namespace :work do
     end
   end
 
+  # AO3-6065
+  desc "Add 'Not Rated' to works missing a rating"
```
sarken, line 29: Since this is intended to be a one-time task, it would be more at home in `lib/tasks/after_tasks.rake`.

It also needs some tests, which can be added to `spec/lib/tasks/after_tasks.rake_spec.rb`.

lib/tasks/work_tasks.rake
```diff
@@ -25,4 +25,14 @@ namespace :work do
...
+  task(:not_rated => :environment) do
+    Work.find_each do |work|
```
sarken, line 31: Since this is going to take a while, can we use `find_in_batches` and then print a progress indicator to `STDOUT` so we can see how things are progressing? (There are a bunch of examples of possible approaches in the other recent after tasks.)

lib/tasks/work_tasks.rake
```diff
@@ -25,4 +25,14 @@ namespace :work do
     end
   end
 
+  # AO3-6065
```
sarken, line 28: We'll be able to check the history or blame if we ever need to know what issue this task was for (as opposed to what it does, which is covered in the description), so this can be left off.

sarken (changes requested):

tlee911:
I spent all morning on the test code, and I still can't figure out what's wrong with it 😕, so help would be appreciated!

tlee911:
@tickinginstant helped me figure it out.  Turns out not all the ratings are available to tests by default and I happened to pick one (mature) that didn't exist.  I've updated spec_helper.rb to have all the usual ratings so this doesn't surprise anyone else in the future.

sarken (changes requested):
We were just talking about potentially making some more changes to this due to [AO3-5988](https://otwarchive.atlassian.net/browse/AO3-5988), so no rush on anything I've mentioned here!

spec/spec_helper.rb
```diff
@@ -51,10 +51,12 @@
     Category.find_or_create_by_name(ArchiveConfig.CATEGORY_SLASH_TAG_NAME).update(canonical: true)
     Rating.find_or_create_by_name(ArchiveConfig.RATING_DEFAULT_TAG_NAME).update(canonical: true)
     Rating.find_or_create_by_name(ArchiveConfig.RATING_EXPLICIT_TAG_NAME).update(canonical: true)
+    Rating.find_or_create_by_name(ArchiveConfig.RATING_MATURE_TAG_NAME).update(canonical: true)
```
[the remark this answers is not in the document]
tlee911: Ok, I can back out this change.  I did end up switching to explicit (was trying to use mature before) which does exist in tests so this is not strictly necessary for my tests to pass.

tlee911:
After looking at #4003, I realize I'm manipulating ratings via `ratings_string` rather than the `ratings` array.  While both seem to work, is there a preference for using `ratings` directly?  That feels more right to me.

sarken:
Manipulating the array directly sounds good to me!

sarken (approved):
I think that's everything now -- thank you!

redsummernight:
This has conflicts now, but let's wait until #4022 is merged first (since that'll also introduce conflicts).

redsummernight:
This has conflicts now, but we can merge it next. You should also pull from latest master to fix the dependency security warnings.

PR #4006, opened 2021-05-17
AO3-6109 Don't pull English from Phrase and do update translations on push

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6109

## Purpose

* Stop uploading default Rails i18n files to Phrase and add a note to the README that says not to edit them
* Stop exporting English locale files from Phrase
* Start exporting translations from Phrase to their own directory, which contains a README that says not to edit them
* Start updating existing English on Phrase when uploading
 
~I'm not sure what we want to do with `views/zh-CN.yml`, which is the Chinese text for the proxy notice, where we display both English and Chinese. Perhaps blacklist `layouts.proxy_notice*` keys in Phrase until we have localization enabled for the interface and no longer need to give it special treatment? If we just want to exclude the file like we've done with the `defaults` directory, that should work, too -- I can change all the `<locale_name>.yml` to `en.yml`.~ We'll just push the `en.yml` files and then we can remove `views/zh-CN.yml` after the first exports are done.

## Testing Instructions

When this is merged, we'll want to export from Phrase via GitHub Sync to get all of the existing translations. Once that is merged, we'll want to import to Phrase to make sure all of the English is correct -- search for "part12" and make sure it doesn't include the term "whitelist" to verify.

More instructions on Jira.


redsummernight (commented):
> I'm not sure what we want to do with `views/zh-CN.yml`, which is the Chinese text for the proxy notice, where we display both English and Chinese. Perhaps blacklist `layouts.proxy_notice*` keys in Phrase until we have localization enabled for the interface and no longer need to give it special treatment?

It's possible to [exclude those keys](https://developers.phrase.com/api/#patch-/projects/{project_id}/keys/exclude), but we need to do it manually (either through Phrase API or website), so it's more trouble than its worth. I prefer having everything in .phrase.yml and nothing else to keep track of.

> If we just want to exclude the file like we've done with the `defaults` directory, that should work, too -- I can change all the `<locale_name>.yml` to `en.yml`.

Let's do this.

redsummernight (approved):

PR #4009, opened 2021-05-26
AO3-2380 Make autocomplete for tags ignore diacritics

tlee911:
# Pull Request Checklist

## Issue

## Purpose

Changes autocomplete functionality so that diacritics are ignored when searching for autocomplete possibilities (_i.e._ accents are ignored when searching, so "e" matches "é" and vice-versa).

## Testing Instructions

- Create (canonical) tags of various types both with and without accented characters (_e.g._ "àbc" and "apple")
- Verify that autocomplete for either the plain character "a" or the accented one "à" (or any accented variation "ä") will suggest both tags as possibilities

**Note** This functionality is accomplished by storing in redis autocomplete strings stripped of diacritics (in addition to the lower casing that was already happening to make it case insensitive) via `ActiveSupport::Inflector.transliterate(string)` and then similarly stripping diacritics from input when invoking the autocomplete.  You will need to run the rake task `autocomplete:reload_tag_data` in order to get this functionality for existing tags.

## References

https://api.rubyonrails.org/v5.2.5/classes/ActiveSupport/Inflector.html#method-i-transliterate


tickinginstant:
The docs for `transliterate` say:

> Default approximations are provided for Western/Latin characters, e.g, “ø”, “ñ”, “é”, “ß”, etc.

What happens with non-Western characters? Does it affect the autocomplete code?

sarken:
fwiw, I looked at the docs a few weeks ago and my takeaway was that you _can_ use it for non-Western characters, but it's not set up to do that by default. I assumed that based on the comment about Russian:

>This method is I18n aware, so you can set up custom approximations for a locale. This can be useful, for example, to transliterate German's “ü” and “ö” to “ue” and “oe”, **or to add support for transliterating Russian to ASCII**.

However, it would be good to be sure.

tickinginstant:
Yeah, it sounds like they're not covered by default. But I don't know whether "not covered" means "the autocomplete for non-Western characters will continue to work the way it used to," or "the autocomplete for non-Western characters will fail in new and interesting ways."

tlee911:
I did test with Chinese characters during development to make sure it didn't mess up anything there, though I have to admit it's not something I use regularly on the real site, so I'm not that familiar with the expected behaviour.

Do we want a test case for that?

sarken:
Yes, please! If you need any specific info about current behavior, you can let me know what questions you have and I'll pass them along to Tag Wrangling.

tlee911:
Ok, so I did mess up the handling of non Latin characters in tags (putting this PR in draft because of this).  Transliterate will turn characters it can't handle into "?", and my test worked because I created only one tag with Chinese characters, the autocomplete lookup matched the question marks.

If I try to make two Chinese language tags, they both get transliterated into question marks and step all over each other in autocomplete.  I think we might need to do some sort of check to ensure that doesn't happen.

Regex like `\p{InBasic_Latin}|\p{InLatin-1_Supplement}` doesn't work because there are characters even in that set that cannot be transliterated (see https://en.wikipedia.org/wiki/Latin-1_Supplement_(Unicode_block) for things like ¬ or ¥) and we definitely don't want to resort to maintaining our own mapping.

I might instead try:
1. Strip all actual question marks from the tag and try transliteration to see if question marks appear
2. Transliterate the name but skip characters which would become question marks

tickinginstant:
Glad that it's working for non-Western characters now!

I did a little digging around and found [the default mapping](https://github.com/ruby-i18n/i18n/blob/95612def98cb9c1d5be660aa77062c9be0b0430c/lib/i18n/backend/transliterator.rb#L43-L72). It might be good to check with the tag wranglers to make sure that all of the necessary mappings are there -- I suspect that there may be some commonly used letter/mark combos that the mapping is missing.

It also occurred to me that it may be difficult to use mappings alone if autocomplete needs to be able to handle two diacritical marks on the same letter, or even [something more extreme](https://en.wikipedia.org/wiki/Zalgo_text). I'm not sure whether that would be necessary to handle for the canonical tag autocomplete (maybe something else to ask the tag wranglers?), but it might be a concern for the pseud autocomplete, since that's controlled by the users.

(BTW, I'm pretty sure that's a real test failure. It's caused by an issue with JS-based tests and redirects like [this one](https://github.com/otwcode/otwarchive/blob/c2001360b865dd23a73d5c5827a8a681e1eb3afb/app/controllers/pseuds_controller.rb#L92). You may need to either change the redirect to use `user_pseud_path`, or change the `I am logged out` step to an `I start a new session` step.)

sarken:
Checking with Tag Wrangling! I have a feeling the multiple diacritics use case will only be an issue for pseuds, not tags, but that still means we'll have to think of what to do about it.

tlee911:
> Glad that it's working for non-Western characters now!
> 
> I did a little digging around and found [the default mapping](https://github.com/ruby-i18n/i18n/blob/95612def98cb9c1d5be660aa77062c9be0b0430c/lib/i18n/backend/transliterator.rb#L43-L72). It might be good to check with the tag wranglers to make sure that all of the necessary mappings are there -- I suspect that there may be some commonly used letter/mark combos that the mapping is missing.
> 
> It also occurred to me that it may be difficult to use mappings alone if autocomplete needs to be able to handle two diacritical marks on the same letter, or even [something more extreme](https://en.wikipedia.org/wiki/Zalgo_text). I'm not sure whether that would be necessary to handle for the canonical tag autocomplete (maybe something else to ask the tag wranglers?), but it might be a concern for the pseud autocomplete, since that's controlled by the users.

Considering we don't to anything with diacritical marks currently as far as autocomplete is concerned (they are treated as entirely separate characters), and that's Tag Wrangling's main complaint, I think this is still better than what we have now.  As long as we don't paint ourselves into a corner with the implementation and I haven't introduced any bugs, this should cover what I assume are the more common accented characters from European languages.  If we're missing some characters and want to go down the path of maintaining our own mapping, those are essentially config values, and can be added in later.

As for someone putting in Zalgo or the like into tags/pseuds, I can't see them thinking that would be handled by autocomplete.

> (BTW, I'm pretty sure that's a real test failure. It's caused by an issue with JS-based tests and redirects like [this one](https://github.com/otwcode/otwarchive/blob/c2001360b865dd23a73d5c5827a8a681e1eb3afb/app/controllers/pseuds_controller.rb#L92). You may need to either change the redirect to use `user_pseud_path`, or change the `I am logged out` step to an `I start a new session` step.)

To note, that test works locally for me, but I'll try pushing a change where I switch the step to `start a new session`.  That seems to still work, and less risky than modifying the pseud controller (as far as the scope of this PR is concerned).

sarken:
We think Vietnamese uses multiple diacritics on a single character. Translation is checking if there's more, and they also were wondering if double-mapping in the other direction is possible (as in typing either "ue" or "u" will turn up results with "ü" -- we know we can have typing "a" turn up results for both "à" and "ä").

tlee911:
> We think Vietnamese uses multiple diacritics on a single character. Translation is checking if there's more, and they also were wondering if double-mapping in the other direction is possible (as in typing either "ue" or "u" will turn up results with "ü" -- we know we can have typing "a" turn up results for both "à" and "ä").

I'll need to test the double diacritics to see how that behaves.  I suspect those will fail to transliterate because they don't exist in the default mapping.

As far as multi character transliterations go, they will work both ways.  The redis cache contains the transliterated value, so in your example we would have "ue" in the autocomplete, so that can be used to search also.

sarken:
Oh, sorry, I didn't mean double characters like "typing oe will turn up results with œ", I meant like "typing u or typing o will both turn up results with ū."

tlee911:
I think I wasn't clear either.

If "ū" transliterates to "oe" then that's what gets stored in redis for autocomplete matching.  Inputs will also be transliterated for search, so in this case typing "u" would not match for "ū" and typing "o" (or "oe") would.  That might be counter-intuitive to people if we have transliterations like that, but I think the remedy in that case will need to involve maintaining our own mappings.

sarken:
Yeah, I've been assuming from the start we'll be using our own mappings, since the default list isn't particularly comprehensive. So if we have this mapping:

```
  ū: "u"
  ū: "o"
```

1. What will the tag "jūs" be stored as in Redis?
2. Will "jūs" appear in the autocomplete suggestions list if someone types
a. "jūs"?
b. "jus"?
c. "jos"?

tlee911:
So we can't have 2 for the same character. In this case it would have to be "u" or "o", pick one.

For example, let's say we use "u", autocomplete will return the tag if user enters either "jūs" or "jus" and redis cache will have "jus".

sarken:
Ah, thanks, that's what we wanted to know! I've passed that along to Translation.

So next question: what's the advantage to using this method instead of the normalization method we use to compare edits to tags?

https://github.com/otwcode/otwarchive/blob/b63fce67a3975ec0614bb535c75d76c8dafbda71/app/models/tag.rb#L193

It seems like that probably covers a more comprehensive list of characters. To avoid inconsistency, would it make sense to use the same method (either `mb_char.normalize(:kd).gsub(/[^\x00-\x7F]/u,'').downcase` or `ActiveSupport::Inflector.transliterate`) in both places?

tlee911:
> So next question: what's the advantage to using this method instead of the normalization method we use to compare edits to tags?

Normalization serves a different purpose and does not remove diacritics.   Normalization handles unicode equivalence by turning different sequences for the same character/string into a standardized one to allow comparisons (see https://en.wikipedia.org/wiki/Unicode_equivalence).

`
a = "Àutobot"
a.mb_chars.normalize.to_s #Àutobot
ActiveSupport::Inflector.transliterate(a) #Autobot
a.mb_chars.normalize.to_s ==  ActiveSupport::Inflector.transliterate(a) #false
`

tickinginstant:
The particular line of code that @sarken linked to actually does strip diacritics, because it normalizes into a decomposed form (hence the `d` in `:kd`), which converts e.g. `á` into `a` + `´`, and then uses the `gsub(/[^\x00-\x7F]/u,'')` regex to remove all non-ASCII characters (including the diacritics).

Unfortunately, it doesn't just strip diacritics -- if you have a tag full of e.g. Chinese characters, you can rename it to a completely different sequence of Chinese characters without being an admin. That line of code should probably be modified to use a regex more like `/[\u0300-\u036F]/` (the [combining diacritical marks](https://en.wikipedia.org/wiki/Combining_Diacritical_Marks) unicode block), or maybe something with [a few more unicode blocks](https://en.wikipedia.org/wiki/Combining_character).

The downside of that method is that it doesn't cover characters like `ø`, because that's not a diacritical mark. But the upside is that it can handle arbitrary numbers of diacritical marks out of the box:
```
2.7.3 :010 > "Àutobot".mb_chars.normalize(:kd).gsub(/[\u0300-\u036F]/, "").to_s
 => "Autobot" 
2.7.3 :011 > "Tiếng Việt".mb_chars.normalize(:kd).gsub(/[\u0300-\u036F]/, "").to_s
 => "Tieng Viet" 
2.7.3 :012 > "z̵̺͍̓̓͌͜a̴̝̞͇̐̕̕l̸̪̝̝̾̚͝g̵̪̠̞͆͊o̴͎̻͙͆͛̓".mb_chars.normalize(:kd).gsub(/[\u0300-\u036F]/, "").to_s
 => "zalgo" 
```

tlee911:
> The particular line of code that @sarken linked to actually does strip diacritics, because it normalizes into a decomposed form (hence the `d` in `:kd`), which converts e.g. `á` into `a` + `´`, and then uses the `gsub(/[^\x00-\x7F]/u,'')` regex to remove all non-ASCII characters (including the diacritics).

I see.  I think I get it now.

> Unfortunately, it doesn't just strip diacritics -- if you have a tag full of e.g. Chinese characters, you can rename it to a completely different sequence of Chinese characters without being an admin. That line of code should probably be modified to use a regex more like `/[\u0300-\u036F]/` (the [combining diacritical marks](https://en.wikipedia.org/wiki/Combining_Diacritical_Marks) unicode block), or maybe something with [a few more unicode blocks](https://en.wikipedia.org/wiki/Combining_character).
> 
> The downside of that method is that it doesn't cover characters like `ø`, because that's not a diacritical mark. But the upside is that it can handle arbitrary numbers of diacritical marks out of the box:
> 
> ```
> 2.7.3 :010 > "Àutobot".mb_chars.normalize(:kd).gsub(/[\u0300-\u036F]/, "").to_s
>  => "Autobot" 
> 2.7.3 :011 > "Tiếng Việt".mb_chars.normalize(:kd).gsub(/[\u0300-\u036F]/, "").to_s
>  => "Tieng Viet" 
> 2.7.3 :012 > "z̵̺͍̓̓͌͜a̴̝̞͇̐̕̕l̸̪̝̝̾̚͝g̵̪̠̞͆͊o̴͎̻͙͆͛̓".mb_chars.normalize(:kd).gsub(/[\u0300-\u036F]/, "").to_s
>  => "zalgo" 
> ```

So which is the preferred solution then?  Do we want to stick with normalize and gsub (as opposed to ActiveSupport::Inflector.transliterate)?  We would have to modify the regex as pointed out here, otherwise autocomplete won't work for things like Asian language characters.

At least the transliteration is done by a single function in the autocomplete module for this PR so it's easy to change the inner workings.  If tag editing also uses the same transliteration method, it might be worth extracting it further so it can be invoked there as well without needing to duplicate the code.

sarken:
I'm talking to Translation and Tag Wrangling to get their input. 

I personally lean slightly toward using the normalize/gsub approach (with the Asian characters fix) because of its ability to handle multiple diacritics, but if we did that, would we be able to use either gsub or transliterate to also cover characters like `ø`? It would probably be slightly annoying to figure out all the characters that were no longer covered, but relying solely on transliterate would involve the same sort of work without the ability to easily handle multiple diacritics.

tlee911:
As a thought, couldn't we do both?

1. normalize.gsub to strip all diacritics and handle multiples on a single character
2. transliterate to handle special characters (like `ø` or `ç`)

Which order to do it in is debatable (I haven't noodled through that enough yet).

sarken:
Right, that's what I meant: normalize and then follow it up with transliterate or more gsub. (Or the other way around is equally fine if necessary.)

tlee911:
I've done it so that we use normalize/gsub first and then transliterate since transliterate does not handle characters with multiple diacritics and will replace them with `?`.

Note that I did not modify the regex for the tag editing functionality.  I think that's outside the scope of this issue, and probably deserves it's own JIRA/PR.

lib/autocomplete_source.rb
```diff
@@ -11,21 +11,33 @@ module AutocompleteSource
...
 
+  def transliterate(input)
+    input = input.to_s.mb_chars.normalize(:kd).gsub(/[\u0300-\u036F]/, "")
+    result = ""
+    input.each_char do |char|
+      tl = ActiveSupport::Inflector.transliterate(char)
+      # Don't transliterate characters which are not supported (e.g. non-Latin characters).
+      # No need for special handling of "?" since it is not modified by transliteration
+      result << (tl == "?" ? char : tl)
+    end
```
sarken, line 22: I think this might be clearer if the code were less concise, particularly since the character we're dealing with is a question mark, which is used in the ternary operator. Maybe something like:

```ruby
if tl == "?"
  # If transliterate returns "?", the original character is either unsupported (e.g. a non-Latin character) or was actually a question mark. In both cases, we should keep the original.
  result << char
else
  result << tl
end
```

sarken (approved):
Looks good to me now -- thank you!

PR #4013, opened 2021-06-02
AO3-5408 Stop storing IP addresses for support requests in the DB

tlee911:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5408

## Purpose

New support requests will no longer store users' IP addresses in the database.  IP will still be sent to Akismet for spam detection, but address is not saved (will be null).

Also adds a DB migration which will drop the column in question.  I've done this in a way that is "reversible" in rails so the column can be added back, but the existing IP address data will be lost if we do need to roll this back (so if we care about recovering the data it should be backed up first).

## Testing Instructions

Automated tests have been added in `feedback_spec.rb`
1. Submit a support request
2. Verify in the database that the value for "ip_address" is null (or the column itself doesn't exist if migrations have been applied)
3. Verify that Akismet data for this request still contains the IP address

### DIFF
--- a/spec/models/feedback_spec.rb
+++ b/spec/models/feedback_spec.rb
@@ -70,4 +70,17 @@ describe Feedback do
       expect(safe_report.save).to be_truthy
     end
   end
+
+  context "with IP address" do
+    let(:ip) { Faker::Internet.ip_v4_address }
+    let(:feedback) { create(:feedback, ip_address: ip) }
+
+    it "has IP in Akismet attributes" do
+      expect(feedback.akismet_attributes[:user_ip]).to eq(ip)
+    end
+
+    it "does not store IP in the database" do
+      expect(Feedback.find_by(id: feedback.id)[:ip_address]).to eq(nil)
+    end
+  end
 end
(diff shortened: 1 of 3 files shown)


sarken:
Let's add the migration to this, since we won't need the column once this is deployed.

spec/models/feedback_spec.rb
```diff
@@ -70,4 +70,17 @@
...
+  context "with IP address" do
+    let(:ip) { Faker::Internet.ip_v4_address }
+    let(:feedback) { create(:feedback, ip_address: ip) }
+
+    it "has IP in Akismet attributes" do
+      expect(feedback.akismet_attributes[:user_ip]).to eq(ip)
+    end
+
+    it "does not store IP in the database" do
+      expect(Feedback.find_by(id: feedback.id)[:ip_address]).to eq(nil)
```
redsummernight, line 83: You can use `Feedback.find(feedback.id)` which is shorter.

redsummernight (commented):
Looks good, thanks! Just some minor style comments.

redsummernight (approved):

PR #4014, opened 2021-06-04
AO3-4932 Only save tags when the work is saved, and other tag-related fixes.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-3455
https://otwarchive.atlassian.net/browse/AO3-3661
https://otwarchive.atlassian.net/browse/AO3-4932
https://otwarchive.atlassian.net/browse/AO3-5041
https://otwarchive.atlassian.net/browse/AO3-5978
https://otwarchive.atlassian.net/browse/AO3-6043

## Purpose

- I rewrote the `Taggable` concern so that modifying tags through `fandom_string=`, `tag_string=`, and such doesn't immediately change the tags. Instead, these functions build taggings and mark them for destruction, so that the taggings will update when the `Taggable` object is saved to the database. This should prevent issues like AO3-4932, where the tags were written as soon as `fandom_string=` was called, even though the result was an invalid work.
- Because of the order that taggings are saved in, this meant I had to rewrite the code limiting the number of tags on a collection. (Basically, the `Tagging` validations are all run before the first `Tagging` is saved to the database, so `tags.count` is always zero regardless of how many tags the user is trying to apply.)
- I refactored the code checking for missing fandoms/ratings/warnings to use validations.
- I modified the `BookmarksController` to fix AO3-5978, because the fact that it wouldn't display errors if `@bookmark.update_attributes` failed made it more difficult to check the behavior of the new `Taggable` code on bookmarks.
- I modified the work form to address AO3-3661 and AO3-6043 (the issues making it impossible to remove the last category/warning), because (a) AO3-6043 masks an issue similar to AO3-4932 with `archive_warning_string=`, and (b) the two issues have the same fix.
- I fixed AO3-3455 and AO3-5041, the issues where the work meta/tags don't update when previewing, to make it easier to check when the tags have changed and when they haven't.

## Testing Instructions

See the instructions in each of the linked issues.


app/models/tagging.rb
```diff
@@ -1,9 +1,8 @@
 class Tagging < ApplicationRecord
-  belongs_to :tagger, polymorphic: true, inverse_of: :taggings
+  belongs_to :tagger, polymorphic: true, inverse_of: :taggings, autosave: true
```
redsummernight, line 2: When would we need to auto-save a tag after modifying its taggings? I'm fine with this change, I'm just wondering.

app/models/concerns/taggable.rb
```diff
@@ -16,185 +13,135 @@ module Taggable
...
-              klass.find_by(name: string, canonical: true)
-            end
-      next unless tag.present? && tag.is_a?(klass)
-
-      if tag.valid?
-        tags << tag if tag.is_a?(klass)
+  # Split the tag_string, and look up the tags for each of those tag names. We
+  # ignore tag type for this one, and create UnsortedTags for any tag names
+  # that we can't find.
+  def parse_tags(tag_string)
```
redsummernight, line 115: Should we keep the comment that we only use this method and create UnsortedTags for bookmarks and collections?

redsummernight (commented):
> Instead, these functions build taggings and mark them for destruction, so that the taggings will update when the Taggable object is saved to the database.

We still haven't made full use of Rails, ha. Great rewrite, thank you!

spec/controllers/works/default_rails_actions_spec.rb
```diff
@@ -233,23 +233,32 @@ def call_with_params(params)
       post :create, params: { work: work_attributes }
       expect(response).to render_template("new")
       expect(assigns[:work].errors.full_messages).to \
-        include "Please add all required tags. Warning is missing."
+        include /Only canonical warning tags are allowed./
```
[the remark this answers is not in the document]
redsummernight: Yeah same, I'm waiting for Rails 6's error formatting to fix [AO3-5635](https://otwarchive.atlassian.net/browse/AO3-5635) in particular.

spec/spec_helper.rb
```diff
@@ -46,15 +46,15 @@
...
-    ArchiveWarning.find_or_create_by_name(ArchiveConfig.WARNING_CHAN_TAG_NAME).update(canonical: true)
-    ArchiveWarning.find_or_create_by_name(ArchiveConfig.WARNING_NONE_TAG_NAME).update(canonical: true)
-    Category.find_or_create_by_name(ArchiveConfig.CATEGORY_SLASH_TAG_NAME).update(canonical: true)
-    Rating.find_or_create_by_name(ArchiveConfig.RATING_DEFAULT_TAG_NAME).update(canonical: true)
-    Rating.find_or_create_by_name(ArchiveConfig.RATING_EXPLICIT_TAG_NAME).update(canonical: true)
+    ArchiveWarning.find_or_create_by!(name: ArchiveConfig.WARNING_CHAN_TAG_NAME, canonical: true)
+    ArchiveWarning.find_or_create_by!(name: ArchiveConfig.WARNING_NONE_TAG_NAME, canonical: true)
+    Category.find_or_create_by!(name: ArchiveConfig.CATEGORY_SLASH_TAG_NAME, canonical: true)
+    Rating.find_or_create_by!(name: ArchiveConfig.RATING_DEFAULT_TAG_NAME, canonical: true)
+    Rating.find_or_create_by!(name: ArchiveConfig.RATING_EXPLICIT_TAG_NAME, canonical: true)
```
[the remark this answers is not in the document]
redsummernight: Makes sense. Can you leave a TODO ("Not Rated" should be adult) here as well?

app/models/tagging.rb
redsummernight: Nice, thanks for explaining this!

redsummernight (approved):

sarken:
I noticed two tests that are erroneously marked pending AO3-3455:

https://github.com/otwcode/otwarchive/blob/b63fce67a3975ec0614bb535c75d76c8dafbda71/features/prompt_memes_b/challenge_promptmeme_posting_fills.feature#L264-L266

https://github.com/otwcode/otwarchive/blob/b63fce67a3975ec0614bb535c75d76c8dafbda71/features/prompt_memes_b/challenge_promptmeme_posting_fills.feature#L338-L340

As far as I could tell, collection names show up correctly when previewing changes to drafts, so they could probably be uncommented.

PR #4017, opened 2021-06-06
AO3-6174 Make titles wrap inside blurbs and news/collection/work pages

niconicosette:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6174

## Purpose

makes obnoxiously long titles wrap.

## Testing Instructions

1. try creating a work with the title "a_really_obnoxiously_long_outlier_of_a_collection_name_for_this_google_code_issue_whee_i_am_so_happy_i_cannot_contain_my_joy_it_is_a_new_bug" and check that the title text wraps inside the blurb. click on the title and check that the title text also wraps inside the work.
2. try creating a collection with the same title. check that the title text wraps inside the blurb. click on the title and check that the title text also wraps inside the collection page.
3. try creating a news post with the same title. check that the title text wraps inside the landing page's news feed blurbs. click on the title and check that the title text also wraps inside the news post.

## References

the changes made to resolve this issue have been undone https://otwarchive.atlassian.net/browse/AO3-3992

## Comments

this fixes the issue where titles run outside of blurbs, but they still scroll inside work skins/news articles/collection pages, which can be fixed with

```
.heading {
    word-wrap: break-word;
}
```

but i wasn't sure if this was a fix to tack onto this issue or if it would require its own issue, or if it's something we should ignore for now

### DIFF
--- a/public/stylesheets/site/2.0/02-elements.css
+++ b/public/stylesheets/site/2.0/02-elements.css
@@ -54,7 +54,6 @@ h1, h2, h3, h4, h5, h6, .heading {
   font-family: Georgia, serif;
   font-weight: 400;
     word-wrap: break-word;
-
 }
 
 h1 {
--- a/public/stylesheets/site/2.0/02-elements.css
+++ b/public/stylesheets/site/2.0/02-elements.css
@@ -53,6 +53,8 @@ a:focus img {
 h1, h2, h3, h4, h5, h6, .heading {
   font-family: Georgia, serif;
   font-weight: 400;
+    word-wrap: break-word;
+
 }
 
 h1 {
--- a/public/stylesheets/site/2.0/13-group-blurb.css
+++ b/public/stylesheets/site/2.0/13-group-blurb.css
@@ -268,10 +268,6 @@ eg collections, users, skins, instead of the 4-icon list
 
 /*various little mods*/
 
-.heading {
-    word-wrap: break-word;
-}
-
 /* mod: ITEM
 blurbs on the manage collection items pages, mostly reseting styles inherited from interactions (forms)
 */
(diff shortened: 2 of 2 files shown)


sarken:
>but i wasn't sure if this was a fix to tack onto this issue or if it would require its own issue, or if it's something we should ignore for now

I think if we do that, we wouldn't need to do the blurb-specific change? So we can just do that instead, although we'll need to update the testing instructions. (Can you keep the `word-wrap` property indented by four spaces when you do? We do that with CSS3 properties. It doesn't make a terrible amount of sense now that they're no longer new, but it did when we started, and since it's harmless, it's easier to keep doing it for consistency than change it 😆.)

niconicosette:
roger! i'll write out the new testing guidelines now

public/stylesheets/site/2.0/13-group-blurb.css
```diff
@@ -268,7 +268,7 @@ eg collections, users, skins, instead of the 4-icon list
 
 /*various little mods*/
 
-.collection .name {
+.heading {
```
sarken, line 271: Since this is no longer specific to blurbs, it should probably be in this block in 02-elements.css: https://github.com/otwcode/otwarchive/blob/fa272977875c757b2cf899743e30beac5ae8a08e/public/stylesheets/site/2.0/02-elements.css#L53-L56

That way, it would also cover any heading tags used in work summaries. The _slight_ downside is it will also apply to heading tags used in works, and it's _possible_ that there are a few users with work skins who count on the headings not breaking, but I think that's such an edge case it's not worth coding an exception for.

public/stylesheets/site/2.0/02-elements.css
```diff
@@ -53,6 +53,8 @@ a:focus img {
 h1, h2, h3, h4, h5, h6, .heading {
   font-family: Georgia, serif;
   font-weight: 400;
+    word-wrap: break-word;
+
```
sarken, line 57: Can you remove this extra blank line that snuck in? Then we can merge :tada:

niconicosette: whoops 👀

all gone! ✌️

PR #4023, opened 2021-06-08
AO3-6181 Refactor ChallengeSignup#unique_tags.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6181

## Purpose

Refactor `ChallengeSignup#unique_tags` to reduce duplication and make it more readable.

It's still not as simple as Code Climate wants, but it's an improvement.

### DIFF
--- a/app/models/challenge_signup.rb
+++ b/app/models/challenge_signup.rb
@@ -102,69 +102,30 @@ class ChallengeSignup < ApplicationRecord
   # make sure that tags are unique across each group of prompts
   validate :unique_tags
   def unique_tags
-    if (challenge = collection.challenge)
-      errors_to_add = []
-      %w(prompts requests).each do |prompt_type|
-        restriction = case prompt_type
-          when "prompts"
-          then challenge.prompt_restriction
-          when "requests"
-          then challenge.request_restriction
+    return unless (challenge = collection.challenge)
+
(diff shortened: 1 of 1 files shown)


app/models/challenge_signup.rb
```diff
@@ -102,70 +102,31 @@ def number_of_prompts
   # make sure that tags are unique across each group of prompts
   validate :unique_tags
   def unique_tags
-    if (challenge = collection.challenge)
-      errors_to_add = []
-      %w(prompts requests).each do |prompt_type|
-        restriction = case prompt_type
-          when "prompts"
-          then challenge.prompt_restriction
```
redsummernight, line 110: `prompt_restriction` look unused. There was a [migration to rename it](https://github.com/otwcode/otwarchive/commit/53fdfda1d553b4d1bd3a628dc12dc1c4976e502e#diff-1459db5e5bb1426fc88b844f351ab93292c39ad14ddf80e2c7b8ccbb2bc0bf14), but the column is still around for some reason.

Not related to this issue, though.

redsummernight (approved):

app/models/challenge_signup.rb
tickinginstant: Huh, I knew that it was mostly unused, but I never realized that there was a migration that was supposed to have removed it.

PR #4025, opened 2021-06-13
AO3-6182 Use enums for CollectionItem's collection_approval_status and user_approval_status.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6182

## Purpose

Modifies `CollectionItem`s so that the `user_approval_status` and `collection_approval_status` fields are enums.

In the process of fixing all of the old references to `CollectionItem::APPROVED`, `CollectionItem::REJECTED`, and `CollectionItem::NEUTRAL`, I also modified some scopes to rely less on raw SQL, and more on scopes defined in `CollectionItem`.

(But not the `fulfilled`/`unfulfilled` scopes in `ChallengeClaim`s and `ChallengeAssignment`s. The `unfulfilled` scope checks an `OR` of constraints, which is easier to do through raw SQL, and because the two scopes are set up to be parallel, I thought it'd be easier to see the parallel construction if I left both of them as SQL.)


app/models/collection_item.rb
```diff
@@ -253,18 +221,17 @@ def user_allowed_to_destroy?(user)
     user.is_author_of?(self.item) || self.collection.user_is_maintainer?(user)
   end
 
-  def approve_by_user ; self.user_approval_status = APPROVED ; end
-  def reject_by_user ; self.user_approval_status = REJECTED ; end
-  def approved_by_user? ; self.user_approval_status == APPROVED ; end
-  def rejected_by_user? ; self.user_approval_status == REJECTED ; end
+  def approve_by_user
```
redsummernight, line 224: Can we replaced all uses of this with `approved_by_user!`? Same for `approved_by_collection!`.

app/controllers/collection_items_controller.rb
```diff
@@ -28,7 +28,7 @@ def index
       @collection_items = @collection.collection_items.include_for_works
       @collection_items = case
                           when params[:approved]
-                            @collection_items.approved_by_collection
+                            @collection_items.approved_by_both
```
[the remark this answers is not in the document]
tickinginstant: No, `approved_by_collection` is currently set to include only items approved both by the user and by the collection:
https://github.com/otwcode/otwarchive/blob/be8787671219a524f81ff25c73bbdbd0af01330e/app/models/collection_item.rb#L68-L70
It's not a bug, just a confusing name.

app/models/collection_item.rb
tickinginstant: No, because `approve_by_user` & `approve_by_collection` are primarily used in `approve_automatically`, which runs `before_save` if the `CollectionItem`. So replacing them with `approved_by_user!` & `approved_by_collection!`, which save the record, results in an infinite loop.

redsummernight (approved):

PR #4026, opened 2021-06-13
 AO3-6177 Fix flaky rate limiting specs

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6177

## Purpose

- Use an in-memory cache instead of memcached.
- Use the `:login` sequence because `Faker::Name.first_name` may be too short.
- Use `Faker::Internet.public_ip_v4_address` because it's guaranteed not to be private IPs (10.0.0.0/8), which are safelisted and never throttled.

## Testing Instructions

None. Tests pass [20x4 times](https://github.com/redsummernight/otwarchive/actions/runs/932672070) on my fork.

### DIFF
--- a/spec/requests/rack_attack_spec.rb
+++ b/spec/requests/rack_attack_spec.rb
@@ -1,14 +1,21 @@
 require "spec_helper"
 
 describe "Rack::Attack" do
+  around do |example|
+    default_store = Rack::Attack.cache.store
+    Rack::Attack.cache.store = ActiveSupport::Cache::MemoryStore.new
+    example.run
+    Rack::Attack.cache.store = default_store
+  end
+
   before { freeze_time }
 
   def unique_ip_env
-    { "REMOTE_ADDR" => Faker::Internet.unique.ip_v4_address }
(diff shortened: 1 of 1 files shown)


spec/requests/rack_attack_spec.rb
```diff
@@ -1,14 +1,21 @@
 require "spec_helper"
 
 describe "Rack::Attack" do
+  around do |example|
+    default_store = Rack::Attack.cache.store
```
sarken, line 5: Just to make sure I follow:

* we're remembering the default store is Rack::Attack.cache.store
* then we're overriding Rack::Attack.cache.store
* then we're running the example
* then we're setting Rack::Attack.cache.store back to its default state?

redsummernight: Yeah. By default we connect to memcached:

https://github.com/otwcode/otwarchive/blob/700ccf065e91b4779370a183f6304bcb0780fd4c/config/environments/test.rb#L23-L24

PR #4030, opened 2021-06-24
AO3-6189 Don't set Time.zone in challenge controllers.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6189

## Purpose

Removes the code in `GiftExchangeController` and `PromptMemeController` that sets `Time.zone` to the challenge's time zone.

Also modifies the datetime string conversion in `lib/challenge_core.rb` so that it formats dates in the challenge's time zone.

## Testing Instructions

See Jira.


redsummernight (approved):
Yeah, those are the only `Time.zone` assignments I can find as well.

PR #4033, opened 2021-06-27
AO3-6126 Fix deleting works with deleted comment placeholder chains

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6126

## Purpose

Don't check comment parents when we're in the middle of deleting the ultimate parent (e.g. works).

## Testing Instructions

See issue.

### DIFF
--- a/lib/plugins/acts_as_commentable/comment_methods.rb
+++ b/lib/plugins/acts_as_commentable/comment_methods.rb
@@ -2,20 +2,12 @@ module CommentMethods
 
   def self.included(comment)
     comment.class_eval do
-      #extend ClassMethods
       include InstanceMethods
 
       before_destroy :fix_threading_on_destroy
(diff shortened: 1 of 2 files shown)


lib/plugins/acts_as_commentable/comment_methods.rb
```diff
@@ -141,6 +133,10 @@ def fix_threading_on_destroy
     # children. If our parent was marked as deleted (but not actually
     # destroyed), we may be able to destroy it.
     def check_can_destroy_parent
+      # We're in the middle of a cascade deletion (e.g. a work is being destroyed),
+      # so don't try to recursively reload and check our parents.
+      return if destroyed_by_association
```
sarken, line 138: TIL! Isn't that clever and handy...

sarken (approved):

PR #4034, opened 2021-06-29
AO3-6116 Use Elasticsearch for collection counts and collection fandoms, and cache the counts/fandoms for 5 minutes.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6116

## Purpose

1. Switch to using Elasticsearch to compute collection work/fandom/bookmark counts.
2. Switch to using Elasticsearch to get the IDs of the collection's fandoms and the counts for each fandom.
3. Cache the counts and fandoms retrieved this way.
4. Move the `WorkSearchForm.count_for_user` and `BookmarkSearchForm.count_for_user` to a new module `SearchCounts`, since they're very similar to the collection work/bookmark/fandom counting code.
3. Switch to Elasticsearch for the 5 works & bookmarks displayed on the `CollectionsController#show` page, so that we can get rid of the (occasionally buggy) SQL-based count functions in the `Collection` model, but don't have a SQL vs. ES mismatch between the number of items displayed, and the number in the link at the bottom.


app/controllers/fandoms_controller.rb
```diff
@@ -5,12 +5,9 @@ def index
     if @collection
       @media = Media.canonical.by_name - [Media.find_by(name: ArchiveConfig.MEDIA_NO_TAG_NAME)] - [Media.find_by(name: ArchiveConfig.MEDIA_UNCATEGORIZED_NAME)]
       @page_subtitle = @collection.title
-      if params[:medium_id]
-        @medium = Media.find_by_name(params[:medium_id])
-        @fandoms = @medium.fandoms.canonical if @medium
-      end
-      @fandoms = (@fandoms || Fandom).where("filter_taggings.inherited = 0").by_name.
-                  for_collections_with_count([@collection] + @collection.children)
+      @medium = Media.find_by_name(params[:medium_id]) if params[:medium_id]
```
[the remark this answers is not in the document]
tickinginstant: It's a tag, and the name we're passing probably had `to_param` called on it at some point, so this is not a good plan.

(I do sometimes think that we should rename the `find_by_name` function to something like `from_param`, which would clear up some of Hound's issues and also make it clearer when it needs to be used vs. `find_by(name: name)`. But that's not in scope for this PR.)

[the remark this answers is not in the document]
sarken: That does sound like a worthwhile idea for the future!

app/controllers/collections_controller.rb
```diff
@@ -70,17 +70,27 @@ def show
 
     if @collection.collection_preference.show_random? || params[:show_random]
       # show a random selection of works/bookmarks
-      @works = Work.in_collection(@collection).visible.random_order.limit(ArchiveConfig.NUMBER_OF_ITEMS_VISIBLE_IN_DASHBOARD).includes(:pseuds, :tags, :series, :language, :approved_collections)
```
redsummernight, line 73: The old `.visible` method returns a different scope for admins (to show hidden works). With Elasticsearch we currently [never show any hidden works in any queries](https://github.com/otwcode/otwarchive/blob/e00d3b137c78306a87bc9cf63a4a47b5cca7efd1/app/models/search/work_query.rb#L120-L122). So there's a functionality change here, though I don't think it's a deal breaker.

features/step_definitions/collection_steps.rb
```diff
@@ -49,6 +49,11 @@
   end
 end
 
+When "the collection counts have expired" do
+  step "all indexing jobs have been run"
+  step "it is currently 5 minutes from now"
```
redsummernight, line 54: Can you use the ArchiveConfig value here?

features/collections/collection_navigation.feature
```diff
@@ -27,6 +27,7 @@ Feature: Basic collection navigation
     And I press "Preview"
     And I press "Post"
     And all indexing jobs have been run
```
redsummernight, line 29: We can remove this redundant step now.

features/step_definitions/collection_steps.rb
```diff
@@ -213,6 +219,7 @@
 end
 
 Then /^the author of "([^\"]*)" should be publicly visible$/ do |title|
+  step "all indexing jobs have been run"
```
redsummernight, line 222: Is this step only relevant once we get to `visit collection_path(work.collections.first)`? Can we move it to just before the visit step?

app/controllers/fandoms_controller.rb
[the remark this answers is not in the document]
redsummernight: We have this method ignored in Rubocop's config:

https://github.com/otwcode/otwarchive/blob/e00d3b137c78306a87bc9cf63a4a47b5cca7efd1/.rubocop.yml#L63-L66

which unfortunately doesn't work out so far.

app/controllers/collections_controller.rb
sarken: tbh it drives me bonkers that hidden works show in such a weird spot and nowhere else, so I consider this an improvement and not a dealbreaker.

app/models/search/query.rb
```diff
@@ -32,6 +32,76 @@ def count
...
+
+    QueryResult.new(klass, response, { page: 1, per_page: count })
+  end
+
+  # Perform a specific aggregation:
+  def aggregation_search(aggregation)
+    $elasticsearch.search(
+      index: index_name,
+      type: document_type,
+      body: {
```
redsummernight, line 60: Not sure if we want this before or after ES7. I did a cursory check anyway and I don't think there's any breaking changes here between 6 and 7.

There's a [minor optimization](https://www.elastic.co/guide/en/elasticsearch/reference/7.x/search-aggregations-bucket-composite-aggregation.html#_early_termination) that might be worth a TODO for ES7:
> In order to optimize the early termination it is advised to set `track_total_hits` in the request to false. The number of total hits that match the request can be retrieved on the first request and it would be costly to compute this number on every page.

tickinginstant: It seems like that optimization only applies when there's an index on the field being summarized (otherwise early termination is impossible).  Are you suggesting adding an index on `fandom_ids`?

redsummernight: Ah gotcha, that optimization has a more specific use case than I thought. The current setup is fine.

redsummernight (approved):

PR #4036, opened 2021-07-03
AO3-6161 Only allow jpg/gif/png as pseud icons.

ThePadawan:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6161

## Purpose
When editing pseud icons, only icons with the mime types for GIF, JPEG, and PNG validate.
Icons with other mime types are rejected, so users will not be able to create new pseuds or update existing ones with icons using other mime types.

I still have some open questions:
* Is there a planned migration for icons which are currently in "invalid" formats? Currently, users could not make changes to such pseuds unless the icons are "fixed" too.
* Is the default error message of "Icon content type is invalid" readable enough? (This came for free - anything more complicated would need actual i18n work)
* Is the allowlist approach (only allowing exactly the mime types for GIF, JPEG and PNG) too restrictive? I chose it because the denylist approach did not match the description on the "Edit Pseud" page (and also because .bmp files can have weirder mime types to check)

## Testing Instructions

* Select or create a user.
    * Create a new pseud, selecting a .bmp file (should fail) or a .jpg/.gif/.png file (should succeed) as an icon.
    * Edit a current pseud, selecting a .bmp file (should fail) or a .jpg/.gif/.png file (should succeed) as an icon.
    * Edit a current pseud, select "Delete your icon and revert to our default" and save. This should still succeed (the default icon is a .png)

## Credit
JIRA account: https://otwarchive.atlassian.net/jira/people/557058%3A0ed2d036-cfdd-4793-9df9-3ec3e141a3cf
Please credit me as "Rudolf M. Schreier (ThePadawan)" using "he/she/they" pronouns.


features/step_definitions/icon_steps.rb
```diff
@@ -14,6 +14,9 @@
 end
 
 ### WHEN
+When /^I attach an icon with the extension "([^"]*)"$/ do |extension|
```
redsummernight, line 17: Can you use [Cucumber expressions](https://cucumber.io/docs/cucumber/cucumber-expressions/) for the new step?

redsummernight (commented):
Looks good to me, thanks! Just one comment on style.

redsummernight:
You should be able to comment on, assign, and transition issues on Jira now!

> Is there a planned migration for icons which are currently in "invalid" formats? Currently, users could not make changes to such pseuds unless the icons are "fixed" too.

If users can fix existing invalid pseuds by simply uploading new icons or removing existing icons, I think we can do without migrations. Can they do that, with the changes in this pull request?

> Is the default error message of "Icon content type is invalid" readable enough? (This came for free - anything more complicated would need actual i18n work)

I'm fine with leaving it like that because (1) it's been like that for a while, (2) it's an error message from the paperclip gem, which we may need to move away from eventually. Other reviewers may think differently, though.

> Is the allowlist approach (only allowing exactly the mime types for GIF, JPEG and PNG) too restrictive? I chose it because the denylist approach did not match the description on the "Edit Pseud" page (and also because .bmp files can have weirder mime types to check)

Yeah, this approach is good and consistent with how we handle various other user content (e.g. CSS image URLs, HTML embed sources). It's easier to add more to an allowlist later, than to a denylist.

features/step_definitions/icon_steps.rb
ThePadawan: Thanks for the link, I wasn't aware of those!
I've replaced the regex with an expression.

ThePadawan:
> You should be able to comment on, assign, and transition issues on Jira now!

Thanks!

> > Is there a planned migration for icons which are currently in "invalid" formats? Currently, users could not make changes to such pseuds unless the icons are "fixed" too.
> 
> If users can fix existing invalid pseuds by simply uploading new icons or removing existing icons, I think we can do without migrations. Can they do that, with the changes in this pull request?

Yes, users can still do that. I was considering a case where a user edits a pseud (but not the icon) and gets a validation error about the icon content type, which might be confusing. However, I think that is also a reasonable burden for a presumably small percentage of user interactions.

redsummernight:
> Yes, users can still do that. I was considering a case where a user edits a pseud (but not the icon) and gets a validation error about the icon content type, which might be confusing. However, I think that is also a reasonable burden for a presumably small percentage of user interactions.

Nice, thanks for checking! If users can self-correct, we should be fine.

redsummernight (approved):

PR #4041, opened 2021-07-09
AO3-6094 Correctly set sent_at during invitation after_save callback

ThePadawan:
# Pull Request Checklist

## Issue
https://otwarchive.atlassian.net/browse/AO3-6094 

## Purpose
This persists the change to invitation.sent_at during send_and_set_date to the database. (Previously, it only occurred on the record in memory)

## Testing Instructions
See JIRA ticket. After this fix, inspecting the sent invitation on e.g. /users/{username}/invitations/{sent_invitation_id} should both have a "Created at" timestamp (as before) and a "Sent at" timestamp listed (new).

## References
AFAICT, originally introduced during https://github.com/otwcode/otwarchive/commit/c328cdd980a5ce985367a9b7ef52e8e2ded2ecd8, which fixed https://code.google.com/archive/p/otwarchive/issues/3115 .

### DIFF
--- a/app/models/invitation.rb
+++ b/app/models/invitation.rb
@@ -74,8 +74,7 @@ class Invitation < ApplicationRecord
           UserMailer.invitation(self.id).deliver_now
         end
 
-        # We cannot simple use "self.sent_at = Time.now" since that change is not persisted when this
-        # function is called during after_save. Using update_column, it is.
+        # Skip callbacks within after_save by using update_column to avoid a callback loop
         self.update_column(:sent_at, Time.now)
       rescue Exception => exception
         errors.add(:base, "Notification email could not be sent: #{exception.message}")
--- a/app/models/invitation.rb
+++ b/app/models/invitation.rb
@@ -73,7 +73,10 @@ class Invitation < ApplicationRecord
           # send invitations actively sent by a user synchronously to avoid delays
           UserMailer.invitation(self.id).deliver_now
         end
-        self.sent_at = Time.now
+
+        # We cannot simple use "self.sent_at = Time.now" since that change is not persisted when this
+        # function is called during after_save. Using update_column, it is.
+        self.update_column(:sent_at, Time.now)
       rescue Exception => exception
         errors.add(:base, "Notification email could not be sent: #{exception.message}")
       end
--- a/spec/models/invitation_spec.rb
+++ b/spec/models/invitation_spec.rb
@@ -81,15 +81,16 @@ describe Invitation, :ready do
 
     # Regression test for AO3-6094
     context "Sending an invitation to a valid email" do
-      let(:invite) {build(:invitation, invitee_email: "foo@example.com")}
+      let(:invite) { build(:invitation, invitee_email: "foo@example.com") }
+
       it "succeeds and 'sent_at' is set" do
         expect(invite.save).to be_truthy
-        expect(invite.id).to_not be_nil
+        expect(invite.id).not_to be_nil
 
         # Reload invite to make sure that change to invite.sent_at is actually persisted
         # on the database (not just on the 'invite' object currently being created).
         reloaded_invite = Invitation.find_by(id: invite.id)
-        expect(reloaded_invite).to_not be_nil
+        expect(reloaded_invite).not_to be_nil
         expect(reloaded_invite.sent_at).not_to be_nil
       end
     end
--- a/spec/models/invitation_spec.rb
+++ b/spec/models/invitation_spec.rb
@@ -78,5 +78,20 @@ describe Invitation, :ready do
         end
       end
     end
+
+    # Regression test for AO3-6094
+    context "Sending an invitation to a valid email" do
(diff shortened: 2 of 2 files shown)


app/models/invitation.rb
```diff
@@ -73,7 +73,10 @@ def send_and_set_date
           # send invitations actively sent by a user synchronously to avoid delays
           UserMailer.invitation(self.id).deliver_now
         end
-        self.sent_at = Time.now
+
+        # We cannot simple use "self.sent_at = Time.now" since that change is not persisted when this
+        # function is called during after_save. Using update_column, it is.
+        self.update_column(:sent_at, Time.now)
```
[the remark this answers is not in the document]
ThePadawan: I am not an expert on what to use here.
I'd prefer not changing two things at once during this bugfix and leaving "Time.now" for now, but if a reviewer has a clear direction which value to use, I'd be open for suggestions.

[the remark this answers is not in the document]
redsummernight: It's used a few more times in this class, let's leave it be for now.

app/models/invitation.rb
```diff
@@ -73,7 +73,10 @@ def send_and_set_date
           # send invitations actively sent by a user synchronously to avoid delays
           UserMailer.invitation(self.id).deliver_now
         end
-        self.sent_at = Time.now
+
+        # We cannot simple use "self.sent_at = Time.now" since that change is not persisted when this
```
redsummernight, line 77: I'd prefer to have a comment like "Skip callbacks within after_save to avoid a callback loop" instead, because my first thought on seeing `update_column` is ["why not `update`?"](https://stackoverflow.com/questions/22196318/rails-after-save-infinite-loop).

spec/models/invitation_spec.rb
```diff
@@ -78,5 +78,21 @@
         end
       end
     end
+
+    # Regression test for AO3-6094
```
redsummernight, line 82: You can leave out this comment. We reference Jira issues in code comments for TODOs or tricky stuff that would come back to bite us, but generally not for regression tests.

spec/models/invitation_spec.rb
```diff
@@ -78,5 +78,21 @@
...
+    context "Sending an invitation to a valid email" do
+      let(:invite) { build(:invitation, invitee_email: "foo@example.com") }
+
+      it "succeeds and 'sent_at' is set" do
+        expect(invite.save).to be_truthy
+        expect(invite.id).not_to be_nil
+
+        # Reload invite to make sure that change to invite.sent_at is actually persisted
+        # on the database (not just on the 'invite' object currently being created).
+        reloaded_invite = Invitation.find_by(id: invite.id)
```
redsummernight, line 92: You can use [`reload`](https://api.rubyonrails.org/v6.1.4/classes/ActiveRecord/Persistence.html#method-i-reload) for this. We do that in a lot of tests so you don't need to comment on its use either.

spec/models/invitation_spec.rb
```diff
@@ -78,5 +78,21 @@
...
+        expect(reloaded_invite).not_to be_nil
```
redsummernight, line 93: `reload` will throw an error if the invite wasn't saved to the database, so you can skip the nil check here.

spec/models/invitation_spec.rb
ThePadawan: Heh, I wasn't expecting something so convenient! Thanks for pointing ``reload`` out :)

app/models/invitation.rb
ThePadawan: I've merged your version of the comment and some of the SO post replies you linked.
I hope this makes the most sense to future readers!

spec/models/invitation_spec.rb
```diff
@@ -78,5 +78,17 @@
         end
       end
     end
+
+    context "Sending an invitation to a valid email" do
+      let(:invite) { build(:invitation, invitee_email: "foo@example.com") }
+
+      it "succeeds and 'sent_at' is set" do
+        expect(invite.save).to be_truthy
+        expect(invite.id).not_to be_nil
```
redsummernight, line 87: Let's leave out this check, which is also implicit in a successful `reload`.

spec/models/invitation_spec.rb
```diff
@@ -78,5 +78,17 @@
...
+
+        invite.reload
+        expect(invite.sent_at).not_to be_nil
```
redsummernight, line 90: Note that `reload` returns the same record for chaining, so you can write this as `expect(invite.reload.sent_at).not_to be_nil`.

redsummernight (approved):
Two more minor comments. Otherwise this looks good, thank you!

spec/models/invitation_spec.rb
ThePadawan: I'm usually not a fan of being that terse (because it makes large tests really tough to read), but with a test this small I think it does make sense.

redsummernight (approved):

PR #4022, opened 2021-06-08
AO3-5988 Replace noncanonical categories and ratings

tlee911:
# Pull Request Checklist

* [x] Have you read ["How to write the perfect pull request"](https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/)?
* [x] Have you read the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [x] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)?
* [x] Have you added the [Jira](https://otwarchive.atlassian.net) issue number
  as the *first* thing in your pull request title (e.g. `AO3-1234 Fix thing`)
* [x] Do you have fewer than 5 pull requests already open? If not, please wait
  until they are reviewed and merged before creating new pull requests.

## Issue

https://otwarchive.atlassian.net/browse/AO3-5988

## Purpose

Updated version of #4003 

> **Replace a specific noncanonical rating tag**
> We accidentally imported a large number of works with a noncanonical rating "Teen & Up Audiences".
> 
> Given that this is our culpa, we want to fix these works and give them the canonical version, "Teen And Up Audiences".
> 
> At the end of the task we make sure that no other works use the erroneous rating, and if so destroy it.
> 
> Must be run before the task that changes the type of all noncanonical rating tags into freeform.
> 
> **Change the type of noncanonical rating and category tags**
> Before we managed to prevent the creation of noncanonical ratings, we ended up with a number of such ratings in the Archive. Same for category tags.
> 
> Change all of them into Additional Tags (AKA freeforms).

## Testing Instructions

Automated tests added in `after_tasks.rake_spec.rb`

Further test details in JIRA.

## References

#4003 

## Credit

@CristinaRO (Enigel, she/her) did most of the work in the original PR

PR #4047, opened 2021-07-14
AO3-5988 Add task to replace all noncanonical ratings 

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5988

## Purpose

Originally we were adding a task to fix just "Teen & Up Audiences", the only non-canonical rating on production. But [there are other non-canonical ratings on staging, so we still need a more generic task](https://github.com/otwcode/otwarchive/pull/4022#discussion_r656790379). 

## Testing Instructions

See issue.

## Credit

@CristinaRO and @tlee911; the task was added in #4022 but removed during review.

### DIFF
--- a/lib/tasks/after_tasks.rake
+++ b/lib/tasks/after_tasks.rake
@@ -792,6 +792,45 @@ namespace :After do
     STDOUT.flush
   end
 
+  desc "Clean up noncanonical rating tags"
+  task(clean_up_noncanonical_ratings: :environment) do
+    canonical_not_rated_tag = Rating.find_by!(name: ArchiveConfig.RATING_DEFAULT_TAG_NAME)
+    noncanonical_ratings = Rating.where(canonical: false)
+    puts "There are #{noncanonical_ratings.size} noncanonical rating tags."
+
+    unless noncanonical_ratings.empty?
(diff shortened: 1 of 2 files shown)


lib/tasks/after_tasks.rake
```diff
@@ -788,15 +788,57 @@ namespace :After do
...
     STDOUT.flush
   end
 
+  desc "Clean up noncanonical rating tags"
+  task(clean_up_noncanonical_ratings: :environment) do
+    canonical_not_rated_tag = Rating.find_by!(name: ArchiveConfig.RATING_DEFAULT_TAG_NAME)
+    noncanonical_ratings = Rating.where(canonical: false)
+    puts "There are #{noncanonical_ratings.size} noncanonical rating tags."
+
+    unless noncanonical_ratings.empty?
```
sarken, line 801: If `noncanonical_ratings.empty?`, do we even need to go through the rest of the task?

redsummernight: We don't, and the rest is no-op. I've added code to [return early from the task](https://stackoverflow.com/questions/2316475/how-do-i-return-early-from-a-rake-task).

sarken (approved):

PR #4049, opened 2021-07-14
AO3-6195 Remove unused database dump & restore scripts, as well as associated config settings & task.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6195

## Purpose

Removes three scripts listed in the issue -- `scripts/seed_dump.rb`, `scripts/seed_restore.rb`, and `scripts/print_test_pages.rb` -- as well as the associated settings in `config/config.yml`.

I also removed a task that wasn't listed in the issue, but uses `ArchiveConfig.DUMP_EMAIL`, one of the three seed-related settings from the issue.


lib/tasks/deploy_tasks.rake
```diff
@@ -13,15 +13,4 @@ namespace :deploy do
       puts "Don't clear subscriptions except on stage!!!"
     end 
   end
-  
-  desc "clear email addresses on stage"
```
sarken, line 17: Hmm. I think we still need a version of this so we can be sure email addresses are wiped from staging the next time we update its data. We don't do that often at all, but Murphy's law being a thing, we'll probably need to do it in the next few days if we merge this. 😆 

Maybe we should leave this task and the `DUMP_EMAIL` config variable for now, and then make an issue for a) creating a new task to dump _all_ emails and b) removing this task and `DUMP_EMAIL`?

cc: @redsummernight

tickinginstant: Oh, I assumed that the email clearing task couldn't be used except in conjunction with the seed dump/restore scripts, because it doesn't touch the collection or comment email fields.

redsummernight: Let's remove it. This task doesn't work cleanly anyway (`DUMP_EMAIL` contains `admin-` users but the task doesn't look at Admin models). If we need to update staging, we'll add a new task.

sarken (approved):

PR #4052, opened 2021-07-16
AO3-6196 Remove actual chapter content from fixtures

nightglyde:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6196

## Purpose

I have run the chapter content from test/fixtures/chapters.yml through a vigenere cipher. This is to remove any adult content, whilst keeping the length of the content the same.

More specific details:
- I have limited changes to the file as much as possible
  - The yaml data structure's keys ("posted:" "word_count:") are unchanged
  - Any html tags (`<p> </p> <em>`) and other symbols (`\r\n \xE2\x80\x93 &nbsp; &rdquo;`) are unchanged
  - Only letters (A-Z, a-z) are changed
- The cipher is run on:
  - "title", "summary", "notes", "content", and "endnotes" fields
  - Any field that continues onto multiple lines (which should only be the above fields anyway)

I wrote a script to encrypt chapters.yml. If you would prefer chapters.yml to be encrypted in a different way (for example not encrypting chapter titles), it would be pretty simple to do. So, let me know.

## Testing Instructions

(I don't know how to test if the changed fixtures are working correctly. I did run chapters.yml through a yaml parser, and didn't get any more errors than the original file had (it doesn't like the tab characters on line 2332). As far as I can tell, it should work fine as is. However, I would really like to test it. I just don't know how.)

## Credit

NightGlyde (she/her)


sarken (approved):
This looks good to me, thank you! (I actually just meant the content of the `content` field, but I see how the issue was unclear, and there's no harm in doing there other fields as well.) 

As for testing, these fixtures are the ones used when [seeding a dev environment](https://github.com/otwcode/otwarchive/wiki/OS-X-Setup#11-seed-your-database-and-load-the-site-styles), so the best way to test it is probably running `bundle exec rake db:otwseed`. However, that _will_ remove any extra data you've made, so I understand if you'd rather not!

nightglyde:
@sarken It would be an easy fix to limit the change to just the `content` field, if you prefer. However, if you're happy with this version then that's cool. :)

sarken:
It's fine the way it is, but thank you!

PR #4053, opened 2021-07-18
AO3-3642 Escape quotes for tag links in wrangling error messages

ThePadawan:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-3642

## Purpose
This PR fixes a bug where a HTML link for a Tag was constructed using manual string concatenation.
This failed for lots of special characters (which are allowed in Tag names).
The new code uses Rails' builtin methods, which no longer fails.
A Cucumber test case was added to assert this.

Note: This already violated layering before, and I have chosen not to change that.
Finding a good generic solution that supported links and l10n seemed like overkill for what is otherwise a 1-line fix.

## Testing Instructions
See JIRA.


app/models/tag.rb
```diff
@@ -995,7 +996,7 @@ def syn_string=(tag_string)
     if new_merger && new_merger == self
       self.errors.add(:base, tag_string + " is considered the same as " + self.name + " by the database.")
     elsif new_merger && !new_merger.canonical?
-      self.errors.add(:base, '<a href="/tags/' + new_merger.to_param + '/edit">' + new_merger.name + '</a> is not a canonical tag. Please make it canonical before adding synonyms to it.')
+      self.errors.add(:base, "<a href=\"#{Rails.application.routes.url_helpers.edit_tag_path(new_merger)}\">#{new_merger.name}</a> is not a canonical tag. Please make it canonical before adding synonyms to it.")
```
redsummernight, line 999: Since you've added `include Rails.application.routes.url_helpers` at the top, just `edit_tag_path(new_merger)` should work here.

features/step_definitions/generic_steps.rb
```diff
@@ -238,6 +238,11 @@
   page.body.should =~ /#{Regexp.escape(text)}/m
 end
 
+Then "I should see a link titled {string} to the URL {string}" do |title, path|
```
redsummernight, line 241: Instead of adding this step, you can test if the link works by following it:
```
When I follow "Evelyn \"Evie\" Carnahan"
Then I should be on the "Evelyn \"Evie\" Carnahan" tag page
```
The quotes may or may not work out though, I'm not sure. But the step definitions are here:
https://github.com/otwcode/otwarchive/blob/8d1456a2b876e5d743c6e71748f8b0cb87f6eb1d/features/step_definitions/web_steps.rb#L46-L50
https://github.com/otwcode/otwarchive/blob/8d1456a2b876e5d743c6e71748f8b0cb87f6eb1d/features/step_definitions/web_steps.rb#L268-L275
https://github.com/otwcode/otwarchive/blob/8d1456a2b876e5d743c6e71748f8b0cb87f6eb1d/features/support/paths.rb#L267-L268

features/tags_and_wrangling/tag_wrangling_characters.feature
```diff
@@ -169,3 +169,18 @@ Scenario: character wrangling - syns, mergers, characters, autocompletes
   When I fill in "Synonym of" with "Doctor Who"
     And I press "Save changes"
   Then I should see "Doctor Who is a fandom. Synonyms must belong to the same category."
+
+
+Scenario: Error messages show correct links even if tags contain special characters
```
redsummernight, line 174: Can you move this scenario to features/tags_and_wrangling/tag_wrangling_special.feature?

features/step_definitions/generic_steps.rb
ThePadawan: Thanks for the input! That'll probably make it easier.

I had to port the fix that allowed double quotes within singly-quoted literals in two places, but that shouldn't be a big problem.

redsummernight (approved):
Thanks for updating the test steps!

PR #4054, opened 2021-07-24
AO3-6103 AO3-5731 Standardize email greetings and closings for i18n

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6103
https://otwarchive.atlassian.net/browse/AO3-5731

## Purpose

Standardizes the greetings and closings in emails. Please refer to AO3-6103 for full details.

Also removes an extra space in the Support email greeting (AO3-5731).

## Testing Instructions

Refer to Jira.


redsummernight (commented):
Can you omit `.deliver` from all the subject declarations?

Without `.deliver`, [the subject is the mailer](https://relishapp.com/rspec/rspec-rails/docs/mailer-specs), but with it, the subject is the message, and in mailer specs we should do the former.

I'll fix the existing subjects when I do [AO3-6165](https://otwarchive.atlassian.net/browse/AO3-6165).

spec/mailers/user_mailer_spec.rb
```diff
@@ -432,4 +434,337 @@
...
+        expect(email).to have_text_part_content("challenge's Matching page:")
+      end
+    end
+  end
+
+  describe "invalid_signup_notification" do
+    let(:collection) { create(:collection) }
+    let(:signup) { create(:challenge_signup) }
+
+    subject(:email) { UserMailer.invalid_signup_notification(collection.id, [signup.id]) }
```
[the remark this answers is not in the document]
redsummernight: Can you move the subject lines to before the `let`s?

redsummernight (approved):

app/views/user_mailer/collection_notification.text.erb
```diff
@@ -1,10 +1,8 @@
 <% content_for :message do %>
-Dear Collection Maintainers!
-
 You have received a message about your collection "<%= @collection.title %>" (<%= collection_url(@collection) %>):
 <%= text_divider %>
 
 <%= raw @message %>"
```
sarken, line 5: *twitches* I just noticed this stray quotation mark.

app/mailers/user_mailer.rb
```diff
@@ -207,15 +207,15 @@ def invalid_signup_notification(collection_id, invalid_signup_ids)
...
+      subject: "[#{ArchiveConfig.APP_SHORT_NAME}][#{@collection.title}] Invalid sign-ups found"
     )
   end
 
   def potential_match_generation_notification(collection_id)
     @collection = Collection.find(collection_id)
     mail(
       to: @collection.get_maintainers_email,
-      subject: "[#{ArchiveConfig.APP_SHORT_NAME}][#{@collection.title}] Potential Assignment Generation Complete"
+      subject: "[#{ArchiveConfig.APP_SHORT_NAME}][#{@collection.title}] Potential assignment generation complete"
```
sarken, line 218: Also, why do we call this "assignment generation"? We call it "potential match(es)" everywhere else.

config/locales/mailers/en.yml
```diff
@@ -209,14 +211,9 @@ en:
         part9: "If you would like Open Doors to update the redirect to point to your pre-existing work, please delete the imported copy, and %{open_doors_link} with your AO3 account name, your account name on the imported archive, and the title and URL of the fanwork you would like the redirect to point to. (If you have multiple works you would like to change the redirects for, you can list these in one email.)"
         part10: "If you had other works on the imported archive under an e-mail address you can no longer access, please %{open_doors_link} with any information that can help verify your identity."
         part11: "For other inquiries, please %{support_link}."
-        part13: "Best,"
-        part14: "The Open Doors team"
-        part15: "Organization for Transformative Works"
     invite_request_declined:
-      subject: "[%{app_name}] Additional Invite Code Request Declined"
-      greeting: "Dear %{user},"
+      subject: "[%{app_name}] Additional invite code request declined"
```
sarken, line 215: This should be "invitation code" or "invitation request"

redsummernight:
This has conflicts now.

redsummernight (approved):

app/views/user_mailer/potential_match_generation_notification.html.erb
```diff
@@ -1,6 +1,4 @@
 <% content_for :message do %>
-  <p>Dear Collection Maintainers!</p>
-
   <p>
     We have finished generating potential assignments for your challenge collection, <%= style_link(@collection.title, collection_url(@collection)) %>.
```
sarken, line 3: Ugh, hang on, gotta fix this to "potential matches" too

tickinginstant: FYI, this message is sent when assignment generation is complete, not when potential match generation is complete:
https://github.com/otwcode/otwarchive/blob/ccb7f351e1736e120ccd2d092ee8fb1749e4217d/app/models/challenge_assignment.rb#L380
So the message probably ought to mention assignments somewhere.

sarken: Yeah, I realized it still said assignments because we were having a discussion about when the email gets sent. But basically, we never say "potential assignments" anywhere else, the side bar still refers to "matching" until assignments are actually sent, and it's the potential matching in progress page (not the assignment generation page) that tells users this email will be coming, so "potential matching" is our best bet short of a total rewrite.

sarken: Actually, I'm just going to revert the change to the subject since it's not translatable yet (and therefore we aren't looking at wasting anyone's effort on an incorrect term)

tickinginstant: But mentioning only potential matches makes the email actively incorrect, not just confusing, if the user has pressed the "Regenerate Assignments" button.

sarken: Oh, I thought it was just the initial generation. Let me add that to the comment so we know to take that into consideration when rewriting... basically all of the text around these pages

sarken:
Flipping the label because I did touch this again.

redsummernight (commented):
Should we update the greeting for this email? (If we do, can we move these keys to mailers/en.yml?)

https://github.com/otwcode/otwarchive/blob/be8787671219a524f81ff25c73bbdbd0af01330e/config/locales/devise/en.yml#L74-L81

config/locales/mailers/en.yml
```diff
@@ -68,7 +76,6 @@ en:
     delete_work_notification:
       subject: "[%{app_name}] Your work has been deleted"
       support: "contact Support"
-      hello: "Dear %{user},"
       part2: "If you have questions, please %{support}."
```
redsummernight, line 79: This key is unused.

sarken:
I have test failures and tests that didn't get committed, so rather than keep guessing and pushing, I'll deal with that in the next couple of days when I have a chance to use both the computer with the uncommitted files and the one that can run the tests locally.

sarken:
I'll get the label right _eventually_.

app/views/user_mailer/collection_notification.text.erb
redsummernight: I won't mind if you deal with it here. We don't even need manual testing for it.

sarken: Whoops, forgot about that. Done now!

PR #4055, opened 2021-07-24
AO3-6206 Update text of email that is sent when an admin hides a work

niconicosette:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6206

## Purpose

Updates the content of the html and text versions, apply `style_bold` to `user`, apply `style_creation_link` to `work`, make `user` username instead of default pseud

## Testing Instructions

1. Log in, making sure to use an account that's hooked up to an email address you can access
2. Post > New Work
3. Fill in required information
4. Press "Post"
5. Make note of the URL
6. Log out
7. Log in as an admin with the role policy_and_abuse or superadmin
8. Go to the work you posted
9. Press "Hide Work"
10. Ensure the email text is as described below

If you have a Jira account with access, please update or comment on the issue with any new or missing testing instructions instead.

### DIFF
--- a/app/views/user_mailer/admin_hidden_work_notification.html.erb
+++ b/app/views/user_mailer/admin_hidden_work_notification.html.erb
@@ -1,7 +1,7 @@
 <% content_for :message do %>
-  <p><%= t '.hello', user: @user.default_pseud.byline %></p>
+  <p><%= t(".hello", name: style_bold(@user.login)).html_safe %></p>
 
-  <p><%= t('.html.part1', title: "<i>#{style_bold(@work.title.html_safe)}</i>".html_safe).html_safe %></p>
+  <p><%= t(".html.part1", title: style_creation_link(@work.title)).html_safe %></p>
 
   <p><%= t '.html.part2' %></p>
 
@@ -12,5 +12,6 @@
   <p><%= t('.html.part5', abuse_link: abuse_link(t '.abuse_committee')).html_safe %></p>
 
   <p><%= t '.sincerely' %></p>
+
   <p><%= t '.sender' %></p>
 <% end %>
--- a/app/views/user_mailer/admin_hidden_work_notification.html.erb
+++ b/app/views/user_mailer/admin_hidden_work_notification.html.erb
@@ -5,7 +5,7 @@
 
   <p><%= t ".access" %></p>
 
-  <p><%= t ".check_mail" %></p>
+  <p><%= t ".check_email" %></p>
 
   <p><%= t(".html.tos_violation", tos_link: tos_link(t ".tos")).html_safe %></p>
 
--- a/app/views/user_mailer/admin_hidden_work_notification.html.erb
+++ b/app/views/user_mailer/admin_hidden_work_notification.html.erb
@@ -13,5 +13,5 @@
 
   <p><%= t '.sincerely' %></p>
 
-  <p><%= t '.sender' %></p>
+  <p><%= style_bold(t '.sender') %></p>
 <% end %>
--- a/app/views/user_mailer/admin_hidden_work_notification.html.erb
+++ b/app/views/user_mailer/admin_hidden_work_notification.html.erb
@@ -1,12 +1,16 @@
 <% content_for :message do %>
-    <p>Dear <%= @user.default_pseud.byline %>,</p>
+  <p><%= t '.hello', user: @user.default_pseud.byline %></p>
 
-    <p>Your work <i><%= style_link(@work.title.html_safe, work_url(@work)) %></i> has been reported as being in violation of the Archive of Our Own's <%= style_link("Terms of Service", tos_url) %>.</p>
+  <p><%= t('.html.part1', title: "<i>#{style_bold(@work.title.html_safe)}</i>".html_safe).html_safe %></p>
 
-    <p>We are investigating the matter and will contact you once the investigation has been completed. While the work is under investigation it has been hidden; only you will be able to access it.</p>
+  <p><%= t '.html.part2' %></p>
 
-    <p>If you have any questions, please <%= abuse_link("contact our Abuse Team") %>.</p>
+  <p><%= t '.html.part3' %></p>
 
-    <p>Sincerely,</p>
-    <p><%= style_bold("AO3 Abuse") %></p>
+  <p><%= t('.html.part4', tos_link: tos_link(t '.tos')).html_safe %></p>
+
+  <p><%= t('.html.part5', abuse_link: abuse_link(t '.abuse_committee')).html_safe %></p>
+
+  <p><%= t '.sincerely' %></p>
+  <p><%= t '.sender' %></p>
 <% end %>
--- a/app/views/user_mailer/admin_hidden_work_notification.html.erb
+++ b/app/views/user_mailer/admin_hidden_work_notification.html.erb
@@ -1,17 +1,13 @@
 <% content_for :message do %>
   <p><%= t(".hello", name: style_bold(@user.login)).html_safe %></p>
 
-  <p><%= t(".html.part1", title: style_creation_link(@work.title, @work)).html_safe %></p>
+  <p><%= t(".html.hidden", title: style_creation_link(@work.title, @work)).html_safe %></p>
 
-  <p><%= t '.html.part2' %></p>
+  <p><%= t '.access' %></p>
 
-  <p><%= t '.html.part3' %></p>
+  <p><%= t '.check_mail' %></p>
 
-  <p><%= t('.html.part4', tos_link: tos_link(t '.tos')).html_safe %></p>
+  <p><%= t('.html.tos_violation', tos_link: tos_link(t '.tos')).html_safe %></p>
 
-  <p><%= t('.html.part5', abuse_link: abuse_link(t '.abuse_committee')).html_safe %></p>
-
-  <p><%= t '.sincerely' %></p>
-
-  <p><%= style_bold(t '.sender') %></p>
+  <p><%= t('.html.help', contact_abuse_link: abuse_link(t '.abuse_committee')).html_safe %></p>
 <% end %>
--- a/app/views/user_mailer/admin_hidden_work_notification.html.erb
+++ b/app/views/user_mailer/admin_hidden_work_notification.html.erb
@@ -3,11 +3,11 @@
 
   <p><%= t(".html.hidden", title: style_creation_link(@work.title, @work)).html_safe %></p>
 
-  <p><%= t '.access' %></p>
+  <p><%= t ".access" %></p>
 
-  <p><%= t '.check_mail' %></p>
+  <p><%= t ".check_mail" %></p>
 
-  <p><%= t('.html.tos_violation', tos_link: tos_link(t '.tos')).html_safe %></p>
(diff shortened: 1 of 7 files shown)


config/locales/mailers/en.yml
```diff
@@ -50,6 +50,25 @@ en:
...
+      tos: "Terms of Service"
+      abuse_committee: "contact Policy & Abuse"
+      sincerely: "Sincerely,"
+      sender: "AO3 Abuse"
+      html:
+        part1: "Your work %{title} has been hidden by the Policy & Abuse team and is no longer publicly accessible."
+        part2: "While your work is hidden, you will still be able to access it through the link provided above, but it will not be listed on your works page, and it won't be available to other users of the Archive."
+        part3: "Please check your email, including your spam folder, as the Policy & Abuse team may have already contacted you explaining why your work was hidden."
+        part4: "If your work was hidden due to being in violation of the Archive of Our Own's %{tos_link}, you will be required to take action to correct the violation. Failure to bring your work into compliance with the Terms of Service may lead to your work being deleted from the Archive."
+        part5: "If you are uncertain why your work was hidden, and you have not received further communication regarding this matter, please %{abuse_link} directly."
```
sarken, line 65: Could you change the key names per [the guidelines](https://github.com/otwcode/otwarchive/wiki/Internationalization-(i18n)-Standards)?

>**Keys should be descriptive and not be numbered sequentially.** Avoid key names like part1 or para1. If you have a five-paragraph email, with each paragraph called para1, para2, and so on, when you insert a new paragraph between first and second paragraphs, it will require the keys for existing paragraphs to be renumbered. This will require our Translation team to reenter the text for all of those paragraphs in every language.

Going in order, they could probably be renamed:
* hidden
* access
* check_email
* tos_violation
* help

It also looks like part2/access and part3/check_email are the same in both the HTML and text versions, so you only need to include it once, directly under `admin_hidden_work_notification` instead of under both `html` and `text`.

app/views/user_mailer/admin_hidden_work_notification.html.erb
```diff
@@ -1,12 +1,17 @@
...
-    <p>Sincerely,</p>
-    <p><%= style_bold("AO3 Abuse") %></p>
-<% end %>
+  <p><%= t('.html.part4', tos_link: tos_link(t '.tos')).html_safe %></p>
+
+  <p><%= t('.html.part5', abuse_link: abuse_link(t '.abuse_committee')).html_safe %></p>
+
+  <p><%= t '.sincerely' %></p>
+
+  <p><%= style_bold(t '.sender') %></p>
```
sarken, line 16: Could you remove this from both emails?

app/views/user_mailer/admin_hidden_work_notification.text.erb
```diff
@@ -1,12 +1,17 @@
...
 
-    We are investigating the matter and will contact you once the investigation has been completed. While the work is under investigation it has been hidden; only you will be able to access it.
+<%= t '.text.part2' %>
 
-    If you have any questions, please contact our Abuse Team (<%= root_url + "abuse_reports/new" %>).
+<%= t '.text.part3' %>
 
-    Sincerely,
-    AO3 Abuse
+<%= t('.text.part4', tos_link: 'http://archiveofourown.org/tos') %>
```
sarken, line 10: Could you use `tos_url` instead of writing out the URL manually? That way, someone running our code elsewhere won't need to update the email.

app/views/user_mailer/admin_hidden_work_notification.text.erb
```diff
@@ -1,12 +1,17 @@
...
+
+<%= t('.text.part5', abuse_link: 'http://archiveofourown.org/abuse_reports/new') %>
```
sarken, line 12: Can you use `new_abuse_report_url` here instead? Same reason as above.

config/locales/mailers/en.yml
```diff
@@ -50,6 +50,25 @@ en:
...
+      text:
+        part1: "Your work \%{title}\ (%{work_url}) has been hidden by the Policy & Abuse team and is no longer publicly accessible."
+        part2: "While your work is hidden, you will still be able to access it through the link provided above, but it will not be listed on your works page, and it won't be available to other users of the Archive."
+        part3: "Please check your email, including your spam folder, as the Policy & Abuse team may have already contacted you explaining why your work was hidden."
+        part4: "If your work was hidden due to being in violation of the Archive of Our Own's Terms of Service (%{tos_link}), you will be required to take action to correct the violation. Failure to bring your work into compliance with the Terms of Service may lead to your work being deleted from the Archive."
+        part5: "If you are uncertain why your work was hidden, and you have not received further communication regarding this matter, please contact Policy & Abuse directly: (%{abuse_link})"
```
sarken, line 71: Oops, my bad: there shouldn't be parentheses around the URL here, although there should be a period at the end.

app/views/user_mailer/admin_hidden_work_notification.html.erb
```diff
@@ -1,12 +1,17 @@
...
 
-    <p>If you have any questions, please <%= abuse_link("contact our Abuse Team") %>.</p>
+  <p><%= t '.html.part3' %></p>
 
-    <p>Sincerely,</p>
-    <p><%= style_bold("AO3 Abuse") %></p>
-<% end %>
+  <p><%= t('.html.part4', tos_link: tos_link(t '.tos')).html_safe %></p>
+
+  <p><%= t('.html.part5', abuse_link: abuse_link(t '.abuse_committee')).html_safe %></p>
```
sarken, line 12: To help Translation out, could you update the keys for the links and URLs per [the guidelines](https://github.com/otwcode/otwarchive/wiki/Internationalization-(i18n)-Standards)?

>**If the variable represents a hyperlink, its name should end with _link.** This helps our translators identify links and the linked text. For example, `<p>If you have questions, please <%= link_to("contact Support", new_feedback_report_url) %>.</p>` can be replaced with `<p><%= t(".questions", support_link: link_to(t(".support", new_feedback_report_url)).html_safe) %></p>`. Note that the variable is the key for the linked text (`support`) plus the suffix` _link`.
>
>**If the variable represents a URL, its name should end with _url.** Particularly with mailers, this ensures links and URLs have distinct keys. For example, `If you have questions, please contact Support: <%= new_feedback_report_url %>.` can be replaced with `<%= t(".questions", support_url: new_feedback_report_url) %>.`

So here instead of `<%= t('.html.part5', abuse_link: abuse_link(t '.abuse_committee')).html_safe %>`, it would be: 

`<%= t(".html.part5", contact_abuse_link: abuse_link(t(".contact_abuse"))).html_safe %>`

(I changed the name because I thought it would be clearer if the variable didn't have the same name as the link helper method.)

And then in the text email, instead of `<%= t('.text.part5', abuse_link: 'http://archiveofourown.org/abuse_reports/new') %>`, it would be: 

`<%= t(".text.part5", contact_abuse_url: new_abuse_report_url) %>`

niconicosette:
on second thought maybe `the Policy & Abuse team may have already contacted you explaining why your work was hidden` would be a better text snippet to use for the test...

sarken (changes requested):
Yay, new email text with i18n! \o/

In addition to the comments I left, if you could
* switch to double quotes consistently in the view files
* add some tests to [user_mailer_spec.rb](https://github.com/otwcode/otwarchive/blob/master/spec/mailers/user_mailer_spec.rb)

If you need help with the specs or if my comments are dreadfully unclear, lmk and I will do my best to clarify!

niconicosette:
i think i got all of them :o

config/locales/mailers/en.yml
```diff
@@ -50,6 +50,23 @@ en:
         part1: "Your work \"%{title}\" was deleted from the Archive by a site admin."
         part2: "If your work was part of an import project managed by our Open Doors team, please contact Open Doors (%{opendoors_link}) with any further questions."
         part3: "If it's possible your work violated the Archive's Terms of Service, please contact our Abuse Committee (%{abuse_link})."
+    admin_hidden_work_notification:
+      subject: "[%{app_name}] Your work has been hidden by the Policy & Abuse Team"
```
sarken, line 54: Could you lowercase "team"?

config/locales/mailers/en.yml
```diff
@@ -50,6 +50,23 @@ en:
...
+      hello: "Dear %{name},"
+      tos: "Terms of Service"
+      abuse_committee: "contact Policy & Abuse"
+      access: "While your work is hidden, you will still be able to access it through the link provided above, but it will not be listed on your works page, and it won't be available to other users of the Archive."
+      check_email: "Please check your email, including your spam folder, as the Policy & Abuse team may have already contacted you explaining why your work was hidden."
+      sincerely: "Sincerely,"
+      sender: "AO3 Abuse"
```
sarken, line 61: These lines can be removed now.

sarken (changes requested):
You got almost all of them! :tada: I left a couple of comments, and it still needs

* single quotes in the view files converted to double
* specs

You can copy [the specs I wrote for this mailer](https://github.com/otwcode/otwarchive/blob/13f84da0c63b8a885f76728c7009e0f279f3d5cd/spec/mailers/user_mailer_spec.rb#L503-L536) (that's from #4054) if you want and just update the text they're looking for.

niconicosette:
> You can copy [the specs I wrote for this mailer](https://github.com/otwcode/otwarchive/blob/13f84da0c63b8a885f76728c7009e0f279f3d5cd/spec/mailers/user_mailer_spec.rb#L503-L536) (that's from #4054) if you want and just update the text they're looking for.

i was mildly intimidated by the specs... i changed `has been reported` to `has been hidden` but was that all it needed?

sarken:
Yep, I think so! [And look at that, they caught a bug](https://github.com/otwcode/otwarchive/pull/4055/checks?check_run_id=3176883362#step:13:1868).

It looks like the view files point to the key `check_mail` but the key in `en.yml` is actually `check_email`. We need to change one of those, then.

sarken (approved):
Yay, that should do it! (The tests are my code, but they're from #4054, a pull request that @redsummernight approved, so I feel semi-okay about hitting the "Approve" button on this PR.)

app/views/user_mailer/admin_hidden_work_notification.text.erb
```diff
@@ -1,12 +1,13 @@
 <% content_for :message do %>
-    Dear <%= @user.default_pseud.byline %>,
+<%= t(".hello", name: @user.login) %>
 
-    Your work "<%= @work.title.html_safe %>" (<%= work_url(@work) %>) has been reported to the Abuse Team as a violation of the Archive of Our Own's Terms of Service (<%= root_url + "tos" %>).
+<%= t(".text.hidden", title: @work.title, work_url: @work) %>
```
sarken, line 4: Ah, found the culprit! This needs to be `work_url: work_url(@work)` to fix [the issue we found on staging](https://otwarchive.atlassian.net/browse/AO3-6206?focusedCommentId=357675).

niconicosette: oh no... how does it look now

niconicosette: https://github.com/otwcode/otwarchive/pull/4059/

PR #4056, opened 2021-07-27
AO3-6165 Skip comment reply notifications for guests with banned emails

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6165

## Purpose

Before sending reply notifications for guest comments (but not logged in comments), we should check if the comment's email is in the banned list. If it is, we should skip the notification.

Also make sure the subject of mailer specs is the mailer object, not the message object (returned by "deliver").

## Testing Instructions

See issue.


sarken (approved):
One of the beautiful things about being chair is I get to make the executive decision to put this in this release. :P

PR #4057, opened 2021-07-31
AO3-6153 Limit maximum number of user-defined tags allowed on works & external works.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6153

## Purpose

- Limit the number of user-defined tags used on works to 75 (configurable).
- Limit the number of user-defined tags used on external works to 75 (configurable).
- Add assorted messages and error messages.
- Modify `BookmarksController` so that it no longer explicitly calls `@bookmarkable.save`. Instead, we rely on Rails' built-in behavior when saving objects that refer to other unsaved objects, and add some checks to the bookmark class to make sure that we only save the bookmark when the new external work is valid and has a fandom.
- Modify `CollectionItem#update_work` so that it skips validations when saving the work. This is so that works can be revealed/de-anoned even if they're over the tag limit.

### DIFF
--- a/app/models/bookmark.rb
+++ b/app/models/bookmark.rb
@@ -25,14 +25,12 @@ class Bookmark < ApplicationRecord
   def check_new_external_work
     return unless bookmarkable.is_a?(ExternalWork) && bookmarkable.new_record?
 
-    if bookmarkable.fandom_string.blank?
-      errors.add(:base, "Fandom tag is required")
-    end
+    errors.add(:base, "Fandom tag is required") if bookmarkable.fandom_string.blank?
 
-    if bookmarkable.invalid?
-      bookmarkable.errors.full_messages.each do |message|
-        errors.add(:base, message)
-      end
+    return if bookmarkable.valid?
+
+    bookmarkable.errors.full_messages.each do |message|
+      errors.add(:base, message)
     end
   end
 
--- a/app/models/bookmark.rb
+++ b/app/models/bookmark.rb
@@ -6,7 +6,7 @@ class Bookmark < ApplicationRecord
   include Responder
   include Taggable
 
-  belongs_to :bookmarkable, polymorphic: true
+  belongs_to :bookmarkable, polymorphic: true, inverse_of: :bookmarks
   belongs_to :pseud
 
   validates_length_of :bookmarker_notes,
@@ -19,6 +19,21 @@ class Bookmark < ApplicationRecord
     return if self.pseud.user.bookmarks.where(bookmarkable: self.bookmarkable).empty?
 
     errors.add(:base, ts("You have already bookmarked that."))
+  end
+
+  validate :check_new_external_work
+  def check_new_external_work
+    return unless bookmarkable.is_a?(ExternalWork) && bookmarkable.new_record?
+
+    if bookmarkable.fandom_string.blank?
+      errors.add(:base, "Fandom tag is required")
+    end
+
+    if bookmarkable.invalid?
+      bookmarkable.errors.full_messages.each do |message|
+        errors.add(:base, message)
+      end
+    end
   end
 
   default_scope -> { order("bookmarks.id DESC") } # id's stand in for creation date
--- a/app/models/external_work.rb
+++ b/app/models/external_work.rb
@@ -32,6 +32,9 @@ class ExternalWork < ApplicationRecord
                                too_long: ts('^Creator must be less than %{max} characters long.',
                                             max: AUTHOR_LENGTH_MAX)
 
+  validates :user_defined_tags_count,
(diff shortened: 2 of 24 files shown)


config/config.yml
```diff
@@ -107,6 +107,9 @@ PASSWORD_LENGTH_MAX: 40
 # The maximum number of tags you can add to a collection
 COLLECTION_TAGS_MAX: 10
 
+# The maximum number of user-defined tags you can add to a work or external work
+USER_DEFINED_TAGS_LIMIT: 75
```
redsummernight, line 111: Can you use `USER_DEFINED_TAGS_MAX` to be consistent with similar options?

app/models/bookmark.rb
```diff
@@ -21,6 +21,19 @@ def not_already_bookmarked_by_user
     errors.add(:base, ts("You have already bookmarked that."))
   end
 
+  validate :check_new_external_work
+  def check_new_external_work
+    return unless bookmarkable.is_a?(ExternalWork) && bookmarkable.new_record?
+
+    errors.add(:base, "Fandom tag is required") if bookmarkable.fandom_string.blank?
```
redsummernight, line 28: Should we make this a validation on ExternalWork?

app/models/external_work.rb
```diff
@@ -32,6 +32,9 @@ class ExternalWork < ApplicationRecord
                                too_long: ts('^Creator must be less than %{max} characters long.',
                                             max: AUTHOR_LENGTH_MAX)
 
+  validates :user_defined_tags_count,
+            at_most: { maximum: proc { ArchiveConfig.USER_DEFINED_TAGS_LIMIT } }
```
redsummernight, line 36: Can `maximum` be an integer instead of a proc?

app/models/bookmark.rb
tickinginstant: The `ExternalWork`s used for `RelatedWork`s don't have an associated fandom, so it'd be tricky to make it a validation directly on `ExternalWork`s. I think I'd have to add an `attr_accessor` indicating whether the `ExternalWork` was being created from a `Bookmark` or a `RelatedWork`, and then have the validation be conditional.

app/models/external_work.rb
tickinginstant: If I pass the value instead of the proc, stubbing `ArchiveConfig.USER_DEFINED_TAGS_LIMIT` in tests won't change anything, which means that all of the tests would have to be rewritten to use 75-76 tags (which takes more time).

app/models/bookmark.rb
redsummernight: Ah, thanks! `RelatedWork` is what I'm missing. The current setup is fine.

app/models/external_work.rb
redsummernight: Oh I see, the validator never calls on `ArchiveConfig`. proc it is!

redsummernight: Oh I see, the validator never calls on `ArchiveConfig` again after it's loaded once. proc it is!

redsummernight (approved):

sarken (commented):
Could you add a test specifically showing it's not possible to add chapters to a work with too many tags? That's probably the main form of updating we want to prevent, since it takes your 500-tag-work and puts it right back at the top of the work listing, annoying everyone.

tickinginstant:
Hmm, I assumed that the work being invalid would prevent chapters from being created, but although posting does re-open the new chapter form (as though the chapter posting failed), the new chapter does get posted. (Of course, in development, the chapter count cache doesn't get expired, so the work I was testing with still had a 1/1 chapter count, but the chapter I added was there.)

Because the work is invalid and the call to `@work.save` doesn't go through, the work's `revised_at` doesn't change, so it won't go to the top of any work listings, but that can result in inconsistent data (like a chapter count of 2/1, or a work where one of the chapters has a publication date newer than its `revised_at`). I'll try to do something to address the inconsistency.

tickinginstant:
Should be fixed now!

Also, I noticed that you updated the issue to include instructions for testing external works (thank you!), but I actually changed the external works error message from "Your work has 75 of these tags" to "You have entered 75 of these tags." I'm guessing most people bookmark other people's works, not their own, so telling the user to fix the tags on "Your work" feels wrong, and potentially confusing.

Should I revert back to "Your work," even though it doesn't have the right connotation? Or is there some other way you'd prefer to word it?

sarken:
"You have entered" is fine! I was actually thinking I almost preferred it to the one we have for works; I just forgot to update it when I copy/pasted the work error message.

sarken (approved):
Looks good to me now -- thank you very much!

PR #4059, opened 2021-08-01
AO3-6206 Fix work URL in text version of admin hiding work email

niconicosette:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6206


sarken (approved):
That should do it!

PR #4061, opened 2021-08-05
AO3-6153 Don't check validity on collection invite.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6153

## Purpose

Makes it so that the "Add to Collection" button doesn't check the validity of the item being added to the collection.

### DIFF
--- a/features/collections/collection_invite.feature
+++ b/features/collections/collection_invite.feature
@@ -88,10 +88,10 @@ Feature: Collection
   Then I should not see "A Death in Hong Kong"
 
   Scenario: A work with too many tags can be invited to a collection
-    Given the user-defined tag limit is 5
+    Given the user-defined tag limit is 2
       And the collection "Favorites"
       And the work "Over the Limit"
-      And the work "Over the Limit" has 6 fandom tags
+      And the work "Over the Limit" has 3 fandom tags
     When I am logged in as "moderator"
       And I add the work "Over the Limit" to the collection "Favorites"
     Then I should see "This work has been invited to your collection (Favorites)."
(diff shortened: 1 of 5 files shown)


features/collections/collection_invite.feature
```diff
@@ -86,3 +86,12 @@ Feature: Collection
     And 0 emails should be delivered
   When I view the approved collection items page for "anon hidden collection"
   Then I should not see "A Death in Hong Kong"
+
+  Scenario: A work with too many tags can be invited to a collection
+    Given the user-defined tag limit is 5
+      And the collection "Favorites"
+      And the work "Over the Limit"
+      And the work "Over the Limit" has 6 fandom tags
```
redsummernight, line 94: Can we make this test even smaller? 2 fandom tags over a limit of 1.

redsummernight (approved):

features/collections/collection_invite.feature
tickinginstant: The default step for creating a work fills in 1 fandom (`Stargate SG-1`) and 1 freeform (`Scary tag`), so a limit of 1 would break that step. I might be able to do a limit of 2, but I did 5 here because I did 5 with 6 fandom tags in several other tests
https://github.com/otwcode/otwarchive/blob/ccb7f351e1736e120ccd2d092ee8fb1749e4217d/features/bookmarks/bookmark_create.feature#L433-L450
https://github.com/otwcode/otwarchive/blob/ccb7f351e1736e120ccd2d092ee8fb1749e4217d/features/collections/collection_anonymity.feature#L536-L545
https://github.com/otwcode/otwarchive/blob/ccb7f351e1736e120ccd2d092ee8fb1749e4217d/features/works/work_delete.feature#L195-L201
Do you want me to change all of them?

redsummernight: Please and thanks!

PR #4063, opened 2021-08-08
AO3-6038 Prevent replying to comments marked as spam

ahiijny:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6038

## Purpose

This helps resolve the bug where a commenter can accidentally unmark a comment as spam by replying to it.

Note: this was already partially fixed by https://github.com/otwcode/otwarchive/pull/3963, which changed the `check_for_spam` validation to run only when new comments are created.

The main thing this PR adds is a validation check on the controller to ensure that the comment you're replying to (i.e. the "commentable", a.k.a. the "immediate parent", not to be confused with the "parent" a.k.a. "ultimate parent") is not marked as spam. And also updates to the rspec test for this.

(If I understand correctly, there is still a possible race condition where the parent comment is marked as spam in that very narrow window of time _after_ the controller check but _before_ we save the new comment to database. But in this case, it's basically the same as if the parent comment were marked as spam _after_ the reply was submitted.)

### Caveat (and question)

Note: This validation check only checks the immediate parent. If there's a nested comment thread like this:

```
Commenter A: Spammy comment
└─ Commenter B: Oh no
   └─ Commenter C: Oh no
        └─ Commenter D: <typing response...>
```

and if Comment A gets marked as spam, then when Commenter D hits "Reply", the comment will still go through and D will get the "Comment created!" confirmation message. But it will not be visible because A was marked as spam. (This is the same as currently live behaviour.)

Do we want to do anything differently here? Or is it okay to only check the immediate parent?

## Testing Instructions

1. Log in
    1. Post > New Work
    2. Fill in required information
    3. Post
    4. Do not log out or close this window/browser
2. In an incognito/private window or another browser, make sure you are logged out
    1. Go to the work you posted
    2. Leave a comment (this will be the spammy one! :3)
    3. Press the "Reply" button on your spammy comment to open the reply form, and get everything ready to leave a reply comment
    4. Switch back to the window/browser where you are logged in
3. Go back to the window where you are logged in.
    1. Mark the spammy comment as spam
4. Go back to the window/browser where you are not logged in
    1. Press the button to submit your reply comment

### Expected behaviour

1. You should get an error message "Sorry, you can't reply to a comment that has been marked as spam."
2. The spammy comment should remain marked as spam and thus should remain hidden.

## Credit

ahiijny (they/them)

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -22,6 +22,7 @@ class CommentsController < ApplicationController
   before_action :check_parent_comment_permissions, only: [:new, :create, :add_comment_reply]
   before_action :check_unreviewed, only: [:add_comment_reply]
   before_action :check_frozen, only: [:new, :create, :add_comment_reply]
+  before_action :check_not_replying_to_spam, only: [:new, :create, :add_comment_reply]
   before_action :check_permission_to_review, only: [:unreviewed]
   before_action :check_permission_to_access_single_unreviewed, only: [:show]
   before_action :check_permission_to_moderate, only: [:approve, :reject]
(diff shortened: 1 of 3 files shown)


sarken:
Yay, thank you for this! I'd say only checking the immediate parent is fine -- it's entirely possible for a legitimate conversation to start off a spam comment before it was actually marked as such. If the work or admin post creator doesn't want the conversation to go on for some reason, they can always freeze the thread.

ahiijny:
Ah okay, sounds good!

PR #4066, opened 2021-08-13
AO3-6188 Don't ignore any non-ASCII character when renaming tags

setupminimal:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6188

## Purpose

This PR tightens the overly-broad regex that was being used to compare previous and new names for a tag. Before this pull request, no non-ASCII charactetrs were compared, so a tag wrangler could rewrite "ø" to "てりやき".

Now, only characters that are latin diacritical marks are stripped.

## Testing Instructions

To test this, please try to rename a tag and ensure that, for example, "ø" can be renamed to "o", but not to "てりやき".

## References

There are no other references that I'm aware of.

## Credit

Daroc Alden, they/them

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -38,6 +38,7 @@ gem 'redis-namespace'
 # Used to convert strings to ascii
 gem 'unicode'
 gem 'unidecoder'
+gem 'unicode_utils', '>=1.4.0'
 
 # Lograge is opinionated, very opinionated.
 gem "lograge" # https://github.com/roidrage/lograge
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -187,12 +187,16 @@ class Tag < ApplicationRecord
     self.errors.add(:unwrangleable, "can't be set on an unsorted tag.") if is_a?(UnsortedTag)
   end
 
+  def normalize_for_tag_comparison(string)
+    UnicodeUtils.casefold(string).mb_chars.normalize(:kd).gsub(/[\u0300-\u036F]/u, "")
+  end
+
   before_validation :check_synonym
   def check_synonym
     if !self.new_record? && self.name_changed?
       # ordinary wranglers can change case and accents but not punctuation or the actual letters in the name
       # admins can change tags with no restriction
-      self.errors.add(:name, "can only be changed by an admin.") unless User.current_user.is_a?(Admin) || (self.name.downcase == self.name_was.downcase) || (UnicodeUtils.casefold(self.name).mb_chars.normalize(:kd).gsub(/[\u0300-\u036F]/u, "").to_s == UnicodeUtils.casefold(self.name_was).mb_chars.normalize(:kd).gsub(/[\u0300-\u036F]/u, "").to_s)
+      self.errors.add(:name, "can only be changed by an admin.") unless User.current_user.is_a?(Admin) || self.normalize_for_tag_comparison(self.name) == self.normalize_for_tag_comparison(self.name_was)
     end
     if self.merger_id
       if self.canonical?
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -190,9 +190,7 @@ class Tag < ApplicationRecord
     if !self.new_record? && self.name_changed?
       # ordinary wranglers can change case and accents but not punctuation or the actual letters in the name
       # admins can change tags with no restriction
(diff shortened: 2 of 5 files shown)


setupminimal:
Hmm. The unit test in tag_spec.rb near line 145 asserts that "Wei Kreuz" can be renamed into "Weiß Kreuz". I don't speak German, so I'll defer to anyone with actual experience, but [Wiktionary](https://en.wiktionary.org/wiki/wei%C3%9F) seems to suggest that "Wei" and "Weiß" are different words.

I'm going to change that unit test, but I'd be equally happy to expand the regex and/or to actually use Unicode data for each character to do this 'properly' if that would be preferable.

Gemfile
```diff
@@ -38,6 +38,7 @@ gem 'redis-namespace'
 # Used to convert strings to ascii
 gem 'unicode'
 gem 'unidecoder'
+gem 'unicode_utils', '>=1.4.0'
```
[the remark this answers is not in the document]
setupminimal: I'm going to ignore this to be consistent with the rest of the Gemfile.

setupminimal:
Hound seems to be suggesting a lot of changes that don't fit with the style used in the rest of the file. Would you like me to update my changes to comply, or ignore Hound?

sarken:
Thank you for working on this! I haven't reviewed it yet, but I've asked for one of our German-speaking team members to take a look at your question about the tag. 

I've also taken a look at the Hound comments and they're correct -- we do prefer to avoid instance variables and use of the word "should." (I don't think we have strong feelings about the encoding comment.) If you could update your tests to comply, that would be great!

spec/models/tag_spec.rb
```diff
@@ -123,53 +121,69 @@
...
 
-        @tag.name = "Amélie"
-        @tag.check_synonym
-        expect(@tag.errors).to be_empty
-        expect(@tag.save).to be_truthy
+        tag.name = "Weiß Kreuz"
+        tag.check_synonym
+        expect(tag.errors).not_to be_empty
+        expect(tag.save).to be_falsey
       end
```
sarken, line 154: I've spoken with the tag wrangling team and some German-speakers, and you're correct: Weiss/Weiß and Wei are quite different, and you should definitely not be able to add an extraneous ß to Wei (or any other word 😆). 

Honestly, the history of this example is so baffling that I would suggest removing it completely. The `ignores the capitalization of ß` example is much more useful and less confusing.

app/models/tag.rb
```diff
@@ -190,9 +192,7 @@ def check_synonym
     if !self.new_record? && self.name_changed?
       # ordinary wranglers can change case and accents but not punctuation or the actual letters in the name
       # admins can change tags with no restriction
-      unless User.current_user.is_a?(Admin) || (self.name.downcase == self.name_was.downcase) || (self.name.mb_chars.normalize(:kd).gsub(/[^\x00-\x7F]/u,'').downcase.to_s == self.name_was.mb_chars.normalize(:kd).gsub(/[^\x00-\x7F]/u,'').downcase.to_s)
-        self.errors.add(:name, "can only be changed by an admin.")
-      end
+      self.errors.add(:name, "can only be changed by an admin.") unless User.current_user.is_a?(Admin) || (self.name.downcase == self.name_was.downcase) || (UnicodeUtils.casefold(self.name).mb_chars.normalize(:kd).gsub(/[\u0300-\u036F]/u, "").to_s == UnicodeUtils.casefold(self.name_was).mb_chars.normalize(:kd).gsub(/[\u0300-\u036F]/u, "").to_s)
```
sarken, line 195: For readability, would it be worthwhile to move the normalization process into its own method?

`(UnicodeUtils.casefold(self.name).mb_chars.normalize(:kd).gsub(/[\u0300-\u036F]/u, "").to_s` is a lot, especially twice in a row. 

@zz9pzza also wanted me to pass along a question: is the `to_s` still necessary?

setupminimal: I've refactored it out into a method `normalize_for_tag_comparison`. I agree that's a lot more readable.

zz9pzza is also totally right: the `to_s` was redundant, so I've removed that.

Thank you!

spec/models/tag_spec.rb
setupminimal: I have removed the `ignores the capitalization of ß` test. I believe the remaining tests cover the expected behavior well enough.

setupminimal:
> Just one tiny question, but otherwise looks good to me!

For whatever reason, GitHub isn't letting me reply to your actual question. As for the encoding line -- that's just what emacs' ruby-mode did because the source file contains Unicode characters. I didn't mean to edit it, but it gets put in there on each save of the file and I missed the change.

I've just pushed a commit that reverts that line, so it should be gone now.

sarken (approved):

PR #4067, opened 2021-08-16
AO3-6120 Add our IPs and secure URL to PERMITTED_HOSTS to hide proxy notice

zz9pzza:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6120

## Purpose

The warning should not occur on https://secure.archiveofourown.org/ 

## Testing Instructions

This can't be tested on staging, apart from noting that we have not got the warning on https://test.archiveofourown.org

### DIFF
--- a/config/config.yml
+++ b/config/config.yml
@@ -457,6 +457,9 @@ FILTER_UPDATE_BATCH_SIZE: 100
 # environment.
 PERMITTED_HOSTS: [
   # Production
+  "104.153.64.122",
+  "208.85.241.152",
+  "208.85.241.157",
   "archiveofourown.org",
   "download.archiveofourown.org",
   "insecure.archiveofourown.org",
--- a/config/config.yml
+++ b/config/config.yml
@@ -460,6 +460,7 @@ PERMITTED_HOSTS: [
   "archiveofourown.org",
   "download.archiveofourown.org",
   "insecure.archiveofourown.org",
+  "secure.archiveofourown.org",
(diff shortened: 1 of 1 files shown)


config/config.yml
```diff
@@ -460,6 +460,7 @@ PERMITTED_HOSTS: [
   "archiveofourown.org",
   "download.archiveofourown.org",
   "insecure.archiveofourown.org",
+  "secure.archiveofourown.org",
```
sarken, line 463: Can we add the IP address mentioned in the issue too? Or is there a reason we shouldn't do that?

zz9pzza: I am moving over to a new ip address before we balance over a new pair, I don't really support the bare ip address although we could for http but not https.

https://208.85.241.152/ is a security warning as expected. I could make http://208.85.241.152/  redirect to https:archiveofourown.org/ I think.

If I did that we still wouldn't want the ip address in it ?

sarken: Support says people use the IP address to get around DNS poisoning. Would redirecting affect their ability to do that?

zz9pzza: I will add the 2 ip's then.

sarken (approved):

config/config.yml
sarken: fwiw, I'm getting 503s on the IP addresses (using HTTP), but I assume you're aware of that

PR #4072, opened 2021-08-29
AO3-5926 Fix chapter re-positioning when deleting chapter one

kwboyd:
~~I'm pretty sure this is a fix for both issues - I'll do my best in JIRA to make note of that.~~ (Initially thought this would fix 2247 as well, but it doesn't quite cover all edge cases from that issue).

# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5926

## Purpose

This fixes the `fix_positions_before_destroy` method in the `before_destroy` chapter hook to make all subsequent chapters go down one position instead of up one.

**Extra details if anyone is curious:**
I stepped through the code a bit to figure out why this wasn't causing issues when a chapter besides the first one was deleted... as it turns out, the `fix_positions` method called in a chapter's `after_save` hook combined with the `save_chapters` hook on works was doing some self-healing for this mis-positioning, except in the case of the first chapter being destroyed.

The reason chapters were getting swapped when you deleted the first chapter (Issue 2247) was because after `fix_positions_before_destroy` is called, that `fix_positions` method moves the chapter immediately following the deleted one like this: 
```
chapters = chapters - [self]
chapters.insert(self.position-1, self)
```

So if you've deleted chapter 1, the new chapter 1/old chapter2 was having its position bumped up to 3. So when `fix_positions` did that `.insert`, it was putting chapter 2 at `chapters[2]`, when it should be put at `[0]`, so it was swapping places with the chapter right after it. This wasn't a problem for deleting mid-work chapters, since essentially there was an empty space before the subsequent chapters. 

## Testing Instructions

Testing instructions are in the JIRA issue.

## References

https://otwarchive.atlassian.net/browse/AO3-2247

## Credit

Kate Boyd, she/her

### DIFF
--- a/spec/controllers/chapters_controller_spec.rb
+++ b/spec/controllers/chapters_controller_spec.rb
@@ -945,13 +945,11 @@ describe ChaptersController do
       end
 
       context "when work has more than one chapter" do
-        before do
-          @chapter2 = create(:chapter, work: work, posted: true, position: 2, authors: [user.pseuds.first])
-        end
+        let!(:chapter2) { create(:chapter, work: work, posted: true, position: 2, authors: [user.pseuds.first]) }
 
         it "updates the work's minor version" do
           expect(work.minor_version).to eq(0)
-          delete :destroy, params: { work_id: work.id, id: @chapter2.id }
+          delete :destroy, params: { work_id: work.id, id: chapter2.id }
           expect(assigns[:work].minor_version).to eq(1)
         end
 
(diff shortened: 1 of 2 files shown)


spec/controllers/chapters_controller_spec.rb
```diff
@@ -945,13 +945,11 @@
       end
 
       context "when work has more than one chapter" do
-        before do
-          @chapter2 = create(:chapter, work: work, posted: true, position: 2, authors: [user.pseuds.first])
-        end
+        let!(:chapter2) { create(:chapter, work: work, posted: true, position: 2, authors: [user.pseuds.first]) }
```
kwboyd, line 948: Hound yelled at me to make this change 🐕

sarken (approved):
This looks good to me! 

I have to admit, though, I'm kind of suspicious of anything chapter-related that's this easy to fix. 😂 I keep wondering why it was originally set to `+ 1` -- was it really just a typo that's stuck around causing trouble for 11 years? 😬 

It looks like the method was also misnamed `fix_positions_after_destroy` until 2 years ago, so it's definitely possible, but I would like a second reviewer to approve this as well and flip the label.

kwboyd:
@sarken Thanks for the review! I 100% get the suspicion, this does seem like a strangely simple fix 😬 

For reference for the next reviewer, here's the commit that added this hook originally: https://github.com/otwcode/otwarchive/commit/6b1a25617cf459e8d54f018ee0184a13b265e365

spec/controllers/chapters_controller_spec.rb
```diff
@@ -962,27 +960,58 @@
...
         end
 
         it "errors and redirects to work when chapter is not deleted" do
           allow_any_instance_of(Chapter).to receive(:destroy).and_return(false)
-          delete :destroy, params: { work_id: work.id, id: @chapter2.id }
+          delete :destroy, params: { work_id: work.id, id: chapter2.id }
           it_redirects_to_with_error(work, "Something went wrong. Please try again.")
         end
+
+        it "does not reorder chapters when deleting the last chapter" do
```
kwboyd, line 985: For reference: this first test and the last test (`"reorders chapters properly when deleting a mid-work chapter"`) pass on both this branch and master. The two in the middle around deleting the first chapter pass on this branch but fail on master.

elzj:
+1 from me - thanks for catching that!

tickinginstant:
This handles AO3-5926 fine. However, there are some edge cases involving reordered chapters that prevent this from fully addressing AO3-2247. For example, try creating a work with 4 chapters (A, B, C, D), and move the first chapter into the third position (B, C, A, D). Then try deleting the new first chapter (B). The new chapter order ought to be C, A, D, but instead it's C, D, A.

If you wanted to tackle AO3-2247, as opposed to making this PR cover only AO3-5926, I think it'd be good to rewrite `fix_positions_before_destroy` to try to reduce the complexity of the code. There might be some fixes that would address AO3-2247 without a complete rewrite, but I feel like I'd need to write a formal proof of correctness to be able to properly review any of them, and I don't think it's healthy to have code that tricky in the codebase.

(Currently, in a work with 10 chapters, deleting the first chapter will cause the `fix_positions` hook to be called a total of 9 times -- 10 times if you include the subsequent call to `@work.save`. That's part of why this code is so tricky to debug and analyze -- only the results of the last position calculation are visible to the user, but each recalculation is affected by the one before it. For the sake of both speed and comprehensibility, there should be only one recalculation. Maybe two, if you count the one for `@work.save`.)

kwboyd:
@tickinginstant Ahhh thank you, I did miss that edge case!

The one quick-fix I can see to make this address 2247 as well would be changing `chapters.each { |c| c.update_attribute(:position, c.position - 1) }` to `chapters.each { |c| c.decrement!(:position) }`, since `decrement!` wouldn't trigger any callbacks on `Chapter`, and it looks like that chapter-swapping bug happens _somewhere_ in one of the `fix_positions` calls. I'm happy to make that change or leave this as-is if you think that could have tricky-to-test side-effects. It doesn't look to me like any of the other callbacks would be necessary to run on chapter deletion, but I'm also too new to this codebase to say that for sure!

If that `decrement` fix doesn't sound like the way to go, then I'll update this PR and the JIRA issues to clarify that it only fixes 5926. I'd be happy to attempt a more thorough refactor of the `fix_positions_before_destroy` and `fix_positions` code in a separate PR, though it'll probably take me a few weeks to make sure I've tested it thoroughly enough :)

tickinginstant:
Unfortunately, skipping callbacks would also skip one of the important parts of `fix_positions`: the cache-busting on the work's comments.

kwboyd:
@tickinginstant Got it, I see that now. Thanks for pointing that out! I updated this PR and the JIRA issues to reflect that this only fully fixes 5926.

PR #4070, opened 2021-08-22
AO3-5372 Add padding to user bookmarks

kwboyd:
# Pull Request Checklist

* [x] Have you read ["How to write the perfect pull request"](https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/)?
* [x] Have you read the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [x] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)?
* [x] Have you added the [Jira](https://otwarchive.atlassian.net) issue number
  as the *first* thing in your pull request title (e.g. `AO3-1234 Fix thing`)
* [x] Do you have fewer than 5 pull requests already open? If not, please wait
  until they are reviewed and merged before creating new pull requests.

## Issue

https://otwarchive.atlassian.net/browse/AO3-5372

## Purpose

This adds padding to the list of bookmarks on the recent bookmarks page `/bookmarks/` and the user bookmarks page `/users/${userName}/bookmarks`.

## Testing Instructions

**New changes:**
1. Go to the `/bookmarks` page. There should be padding within the bookmark so the content isn't directly up against the borders anymore.
2. Go to the bookmark's list for any specific user (`/users/${userName}/bookmarks`). Same check as above ^
3. Bookmark a work. The gray "bookmarked by `{myUsername}`" bookmark should also have padding now.

**Things that should be unchanged:**
1. Go to the bookmarks for any specific work (`works/${workId}/bookmarks`). The padding within those bookmarks should be unchanged.
2. Bookmark a work. Click "Show Most Recent Bookmarks". The padding within those bookmarks should also be unchanged.

## References

Are there other relevant issues/pull requests/mailing list discussions?
Not that I know of.

## Credit

What name and pronouns should we use to credit you in the [Archive of Our Own's Release Notes](https://archiveofourown.org/admin_posts?tag=1)?
Kate Boyd, she/her

This is my first PR to the archive, I appreciate your patience if I've missed a step!

PR #4077, opened 2021-09-03
AO3-5372 Fix small-screen bookmark byline wrapping, increase top/right on bookmark datetimes

kwboyd:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5372

## Purpose

This addresses the QA feedback left on the JIRA issue. With the margin-right in place for bookmark bylines, they should wrap on small screens now. I also increased the top and right properties on the datetimes in recent and user bookmarks - I forgot about how the datetimes having a smaller font-size would mean their `em` value is lower, oops. They should better match the positioning of the datetimes in bookmarks with an icon now.

Screenshots that have both types of bookmarks!
Screen Shot 2021-09-02 at 8 32 29 PM
Screen Shot 2021-09-02 at 8 32 59 PM

## Testing Instructions

In JIRA.

## References

This is addressing QA feedback on https://github.com/otwcode/otwarchive/pull/4070

## Credit

What name and pronouns should we use to credit you in the [Archive of Our Own's Release Notes]
Kate Boyd, she/her

### DIFF
--- a/public/stylesheets/site/2.0/13-group-blurb.css
+++ b/public/stylesheets/site/2.0/13-group-blurb.css
@@ -369,8 +369,9 @@ blurbs on the manage collection items pages, mostly reseting styles inherited fr
 	min-height: 30px;
 }
 
-.bookmark .user .heading {
+.bookmark .user .byline {
   margin-top: 0.375em;
+  margin-right: 5.25em;
 }
 
 .bookmark .user .datetime, .bookmark .work .datetime {
@@ -378,8 +379,8 @@ blurbs on the manage collection items pages, mostly reseting styles inherited fr
 }
 
 .bookmark div.user .datetime {
-  top: 0.75em;
-  right: 0.75em;
+  top: 0.875em;
+  right: 0.875em;
 }
 
 .bookmark .short .status {


public/stylesheets/site/2.0/13-group-blurb.css
```diff
@@ -369,17 +369,18 @@ blurbs on the manage collection items pages, mostly reseting styles inherited fr
 	min-height: 30px;
 }
 
-.bookmark .user .heading {
+.bookmark .user .byline.heading {
```
sarken, line 372: For the sake of simplicity (quite necessary with as many CSS files as we have! 😆) we generally try to avoid chaining class names. Is there a specific page where there's a problem if we just use `.bookmark .user .byline`?

sarken (commented):
Just one quick question!

public/stylesheets/site/2.0/13-group-blurb.css
kwboyd: Haha makes sense! It's been a while since I've worked in pure CSS and I forget how careful you've gotta be with the selectors sometimes. I looked around and it seems like just `.bookmark .user .byline` works completely fine, so I just pushed an update to make it that. Thanks!

sarken (approved):
Yay, thank you!

PR #4078, opened 2021-09-03
AO3-6168 Introduce kudos heading

korrien:
# Pull Request Checklist

We do not test for visual changes

## Issue

https://otwarchive.atlassian.net/browse/AO3-6168

## Purpose

Wanted to fix the issue as it's a consequence of my old PR 😅 
Issue was set as Medium difficulty, but with the great suggestions in the description I can't find anything else that needs fixing. Might be wrong as usual 😄 
- all looks the same with javascript disabled
- heading doesn't directly have `id="kudos"`, but that's ok (I believe id is used now only when navigating directly to kudos section from work blurb and for rendering new kudo)
- I can't figure out how to apply other existing skins other than Default and Archive 2.0, but I believe they are not affected as they mainly deal with color schemes

## References

Before [AO3-4008](https://otwarchive.atlassian.net/browse/AO3-4008)
image

After [AO3-4008](https://otwarchive.atlassian.net/browse/AO3-4008)
image

Now - no messages
image
Now - with messages
image

## Credit

korrien, she/her

### DIFF
--- a/app/views/kudos/_kudos.html.erb
+++ b/app/views/kudos/_kudos.html.erb
@@ -1,4 +1,4 @@
-<%# expects local variables kudos, guest_kudos_count, commentable %>
+<% # expects local variables kudos, guest_kudos_count, commentable %>
 <h3 class="landmark heading"><%= ts("Kudos") %></h3>
 <div id="kudos">
   <% has_user_kudos = kudos.exists? %>
--- a/app/views/kudos/_kudos.html.erb
+++ b/app/views/kudos/_kudos.html.erb
@@ -1,4 +1,5 @@
-<% # expects local variables kudos, guest_kudos_count, commentable %>
+<%# expects local variables kudos, guest_kudos_count, commentable %>
+<h3 class="landmark heading"><%= ts("Kudos") %></h3>
 <div id="kudos">
   <% has_user_kudos = kudos.exists? %>
   <% if has_user_kudos || guest_kudos_count > 0 %>


app/views/kudos/_kudos.html.erb
```diff
@@ -1,4 +1,5 @@
-<% # expects local variables kudos, guest_kudos_count, commentable %>
+<%# expects local variables kudos, guest_kudos_count, commentable %>
```
sarken, line 1: Could you add the space back between `%` and `#`?

sarken (commented):
Woohoo, thank you! Just one tiny comment. 

>Issue was set as Medium difficulty, but with the great suggestions in the description I can't find anything else that needs fixing. Might be wrong as usual 😄

Nah, I think you're fine! Medium is just our default and it doesn't always get changed when it should.

>I can't figure out how to apply other existing skins other than Default and Archive 2.0, but I believe they are not affected as they mainly deal with color schemes

You're right; they should indeed be fine. Unfortunately, it's been a long time since I looked at a dev environment with the basic data, so I'm not 100% sure what the situation is with the other skins. 

If they're not listed on localhost:3000/skins, you might need to log in as an admin and approve them from localhost:3000/admin/skins (and check "Chooser" if you want them listed in the footer).  

If they're not on the approval page either, your best bet might be to
* log in as a user
* go to localhost:3000/skins/new?skin_type=Skin
* give your skin a title
* check the "Apply to make public" box
* copy and paste [the desired CSS from the repo](https://github.com/otwcode/otwarchive/tree/master/public/stylesheets/masters) into the CSS box
* press "Submit"

That should put it in the admin approval queue, where you can make it a public skin available to any user on your dev environment and/or add it to the chooser in the footer.

app/views/kudos/_kudos.html.erb
korrien: Done, but may I ask why?
erb linter was complaining about that space.
I also found [several](https://stackoverflow.com/questions/2774841/best-way-to-add-comments-in-erb#comment2811714_2774858) [posts](https://stackoverflow.com/a/3901665) saying that propper comment syntax is `<%#`

And thanks for the skin tips! Had to add all of them by copying, but all looks good.

sarken: Sure! I don't know if there's a particular reason we chose it, but the space is our standard -- we have one use without the space and 260 with it -- so I just asked for consistency's sake. 

And I'm glad the skin tips worked!

sarken (approved):

PR #4082, opened 2021-09-08
AO3-6233 Highlight autocomplete suggestions with diacritics

tlee911:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6233

## Purpose

Fixes the autocomplete highlighting to handle diacritic marks so that matching substrings in suggestions are bolded regardless of diacritic marks.  Makes the changes from #4009 more obvious to the user.

Screen Shot 2021-09-08 at 12 16 56 AM
Screen Shot 2021-09-08 at 12 17 39 AM

## Testing Instructions

Cucumber tests have been added to `autocomplete.feature`.

For manual testing:
- Create canonical tags of various types, both with and without diacritics
- Reload the tag data for the autocomplete `rake autocomplete:reload_tag_data`
- Go to the form to post a new work (or edit an existing one)
- Enter some values in the tag fields and ensure the autocomplete brings up suggestions with the appropriate parts highlighted.

The updated method for highlighting could be slow if there are a lot of long potentially matching tags.  Performance seems reasonable to me, but that might be dependent on device, and I did not run a test with more than a few tag suggestions at a time.

I've also incidentally changed a behaviour which I think was unintentional before.  Now only the first occurence of match in a suggestion will be highlighted.  For example, if you type `f` and "Fluff" is in the suggestions, only the first "F" will be bolded, instead of all three.  I think this makes more sense as the autocomplete is only keying in on matches at the start of words anyway.

## References

#4009


tickinginstant:
This code appears to treat e.g. `â` and `î` similarly (replacing both with a `.` character). What happens if you have a canonical tag `Pîtter Pâtter` and search for `Pâtter` or `Patter`? Can it tell the difference, or will it highlight `Pîtter` instead of `Pâtter`?

(Also, you might want to avoid `let`. I think that might be why you're getting javascript errors in the automated tests -- phantomjs isn't really up to the standards of modern browsers.)

tlee911:
> This code appears to treat e.g. `â` and `î` similarly (replacing both with a `.` character). What happens if you have a canonical tag `Pîtter Pâtter` and search for `Pâtter` or `Patter`? Can it tell the difference, or will it highlight `Pîtter` instead of `Pâtter`?

You are correct, it will highlight the first occurrence in that situation.  This is my workaround for dealing with inputs that do not include diacritics but trying to highlight tags which do have the diacritics without having something like transliterate available in the javascript.  I feel this would be acceptable since only the highlighting will be off, the autocomplete itself would suggest only tags where the transliteration matched.  As compared to the current behaviour of no highlighting except for exact matches, this is still an improvement.  Of course, I'm open to any suggestions for a better way of doing this, as this is the best I could come up with so far.

> (Also, you might want to avoid `let`. I think that might be why you're getting javascript errors in the automated tests -- phantomjs isn't really up to the standards of modern browsers.)

Thanks, let me try that.  All the tests worked locally for me, so I had thought it was something else.

bingeling:
tickinginstant makes a good point, we don't want to highlight the wrong thing. It looks to me like a better way would be to use [String.prototype.normalize()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize), which is ES6, but [a polyfill seems to exist](https://github.com/walling/unorm). Could you look into if this is a workable solution?

tickinginstant:
I don't think `String.prototype.normalize()` will handle non-diacritics like `ø`. If you do want to handle those as well as diacritical marks, something like [this](https://github.com/valudio/diacritic-regex-helper/blob/master/index.js) seems like a fairly lightweight approach, although it can't handle two-character sequences like `æ`. (It's also missing some equivalences from [the default i18n mappings in Rails](https://github.com/ruby-i18n/i18n/blob/3d0f144038eebe3ff6bc7699de00736abd5ef634/lib/i18n/backend/transliterator.rb#L43-L72).)

Alternatively, maybe the highlighting could be done server-side, so that you have access to `transliterate`?

bingeling:
Hm, the issue specifically mentions diacritics, so I think we only need to cover those. Looking for stuff with `æ` or `Ø` with our current autocomplete code on production gives the right results and the highlighting works. If I understand correctly, the issue is specifically with Unicode characters with decomposition properties. Whereas `Ø` doesn't have any.

tickinginstant:
The pull request (#4009) that changes the autocomplete behavior for both diacritics and characters like `ø` hasn't been deployed to production yet, so the issues with autocomplete highlighting in production don't give a complete perspective of what the issue is. On staging, where the changes have been made, you can e.g. search for the tag "Original Work" by entering `Øriginal` in the autocomplete, but the highlighting doesn't reflect the fact that the autocomplete views `O` and `Ø` as the same character.

(AO3-2380 also mentions only diacritics, but in #4009 it was decided that the issue should also encompass characters like `ø`. And since AO3-6233 was created to fix the issues with highlighting introduced by #4009, I think `ø` is in scope.)

bingeling:
Okay, I see. Thanks for explaining this to me, @tickinginstant. In this case, @tlee911, the way to go is to do the highlighting server-side. It's the best way to ensure search and highlighting are using the same logic and to always match up.

tlee911:
I will close this PR as the desired approach doesn't fit this implementation.

PR #4069, opened 2021-08-19
AO3-6220 Update collection FAQ links in emails, TOS, and TOS FAQ

sharra44:
# Pull Request Checklist

* [x] Have you read ["How to write the perfect pull request"](https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/)?
* [x] Have you read the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [x] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)?
* [x] Have you added the [Jira](https://otwarchive.atlassian.net) issue number
  as the *first* thing in your pull request title (e.g. `AO3-1234 Fix thing`)
* [x] Do you have fewer than 5 pull requests already open? If not, please wait
  until they are reviewed and merged before creating new pull requests.

## Issue

https://otwarchive.atlassian.net/browse/AO3-6220

## Purpose

Updates outdated FAQ references for the collections feature

## Testing Instructions

    Create a work

    Create a collection

    Add the work to the collection

    Change the collection status to 'unrevealed' or to 'anonymous' (or both)

    Follow the “Manage Items” link in the collection sidebar

    Follow the “Approved” link near the top of the page

    Check the “Anonymous” or “Unrevealed” (or both) boxes for the work you created

    Press “Submit”

## References

N/A

## Credit

Sharra Neely, she/her

PR #4083, opened 2021-09-10
AO3-6220 Update remaining collection FAQ links

korrien:
# Pull Request Checklist

No tests added. As far as I understand, we do not verify exact content of sent emails and do not test any user-facing large bodies of text like TOS FAQ

## Issue

https://otwarchive.atlassian.net/browse/AO3-6220

## Purpose

Added more changes to existing ones as per information from tester.

- I believe FAQ itself is not available on development environments, so couldn't check the links by following them, but they look consistent with FAQ provided on AO3.
- both text/html is updated in emails (English version as the only one present in the source code)
- could check it only through developer log, not actual email (didn't figure out which setting to change, `PERFORM_DELIVERIES: true` seemed to have no effect).
- I assume we supply two versions of email (text/html) as we want to support email clients which do not support html

Only remaining "Collections and Challenges" labels that are still left are on the site_map, but they seem reasonable enough and anyway are not covered by the FAQ issue.
Search for "collections-and-challenges" doesn't give any results now.

## References

The original PR #4069.

## Credit

korrien, she/her

### DIFF
--- a/app/views/home/tos_faq.html.erb
+++ b/app/views/home/tos_faq.html.erb
@@ -411,7 +411,7 @@ Please use our search functions for this rather than creating a separate work.</
 			<h3 class="heading" id="specialized_faq">Assorted Specialized Policies</h3>
 			<h4 class="heading" id="collections">Collections</h4>
 			<h5 class="heading" id="collections_def">What do you mean by "collections"?</h5>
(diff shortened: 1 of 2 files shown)


sarken (approved):
Thank you!

>As far as I understand, we do not verify exact content of sent emails and do not test any user-facing large bodies of text like TOS FAQ

Yup, correct.

>I believe FAQ itself is not available on development environments, so couldn't check the links by following them, but they look consistent with FAQ provided on AO3.

Also correct.

> could check it only through developer log, not actual email (didn't figure out which setting to change, PERFORM_DELIVERIES: true seemed to have no effect).

Yeah, afaik, the development log is the only way to test the emails. Well. That or forcing the specs to fail by looking for text the emails don't contain, which will also output the contents where you can easily see it. 

> I assume we supply two versions of email (text/html) as we want to support email clients which do not support html

Also correct!

PR #4087, opened 2021-09-17
AO3-6225 Make comparison of email addresses case-insensitive

setupminimal:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6225

## Purpose

This PR makes it so that the email address confirmation is compared case-insensitively when a user changes their email.

## Testing Instructions

Follow the Steps to Reproduce from the JIRA ticket, and ensure that no error message is seen.

## References

None that I'm aware of.

## Credit

Daroc Alden, they/them

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -170,17 +170,14 @@ class UsersController < ApplicationController
     if !params[:new_email].blank? && reauthenticate
       new_email = params[:new_email]
 
-      # Please note: This comparison is not technically
-      # correct. According to RFC 5321, the local part of
-      # an email address is case sensitive, while the
-      # domain is case insensitive. That said, all major
-      # email providers treat the local part as case
-      # insensitive, so it would probably cause more confusion
-      # if we did this correctly.
+      # Please note: This comparison is not technically correct. According to
+      # RFC 5321, the local part of an email address is case sensitive, while the
+      # domain is case insensitive. That said, all major email providers treat
+      # the local part as case insensitive, so it would probably cause more
+      # confusion if we did this correctly.
       #
-      # Also, email addresses are validated on the client, and
-      # will only contain a limited subset of ASCII, so we don't
-      # need to do a unicode casefolding pass.
+      # Also, email addresses are validated on the client, and will only contain
+      # a limited subset of ASCII, so we don't need to do a unicode casefolding pass.
       if new_email.downcase != params[:email_confirmation].downcase
         flash.now[:error] = ts("Email addresses don't match! Please retype and try again")
         render :change_email and return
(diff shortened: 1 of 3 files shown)


app/controllers/users_controller.rb
```diff
@@ -170,7 +170,18 @@ def changed_email
     if !params[:new_email].blank? && reauthenticate
       new_email = params[:new_email]
 
-      if new_email != params[:email_confirmation]
+      # Please note: This comparison is not technically
```
redsummernight, line 173: You can wrap your comments at around 80 characters, like a few other places in this file. I think this is a bit narrow.

Thanks for looking into this! I didn't know that the local part of emails is supposed to be case sensitive. The pragmatic choice makes sense.

features/other_a/profile_edit.feature
```diff
@@ -44,6 +44,22 @@ Scenario: Changing email address - entering an invalid email address
   Then I should see "Email does not seem to be a valid address"
     And 0 emails should be delivered
 
+Scenario: Changing email address - case-insensitive comparison
```
redsummernight, line 47: Can you change the last part to "case-insensitive confirmation"?

features/other_a/profile_edit.feature
```diff
@@ -44,6 +44,22 @@ Scenario: Changing email address - entering an invalid email address
...
+
+  When I follow "Change Email"
+    And I fill in "New Email" with "foo@example.com"
+    And I fill in "Confirm New Email" with "FoO@example.com"
+    And I enter my password
```
redsummernight, line 52: You can use `And I fill in "Password" with "password"` instead of adding a new step here.

app/controllers/users_controller.rb
```diff
@@ -170,7 +170,15 @@ def changed_email
...
+      # Please note: This comparison is not technically correct. According to
+      # RFC 5321, the local part of an email address is case sensitive, while the
+      # domain is case insensitive. That said, all major email providers treat
+      # the local part as case insensitive, so it would probably cause more
+      # confusion if we did this correctly.
+      #
+      # Also, email addresses are validated on the client, and will only contain
+      # a limited subset of ASCII, so we don't need to do a unicode casefolding pass.
+      if new_email.downcase != params[:email_confirmation].downcase
         flash.now[:error] = ts("Email addresses don't match! Please retype and try again")
```
redsummernight, line 182: One last thing I just noticed, can you add a period to the end of this error message? Otherwise this looks good.

setupminimal: Certainly, good catch!

redsummernight (approved):
Thanks!

PR #4090, opened 2021-09-24
AO3-6109 Update list of locales to pull from Phrase

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6109

## Purpose

Make sure we're pulling locales from Phrase into files that match the locale ISO codes in our database, and then make sure the application configuration also matches the ISO codes.

~Should we wipe out the default locale files? They're now included in the files exported from Phrase, plus they no longer match the available_locales.~


.phrase.yml
```diff
@@ -88,12 +91,12 @@ phrase:
       - file: ./config/locales/phrase-exports/ja.yml
         params:
           locale_id: ja
-      - file: ./config/locales/phrase-exports/ka.yml
```
redsummernight, line 91: We should update the defaults to match, which means:

- removing config/locales/defaults/ka.yml
- renaming config/locales/defaults/mr-IN.yml
- removing config/locales/defaults/tl.yml

fil.yml and ky.yml are not provided by rails-i18n or we'd be adding those as well.

sarken: Wait, I'm confused. I thought Translation started using `tl` rather than `fil` on Phrase specifically to take advantage of `config/locales/defaults/tl.yml`?

redsummernight: I see. In that case we should rename tl.yml to its new name. Can you also add the rename cases to https://github.com/otwcode/otwarchive/blob/master/config/locales/README.md so we don't lose the changes in future rails-i18n updates?

sarken: Yup, already planning to! This is not a headache I want to repeat 😆

ChasNord (approved):

.phrase.yml
sarken: Ugh, one more thought: doesn't renaming these files mean that when we import to Phrase, the defaults will be added to Phrase as fil and mr, rather than updating what's already on Phrase as tl and mr-IN?

redsummernight: We don't push the defaults files after #4006.

sarken: Right, sorry, I'm thinking about the hypothetical catastrophe we mention in the README rather than routine work. I suppose we could just manually rename the files before uploading if it ever happens, though.

redsummernight: I see what you mean. In Phrase's UI looks like we can use "Edit Language" to change the name and code. We should try similarly renaming there as well, e.g. `tl` to `fil`, that way the pull settings can now be consistent:
```yml
      - file: ./config/locales/phrase-exports/fil.yml
        params:
          locale_id: fil  
```

sarken: I'll double check with Translation that that'll work for them.

sarken: Done on their end and done on my end!

PR #4091, opened 2021-09-24
AO3-5127 Do not show unapproved collections in mystery blurb

korrien:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5127

## Purpose

- Removes some anonymous mystery blurb code, which seems unused
- Updates frontend to show only approved unrevealed collections in Mystery Blurb

## Credit

korrien, she/her


sarken:
ticking raises a good point, and I think it's definitely in scope to do something about that awkwardness. For now, if there are no approved collections, let's not include anything where the "Part of" text would be.

korrien:
Yes, thank you! I should have thought of it.

Now it looks like this:

app/views/works/_mystery_blurb.html.erb
```diff
@@ -1,15 +1,11 @@
 <div class="mystery header picture module">
   <h4 class="heading"><%= item.is_a?(Work) ? ts("Mystery Work") : ts("Mystery Series") %></h4>
-  <% if item.is_a?(Work) %>
-    <h5 class="heading"><%= h(ts("Part of ")) + (unrevealed ? item.collections.unrevealed : item.collections.anonymous).map {|challenge| link_to(challenge.title, collection_path(challenge))}.join(", ").html_safe %></h5>
+  <% if item.is_a?(Work) && !item.approved_collections.unrevealed.empty? %>
```
sarken, line 3: We could probably get away with just `<% if item.approved_collections.unrevealed.present? %>` or `<% unless item.approved_collections.unrevealed.empty? %>` here: despite what line two says, I don't see any indication this blurb is actually ever used for series.

sarken:
I'm happy to leave the decision up to @korrien! If she's up to doing more work, we can tackle that now. If not, it can wait.

korrien:
Thanks much for the tips! Now I know how to run security locally as well! 🥳 
Maybe we can update some section in the wiki with a short note about it? (don't know where it would fit though)
For a such Ruby/Rails newbie like me it's very nice to know I can preview these things locally. 

I actually thought that brakman complained about `link_to` (which should be safe as we perform validations on collection input in the controller), but it was complaining even when I removed the call. It looks to me like it got confused already at the collecting stage.
I do not really get why putting this code to the controller helps with this xss warning (are we now somehow sure that our code will always be treated as a string, not executable???), but it worked, so I decided it would be (probably) better than ignoring the warning. 

-----
On unrelated note: Mysterious work leaks a bit of its mystery as when I anonymously directly navigate to the work page I may receive an adult warning notice.

sarken:
I haven't looked at the code yet, but

> Maybe we can update some section in the wiki with a short note about it?

Sounds like a good idea to me! We'll probably end up making a new page for Hakiri/Brakeman. 

>On unrelated note: Mysterious work leaks a bit of its mystery as when I anonymously directly navigate to the work page I may receive an adult warning notice.

Thanks for the heads up! I've made https://otwarchive.atlassian.net/browse/AO3-6247 -- please take a look when you have time and make sure it accurately reflects the issue you ran into.

korrien:
> please take a look when you have time and make sure it accurately reflects the issue you ran into

Yes, you translated it perfectly! 😃

sarken (approved):

PR #4094, opened 2021-10-02
AO3-4311 Update FAQ translations when their locale changes ISO code

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4311

## Purpose

Translated FAQ and question records (from table updated by Globalize) save their own locale code values, which need to be updated if we want to rename locale codes.

Also enable I18n locale fallbacks in test environment to match the behavior in staging and production: viewing FAQs with invalid locales shows the default English FAQ title.

## Testing Instructions

See issue.


features/admins/admin_post_faqs.feature
```diff
@@ -2,85 +2,88 @@
...
+    Then I should see "Deutsch" in the "Name" input
+    When I fill in "locale_iso" with "ger"
+      And I press "Update Locale"
+    Then I should see "Your locale was successfully updated."
+      And I should see "Deutsch ger"
+
+    # The current session preference remains "de", the category name falls back to the default locale,
+    # but no "de" questions will be found.
+    When I go to the archive_faqs page
+      And I follow "New subsection"
```
redsummernight, line 72: The test dies here without `config.i18n.fallbacks = true`. The FAQ title will be `nil` for locale "ger", which means:

- This link to the FAQ (on the FAQ index) is bare, appearing as `/faq/new-subsection?language_id=de`.
- Following that link (or "Show") will crash the request when adding the title to the page subtitle:
  https://github.com/otwcode/otwarchive/blob/6b99402ac54eba0b220e85163408c777322c2251/app/controllers/archive_faqs_controller.rb#L34

features/admins/admin_post_faqs.feature
```diff
@@ -2,85 +2,88 @@
...
-#
-#    # Toggle the languages at the top and see the correct data
-#    When I follow "Archive FAQ"
-#      And I select "English" from "language_id"
-#      And I press "Go" within "div#inner.wrapper"
-#    Then I should see "New subsection"
-#      And I should not see "Neuer Abschnitt"
+    # Log out and view FAQs; the default locale should be used
+    When I log out
+      And I go to the archive_faqs page
```
redsummernight, line 78: Without the fix (`after_update` for locales), the test dies here as described in https://otwarchive.atlassian.net/browse/AO3-4311?focusedCommentId=357975.

config/environments/test.rb
```diff
@@ -37,6 +37,10 @@
   # Inline ActiveJob when testing:
   config.active_job.queue_adapter = :inline
 
+  # Enable locale fallbacks for I18n (makes lookups for any locale fall back to
+  # the I18n.default_locale when a translation can not be found)
+  config.i18n.fallbacks = true
```
sarken, line 42: Is there a benefit to not enabling fallbacks for the dev environment as well? (That's the only place it won't be enabled once this is merged.)

redsummernight: Not enabling fallbacks means we get errors when translation keys are missing. So that doesn't add much because:

- We plan to do https://otwarchive.atlassian.net/browse/AO3-6003 and have automated checks for missing keys.
- I was worried that `it_behaves_like "a translated email"` would become useless, but even with fallbacks, such assertions would still fail if we're running in the default locale and have no keys for "en" (there's nothing else to fall back on).

Should I enable fallbacks in config/application.rb for all environments?

redsummernight: Oh lol:

https://github.com/otwcode/otwarchive/blob/6b99402ac54eba0b220e85163408c777322c2251/config/environments/development.rb#L31-L33

config/application.rb it is!

sarken: ... y'know, I am getting _quite_ tired of being failed by my text editor's multi-file search. 😒

sarken (approved):

PR #4095, opened 2021-10-10
AO3-6164 Validate locale parameter on all admin FAQ actions

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6164

## Purpose

For guests and registered users, display the (empty) FAQ pages anyway for invalid locales.

For admins, redirect to the same page for the default locale, with an error message.

Also fix the infinite redirect loop for all user types when the locale is an empty string.

## Testing Instructions

See issue.

### DIFF
--- a/features/support/hooks.rb
+++ b/features/support/hooks.rb
@@ -16,6 +16,9 @@ Before do
   # Clears used values for all generators.
   Faker::UniqueGenerator.clear
 
+  # Reset global locale setting.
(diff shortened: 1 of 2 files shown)


sarken:
It might make sense to keep the controller change as well, since it prevents an admin from wasting effort and entering an entire FAQ of info in the form for the wrong locale.

PR #4099, opened 2021-11-08
AO3-6252 Add error message for kudos rate limit

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6252

## Purpose

If you submit kudos using JavaScript and run into a rate limit, you should get a meaningful error.

## Testing Instructions

As a guest, submit more kudos than allowed in a given time period. Verify the error message appears. Also verify you get the usual "You have already left kudos here. :)" error if submitting duplicate kudos without going over the rate limit.

### DIFF
--- a/public/javascripts/application.js
+++ b/public/javascripts/application.js
@@ -442,7 +442,7 @@ $j(document).ready(function() {
       error: function(jqXHR, textStatus, errorThrown) {
         var msg = 'Sorry, we were unable to save your kudos';
 
-        // When we hit the rate limit, the response is a plain text 429.
+        // When we hit the rate limit, the response from Rack::Attack is a plain text 429.
         if (jqXHR.status == "429") {
           msg = "Sorry, you can't leave more kudos right now. Please try again in a few minutes.";
         } else {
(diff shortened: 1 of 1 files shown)


public/javascripts/application.js
```diff
@@ -439,19 +439,24 @@ $j(document).ready(function() {
...
       data: jQuery('#new_kudo').serialize(),
-      error: function(jqXHR, textStatus, errorThrown ) {
+      error: function(jqXHR, textStatus, errorThrown) {
         var msg = 'Sorry, we were unable to save your kudos';
-        var data = $j.parseJSON(jqXHR.responseText);
 
-        if (data.errors && (data.errors.ip_address || data.errors.user_id)) {
-          msg = "You have already left kudos here. :)";
-        }
+        // When we hit the rate limit, the response is a plain text 429.
```
redsummernight, line 445: Can you specify "the response from Rack::Attack"?

sarken: Are we definitely using Rack::Attack for this? I thought it was unrelated since james was doing it on the Systems side. (If it is Rack::Attack, should it be in the codebase with the other Rack::Attack settings?)

redsummernight: Yeah we are, it's the default setting.

https://github.com/rack/rack-attack/blob/2257f008763b405f6a1515b460f80ed5d0f06c2c/lib/rack/attack/configuration.rb#L10-L20

redsummernight (approved):

PR #4100, opened 2021-11-11
AO3-6254 AO3-6256 Fix kudos counts for pluralization and kudos email capitalization

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6254
https://otwarchive.atlassian.net/browse/AO3-6256

## Purpose

Fixes the counts we use for the pluralization of the individual "[whoever] left kudos on [whatever]" sections of the kudos emails. 

Stops using `capitalize` on `a guest` when we need the sentence to be "A guest left kudos on [whatever]," because this did things like turn "ein Gast" into "Ein gast." Now we handle that case separately.

## Testing Instructions

Refer to Jira.


app/mailers/kudo_mailer.rb
```diff
@@ -22,26 +23,30 @@ def batch_kudo_notification(user_id, user_kudos)
...
-          kudo_givers = names
+          # dup so we don't add "a guest" or "5 guests" to the original names
+          # array. If we do that, our kudo_count will be too high whenever we
+          # have both named and guest kudos.
+          kudo_givers = names.dup
           kudo_givers << guest_kudos(guest_count) unless guest_count == 0
-          @kudos_count = names.size + guest_count
+          kudo_count = names.size + guest_count
         else
           kudo_givers << guest_kudos(guest_count).capitalize unless guest_count == 0
```
redsummernight, line 38: We might want to do https://otwarchive.atlassian.net/browse/AO3-6256 at the same time. Depending on how we fix that the sentence structure may change entirely.

app/views/kudo_mailer/batch_kudo_notification.html.erb
```diff
@@ -1,19 +1,19 @@
...
+    %>
 
-        <%= t('mailer.kudos.html.left_kudos', givers_list: givers_list, commentable_link: commentable_link.html_safe, count: @kudos_count.to_i).html_safe %>
-
-        <% if (index < @commentables.length-1) %>
-            <%= styled_divider %>
-        <% end %>
+    <%= t("mailer.kudos.html.left_kudos", givers_list: givers_list, commentable_link: commentable_link.html_safe, count: kudo_count).html_safe %>
 
+    <% if (index < @commentables.length-1) %>
```
redsummernight, line 14: Can you add spaces around the minus in both templates?

app/mailers/kudo_mailer.rb
sarken: Do you want me to add the fix to this pull request? It won't require any changes to the sentence structure with the way I plan to do it.

redsummernight: Let's make it a doubleheader. Efficient testing!

app/mailers/kudo_mailer.rb
```diff
@@ -10,7 +10,11 @@ class KudoMailer < ActionMailer::Base
   #   guest_count: number of guest kudos
   def batch_kudo_notification(user_id, user_kudos)
     @commentables = []
+    @kudo_counts = {}
     @kudo_givers = {}
+    # Does this work have only one kudos, and is it from a guest? We need this
+    # information in the view.
+    @single_guest_only = {}
```
sarken, line 17: I will gladly take suggestions for a better name.

sarken: Oh, also: I considered making an array containing the commentables that matched this criteria and then, in the view, checking if the current commentable was in the array to determine which translation we wanted to use. It doesn't really save any code and is just as difficult to name, but if it seems like it would be clearer, I can switch to that.

features/comments_and_kudos/kudos.feature
```diff
@@ -152,10 +153,13 @@ Feature: Kudos
       And I am logged out
       And I leave kudos on "Awesome Story"
       And I leave kudos on "Another Awesome Story"
+      And I leave kudos on "Meh Story"
     When kudos are sent
     Then 1 email should be delivered to "myname1@foo.com"
       And the email should contain "myname2"
       And the email should contain "someone_else"
       And the email should contain "guest"
```
sarken, line 161: I can adjust this to

```
And the email should contain "a guest"
And the email should contain "A guest left"
```

Once I have feedback on the variable name/approach.

app/mailers/kudo_mailer.rb
redsummernight: I think we can have the mailer do one pass through `kudos_hash` to build a list of commentables that exist. The template gets `@commentables` and `user_kudos` (hash from `[commentable_type]_[commentable_id]` to: list of registered user names, a guest count) and decides which translation to pick from there.

So most of the logic will be in the templates.

app/mailers/kudo_mailer.rb
```diff
@@ -10,43 +10,24 @@ class KudoMailer < ActionMailer::Base
...
-        end
-        next if kudo_givers.empty?
-
         @commentables << commentable
-        @kudo_givers[commentable_info] = kudo_givers
       end
       mail(
         to: user.email,
-        subject: t('mailer.kudos.you_have', app_name: ArchiveConfig.APP_SHORT_NAME)
+        subject: t("mailer.kudos.you_have", app_name: ArchiveConfig.APP_SHORT_NAME)
```
sarken, line 29: Lowkey want to fix the key names for this...

app/views/kudo_mailer/batch_kudo_notification.html.erb
```diff
@@ -1,19 +1,29 @@
...
+    <%
+      commentable_link = style_creation_link(commentable.commentable_name, polymorphic_url(commentable))
+      givers_hash = @user_kudos["#{commentable.class.name}_#{commentable.id}"]
+      # dup so we don't add "a guest" or "5 guests" to the original array. If
+      # that happens, kudo_count ends up too high whenever both named and guest
+      # kudos are present.
+      names = givers_hash["names"].dup
+      guest_count = givers_hash["guest_count"].to_i
+      kudo_count = names.size + guest_count
+      givers = names
```
redsummernight, line 13: This makes sense but confuses me a little bit at a glance. I think it's clearer if you have:
```rb
names = givers_hash["names"]
# more code here

# dup so we don't add "a guest" or "5 guests" to the original array. [...]
givers = names.dup
```

app/mailers/kudo_mailer.rb
redsummernight: Please do!

config/locales/mailers/en.yml
```diff
@@ -26,19 +26,22 @@ en:
             part1: "If you've received this message in error, please contact Support at %{support_url}."
             part2: "The Archive of Our Own is a fan-run and fan-supported archive that relies on your donations: %{donate_url}."
         sent_at: "Sent at %{sent_at}."
-    kudos:
+  kudo_mailer:
```
redsummernight, line 29: These keys are renamed, maybe we need to deal with https://otwarchive.atlassian.net/browse/AO3-6240 first.

Otherwise this looks good.

redsummernight (approved):

redsummernight:
Set to ready. We'll think about key renames later.

PR #4102, opened 2021-11-16
AO3-4590 Fixing display of parentheses after RTL language titles in notification emails

setupminimal:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-4590

## Purpose

This PR adds bdi (bi-directional isolation) tags to the HTML for notification emails, and a Unicode left-to-right mark to the text form of notification emails to prevent titles in right-to-left languages from causing weakly directional characters (such as numbers and parentheses) from displaying incorrectly.

This does not fix all possible locations where this could be a problem -- I think fixing this everywhere on the site probably requires doing something tricky with Rails' string interpolation -- but it does fix the places where the error was initially reported in the original ticket. We can use the same fix in other places, as it comes up.

## Testing Instructions

This can be tested by generating an update notification email for a work with a title written in a right-to-left language like Hebrew or Arabic, and checking that the word count displayed after the title appears correct. This should be repeated for an account that gets HTML emails and an account that gets text emails.

## References

None

## Credit

Daroc Alden, they/them

### DIFF
--- a/app/views/user_mailer/recipient_notification.text.erb
+++ b/app/views/user_mailer/recipient_notification.text.erb
@@ -3,7 +3,7 @@
 
 A gift work has been posted for you<% if @collection %> in the "<%= @collection.title %>" collection (<%= collection_url(@collection) %>)<% end %> at the Archive of Our Own!
 
-"<%= @work.title.html_safe %>" (<%= @work.word_count %> words)
+"<%= @work.title.html_safe %>"‎ (<%= @work.word_count %> words)
 <%= @collection ? collection_work_url(@collection, @work) : work_url(@work) %>
 
 by <%= @work.anonymous? ? "an anonymous responder" : (@work.pseuds.map{|p| text_pseud(p)}.to_sentence) %>
--- a/app/helpers/mailer_helper.rb
+++ b/app/helpers/mailer_helper.rb
@@ -12,11 +12,13 @@ module MailerHelper
   # For work, chapter, and series links
   def style_creation_link(title, url, html_options = {})
     html_options[:style] = "color:#990000"
+    title = "<bdi>" + title + "</bdi>"
     ("<i><b>" + link_to(title.html_safe, url, html_options) + "</b></i>").html_safe
   end
 
   # For work, chapter, and series titles
   def style_creation_title(title)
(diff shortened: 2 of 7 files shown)


sarken (changes requested):
Thanks for the pull request! I agree we can't fix it everywhere at once, but it looks like the issue mentions a couple of emails that are affected and are not currently covered by this pull request:

* Subscription emails (batch_subscription_notification.text.erb & batch_subscription_notification.html.erb)
* Prompt fill emails (prompter_notification.html.erb & prompter_notification.text.erb)

Could you update those as well?

setupminimal:
I'm not sure how I missed those, thank you for pointing them out. I'll add them.

app/views/user_mailer/recipient_notification.text.erb
```diff
@@ -3,7 +3,7 @@
 
 A gift work has been posted for you<% if @collection %> in the "<%= @collection.title %>" collection (<%= collection_url(@collection) %>)<% end %> at the Archive of Our Own!
 
-"<%= @work.title.html_safe %>" (<%= @work.word_count %> words)
+"<%= @work.title.html_safe %>"‎ (<%= @work.word_count %> words)
```
sarken, line 6: This seems to be using the RTL mark (U+200F) while the other emails use the LTR mark (U+200E).

sarken (commented):
Thanks for working on this! Unfortunately, after some research and discussion, we don't think this is the best way to tackle this issue.

For the HTML emails, it looks like the bdi tag doesn't have great support:
* https://www.caniemail.com/search/?s=bdi
* https://www.pinpointe.com/blog/email-campaign-html-and-css-support

And with the text emails, there are a couple of potential issues:
* While a LTR mark might work for English emails, we're in the process of translating emails, and we'll probably need a RTL mark for the opposite case (e.g., Hebrew emails with English titles).
* Using invisible characters is difficult to maintain, particularly with 40 language teams who will be handling strings like `"%{work.title}"‏ (%{word_count} words)`.

If you have any ideas for a different way to handle this, we'd love to hear them! However, we're going to close this pull request for now, until we decide on an approach that will work for us.

PR #4105, opened 2021-11-28
AO3-6260 Remove setup files for Codeship tests

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6260

## Purpose

Remove some unused files.

## Testing Instructions

None.


sarken (commented):
Should we also take the Codeship build status badge off the README?

redsummernight:
We should not because we're still using it for staging deploys.

sarken:
Ah, so it'll let us know if the deploy to staging fails, not just the tests?

redsummernight:
Yeah. Eventually I'd like to move away from Codeship for deploys as well ([AO3-6133](https://otwarchive.atlassian.net/browse/AO3-6133)), but for now we keep the badge.

PR #4108, opened 2021-12-01
AO3-5717 Fix expected chapter count lower than actual count

luis-pabon-tf:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5717

## Purpose

Uses an existing virtual attribute so that the chapter count safeguards happen even if the chapter count provided is wrong.

## Testing Instructions

In the JIRA ticket.

## Credit

bird - (they)

### DIFF
--- a/spec/models/work_spec.rb
+++ b/spec/models/work_spec.rb
@@ -464,22 +464,7 @@ describe Work do
     end
   end
 
-  describe "#wip_length_equals" do
-    it "updating chapter count via expected_number_of_chapters ignores any safeguard logic" do
-      work = create(:work)
-      create(:chapter, work: work)
-      work.reload
-
-      work.expected_number_of_chapters = 1
-      expect(work.wip_length).to eq(work.expected_number_of_chapters)
-
-      work.expected_number_of_chapters = 2
-      expect(work.wip_length).to eq(work.expected_number_of_chapters)
-
-      work.expected_number_of_chapters = 3
-      expect(work.wip_length).to eq(work.expected_number_of_chapters)
-    end
-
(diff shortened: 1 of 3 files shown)


redsummernight (commented):
Can you add a test for this behavior of `wip_length`? In spec/models/work_spec.rb, we can have a block for `describe "#wip_length="` and check that setting the attribute to e.g. 2 for a work with 1 chapter will end up with both `expected_number_of_chapters` and `wip_length` at 1.

spec/models/work_spec.rb
```diff
@@ -464,6 +464,41 @@
     end
   end
 
+  describe "#wip_length_equals" do
```
redsummernight, line 467: Can you change this to `#wip_length=` to match the method name?

spec/models/work_spec.rb
```diff
@@ -464,6 +464,41 @@
...
+    it "updating chapter count via expected_number_of_chapters ignores any safeguard logic" do
```
redsummernight, line 468: I think we can skip this test, we don't have to check that the default `expected_number_of_chapters` setter works as expected.

tickinginstant:
Would you mind removing `expected_number_of_chapters` from the `work_params` function in `WorksController`, as well?
https://github.com/otwcode/otwarchive/blob/72b975cf5ed90ce28cf212e84420aaccea871372/app/controllers/works_controller.rb#L893-L896
Otherwise someone would still be able to manipulate the HTML to get an invalid chapter count.

redsummernight (approved):
Looks good, thanks!

PR #4110, opened 2021-12-05
AO3-6257 Add gift preference

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6257

## Purpose

Allows users to turn off/on the ability to receive gifts in certain circumstances. (Sorry about the huge diff for preferences; the whitespace was in desperate need of fixing.)

## Testing Instructions

Refer to Jira.

## References

This is #3981 updated to work with a preference rather than protected users. 

Changes start at https://github.com/otwcode/otwarchive/pull/4110/commits/fda07eeb35f7a8e842c63e4b67d4a4e6d63c0d15

## Credit

Sarken, she/her


sarken:
Oops, I was premature. I just realized I need to update all the existing tests.

app/models/collection_item.rb
```diff
@@ -191,12 +191,10 @@ def send_work_invitation
 
   after_update :notify_of_status_change
   def notify_of_status_change
-    if saved_change_to_unrevealed? && item.respond_to?(:new_recipients)
-      # making sure notify_recipients in the work model has not already notified
-      if item.new_recipients.present?
-        notify_of_reveal
-      end
-    end
+    return unless saved_change_to_unrevealed? && item.respond_to?(:new_gifts)
```
redsummernight, line 194: Should this method still be here, per @tickinginstant's [comment on the old PR](https://github.com/otwcode/otwarchive/pull/3981#discussion_r668238836)?

features/gift_exchanges/challenge_giftexchange.feature
```diff
@@ -596,3 +596,53 @@ Feature: Gift Exchange Challenge
...
       And I follow "My Assignments" within "#dashboard"
     Then I should not see the image "src" text "/images/envelope_icon.gif"
+
+  Scenario: A user who disallows gift works is cautioned about signing up for
+  an exchange.
+    Given the gift exchange "Some Gift Exchange" is ready for signups
+      And I am logged in as "participant"
+      And the user "participant" disallows gifts
+    When I go to "Some Gift Exchange" collection's page
+     And I follow "Sign-up Form"
```
redsummernight, line 606: There should be an extra leading space here.

features/gift_exchanges/challenge_giftexchange.feature
```diff
@@ -596,3 +596,53 @@ Feature: Gift Exchange Challenge
...
+    Then I should see "Signing up for this challenge will allow users to give you gift works"
+
+  Scenario: A user who allows gift works is not cautioned about signing up for
```
redsummernight, line 609: I think we can combine this scenario with the previous one, same setup.

app/models/collection_item.rb
sarken: I cannot remember why I didn't take it out, so I'll try it and we can see what happens.

db/migrate/20211204001151_add_allow_gifts_to_preference.rb
```diff
@@ -0,0 +1,5 @@
+class AddAllowGiftsToPreference < ActiveRecord::Migration[5.2]
+  def change
+    add_column :preferences, :allow_gifts, :boolean, default: true, null: false
```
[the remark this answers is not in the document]
redsummernight: Also applies to the other migration ChangePreferenceAllowGiftsDefault.

db/migrate/20211205115025_change_preference_allow_gifts_default.rb
```diff
@@ -0,0 +1,49 @@
+class ChangePreferenceAllowGiftsDefault < ActiveRecord::Migration[5.2]
+  def up
+    if Rails.env.staging? || Rails.env.production?
+      database = Preference.connection.current_database
+
+      puts <<~PTOSC
+        Schema Change Command:
+
+        pt-online-schema-change D=#{database},t=preferences \\
+          --alter "ALTER allow_gifts SET DEFAULT 0" \\
```
sarken, line 10: I'm pretty sure this is right, since [the documentation](https://www.percona.com/doc/percona-toolkit/2.0/pt-online-schema-change.html#options) says everything that comes after the initial `ALTER table`, but it feels wrong.

redsummernight: Should be `ALTER COLUMN allow_gifts SET DEFAULT 0` (https://stackoverflow.com/a/27907555).

sarken: Ah, that makes it seem less wrong, thanks!

db/migrate/20211204001151_add_allow_gifts_to_preference.rb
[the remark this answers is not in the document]
tickinginstant: I'm pretty sure changing the default from one non-null value to another is metadata-only, so I think PTOSC is overkill for `ChangePreferenceAllowGiftsDefault`.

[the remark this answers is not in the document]
sarken: I'm happy to revert it if we'd rather.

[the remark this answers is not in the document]
redsummernight: Nice, I didn't know that: https://dev.mysql.com/doc/refman/5.7/en/innodb-online-ddl-operations.html#online-ddl-column-operations. Let's revert PTOSC for the second migration.

app/models/work.rb
```diff
@@ -436,40 +461,33 @@ def recipients=(recipient_names)
       gift = self.gifts.for_name_or_byline(name).first
       if gift
         gifts << gift # new gifts are added after saving, not now
-        new_recipients << name unless self.posted # all recipients are new if work not posted
+        new_gifts << gift unless self.posted # all gifts are new if work not posted
       else
-        # check that the gift would be valid
-        g = Gift.new(work: self, recipient: name)
+        g = self.gifts.new(work: self, recipient: name)
```
redsummernight, line 466: You can omit `work: self` here, `self.gifts.new` implies that.

app/models/work.rb
```diff
@@ -436,40 +461,33 @@ def recipients=(recipient_names)
...
-          g.save
-        end
-      end
+  def save_new_gifts
+    return if self.new_gifts.blank?
+
+    self.new_gifts.each do |gift|
+      next if self.gifts.for_name_or_byline(gift.recipient).present?
+
+      Gift.create(recipient: gift.recipient, work: self)
```
redsummernight, line 490: This line works, though it looks a bit odd. Same question as https://github.com/otwcode/otwarchive/pull/3981#discussion_r668239587.

sarken: I don't remember at all, so I've pushed the broken version (https://github.com/otwcode/otwarchive/pull/4110/commits/88c12e48d16f9201eded3d8ac1cac376270cfac0) so we can see what fails.

tickinginstant: I think the issue is that `work_id` doesn't get set properly on gifts created before the work is saved (or at least, gifts created with this particular series of steps). Hence re-creating the gift after the work has been saved.

redsummernight (approved):

PR #4115, opened 2021-12-06
AO3-6190 Don't use Time.now and Date.today in the work/chapter date calculations.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6190

## Purpose

Stop using Time.now and Date.today in the work/chapter date calculations, because they return the time/date in the system time zone (UTC, for production), which doesn't match all of the times/dates that use the time zone set here:
https://github.com/otwcode/otwarchive/blob/72b975cf5ed90ce28cf212e84420aaccea871372/config/application.rb#L50-L53
This should prevent all future works from encountering https://otwarchive.atlassian.net/browse/AO3-5392, but won't address any pre-existing works with a `revised_at`/`published_at` mismatch.

I also modified `set_revised_at` so that redating to the current day will no longer set `revised_at` to noon UTC, even if noon UTC is in the future.

### DIFF
--- a/app/models/chapter.rb
+++ b/app/models/chapter.rb
@@ -172,8 +172,8 @@ class Chapter < ApplicationRecord
   # Checks the chapter published_at date isn't in the future
   def validate_published_at
     if !self.published_at
-      self.published_at = Date.today
-    elsif self.published_at > Date.today
+      self.published_at = Date.current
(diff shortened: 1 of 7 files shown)


redsummernight (approved):
Thank you!

PR #4116, opened 2021-12-06
AO3-6132 Fix intermittent failure in series blurb helper tests

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6132

## Purpose

Fixes an intermittent test failure introduced in the tests for AO3-6132. Adding `user2` as a creator of the work in `spec/helpers/application_helper_spec.rb:248` causes the series to be touched, but the series object that has `touch` called on it is a different in-memory copy of the series defined in the `let` above. This causes the initial cache key to be computed incorrectly, which breaks the test.

### DIFF
--- a/spec/helpers/application_helper_spec.rb
+++ b/spec/helpers/application_helper_spec.rb
@@ -246,7 +246,10 @@ describe ApplicationHelper do
         end
 
         context "when user is removed from series' work" do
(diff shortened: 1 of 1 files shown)


sarken (approved):
Successfully ran the tests 5 or 6 times. Thanks!

PR #4117, opened 2021-12-11
AO3-6262 Update pluralization rule for Croatian

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6262

## Purpose

Phrase follows CLDR rules, where Croatian uses [one/few/other](https://unicode-org.github.io/cldr-staging/charts/latest/supplemental/language_plural_rules.html#hr) but rails-i18n expects [one/few/other/many](https://github.com/svenfuchs/rails-i18n/blob/f6cc43ec8cfc08d0db7a06ca8299d7834c517a05/lib/rails_i18n/common_pluralizations/east_slavic.rb#L37).

## Testing Instructions

See issue.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -107,6 +107,8 @@ RSpec/ContextWording:
 
 RSpec/DescribeClass:
   Exclude:
+    # Exception for specs about I18n configurations
+    - 'spec/lib/i18n/**/*.rb'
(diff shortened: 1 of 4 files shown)


sarken (approved):
Team Croatian was happy, so I'm happy!

PR #4127, opened 2021-12-30
AO3-6249 Make Pseuds dropdown on user page sidebar always label itself as Pseuds

AuroraAri:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6249

## Purpose

Changes the "Pseuds" dropdown on /users/USERNAME/* pages to always read as "Pseuds" rather than showing the pseud that is currently being acted on

## Testing Instructions

1. Log in as a user who has more than one pseud
2. Click the "Pseuds ↓" dropdown
3. Switch to another pseud
4. See that the "Pseuds ↓" dropdown is no longer changing :)

## Credit

Arianna Story (she/her)

### DIFF
--- a/app/views/users/_sidebar.html.erb
+++ b/app/views/users/_sidebar.html.erb
@@ -5,8 +5,7 @@
   <li><%= span_if_current ts("Profile"), user_profile_path(@user) %></li>
   <% if @user.pseuds.size > 1 %>
     <li class="pseud" aria-haspopup="true">
-      <% pseud_link_text = ts("Pseuds") %>
-      <a href="#" title="<%= ts("Pseud Switcher") %>"><%= pseud_link_text %></a>
+      <a href="#" title="<%= ts("Pseud Switcher") %>"><%= ts("Pseuds") %></a>
       <ul class="expandable secondary">
         <%= print_pseud_selector(@user.pseuds) %>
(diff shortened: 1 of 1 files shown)


app/views/users/_sidebar.html.erb
```diff
@@ -5,7 +5,7 @@
   <li><%= span_if_current ts("Profile"), user_profile_path(@user) %></li>
   <% if @user.pseuds.size > 1 %>
     <li class="pseud" aria-haspopup="true">
-      <% pseud_link_text = current_page?(@user) ? ts("Pseuds") : (@author ? @author.name : @user.login) %>
+      <% pseud_link_text = ts("Pseuds") %>
```
sarken, line 8: Can we do `<a href="#" title="<%= ts("Pseud Switcher") %>"><%= ts("Pseuds") %></a>` instead? I don't think there's much value to continuing to use the `pseud_link_text` variable, since it looks like it's only used in that one place.

sarken (changes requested):
Thanks for the pull request! While this would definitely work fine, I think we can simplify things a bit.

app/views/users/_sidebar.html.erb
AuroraAri: Done! ☺️ It's been a bit since I've worked with Ruby, so I'll just ascribe that to growing pains. Thanks for the kind review!

sarken (approved):
Looks good now -- thank you again!

PR #4130, opened 2022-01-03
AO3-5601 Use DownloadWriter for deleted work email attachments

tlee911:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5601

## Purpose

Switched work deletion emails (both by author or admins) to use the DownloadWriter to create attachments rather than the custom function from the mailer.

~I've also made the TXT attachment plain text by stripping the HTML tags from the HTML version of work (previously was just putting two copies of the HTML content with different filenames).~

## Testing Instructions

Verify that works deleted by both authors and admins still generate their attachments correctly.

Automated tests have been added in `spec/mailers/user_mailer_spec.rb` (these notifications were previously not covered).


app/mailers/user_mailer.rb
```diff
@@ -353,11 +353,11 @@ def prompter_notification(work_id, collection_id=nil)
   def delete_work_notification(user, work)
     @user = user
     @work = work
-    work_copy = generate_attachment_content_from_work(work)
-    work_copy = ::Mail::Encodings::Base64.encode(work_copy)
```
redsummernight, line 357: I think we would still want base64 encoding (plus the `encoding` option on `attachments`) to avoid long lines (https://otwarchive.atlassian.net/browse/AO3-5453).

app/mailers/user_mailer.rb
```diff
@@ -353,11 +353,11 @@ def prompter_notification(work_id, collection_id=nil)
...
-    filename = work.title.gsub(/[*:?<>|\/\\\"]/,'')
-    attachments["#{filename}.txt"] = { content: work_copy, encoding: "base64" }
-    attachments["#{filename}.html"] = { content: work_copy, encoding: "base64" }
+    @download = Download.new(@work, mime_type: "text/html")
+    @download.generate
```
redsummernight, line 357: Download writes to a file, and then we have to `File.read` from it. I think the part that we need is:
https://github.com/otwcode/otwarchive/blob/5161fd10c15abd943d9c50158c06aa0fe55e420d/app/models/download_writer.rb#L23-L34
That gives the string right away. You'll need to refactor a little to get a separate public method for it, then you can do something like:
```
DownloadWriter.new(@download).html
```

app/mailers/user_mailer.rb
```diff
@@ -353,11 +353,11 @@ def prompter_notification(work_id, collection_id=nil)
   def delete_work_notification(user, work)
     @user = user
     @work = work
-    work_copy = generate_attachment_content_from_work(work)
-    work_copy = ::Mail::Encodings::Base64.encode(work_copy)
-    filename = work.title.gsub(/[*:?<>|\/\\\"]/,'')
-    attachments["#{filename}.txt"] = { content: work_copy, encoding: "base64" }
-    attachments["#{filename}.html"] = { content: work_copy, encoding: "base64" }
+    @download = Download.new(@work, mime_type: "text/html")
```
redsummernight, line 356: You can use a local variable for this. Instance variables are accessible from views, but we don't need that here.

redsummernight (commented):
> Note that I haven't fully looked into how to get emails from local development (doesn't send emails). I've just been inspecting mail contents from logs. I'm assuming staging environment does have mailing set up.

You can set `config.action_mailer.delivery_method` to `:file` and mails will be saved to `tmp/mails` by default (https://api.rubyonrails.org/v5.2/classes/ActionMailer/Base.html).

The staging environment does have mailing set up similar to production.

app/mailers/user_mailer.rb
redsummernight: We should keep the calls `::Mail::Encodings::Base64.encode` on the final content strings as well.

app/mailers/user_mailer.rb
```diff
@@ -353,11 +353,14 @@ def prompter_notification(work_id, collection_id=nil)
...
     @user = user
     @work = work
-    work_copy = generate_attachment_content_from_work(work)
-    work_copy = ::Mail::Encodings::Base64.encode(work_copy)
-    filename = work.title.gsub(/[*:?<>|\/\\\"]/,'')
-    attachments["#{filename}.txt"] = { content: work_copy, encoding: "base64" }
-    attachments["#{filename}.html"] = { content: work_copy, encoding: "base64" }
+    download = Download.new(@work, mime_type: "text/html")
+    html = DownloadWriter.new(download).generate_html
+    txt = ActionController::Base.helpers.strip_tags(html)
```
redsummernight, line 358: I just noticed this after reading old comments on the issue: would this remove all HTML tags in the output, including user-created ones in the work body?

We can keep doing what we're doing now, which is the same output for both types of attachments.

tlee911: Yes, the previous behaviour was to attach the same file twice but with different filenames, which didn't seem particularly useful to me.  Do we want to go back to that behaviour?  I feel like this gives you the option of a true plain text version, versus HTML just saved as text (which you technically have anyway, if you rename the HTML file when you download it).

Do we want to put it back to the way it was before?

redsummernight: Let's put it back to the way it was.

I don't think users would want a version where their formatting gets removed.

Having the plain text file with the same content makes sure that people have a way to open and view works with the HTML visible (easy to repost elsewhere). Some devices may only display the HTML file as rich text, or [have problems opening it at all](https://support.google.com/chrome/thread/32010024/cant-open-local-html-file-on-chrome?hl=en).

redsummernight (approved):
Looks good, thanks!

PR #4131, opened 2022-01-04
AO3-6233 Server-side highlight for autocomplete (tags)

tlee911:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6233

## Purpose

Reimplements the autocomplete tag highlighting to be done in the `autcomplete_controller.rb` rather than Javascript.  This will fix the highlighting for diacritics in the autocomplete.

## Testing Instructions

Cucumber tests have been added to `autocomplete.feature`.  (Now that the functionality is in the controller though, maybe some of these scenarios should be moved to rspec.)

For manual testing:
- Create canonical tags of various types, both with and without diacritics
- Reload the tag data for the autocomplete `rake autocomplete:reload_tag_data`
- Go to the form to post a new work (or edit an existing one)
- Enter some values in the tag fields and ensure the autocomplete brings up suggestions with the appropriate parts highlighted.

The updated method for highlighting could be slow if there are a lot of long potentially matching tags.  Performance seems reasonable to me, but that might be dependent on device, and I did not run a test with more than a few tag suggestions at a time.

I've also incidentally changed a behaviour which I think was unintentional before.  Now only the first occurence of match in a suggestion will be highlighted.  For example, if you type `f` and "Fluff" is in the suggestions, only the first "F" will be bolded, instead of all three.  I think this makes more sense as the autocomplete is only keying in on matches at the start of words anyway.

## References

#4009 (Original implementation of diacritics handling in autocomplete)
#4089 (Closed not merged PR for this highlighting done via JS)


tlee911:
Putting in draft.  The test failures are legit and all due to the same thing.  I've inserted <b> tags into API responses essentially, but this breaks the "in fandom" tag search because then something "<b>Fan</b>dom Name" gets inserted into the fandom, and that doesn't trigger the right filtering for the tag search since the fandom doesn't seem like it matches.

PR #4134, opened 2022-01-09
AO3-4037 Retry comment notifications when comment does not exist.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4037

## Purpose

Modify all mailers to retry on a deserialization error or a record not found error. It will try to send the email a total of 3 times, with 1 minute of delay between each attempt. If all three attempts fail, it will silently delete the mailer job, instead of bubbling up to the Resque failures list.


redsummernight (approved):
Thanks!

PR #4135, opened 2022-01-09
AO3-6216 Fix link for related works in downloads

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6216

## Purpose

* Update links to related works in downloaded works to us the correct archive URL, rather than `file:///work/{id}`
* Redact restricted related works in downloaded works

## Testing Instructions

See JIRA ticket

## Credit

**Brian Austin (he/they)**


brianjaustin:
I noticed the issue with restricted works while working on the original issue. If y'all would prefer that be fixed separately, I can revert those changes and make a separate bug for it

tickinginstant:
>I noticed the issue with restricted works while working on the original issue. If y'all would prefer that be fixed separately, I can revert those changes and make a separate bug for it

I am not a volunteer, so I can't provide an official ruling, but that behavior is very likely deliberate, not a bug. It's possible that they've changed it since I last looked, but there are certain quirks in the way the system handles downloads that mean calling `current_user` in the download code will not work in staging and production. So there are some compromises in the download code.

Even if it does count a bug, I'm pretty sure it needs to be in a different issue (to handle the quirks).

redsummernight:
On production, downloads are heavily cached by nginx, the same files are served to all users. So any visibility checks depending on the logged in status of users don't work. There's more context at #3522, where we decided that guest users can download restricted works (even if there's no easy UI and they would have to find the URL themselves).

So for restricted related works in downloads, we have these options:

1. Show them to no one.
2. Show them to all users, with title (current behavior).
3. Show them to all users, with title redacted.

We could switch to 3, because that's [how restricted related works are already displayed to guest users](https://github.com/otwcode/otwarchive/blob/6032d99dd65594e8e471d7296be7bb3b9967ccac/app/views/works/_work_approved_children.html.erb#L13-L18), but I'd wait for other reviewers.

brianjaustin:
Ah totally forgot about caching, my bad! Personally I'd lean more towards 3 than the current behaviour, but in any case I agree with @tickinginstant that it should be a separate issue if changes are made. I've reverted on this PR to keep it on topic

features/step_definitions/work_download_steps.rb
```diff
@@ -1,3 +1,15 @@
+When /^I download "(.*?)"$/ do |title|
+  work = Work.find_by(title: title)
+  visit work_url(work)
+  step %{I follow "HTML"}
+end
+
+Then /^I should see the inspiring parent work link$/ do
+  parent = Work.find_by(title: "Worldbuilding")
+  inspired_link = "<a href=\"#{work_url(parent)}\">#{parent.title.html_safe}</a>"
```
sarken, line 9: It looks like the issue was for a problem that occurs specifically with external works, not works. Could we test that case as well (or instead)?

sarken (changes requested):

features/step_definitions/work_download_steps.rb
brianjaustin: Fair enough, test added!

features/step_definitions/work_download_steps.rb
```diff
@@ -1,3 +1,21 @@
+When /^I download "(.*?)"$/ do |title|
```
sarken, line 1: Sorry, one more thing! This new step should also take the download format as an argument (using [Cucumber expressions](https://github.com/cucumber/cucumber-expressions#readme) rather than a regular expression) so we can use it in all of the other scenarios that currently do: 

```
I view the work "Work Title"
I follow "HTML/EPUB/Whatever"
```

Or, even easier, we can just use those two steps in the new tests as well.

No rush, though! We're saving up all of the download-related issues for [a single release](https://otwarchive.atlassian.net/issues/?filter=10601&jql=project%20%3D%20AO3%20AND%20fixVersion%20%3D%20%220.9.3xx%20(Downloads)%22), and there are a lot that still need fixing before we're ready. (All of our download files are cached, so we're going to have dump the cache to ensure the fixes get applied across the board, not just to new works, and we'd prefer to do that as infrequently as possible.)

brianjaustin: No worries! I've updated to use the existing steps. I was planning to just go through that release (either either 1 issue per PR or grouping similar issues together, whichever is preferred) next, but if you'd prefer I can go back to the wiki page.

sarken (approved):

redsummernight:
Would you be able to update this pull request to fix the merge conflicts for this upcoming release? Let us know if you'd like us to handle them instead.

PR #4136, opened 2022-01-12
AO3-5392 Add task to reset revised_at on backdated works

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5392

## Purpose

The field `revised_at` is used in work blurbs.

Also reset revised_at for non-backdated works if it's arbitrarily set in the past (possibly because of [AO3-6187](https://otwarchive.atlassian.net/browse/AO3-6187)).

Note that the actual bug was fixed in #4115. This is just to fix the dates on existing works.

## Testing Instructions

Find an existing work from testy affected by the bug, run the task and check that the work now has consistent dates in blurb and meta.

### DIFF
--- a/lib/tasks/after_tasks.rake
+++ b/lib/tasks/after_tasks.rake
@@ -876,21 +876,28 @@ namespace :After do
 
   desc "Reset blurb date (revised_at) on backdated works"
   task(reset_revised_at_on_backdated_works: :environment) do
-    # Fix all backdated works, as well as works with an arbitrary revised_at
-    # set in the past (possibly because of AO3-6187).
+    # Fix all backdated works, as well as works with an arbitrary revised_at set
+    # in the past (possibly because of AO3-6187). Skip works failing validations.
     works = Work.where("backdate OR revised_at < created_at")
     work_count = works.count
     total_batches = (work_count + 999) / 1000
     puts("Checking #{work_count} works in #{total_batches} batches") && STDOUT.flush
 
+    invalid_work_ids = []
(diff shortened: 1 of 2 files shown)


spec/lib/tasks/after_tasks.rake_spec.rb
```diff
@@ -301,3 +301,48 @@
...
+
+  context "on non-backdated works" do
+    it "does nothing" do
+      expect do
+        subject.invoke
+      end.to avoid_changing { work.reload.updated_at }
+        .and avoid_changing { work.reload.revised_at }
+    end
+
+    it "resets revised_at if arbitrary set in the past" do
```
sarken, line 320: I think we want the adverbial form "arbitrarily" here.

lib/tasks/after_tasks.rake
```diff
@@ -874,6 +874,26 @@ namespace :After do
     puts("Added default rating to works: #{updated_works}") && STDOUT.flush
   end
 
+  desc "Reset blurb date (revised_at) on backdated works"
+  task(reset_revised_at_on_backdated_works: :environment) do
+    # Fix all backdated works, as well as works with an arbitrary revised_at
+    # set in the past (possibly because of AO3-6187).
```
sarken, line 880: Do we want to note for future reference that works failing validations (e.g., because they're over the tag limit) will not be fixed? Or should we have a list of work IDs we couldn't fix?

redsummernight: Let's do both.

sarken (approved):

PR #4137, opened 2022-01-17
AO3-5763 AO3-5348 Add more info to gift notification email and internationalize it

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5348
https://otwarchive.atlassian.net/browse/AO3-5763

## Purpose

Make sure the gift notification email includes the same info we'd send for a subscription notification. Accomplishes this using a partial we can reuse across many emails. I did not update the subscription emails to use the partial, even though it's the same info and formatting -- I can add that change if we'd like.

Also makes the email translatable and fixes an issue where the subject was always in English.

## Testing Instructions

Give people gifts! Check how the emails look in various languages, HTML and plain text! 

## Credit

Sarken, she/her


app/helpers/translation_helper.rb
```diff
@@ -60,4 +60,11 @@ def time_ago_in_words(from_time, include_seconds = false)
 
 alias distance_of_time_in_words_to_now time_ago_in_words
 
+  # Take some text and add whatever punctuation, symbols, and/or spacing
+  # we use to separate a metadata property from its value, e.g., "Property: ",
+  # "Propriété : ".
```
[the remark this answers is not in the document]
sarken: Va te faire enculer, Hound. It's French.

app/views/user_mailer/_work_info.text.erb
```diff
@@ -0,0 +1,12 @@
+<% # expects work %>
+<%= work_metadata_label(Work.human_attribute_name("chapter_total_display"))%><%= work.chapter_total_display %>
+<%= work_tag_metadata(work.fandoms) %>
+<%= work_tag_metadata(work.ratings) %>
+<%= work_tag_metadata(work.archive_warnings) %><% unless work.relationship_string.blank? %>
```
sarken, line 5: (It's like this so we don't end up with extra blank lines in the email when a tag type is not present.)

app/views/user_mailer/recipient_notification.html.erb
```diff
@@ -7,23 +7,13 @@
   </p>
 
   <p>
-    <% if @collection.nil? %>
-      <%= style_creation_link(@work.title, work_url(@work)) %> <%= "(#{@work.word_count} words)" %>
-    <% else %>
-      <%= style_creation_link(@work.title, collection_work_url(@collection, @work)) %> <%= "(#{@work.word_count} words)" %>
-    <% end %>
+    <% url = @collection ? collection_work_url(@collection, @work) : work_url(@work) %>
```
sarken, line 10: (The work title/byline is separate from the other metadata because subscription emails display this info differently.)

app/views/user_mailer/_work_info.text.erb
tickinginstant: Counterintuitively enough, I think you can actually prevent extra blank lines in the output by adding newlines. A template like this produces three blank lines:
```
<% if false %><%= "Text." %><% end %>
<% if false %><%= "Text." %><% end %>
<% if false %><%= "Text." %><% end %>
```
Whereas a template like this produces no blank lines:
```
<% if false %>
<%= "Text." %>
<% end %>
<% if false %>
<%= "Text." %>
<% end %>
<% if false %>
<%= "Text." %>
<% end %>
```
(This is because a `<% %>` tag that's alone on the line doesn't affect the output. But multiple tags in one line are trickier.)

sarken: Oh, that would be much better -- thanks!

app/helpers/translation_helper.rb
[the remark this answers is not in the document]
CristinaRO: Duo serait très fier de toi :P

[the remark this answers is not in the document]
redsummernight: Yeah we don't have to do anything here. This check will eventually be phased out (https://github.com/rubocop/rubocop/issues/9674).

app/views/user_mailer/_work_info.text.erb
```diff
@@ -0,0 +1,22 @@
...
+<% unless work.freeform_string.blank? %>
+<%= work_tag_metadata(work.freeforms) %>
+<% end %>
+<% unless work.series.count.zero? %>
+<%= Series.model_name.human(count: work.series.count) %><%= raw to_plain_text(series_list_for_feeds(work)) %>
+<% end %>
+<% unless work.summary.blank? %>
+
+<%= work_metadata_label(Work.human_attribute_name("summary")) %>
+    <%= raw to_plain_text(sanitize_field(work, :summary)) %>
```
redsummernight, line 21: Should this be indented 2 spaces?

app/helpers/mailer_helper.rb
```diff
@@ -127,4 +127,54 @@ def creator_text(work)
...
+  def work_tag_metadata_label(tags)
+    return if tags.empty?
+
+    type = tags.first.type
+    t("activerecord.models.#{type.underscore}", count: tags.count) + t("mailer.general.metadata_label_indicator")
+  end
+
+  # We don't use .to_sentence because these aren't links and we risk making any
+  # connector word (e.g., "and") look like part of the final tag.
+  def work_tag_metadata_list(tags)
```
redsummernight, line 144: Is this used directly in a view? If so, we should make it private and keep the list of available helpers shorter.

Then we can have specs just for the public ones as well.

app/views/user_mailer/_work_info.text.erb
sarken: Nope, I wanted to match the indenting from the subscription emails (I assume we went with four spaces because it's more obviously done for a pretty formatting effect than two -- two just looks kinda like we misaligned the text): 
https://github.com/otwcode/otwarchive/blob/5cd77dd6d6299fcd8b578115af460cad22ea3e78/app/views/user_mailer/batch_subscription_notification.text.erb#L31-L32

redsummernight (approved):

PR #4150, opened 2022-02-07
AO3-4728 Fix 500 error on subscriptions page for invalid items

tlee911:
## Issue

## Purpose

Subscriptions page will no longer give 500 error if user has invalid subscribables (will show up as "Deleted item" in the subscriptions list, and can be removed).

## Testing Instructions

I was unable to reproduce the problem without direct manipulation of the database, as all the ways listed in the Jira ticket for getting into this state seem to have been fixed previously (e.g. can no longer subscribe to a deleted work by keeping the work page open before the deletion).

I've added unit test in `models/subscription_spec.rb`, but haven't found a way to reliably reproduce without mucking around in the DB, and I'm not sure we want to be doing that in controller or cucumber tests.

For local testing, I created a subscription then modified it via SQL to point to a subscribable ID which does not exist, and verified the subscriptions page would still render, as well as allowing deletion of the invalid subscription.


app/views/subscriptions/index.html.erb
```diff
@@ -38,20 +38,24 @@
 <dl class="subscription index group">
   <% @subscriptions.each do |subscription| %>
     <dt>
-      <%= link_to(subscription.name, subscription.subscribable) %>
+      <% if subscription.subscribable.nil? %>
+      <%= t(".deleted") %>
```
sarken, line 42: Can you indent this two spaces? And can we use `<%= subscription.name %>` here instead now that deleted subscriptions have a `name`? That would save Translation from having to translate the same thing twice.

config/locales/models/en.yml
```diff
@@ -109,3 +109,5 @@ en:
         user_defined_tags_count: "Fandom, relationship, and character tags"
       work:
         user_defined_tags_count: "Fandom, relationship, character, and additional tags"
+  subscriptions:
+    deleted: "DELETED"
```
sarken, line 113: The placeholders in history use "Deleted work," so let's go with "Deleted item" here.

sarken (changes requested):

sarken:
>I've added unit test in models/subscription_spec.rb, but haven't found a way to reliably reproduce without mucking around in the DB, and I'm not sure we want to be doing that in controller or cucumber tests.

Oh, sorry, I forgot to mention -- this is fine! We muck around to create invalid items for controller or Cucumber tests all the time.

app/views/subscriptions/index.html.erb
```diff
@@ -38,20 +38,24 @@
...
+      <% else %>
+        <%= link_to(subscription.name, subscription.subscribable) %>
+      <% end %>
       <% if params[:type].blank? && subscription.subscribable_type == "Work" %>
```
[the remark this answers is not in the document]
tlee911: Can the subscription be nil? I meant to handle the case when the subscribable is nil (due to pointing to an entity that doesn't exist anymore).  If subscription is nil, wouldn't that mean there's no record to render?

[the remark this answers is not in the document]
sarken: Oh, right -- should be fine, then!

config/locales/models/en.yml
```diff
@@ -109,3 +109,5 @@ en:
         user_defined_tags_count: "Fandom, relationship, and character tags"
       work:
         user_defined_tags_count: "Fandom, relationship, character, and additional tags"
+  subscriptions:
+    deleted: "Deleted item"
```
sarken, line 113: Just one more tiny nitpick -- can you add the missing newlines to the end of this file and locales/views/en.yml?

sarken (commented):
The changes look good, but I think it's still missing tests to prove that the page will load and the subscription can be deleted. (If you'd like an example of how we set up invalid objects for Cucumber tests, [here's a step where we set up an invalid sign-up for gift exchanges](https://github.com/otwcode/otwarchive/blob/b10cff21682ae5a401a5f08488f6c448886a4070/features/step_definitions/challege_gift_exchange_steps.rb#L236-L241).)

tlee911:
> The changes look good, but I think it's still missing tests to prove that the page will load and the subscription can be deleted. (If you'd like an example of how we set up invalid objects for Cucumber tests, [here's a step where we set up an invalid sign-up for gift exchanges](https://github.com/otwcode/otwarchive/blob/b10cff21682ae5a401a5f08488f6c448886a4070/features/step_definitions/challege_gift_exchange_steps.rb#L236-L241).)

Yup, I'm still meaning to do that.  Haven't had the time the last few days to sit down properly and work on it.  But the example is helpful.

spec/controllers/subscriptions_controller_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  context "when logged in" do
+    before { fake_login_known_user(user) }
+
+    context "with valid subscriptions" do
+      let(:author) { create(:user) }
+      let(:work) { create(:work) }
+      let(:series) { create(:series) }
+      let!(:sub_series) { create(:subscription, user: user, subscribable_type: "Series", subscribable_id: series.id) }
```
redsummernight, line 24: You should be able to use `create(:subscription, user: user, subscribable: series)`, etc.

spec/controllers/subscriptions_controller_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+      let!(:sub_work) { create(:subscription, user: user, subscribable_type: "Work", subscribable_id: work.id) }
+      let!(:sub_user) { create(:subscription, user: user, subscribable_type: "User", subscribable_id: author.id) }
+
+      it "renders the user subscriptions" do
+        get :index, params: { user_id: user.login }
+        expect(response).to render_template("index")
+        expect(assigns(:subscriptions)).to satisfy { |subs| subs.size == 3 }
```
redsummernight, line 31: You can write `expect(assigns(:subscriptions)).to contain_exactly(sub_series, sub_work, sub_user)`.

PR #4156, opened 2022-02-12
AO3-6296 Use WebMock to stub external work URLs in specs.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6296

## Purpose

Stub example.org for all rspec tests, and stub the URL used in `./spec/models/external_work_spec.rb:34`.


sarken (approved):
It's so delightfully green!

PR #4158, opened 2022-02-12
AO3-6287 Run Brakeman scan in Github Actions.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6287

## Purpose

- Adds a new Github Actions workflow to run Brakeman.
- Updates Brakeman to version 5.2.1 so that it can produce SARIF files compatible with Github.
- Deletes the `config/brakeman.ignore` file, since Github ignores it.
- Adds a new `config/brakeman.yml` file that instructs brakeman to treat `sanitize_field` as a "safe" method, so that it doesn't flag all of the calls to `raw sanitize_field(object, :field)`.


redsummernight:
I think the new Brakeman check only failed because the base branch has no results yet. But it should be fine once merged, and only alerts applying to a PR's diff will be shown.

redsummernight (approved):

PR #4159, opened 2022-02-13
AO3-5628 Invalidate chapter count cache when saving drafts

melinath:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5628

## Purpose

Ensures that work chapter counts get properly updated when a draft chapter is added, which is important to ensuring that the "manage chapters" button shows up if there are two chapters and one is draft.

## Testing Instructions

On Jira issue

## Credit

Stephen Lewis, he/him

### DIFF
--- a/spec/controllers/chapters_controller_spec.rb
+++ b/spec/controllers/chapters_controller_spec.rb
@@ -465,7 +465,7 @@ describe ChaptersController do
               post :create, params: { work_id: work.id, chapter: chapter_attributes, post_without_preview_button: true }
             end.to change { work.number_of_chapters }
               .from(1).to(2)
-              .and change {work.number_of_posted_chapters}
+              .and change { work.number_of_posted_chapters }
               .from(1).to(2)
           end
 
@@ -515,7 +515,7 @@ describe ChaptersController do
               post :create, params: { work_id: work.id, chapter: chapter_attributes, preview_button: true }
             end.to change { work.number_of_chapters }
               .from(1).to(2)
-              .and avoid_changing {work.number_of_posted_chapters}
+              .and avoid_changing { work.number_of_posted_chapters }
           end
 
           it "gives a notice that the chapter is a draft and redirects to the chapter preview" do
@@ -990,7 +990,7 @@ describe ChaptersController do
             delete :destroy, params: { work_id: work.id, id: chapter2.id }
           end.to change { work.number_of_chapters }
             .from(2).to(1)
-            .and change {work.number_of_posted_chapters}
+            .and change { work.number_of_posted_chapters }
             .from(2).to(1)
         end
 
@@ -1064,7 +1064,7 @@ describe ChaptersController do
             delete :destroy, params: { work_id: work.id, id: chapter2.id }
           end.to change { work.number_of_chapters }
             .from(2).to(1)
-            .and avoid_changing {work.number_of_posted_chapters}
+            .and avoid_changing { work.number_of_posted_chapters }
         end
       end
     end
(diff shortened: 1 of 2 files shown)


melinath:
I was a little surprised to find that the `destroy` tests I added passed without me needing to do anything - I don't see any logic in that part of the controller to invalidate the cache, but maybe I was missing something?

Also, I didn't add a feature test since this seemed like a pretty straightforward issue with the controller (is number_of_chapters correct after a draft is created) and the feature tests are slower - but I'd be happy to add one if that would be preferred.

sarken:
Yeah, that's a flaky one. I've added it to the list on the wiki and restarted the tests.

spec/controllers/chapters_controller_spec.rb
```diff
@@ -473,6 +473,15 @@
             expect(assigns[:chapter].posted).to be true
           end
 
+          it "updates cached chapter counts" do
+            work.invalidate_work_chapter_count(work)
```
redsummernight, line 477: You should be able to remove the `invalidate_work_chapter_count` calls from the tests, as they should be correctly called by the controller.

spec/controllers/chapters_controller_spec.rb
```diff
@@ -473,6 +473,15 @@
...
+            expect(work.number_of_chapters).to eq 1
+            expect(work.number_of_posted_chapters).to eq 1
+            post :create, params: { work_id: work.id, chapter: chapter_attributes, post_without_preview_button: true }
+            expect(work.number_of_chapters).to eq 2
+            expect(work.number_of_posted_chapters).to eq 2
```
redsummernight, line 482: Since you're checking the same fields before and after an action, you can use the change matcher, like this:

https://github.com/otwcode/otwarchive/blob/1f8558452e6984a41ace47074324013d2a3454f8/spec/controllers/admin/admin_users_controller_spec.rb#L109-L122

`.from` and `.to` should be short enough to live on the same line as the `change` block in this case.

melinath: done in [c9b6691](https://github.com/otwcode/otwarchive/pull/4159/commits/c9b66913424ce335e9aacdf475005d93aed55a09)

spec/controllers/chapters_controller_spec.rb
melinath: Done in https://github.com/otwcode/otwarchive/pull/4159/commits/2fac5f9549464472387c72e0c29721551e08ec78

spec/controllers/chapters_controller_spec.rb
```diff
@@ -460,6 +460,15 @@
             expect(assigns[:chapter].posted).to be true
           end
 
+          it "updates cached chapter counts" do
+            expect do
+              post :create, params: { work_id: work.id, chapter: chapter_attributes, post_without_preview_button: true }
+            end.to change { work.number_of_chapters }
+              .from(1).to(2)
+              .and change { work.number_of_posted_chapters }
```
sarken, line 468: Should we `reload` the work? It seems like we always use that with `avoid_changing`.

tickinginstant: In this specific case, `work` doesn't need reloading to reflect the changes. `number_of_chapters` and `number_of_posted_chapters` are set up with cache keys that only include the work's ID (so reloading the work won't change the cache key), and the contents of the cache are set using `.count` (which forces a DB query even if the association is already loaded).

sarken (approved):

PR #4160, opened 2022-02-13
AO3-4488 Add VCR back to some tests, and cache the cassettes in the automated tests.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4488

## Purpose

- Removes all files from the `features/cassette_library` directory, and adds the directory to the ignore file.
- Adds `main` as an alternative to `master` in the automated tests, to make it easier for people who have renamed their default branch from `master` to `main`.
- Modify the automated tests so that the command and the arguments are passed separately. This allows us to eliminate the `cucumber` setting (since we can just check whether the command is `cucumber` instead), and allows us to use `hashFiles(matrix.tests.arguments)` as part of a cache key, to invalidate the cache when one of the tests in the group is modified.
- Adds an extra step, used only for the `cucumber features/importing` tests, that loads the contents of the `features/cassette_library` directory from cache, and writes it back to the cache when it's done.
- Adds VCR back to some tests that used to have it.


.github/workflows/automated-tests.yml
```diff
@@ -86,20 +105,40 @@ jobs:
...
+          # key to be overwritten if there's an exact match. So instead we add
+          # a unique identifier to the key to always force a "cache miss", and
+          # restore from related keys to make sure that we still get to load a
+          # recent version of the cached folder. More explanation here:
+          #
+          # https://github.com/actions/cache/issues/109#issuecomment-558771281
+          #
+          # It's hacky, but until the cache action is more configurable, it's
+          # the best we can do.
+          key: cassette-library-${{ hashFiles(matrix.tests.arguments) }}-${{ github.run_id }}-${{ github.run_attempt }}
```
redsummernight, line 138: +1, unique across re-runs for the same PR as well.

redsummernight (approved):

PR #4161, opened 2022-02-13
AO3-6146 Limit username changes

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6146

## Purpose

Allow limiting username changes to once every 7 days (configurable)

## Testing Instructions

Added to the Jira ticket

## References

Rake task from a comment in the Jira ticket (see above)

## Credit

**Brian Austin (he/they)**

they/them pronouns.*


app/views/users/change_username.html.erb
```diff
@@ -1,12 +1,23 @@
...
+    <%= ts("You can change your user name %{times} time every %{days} days.",
+      times: ArchiveConfig.USERNAME_CHANGE["TIMES"],
+      days: ArchiveConfig.USERNAME_CHANGE["INTERVAL_DAYS"]) %>
+    <% if @user.renamed_at %>
+      <%= ts("You last changed your user name on %{renamed_at}.", renamed_at: l(@user.renamed_at, format: :long)) %>
+    <% end %>
+  </p>
+  <p>
+    <%= ts("For information on how changing your user name will affect your account, please check out the %{account_faq}. Note that changes to your username may take several days or longer to appear. If you are still seeing your old user name on your works, bookmarks, series, or collections after a week, please %{contact_support}.",
+    account_faq: (link_to ts("Account FAQ"), archive_faqs_path + "/your-account#namechange"),
```
redsummernight, line 16: Because the variable represents a hyperlink, can you rename it [with the suffix _link](https://github.com/otwcode/otwarchive/wiki/Internationalization-(i18n)-Standards)?

redsummernight, line 16: The link should be constructed with `archive_faq_path("your-account", anchor: "namechange")`.

features/other_a/autocomplete.feature
```diff
@@ -98,15 +98,19 @@ Feature: Display autocomplete for tags
...
-    Then the pseud autocomplete should not contain "funny (different_user)"
-      And the pseud autocomplete should contain "funny"
-      And the pseud autocomplete should contain "different_user (funny)"
-    When I try to delete my account as funny
+    When I try to delete my account as different_user
     Then a user account should not exist for "funny"
       And the pseud autocomplete should not contain "funny"
       And the pseud autocomplete should not contain "different_user (funny)"
 
+  Scenario: Pseuds should be listed in autocomplete when username changes to pseud
```
redsummernight, line 106: I think we can remove this scenario. We're testing much more for autocomplete here. Plus there's already:

https://github.com/otwcode/otwarchive/blob/d42d1bbe5b6ff49fa4bf75e6a5b11738341eb52f/features/users/user_rename.feature#L95

spec/models/user_spec.rb
```diff
@@ -127,6 +127,58 @@
     end
   end
 
+  describe "#update" do
+    let(:existing_user) { create(:user) }
+
+    it "sets renamed_at if username changed" do
+      existing_user.update(login: "new_username")
+      expect(existing_user.renamed_at).to be_within(1.minute).of(Time.now.utc)
```
redsummernight, line 135: You can [freeze time](https://api.rubyonrails.org/v5.2/classes/ActiveSupport/Testing/TimeHelpers.html#method-i-freeze_time) here as well. Then you can just check that it's equal to `Time.now`.

spec/models/user_spec.rb
```diff
@@ -127,6 +127,58 @@
...
+    end
+
+    it "does not set renamed_at when username not changed" do
+      existing_user.update(email: "newemail@example.com")
+      expect(existing_user.renamed_at).to be_nil
+    end
+
+    context "username recently changed" do
+      before do
+        Timecop.freeze
```
redsummernight, line 145: You can use [Rails' time helpers](https://api.rubyonrails.org/v5.2/classes/ActiveSupport/Testing/TimeHelpers.html#method-i-travel_to), which automatically remove the stubs at the end of specs, so you don't need an `after`.

spec/models/user_spec.rb
```diff
@@ -127,6 +127,58 @@
...
+        existing_user.update(login: "#{existing_user.login}2")
```
redsummernight, line 146: I'd prefer if the new values are simply typed out in full. They do not need to strictly be variations of the old value, so they do not need interpolation.

config/config.yml
```diff
@@ -473,3 +473,7 @@ PERMITTED_HOSTS: [
   "test.archiveofourown.org",
   "testdownload.archiveofourown.org"
 ]
+
+USERNAME_CHANGE:
```
redsummernight, line 477: Can you use `USER_RENAME_LIMIT` and `USER_RENAME_LIMIT_DAYS` for these options?

Generally we avoid nesting options.

(`RATE_LIMIT_PER_NGINX_UPSTREAM` is an outlier because we need potentially add new limits on the fly.)

app/models/user.rb
```diff
@@ -521,6 +523,10 @@ def reindex_user_creations_after_rename
     reindex_user_creations
   end
 
+  def add_renamed_at
+    self.renamed_at = Time.now.utc
```
redsummernight, line 527: With #4169 merged, you can [use `Time.current` here](https://rails.rubystyle.guide/#time-now).

app/models/user.rb
```diff
@@ -530,4 +536,14 @@ def remove_stale_from_autocomplete
     self.class.remove_from_autocomplete(self.autocomplete_search_string_was, self.autocomplete_prefixes, self.autocomplete_value_was)
   end
 
+  def check_username_already_changed
```
redsummernight, line 539: Can you change this method name to `username_is_not_recently_changed`?

redsummernight (changes requested):

app/models/user.rb
```diff
@@ -530,4 +536,14 @@ def remove_stale_from_autocomplete
     self.class.remove_from_autocomplete(self.autocomplete_search_string_was, self.autocomplete_prefixes, self.autocomplete_value_was)
   end
 
+  def username_is_not_recently_changed
+    change_interval_days = ArchiveConfig.USER_RENAME_LIMIT_DAYS
+    return unless renamed_at && change_interval_days.days.ago <= renamed_at
```
tickinginstant, line 541: It doesn't look like `USER_RENAME_LIMIT` is used in this calculation, only in the error message. So if you did set it to 2, you'd get the "Login can only be changed 2 times every 7 days." message, but you'd still only be able to rename yourself once every seven days. Or am I missing something?

(Actually, thinking about it ... I think in order to make `USER_RENAME_LIMIT` configurable, you have to use the audits table to check the number of time the user has renamed in the last `USER_RENAME_LIMIT_DAYS` days, since the time of the most recent rename won't tell you how many other renames the user has done recently. So there's not much point in introducing the `renamed_at` column, since the calculation will be based on the audits table regardless. I think the description of the issue needs to be revisited — either get rid of the `renamed_at` column and rely on the audits table, or get rid of the ability to configure `USER_RENAME_LIMIT`.)

sarken: I'd lean toward getting rid of `USER_RENAME_LIMIT` -- I don't really see us going with twice a week or three times every two weeks or something weird like that. Once per a configurable period of time seems fine.

lib/tasks/after_tasks.rake
```diff
@@ -362,6 +362,26 @@ namespace :After do
 
   #### Add your new tasks here
 
+  desc "Backfill renamed_at for existing users"
```
redsummernight, line 365: Can you move the new task to the end of the list?

https://github.com/otwcode/otwarchive/blob/086e6f08bd1578943cf1ddf2ca14c0cdcdb4e210/lib/tasks/after_tasks.rake#L877

app/views/users/change_username.html.erb
```diff
@@ -1,12 +1,23 @@
...
+    <%= t("users.change_username.change_window",
+      times: t("users.change_username.times", count: ArchiveConfig.USER_RENAME_LIMIT),
+      days: distance_of_time_in_words_to_now(ArchiveConfig.USER_RENAME_LIMIT_DAYS.days.ago)) %>
+    <% if @user.renamed_at %>
+      <%= t("users.change_username.last_renamed", renamed_at: l(@user.renamed_at, format: :long)) %>
+    <% end %>
+  </p>
+  <p>
+    <%= t("users.change_username.more_info",
+    account_faq_link: (link_to t("users.change_username.account_faq"), archive_faq_path("your-account", anchor: "namechange")),
```
redsummernight, line 16: Can you indent these lines similar to the `users.change_username.change_window` section?

app/views/users/change_username.html.erb
```diff
@@ -1,12 +1,23 @@
...
-<p class="caution">
-  <%= ts("Please use this feature with caution. For information on how changing your user name will affect your account, please check out the %{account_faq}. Note that changes to your username may take several days or longer to appear. If you are still seeing your old user name on your works, bookmarks, series, or collections after a week, please %{contact_support}.",
-  account_faq: (link_to ts("Account FAQ"), archive_faqs_path + "/your-account#namechange"),
-  contact_support: (link_to ts("contact Support"), new_feedback_report_path)
-  ).html_safe %>
-</p>
+<div class="caution">
+  <p>
+    <strong><%= t("users.change_username.caution") %></strong>
+    <%= t("users.change_username.change_window",
```
redsummernight, line 7: In general we try to avoid concatenation when setting up I18n keys. In this case, the way "every X days" gets translated will be dependent on the language (e.g. because of plural rules) as well as the sentence containing it. The same applies to the "time ago" (e.g. [German](https://stackoverflow.com/questions/53013887/rails-is-there-a-built-translation-for-ago#comment113736195_53014319)) string on the inbox comments.

So I think we should:

- Set up the I18n keys for full sentences, like this (we skip "x times" as well since we're getting rid of `USER_RENAME_LIMIT`):
  ```yml
  change_window:
    one: "You can change your user name once every day."
    other: "You can change your user name once every %{count} days."
  ```
- Leave the custom helpers in `app/helpers/translation_helper.rb` and the inbox comment view alone for now.

app/models/user.rb
brianjaustin: Ah, thanks for asking -- I meant to bring this up but totally forgot!

spec/models/user_spec.rb
```diff
@@ -127,6 +127,54 @@
...
+      existing_user.update(login: "new_username")
+      expect(existing_user.renamed_at).to eq(Time.current)
+    end
+
+    it "does not set renamed_at when username is not changed" do
+      existing_user.update(email: "newemail@example.com")
+      expect(existing_user.renamed_at).to be_nil
+    end
+
+    context "username recently changed" do
```
redsummernight, line 144: Can you write this as "when username was recently changed"?

spec/models/user_spec.rb
```diff
@@ -127,6 +127,54 @@
...
+      before do
+        freeze_time
+        existing_user.update(login: "new_login")
+      end
+
+      it "does not allow another rename" do
+        expect(existing_user.update(login: "new")).to be_falsey
+        localized_renamed_at = I18n.l(Time.now.in_time_zone(existing_user.renamed_at.time_zone), format: :long)
```
redsummernight, line 152: Since we froze time before the test, instead of `Time.now.in_time_zone(existing_user.renamed_at.time_zone)`, does `existing_user.renamed_at` work?

redsummernight (commented):
Approach looks good to me! Just some minor comments.

redsummernight (approved):
Thank you!

redsummernight:
Would you be able to update this pull request to fix the merge conflicts for this upcoming release? The newest task should be kept at the bottom of the file.

Let us know if you'd like us to handle them instead.

brianjaustin:
> Would you be able to update this pull request to fix the merge conflicts for this upcoming release? The newest task should be kept at the bottom of the file.
> 
> Let us know if you'd like us to handle them instead.

I can take a look, hopefully sometime this week/weekend

redsummernight:
Thank you!

PR #4168, opened 2022-02-18
AO3-6253 Localize sent_at time in emails

ceithir:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6253

## Purpose

Translate the "sent at" date to the user locale.

## Testing Instructions

The testing instructions in the Jira ticket should work fine.

## References

The chosen solution is by the book Rails date/time localization: https://guides.rubyonrails.org/v2.3/i18n.html#adding-datetime-formats

## Missing translations (?)

While the date will technically be translated with that PR, the formatting is likely to still be on the weird side in languages other than English. No real solution but to fix this language per language I fear.

## Credit

Ceithir (he/him)


features/step_definitions/preferences_steps.rb
```diff
@@ -140,3 +140,7 @@
   user.preference.adult = false
   user.preference.save
 end
+
+Given /^the user locale is "([^"]*)"$/ do |locale|
+  I18n.locale = locale
```
redsummernight, line 145: Our translated emails use a setup like this:

https://github.com/otwcode/otwarchive/blob/5cd77dd6d6299fcd8b578115af460cad22ea3e78/app/mailers/user_mailer.rb#L61-L67

I think setting `I18n.locale` to test the abuse report confirmation email only works because it doesn't use that setup:

https://github.com/otwcode/otwarchive/blob/5cd77dd6d6299fcd8b578115af460cad22ea3e78/app/mailers/user_mailer.rb#L422-L431

So I'd rather not add this step for now. If we do add it, it should set up `preferred_locale` on the user's preferences.

Also in this case, we would be using `l(Time.now)` for a well-known format, so we can get away with not adding a specific test for it.

redsummernight (changes requested):

config/initializers/archive_config/time_formats.rb
```diff
@@ -1,5 +1,2 @@
 Date::DATE_FORMATS[:date_for_comment_test] = "%a, %d %b %Y"
```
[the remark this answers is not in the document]
ceithir: https://github.com/otwcode/otwarchive/pull/4168/commits/7533f38e63edcb8617cd36a709931add94fb0328

features/step_definitions/preferences_steps.rb
ceithir: Change to Cucumber tests reverted and replaced with a simpler, shorter and more explicit Rspec test: https://github.com/otwcode/otwarchive/pull/4168/commits/a2c477f5717bf4c8b03ad718aec1b7f019f556f1

app/views/admin_mailer/feedback.html.erb
```diff
@@ -26,7 +26,7 @@
 <% end %>
 
 <% content_for :sent_at do %>
-  <%= Time.now.to_s(:time_for_mailers) %>
+  <%= l Time.now, :format => :time_for_mailers %>
```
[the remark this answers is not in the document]
redsummernight: Yeah, though we haven't really been trying to use `localize` until recently. So whichever format we use would need tweaking at some point.

For this issue, I think we can just use `:default`. Later we can override `time.formats.default` on a per-locale basis as needed.

spec/mailers/user_mailer_spec.rb
```diff
@@ -608,6 +608,28 @@
...
+
+      it "formats the date rightfully in English" do
+        expect(email).to have_html_part_content("Sent at 01:27PM UTC Mon 14 March 2022.")
+        expect(email).to have_text_part_content("Sent at 01:27PM UTC Mon 14 March 2022.")
+      end
+
+      context "french" do
+        before { I18n.locale = "fr" }
+
+        after { I18n.locale = I18n.default_locale }
```
redsummernight, line 625: I think you can skip this. We reset it here:

https://github.com/otwcode/otwarchive/blob/1f8558452e6984a41ace47074324013d2a3454f8/spec/spec_helper.rb#L67-L68

spec/mailers/user_mailer_spec.rb
```diff
@@ -608,6 +608,28 @@
         expect(email).to have_text_part_content(report.url)
       end
     end
+
+    describe "translation" do
+      before { Timecop.freeze("2022-03-14 13:27:09 +0000") }
```
redsummernight, line 613: Can you use Rails' [time helpers](https://api.rubyonrails.org/v6.0/classes/ActiveSupport/Testing/TimeHelpers.html) instead?

app/views/admin_mailer/feedback.html.erb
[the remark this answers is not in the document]
ceithir: Sure: https://github.com/otwcode/otwarchive/pull/4168/commits/403b628cb957101e2c3de3340255c18496aaa830

features/step_definitions/email_custom_steps.rb
```diff
@@ -40,8 +40,9 @@
 end
 
 Then /^the email should say what time it was sent$/ do
-  # mailer time format: 11:38PM EDT Thu 29 September 2016
-  nowish = Time.now.strftime('%I:%M%p %Z %a %d %B %Y')
+  # Rails default: https://github.com/rails/rails/blob/v7.0.2.3/activesupport/lib/active_support/locale/en.yml#L25
+  # But truncated to the day so the test don't need to run under a second to pass
+  nowish = Time.now.utc.strftime('%a, %d %b %Y')
   step %{the email should contain "Sent at #{nowish}"}
   step %{the email should not contain "sent_at"}
```
ceithir, line 47: I'm wondering if that step shouldn't simply be removed instead of maintained. It's only used in a single test, and that test might not be relevant anymore, as the original [commit](https://github.com/otwcode/otwarchive/commit/40bce3aa59b799d5f6724d80f88197e08dc8bb85) references timezones, and whether timezones appear or not in the text is now dependent on Rails' default localization (cf the example, where they appear in the English version but not the French one).

redsummernight: Yeah that makes sense. Let's remove the step. For the test, I think you can freeze time at the start and do a simple "should see" check for the time string, however it's formatted right now. As an example:

https://github.com/otwcode/otwarchive/blob/1f8558452e6984a41ace47074324013d2a3454f8/features/other_a/reading.feature#L29-L34

ceithir: https://github.com/otwcode/otwarchive/pull/4168/commits/23ed696266a8f846f21f441eb98a94d0af942aac

redsummernight (approved):
Looks good, thank you!

sarken:
Hi, Ceithir! 

Thank you so much for this pull request. I think we might've forgotten to mention that if you'd like the ability to comment on, assign, and transition issues in the future, you're welcome to create a Jira account! You can just reply here with the account name and we'll set up the permissions for you. (It makes things a bit easier for us on the organizational side if the Full Name on your Jira account either closely matches the name you'd like us to credit in the release notes or includes it in parentheses, e.g. "Nickname (PREFERRED NAME).")

PR #4175, opened 2022-02-27
AO3-6306 Can't bookmark URLs with response codes of 308 or 307

dark-veracity:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6306 

## Purpose

When a user attempts to bookmark an external work, the backend checks that the URL they entered is not a dead link by sending it a HTTP HEAD request. The code to handle this accepts a response of 200, 301, or 302 as a live URL but since it was written before 307 and 308 redirects were added to the HTTP spec, it fails on websites that return those response codes. This PR fixes that.

## Testing Instructions

1. Log in
2. Hi, username! > My Bookmarks > Bookmark External Work
3. Enter a URL that will give a response code of either 307 or 308, e.g. something from https://mangadex.org/ will give a 308
4. Fill in other required fields
5. Press "Create"
6. The bookmark should be created successfully

## References

N/A

## Credit

dark-veracity (She/Her)

### DIFF
--- a/lib/url_helpers.rb
+++ b/lib/url_helpers.rb
@@ -11,7 +11,7 @@ module UrlHelpers
       begin
         url = Addressable::URI.parse(url)
         response_code = Net::HTTP.start(url.host, url.port) {|http| http.head(url.path.blank? ? '/' : url.path).code}
-        active_status = %w(200 301 302 307 308)
+        active_status = %w[200 301 302 307 308]
         active_status.include? response_code
       rescue
         false
--- a/lib/url_helpers.rb
+++ b/lib/url_helpers.rb
@@ -11,7 +11,7 @@ module UrlHelpers
       begin
         url = Addressable::URI.parse(url)
         response_code = Net::HTTP.start(url.host, url.port) {|http| http.head(url.path.blank? ? '/' : url.path).code}
-        active_status = %w(200 301 302)
+        active_status = %w(200 301 302 307 308)
         active_status.include? response_code
       rescue
         false
--- a/lib/url_helpers.rb
+++ b/lib/url_helpers.rb
@@ -2,23 +2,6 @@ require 'timeout'
 require 'uri'
 
 module UrlHelpers
-  # Checks the status of the webpage at the given url
-  # To speed things up we ONLY request the head and not the entire page.
-  # Bypass check for fanfiction.net because of ip block
-  def url_active?(url, timeout_in_seconds=60)
-    return true if url.match("fanfiction.net")
-    Timeout::timeout(timeout_in_seconds) {
-      begin
-        url = Addressable::URI.parse(url)
-        response_code = Net::HTTP.start(url.host, url.port) {|http| http.head(url.path.blank? ? '/' : url.path).code}
-        active_status = %w[200 301 302 307 308]
-        active_status.include? response_code
-      rescue
-        false
-      end
-    }
-  end
-
   # Make urls consistent
   def reformat_url(url)
     url = url.gsub(/https/, "http")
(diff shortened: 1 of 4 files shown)


lib/url_helpers.rb
```diff
@@ -11,7 +11,7 @@ def url_active?(url, timeout_in_seconds=60)
       begin
         url = Addressable::URI.parse(url)
         response_code = Net::HTTP.start(url.host, url.port) {|http| http.head(url.path.blank? ? '/' : url.path).code}
-        active_status = %w(200 301 302)
+        active_status = %w[200 301 302 307 308]
```
sarken, line 14: It looks like this is actually the wrong place for this change: I pulled the code and I still got an error message when trying to bookmark something on Mangadex. (It's possible this code is unused and can be removed.) However, bookmarking after updating app/validators/url_active_validator.rb instead works just fine.

It would be a good idea to add automated tests to spec/models/external_work_spec.rb to prove that the code works.

sarken (changes requested):
Thanks for working on this! I've assigned the issue to you on Jira and given your account permission to claim, transition, and comment on issues.

lib/url_helpers.rb
tickinginstant: > It's possible this code is unused and can be removed.

For the record, the `url_active?` method *is* called in `ExternalWork#set_url_status`, but `set_url_status` isn't called anywhere, so both methods should probably be deleted sooner or later.

dark-veracity: Oh that makes sense. Would you like me to remove the unused code in this PR or should I leave it and it can be taken out in a separate cleanup PR?

redsummernight: Let's remove them in this PR.

dark-veracity:
I made the changes in url_active_validator, removed the unused code, and added some automated tests for bookmarking an external work that sends back http response codes other than 200. This should be ready to get reviewed again

spec/models/external_work_spec.rb
```diff
@@ -30,13 +30,32 @@
     URLS = ["http://the--ivorytower.livejournal.com/153798.html"]
 
     URLS.each do |url|
-      let(:valid_url) {build(:external_work, url: url)}
+      # Test each of the possible valid response codes
+      WebMock.stub_request(:any, url).to_return({ status: 200, body: "Success" }, { status: 301, body: "Moved Permanently" }, { status: 302, body: "Found" }, { status: 307, body: "Temporary Redirect" }, { status: 308, body: "Permanent Redirect" })
```
tickinginstant, line 34: This line doesn't seem to work the way it's intended to. If I check out this commit but revert the changes to `app/validators/url_active_validator.rb`, it should fail on the 307/308 tests. Instead, all five tests pass.

This line of code isn't in a `let`, a `before`, or inside a test. That means it gets run once when the file is first loaded. And I think WebMock cleans up any mocked requests after each test, so this test is actually performing an external request each time. (With the possible exception of the very first test.)

spec/models/external_work_spec.rb
```diff
@@ -30,13 +30,35 @@
     URLS = ["http://the--ivorytower.livejournal.com/153798.html"]
 
     URLS.each do |url|
-      let(:valid_url) {build(:external_work, url: url)}
```
redsummernight, line 33: We don't really use `URLS`, so let's make it more concise. You can replace the entire `context "valid urls"` block with:
```rb
context "for valid URLs" do
  [200, 301, 302, 307, 308].each do |status|
    context "returning #{status}" do
      let(:external_work) { build(:external_work) }

      before do
        WebMock.stub_request(:any, external_work.url).to_return(status: status)
      end

      it "saves" do
        expect(external_work.save).to be_truthy
      end
    end
  end
end
```

dark-veracity: Done! That's a much cleaner way of handling it

redsummernight (approved):
Thank you!

PR #4144, opened 2022-02-01
AO3-5830 Tag search: Allow searching by fandom, searching for non-canonical tags, and sorting results

weeklies:
# Pull Request Checklist

* [x] Have you read ["How to write the perfect pull request"](https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/)?
* [x] Have you read the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [x] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)?
* [x] Have you added the [Jira](https://otwarchive.atlassian.net) issue number
  as the *first* thing in your pull request title (e.g. `AO3-1234 Fix thing`)
* [x] Do you have fewer than 5 pull requests already open? If not, please wait
  until they are reviewed and merged before creating new pull requests.

## Issue

https://otwarchive.atlassian.net/browse/AO3-5830

## Purpose

1. Create new Tag Search form with sorting and searching by fandom
2. Update and create new tests 
3. Remove old form and update Browse Tags page

## Testing Instructions

- Try searching by tag name, tag type, wrangling status 
- Try sorting by uses and kudos
- Try filtering by fandom(s) (Test for single, multiple, non-existent and non-canonical fandoms)
- Check the Browse Tags page for new description of tag cloud and link to Search Tags

## Credit

weeklies (she/her)

PR #4181, opened 2022-03-11
AO3-6291 Resolve ES sorting ties

korrien:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6291

## Purpose

Adds secondary sort criteria to existing queries.
Updated:
- Works as most affected
- Bookmarks as possible in theory (though I doubt it would cause problems in practice)
- Tags for company (I didn't find where we search and sort for tags, but I think it wouldn't hurt)

PseudsQuery is missing sort, so did nothing there.
Default one also left as is as it is too unclear what benefits and consequences change could bring. 

Had to use _id for bookmarkables. ES discourages searching by _id, but I don't see other options. But as it's my first real encountered with ES, I might have misunderstood something.

## Credit

korrien, she/her


app/models/search/bookmarkable_query.rb
```diff
@@ -98,9 +98,9 @@ def bookmarkable_query_text
   # field and sort by score).
   def sort
     if sort_column == "bookmarkable_date"
-      { revised_at: { order: sort_direction, unmapped_type: "date" } }
+      { revised_at: { order: sort_direction, unmapped_type: "date" }, _id: { order: sort_direction } }
```
redsummernight, line 101: > Had to use _id for bookmarkables. ES discourages searching by _id, but I don't see other options.

I'm not sure if we should. ES has plans to eventually prevent using `_id` in sorting (https://github.com/elastic/elasticsearch/issues/64511). They added automatic tie breakers for `search_after`/point-in-time requests (https://github.com/elastic/elasticsearch/pull/68833), but for other use cases like ours, here's [what they recommend](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/paginate-search-results.html):

> If you cannot use a PIT, we recommend that you include a tiebreaker field in your `sort`. This tiebreaker field should contain a unique value for each document. If you don’t include a tiebreaker field, your paged results could miss or duplicate hits.

I'd wait for a second reviewer, but I think we should index bookmarkables with an "id" field (like we already did with works and the others). Because bookmarkables can be different object types with the same primary key, "id" should have the type for disambiguation, and we can probably reuse `document_id`:

https://github.com/otwcode/otwarchive/blob/1f8558452e6984a41ace47074324013d2a3454f8/app/models/search/bookmarkable_indexer.rb#L31-L33

If we use that approach, we'll need to reindex all works and bookmarkables (there are already rake tasks for that from #3987).

app/models/search/tag_query.rb
```diff
@@ -54,7 +54,7 @@ def sort
       column = "name.keyword"
       direction ||= "asc"
     end
-    sort_hash = { column => { order: direction } }
+    sort_hash = { column => { order: direction }, "id" => { order: direction } }
```
redsummernight, line 57: > I didn't find where we search and sort for tags, but I think it wouldn't hurt

We can't sort tags from the UI yet, but that's a planned feature (#4144) so let's definitely add this.

spec/models/search/bookmark_search_form_spec.rb
```diff
@@ -62,13 +62,33 @@
...
-          work2.update_attribute(:revised_at, Time.now)
+          work2.update_attribute(:revised_at, Time.current)
           run_all_indexing_jobs
           results = BookmarkSearchForm.new(
             parent: tag, sort_column: "bookmarkable_date"
           ).bookmarkable_search_results
           expect(results.map(&:title)).to eq ["Two", "One", "Three"]
         end
+
+        it "doesn't change tied bookmarkables order on work update" do
```
redsummernight, line 73: Without the tie breaker fixes, do these tests always fail or only sometimes fail?

app/models/search/bookmarkable_query.rb
zz9pzza: I think red's comment makes sense.

app/models/search/bookmark_query.rb
```diff
@@ -124,7 +124,7 @@ def default_sort
   end
 
   def sort
-    sort_hash = { sort_column => { order: sort_direction } }
+    sort_hash = { sort_column => { order: sort_direction }, "id" => { order: sort_direction } }
```
tickinginstant, line 127: I'm not sure if I like having both sort columns as keys in the same hash. It means that we're relying on both the ruby end and the elasticsearch end to respect the order of the keys, and while it seems that both ends do respect the order in the current implementation, I'm not sure whether that's guaranteed as part of the official specifications. I'd prefer if this was an array of hashes, since the order there is guaranteed to be correct.

app/models/search/bookmarkable_query.rb
korrien: Thanks for explanations!
Added something, looks to be working.

- Named field "id" as suggested. Hope there won't be any clashes with bookmarks "id" in the same index as these are separate objects.
- Added mapping in the main bookmarks mapping as other places complained. Can't find in ES documents some values from the same mapping, like "creators", so not sure if there are not used anymore or if I put my code into incorrect place.
- I understand that rake jobs would be run after code deployment. But I am not sure why works must be reindexed if "_id" update touches only bookmarks index.

spec/models/search/bookmark_search_form_spec.rb
korrien: They failed every time I ran them.

app/models/search/bookmarkable_query.rb
redsummernight: > Can't find in ES documents some values from the same mapping, like "creators", so not sure if there are not used anymore or if I put my code into incorrect place.

The bookmarks index contains 4 types of documents: bookmarks + 3 types of bookmarkables (works, series, external works). The mapping is shared between all 4 types:

https://github.com/otwcode/otwarchive/blob/1f8558452e6984a41ace47074324013d2a3454f8/app/models/search/bookmarkable_indexer.rb#L11-L13

So `creators` is not used by bookmarks, but is used by the other 3.

> But I am not sure why works must be reindexed if "_id" update touches only bookmarks index.

Works appear in 2 indices: one for work search (using `WorkIndexer`/`WorkCreatorIndexer`) and one for bookmark search (as a bookmarkable type, using `BookmarkedWorkIndexer`). Because we change how bookmarkables are indexed, we need to reindex works, but only with `BookmarkedWorkIndexer`.

We'll run this rake task specifically:

https://github.com/otwcode/otwarchive/blob/1f8558452e6984a41ace47074324013d2a3454f8/lib/tasks/search.rake#L33-L38

spec/models/search/work_query_spec.rb
```diff
@@ -176,4 +176,20 @@
     expect(q.date_range_filter.dig(:range, :revised_at, :gte)).to eq(Date.new(0, 12, 26))
     expect(q.date_range_filter.dig(:range, :revised_at, :lte)).to eq(Date.new(9999, 11, 27))
   end
+
+  it "keeps sort order of tied works the same when work info is updated" do
```
redsummernight, line 180: Can you move this spec next to similar tests in `work_search_form_spec.rb`?

spec/models/search/bookmark_search_form_spec.rb
```diff
@@ -62,13 +62,33 @@
...
+          end
+          run_all_indexing_jobs
+          search = BookmarkSearchForm.new(
+            parent: tag, sort_column: "bookmarkable_date"
+          )
+          res = search.bookmarkable_search_results.map(&:title)
+
+          works.each do |work|
+            work.update(summary: "Updated")
+            work.update_attribute(:revised_at, revised_at)
```
redsummernight, line 87: Why do we set `revised_at` for a second time here? Does it get changed otherwise?

app/models/search/bookmark_query.rb
```diff
@@ -124,13 +124,13 @@ def default_sort
   end
 
   def sort
-    sort_hash = { sort_column => { order: sort_direction } }
+    sort_array = [{ sort_column => { order: sort_direction } }, { "id" => { order: sort_direction } }]
```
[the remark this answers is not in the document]
tickinginstant: Alternatively, you could consider reverting the changes to this method, and then making the last line:
```ruby
[sort_hash, { id: { order: sort_direction } }]
```
I think that should make the braces a little clearer, while also making it easier to modify the hash (without having to call `sort_array[0]`).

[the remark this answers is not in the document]
redsummernight: +1, this is cleaner.

spec/models/search/bookmark_search_form_spec.rb
korrien: It's not really seemed to be needed now, but I added it on purpose.

As I go for update through controller, some functionality regarding updating revised date might be touched. Maybe it won't ever be, maybe it will in some rare cases. If something is changed in the future, I do not want tests to silently become invalid.
As prerequisite for this test is that revised_date is always the same, I want to always assure it explicitly.

redsummernight: Gotcha. In that case, instead of setting `revised_at` a second time, let's add an expect that `revised_at` is the same as the initial value.

spec/models/search/bookmark_search_form_spec.rb
```diff
@@ -244,6 +284,41 @@
...
+
+    context("when sorting") do
+      let!(:time) { Time.current }
+      let(:work1) { create(:work, revised_at: time) }
+      let(:work2) { create(:work, revised_at: time) }
+      let!(:bookmark1) { create(:bookmark, bookmarkable: work1, pseud: bookmarker.default_pseud, created_at: time) }
+      let!(:bookmark2) { create(:bookmark, bookmarkable: work2, pseud: bookmarker.default_pseud, created_at: time) }
+
+      %w[created_at bookmarkable_date].each do |sort_column|
+        it "doesn't change tied order on work update when sorted by #{sort_column}" do
```
redsummernight, line 296: Can you change these spec descriptions to mention "tied bookmark order"?

spec/models/search/bookmark_search_form_spec.rb
```diff
@@ -244,6 +284,41 @@
         expect(result).to eq bookmark
       end
     end
+
+    context("when sorting") do
```
redsummernight, line 288: There is a previous block that's structured like this:
```rb
describe "bookmarkable_search_results" do
  describe "sorting" do
  end
end
```
So for the bookmark sorting specs, can you put them under this structure?
```rb
describe "search_results" do
  describe "sorting" do
  end
end
```

spec/models/search/bookmark_search_form_spec.rb
```diff
@@ -215,6 +215,75 @@
     end
   end
 
+  describe "search_results" do
```
redsummernight, line 218: Sorry, I should have made it clear that I meant for the describe blocks to use the name of the methods being tested. So the bookmarkable specs (calling `bookmarkable_search_results`) were fine where they were, only the bookmark specs (calling `search_results` should be moved to the new block.

spec/models/search/bookmark_search_form_spec.rb
korrien: Done, though I think "expects" is about expected outcome of what is tested and "revised_at" is not the point of the test. So I like it a bit less then setting the value.
But as we don't have any other prerequisite assertion, I am fine with it.

redsummernight: Yeah, you're right. Plus I forgot how annoying it is to test date time equality instead of freezing time.

Setting the value should be fine, if you add a comment mentioning that `revised_at` should stay the same, and is the sorting option being tested.

tickinginstant: I think it'd be cleaner to just freeze time. If you place the freeze in the block before the `let!`s, it should mean that you don't have to manually set any dates.

korrien: I've updated for now with a comment. Not sure how freeze should work.
```
Timecop.freeze do 
 # all the lets and stuff but without any reference to time (no revised_at, no created_at)
```
apparently didn't freeze time without explicitly setting it in lets. So maybe I misunderstood.
Also if it works like that it will clash with bookmarkable_search_result tests and I am already not sure what I would be expected to move where 😅

spec/models/search/bookmark_search_form_spec.rb
korrien: Nope, that's me who can't get a grasp on it 😄 
Thank you, done.

spec/models/search/bookmark_search_form_spec.rb
tickinginstant: It should be pretty straightforward to freeze time in the `search_results` tests (try calling just `before { freeze_time }`).

As for the tests in this block, I'd be inclined to move them into a separate context where you can freeze time (e.g. `"when there are a variety of times"` vs `"when everything was created at the same time"`). The existing tests -- and all of the objects defined in them -- were designed around the exact opposite of the thing that you're trying to test for, which is contributing to the verboseness of the tests you're writing.

korrien: Thank you! All good with correct freeze.

redsummernight (approved):
Thank you!

sarken:
Does this need to be marked deferred until we update the reindexing task? ([AO3-6315](https://otwarchive.atlassian.net/browse/AO3-6315))

redsummernight:
I put up a quick PR for that: #4184.

redsummernight:
Would you be able to update this pull request to fix the merge conflicts for this upcoming release? Let us know if you'd like us to handle them instead.

redsummernight:
Thanks!

PR #4182, opened 2022-03-17
AO3-5955 AO3-5956 AO3-6294 Retry email sending on NoMethodError and ActionView::Template::Error

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5955
https://otwarchive.atlassian.net/browse/AO3-5956
https://otwarchive.atlassian.net/browse/AO3-6294

## Purpose

Sometimes, if an object is deleted around the time that an email is sent about it, the email will successfully load certain objects (thus preventing an `ActiveRecord::RecordNotFound` error that might trigger a retry), but will then fail to load that object's associations. For example, if a user deletes their account around the time that they are supposed to receive an email, it's possible that the email job will be able to load the `User` object, but not the associated `Preference` object. This produces an error when we subsequently try to access the locale associated with the user.

This PR causes email jobs to be retried on two types of error: `NoMethodError` (which is the error produced when calling an invalid method on `nil`), and `ActionView::Template::Error` (which is the error that gets wrapped around the `NoMethodError` when it occurs within a template).

It doesn't discard after three tries, unlike the retry for `ActiveRecord::RecordNotFound`, because we don't want to ignore errors that might be a sign of something wrong with the code. The hope is that the error won't recur, because the second time around, both the `User` object and the associated `Preference` object should be unavailable (triggering an `ActiveRecord::RecordNotFound` error, which will get cleaned up).

### DIFF
--- a/spec/mailers/user_mailer_spec.rb
+++ b/spec/mailers/user_mailer_spec.rb
@@ -1,13 +1,33 @@
 require "spec_helper"
 
 describe UserMailer do
+  shared_examples "it retries and fails on" do |error|
+    it "retries 3 times and ultimately fails with a #{error}" do
+      assert_performed_jobs 3, only: ApplicationMailerJob do
+        expect { subject.deliver_later }.to raise_exception(error)
+      end
+    end
+  end
+
   describe "creatorship_request" do
-    subject(:email) { UserMailer.creatorship_request(work_creatorship.id, author.id).deliver }
+    subject(:email) { UserMailer.creatorship_request(work_creatorship.id, author.id) }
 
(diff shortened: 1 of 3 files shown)


spec/mailers/user_mailer_spec.rb
```diff
@@ -1,14 +1,34 @@
 require "spec_helper"
 
 describe UserMailer do
+  shared_examples "it retries and fails on" do |error|
```
redsummernight, line 4: Can you move this to mailer_shared_examples.rb?

redsummernight (approved):

PR #4183, opened 2022-03-18
AO3-5819 Automatically label pull requests.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5819

## Purpose

This PR sets up the automatic pull request labeler action. It will run whenever a pull request is submitted, updated, or reopened, and it will automatically add ~~and/or remove~~ labels according to which paths are being modified.

~~It will also remove the two "Reviewed" labels and add back the "Awaiting Review" label any time the pull request is modified, which isn't exactly the desired behavior (it'd be better to add "Coder Has Actioned Review" instead of "Awaiting Review"), but I don't think it's possible to configure it to only add "Coder Has Actioned Review" if it previously had one of the "Reviewed" labels and "Awaiting Review" otherwise.~~

~~I picked this approach because it makes the set-up simpler, and I think re-adding "Awaiting Review" when a PR Is modified makes sense, but I can also see the argument for leaving all "Reviewed" => "Awaiting Review" changes manual. If you'd prefer, I can modify the labeler so that it only runs when the PR is first opened. Or I can try running two labeler actions, one to add "Awaiting Review" (which only runs when the PR is first opened), and one to add/remove the rest of the labels (which runs every time the PR is updated).~~

The labels are changed in three ways:
1. When a pull request is submitted, updated, or reopened, the labels "Gem Updates," "Has Migrations," "Scope: i18n Only," and "Scope: Tests Only" are added according to which files the pull request modifies. *They will not be removed automatically.* (I wanted to avoid situations where someone manually adds a "Scope: Tests Only" label, and has to keep re-adding it every time the pull request is updated.) 
2. When a pull request is opened, the "Awaiting Review" label is added automatically.
3. When a pull request with the tag "Reviewed: Action Needed" is updated or reopened, the "Reviewed: Action Needed" label is removed and the "Coder Has Actioned Review" label is added.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,26 @@
+# Label to be added when the PR is updated:
+"Awaiting Review":
+  - "**/*"
+
+# Labels to be removed when the PR is updated:
+"Reviewed: Ready to Merge": []
+"Reviewed: Action Needed": []
+
+# Labels based on which files are modified:
+"Gem Updates":
+  - "Gemfile"
+  - "Gemfile.lock"
+
+"Has Migrations":
+  - "db/migrate/**/*"
+
+"Has Production Config Changes":
+  - "config/config.yml"
+
+"Scope: i18n Only":
+  - all:
+    - "config/locales/**/*"
+
+"Scope: Tests Only":
+  - all:
+    - "+(factories|features|spec|test)/**/*"
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -14,13 +14,10 @@
 "Has Migrations":
   - "db/migrate/**/*"
 
-"Has Production Config Changes":
-  - "config/config.yml"
-
 "Scope: i18n Only":
   - all:
     - "config/locales/**/*"
 
 "Scope: Tests Only":
   - all:
-    - "+(factories|features|spec|test)/**/*"
+    - "+(factories|features|spec|test|.github)/**/*"
--- a/.github/labeler/awaiting.yml
+++ b/.github/labeler/awaiting.yml
@@ -1,0 +1,2 @@
+"Awaiting Review":
+  - "**/*"
--- a/.github/labeler/contents.yml
+++ b/.github/labeler/contents.yml
@@ -1,12 +1,3 @@
-# Label to be added when the PR is updated:
-"Awaiting Review":
-  - "**/*"
-
-# Labels to be removed when the PR is updated:
-"Reviewed: Ready to Merge": []
-"Reviewed: Action Needed": []
-
-# Labels based on which files are modified:
 "Gem Updates":
   - "Gemfile"
   - "Gemfile.lock"
--- a/.github/labeler/review.yml
+++ b/.github/labeler/review.yml
@@ -1,0 +1,6 @@
+# Label to be added when the PR is updated:
+"Coder Has Actioned Review":
+  - "**/*"
+
+# Label to be removed when the PR is updated:
+"Reviewed: Action Needed": []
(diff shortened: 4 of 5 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,26 @@
...
+
+# Labels based on which files are modified:
+"Gem Updates":
+  - "Gemfile"
+  - "Gemfile.lock"
+
+"Has Migrations":
+  - "db/migrate/**/*"
+
+"Has Production Config Changes":
```
redsummernight, line 17: I think you can leave this out. We only need this label if we expect the added values to be overridden in local.yml on staging/production (e.g. Zoho configs), in order words follow-up actions for our sysadmins. Most of the time we don't.

.github/labeler.yml
```diff
@@ -0,0 +1,26 @@
+# Label to be added when the PR is updated:
+"Awaiting Review":
```
redsummernight, line 2: > I picked this approach because it makes the set-up simpler, and I think re-adding "Awaiting Review" when a PR Is modified makes sense, but I can also see the argument for leaving all "Reviewed" => "Awaiting Review" changes manual. If you'd prefer, I can modify the labeler so that it only runs when the PR is first opened. Or I can try running two labeler actions, one to add "Awaiting Review" (which only runs when the PR is first opened), and one to add/remove the rest of the labels (which runs every time the PR is updated).

I do want to keep the distinction between "Coder Has Actioned Review" (someone's already reviewed this, should be prioritized, some level of commitment from the past reviewer) vs "Awaiting Review" (no one touched it yet). So if it's possible to run 2 actions, we can have one for sticking "Awaiting Review" on PRs when first opened, and the other for switching "Reviewed" => "Coder Has Actioned Review".

I'd wait for @sarken to take a look, though.

(If it's too annoying though, it's fine if we leave all the review status labels alone.)

.github/labeler.yml
```diff
@@ -0,0 +1,26 @@
...
+"Has Production Config Changes":
+  - "config/config.yml"
+
+"Scope: i18n Only":
+  - all:
+    - "config/locales/**/*"
+
+"Scope: Tests Only":
+  - all:
+    - "+(factories|features|spec|test)/**/*"
```
redsummernight, line 26: Can you also add `.github` here? No worries about the CI files under config/ and script/ though.

.github/labeler.yml
tickinginstant: Splitting it into "Awaiting Review" when it's first opened and "Coder Has Actioned Review" when it's updated means that if someone updates the PR immediately after submitting it (say, to make changes suggested by Hound, but before a reviewer has looked at it), it'll be marked as "Coder Has Actioned Review," which doesn't sound like what you want.

It sounds like I should just have it add "Awaiting Review" at the start and otherwise ignore review labels.

tickinginstant: It turns out that I can make it so that it only adds "Coder Has Actioned Review" if it's been previously marked as "Reviewed: Action Needed." Alternatively, there might be a way to make it add "Coder Has Actioned Review" if someone has used Github's built-in system to request a review, although I'm having trouble testing that because I can't figure out how to request a review from myself. Do either of those sound like they might be appealing options?

redsummernight: We don't really use GitHub's built-in review requests consistently, I'd say the first one.

You can request reviews from me if you need to test things out.

tickinginstant: Okay, I split the label action into three steps, and I updated the text of the PR to explain the intended behavior in detail.

(I didn't bother playing around with review requests because it sounds like that's not a feature you're interested in.)

redsummernight (approved):
Thanks!

PR #4185, opened 2022-03-19
AO3-5968 Make the chapter factory create posted chapters by default.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5968

## Purpose

Modifies the default chapter factory to make chapters posted by default, and adds a `:draft` trait for chapters.


redsummernight (approved):
Thank you!

PR #4187, opened 2022-03-22
AO3-6316 Use separate path for staging's pseud icons, add task to fix invalid pseuds

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6316

## Purpose

* Puts the pseud icons for staging under a different path so we can change those when testing without affecting production
* Add a task to fix invalid pseud icons, icon alt text, and icon comment text

## Testing Instructions

Refer to Jira 

## References

https://github.com/otwcode/otw-scripts/pull/18

## Credit

Sarken, she/her

### DIFF
--- a/app/models/pseud.rb
+++ b/app/models/pseud.rb
@@ -8,9 +8,9 @@ class Pseud < ApplicationRecord
     path: if Rails.env.production?
             ":attachment/:id/:style.:extension"
           elsif Rails.env.staging?
-            "staging/:attachment/:id/:style.:extension"
+            ":rails_env/:attachment/:id/:style.:extension"
           else
-            ":rails_root/public:url"
+            ":rails_root/public/system/:rails_env/:class/:attachment/:id_partition/:style/:filename"
           end,
     storage: %w(staging production).include?(Rails.env) ? :s3 : :filesystem,
     s3_protocol: "https",
--- a/app/models/pseud.rb
+++ b/app/models/pseud.rb
@@ -5,7 +5,13 @@ class Pseud < ApplicationRecord
 
   has_attached_file :icon,
     styles: { standard: "100x100>" },
-    path: %w(staging production).include?(Rails.env) ? ":attachment/:id/:style.:extension" : ":rails_root/public:url",
+    path: if Rails.env.production?
+            ":attachment/:id/:style.:extension"
+          elsif Rails.env.staging?
+            "staging/:attachment/:id/:style.:extension"
+          else
+            ":rails_root/public:url"
+          end,
     storage: %w(staging production).include?(Rails.env) ? :s3 : :filesystem,
     s3_protocol: "https",
     s3_credentials: "#{Rails.root}/config/s3.yml",
--- a/lib/tasks/after_tasks.rake
+++ b/lib/tasks/after_tasks.rake
@@ -874,8 +874,8 @@ namespace :After do
     puts("Added default rating to works: #{updated_works}") && STDOUT.flush
   end
 
-  desc "Delete invalid icon data from pseuds"
-  task(delete_invalid_pseud_icon_data: :environment) do
+  desc "Fix pseuds with invalid icon data"
+  task(fix_invalid_pseud_icon_data: :environment) do
     # From validates_attachment_content_type in pseuds model.
     valid_types = %w[image/gif image/jpeg image/png]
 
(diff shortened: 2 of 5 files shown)


lib/tasks/after_tasks.rake
```diff
@@ -874,6 +874,42 @@ namespace :After do
...
+        pseud.icon = nil
+      end
+
+      # Delete the icon alt text if it's too long.
+      if pseud.icon_alt_text.length > ArchiveConfig.ICON_ALT_MAX
+        pseud.icon_alt_text = ""
+      end
+
+      # Delete the icon comment if it's too long.
+      if pseud.icon_comment_text.length > ArchiveConfig.ICON_COMMENT_MAX
```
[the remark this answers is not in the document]
sarken: I did it for the line length, dog.

app/models/pseud.rb
```diff
@@ -5,7 +5,13 @@ class Pseud < ApplicationRecord
 
   has_attached_file :icon,
     styles: { standard: "100x100>" },
-    path: %w(staging production).include?(Rails.env) ? ":attachment/:id/:style.:extension" : ":rails_root/public:url",
+    path: if Rails.env.production?
+            ":attachment/:id/:style.:extension"
+          elsif Rails.env.staging?
+            "staging/:attachment/:id/:style.:extension"
+          else
+            ":rails_root/public:url"
```
redsummernight, line 13: All the uploaded files end up in the public folder. So I think we also need to:

- Change the path to include the environment, so we can clean up test data without touching development data. `url` is `/system/:class/:attachment/:id_partition/:style/:filename`, so we can use `:rails_root/public/system/:rails_env/:class/:attachment/:id_partition/:style/:filename`.
- Add `public/system/test` and `public/system/development` to .gitignore.
- Add [an after suite hook to `FileUtils.rm_rf` the test folder](https://github.com/thoughtbot/paperclip/#testing).

lib/tasks/after_tasks.rake
```diff
@@ -874,6 +874,34 @@ namespace :After do
     puts("Added default rating to works: #{updated_works}") && STDOUT.flush
   end
 
+  desc "Delete invalid icon data from pseuds"
+  task(delete_invalid_pseud_icon_data: :environment) do
+    # From validates_attachment_content_type in pseuds model.
+    valid_types = %w[image/gif image/jpeg image/png]
```
redsummernight, line 880: We have a tiny few entries with "image/jpg", so we should edit them to "image/jpeg".

lib/tasks/after_tasks.rake
```diff
@@ -874,6 +874,34 @@ namespace :After do
...
+    puts("Updating #{invalid_pseuds_count} pseuds") && STDOUT.flush
+
+    invalid_pseuds.each do |pseud|
+      # Delete the icon if it's not a valid type.
+      pseud.icon = nil unless valid_types.include?(pseud.icon_content_type)
+      # Delete the icon alt text if it's too long.
+      pseud.icon_alt_text = "" if pseud.icon_alt_text.length > ArchiveConfig.ICON_ALT_MAX
+      # Delete the icon comment if it's too long.
+      pseud.icon_comment_text = "" if pseud.icon_comment_text.length > ArchiveConfig.ICON_COMMENT_MAX
+      pseud.save
```
redsummernight, line 900: We should keep a list of pseud IDs that failed to save and print that out at the end.

spec/lib/tasks/after_tasks.rake_spec.rb
```diff
@@ -301,3 +301,66 @@
...
+  end
+  let(:invalid_pseud) do
+    create(
+      :pseud,
+      icon: File.new(Rails.root.join("public/images/feed-icon-14x14.png"))
+    )
+  end
+
+  before do
+    ArchiveConfig.ICON_ALT_MAX = 5
```
redsummernight, line 322: Can you stub `ArchiveConfig` instead of editing it? These changes will last after this spec.

lib/tasks/after_tasks.rake
tickinginstant: Hound ignores line length because we told it to: https://github.com/otwcode/otwarchive/blob/678bf364c791a503d4c9647cf9bed67daace3bea/.rubocop.yml#L24-L25

app/models/pseud.rb
```diff
@@ -5,7 +5,13 @@ class Pseud < ApplicationRecord
...
+          end,
```
tickinginstant, line 14: Does this need to be done for collections/skins, too? Or is that not in scope?

spec/lib/tasks/after_tasks.rake_spec.rb
```diff
@@ -301,3 +301,66 @@
...
+      :pseud,
+      icon_alt_text: "hi",
+      icon_comment_text: "okay",
+      icon: File.new(Rails.root.join("public/images/feed-icon-14x14.png"))
+    )
+  end
+  let(:invalid_pseud) do
+    create(
+      :pseud,
+      icon: File.new(Rails.root.join("public/images/feed-icon-14x14.png"))
```
redsummernight, line 317: Let's use one of the test files instead, e.g. features/fixtures/icon.bmp.

app/models/pseud.rb
redsummernight: Yeah, we need to fix this for collections and skins too, but I think in a separate issue.

redsummernight (approved):

app/models/pseud.rb
redsummernight: I made [AO3-6324](https://otwarchive.atlassian.net/browse/AO3-6324) for that.

PR #4195, opened 2022-04-11
AO3-3293 Touch comments when pseud is changed

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-3293

## Purpose

When we modify a pseud, touch any comments that belong to it to ensure the name and icon are updated.

## Testing Instructions

Refer to Jira.

### DIFF
--- a/spec/models/pseud_spec.rb
+++ b/spec/models/pseud_spec.rb
@@ -10,7 +10,7 @@ describe Pseud do
   end
 
   it "is invalid if there are special characters" do
-    expect(build(:pseud, name: '*pseud*')).to be_invalid
+    expect(build(:pseud, name: "*pseud*")).to be_invalid
   end
 
   describe "save" do
--- a/spec/models/pseud_spec.rb
+++ b/spec/models/pseud_spec.rb
@@ -10,7 +10,7 @@ describe Pseud do
   end
 
   it "is invalid if there are special characters" do
-      expect(build(:pseud, name: '*pseud*')).to be_invalid
+    expect(build(:pseud, name: '*pseud*')).to be_invalid
   end
 
   describe "save" do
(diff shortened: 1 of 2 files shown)


spec/models/pseud_spec.rb
```diff
@@ -48,4 +48,19 @@
...
+    let(:pseud) { create(:pseud) }
+    let!(:comment) { create(:comment, pseud_id: pseud.id) }
+
+    it "modifies the updated_at of associated comments" do
+      # Without this, the in-memory pseud has 0 comments and the test fails.
+      pseud.reload
+      original_comment_updated_at = comment.updated_at
+      travel(1.day)
+      pseud.name = "New Name"
+      pseud.save
```
redsummernight, line 62: You can use `.update(name: "New Name")` here.

spec/models/pseud_spec.rb
```diff
@@ -48,4 +48,19 @@
...
+      expect(comment.reload.updated_at).not_to eq(original_comment_updated_at)
```
redsummernight, line 63: You can use
```ruby
expect do
end.to change { comment.reload.updated_at }
```
and skip saving the old timestamp.

spec/models/pseud_spec.rb
```diff
@@ -48,4 +48,19 @@
       @pseud.errors[:icon_comment_text].should_not be_empty
     end
   end
+
+  describe "touch_comments" do
+    let(:pseud) { create(:pseud) }
+    let!(:comment) { create(:comment, pseud_id: pseud.id) }
```
redsummernight, line 54: `pseud: pseud` also works and is more concise here.

spec/models/pseud_spec.rb
```diff
@@ -48,4 +48,19 @@
...
+
+    it "modifies the updated_at of associated comments" do
+      # Without this, the in-memory pseud has 0 comments and the test fails.
```
redsummernight, line 57: Not the first time we encounter this problem (#3239).

redsummernight (approved):

PR #4200, opened 2022-04-15
AO3-3357 Add a button to clear filters

weeklies:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-3357
## Purpose

Add a button to clear applied filters for works, bookmarks and collections. 

## Testing Instructions
See Jira issue.

## Credit
weeklies (she/her)

### DIFF
--- a/app/helpers/search_helper.rb
+++ b/app/helpers/search_helper.rb
@@ -42,22 +42,19 @@ module SearchHelper
     ArchiveConfig.SEARCH_TIPS[rand(ArchiveConfig.SEARCH_TIPS.size)]
   end
 
-  def original_path(var)
-    case var
-    when Collection
-      return collection_works_path(var) if params[:work_search].present?
-      return collection_bookmarks_path(var) if params[:bookmark_search].present?
-    when Tag, Fandom
-      return tag_works_path(var) if params[:work_search].present?
-      return tag_bookmarks_path(var) if params[:bookmark_search].present?
-    when Pseud
-      return user_pseud_works_path(var.user, var) if params[:work_search].present?
-      return user_pseud_bookmarks_path(var.user, var) if params[:bookmark_search].present?
-    when User
-      return user_works_path(var) if params[:work_search].present?
-      return user_bookmarks_path(var) if params[:bookmark_search].present?
-    else
-      "/"
-    end
+  def works_original_path
+    url_for(
+      controller: :works,
+      action: :index,
+      **params.slice(:tag_id, :fandom_id, :collection_id, :pseud_id, :user_id).permit!
+    )
+  end
+
+  def bookmarks_original_path
+    url_for(
+      controller: :bookmarks,
+      action: :index,
+      **params.slice(:tag_id, :collection_id, :pseud_id, :user_id).permit!
+    )
   end
 end
--- a/app/helpers/search_helper.rb
+++ b/app/helpers/search_helper.rb
@@ -42,20 +42,19 @@ module SearchHelper
     ArchiveConfig.SEARCH_TIPS[rand(ArchiveConfig.SEARCH_TIPS.size)]
   end
 
-  def original_path
-    if @collection
-      collections_path
-    elsif @tag
-      return tag_works_path(@tag) if params[:work_search].present?
-      return tag_bookmarks_path(@tag) if params[:bookmark_search].present?
-    elsif @pseud
-      return user_pseud_works_path(@user, @pseud) if params[:work_search].present?
-      return user_pseud_bookmarks_path(@user, @pseud) if params[:bookmark_search].present?
-    elsif @user
-      return user_works_path(@user) if params[:work_search].present?
-      return user_bookmarks_path(@user) if params[:bookmark_search].present?
-
-      collected_user_works_path(@user)
+  def original_path(var)
+    case var
+    when Tag, Fandom
+      return tag_works_path(var) if params[:work_search].present?
+      return tag_bookmarks_path(var) if params[:bookmark_search].present?
+    when Pseud
+      return user_pseud_works_path(var.user, var) if params[:work_search].present?
+      return user_pseud_bookmarks_path(var.user, var) if params[:bookmark_search].present?
+    when User
+      return user_works_path(var) if params[:work_search].present?
+      return user_bookmarks_path(var) if params[:bookmark_search].present?
+    else
+      "/"
     end
   end
 end
--- a/app/helpers/search_helper.rb
+++ b/app/helpers/search_helper.rb
@@ -42,4 +42,16 @@ module SearchHelper
     ArchiveConfig.SEARCH_TIPS[rand(ArchiveConfig.SEARCH_TIPS.size)]
   end
 
+  def original_path
+    if @tag
+      return tag_works_path(@tag) if params[:work_search].present?
+      return tag_bookmarks_path(@tag) if params[:bookmark_search].present?
+    elsif @pseud
+      return user_pseud_works_path(@user, @pseud) if params[:work_search].present?
+      return user_pseud_bookmarks_path(@user, @pseud) if params[:bookmark_search].present?
+    elsif @user
+      return user_works_path(@user) if params[:work_search].present?
+      return user_bookmarks_path(@user) if params[:bookmark_search].present?
+    end
+  end
 end
--- a/app/helpers/search_helper.rb
+++ b/app/helpers/search_helper.rb
@@ -44,6 +44,9 @@ module SearchHelper
 
   def original_path(var)
     case var
+    when Collection
+      return collection_works_path(var) if params[:work_search].present?
+      return collection_bookmarks_path(var) if params[:bookmark_search].present?
     when Tag, Fandom
       return tag_works_path(var) if params[:work_search].present?
       return tag_bookmarks_path(var) if params[:bookmark_search].present?
--- a/app/helpers/search_helper.rb
+++ b/app/helpers/search_helper.rb
@@ -46,6 +46,7 @@ module SearchHelper
     url_for(
       controller: :works,
       action: :index,
+      only_path: true,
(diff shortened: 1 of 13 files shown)


app/views/bookmarks/_filters.html.erb
```diff
@@ -128,6 +128,7 @@
         </dl>
       </dd>
       <dt class="landmark"><%= ts("Submit") %></dt>
+      <dd class="submit actions"><%= f.submit ts("Clear Filters") %></dd>
```
redsummernight, line 131: Instead of a callback that redirects, I think this button can be a link to whatever the original form / starting URL was.

For the original form of the works and bookmarks filters, you'll have to handle 5 different kinds of owning objects:

https://github.com/otwcode/otwarchive/blob/808b34b1d21be65c26fc9ae4a20621c40955a37d/app/views/bookmarks/_filters.html.erb#L135-L139

You can add a helper to determine the original URL in app/helpers/search_helper.rb.

----

The original form of the collections filters can just be `collections_path`. On that note, can you also change this part to `collections_path`?

https://github.com/otwcode/otwarchive/blob/808b34b1d21be65c26fc9ae4a20621c40955a37d/app/views/collections/_filters.html.erb#L1

----

There's one more filters sidebar at app/views/works/_collection_filters.html.erb. The reset path should just be `collected_user_works_path(@user)`.

weeklies: Thank you for explaining!

app/views/bookmarks/_filters.html.erb
```diff
@@ -128,6 +128,11 @@
         </dl>
       </dd>
       <dt class="landmark"><%= ts("Submit") %></dt>
+      <dd class="submit actions"><%= link_to ts("Clear Filters"), original_path(@collection) if @collection %></dd>
+      <dd class="submit actions"><%= link_to ts("Clear Filters"), original_path(@tag) if @tag %></dd>
+      <dd class="submit actions"><%= link_to ts("Clear Filters"), original_path(@fandom) if @fandom %></dd>
+      <dd class="submit actions"><%= link_to ts("Clear Filters"), original_path(@pseud) if @pseud %></dd>
+      <dd class="submit actions"><%= link_to ts("Clear Filters"), original_path(@user) if @user and not @pseud %></dd>
```
sarken, line 135: Let's deemphasize the "Clear Filters" links a bit by moving it out of the `<dl>` and making it into a small plain text link. Putting `<p class="footnote"><a href="URL">Clear Filters</a></p>` under the `</dl>` should work. 

You'll probably also need to reposition the logic regarding which link to display -- I noticed the current approach was leaving a lot of empty `<dd>` tags in the code, and we'd rather not have empty elements lying around.

sarken, line 135: Let's deemphasize the "Clear Filters" links a bit by moving it out of the `<dl>` and making it into a small plain text link. Putting `<p class="footnote"><a href="URL">Clear Filters</a></p>` under the `</dl>` should work. 

You'll probably also need to reposition the logic regarding which link to display -- I noticed the current approach was leaving a lot of empty `<dd>` tags in the code, and we'd rather not have empty elements lying around.

app/views/works/_filters.html.erb
```diff
@@ -162,6 +162,28 @@
...
+      <%= link_to ts("Clear Filters"), original_path(@collection) %>
+    </p>
+    <% elsif @tag %>
+    <p class="footnote">
+      <%= link_to ts("Clear Filters"), original_path(@tag) %>
+    </p>
+    <% elsif @fandom %>
+    <p class="footnote">
+      <%= link_to ts("Clear Filters"), original_path(@fandom) %>
+    </p>
```
tickinginstant, line 176: In the `WorksController`, the `@fandom` variable is only used in conjunction with another owner. In theory it could be used with any of the other four owners (`@tag`, `@collection`, `@pseud`, `@user`), but in practice there's only one link with `fandom_id` set:
https://github.com/otwcode/otwarchive/blob/59bb5ebc3ca4fa86056b410d922152905b8cc179/app/views/users/_contents.html.erb#L13-L17
So `fandom_id` needs to be passed as a param for at least the `@user` and `@pseud` "Clear Filters" links, and you should probably avoid having a separate branch for `@fandom`.

(You might also want to think about using the `@owner` variable to simplify your code: https://github.com/otwcode/otwarchive/blob/59bb5ebc3ca4fa86056b410d922152905b8cc179/app/controllers/works_controller.rb#L770
Both here and for the bookmark filters.)

app/helpers/search_helper.rb
```diff
@@ -42,4 +42,22 @@ def random_search_tip
     ArchiveConfig.SEARCH_TIPS[rand(ArchiveConfig.SEARCH_TIPS.size)]
   end
 
+  def original_path(var)
+    case var
+    when Collection
+      return collection_works_path(var) if params[:work_search].present?
```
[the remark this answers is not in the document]
weeklies: I think I'll go with `url_for`, thanks for the info.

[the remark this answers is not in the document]
sarken: Does `url_for` default to `only_path` in helpers? iirc, it includes the host in views and controllers, but I am not sure what the story is with helpers, and my Google skills are failing me.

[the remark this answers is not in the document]
weeklies: When I was trying it out, `url_for` gave the host name as well.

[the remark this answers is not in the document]
tickinginstant: You can pass `only_path: true` as an argument to `url_for` to get it to skip the host. (And probably should -- there are [some](https://otwarchive.atlassian.net/browse/AO3-5908) [issues](https://otwarchive.atlassian.net/browse/AO3-5945) with having full URLs as links in production. I didn't think about it when suggesting `url_for`, sorry.)

app/views/bookmarks/_filters.html.erb
```diff
@@ -131,6 +131,12 @@
       <dd class="submit actions"><%= f.submit ts("Sort and Filter") %></dd>
     </dl>
 
+    <% if @pseud || @user || @collection || @tag %>
+    <p class="footnote">
```
redsummernight, line 135: In most places in our views we indent 2 spaces inside a conditional, let's also do that here and in works/_filters.

app/views/bookmarks/_filters.html.erb
```diff
@@ -131,6 +131,12 @@
       <dd class="submit actions"><%= f.submit ts("Sort and Filter") %></dd>
     </dl>
 
+    <% if @pseud || @user || @collection || @tag %>
```
redsummernight, line 134: You can use `@owner` here.

https://github.com/otwcode/otwarchive/blob/6551990716b90f441ce9cbbb476a42a6d4142185/app/controllers/bookmarks_controller.rb#L332

It includes `@bookmarkable` as well, but if there's a `@bookmarkable` then the filters won't be loaded anyway, so we're still good.

app/views/works/_filters.html.erb
```diff
@@ -162,6 +162,12 @@
       <dd class="submit actions"><%= f.submit ts("Sort and Filter") %></dd>
     </dl>
 
+    <% if @pseud || @user || @collection || @tag %>
```
redsummernight, line 165: You can use `@owner` here:

https://github.com/otwcode/otwarchive/blob/6551990716b90f441ce9cbbb476a42a6d4142185/app/controllers/works_controller.rb#L773

redsummernight (approved):
Approach is good to me!

app/helpers/search_helper.rb
```diff
@@ -42,4 +42,22 @@ def random_search_tip
...
+      return tag_works_path(var) if params[:work_search].present?
+      return tag_bookmarks_path(var) if params[:bookmark_search].present?
+    when Pseud
+      return user_pseud_works_path(var.user, var) if params[:work_search].present?
+      return user_pseud_bookmarks_path(var.user, var) if params[:bookmark_search].present?
+    when User
+      return user_works_path(var) if params[:work_search].present?
+      return user_bookmarks_path(var) if params[:bookmark_search].present?
+    else
+      "/"
```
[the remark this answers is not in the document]
redsummernight: Never mind this old review comment.

features/search/filters.feature
```diff
@@ -40,13 +40,11 @@ Feature: Filters
     Then I should see "Bilbo Does the Thing"
       And I should not see "A Hobbit's Meandering"
       And I should not see "Roonal Woozlib and the Ferrets of Nimh"
+    When I follow "Clear Filters"
+    Then I should see "3 Works by meatloaf"
     When I press "Fandoms" within "dd.include"
-      And I uncheck "The Hobbit (1)" within "#include_fandom_tags"
-      And I uncheck "Legend of Korra (1)" within "#include_fandom_tags"
-      And I press "Sort and Filter"
-    Then I should see "Roonal Woozlib and the Ferrets of Nimh"
```
redsummernight, line 47: Let's keep these assertions at the end so we know that "Clear Filters" takes effect:
```gherkin
    Then I should see "Roonal Woozlib and the Ferrets of Nimh"
      And I should see "A Hobbit's Meandering"
      And I should see "Bilbo Does the Thing"
```
The same goes for the exclusion filters scenario as well.

app/views/bookmarks/_filters.html.erb
```diff
@@ -131,6 +131,12 @@
       <dd class="submit actions"><%= f.submit ts("Sort and Filter") %></dd>
     </dl>
 
+    <% if @owner = @bookmarkable || @pseud || @user || @collection || @tag %>
```
redsummernight, line 134: You should be able to use `<% if @owner %>` here, since the variable is set in the controller. Same goes for the bookmark filters.

redsummernight (commented):
Final comments!

redsummernight (approved):
Thank you!

PR #4205, opened 2022-04-21
AO3-6323 Add user ID to comments

EchoEkhi:
## To do

- [x] update view erb files
- [x] write spec

## Issue

https://otwarchive.atlassian.net/browse/AO3-6323

## Purpose

Add the user ID of a comment’s creator to the comment’s classes, in the format user-123, allowing users to highlight or hide comments with skins.

## Testing Instructions

See JIRA ticket.

## Credit

EchoEkhi (he/him)


EchoEkhi:
It works on my machine ¯\_(ツ)_/¯

app/helpers/comments_helper.rb
```diff
@@ -286,6 +286,17 @@ def tag_comment_as_spam_link(comment)
...
+  def get_comment_user_css_class(comment)
+    if comment.pseud.nil?
+      "visitor"
+    elsif comment.pseud.user.is_author_of?(comment.ultimate_parent) && comment.ultimate_parent.anonymous?
+      "anonymous-creator"
+    else
+      comment.pseud.user.id
+    end
+  end
+
```
sarken, line 299: You can skip the special cases for anonymous and guests. We just want to add the user ID like we did with work blurbs: 

https://github.com/otwcode/otwarchive/blob/808b34b1d21be65c26fc9ae4a20621c40955a37d/app/helpers/application_helper.rb#L588-L599

app/helpers/comments_helper.rb
```diff
@@ -286,6 +286,14 @@ def tag_comment_as_spam_link(comment)
     end
   end
 
+  # gets the css user-<id> class name for the comment
+  def comment_user_css_class(comment)
+    return "" if comment.pseud.nil?
+    return "" if comment.pseud.user.is_author_of?(comment.ultimate_parent) && comment.ultimate_parent.anonymous?
```
sarken, line 292: I think you can skip the empty string and just say `return if [condition]`

app/helpers/comments_helper.rb
```diff
@@ -286,6 +286,14 @@ def tag_comment_as_spam_link(comment)
...
+
+    " user-#{comment.pseud.user.id}"
```
sarken, line 294: Let's leave out the leading space here for consistency with [css_classes_for_creation_blurb](https://github.com/otwcode/otwarchive/blob/84f2f1eed7d5e254411116417c1500b4a31ae7e7/app/helpers/application_helper.rb#L601-L609). We can even stick all the reused classes in here, too, to further reduce redundancy.

So this could be 

```ruby
def css_classes_for_comment(comment)
  return if comment.pseud.nil?
  return if comment.pseud.user.is_author_of?(comment.ultimate_parent) && comment.ultimate_parent.anonymous?
  
  "comment group user-#{comment.pseud.user.id}"
end
```

I know right now the single comment view does `comment group odd/even user-000`, but it's okay if it gets switched to `odd/even comment group user-000` -- the odd/even classes really should've been out front to begin with.

sarken (changes requested):
The helper should also have tests. You can check the [application helper spec](https://github.com/otwcode/otwarchive/blob/84f2f1eed7d5e254411116417c1500b4a31ae7e7/spec/helpers/application_helper_spec.rb#L181) to see how we did it for the other helpers that handle CSS classes.

app/helpers/comments_helper.rb
sarken: I just realized this will clash a little bit with #4208, but I have to go afk -- I'll get back to you about the best way to handle that as soon as I can.

sarken (commented):
I've marked my previous comments resolved, so you can ignore them. Sorry about that!

sarken (commented):
Just one more tiny tweak to the helper while you add specs and update the views.

spec/helpers/application_helper_spec.rb
```diff
@@ -364,4 +364,50 @@
...
+
+      it "returns nil" do
+        expect(helper.commenter_id_for_css_classes(comment)).to eq(nil)
+      end
+    end
+
+    context "when commenter is creater of work inside anonymous collection" do
+      let(:user) { create(:user) }
+      let(:anonymous_collection) { create(:anonymous_collection) }
+      let(:work) { create(:work, authors: [user.default_pseud]) }
```
tickinginstant, line 389: The code for work anonymity is very finicky at the moment. I'm not 100% certain, but I'm guessing that it'd help to change the definition here to:
```
      let(:work) { create(:work, authors: [user.default_pseud], collections: [anonymous_collection]) }
```
The other alternative would be calling `work.reload` after the anonymous collection is added, but before creating the comment. The issue is effectively that you've got multiple copies of the work in memory, and one of them is anonymous and the other isn't. (This doesn't happen in production because each request loads items from scratch, but tests sometimes need a `reload` call to make sure all objects have access to the same data.)

spec/helpers/application_helper_spec.rb
```diff
@@ -364,4 +364,50 @@
       end
     end
   end
+
```
sarken, line 367: The methods we're testing are in the comments helper, so let's make spec/helpers/comment_helper_spec.rb and put the tests there instead.

spec/helpers/application_helper_spec.rb
```diff
@@ -364,4 +364,50 @@
...
+    
+    context "when commenter is a visitor" do
+      let(:comment) { create(:comment, :by_guest) }
+
+      it "returns nil" do
+        expect(helper.commenter_id_for_css_classes(comment)).to eq(nil)
+      end
+    end
+
+    context "when commenter is creater of work inside anonymous collection" do
```
sarken, line 386: Should be "creator." This could even go under `context "when commenter is a user"`, since the commenter here is indeed a user.

spec/helpers/application_helper_spec.rb
```diff
@@ -364,4 +364,50 @@
...
+  describe "#css_classes_for_comment" do
+    context "when comment exists" do
+      let(:user) { create(:user) }
+      let(:comment) { create(:comment, pseud: user.default_pseud) }
+      
+      it "has classes" do
+        expect(helper.css_classes_for_comment(comment)).to eq("comment group user-#{user.id}")
+      end
+
+      it "is unreviewed" do
```
sarken, line 407: For clarity, perhaps `it "includes unreviewed class when comment is unreviewed" do`.

app/models/comment.rb
```diff
@@ -36,6 +36,8 @@ def check_for_spam
   scope :ordered_by_date, -> { order('created_at DESC') }
   scope :top_level,       -> { where.not(commentable_type: "Comment") }
   scope :include_pseud,   -> { includes(:pseud) }
+  scope :include_user,    -> { includes(pseud: [:user]) }
+  scope :include_parent,  -> { includes(:parent) }
```
sarken, line 40: Since the changes to the comments controller were removed, these are now unused and should be removed as well.

sarken (approved):
Looks good -- thank you!

PR #4207, opened 2022-04-21
AO3-6319 Allow hsla colors in skins

EchoEkhi:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6319

## Purpose

It allows hsl and hsla colors in skins.

## Testing Instructions

See JIRA ticket.

## Credit

EchoEkhi (he/him)

### DIFF
--- a/spec/models/skin_spec.rb
+++ b/spec/models/skin_spec.rb
@@ -29,8 +29,8 @@ describe Skin do
         /* end comment */",
 
       "should allow hsl(a) colours" =>
-        "ol {color: hsl(100, 100, 100);}
-        li {color: hsla(100%, 100%, 100%, 1);}",
+        "ol {color: hsl(180, 100%, 50%);}
+        li {color: hsla(90, 30%, 70%, 50%);}",
       
       "should allow through border-radius (CSS3 property)" =>
         ".profile { border-radius: 5px }",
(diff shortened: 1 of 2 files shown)


redsummernight (changes requested):
If you could leave the pull request template checklist in until you've had at least a few pull requests merged, we'd appreciate it.

For this issue, you should add tests to make sure that site skins accept the new color format. You can add to this existing section:

https://github.com/otwcode/otwarchive/blob/808b34b1d21be65c26fc9ae4a20621c40955a37d/spec/models/skin_spec.rb#L57-L77

spec/models/skin_spec.rb
```diff
@@ -28,6 +28,10 @@
         dd {color: blue;}
         /* end comment */",
 
+      "should allow hsl(a) colours" =>
+        "ol {color: hsl(100, 100, 100);}
```
redsummernight, line 32: Let's use [the right syntax for `hsl`](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/hsl) here: the first value can be a unitless number, but the second and third values for saturation and lightness should be percentages. The same applies to `hsla` as well.

redsummernight (approved):
Thanks!

PR #4208, opened 2022-04-21
AO3-5986 Create an 'Official' role

EchoEkhi:
## To do (maybe)

- [ ] Add tests
- [ ] Apply changes on comment notification emails

## Issue

https://otwarchive.atlassian.net/browse/AO3-5986

## Purpose

Introduces the 'official' role, and alters the appearance of comments left by official accounts to make clear that this is official AO3 communication and to make it harder for users to imitate official accounts.

## Testing Instructions

Run `rake defaults:create_roles`
Then follow JIRA ticket.

## References

Maybe we should merge this before implementing tests and email changes, as it is blocking [AO3-6327](https://otwarchive.atlassian.net/browse/AO3-6327).

## Credit

EchoEkhi (He/him)

### DIFF
--- a/app/helpers/comments_helper.rb
+++ b/app/helpers/comments_helper.rb
@@ -49,6 +49,8 @@ module CommentsHelper
     if comment.pseud_id
       if comment.pseud.nil?
         ts("Account Deleted")
+      elsif comment.pseud.user.official
+        link_to (comment.pseud.byline + content_tag(:span, " (Official)", :class => "role")).html_safe, [comment.pseud.user, comment.pseud]
       else
         link_to comment.pseud.byline, [comment.pseud.user, comment.pseud]
       end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -388,6 +388,15 @@ class User < ApplicationRecord
     has_role?(:archivist)
   end
 
+  # Is this user an authorized official?
+  def official
+    self.is_official?
+  end
+
+  def is_official?
+    has_role?(:official)
+  end
+
   # Is this user a protected user? These are users experiencing certain types
   # of harassment. For now, this is only used to prevent harassment via repeated
   # password reset requests.
--- a/app/views/comments/_single_comment.html.erb
+++ b/app/views/comments/_single_comment.html.erb
@@ -1,5 +1,5 @@
 <!-- START of single comment -->
-<li class="<% if single_comment.unreviewed? %>unreviewed <% end %>comment group <%= cycle :odd, :even %>" id="comment_<%= single_comment.id %>" role="article">
+<li class="<% if !single_comment.pseud.nil? && single_comment.pseud.user.official %>official <% end %><% if single_comment.unreviewed? %>unreviewed <% end %>comment group <%= cycle :odd, :even %>" id="comment_<%= single_comment.id %>" role="article">
   <% if params[:edit_comment_id] && params[:edit_comment_id] == single_comment.id.to_s && can_edit_comment?(single_comment) %>
     <!-- we're editing this comment -->
     <%= render 'comments/comment_form',
--- a/public/stylesheets/site/2.0/15-group-comments.css
+++ b/public/stylesheets/site/2.0/15-group-comments.css
@@ -61,7 +61,7 @@ div.comment, li.comment {
 fieldset.comments, .comment .userstuff {
   background: transparent;
   border: none;
-    box-shadow: none;
+  box-shadow: none;
 }
 
 .comment .userstuff {
@@ -80,6 +80,15 @@ fieldset.comments, .comment .userstuff {
 
 .comment .edited {
   margin: 1.286em 0.643em 0.25em;
+}
+
+.comment span.role {
+  font-weight: bold;
+  color: #900;
+}
+
+.comment.official {
+  background-color: #FCC;
 }
 
 /* mods */
--- a/app/views/home/_inbox_module.html.erb
+++ b/app/views/home/_inbox_module.html.erb
@@ -8,7 +8,7 @@
   <p class="note"><%= ts('The latest unread items from your inbox.') %></p>
   <ul class="abbreviated comment group">
     <% inbox_comments.each do |inbox_comment| %>
-      <li class="<% if inbox_comment.feedback_comment.unreviewed? %>unreviewed <% end %>unread comment group" id="feedback_comment_<%= inbox_comment.feedback_comment_id %>">
+      <li class="<% if !inbox_comment.feedback_comment.pseud.nil? && inbox_comment.feedback_comment.pseud.user.official %>official <% end %><% if inbox_comment.feedback_comment.unreviewed? %>unreviewed <% end %>unread comment group" id="feedback_comment_<%= inbox_comment.feedback_comment_id %>">
         <%= render 'inbox/inbox_comment_contents', feedback_comment: inbox_comment.feedback_comment %>
         <ul class="actions">
           <% if inbox_comment.feedback_comment.iced? %>
--- a/app/views/inbox/show.html.erb
+++ b/app/views/inbox/show.html.erb
@@ -37,7 +37,7 @@
           <% feedback_comment = inbox_comment.feedback_comment %>
 
           <!-- START of single comment -->
-          <li class="<% if feedback_comment.unreviewed? %>unreviewed <% end %><%= inbox_comment.read? ? 'read' : 'unread' %> comment group" role="article" id="feedback_comment_<%= feedback_comment.id %>">
+          <li class="<% if !feedback_comment.pseud.nil? && feedback_comment.pseud.user.official %>official <% end %><% if feedback_comment.unreviewed? %>unreviewed <% end %><%= inbox_comment.read? ? 'read' : 'unread' %> comment group" role="article" id="feedback_comment_<%= feedback_comment.id %>">
 
             <%= render "inbox_comment_contents", feedback_comment: feedback_comment %>
 
--- a/features/fixtures/roles.yml
+++ b/features/fixtures/roles.yml
@@ -10,3 +10,6 @@ translation_admin:
 archivist:
   id: 4
   name: archivist
+official:
+  id: 5
+  name: official
--- a/lib/tasks/defaults.rake
+++ b/lib/tasks/defaults.rake
@@ -1,7 +1,7 @@
 namespace :defaults do
   desc "Create default roles by name"
   task(create_roles: :environment) do
-    %w[archivist opendoors protected_user tag_wrangler translation_admin translator].each do |role|
+    %w[archivist opendoors protected_user tag_wrangler translation_admin translator official].each do |role|
       Role.find_or_create_by(name: role)
     end
   end
--- a/spec/lib/tasks/defaults.rake_spec.rb
+++ b/spec/lib/tasks/defaults.rake_spec.rb
@@ -4,7 +4,7 @@ describe "rake defaults:create_roles" do
   it "creates roles" do
     subject.invoke
 
-    role_names = %w[archivist opendoors protected_user tag_wrangler translation_admin translator]
+    role_names = %w[archivist opendoors protected_user tag_wrangler translation_admin translator official]
     expect(Role.all.pluck(:name)).to eq(role_names)
   end
 end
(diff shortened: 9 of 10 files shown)


public/stylesheets/site/2.0/15-group-comments.css
```diff
@@ -82,6 +82,15 @@ fieldset.comments, .comment .userstuff {
...
 }
 
+.comment span.role {
+  font-weight: bold;
+  color: #900;
+}
+
+.comment.official {
+  background-color: #FCC;
+}
```
sarken, line 92: We use lowercase letters in hex color codes and try to avoid chaining classes. However, this should be removed, since it wasn't requested in the issue.

public/stylesheets/site/2.0/15-group-comments.css
```diff
@@ -82,6 +82,15 @@ fieldset.comments, .comment .userstuff {
   margin: 1.286em 0.643em 0.25em;
 }
 
+.comment span.role {
+  font-weight: bold;
+  color: #900;
```
sarken, line 87: Please remove this -- we just want the bold.

app/helpers/comments_helper.rb
```diff
@@ -49,6 +49,8 @@ def get_commenter_pseud_or_name(comment)
     if comment.pseud_id
       if comment.pseud.nil?
         ts("Account Deleted")
+      elsif comment.pseud.user.official
```
sarken, line 52: We'd like the span containing the role to be outside the link. Sorry for not being totally clear about that in the issue.

app/views/comments/_single_comment.html.erb
```diff
@@ -1,5 +1,5 @@
 <!-- START of single comment -->
-<li class="<% if single_comment.unreviewed? %>unreviewed <% end %>comment group <%= cycle :odd, :even %>" id="comment_<%= single_comment.id %>" role="article">
+<li class="<% if !single_comment.pseud.nil? && single_comment.pseud.user.official %>official <% end %><% if single_comment.unreviewed? %>unreviewed <% end %>comment group <%= cycle :odd, :even %>" id="comment_<%= single_comment.id %>" role="article">
```
sarken, line 2: It might be possible to shorten or tidy this logic a bit, e.g. `if single_comment&.pseud&.user&.official` or `if single_comment.pseud.present? && single_comment.pseud.user.official`

Even better, though, would be defining a helper method. That way, we wouldn't need to duplicate so much across the various views or edit so many places if the logic for applying the `official` class changes.

sarken (changes requested):
Thanks for working on this! We definitely want tests before we merge it, and if you're up for doing the i18n work, that would be great. However, we agree that any changes to emails are outside the initial scope. 

Even though this is blocking AO3-6327, we don't need to rush. There's other non-coding work blocking that from being deployed, too. As long as this pull request ready within the next week or two, it will be fine.

EchoEkhi:
I was just trying to write tests for the feature, but I'm seeing some strange behaviour in the test environment.

Is it intentional that admins in the test environment don't have pseuds?

sarken:
admins are different from regular users and don't have pseuds at all, only logins.

public/stylesheets/site/2.0/15-group-comments.css
```diff
@@ -61,7 +61,7 @@ div.comment, li.comment {
 fieldset.comments, .comment .userstuff {
   background: transparent;
   border: none;
-    box-shadow: none;
+  box-shadow: none;
```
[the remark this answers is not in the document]
EchoEkhi: Sorry, that was VSCode auto formatting.

public/stylesheets/site/2.0/15-group-comments.css
EchoEkhi: Sorry that was me testing the .official class, same with the colour in .role

EchoEkhi:
How does i18n work on the archive exactly? I've been trying to change the site's language to test the ts() function, but I couldn't figure out how.

sarken:
It's not currently possible to change the language of the Archive interface. 

However, we do have a locale preference for users that currently applies to emails. It's behind a [rollout feature flag](https://github.com/fetlife/rollout), so you'll need to enable that first by running `$rollout.activate_group(:set_locale_preference, :all)` from the Rails console, and then the "Your locale" menu will show up on the preferences page for all users. You may need to [log in as an admin](https://github.com/otwcode/otwarchive/wiki/Development-Data#creating-admins) and configure the available locales from the Locales page to choose which options appear in the select menu.

For automated tests, you can set `I18n.locale`: https://github.com/otwcode/otwarchive/blob/59bb5ebc3ca4fa86056b410d922152905b8cc179/spec/mailers/user_mailer_spec.rb#L669

Even with emails, though, the `ts()` function won't produce translated emails. We're switching to [Rails i18n](https://guides.rubyonrails.org/i18n.html) instead ([our i18n guidelines are on the wiki](https://github.com/otwcode/otwarchive/wiki/Internationalization-(i18n)-Standards)), and only emails that use those are actually available in other languages.

sarken (changes requested):
#4205 has been merged, tested, and declared ready for release, so I think we're safe to come back to this. No rush, though -- we're not adding anything else to this release, and we probably won't get getting started on the next one until next weekend at the earliest.

sarken (changes requested):
Hmm... I feel like something weird happened here. For example, in the comments_helper, we have `css_classes_for_comment(comment)` on line 299: 

https://github.com/otwcode/otwarchive/blob/fcb6a8e6eeec7f5001d31da07d72219d82ee5ded/app/helpers/comments_helper.rb#L299-L305

Here it's on line 47, but there's no sign of it being removed from line 299-305. (It should't be, imo; it's fine on 299, directly after the `commenter_id_for_css_classes(comment)` method.) I'm concerned we're going to end up with duplicate methods. Also, the version of `css_classes_for_comment(comment)` in this PR is missing some things.

Could you make sure you've pulled from upstream master and then merged those updates into this branch? Hopefully that'll fix things.

EchoEkhi:
Ok something bad happened...

sarken:
Try a hard reset to the last okay commit (https://github.com/otwcode/otwarchive/pull/4208/commits/edffdf3f973961eae08b3c12a67f0e04af4c4d77): 

`git reset --hard edffdf3f973961eae08b3c12a67f0e04af4c4d77`

Then you can push that (you might need to `--force`), make sure everything in the PR looks like it got reset okay, and try merging with mater again:

```
git checkout master
git pull upstream master
git push origin master
git checkout AO3-5986
git merge master
```

sarken:
I just want to say thank you for continuing to work on this! Things are a bit hectic right now due to an ongoing situation you can read about in [this news post](https://archiveofourown.org/admin_posts/22557), so my apologies for the radio silence, and the fact that code review will probably continue to be delayed.

app/helpers/comments_helper.rb
```diff
@@ -301,7 +303,9 @@ def css_classes_for_comment(comment)
 
     unreviewed = "unreviewed" if comment.unreviewed?
     commenter = commenter_id_for_css_classes(comment)
-    "#{unreviewed} comment group #{commenter}".strip
+    official = "official" if comment&.pseud&.user&.official
+
+    "#{unreviewed} comment group #{commenter} #{official}".strip
```
sarken, line 308: Could you move `#{official}` to the front?

factories/admin.rb
```diff
@@ -26,6 +26,10 @@
     factory :open_doors_admin do
       roles { ["open_doors"] }
     end
+
+    factory :official_admin do
+      roles { ["official"] }
+    end
```
sarken, line 32: Can you remove this, since we don't have an official role for admins? (Also, my apologies for hitting the single comment button rather than start a review. :facepalm:)

app/helpers/comments_helper.rb
```diff
@@ -299,9 +301,11 @@ def commenter_id_for_css_classes(comment)
   def css_classes_for_comment(comment)
     return if comment.nil?
 
+    official = "official" if comment&.pseud&.user&.official
     unreviewed = "unreviewed" if comment.unreviewed?
     commenter = commenter_id_for_css_classes(comment)
-    "#{unreviewed} comment group #{commenter}".strip
+
+    "#{official} #{unreviewed} comment group #{commenter}".strip
```
tickinginstant, line 308: Would it make sense to use `squish` here instead?

app/models/user.rb
```diff
@@ -390,6 +390,14 @@ def is_archivist?
     has_role?(:archivist)
   end
 
+  # Is this user an authorized official?
+  def official
+    roles.each do |role|
+      return true if role.name == "official"
+    end
+    false
```
tickinginstant, line 398: I think it'd be cleaner to rewrite this using [any?](https://ruby-doc.org/core-2.7.0/Enumerable.html#method-i-any-3F).

(You could have continued using `has_role?` here and instead modified the [definition](https://github.com/otwcode/otwarchive/blob/aa9204948ca8c5104ef00f965f46dd4276f2b549/config/initializers/gem-plugin_config/permit_yo_config.rb#L23-L28) of `has_role?`, possibly with a call to `if roles.loaded?` if you wanted to keep the existing behavior when the roles aren't loaded. But I don't think there are any more examples of N+1 queries involving roles in the current codebase, so I'm not sure if that kind of future-proofing is necessary.)

app/helpers/comments_helper.rb
```diff
@@ -299,9 +301,11 @@ def commenter_id_for_css_classes(comment)
   def css_classes_for_comment(comment)
     return if comment.nil?
 
+    official = "official" if comment&.pseud&.user&.official
```
tickinginstant, line 304: Does this also need to check whether the comment is an anonymous creator comment? Or is the assumption that official users won't have any anonymous works?

sarken: It doesn't hurt to cover that case, but official accounts should abso-freaking-lutely not have anonymous works.

app/helpers/comments_helper.rb
sarken: That seems like a good idea. It'll get rid of extra whitespace when a comment is official but not unreviewed.

spec/helpers/comments_helper_spec.rb
```diff
@@ -60,5 +60,24 @@
         expect(helper.css_classes_for_comment(unreviewed_comment)).to eq("unreviewed comment group user-#{user.id}")
       end
     end
+
+    context "when comment author is an official" do
```
sarken, line 64: We can say "commenter" instead of "comment author" in these descriptions for brevity and consistency with the other tests.

spec/helpers/comments_helper_spec.rb
```diff
@@ -60,5 +60,24 @@
...
+      let(:official_user) { create(:official) }
+      let(:comment) { create(:comment, pseud: official_user.default_pseud) }
+  
+      it "has official class" do
+        expect(helper.css_classes_for_comment(comment)).to match("official")
+      end
+
+      context "comment author is also creator of a work inside anonymous collection" do
+        let(:anonymous_collection) { create(:anonymous_collection) }
+        let(:work) { create(:work, authors: [official_user.default_pseud], collections: [anonymous_collection]) }
```
sarken, line 74: Since we're not referencing the anon collection elsewhere, we could just do

```
let(:work) { create(:work, authors: [official_user.default_pseud], collections: [create(:anonymous_collection)]) }
```

factories/users.rb
```diff
@@ -46,5 +46,9 @@
     factory :translation_admin do
       roles { [Role.find_or_create_by(name: "translation_admin")] }
     end
+
+    factory :official do
```
sarken, line 50: Can we call this `official_user`, just in case we end up calling other things `official` in the future?

sarken (commented):
Just a few finishing touches and this should be ready to go!

app/helpers/comments_helper.rb
sarken: Just this one left -- I think it probably got overlooked because it was collapsed.

EchoEkhi: huh, I remember changing it to squash - must have been overwritten when I merged commits

sarken (approved):

PR #4212, opened 2022-04-24
AO3-6335 Fix comment cache expiration.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6335

## Purpose

Fix the arguments to `cache` in `comments/single_comment`. I also switched to a simpler cache key, because we now have better tools to handle the intermittent test failures that `edited_at` was added to prevent.

### DIFF
--- a/features/comments_and_kudos/add_comment.feature
+++ b/features/comments_and_kudos/add_comment.feature
@@ -30,7 +30,7 @@ Scenario: When logged in I can comment on a work
     When a chapter is added to "The One Where Neal is Awesome"
       And I view the work "The One Where Neal is Awesome" in full mode
       And I follow "Comments (1)"
-    When "comment expiration" is fixed
+    When "AO3-4214" is fixed
     # Then I should see "commenter on Chapter 1" within "h4.heading.byline"
 
   Scenario: IP address of the commenter are displayed only to an admin
--- a/features/comments_and_kudos/add_comment.feature
+++ b/features/comments_and_kudos/add_comment.feature
@@ -118,7 +118,6 @@ Scenario: Comment threading, comment editing
     And I fill in "Comment" with "Actually, I meant something different"
     And I press "Update"
   Then I should see "Comment was successfully updated"
-    #TODO Someone should figure out why this fails intermittently on Travis. Caching? The success message is there but the old comment text lingers.
     And I should see "Actually, I meant something different"
     And I should not see "Mistaken comment"
     And I should see Last Edited in the right timezone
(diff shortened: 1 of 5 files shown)


features/comments_and_kudos/add_comment.feature
```diff
@@ -6,29 +6,32 @@ Feature: Comment on work
...
     And I fill in "Comment" with "I loved this! 😍🤩"
     And I press "Comment"
   Then I should see "Comment created!"
     And I should see "I loved this! 😍🤩" within ".odd"
-    And I should not see "on Chapter 1" within ".odd"
-  When I am logged in as "author"
-    And a chapter is added to "The One Where Neal is Awesome"
-    And I follow "Entire Work"
-    And I follow "Comments (1)"
-  Then I should see "commenter on Chapter 1" within "h4.heading.byline"
```
tickinginstant, line 31: All right, this one takes some explaining:

1. In 2011, this part of the test was introduced (#282). Comment caching, with `single_comment` as the cache key, was introduced in the same PR. At the time, the Rails version was 3.0.4, so the `chapter.position_changed?` check in `app/models/chapter.rb` returned true even after the new chapter was saved. So the test passed, and the cache was expired every time a new chapter was added.
2. In 2016, the comment cache was set to expire after a week (#2388). So far so good.
3. In March 2017, the cache in `comments/single_comment` was accidentally changed to use keyword arguments (#2737). This meant that caches were no longer being expired after a week. It also introduced a peculiarity in the way that comments were cached: the cache depended on the path being viewed (with query string removed). Because posting a comment on a one-shot [redirects to the chapter instead of the work](https://otwarchive.atlassian.net/browse/AO3-2815), this means that the cache that gets written in the first half of the test is for `/chapters/:id`, while the cache that gets read in the second half of the test is for `/works/:id`. This means that the test will pass even if the cache doesn't get expired.
4. In August 2017, the pull request for the upgrade to Rails 5.1 (#2980) changed the chapter cache expiration check from `chapter.position_changed?` to `chapter.position != i+1`, presumably to avoid the deprecation warnings from continuing to use `position_changed?` in an `after_save` hook. This meant that comment caches were no longer being expired properly when a new chapter was added.
5. When I switched the cache key back to using positional arguments, it stopped keeping separate cache keys for each page the comment was viewed on. That meant that the issue introduced by the Rails 5.1 upgrade now causes the test failure that it should.

I didn't want to throw too many changes into this pull request, so I split the broken part off into a separate test to make it clearer what it's trying to test, and disabled the last line. I also modified the test so that it fails on the main branch, too.

redsummernight: Whoa, awesome detective work!

I'll see about making an issue for expiring comment caches when chapter position is changed, so we can reference it at the pending test.

tickinginstant: If you're busy with other things and haven't had a chance to make the issue, I can take a crack at it. I didn't make one before because it seemed almost too minor to bother with, but I'm willing to make it if that'd be helpful for you.

redsummernight: Thank you! I was looking for an existing issue and @sarken found [AO3-4214](https://otwarchive.atlassian.net/browse/AO3-4214), which says:

> The "Chapter X" title on both the chapter and the comments does not necessarily change properly when chapters are re-ordered.

Let's use it.

redsummernight (approved):

PR #3985, opened 2021-03-22
AO3-4313 Load work's first posted chapter for everyone but admins and co-creators

britnesissom:
# Pull Request Checklist

* [X] Have you read ["How to write the perfect pull request"](https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/)?
* [X] Have you read the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [X] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)?
* [X] Have you added the [Jira](https://otwarchive.atlassian.net) issue number
  as the *first* thing in your pull request title (e.g. `AO3-1234 Fix thing`)
* [X] Do you have fewer than 5 pull requests already open? If not, please wait
  until they are reviewed and merged before creating new pull requests.

## Issue

https://otwarchive.atlassian.net/browse/AO3-4313

## Purpose

Invited/approved co-creators and admins can now view draft works

## Testing Instructions

1. create a multi-chapter draft work and add a co-creator
2. Log in as co-creator and go to co-creator requests page
3. Open draft work you were invited to (without accepting the invite)
4. You should see the work instead of getting an error

## References

original PR: https://github.com/otwcode/otwarchive/pull/3938

## Credit

*What name do you want us to use to credit your work in the [Archive of Our Own's Release Notes](https://archiveofourown.org/admin_posts?tag=1)?*
agenderdanvers

*What pronouns do you prefer (she/her, he/him, zie/hir etc)?*
they/them

PR #4213, opened 2022-04-24
AO3-6333 Try to reduce database queries when viewing comments.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6333

## Purpose

When displaying a bunch of comments, try to avoid any database queries involving single comments. This should avoid the n+1 queries problem, where we load n comments in a single query, then perform 1 query for each of the n comments.

In particular, the actions that we address this for are:
- `CommentsController#show`
- `CommentsController#index`
- `CommentsController#unreviewed`
- `CommentsController#show_comments`

Because `CommentsController#index` uses the `comments/commentable` view, this should also reduce the number of comment-related queries in the actions that use that view:
- `WorksController#show` with the `show_comments` param
- `ChaptersController#show` with the `show_comments` param
- `AdminPostsController#show` with the `show_comments` param

But I don't make any guarantees about any non-comment-related queries produced by those controllers.

### DIFF
--- a/spec/controllers/comments_controller_spec.rb
+++ b/spec/controllers/comments_controller_spec.rb
@@ -76,16 +76,15 @@ describe CommentsController do
   describe "GET #unreviewed" do
     let!(:user) { create(:user) }
     let!(:work) { create(:work, authors: [user.default_pseud], moderated_commenting_enabled: true) }
-    let(:comment) { create(:comment, :unreviewed, commentable: work.first_chapter) }
 
     it "redirects logged out users to login path with an error" do
-      get :unreviewed, params: { comment_id: comment.id, work_id: work.id }
+      get :unreviewed, params: { work_id: work.id }
       it_redirects_to_with_error(new_user_session_path, "Sorry, you don't have permission to see those unreviewed comments.")
     end
 
     it "redirects to root path with an error when logged in user does not own the commentable" do
       fake_login
-      get :unreviewed, params: { comment_id: comment.id, work_id: work.id }
+      get :unreviewed, params: { work_id: work.id }
       it_redirects_to_with_error(root_path, "Sorry, you don't have permission to see those unreviewed comments.")
     end
 
@@ -99,6 +98,26 @@ describe CommentsController do
       fake_login_admin(create(:admin))
       get :unreviewed, params: { work_id: work.id }
       expect(response).to render_template("unreviewed")
+    end
+
+    context "when displaying multiple comments", n_plus_one: true do
+      before do
+        # Skip authorization because it's a headache and a half:
+        allow(controller).to receive(:check_permission_to_review)
+      end
+
+      populate do |n|
+        create_list(:comment, n, :unreviewed, commentable: work.first_chapter)
+      end
+
+      render_views
+
+      it "produces a constant number of queries" do
+        expect do
+          Rails.cache.clear
+          get :unreviewed, params: { work_id: work.id }
+        end.to perform_constant_number_of_queries
+      end
     end
   end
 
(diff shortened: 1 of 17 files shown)


spec/controllers/comments_controller_spec.rb
```diff
@@ -100,6 +99,26 @@
...
+      before do
+        # Skip authorization because it's a headache and a half:
+        allow(controller).to receive(:check_permission_to_review)
+      end
+
+      populate do |n|
+        create_list(:comment, n, :unreviewed, commentable: work.first_chapter)
+      end
+
+      render_views
```
redsummernight, line 113: If we have to `render_views`, let's use request specs instead (like #3985).

spec/controllers/comments_controller_spec.rb
```diff
@@ -100,6 +99,26 @@
       get :unreviewed, params: { work_id: work.id }
       expect(response).to render_template("unreviewed")
     end
+
+    context "when displaying multiple comments", n_plus_one: true do
```
redsummernight, line 103: Neat gem, thanks! I'll keep it in mind when reviewing n+1 fixes going forward.

spec/requests/comments_n_plus_one_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+  end
+
+  describe "#unreviewed" do
+    let!(:work) { create(:work, moderated_commenting_enabled: true) }
+
+    context "when displaying multiple comments", n_plus_one: true do
+      before { fake_login_known_user(work.users.first) }
+
+      populate do |n|
+        User.current_user = nil # prevent bugs with unreviewed check
```
redsummernight, line 100: Which bugs does this comment refer to?

spec/requests/comments_n_plus_one_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+        create_list(:comment, n, commentable: work.first_chapter)
+      end
+
+      warmup { get unreviewed_work_comments_path(work) }
+
+      it "produces a constant number of queries" do
+        expect do
+          get unreviewed_work_comments_path(work)
+          expect(assigns(:comments).size).to eq(current_scale)
```
redsummernight, line 109: Why does `#unreviewed` in particular need a `current_scale` check, but not the others?

I ask partly because if we can use request specs, we should probably avoid `assigns` from [rails-controller-testing](https://github.com/rails/rails-controller-testing).

spec/requests/comments_n_plus_one_spec.rb
tickinginstant: The unreviewed check in comments relies on `User.current_user`, which is set when making a request through the `ApplicationController`, but not cleared after the request is complete. This means that any checks that rely on `User.current_user` that run after a request by a logged-in user will behave as if it's the logged-in user performing those actions, even if those actions are a call to `FactoryBot.create`.

In this case, the sequence of events is:
1) 2 comments are created on the moderated work. `User.current_user` is nil, so they're marked as unreviewed.
2) The owner of the work views those two unreviewed comments, causing `User.current_user` to become set.
3) The 2 comments are deleted.
4) 3 new comments are created on the moderated work. Because `User.current_user` is set to the owner of the work, these comments count as creator comments, even though the pseuds on the comments are not creator pseuds. As a result, none of the comments are unreviewed. So you end up comparing the display of 2 comments vs. the display of 0 comments, instead of 2 vs. 3 as it should be. (This is, incidentally, why I added the check involving `current_scale`.)

The issue is widespread -- that's why most of the calls to `FactoryBot.create(:work)` in the feature tests are preceded by a call to `step %{I start a new session}`, because if `User.current_user` was still set, they'd fail with an error about the creator not allowing co-creator requests.

At some point I'd like to convert the `set_current_user` hook to an `around_action`, so that it can clear `User.current_user` after the request is over. But until then, it's sometimes going to be necessary to either log out, or manually set `User.current_user` to nil.

redsummernight: Makes sense, thanks! Yeah, we'll need to deal with `User.current_user` before we [start using request specs more heavily](https://stackoverflow.com/questions/40851705/controller-specs-vs-request-specs).

Can you change the comment to "log out to prevent commenting as work creator"?

redsummernight (approved):

PR #4216, opened 2022-04-25
AO3-6325 Add creator, creation, and bookmarker IDs to bookmark blurbs

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6325

## Purpose

Add classes to various bookmark blurbs identifying the bookmarkable/creation, its creator, and the bookmarker.


app/helpers/bookmarks_helper.rb
```diff
@@ -84,4 +84,52 @@ def get_count_for_bookmark_blurb(bookmarkable)
...
+  # Array of unique creator and bookmarker ids, formatted user-123, user-126.
+  # If the user has bookmarked their own work, we don't need their id twice.
+  def user_ids_for_bookmark_blurb(bookmark)
+    user_ids = creator_ids_for_css_classes(bookmark.bookmarkable)
+    user_ids << bookmarker_id_for_css_classes(bookmark)
+    user_ids.uniq
+  end
+
+  # The bookmark blurb contains the bookmarkable and a single user's bookmark.
+  # If we cache this, it probably only needs to use the creation cache key.
```
tickinginstant, line 102: Doesn't it also need to use the bookmark's cache key, because it also includes the bookmarker's user ID?

I think it probably should be cached, at least until we get a bunch of N+1 errors fixed up. And that's probably not going to happen before the next release.

app/helpers/bookmarks_helper.rb
```diff
@@ -84,4 +84,52 @@ def get_count_for_bookmark_blurb(bookmarkable)
...
+    user_ids = user_ids_for_bookmark_blurb(bookmark).join(" ")
+    "bookmark blurb group #{creation_id} #{user_ids}".strip
+  end
+
+  # I'm not using this yet.
+  # The bookmarkable blurb contains the bookmarkable and many users' bookmarks.
+  # This is equivalent to sticking the word "bookmark" in front of
+  # css_classes_for_creation_blurb(creation), but I don't know if we want to
+  # use the existing method because it's cached (the key is
+  # "#{creation.cache_key_with_version}/blurb_css_classes-v2").
```
tickinginstant, line 116: I like the approach you've done in the view file, where you're directly using `css_classes_for_creation_blurb`. I'm pretty neutral on whether to move that into a helper function, but I do think you should share the cache between this blurb and the creation blurb.

app/helpers/bookmarks_helper.rb
```diff
@@ -84,4 +84,52 @@ def get_count_for_bookmark_blurb(bookmarkable)
...
+    return if bookmark.nil?
+
+    creation_id = creation_id_for_css_classes(bookmark.bookmarkable)
+    creator_ids = creator_ids_for_css_classes(bookmark.bookmarkable)
+    "bookmark blurb group #{creation_id} #{creator_ids}".strip
+  end
+
+  # Also not using this yet.
+  # We could cache this if we kept the "own" logic out, but I don't know how
+  # useful that is.
```
tickinginstant, line 127: Yeah, not much point in caching this when the short blurb isn't cached.

app/views/bookmarks/_bookmark_blurb_short.html.erb
```diff
@@ -1,5 +1,6 @@
 <% # This partial requires local variable 'bookmark' %>
-<li class="user<% if is_author_of?(bookmark) %> own<% end %> short blurb group" role="article">
+<% # user own short blurb group bookmarker-id %>
+<li class="user<% if is_author_of?(bookmark) %> own<% end %> short blurb group <%= bookmarker_id_for_css_classes(bookmark) %>" role="article">
```
tickinginstant, line 3: This line is long enough that I think it might make sense to use the helper method, though I think you could make an argument the other way as well.

app/helpers/bookmarks_helper.rb
sarken: It occurs to me that if the creation (bookmarkable) is nil, we're going to be left with just "bookmark" and not "bookmark blurb group" as the class. 

I don't think we intentionally display bookmarkable blurbs for deleted items anywhere (as opposed to bookmark blurbs, which are on the user's bookmarks page), _but_ I assume there's a possibility of them being visible a few moments while the index updates. 

So I think I might update `css_classes_for_creation_blurb` to return `blurb group` when the creation is nil.

app/helpers/bookmarks_helper.rb
sarken: I was going to say yes, but... does it? The user ID for the bookmarker will never change. You can't transfer a bookmark to a different account, nor can you make it anonymous... you can only delete it.

tickinginstant: It can't change, that's true, but the contents of the cache includes the bookmarker ID, so the bookmark has to be used somewhere in the cache key. You could get away with using `bookmark.id` instead of the full `bookmark.cache_key`, though.

app/helpers/bookmarks_helper.rb
tickinginstant: I think what happens is that the deleted bookmarkable is skipped and you get a header like "1 - 19 of 100 Bookmarked Items in Testing." So I don't think you can get `nil` bookmarkables in listings like that. But I haven't tested it recently.

spec/helpers/bookmarks_helper_spec.rb
```diff
@@ -0,0 +1,223 @@
...
+  describe "#css_classes_for_bookmark_blurb" do
+    let(:default_classes) { "bookmark blurb group" }
+
+    context "when bookmarkable is ExternalWork" do
+      it "returns string with default classes, creation info, and bookmarker info" do
+        result = helper.css_classes_for_bookmark_blurb(external_work_bookmark)
+        expect(result).to eq("#{default_classes} external-work-#{external_work.id} user-#{bookmarker.id}")
+      end
+
+      context "when bookmark is updated" do
```
sarken, line 23: I didn't test updating the external work because that's... not really a thing we do, and it definitely wouldn't be updated in a way that would change the expected classes. I can add it if we want, though.

spec/helpers/bookmarks_helper_spec.rb
```diff
@@ -0,0 +1,223 @@
...
+        end
+
+        it "only includes the user id once" do
+          result = helper.css_classes_for_bookmark_blurb(series_bookmark)
+          expect(result).to eq("#{default_classes} series-#{series.id} user-#{bookmarker.id}")
+        end
+      end
+
+      context "when series is updated" do
+        context "when new creator is added" do
```
sarken, line 56: I only tested one scenario for updating series and works, but I can add more (e.g., removing creators and making the work unrevealed or anonymous).

spec/helpers/bookmarks_helper_spec.rb
redsummernight: Let's skip that.

spec/helpers/bookmarks_helper_spec.rb
redsummernight: Let's not. If those scenarios fail to change the cache keys, I expect the `css_classes_for_creation_blurb` specs in spec/helpers/application_helper_spec.rb to fail first.

redsummernight (commented):
I now experience [semantic satiation](https://en.wikipedia.org/wiki/Semantic_satiation) for "blurbs" and variants of "bookmark".

PR #4221, opened 2022-04-29
AO3-5348 Gift notification i18n

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5348

## Purpose

* Add missing formatting to series metadata label in text email
* Finish i18n, except for "by creators," which we're going to deal with for subscriptions first

## Testing Instructions

Give a gift that's a work in a series.

### DIFF
--- a/app/helpers/mailer_helper.rb
+++ b/app/helpers/mailer_helper.rb
@@ -110,6 +110,27 @@ module MailerHelper
     end
   end
 
+  # e.g., 1 word, 50 words
+  def creation_word_count(creation)
+    t("mailer.general.creation.word_count", count: creation.word_count)
+  end
+
+  # e.g., Title (x words), where Title is a link
+  def creation_link_with_word_count(creation, creation_url)
+    creation_link = style_creation_link(creation.title, creation_url)
+    word_count = creation_word_count(creation)
+    t("mailer.general.creation.link_with_word_count",
+      creation_link: creation_link, word_count: word_count).html_safe
+  end
+
+  # e.g., "Title" (x words), where Title is not a link
+  def creation_title_with_word_count(creation)
+    creation_title = creation_title(creation)
+    word_count = creation_word_count(creation)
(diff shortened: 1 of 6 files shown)


app/helpers/mailer_helper.rb
```diff
@@ -110,6 +110,34 @@ def creation_title(creation)
     end
   end
 
+  # e.g., 1 word or 50 words
```
sarken, line 113: We do something quite similar to all of this in the share code:

```
<a href="https://archiveofourown.org/works/###"><strong>Title</strong></a> (x words) by <a href="https://archiveofourown.org/users/user"><strong>creator</strong></a>
```

So I thought about not putting this in the mailer helper and not using mailer keys for the translations... but I decided I didn't have much faith in the assumption the mailers and share code would always match (whereas having similar text with similar formatting throughout the mailers is something we should always strive for). The share code should probably just be handled as its own separate thing.

spec/helpers/mailer_helper_spec.rb
```diff
@@ -1,13 +1,76 @@
...
+  let(:series) { create(:series) }
+
   describe "style_creation_link" do
     it "nests red link inside bold inside italics" do
-      work = create(:work)
       expect(style_creation_link(work.title, work_url(work))).to eq("<i><b><a style=\"color:#990000\" href=\"#{work_url(work)}\">#{work.title}</a></b></i>")
     end
   end
 
+  describe "#creation_word_count" do
```
redsummernight, line 14: I think we can skip these specs - they only make sure that `I18n.t` works. We're also covered by the `creation_*_with_word_count` specs.

app/helpers/mailer_helper.rb
redsummernight: Let's make `creation_word_count` private, so no one tries to use it elsewhere.

redsummernight (approved):

PR #4222, opened 2022-04-30
AO3-6291 Rename bookmarkable id indexing field

korrien:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6291

## Purpose

Refer to the issue's comments.

Unsure if I implemented suggestion correctly, as I am unable to reproduce the original failure and not sure of the setup.

Not sure if the "id - keyword" field simply can be renamed or if I should keep id with type "long" there as well.
Locally with dev data from testing I now had the opposite problem from reported - re-indexing bookmarks jobs failed because id was not found, so got error on fandom/bookmarks pages. Had to run indexing job to fix.
Not sure if same could happen on staging if it was possible to create new bookmarks, or if failing re-indexing job was enough to prevent any inconsistencies.

## Credit

korrien, she/her

### DIFF
--- a/app/models/search/bookmark_indexer.rb
+++ b/app/models/search/bookmark_indexer.rb
@@ -51,7 +51,7 @@ class BookmarkIndexer < Indexer
           type: "text",
           analyzer: "simple"
         },
-        id: {
+        sort_id: {
           type: "keyword"
         }
       }
--- a/app/models/search/bookmarkable_indexer.rb
+++ b/app/models/search/bookmarkable_indexer.rb
@@ -26,7 +26,7 @@ class BookmarkableIndexer < Indexer
 
   def document(object)
     object.bookmarkable_json.merge(
-      id: document_id(object.id)
+      sort_id: document_id(object.id)
     )
   end
 
--- a/app/models/search/bookmarkable_query.rb
+++ b/app/models/search/bookmarkable_query.rb
@@ -103,7 +103,7 @@ class BookmarkableQuery < Query
       sort_hash = { _score: { order: sort_direction } }
     end
 
-    [sort_hash, { id: { order: sort_direction } }]
+    [sort_hash, { sort_id: { order: sort_direction } }]
   end
 
   # Define the aggregations for the search


redsummernight:
> Unsure if I implemented suggestion correctly, as I am unable to reproduce the original failure and not sure of the setup.
> [...]
> Locally with dev data from testing I now had the opposite problem from reported

Elasticsearch infers the type of a field from the first value indexed for that field, so which problem you have depends on when you first index anything in your local development environment.

I did this to reproduce the error:

- Force rebuild the container "es" in the Docker setup, to get a blank slate (`docker-compose up --build --force-recreate --no-deps -d es`).
- Run the site at a commit before 208434f42a6a58bf7437ef6aa2f9a2f3e1913a2d and index a bookmark (`Bookmark.last.reindex_document`) and a bookmarkable (`Work.last.reindex_document`); these get processed immediately without needing Resque.
- Run `pp $elasticsearch.indices.get_mapping(index: BookmarkIndexer.index_name)` to see what the mapping actually looks like in ES, and it will include `"id"=>{"type"=>"long"}`.
- Run the site at a commit after 208434f42a6a58bf7437ef6aa2f9a2f3e1913a2d (but without your fix - this pull) and update the mapping (`BookmarkIndexer.create_mapping`). I get the error "mapper [id] cannot be changed from type [long] to [keyword]".

If you run the site with your fix, you know it works if `BookmarkIndexer.create_mapping` gives you no error, and `Bookmark.last.reindex_document` and `Work.last.reindex_document` both succeed.

> Not sure if the "id - keyword" field simply can be renamed or if I should keep id with type "long" there as well.

`"id"=>{"type"=>"long"}` will already be in the mapping in any case, so you're not renaming it, you're adding a new field to avoid conflicts with the old one.

We can't see all the inferred mappings, but we need to stay compatible with them when making changes. Our CI setup will make fresh indexes every time, so it won't catch issues like this.

I think it's a good idea to make all the mappings explicit in our code (from the output of `$elasticsearch.indices.get_mapping` above), but that can be a separate issue.

korrien:
As usual thank you very much for detailed explanation! Got it.

Checked it locally, everything went as you described it. Was able to reproduce the error and verified that with this PR `BookmarkIndexer.create_mapping` gives error no more.

redsummernight (approved):
Thanks for checking!

PR #4225, opened 2022-05-03
AO3-6200 Make bookmark_if_exists faster, so that it can handle orphan_account.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6200

## Purpose

Rewrites the query used in `bookmark_if_exists` so that it uses a join instead of passing all of the pseud IDs. This should mean that when the orphan account is logged in and browsing, it won't produce queries involving 20k pseud IDs. Also reduces the number of overall queries from 2 to 1, and adds some specs to make sure that it produces the correct results.


redsummernight (approved):
Nice, thanks! We'll get some decent mileage out of `create_invalid` for sure.

PR #4228, opened 2022-05-09
AO3-6339 Bump Rails to 6.0.5

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6339

## Purpose

Bump Rails version to address vulnerabilities in Rails sub-gems. 

## Testing Instructions

See if the tests pass.


Gemfile
```diff
@@ -6,7 +6,7 @@ gem 'test-unit', '~> 3.2'
 
 gem 'bundler'
 
-gem "rails", "~> 6.0.4"
+gem "rails", "~> 6.0.4.8"
```
redsummernight, line 9: Let's keep this 6.0.4, so Dependabot can work once https://github.com/dependabot/dependabot-core/issues/4867 is fixed.

sarken:
james wants this by itself, so no rush. I think he wanted to try a new method on the Systems side of things.

sarken:
Okay, so, for some godforsaken reason, bumping to 6.0.4.8 was no longer working. However, bumping to 6.0.5 works? We good with that?

redsummernight:
> However, bumping to 6.0.5 works? We good with that?

Yeah, it's fine. Patch version update, we can trust the test suite on this one.

redsummernight (approved):

PR #4229, opened 2022-05-12
AO3-6148 Limit admin roles that can see IPs on works/comments, and show email for guest comments.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6148

## Purpose

1. Restrict the display of the IP addresses on works and comments so that only Superadmins and Policy & Abuse can see them.
2. Add a line displaying the email address of guest comments to Superadmins, Policy & Abuse, and Support.

## Testing Instructions

See JIRA.

### DIFF
--- a/features/comments_and_kudos/admin_info.feature
+++ b/features/comments_and_kudos/admin_info.feature
@@ -1,0 +1,40 @@
+Feature: Some admins can see IP addresses and emails for comments
+  Scenario: Admin info for comments isn't visible to the work's owner
+    Given the work "Random Work"
+      And a guest comment on the work "Random Work"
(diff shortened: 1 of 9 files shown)


features/comments_and_kudos/admin_info.feature
```diff
@@ -0,0 +1,40 @@
+Feature: Some admins can see IP addresses and emails for comments
+  Scenario: Admin info for comments isn't visible to the work's owner
```
sarken, line 2: Work owner covers logged in users, but can we also test that admin info isn't visible to a guest?

sarken (approved):

PR #4230, opened 2022-05-12
AO3-6328 Allow tags to be sorted by number of uses

tararosenthal:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6328

## Purpose

This pull request is to allow tags to be sorted by number of uses in the recently created tag search. To do this, we need to allow reindexing when the use counts change and to reduce redundant writes to the database when counts do not change.

## Testing Instructions

See Jira.

## References

Initial creation of the tag search: https://otwarchive.atlassian.net/browse/AO3-5830
Unmerged PR attempting to reduce redundant writes: https://github.com/otwcode/otwarchive/pull/2868/files

## Credit

Tara Rosenthal (she/her)


tararosenthal:
I plan on editing the added cucumber test to increase efficiency based on some suggestions from ticking instant. If someone views this in the meantime, please let me know if the spec test I added `should only update the database when counts change` should be divided into separate tests. Thank you!

tararosenthal:
I plan on editing the added cucumber test to increase efficiency based on some suggestions from ticking instant. If someone views this in the meantime, please let me know if the spec test I added `should only update the database when counts change` should be divided into separate tests. Thank you!

tararosenthal:
I went ahead and edited the cucumber test based on the recommendation from ticking instant and separated out the rspec test into separate tests. I fixed most of the Hound issues except avoiding instance variables in the rspec test as I was following the formatting from the other related tests. I can change this if you'd like. Also not sure why that test failed, but it passes on my end. It only failed on my most recent commit which was just minor formatting changes.

tararosenthal:
I switched from using instance to local variables.

tararosenthal:
I couldn't stop wondering if the changes I made affected performance, so I changed the search and sort by uses test back to the version where different works were created to increase the tag count. The test passed in every case. It seems like Elasticsearch may be reindexed whether or not update_all is used? 

If there's some sort of caching going on when you run a test multiple times, these numbers might not mean anything. But here is what I found in the order in which I ran the tests.

A. Kept code to reduce redundant writes to the database but used update_all instead of update to prevent callbacks:
1. 0m21.086s
2. 0m18.603s

B. Current code of the pull request with all the changes:
1. 0m17.490s
2. 0m16.639s

C. Reverted all the code changed in the tag.rb back to the original (allowed redundant writes and update_all)
1. 0m17.427s
2. 0m18.911s

Then I reran all versions one final time in the same order:
A3: 0m17.154s
B3: 0m16.605s
C3: 0m15.893s

tickinginstant:
A reindex is not triggered when `update_all` is used. Small tags are enqueued to be indexed when a tagging is added:
https://github.com/otwcode/otwarchive/blob/7e14cc63ce268719163d572a055aedf28a7b6afc/app/models/tagging.rb#L43
So small tags will get a reindex some time after they increase or decrease in size. But that reindex isn't guaranteed to occur after the `Tag.write_redis_to_database` task runs, so it's not guaranteed to have access to the new `taggings_count_cache` values. And large tags have no reindexing at all currently.

In the "Search and sort by Uses" test, you call the step `the periodic tag count task is run` in the tag creation step. This means that when reindexing occurs in the second step, the correct `taggings_count_cache` values are available. But if you modified the tag creation step to insert an extra `all reindexing jobs have been run` step before the `the periodic tag count task is run` step, you'd get test failures when using `update_all`, but not when using `update`. That's because performing a reindex clears the reindexing queue, and `update` re-enqueues, but `update_all` does not.

tararosenthal:
I tried to add tests to check if large and small tags are being enqueued to be reindexed after their counts are updated and `Tag.write_redis_to_database` is called, but I couldn't find the tags in either the background or main queue. I must be missing something, but I'm not sure if it's important to test. I added the indirect test for this that ticking instant suggested in the cucumber test.

Here's an example of the test I was trying to write (I tried many different variations):

```
it "triggers reindexing of tags which aren't used much" do
  tag = FactoryBot.create(:fandom)
  FactoryBot.create(:work, fandom_string: tag.name)
  Tag.write_redis_to_database
  tag.reload
  run_all_indexing_jobs

  expect do
    FactoryBot.create(:work, fandom_string: tag.name)
    Tag.write_redis_to_database
    tag.reload
  end.to(add_to_reindex_queue(tag, :background) &
         add_to_reindex_queue(tag, :main))
end
```

tararosenthal:
Thank you again so much for your help! I tried using the `base_class` method in `add_to_reindex_queue`; however, I got this error: 
`NoMethodError:
       undefined method 'base_class' for #<Fandom:0x000055c4cca345d0>`

Also found out that `IndexQueue.get_key` is sometimes passed strings of the class names instead of the classes themselves. Was able to get the test to work adding this code:
```
if (klass.is_a?(Class))
  klass = klass < Tag ? Tag : klass
end
```

Was also able to fix it in `add_to_reindex_queue` using this code:
```
klass = item.is_a?(Tag) ? Tag : item.class
queue_key = IndexQueue.get_key(klass, queue)
```
I decided to go with the fix in `IndexQueue.get_key`, but let me know if you prefer the other.

tickinginstant:
`base_class` is a class method, not an instance method, so you'd have to call `item.class.base_class` or `klass.base_class` to make it work.

(I think that's preferable to checking Tag because it's more general -- if we ever wanted to, say, index prompts, which is the other class that currently uses STI, we'd have to special-case the queue used for that as well.)

tararosenthal:
Ah, thank you again for your help. That is a silly oversight on my part. I agree that using `base_class` makes the most sense, so I went ahead and updated that.

spec/models/tag_spec.rb
```diff
@@ -77,6 +78,81 @@
...
+        run_all_indexing_jobs
+
+        expect do
+          write_tag_to_database(tag)
+        end.to(add_to_reindex_queue(tag, :main))
+      end
+
+      it "triggers reindexing of tags which are used significantly" do
+        tag = fandom_tag_with_one_work
+        (1..ArchiveConfig.TAGGINGS_COUNT_MIN_CACHE_COUNT).each do
```
redsummernight, line 147: I'm not sure this test and the previous one are different. In both cases, you're using tags with use count under `ArchiveConfig.TAGGINGS_COUNT_MIN_CACHE_COUNT` ("small").

Note that in:
https://github.com/otwcode/otwarchive/blob/e7c7da0a33a554cfde124ef1ad2e3df5ce005470/spec/models/tag_spec.rb#L28-L35
It's a large tag because one more work was added outside the loop. The test wants to check the difference when a tag goes from "small" to "large".

If you don't care about that transition, you can simply use `ArchiveConfig.TAGGINGS_COUNT_MIN_CACHE_COUNT.times do` to create enough works for "large".

spec/models/tag_spec.rb
```diff
@@ -77,6 +78,81 @@
...
+
+        write_tag_to_database(tag)
+        expect(tag.taggings_count_cache).to eq 1
+      end
+
+      it "triggers reindexing of tags which aren't used much" do
+        tag = fandom_tag_with_one_work
+        FactoryBot.create(:work, fandom_string: tag.name)
+        # Run indexing jobs to prevent automatic reindexing with small tags
+        run_all_indexing_jobs
```
redsummernight, line 138: I'm not sure why `run_all_indexing_jobs` is used here, is it to clear out any existing indexing jobs?

sarken:
Thank you for all your work on this! weeklies has updated the PR in #4632, so I'm going to close this one.

PR #4235, opened 2022-05-19
AO3-6290 "Chapter # of [title]" in co-creation emails is not translatable

EchoEkhi:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6290

## Purpose

Translate "Chapter # of [title]" in co-creation emails.

## Testing Instructions

See JIRA ticket.

## Credit

EchoEkhi (He/him)

### DIFF
--- a/spec/helpers/mailer_helper_spec.rb
+++ b/spec/helpers/mailer_helper_spec.rb
@@ -14,7 +14,7 @@ describe MailerHelper do
   describe "#creation_title" do
     context "when creation is a work" do
       it "returns the work title" do
-        expect(creation_title(work)).to eq("#{work.title}")
+        expect(creation_title(work)).to eq(work.title)
       end
     end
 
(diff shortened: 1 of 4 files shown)


sarken (changes requested):
I think we need some tests for the `creation_title(creation)` method in mailer_helper_spec.rb as well, just to be sure we don't mess up the key and have a missing translation. (The existing specs for the co-creator emails wouldn't catch a missing translation here because it looks like they only test the emails when the creation is a work.)

spec/helpers/mailer_helper_spec.rb
```diff
@@ -11,6 +11,20 @@
     end
   end
 
+  describe "#creation_title" do
```
sarken, line 14: If you could also check when creation is a series, everything will be covered and this should be good to go!

sarken (approved):
Thanks!

PR #4244, opened 2022-05-31
AO3-6144 Prevent protected users' works from being cited as related works

weeklies:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6144

## Purpose

Prevent protected users from having their works cited as related works

## Testing Instructions

See Jira issue. Added another test case in comment.

## Credit
weeklies (she/her)


app/models/work.rb
```diff
@@ -180,6 +180,22 @@ def new_recipients_allow_gifts
     end
   end
 
+  validate :parent_users_allow_related_works
+
+  def parent_users_allow_related_works
```
sarken, line 185: Maybe `new_parents_can_be_cited`? I feel like `parent_users` might throw someone when seeing this validation without the context of the pull request title.

app/models/work.rb
```diff
@@ -180,6 +180,22 @@ def new_recipients_allow_gifts
...
+    return if self.new_parent.blank?
+
+    parent = self.new_parent[:parent]
+    return if !parent.respond_to?(:pseuds)
```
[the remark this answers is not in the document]
sarken: Hound makes a good point here and below.

app/models/work.rb
```diff
@@ -180,6 +180,22 @@ def new_recipients_allow_gifts
...
+
+    users = parent.pseuds.collect(&:user).uniq
+    users.each do |user|
+      if user.protected_user
+        self.errors.add(:base, ts("You can't use the related works function to cite works by #{user.default_pseud.name}."))
```
sarken, line 194: The issue wasn't entirely clear about what name to use here, which is totally my fault. However, we should probably use the user's login, because the default pseud a) doesn't always match the user login and b) might not be the pseud on the work.

It also belatedly occurs to me this could break anonymity in the case of unrevealed or anonymous works (e.g., if I'm a protected user and I post an anon work, when someone tries to cite it, they'll get the error "You can't use the related works function to cite works by sarken"). 

We probably only want to do this validation if the work is neither anon or unrevealed.

sarken (changes requested):

sarken (approved):
Thank you!

PR #4245, opened 2022-05-31
AO3-6327 Assorted fixes for the comment blocking code.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6327

## Purpose

Minor fixes for comment blocking, as suggested in the post-merge review on #4237:
- a few tweaked comments
- order blocks by both `created_at` and `id`, instead of just `id`
- remove the unused info from the error added by `NotBlockedValidator`
- remove an unnecessary lazy quantifier
- change the title of the Blocked Users page


redsummernight (approved):

PR #4246, opened 2022-06-01
AO3-6327 Fix admin view of Blocked Users page, and add a Cancel button to the Confirm Block/Unblock pages.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6327

## Purpose

- Fix the 500 error that occurs when admins try to view a user's Blocked Users page.
- Add a Cancel button to the Confirm Block page and the Confirm Unblock page.

### DIFF
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -73,7 +73,7 @@ class Comment < ApplicationRecord
   scope :for_display, lambda {
     includes(
       pseud: { user: [:roles, :block_of_current_user, :block_by_current_user] },
-      parent: { work: :pseuds }
+      parent: { work: [:pseuds, :users] }
     )
   }
 
--- a/app/helpers/block_helper.rb
+++ b/app/helpers/block_helper.rb
@@ -1,11 +1,16 @@
 module BlockHelper
   def block_link(user, block: nil)
-    block ||= user.block_by_current_user
+    if block.nil?
+      block = user.block_by_current_user
+      blocking_user = current_user
+    else
+      blocking_user = block.blocker
+    end
 
     if block
-      link_to(t("blocked.unblock"), confirm_unblock_user_blocked_user_path(current_user, block))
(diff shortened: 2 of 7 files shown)


app/models/comment.rb
```diff
@@ -73,7 +73,7 @@ def check_for_spam
   scope :for_display, lambda {
     includes(
       pseud: { user: [:roles, :block_of_current_user, :block_by_current_user] },
-      parent: { work: :pseuds }
+      parent: { work: [:pseuds, :users] }
```
redsummernight, line 76: Which action needs this btw? spec/requests/comments_n_plus_one_spec.rb didn't catch its absence.

tickinginstant: I noticed that there's an N+1 error when a user is viewing comments on their own work. It comes from calling `blocked_by?`:
https://github.com/otwcode/otwarchive/blob/27f1b745078139bd9aadf285de1feb1c8c0a4a70/app/helpers/block_helper.rb#L17-L21
When `blocked_by?` is called on a work, `blocker` is equal to the association `work.users`. When someone other than the work's owner views comments, the first call to `blocked_by?(work)` will produce these queries:
- one for calling `include?` on the unloaded `work.users` association
- one to load the `work.users` association as a result of calling `any?` on the unloaded association
- queries loading `block_of_current_user` for all users loaded

When `blocked_by?(work)` gets called again, this means that there are no additional queries, because the `work.users` association is now loaded.

When the work's owner views the comments, however, the function returns false before calling `any?`, so the `work.users` association is never loaded. This means that every call to `blocked_by?(work)` results in calling `includes?` on an unloaded association, and therefore a database query.

redsummernight: It's fine as is. Thank you for the breakdown!

redsummernight (approved):

PR #4247, opened 2022-06-02
AO3-5239 AO3-4659 Restyle collection member and blocked user autocomplete field

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4659
https://otwarchive.atlassian.net/browse/AO3-5239

## Purpose

Modifies the user autocomplete fields on the collection participant and blocked user page to make them look better and work better on mobile.

## Testing Instructions

Refer to Jira.

## Credit

Sarken, she/her


app/views/blocked/users/index.html.erb
```diff
@@ -17,13 +17,19 @@
 </div>
 
 <% # form for blocking users %>
-<div class="actions">
-  <%= form_tag confirm_block_user_blocked_users_path(@user), method: :get do %>
-    <%= text_field_tag :blocked_id, "", autocomplete_options("pseud", size: 60, data: { autocomplete_token_limit: 1 }) %>
```
redsummernight, line 22: Discussed elsewhere: we'll remove the ["size" HTML attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/size) because we don't really need it for presentation (it's on the hidden input elements and jquery.tokeninput.js doesn't copy it to the displayed input elements, plus CSS takes precedence) or validation (that needs to happen server-side regardless).

redsummernight (approved):

sarken:
I've copied the Reversi and Low Vision Default changes over to staging. The skin recaching that happens on deploy should update the skins and this will then be ready for testing without further poking.

PR #4249, opened 2022-06-04
Skip Dreamwidth importing tests

redsummernight:
## Issue

None.

## Purpose

Disable failing tests until we can fix them ([AO3-6353](https://otwarchive.atlassian.net/browse/AO3-6353)).

Also delete an old test with unclear checks.

## Testing Instructions

None.

### DIFF
--- a/features/importing/archivist.feature
+++ b/features/importing/archivist.feature
@@ -45,6 +45,8 @@ Feature: Archivist bulk imports
       And I press "Import"
     Then I should see "We couldn't successfully import that work, sorry: No external author name or email specified"
 
+  # TODO: Enable after AO3-6353.
+  @wip
   Scenario: Importing for an author without an account should have the correct byline and email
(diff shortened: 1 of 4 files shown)


features/importing/archivist.feature
```diff
@@ -45,26 +45,34 @@ Feature: Archivist bulk imports
       And I press "Import"
     Then I should see "We couldn't successfully import that work, sorry: No external author name or email specified"
 
+  # TODO: Enable after AO3-6353.
```
sarken, line 48: Should  this LJ one be re-enabled too or is it failing for reasons other than LJ being down for maintenance?

redsummernight: Just that reason. It fails on my fork, but the main repo should have the page cached by the VCR gem so it should still pass for now.

sarken (approved):

PR #4250, opened 2022-06-04
AO3-6346 Return HTML error pages for non-registered MIME types

potpotkettle:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6346

## Purpose

This will prevent raw HTML pages from showing up when non-existing work IDs and unknown file extensions are specified.

## Testing Instructions

This issue didn't reproduce in the development mode. I had to set ``config.consider_all_requests_local = false`` in config/environments/development.rb to reproduce it with Docker.

## References

n/a

## Credit

Potpotkettle (they/them)


sarken:
Hi, Potpotkettle!

app/controllers/errors_controller.rb
```diff
@@ -4,7 +4,7 @@ class ErrorsController < ApplicationController
     define_method error_code.to_sym do
       respond_to do |format|
         format.any(:html, :text, :pdf, :mobi, :epub, :azw3) { render error_code, status: error_code.to_i }
```
redsummernight, line 6: Looking at the existing code, the intent seems to be:

- for known formats (html, pdf, etc.) return the 404 page
- for all else return an empty response with just the status code

But it doesn't work out like that. `format.any(:html, :text, :pdf, :mobi, :epub, :azw3)` implies we have pdf versions of the error templates, which we don't.

I think you can replace the whole `respond_to` block with just:
```rb
render error_code, status: error_code.to_i, formats: :html
```
and it should work for the .htm case as well.

app/controllers/errors_controller.rb
```diff
@@ -4,7 +4,7 @@ class ErrorsController < ApplicationController
     define_method error_code.to_sym do
```
redsummernight, line 4: Can you also remove 502 from the list above? We don't have a template for it in app/views/errors, or a route for it in config/routes.rb. It's a static error page that doesn't need to be processed by this controller.

app/controllers/errors_controller.rb
potpotkettle: Thanks a lot for the review. It looks like the header-only response for unknown formats was indeed part of the original intent, as seen in these changes in f03dbb8aa63bdb7ce8f3269fed7ee6bf518c8d47:
```diff
-        format.any(:html, :text, :pdf, :mobi, :epub) { render error_code, :status => error_code }
-        format.any(:js) { head error_code }
+        format.any(:html, :text, :pdf, :mobi, :epub) { render error_code, :status => error_code.to_i }
+        format.all { render :nothing => true, :status => error_code.to_i }
```

Just to make sure, are we going to drop this part as well

> for all else return an empty response with just the status code

and make it always render an HTML (as long as there is a template file for the given error code)?

redsummernight: Yeah, I think it's friendlier to the user to render HTML if all else fails.

potpotkettle: Understood, thanks! The next commit should address this.

app/controllers/errors_controller.rb
potpotkettle: Thanks, this will be done in the next commit (and included in the commit message).

spec/controllers/errors_controller_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+      expect(response.header["Content-Type"]).to eq("text/html; charset=utf-8")
+    end
+  end
+
+  describe "auth_error" do
+    render_views
+
+    it "returns an auth error page" do
+      get "auth_error"
+      expect(response.status).to eq(200)
```
potpotkettle, line 37: Incidentally, should the "Auth Error" page have the status code 401 or something, instead of 200? (Maybe this is a separate issue anyway, though.)

app/controllers/errors_controller.rb
redsummernight: Since we're responding to all formats in the same way, you should be able to write:
```rb
%w(403 404 422 500).each do |error_code|
  define_method error_code.to_sym do
    render error_code, status: error_code.to_i, formats: :html
  end
end
```
You can also leave out the content type, since `formats: :html` is already specified.

When you're making changes to pull requests, we generally prefer that you don't force push. That way it's easier for reviewers to view changes compared to the previous round of review.

spec/controllers/errors_controller_spec.rb
redsummernight: Maybe, but let's see how returning non-200 status codes behaves on staging/production environments first.

spec/controllers/errors_controller_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+  describe "500" do
+    it "returns an HTML 500 page" do
+      get :"500"
+      expect(response.status).to eq(500)
+      expect(response.header["Content-Type"]).to eq("text/html; charset=utf-8")
+    end
+  end
+
+  describe "auth_error" do
+    render_views
```
redsummernight, line 33: We generally avoid using `render_views` in controller specs. If we do need to examine the response page, we would either use request specs (spec/requests/) or Cucumber tests (features/). For this test, I think it's fine to skip `render_views` and the page title check - or even skip it entirely (until we change it more, e.g. return status code 401).

potpotkettle: > even skip it entirely (until we change it more, e.g. return status code 401).

I'll take this advice and remove additions related to "auth_error".

potpotkettle:
It looks like the recent test failures are related to (and solved by) #4262.
```
Failures:
  1) Download author_names when the pseud is the same as the username returns the pseud by itself
Failure/Error: let(:simple_author) { FactoryGirl.build(:pseud, name: "SimpleAuthor", user: simple_user) }
NameError:
  uninitialized constant FactoryGirl
  Did you mean?  FactoryBot
```

potpotkettle:
[The recent test results](https://github.com/otwcode/otwarchive/runs/6848255980?check_suite_focus=true) here say:

```
Failing Scenarios:
cucumber features/works/chapter_edit.feature:533 # Scenario: You should be able to invite a co-creator to a chapter if they allow it.
/home/runner/work/otwarchive/otwarchive/vendor/bundle/ruby/2.7.0/gems/cucumber-core-3.2.1/lib/cucumber/core/test/result.rb:12: warning: calling private without arguments inside a method may not have the intended effect
```

I tried to replicate it by running ``docker-compose exec -e RAILS_ENV=test web bundle exec cucumber features/works/chapter_edit.feature`` locally, but I couldn't. (All checks passed.)

redsummernight:
Yeah, [some of our tests are flaky](https://github.com/otwcode/otwarchive/wiki/Automated-Testing#Known_Test_Failures) when run by GitHub Actions. Usually, but not always, because they are run fast enough that timestamp-based caches don't have the chance to expire. I've added features/works/chapter_edit.feature:533 to the list.

redsummernight (approved):

potpotkettle:
That makes sense. Thanks for the explanation and for reviewing this, @redsummernight.

redsummernight:
Thanks for working on it!

PR #4251, opened 2022-06-05
AO3-5521 Record a work's original creator on orphaning

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5521

## Purpose

When a work is orphaned, temporarily store the original creator's ID and make it visible to select admins.

## Testing Instructions

Added to the Jira ticket.

**Before merging, post two works and orphan one of them.** This should be done within the 72-hour window.

## Credit

What name and pronouns should we use to credit you in the [Archive of Our Own's Release Notes](https://archiveofourown.org/admin_posts?tag=1)? **Brian (they/he)**


sarken:
Thanks for this PR! I just wanted to note that I added some text under Testing Instructions, per request on the Jira ticket.

redsummernight (commented):
Generally we use Redis for stuff that we need to read/write quickly/frequently, that we can afford to lose and retrieve from the database if Redis dies for whatever reasons.

I think we should save the original creator ID in the database, along with the timestamp of the orphaning action. We'll have better control of when to purge the ID.

brianjaustin:
@redsummernight I saw that legal was involved for questions about the expiration time on the ticket, so my rationale was that it would be better to risk losing the information early and rely on Redis's built-in expiration mechanism to avoid something like a background job erroring and failing to remove the info after 1+ days (for example). If that's not as much of a concern as I've been assuming, I'd be happy to use a DB-based approach, although that might need a bit of tweaking depending on how important it is that we handle the multiple orphaneers edge case I mentioned above

redsummernight:
> I saw that legal was involved for questions about the expiration time on the ticket, so my rationale was that it would be better to risk losing the information early and rely on Redis's built-in expiration mechanism to avoid something like a background job erroring and failing to remove the info after 1+ days (for example).

For this use case, we care that we have the info reliably, and that we don't have it forever. So yeah, the actual expiration time does not need to that strict.

> The ticket specifically mentions works, but I think this would be fairly straightforward to extend to other Createables if the approach looks good and there's a desire to do so.

We only care about this info on works for now, so you don't have to generalize the approach in Creatable yet.

> Additionally, there is an edge case where, if multiple creators orphan the same work within 72 hours, the information will disappear 72 hours after the first one orphaned. I didn't think this would be common enough to cause issues, but if that's not the case I can come up with something else.

I talked to @sarken about this - we could have an array column for all the original creator IDs. The column gets deleted after orphaned_at timestamp + TTL, and orphaned_at is bumped any time a creator on the work orphans it. For example, if there are 2 creators, and creator B orphans the work 24h after creator A, then creator A's ID is technically kept around for 96h.

Thank you for working on this, especially when the requirements are not fully clear from the start.

brianjaustin:
@redsummernight thanks for the suggestions! I think this should work and fit the requirements now (although I had to do some fakery to get things scheduled locally)

db/migrate/20220717185757_add_original_creators_to_work.rb
```diff
@@ -0,0 +1,51 @@
+class AddOriginalCreatorsToWork < ActiveRecord::Migration[6.0]
+  def up
+    if Rails.env.staging? || Rails.env.production?
+      database = User.connection.current_database
+
+      puts <<~PTOSC
+        Schema Change Command:
+
+        pt-online-schema-change D=#{database},t=works \\
```
brianjaustin, line 9: @redsummernight @tickinginstant how comfortable are you folks with
1. Adding new fields to this table in general; and
2. Scanning through in the background cleanup job?
For the second, an index might be helpful although maybe not necessary since it is, after all, a background job. Another option I've been mulling over is to add a new table with this information separately; that could also be a slightly nicer way to store multiple orphaners. Thoughts?

config/config.yml
```diff
@@ -479,3 +479,7 @@ PERMITTED_HOSTS: [
 ]
 
 USER_RENAME_LIMIT_DAYS: 7
+
+# Number of seconds to retain the original creator for works that have been
```
zz9pzza, line 483: The issue says 72 hours, what is the units ?

db/migrate/20220717185757_add_original_creators_to_work.rb
tickinginstant: I'm personally kind of iffy about adding more columns to the works table in this case, and would prefer it to be stored in a different table with one row per former creator (which is what I think you're suggesting as an alternative). As you say, it handles multiple orphanings in a cleaner way, because you can have a separate date associated with each former creator.

tickinginstant:
A thought: because `orphan_account` automatically accepts creatorship invitations, it might be possible to bypass this logging by manually adding `orphan_account` as a co-creator before removing yourself as a co-creator. It wouldn't transfer comments like normal orphaning, but otherwise it'd have the same effect of making it difficult for PAC to see who the original creator was.

Maybe instead of logging whenever someone clicks the "Orphan Work" button, it'd be better to log user IDs whenever a creatorship is deleted? Or log when orphaning, but also log when using the "Remove Me As Co-Creator" button?

config/config.yml
brianjaustin: Whoops, should be hours. I've updated the name to make that clear

db/migrate/20220717185757_add_original_creators_to_work.rb
brianjaustin: Ok, I've updated here with a somewhat basic implementation using a new table. I know it was mentioned earlier that we only need this for `Work`, but since this method would be harder to extend to other `Createable`s I've done some exploration of a potential polymorphic approach in https://github.com/brianjaustin/otwarchive/tree/orphan-alt-polymorphic. What do you think?

tickinginstant: I can't speak for PAC, so you should probably check with someone who can, but I suspect that it won't need to handle chapters or series -- you can't orphan just a chapter, and orphaning a series also orphans the works in it, so in 99.9% of cases all the critical information will be available on the works.

Also, I think aside from table names/column names, it's not too hard to transition to polymorphism if it ends up being needed for more than works. (Just add a new non-null column with a default value `Work`, and run the migration before deploy. The table's probably going to pretty small, so we probably wouldn't even need to do ptosc for the migration.)

tickinginstant (commented):
This isn't a thorough review, mostly just some questions that arose when I read the code (and one suggestion).

db/migrate/20220717185757_add_original_creators_to_work.rb
brianjaustin: Fair enough! From @redsummernight's previous comment
> We only care about this info on works for now, so you don't have to generalize the approach in Creatable yet.

so I'm going to resolve this thread for now.

db/migrate/20220911173601_create_work_original_creators.rb
```diff
@@ -0,0 +1,12 @@
+class CreateWorkOriginalCreators < ActiveRecord::Migration[6.0]
+  def change
+    create_table :work_original_creators do |t|
+      t.references :work, null: false, dependent: :destroy
```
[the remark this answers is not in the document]
brianjaustin: I don't think so now I think about it... I think I saw that somewhere and just blindly trusted it :facepalm:

db/migrate/20220911173601_create_work_original_creators.rb
```diff
@@ -0,0 +1,12 @@
...
+      t.integer :user_id, null: false
```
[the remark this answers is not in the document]
brianjaustin: I still sometimes get that confused with foreign keys, so I thought it would be weird with deleted users. I've fixed it now

app/models/creatorship.rb
```diff
@@ -177,6 +178,17 @@ def check_not_last
     raise ActiveRecord::RecordInvalid, self
   end
 
+  # Record the original creator if the creation is a work.
+  # This information is stored temporarily to make it available for
+  # Policy and Abuse on orphaned works.
+  def save_original_creator
+    return unless creation.is_a?(Work)
+
+    creation.original_creators << WorkOriginalCreator.new(user_id: pseud.user.id)
```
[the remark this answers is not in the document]
brianjaustin: Fancy, thanks!

app/models/creatorship.rb
```diff
@@ -77,6 +77,7 @@ def check_approved_becoming_false
 
   before_destroy :expire_caches
   before_destroy :check_not_last
+  before_destroy :save_original_creator
```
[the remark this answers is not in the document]
brianjaustin: Yep; I don't _think_ we want that (although I could see an argument the other way) so I've added a `creation.destroyed?` guard. If you know of a cleaner way I'd be open to that too

app/models/work_original_creator.rb
```diff
@@ -0,0 +1,16 @@
+class WorkOriginalCreator < ApplicationRecord
+  belongs_to :work
+
+  # Get the id and username (if still available) for the associated user.
+  def display
+    user = User.find_by(id: user_id)
```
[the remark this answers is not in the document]
brianjaustin: Not really, I just wasn't sure how it would work for deleted users. After some local testing to satisfy my concern I've updated, and it's a lot nicer now

config/config.yml
tickinginstant, line 483: Can you change the units in the comment to hours? It looks like you added the units to the variable name, but the comment still says seconds.

tickinginstant (changes requested):
Looking pretty good! There's an issue with the scheduled job that needs resolving, but otherwise my comments are pretty minor.

config/config.yml
brianjaustin: Thanks for the catch! Can't believe I missed it earlier :facepalm:

config/resque_schedule.yml
```diff
@@ -75,3 +75,12 @@ check_invite_queue:
...
   description: "Invite users from the queue if it's time to do so."
+
+cleanup_work_original_creators:
+  every: 1h
+  class: "WorkOriginalCreator"
+  queue: utilities
+  args: cleanup
+  description: >-
+    Remove original_creators for works orphaned/moved more than
+    ORIGINAL_CREATOR_TTL_HOURS hours ago.
```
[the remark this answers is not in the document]
brianjaustin: Somehow I got it in my head I'd tested this... but I definitely didn't. I've `include`d `AsyncWithResque` since I don't think we need anything custom in `perform`. How does it look?

tickinginstant (approved):
Thanks!

sarken:
Hey, @brianjaustin! We'd like to include this in our next release if you have time to resolve the merge conflicts in the next few days. Do you think you could do that? If not, no worries, we'll just merge #[4348](https://github.com/otwcode/otwarchive/pull/4348) and leave this for another day.

brianjaustin:
@sarken should be all set now!

sarken:
Yay, thank you!

PR #4252, opened 2022-06-06
AO3-6354 Ignore hidden elements with Capybara in feature tests

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6354

## Purpose

- Update Capybara to 3.x.
- Remove setting `Capybara.ignore_hidden_elements` to false in feature tests, so matchers no longer find text or elements that are not visible to users. I kept different issues in separate commits.

## Testing Instructions

None, automated.


features/step_definitions/skin_steps.rb
```diff
@@ -178,28 +178,36 @@
...
   skin = Skin.find_by(title: skin_name)
-  page.should have_xpath("//link[contains(@href, '#{skin.skin_dirname}')]")
+  expect(page).to have_css("link[href*='#{skin.skin_dirname}']", visible: false)
 end
 
-Then /^the page should not have the cached skin "([^"]*)"$/ do |skin_name|
+Then "the page should not have the cached skin {string}" do |skin_name|
   skin = Skin.find_by(title: skin_name)
-  page.should_not have_xpath("//link[contains(@href, '#{skin.skin_dirname}')]")
+  expect(page).not_to have_css("link[href*='#{skin.skin_dirname}']", visible: false)
```
redsummernight, line 196: Note that both `.not_to have_css` and `.to have_no_css` RSpec matchers [can wait for AJAX requests](https://github.com/teamcapybara/capybara#asynchronous-javascript-ajax-and-friends); they're functionally equivalent.

Gemfile
```diff
@@ -123,8 +123,6 @@ group :test do
   gem "rspec-rails", "~> 4.0.1"
   gem 'pickle'
   gem 'shoulda'
-  gem 'capybara', '~> 2.16.1'
-  gem 'cucumber', '~> 3.1'
```
tickinginstant, line 127: I think I'd prefer it if you deleted just the version numbers here, instead of the whole lines. They're included either way, but it's nice to be able to look at the Gemfile and get an idea of which gems are being used.

features/support/capybara.rb
```diff
@@ -7,10 +7,16 @@
...
-
   # Increased timeout to minimise failures on CI servers.
   config.default_max_wait_time = 25
+
+  # Capybara 2.x behavior: match rendered text, squish whitespace by default.
+  config.default_normalize_ws = true
+
+  # Capybara 3.x changes the default server to Puma; we have WEBRick
+  # (a dependency of Mechanize, used for importing; also used for the
+  # Rails development server), so we;ll stick with that for now.
```
tickinginstant, line 18: `we;ll` should be `we'll`

features/step_definitions/tag_set_steps.rb
```diff
@@ -13,7 +13,9 @@
       tags = scanned_tags.split(/, ?/)
       tags.each { |tag| check(tag) }
     else
-      fill_in("owned_tag_set_tag_set_attributes_#{type}_tagnames_to_add", with: scanned_tags)
+      field_name = "owned_tag_set_tag_set_attributes_#{type}_tagnames_to_add"
+      field_name += "_autocomplete" if @javascript
```
sarken, line 17: Nice.

PR #4254, opened 2022-06-06
AO3-5780 Add thousands separators to long numbers on series pages

irrationalpie7:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5780

## Purpose

This uses `number_with_delimiter` to display bookmark counts on the series page or on the series bookmarks page with delimiters if they have more than 3 digits.

Note: it was unclear to me whether this issue also referred to number of works in a series, so I left that alone, but I can add that in if it should be included!

## Testing Instructions

If you just want to see that the bookmark number still exists after this change, you can

1. create or find a work in a series
2. go to the series page
3. bookmark the series
  a. verify that after "bookmarks:" there's a 1 that's a link
4. go back to the series page
  a. verify that after "bookmarks:" there's a 1 that's a link

... I'm very new to docker, ruby/rails, *and* this project and could not figure out how to get existing automated tests to pass even before my changes, unfortunately, let alone find the relevant tests and modify or add to them. I'd love to learn! I got a ton of errors that looked like the following when I tried to run the spec tests, which I'm not even sure is the correct place to be looking for tests of these files.
```
NameError:
        undefined local variable or method `renamed_at' for #<User:0x000055ad2c583c78>
        Did you mean?  created_at
```
I suspect I must have failed to set up the test database correctly (`docker exec web bundle exec rake db:test:prepare`?)? I notice that `test/fixtures/users.yml` seems to have "created_at" info on users but not "renamed_at" info, but I have no idea if that's relevant.

## Credit

irrationalpie, they/them

### DIFF
--- a/app/views/series/_series_module.html.erb
+++ b/app/views/series/_series_module.html.erb
@@ -47,6 +47,6 @@
   <dd><%= series.visible_work_count %></dd>
   <% if (bookmark_count = series.bookmarks.is_public.count) > 0 %>
     <dt><%= ts('Bookmarks:') %></dt>
-    <dd><%= link_to_bookmarkable_bookmarks(series, bookmark_count.to_s) %></dd>
+    <dd><%= link_to_bookmarkable_bookmarks(series, number_with_delimiter(bookmark_count)) %></dd>
   <% end %>
 </dl>
--- a/app/views/series/_series_module.html.erb
+++ b/app/views/series/_series_module.html.erb
@@ -44,7 +44,7 @@
   <dt><%= ts('Words:') %></dt> 
   <dd><%= number_with_delimiter(series.visible_word_count) %></dd>
   <dt><%= ts('Works:') %></dt> 
-  <dd><%= series.visible_work_count %></dd>
+  <dd><%= number_with_delimiter(series.visible_work_count) %></dd>
   <% if (bookmark_count = series.bookmarks.is_public.count) > 0 %>
     <dt><%= ts('Bookmarks:') %></dt>
     <dd><%= link_to_bookmarkable_bookmarks(series, number_with_delimiter(bookmark_count)) %></dd>
--- a/app/views/series/show.html.erb
+++ b/app/views/series/show.html.erb
@@ -61,7 +61,7 @@
         <dt><%= ts("Words:") %></dt>
         <dd><%= number_with_delimiter(@series.visible_word_count) %></dd>
         <dt><%= ts("Works:") %></dt>
-        <dd><%= @series.visible_work_count %></dd>
+        <dd><%= number_with_delimiter(@series.visible_work_count) %></dd>
         <dt><%= ts("Complete:") %></dt>
         <dd><%= @series.complete? ? ts("Yes") : ts("No") %></dd>
         <% if (bookmark_count = @series.bookmarks.is_public.count) > 0 %>
--- a/app/views/series/show.html.erb
+++ b/app/views/series/show.html.erb
@@ -66,7 +66,7 @@
         <dd><%= @series.complete? ? ts("Yes") : ts("No") %></dd>
         <% if (bookmark_count = @series.bookmarks.is_public.count) > 0 %>
           <dt><%= ts("Bookmarks:") %></dt>
-          <dd><%= link_to_bookmarkable_bookmarks(@series, bookmark_count.to_s) %></dd>
+          <dd><%= link_to_bookmarkable_bookmarks(@series, number_with_delimiter(bookmark_count)) %></dd>
         <% end %>
       </dl>
     </dd>
--- a/features/other_b/series.feature
+++ b/features/other_b/series.feature
@@ -240,6 +240,15 @@ Feature: Create and Edit Series
     Then I should see "Part 2 of the Ponies series" within "dd.series"
       And I should see "Part 1 of the Black Beauty series" within "dd.series"
 
+  Scenario: A series's metadata is visible when viewing the series
+    Given I am logged in as a random user
+      And I post the work "Story" as part of a series "Excellent Series"
+      And I bookmark the series "Excellent Series"
+    When I view the series "Excellent Series"
+    Then I should see "Words: 6" within ".series.meta"
+      And I should see "Bookmarks: 1" within ".series.meta"
+      And I should see "Works: 1" within ".series.meta"
+
   Scenario: When editing a series, the title field should not escape HTML
     Given I am logged in as "whoever"
       And I post the work "whatever" as part of a series "What a title! :< :& :>"


sarken:
Hi, irrationalpie!

Thank you so much for this pull request. Someone will be along to review it soon and help with your questions.

I've updated the Jira issue status to In Review, so no one will mistakenly create a duplicate pull request. If you'd like the ability to comment on, assign, and transition issues in the future, you're welcome to create a Jira account! You can just reply here with the account name and we'll set up the permissions for you. (It makes things a bit easier for us on the organizational side if the Full Name on your Jira account either closely matches the name you'd like us to credit in the release notes or includes it in parentheses, e.g. "Nickname (PREFERRED NAME).")

irrationalpie7:
I made a jira account as "irrationalpie" just now. Let me know if you need more info to find me!

sarken:
Found you! 😄 I've assigned the issue to you and set up your permissions, so you should be all set on the Jira front.

>Note: it was unclear to me whether this issue also referred to number of works in a series, so I left that alone, but I can add that in if it should be included!

If you could include that, that would be great, thanks!

>I suspect I must have failed to set up the test database correctly

I don't use Docker, so I'm afraid you'll need someone else to tell you the best way to fix it, but I can confirm your suspicions are correct and the test database isn't quite up to date. The error you're getting most likely means that recent migrations weren't run on the test database. (It's missing the column added by [20220212222625_add_renamed_at_to_users.rb](https://github.com/otwcode/otwarchive/blob/df88c2f52872b3cbddaceac54e773274e461cb0e/db/migrate/20220212222625_add_renamed_at_to_users.rb), which means it's probably also missing the table from the migration after that as well.)

irrationalpie7:
I got the tests running locally!

I will try to find the right place to update the documentation in a bit, but just as a note to myself for later, I found a line in the non-docker documentation (`bundle exec rake db:migrate`) and changed it to `docker exec -e RAILS_ENV=test otwarchive_web_1 bundle exec rake db:migrate` (prepending the command with `docker exec [...] otwarchive_web_1` makes it run in the docker container that is running my local archive, and inserting `-e RAILS_ENV=test` makes it apply to the test database) and got a bunch of output that included the following, so I suspect that command was the missing piece:
```
== 20220212222625 AddRenamedAtToUsers: migrating ==============================
-- add_column(:users, :renamed_at, :datetime, {:null=>true})
   -> 0.0061s
== 20220212222625 AddRenamedAtToUsers: migrated (0.0064s) =====================
```

Also the series tests now pass locally (command `docker exec -e RAILS_ENV=test otwarchive_web_1 bundle exec rspec spec/models/series_spec.rb`) which is another good sign :joy: 

I also ran the `db:migrate` command without `-e RAILS_ENV=test` before I tried it with, but that time there wasn't any output, so I'm not sure whether that did anything.

irrationalpie7:
I've wrapped the work counts with `number_with_delimiter` as well now! :tada:

I'm still not sure if/where there are existing tests for these pages, so if you know and can share that would be great! In the meantime I'll continue to poke around at the codebase/docs to try to learn more about it

irrationalpie7:
After further investigation into why my tests initially failed locally, it turns out that `script/docker/init.sh` should have run the `db:migrate` thing when it ran `script/reset_database.sh` for `RAILS_ENV=development` and `RAILS_ENV=test`. I no longer have the logs from running that the first time so I'm not sure why it didn't work then, but I reran the reset database portions of `script/docker/init.sh` and it doesn't seem to have failed this time

sarken:
Yay, I'm glad the tests are working! It sounds like we can probably write this off as a random hiccup, but I'm going to ping @zz9pzza and @redsummernight just so they're aware it happened and can take a look a Docker if they think it might be an actual issue.

In general, series tests are buried in `features/other_b`, but you can also find some scattered around by searching the features for terms like `view the series`. 

In terms of tests for this pull request, we're making use of a Rails helper here, so we don't need to confirm that the actual delimiter is working. However, it would probably be good to make sure we don't have a typo or something that causes the page to 500. 

This bookmark test would be erroring if the changes to the series module weren't right (the module is used on the page that displays after you create the bookmark, among other places): https://github.com/otwcode/otwarchive/blob/742827a740dee71d3abcfd945822b1fe2fd60973/features/bookmarks/bookmark_create.feature#L334-L341

But I don't see one that checks the series show page when the series has a bookmark, so it wouldn't hurt to add a test to `features/other_b/series.feature` that does a general check of the stats metadata (i.e., create a series, bookmark it, and then check that when you view the series, you see Words: X, Bookmarks: Y, Works: Z).

irrationalpie7:
I'll take a look at the tests!

---


Regarding the docker issue, I fought with it a bit more today so I can give a bit more context:
1. The first issue I had with `script/docker/init.sh` was either that the custom root password I chose for the database had some special characters the script didn't like, or the following:
```
ERROR: for otwarchive_web_1  UnixHTTPConnectionPool(host='localhost', port=None): Read timed out. (read timeout=60)

ERROR: for web  UnixHTTPConnectionPool(host='localhost', port=None): Read timed out. (read timeout=60)
ERROR: An HTTP request took too long to complete. Retry with --verbose to obtain debug information.
If you encounter this issue regularly because of slow network conditions, consider setting COMPOSE_HTTP_TIMEOUT to a higher value (current value: 60).
```
2. I fixed the second issue by running the following lines, found on stack exchange somewhere:
```
export DOCKER_CLIENT_TIMEOUT=120
export COMPOSE_HTTP_TIMEOUT=120
```
3. At that point, I was getting a bunch of `Mysql2::Error::ConnectionError: Access denied for user 'root'@'172.23.0.6' (using password: YES)` errors when I tried to either rerun the `init.sh` script or just do `docker-compose up -d web`. It took me a while to realize that fixing the password in the two files mentioned in the wiki (`config/docker/database.yml` and `docker-compose.yml`) was insufficient, since `config/docker/database.yml` is only copied to `config/database.yml` if it doesn't exist yet.
4. One of the commands I ended up running before I could finally get `init.sh` to work was `docker-compose down -v`. Today I was trying to change the password back to the default and didn't know what I was doing, and this line ended up being load-bearing. After that I was able to re-run `script/docker/init.sh` and both see the local archive in the browser, and have local spec tests pass.

Let me know if there's a better place to continue this conversation if needed!

irrationalpie7:
I've added a test! :tada: Now, is it a good test? Idk, feedback/tips much appreciated :joy: 

I read through some of the docs but it wasn't totally clear to me how to make sure things I need to exist exist when testing. It seems odd to pretend to post a work in a test that's just trying to view a series with a pre-defined number of words, works, and bookmarks. And should I be explicitly setting the work content so anyone looking at the test can know where the word count "6" came from?

Also do you have any tips or tricks for figuring out which step file a particular step in a feature file is defined in? Like is there an IDE that lets me navigate from a line in a `*.feature` file to the definition for that step in a `*_steps.rb` file? I'm currently using vscode but about all I've figured out for it to do for me so far is syntax highlighting and formatting.

sarken:
The test looks good to me! Thanks for adding it. I'll mark this Ready to Merge now. :tada: 

>I read through some of the docs but it wasn't totally clear to me how to make sure things I need to exist exist when testing. It seems odd to pretend to post a work in a test that's just trying to view a series with a pre-defined number of words, works, and bookmarks.

We do this in the majority of our feature tests, so it's perfectly fine by our standards! However, you're right that it's not super efficient, so other projects might have different preferences. 

Another way to create data would be to write a new step definition that uses FactoryBot, but the downside to that approach, imo, is you end up with a lot of step definitions that may only get used once. We do use it sometimes, though, like when we want to create a whole bunch of data: https://github.com/otwcode/otwarchive/blob/742827a740dee71d3abcfd945822b1fe2fd60973/features/step_definitions/bookmark_steps.rb#L31-L103

>And should I be explicitly setting the work content so anyone looking at the test can know where the word count "6" came from?

Relying on the default text also means the test will start to fail if someone changes that, so it's not a _bad_ idea to set the text explicitly, but it's also not super important in the context of this particular test, since we're just checking that the word count is present. If we wanted to make sure the word count was _correct_, it would be more important to know the text... but it would be better to test that with Rspec tests for the `visible_word_count` method.

>Also do you have any tips or tricks for figuring out which step file a particular step in a feature file is defined in?

I just use a combination of my own memory and Sublime Text's Find in Files function to search the step definition folder. However, I did a quick Google, and [this StackOverflow answer](https://stackoverflow.com/a/45281180) might help!

sarken (approved):

redsummernight:
> 3. At that point, I was getting a bunch of `Mysql2::Error::ConnectionError: Access denied for user 'root'@'172.23.0.6' (using password: YES)` errors when I tried to either rerun the `init.sh` script or just do `docker-compose up -d web`. It took me a while to realize that fixing the password in the two files mentioned in the wiki (`config/docker/database.yml` and `docker-compose.yml`) was insufficient, since `config/docker/database.yml` is only copied to `config/database.yml` if it doesn't exist yet.

Thanks for looking into this issue! I made [AO3-6358](https://otwarchive.atlassian.net/browse/AO3-6358) and you're welcomed to claim it.

PR #4256, opened 2022-06-07
AO3-4659 Tweaks to block user and collection participant forms

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4659

## Purpose

* Makes the form prettier on smaller screens, e.g. tablets and phones -- the button and the text input were overlapping a bit because the min-width on the text input was too big for the space, so I've put them on separate lines to give the text input enough room.
* Reorders the bullet points on the block user form because we wanted the bit about not hiding works to be at the top.

## Testing Instructions

On a phone or a tablet in portrait mode (or a browser with the font size cranked up and the window width narrowed), check the collection participant and blocked users pages and make sure the forms are decent looking and functional.

Also check the order of the bullet points on the block user form; the one about hiding works should be at the top.

### DIFF
--- a/app/views/blocked/users/index.html.erb
+++ b/app/views/blocked/users/index.html.erb
@@ -11,9 +11,9 @@
   <p><%= t(".will_not.intro") %></p>
 
   <ul>
+    <li><%= t(".will_not.hide_works") %></li>
     <li><%= t(".will_not.comments_on_works") %></li>
     <li><%= t(".will_not.comments_elsewhere") %></li>
-    <li><%= t(".will_not.hide_works") %></li>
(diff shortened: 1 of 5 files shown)


app/views/blocked/users/index.html.erb
```diff
@@ -11,9 +11,9 @@
   <p><%= t(".will_not.intro") %></p>
 
   <ul>
+    <li><%= t(".will_not.hide_works") %></li>
```
redsummernight, line 14: Should we do this for the confirm block page as well?

redsummernight (approved):

PR #4262, opened 2022-06-12
AO3-5709 Remove factory_girl references from download specs

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5709

## Purpose

#3609 predates #3615, so it's factory_bot now. Download specs are currently broken.

## Testing Instructions

None.

PR #4263, opened 2022-06-12
AO3-6192 Turn off comments on news posts after 2 weeks

hunternet93:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6192

## Purpose

This PR prevents comments on admin posts after an archive-configurable period has passed.

The archive configuration setting `ADMIN_POST_COMMENTING_EXPIRATION_DAYS` controls the number of days after which commenting will be prevented. The setting's default value is `14`. A value of `0` will disable commenting expiration and allow commenting indefinitely.

#### Test that commenting on posts within the expiration period is allowed:
* As an admin user, create a new admin post
* As a guest or regular user, attempt to comment on the new admin post
* You should be able to comment on the post normally

#### Test that commenting on posts beyond the expiration period is not allowed:
* As an admin user, create a new admin post
* Use the Rails console to set the admin post's creation date to before the expiration period:
    ```ruby
    ap_id = 1 # Set to the new admin post's ID. This can be found at the end of the admin post's URL.
    ap = AdminPost.find(ap_id)
    ap.created_at = Time.now - 14.days
    ap.save
    ```
* Run the new Rake task with `rake admin_post:expire_commenting`
* As a guest or regular user, attempt to comment on the new admin post
* You should see a message stating that commenting is no longer allowed

#### Test that commenting expiration is disabled when the new setting is set to 0
* Set the `ADMIN_POST_COMMENTING_EXPIRATION_DAYS` setting to 0 in your test instance's configuration
* As a guest or regular user, attempt to comment on the admin post with a creation date any amount of time in the past.
* You should be able to comment on the post normally, regardless of the post's age

## Note

This is my first time working on the Archive, so please let me know if I missed any steps submitting the PR!

## Credit

Hunter Ada Smith (she/her)

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -49,6 +49,11 @@ every 1.day, at: '7:40 am' do
   rake "work:purge_old_drafts"
 end
 
+# Expire commenting for old admin posts
+every 1.day, at: "8:10 am" do
+  rake "admin_post:expire_commenting"
+end
+
 # Send kudos notifications
 every 1.day, at: '10:00 am' do
   rake "notifications:deliver_kudos"
--- a/features/step_definitions/admin_steps.rb
+++ b/features/step_definitions/admin_steps.rb
@@ -108,6 +108,10 @@ Given "the admin post {string}" do |title|
   FactoryBot.create(:admin_post, title: title)
 end
 
+Given "an admin post with expired commenting" do
+  FactoryBot.create(:admin_post, created_at: Time.current - ArchiveConfig.ADMIN_POST_COMMENTING_EXPIRATION_DAYS.days)
+end
+
 Given /^the fannish next of kin "([^\"]*)" for the user "([^\"]*)"$/ do |kin, user|
   step %{the user "#{kin}" exists and is activated}
   step %{the user "#{user}" exists and is activated}
--- a/config/config.yml
+++ b/config/config.yml
@@ -265,6 +265,10 @@ ANONYMOUS_THRESHOLD_COUNT: 10
 # whether a comment should go back into moderation after editing or not
 COMMENT_MODERATION_THRESHOLD: 10
 
+# This is the number of days since admin post creation until commenting on an admin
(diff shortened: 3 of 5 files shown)


hunternet93:
The test failure appears to be related to #4262, I'll update my branch once that PR gets merged.

sarken:
Hi, Hunter!

Thank you so much for this pull request. Someone will be along to review it once it's no longer a WIP (unless there's something you'd like help with now -- if so, please let us know).

sarken (changes requested):
While this would work, it's not quite what we had in mind. We want an approach that will actually change the `comment_permissions` of a post (and its translations) in the database, e.g., a daily task that checks and updates admin posts, much like the task that deletes draft works. Sorry for not being clearer in the issue!

features/step_definitions/admin_steps.rb
```diff
@@ -108,6 +108,10 @@
   FactoryBot.create(:admin_post, title: title)
 end
 
+Given "an admin post with expired commenting" do
```
redsummernight, line 111: Instead of adding this new step, you can use the existing steps for creating posts, then use [cucumber-timecop steps](https://gitlab.com/zedtux/cucumber-timecop/-/blob/master/lib/cucumber/timecop.rb) to control time, e.g. `it is currently 14 days from now`.

config/schedule.rb
```diff
@@ -49,6 +49,11 @@
   rake "work:purge_old_drafts"
 end
 
+# Expire commenting for old admin posts
+every 1.day, at: "8:10 am" do
```
redsummernight, line 53: We're [moving away from the whenever gem](https://otwarchive.atlassian.net/browse/AO3-5794) for scheduling. Can you add this to resque_schedule.yml instead?

#4280 is an example of how to do it. You can skip adding a rake as well.

hunternet93:
Closing this PR since it looks like this wasn't quite the right approach.

PR #4265, opened 2022-06-12
AO3-5522 Limit number of password resets

hunternet93:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5522

## Purpose

This adds a limit to the number of password resets that can be requested within a period of time. If the limit is reached, then reset requests will be prevented until the cooldown expires.

A new `users` table column is added: 
* `resets_requested`: The number of resets that have been requested. Integer, non-nullable, indexed, defaults to 0.

Two new archive configuration settings are added:
* `PASSWORD_RESET_LIMIT`: The limit to the number of resets within the `PASSWORD_RESET_COOLDOWN_HOURS` period. Defaults to 3.
* `PASSWORD_RESET_COOLDOWN_HOURS`: The period of time in which the reset request limit will be enforced. Defaults to 12.

#### Test that password resets can be requested
* Open the login page (or login drop-down form)
* Click the "reset password" link
* Enter a valid, non-protected username or email address
* You should see a successful response and receive a password reset email

#### Test that password reset requests beyond the limit cannot be requested
* Follow the above instructions two more times
  * (or three times if it has been more than 12 hours since the last reset request)
* Each attempt should be successfull
* Follow the above instructions a fourth time
* You should see a response indicating that you cannot reset your password at this time. Response should indicate the date and time after which password resetting will again be available
* You should not receive a password reset email

#### Test that password resets can again be requested after waiting out the cooldown period
* Wait until after the time specified in the password reset error message
* Attempt to reset your password
* You should see a successful response and receive a password reset email

## Credit

Hunter Ada Smith (she/her)


hunternet93:
Test failure appears to be a transient issue, the test runs fine locally and it isn't related to my changes.

app/controllers/users/passwords_controller.rb
```diff
@@ -6,12 +6,27 @@ class Users::PasswordsController < Devise::PasswordsController
...
+      flash[:error] = t(".reset_cooldown", reset_available_time: available_time).html_safe
+      redirect_to root_path and return
     end
+
     super do |user|
-      flash.now[:notice] = ts("We couldn't find an account with that email address or username. Please try again?") if user.nil? || user.new_record?
+      if user.nil? || user.new_record?
+        flash.now[:notice] = ts("We couldn't find an account with that email address or username. Please try again?")
+      else
+        user.update_password_resets_requested
```
brianjaustin, line 27: Would it make sense to shift this into the `User` model, something like the following?
```ruby
after_update :update_password_resets_requests, if: :will_save_change_to_password?
```

app/models/user.rb
```diff
@@ -416,6 +416,22 @@ def is_protected_user?
     has_role?(:protected_user)
   end
 
+  def password_resets_limit_reached?
```
brianjaustin, line 419: I think the ticket mentions something about pulling this out into a [`concern`](https://api.rubyonrails.org/classes/ActiveSupport/Concern.html); I could see that being useful if `Admin` also needs this behaviour at some point, but I don't know if that's planned at the moment

brianjaustin (commented):
Not necessarily a formal review, just a few thoughts I had

db/migrate/20220612064434_add_resets_requested_to_users.rb
```diff
@@ -0,0 +1,6 @@
+class AddResetsRequestedToUsers < ActiveRecord::Migration[6.0]
```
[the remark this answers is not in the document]
zz9pzza: I agree with Brian this will have to be done online ( and we will run it before the code is deployed ).

zz9pzza:
This has merge conflicts and needs the migration changing at least. Added Action needed.

db/migrate/20220612064434_add_resets_requested_to_users.rb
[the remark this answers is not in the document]
hunternet93: I added `pt-online-schema-change` outputs to the migration for staging and production environments, though I'm not entirely sure I did it correctly.

app/controllers/users/passwords_controller.rb
hunternet93: I don't believe so, since `resets_requested` is updated when the reset email is sent rather than when the password is changed.

brianjaustin: Ah, yes, I think I got those 2 mixed together in my head. Feel free to mark this thread as resolved!

app/models/concerns/password_resets_limitable.rb
```diff
@@ -0,0 +1,28 @@
...
+      self.resets_requested >= ArchiveConfig.PASSWORD_RESET_LIMIT && self.last_reset_within_cooldown?
+    end
+
+    def password_resets_available_time
+      self.reset_password_sent_at + ArchiveConfig.PASSWORD_RESET_COOLDOWN_HOURS.hours
+    end
+
+    def update_password_resets_requested
+      if self.resets_requested.positive? && !self.last_reset_within_cooldown?
+        self.resets_requested = 1
```
brianjaustin, line 15: Do you think it would make sense to do something like
```rb
self.update!(resets_requested: 1)
```
(and similarly for below as well)? That way, we won't lose the change in case someone forgets to call `save` later

app/controllers/users/passwords_controller.rb
```diff
@@ -6,12 +6,27 @@ class Users::PasswordsController < Devise::PasswordsController
...
+        user.save
```
brianjaustin, line 28: Hmm, I'm not sure if this should be `user.save` or `user.save!`. On the one hand, it could be a good idea to make sure the reset counter is incremented before sending out the email, but on the other it seems a bit draconian to not send the email if something goes wrong saving... @zz9pzza (if you have time) what do you think?

sarken:
Updated in #4626 and merged there, so I'm closing this one. Thank you!

PR #4266, opened 2022-06-17
AO3-5734 Stop truncating and sanitizing page titles of HTML downloads

potpotkettle:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5734

## Purpose

To stop doing truncation and cleaning of the page title shown in the downloaded HTML file which seem unnecessary.

## Testing Instructions

(Using Docker) Go to http://localhost:3000/works/107 and choose Download > HTML. Open the downloaded HTML file. The title (as shown in a browser tab) should read "Addding tags for stats purposes" instead of "Addding tags for stats"

Note: Title lines shown inside the screen (in bold text) should be fine before this set of changes and should remain that way. Also, the file name should be in ASCII before and after.

EDIT: I actually don't know if work IDs (107 here) of test data are same across installations.

## Credit

Potpotkettle (they/them)

### DIFF
--- a/app/models/download.rb
+++ b/app/models/download.rb
@@ -90,16 +90,13 @@ class Download
     @tmpdir
   end
 
-  # Mimic roughly the default way of producing HTML page titles
-  # @see WorksController#show
-  # @see ApplicationController#get_page_title
   def page_title
     fandom = if work.fandoms.size > 3
                "Multifandom"
              elsif work.fandoms.empty?
                "No fandom specified"
              else
-               work.fandoms[0].name
+               work.fandoms.string
              end
     [work.title, authors, fandom].join(" - ")
   end
--- a/app/models/download.rb
+++ b/app/models/download.rb
@@ -90,28 +90,26 @@ class Download
     @tmpdir
   end
 
-  # Utility methods which clean up work data for use in downloads
-
-  def fandoms
-    string = work.fandoms.size > 3 ? "Multifandom" : work.fandoms.string
-    clean(string)
+  # Mimic roughly the default way of producing HTML page titles
+  # @see WorksController#show
+  # @see ApplicationController#get_page_title
+  def page_title
+    fandom = if work.fandoms.size > 3
+               "Multifandom"
+             elsif work.fandoms.empty?
+               "No fandom specified"
+             else
+               work.fandoms[0].name
+             end
+    [work.title, authors, fandom].join(" - ")
   end
 
   def authors
-    author_names.join(', ').to_ascii
+    author_names.join(", ")
   end
 
   def author_names
     work.anonymous? ? ["Anonymous"] : work.pseuds.sort.map(&:byline)
-  end
-
-  # need the next two to be filesystem safe and not overly long
-  def file_authors
-    clean(author_names.join('-'))
-  end
-
-  def page_title
-    [file_name, file_authors, fandoms].join(" - ")
   end
 
   def chapters
--- a/app/models/download.rb
+++ b/app/models/download.rb
@@ -96,7 +96,7 @@ class Download
              elsif work.fandoms.empty?
                "No fandom specified"
              else
-               work.fandoms.string
+               work.fandom_string
              end
     [work.title, authors, fandom].join(" - ")
   end
--- a/spec/models/download_spec.rb
+++ b/spec/models/download_spec.rb
@@ -104,15 +104,21 @@ describe Download do
     let(:work) { create(:work, fandoms: [fandom1, fandom2], title: "Foo bar") }
     let(:subject) { Download.new(work) }
 
-    it "includes first fandom name" do
+    it "includes fandom names" do
       expect(subject.page_title).to include(fandom1.name)
-      expect(subject.page_title).not_to include(fandom2.name)
+      expect(subject.page_title).to include(fandom2.name)
     end
 
     it "leaves emojis alone" do
       work.title = "emoji 🥳 is 🚀 awesome"
 
       expect(subject.page_title).to include("emoji 🥳 is 🚀 awesome")
+    end
+
+    it "leaves long titles alone" do
+      work.title = "no title is too long to print nor to read"
+
+      expect(subject.page_title).to include("no title is too long to print nor to read")
     end
 
     context "for a work with multiple authors" do
--- a/spec/models/download_spec.rb
+++ b/spec/models/download_spec.rb
@@ -97,10 +97,10 @@ describe Download do
   end
 
   describe "page_title" do
-    let(:fandom1) { create(:canonical_fandom) }
-    let(:fandom2) { create(:canonical_fandom) }
-    let(:pseud1) { create(:pseud, name: "First", user: create(:user, login: "Zeroth")) }
-    let(:pseud2) { create(:pseud, name: "Second", user: create(:user)) }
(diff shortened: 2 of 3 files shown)


potpotkettle:
I first attempted to extract code from ``ApplicationController#get_page_title`` and ``WorksController#show`` to create a new common utility function under ``lib``, and include it from those files and Download, but eventually decided against that because the end result would seem more complex that way (mostly because user preferences can apply to `ApplicationController` and `WorksController`, but they should not to Download, at least that's what I assume).

I took the liberty to remove apparently unnecessary parts from the related code, namely ``file_author`` and associated tests. It seems that it was used as part of URLs in earlier versions (https://github.com/otwcode/otwarchive/commit/3463d1a8f9b349bd3f827098d43c10c786777f9b#diff-959bc9abc46a55332bb64d5155a79323afa75a50ec1a2137ddd22d926f62c6c5) but probably not any more as far as I can see.

(If I made wrong assumptions, I apologize for my ignorance. I'd be happy to be corrected and I'll adjust the PR accordingly.)

app/models/download.rb
```diff
@@ -90,30 +90,28 @@ def dir
...
   end
 
-  # Utility methods which clean up work data for use in downloads
-
-  def fandoms
-    string = work.fandoms.size > 3 ? "Multifandom" : work.fandoms.string
-    clean(string)
+  # Mimic roughly the default way of producing HTML page titles
+  # @see WorksController#show
+  # @see ApplicationController#get_page_title
```
redsummernight, line 95: I think you can remove the comment here. Mimicking `ApplicationController#get_page_title` is a starting point but it's not important for us to keep the two implementations totally in sync going forward.

spec/models/download_spec.rb
```diff
@@ -99,35 +99,53 @@
   describe "authors" do
     let(:work) { Work.new }
     let(:subject) { Download.new(work) }
+    let(:pseud1) { create(:pseud, name: "First", user: create(:user)) }
+    let(:pseud2) { create(:pseud, name: "Second", user: create(:user)) }
 
     it "joins the author names separated by a comma and a space" do
```
redsummernight, line 105: You can remove this test, since it's checking the same thing as "for a work with multiple authors" below.

spec/models/download_spec.rb
```diff
@@ -99,35 +99,53 @@
...
+      allow(pseud1).to receive(:byline).and_return("First (Zeroth)")
+      allow(pseud2).to receive(:byline).and_return("Second")
+      allow(work).to receive(:pseuds).and_return([pseud1, pseud2])
 
       expect(subject.authors).to eq("First (Zeroth), Second")
     end
 
-    it "transliterates non-ASCII characters" do
-      allow(subject).to receive(:author_names).and_return(["我哥好像被奇怪的人盯上了怎么破"])
+    it "leaves Chinese characters alone" do
```
redsummernight, line 113: Can you move this under `describe "page_title"` and remove `describe "authors"` entirely?

spec/models/download_spec.rb
```diff
@@ -99,35 +99,53 @@
...
+    let(:fandom2) { create(:canonical_fandom) }
+    let(:pseud1) { create(:pseud, name: "First", user: create(:user)) }
+    let(:pseud2) { create(:pseud, name: "Second", user: create(:user)) }
+    let(:work) { create(:work, fandoms: [fandom1, fandom2]) }
     let(:subject) { Download.new(work) }
 
-    it "truncates if too long" do
-      allow(subject).to receive(:author_names).and_return(["ComplexAuthor (ComplexUser)", "SimpleAuthor"])
+    it "includes first fandom name" do
+      work.title = "Foo bar"
```
redsummernight, line 130: You can set a title in `let(:work)`.

spec/models/download_spec.rb
```diff
@@ -99,35 +99,53 @@
...
+      expect(subject.authors).to eq("我哥好像被奇怪的人盯上了怎么破")
     end
   end
 
-  describe "file_authors" do
-    let(:work) { Work.new }
+  describe "page_title" do
+    let(:fandom1) { create(:canonical_fandom) }
+    let(:fandom2) { create(:canonical_fandom) }
+    let(:pseud1) { create(:pseud, name: "First", user: create(:user)) }
```
redsummernight, line 124: You can set the user name here, so you don't need to fake the byline below.

app/models/download.rb
potpotkettle: I'm not sure if I understand. It is not important because we should eventually have one implementation instead of two implementations? Or because we don't exactly care how the precise formatting of the titles of downloads and online versions may differ, as long as information is not lost?

I think in the former case, some comment would be useful. Maybe a "TODO" comment to ask merging the two?

redsummernight: It's ok if they differ, yeah. For the work page we may want to keep the title more concise. For downloads it's fine if we keep more information. On that note, let's keep using `work.fandoms.string` instead of just the first one (`work.fandoms[0].name`) for the else case.

potpotkettle: That makes sense, thanks. I'll change the part of the code accordingly.

potpotkettle:
I believe all the points raised have been addressed by the two last changes (Jul 5 & Jul 7).

spec/models/download_spec.rb
```diff
@@ -96,39 +96,44 @@
     end
   end
 
-  describe "authors" do
-    let(:work) { Work.new }
+  describe "page_title" do
+    let(:fandom1) { create(:canonical_fandom) }
```
redsummernight, line 100: Most of these tests don't actually require works, pseuds, fandoms, etc. to be saved to the database, so you should be able to write them using [`build`](https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#build-strategies) and make them faster, e.g.

https://github.com/otwcode/otwarchive/blob/14fbeb6ae13f9637cfdd3c4f586620d8863ff32a/spec/models/download_spec.rb#L60-L66

Otherwise this looks good to me.

potpotkettle: Thanks for the comment, and that's true. I've managed to use ``build`` in most of them, but not  ``let(:work)``. (To be honest, I only understand the Work class on a very shallow level, and I'm not sure which part of it around saving/validation is the issue.)

redsummernight: I figured this out. If we use `work.fandoms.string`, we call this method:

https://github.com/otwcode/otwarchive/blob/015ebc1d01be7c3962d59082b5f9515ec6e7c599/app/models/tag.rb#L459-L463

which will force a database query.

You can switch from `work.fandoms.string` to `work.fandom_string` instead, which will use `.fandoms` on an unsaved work without hitting the database:

https://github.com/otwcode/otwarchive/blob/015ebc1d01be7c3962d59082b5f9515ec6e7c599/app/models/concerns/taggable.rb#L27-L29

Then you should be able to avoid saving the work:
```rb
let(:work) { build(:work, fandoms: [fandom1, fandom2], title: "Foo bar") }
```

potpotkettle: Thanks again! This has been very illuminating. I'll revise it again (hopefully for the last time).

redsummernight (approved):
Looks good, thanks!

PR #4267, opened 2022-06-20
AO3-6318 Include series in the noindex privacy preference

hunternet93:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6318

## Purpose

This makes the "Hide my work from search engines when possible" user privacy preference take effect for series in addition to works. The `<meta>` tags that indicates that search engines should not index a page is now included on the series pages of users with this preference enabled.

#### Test that the works and series of a user hidden from search includes the noindex tag
* Log in as a user
* In that user's preferences page, enable the "Hide my work from search engines when possible" and save the change
* Ensure that the user has at least one work and series for testing
* Log out from the user, or open a new incognito/private browser window for the following steps
* Open any of the user's series
* Use your browser's dev tools to verify the presence of the noindex meta tags. These look like:
    ```html
    <meta name="robots" content="noindex" />
    <meta name="googlebot" content="noindex" />
    ```
    For easier checking, use this line of Javascript. A response of robot emojis means the tags are present, the no entry emojis mean the tags aren't present
    ```javascript
    console.log(document.querySelector('meta[name="robots"]') ? '🤖🤖🤖🤖🤖' : '🚫🚫🚫🚫🚫');
    ```
* Open any of the user's works and perform the above step again to verify the noindex meta tags are present here as well

#### Test that the works and series of a user not hidden from search does not include the noindex tag
* Perform the same steps as above, but instead use a user with the "Hide my work from search engines when possible" privacy preference disabled
* The noindex meta tags should not be present on any of that user's works or series

## Credit

Hunter Ada Smith (she/her)


hunternet93:
Test failure appears to be transient, Cucumber tests passed for me locally after running them a couple times. Can someone please retry it?

app/views/layouts/application.html.erb
```diff
@@ -11,7 +11,10 @@
     <meta name="distribution" content="GLOBAL" />
     <meta name="classification" content="transformative works" />
     <meta name="author" content="Organization for Transformative Works" />
-    <% if @user && @user.try(:preference).try(:minimize_search_engines?) || @work.is_a?(Work) && @work.users.all? {|u| u.try(:preference).try(:minimize_search_engines?)} %>
+    <% if @user && @user.try(:preference).try(:minimize_search_engines?) ||
```
sarken, line 14: The code looks good! It's getting a bit long, though. Maybe we could put the logic in a helper method in application_helper.rb?

sarken (changes requested):
Thank you for the pull request! I love the handy bit of JavaScript to help with manual testing. 😍 

I do think we should add an automated test that checks the meta tags for series (while it would be nice to have tests for work and user pages, they're a bit outside the issue's scope, so it's up to you whether you'd like to add them). It should be possible to do this with a Cucumber test if you create a step definition that [uses Capybara to check the page head for meta tags](https://codehunter.cc/a/ruby-on-rails/capybara-not-finding-meta-tags).

app/helpers/application_helper.rb
```diff
@@ -650,4 +650,10 @@ def permit_all_except(params, blocked_keys)
       params
     end
   end
+
+  def disallow_robots?(user, creation)
```
sarken, line 654: I don't think we need two arguments here, since we're never going to have both a user and a creation at the same time. We can also use Ruby's safe operator (`&.`) instead of Rails's `try` method:

```ruby
  def disallow_robots?(item)
    return unless item

    if item.is_a?(User)
      item&.preference&.minimize_search_engines?
    else
      item.users.all? { |u| u&.preference&.minimize_search_engines? }
    end
  end
 ```
 
 Then in the view, we'd do:
 ```ruby
<% item = @user || @work || @series %>
<% if disallow_robots?(item) %>
```

features/step_definitions/user_steps.rb
```diff
@@ -177,6 +177,11 @@
   user.roles = [Role.find_or_create_by(name: "protected_user")]
 end
 
+Given "the user {string} is hidden from search engines" do |login|
```
sarken, line 180: This should probably go in `features/step_definitions/preferences_steps.rb` instead. The tests themselves could probably also be moved to `features/other_a/preferences_edit_more.feature`.

features/users/hide_from_search_engines.feature
```diff
@@ -0,0 +1,49 @@
+Feature: Hiding users and their content from search engines
+  Scenario: Hidden users' user pages are disallowed for search engine indexing
+    Given I have no users
```
sarken, line 3: Just wondering -- why do we need to start with no users?

features/step_definitions/generic_steps.rb
```diff
@@ -92,25 +92,47 @@
   step %{I should see "success"}
 end
 
-def assure_xpath_present(tag, attribute, value, selector)
+def xpath_for_tag_with_attributes(tag, attributes)
```
sarken, line 95: If we anticipated frequently needing to check multiple attributes, this would probably be the way to go, but I think we're getting a bit ahead of ourselves here. 

Rather than modify all these existing steps, we can define one new step to check for these meta tags (and a corresponding one to check that the tags aren't present):

```gherkin
Then "the page should be hidden from search engines" do
  page.should have_xpath('//head/meta[@name="robots"][@content="noindex"]', visible: false)
  page.should have_xpath('//head/meta[@name="googlebot"][@content="noindex"]', visible: false)
end
```

redsummernight: +1. Additionally, we should prefer [the expect syntax](https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers), as well as CSS selectors (more familiar than XPath), so something like:
```rb
expect(page).to have_css("meta[name=robots][content=noindex]", visible: false)
```

features/users/hide_from_search_engines.feature
hunternet93: A misunderstanding on my part, it wasn't actually needed.

features/other_a/preferences_edit_more.feature
```diff
@@ -48,3 +48,42 @@ Feature: Preferences
...
+    When I view the work "Hidden Work"
+    Then the page should be hidden from search engines
+
+  Scenario: Unhidden users' works pages are allowed for search engine indexing
+    Given the user "unhidden" exists and is activated
+      And the work "Unhidden Work" by "unhidden"
+    When I view the work "Unhidden Work"
+    Then the page should not be hidden from search engines
+
+  Scenario: Hidden users' series pages are disallowed for search engine indexing
```
redsummernight, line 76: I think you can combine these 6 tests into just 2, one hidden one default, since their setup is similar.

After `I post the work "Hidden Work" as part of a series "Hidden Series"` you can check the user dashboard, the work, and the series in one go.

redsummernight (approved):
Looks good to me! Just one minor comment. I'll leave the label "Reviewed: Ready to Merge" to @sarken.

sarken (changes requested):
(Flipping this to Changes Requested/Reviewed: Needs Coder Action due to red's last comment. No rush, though!)

sarken (approved):

PR #4269, opened 2022-06-24
AO3-4355 Set language for admin posts RSS

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-4355

## Purpose

Show localized RSS feeds when clicking on the "RSS Feed" button for admin posts

## Testing Instructions

See Jira ticket.

## References

## Credit

What name and pronouns should we use to credit you in the [Archive of Our Own's Release Notes](https://archiveofourown.org/admin_posts?tag=1)? **Brian Austin (they/he)**

### DIFF
--- a/app/views/admin_posts/index.html.erb
+++ b/app/views/admin_posts/index.html.erb
@@ -11,7 +11,7 @@
       <!--subnav-->
       <ul class="navigation actions" role="navigation">
         <li><%= render 'filters' %></li>
-        <li><%= link_to_rss admin_posts_path(:rss) %></li>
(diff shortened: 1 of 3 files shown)


sarken (approved):
Looks good to me! The only note I have is that we're going to have a feed specifically for English (the one with `en` params) and a feed that is _effectively_, but not exactly, for English (the one with no params that you get if you subscribe from the admin index without first filtering by language). I can't think of a reason that would cause much of a problem, but if we run into something later, we can adjust.

PR #4270, opened 2022-06-24
AO3-4494 Remove time of day from draft deletion warning message

potpotkettle:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-4494

## Purpose

The timestamps for scheduled deletion of drafts are overly precise and not accurately tells when a draft will actually be deleted. This pull request implements the improved wording that has been decided:  "scheduled for deletion on [time of creation + 29 days]".

## Testing Instructions

Added as a comment to the Jira issue.

## Credit

Potpotkettle (they/them)


app/helpers/application_helper.rb
```diff
@@ -325,27 +325,47 @@ def link_to_remove_section(linktext, form, class_of_section_to_remove="removeme"
...
+ 
+    zone ||= (user&.is_a?(User) && user&.preference&.time_zone) ? user.preference.time_zone : Time.zone.name
     time_in_zone = time.in_time_zone(zone)
-    time_in_zone_string = time_in_zone.strftime('<abbr class="day" title="%A">%a</abbr> <span class="date">%d</span>
-                                                 <abbr class="month" title="%B">%b</abbr> <span class="year">%Y</span>
-                                                 <span class="time">%I:%M%p</span>').html_safe +
-                                          " <abbr class=\"timezone\" title=\"#{zone}\">#{time_in_zone.zone}</abbr> ".html_safe
-
-    user_time_string = "".html_safe
-    if user.is_a?(User) && user.preference.time_zone
```
potpotkettle, line 338: I think the addition of "set timezone" here has been broken, so I tried to fix it here, although it's not part of the reported issue on Jira. I commented about it in the Jira issue.

My current thought is that it seems like something that needs to be fixed eventually, and if I change a function to solve one issue anyway, I could fix something else about the same function as well. (And it's a short function.)

If reviewers don't like it, I'm happy to remove the additional part. I'd suggest leaving relevants tests as pending, and creating a new Jira issue about it, though.

spec/helpers/application_helper_spec.rb
```diff
@@ -364,4 +364,91 @@
...
+          allow(pref).to receive(:time_zone).and_return(nil)
+          allow(user).to receive(:preference).and_return(pref)
+          result = helper.time_in_zone(time, utc, user)
+
+          expect(result.html_safe?).to eq(true)
+          expect(strip_tags(result)).to eq("Fri 31 Dec 1999 04:00PM UTC (set timezone)")
+        end
+      end
+    end
+  end
```
potpotkettle, line 440: I added these tests for ``time_in_zone`` after finding out what I think is a bug of the function incidentally. I think I shouldn't try to fix ``time_in_zone`` here, but I hope (pending) tests are acceptable. A previous PR #2453 had to be reverted for changes in that function. Tests like these would help preventing that.

app/views/works/show.html.erb
```diff
@@ -5,7 +5,7 @@
 </ul>
 <% if !@work.posted? %>
   <p class="notice">
-    <%= ts("This work is a draft and has not been posted. The draft will be") %> <strong><%= ts("automatically deleted") %></strong> <%= ts("on") %> <%= time_in_zone(@work.created_at + 1.month) %>
+    <%= t(".unposted_deletion_notice", deletion_date: date_in_zone(@work.created_at + 29.days)).html_safe %>
```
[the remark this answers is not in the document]
tickinginstant: To fix the Brakeman warning, you can [rename the i18n key](https://guides.rubyonrails.org/i18n.html#using-safe-html-translations) so that it ends in `_html` or `.html`, and then you should be able to delete the call to `html_safe`.

(It might be nice to do that for the message in `works/work_module`, too.)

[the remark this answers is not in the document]
potpotkettle: Thanks! I applied your suggestion.

potpotkettle:
I've realized there is app/helpers/date_helper.rb . Would it be a good idea to move ``time_in_zone`` and ``date_in_zone`` there or at least the latter?

potpotkettle:
While resolving conflicts, I rebased out of habit elsewhere. If it messed up with the discussion log here, sorry.

spec/helpers/application_helper_spec.rb
redsummernight: Let's definitely not touch `time_in_zone` for now.

I have mixed feelings about pending tests, because we have a few that end up not being enabled for a long time, but still needs to be updated once in a while (e.g. when we need update Ruby or some test libraries). We do have a copy of the pending tests in your commit history here to revisit if necessary.

redsummernight:
> I've realized there is app/helpers/date_helper.rb . Would it be a good idea to move time_in_zone and date_in_zone there or at least the latter?

Let's move the new method there while leaving the existing one alone.

> I rebased out of habit elsewhere. If it messed up with the discussion log here, sorry.

No worries!

spec/helpers/application_helper_spec.rb
potpotkettle: I still believe that function [needs to be fixed](https://otwarchive.atlassian.net/browse/AO3-4494?focusedCommentId=358854) (as shown by the failures of these tests), but I agree that it should be dealt with separately, not here. If it is reported (or gets reported) somewhere, I'd be interested in working on it.

For now, I have removed them per suggestion.

potpotkettle:
Thanks a lot for reviewing. I have added a commit to address the review comments.

app/helpers/date_helper.rb
```diff
@@ -23,4 +23,13 @@ def date_in_user_time_zone(datetime)
     end
   end
 
+  # show date in the time zone specified
+  # note: this does *not* append timezone and does *not* reflect user preferences
+  def date_in_zone(time, zone = nil)
+    zone ||= Time.zone.name
+    return I18n.t("time.no_date_specified") if time.blank?
```
redsummernight, line 30: It might not be easy for Translation to handle cases where `no_date_specified` is filled in where a sentence expects an actual date. I think we can simply return nil if time is not provided. It's up to the views to deal with nil cases correctly.

(For drafts, we use `created_at` which should always be set, so we should be fine.)

app/controllers/chapters_controller.rb
```diff
@@ -97,7 +97,7 @@ def edit
   end
 
   def draft_flash_message(work)
-    flash[:notice] = work.posted ? ts("This is a draft chapter in a posted work. It will be kept unless the work is deleted.") : ts("This is a draft chapter in an unposted work. The work will be <strong>automatically deleted</strong> on #{view_context.time_in_zone(work.created_at + 1.month)}.").html_safe
+    flash[:notice] = work.posted ? t("chapters.draft_flash.posted_work") : t("chapters.draft_flash.unposted_work_html", deletion_date: view_context.date_in_zone(work.created_at + 29.days)).html_safe
```
redsummernight, line 100: Since the key [ends with `_html`](https://guides.rubyonrails.org/i18n.html#using-safe-html-translations), you might not need `.html_safe` here.

redsummernight (commented):
Looks good to me! Just two minor comments.

app/controllers/chapters_controller.rb
tickinginstant: Unfortunately, the code for HTML-safe translation in controllers was [first introduced in Rails 7.0](https://github.com/rails/rails/commit/fee61e3abc878e4d4930a0be0accf8e4569009a8). So we have a ways to go before we can use that.

redsummernight: Whoops, good catch, thanks!

app/helpers/date_helper.rb
potpotkettle: Just to make sure, would it be fine if it would be a bit different from `time_in_zone`? `time_in_zone` has a similar string return value for "no time specified": https://github.com/otwcode/otwarchive/blob/4d661d5cf70afb32dfa6df84419b3de6b9dd6e27/app/helpers/application_helper.rb#L307

I have no personal preference, I just tried to make `date_in_zone` similar to `time_in_zone`.

redsummernight: Yeah, it's fine if they're different. We'll leave `time_in_zone` as is but make `date_in_zone` simpler.

redsummernight (approved):
Thank you!

potpotkettle:
Is the latest check failure a false positive? I believe the key below is used.

```
Unused keys (1) | i18n-tasks v1.0.12
+--------+------------------------------+-------------------------------------------------------------------------------------------------------------------------------------------+
| Locale | Key                          | Value                                                                                                                                     |
+--------+------------------------------+-------------------------------------------------------------------------------------------------------------------------------------------+
|   en   | time.formats.date_short_html | <abbr class="day" title="%A">%a</abbr> <span class="date">%d</span> <abbr class="month" title="%B">%b</abbr> <span class="year">%Y</span> |
+--------+------------------------------+-------------------------------------------------------------------------------------------------------------------------------------------+
```

Bilka2:
> Is the latest check failure a false positive? I believe the key below is used.
> [...] time.formats.date_short_html 

Mark the key as used by adding this comment above the `I18n.l` call in [`date_in_zone`](https://github.com/otwcode/otwarchive/pull/4270/files#diff-739f6966968769aa8d095b49bc52856feafef7816fd80ed75e55a101a1625073R33):
```# i18n-tasks-use t('time.formats.date_short_html')```

potpotkettle:
@Bilka2 thanks a lot for the tip. Looks like it did the trick.

sarken:
Just a heads up that this has merge conflicts now -- but no rush!

potpotkettle:
The merge conflicts are resolved. :)

sarken:
Thank you!

sarken:
It looks like this has some merge conflicts now -- would you be able to update it in the next couple of days? If so, we can include it in this release.

PR #4272, opened 2022-06-25
AO3-6158 Remove US$ before banner goal/progress values

saltlas:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6158

## Purpose

Makes the thermometer not use currency values for progress and goal by default.

## Testing Instructions

As per Jira instructions.

## References

## Credit


designateddolt:
Spent about 4 hours trying to get a Cucumber test to work on this, but "I should see" and "I should see text with tags" both do not actually see the numerical values on the thermometer widget at all. I figure since it's a 6-character change it doesn't actually need to be tested, but if anyone wants tests for it please guide me towards how to figure out how to get Rspec to actually see the thermometer.

redsummernight:
> I assume this is because it's JavaScript.

Yeah, Cucumber tests by default browse the site without JavaScript. If you want the test to have JavaScript enabled, you need to add a `@javascript` tag on the scenario. This is [a feature of the cucumber-rails gem we use](https://github.com/cucumber/cucumber-rails/blob/5bfb4219abc50a5d1c8b2fa0a04c441bbc6e5288/features/capybara_javascript_drivers.feature).

We don't have existing tests for the thermometer, so I'm fine without new ones just for this change.

redsummernight (approved):

tickinginstant:
Does this need extra code to handle the case when you do want to specify a unit? E.g. if you choose to enter `US$100` as the goal and `US$50` as the current value, will it set the progress bar to 50%?

redsummernight:
Good catch, it does need extra. If we specify `<p>Money, money, money. We want <span class="goal">US$1,000</span> and we've gotten <span class="progress">$US900</span>.99.</p>`, then `parseFloat` will return NaN.

@sarken should we have extra markup for the units which makes it easier to parse?

sarken:
I'd like to avoid extra markup so we don't make it too complicated for the committees making the banners. If possible, I'd rather use a regex to grab just the numbers from the goal and progress spans.

public/javascripts/application.js
```diff
@@ -585,12 +585,12 @@ function thermometer() {
           stretch = true
         } else { stretch = false }
 
-        goal_amount = parseFloat(banner_goal_text.replace(/,/g, ''))
-        progress_amount = parseFloat(banner_progress_text.replace(/,/g, ''))
+        goal_amount = parseFloat(banner_goal_text.replace(/[^0-9]/g, ''))
+        progress_amount = parseFloat(banner_progress_text.replace(/[^0-9]/g, ''))
```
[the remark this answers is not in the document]
redsummernight: Can you add the trailing semicolons for the changed lines?

redsummernight, line 589: Does this handle thousand separators?

designateddolt: yep! i've tested it as shown for "<p>Members, members, members. We want <span class="goal">ssssssssssssssssssss1,000</span> and we've gotten <span class="progress">900</span>.</p>", progress  bar still shows up.
image

designateddolt: Actually it would potentially cause problems if anyone specified any decimal places such as $9.99 (would show as $999) - is that something that's likely to happen or should it be left alone? Decimal points in some languages such as Spanish are commas, also, and thousand seperators are full-stops, so I'm not sure.

redsummernight: Yeah, i18n for this banner will be annoying. But AFAIK we've always implicitly used the en-US (removing the comma), so it's fine to stick with that.

Let's update the regex to keep the decimal point though, since we do support that currently.

designateddolt:
Sorry for the delay! Should hopefully be fine now.

designateddolt:
The only thing I can find that this still can't handle is if a number with two decimal points is entered - as far as I can tell it won't break the bar, it'll just ignore anything after the second decimal place (e.g. 0.1.9 will show up as 0.1).

sarken (approved):

PR #4274, opened 2022-06-25
AO3-6160 Lazy pluralization of bookmark count on delete pseud page

designateddolt:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6160

## Purpose

Uses i18n pluralize function to pluralize "bookmark" when appropriate on delete pseud page. (Previously, the page said "bookmark(s)")

## Testing Instructions

As per Jira

## References

## Credit

### DIFF
--- a/app/views/pseuds/delete_preview.html.erb
+++ b/app/views/pseuds/delete_preview.html.erb
@@ -7,7 +7,7 @@
 
 <% if @pseud.bookmarks %>
   <%= form_tag({:action => "destroy"}, :method => :delete) do %>
-    <p>You have saved <%= @pseud.bookmarks.count %> bookmark(s) using this pseud. You can choose whether to
+    <p>You have saved <%= pluralize(@pseud.bookmarks.count, "bookmark") %> using this pseud. You can choose whether to
     delete them or transfer them to your default pseud.
     </p>
 


sarken (changes requested):
It looks like this doesn't quite fully address the issue, which says at the bottom: 

> The rest of the text for this page should be internationalized as well.

If you could finish internationalizing the page, that would be great!

designateddolt:
Very sorry about that! Didn't read the issue properly. Should be good now? Let me know if I missed anything.

app/views/pseuds/delete_preview.html.erb
```diff
@@ -1,28 +1,27 @@
-<h2 class="heading"><%=h 'Pseud deletion' %></h2>
+<h2 class="heading"><%=h ts("Pseud deletion") %></h2>
 
 <p>
-  When you delete your pseud, any works, series, or comments you have created under it will be 
-  transferred to your default pseud.
+  <%= ts("When you delete your pseud, any works, series, or comments you have created under it will be
```
redsummernight, line 4: We're moving away from using `ts`, which is [a custom setup we never quite finished](https://github.com/otwcode/otwarchive/blob/14fbeb6ae13f9637cfdd3c4f586620d8863ff32a/config/initializers/monkeypatches/translate_string.rb#L121), towards doing i18n [how Rails would expect](https://guides.rubyonrails.org/i18n.html). That means we use the `t` helper, and set up the English strings in config/locales/views/en.yml.

You can see how we did that for user deletion preview:

- https://github.com/otwcode/otwarchive/blob/14fbeb6ae13f9637cfdd3c4f586620d8863ff32a/app/views/users/delete_preview.html.erb
- https://github.com/otwcode/otwarchive/blob/14fbeb6ae13f9637cfdd3c4f586620d8863ff32a/config/locales/views/en.yml#L142-L162

Our i18n house style is documented at https://github.com/otwcode/otwarchive/wiki/Internationalization-%28i18n%29-Standards.

app/views/pseuds/delete_preview.html.erb
```diff
@@ -1,28 +1,27 @@
...
-  transferred to your default pseud.
+  <%= ts("When you delete your pseud, any works, series, or comments you have created under it will be 
+  transferred to your default pseud.") %>
 </p>
 
 <% if @pseud.bookmarks %>
   <%= form_tag({:action => "destroy"}, :method => :delete) do %>
-    <p>You have saved <%= @pseud.bookmarks.count %> bookmark(s) using this pseud. You can choose whether to
-    delete them or transfer them to your default pseud.
+    <p> <%= ts("You have saved  #{pluralize(@pseud.bookmarks.count, "bookmark")} using this pseud. You can choose whether to delete them or transfer them to your default pseud.") %>
```
redsummernight, line 10: For [pluralization](https://guides.rubyonrails.org/i18n.html#pluralization), you'll need to add at least 2 translations, "one" and "other", for the default English language.

redsummernight (dismissed):

app/views/pseuds/delete_preview.html.erb
redsummernight: Also, you shouldn't need to use `pluralize` directly. Once you set up the correct translation strings, you can simply use the `%{count}` placeholder in them.

designateddolt:
Sorry it took me so long to update this! I think I've finally got this one right. Let me know if I've bungled anything :)

config/locales/views/en.yml
```diff
@@ -380,3 +380,18 @@ en:
       title: "Invitation Status for %{email}"
       position_html: "You are currently number %{position} on our waiting list!"
       date: "At our current rate, you should receive an invitation on or around: %{date}."
+  pseuds:
+    delete_preview:
+      heading: "Pseud deletion"
+      notice: "When you delete your pseud, any works, series, or comments you have created under it will be transferred to your default pseud."
+      saved_bookmarks: 
+        one: "You have saved %{count} bookmark using this pseud. You can choose whether to delete them or transfer them to your default pseud."
```
sarken, line 388: Not to step on @redsummernight's toes here, but the second sentence here should probably use "it" rather than "them," since we're just talking about one bookmark. It also looks like this file needs a newline at the end.

designateddolt: My bad!

sarken (approved):

PR #4280, opened 2022-07-04
AO3-5962 Drop position from invite requests to make sending invitations faster.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5962

## Purpose

- Add a migration to drop the `position` column from the `invite_requests` table.
  - Add `position` to the ignored columns for the `InviteRequest` class to try to reduce errors when dropping.
- Modify the table of invite requests so that the position is either:
  - Computed from the current page and the index of the request on the page.
  - Or, when the list is filtered, computed by checking the number of smaller IDs in the database.
- Tidy up the `InviteRequestsController`:
  - Remove the `reorder` action, since it was only used to fix broken positions.
  - Modify the routes file to clean up some unused routes.
  - Remove some unused view files.
- Make the `Admin::AdminInvitationsController#invite_new_users` action send invitations asynchronously.
  - Add an `InviteFromQueueJob` class to handle the job, and an `ApplicationJob` class to use as the parent.
  - Change the text of the message to say "are being invited" instead of "were invited," since it's more accurate.
  - Move the message into `config/locales/controllers/en.yml` to make pluralization easier.


tickinginstant:
I'm not 100% sure what to do about the issue flagged by Brakeman. It's not showing up as a comment, because the line associated with the issue is not modified in this PR:
https://github.com/otwcode/otwarchive/blob/14fbeb6ae13f9637cfdd3c4f586620d8863ff32a/app/views/invite_requests/_invite_request.html.erb#L9
The issue comes from changing how `@position_in_queue` is set -- it's roughly the same computation under the hood, but because it's computed by calling a method on the `InviteRequest` model, instead of through code in the controller, Brakeman flags the value as a model attribute, and therefore potentially unsafe.

I could leave it, since it's not a real security risk (it's guaranteed to be a number, so it's fine to call `html_safe` on). I could also try to edit `app/views/invite_requests/_invite_request.html.erb` to address the issue, either by calling `h(@position_in_queue)`, or by moving the text into I18n with a key ending in `_html`. Any preferences?

redsummernight:
> by moving the text into I18n with a key ending in `_html`

If it gets rid of the warning, let's do that. We'll need to i18n-ize the view eventually.

app/models/admin_setting.rb
```diff
@@ -60,7 +60,7 @@ def self.check_queue
       if Date.today >= self.invite_from_queue_at.to_date
         new_date = Time.now + self.invite_from_queue_frequency.days
         self.first.update_attribute(:invite_from_queue_at, new_date)
-        InviteRequest.invite
+        InviteFromQueueJob.perform_now(count: invite_from_queue_number)
```
redsummernight, line 63: Did you also find that ActiveJob doesn't work well with resque-scheduler [out of the box](https://github.com/JustinAiken/active_scheduler)?

(This change is fine as is, anyway.)

tickinginstant: Yeah, it's annoying that you need to either add very complex arguments to the resque-scheduler file, or add a package to do it for you.

Is leaving `check_queue` as a rake task okay for now, or did you want me to move it into resque-scheduler?

redsummernight: If you don't mind, please and thanks! [AO3-5794](https://otwarchive.atlassian.net/browse/AO3-5794) still has the right cron schedule.

tickinginstant: I wasn't sure whether you also wanted me to move the code into a second `ActiveJob`, but I can if you want.

redsummernight: Would that run into resque-scheduler compatibility issues?

tickinginstant: Yes, I think I'd have to add a gem or use Rails internal classes to get it to work.

redsummernight: Nah, it's not worth it. This is good as is.

redsummernight (approved):

PR #4285, opened 2022-07-08
AO3-6166 Prevent unrevealed/hidden works title exposure

korrien:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6166

## Purpose

Attempts to make sure that users can't access info in unrevealed/hidden work from most cases.
Had a lot of uncertainty when trying to do this, so mostly ended up writing tests...

- Originally wanted to model bookmarks after comments (as per issue), but it didn't seem to work nicely as for bookmarks works would be visible in various bookmarks indices.
- There are some differences between works hidden by admin/unrevealed works which I am not sure are expected.
- Currently due to various pre-checks redirects happen to various places with various messages. Didn't know if we want to unify them and what to. So at the moment left as they are.
- Tests are probably too extensive, but it is caused by noticed different behavior in various cases. When originally tried to make bookmarks more like comments, noticed that different paths (/bookmarks, /work/id/bookmarks) can cause different behavior. So added it to tests.
- If expected behavior is unified, tests probably can be smaller in size (and faster if we decide that some make no sense).
- Not sure if feature tests are needed. They would nicely show that titles are not exposed (don't think that's possible with rspec), but I wouldn't know which configurations to choose. Running everything would be too slow and choosing just some configurations do not prove "we never disclose the title"
-  Uncertain about some error messages (for example message for author editing bookmark modelled after same in comments)
- Added some paths not mentioned in the issue (like navigate). There may be more, but that's what seemed to make sense.

(also tests for /users/id/bookmarks seem to fail on CI, don't know why. Locally they work, but on CI ES complains)

## Credit

korrien, she/her

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -462,6 +462,18 @@ public
     end
   end
 
+  # Checks if user is allowed to see related page if parent page is hidden or in hidden collection
+  def check_parent_visibility(parent)
+    # Only admins and the owner can see related pages on something hidden by an admin.
+    if parent.respond_to?(:hidden_by_admin) && parent.hidden_by_admin
+      logged_in_as_admin? || current_user_owns?(parent) || access_denied(redirect: root_path)
+    end
+    # Only admins and the owner can see related pages on unrevealed works.
+    if parent.respond_to?(:in_unrevealed_collection) && parent.in_unrevealed_collection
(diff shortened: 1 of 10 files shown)


Bilka2:
Work continued in the linked PR. Apologies that nobody ended up reviewing this in nearly 3 years!

PR #4287, opened 2022-07-10
AO3-6364 Further restrict admin access for managing invitation queue

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6364

## Purpose

Restricts ability to manage invitation queue to PAC, Support, and superadmins.

## Testing Instructions

Refer to Jira.


redsummernight:
Let's do #4280 first, which removes the reorder action.

app/controllers/invite_requests_controller.rb
```diff
@@ -63,6 +67,8 @@ def reorder
 
   def destroy
     @invite_request = InviteRequest.find_by(id: params[:id])
+    authorize @invite_request || authorize(InviteRequest)
```
sarken, line 70: Do we want to switch from `find_by` to `find` in the line above, so we get a 404 when the request can't be found? Otherwise the `find_by` means `@invite_request` can be `nil` and we either need to do `authorize(InviteRequest)` or `authorize @invite_request || authorize(InviteRequest)` to handle cases where the ID passed is for a nonexistent request.

app/controllers/invite_requests_controller.rb
```diff
@@ -63,7 +67,13 @@ def reorder
 
   def destroy
     @invite_request = InviteRequest.find_by(id: params[:id])
-    if @invite_request.nil? || @invite_request.destroy
+    if !policy(InviteRequest).can_destroy?
+      error_message = [ts("Sorry, only an authorized admin can do that.")]
+      respond_to do |format|
+        format.html { admin_only_access_denied }
+        format.json { render json: { errors: error_message }, status: :forbidden }
+      end
```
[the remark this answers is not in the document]
sarken: Sure! I just wasn't sure if we wanted to modify `admin_only_access_denied` for a single use case, but we'll probably have more of these.

app/controllers/invite_requests_controller.rb
sarken: ...okay, well, I didn't mean to commit that, but I did, so I guess I'll just finish it.

app/controllers/application_controller.rb
```diff
@@ -292,9 +292,15 @@ def access_denied(options ={})
...
-    flash[:error] = ts("Sorry, only an authorized admin can access the page you were trying to reach.")
-    redirect_to root_path
-    false
+    respond_to do |format|
+      format.html do
+        flash[:error] = ts("Sorry, only an authorized admin can access the page you were trying to reach.")
+        redirect_to root_path
+        false
+      end
+      errors = [ts("Sorry, only an authorized admin can do that.")]
```
redsummernight, line 301: Can you move this line inside the `format.json` block, since it's only used there?

app/controllers/application_controller.rb
```diff
@@ -292,9 +292,15 @@ def access_denied(options ={})
...
 
   def admin_only_access_denied
-    flash[:error] = ts("Sorry, only an authorized admin can access the page you were trying to reach.")
-    redirect_to root_path
-    false
+    respond_to do |format|
+      format.html do
+        flash[:error] = ts("Sorry, only an authorized admin can access the page you were trying to reach.")
+        redirect_to root_path
+        false
```
redsummernight, line 299: You can remove `false`. In Rails 6, [renders or redirects will already halt the request cycle](https://guides.rubyonrails.org/v6.0/action_controller_overview.html#filters).

spec/models/invite_request_spec.rb
```diff
@@ -1,5 +1,9 @@
 require 'spec_helper'
 describe InviteRequest, :ready do
```
redsummernight, line 2: Can you remove the unused "ready" tag?

app/policies/invite_request_policy.rb
```diff
@@ -0,0 +1,11 @@
+class InviteRequestPolicy < ApplicationPolicy
+  MANAGE_ROLES = %w[superadmin policy_and_abuse support].freeze
+
+  def can_manage?
+    user_has_roles?(MANAGE_ROLES)
+  end
+
+  alias can_destroy? can_manage?
```
tickinginstant, line 8: Is `can_destroy?` called anywhere?

sarken: I have utterly no idea why I put that there. I'll remove it.

redsummernight (approved):

PR #4289, opened 2022-07-12
AO3-6337 Remove unused CollectionItemsController#destroy action

gmjimenezc:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6337

## Purpose

What does this PR do?

* It removes the unused  #destroy action in CollectionItemsController and its related functions/code.

## Testing Instructions

Automated tests are enough

## References

## Credit

warlockmel (she/her)

### DIFF
--- a/app/controllers/collection_items_controller.rb
+++ b/app/controllers/collection_items_controller.rb
@@ -3,7 +3,6 @@ class CollectionItemsController < ApplicationController
   before_action :load_user, only: [:update_multiple]
   before_action :load_item_and_collection, only: [:update_multiple]
   before_action :load_collectible_item, only: [:new, :create]
-  before_action :allowed_to_destroy, only: [:update_multiple]
 
   cache_sweeper :collection_sweeper
 
@@ -15,11 +14,6 @@ class CollectionItemsController < ApplicationController
     end
     not_allowed(@collection) and return unless @collection_item
     @collection = @collection_item.collection
-  end
-
-
-  def allowed_to_destroy
-    @collection_item.user_allowed_to_destroy?(current_user) || not_allowed(@collection)
   end
 
   def index
--- a/app/controllers/collection_items_controller.rb
+++ b/app/controllers/collection_items_controller.rb
@@ -1,7 +1,6 @@
 class CollectionItemsController < ApplicationController
   before_action :load_collection
   before_action :load_user, only: [:update_multiple]
-  before_action :load_item_and_collection, only: [:update_multiple]
   before_action :load_collectible_item, only: [:new, :create]
 
   cache_sweeper :collection_sweeper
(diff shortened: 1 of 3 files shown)


app/controllers/collection_items_controller.rb
```diff
@@ -1,9 +1,9 @@
 class CollectionItemsController < ApplicationController
   before_action :load_collection
   before_action :load_user, only: [:update_multiple]
-  before_action :load_item_and_collection, only: [:destroy]
+  before_action :load_item_and_collection, only: [:update_multiple]
   before_action :load_collectible_item, only: [:new, :create]
-  before_action :allowed_to_destroy, only: [:destroy]
+  before_action :allowed_to_destroy, only: [:update_multiple]
```
gmjimenezc, line 6: I'm not sure about this line, should I delete it altogether? Because it would prompt the allowed_to_destroy everytime and :update_multiple in this Collection Items is triggered. Not sure if that's the intended behavior.

app/controllers/collection_items_controller.rb
```diff
@@ -192,7 +186,7 @@ def update_multiple_with_params(allowed_items:, update_params:, success_path:)
     allowed_items.where(id: update_params.keys).each do |item|
       item_data = update_params[item.id]
       if item_data[:remove] == "1"
-        @collection_items << item unless item.destroy
+        @collection_items << item unless item.update_multiple
```
sarken, line 189: I'm pretty sure this should still be `destroy`, which doesn't refer to the controller action, but rather to  the method for destroying the object -- which is what we want to do when removing a collection item.

app/controllers/collection_items_controller.rb
```diff
@@ -1,9 +1,8 @@
 class CollectionItemsController < ApplicationController
   before_action :load_collection
   before_action :load_user, only: [:update_multiple]
-  before_action :load_item_and_collection, only: [:destroy]
+  before_action :load_item_and_collection, only: [:update_multiple]
```
sarken, line 4: Your first instinct was right. This should be removed, and so should `load_item_and_collection`.

sarken (changes requested):
I think you also need to update routes.rb:

https://github.com/otwcode/otwarchive/blob/cb53d18ded95d9ece8c3b7314bf51c0e828c0755/config/routes.rb#L235

app/controllers/collection_items_controller.rb
```diff
@@ -1,9 +1,7 @@
 class CollectionItemsController < ApplicationController
   before_action :load_collection
   before_action :load_user, only: [:update_multiple]
-  before_action :load_item_and_collection, only: [:destroy]
```
tickinginstant, line 4: Like `allowed_to_destroy`, you should be able to delete `load_item_and_collection` now that it's unused. (And as a consequence, you should be able to delete `collection_item_params` as well.)

sarken (approved):

PR #4291, opened 2022-07-15
AO3-6349 When reindexing works, try to perform fewer database queries.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6349

## Purpose

This PR uses `includes()` to try to reduce the number of database queries performed when reindexing works, in an attempt to make reindexing faster. On my machine this makes reindexing works 3-5x faster, but your mileage may vary.


redsummernight (approved):
Great specs!

PR #4293, opened 2022-07-17
AO3-2197 Fix edge cases of chapter kudos cleanup

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-2197

## Purpose

- Expire kudos-related caches after re-saving kudos. Move kudos sweeper logic into model callbacks so we can use it outside controllers.
- Set some kudos validations (not on own works, not guest on restricted) to run only on create, and not on cleanup. It's possible for e.g. guest kudos to exist on works that have not always been restricted.
- Log chapter kudos that cannot be destroyed and chapter kudos that cannot be updated separately.
- Destroy chapter kudos if the chapter's work does not exist.

## Testing Instructions

See issue.


lib/tasks/after_tasks.rake
```diff
@@ -962,26 +962,34 @@ namespace :After do
 
     puts("Updating #{kudos_count} chapter kudos") && STDOUT.flush
 
-    skipped_kudo_ids = []
+    indestructible_kudo_ids = []
+    unupdatable_kudo_ids = []
```
sarken, line 966: This amuses me greatly.

sarken (approved):

PR #4298, opened 2022-07-20
AO3-6344 Add retry-after header when rate limited

nianeyna:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6344

## Purpose

Add "Retry-After" response header to "too many requests" responses so that polite clients know how long to back off for

## Testing Instructions

- Change RATE_LIMIT_NUMBER in config.yml to 1 and navigate to the root page with the network tab open
- Refresh the page, you should see "Retry later"
- Check the response headers in the network tab to ensure that "Retry-After" is present

## Credit

Name: nianeyna
Jira name: Nia Neyna
Pronouns: she/her

### DIFF
--- a/spec/requests/rack_attack_spec.rb
+++ b/spec/requests/rack_attack_spec.rb
@@ -63,7 +63,7 @@ describe "Rack::Attack" do
       expect(response).to have_http_status(:too_many_requests)
       expect(response.headers["Retry-After"]).not_to be_nil
       expect(response.headers["Retry-After"].to_i).to be > 0
-      expect(response.headers["Retry-After"].to_i).to be <= ArchiveConfig.RATE_LIMIT_PERIOD
+      expect(response.headers["Retry-After"].to_i).to be <= ArchiveConfig.RATE_LIMIT_LOGIN_PERIOD
     end
 
     it "throttles the next attempt from the same IP" do
@@ -97,7 +97,7 @@ describe "Rack::Attack" do
       expect(response).to have_http_status(:too_many_requests)
       expect(response.headers["Retry-After"]).not_to be_nil
       expect(response.headers["Retry-After"].to_i).to be > 0
-      expect(response.headers["Retry-After"].to_i).to be <= ArchiveConfig.RATE_LIMIT_PERIOD
+      expect(response.headers["Retry-After"].to_i).to be <= ArchiveConfig.RATE_LIMIT_LOGIN_PERIOD
     end
 
     it "throttles the next attempt for the same username" do
--- a/spec/requests/rack_attack_spec.rb
+++ b/spec/requests/rack_attack_spec.rb
@@ -25,6 +25,30 @@ describe "Rack::Attack" do
     expect(response).to have_http_status(:redirect)
   end
 
+  it "successful response does not include retry-after header" do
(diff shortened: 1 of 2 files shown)


sarken:
Hi, Nia!

spec/requests/rack_attack_spec.rb
```diff
@@ -34,6 +58,14 @@ def unique_user_params
       end
     end
 
+    it "response to the next attempt from the same IP includes retry-after header" do
+      post user_session_path, params: unique_user_params.to_query, env: { "REMOTE_ADDR" => ip }
+      expect(response).to have_http_status(:too_many_requests)
+      expect(response.headers["Retry-After"]).not_to be_nil
+      expect(response.headers["Retry-After"].to_i).to be > 0
+      expect(response.headers["Retry-After"].to_i).to be <= ArchiveConfig.RATE_LIMIT_PERIOD
```
sarken, line 66: I think this and line 100 both need to use `RATE_LIMIT_LOGIN_PERIOD`.

sarken (changes requested):
Thanks for bearing with us!

spec/requests/rack_attack_spec.rb
```diff
@@ -25,6 +25,30 @@ def unique_user_params
     expect(response).to have_http_status(:redirect)
   end
 
+  it "successful response does not include retry-after header" do
+    get root_path, env: { "REMOTE_ADDR" => Faker::Internet.unique.ip_v4_address }
```
redsummernight, line 29: Can you use `public_ip_v4_address` instead? `ip_v4_address` sometimes return an internal address which is not rate limited because of the `safelist_ip` rule, causing intermittent test failures. (Please feel free to update the existing uses of `ip_v4_address` in this file as well.

spec/requests/rack_attack_spec.rb
```diff
@@ -25,6 +25,30 @@ def unique_user_params
...
+    before do
+      ArchiveConfig.RATE_LIMIT_NUMBER.times do
+        get root_path, env: { "REMOTE_ADDR" => ip }
+      end
+    end
+
+    it "response to the next attempt from the same IP includes retry-after header" do
+      get root_path, env: { "REMOTE_ADDR" => ip }
+      expect(response).to have_http_status(:too_many_requests)
+      expect(response.headers["Retry-After"]).not_to be_nil
```
redsummernight, line 46: You can omit this check since `nil..to_i` is 0 and covered by the check after.

Also, since time is frozen in these tests, is this header value equal exactly e.g. `ArchiveConfig.RATE_LIMIT_PERIOD`? If so, we can check for that.

spec/requests/rack_attack_spec.rb
nianeyna: I think you are correct! Done.

spec/requests/rack_attack_spec.rb
nianeyna: Checking for exact equality resulted in failed tests, so I guess they are not (always?) equal. I'm not familiar enough with the code to know why that is.

redsummernight: I don't know either, but checking for a range is fine.

redsummernight (approved):
Looks good, thanks!

sarken (approved):

PR #4302, opened 2022-07-24
AO3-6361 Record admin email change

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6361

## Purpose

Record a log item when admins change a user's email.

## Testing Instructions

See bug replication steps in Jira.

## References

## Credit

Brian Austin (they/he)

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -79,6 +79,7 @@ class User < ApplicationRecord
   before_update :add_renamed_at, if: :will_save_change_to_login?
   after_update :update_pseud_name
   after_update :log_change_if_login_was_edited
+  after_update :log_email_change, if: :saved_change_to_email?
 
   after_commit :reindex_user_creations_after_rename
 
@@ -542,9 +543,15 @@ class User < ApplicationRecord
     self.renamed_at = Time.current
   end
 
-   def log_change_if_login_was_edited
-     create_log_item(options = { action: ArchiveConfig.ACTION_RENAME, note: "Old Username: #{login_before_last_save}; New Username: #{login}" }) if saved_change_to_login?
-   end
+  def log_change_if_login_was_edited
+    create_log_item(options = { action: ArchiveConfig.ACTION_RENAME, note: "Old Username: #{login_before_last_save}; New Username: #{login}" }) if saved_change_to_login?
+  end
+
+  def log_email_change
+    current_admin = User.current_user if User.current_user.is_a?(Admin)
+    note = current_admin ? "Change made by admin #{current_admin&.login}" : "Change made by user"
+    create_log_item(action: ArchiveConfig.ACTION_NEW_EMAIL, admin_id: current_admin&.id, note: note)
+  end
 
   def remove_stale_from_autocomplete
     Rails.logger.debug "Removing stale from autocomplete: #{autocomplete_search_string_was}"
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -549,8 +549,12 @@ class User < ApplicationRecord
 
   def log_email_change
     current_admin = User.current_user if User.current_user.is_a?(Admin)
-    note = current_admin ? "Change made by admin #{current_admin&.login}" : "Change made by user"
-    create_log_item(action: ArchiveConfig.ACTION_NEW_EMAIL, admin_id: current_admin&.id, note: note)
+    options = {
+      action: ArchiveConfig.ACTION_NEW_EMAIL,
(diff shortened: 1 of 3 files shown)


app/models/user.rb
```diff
@@ -542,9 +543,15 @@ def add_renamed_at
...
-   def log_change_if_login_was_edited
-     create_log_item(options = { action: ArchiveConfig.ACTION_RENAME, note: "Old Username: #{login_before_last_save}; New Username: #{login}" }) if saved_change_to_login?
-   end
+  def log_change_if_login_was_edited
+    create_log_item(action: ArchiveConfig.ACTION_RENAME, note: "Old Username: #{login_before_last_save}; New Username: #{login}") if saved_change_to_login?
+  end
+
+  def log_email_change
+    current_admin = User.current_user if User.current_user.is_a?(Admin)
+    note = current_admin ? "Change made by admin #{current_admin&.login}" : "Change made by user"
```
sarken, line 552: I'm checking the notes for other actions/current notes for email changes and let's go with: 
- `"Change made by #{current_admin&.login}"` when it's an admin, for consistency with the role change note
- `System Generated` when it's the user, for consistency with the existing message (this is also consistent with user-made password changes)

spec/models/user_spec.rb
```diff
@@ -136,9 +136,22 @@
       expect(existing_user.renamed_at).to eq(Time.current)
     end
 
-    it "does not set renamed_at when username is not changed" do
-      existing_user.update(email: "newemail@example.com")
-      expect(existing_user.renamed_at).to be_nil
+    context "email is changed" do
```
sarken, line 139: As [recommended by Better Specs](https://www.betterspecs.org/#contexts), we generally start our `context` descriptions with `when` (`with` and `without` are also very good).

It would probably also make sense to move these tests after the ones relating to renames -- right now it's separating `it "sets renamed_at if username is changed"` from all of the other rename-related tests.

spec/models/user_spec.rb
```diff
@@ -173,6 +186,23 @@
         expect(existing_user.login).to eq("new")
       end
     end
+
+    context "as an admin" do
+      let(:admin) { create(:admin) }
+
+      before do
+        User.current_user = admin
+        existing_user.update!(email: "new_email@example.com", profile: nil)
```
sarken, line 195: Can you explain why `profile: nil` is necessary here?

sarken (changes requested):

spec/models/user_spec.rb
brianjaustin: It's a workaround for the requirement that a ticket be associated with this admin change (otherwise updating fails with a validation error). I could do some mocking here similar to https://github.com/otwcode/otwarchive/blob/master/spec/models/concerns/justifiable_spec.rb if you'd like, but I didn't want to add too much that's not directly related to the test

sarken: Hmm, I'm not sure that's what the problem is. If I reload the user or even just use `let!` to create it before trying to update it, the tests pass without `profile: nil`. I also don't have that problem when trying to update the user from the console, even if I have `User.current_user` set to an admin.

My _guess_ is it relates to `before_create :create_default_associateds` not firing for some reason -- it's responsible for creating the user profile, and the error I get is that the profile doesn't exist, making the user invalid, not that I'm missing a ticket number... but if that's the case, I'm not sure why the other specs don't have that issue.

@redsummernight or @tickinginstant, do you know what might be up here? I feel like I've run into something along this line before and one of you has set me straight, but I am drawing a blank...

tickinginstant: Using `let` for `existing_user` means that the user object is created the first time that `existing_user` is accessed. In this test case (but not any of the others), that occurs after setting `User.current_user = admin`, which means that the system is treating it like someone logged in as an `Admin` is trying to create a `User` account. That's causing the `Justifiable` callbacks to trigger (because admins aren't supposed to be modifying user profiles willy-nilly), which prevents the profile from being saved alongside the user.

Switching to `let!` seems like a reasonable solution to me, since the test failure here effectively comes from the fact that `existing_user` doesn't exist before the admin "logs in" to try to update it.

brianjaustin: Thank you both for the explanations! I've updated the tests to use `let!` so now the weird `nil` update isn't needed anymore

sarken (approved):
Looks good to me -- thank you!

PR #3847, opened 2020-06-20
AO3-5974 Update sanitizer configuration with workaround for CVE-2020-4054

zz9pzza:
# Pull Request Checklist

* [ ] Have you read ["How to write the perfect pull request"](https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/)?
* [ ] Have you read the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [ ] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)?
* [ ] Have you added the [Jira](https://otwarchive.atlassian.net) issue number
  as the *first* thing in your pull request title (e.g. `AO3-1234 Fix thing`)
* [ ] Do you have fewer than 5 pull requests already open? If not, please wait
  until they are reviewed and merged before creating new pull requests.

## Issue

https://otwarchive.atlassian.net/browse/AO3-5974

## Purpose

There is a security issue with sanitizer and this is the work around.

## Testing Instructions

To test, we’ll probably want to try creating a work with embeds (iframe in particular) and our HTML regression test doc.

PR #2217, opened 2015-11-07
AO3-4365 XHTML to HTML5

sarken:
https://otwarchive.atlassian.net/browse/AO3-4365

Change from XHTML strict doctype to HTML5. Remove http://archiveofourown.org/auth_error.html and http://archiveofourown.org/placeholder.html which we were hopefully not ever showing to anyone because good lord are they ancient and broken relics.

PR #4303, opened 2022-07-25
AO3-5801 Do not close self-closing tags for better HTML5 compliance

potpotkettle:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5801

## Purpose

- Some multimedia tags are self-closing in HTML5. So don't try to close them. ``<source ...>`` should remain ``<source ...>``, not ``<source ...></source>``, for example.
- This is an HTML4 vs HTML5 difference, and the current sanitizer uses HTML4 parser. So more broadly, move away from HTML4 and XHTML to HTML5. The HTML output has been declared to be HTML5 as a whole since AO3-4365 (#2217).
- To do so, prefer ``Nokogiri::HTML5`` over ``Nokogiri::HTML`` and ``Nokogiri::HTML4`` (but still use ``Nokogiri::HTML4`` where necessary). Note that ``Nokogiri::HTML`` and ``Nokogiri::HTML4`` are synonymous.
- The shift to HTML5 will bring other miscelleneous changes in how the sanitizer behaves. I hope to make the impact as small as possible and added changes that I believe are necessary, so this set of commits may be bigger than anticipated.(But I'll be honest, I might have been overzealous to make Rubocop happier in parts of the code that are only marginally related to this issue.)

## Testing Instructions

Trying to post and import works with audio/video tags would be a good place to start. However, I'd go beyond that.

This set of changes (at least its initial version) changes the inner working of the sanitization process at some fundamental level, so the scope of testing would probably have to be large. I tried to make the changes minimal and reasonable, but some unintended consequences might still happen. Known side effects are written above.

Posting and importing trigger sanitization, and I think that's mainly where testing is needed. It would be wise to test this as thoroughly as possible, using short and long, old and new content.

If possible, I would check performance metrics (like average time needed to complete one posting action) before and after this change. During my own (limited amount of) manual testing, there was no noticeable difference, though.

## References

- https://github.com/otwcode/otwarchive/pull/3678 by @sarken - I reused some of the test code.
- [AO3-5693](https://otwarchive.atlassian.net/browse/AO3-5693) (#3679) - a hack that should be made obsolete by this change
- [AO3-5974](https://otwarchive.atlassian.net/browse/AO3-5974) (#3847) - same as above

The Jira issue contains some preliminary discussion, too.
 
## Credit

Potpotkettle (they/them)

### DIFF
--- a/app/models/story_parser.rb
+++ b/app/models/story_parser.rb
@@ -545,10 +545,25 @@ class StoryParser
   def parse_common(story, location = nil, encoding = nil, detect_tags = true)
     work_params = { title: "Untitled Imported Work", chapter_attributes: { content: "" } }
 
+    # check if the given encoding is valid, otherwise fallback to nil
+    begin
+      Encoding.find(encoding) if encoding
+    rescue ArgumentError
+      encoding = nil
+    end
+
     # Encode as HTML - the dummy "foo" tag will be stripped out by the sanitizer but forces Nokogiri to
     # preserve line breaks in plain text documents
     # Rescue all errors as Nokogiri complains about things the sanitizer will fix later
-    @doc = Nokogiri::HTML.parse(story.prepend("<foo/>"), nil, encoding) rescue ""
+    story.prepend("<foo><foo/>")
+    @doc = begin
+             # Using Nokogiri::HTML4 here for its better encoding handling.
+             # Sanitizer will make content HTML5 later anyway.
+             # https://nokogiri.org/rdoc/Nokogiri/HTML4/Document.html#method-c-parse
+             Nokogiri::HTML4.parse(story, nil, encoding)
+           rescue Object
+             Nokogiri::HTML4.parse("")
+           end
 
     # Try to convert all relative links to absolute
     base = @doc.at_css("base") ? @doc.css("base")[0]["href"] : location.split("?").first
@@ -701,7 +716,7 @@ class StoryParser
 
     # cleanup the notes
     notes.gsub!(%r{<br\s*\/?>}i, "\n") # replace the breaks with newlines
-    notes = clean_storytext(notes)
+    notes = clean_storytext(notes, "notes")
     work_params[:notes] = notes
 
     work_params.merge!(scan_text_for_meta(notes, detect_tags))
@@ -842,9 +857,9 @@ class StoryParser
   end
 
   # We clean the text as if it had been submitted as the content of a chapter
-  def clean_storytext(storytext)
+  def clean_storytext(storytext, field = "content")
     storytext = storytext.encode("UTF-8", invalid: :replace, undef: :replace, replace: "") unless storytext.encoding.name == "UTF-8"
-    sanitize_value("content", storytext)
+    sanitize_value(field, storytext)
   end
 
   # works conservatively -- doesn't split on
--- a/config/initializers/gem-plugin_config/sanitizer_config.rb
+++ b/config/initializers/gem-plugin_config/sanitizer_config.rb
@@ -36,10 +36,6 @@ class Sanitize
         'img' => {'src' => ['http', 'https', :relative]},
         'q' => {'cite' => ['http', 'https', :relative]}
       },
-      
-      # TODO: This can be removed once we upgrade sanitizer gem, AO3-5801
-      # I would leave the tests we added in AO3-5974 though.
-      remove_contents: %w[iframe math noembed noframes noscript plaintext script style svg xmp]
     )
 
     CLASS_ATTRIBUTE = freeze_config(
--- a/features/other_b/sanitizer.feature
+++ b/features/other_b/sanitizer.feature
@@ -5,7 +5,7 @@ Feature: Sanitizing HTML
 
     When I am logged in as "newbie" with password "password"
     And I set up the draft "My Awesome Story"
-    And I fill in "content" with 
+    And I fill in "content" with
     """
     The quick brown fox jumps over the lazy dog.
     <!-- #exec cmd=\"/bin/echo \" -->
@@ -15,6 +15,22 @@ Feature: Sanitizing HTML
     Then I should see "Preview"
     And I should not see the text with tags '<!-- #exec cmd='
     And I should not see the text with tags '<script src=http://ha.ckers.org/xss.js></script>'
+
+
+  Scenario: Sanitizer should leave safe HTML alone
+
+    When I am logged in as "newbie" with password "password"
+    And I set up the draft "My Awesome Story"
+    And I fill in "content" with
+    """
+    1. Lorem ipsim
+
+    2. <a href="https://en.wikipedia.org/wiki/The_quick_brown_fox_jumps_over_the_lazy_dog"><em>The quick brown fox jumps over the lazy dog</em></a>.
+    """
+    And I press "Preview"
+    Then I should see "Preview"
+    And I should see "The quick brown fox jumps over the lazy dog" within "p a em"
+    And I should see the text with tags '"https://en.wikipedia.org/wiki/The_quick_brown_fox_jumps_over_the_lazy_dog"'
 
 
   Scenario: XSS hacks in works should be blocked by sanitizing
--- a/lib/html_cleaner.rb
+++ b/lib/html_cleaner.rb
@@ -38,6 +38,10 @@ module HtmlCleaner
       flatten.include?("p")
     end
 
+    def immediately_inside_paragraph?
+      flatten.select { |x| x.is_a? String }.last == "p"
+    end
+
     def ignore_tag?(tag)
       ["text", "myroot", "#cdata-section"].include?(tag)
     end
@@ -67,13 +71,18 @@ module HtmlCleaner
       end
     end
 
-    def close_and_pop_last
+    def close_and_pop_last(num = nil)
       result = ""
-      pop.reverse.each do |tag, attributes|
+      popped = pop.reverse
+      num ||= popped.size
+      popped[0...num].each do |tag, _|
         next if ignore_tag?(tag)
         result += "</#{tag}>"
       end
-      return result
+
+      push(popped[num..].reverse) unless popped[num..].empty?
+
+      result
     end
 
     def add_p
@@ -169,14 +178,9 @@ module HtmlCleaner
         unfrozen_value = add_paragraphs_to_text(Sanitize.clean(fix_bad_characters(unfrozen_value),
                                Sanitize::Config::ARCHIVE.merge(transformers: transformers))) unless field.to_s == "screencast"
       end
-      doc = Nokogiri::HTML::Document.new
+      doc = Nokogiri::HTML5::Document.new
       doc.encoding = "UTF-8"
-      unfrozen_value = doc.fragment(unfrozen_value).to_xhtml
-
-      # Hack! the herald angels sing
-      # TODO: AO3-5801 Switch to an HTML5 serializer that doesn't add invalid closing tags
-      # to track and source elements.
-      unfrozen_value.gsub!(%r{</(source|track)>}, "")
+      unfrozen_value = doc.fragment(unfrozen_value).to_html
     else
       # clean out all tags
       unfrozen_value = Sanitize.clean(fix_bad_characters(unfrozen_value))
@@ -241,14 +245,14 @@ module HtmlCleaner
 
   # Tags that can't be inside p tags
   def put_outside_p_tag?(tag)
-    %w(audio dl h1 h2 h3 h4 h5 h6 hr ol p pre source table track ul video).include?(tag)
+    %w[address audio blockquote center div dl h1 h2 h3 h4 h5 h6 hr ol p pre source table track ul video].include?(tag)
   end
 
   # Tags before and after which we don't want to convert linebreaks
   # into br's and p's
   def no_break_before_after_tag?(tag)
-    %w(audio blockquote br center dl div h1 h2 h3 h4 h5 h6
-       hr ol p pre source table track ul video).include?(tag)
+    %w[audio blockquote br center div dl h1 h2 h3 h4 h5 h6
+       hr ol p pre source table track ul video].include?(tag)
   end
 
   # Traverse a Nokogiri document tree recursively in order to insert
@@ -258,8 +262,8 @@ module HtmlCleaner
   # creating a string with the resulting html and keep track of the
   # changed path to the current element via a stack.
   def traverse_nodes(node, stack=nil, out_html=nil)
-    stack = stack || TagStack.new
-    out_html = out_html || ""
+    stack ||= TagStack.new
+    out_html ||= ""
 
     # Convert double and triple br tags into paragraph breaks
     if node.name == "br" && node.previous_sibling && node.previous_sibling.name == "br" && node.previous_sibling.previous_sibling && node.previous_sibling.previous_sibling.name == "br"
@@ -278,7 +282,7 @@ module HtmlCleaner
     # this kind of tag
     if dont_touch_content_tag?(node.name)
       if put_inside_p_tag?(node.name) && !stack.inside_paragraph?
-        return [stack, out_html + "<p>#{node.to_s}</p>"]
+        return [stack, out_html + "<p>#{node}</p>"]
       end
 
       if put_outside_p_tag?(node.name) && stack.inside_paragraph?
@@ -290,20 +294,17 @@ module HtmlCleaner
     end
 
     if !node.text? && !node.cdata?
+      out_html += stack.close_and_pop_last(1) if put_outside_p_tag?(node.name) && stack.immediately_inside_paragraph?
+
       out_html += stack.add_p if put_inside_p_tag?(node.name) && !stack.inside_paragraph?
 
       stack << [[node.name, Hash[*(node.attribute_nodes.map { |n| [n.name, n.value] }.flatten)]]]
       out_html += open_tag(node)
 
       # If we are the root node, pre-emptively open a paragraph
-      if node.name == "myroot"
-        out_html += stack.add_p
-      end
+      out_html += stack.add_p if node.name == "myroot" && node.children.any? { |x| put_inside_p_tag?(x.name) || x.text? || x.cdata? }
 
-      if no_break_before_after_tag?(node.name) and !stack.last.include?("p")
-        out_html += stack.add_p
-      end
-
+      out_html += stack.add_p if no_break_before_after_tag?(node.name) && !stack.last.include?("p")
     else
       text = node.to_s
 
@@ -361,8 +362,9 @@ module HtmlCleaner
     end
 
     # Adding paragraphs in place of linebreaks
-    doc = Nokogiri::HTML.fragment("<myroot>#{text}</myroot>")
+    doc = Nokogiri::HTML5.fragment("<myroot>#{text}</myroot>")
     out_html = traverse_nodes(doc.at_css("myroot"))[1]
+
     # Remove empty paragraphs
     out_html.gsub!(/<p>\s*?<\/p>/, "")
     out_html.gsub!(/(\A<myroot>)|(<\/myroot>\Z)|(\A<myroot\/>\Z)/, "")
--- a/lib/otw_sanitize/embed_sanitizer.rb
+++ b/lib/otw_sanitize/embed_sanitizer.rb
@@ -155,7 +155,7 @@ module OtwSanitize
       end
       return if mp3_urls.blank?
 
-      audio_fragment = Nokogiri::HTML::DocumentFragment.parse ""
+      audio_fragment = Nokogiri::HTML5::DocumentFragment.parse ""
       Nokogiri::HTML::Builder.with(audio_fragment) do |fragment|
         fragment.p do
           mp3_urls.each_with_index do |url, i|
--- a/spec/lib/html_cleaner_spec.rb
+++ b/spec/lib/html_cleaner_spec.rb
@@ -5,6 +5,10 @@ require 'nokogiri'
 describe HtmlCleaner do
   include HtmlCleaner
 
+  def one_cell_table(content)
+    "<table><tr><td>#{content}</td></tr></table>"
+  end
+
   describe "TagStack" do
     let(:stack) { HtmlCleaner::TagStack.new }
 
@@ -123,11 +127,6 @@ describe HtmlCleaner do
       expect(result).to eq("some <s><i>more text</i></s>")
     end
 
-    it "should close tag before next closing tag" do
-      result = close_unclosed_tag("some <s><i>more text</s>", "i", 1)
-      expect(result).to eq("some <s><i>more text</i></s>")
-    end
-
     it "should close second opening tag" do
       result = close_unclosed_tag("some <i>more</i> <i>text", "i", 1)
       expect(result).to eq("some <i>more</i> <i>text</i>")
@@ -172,10 +171,11 @@ describe HtmlCleaner do
           end
         end
 
-        it "keeps google player embeds" do
-          html = '<embed type="application/x-shockwave-flash" flashvars="audioUrl=http://dl.dropbox.com/u/123/foo.mp3" src="http://www.google.com/reader/ui/123-audio-player.swf" width="400" height="27" allowscriptaccess="never" allownetworking="internal"></embed>'
-          result = sanitize_value(field, html)
-          expect(result).to include(html)
+        it "keeps google player embeds without closing tag" do
+          html1 = '<embed type="application/x-shockwave-flash" flashvars="audioUrl=http://dl.dropbox.com/u/123/foo.mp3" src="http://www.google.com/reader/ui/123-audio-player.swf" width="400" height="27" allowscriptaccess="never" allownetworking="internal">'
+          html2 = "#{html1}</embed>"
+          result = sanitize_value(field, html2)
+          expect(result).to include(html1)
         end
 
         it "converts an Archive-hosted Dewplayer embed into an audio tag" do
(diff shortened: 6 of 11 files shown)


config/initializers/gem-plugin_config/sanitizer_config.rb
```diff
@@ -36,10 +36,6 @@ module Config
         'img' => {'src' => ['http', 'https', :relative]},
         'q' => {'cite' => ['http', 'https', :relative]}
       },
-      
-      # TODO: This can be removed once we upgrade sanitizer gem, AO3-5801
-      # I would leave the tests we added in AO3-5974 though.
-      remove_contents: %w[iframe math noembed noframes noscript plaintext script style svg xmp]
```
potpotkettle, line 42: I believe this issue was resolved earlier than my PR, when `sanitize` was updated to 6.0.0 in AO3-6085 (Upgrade to Rails 6). Should we update Gemfile as well? In Gemfile `sanitize` is still '>= 4.6.5'.

lib/html_cleaner.rb
```diff
@@ -290,20 +294,17 @@ def traverse_nodes(node, stack=nil, out_html=nil)
...
 
       # If we are the root node, pre-emptively open a paragraph
-      if node.name == "myroot"
-        out_html += stack.add_p
-      end
-
-      if no_break_before_after_tag?(node.name) and !stack.last.include?("p")
-        out_html += stack.add_p
-      end
+      out_html += stack.add_p if node.name == "myroot" && node.children.any? { |x| put_inside_p_tag?(x.name) || x.text? || x.cdata? }
```
potpotkettle, line 305: I believe the condition ` node.name == "myroot"` is too loose and adds unnecessary `p` tags a lot.

However, the additional conditions may degrade performance.

lib/html_cleaner.rb
```diff
@@ -290,20 +294,17 @@ def traverse_nodes(node, stack=nil, out_html=nil)
     end
 
     if !node.text? && !node.cdata?
+      out_html += stack.close_and_pop_last(1) if put_outside_p_tag?(node.name) && stack.immediately_inside_paragraph?
```
potpotkettle, line 297: I added this because the HTML5 parser is not happy with structures like `<p><h1>foo</h1><p>`. (Apparently it's stricter about not having these impossible combinations.)

potpotkettle:
The recent changes in #4382 resolved some of the standing concerns here, too. I have marked the applicable posts here resolved after working on the merge conflicts (they are mostly trivial, thankfully).

I believe the ones I have not marked, higher up in this page & written on June 25, still stand.

potpotkettle:
The recent changes in #4382 resolved some of the standing concerns here, too. I have marked the applicable posts here resolved after working on the merge conflicts (they were mostly trivial, thankfully).

I believe the ones I have not marked, higher up in this page & written on June 25, still stand.

potpotkettle:
It looks like with https://github.com/otwcode/otwarchive/commit/09d8ba39c8ec82c070a774bfd4ea479c2ee20c3c we need resolve merge conflicts again. I wonder if I should wait longer this time (because there might be another related commit coming soon).

sarken:
Waiting is definitely a good idea -- we'll be doing some more work in this area over the coming week or so.

potpotkettle:
Just to check the current status, what's the blocker here exactly? [AO3-6421](https://otwarchive.atlassian.net/browse/AO3-6421)?

redsummernight:
[AO3-6421](https://otwarchive.atlassian.net/browse/AO3-6421) is good saved for some issues with the task to re-sanitize all existing content. We probably should fix and run the task before deploying other changes to the sanitizer, but it shouldn't block reviewing other changes.

I'll update the labels.

config/initializers/gem-plugin_config/sanitizer_config.rb
potpotkettle: Now ParagraphMaker (in a2dafbcc43aadeeed377ea7e4365883f272688b6) uses this. In July I thought it was unnecessary, but I'm now less confident. If I recall it correctly, it wasn't exactly related to HTML5, but it was merely something I found while working on the HTML5 issue. I'll revert it for now.

potpotkettle:
Thanks! The second merge is now done and this is ready for review (I think).

app/models/story_parser.rb
```diff
@@ -545,10 +545,25 @@ def download_chaptered_from_efiction(location)
...
+      encoding = nil
+    end
+
     # Encode as HTML - the dummy "foo" tag will be stripped out by the sanitizer but forces Nokogiri to
     # preserve line breaks in plain text documents
     # Rescue all errors as Nokogiri complains about things the sanitizer will fix later
-    @doc = Nokogiri::HTML.parse(story.prepend("<foo/>"), nil, encoding) rescue ""
+    story.prepend("<foo><foo/>")
+    @doc = begin
+             # Using Nokogiri::HTML4 here for its better encoding handling.
```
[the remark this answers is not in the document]
potpotkettle: Thank you for the investigation.

The tests were introduced in https://github.com/otwcode/otwarchive/commit/0e59af31c028404c932acbafffb5a8a94dde7590 / [AO3-4628](https://otwarchive.atlassian.net/browse/AO3-4628) and haven't been changed since then. A straightforward implementation of using the HTML5 parser would break one of the tests, if I recall it correctly.

It seems like the intention was to prevent crashing against exoteric encodings. I don't know if that included outright wrong encodings, but I think it's good to prepare for that possibility, too. I believe [Ruby 3](https://otwarchive.atlassian.net/browse/AO3-6409) changed a lot about string encodings, so some updating might be necessary here because of that, too.

I think we could do better with HTML5 than I initially thought with a small change. In particular, we can rescue from `Encoding::UndefinedConversionError` and specify `binary` or `ASCII-8bit`, instead of leaving the encoding `nil`. It might not get you correct conversion, but it looks like it would at least avoid crashing. (I think I will add a commit that does this later.)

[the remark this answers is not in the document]
potpotkettle: Perhaps I should ask this first: given input with wrong encoding (a UTF-8 string but Windows-1252 is specified, for example), which do we prefer? This is mostly about importing.

1. Go on with potentially [malformed](https://en.wikipedia.org/wiki/Mojibake) content - I think this is what we do currently 
2. Abort with an error
3. Guess and recover the correct encoding statistically (ignoring the specified encoding) - good user experience but, requires significantly more code

potpotkettle:
Merge done again. I hope I didn't make mistakes because this time I had to a bit more of copy and paste than in the last 2 merges. (I can only blame myself - if I made my earlier commits minimal, it would have been easier.)

potpotkettle:
It looks like the last conflicts were with #4427 (Dewplayer). They seemed straightforward so I resolved on GitHub.

potpotkettle:
Why has this been stalled for years?

If the patch is difficult to review (I think that's fair), I wonder if it helps if I try starting from scratch, and submit a new, hopefully simpler, pull request.

If there is a blocker that I don't know (and is out of my control), I guess I can just continue waiting.

app/models/story_parser.rb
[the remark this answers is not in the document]
Bilka2: I would go with 1 - leaving the encoding empty will make it default to something sensible, so if the user specified it I would assume that they really do mean to use that encoding.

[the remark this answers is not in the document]
Bilka2: Because of that, I would agree with irrationalpie7 that it would be good to actually encode the body in the right encoding for the tests. Assuming that it doesn't break the importing

lib/otw_sanitize/embed_sanitizer.rb
```diff
@@ -155,7 +155,7 @@ def replace_dewplayer
       end
       return if mp3_urls.blank?
 
-      audio_fragment = Nokogiri::HTML::DocumentFragment.parse ""
+      audio_fragment = Nokogiri::HTML5::DocumentFragment.parse ""
       Nokogiri::HTML::Builder.with(audio_fragment) do |fragment|
```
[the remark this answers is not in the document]
Bilka2: Since this was written, [nokogiri 1.17.0](https://github.com/sparklemotion/nokogiri/blob/main/CHANGELOG.md#v1170--2024-12-08) came out and added a html builder:
> [CRuby] Nokogiri::HTML5::Builder is similar to HTML4::Builder but returns an HTML5::Document. [#3119]

master branch is currently using nokogiri 1.18.8 so this should be available to us now

potpotkettle:
Thanks a lot! I mostly resolved merge conflicts, and now resolving new test failures. (Not pushed yet.)

I think continuiing incremental changes would be easier for me. The only problem is that I'm not completely clear about some of the earlier changes I made years ago. (I'm sure some of them might be just my ignorance at that time, though.)

potpotkettle:
I think this is ready for review now.

lib/html_cleaner.rb
```diff
@@ -132,10 +127,9 @@ def add_paragraphs_to_text(text)
     doc = Nokogiri::HTML.fragment("<myroot>#{text}</myroot>")
```
Bilka2: Should this be HTML5?

app/models/story_parser.rb
```diff
@@ -551,14 +551,23 @@ def download_chaptered_from_efiction(location)
...
+
     # Encode as HTML - the dummy "foo" tag will be stripped out by the sanitizer but forces Nokogiri to
     # preserve line breaks in plain text documents
     # Rescue all errors as Nokogiri complains about things the sanitizer will fix later
-    begin
-      @doc = Nokogiri::HTML.parse(story.prepend("<foo/>"), encoding: encoding)
-    rescue StandardError
-      @doc = ""
-    end
+    story.prepend("<foo><foo/>")
```
Bilka2: Shouldn't this be `<foo></foo>`?

app/models/story_parser.rb
```diff
@@ -551,14 +551,23 @@ def download_chaptered_from_efiction(location)
   def parse_common(story, location = nil, encoding = nil, detect_tags = true)
     work_params = { title: "Untitled Imported Work", chapter_attributes: { content: "" } }
 
+    # check if the given encoding is valid, otherwise fallback to nil
+    begin
+      Encoding.find(encoding) if encoding
+    rescue ArgumentError
+      encoding = nil
+    end
```
Bilka2: When would this code be hit? From what I can see, the import form only allows Encodings from this same list, so in normal operation the encoding should always be found. Are you guarding against a bad test maybe? Or someone making a bad request (bypassing the form) directly to `/works/import`?

features/other_b/sanitizer.feature
```diff
@@ -17,6 +17,22 @@ Feature: Sanitizing HTML
     And I should not see the text with tags '<script src=http://ha.ckers.org/xss.js></script>'
 
 
+  Scenario: Sanitizer should leave safe HTML alone
+
+    When I am logged in as "newbie" with password "password"
```
Bilka2: You can remove the password part

features/other_b/sanitizer.feature
```diff
@@ -17,6 +17,22 @@ Feature: Sanitizing HTML
     And I should not see the text with tags '<script src=http://ha.ckers.org/xss.js></script>'
 
 
+  Scenario: Sanitizer should leave safe HTML alone
```
Bilka2, line 20: Please indent the `And`s one level

spec/models/story_parser_spec.rb
```diff
@@ -284,7 +284,7 @@ class StoryParser
   def mock_external
     curly_quotes = "String with non-ASCII “Curly quotes” and apostrophes’"
 
-    body = "
+    body = <<~STUB
```
Bilka2, line 287: Nice!

spec/lib/html_cleaner_spec.rb
Bilka2, line 1: There are instances of `Nokogiri::HTML` still in this file, should these also be converted to HTML5?

app/models/story_parser.rb
Bilka2, line 1: In `parse_author_from_lj` there are two instances of `Nokogiri.parse`, I think those should also be `Nokogiri::HTML5.parse`.

spec/lib/html_cleaner_spec.rb
```diff
@@ -40,10 +44,11 @@
...
 
-        it "keeps google player embeds" do
-          html = '<embed type="application/x-shockwave-flash" flashvars="audioUrl=http://dl.dropbox.com/u/123/foo.mp3" src="http://www.google.com/reader/ui/123-audio-player.swf" width="400" height="27" allowscriptaccess="never" allownetworking="internal"></embed>'
-          result = sanitize_value(field, html)
-          expect(result).to include(html)
+        it "keeps google player embeds without closing tag" do
+          html1 = '<embed type="application/x-shockwave-flash" flashvars="audioUrl=http://dl.dropbox.com/u/123/foo.mp3" src="http://www.google.com/reader/ui/123-audio-player.swf" width="400" height="27" allowscriptaccess="never" allownetworking="internal">'
+          html2 = "#{html1}</embed>"
+          result = sanitize_value(field, html2)
+          expect(result).to include(html1)
```
Bilka2: Should this also check that the result doesn't include `</embed>`?

spec/lib/html_cleaner_spec.rb
```diff
@@ -587,51 +592,92 @@
...
-      doc = Nokogiri::HTML.fragment(result)
-      expect(doc.xpath(".//p").size).to eq(3)
+      doc = Nokogiri::HTML5.fragment(result)
+      expect(doc.xpath("./p").size).to eq(3)
       expect(doc.xpath(".//br")).to be_empty
     end
 
+    it "does not wrap table in p tags" do
+      result = add_paragraphs_to_text("aa #{one_cell_table('foo')} bb")
+      doc = Nokogiri::XML.fragment(result)
```
Bilka2, line 675: Could you add a comment here why this uses XML?

spec/lib/html_cleaner_spec.rb
```diff
@@ -657,62 +703,62 @@
...
-      doc = Nokogiri::HTML.fragment(result)
-      expect(doc.xpath("./table/tr[1]/th[1]").children.to_s.strip).to eq("A")
-      expect(doc.xpath("./table/tr[1]/th[2]").children.to_s.strip).to eq("B")
-      expect(doc.xpath("./table/tr[2]/td[1]").children.to_s.strip).to eq("C")
-      expect(doc.xpath("./table/tr[2]/td[2]").children.to_s.strip).to eq("D")
+      doc = Nokogiri::HTML5.fragment(result)
+      expect(doc.xpath("./table//tr[1]/th[1]/node()").to_s.strip).to eq("A")
+      expect(doc.xpath("./table//tr[1]/th[2]/node()").to_s.strip).to eq("B")
+      expect(doc.xpath("./table//tr[2]/td[1]/node()").to_s.strip).to eq("C")
+      expect(doc.xpath("./table//tr[2]/td[2]/node()").to_s.strip).to eq("D")
```
Bilka2: Could these go via the inserted `tbody` tag instead of the `//`?

spec/lib/html_cleaner_spec.rb
```diff
@@ -802,85 +848,86 @@
...
     end
 
-    %w(address h1 h2 h3 h4 h5 h6 p pre).each do |tag|
-      it "should not wrap in p and not convert linebreaks inside #{tag} tags" do
+    %w[address h1 h2 h3 h4 h5 h6 p pre].each do |tag|
+      it "does not wrap in p and not convert linebreaks inside #{tag} tags" do
         result = add_paragraphs_to_text("<#{tag}>A\nB\n\nC\n\n\nD</#{tag}>")
-        doc = Nokogiri::HTML.fragment(result)
-        expect(doc.xpath("./#{tag}[1]").children.to_s.strip).to eq("A\nB\n\nC\n\n\nD")
+        doc = Nokogiri::XML.fragment(result)
```
Bilka2, line 868: Could you add a comment here why this uses XML?

spec/lib/html_cleaner_spec.rb
```diff
@@ -986,82 +1033,81 @@
...
       expect(doc.xpath("./table/colgroup[@align='left']/col[@width='20']").size).to eq(1)
       expect(doc.xpath("./table/colgroup[@align='right']").size).to eq(1)
-      expect(doc.xpath("./table/tr[1]/th[1]").children.to_s.strip).to eq("A")
-      expect(doc.xpath("./table/tr[1]/th[2]").children.to_s.strip).to eq("B")
-      expect(doc.xpath("./table/tr[2]/td[1]").children.to_s.strip).to eq("C")
-      expect(doc.xpath("./table/tr[2]/td[2]").children.to_s.strip).to eq("D")
+      expect(doc.xpath("./table//tr[1]/th[1]/node()").to_s.strip).to eq("A")
+      expect(doc.xpath("./table//tr[1]/th[2]/node()").to_s.strip).to eq("B")
+      expect(doc.xpath("./table//tr[2]/td[1]/node()").to_s.strip).to eq("C")
+      expect(doc.xpath("./table//tr[2]/td[2]/node()").to_s.strip).to eq("D")
```
Bilka2: Same tbody note here as above somewhere

spec/lib/html_cleaner_spec.rb
potpotkettle: I think my reasoning was that the closing tag was optional, and checking the absence of it might be too strict. "keeps google player embeds without closing tag" can be better phrased, though, for that. What do you think?

lib/html_cleaner.rb
potpotkettle: The change will break 2 tests.
```
spec './spec/lib/html_cleaner_spec.rb[1:3:49]' # HtmlCleaner add_paragraphs_to_text does not wrap table in p tags
rspec ./spec/lib/html_cleaner_spec.rb:1082 # HtmlCleaner add_paragraphs_to_text removes closing br tags at the beginning
```

I think we can safely remove the 1st one (in the block starting L675). Table tag is a special case and is dealt with another test above this. In Nokogiri::HTM5, `<table>text</table>` would be illegal and the text will be pushed to the left of `<table>`.

The 2nd one is a bit tricky. I don't know what is intended to verify here. It looks like it was added in https://github.com/otwcode/otwarchive/commit/6b4e2f7775c3cc56341914e58af524909f4b5886#diff-37d9e7e486d0ad52451c860f0451345d5c01561903658f496e43fd53ac3f6c28R735 . In Nokogiri::HTML5, ``<br/>`` is converted into ``<br>``, and remains in the output, if we do nothing about it. Nokogiri::HTML4 silently removes it, making the test pass. Is this removal an important feature and we want HTML5 here, I can add a simple replacement to do that. If not, maybe the test is not needed?

Or maybe we can keep HTML4 here. I guess we can make things more HTML5 compliant than before, while keeping some tricky legacy code to be dealt with later.

app/models/story_parser.rb
potpotkettle: Definitely. Thanks for pointing out!

spec/lib/html_cleaner_spec.rb
potpotkettle: I think (similar to `</embed>`) my reasoning was that `tbody` was optional and we don't have to verify its presence. However, `//` can accept wrong content in between, so I now agree that it seems better to specify `tbody`.

spec/lib/html_cleaner_spec.rb
Bilka2: [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/embed#technical_summary) defines embed as "must not have an end tag", just like [source](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/source) which is getting fixed here, so I think this would be good to check.

lib/html_cleaner.rb
Bilka2: Agreed on the table test.

The original test with `<br/></br>text` still looks to be passing with HTML5. So I would not consider this an important feature and would accept this as a behavior change.

app/models/story_parser.rb
potpotkettle: To be honest, I don't remember exactly why the check is here.  My best guess is that I might have wanted to guard against some tests. If the tests pass without it, I think it's okay to remove it.

app/models/story_parser.rb
potpotkettle: Yes, we can convert them. But it doesn't seem necessary because (I believe) the parsed result will only be used locally, and I think the goal here is not to remove all HTML4 function calls. That said, if you want them converted for consistency, I wouldn't mind doing that.

Bilka2: I see, you're right. You can keep them as HTML4 then.

spec/lib/html_cleaner_spec.rb
potpotkettle: We need to be a bit careful because HTML5 parser can sometimes silently "fix" issues (to conform the parser's strictness), possibly making tests meaningless in an unexpected way.
```
irb(main):017> Nokogiri::HTML5.fragment('<p><div>foo</div</p>').to_s
"<p></p><div>foo</div>"
irb(main):018> Nokogiri::XML.fragment('<p><div>foo</div</p>').to_s
"<p>\n  <div>foo</div>\n</p>"
```

In fact, it might safer to use XML parser almost everywhere in assertions unless there is reason to use HTML5. XML parser doesn't seem to have these silent conversions.

With XML, it seems that unwanted indentation can be introduced, though, and we will need a lot more of `strip`.

Bilka2: I see, makes sense that the spec that's tighter in what's allowed would do some auto-conversions of disallowed syntax. I think the best way forward here would be to keep the parser for the so far untouched tests as is then, since that way the tests will keep testing the same thing as originally written (mostly...).

potpotkettle:
I think the review comments are addressed by now. (Sorry if I missed something, please point it out again!)

I made new changes for line breaks. [b739949](https://github.com/otwcode/otwarchive/pull/4303/commits/b73994914ca794a712063466aa6c57c0fd78d662) Does this look good?

Explanation: I noticed that sometimes paragraphs get connected into one line, like from ``foo\n\nbar`` to ``<p>foo></p><p>bar</p>`` as a result of sanitization. As HTML it's still fine because linebreaks are meaningless, but the source code can be harder for the user to edit if linebreaks are removed. I believe line breaks are preserved similarly before this PR.

It seems that this can be fixed by re-introducing linebreaks in ParagraphMaker, although it's kind of ad-hoc. I tried to investigate why ParagraphMaker removes linebreaks like this (I think only for HTML5 input) - I couldn't figure it out why. The ad-hoc fix seems to work, though.

I made a related change to test textarea instead of HTML in parser.feature. Before this PR, linebreaks are not proved to be preserved in the edit screen.

features/step_definitions/generic_steps.rb
```diff
@@ -132,6 +132,10 @@ def assure_xpath_not_present(tag, attribute, value, selector)
   find_field("#{labeltext}").value.should == content
 end
 
+Then "I should see in the {string} input" do |labeltext, content|
+  find_field(labeltext).value.should == content
```
Bilka2, line 136: How does this match two things out of one string param?

Bilka2 (commented):
This looks really good! Thank you also for the test with saving twice - I tested this locally as well because this was my one concern with inserting extra linebreaks as well. I recall we had (have?) a bug where resaving blockquotes would keep inserter extra linebreaks, so it's good that this makes sure it won't happen here.

features/step_definitions/generic_steps.rb
potpotkettle: I first left it undefined and cucumber suggested a stub for this.

It looks like this is what is happening:
https://cucumber.io/docs/gherkin/reference/#step-arguments
> In your step definition, there’s no need to find this text and match it in your pattern. It will automatically be passed as the last argument in the step definition.

Bilka2: Ohh that is nice!

Bilka2:
I tested the 5 linebreaks that you have in the spec locally and I have bad news.
Test text:
```
a paragraph




another paragraph
```
If I preview this, HTML has the p tags with nbsp as expected. If I then press edit again, the textbox doesn't have those tags anymore, and when previewing again they're also gone from the output.

On production, the paragraph tags with the nbsp are inserted in the preview and stay through the editing cycles without changes, as expected.

potpotkettle:
It looks like config/initializers/monkeypatches/textarea_convert_html_to_newlines.rb is supposed to reverse what PargraphMaker currently does. As far as I can see, the current state of the reversion conversion is half working and half broken, and it has been the case for a long time. I think the original intention was to convert back `<p>` into double line break in the textare (after rendering the HTML), which clearly is not happening in production now.

The monkey patch *appears* to want to convert `<p>&nbsp;</p>` into `\n\n\n`, but it doesn't happen in production now, because `&nbsp;` becomes 0x0A00 before reaching there.

I think it's reasonable to copy production for now. But doing so makes the code more mysterious than necessary so I added some comments into the relevant code. There is definitely something more to be done about it later, though.

Bilka2:
Hm, since this dive into this newline code started with the change to HTML5 in add_paragraphs_to_text, would undoing that change solve all these issues (which now have tests) without any new workarounds? If yes, that might be the better solution than adding workarounds to workarounds

potpotkettle:
> Hm, since this dive into this newline code started with the change to HTML5 in add_paragraphs_to_text, would undoing that change solve all these issues (which now have tests) without any new workarounds?

Thanks for the suggestion, this is my attempt. https://github.com/otwcode/otwarchive/pull/4303/commits/e5910dd3cc03d0fc61bb344ba3b4a058116def06 It looks like the answer is no unfortunately.

The additional conversion of nbsp is not too bad in my opinion, in the sense that it keeps the existing behavior, whether the behavior is truly ideal or not (which can be examined later separately). I added some more comment.

Bilka2:
I noticed that this also fixes https://otwarchive.atlassian.net/browse/AO3-3282 so I've added the issue to the PR title and assigned it to you on Jira so we don't forget to test it!

Bilka2:
@potpotkettle pinging you with some positivity: Many thanks from users for fixing AO3-3282 in the course of this work! There are a lot of positive comments about it on the [current release notes](https://archiveofourown.org/admin_posts/33176). And from the AD&T side, once again thank you for being willing to pick this up again and finish it off after we left you without a review for so long :heart:

potpotkettle:
@Bilka2 I'm glad that it worked out in the end. Your review feedback helped a lot, too!

PR #4312, opened 2022-08-04
AO3-5535 Remove series index

Bilka2:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5535

## Purpose

Remove series index, so accessing `/series` redirects to the homepage and gives a red flash error that asks, "Whose series did you want to see?"

## Notes

As visible in the tests, trying to access the series of a user that does not exist like `/users/invalid_user/series` also redirects.
Previously, this page would throw a `ActiveRecord::RecordNotFound` error and 404. The behaviour I implemented here mirrors that of gift_controller, which also redirects for an invalid user name.

Trying to access the series of an invalid pseud (`users/valid_user/pseuds/invalid_pseud/series`) does not redirect. I left it with the original behaviour because I don't know if it should be changed, since gift_controller doesn't handle pseuds. However, I think it would make sense if this also redirected.

## References

I referenced the gift_controller for code and behaviour.

## Credit

Bilka (he/him)

### DIFF
--- a/app/controllers/series_controller.rb
+++ b/app/controllers/series_controller.rb
@@ -16,17 +16,15 @@ class SeriesController < ApplicationController
   # GET /series
   # GET /series.xml
   def index
-    @user = User.find_by!(login: params[:user_id]) if params[:user_id]
-    unless @user
+    unless params[:user_id]
       flash[:error] = ts("Whose series did you want to see?")
       redirect_to(root_path) and return
     end
-
+    @user = User.find_by!(login: params[:user_id])
     @page_subtitle = ts("%{username} - Series", username: @user.login)
     pseuds = @user.pseuds
     if params[:pseud_id]
-      @pseud = @user.pseuds.find_by!(name: params[:pseud_id])
-      
+      @pseud = @user.pseuds.find_by!(name: params[:pseud_id])   
       @page_subtitle = ts("by ") + @pseud.byline
       pseuds = [@pseud]
     end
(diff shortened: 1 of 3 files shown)


app/controllers/series_controller.rb
```diff
@@ -16,21 +16,20 @@ def load_series
...
+      redirect_to(root_path) and return
+    end
+
+    @page_subtitle = ts("%{username} - Series", username: @user.login)
+    pseuds = @user.pseuds
+    if params[:pseud_id]
+      @pseud = @user.pseuds.find_by(name: params[:pseud_id])
+      raise ActiveRecord::RecordNotFound, "Couldn't find pseud '#{params[:pseud_id]}'" unless @pseud
+      
+      @page_subtitle = ts("by ") + @pseud.byline
```
Bilka2, line 31: This page_subtitle is only "by pseud (user)", it should probably include the word "series" somewhere, but that is outside the scope of this PR. Is the [support form](https://archiveofourown.org/support) the right place to report issues like this?

sarken: Here is fine! I've made [an issue](https://otwarchive.atlassian.net/browse/AO3-6378).

sarken (changes requested):
Thanks for bearing with us! I had to do some research to find out what our actual desired behavior was for instances where the user or pseud wasn't found. While we do want to redirect when no user params are provided, we'd like to give 404 errors when the user or pseud is nonexistent -- that gives the person who entered the URL a chance to fix any typos. 

(The gifts behavior is incorrect and I'll have to make an issue for updating that.)

sarken (approved):
Thank you!

PR #4317, opened 2022-08-09
AO3-2426 Allow admin to change the language of an external work

potpotkettle:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-2426

## Purpose

External works have language ID which can only be assigned when creating it. This set of changes will add UI elements to allow admins to change the language ID of an external work.

## Testing Instructions

See the Jira issue.

## References

#197

## Credit

Potpotkettle (they/them)

### DIFF
--- a/app/views/external_works/_blurb.html.erb
+++ b/app/views/external_works/_blurb.html.erb
@@ -14,6 +14,11 @@
       <% fandoms = external_work.tag_groups["Fandom"] %>
       <%= fandoms.collect{ |tag| link_to_tag_works(tag) }.join(", ").html_safe if fandoms %>
     </h5>
+    <% unless external_work.language.blank? %>
+      <h5 class="language heading">
+        <%= ts("Language:") %> <%= external_work.language.name %>
+      </h5>
+    <% end %>
 
     <p class="notice"><%= ts("This work isn't hosted on the Archive so this blurb might not be complete or accurate.") %></p>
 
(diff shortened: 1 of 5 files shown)


potpotkettle:
Here is a screenshot.
image

app/views/external_works/_blurb.html.erb
```diff
@@ -14,6 +14,11 @@
       <% fandoms = external_work.tag_groups["Fandom"] %>
       <%= fandoms.collect{ |tag| link_to_tag_works(tag) }.join(", ").html_safe if fandoms %>
     </h5>
+    <% unless external_work.language.blank? %>
```
sarken, line 17: Just a drive-by comment: this should probably be at the start of the stats `dl`, since that's where language is on work blurbs.

potpotkettle: The best I come up with will look like this. Does it look fine, or should language be in a separate row from the rest of stats?

Screenshot 2022-08-15 at 17-57-13 (development) Show External Work Example Archive

Its ERB code looks like this:
```
  <!--stats-->
  <dl class="stats">
    <% unless external_work.language.blank? %>
      <dt class="language"><%= ts("Language:") %></dt>
      <dd class="language"><%= external_work.language.name %></dd>
    <% end %>
    <% if Bookmark.count_visible_bookmarks(external_work) > 0 %>
    <dt><%= ts("Bookmarks") + ": " %></dt>
    <dd><%= link_to_bookmarkable_bookmarks(external_work) %></dd>
    <% end %>
```

On second thought, if we are to add language into the external work blurb, we might want to add it to the bookmarks view, too. What do you think?

In fact, I've just realized that the task description doesn't seems to ask to show language anywhere outside of the edit screen. I think it's something helpful to have, but I'm fine with dropping this part, too, if it's unnecessary.

sarken: That's perfect! If you could add it to the bookmark version as well, I think that would make sense.

sarken (approved):
This looks good -- thanks!

PR #4322, opened 2022-08-20
AO3-6243 Remove condition for comment form to be 'remote' to avoid CSRF token missing error

EchoEkhi:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6243

## Purpose

Fix editing a comment without JavaScript results in a session expired auth error.

## Testing Instructions

Follow instructions on JIRA.

## Credit

EchoEkhi (He/Him/{/([A-Z][a-z]*)/g})


sarken:
Thank you for the pull request! However, I'm going to temporarily close this, since it looks like you're currently over the limit of 5 open PRs. If you could pick another of your Awaiting Review PRs to close until you've had some PRs merged, we'd appreciate it.

EchoEkhi:
Could you please re-open this? I'm not allowed to re-open it apparently.

app/views/comments/_comment_form.html.erb
```diff
@@ -3,7 +3,7 @@
   <% commentable = @commentable %>
 <% end %>
 <div class="post comment" id="comment_form_for_<%= commentable.id %>">
-  <%= form_for value_for_comment_form(commentable, comment), :remote => !comment.new_record?, :html => {:id => "comment_for_#{commentable.id}"} do |f| %>
+  <%= form_for value_for_comment_form(commentable, comment), :html => {:id => "comment_for_#{commentable.id}"} do |f| %>
```
sarken, line 6: Can you update the syntax here to `html: { id: "comment_for_#{commentable.id}" }`? Then this should be ready to go!

sarken (changes requested):
Just one little syntax change -- the actual fix looks good!

sarken:
I'm not sure _why_, since you didn't touch the locale files, but it looks like you'll need to run `bundle exec i18n-tasks normalize -l en` to make the tests pass on this.

tickinginstant:
I'm not convinced that this approach to fixing the issue is correct, since I believe this disables the remote form in all cases, even when editing a comment with JS enabled (which currently uses JS with a remote form). You might want to look into the `authenticity_token` argument on form helpers to try to resolve the CSRF error without otherwise changing the functionality.

If this approach is correct and the goal is to make it so that updating a comment never uses JS, even when JS is enabled, then you should probably also delete the code for updating a comment using JS, rather than leave the newly unused code lying around:
https://github.com/otwcode/otwarchive/blob/25150893db066858d83609cc52f36abf288d5071/app/controllers/comments_controller.rb#L376
https://github.com/otwcode/otwarchive/blob/25150893db066858d83609cc52f36abf288d5071/app/views/comments/update.js.erb#L1-L11
It would also be good to add some additional QA steps to cover the change in functionality.

sarken:
Thanks for catching that, ticking! I didn't notice the change in behavior at all when I pulled this and tried it out.

sarken (approved):

PR #4324, opened 2022-08-23
AO3-6365 Add interface for muting users via CSS

EchoEkhi:
Just so everybody knows, this pull request is *way* above my skill level, so if anybody wants to submit a competing PR on this issue, that's more than welcome by me.

# Pull Request Checklist

## To Do List
* [x] Change strings
* [x] Add tests
* [x] Add tests to check muted items are not visible by CSS
* [x] Cache
* [x] Add button to dashboards

## Issue

https://otwarchive.atlassian.net/browse/AO3-6365

## Purpose

Add interface for muting users via CSS

## Testing Instructions

See JIRA ticket.

## Credit

EchoEkhi (He/Him/{/([A-Z][a-z]*)/g}), tickinginstant (for making the blocking code which I copied from)


app/views/muted/_muted_items_notice.html.erb
```diff
@@ -0,0 +1,3 @@
+<% if user_has_muted_users? %>
+  <p><%= t('muted.muted_items_notice', muted_users_link: link_to(t("muted.muted_users_link_text"), user_muted_users_path(current_user))).html_safe %></p>
+<% end %>
```
sarken, line 3: Can you add the missing newline at the end of the file here, and all use double quotes around the translation key name?

sarken (changes requested):
Thank you for working on this! Other folks might have some comments as well, but here are some to get started.

app/helpers/mute_helper.rb
```diff
@@ -0,0 +1,34 @@
...
+      link_to(t("muted.unmute"), confirm_unmute_user_muted_user_path(muting_user, mute))
+    else
+      link_to(t("muted.mute"), confirm_mute_user_muted_users_path(muting_user, muted_id: user))
+    end
+  end
+
+  def mute_css
+    return unless current_user
+
+    Rails.cache.fetch("muted/#{current_user.id}/mute_css") do
```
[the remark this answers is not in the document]
EchoEkhi: I think this is in case of a cache miss. If the cache has been expired or reset, it is regenerated and re-cached in the do block.

app/helpers/mute_helper.rb
```diff
@@ -0,0 +1,34 @@
...
+      muted_users_css_classes = current_user.muted_users.map { |muted_user| ".user-#{muted_user.id}" }
+      
+      "<style>#{muted_users_css_classes.join(', ')} {display: none !important; visibility: hidden !important;}</style>".html_safe
+    end
+  end
+
+  def user_has_muted_users?
+    return unless current_user
+
+    !Rails.cache.fetch("muted/#{current_user.id}/mute_css").nil?
```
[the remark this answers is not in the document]
EchoEkhi: This is checked every time the "items muted" notice is meant to be displayed, which is every work/series/etc search results page. That's a lot of SQL requests for something that doesn't change very often.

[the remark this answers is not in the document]
sarken: That's okay. We can handle them.

sarken (changes requested):
We'd also like to cap mutes at 1000 per user (we're going to go back and cap blocks as well). The number should be in config/config.yml so it's easy to change.

app/models/mute.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  belongs_to :muter, class_name: "User"
+  belongs_to :muted, class_name: "User"
+
+  validates :muter, :muted, presence: true
+  validates :muted_id, uniqueness: { scope: :muter_id }
+
+  validate :check_self
+  validate :check_official, if: :muted
+  validate :check_mute_limit
```
sarken, line 12: Can you add a test for this? You can temporarily lower the limit; no need to actually make a thousand muted users.

https://github.com/otwcode/otwarchive/blob/09d8ba39c8ec82c070a774bfd4ea479c2ee20c3c/features/step_definitions/kudos_steps.rb#L16-L18

features/users/mute.feature
```diff
@@ -0,0 +1,152 @@
+Feature: Muting
```
sarken, line 1: I just noticed the blocking feature is named blocking -- can you change this to muting to match?

config/locales/views/en.yml
```diff
@@ -372,9 +378,62 @@ en:
...
+        title: "Muted Users"
+        button: "Mute"
+        none: "You have not muted any users."
+        legend: "Mute a user"
+        label: "User"
+        heading:
+          landmark:
+            muted_users: "Listing Muted Users"
+        will:
+          intro: "Muting a user:"
```
sarken, line 397: PAC and Support would like one bit of text added here (just on the index, not the confirmation page). They'd like it to mention the limit and say: "You can mute up to [limit, with delimiter] users. Muting a user:"

sarken (changes requested):
Once again, thank you so much! I think we really are down to the final details now. 🤞

sarken (changes requested):

app/models/mute.rb
```diff
@@ -0,0 +1,37 @@
...
+  def check_self
+    errors.add(:muted, :self) if muted == muter
+  end
+
+  def check_official
+    errors.add(:muted, :official) if muted.official
+  end
+
+  def check_mute_limit
+    errors.add(:muted, :limit) if muter.muted_users.length >= ArchiveConfig.MAX_MUTED_USERS
```
tickinginstant, line 26: `length` is not a good function to use here, as it'll force all of the muted users to be loaded. `count` is probably the right choice.

app/views/blocked/users/confirm_block.html.erb
```diff
@@ -19,6 +19,8 @@
       <li><%= t(".will_not.comments_on_works") %></li>
       <li><%= t(".will_not.comments_elsewhere") %></li>
     </ul>
+
+    <p><%= t(".mute_users_instead", muted_users_link: link_to(t(".muted_users_link_text"), user_muted_users_path(@user))).html_safe %></p>
```
tickinginstant, line 23: Rather than explicitly calling `.html_safe` on the result of the call to `t`, would you mind renaming the translation key to end in `_html` and dropping the `.html_safe` call? Here and everywhere else.

(In views, using a translation key that ends in `_html` or `.html` will [automatically handle marking as html-safe](https://guides.rubyonrails.org/i18n.html#using-safe-html-translations). That's why the `.sure_html` translation above doesn't need a call to `html_safe`, even though it does contain a `<strong>` tag.)

sarken (approved):

app/models/user.rb
```diff
@@ -69,6 +69,22 @@ class User < ApplicationRecord
...
 
+  has_many :mutes_as_muted, class_name: "Mute", dependent: :delete_all, inverse_of: :muted, foreign_key: :muted_id
+  has_many :mutes_as_muter, class_name: "Mute", dependent: :delete_all, inverse_of: :muter, foreign_key: :muter_id
+  has_many :muted_users, through: :mutes_as_muter, source: :muted
+
+  # The mute (if it exists) with this user as the muter and
+  # User.current_user as the muted:
+  has_one :mute_of_current_user,
+          -> { where(muted: User.current_user) },
+          class_name: "Mute", foreign_key: :muter_id, inverse_of: :muter
```
tickinginstant, line 80: This doesn't appear to be used. Can you remove it?

app/views/blocked/users/index.html.erb
```diff
@@ -1,5 +1,12 @@
 <h2 class="heading"><%= t(".title") %></h2>
 
+<!--subnav-->
+<ul class="navigation actions" role="navigation">
+  <h3 class="landmark heading"><%= t("a11y.navigation") %></h3>
+  <li><%= span_if_current t(".blocked_users"), user_blocked_users_path %></li>
+  <li><%= span_if_current t(".muted_users"), user_muted_users_path %></li>
```
tickinginstant, line 7: The translation keys for the text of these links (and the similar ones in the muted users index) don't appear to be defined in `config/locales/views/en.yml`. I'm pretty sure they're just using the default translation (replacing underscores with spaces and capitalizing). Have you tried hovering over them in development to see if they have the "translation missing" markup?

It would be good to either define the text associated with these keys, or use different keys that are already defined. I'm not sure which is preferred in this case.

app/views/muted/users/confirm_mute.html.erb
```diff
@@ -0,0 +1,34 @@
...
+      <li><%= t(".will_not.prevent_emails") %></li>
+      <li><%= t(".will_not.hide_content_for_others") %></li>
+    </ul>
+
+    <p>
+      <%= t(".block_users_instead_html", blocked_users_link: link_to(t(".blocked_users_link_text"), user_blocked_users_path(@user))) %>
+    </p>
+
+    <p>
+      <%= t(".site_skin_warning", restore_site_skin_faq_link: link_to(t(".restore_site_skin_faq_link_test"), archive_faq_url("skins-and-archive-interface", anchor: :restoresiteskin))).html_safe %>
```
tickinginstant, line 26: There's one more call to `.html_safe` that you can get rid of here (and below in the muted users index).

PR #4331, opened 2022-09-05
AO3-6389 Stop using Dreamwidth URLs in external work tests.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6389

## Purpose

- Use example.org instead of dreamwidth.org in external work tests.
- Use `WebMock` to stub the URLs to prevent connection errors.

## Testing Instructions

Make sure the automated tests pass.


sarken (approved):
omg, thank you <3

PR #4335, opened 2022-09-11
AO3-6394 Restrict admin access to banned emails (formerly blacklist)

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6394

## Purpose

* Updates user-facing text to refer to banned emails rather than blacklisted emails
* Restricts access to banned email to admins with the superadmin, policy_and_abuse, and support roles
* Removes an unused action from the controller

## Testing Instructions

Refer to Jira

## References

[AO3-5982 Rename "Blacklisted Emails" to "Banned Emails"
](https://otwarchive.atlassian.net/browse/AO3-5982) is where I grabbed the text from and can now be just for renaming the code side of things

## Credit

Sarken, she/her


features/admins/admin_email_blacklist.feature
```diff
@@ -2,35 +2,35 @@
...
-  When I follow "Blacklist"
-  Then I should see "Find blacklisted emails"
-    And I should see "Add email address to blacklist"
+Scenario: Ban email address
+  Given I am logged in as a "policy_and_abuse" admin
+  Then I should see "Banned Emails"
+  When I follow "Banned Emails"
+  Then I should see "Find banned emails"
+    And I should see "Ban email address"
     And I should see "Email"
```
tickinginstant, line 13: Since you're changing the `I have banned the address {string}` step below to fill in the "Email to ban" field instead of just "Email," would you mind changing it here as well (and on the line below)? It's not necessary if you're too busy to fix it, but since both of the forms on that page have the word "Email" in them, I think it'd be nice to keep it unambiguous.

tickinginstant (approved):
One minor (optional) nitpick, but otherwise it looks good!

PR #4340, opened 2022-09-14
AO3-6381 Restrict admin access to spam page

Cesium-Ice:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6381

## Purpose

What does this PR do?
Restrict admin access to spam

## Testing Instructions

Description 
Log in as an admin with the policy_and_abuse or superadmin role

You should see Spam in the admin navigation and it should take you to https://test.archiveofourown.org/admin/spam

You should be able to perform the various marking/unmarking as spam activities on this page as well as the Confirmed Spam and Approved Works pages

 

Log in as an admin without the policy_and_abuse or superadmin role

You should not see Spam in the admin navigation and should not be able to access https://test.archiveofourown.org/admin/spam 

## References

## Credit

Cesium-Ice,  she/her

they/them pronouns.*

### DIFF
--- a/app/policies/spam_policy.rb
+++ b/app/policies/spam_policy.rb
@@ -1,9 +1,9 @@
 class SpamPolicy < ApplicationPolicy
-    MANAGE_SPAM = %w[superadmin policy_and_abuse].freeze
+  MANAGE_SPAM = %w[superadmin policy_and_abuse].freeze
 
-   def index?
-     user_has_roles?(MANAGE_SPAM)
-   end
+  def index?
+    user_has_roles?(MANAGE_SPAM)
+  end
 
-   alias bulk_update? index?
- end
+  alias bulk_update? index?
+end
--- a/app/policies/spam_policy.rb
+++ b/app/policies/spam_policy.rb
@@ -1,9 +1,0 @@
-class SpamPolicy < ApplicationPolicy
-  MANAGE_SPAM = %w[superadmin policy_and_abuse].freeze
-
-  def index?
-    user_has_roles?(MANAGE_SPAM)
-  end
-
-  alias bulk_update? index?
-end
--- a/app/policies/spam_policy.rb
+++ b/app/policies/spam_policy.rb
@@ -1,0 +1,9 @@
+class SpamPolicy < ApplicationPolicy
+    MANAGE_SPAM = %w[superadmin policy_and_abuse].freeze
+
+   def index?
+     user_has_roles?(MANAGE_SPAM)
+   end
+
+   alias bulk_update? index?
+ end
--- a/app/views/admin/_header.html.erb
+++ b/app/views/admin/_header.html.erb
@@ -36,9 +36,9 @@
     <li><%= link_to ts("Banned Emails", key: "header"), admin_blacklisted_emails_path %></li>
   <% end %>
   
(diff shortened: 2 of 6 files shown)


app/policies/spam_policy.rb
```diff
@@ -0,0 +1,9 @@
+class SpamPolicy < ApplicationPolicy
```
sarken, line 1: To follow up from the other day, we've decided we'd prefer to call this `ModeratedWorkPolicy` for clarity.

app/views/admin/_header.html.erb
```diff
@@ -35,7 +35,11 @@
   <% if policy(AdminBlacklistedEmail).index? %>
     <li><%= link_to ts("Banned Emails", key: "header"), admin_blacklisted_emails_path %></li>
   <% end %>
-  <li><%= link_to ts("Spam", key: "header"), admin_spam_index_path %></li>
+  
+  <% if policy(:spam).index? %>
+     <li><%= link_to ts("Spam", key: "header"), admin_spam_index_path %></li>
+   <% end %>
```
sarken, line 41: I think the indenting acquired some extra spaces on lines 40 and 41

sarken (changes requested):

app/policies/spam_policy.rb
Cesium-Ice: Do I need to replace all instances of “spam” with “moderatedwork” then?

sarken: Pretty much, yup! (You should use the class `ModeratedWork` where you're currently using the symbol `:spam`, but for renaming the policy file and the `MANAGE_SPAM` constant, you should use `moderated_work_policy.rb` and `MANAGE_MODERATED_WORK` instead.)

app/views/admin/_header.html.erb
Cesium-Ice: fixed

app/policies/spam_policy.rb
Cesium-Ice: I guess spam_controller should also be moderated_work_controller, then?
And "Spam" in the admin banner should be "Moderated Works" instead?

sarken: No, those can stay the same. We're just changing the policy because Pundit policies typically correspond with existing classes.

spec/controllers/admin/spam_controller_spec.rb
```diff
@@ -18,21 +16,29 @@
...
-    context "when logged in as admin without correct authorization" do
-      xit "redirects with notice" do
-        fake_login_admin(admin)
-        get :index, params: { reviewed: false, approved: false }
+    %w[board communications translation tag_wrangling docs support open_doors].each do |role|
+      context "when logged in as an admin with #{role} role" do
+        let(:admin) { create(:admin, roles: [role]) }
 
-        it_redirects_to_with_error(root_url, "Sorry, only an authorized admin can access the page you were trying to reach.")
+        it "redirects with notice" do
```
sarken, line 23: Just one last thing! It should probably say `redirects with error` here and on line 66.

Cesium-Ice: done

sarken (approved):

PR #4348, opened 2022-09-21
AO3-6118 Require opt-in for work collection invitations

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6118

## Purpose

* Remove automatic approvals for when a work is invited to a collection.
* Add a preference to opt works in to collection invitations.

## Testing Instructions

Preliminary steps listed in a Jira comment.

## References

https://github.com/otwcode/otwarchive/pull/4348#discussion_r979325697

> As I mentioned above, I think the intention is to not change the behavior of the "Edit Work" form, so it's not an issue if the work's owners can't use the "Add to Collections" form.

Modify the "Add to Collections" form so that only collection maintainers can use it, and rename it to "Invite to Collections."

## Credit

Brian (they/he)


spec/helpers/application_helper_spec.rb
```diff
@@ -336,30 +356,44 @@
...
+          travel_to(1.day.ago) do
+            original_cache_key = "#{work.cache_key_with_version}/blurb_css_classes-v2"
+            expect(helper.css_classes_for_creation_blurb(work)).to eq("#{default_classes} work-#{work.id} user-#{user1.id}")
+          end
 
-          travel_back
-          work.collections << collection
-          expect(helper.css_classes_for_creation_blurb(work)).to eq("#{default_classes} work-#{work.id}")
-          expect(original_cache_key).not_to eq("#{work.cache_key_with_version}/blurb_css_classes-v2")
+          travel_to(1.second.from_now) do
```
brianjaustin, line 370: I'm not really sure why this became necessary to get the test passing; I suspect it has something to do with how `in_unrevealed_collection?` gets updated, but I couldn't find exactly where it happens. Just wanted to flag this in case it's a sign of a bigger issue.

public/help/collection-preferences.html
```diff
@@ -2,8 +2,8 @@ <h4>Collections, Challenges and Gifts Preferences</h4>
 
 <dl id="help">
   
-  <dt>Automatically agree to your work being collected by others in the Archive</dt>
-  <dd>By default, if another user adds your work to a collection, you will be asked to confirm whether you are willing for your work to be displayed as part of that collection. If you'd like people to be able to add your work to collections without waiting for confirmation from you, enable this option.</dd>
+  <dt>Allow others to invite my works to collections</dt>
+  <dd>By default, other users may not add your work to a collection. If you'd like people to be able to add your work to collections, enable this option. You will also need to approve invitations to a collection.</dd>
```
brianjaustin, line 6: @sarken the ticket mentioned potential new wording from the Docs team; I just made this up, so maybe someone from there would want to review it?

app/models/collection_item.rb
```diff
@@ -124,20 +130,7 @@ def approve_automatically
...
-            approve_by_user
-            users.each do |email_user|
-              unless email_user.preference.collection_emails_off
-                UserMailer.added_to_collection_notification(email_user.id, item.id, collection.id).deliver_now
-              end
-            end
-            break
-          end
-        end
+        approve_by_user if users == [User.current_user]
```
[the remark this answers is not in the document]
tickinginstant: Bookmarks are handled by the `approve(User.current_user)` check above, so you can delete the entire `if !approved_by_user?` block.

app/models/collection_item.rb
```diff
@@ -34,6 +34,13 @@ def collection_is_open
     end
   end
 
+  validate :creators_allow_invitation, on: :create
+  def creators_allow_invitation
+    return unless item.respond_to?(:allow_collection_invitation)
+
+    errors.add(:base, :creator_invitation_not_allowed) unless item.allow_collection_invitation
+  end
```
tickinginstant, line 42: I don't think it's a good idea to put this check in the model, particularly with the way you've defined `Work#allow_collection_invitation`.

Currently, works can have collection items created in two different controllers: the `CollectionItemsController` (the "Add to Collection" form), and the `WorksController` (the "Edit Work" form). From the description of the issue, I think the intention is that the behavior in the "Edit Work" form should be unchanged, and this PR should affect only the behavior of the `CollectionItemsController`:
> When none of the creators of a work has checked "Allow others to invite my works to collections.", the "Add To Collections" button should be removed from the work. (**Creators can always edit the work itself to add it to a collection.**) 

(Emphasis mine.)

spec/helpers/application_helper_spec.rb
tickinginstant: Variables defined with `let` are lazily evaluated. In the original version of this test, the first time `work` was referenced was in the `travel_to(1.day.ago)` block, meaning that the work got created (and most recently updated) 1 day ago. That meant that when the work was updated again in the present, its update date changed, and the cache on its CSS classes expired.

By adding a `before` hook that references `work`, you're causing the work to be created before the `travel_to(1.day.ago)` hook, making the work be created in the present. This makes the first `travel_to` block a no-op, and means that if the test runs fast enough you're just creating the work and then updating it all in the same second, and nothing gets expired.

(Luckily, you should be able to revert almost all of the spec changes once you pull the validation out of the model.)

public/help/collection-preferences.html
sarken: Checking with them now!

app/models/work.rb
```diff
@@ -1230,4 +1230,9 @@ def nonfiction
     nonfiction_tags = [125773, 66586, 123921, 747397] # Essays, Nonfiction, Reviews, Reference
     (filter_ids & nonfiction_tags).present?
   end
+
+  # If the creator(s) allow this work to be invited to collections.
+  def allow_collection_invitation
+    users.all? { |user| user == User.current_user || user.preference.allow_collection_invitation }
```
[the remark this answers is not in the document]
tickinginstant: As I mentioned above, I think the intention is to not change the behavior of the "Edit Work" form, so it's not an issue if the work's owners can't use the "Add to Collections" form. That said, I think there are a few different ways to approach the behavior of the "Add to Collections" form for work owners:

1. Make it so that the creator of the work can always use the "Add to Collections" form to add their own work to collections, regardless of the number of co-creators or their preferences. This means that the "Add to Collections" form is as permissive as the "Edit Work" page for work owners.
2. Make it so that the creator of the work can only use the "Add to Collections" form if at least one of the work's creators allows collection invitations. I think this approach most closely matches the issue description, but I'm not sure if that's the intended approach.
3. Modify the "Add to Collections" form so that only collection maintainers can use it, and rename it to "Invite to Collections."

My personal preference is for option 3, because I feel like making the form available to both work owners and collection maintainers with the same wording used for both makes it more confusing. But you should probably get someone else's feedback on the desired behavior.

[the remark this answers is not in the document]
sarken: I like 3 as well, but I will run it by Support and make sure they don't have any objections.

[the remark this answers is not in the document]
sarken: Support is also good with 3!

public/help/collection-preferences.html
sarken: Here we go:

> Enabling this option will allow other AO3 users to invite your work into their collection. Your work won't be added to the collection until you have accepted the request. If you have this option enabled, you can find any collection invitations in your Manage Collected Works page. You will also receive an email notifying you of the request.
>
> Leaving this option disabled will prevent other users from inviting your works into their collections at all, and you won't receive any notifications.
>
> Changing this setting won't affect any existing works in collections.

If there are any inaccuracies in that where the changes may have deviated from the Jira issue, please lmk and I'll take it back to Docs for edits.

sarken: ...instead, Docs has edited their own text. 😆 Sorry about that. This should be the text:

>Enabling this option will allow other AO3 users to invite your work into their collection. Your work won't be added to the collection until you have accepted the request. For more information on accepting collection invitations, refer to <a href="/faq/collections#collectionitems>How do I approve or reject an invitation to include my work in a collection?</a>
>
>Leaving this option disabled will prevent other users from inviting your works into their collections at all, and you won't receive any notifications.
>
>Changing this setting won't affect any existing works in collections.

brianjaustin: No worries! I did my version as a placeholder so I didn't forget to ask you folks. Here's what that text looks on my local setup; does anything seem out of place here?

updated-collections-preferences-docs

app/models/collection_item.rb
brianjaustin: You know, I think my brain just skipped over that parenthetical completely... thanks for calling me out and the push in the right direction!

spec/helpers/application_helper_spec.rb
brianjaustin: Thank you so much for the explanation!

public/help/collection-preferences.html
sarken: I think it needs paragraph tags, since Docs intended it to be three separate paragraphs, as it is above.

brianjaustin: Sure thing! In terms of appearance, do you have a preference for either of these?

Extra line breaks:
Collection invite settings help with line breaks

Just `<p>` tags
Collection invite settings help with no line breaks

I'm leaning more towards the top one, but the line breaks seem a bit large. I suppose I could tinker with the CSS to make them smaller (or add margin between `<p>`s) if you would prefer that too.

app/models/work.rb
```diff
@@ -1230,4 +1230,9 @@ def nonfiction
     nonfiction_tags = [125773, 66586, 123921, 747397] # Essays, Nonfiction, Reviews, Reference
     (filter_ids & nonfiction_tags).present?
   end
+
+  # If the creator(s) allow this work to be invited to collections.
+  def allow_collection_invitation?
+    users.all? { |user| user.preference.allow_collection_invitation }
```
Bilka2, line 1236: This likely needs an official opinion, but I noticed that the Jira issue says
> When none of the creators of a work has checked "Allow others to invite my works to collections.", the "Add To Collections" button should be removed from the work. 

Which I would read as "show if at least one person allows it", so in this case I think it should be `any?` instead of `all?`.

features/step_definitions/work_steps.rb
```diff
@@ -709,6 +709,12 @@
   end
 end
 
+When "I set the collections of work {string} to {string}" do |work, collection|
```
redsummernight, line 712: "I edit the work X to be in the collection Y"? Let's move this over to collection_steps.rb as well, next to the invite version.

features/works/work_view.feature
```diff
@@ -42,3 +42,20 @@ Feature: View a work with various options
...
+  When I am logged in as "moderator"
+    And I view the work "Whatever"
+  Then I should not see a link "Invite To Collections"
+
+  Scenario: other users can collect a work when the creator has opted-in
+  Given the work "Whatever"
+    And I am logged in as the author of "Whatever"
+    And I set my preferences to allow collection invitations
+    And I have the collection "test collection" with name "test_collection"
+  When I am logged in as "moderator"
```
redsummernight, line 59: Let's not use "moderator" here, since this is just a random user with no ties to the collection.

public/help/collection-preferences.html
redsummernight: Just paragraph tags inside `<dd>` should be fine, similar to this existing help:

https://github.com/otwcode/otwarchive/blob/2d92ec1e100d2db8d98c791258b8ea115044975d/public/help/privacy-preferences.html#L28

We can deal with the spacing separately.

app/models/work.rb
redsummernight: Yeah, this should be `any?`. Good catch!

brianjaustin: Ok! I will update this. I think I went `all?` originall because it seemed odd to have one creator's opt-in override their co-creator's (or creators') opt-out -- or at least lack of opt-in -- but I guess that can be solved with good communication :+1:

features/step_definitions/work_steps.rb
```diff
@@ -295,8 +295,8 @@
   step %{I edit the work "#{draft}"}
 end
 
-When /^I post the chaptered work "([^"]*)"$/ do |title|
-  step %{I post the work "#{title}"}
+When /^I post the chaptered work "([^"]*)"(?: (?:in|to) (?:the )?collection "([^"]*)")?$/ do |title, collection|
```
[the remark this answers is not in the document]
brianjaustin: Oh thank goodness, that regex was getting wild. Done!

features/step_definitions/work_steps.rb
brianjaustin: Thanks for the better words! I've updated that step, and I also added an optional `s` to collections since some features are adding more than one

features/works/work_view.feature
brianjaustin: I might be confusing things, but I think the user _is_ tied to the collection (and needs to be):
* The `And I have a collection` step creates a collection belonging to `moderator`: https://github.com/otwcode/otwarchive/blob/master/features/step_definitions/collection_steps.rb#L65
* The `Invite to Collections` link should only be visible if the current user is a collection maintainer: https://github.com/brianjaustin/otwarchive/blob/AO3-6118-collection-invitation-opt-in/app/views/comments/_commentable.html.erb#L40

If that's right, then the wording/order is a bit weird. Maybe the following would be slightly better?
```cucumber
    And I am logged in as the author of "Whatever"
    And I set my preferences to allow collection invitations
  When I have the collection "test collection" with name "test_collection"
    And I am logged in as "moderator"
```

redsummernight: Sounds good!

app/models/work.rb
redsummernight: I think we went with `any?` to be consistent with how the `disable_share_links?` preference works: opting out only if everyone opts out.

https://github.com/otwcode/otwarchive/blob/2d92ec1e100d2db8d98c791258b8ea115044975d/app/views/works/_work_header_navigation.html.erb#L99-L103

We'll see if this (and/or the FAQ text) needs tweaking further.

redsummernight (approved):
Looks good to me, thanks!

I'll set the ready label once we hear from a second reviewer (@sarken, @tickinginstant).

app/views/collection_items/_collection_item_form.html.erb
```diff
@@ -2,7 +2,11 @@
 <% @collection_item ||= CollectionItem.new %>
 <% in_page ||= false %>
 <div id="collection-form">
-  <h3 class="heading"><%= ts("Add %{title} to collections", :title => item.respond_to?(:title) ? item.title : item.class.name).html_safe %></h3>
+  <% if item.is_a?(Work) %>
+    <h3 class="heading"><%= t(".invite_header", title: item.title).html_safe %></h3>
```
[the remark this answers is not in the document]
tickinginstant: You should be able to fix this by changing the translation key to `.invite_header_html`, and dropping the `.html_safe` at the end.

app/views/collection_items/_collection_item_form.html.erb
```diff
@@ -2,7 +2,11 @@
...
+  <% else %>
+    <h3 class="heading"><%= t(".add_header", title: item.class.name) %></h3>
```
tickinginstant, line 8: I think the Translation Committee [doesn't generally like passing class names](https://otwarchive.atlassian.net/browse/AO3-5867) like "Bookmark" as interpolated variables. But since this is only for bookmarks at the moment, you can probably just specialize `.add_header` to be for bookmarks, instead of having `".add_header_#{item.model_name.i18n_key}"` as the translation key.

tickinginstant (commented):
I have a couple of minor quibbles, but it looks good otherwise!

app/views/collection_items/_collection_item_form.html.erb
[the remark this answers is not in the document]
brianjaustin: :facepalm: I totally forgot about this, even though I have been using it in other PRs. Thanks for reminding me it exists, ticking!!

sarken:
Hey, Brian! Do you have a few minutes to resolve the merge conflict on this?

brianjaustin:
@sarken all set!

redsummernight:
Thanks!

PR #4349, opened 2022-09-22
AO3-6403 Fix prompt meme date test.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6403

## Purpose

There's a prompt meme test that assumes that September 20, 2022 is in the future. That's no longer true, so the test is failing. This PR uses time travel to make the test work again.

## Testing Instructions

No QA required as long as this passes the automated tests.


brianjaustin (approved):
Thanks! I was actually just running into this on my PR and about to mention it :laughing:

sarken (approved):

PR #4353, opened 2022-09-25
AO3-5837 Remove unused code from bookmarks controller

designateddolt:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5837 

## Purpose

What does this PR do?
As per Jira issue

## Testing Instructions

As per Jira issue

## References

## Credit


designateddolt:
Can someone please let me know if the RSpec tests i've modified are still relevant/checking what we actually want them to check? I deleted one that tested solely this branch of the code, and modified two others to use work id rather than chapter id, but I just want to make sure that's alright to do for those.

spec/controllers/bookmarks_controller_spec.rb
```diff
@@ -28,7 +28,7 @@ def it_redirects_to_user_login
 
         it "renders the new form for a multi-chapter work" do
           fake_login
-          get :new, params: { chapter_id: chapter2.id }
+          get :new, params: { work_id: chaptered_work.id }
```
sarken, line 31: Hmm. I _thought_ what we were testing here would be the new bookmark path for a chapter (`https://archiveofourown.org/chapters/000/bookmarks/new`), but that gives a 404 even on production. 

However, it looks like it was actually the new bookmark path with chapter parameters (`https://archiveofourown.org/bookmarks/new?chapter_id=000`), which works on production, but gives a 500 error on development when these code changes are made.

I'm pretty sure we can remove this test because I can't think of anywhere we use `bookmarks/new?some_param=000`. 

In fact, we don't seem to use `new_bookmark_path` anywhere, so I think we can get rid of the other test you modified as well. What we use to take people to the new bookmark form seems to be our own `get_new_bookmark_path` helper:
https://github.com/otwcode/otwarchive/blob/de0fa361add04f0be7fed8a91d2651373eeb48d4/app/helpers/bookmarks_helper.rb#L13-L25

redsummernight (approved):

PR #4358, opened 2022-10-15
AO3-5934 Allow <figure> and <figcaption> HTML tags

potpotkettle:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5934

## Purpose

The purpose here is to add `<figure> `and `<figcaption>` into the set of allowed HTML tags. These tags are used to show pictures with captions (and more complex content).

For the expected behavior, see the Jira issue. Some of the aspects of the feature as written in the Jira issue are not satisfied exactly in this implementation (at least in its initial version), though.
- If a user tries to wrap `<figure>` with a `<p>`, they can't. It will result in 2 emtpty `<p>`s, one before `<figure>`, another after it. This may not be ideal, but I don't believe we can do much better. Ultimately, most browsers don't support the structure like `<p><figure>`...`</figure></p>`.
- Bare `<figcaption>` is not stripped, they will be left just as entered by the user. I don't think this stripping feature will be useful enough to justify the additional complexity, so I would suggest going without it.

## Testing Instructions

A simple code to be supported looks like this:

```html
<figure>
<img src="https://upload.wikimedia.org/wikipedia/commons/a/aa/Camera-icon.svg" alt="camera icon">
<figcaption>Camera icon, black and white</figcaption>
</figure>
```

## Credit

- Potpotkettle (they/them)

### DIFF
--- a/config/initializers/gem-plugin_config/sanitizer_config.rb
+++ b/config/initializers/gem-plugin_config/sanitizer_config.rb
@@ -4,12 +4,12 @@ class Sanitize
   # This defines the configuration we use for HTML tags and attributes allowed in the archive.
   module Config
     ARCHIVE = freeze_config(
-      elements: [
-        'a', 'abbr', 'acronym', 'address', 'b', 'big', 'blockquote', 'br', 'caption', 'center', 'cite', 'code', 'col',
-        'colgroup', 'dd', 'del', 'dfn', 'div', 'dl', 'dt', 'em', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'hr',
-        'i', 'img', 'ins', 'kbd', 'li', 'ol', 'p', 'pre', 'q', 's', 'samp', 'small', 'span', 'strike', 'strong',
-        'sub', 'sup', 'table', 'tbody', 'td', 'tfoot', 'th', 'thead', 'tr', 'tt', 'u', 'ul', 'var'],
-
+      elements: %w[
+        a abbr acronym address b big blockquote br caption center cite code col
+        colgroup figcaption figure dd del dfn div dl dt em h1 h2 h3 h4 h5 h6 hr
+        i img ins kbd li ol p pre q s samp small span strike strong
+        sub sup table tbody td tfoot th thead tr tt u ul var
+      ],
       attributes: {
         all: ['align', 'title', 'dir'],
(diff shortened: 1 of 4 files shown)


spec/lib/html_cleaner_spec.rb
```diff
@@ -786,6 +786,39 @@
...
+        </figure>
+      HTML
+
+      result = add_paragraphs_to_text(html)
+      doc = Nokogiri::HTML.fragment(result)
+
+      expect(doc.xpath("./figure/p")).to be_empty
+    end
+
+    it "keeps figure/figcaption and titles" do
```
sarken, line 806: It looks like we're testing two different things here. Maybe break this into two examples:
- `it "allows alt and title attributes on elements inside figure" do`
- `it "allows other HTML elements inside figcaption" do`

config/initializers/gem-plugin_config/sanitizer_config.rb
```diff
@@ -4,12 +4,12 @@ class Sanitize
   # This defines the configuration we use for HTML tags and attributes allowed in the archive.
   module Config
     ARCHIVE = freeze_config(
-      elements: [
-        'a', 'abbr', 'acronym', 'address', 'b', 'big', 'blockquote', 'br', 'caption', 'center', 'cite', 'code', 'col',
-        'colgroup', 'dd', 'del', 'dfn', 'div', 'dl', 'dt', 'em', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'hr',
-        'i', 'img', 'ins', 'kbd', 'li', 'ol', 'p', 'pre', 'q', 's', 'samp', 'small', 'span', 'strike', 'strong',
-        'sub', 'sup', 'table', 'tbody', 'td', 'tfoot', 'th', 'thead', 'tr', 'tt', 'u', 'ul', 'var'],
-
+      elements: %w[
```
sarken, line 7: Much neater -- thanks!

sarken (changes requested):
Looks good! Just a minor suggestion regarding the tests.

spec/lib/html_cleaner_spec.rb
potpotkettle: Sure, I will do that shortly.

sarken (approved):

PR #4365, opened 2022-10-16
AO3-6410 Update the automated PR labeller to handle the .github directory better.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6410

## Purpose

- Make sure to include files that start with a `.` in the labeller checks for the Awaiting Review and Coder Has Actioned Review labels.
- Switch to a different syntax for the Tests Only label to try to improve handling of the `.github` directory.

## Testing Instructions

Once this PR has been merged, set one of the dependabot PRs to "Reviewed: Action Needed," and then tell dependabot to rebase the PR. If it adds the "Coder Has Actioned Review" label and the "Scope: Tests Only" label, that should indicate that the two checks will work on future PRs.

Note that because the PR labeller is set to run as `pull_request_target`, the changes in this PR will not be visible on this PR, only on PRs labelled after this PR is merged.


.github/labeler/contents.yml
```diff
@@ -11,4 +11,4 @@
 
 "Scope: Tests Only":
   - all:
-    - "+(factories|features|spec|test|.github)/**/*"
+    - "{factories,features,spec,test,.github}/**/*"
```
tickinginstant, line 14: When minimatch is converting a glob to a regex, it checks whether the pattern starts with a `.`. If it doesn't start with a dot, then it [adds a negative lookahead to the regex](https://github.com/isaacs/minimatch/blob/6410ef32f59e4842121ca13eefacdf0b3da8533c/minimatch.js#L440-L445) to ensure that the pattern won't match to anything that starts with a dot. This prevents the `+(factories|...|.github)` pattern from matching to anything in the `.github` directory, because the pattern doesn't start with a dot but the `.github` directory does.

Braces, however, are processed before the pattern is turned into a regex. Specifically, the braces above turn the pattern into five separate patterns:
- `factories/**/*`
- `features/**/*`
- `spec/**/*`
- `test/**/*`
- `.github/**/*`

Then each of those patterns is turned into a separate regex, and if any one of those regexes matches, it counts as matching the glob. Because the `.github/**/*` pattern *does* start with a `.`, it doesn't have the special negative lookahead to prevent it from matching dotfiles, so it will happily match to the `.github` directory.

sarken (approved):

sarken:
Do the dots need to be escaped or something? The Scope label worked, but not the Actioned Review.

tickinginstant:
Looks like I screwed up the globs a bit -- I think it needs to be `.*/**/*` instead of `.**/*`, to make sure that it catches everything in all descendant directories of `.github`, and not just files in `.github`. (I was testing with a file at `.github/test` file, so I didn't realize that this glob didn't cover files in the subdirectories, as well.)

PR #4366, opened 2022-10-16
AO3-2228 Add color picker for skin wizard

weeklies:
## Issue

https://otwarchive.atlassian.net/browse/AO3-2228

I've removed the handy list of colors link and the footnotes stating the default value as I believe they are no longer necessary. 

## Testing Instructions

See Jira issue.

## Credit

weeklies (she/her)

### DIFF
--- a/app/views/skins/_wizard_form.html.erb
+++ b/app/views/skins/_wizard_form.html.erb
@@ -60,38 +60,26 @@
 
 <fieldset>
   <legend><%= ts("Colors") %></legend>
-  <p class="notes">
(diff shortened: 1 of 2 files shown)


sarken (approved):
Looks good to me! There might be some feedback on regarding help text once we get to testing -- it's not super intuitive, imo, and it took me a while to figure out how to enter a hex, which I think a lot of people are still going to want to do -- but we'll see when we get there.

weeklies:
Just a note, it looks like Firefox for Android 27 doesn't support custom colors yet: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/color#browser_compatibility

sarken:
Not a problem! I'll mention that on the issue so we can find it again, and make sure Support knows it's a Firefox thing not an us thing.

PR #4373, opened 2022-10-23
AO3-6414 Clean up unnecessary calls to include ActiveModel::ForbiddenAttributesProtection.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6414

## Purpose

I removed all lines like the following:
```
```
These were added when AO3 was still using Rails 3, to prepare for the transition to Rails 4. But Rails 4+ includes it automatically in `ActiveRecord::Base`, so the lines are no longer necessary.
 
I also called `rubocop -a --only "Layout/EmptyLinesAroundClassBody"` on all affected files, to clean up extra newlines.

### DIFF
--- a/app/models/abuse_report.rb
+++ b/app/models/abuse_report.rb
@@ -1,6 +1,4 @@
 class AbuseReport < ApplicationRecord
-  include ActiveModel::ForbiddenAttributesProtection
-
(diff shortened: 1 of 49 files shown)


sarken (approved):
Grazie!

PR #4377, opened 2022-11-08
AO3-6387 Prevent editing existing tags when creating new tags

ceithir:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6387

## Purpose

Prevent accidentally uncanonizing an existing tag when creating a new one.

## Testing Instructions

As per the JIRA walkthrough.

## Credit

Ceithir (he/him)

### DIFF
--- a/app/controllers/tags_controller.rb
+++ b/app/controllers/tags_controller.rb
@@ -198,13 +198,13 @@ class TagsController < ApplicationController
       return
     end
 
-    if (@tag.name != tag_params[:name]) && @tag.name.casecmp(tag_params[:name].downcase).zero? # only capitalization different
-      @tag.update_attribute(:name, tag_params[:name]) # use the new capitalization
-      flash[:notice] = ts('Tag was successfully modified.')
-    else
+    if @tag.id_previously_changed? # i.e. tag is new
+      @tag.update_attribute(:canonical, tag_params[:canonical])
       flash[:notice] = ts('Tag was successfully created.')
+    else
+      flash[:notice] = ts('Tag already existed and was not modified.')
     end
-    @tag.update_attribute(:canonical, tag_params[:canonical]) unless @tag.canonical? # If tag already canonical, do not uncanonize it
+
     redirect_to edit_tag_path(@tag)
   end
 
--- a/app/controllers/tags_controller.rb
+++ b/app/controllers/tags_controller.rb
@@ -178,9 +178,9 @@ class TagsController < ApplicationController
     type = tag_params[:type] if params[:tag]
 
     unless type
-      flash[:error] = ts('Please provide a category.')
+      flash[:error] = ts("Please provide a category.")
       @tag = Tag.new(name: tag_params[:name])
-      render(action: 'new')
+      render(action: "new")
       return
     end
 
@@ -188,21 +188,21 @@ class TagsController < ApplicationController
 
     model = begin
               type.classify.constantize
-            rescue
+            rescue StandardError
               nil
             end
     @tag = model.find_or_create_by_name(tag_params[:name]) if model.is_a? Class
 
-    unless @tag && @tag.valid?
-      render(action: 'new')
+    unless @tag&.valid?
+      render(action: "new")
       return
     end
 
     if @tag.id_previously_changed? # i.e. tag is new
       @tag.update_attribute(:canonical, tag_params[:canonical])
-      flash[:notice] = ts('Tag was successfully created.')
+      flash[:notice] = ts("Tag was successfully created.")
     else
-      flash[:notice] = ts('Tag already existed and was not modified.')
+      flash[:notice] = ts("Tag already existed and was not modified.")
     end
 
     redirect_to edit_tag_path(@tag)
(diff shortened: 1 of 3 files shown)


app/controllers/tags_controller.rb
```diff
@@ -196,7 +196,7 @@ def create
       else
         flash[:notice] = ts('Tag was successfully created.')
       end
-      @tag.update_attribute(:canonical, tag_params[:canonical])
+      @tag.update_attribute(:canonical, tag_params[:canonical]) unless @tag.canonical? # If tag already canonical, do not uncanonize it
```
brianjaustin, line 199: I'm wondering if this makes more sense to do at the model layer, maybe something like
```ruby
before_update :prevent_un_canonicalization, if: :will_save_change_to_canonical?
def prevent_un_canonicalization
  return unless canonical_was
  self.canonical = true
end
```
(caveat: I think there might be a better way to do that, but I didn't want to go too far down the rabbit hole.)

Personally, erroring here seems more user-friendly as well (it also is kinda weird to me that tag creation could update at all...), but someone more official should probably weigh in on that question.

features/tags_and_wrangling/tag_wrangling_special.feature
```diff
@@ -180,3 +180,16 @@ Feature: Tag Wrangling - special cases
     Then I should see "is not a canonical tag. Please make it canonical before adding synonyms to it."
     When I follow 'Evelyn "Evie" Carnahan'
     Then I should be on the 'Evelyn "Evie" Carnahan' tag edit page
+
+  Scenario: Cannot uncanonize a tag on creation (AO3-6387)
```
brianjaustin, line 184: Independent of what happens above, couldn't this be pulled down into the unit test specs? I tend to only reach for integration tests when absolutely necessary given that they tend to run slower

app/controllers/tags_controller.rb
tickinginstant: I don't think putting that callback in the model is a good plan, as it'll prevent anyone from ever making a tag non-canonical through the tag edit form (even for tags where that's allowed, and for admins who are always permitted to make tags non-canonical).

brianjaustin: Fair enough! I think it would be possible to add an exception for that, but I agree it's murkier taking that into consideration

features/tags_and_wrangling/tag_wrangling_special.feature
ceithir: Cannot be a unit if the check remains in the controller (as per the conclusion of the above discussion). But I guess going for a more lightweight controller test for a controller change does make a lot of sense.

app/controllers/tags_controller.rb
```diff
@@ -196,7 +196,7 @@ def create
       else
         flash[:notice] = ts('Tag was successfully created.')
```
redsummernight, line 197: Thinking about it more, let's avoid making changes to existing tags altogether - wranglers should use the edit form for that and go through all the validations, instead of running some one off `update_attribute` here.

We can rewrite this section like so:
```ruby
if @tag && @tag.valid?
  if @tag.id_previously_changed?
    # Set canonical to checkbox (it's fine here because this is a new tag).
    # Flash "Tag was successfully created."
  else
    # Flash "Tag already existed and was not modified."
  end
else
  # Go back to new
end
```

ceithir: Oh, didn't know about the "id_previously_changed?" trick to differentiate between a find or a create in that case. Thanks.

app/controllers/tags_controller.rb
```diff
@@ -176,31 +176,36 @@ def new
...
+    end
+
+    if @tag.id_previously_changed? # i.e. tag is new
       @tag.update_attribute(:canonical, tag_params[:canonical])
-      redirect_to edit_tag_path(@tag)
+      flash[:notice] = ts("Tag was successfully created.")
     else
-      render(action: 'new') && return
+      flash[:notice] = ts("Tag already existed and was not modified.")
     end
```
ceithir, line 206: The only real change is within this block, that replaces this one:
https://github.com/otwcode/otwarchive/blob/master/app/controllers/tags_controller.rb#L192..L201

The rest is a small syntax change* to keep the number of imbricated if low, and accidentally awakening the hound while doing so.

*Replacing:
```
if condition
  [normal behavior]
else
  [error handling]
  return
end
```
With:
```
unless condition
  [error handling]
  return
end

[normal behavior]
```

redsummernight: Yeah that's totally fine.

FYI when linking to code, you can [get a link to specific commits](https://docs.github.com/en/repositories/working-with-files/using-files/getting-permanent-links-to-files) by pressing Y.

spec/controllers/tags_controller_spec.rb
```diff
@@ -22,8 +22,37 @@
     context "successful creation" do
       before { post :create, params: { tag: tag_params } }
 
+      it "creates a new, non-canonical, tag" do
+        expect(flash[:notice]).to eq "Tag was successfully created."
```
redsummernight, line 26: You can use the `it_redirects_to_with_notice` helper which also checks the destination path.

redsummernight (approved):
Looks good, thanks!

PR #4378, opened 2022-11-09
AO3-5860 commenting or leaving kudos with admin accounts

ceithir:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5860

## Purpose

Enjoy admins to log out before leaving a comment or kudo.

More exactly:
Hide the relevant buttons as long as they're logged as admin
Return an error message in case such a request reaches the back regardless

## Testing Instructions

See JIRA issues (including comments).

## Credit

Ceithir (he/him)

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -28,6 +28,7 @@ class CommentsController < ApplicationController
   before_action :check_permission_to_moderate, only: [:approve, :reject]
   before_action :check_permission_to_modify_frozen_status, only: [:freeze, :unfreeze]
   before_action :check_permission_to_modify_hidden_status, only: [:hide, :unhide]
+  before_action :admin_logout_required, only: [:create]
 
   include BlockHelper
 
--- a/app/views/comments/_comment_form.html.erb
+++ b/app/views/comments/_comment_form.html.erb
@@ -94,7 +94,9 @@
               :maximum_length => ArchiveConfig.COMMENT_MAX,
               :tooLongMessage => ts("must be less than %{count} characters long.", :count => ArchiveConfig.COMMENT_MAX)) %>
       <p class="submit actions">
-        <%= f.submit button_name, :id => "comment_submit_for_#{commentable.id}", data: {disable_with: ts("Please wait...")} %>
+        <% if !logged_in_as_admin? %>
+          <%= f.submit button_name, :id => "comment_submit_for_#{commentable.id}", data: {disable_with: ts("Please wait...")} %>
+        <% end %>
         <% if controller.controller_name == 'inbox' %>
           <a name="comment_cancel" id="comment_cancel"><%= ts("Cancel") %></a>
         <% elsif comment.persisted? %>
--- a/app/views/comments/_comment_form.html.erb
+++ b/app/views/comments/_comment_form.html.erb
@@ -64,13 +64,6 @@
           <p class="footnote">(<%= allowed_html_instructions %>)</p>
         <% end %>
 
-      <% elsif logged_in_as_admin? %>
-        <h4 class="heading"><%= ts("Comment as") %> <span class="byline"><%= current_admin.login %></span>
-          <%= f.hidden_field :name, :value => "#{current_admin.login}", :id => "comment_name_for_#{commentable.id}" %>
-          <%= f.hidden_field :email, :value => "#{current_admin.email}", :id => "comment_email_for_#{commentable.id}" %>
-        </h4>
-        <p class="footnote">(<%= allowed_html_instructions %>)</p>
-
       <% else %>
         <dl>
           <dt class="landmark"><%= ts("Note") %>:</dt>
--- a/app/views/comments/_commentable.html.erb
+++ b/app/views/comments/_commentable.html.erb
@@ -28,7 +28,7 @@
       <li><%= link_to ts('Back to AO3 News Index'), admin_posts_path %></li>
     <% end %>
 
-    <% if @work && !is_author_of?(@work) %>
+    <% if @work && !is_author_of?(@work) && !logged_in_as_admin? %>
       <li>
         <%= form_for(:kudo, url: kudos_path, html: { method: 'post', id: 'new_kudo' }) do |kudo_form| %>
           <%= kudo_form.hidden_field :commentable_id, :value => @work.id %>
--- a/features/comments_and_kudos/add_comment.feature
+++ b/features/comments_and_kudos/add_comment.feature
@@ -208,12 +208,10 @@ Scenario: Try to post a comment with a < angle bracket before a linebreak, with 
       And I press "Comment"
     Then I should see "Comment created!"
 
-Scenario: Asked to logged out first if trying to post a comment as an admin
+Scenario: Cannot comment (no button) while logged as admin
 
     Given the work "Generic Work"
       And I am logged in as an admin
       And I view the work "Generic Work"
-    When I fill in "Comment" with "Nice work!"
-      And I press "Comment"
-    Then I should see "Please log out of your admin account first!"
-      And I should not see "Comment created!"
+    Then I should see "Generic Work"
+      And I should not see a "Comment" button
--- a/features/comments_and_kudos/add_comment.feature
+++ b/features/comments_and_kudos/add_comment.feature
@@ -207,3 +207,13 @@ Scenario: Try to post a comment with a < angle bracket before a linebreak, with 
       """
       And I press "Comment"
     Then I should see "Comment created!"
+
+Scenario: Asked to logged out first if trying to post a comment as an admin
+
+    Given the work "Generic Work"
+      And I am logged in as an admin
+      And I view the work "Generic Work"
+    When I fill in "Comment" with "Nice work!"
+      And I press "Comment"
+    Then I should see "Please log out of your admin account first!"
+      And I should not see "Comment created!"
--- a/spec/controllers/comments_controller_spec.rb
+++ b/spec/controllers/comments_controller_spec.rb
@@ -247,16 +247,9 @@ describe CommentsController do
       context "when logged in as an admin" do
         before { fake_login_admin(create(:admin)) }
 
-        it "posts the comment and shows it in context" do
+        it "asks to log out first" do
           post :create, params: { tag_id: fandom.name, comment: anon_comment_attributes }
-          comment = Comment.last
-          expect(comment.commentable).to eq fandom
-          expect(comment.name).to eq anon_comment_attributes[:name]
-          expect(comment.email).to eq anon_comment_attributes[:email]
-          expect(comment.comment_content).to include anon_comment_attributes[:comment_content]
-          path = comments_path(tag_id: fandom.to_param,
-                               anchor: "comment_#{comment.id}")
-          expect(response).to redirect_to path
+          expect(flash[:notice]).to eq("Please log out of your admin account first!")
         end
       end
 
--- a/spec/controllers/kudos_controller_spec.rb
+++ b/spec/controllers/kudos_controller_spec.rb
@@ -182,5 +182,17 @@ describe KudosController do
         end
       end
     end
+
+    context "when kudos giver is admin" do
+      let(:work) { create(:work) }
+      let(:admin) { create(:admin) }
+
+      before { fake_login_admin(admin) }
+
+      it "asks to log out first" do
+        post :create, params: { kudo: { commentable_id: work.id, commentable_type: "Work" } }
+        expect(flash[:notice]).to eq("Please log out of your admin account first!")
+      end
+    end
   end
 end
--- a/app/controllers/kudos_controller.rb
+++ b/app/controllers/kudos_controller.rb
@@ -1,5 +1,6 @@
 class KudosController < ApplicationController
   skip_before_action :store_location
+  before_action :admin_logout_required, only: [:create]
 
   def index
     @work = Work.find(params[:work_id])
--- a/config/locales/views/en.yml
+++ b/config/locales/views/en.yml
@@ -99,7 +99,7 @@ en:
       page_title: "Hi, %{login}!"
       confidentiality_reminder: "You are now logged in as an admin. That means you will probably encounter information that is personal or confidential (e.g. usernames, email and IP addresses, creator names on anonymous works, etc). Please do not use this information in ways unrelated to your OTW role. If you have questions about what you can or cannot do with information you see here, contact your committee chair(s)."
       responsibility: "With great power comes great responsibility."
-      log_out_reminder: "Please remember to log out before resuming your normal site activity and leaving comments and kudos!"
+      log_out_reminder: "Please remember to log out before resuming your normal site activity!"
       roles:
         heading: "Your admin roles:"
         none: "You currently have no admin roles assigned to you."
--- a/features/comments_and_kudos/kudos.feature
+++ b/features/comments_and_kudos/kudos.feature
@@ -167,3 +167,11 @@ Feature: Kudos
       And the email should contain "Meh Story"
       And the email should not contain "0 guests"
       And the email should not contain "translation missing"
+
+Scenario: Asked to log out first when trying to leave kudos as admin
+
+    Given I am logged in as an admin
+      And I view the work "Awesome Story"
+    When I press "Kudos ♥"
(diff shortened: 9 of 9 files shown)


ceithir:
As of now, I've restricted admins only from commenting on works. Not sure if this should be extended to some other types of commentable (chapters?).

sarken:
Thanks for working on this! And my apologies; I thought I had replied to this days ago. 

It looks like the issue wasn't entirely clear about the scope -- admins shouldn't be able to leave comments or kudos on anything, period. (Volunteers are expected to use user accounts with the official role for communication with users.) The comment form should also be hidden, not just the button.

app/views/comments/_comment_form.html.erb
```diff
@@ -2,115 +2,120 @@
...
+                <% else %>
+                  <span><%= ts("to") %> <%= get_commenter_pseud_or_name(commentable) %> <%= ts("on") %> <%= commentable_description_link(commentable) %></span>
+                <% end %>
               <% end %>
-            <% end %>
+            </h4>
+            <p class="footnote">(<%= allowed_html_instructions %>)</p>
+          <% end %>
+
+        <% elsif logged_in_as_admin? %>
```
sarken, line 71: The admin-only code should be removed.

app/views/comments/_comment_form.html.erb
```diff
@@ -2,115 +2,120 @@
...
-    <fieldset>
-      <legend><%= ts("Post Comment") %></legend>
 
-      <% # here come the hacks (hidden fields to transmit various info to the create action) %>
-      <% if commentable.is_a?(Tag) %>
-        <%= hidden_field_tag :tag_id, commentable.name %>
-      <% end %>
+<% if logged_in_as_admin? %>
+  <% # Hide form for admins %>
+<% else %>
```
sarken, line 8: Would `unless logged_in_as_admin?` work here, rather than including an empty `if` for admins?

spec/controllers/comments_controller_spec.rb
```diff
@@ -340,6 +335,17 @@
           expect(cookies[:flash_is_set]).to eq(1)
         end
       end
+
+      context "when logged in as an admin" do
+        let(:work) { create(:work) }
+
+        before { fake_login_admin(create(:admin)) }
+
+        it "asks to log out first" do
```
sarken, line 344: Could you check the redirects in these tests as well, e.g., 

```
it "redirects to root with notice to log out" do
  post :create, params: { work_id: work.id, comment: anon_comment_attributes }
  it_redirects_to_with_notice(root_path, "Please log out of your admin account first!")
end
```

sarken (changes requested):

app/views/comments/_comment_form.html.erb
tickinginstant: Maybe it'd be better to move the conditional out of this file? Looking at the places where `comments/comment_form` is rendered, most of those should be inaccessible to admins with a few other (probably necessary) changes like:
1. Removing the Reply button for admins by modifying `can_reply_to_comment?`.
2. Preventing admins from accessing the `CommentsController#new` action.
3. Preventing admins from accessing the `CommentsController#add_comment_reply` action.

Since admins can't access user inboxes, and can't edit comments, that would leave only the call in `comments/commentable` to modify.

ceithir: @tickinginstant 

Unsure about such an approach in the midterm. We know the various ways to get to the comment form as of now. But if a new one is added later on, there's a good chance for the "not admin check" to be forgotten and the issue to return.

tickinginstant: IMO it's not a big deal if `comments/comment_form` is added to a new place later on, because the actual creation of comments still goes through the `create` action, which will be blocked. So it'll be a visual bug at most (and a very obvious visual bug, which is generally pretty easy to find/fix/test).

If you stick to wrapping `comments/comment_form` in a giant if statement, and don't add checks elsewhere, you're going to leave behind empty `<div>`s in places like this:
https://github.com/otwcode/otwarchive/blob/23ffdc103362f4c3c61167a31066d1fe87ed7167/app/views/comments/_commentable.html.erb#L109-L113
Which feels a bit messy to me. And if you do have checks elsewhere, you don't need them in `comments/comment_form` as well.

features/comments_and_kudos/add_comment.feature
```diff
@@ -207,3 +207,24 @@ Scenario: Try to post a comment with a < angle bracket before a linebreak, with
...
+
+    Given the work "Generic Work"
+    When I am logged in as "commenter"
+      And I view the work "Generic Work"
+      And I post a comment "Woohoo"
+    When I am logged in as an admin
+      And I view the work "Generic Work"
+      And I follow "Comments (1)"
+    Then I should see "Woohoo"
+      And I should not see a "Reply" button
```
tickinginstant, line 230: "Reply" looks like a button, but it's actually a link, so this test is not checking what it's supposed to.

app/controllers/comments_controller.rb
```diff
@@ -28,6 +28,7 @@ class CommentsController < ApplicationController
   before_action :check_permission_to_moderate, only: [:approve, :reject]
   before_action :check_permission_to_modify_frozen_status, only: [:freeze, :unfreeze]
   before_action :check_permission_to_modify_hidden_status, only: [:hide, :unhide]
+  before_action :admin_logout_required, only: [:create]
```
tickinginstant, line 31: I really do think you should add `:new` and `:add_comment_reply` here. Otherwise, an admin who opens one of those pages is going to get this behavior:
1) The new comments page still has "New comment on TITLE", but there's nothing at all on the page.
2) Opening the comments section logged-out in one tab, logging in as admin in another tab, and then clicking "Reply" in the first tab makes the navigation for that comment disappear. (Normally it would make the navigation disappear and replace it with a comment box. But with the comment box hidden, it just inexplicably takes away the Thread/Parent Thread links.)

(In general, any filters that apply to the comment `create` action should probably also be applied to `new` and `add_comment_reply`, unless there's a specific reason that it should apply only when submitting the new comment.)

app/views/comments/_comment_form.html.erb
ceithir: Just modifying the CommentsController isn't enough to prevent the form from being displayed in all cases (this controller is not called when the template is just rendered in another template). I do agree however it makes more sense to perform the check in the `comments/_commentable.html.erb`file, alongside all the others, than adding a gigantic wrapper one step later.

tickinginstant: Thanks for making the changes! Which of the actions/templates hasn't been handled yet?

ceithir: Based on the files `comments/comment_form` is rendered in, I think we should be good?

* `app/views/comments/_commentable.html.erb` => Already updated, so good
* `app/views/comments/_single_comment.html.erb`, `app/views/comments/edit.html.erb`, `app/views/comments/edit.js.erb` => Only for existing comment edition, with a check on ownership, so should be good (see no reason to prevent admins from editing their old comments they created before this PR, on the contrary)
* `app/views/comments/_comment_actions.html.erb`, `app/views/comments/add_comment_reply.js.erb`, `app/views/inbox/reply.js.erb` -> Behind a `can_reply_to_comment?` check, an already updated method, so good
* `app/views/comments/new.html.erb` => Behind the new controller check, so good

Unless I'm missing something here?

spec/controllers/comments_controller_spec.rb
```diff
@@ -154,10 +154,9 @@
       context "when logged in as an admin" do
         before { fake_login_admin(create(:admin)) }
 
-        it "renders the :new template" do
+        it "asks to log out first" do
           post :new, params: { tag_id: fandom.name }
```
sarken, line 158: I think we need to add a test for the `add_comment_reply` action as well. And can we tweak the description to better describe the behavior, i.e., `it "redirects to root with notice prompting log out"` or similar?

app/views/comments/_commentable.html.erb
```diff
@@ -105,6 +105,10 @@
     <p class="notice">
       <%= t(".blocked") %>
     </p>
+  <% elsif logged_in_as_admin? %>
+    <p class="notice">
+      <%= t("admins.index.log_out_reminder") %>
```
sarken, line 110: How about a new message that says: "Please log out of your admin account to comment."

sarken (commented):
Just a couple small things and I think this will be ready -- thank you for working on it!

spec/controllers/kudos_controller_spec.rb
```diff
@@ -208,5 +208,17 @@
...
     end
+
+    context "when kudos giver is admin" do
+      let(:work) { create(:work) }
+      let(:admin) { create(:admin) }
+
+      before { fake_login_admin(admin) }
+
+      it "asks to log out first" do
+        post :create, params: { kudo: { commentable_id: work.id, commentable_type: "Work" } }
```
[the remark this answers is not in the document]
ceithir: Definitely not great, as `<html><body>You are being <a href="http://www.example.com/">redirected</a>.</body></html>` is not valid JSON.

Could be handled more nicely like https://github.com/otwcode/otwarchive/pull/4378/commits/5ddfb2e209bab38abe496ca673976ce91ff46b0c... But seems a bit overkill for what I understand is an edge case (maybe I'm missing something?), where the user clicks in the Kudos button in one tab after having logged as admin in another.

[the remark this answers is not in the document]
sarken: Yeah, that does seem like overkill for the situation. Maybe just add a test showing no kudos is created and we'll leave it at that?

ceithir:
Conflict with https://github.com/otwcode/otwarchive/pull/4492 technically solved. Unsure if it also implies some behavior changes here however.

sarken:
I can't think of any changes we'd need off the top of my head!

Bilka2 (changes requested):
All of the code looks good, but this has merge conflicts with master that need to be solved.

app/views/comments/_comment_form.html.erb
```diff
@@ -64,46 +64,39 @@
...
-          <%= f.hidden_field :email, :value => "#{current_admin.email}", :id => "comment_email_for_#{commentable.id}" %>
-        </h4>
-        <p class="footnote">(<%= allowed_html_instructions %>)</p>
-
       <% else %>
         <dl>
-          <dt class="landmark"><%= ts("Note") %>:</dt>
-          <dd class="instructions comment_form"><%=ts("All fields are required. Your email address will not be published.") %></dd>
-          <dt><%= f.label "name_for_#{commentable.id}", ts("Guest name: ") %></dt>
+          <dt class="landmark"><%= t(".note") %>:</dt>
```
Bilka2, line 69: Another landmark! A key like `.landmark.note` would be clearer to its function.

app/views/comments/_comment_form.html.erb
```diff
@@ -5,7 +5,7 @@
 <div class="post comment" id="comment_form_for_<%= commentable.id %>">
   <%= form_for value_for_comment_form(commentable, comment), remote: !comment.new_record?, authenticity_token: true, html: { id: "comment_for_#{commentable.id}" } do |f| %>
     <fieldset>
-      <legend><%= ts("Post Comment") %></legend>
+      <legend><%= t(".form_title") %></legend>
```
Bilka2, line 8: Based on other places where there is a `<legend>`, maybe the key should be `.legend`. Once again this is meta information that is not visible usually, so it's nice to make that clear.

app/views/comments/_comment_form.html.erb
```diff
@@ -28,19 +28,19 @@
...
         <% if current_user_is_anonymous_creator(commentable) %>
           <p class="notice">
-           <%= ts("While this work is anonymous, comments you post will also be listed anonymously.") %>
+          <%= t(".anonymous_forewarning") %>
           </p>
         <% end %>
 
         <% if current_user.pseuds.count > 1 %>
-          <h4 class="heading"><%= ts("Comment as") %> <%= f.collection_select :pseud_id, current_user.pseuds, :id, :name, {:selected => (comment.pseud ? comment.pseud.id.to_s : current_user.default_pseud.id.to_s)}, :id => "comment_pseud_id_for_#{commentable.id}", :title => ts("Choose Name") %>
+          <h4 class="heading"><%= t(".comment_as") %> <%= f.collection_select :pseud_id, current_user.pseuds, :id, :name, { selected: (comment.pseud ? comment.pseud.id.to_s : current_user.default_pseud.id.to_s) }, id: "comment_pseud_id_for_#{commentable.id}", title: t(".choose_name") %>
```
Bilka2, line 43: There is the `.comment_field_title` key, but this field title doesn't have `field_title` in its key. So maybe `.choose_name_field_title` is better? Not sure though.

app/views/comments/_comment_form.html.erb
```diff
@@ -64,46 +64,39 @@
...
-              :failureMessage => ts('Brevity is the soul of wit, but we need your comment to have text in it.'),
-              :maximum_length => ArchiveConfig.COMMENT_MAX,
-              :tooLongMessage => ts("must be less than %{count} characters long.", :count => ArchiveConfig.COMMENT_MAX)) %>
-      <p class="submit actions">
-        <%= f.submit button_name, :id => "comment_submit_for_#{commentable.id}", data: {disable_with: ts("Please wait...")} %>
+      <%= live_validation_for_field "comment_content_for_#{commentable.id}",
+            failureMessage: t(".comment_too_short"),
+            maximum_length: ArchiveConfig.COMMENT_MAX,
+            tooLongMessage: t(".comment_too_long", count: ArchiveConfig.COMMENT_MAX) %>
+    <p class="submit actions">
```
Bilka2, line 96: I think this needs to be indented two spaces more, assuming ERB Lint doesn't complain.

app/views/comments/_comment_form.html.erb
```diff
@@ -28,19 +28,19 @@
...
-          <%= ts("This work's creator has chosen to moderate comments on the work. Your comment will not appear until it has been approved by the creator.") %>
+          <%= t(".moderation_forewarning") %>
         </p>
       <% end %>
 
       <% if logged_in? %>
         <% if current_user_is_anonymous_creator(commentable) %>
           <p class="notice">
-           <%= ts("While this work is anonymous, comments you post will also be listed anonymously.") %>
+          <%= t(".anonymous_forewarning") %>
```
Bilka2, line 38: Indent two more spaces.

Bilka2 (changes requested):
In for a penny, in for a pound, indeed.

app/views/comments/_comment_form.html.erb
```diff
@@ -28,88 +28,77 @@
...
         <% if current_user.pseuds.count > 1 %>
-          <h4 class="heading"><%= ts("Comment as") %> <%= f.collection_select :pseud_id, current_user.pseuds, :id, :name, {:selected => (comment.pseud ? comment.pseud.id.to_s : current_user.default_pseud.id.to_s)}, :id => "comment_pseud_id_for_#{commentable.id}", :title => ts("Choose Name") %>
+          <h4 class="heading"><%= t(".comment_as") %> <%= f.collection_select :pseud_id, current_user.pseuds, :id, :name, { selected: (comment.pseud ? comment.pseud.id.to_s : current_user.default_pseud.id.to_s) }, id: "comment_pseud_id_for_#{commentable.id}", title: t(".choose_name_field_title") %>
             <% if controller.controller_name == "inbox" %>
-              <% if commentable.by_anonymous_creator? %>
-                <span><%= ts("to") %> <%= "Anonymous Creator" %> <%= ts("on") %> <%= commentable_description_link(commentable) %></span>
-              <% else %>
-                <span><%= ts("to") %> <%= get_commenter_pseud_or_name(commentable) %> <%= ts("on") %> <%= commentable_description_link(commentable) %></span>
-              <% end %>
+              <span><%= t ".inbox_reference",
```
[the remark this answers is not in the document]
ceithir: > According to the i18n wiki page, the t function call should use parens ().

Might be good to add it to Rubocop then? I'm not sure if there's an elegant way to force https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/MethodCallWithArgsParentheses for a single specific function.

Bilka2 (approved):
:tada:

PR #4381, opened 2022-11-10
AO3-6392 Comment actions when replying in full page

ceithir:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6392

## Purpose

As per the JIRA, fix extra actions showing up and the Cancel button being "stuck" when answering a comment in full page.

## Testing Instructions

Follow the JIRA instructions.

At the end, also try to "Cancel" the reply in the new page. ~~The page should reload instead of the wrong buttons sticking around.~~ The comment box should now collapse properly, the button "Reply" shows up, and hitting it open the box once again.

## Credit

Ceithir (he/him)


ceithir:
No, I didn't forget to add tests to this PR for over six months, it's just a trick of the light.

features/step_definitions/comment_steps.rb
```diff
@@ -302,3 +302,24 @@
...
+end
+
+When /^I reply to the work comment "([^"]*)" on a new page$/ do |comment_content|
+  comment = Comment.find_by(comment_content: comment_content)
+  visit work_path(
+    comment.commentable,
+    add_comment_reply_id: comment.id,
+    show_comments: true
+  )
+end
```
ceithir, line 325: Was originally:
```rb
new_window = window_opened_by do
  find(:link, "Reply").click(:control)
end
switch_to_window(new_window)
```
To be closer to real life behavior. But I encountered issues with domains (see below) and moved to a more direct approach.

features/step_definitions/web_steps.rb
```diff
@@ -319,3 +319,9 @@ def with_scope(locator)
   cookie_value = Capybara.current_session.driver.request.cookies.[](cookie)
   assert cookie_value.nil?
 end
+
+When "I fix the domain name" do
+  parsed = URI.parse(current_url)
+  dest = "#{parsed.path}?#{parsed.query || ''}##{parsed.fragment || ''}"
+  visit dest
+end
```
ceithir, line 327: I'm not sure exactly what happens, but every time there's a non Capybara page change with JS enabled (clicking a link, getting redirected...), the domain is replaced with `example.com`.

No idea how to get around that issue cleanly honestly.

features/step_definitions/comment_steps.rb
weeklies: I ran into the same issue with [domains in testing](https://github.com/otwcode/otwarchive/pull/4558/files#diff-bff3e97645581265ef40912813188da582223b750b875fa0c24c1d4f40f1e948R2) and it's due to the use of URLs rather than paths. You might be able to resolve it by modifying this [helper](https://github.com/otwcode/otwarchive/blob/801ce0931c0dbb144f9631eed5e9cf743a6c8d00/app/helpers/comments_helper.rb#L192)

app/controllers/comments_controller.rb
```diff
@@ -657,7 +657,8 @@ def redirect_to_all_comments(commentable, options = {})
                   delete_comment_id: options[:delete_comment_id],
                   view_full_work: options[:view_full_work],
                   anchor: options[:anchor],
-                  page: options[:page]
+                  page: options[:page],
+                  only_path: true
```
ceithir, line 661: Conflicted about this.

On one hand, it does fix every issue I had in tests about being redirect to the wrong domain (example.com).

On the other, changing such a critical piece of code just so tests pass trigger every warning in my brain.

app/helpers/comments_helper.rb
```diff
@@ -361,4 +364,8 @@ def comments_are_moderated(commentable)
     parent = find_parent(commentable)
     parent.respond_to?(:moderated_commenting_enabled) && parent.moderated_commenting_enabled?
   end
+
+  def on_js_less_comment_form(commentable)
```
brianjaustin, line 368: I think there are scenarios where this will be the case when JS is enabled, for example using the reply link from an email (or opening in a new tab like the issue mentions). Perhaps something like `focused_on_comment` would be a better description of what this represents?

app/views/comments/_comment_form.html.erb
```diff
@@ -107,7 +107,7 @@
         <% elsif comment.persisted? %>
           <%= cancel_edit_comment_link(comment) %>
         <% elsif commentable.is_a?(Comment) || commentable.is_a?(CommentDecorator) %>
-          <%= cancel_comment_reply_link(commentable) %>
+          <%= cancel_comment_reply_link(commentable, remote: !on_js_less_comment_form(commentable)) %>
```
brianjaustin, line 110: It looks like this results in redirecting to the work page (assuming the parent comment is on a work) with the comments section collapsed. That does the job, but I'm not sure if it's the intended behaviour. Thoughts @sarken ?

sarken: Do you mean when replying from the comment section or somewhere else?

I'm looking at the current behavior of the Cancel button on the reply form and what I see is:

* if you're replying to a comment from an email link, e.g., `http://test.archiveofourown.org/comments/5209801?add_comment_reply_id=5209801`, it 
  * redirects to the comment you're replying to without the reply form open if JavaScript is disabled
  * closes the reply form without reloading if JavaScript is enabled
* if you're replying to a comment from the comment section on the work, it
  * redirects to the work with the comment section closed if JavaScript is disabled
  * closes the reply form without reloading if JavaScript is enabled

So if we're talking about from the comment section on the bottom of a work and not a page like the URL I provided, it's not ideal, but it does seem consistent. I guess in an ideal world it would leave the comment section expanded when it redirected to the work, but I'd probably call that out of scope.

app/controllers/comments_controller.rb
sarken: Yeah, I'm normally not a fan of that either, and the alarm bells are screaming. But ignoring my knee-jerk "let's not do this," I can't think of a reason it _shouldn't_ be `only_path: true`. We definitely prefer paths to URLs. I think the only reason we hadn't already changed it is [we do something in nginx that means this wasn't actively causing issues](https://otwarchive.atlassian.net/browse/AO3-4630) when we were supporting both HTTP and HTTPS. So this is probably fine.

app/views/comments/_comment_form.html.erb
ceithir: Well...

While attempting to clean up a seemingly unused method, I accidentally found the source of the bug that I had tried to circumvent by forcing a redirection there. So it's not needed anymore as the form now collapses down properly even if the page started in "focused" mode, and everything is smooth in the best of worlds.

Hooray for serendipity?

brianjaustin: Hah, I love when deleting code fixes bugs :laughing:

I guess it's less relevant now, but the case I was talking about was a special case of 2b: opening the reply link in a new tab (so ending up on a page like http://localhost:3000/chapters/170?add_comment_reply_id=42&show_comments=true#comment_42) and then clicking "Cancel"

brianjaustin (approved):

sarken (changes requested):

app/views/comments/_comment_form.html.erb
ceithir: That should work now actually. Cf https://github.com/otwcode/otwarchive/pull/4381/files#diff-6adcc86fa92c4daa39cee6c5bbac11ad8cc3a746852f42c98e8a4f1301cb0eb2R233-R246

sarken (approved):

PR #4382, opened 2022-11-15
AO3-6421 Rewrite add_paragraphs_to_text to use Nokogiri nodes.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6421

Also fixes:
https://otwarchive.atlassian.net/browse/AO3-3769

## Purpose

- Rewrite `add_paragraphs_to_text` to use Nokogiri nodes to perform the HTML manipulation.
- Reorder the calls to `Sanitize.clean` vs. `add_paragraphs_to_text` for safety.


lib/paragraph_maker.rb
```diff
@@ -0,0 +1,260 @@
...
+
+  # Traverse the tree and delete any paragraph nodes that have no children.
+  def delete_empty_paragraphs(root)
+    traverse(root) do |node|
+      node.unlink if node.name == "p" && node.children.empty?
+    end
+  end
+
+  # Process the given node to add paragraphs to it and all of its children.
+  def process(root)
```
redsummernight, line 251: Can you make the other methods private?

lib/paragraph_maker.rb
```diff
@@ -0,0 +1,260 @@
...
+    # there's no point in a proper split, because one of the two tags would just
+    # be empty. Instead, we rearrange ourselves relative to our parent.
+    return node.parent.add_previous_sibling(node) if node.previous_sibling.nil?
+    return node.parent.add_next_sibling(node) if node.next_sibling.nil?
+
+    # Create a shallow copy of the parent:
+    new_parent = node.parent.dup(0)
+
+    # Copy over all of the attributes:
+    node.parent.attributes.each do |name, attribute|
```
redsummernight, line 199: Should we have a spec for keeping the attributes on the new split parent?

lib/html_cleaner.rb
```diff
@@ -220,153 +140,12 @@ def walk_array(array)
...
     # Adding paragraphs in place of linebreaks
     doc = Nokogiri::HTML.fragment("<myroot>#{text}</myroot>")
-    out_html = traverse_nodes(doc.at_css("myroot"))[1]
-    # Remove empty paragraphs
-    out_html.gsub!(/<p>\s*?<\/p>/, "")
-    out_html.gsub!(/(\A<myroot>)|(<\/myroot>\Z)|(\A<myroot\/>\Z)/, "")
-    out_html
+    myroot = doc.children.first
+    ParagraphMaker.process(myroot)
+    myroot.children.to_xhtml
```
redsummernight, line 148: Do you choose `to_xhtml` to be consistent with:

https://github.com/otwcode/otwarchive/blob/d74d404f052b4e60e223c712fff619844695806d/lib/html_cleaner.rb#L172-L174

or is there another reason?

redsummernight (commented):
Great work, thanks! Very happy to see some long standing TODOs in the parser cleaned up as well.

potpotkettle:
Can you also fix [AO3-5801](https://otwarchive.atlassian.net/browse/AO3-5801) ? I have #4303 but the approach here may be better.

lib/paragraph_maker.rb
```diff
@@ -0,0 +1,260 @@
...
+  end
+
+  # Strip whitespace before and after tags in the TAG_NAMES_STRIP_WHITESPACE
+  # list.
+  def strip_whitespace(root)
+    traverse(root) do |node|
+      next unless node.text? || node.cdata?
+
+      # This text node immediately follows the closing tag of either its
+      # previous sibling, or its parent if it has no previous sibling. If we're
```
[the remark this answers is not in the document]
tickinginstant: Right, sorry! The comment is incorrect, it's either the closing tag of the previous sibling or the opening tag of the parent.

lib/paragraph_maker.rb
tickinginstant: There's this pre-existing spec: https://github.com/otwcode/otwarchive/blob/d74d404f052b4e60e223c712fff619844695806d/spec/lib/html_cleaner_spec.rb#L961-L966
Do you want me to add more?

lib/paragraph_maker.rb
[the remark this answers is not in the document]
redsummernight: You should update the `rstrip` comment as well.

lib/paragraph_maker.rb
redsummernight: Oh this works! No need to add more, I just didn't find it.

lib/html_cleaner.rb
tickinginstant: I chose `to_xhtml` because when I first wrote this with `to_s`, it introduced a test failure for this test:
https://github.com/otwcode/otwarchive/blob/d74d404f052b4e60e223c712fff619844695806d/features/other_a/parser.feature#L6-L21
The failure occurred not because it was parsing incorrectly, but because `to_s` adds a newline between the two characters when they're wrapped in a `<myroot>` tag:
```
2.7.3 :054 > Nokogiri::HTML.fragment("<myroot><p>1</p><p>2</p></myroot>").children.first.children.to_s
 => "<p>1</p>\n<p>2</p>" 
2.7.3 :055 > Nokogiri::HTML.fragment("<myroot><p>1</p><p>2</p></myroot>").children.first.children.to_xhtml
 => "<p>1</p><p>2</p>" 
```
Since `to_xhtml` matched the other code and didn't introduce the extra newline, it seemed like a reasonable choice to make. If you'd prefer I can switch to `to_s` (or some other output method) and fix up the test instead.

redsummernight: I'm trying to see if relying on XHTML output is necessary, because we may eventually want to move away from it. This is related to @potpotkettle's comment below:

> Can you also fix [AO3-5801](https://otwarchive.atlassian.net/browse/AO3-5801)? I have #4303 but the approach here may be better.

The way this module works could probably be simplified even more. Right now, it does:

- `add_paragraphs_to_text` (which uses `Nokogiri::HTML.fragment`)
- `Sanitize.clean` (which [uses `Nokogiri::HTML5.fragment`](https://github.com/rgrove/sanitize/blob/v6.0.0/lib/sanitize.rb#L135-L144))
- then another round of `Nokogiri::HTML::Document.fragment`

If you remove the third part and avoid `to_xhtml`, it might help with AO3-5801.

But this change should be fine as is. We can tackle AO3-5801 separately.

tickinginstant: I agree that it might be good to eventually move away from XHTML, since I suspect it's responsible for https://otwarchive.atlassian.net/browse/AO3-4599 and possibly some other known issues. I think that XHTML is probably most useful for pretty formatting, because otherwise you'd get an unreadable blob of HTML tags in the Edit Work form, but that's also what the [textarea_convert_html_to_newlines.rb](https://github.com/otwcode/otwarchive/blob/master/config/initializers/monkeypatches/textarea_convert_html_to_newlines.rb) monkeypatch is for, so I could be completely off-base.

👍  for tackling AO3-5801 separately.

redsummernight (approved):

PR #4390, opened 2022-11-21
AO3-3997 Replace email_veracity with email_format

potpotkettle:
# Pull Request Checklist

Since this PR mainly reduces code, existing tests seem to suffice. But if someone has an idea of additional tests, I'm happy to add.

## Issue

https://otwarchive.atlassian.net/browse/AO3-3997

## Purpose

- It appears like there is no reason to keep 2 email address validators so let's use the stricter one only.
- There is one related test that is a duplicate of another test. ("Create Invitation for invalid emails") While strictly not related to the issue, I believe it's a good opportunity to remove it.
- app/validators/email_validator.rb is probably not used, unless it's automagically used somewhere. This can be removed, too.

## Testing Instructions

See the JIRA link above.

## References

As to why there were validators doing seemingly similar things, I'm guessing that they were introduced independently from each other, creating an unintentional overlap. (I have no knowledge beyond the commit history, so I might be way off in this assessment, though.)

- AO3-2839 / #3378  <--- this is where email_format_validator.rb comes from (taken from authlogic)
- commit 1b30ebed3da3c54ae904761f7c9e66b038df8d0c <--- this is where email_veracity_validator.rb comes from (much older than email_format)
- ...which in turn came from a plugin called [Validates Email Veracity Of (validates_email_veracity_of)](https://github.com/Empact/validates_email_veracity_of) in commit dfc84d4db5edabc5e1542d95d1c90472626cd363

## Credit

Potpotkettle (they/them)


spec/spec_helper.rb
```diff
@@ -136,10 +136,7 @@
   # instead of true.
   config.use_transactional_fixtures = true
 
-  # For email veracity checks
-  BAD_EMAILS = ['Abc.example.com', 'A@b@c@example.com', 'a\"b(c)d,e:f;g<h>i[j\k]l@example.com', 'this is"not\allowed@example.com', 'this\ still\"not/\/\allowed@example.com', 'nodomain', 'foo@oops'].freeze
-  # For email format checks
-  BADLY_FORMATTED_EMAILS = ['ast*risk@example.com', 'asterisk@ex*ample.com'].freeze
+  BAD_EMAILS = ['Abc.example.com', 'A@b@c@example.com', 'a\"b(c)d,e:f;g<h>i[j\k]l@example.com', 'this is"not\allowed@example.com', 'this\ still\"not/\/\allowed@example.com', 'nodomain', 'foo@oops', 'ast*risk@example.com', 'asterisk@ex*ample.com'].freeze
```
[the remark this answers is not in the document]
potpotkettle: I think some of them (`a\"... ` and `this is"...`) are better left in single quotes. Others can be converted into double quotes, but would the mixture be preferable?

Bilka2:
> app/validators/email_validator.rb is probably not used, unless it's automagically used somewhere. This can be removed, too.

Indeed, it's already set to be removed with #4351 -> https://otwarchive.atlassian.net/browse/AO3-6368. I don't know if it's worth the effort to remove the file deletion from this PR.

potpotkettle:
@Bilka2 Thank you for pointing that out. I didn't know about the existing (and already approved) PR. I hope it gets merged first - after that either Git will automatically sort it out or I can resolve conflict. Better to separate things especially if they were separately reported in the first place.

config/locales/phrase-exports/pt-BR.yml
```diff
@@ -979,7 +979,3 @@ pt-BR:
         allow_blank: deveria ter o formato de um endereço de e-mail. Por favor utilize
           um endereço diferente ou deixe em branco.
         no_blank: deveria ter o formato de um endereço de e-mail.
-      veracity:
```
redsummernight, line 982: You should leave out changes to phrase-exports/ files - we let Phrase / the Translation team resolve the differences, for now.

app/models/admin_blacklisted_email.rb
```diff
@@ -1,7 +1,7 @@
 class AdminBlacklistedEmail < ApplicationRecord
   before_validation :canonicalize_email
 
-  validates :email, presence: true, uniqueness: true, email_veracity: true
+  validates :email, presence: true, uniqueness: true, email_format: { case_sensitive: true }
```
redsummernight, line 4: `case_sensitive` is not an option in EmailFormatValidator (only `options[:allow_blank]` is checked there). Just `true` should be fine.

spec/spec_helper.rb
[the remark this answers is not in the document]
redsummernight: Yeah, a mixture is fine.

redsummernight:
I linked the 2 issues in Jira so we remember to merge #4351 first.

app/models/admin_blacklisted_email.rb
potpotkettle: Thanks for pointing out the problem. That was my failed attempt at addressing this deprecation warning:
```   DEPRECATION WARNING: Uniqueness validator will no longer enforce case sensitive comparison in Rails 6.1. To continue case sensitive comparison on the :email attribute in AdminBlacklistedEmail model, pass `case_sensitive: true` option explicitly to the uniqueness validator. (called from validate_each at /usr/local/bundle/bundler/gems/globalize-76b808b13d15/lib/patches/active_record/rails5_1/uniqueness_validator.rb:38)```

However, I mistakenly added `case_sensitive` to `email_format` where I should added it to `uniqueness`. So it seems like a backward compatible validation (assuming it is supposed to be case-sensitive) would be:

`  validates :email, presence: true, uniqueness: { case_sensitive: true }, email_format: true`

redsummernight: Let's use `case_sensitive: false` for the uniqueness check. That's our usual approach for emails, for example:

https://github.com/otwcode/otwarchive/blob/e7c7da0a33a554cfde124ef1ad2e3df5ce005470/app/models/invite_request.rb#L5

https://github.com/otwcode/otwarchive/blob/e7c7da0a33a554cfde124ef1ad2e3df5ce005470/app/models/user.rb#L201

potpotkettle: Sure. It's been done in commits updated today.

config/locales/phrase-exports/pt-BR.yml
potpotkettle: Thanks - removed.

redsummernight (approved):

PR #4389, opened 2022-11-18
AO3-6421 Temporarily add an extra sanitizing step.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6421

## Purpose

- Adds a stripped down version of the `sanitize_value` function that removes all of the code that changes formatting, leaving only the code that sanitizes, and call it in `sanitize_field` to add an extra sanitizing step right before display.

PR #4391, opened 2022-11-21
AO3-6421 Rewrite add_paragraphs_to_text, and add a rake task for resanitizing all fields.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6421

Also fixes:
https://otwarchive.atlassian.net/browse/AO3-3769

## Purpose

- Added back the `ParagraphMaker` code introduced in #4382 , with some tweaks to which tags are in which categories to try to make the parser behave better on some examples used to test the parser. (In particular: `img` should be wrapped in paragraph tags, `address` should not be, various invalid tags like `style`, `script`, and `noscript` should be skipped and should ignore whitespace nearby, a bunch of other invalid tags like `button`, `input`, `label`, and `textarea` should be wrapped in paragraph tags, etc.)
- Removed the `just_sanitize` function introduced in #4389, as well as a couple of other unused functions.
- Modified `sanitize_field` so that if the `sanitizer_version` is out of date, it'll still rerun the sanitizer, but will *never* save the results to the database, because that apparently crashes the servers.
- Fixed up a few existing tasks that broke as a result (because they relied on `sanitize_field` saving when sanitizing).
- Added a rake task + job for resanitizing all existing fields. The job uses a new `resanitize` resque queue, so that it can be throttled separately from all other resque jobs.
- Removed `screencast` from the list of fields allowing HTML, since it's apparently supposed to be a URL.
- Re-enabled the `update_sanitizer_version` callback for the `Question` class.
- Added a temporary fix for `AbuseReport`s having the wrong type for `summary_sanitizer_version`.


lib/tasks/after_tasks.rake
```diff
@@ -572,8 +572,8 @@ namespace :After do
       end
       if chapter.content.match /<(embed|iframe)/
         begin
-          chapter.content_sanitizer_version = -1
-          chapter.sanitize_field(chapter, :content)
+          content = chapter.sanitize_value(:content, chapter.content)
+          chapter.update_attribute(:content, content)
```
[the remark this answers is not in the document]
redsummernight: We generally keep them around in case someone needs to update an older instance of the site (admittedly probably pretty rare). Your fixes for the tasks look good, the tasks can stay.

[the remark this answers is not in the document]
tickinginstant: Since the three tasks are centered around resanitizing chapters, you can actually update an older instance of the site by running `rake resanitize:all` instead of the three individual tasks. But I've already done the work of updating them, so I don't mind leaving them in.

spec/lib/tasks/resanitize.rake_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+
+    context "when the #{field} attribute has an old sanitizer version" do
+      it "removes invalid content and updates the sanitizer version" do
+        record.update_columns(field => "<invalid>",
+                              field_sanitizer_version => ArchiveConfig.SANITIZER_VERSION - 1)
+
+        expect do
+          perform_enqueued_jobs { subject.invoke }
+          record.reload
+        end.to output.to_stdout
```
redsummernight, line 17: `output.to_stdout` is not checking for any string specifically, can we remove it?

spec/lib/tasks/resanitize.rake_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+      it "doesn't change the content in other languages when resanitizing" do
+        translation.update_columns(content: "<invalid>")
+        record.update_columns(content_sanitizer_version: ArchiveConfig.SANITIZER_VERSION - 1)
+
+        expect do
+          perform_enqueued_jobs { subject.invoke }
+          record.reload
+          translation.reload
+        end.to output.to_stdout
+          .and change { record.content_sanitizer_version }
```
redsummernight, line 115: Can you add `.to(ArchiveConfig.SANITIZER_VERSION)` here?

spec/lib/tasks/resanitize.rake_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+      let(:locale) { create(:locale) }
+      let(:translation) { question.translations.find_by(locale: locale.iso) }
+
+      before do
+        I18n.with_locale(locale.iso) do
+          question.update(content: Faker::Lorem.paragraph)
+        end
+      end
+
+      it "doesn't change the content in other languages when resanitizing" do
```
redsummernight, line 106: Should this work, considering we're iterating over all `Question::Translation` records?

spec/lib/tasks/resanitize.rake_spec.rb
tickinginstant: I added it because it suppresses the rake task's printouts when running tests, but if you're okay with more verbose tests I can drop it.

lib/paragraph_maker.rb
```diff
@@ -0,0 +1,270 @@
...
+
+  # Tags whose content we don't touch
+  TAG_NAMES_TO_SKIP = %w[
+    a abbr acronym address audio button dl embed figure h1 h2 h3 h4 h5 h6 hr
+    iframe img input label ol map math noscript object p pre script select
+    source style svg table textarea track video ul
+  ].freeze
+
+  # Tags that need to go inside p tags
+  TAG_NAMES_TO_WRAP = %w[
```
redsummernight, line 15: > various invalid tags like `style`, `script`, and `noscript` should be skipped and should ignore whitespace nearby, a bunch of other invalid tags like `button`, `input`, `label`, and `textarea` should be wrapped in paragraph tags, etc.

Is this to make sure there's no superfluous paragraphs / breaks around the invalid tags once they're removed by Sanitize?

spec/lib/tasks/resanitize.rake_spec.rb
redsummernight: Nice use of that. It's fine to stay - there's a lot more stdout prints than most tested rake tasks right?

spec/lib/tasks/resanitize.rake_spec.rb
tickinginstant: We're iterating over all `Question::Translation` records, but we only resanitize when the `content_sanitizer_version` is lower than the `ArchiveConfig.SANITIZER_VERSION`. This test is checking that if you have two translations of the same question, and one of them has an old sanitizer version, only the translation with the old sanitizer version will be resanitized. Should I rephrase the test to try to make it clearer what the purpose is?

redsummernight: Yeah, I missed that. Something like "doesn't change translations with an up-to-date content sanitizer version"?

spec/lib/tasks/resanitize.rake_spec.rb
tickinginstant: It's not too big (~5-10 lines per test), but because it's being run 60+ times, it adds up.

lib/paragraph_maker.rb
tickinginstant: Yeah, that was my attempt to make it so that invalid tags wouldn't produce so many empty paragraph tags. My approach was to skip and ignore whitespace for any tags that would be completely removed, and skip and add paragraphs around most of the inline tags that might leave text behind.

(`input` can't leave text behind, but I put it in the latter category rather than the former because it's more similar to `button` than `script`.)

redsummernight: I feel like we shouldn't have a mix of tags allowed by the Sanitize config, and tags that are not. Should we keep them in their own lists and add them into `TAG_NAMES_TO_WRAP` as needed?

In the case of

> any tags that would be completely removed

I think you can use `Sanitize::Config::ARCHIVE[:remove_contents]` (or `Sanitize::Config::DEFAULT[:remove_contents]`).

tickinginstant: How's this reorganization?

redsummernight: Looks great, thanks!

redsummernight (approved):

redsummernight:
We need to do this [testing step](https://otwarchive.atlassian.net/browse/AO3-6421):
> Before deployment, find or create a work with complex HTML.

before merging.

sarken:
We've got testy's Sliding Doors work and I just posted fresh copies of the sanitizer test and the RTE test, and then threw together a chapter that's got multiple paragraphs, a heading, nested lists, and address with abbr tags all inside a blockquote. Is there anything else we want to cover?

app/jobs/resanitize_batch_job.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    # Go through each record in the batch, and check the sanitizer version of
+    # all the fields that need processing. If any field has a sanitizer version
+    # less than the archive's SANITIZER_VERSION, run the sanitizer on the field
+    # and re-save it to the database.
+    klass.where(id: ids).each do |record|
+      record.with_lock do
+        fields_to_sanitize.each do |field, sanitizer_version|
+          next if record[sanitizer_version] &&
+                  record[sanitizer_version] >= ArchiveConfig.SANITIZER_VERSION
```
redsummernight, line 34: Because #4382 was live on production for a short time, there are a number of records in the database with the sanitizer version at 2 right now. Those won't be reprocessed by this rake.

Should we change `SANITIZER_VERSION` to 3?

redsummernight:
> We've got testy's Sliding Doors work and I just posted fresh copies of the sanitizer test and the RTE test, and then threw together a chapter that's got multiple paragraphs, a heading, nested lists, and address with abbr tags all inside a blockquote. Is there anything else we want to cover?

@sarken you can add the attributes example from [AO3-6421](https://otwarchive.atlassian.net/browse/AO3-6421). I can't think of anything else.

app/jobs/resanitize_batch_job.rb
tickinginstant: Sure, I can do that if you like.

sarken:
@redsummernight Added the title attributes from that issue!

app/jobs/resanitize_batch_job.rb
redsummernight: Thanks!

PR #4397, opened 2022-11-25
AO3-6278 Fix word counter to not count pipe characters as words

GoodGithubAccount:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6278

## Purpose

Stops matching pipe characters as words.

## Testing Instructions

Follow the guideline outlined in the jira issue, though do note that "foo bar |" should result in 2, not 3 as the issue says.
The new word_counter_spec test, which has been updated to include pipe characters in the symbol test, should also verify everything works as intended.

## Credit

GoodGithubAccount (he/him)


sarken:
Hi, GoodGithubAccount!

spec/lib/word_counter_spec.rb
```diff
@@ -45,7 +45,7 @@
     expect(word_counter.count).to eq(2)
   end
 
-  %w(* ~ !? - ~* ~!).each do |char|
+  %w(* ~ !? - ~* ~! |+).each do |char|
```
redsummernight, line 48: Why is it `|+` in this case, instead of just `|`?

spec/lib/word_counter_spec.rb
```diff
@@ -85,4 +85,9 @@
     word_counter.text = "\“嘿Bob,\” Alice说，‘啊？！？’"
     expect(word_counter.count).to eq(5)
   end
+
+  it "should count words in mixed languages with symbols correctly" do
+    word_counter.text = "\“嘿Bob,\” Alice说，‘啊？！？’ 说—嘿 嘿|bob|说"
```
redsummernight, line 90: Let's keep the text shorter so it's easier to see the intent, e.g. just `嘿|bob|嘿`.

redsummernight (commented):
The fix makes sense to me! Just a few more things:

- Can you fix the Hound/Rubocop warnings?
- We'll need to reset the word count on every work for the update to apply. There's a task for that in lib/tasks/work_tasks.rake, but it requires a specific language. Can you update the task so that if a language argument is given, it should process works with that language, but if not, it should process all works?

spec/lib/word_counter_spec.rb
GoodGithubAccount: > The fix makes sense to me! Just a few more things:
> 
> * Can you fix the Hound/Rubocop warnings?
> * We'll need to reset the word count on every work for the update to apply. There's a task for that in lib/tasks/work_tasks.rake, but it requires a specific language. Can you update the task so that if a language argument is given, it should process works with that language, but if not, it should process all works?


Yeah both should be no problem, I'll fix it after exams. Wasn't sure if you all wanted the hounds things updated, since I didn't really introduce new code, but just followed the style that was already present.

GoodGithubAccount: > Why is it `|+` in this case, instead of just `|`?

There's really no reason, just followed the previous examples. Now that I look at it again and think about it, I don't even understand why `!?` `~*` and `~!` are there since it goes by char anyway.

redsummernight: > Yeah both should be no problem, I'll fix it after exams. Wasn't sure if you all wanted the hounds things updated, since I didn't really introduce new code, but just followed the style that was already present.

In general we fixed warnings on lines added/modified and leave the rest alone. No rush and good luck on your exams!

redsummernight (approved):
Looks good, thank you!

PR #4400, opened 2022-11-28
AO3-6317 Allow support admins to delete works and edit FNOK

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6317

## Purpose

- Remove editing FNOK from UserManager, and add a new endpoint/form covered by a new permission in UserPolicy.
- Update the FNOK model to use Rails associations.
- Add a new policy class for UserManager, to check that PAC admins can perform all actions (add notes, suspend, etc.) but support admins can only add notes.
- Allow support admins to delete works, but not other types of user creations (bookmarks, external works, series).
- Fix redirect after an admin successfully deletes a bookmark.
- Fix redirect after an admin successfully adds notes, etc.
- Enforce admin permissions on editing work tags and external works.
- Update text on both the PAC form and the Support form.

## Testing Instructions

See issue.

### DIFF
--- a/app/controllers/admin/admin_users_controller.rb
+++ b/app/controllers/admin/admin_users_controller.rb
@@ -80,7 +80,8 @@ class Admin::AdminUsersController < Admin::BaseController
       return
     end
 
-    fnok = @user.build_fannish_next_of_kin(kin: kin, kin_email: kin_email)
+    fnok = @user.build_fannish_next_of_kin if fnok.blank?
+    fnok.assign_attributes(kin: kin, kin_email: kin_email)
     if fnok.save
       flash[:notice] = ts("Fannish next of kin was updated.")
       redirect_to admin_user_path(@user)
--- a/app/controllers/admin/admin_users_controller.rb
+++ b/app/controllers/admin/admin_users_controller.rb
@@ -43,7 +43,6 @@ class Admin::AdminUsersController < Admin::BaseController
   end
 
   # GET admin/users/1
-  # GET admin/users/1.xml
   def show
     @user = authorize User.find_by!(login: params[:id])
     @hide_dashboard = true
--- a/app/controllers/admin/admin_users_controller.rb
+++ b/app/controllers/admin/admin_users_controller.rb
@@ -45,9 +45,8 @@ class Admin::AdminUsersController < Admin::BaseController
   # GET admin/users/1
   # GET admin/users/1.xml
   def show
+    @user = authorize User.find_by!(login: params[:id])
     @hide_dashboard = true
-    @user = User.find_by!(login: params[:id])
-    authorize @user
     @log_items = @user.log_items.sort_by(&:created_at).reverse
   end
 
@@ -67,20 +66,47 @@ class Admin::AdminUsersController < Admin::BaseController
     redirect_to request.referer || root_path
   end
 
+  def update_next_of_kin
+    @user = authorize User.find_by!(login: params[:user_login])
+    fnok = @user.fannish_next_of_kin
+    kin = User.find_by(login: params[:next_of_kin_name])
+    kin_email = params[:next_of_kin_email]
+
+    if kin.blank? && kin_email.blank?
+      if fnok.present?
+        fnok.destroy
+        flash[:notice] = ts("Fannish next of kin was removed.")
+      end
+      redirect_to admin_user_path(@user)
+      return
+    end
+
+    fnok = @user.build_fannish_next_of_kin(kin: kin, kin_email: kin_email)
+    if fnok.save
+      flash[:notice] = ts("Fannish next of kin was updated.")
+      redirect_to admin_user_path(@user)
+    else
+      @hide_dashboard = true
+      @log_items = @user.log_items.sort_by(&:created_at).reverse
+      render :show
+    end
+  end
+
   def update_status
-    @user = User.find_by(login: params[:user_login])
-    authorize @user
-    @user_manager = UserManager.new(current_admin, params)
+    @user = User.find_by!(login: params[:user_login])
+
+    # Authorize on the manager, as we need to check which specific actions the admin can do.
+    @user_manager = authorize UserManager.new(current_admin, @user, params)
     if @user_manager.save
       flash[:notice] = @user_manager.success_message
-      if params[:admin_action] == "spamban"
+      if @user_manager.admin_action == "spamban"
         redirect_to confirm_delete_user_creations_admin_user_path(@user)
       else
-        redirect_to request.referer || root_path
+        redirect_to admin_user_path(@user)
       end
     else
       flash[:error] = @user_manager.error_message
-      redirect_to request.referer || root_path
+      redirect_to admin_user_path(@user)
     end
   end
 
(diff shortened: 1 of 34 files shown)


app/policies/user_creation_policy.rb
```diff
@@ -1,35 +1,32 @@
...
-  EDIT_ROLES = %w[superadmin support policy_and_abuse].freeze
-  HIDE_ROLES = %w[superadmin policy_and_abuse].freeze
-  SPAM_ROLES = %w[superadmin policy_and_abuse].freeze
 
-  def self.can_destroy_creations?(user)
-    self.new(user, nil).can_destroy_creations?
-  end
+  # Roles that allow destroying only Works.
+  #
+  # Include support admins for handling FNOK requests.
```
sarken, line 9: I was like, "Wait, what? FNOK doesn't relate to deleting works," but I think I see where the confusion came from. The issue wants us to allow Support to destroy works not because they're taking over FNOK management, but because they're also taking over management of duplicate works.

However, I do have a little lingering confusion -- why does PAC need to be here? Does `DESTROY_ROLES` not cover destroying works, only other types of creations?

redsummernight: I'll update the comment.

`DESTROY_ROLES` covers all types of user creations, so having PAC and super admins in `DESTROY_WORK_ROLES` is to be technically true, but it's not necessary. Should I remove both?

sarken: I think the relationship between `DESTROY_ROLES` and `DESTROY_WORK_ROLES` is clearer if we just let it be support, yeah. Maybe we should also clarify the comment on `DESTROY_ROLES` to say it's for all types of user creations just to be extra clear.

features/admins/users/admin_fnok.feature
```diff
@@ -9,11 +9,11 @@ Feature: Admin Fannish Next Of Kind actions
       | login    | password   |
       | harrykim | diesalot   |
       | libby    | stillalive |
-      And I am logged in as a "policy_and_abuse" admin
+      And I am logged in as a "support" admin
```
sarken, line 12: Should we be using scenario outlines instead so we can test every kind of admin (superadmin, support, policy_and_abuse) who is supposed to be able to edit FNOKs? Or am I just not seeing the tests that check that?

sarken (commented):
Couple more nitpicks for you! No rush.

tickinginstant:
There are so many "applies only to X" comments in the `UserCreationPolicy` class that I can't help but wonder if it'd be simpler to start subclassing it for each of the user creation types. e.g.

```ruby
class UserCreationPolicy < ApplicationPolicy
  def show_admin_options?
    destroy? || hide? || edit?
  end

  def destroy?
    user_has_roles?(%w[superadmin policy_and_abuse])
  end

  def hide?
    user_has_roles?(%w[superadmin policy_and_abuse])
  end

  def show_ip_address?
    user_has_roles?(%w[superadmin policy_and_abuse])
  end
end

class BookmarkPolicy < UserCreationPolicy
end

class ExternalWorkPolicy < UserCreationPolicy
  def update?
    user_has_roles?(%w[superadmin policy_and_abuse])
  end
end

class SeriesPolicy < UserCreationPolicy
end

class WorkPolicy < UserCreationPolicy
  def show_admin_options?
    super || edit_tags? || set_spam?
  end

  # Allow admins to edit work tags and languages.
  # Include support admins due to AO3-4932.
  def update_tags?
    user_has_roles?(%w[superadmin policy_and_abuse support])
  end

  alias edit_tags? update_tags?

  # Support needs to be able to delete duplicate works:
  def destroy?
    super || user_has_roles?(%[support])
  end

  def set_spam?
    user_has_roles?(%w[superadmin policy_and_abuse])
  end
end
```
That'd also let you use more idiomatic Pundit syntax (`policy(item).destroy?`), instead of having to use a helper.

(Also, not really in scope, but I noticed it while looking through this PR: it looks like the policy on editing the work tags only applies to whether the link appears, not whether the action is actually possible. If the admin manually enters the `/works/:id/edit_tags` link, they're still allowed to use it even if they're not supposed to have permission.)

redsummernight:
> start subclassing it for each of the user creation types

Good idea, let me make these changes. Thanks!

features/admins/users/admin_fnok.feature
redsummernight: This was missed, I'll add specs.

app/views/feedbacks/new.html.erb
```diff
@@ -25,6 +25,7 @@
     <p><%= t(".abuse.reports", contact_link: link_to(t(".abuse.contact"), new_abuse_report_path)).html_safe %></p>
     <p><%= t(".reportable.intro") %></p>
     <ul>
+      <li><%= t(".reportable.fnok.concerning", fnok_link: link_to(t(".reportable.fnok.fnok"), tos_faq_path(anchor: "next_of_kin"))).html_safe %></li>
```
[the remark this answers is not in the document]
redsummernight: I've added this to [Internationalization (i18n) Standards](https://github.com/otwcode/otwarchive/wiki/Internationalization-(i18n)-Standards) as well.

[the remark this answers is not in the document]
sarken: Good idea -- thanks!

app/controllers/admin/admin_users_controller.rb
```diff
@@ -67,20 +65,47 @@ def update
...
+      if fnok.present?
+        fnok.destroy
+        flash[:notice] = ts("Fannish next of kin was removed.")
+      end
+      redirect_to admin_user_path(@user)
+      return
+    end
+
+    fnok = @user.build_fannish_next_of_kin(kin: kin, kin_email: kin_email)
+    if fnok.save
```
tickinginstant, line 84: I just did some experiments to check out how this works. Because `fannish_next_of_kin` is a `has_one`, calling `build_fannish_next_of_kin` will immediately delete the old fannish next-of-kin from the database, in a separate transaction from the call to `fnok.save`. So if the new fannish next-of-kin is invalid, it will result in the old next of kin being deleted without the new one being saved. Maybe there should be some kind of message telling the admin that the old next of kin has been removed? Or maybe this should be modified to avoid removing the old next of kin unless the new next of kin saves properly?

tickinginstant (changes requested):
Ooh, I didn't notice that the `ExternalWork` check only applied to the edit action, not the update action. Good catch!

Unfortunately, I think there might need to be some changes to the new fannish next-of-kin code. But it's looking pretty good otherwise.

app/controllers/admin/admin_users_controller.rb
redsummernight: Thanks for catching that! I'll change it to

> Or maybe this should be modified to avoid removing the old next of kin unless the new next of kin saves properly?

features/step_definitions/generic_steps.rb
```diff
@@ -125,6 +125,10 @@ def assure_xpath_not_present(tag, attribute, value, selector)
   find_field("#{labeltext}").value.should == content
 end
 
+Then "the {string} input should be blank" do |label|
+  expect(find_field(label).value).to be_blank
```
redsummernight, line 129: `value` can be nil for Rack::Test if [the attribute is omitted](https://web.archive.org/web/20220801180017/http://www.opinionatedprogrammer.com/2011/03/capybara-input-tag-without-value-attribute/), so I can't just do [`field should contain "^$"`](https://github.com/otwcode/otwarchive/blob/592328d48c4caaaaca9e3e87ae759cfb71a8e26b/features/step_definitions/web_steps.rb#L228-L248).

tickinginstant (approved):
LGTM!

PR #4401, opened 2022-11-29
AO3-6433 Clean up old After tasks

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6433

## Purpose

Update code comments and remove tasks (and associated tests) more than two years old from `after_tasks.rake`, which were pretty unnecessary and dragging our test coverage percentage down.

However, we did want to keep `reset_word_counts` task, which might be useful and has been relocated to `work_tasks.rake` (its tests have been relocated as well).

I've also updated the [AO3 Coding: Your First Pull Request](https://github.com/otwcode/otwarchive/wiki/AO3-Coding:-Your-First-Pull-Request) page to remove references to running `bundle exec rake After` whenever you pull from upstream, because... we kinda stopped using it that way ten years ago. I also removed the bits at the bottom that assumed we'd be using it that way.

## Testing Instructions

No testing!


tickinginstant (approved):
🥳

PR #4403, opened 2022-11-30
Improve the installable web app experience

vaaas:
# Pull Request Checklist

## Issue

N/A (this is an interest check)

## Purpose

AO3 does not have a native app, and if the user desires a more dedicated shortcut on a mobile device, they are prompted to use the "Add to Home Screen" functionality that many mobile web browsers provide.

This experience can be improved by providing a minimal web manifest and service worker (for Chrome):

- browser and device chrome (status, tab bars) will have AO3's theme colours
- a more high-res icon can be used
- some browser chrome is hidden when installed, giving a more "native" in-app look and feel
- other goodies, such as optional splash screen, better app name, etc.

The overall idea is meeting some of the users' requests halfway by providing a more "app" like experience without committing resources to native Android and iOS app development.

## Notes

- I've added the manifest in `application.html.erb` and the service worker loader in `application.js`, so they load in every (or most) page. Technically serving them just in the home page is fine, so we could go for a more surgical change.
- Service worker is not added in `javascripts` because of security reasons in the service worker API: SWs have scopes defined by `scope` upon registration. It can still be in `javascripts` if the server provides a relevant header, but I considered that a less clean solution than simply putting it in the root folder.
- service worker does nothing, it's the most minimal implementation I'm aware of so that we pass Chrome's requirements and it doesn't complain
- I had to add a new logo based on `logo.png` because Chrome requires web apps to have square icons. I did my best to minimise its size by using just enough indexed colours that the differences are imperceptible. It ends up being smaller than `logo.png` by a little, so it shouldn't be a huge burden when serving...

- perhaps we don't want the third commit adding theme colours, as it may interfere visually with the user's selected theme?
- perhaps we want to flip background and foreground colours so that the icon is white against red. For users in dark mode, the white background will flash brightly, which can be painful depending on light sensitivity.
- splash screen is automatically generated, but we can provide an even larger (or svg) icon for it that looks cooler

## Testing Instructions

Given:

- Website is served in a HTTPS context with a valid SSL certificate.
- User agent supports Service Workers.

Then:

- User can add the site to Home Screen.
- An icon is created with the AO3 logo and a white background
- The app has the short name "AO3"
- While loading, a longer name can be seen in a splash screen: "An Archive of Our Own"
- While using the app, the status bar is red in colour

## References

https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Installable_PWAs

## Credit

Vasileios Pasialiokis (he/him)


sarken:
Due to a backlog of high-priority issues, we are not currently accepting pull requests for anything that isn't listed in our [Jira issue tracker](https://otwarchive.atlassian.net/). Please review and follow the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md) before opening any additional pull requests.

PR #4411, opened 2022-12-09
AO3-6436 Disallow Common Crawl bot

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6436

## Purpose

Prevents CCBot from crawling the archive.

## Testing Instructions

Refer to Jira.

## Credit

Sarken, she/her

### DIFF
--- a/public/robots.public.txt
+++ b/public/robots.public.txt
@@ -31,8 +31,8 @@ Disallow: /*?*user_id
 Disallow: /*?*pseud=
 Disallow: /people?*show=
 
-User-agent: Slurp
-Crawl-delay: 30
-
 User-agent: CCBot
 Disallow: /
+
+User-agent: Slurp
+Crawl-delay: 30
--- a/public/robots.public.txt
+++ b/public/robots.public.txt
@@ -33,3 +33,6 @@ Disallow: /people?*show=
 
 User-agent: Slurp
 Crawl-delay: 30
+
+User-agent: CCBot
+Disallow: /


public/robots.public.txt
```diff
@@ -33,3 +33,6 @@ Disallow: /people?*show=
 
 User-agent: Slurp
 Crawl-delay: 30
+
+User-agent: CCBot
+Disallow: /
```
potpotkettle, line 38: I could be wrong but isn't it better to keep sections with `Crawl-delay` (and without `Disallow`) at the end of the file?

The rationale (if I recall it correctly) is that because `Crawl-delay` is not universally supported, some crawlers may skip lines containing it, and may interpret the previous line and the next line as a connected section in an unexpected way.

(I'm not completely sure about this - the only document I can find about it is in Japanese only: https://webtan.impress.co.jp/e/2022/11/04/43611 )

sarken: I also can't find any other source for that, but it won't hurt to change it, so I'll do that just in case!

redsummernight: Good call, this problem definitely applies to [Googlebot](https://developers.google.com/search/docs/crawling-indexing/robots/robots_txt#file-format).

> Google ignores invalid lines in robots.txt files, including the Unicode [Byte Order Mark](https://en.wikipedia.org/wiki/Byte_order_mark) (BOM) at the beginning of the robots.txt file, and use only valid lines.

So a file like:
```
User-agent: Googlebot
Crawl-delay: 30

User-agent: CCBot
Disallow: /
```
will be interpreted by Google as [one rule applying to multiple agents](https://developers.google.com/search/docs/crawling-indexing/robots/robots_txt#grouping-of-lines-and-rules):
```
User-agent: Googlebot
User-agent: CCBot
Disallow: /
```
CCBot shouldn't have this problem because it does recognize [`crawl-delay`](https://commoncrawl.org/big-picture/frequently-asked-questions/), but it's good to keep that in mind and avoid ambiguity.

redsummernight (approved):

public/robots.public.txt
sarken: I'm going to update the commit messages with that link.

PR #4415, opened 2022-12-21
AO3-6440 Add Recoverable, Lockable, and separate password length config to admins

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6440

## Purpose

As described:
* Allows admins to reset their password using the Devise Recoverable module.
* Abuses aforementioned module to send email requesting that admins set a new password before their first login. 
* Updates the admin creation script to take into account that admins will no longer be given account information by database admins.

Not exactly as planned:
* Allows database admins to lock admins out of their accounts, _but_ uses the Devise Lockable module to do it rather than rolling our own solution with an `allow_resets` column. Lockable comes with the wonderful bonus of actually signing the locked account out.

Not even in the issue:
* Allows the minimum and maximum lengths of admin passwords to be configured separately from user passwords _and_ ups the minimum to something james_ and I were happier with.
* Updates user password view files (removes a colon in a label, adds some classes, removes a label that shouldn't have existed, fixes spacing, fixes heading level) so it's consistent with the admin password view code, which is more in line with our actual standards.

Questions:
* Do we like the Lockable approach? 
* Any changes we want to the scripts? We didn't have many specifics for what we wanted there.

## Testing Instructions

For the most part, Jira. I'll update the issue after this code has been reviewed to more accurately reflect the behavior.

## References

* [Devise Recoverable Docs](http://www.rubydoc.info/github/heartcombo/devise/main/Devise/Models/Recoverable)
* [Devise Lockable Docs](https://www.rubydoc.info/github/heartcombo/devise/main/Devise/Models/Lockable)
* [StackOverflow - Rails Devise: Set password reset token and redirect user](https://stackoverflow.com/q/5894191)

## Credit

Sarken, she/her


script/create_admin.rb
```diff
@@ -40,34 +40,27 @@ def multi_gets(all_text = "")
...
   a = Admin.find_or_initialize_by(login: "admin-#{login}")
+  success_message = a.new_record? ? "Created and notified" : "Updated"
   a.email = email if email.present?
   a.roles = roles if roles.present?
 
+  # If this is a new admin, we need to set a temporary password.
   if a.new_record?
-    # Create password only for new admins
-    password = `pwgen 8 1`.strip
+    password = `pwgen 10 1`.strip
```
zz9pzza, line 56: As an aside and as out of scope this pains me :)

```
2.7.3 :007 > (0...10).map{ [*'0'..'9',*'A'..'Z',*'a'..'z'].sample}.join
 => "Cj1Q0JCp0I"
``` 
from https://caseyscarborough.com/blog/2013/07/23/generating-alphanumeric-strings-in-ruby/

zz9pzza (changes requested):

script/create_admin.rb
redsummernight: I'd use [`SecureRandom.alphanumeric`](https://ruby-doc.org/stdlib-2.7.3/libdoc/securerandom/rdoc/SecureRandom.html) and avoid implementing our own.

If we do this, let's remove `pwgen` as a dev dependency as well:

https://github.com/otwcode/otwarchive/blob/b63fce67a3975ec0614bb535c75d76c8dafbda71/config/docker/Dockerfile#L9

sarken: Definitely game to do that! I saw someone use Devise's token generator to create the temporary, unshared keysmash password and I thought that was very clever, but SecureRandom is actually meant for such purposes, so we'll go with that.

script/create_admin.rb
```diff
@@ -40,34 +40,27 @@ def multi_gets(all_text = "")
 
 list = CSV.parse(multi_gets)
 
-puts "\nFor new admins, copy and paste each section into a separate file and upload to the admin's Vault:\n"
-
 admins = []
 list.each do |user|
   login = user[0].gsub(/\s+/, "")
   email = user[1]&.strip
```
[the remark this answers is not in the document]
sarken: In our case, human review! Database admins are the only folks who set admin emails (admins can't change their email addresses), and they will all be internal org addresses controlled by and on record with VolCom (who send us requests with usernames and emails). I'll add a clarifying comment like we discussed in Slack.

sarken:
Well. Those extra commits sure explain the weird complain I got when pulling from `upstream`. 

brb, unfucking...

sarken:
All better, I think!

redsummernight (approved):
Looks good to me! Let's update the labels once @zz9pzza re-reviews.

PR #4418, opened 2022-12-27
AO3-6442 Bypass URL validation for ficbook.net due to ip block

ellieyhcheng:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6442

## Purpose
Add bypass for `ficbook.net` because of IP block, similar to what is already done for `fanfiction.net`

## Testing Instructions
Same as in issue:

- Find a work on [Ficbook.net](http://ficbook.net/), such as https://ficbook.net/readfic/12909810.

- Go to https://test.archiveofourown.org/external_works/new and attempt to bookmark the Ficbook.net work; you receive the error "Url could not be reached. If the URL is correct and the site is currently down, please try again later."

- Go to https://test.archiveofourown.org/works/new and attempt to create a work with the Ficbook.net work in “Inspired By”; you receive the same error.

## References
N/A

## Credit
ellieyhc (she/her)


ellieyhcheng:
Are there any tests I should add for this bypass? I can't find any for bookmarking from `fanfiction.net`. I'm more than happy to add tests for both, but was wondering if that's desired since it's a bypass and not really an intended feature.

brianjaustin (commented):
Looks good to me! A couple thoughts for someone more official to weigh in on:
* It might be useful to extract this to a config option?
* Especially if the above happens, I think it would be good to add a test to https://github.com/otwcode/otwarchive/blob/master/spec/models/external_work_spec.rb. Maybe an easy way to poke the desired behaviour would be `WebMock.stub_request(:any, external_work.url).to_return(status: 500)`?

redsummernight (commented):
Yeah, let's add tests for bypassing both of these domains. +1 to @brianjaustin, you can add tests similar to https://github.com/otwcode/otwarchive/blob/b8f0950323f2b82eaf89ace957d0e739b6ad8051/spec/models/external_work_spec.rb#L29

We can stub a request to return 403, which is what ff.net returned for the Cloudflare check.

> It might be useful to extract this to a config option?

I'm fine with hardcoding for now, since we (knocks on wood) do this infrequently enough.

ellieyhcheng:
Sorry for the delay. But I've added the tests for the bypass, using hardcoded strings.

redsummernight (approved):
Thank you!

PR #4419, opened 2022-12-29
AO3-4120 Prevent comments caching stale username and pseud

brianjaustin:
Please do not review yet. I want to see if the CI behaviour matches something strange I see locally before marking as review-ready.

# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-4120

## Purpose

Update the cache key for single comments to avoid displaying a stale commenter name

## Testing Instructions

See Jira bug description for STR

## References

Epic: https://otwarchive.atlassian.net/browse/AO3-5051
Comment by @tickinginstant with suggested solution: https://otwarchive.atlassian.net/browse/AO3-4120?focusedCommentId=358747

## Credit

**Brian (they/he)**

### DIFF
--- a/features/other_a/pseuds.feature
+++ b/features/other_a/pseuds.feature
@@ -130,6 +130,7 @@ Scenario: Comments reflect pseud changes immediately
   Then I should see "before (myself)" within "h4.heading.byline"
 
   When I change the pseud "before" to "after"
+    And it is currently 10 seconds from now
     And I view the work "Interesting" with comments
   Then I should see "after (myself)" within "h4.heading.byline"
     And I should not see "before (myself)"
--- a/features/other_a/pseuds.feature
+++ b/features/other_a/pseuds.feature
@@ -117,3 +117,19 @@ Scenario: Manage pseuds - add, edit
     And I should see "My new fancy name (editpseuds)"
     And I should see "I wanted to add another fancy name"
     And I should not see "My new name (editpseuds)"
+
+Scenario: Comments reflect pseud changes immediately
+
+  Given the work "Interesting"
+    And I am logged in as "myself"
+    And I add the pseud "before"
+  When I set up the comment "Wow!" on the work "Interesting"
+    And I select "before" from "comment[pseud_id]"
+    And I press "Comment"
+    And I view the work "Interesting" with comments
+  Then I should see "before (myself)" within "h4.heading.byline"
+
+  When I change the pseud "before" to "after"
+    And I view the work "Interesting" with comments
+  Then I should see "after (myself)" within "h4.heading.byline"
+    And I should not see "before (myself)"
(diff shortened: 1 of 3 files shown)


features/other_a/pseuds.feature
```diff
@@ -117,3 +117,20 @@ Scenario: Manage pseuds - add, edit
...
+    And I am logged in as "myself"
+    And I add the pseud "before"
+  When I set up the comment "Wow!" on the work "Interesting"
+    And I select "before" from "comment[pseud_id]"
+    And I press "Comment"
+    And I view the work "Interesting" with comments
+  Then I should see "before (myself)" within "h4.heading.byline"
+
+  When I change the pseud "before" to "after"
+    And it is currently 10 seconds from now
```
brianjaustin, line 133: I added this because running this and the tests in `user_rename.feature` made this fail. I suspect something weird is happening in the cache (running `flush_all` from telnet made things happier), but if I add breakpoints and poke around, whatever timing issue it is doesn't occur. Seems a bit hacky to me, but also caching is hard :shrug:

tickinginstant: Yeah, it's because the cache is expired when `updated_at` changes, but `updated_at` is only accurate down to the second. So if you create a pseud and then change the pseud name within the same second, the code to expire the comment cache won't work without some kind of artificial delay.


Eventually I'd like to increase the precision of all timestamps, so that we don't have to keep adding delays to make the tests pass more consistently. But practically speaking, most people aren't going to experience that in production, so changing the precision on all those timestamps

brianjaustin: Thanks as always for the thorough explanation ticking! I've made your suggested change, and the tests are much happier now :tada:

features/other_a/pseuds.feature
```diff
@@ -117,3 +117,20 @@ Scenario: Manage pseuds - add, edit
...
+  Given the work "Interesting"
+    And I am logged in as "myself"
+    And I add the pseud "before"
+  When I set up the comment "Wow!" on the work "Interesting"
+    And I select "before" from "comment[pseud_id]"
+    And I press "Comment"
+    And I view the work "Interesting" with comments
+  Then I should see "before (myself)" within "h4.heading.byline"
+
+  When it is currently 2 seconds from now
```
redsummernight, line 132: You can use "1 second from now" to make it clear that the timestamp only needs to be different. 2 seconds leave us wondering. :)

features/other_a/pseuds.feature
```diff
@@ -117,3 +117,20 @@ Scenario: Manage pseuds - add, edit
...
+Scenario: Comments reflect pseud changes immediately
+
+  Given the work "Interesting"
+    And I am logged in as "myself"
+    And I add the pseud "before"
+  When I set up the comment "Wow!" on the work "Interesting"
+    And I select "before" from "comment[pseud_id]"
+    And I press "Comment"
+    And I view the work "Interesting" with comments
+  Then I should see "before (myself)" within "h4.heading.byline"
```
redsummernight, line 130: You should use `.comment h4.byline` here. `h4.heading.byline` can match the chapter byline as well.

redsummernight (approved):
Thank you!

PR #4422, opened 2023-01-03
AO3-6409 Make Mail::SMTP respect the enable_starttls_auto option even when it's false.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6409

## Purpose

Add a [monkeypatch](https://github.com/rails/rails/issues/44698#issuecomment-1069675285) to try to make `mail` 2.7.1 compatible with `net-smtp` 0.3.1  by calling `disable_starttls` when the `enable_starttls_auto` setting is set to false.

## Testing Instructions

Do something on staging that causes mail to be sent.

## Credit

h/t @redsummernight


redsummernight (approved):
Thanks!

PR #4424, opened 2023-01-05
AO3-6428 Fix downloads and attachments of works with no posted chapters

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6428

## Purpose

Fixes 500 error when deleting or downloading a posted work with zero posted chapters, with the side effect of always including draft chapters in deleted work downloads.

## Testing Instructions

Refer to Jira.

## Credit

Sarken, she/her

### DIFF
--- a/app/models/download.rb
+++ b/app/models/download.rb
@@ -7,6 +7,7 @@ class Download
     # TODO: Our current version of the mime-types gem doesn't include azw3, but
     # the gem cannot be updated without updating rest-client
     @mime_type = @file_type == "azw3" ? "application/x-mobi8-ebook" : MIME::Types.type_for(@file_type).first
+    @include_draft_chapters = options[:include_draft_chapters]
   end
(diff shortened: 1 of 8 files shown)


app/models/download.rb
```diff
@@ -115,7 +116,11 @@ def page_title
   end
 
   def chapters
-    work.chapters.order('position ASC').where(posted: true)
+    if @include_draft_chapters
+      work.chapters.order("position ASC")
+    else
+      work.chapters.order("position ASC").where(posted: true)
```
redsummernight, line 122: You can use the posted scope in some places:

https://github.com/otwcode/otwarchive/blob/b8f0950323f2b82eaf89ace957d0e739b6ad8051/app/models/chapter.rb#L51

but it's not a blocker.

redsummernight (approved):

PR #4425, opened 2023-01-06
AO3-6075 Make download chapter count display consistent

designateddolt:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6075 

## Purpose

What does this PR do?

As per Jira issue

## Testing Instructions

As per Jira issue

## References

## Credit


designateddolt:
Not sure how to internationalize a downloads page with t() as downloads aren't connected to the internet. Let me know how to do so if the internationalization needs modifying.

app/views/works/_work_module.html.erb
```diff
@@ -99,7 +99,7 @@
 
   <% if work.count_visible_comments > 0 %>
     <dt class="comments"><%= ts('Comments') %>:</dt>
-    <dd class="comments"><%= link_to work.count_visible_comments, work.chapter_total_display.to_i > 1 ? work_path(work, anchor: "comments", show_comments: true, view_full_work: "true") : work_path(work, anchor: "comments", show_comments: true) %></dd>
+    <dd class="comments"><%= link_to number_with_delimiter(work.count_visible_comments), work.chapter_total_display.to_i > 1 ? work_path(work, anchor: "comments", show_comments: true, view_full_work: "true") : work_path(work, anchor: "comments", show_comments: true) %></dd>
```
Bilka2, line 102: Accidental extra change?

designateddolt: i think i did it on purpose (although it was a while ago so i can't remember) because it looked weird without the delimiter and the actual non-download page has delimiters. can remove if unwanted though!

Bilka2: As far as I understand, the cache key in line 94 would have to be changed for this to work properly. So I think it would be better to remove the change. Additionally, there is AO3-5777 for adding the delimiters in this file.

sarken (changes requested):
>Not sure how to internationalize a downloads page with t() as downloads aren't connected to the internet. Let me know how to do so if the internationalization needs modifying.

You can leave it as-is for now, but `t()` should work just like any other Ruby code in these files. The downloads are generated by the app code, which can pull text from the locale files. The resultant static files should have the text filled in where the `t()` method was, just like they'll have, e.g., the chapter total filled in where `@work.chapter_total_display` was.

features/works/work_download.feature
```diff
@@ -109,8 +109,8 @@ Feature: Download a work
     And "Published:" should appear before "Completed"
     And "Completed:" should appear before "Chapters"
     # note difference with the work: in the work "Words" come before "Chapters"
-    And "Chapters:" should appear before "Words"
-    And "Words:" should appear before "Could be downloaded"
+    And "Words:" should appear before "Chapters:"
```
sarken, line 112: Can you remove the now-inaccurate comment above this, and the one on line 97 as well?

sarken (approved):

PR #4431, opened 2023-01-10
AO3-6455 Try to address deadlocks in readings code.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6455

## Purpose

1. Introduces the job `ReadingsToDatabaseJob` which handles the majority of the code writing readings to the database.
   - Modifies the code so that it calls `sscan` instead of `smembers`, to make it scaleable.
   - Reorders the batch by `(user_id, work_id)` so that the locks occur in a consistent order.
   - Calls `retry_on ActiveRecord::Deadlocked` so that if there are any deadlocks left in the code, we don't give up immediately.
2. Adds an extra `rename` step in the `Reading.update_or_create_in_database` method so that if the first call takes a long time, we won't end up with two workers trying to iterate through the same data at the same time. Instead, the second worker will work on the data added since the first worker started working.
3. Modifies `Reading.reading_object` in a few ways:
   - Make it take six arguments instead of a single JSON-encoded array (because `ReadingsToDatabaseJob` already has to parse the JSON to be able to sort, and there's no point in doing it twice).
   - Change the logic on the `last_viewed`/`major_version_read`/`minor_version_read` updates so that readings can be processed in any order.
   - Remove the code that updates `Reading:new`, because (a) we no longer have access to the JSON-encoded version, and (b) it used to delete the reading before the end of the transaction, which meant that if the transaction failed, it would produce https://otwarchive.atlassian.net/browse/AO3-5064. Now the deletion is a batch `srem` call in `ReadingsToDatabaseJob`, outside of the transaction, so this should hopefully reduce the incidence of lost readings.
4. Modifies the way that readings are deleted when users delete their accounts to try to make sure they're locked in the correct order.
5. Removes the unused work functions `update_major_version` and `update_minor_version`, because originally I handled version updates in `Reading.reading_object` in a different way that was incompatible with this implementation. I switched to a more general-purpose version later, but left in the code deletion because I like code deletion.


app/jobs/readings_to_database_job.rb
```diff
@@ -0,0 +1,57 @@
...
+    scan_readings_in_batches(key) do |batch|
+      process_batch(batch)
+      remove_readings(key, batch)
+    end
+  end
+
+  private
+
+  def batch_size
+    ArchiveConfig.READING_BATCHSIZE || 1000
```
redsummernight, line 14: Can you add `READING_BATCHSIZE` to config.yml so we don't need to set a default here?

app/models/reading.rb
```diff
@@ -25,27 +25,34 @@ def self.mark_to_read_later(work, user, toread)
...
-      Reading.transaction do
-        batch.each do |reading_json|
-          Reading.reading_object(reading_json)
-        end
-      end
-    end
+    return unless REDIS_GENERAL.exists("Reading:new")
+
+    # Get a unique key associated with this run:
+    counter = REDIS_GENERAL.incrby("Reading:job_id", 1)
```
redsummernight, line 31: Is this the same as `.incr("Reading:job_id")`?

tickinginstant: Yep, good catch!

redsummernight (approved):

redsummernight:
Thank you!

PR #4432, opened 2023-01-14
AO3-6456 AO3-6438 Add missing primary keys and exclude innodb_monitor from schema dumps 

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6456
https://otwarchive.atlassian.net/browse/AO3-6438

## Purpose

Add missing primary keys to:

- `roles_users` (no primary keys in any environment)
  https://github.com/otwcode/otwarchive/blob/17384eb5024f0658c3f307b451096e2d2301b7ba/db/schema.rb#L985

- `schema_migrations` (no primary keys but only in staging/production; Rails [sets column "version" as the primary key](https://github.com/rails/rails/blob/c7b76db7784151c4b47c22e5b234fb2b4c16021f/activerecord/lib/active_record/schema_migration.rb#L38-L40) these days)

Also exclude [`innodb_monitor`](https://mariadb.com/kb/en/xtradb-innodb-monitors/) from schema.rb dumps.

## Testing Instructions

Migrate up/down/up on staging.

### DIFF
--- a/config/initializers/active_record_schema_ignore_tables.rb
+++ b/config/initializers/active_record_schema_ignore_tables.rb
@@ -1,0 +1,4 @@
+ActiveRecord::SchemaDumper.ignore_tables += [
+  # MySQL/MariaDB internals
+  "innodb_monitor"
+]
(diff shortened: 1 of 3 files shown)


sarken (approved):
Looks good! (Found an existing issue for ignoring `innodb_monitor` and updated the PR/issue accordingly for you.)

PR #4433, opened 2023-01-14
AO3-6457 Improve the scalability of the StatCounter job.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6457

## Purpose

1. Introduce a library to make it easier to scan through redis sets in batches, and modify the existing `ReadingsToDatabaseJob` to use it.
2. Create a new abstract job class `RedisSetJob` with a linked job `RedisSetJobSpawner` that's designed around breaking Redis sets down into smaller batches, and then running one job per batch. This new system should be easy to extend to [tag counts](https://github.com/otwcode/otwarchive/commit/1556ede32fc7f2e1e97394e8d51ab0e677b8454f) and [readings](https://github.com/otwcode/otwarchive/commit/8691fea11e5dc1f1f5409693f16c213a1414fa14).
3. Create a `StatCounterJob` class extending the `RedisSetJob` class that calls `update_stat_counter` for all works with IDs in the `works_to_update_stats` set.

### DIFF
--- a/app/jobs/redis_set_job_spawner.rb
+++ b/app/jobs/redis_set_job_spawner.rb
@@ -1,0 +1,34 @@
+# An ActiveJob designed to spawn a bunch of subclasses of RedisSetJob, using
+# the contents of a Redis set to generate the jobs.
+class RedisSetJobSpawner < ApplicationJob
+  include RedisScanning
+
+  def perform(job_name, *args, key: nil, redis: nil, job_size: nil, **kwargs)
+    job_class = job_name.constantize
+
+    # Read settings from the job class:
+    redis ||= job_class.redis
+    job_size ||= job_class.job_size
+    key ||= job_class.base_key
+
+    # Bail out early if there's nothing to process:
+    return if redis.scard(key).zero?
+
+    # Rename the job to a unique name to avoid conflicts when this is called
+    # multiple times in a short period:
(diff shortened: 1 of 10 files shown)


app/jobs/redis_set_job_spawner.rb
```diff
@@ -0,0 +1,34 @@
+# An ActiveJob designed to spawn a bunch of subclasses of RedisSetJob, using
+# the contents of a Redis set to generate the jobs.
+class RedisSetJobSpawner < ApplicationJob
+  include RedisScanning
+
+  def perform(job_name, *args, key: nil, redis: nil, job_size: nil, **kwargs)
```
redsummernight, line 6: Let's disable the Metrics/ParameterLists cop (or maybe bump the max to 10).

redsummernight (approved):
Looks good, thanks!

app/jobs/redis_set_job_spawner.rb
tickinginstant: Would you mind if I set `CountKeywordArgs: false` instead? I think there's some value in making sure there aren't too many arguments, but I do think keyword arguments are better because you don't get as confused about "does this value belong in position 4, or position 5?"

redsummernight: Not at all, that's actually better.

PR #4434, opened 2023-01-15
AO3-6458 Use new batch code for tag count updates to try to address LockWaitTimeout errors.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6458

## Purpose

Use the new `RedisSetJob` and `RedisSetJobSpawner`, introduced in #4433, to handle taggings count updates. Retries the job on deadlocks and on lock wait timeouts, so that hopefully the jobs will run more consistently.

### DIFF
--- a/app/jobs/tag_count_update_job.rb
+++ b/app/jobs/tag_count_update_job.rb
@@ -20,7 +20,7 @@ class TagCountUpdateJob < RedisSetJob
     Tag.transaction do
       tag_ids.each do |id|
         value = REDIS_GENERAL.get("tag_update_#{id}_value")
-        Tag.where(id: id).update_all(taggings_count_cache: value) unless value.blank?
+        Tag.where(id: id).update_all(taggings_count_cache: value) if value.present?
       end
     end
   end
--- a/app/jobs/tag_count_update_job.rb
+++ b/app/jobs/tag_count_update_job.rb
@@ -1,0 +1,27 @@
+class TagCountUpdateJob < RedisSetJob
+  queue_as :utilities
+
+  retry_on ActiveRecord::Deadlocked, attempts: 10
+  retry_on ActiveRecord::LockWaitTimeout, attempts: 10
+
+  def self.base_key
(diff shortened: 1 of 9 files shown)


app/jobs/tag_count_update_job.rb
```diff
@@ -0,0 +1,27 @@
+class TagCountUpdateJob < RedisSetJob
+  queue_as :utilities
```
tickinginstant, line 2: Do you want to queue this as utilities? I made the stat counter PR use a separate queue, but this job happens every 2 minutes anyway, so it's not like it's going to swamp everything every half-an-hour. (It might swamp a little if the job gets backed up, though.)

sarken: I'm inclined to move it to a different queue, but let's see how @zz9pzza and @redsummernight feel.

redsummernight: Yeah I think different queues that can be throttled differently are good.

zz9pzza: Perfect world would be config configurable :) but yes different queues

tickinginstant: Unfortunately, one of the jobs involved has its queue set in `config/resque_schedule.yml`, so we can't set that queue with a config variable.

features/step_definitions/work_steps.rb
```diff
@@ -572,25 +572,25 @@
...
   step %{all indexing jobs have been run}
 
-  Tag.write_redis_to_database
+  RedisSetJobSpawner.perform_now("TagCountUpdateJob")
 end
 When /^I post the work without preview$/ do
   click_button "Post"
   step %{all indexing jobs have been run}
 
-  Tag.write_redis_to_database
```
redsummernight, line 593: Can you change these steps to use "the periodic tag count task is run"?

redsummernight (approved):
Looks good, thank you! Just a minor comment.

PR #4435, opened 2023-01-15
AO3-6459 Use new batch code for readings to try to address LockWaitTimeout errors.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6459

## Purpose

Modify the readings code to use `RedisSetJob` and `RedisSetJobSpawner`, introduced in #4433, to try to get the reading job to recover better when it encounters errors.

### DIFF
--- a/app/jobs/readings_job.rb
+++ b/app/jobs/readings_job.rb
@@ -1,24 +1,22 @@
-class ReadingsToDatabaseJob < ApplicationJob
-  include RedisScanning
+class ReadingsJob < RedisSetJob
+  queue_as :readings
 
   retry_on ActiveRecord::Deadlocked, attempts: 10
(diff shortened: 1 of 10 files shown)


app/jobs/readings_job.rb
```diff
@@ -1,24 +1,22 @@
-class ReadingsToDatabaseJob < ApplicationJob
-  include RedisScanning
+class ReadingsJob < RedisSetJob
```
tickinginstant, line 1: I renamed this so that we wouldn't accidentally end up running jobs enqueued by the old code with the new code, since they're not compatible.

redsummernight (approved):
ReadingsJob is very succinct now, nice!

PR #4439, opened 2023-01-26
AO3-6469 Upgrade to Rails 6.1.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6469

## Purpose

- Upgrade Rails to version 6.1.7.2.
- Remove `case_sensitive: false` settings on uniqueness validations, since that's the default now.
- Add the names of all autocomplete actions to the routes file, instead of using syntax which was deprecated in 6.0 and removed in 6.1.
- Add the default `config/storage.yml` file to keep `ActiveStorage` from complaining.
- Monkeypatch `phraseapp-in-context-editor-ruby` to fix a bug that occurs when it's used with Ruby 3.0 + Rails 6.1. The original of this method is defined here: https://github.com/phrase/phraseapp-in-context-editor-ruby/blob/6bf481f672d86c9ad221568f8ec74eea50a13d1a/lib/phraseapp-in-context-editor-ruby/backend_service.rb#L15-L41
  ```rb
      def translate(*args)
        if to_be_translated_without_phraseapp?(args)
          # *Ruby 2.7+ keyword arguments warning*
          #
          # This method uses keyword arguments.
          # There is a breaking change in ruby that produces warning with ruby 2.7 and won't work as expected with ruby 3.0
          # The "hash" parameter must be passed as keyword argument.
          #
          # Good:
          #  I18n.t(:salutation, :gender => 'w', :name => 'Smith')
          #  I18n.t(:salutation, **{ :gender => 'w', :name => 'Smith' })
          #  I18n.t(:salutation, **any_hash)
          #
          # Bad:
          #  I18n.t(:salutation, { :gender => 'w', :name => 'Smith' })
          #  I18n.t(:salutation, any_hash)
          #
          kw_args = args[1]
          if kw_args.present?
            I18n.translate_without_phraseapp(args[0], **kw_args)
          else
            I18n.translate_without_phraseapp(args[0])
          end
        else
          phraseapp_delegate_for(args)
        end
      end
  ```
  The method assumes that the first argument in `args` is always the key to be translated, but calling `translate` with no key and only keyword arguments is valid. Rails 6.1 actually does this by default in views when it can't find a translation for the first key, and the first key is not a symbol: https://github.com/rails/rails/blob/5c835bd669ae996a30bcc914672c9941539a496e/actionview/lib/action_view/helpers/translation_helper.rb#L96
- Fix `errors[:base] << message` syntax, which is newly deprecated in 6.1.

## Testing Instructions

See Jira.


Gemfile.lock
```diff
@@ -274,7 +278,7 @@ GEM
       et-orbi (~> 1.1, >= 1.1.8)
       raabro (~> 1.4)
     gherkin (5.1.0)
-    globalid (1.0.0)
+    globalid (1.0.1)
```
redsummernight, line 281: This can supersede #4437.

redsummernight:
> Remove `case_sensitive: false` settings on uniqueness validations, since that's the default now.

Going to Rails 6.1 technically changes the case sensitivity of

https://github.com/otwcode/otwarchive/blob/d13d1c217534ee5714e4d413534b61298d561039/app/models/admin_blacklisted_email.rb#L4

but it shouldn't matter because we were down-casing all values for case insensitivity anyway.

Gemfile.lock
tickinginstant: I was actually assuming #4437 would be merged as part of the current release -- I figured I would resolve merge conflicts once it was in. But it's up to you.

config/routes.rb
```diff
@@ -654,7 +654,31 @@
 
   # TODO: rewrite the autocomplete controller to deal with the fact that
   # there are fifty different actions going on in there
```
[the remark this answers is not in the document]
tickinginstant: I left it in because 21 is still a lot of actions, and we could consider refactoring it. But I can delete it if that's your preference.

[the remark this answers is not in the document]
redsummernight: Let's keep the comment then, with the count updated.

Gemfile.lock
redsummernight: Yeah, I've merged that now.

redsummernight (approved):
Looks good to me, thank you!

PR #4319, opened 2022-08-17
AO3-5777 Add thousands separators to long numbers in work blurbs

Bilka2:
# Pull Request Checklist

* [X] Have you read ["How to write the perfect pull request"](https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/)?
* [X] Have you read the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [ ] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)?
* [X] Have you added the [Jira](https://otwarchive.atlassian.net) issue number
  as the *first* thing in your pull request title (e.g. `AO3-1234 Fix thing`)
* [X] Do you have fewer than 5 pull requests already open? If not, please wait
  until they are reviewed and merged before creating new pull requests.

## Issue

https://otwarchive.atlassian.net/browse/AO3-5777

## Purpose

Add thousands separators to long numbers in work blurbs

## Testing Instructions

Testing this manually was somewhat iffy because the easy stat to get to over 1000 is word count, but that already used `number_with_delimiter`. I opted to just modify the display code to add +1000 to the numbers before display, things worked as expected.

One thing to note for testing is that for works with multiple chapters but `?` planned total chapters, nothing should error because it is trying to delimit the `?`.

## References

I used the changes in #2514 as an indication that I should increment the cache keys.

## Credit

Bilka (he/him)

PR #4440, opened 2023-01-27
AO3-5778 AO3-5782 Add thousands separators to long numbers in work meta and downloads

Bilka2:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5778
https://otwarchive.atlassian.net/browse/AO3-5782

## Purpose

Add delimiters to long numbers in work meta.
Add delimiters to long numbers in downloads.

## Testing Instructions

No automated tests as per previous delimiter PRs.

One thing to note for testing is that for works with multiple chapters but `?` planned total chapters, nothing should error because it is trying to delimit the `?`.

## Notes

This commit adds the delimiters to all `Chapter: n/n` usages, namely also in:
* The RSS feed of tags
* Sharing works/bookmarks, the html and tumblr options
* Gift notification email
* Subscription notification email

## Credit

Bilka (he/him)


Bilka2:
I just grabbed [AO3-5782](https://otwarchive.atlassian.net/browse/AO3-5782) (delimiters for word count in downloads) and saw the note about the downloads cache needing to be cleared which delays merging. Well, this PR changes downloads, so it would be affected by that too. Possible solutions:
a) Merge the above issue into this PR so all the download changes are done at once. Disadvantage: Delays this PR until there is a downloads release.  
b) Restore the old chapter_total_display method on the work model with a comment that it is deprecated, make downloads use that for now. Then, change over to the new function + delete the old one together with the other download delimiters in AO3-5782. Disadvantages: Keeping old code around. This issue blocks the downloads issue.  

I originally picked this so that it can be merged at the same time as the addition of delimiters in the work blurb, so I'd prefer option b).

The other places where chapter count is displayed do not look like they'd affect the deploy. However, RSS feeds of tags are missing another delimiter on word count. I'd propose to add the delimiter to the word count with this PR, then it's all done. Disadvantage: Lots of things to test at once.

redsummernight:
Adding [AO3-5782](https://otwarchive.atlassian.net/browse/AO3-5782) to this pull request sounds good to me, then we don't have to remember the merge order or keep the old `chapter_total_display` code around, and both issues can be in the same release requiring a download cache clear.

Bilka2:
Okay, added to this PR.

sarken (commented):
This looks good to me! We've put other download changes in this release... do you have a few minutes to resolve the merge conflict so we can add this as well?

Bilka2:
Merge conflict solved, should be good to go!

app/views/works/_work_module.html.erb
```diff
@@ -99,12 +99,12 @@
 
   <% if work.count_visible_comments > 0 %>
     <dt class="comments"><%= ts('Comments') %>:</dt>
-    <dd class="comments"><%= link_to work.count_visible_comments, work.chapter_total_display.to_i > 1 ? work_path(work, anchor: "comments", show_comments: true, view_full_work: "true") : work_path(work, anchor: "comments", show_comments: true) %></dd>
+    <dd class="comments"><%= link_to work.count_visible_comments, work.number_of_posted_chapters > 1 ? work_path(work, anchor: "comments", show_comments: true, view_full_work: "true") : work_path(work, anchor: "comments", show_comments: true) %></dd>
```
sarken, line 102: Oh, you know what, I think there's a cache key on line 94 that needs to be bumped from v2 to v3. Sorry for not noticing that before!

sarken (changes requested):

app/views/works/_work_module.html.erb
Bilka2: The resulting HTML should not be affected by this change, it's just about how it gets there. Does the cache key still have to be updated then?

For clarity, the blurb is not affected be the delimiters in the chapter counts because it goes through chapter_total_display_with_link which is not changed here but in #4319 (and that would also be a different cache key, line 27).

Bilka2: I just realized that the chapter display in work blurbs _is_ affected, in a specific case: One posted chapter, some amount of planned chapters. E.g. 1/2000 chapters should now be displayed as to 1/2,000 chapters. But something like 300/5000 will stay the same. Sorry about the confusion!

So I updated both keys, the one in line 94 for this change and the one in line 27 for the chapter count display change.

sarken: All good, and thanks!

sarken (approved):

PR #4441, opened 2023-01-27
AO3-5356 Add lang attribute to language names

korrien:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5356

## Purpose

Try to make sure most places where we have language different than English are marked with lang attribute. Browser translation tools seemed to work same with or without the attribute (Chrome for some reason struggled in both cases, Edge was better). Experience with screen reader has improved though. I tested with screen reader just one of the changes, but I believe it should be the same for everything else too. 

Updated: almost everything I could find. Divided into many commits for easier overview during review.

Not Updated:
1. Download language under `app/views/downloads/_download_preface.html.erb`
I recall that everything that touches download should be merged separately, so I assume it requires its own issue?
2. Feed summary under `app/helpers/works_helper.rb`
I know nothing about RSS/atom feed readers and wasn't able to set up an environment to test the behavior. I can update the html that we send, but I am not sure if that would be enough. I found some references that we need to use `xml:lang` instead, but as I don't know how that is supposed to work, I just left it as is.

Remarks:
1. I am not familiar with html code safety, but I assume all places where I added `html_safe` are fine as we validate language on input. (If there was a way to avoid that, I don't know. My Ruby skills are still pretty much nonexistent).
    - Though it is possible to create weird languages with for example abbreviation `<p>en`. Something gets saved and language show/edit page doesn't load for me. Hope it's still ok...
2. Wasn't able to find how to properly reach "translation into" code in `app/views/works/_work_approved_children.html.erb`, so had to test rendering by disabling `child_work.translation`. Also couldn't properly test FAQ filter for users, but that was probably due to my broken setup.

## Credit

korrien, she/her

### DIFF
--- a/app/views/abuse_reports/new.html.erb
+++ b/app/views/abuse_reports/new.html.erb
@@ -21,7 +21,8 @@
     <li><%= t(".reportable.dmca.claim", dmca_link:
             (link_to t(".reportable.dmca.policy"), dmca_path)).html_safe %></li>
   </ul>
-  <p><%= t(".languages_html", list: @abuse_languages.map(&:name).to_sentence).html_safe %></p>
+  <p><%= t(".languages_html", list:
+         @abuse_languages.map { |language| tag.span(language.name, lang: language.short) }.to_sentence.html_safe).html_safe %></p>
 </div>
 <%= form_for @abuse_report, class: "post" do |f| %>
   <fieldset>
@@ -40,7 +41,8 @@
         <%= f.label :language, t(".form.language.label") %>
       </dt>
       <dd class="required">
-        <%= f.collection_select(:language, @abuse_languages, :name, :name, { selected: @abuse_report.language || Language.default.name }) %>
+        <%= f.select(:language, @abuse_languages.map { |language| [language.name, language.name, { lang: language.short }] },
+              { selected: @abuse_report.language || Language.default.name }) %>
       </dd>
       <dt class="required">
         <%= f.label :summary, t(".form.summary.label") %>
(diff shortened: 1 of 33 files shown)


app/views/abuse_reports/new.html.erb
```diff
@@ -40,7 +41,8 @@
         <%= f.label :language, t(".form.language.label") %>
       </dt>
```
sarken, line 42: It looks like there's a lot of selects where we take an array of languages or locales and turn them into options like:
```
@abuse_languages.map { |language| [language.name, language.name, { lang: language.short }] }
@news_languages.map { |language| [language.name, language.id, { lang: language.short }] }
@news_languages.map { |language| [language.name, language.short, { lang: language.short }] }
@languages.collect { |language| [language.name, language.id, { lang: language.short }] }

Locale.default_order.map { |locale| [locale.name, locale.iso, { lang: locale.language.short }] }
@available_locales.collect { |locale| [locale.name, locale.id, { lang: locale.language.short }] }
```

Is there any way we could use a helper or two to make that a little less repetitive, e.g., a method that takes an array of languages (or locales) and lets you specify whether to use `id`, `short`, or `name` as the option's value?

sarken (commented):
Thank you so much! This looks pretty good to me -- just two places where I think we have some caching issues. RSS feeds and downloads can definitely be left for later.

korrien:
Oh, thanks! Wasn't aware that's a thing one should pay attention to.
FYI if more fixes are needed I won't be available for a few weeks.

sarken (approved):

sarken:
Hi, korrien! If you have a few minutes to resolve the merge conflict on this, we'd love to add it to this release. (Or if you don't have time, we can handle the conflict -- just let us know!)

PR #4443, opened 2023-01-28
AO3-6464 Allow ruby, rt, and rp HTML tags where HTML is allowed

potpotkettle:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6464

## Purpose

The `ruby`, `rt`, and `rp` tags are used to represent [ruby annotation](https://en.wikipedia.org/wiki/Ruby_character). With these tags, the small text marked with `rt` will be located above the base text marked with `ruby`. Although there are various complexities that come with more advanced features of ruby annotation (e.g. the tabular syntax, CSS rules to control how to render ruby characters, etc), the intention behind the linked Jira issue seems to be to support the most simple (and common) use cases.

## Testing Instructions

To start with, please try posting a new work with the new tags (and with the usual required fields).

Suggested content: `<ruby>BigText<rp> (</rp><rt>short_text</rt><rp>)</rp></ruby>`

This should work in other places where the limited HTML is allowed (like comments).

## References

`rb` is not included here (thus removed by the sanitizer when entered). I think this could potentially be revisited if and when https://github.com/whatwg/html/issues/1771 is resolved. See also the linked Jira issue for related discussion.

## Credit

Potpotkettle (they/them)


potpotkettle:
Here is a current screenshot.
Screenshot 2023-01-28 at 12-39-22 (development) Work with annotations - pot - Original work Example Archive


I have also attached [downloaded versions in PDF, AZW3 and MOBI](https://github.com/otwcode/otwarchive/files/10525243/files.zip). When viewed from an app called Foliate, the mobi version was rendered with parentheses. The rest were the same as the web site version (annotations above the main text).

public/help/html-help.html
```diff
@@ -4,7 +4,7 @@ <h3>Allowed HTML</h3>
 <p>
   <code>a, abbr, acronym, address, [align], [alt], [axis], b, big, blockquote, br, caption, center, cite, [class], code, 
     col, colgroup, dd, del, dfn, div, dl, dt, em, figcaption, figure, h1, h2, h3, h4, h5, h6, [height], hr, [href], i, img, 
-    ins, kbd, li, [name], ol, p, pre, q, s, samp, small, span, [src], strike, strong, sub, sup, table, tbody, td, 
+    ins, kbd, li, [name], ol, p, pre, q, rp, rt, ruby, s, samp, small, span, [src], strike, strong, sub, sup, table, tbody, td,
```
redsummernight, line 7: There's another list at

https://github.com/otwcode/otwarchive/blob/2d92ec1e100d2db8d98c791258b8ea115044975d/app/helpers/application_helper.rb#L67-L74

but I don't think we ever set `show_list = true` anywhere. Can you remove that option and the conditional that uses it?

spec/lib/word_counter_spec.rb
```diff
@@ -85,4 +85,9 @@
     word_counter.text = "\“嘿Bob,\” Alice说，‘啊？！？’"
     expect(word_counter.count).to eq(5)
   end
+
+  it "doesn't count parentheses in rp" do
```
redsummernight, line 89: You can leave out this test, which is not too related to the current issue (the word counter ignores all HTML tags).

public/help/html-help.html
potpotkettle: For all I could find, it was disabled in 1d9f656ba594d9f2940f26fac8fd99d67409053d (in 2011).

The option is used in app/views/bookmarks/_bookmark_form.html.erb and app/views/comments/_comment_form.html.erb (with show_list=false). I feel like I would be touching parts of the code that's not particularly related to the issue (ruby annotations), but if that's okay, sure, I'll do that.

spec/lib/word_counter_spec.rb
potpotkettle: It's true that this is not a test that justifies the other changes in the PR. It would have passed before. So I agree with leaving it out if that's what you mean.

There doesn't seem to be tests that involve parenthesized text, full-width or not, and parentheses might be common things that would warrant tests, but it seems like that's a separate issue to think about.

(That's a long way to say "yes" but I wanted to make sure I have it right.)

redsummernight: Yeah, let's leave this alone for now as it's not related to the main changes in the PR.

public/help/html-help.html
redsummernight: I think it's fine to make these changes. It's unused code that looks puzzling if not updated to include to new tags.

redsummernight (approved):
Thank you!

sarken:
Hey, @potpotkettle! Do you have a few minutes to resolve the merge conflicts on this? If not, no worries, someone else can take them.

potpotkettle:
Sure, done.

redsummernight:
Thanks!

PR #4446, opened 2023-01-31
AO3-6365 Text tweaks for muting and blocking

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6365

## Purpose

Add a missing Oxford comma and change "official account" to "official user" because we realized we generally prefer "user" to "account."

## Testing Instructions

- Try to both block and mute an official user and confirm the message says "user" and not "account"
- Mute someone, and then unmute them and make sure the confirmation page says "see their works, series, bookmarks, and comments on the site"
- Check the FAQ link on your muted users page and the mute confirmation page and ensure it is relative

### DIFF
--- a/app/views/muted/users/confirm_mute.html.erb
+++ b/app/views/muted/users/confirm_mute.html.erb
@@ -23,7 +23,7 @@
     </p>
 
     <p>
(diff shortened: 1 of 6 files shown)


sarken:
Hang on, I'll fix the link, too.

redsummernight (approved):

PR #4448, opened 2023-02-04
AO3-6466 Allow <details> and <summary> HTML tags

irrationalpie7:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6466

## Purpose

Allow the [`<details>` element](https://html.spec.whatwg.org/multipage/interactive-elements.html#the-details-element) and [`<summary>` element](https://html.spec.whatwg.org/multipage/interactive-elements.html#the-summary-element).

Interpretation:
1. Make sure both `<details>` and `<summary>` elements are accepted
2. Make sure neither gets wrapped in a `<p>`
3. Don't break the contents of `<summary>` up into paragraphs 
4. *Do* break the contents of `<details>` other than `<summary>` up into paragraph(s), as needed
6. Don't bother requiring a `<summary>` to be the child of a `<details>`

## Testing Instructions

See Jira issue.

## References

I modeled this heavily on "Allow figure and figcaption HTML tags" ([Jira](https://otwarchive.atlassian.net/browse/AO3-5934)/[Pull request](https://github.com/otwcode/otwarchive/pull/4358/files)), and also referenced the mdn pages on [`<details>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) and [`<summary>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/summary)

## Credit

irrationalpie, they/them

## Comments

There's a note on Jira about potentially making sure margins for `details` match `.userstuff p`, but I'm not sure how to verify whether that needs to happen/where that would go. (edited to add: found a place to put it, but not totally sure it's the correct place)

### DIFF
--- a/config/initializers/gem-plugin_config/sanitizer_config.rb
+++ b/config/initializers/gem-plugin_config/sanitizer_config.rb
@@ -6,9 +6,9 @@ class Sanitize
     ARCHIVE = freeze_config(
       elements: %w[
         a abbr acronym address b big blockquote br caption center cite code col
-        colgroup figcaption figure dd del dfn div dl dt em h1 h2 h3 h4 h5 h6 hr
+        colgroup details figcaption figure dd del dfn div dl dt em h1 h2 h3 h4 h5 h6 hr
         i img ins kbd li ol p pre q s samp small span strike strong
-        sub sup table tbody td tfoot th thead tr tt u ul var
+        sub summary sup table tbody td tfoot th thead tr tt u ul var
       ],
       attributes: {
         all: ['align', 'title', 'dir'],
--- a/lib/paragraph_maker.rb
+++ b/lib/paragraph_maker.rb
@@ -19,7 +19,7 @@ module ParagraphMaker
   # Tags whose content we don't touch
   TAG_NAMES_TO_SKIP = (%w[
     a abbr acronym address audio dl embed figure h1 h2 h3 h4 h5 h6 hr img ol
-    object p pre source table track video ul
+    object p pre source summary table track video ul
   ] + INLINE_INVALID_TAGS + REMOVED_INVALID_TAGS).freeze
 
   # Tags that need to go inside p tags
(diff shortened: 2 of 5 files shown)


potpotkettle:
I tested it manually and all looked good. I was curious about what happens to the details element in the downloaded files. In my environment (Ubuntu Linux, Foliate and Firefox):

* PDF - details collapsed (cannot be uncollapsed)
* MOBI - details uncollapsed (cannot be collapsed)
* EPUB - details collapsed (*can* be uncollapsed)

I don't meant to say something is broken here. (The Jira issue doesn't even mention downloads.) The PDF vs MOBI difference was a bit surprising, though, and it might be a good idea to cover that somewhere in the documentation, anticipating users' questions.

irrationalpie7:
Huh, I hadn't considered downloads! Good idea to check. Where do you think it would make most sense to document this? I looked over the [archive FAQ](https://archiveofourown.org/faq?language_id=en) just now, but it wasn't clear to me. Some candidate locations:
- [Downloading Fanworks](https://archiveofourown.org/faq/downloading-fanworks?language_id=en)
    - (related sibling) [Will images/audio/video load in downloaded works?](https://archiveofourown.org/faq/downloading-fanworks?language_id=en#downloadembeds)
    - How will collapsible details elements appear in downloaded works? (explain epub, pdf, mobi differences)
- 
    -

sarken:
> PDF - details collapsed (cannot be uncollapsed)
> MOBI - details uncollapsed (cannot be collapsed)
> EPUB - details collapsed (can be uncollapsed)

Thanks for the summary! That all makes sense to me: our PDFs are effectively images of the HTML file, so it makes sense you can't change the element from its initial state; MOBI is an old format that almost certainly predates the existence of the tag (in addition to allowing limited formatting) and doesn't even know how to collapse things in the first place; EPUB is pretty interactive and can handle most HTML tags.

The [Known Issues page](https://archiveofourown.org/known_issues) is probably where this will get documented, but Support and Docs will make the final decision and do the actual documenting.

sarken (approved):

PR #4449, opened 2023-02-04
AO3-6475 Let QueryResults use scopes, so that we can start work on reducing N+1 queries for works & bookmarks.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6475

## Purpose

Add a `scope()` function to the `QueryResult` class that lets you add scopes that you want to use when retrieving records from the database. This is passed as an optional `scopes:` argument to the `load_from_elasticsearch` function, and applied to the `ActiveRecord` relation before loading.

Also changes the `WorksController` and the `BookmarksController` to use the `for_blurb` scope for works, bookmarks, series, and external works.

Works are heavily cached, so by default we only call `includes(:pseuds)` for the work's `for_blurb` scope (this is necessary for `set_own_works`, which calls `work.pseuds.pluck(:id)` for each work in the listing when the viewer is logged-in, even when everything else is fully cached).

The other three types aren't as heavily cached, so I added a couple of associations to start with, but this is far from done. (Series, in particular, cause a lot of queries that can't currently be handled by preloading more things.)

### DIFF
--- a/app/controllers/bookmarks_controller.rb
+++ b/app/controllers/bookmarks_controller.rb
@@ -53,7 +53,7 @@ class BookmarksController < ApplicationController
       if @search.query.present?
         @page_subtitle = ts("Bookmarks Matching '%{query}'", query: @search.query)
       end
-      @bookmarks = @search.search_results
+      @bookmarks = @search.search_results.scope(:for_blurb)
       flash_search_warnings(@bookmarks)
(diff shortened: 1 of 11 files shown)


redsummernight (approved):
Looks good to me, thanks! We'll see how it does on production.

PR #4450, opened 2023-02-05
AO3-6431 Make sure work in anonymous collection generates subscription notifications to someone also subscribed to the creator

irrationalpie7:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6431

## Purpose

Users can subscribe to a work in an anonymous collection and should receive notifications when it updates.  However, if any of those users are also subscribed to the creator of the anonymous work, they won’t get a notification when that work updates.

(Note: initially the test I added passed without changing any code so this section only described the test, but after some discussion and investigation I figured out what was up with that)

## Testing Instructions

See Jira.

## References

## Credit

irrationalpie, they/them


features/other_b/subscriptions.feature
```diff
@@ -193,6 +193,22 @@
     And the email should contain "Anonymous"
     And the email should not contain "creator"
 
+  Scenario: When a chapter is added to an anonymous work, and a user is subscribed
```
sarken, line 196: Hm. Is something in a previous test causing this to pass when it shouldn't? I just manually tested the issue a few times on staging and definitely didn't get an email, so the bug is still with us. 

I also tried running this test locally and it seems like it fails pretty consistently when I run it alone (`RAILS_ENV=test bundle exec cucumber features/other_b/subscriptions.feature:196`) and passes pretty consistently when I run all the tests in this file. The step it fails on is `Then "subscriber" should be emailed`, which is what I'd expect. The only inconsistencies seem to happen when I switch between the two approaches (i.e., the first time I switch from running just this scenario to running the whole file, the scenario passes).

irrationalpie7: Huh, I can't seem to get it to fail locally, even when running alone, either before or after changing author/subscriber/work/etc names. Do you know of any commands I might need to run to get my local development environment in a working state? I noticed in the [testing docs](https://github.com/otwcode/otwarchive/wiki/Automated-Testing#prepare-the-database) a line about preparing the database, but after running `bundle exec rake db:test:prepare` it just aborted and made me run `rails db:migrate RAILS_ENV=test` before going right back to passing.

irrationalpie7: I added a duplicate scenario on line 213 with modified names, but given that I still can't get it to fail I'm not positive it's internally consistent.

irrationalpie7: I continue to be baffled by my local tests, but my current hypothesis as to why sending the email fails is that it's something to do with the [`Subscription.for_work(work)`](https://github.com/otwcode/otwarchive/blob/2f0f0d43c73c8200ed65f02558074d2a1c4d791d/app/models/subscription.rb#L19) scope in `app/models/subscription.rb`, which groups subscriptions by user id, presumably so you don't send duplicate emails to a user that is subscribed to a work *and* its series *and* its creator. I'm not totally confident on what `group` does--I thought I found a source that said it made a list out of the items so grouped? However, where `for_work` gets used later in [`lib/creation_notifier.rb`](https://github.com/otwcode/otwarchive/blob/2f0f0d43c73c8200ed65f02558074d2a1c4d791d/lib/creation_notifier.rb#L52), it seems to be assuming that everything it's getting is a subscription, sending it to [`RedisMailQueue.queue_subscription`](https://github.com/otwcode/otwarchive/blob/2f0f0d43c73c8200ed65f02558074d2a1c4d791d/app/models/redis_mail_queue.rb#L49) which gets a single `subscription.id` from it. So clearly I'm missing something here.

That aside, conceptually it makes sense that you'd be trying to batch all the works that a creator has posted into the same notification email, but then it turns out some of those works the creator posted were anonymous. At the point where you're batching the notifications into a single email, you don't want to include the anonymous ones in the creator email, so your only real choice is to discard them, which I think is what's happening on [line 132 of `UserMailer.batch_subscription_notification`](https://github.com/otwcode/otwarchive/blob/2f0f0d43c73c8200ed65f02558074d2a1c4d791d/app/mailers/user_mailer.rb#L132).

So at that point it's kind of too late to be processing an anonymous work. Though maybe you could somehow split it into multiple emails? One for the creator, one per non-creator subscription type (if you even have that information available at that point--again, I'm not clear on where the group (list?) of subscriptions becomes a single subscription). 

That seemed kinda tricky though, so instead I thought, what if at the point you enqueue the subscriptions for a work, you check if it's an anonymous work and only look for work or series subscriptions. Then you shouldn't end up grouping the creator's works with their anonymous works, unless a subscription to a work can somehow be replaced with a subscription to a creator somewhere in this flow that I haven't found. However, if a creator posts a new chapter, then edits the work to make it anonymous, then the subscription email goes out, I think the notification would still get dropped even with this fix. So maybe it's not sufficient.

Sorry this is kinda long and rambly, but the tests continue to pass locally no matter what I do (well, except when I change "should be emailed" to "should not be emailed", but that doesn't exactly help) so I thought maybe writing out my thoughts would at least help me pick this up again when I do have the tests to verify whether what I'm doing works.

Since this draft theoretically could be working because I actually fixed the bug, I'll remove the duplicate test from line 213 so it's just the one at 196 again.

irrationalpie7: I still can't get this test to fail locally when I run it alone (with line number) at either of the first two commits, so the fact that the test passes after the changes I made in the third commit doesn't really give me much confidence that it's actually fixed. Do you have any idea what I might need to change to get the tests to fail?

sarken: Let me see if I can find a third person to run it -- it's possible the result in _my_ environment is wrong, so it would be good to have a tiebreaker.

irrationalpie7: Okay, I think it's the [`group` in the `for_work` scope](https://github.com/otwcode/otwarchive/blob/b63fce67a3975ec0614bb535c75d76c8dafbda71/app/models/subscription.rb#L26) that I wasn't sure how to understand before that's causing the testing inconsistency. The [first `group` example here](https://guides.rubyonrails.org/active_record_querying.html#group) suggests that what's being returned is a single subscription to the work that satisfies that condition per subscribing user, picking one and throwing the rest away. And on my machine that seems to consistently return the user's subscription to the work rather than their subscription to the author. Thus, the test passes. I don't know if there's a way to manipulate the test case to induce it to return the work subscription instead, but I'm going to have to either figure that out or have the final fix work in such a way that the final set of tests isn't impacted by this behavior

redsummernight: > And on my machine that seems to consistently return the user's subscription to the work rather than their subscription to the author.

This is also the case for me (Docker/WSL), the test you added never fails.

When we use `group`, picking the subscribable columns per subscribing user without an aggregate function is considered undefined behavior by MySQL, so we have no guarantee which subscribable we get, which I think leads to inconsistent behavior between different environments (this kind of query is not allowed by default but we're running explicitly without `ONLY_FULL_GROUP_BY` mode, e.g. #3420).

So we do need to rewrite `for_work` to guarantee that for each user + work, we pick the subscribable of type, in descending order of preference: work, series, user.

irrationalpie7: Thanks for that context @redsummernight! In the end I couldn't figure out how to make the query do that all at once, so I went with making the query give us at most one subscription of each type per user, and then wrote some code to group those by user and pick the right subscription from that set.

Querying really seems like it should be possible--we essentially want to apply a custom `min` operation to the `group`, where the `min` of two subscriptions is the one that has a type that comes earlier in the list `Work`, `Series`, `User`, right? But nothing I found by googling suggests that you can make your own custom aggregation (calculation?) function for `group`. You could theoretically have the query sort by id, and then within that sort by type, and then the code just has to pick the first subscription it sees for each user, but that seemed like it would require needless extra work from the query, since we don't need the resulting subscriptions to be in `user_id` order. `coalesce` initially seemed promising, because you can give it a list of conditions and it falls back to the first non-null one, but I think that would only really work if we're querying for all users, and then asking for each user whether they have, successively, any relevant work, series, or author subscriptions. And given that the set of all users is probably much much larger than the set of users to notify, that seems like it would be inefficient.

I considered making the new `for_work` essentially act as an iterator and `yield` subscriptions, instead of using `map` to make a transformed array, but it ended up being slightly more complicated to test so I dropped that idea. I can go back to it though if that seems better! It looked something like this:
```
def self.for_work(work)
    Subscription.for_work_with_duplicates(work).group_by { |u| u.user_id }.values
      .each { |subscriptions| yield Subscription.pick_most_relevant_of(subscriptions) }
  end
```
and then that would have required removing the `.each` in `lib/creation_notifier.rb` to make it:
```
      Subscription.for_work(work) do |subscription|
        RedisMailQueue.queue_subscription(subscription, self)
      end
```

irrationalpie7:
While randomly looking through jira issues, I found a comment on [AO3-3620 Subscriptions to orphan_account don't work and having orphan_account as a co-author will break subscriptions to the other creator(s)](https://otwarchive.atlassian.net/browse/AO3-3620) that said "If a co-authored work is orphaned by one creator, users subscribed to the work or to the other creator(s) won't get subscription notifications when it's updated by the remaining creator(s)."

That seems like it *could* be a special case of the bug I'm trying to fix here. In my fix I assume all user-type subscriptions are equally valid and just have it pick an arbitrary one; if I should be prioritizing a non-orphan user or non-orphaned user, that would require not grouping by `subscribable_type`, and then adding some extra logic when picking the best subscription to use as the basis for notifying a subscriber.

sarken:
I'm not sure if the orphaning issue is relevant. As ticking instant recently pointed out on the issue, part of the problem is tied to a check in the mailer itself that prevents sending notifications if orphan_account is a creator on the subscribable:

https://github.com/otwcode/otwarchive/blob/f7cd4fc80c1482e12d216422759b3eee670b45fb/app/mailers/user_mailer.rb#L127

The check was implemented to fix [AO3-3186](https://otwarchive.atlassian.net/browse/AO3-3186), where if I posted a work and orphaned it before the hour's subscription emails were sent, people subscribed to me would still get a notification.

irrationalpie7:
@sarken I was looking through my old issues and found this. I've merged in master and shuffled some of the tests around, so it should be ready for re-review!

PR #4452, opened 2023-02-07
AO3-6471 Add word count search, filter, and sorting to interface for bookmarks

irrationalpie7:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6471

## Purpose

Add the ability to search, filter, and sort by bookmark word counts.

One point mentioned in the Jira issue is that some but not all works in a series can be restricted, so we need to index two separate word counts, one for logged in users (including restricted works) and one for guests (excluding restricted works). In order to support sorting of both bookmarks and bookmarkables, I ended up going with the following index structure:
- Bookmark index:
    - \+ bookmarkable_word_count
    - \+ bookmarkable_guest_word_count
- Work index:
    - word_count (pre-existing)
    - \+ guest_visible_word_count (duplicate of word_count)
- Series index:
    - word_count (pre-existing)
    - \+ guest_visible_word_count (only includes word counts of unrestricted works)

I'm not sure what effect this index duplication will have at archive-scale and welcome feedback on consolidating this somehow!

It's also possible that the way filtering works in this pull request, if you're looking at a *bookmarkable* search it has to look at all of the bookmarkable's bookmarks to determine whether the bookmarkable should be kept or filtered out based on its word count, which if true seems suboptimal.

## Testing Instructions

See Jira.

I tried to be thorough with the unit testing to reduce the need for manual testing, but ran out of steam for integration tests so just have the one of those at the moment, though I'm happy to add more if pointed in the right direction. Even if we end up needing to take a different approach re: indexing, the tests I've added should be useful for making sure the new approach works.

## References

These pull requests aren't super recent, but I found #3194, which added word count filtering to work search, and #3252, which overhauled bookmark index pages to arrange by bookmarkable, to be useful for figuring out what I needed to change where.

## Credit

irrationalpie, they/them


irrationalpie7:
> It's also possible that the way filtering works in this pull request, if you're looking at a _bookmarkable_ search it has to look at all of the bookmarkable's bookmarks to determine whether the bookmarkable should be kept or filtered out based on its word count, which if true seems suboptimal.

Since we already have word count indexed under both bookmarks and bookmarkables, I was able to move the word count filter to bookmarkable_query which at least conceptually seems like it would be more efficient. And after moving some of them around, the tests still pass! :tada:

irrationalpie7:
> * Bookmark index:
>   
>   * \+ bookmarkable_word_count
>   * \+ bookmarkable_guest_word_count
> * Work index:
>   
>   * word_count (pre-existing)
>   * \+ guest_visible_word_count (duplicate of word_count)
> * Series index:
>   
>   * word_count (pre-existing)
>   * \+ guest_visible_word_count (only includes word counts of unrestricted works)
> 
> I'm not sure what effect this index duplication will have at archive-scale and welcome feedback on consolidating this somehow!

To clarify on how these indexes are used, `bookmarkable_word_count` and `bookmarkable_guest_word_count` are used for sorting a bookmark search (I think an example of this is a user's bookmarks page). `word_count` and `guest_visible_word_count` are used for sorting a bookmarkable search (I think an example of this is a tag's bookmarks page), or for filtering either kind of search. This difference between bookmark(able) sorting and filtering is similar to how a bookmark's `bookmarkable_date` and bookmarkable's `revised_at` indexes are used.

Bilka2:
We've decided and coded the design of [AO3-6304](https://otwarchive.atlassian.net/browse/AO3-6304) which was blocking this issue. We haven't merged the associated PR yet because it needs a reindex. Ideally, we would like combine it with the reindex for this PR. So it would be nice if you could already go ahead and adjust this pull request to match the design in #5179, especially in regards to the mappings and their fields.

[AO3-6304]: https://otwarchive.atlassian.net/browse/AO3-6304?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

irrationalpie7:
> We've decided and coded the design of [AO3-6304](https://otwarchive.atlassian.net/browse/AO3-6304) which was blocking this issue. We haven't merged the associated PR yet because it needs a reindex. Ideally, we would like combine it with the reindex for this PR. So it would be nice if you could already go ahead and adjust this pull request to match the design in #5179, especially in regards to the mappings and their fields.

@Bilka2 ooh I'll take a look 👀

[AO3-6304]: https://otwarchive.atlassian.net/browse/AO3-6304?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

app/models/search/bookmarkable_query.rb
```diff
@@ -189,8 +190,13 @@ def bookmark_filter
...
     # If we're sorting by created_at, we actually need to fetch the bookmarks'
     # created_at as the score of this query, so that we can sort by score (and
-    # therefore by the bookmarks' created_at).
+    # therefore by the bookmarks' created_at). Similarly for word_count.
     bool = field_value_score("created_at", bool) if sort_column == "created_at"
+    if sort_column == "word_count"
+      bool = include_restricted? ?
+        field_value_score("general_word_count", bool) :
+        field_value_score("public_word_count", bool)
+    end
```
irrationalpie7: I'm really not sure what rubocop wants me to do here. Any ideas? I originally had the ternary as an inner if/else, but it also complained about that (I think the message before was "Use the return of the conditional for variable assignment and comparison")

Bilka2:
```
bool = if include_restricted? 
        field_value_score("general_word_count", bool) 
     else
        field_value_score("public_word_count", bool)
   end
```
with correct indentation (instead of my mess) should make it happy. (Or maybe it even wants you to combine it with the sortcolumn condition.) 

The first message sounds like you had something like this:
```
if include_restricted? 
  bool = field_value_score("general_word_count", bool) 
else
  bool = field_value_score("public_word_count", bool)
end
```

irrationalpie7: Ah okay, that syntax is new to me. I'll try that out, thanks! 💙

docker-compose.yml
```diff
@@ -56,7 +56,7 @@ services:
     environment:
       - transport.host=localhost
       - bootstrap.memory_lock=false
-      - "ES_JAVA_OPTS=-Xms1500m -Xmx1500m"
+      - "ES_JAVA_OPTS=-Xms1000m -Xmx1000m"
```
irrationalpie7: Before this change, elastic search was constantly crashing (possibly either because of my docker settings or my computer not having that much ram?). Capping this at a lower amount helped elastic search better manage its memory usage. Here's a [thread I found useful while investigating](https://discuss.elastic.co/t/elasticsearch-process-ended-by-code-137/345399) + the [official elasticsearch docs on setting the heap size](https://www.elastic.co/docs/reference/elasticsearch/jvm-settings#set-jvm-heap-size)

test/fixtures/bookmarks.yml
```diff
@@ -627,3 +627,16 @@ bookmark_00016:
...
+  created_at: 2009-07-18 00:14:46 Z
+  bookmarkable_id: 5
+  delta: true
+  updated_at: 2009-07-18 00:14:46 Z
+  private: false
+  bookmarker_notes: mixed visibility series yay
+  rec: false
+  id: 47
+  user_id: 
+  hidden_by_admin: false
```
irrationalpie7, line 642: It turns out we didn't have any series bookmarks in the local data you can view at localhost:3000, so I added one of a mixed-visibility series (can be seen at http://localhost:3000/users/reccer/pseuds/reccer/bookmarks if you have the local dev environment going)

app/models/search/bookmark_search_form.rb
```diff
@@ -141,8 +145,9 @@ def bookmarkable_search_results
 
   def sort_options
     [
-      ['Date Bookmarked', 'created_at'],
-      ['Date Updated', 'bookmarkable_date'],
+      ["Date Bookmarked", "created_at"],
+      ["Date Updated", "bookmarkable_date"],
+      ["Word Count", "word_count"]
```
irrationalpie7, line 150: FYI these options aren't in sentence case--do we want to change it? Or leave as-is? They're dropdown options, rather than form labels, if that makes a difference

app/models/series.rb
```diff
@@ -256,6 +249,14 @@ def word_count
     self.works.posted.pluck(:word_count).compact.sum
   end
 
+  def public_word_count
+    self.works.posted.unrestricted.pluck(:word_count).compact.sum
+  end
+
+  def general_word_count
+    self.works.posted.pluck(:word_count).compact.sum
+  end
```
irrationalpie7, line 266: Note: double-check that we're computing this value correctly. I'm not sure where we test visible_word_count, so it's possible we already have some tests? But some edge cases off the top of my head would be series containing:
- draft works
- works with draft chapters
- mystery works (unrevealed)
- anonymous works
- archive-locked works
- works hidden by an admin

(I'm also curious if this value is ever different if the creator of the series is the user looking at the value?)

I call this out for this PR specifically, because if we're about to reindex all series it would be good to make sure we have the right numbers computed. My understanding is that it's comparatively much easier to change the word count display values

app/models/work.rb
```diff
@@ -1213,8 +1213,10 @@ def bookmarkable_json
       unrevealed: unrevealed?,
       pseud_ids: anonymous? || unrevealed? ? nil : pseud_ids,
       user_ids: anonymous? || unrevealed? ? nil : user_ids,
-      bookmarkable_type: 'Work',
-      bookmarkable_join: { name: "bookmarkable" }
+      bookmarkable_type: "Work",
+      bookmarkable_join: { name: "bookmarkable" },
+      public_word_count: word_count,
+      general_word_count: word_count
```
irrationalpie7, line 1227: For symmetry with the series tag PRs, I have added `public_word_count` and `general_word_count` fields. However, I wasn't sure if we should remove `word_count`, as I wasn't sure where else it might be in use. Alternatively, we could use `word_count` and `public_word_count` as the field names--let me know!

spec/models/search/bookmark_search_form_spec.rb
```diff
@@ -13,6 +13,45 @@
...
+        let!(:work10r) { create(:work, fandom_string: tag.name, restricted: true, chapter_attributes: { content: "one two three four five six seven eight nine ten" }) }
+        let!(:work10) { create(:work, fandom_string: tag.name, title: "Ten", chapter_attributes: { content: "one two three four five six seven eight nine ten" }) }
+        # work "Ten" has word_count 10
+        let!(:work_bookmark) { create(:bookmark, bookmarkable: work10) }
+
+        let!(:series) { create(:series, title: "Series to be bookmarked") }
+        let!(:serial_work1) { create(:serial_work, series: series, work: work5) }
+        let!(:serial_work2) { create(:serial_work, series: series, work: work10r) }
+        # series "Series to be bookmarked" word_count is 5 or 15
+        let!(:series_bookmark) { create(:bookmark, bookmarkable: series) }
```
irrationalpie7, line 28: Is this the recommended way to set up the series/works I want for a test? I'm not sure whether the reason the associated test fails is due to a bug in the code, a problem with the test, or a problem with the test setup :thinking:

features/bookmarks/bookmark_filter.feature
```diff
@@ -17,3 +17,55 @@ Feature: Filter bookmarks
...
+    Then I should see "1 Bookmark by awesome_posterrr"
+      And I should see "Mixed Visibility"
+    When I am logged out
+      And I go to awesome_posterrr's bookmarks page
+    Then I should see "1 Bookmark by awesome_posterrr"
+      And I should see "Words: 6"
+    When I fill in "To" with "10" within "#work_words"
+      And I press "Sort and Filter"
+    Then I should see "1 Bookmark by awesome_posterrr"
+      And I should see "Mixed Visibility"
```
irrationalpie7, line 46: The last two steps here keep failing, and I can't figure out why. The first part of the test works, and filtering works for localhost :thinking:

spec/models/search/bookmark_search_form_spec.rb
irrationalpie7: Managed to fix the setup! It was indeed the problem (the series was being created with an additional work, so the word count wasn't what I expected)

app/models/series.rb
irrationalpie7: I asked [on slack](https://transformativeworks.slack.com/archives/G2CRYKP8W/p1760858971643929) and learned we should exclude works hidden by admin, but include unrevealed works. I've added tests for `visible_word_count` after this change

docker-compose.yml
irrationalpie7: I ended up having to bump it even lower to prevent crashes for me, but I'm not sure to what extent this might cause a bad experience for people working with more powerful hardware. I'm happy to revert and/or file a separate PR, just let me know!

features/bookmarks/bookmark_filter.feature
irrationalpie7: Is it possible a caching thing???

spec/models/search/bookmark_search_form_spec.rb
```diff
@@ -13,6 +13,81 @@
...
+          # Check titles of returned results
+          expect(results.map(&:title)).to eq ["Ten", "Series to be bookmarked"]
+        end
+
+        it "changes when the work wordcount changes" do
+          # Update the one public work in the series to have 15 words instead of 5
+          work5.chapters.first.update(content: "This is a work with a word count of fifteen which is more than ten.")
+          work5.save
+          series.save
+          run_all_indexing_jobs
```
irrationalpie7, line 57: Note: without the `series.save` line, the test fails--the word count of the returned results is correct, but they're returned in the wrong order. Is this a sign of something that's a problem? Like, if we update a work, do we need to mark its series for reindexing somehow?

features/bookmarks/bookmark_filter.feature
```diff
@@ -17,3 +17,61 @@ Feature: Filter bookmarks
     Then I should see "2 Bookmarks by recengine"
       And I should see "english work"
       And I should see "german work"
+
+  # FIXME: currently failing
+  Scenario: Filtering series bookmarks by word count with restricted works
+    Given a bookmark by "awesome_posterrr" of a mixed visibility series with fandom "What a fandom"
```
irrationalpie7: @irrationalpie7 : fix the tests in this file

irrationalpie7:
@Bilka2 One of the tests still has some todos/fixmes, but they're not actually failing anymore, just in need of a refactor to make it clearer what's happening and remove some extraneous comments. This PR is ready for review! :tada:

docker-compose.yml
```diff
@@ -56,7 +56,7 @@ services:
     environment:
       - transport.host=localhost
       - bootstrap.memory_lock=false
-      - "ES_JAVA_OPTS=-Xms1500m -Xmx1500m"
+      - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
```
zz9pzza: Why are you making this change ?

irrationalpie7: I've been developing locally with docker, and kept getting elasticsearch crashes before I changed this line (I left a pr comment about it). I'm happy to revert this file and/or break it out into a separate pr, I just left the change in for now to make running tests locally easier for me

features/bookmarks/bookmark_filter.feature
irrationalpie7: It was due to a lack of reindexing, I'm pretty sure

spec/models/search/bookmark_search_form_spec.rb
irrationalpie7: Yep, that was a problem, and I have added code to mark series for reindexing

app/models/search/bookmark_search_form.rb
```diff
@@ -157,7 +162,7 @@ def sort_direction(sort_column)
   private
 
   def processed_options(opts = {})
-    [:date, :bookmarkable_date].each do |countable|
+    [:date, :bookmarkable_date, :word_count].each do |countable|
```
brianjaustin: Nitpick:
```suggestion
    %i[date bookmarkable_date word_count].each do |countable|
```

app/models/work.rb
```diff
@@ -315,6 +315,19 @@ def should_update_pseud_and_collection_indexes?
...
 
+  def update_series_indexes
+    return unless should_update_series_indexes?
+
+    series_ids = SerialWork.where(work_id: id).pluck(:series_id)
+    IndexQueue.enqueue_ids(Series, series_ids, :background)
+  end
+
+  # Visibility or word count has changed, so we need to update series word counts
+  def should_update_series_indexes?
```
brianjaustin: Can this helper be private so we don't expose it for unintended use elsewhere?

app/models/work.rb
brianjaustin: Let's remove `word_count` to avoid confusion

public/help/bookmark-search-numerical-help.html
```diff
@@ -0,0 +1,14 @@
+<h4>Bookmark Search: Numerical Values</h4>
```
brianjaustin: Should leave for docs (I didn't see this in the issue, may have missed though)

app/models/search/bookmarkable_query.rb
```diff
@@ -99,16 +100,22 @@ def bookmarkable_query_text
...
-  # results. When sorting by created_at, we use _score to sort (because the
-  # only way to sort by a child's fields is to store the value in the _score
-  # field and sort by score).
   def sort
-    if sort_column == "bookmarkable_date"
-      sort_hash = { revised_at: { order: sort_direction, unmapped_type: "date" } }
-    else
-      sort_hash = { _score: { order: sort_direction } }
-    end
+    sort_hash = if sort_column == "bookmarkable_date"
```
zz9pzza: Could you refactor this in to a case statement please ?

app/models/search/bookmarkable_query.rb
```diff
@@ -99,16 +100,22 @@ def bookmarkable_query_text
...
+                  # bookmarkable_date corresponds to the bookmarkable's revised_at date
+                  { revised_at: { order: sort_direction, unmapped_type: "date" } }
+                elsif sort_column == "created_at"
+                  # When sorting by created_at, we use _score to sort (because the
+                  # only way to sort by a child's fields is to store the value in
+                  # the _score field and sort by score).
+                  { _score: { order: sort_direction } }
+                # We check the word_count cases last, as they are different depending
+                # on whether we include restricted works.
+                elsif include_restricted?
```
zz9pzza: Can we have the sort_column in the conditional ?

app/views/bookmarks/_filters.html.erb
```diff
@@ -85,6 +85,16 @@
       </dt>
       <dd class="more group">
         <dl>
+          <dt id="toggle_work_words" class="filter-toggle words"><%= t("bookmarks.filters.word_count") %></dt>
```
brianjaustin: Is it possible to use lazy i18n keys here, or is it weird because of this being a partial?

(If this is for re-use, our translators generally prefer repetition instead because it fits their process better.)

app/views/bookmarks/_search_form.html.erb
```diff
@@ -27,6 +27,13 @@
                      options_for_select(["", "Work", "Series", "External Work"],
                                         @search.bookmarkable_type) %>
       </dd>
+      <dt>
+        <%= f.label :word_count, t("bookmarks.filters.word_count") %>
```
brianjaustin: Same question here

features/bookmarks/bookmark_filter.feature
zz9pzza: Can we help at all ?

features/bookmarks/bookmark_filter.feature
```diff
@@ -17,3 +17,61 @@ Feature: Filter bookmarks
...
+      And I should see "Mixed Visibility"
+    # Logged out behavior
+    When I am logged out
+      And I go to awesome_posterrr's bookmarks page
+    Then I should see "1 Bookmark by awesome_posterrr"
+      And I should see "Words: 8"
+    When I fill in "To" with "10" within "#work_words"
+      And I press "Sort and Filter"
+    Then I should see "1 Bookmark by awesome_posterrr"
+      And I should see "Mixed Visibility"
```
zz9pzza, line 46: Is it worth adding a test for the 0 bookmarks eg To 4 ?

(non blocking)

features/bookmarks/bookmark_filter.feature
```diff
@@ -17,3 +17,61 @@ Feature: Filter bookmarks
...
+
+  # FIXME: currently failing
```
zz9pzza: Does this still fail ?

features/bookmarks/bookmark_search.feature
```diff
@@ -65,6 +65,31 @@ Feature: Search Bookmarks
...
+      And I should see "Ten"
+    When I follow "Edit Your Search"
+    Then the field labeled "Word count" should contain "<15"
+
+  Scenario: Sort bookmarks by word count
+    Given bookmarks with various word counts in fandom "Totally a fandom" to search
+    When I select "Word Count" from "Sort by"
+      And I fill in "Work tags" with "Totally a fandom"
+      And I press "Search Bookmarks"
+    Then I should see the page title "Search Bookmarks"
```
zz9pzza, line 85: Is it worth having a test of the reverse sort order ?

public/help/bookmark-search-numerical-help.html
zz9pzza: Can we remove this, there is work ongoing with i18n docs.

spec/models/search/bookmark_query_spec.rb
```diff
@@ -129,6 +159,32 @@ def find_parent_filter(query_list)
...
+    it "allows sorting by public word count" do
+      q = BookmarkQuery.new(sort_column: "word_count")
+      # Sets sorting by score
+      expect(q.generated_query[:sort]).to eq([{ _score: { order: "desc" } }, { id: { order: "desc" } }])
+      # Computes score by public word count
+      expect(q.generated_query.dig(:query, :bool, :must, :has_parent, :query, :function_score, :field_value_factor, :field)).to \
+        eq("public_word_count")
+    end
+
+    it "allows sorting by general word count when logged in" do
```
zz9pzza, line 172: I think we should also test sort order here ( with a preference to here rather than the cucumber, ( You don't have to do both ) )

spec/models/search/bookmark_search_form_spec.rb
```diff
@@ -13,6 +13,126 @@
...
+          results = BookmarkSearchForm.new(parent: tag, sort_column: "word_count").bookmarkable_search_results
+          # "Series to be bookmarked": 15, "Ten": 10
+          # Check word count of returned results
+          expect(results.map { |item| item.respond_to?(:general_word_count) ? item.general_word_count : item.word_count }).to eq [15, 10]
+          # Check titles of returned results
+          expect(results.map(&:title)).to eq ["Series to be bookmarked", "Ten"]
+        end
+
+        it "sorts bookmarkables correctly when not logged in" do
+          User.current_user = nil
```
brianjaustin: Is this to fix a failure? If yes, then we have an issue with the inter-test cleanup. Otherwise, I would prefer a `context` block to clarify that this is for logged out users/guests

spec/models/search/bookmark_search_form_spec.rb
```diff
@@ -13,6 +13,126 @@
...
+          # "Series to be bookmarked": 15, "Ten": 10
+          # Check word count of returned results
+          expect(results.map { |item| item.respond_to?(:public_word_count) ? item.public_word_count : item.word_count }).to eq [15, 10]
+          # Check titles of returned results
+          expect(results.map(&:title)).to eq ["Series to be bookmarked", "Ten"]
+        end
+
+        it "reindexes when a work is deleted" do
+          # Destroy one of the works in the series
+          work10r.destroy
```
brianjaustin: Could you do the following to make sure we don't silently fail to destroy it?
```suggestion
          work10r.destroy!
```

spec/models/search/bookmark_search_form_spec.rb
```diff
@@ -13,6 +13,126 @@
 
   describe "bookmarkable_search_results" do
     describe "sorting" do
+      context "by word count" do
+        let(:tag) { create(:canonical_fandom) }
+
+        let!(:work5) { create(:work, fandom_string: tag.name, chapter_attributes: { content: "one two three four five" }) }
+        let!(:work10r) { create(:work, fandom_string: tag.name, restricted: true, chapter_attributes: { content: "one two three four five six seven eight nine ten" }) }
```
zz9pzza: Can we have a unique number of words for the restricted works so it's clear which one we are finding ?

spec/models/search/bookmark_search_form_spec.rb
```diff
@@ -13,6 +13,126 @@
...
+          # "Series to be bookmarked": 5, "Ten": 10
+          # Check word count of returned results
+          expect(results.map { |item| item.respond_to?(:public_word_count) ? item.public_word_count : item.word_count }).to eq [10, 5]
+          # Check titles of returned results
+          expect(results.map(&:title)).to eq ["Ten", "Series to be bookmarked"]
+        end
+
+        it "reindexes when a work is hidden by admin" do
+          # Hide one of the works in the series
+          work10r.update(hidden_by_admin: true)
```
brianjaustin: Similar here
```suggestion
          work10r.update!(hidden_by_admin: true)
```

app/models/series.rb
```diff
@@ -256,6 +251,18 @@ def word_count
     self.works.posted.pluck(:word_count).compact.sum
   end
 
+  # Word count as seen by guests
+  def public_word_count
+    # Exclude restricted works and those hidden by admin from the word count
+    self.works.posted.unrestricted.unhidden.pluck(:word_count).compact.sum
+  end
+
+  # Word count as seen by registered users
```
zz9pzza, line 260: Can we add a comment that if a work is hidden by an admin then normally we would expect a creator to be able to see the work but this will not happen in this case ?

zz9pzza (changes requested):

app/models/series.rb
lydia-theda: When viewing your series page as the creator you do currently see the blurbs of all works in the series, including any hidden works.

features/bookmarks/bookmark_filter.feature
irrationalpie7: I think my comment is outdated (CI seems to be passing) so I'll remove this comment

app/views/bookmarks/_filters.html.erb
irrationalpie7: What are lazy i18n keys? I'm happy to change the approach if that makes more sense!

spec/models/search/bookmark_search_form_spec.rb
irrationalpie7: I think I was just confused at the time I wrote this about how inter-test cleanup works. I can remove the line setting current user to nil!

features/bookmarks/bookmark_filter.feature
irrationalpie7: No, I think I just forgot to clean up this comment

features/bookmarks/bookmark_search.feature
irrationalpie7: Do we want to support reverse sort order? Would that require an extra dropdown option, or can it be accessed some other way?

public/help/bookmark-search-numerical-help.html
irrationalpie7: Sure! Should there just be no helper link for the word count inputs then?

app/models/series.rb
irrationalpie7: I think the creator will still see the hidden work's blurb if they look at the series, but the series blurb won't include the hidden work's word count, even for creators. Does that distinction make sense? Is that different from the desired behavior?

app/models/work.rb
irrationalpie7: I think I have now made it private (added `private` on a separate line above the method) but tbh I found the documentation on private methods for ruby on rails a bit confusing, so please let me know if I didn't quite do it right!

spec/models/search/bookmark_query_spec.rb
irrationalpie7: The existing tests in this file seems to be strictly for the query shape, while bookmark_search_form_spec.rb checks sort order. Would you like any tests moved from one spec file to a different one? Are there spec test cases you feel are missing? Are there any cucumber tests that feel unnecessary/superfluous given the spec tests I have?

app/models/series.rb
irrationalpie7: I've updated the comment

docker-compose.yml
irrationalpie7: Reverted!

app/models/work.rb
irrationalpie7: That appears to have broken everything lol, so I have instead moved the method to the end of the file with the other private method :joy:

app/models/search/bookmarkable_query.rb
```diff
@@ -99,16 +100,25 @@ def bookmarkable_query_text
...
-    end
+    sort_hash = case sort_column
+                when "bookmarkable_date"
+                  # bookmarkable_date corresponds to the bookmarkable's revised_at date
+                  { revised_at: { order: sort_direction, unmapped_type: "date" } }
+                when "created_at", "_score"
+                  # When sorting by created_at, we use _score to sort (because the
+                  # only way to sort by a child's fields is to store the value in
+                  # the _score field and sort by score).
+                  { _score: { order: sort_direction } }
```
irrationalpie7, line 112: I think there was a bug pre-refactor to `case` statement, because I didn't realize the `sort_column` might be "_score" (I don't think it's directly selectable but it appears to be the default under some circumstances?). This resulted in a test failure after refactoring it to a case statement, because I didn't include it directly or include a default case for the case statement, and that indirectly resulted in an invalid query. It feels like there should have been a test to catch the original bug (which would have replaced the sort order with `word_count` instead of `_score`) but I'm not sure where that test should go? Probably in bookmarkable_query.rb I guess :thinking:

irrationalpie7: I added some tests in bookmarkable_query_spec to verify that the sort_column gets respected in the default cases

public/help/bookmark-search-numerical-help.html
irrationalpie7: Removed, and removed the link

spec/models/search/bookmark_search_form_spec.rb
irrationalpie7: Removed in a few other of my tests too

spec/models/search/bookmark_search_form_spec.rb
irrationalpie7: Done!

irrationalpie7:
@zz9pzza and @brianjaustin I believe I've addressed or asked follow-ups for all your comments. I've left comment threads open so you can address if I've misunderstood your comments, but please feel free to resolve any threads that seem done!

I think the only open question I still have is around translation because I didn't quite understand the comment about lazy i18n keys/reusing translations (Edit: I think I figured this one out after all :tada:)

Edit: oh and also rubocop keeps complaining about a part of a file I didn't touch--am I okay to ignore that?

app/views/bookmarks/_filters.html.erb
irrationalpie7: I think I figured out after all what you meant and changed it like you wanted, but please let me know if I misunderstood!

app/models/work.rb
brianjaustin: Yeah, usually easiest to just put all of them together (_technically_, you can go back via `public`, but it's a bit messy to go back and forth)

app/models/search/bookmarkable_query.rb
brianjaustin: That works; thanks!

app/models/series.rb
brianjaustin: @irrationalpie7 sorry for the late reply
>  but the series blurb won't include the hidden work's word count, even for creators. Does that distinction make sense? Is that different from the desired behavior?

that's fine for now, and similar to behaviour we'll be having in the other PR as well

features/bookmarks/bookmark_search.feature
lydia-theda: this is how reverse sorting currently functions/can be accessed: https://archiveofourown.org/admin_posts/10851 
> `sort:>words` to reverse the sorting order and show shortest works first

(personally, I always forget whether I need the < or the > for a given sort type; doesn't help that what is entered doesn't persist/remain in the Sort within results field after the search is run)

irrationalpie7: I'd have to double check to verify, but I don't think bookmark search supports the "sort" syntax at the moment. At least, I tried "sort:>updated" and the only thing it did was return 0 results

spec/models/search/bookmark_query_spec.rb
brianjaustin: I think the existing split is fine 👍

PR #4453, opened 2023-02-08
AO3-6440 Restrict access to password reset pages

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6440

## Purpose

If you're logged in as an admin and try to access the new or edit user password pages, redirect to homepage with reminder to log out.

If you're logged in as a user and try to access the new or edit admin password pages, redirect to homepage with reminder to log out.

## Testing Instructions

Log in as an admin and try to go to /users/password/new and /users/password/edit. You should be redirected with a friendly blue log out reminder.

Log in as a user and try to go to /admin/password/new and /admin/password/edit. You should be redirected with a friendly blue log out reminder.

### DIFF
--- a/app/controllers/admin/passwords_controller.rb
+++ b/app/controllers/admin/passwords_controller.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class Admin::PasswordsController < Devise::PasswordsController
-  before_action :user_logout_required
+  before_action :user_logout_required, only: [:new, :edit]
   skip_before_action :store_location
   layout "session"
 end
--- a/app/controllers/admin/passwords_controller.rb
+++ b/app/controllers/admin/passwords_controller.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class Admin::PasswordsController < Devise::PasswordsController
-  before_action :user_logout_required, only: [:new, :edit]
+  before_action :user_logout_required
   skip_before_action :store_location
   layout "session"
 end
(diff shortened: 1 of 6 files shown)


app/controllers/admin/passwords_controller.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 class Admin::PasswordsController < Devise::PasswordsController
+  before_action :user_logout_required, only: [:new, :edit]
```
redsummernight, line 4: Should we limit this to just new and edit? Can we have the callback on all actions on the controller?

app/models/admin.rb
```diff
@@ -9,6 +9,7 @@ class Admin < ApplicationRecord
          :validatable,
          password_length: ArchiveConfig.ADMIN_PASSWORD_LENGTH_MIN..ArchiveConfig.ADMIN_PASSWORD_LENGTH_MAX,
          reset_password_within: ArchiveConfig.DAYS_UNTIL_ADMIN_RESET_PASSWORD_LINK_EXPIRES.days,
+         reset_password_keys: [:email, :login],
```
redsummernight, line 12: Would this require both keys to be provided for a reset?

sarken: Oops, sorry, that wasn't supposed to be here -- I was playing yesterday and forgot to check it out when I was done.

app/controllers/admin/passwords_controller.rb
sarken: Can do! I wasn't sure which we wanted. Should it be on all actions in both controllers?

redsummernight: I'd prefer all actions, yeah.

redsummernight (approved):

PR #4454, opened 2023-02-09
AO3-6476 Update styling of muted users notice

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6476

## Purpose

Reposition and restyle the notice about muting affecting the number of items on a page so it better resembles the existing notice about items on a page.  

## Testing Instructions

Refer to Jira.

## Credit

Sarken, she/her

### DIFF
--- a/app/views/muted/_muted_items_notice.html.erb
+++ b/app/views/muted/_muted_items_notice.html.erb
@@ -1,3 +1,3 @@
 <% if user_has_muted_users? %>
-  <p class="notes"><%= t("muted.muted_items_notice_html", muted_users_link: link_to(t("muted.muted_users_link_text"), user_muted_users_path(current_user))) %></p>
+  <p class="muted notice"><%= t("muted.muted_items_notice_html", muted_users_link: link_to(t("muted.muted_users_link_text"), user_muted_users_path(current_user))) %></p>
 <% end %>
--- a/app/views/bookmarks/index.html.erb
+++ b/app/views/bookmarks/index.html.erb
@@ -1,4 +1,6 @@
 <!--SEARCHBROWSE descriptive page name and system messages, descriptions, and instructions.-->
+<%= render "muted/muted_items_notice" %>
+
 <h2 class="heading">
   <% if @bookmarkable_items %>
     <%= search_header @bookmarkable_items, @search, "Bookmarked Item", @owner %>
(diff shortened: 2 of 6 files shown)


redsummernight (approved):

app/views/muted/_muted_items_notice.html.erb
```diff
@@ -1,3 +1,3 @@
 <% if user_has_muted_users? %>
-  <p class="notes"><%= t("muted.muted_items_notice_html", muted_users_link: link_to(t("muted.muted_users_link_text"), user_muted_users_path(current_user))) %></p>
+  <p class="muted notice"><%= t("muted.muted_items_notice_html", muted_users_link: link_to(t("muted.muted_users_link_text"), user_muted_users_path(current_user))) %></p>
```
tickinginstant, line 2: As I understand it, the eventual plan is to add placeholders for muted comments with unmuted replies, so that if someone replies to someone you've muted, you can see that they're replying to something you can't see, and not the previous comment. Those placeholders should probably be marked with a specific CSS class, and IMHO `muted` would be the best match for the style of [other](https://github.com/otwcode/otwarchive/blob/850b858b8034bd44d5fb851f2cfbfec3ae3766b0/app/helpers/comments_helper.rb#L321-L324) [classes](https://otwarchive.atlassian.net/browse/AO3-6034) on comments.

But using the same class as this notice could cause issues, especially if people start using site skins to style or hide this notice. Would it maybe make sense to use a different class name here? `mute-notice` or `has-muted-users` or some such? Or maybe even an ID instead of a class, since it's only displayed once per page?

sarken: So [our class system](http://otwcode.github.io/docs/front_end_coding/classes.html) is to use single-word class names in different combinations rather than unique classes or IDs. We rarely define styles for a single class without context unless we're _very_ sure the style should apply everywhere, and even then, sometimes we need to go back and adjust later as we find new uses for the class name. 

Skins basically come with an implied caveat emptor (or user) warning -- we've always reused existing class names in new locations and adjusted existing styles, and sometimes those changes interact poorly with existing skins. Here, though, `p.muted.notice { display: none; }` will avoid affecting other notices or other elements with the muted class (e.g., comments).

PR #4459, opened 2023-02-22
AO3-6460 Avoid caching stale AdminSetting

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6460

## Purpose

Write the modified record to the cache, instead of clearing the cache. Then the next read will get the modified version, instead of possibly fetching an old version from the database.

In specs, save the skin ID to the AdminSetting record instead of the whole Skin record. A newly created Skin record will have a closed file handle from generating a preview image and cannot be serialized for caching (IOError: closed stream).

https://github.com/otwcode/otwarchive/blob/d2e0c292d761e1f39491f6edd6c95a6097c1cc30/app/models/skin.rb#L584-L586

## Testing Instructions

See issue.

### DIFF
--- a/app/models/admin_setting.rb
+++ b/app/models/admin_setting.rb
@@ -6,7 +6,7 @@ class AdminSetting < ApplicationRecord
 
   before_save :update_invite_date
   before_update :check_filter_status
-  after_save :expire_cached_settings
+  after_commit :recache_settings
 
   belongs_to :default_skin, class_name: 'Skin'
 
@@ -71,8 +71,8 @@ class AdminSetting < ApplicationRecord
 
   private
 
-  def expire_cached_settings
-    Rails.cache.delete("admin_settings")
+  def recache_settings
+    Rails.cache.write("admin_settings", self)
   end
 
   def check_filter_status
--- a/app/models/admin_setting.rb
+++ b/app/models/admin_setting.rb
@@ -1,4 +1,6 @@
 class AdminSetting < ApplicationRecord
+  include AfterCommitEverywhere
+
   belongs_to :last_updated, class_name: 'Admin', foreign_key: :last_updated_by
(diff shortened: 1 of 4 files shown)


tickinginstant:
The bug with writing a settings object with a modified skin seems like the sort of thing that might come up again. Maybe it'd make sense to add a comment about it? (Alternatively, you could try wiping the `@association_cache` before writing the cache, or loading a fresh copy from the database, to avoid the bug altogether.)

redsummernight:
The `@association_cache` is not part of Rails' public API, I'll reload the record instead. Thanks!

app/models/admin_setting.rb
```diff
@@ -71,8 +71,11 @@ def self.perform(method, *args)
 
   private
 
-  def expire_cached_settings
-    Rails.cache.delete("admin_settings")
+  def recache_settings
+    # Reload the record; if the default skin was just created and has a
+    # closed file handle from generating a preview image, it cannot be
+    # serialized for caching.
+    Rails.cache.write("admin_settings", self.reload)
```
tickinginstant, line 78: Sorry, I guess I should have mentioned that reloading is a little more complex than it seems. You can't reload outside of the transaction, or you might get stale data. But to avoid writing incorrect data in the event that the transaction fails, you probably want to write to the cache outside of the transaction. So you probably either need multiple callbacks (an `after_save` one to load the fresh copy, and an `after_commit` one to actually write it) or a single `after_save` hook that calls `AfterCommitEverywhere.after_commit` to execute a block after the commit has finished.

(It might be simpler to just stick with the comment.)

redsummernight: I'll give the 2-callback approach a try, it could be more foolproof than only adjusting the specs and leaving comments.

tickinginstant (approved):

PR #4460, opened 2023-02-22
AO3-6474 Only load media for the menu when needed.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6474

## Purpose

Remove `ApplicationController#set_media` so that we only load the media list for the menu when we actually need to load the tags to regenerate the cache.

### DIFF
--- a/app/models/media.rb
+++ b/app/models/media.rb
@@ -15,9 +15,9 @@ class Media < Tag
     tag
   end
 
-  # The list of media used for the menu on every page. All media except No
-  # Media and Uncategorized Fandoms are listed in order, and then Uncategorized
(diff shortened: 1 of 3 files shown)


app/models/media.rb
```diff
@@ -14,4 +14,14 @@ def self.uncategorized
     tag.update(canonical: true) unless tag.canonical
     tag
   end
+
+  # The list of media used for the menu on every page. All media except No
+  # Media and Uncategorized Fandoms are listed in order, and then Uncategorized
```
redsummernight, line 19: Minor comment: can you quote the media tag names in the comment?

redsummernight (approved):
Thanks!

PR #4469, opened 2023-03-02
AO3-5499 Revamp user admin page

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5499

## Purpose

* Changes the subnavigation: 
    * repositions it so it's above the page text
    * shortens text
    * adds profile link
    * removes link for sending activation email
    * updates destination of the Manage Roles link so it uses exact search
* Removes action for sending activation email
* Adds new meta section with user email address and a link to the invitation used to create the account, if applicable (otherwise states no invitation was used)
* Stick the forms' submit buttons inside the field sets to strengthen visual association
* Where we disable all the form items inside a field set based on admin role, use the `field_set_tag` for tidier code
* Wrap the two separate field sets for suspensions, notes, and warnings in a surrounding field set to strengthen visual association
* i18n all the things
* Change the page title format
* Remove the now-unused controller action for sending activation emails
* Updates the tests to call this the "user administration" page 'cause "abuse administration" hasn't been accurate in a while

## Testing Instructions

Refer to Jira.

## Credit

Sarken, she/her


app/views/admin/admin_users/show.html.erb
```diff
@@ -1,110 +1,119 @@
...
+      <dd><%= @user.invitation ? link_to(@user.invitation.id, invitation_path(@user.invitation.id)) : t(".info.no_invitation") %></dd>
+    </dl>
+  </div>
+
+  <h3 class="heading"><%= t(".fnok.heading") %></h3>
   <%= form_tag action: "update_next_of_kin", controller: "admin/admin_users" do %>
     <%= error_messages_for @user.fannish_next_of_kin %>
     <%= hidden_field_tag "user_login", @user.login %>
-    <fieldset<% unless policy(@user).can_manage_next_of_kin? %> disabled="disabled"<% end %>>
-      <legend><%= ts("Fannish Next of Kin") %></legend>
```
redsummernight, line 35: Do we still need a `<legend>` here, or is it not needed if it's just one field set?

sarken: I was going back and forth on it myself. It's not required, and we have the heading, so it could be redundant, especially since we usually include a backup heading under the legend.

But I think I'll add legends to both field sets and let the existing headings serve as the backups.

redsummernight (approved):

PR #4470, opened 2023-03-03
AO3-6391 Improve the tag callbacks for updating TagNominations.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6391

## Purpose

- Cleans up the `update_tag_nominations` function by (a) splitting it into two separate callbacks, one for updating and one for clearing, and (b) avoiding passing around `tag` as a variable (which is vestige of the fact that this code used to be in a sweeper at one point).
- Modifies the `update_tag_nominations` function so that it now sets `parented` similarly to the way that `TagNomination#set_parented` sets the value of `parented`. In particular: https://github.com/otwcode/otwarchive/blob/8e03660aaa2b470011109a0af187716bf34e575c/app/models/tagset_models/tag_nomination.rb#L77-L81
- Fixes the test "If a nominated tag and its parent are wrangled after approval it should still be possible to associate them" so that it now accurately reflects reality (namely, that if a nominated tag is wrangled into its parent tag after approval, the association will disappear from the Review Associations page, and the two tags will appear associated in the tag set — although of course https://otwarchive.atlassian.net/browse/AO3-5714 is an ever-present problem).
- Adds a new test to check that when the nominated tag is wrangled into a *different* fandom, the nominated association won't be removed from the Review Associations page.

### DIFF
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -1070,7 +1070,6 @@ class Tag < ApplicationRecord
     if tag.canonical
       tag.add_to_autocomplete
     end
-    update_tag_nominations(tag)
   end
 
   after_update :after_update
@@ -1113,8 +1112,6 @@ class Tag < ApplicationRecord
     if tag.saved_change_to_unwrangleable?
       tag.reindex_document
     end
-
-    update_tag_nominations(tag)
   end
 
   before_destroy :before_destroy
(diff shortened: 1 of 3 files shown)


features/tag_sets/tag_set_associations.feature
```diff
@@ -14,15 +14,33 @@ Feature: Reviewing tag set associations
   Then I should see "Nominated associations were added"
     And I should not see "don't seem to be associated"
 
-  Scenario: If a nominated tag and its parent are wrangled after approval it should still be possible to associate them
+  Scenario: If a nominated tag is wrangled into its nominated parent after approval, it should be automatically associated with the parent
   Given I nominate and approve fandom "Floobry" and character "Zarrr" in "Nominated Tags"
-    And a canonical character "Zarrr" in fandom "Floobry"
```
tickinginstant, line 19: The reason this test didn't work before comes down to this step. It's defined here:
https://github.com/otwcode/otwarchive/blob/8e03660aaa2b470011109a0af187716bf34e575c/features/step_definitions/tag_steps.rb#L76-L80
If the tags in question exist in the database and aren't already canonical, then calling `where(name: name, canonical: true).first_or_create` will create a new tag with the same name, but with `canonical` set to true. These tags will fail to save, because they have duplicate tag names. But there are no calls to `save!` here, so everything will fail silently instead of erroring. Which makes this step a no-op.

redsummernight:
There was [a GitHub incident earlier](https://www.githubstatus.com/incidents/qfby3shb6r3w) that left the checks hanging, I closed and reopened the PR and that revived them.

app/models/tag.rb
```diff
@@ -1123,25 +1120,40 @@ def before_destroy
...
-      values[:parented] = tag.parents.any? {|p| p.is_a?(Fandom)}
-      values[:exists] = true
-    end
-    TagNomination.where(tagname: tag.name).update_all(values)
+  after_save :update_tag_nominations
+  def update_tag_nominations
+    TagNomination.where(tagname: name).update_all(
+      canonical: canonical,
+      synonym: merger.nil? ? nil : merger.name,
+      parented: false, # we'll fix this later
```
sarken, line 1132: This threw me for a sec -- I was like, "Well, why isn't this a TODO referencing an issue?"

Could you change it to something like, "We'll fix this later in the callback" instead?

features/tag_sets/tag_set_associations.feature
```diff
@@ -14,15 +14,33 @@ Feature: Reviewing tag set associations
...
+    And the tag "Floobry" is canonized
+    And I add the fandom "Floobry" to the character "Zarrr"
+    And the tag "Zarrr" is canonized
+  When I review associations for "Nominated Tags"
+  Then I should not see "Zarrr → Floobry"
+  # Remove this step when AO3-3757 is fixed:
```
sarken, line 24: Can you throw a `TODO:` on here for good measure?

sarken (approved):
Looks good! Just two comments about comments 😆

PR #4472, opened 2023-03-05
AO3-6003 Add i18n-tasks to normalize locale files and add tests to detect missing and unused translation keys

Bilka2:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6003

## Purpose

Install and configure the i18n-tasks gem.
Enable tests for missing, unused locale keys and normalized locale files (all English only).
Move all default locale inside `t()` calls to locale files, to prevent it from being reported as missing.
Fix 3 broken locales: [`comments.check_permission_to_edit.error.frozen`](https://github.com/otwcode/otwarchive/commit/310949703443e8684a556d1c634c73173273af7d), [`blocked.users.index.heading.landmark.blocked_users` and `muted.users.index.heading.landmark.muted_users`](https://github.com/otwcode/otwarchive/commit/fe2a1826a33f8397fe9cdcdc050070834d10551a).
Ignore incorrect reporting of unused locales/mark locales as used where possible.
Add test for ensuring no locale values contain newlines. (English only)
Remove newlines from [signup_notification mailer locale](https://github.com/otwcode/otwarchive/commit/23def42a66029234e08cf55b58ab20098f9a601c).
Normalize English locale files.

## Testing Instructions

I don't think it was expected that this would touch so many files, so this may require more testing than what the Jira issue states. I would request that at least it is tested that all pages/emails that I touched load, don't care if automated or manual.

I manually verified that the 3 locales I fixed are broken on production and fixed in dev. I also manually verified that 5c98a61e62491ff1dad8b0b58ce263eaee7c9b5a is correct.

## Notes

The tests are configured to run only for the English locale. If they would run for all locales, "missing" would simply report missing translations (about [1000](https://gist.github.com/Bilka2/c96920d14599186c8480c49f73334e7d#file-missing-txt)) and "normalized" expects different formatting than the phrase export produces, more info below. The tests that make sense to run for all locales are the "unused" test and the "newlines" test. However the problems would have to be fixed, probably in phrase?

### Test: Unused keys (all locales)
Reports 120 unused keys. Some of these are related to 83e6e2d730135cdbb16959d250645e3c5f727199, but most are legitimate unused keys, mostly in the mailers. Output from the task: https://gist.github.com/Bilka2/c96920d14599186c8480c49f73334e7d#file-unused-txt

### Test: Newlines in values (all locales)

The following locale keys have values that contain newlines:
`["fa.user_mailer.signup_notification.text.part3", "fr.user_mailer.signup_notification.text.part3", "id.user_mailer.invitation_to_claim.text.part2", "pt-BR.user_mailer.invitation_to_claim.text.part7", "uk.user_mailer.anonymous_or_unrevealed_notification.anonymous_info"]`

### Multi locale configuration problems

The gem does not allow routing locale writes by language. This means that [moving locales](https://otwarchive.atlassian.net/browse/AO3-6240) will place the locale in the wrong files. The config has a commented out line to make it work correctly for any locale besides en.

Moving locales is further complicated by the fact that the gem normalizes files that locale is moved to/in. The gem disagrees how files should be normalized compared to phrase: Alphabetical sorting is slightly different (`users` vs `user_mailer`) and handling of line_length (when configured) is different.

So [AO3-6240](https://otwarchive.atlassian.net/browse/AO3-6240) will not work as is, but may work with some workarounds:
* Uncommenting the line in the config
* Running the move tasks
* Manually merging the produced en.yml file with the mailers/en.yml file
* (optional) manually undoing the different normalization of the other locale files before the phrase import.
* Removing the line in the config again 

## Credit

Bilka (he/him)

[AO3-6240]: https://otwarchive.atlassian.net/browse/AO3-6240?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.phrase.yml
+++ b/.phrase.yml
@@ -11,6 +11,10 @@ phrase:
       - file: ./config/locales/devise/en.yml
         params:
           update_translations: true
+      # helpers
+      - file: ./config/locales/helpers/en.yml
+        params:
+          update_translations: true
       # mailers
       - file: ./config/locales/mailers/en.yml
         params:
--- a/Gemfile
+++ b/Gemfile
@@ -144,6 +144,7 @@ group :test, :development do
   gem 'whiny_validation'
   gem "factory_bot_rails"
   gem 'minitest'
+  gem 'i18n-tasks', require: false
 end
 
 group :development do
--- a/Gemfile
+++ b/Gemfile
@@ -144,7 +144,7 @@ group :test, :development do
   gem 'whiny_validation'
   gem "factory_bot_rails"
   gem 'minitest'
-  gem 'i18n-tasks', require: false
+  gem "i18n-tasks", require: false
 end
 
 group :development do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -291,6 +291,17 @@ GEM
       multi_xml (>= 0.5.2)
     i18n (1.12.0)
       concurrent-ruby (~> 1.0)
+    i18n-tasks (1.0.12)
+      activesupport (>= 4.0.2)
+      ast (>= 2.1.0)
+      better_html (>= 1.0, < 3.0)
+      erubi
+      highline (>= 2.0.0)
+      i18n
+      parser (>= 2.2.3.0)
+      rails-i18n
+      rainbow (>= 2.2.2, < 4.0)
+      terminal-table (>= 1.5.1)
     jmespath (1.6.1)
     json (2.6.1)
     kgio (2.10.0)
@@ -531,6 +542,8 @@ GEM
       activesupport (>= 5.2)
       sprockets (>= 3.0.0)
     stringex (2.8.5)
+    terminal-table (3.0.2)
+      unicode-display_width (>= 1.1.1, < 3)
     terrapin (0.6.0)
       climate_control (>= 0.0.3, < 1.0)
     test-unit (3.5.3)
@@ -622,6 +635,7 @@ DEPENDENCIES
   google_visualr!
   htmlentities
   httparty
+  i18n-tasks
   kgio (= 2.10.0)
   kt-paperclip (>= 5.2.0)
   launchy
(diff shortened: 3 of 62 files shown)


redsummernight:
> I don't think it was expected that this would touch so many files, so this may require more testing than what the Jira issue states.

Yeah, that's a big reason why this hasn't been done yet. I was looking for a way to install the check partially, to only prevent new bad keys while allowing us to slowly to work through the existing bad keys (same idea as [a .rubocop_todo.yml file](https://github.com/rubocop/rubocop/blob/master/.rubocop_todo.yml) for style violations). But we may have to bite the bullet and fix everything once.

Bilka2:
The change that touches a lot of files and could be left for later is this one:
> Move all default locale inside `t()` calls to locale files, to prevent it from being reported as missing.

There is no "todo" config in the gem, however it would be possible to [ignore](https://github.com/otwcode/otwarchive/pull/4472/files#diff-c1ded81b69dea6a1b03b09ed3a03e8b5a56a9fe604eff2febb6099931ffa6009R116) those keys.

So, that's an option, but I'd still bite the bullet. It looks like locales that aren't in the yml files aren't available for translation in Phrase, so once the goal is to translate the full site, there will have to be a cleanup of these locale keys. Might as well do it now then, since the work already done. But that just my view, I'm not the person reviewing or testing this :sweat_smile: (Though I would help out with testing on staging if that is somehow possible.)

Anyway, if you'd rather use the ignore list for these keys, let me know. It shouldn't be a problem for me to cherry pick the changes and to fill the ignore list with the keys that use default locales inside `t()` calls.

Bilka2:
Version with the ignore list: #4497

PR #4473, opened 2023-03-06
AO3-5666 changed invalid username error text

ciaraou:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5666

## Purpose

Changes all invalid username errors from "Login ..." to "User name ..." for consistency of language. Also switched out more HTML heavy error display logic for user creation, in favor of calling a function in a .rb file.

## Testing Instructions

Jira ticket explains how to reach the error and verify the changed text — should look like this
Screenshot 2023-03-05 at 3 56 47 PM
Screenshot 2023-03-05 at 4 00 57 PM
image

You also may want to run user_rename.feature, user_create.feature, and user_spec.rb

## Credit

pinkpurpleblue (she/her)

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -203,7 +203,9 @@ class User < ApplicationRecord
                                    max_pwd: ArchiveConfig.PASSWORD_LENGTH_MAX)
 
   validates_format_of :login,
-                      message: ts("must be 3 to 40 characters (A-Z, a-z, _, 0-9 only), no spaces, cannot begin or end with underscore (_)."),
+                      message: ts("must be %{min_pwd} to %{max_pwd} characters (A-Z, a-z, _, 0-9 only), no spaces, cannot begin or end with underscore (_).",
+                                    min_pwd: ArchiveConfig.LOGIN_LENGTH_MIN,
+                                    max_pwd: ArchiveConfig.LOGIN_LENGTH_MAX),
                       with: /\A[A-Za-z0-9]\w*[A-Za-z0-9]\Z/
   validates :login, uniqueness: { message: ts("has already been taken") }
   validate :login, :username_is_not_recently_changed, if: :will_save_change_to_login?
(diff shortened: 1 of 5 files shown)


ciaraou:
I changed the Create Account error message because I couldn't figure out how to get rid of the extra "Login" word with the way it was structured before... it functions the same but it has a different header message ("Oops, there's some problems with the stuff you told us" -> "Sorry! We couldn't save this user because:") — I can switch it back if it's preferred to keep the "oops" version
Screenshot 2023-03-05 at 3 56 52 PM
Screenshot 2023-03-05 at 3 56 47 PM

app/models/user.rb
```diff
@@ -203,9 +203,11 @@ def unread_inbox_comments_count
                                    max_pwd: ArchiveConfig.PASSWORD_LENGTH_MAX)
 
   validates_format_of :login,
-                      message: ts("must be 3 to 40 characters (A-Z, a-z, _, 0-9 only), no spaces, cannot begin or end with underscore (_)."),
+                      message: ts("^User name must be %{min_pwd} to %{max_pwd} characters (A-Z, a-z, _, 0-9 only), no spaces, cannot begin or end with underscore (_).",
+                                  min_pwd: ArchiveConfig.LOGIN_LENGTH_MIN,
```
sarken, line 207: Can you change `pwd` to `login` here, since we're describing the login length, not password length?

sarken:
Getting rid of the "Oops, there's some problems" text is _great_ -- thank you!

app/views/users/registrations/new.html.erb
```diff
@@ -1,14 +1,7 @@
 <!--Descriptive page name, messages and instructions-->
 <h2 class="heading"><%= ts("Create Account") %></h2>
 <% if resource.errors.any? %>
```
sarken, line 3: You _might_ be able to get rid of the surrounding `resource.errors.any?` logic as well -- could you check if you haven't?

ciaraou: When I tried before it broke so I just put it back — I can investigate a little more into it to see if there's anything I can do about it though

app/models/user.rb
ciaraou: Got it — thanks for catching that 😅

app/views/users/registrations/new.html.erb
ciaraou: Hm I'm not sure what broke before but it works now haha — I'll get these changes pushed shortly

sarken (approved):
Thank you!

PR #4477, opened 2023-03-10
AO3-6482 Improve reliability of skins:cache_all_site_skins

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6482

## Purpose

Makes sure that all of the skins in the chooser get cached when `rake skins:cache_all_site_skins` is run, because you can't use a skin in the chooser if it's not cached. 

## Testing Instructions

Refer to Jira.

## Credit

Sarken, she/her

### DIFF
--- a/lib/tasks/skin_tasks.rake
+++ b/lib/tasks/skin_tasks.rake
@@ -160,9 +160,10 @@ namespace :skins do
   desc "Cache all site skins"
   task(cache_all_site_skins: :environment) do
     skins = Skin.in_chooser + [Skin.default]
-    successes, failures = [], []
+    successes = []
+    failures = []
 
-    skins.each do  |skin|
+    skins.each do |skin|
       if skin.cache!
         successes << skin.title
       else
--- a/lib/tasks/skin_tasks.rake
+++ b/lib/tasks/skin_tasks.rake
@@ -159,7 +159,9 @@ namespace :skins do
 
   desc "Cache all site skins"
   task(cache_all_site_skins: :environment) do
-    skins = Skin.in_chooser + [Skin.default]
+    # The default skin can be changed to something other than Skin.default via
+    # admin settings, so we want to cache that skin, not Skin.default.
+    skins = Skin.where(id: AdminSetting.default_skin_id).or(Skin.in_chooser)
     successes = []
     failures = []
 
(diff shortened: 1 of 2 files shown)


lib/tasks/skin_tasks.rake
```diff
@@ -158,8 +158,21 @@ namespace :skins do
   end
 
   desc "Cache all site skins"
-  task(:cache_all_site_skins => :environment) do
-    Skin.where(cached: true).each{|skin| skin.cache!}
+  task(cache_all_site_skins: :environment) do
+    skins = Skin.in_chooser + [Skin.default]
+    successes, failures = [], []
```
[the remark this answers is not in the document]
sarken: But it's nicer here 😢

lib/tasks/skin_tasks.rake
```diff
@@ -158,8 +158,22 @@ namespace :skins do
   end
 
   desc "Cache all site skins"
-  task(:cache_all_site_skins => :environment) do
-    Skin.where(cached: true).each{|skin| skin.cache!}
+  task(cache_all_site_skins: :environment) do
+    skins = Skin.in_chooser + [Skin.default]
```
redsummernight, line 162: The "default" option in the skin chooser is the unset action, which does this:

https://github.com/otwcode/otwarchive/blob/73ee2a428a0a2ce364fd73133e108251ecbd9121/app/controllers/skins_controller.rb#L140-L145

So we should probably pick the skin ID from the admin setting instead:
```rb
Skin.where(id: AdminSetting.default_skin_id).or(Skin.in_chooser).each do |skin|
```
You'll need to set `default_skin: true` on the describe block to set up the admin setting as well.

sarken: Isn't `Skin.default` the actual [default skin](https://test.archiveofourown.org/skins/1), though, distinct from the "Default"/unset link in the footer? 

https://github.com/otwcode/otwarchive/blob/73ee2a428a0a2ce364fd73133e108251ecbd9121/app/models/skin.rb#L540-L542

...although I guess `AdminSetting.default_skin_id` could be set to Panda Madness rather than Default, and in that case, we'd want to recache Panda Madness, not Default, so the concern still applies.

redsummernight (approved):

PR #4483, opened 2023-03-19
AO3-6425 Don't mangle reported chapter URLs without schemes

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6425

## Purpose

When you reported a chapter URL without including http or https at the start, the URL was coming back with the work ID at the front: `/works/1080706archiveofourown.org/chapters/2165131/`

This adds the scheme to the start of URLs so that won't happen.

I didn't have much time so I wanted to get something functional, but it is ugly. Please help me make it less ugly.

## Testing Instructions

Submit an abuse report with a URL like archiveofourown.org/chapters/123 and make sure it doesn't error.

### DIFF
--- a/app/models/abuse_report.rb
+++ b/app/models/abuse_report.rb
@@ -48,14 +48,13 @@ class AbuseReport < ApplicationRecord
   def standardize_url
     return unless url =~ %r{((chapters|works)/\d+)} || url =~ %r{(users\/\w+)}
 
-    url_with_scheme = add_scheme_to_url(url)
-    clean_url = clean_url(url_with_scheme)
-    standardized_url = add_work_id_to_url(clean_url)
-    self.url = standardized_url
+    self.url = add_scheme_to_url(url)
+    self.url = clean_url(url)
+    self.url = add_work_id_to_url(self.url)
   end
 
   def add_scheme_to_url(url)
-    uri = Addressable::URI.parse url
+    uri = Addressable::URI.parse(url)
     return url unless uri.scheme.nil?
 
     "https://#{uri}"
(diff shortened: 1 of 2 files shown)


app/models/abuse_report.rb
```diff
@@ -39,42 +39,57 @@ def akismet_attributes
...
+  before_validation :standardize_url, on: :create
+  def standardize_url
+    return unless url =~ %r{((chapters|works)/\d+)} || url =~ %r{(users\/\w+)}
+
+    self.url = add_scheme_to_url(url)
+    self.url = clean_url(url)
+    self.url = add_work_id_to_url(self.url)
+  end
+
+  def add_scheme_to_url(url)
```
redsummernight, line 56: You can combine this with `clean_url` so you have to run `Addressable::URI.parse` one less time. When you have the parsed uri object I think you can do `uri.scheme = "https"` if it doesn't have a scheme.

sarken: Unfortunately, that didn't work. If you have a URL without a scheme and try to add one with `uri.scheme`, you just get, e.g., `https:archiveofourown.org`.  It has something to do with it not properly recognizing the host and treating it like a path, I think?

```ruby
irb(main):012:0> url = "archiveofourown.org/chapters/000"
=> "archiveofourown.org/chapters/000"
irb(main):013:0> uri = Addressable::URI.parse(url)
=> #<Addressable::URI:0x11044 URI:archiveofourown.org/chapters/000>
irb(main):014:0> uri.scheme = "https"
=> "https"
irb(main):015:0> uri.to_s
=> "https:archiveofourown.org/chapters/000"
irb(main):016:0> uri.host
=> nil
irb(main):017:0> uri.path
=> "archiveofourown.org/chapters/000"
```

redsummernight: Gotcha, then this is fine as is.

redsummernight (approved):

PR #4488, opened 2023-04-01
AO3-6044 Use PERMITTED_HOSTS to validate abuse report URLs

brianjaustin:
# Pull Request Checklist

* [ ] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)? (n/a)

## Issue

https://otwarchive.atlassian.net/browse/AO3-6044

## Purpose

Move the regex used to validate URLs in abuse reports to config.yaml. This will allow us to override it on test environment(s).

## Testing Instructions

Listed in the Jira description.

## References

## Credit

Brian Austin (they/he)

### DIFF
--- a/config/config.yml
+++ b/config/config.yml
@@ -483,7 +483,11 @@ PERMITTED_HOSTS: [
   "ao3.org",
   "www.archiveofourown.com",
   "www.archiveofourown.net",
-  "www.archiveofourown.org"
+  "www.archiveofourown.org",
+  # Staging
+  "insecure-test.archiveofourown.org",
+  "test.archiveofourown.org",
+  "testdownload.archiveofourown.org"
 ]
 
 USER_RENAME_LIMIT_DAYS: 7
--- a/config/config.yml
+++ b/config/config.yml
@@ -480,7 +480,7 @@ PERMITTED_HOSTS: [
   "download.archiveofourown.org",
   "insecure.archiveofourown.org",
   "secure.archiveofourown.org",
-  "ao3.org",
+  "www.ao3.org",
   "www.archiveofourown.com",
   "www.archiveofourown.net",
   "www.archiveofourown.org",
--- a/config/config.yml
+++ b/config/config.yml
@@ -503,3 +503,7 @@ STAT_COUNTER_BATCH_SIZE: 100
 # Number of hours to retain the original creator for works that have been
 # orphaned.
 ORIGINAL_CREATOR_TTL_HOURS: 72
+
+# Regex for the site's domains, used to validate that user-supplied
+# links are located on the Archive when appropriate.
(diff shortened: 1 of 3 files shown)


brianjaustin:
To fully achieve the ticket's goal, we will need systems to add an override regex on staging. I tried this out and it seems to work
```regex
^(?:https?:\\/\\/)?(?:www\\.)?(?:insecure-)?test\\.archiveofourown\\.org.*
```

config/config.yml
```diff
@@ -503,3 +503,7 @@ STAT_COUNTER_BATCH_SIZE: 100
 # Number of hours to retain the original creator for works that have been
 # orphaned.
 ORIGINAL_CREATOR_TTL_HOURS: 72
+
+# Regex for the site's domains, used to validate that user-supplied
+# links are located on the Archive when appropriate.
+APP_URL_REGEX: "^(https?:\\/\\/)?(www\\.|(insecure\\.))?(archiveofourown|ao3)\\.(org|com).*"
```
brianjaustin, line 509: I went to `archiveofourown.com` and noticed that it has a proxy warning; should it be added to the allowed hosts, or do we want to remove it from this regex? Similarly, `ao3.com` does not resolve at all

sarken: Yeah, let's add `archiveofourown.com` (and `archiveofourown.net`, which is also displaying the proxy notice) to `PERMITTED_HOSTS`. 

I think `ao3.org` is the only `ao3` we have, no `.com` or `.net`, but we can check with Systems.

redsummernight, line 509: I'm not sure about having a regex in the config file. I think configs should be simple values. Regexes are complicated and you won't know they don't work until run time, so it's easy for us to mistype it when overriding in local.yml and not notice because it doesn't go through our usual testing process.

Following the discussion in [AO3-6496](https://otwarchive.atlassian.net/browse/AO3-6496), I think we can use `Addressable::URI.heuristic_parse` (which handles URLs without schemes for us, see test cases in #4483). Once parsed, we can match `.host` against the existing config value `PERMITTED_HOSTS`.

sarken: Just a heads up if you didn't see the convo, only ao3.org is ours.

brianjaustin: Oh that's a neat method! I agree, that's a much cleaner way to do things. Would we want systems to override `PERMITTED_HOSTS` in production to avoid letting staging URLs through there?

redsummernight: Yeah, we should have `PERMITTED_HOSTS` be different between staging and production in that case.

brianjaustin: I've pulled out the staging values, since that way would impact fewer people if the override was delayed for some reason

redsummernight: You can leave the staging values in, for public tracking purposes. Currently, `PERMITTED_HOSTS` is always overridden by our deploy process for staging/production, so it needs to be handled before we merge regardless.

redsummernight (approved):
Thanks!

PR #4490, opened 2023-04-05
AO3-6496 Add the --rm option to docker-compose run commands.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6496

## Purpose

- Adds the `--rm` option when calling `docker-compose run`, to avoid leaving behind dead containers.
- Changes the init scripts to use `rake db:migrate` instead of `script/reset_database.sh` for the test environment, because the development script already creates the database and loads the schema for the test environment.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,7 +18,7 @@
 # You can run tests like this:
 #
 # ```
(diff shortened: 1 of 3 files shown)


redsummernight (approved):
Nice, thanks!

PR #4491, opened 2023-04-08
AO3-6504 Avoid resetting dev databases twice for Docker

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6504

## Purpose

The task `db:otwseed` depends on `db:drop`, `db:create`, etc. so we shouldn’t be [running both in the development environment](https://github.com/otwcode/otwarchive/blob/ccd24c3697144cc303f17c0e19cc65ff24915a5b/script/reset_database.sh#L13-L21).

## Testing Instructions

Tested locally by:
- Running the script from scratch: `./script/docker/init.sh`;
- Running `docker-compose run --rm test script/reset_database.sh` after the development environment has been set up and checking that only the test database is reset (and only once);
- Running `docker-compose run -e RAILS_ENV=staging --rm test script/reset_database.sh` and getting the error for an unsupported environment.

### DIFF
--- a/script/reset_database.sh
+++ b/script/reset_database.sh
@@ -3,24 +3,21 @@
 set -ex
 
 case "${RAILS_ENV}" in
-test) ;;
-development) ;;
+test)
+  bundle install
+  bundle exec rake db:reset_and_migrate
+  ;;
+development)
+  bundle install
+  bundle exec rake db:otwseed
+  bundle exec rake skins:load_site_skins
+  bundle exec rake search:index_tags
+  bundle exec rake search:index_works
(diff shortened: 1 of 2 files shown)


script/reset_database.sh
```diff
@@ -10,19 +12,14 @@ development) ;;
 esac
 
 bundle install
```
tickinginstant, line 14: With the changes you've made, it appears that this `bundle install` line is the only code shared between the two environments. Would it make sense to move all of the code into the case statement above?

tickinginstant (approved):
Thanks!

PR #4492, opened 2023-04-08
AO3-6503 Allow superadmins to disable guest comments across the site

EchoEkhi:
# Pull Request Checklist

# Todo

* [x] Fix weird bug where the reply to comment function doesn't work
## Issue

https://otwarchive.atlassian.net/browse/AO3-6503

## Purpose

Allow superadmins to disable guest comments across the site.

## Testing Instructions

See JIRA ticket.

## Credit

EchoEkhi


app/controllers/comments_controller.rb
```diff
@@ -19,6 +19,7 @@ class CommentsController < ApplicationController
   before_action :check_permission_to_edit, only: [:edit, :update ]
   before_action :check_permission_to_delete, only: [:delete_comment, :destroy]
   before_action :check_parent_comment_permissions, only: [:new, :create, :add_comment_reply]
+  before_action :check_guest_comment_permissions, only: [:new, :create, :add_comment_reply]
```
sarken, line 22: Could we change this to something like `check_site_comment_permissions` or `check_comment_admin_setting` for clarity?

spec/controllers/comments_controller_spec.rb
```diff
@@ -446,6 +446,32 @@
         end
       end
     end
+
+    context "guest comments are turned on in admin settings" do
+      let(:work) { create(:work) }
```
sarken, line 451: Let's set the `AdminSetting` here as well -- even though `guest_comments_off: false` is the default, it's probably safer/a good idea to make it explicit.

spec/controllers/comments_controller_spec.rb
```diff
@@ -446,6 +446,32 @@
...
+
+      it "allows guest comments" do
+        post :create, params: { work_id: work.id, comment: anon_comment_attributes }
```
sarken, line 454: I think there are some other things we need to be testing here:
* the other actions affected by `check_guest_comment_permissions` (so `new` and `add_comment_reply`)
* how this interacts with other settings on commentables (i.e., when the creator of a work disallows all comments) -- I'm okay with just checking works in this case, although someone else might have a good argument for also testing admin posts
* that logged-in users can still comment (it's probably okay to just test this for when guest comments are off)

sarken (changes requested):

db/migrate/20230408145819_add_guest_comment_off_to_admin_settings.rb
```diff
@@ -0,0 +1,5 @@
+class AddGuestCommentOffToAdminSettings < ActiveRecord::Migration[6.1]
+  def change
+    add_column :admin_settings, :guest_comments_off, :boolean, default: false, null: false
```
tickinginstant, line 3: Since you're adding a new column here, I think it might be good to change the key used to cache the admin settings. That way we can keep the deploy steps simpler:
1. Run the migration.
2. Do a rolling restart of the servers and workers with the new code.

The alternative would be to add some extra steps to the deploy to prevent errors:
1. Make sure that all admins know not to update the admin settings during this period, and make sure that the deploy isn't going to overlap with 6:21 AM UTC, when the invitations code changes the admin settings.
2. Run the migration.
3. Open a copy of the rails console with the new code and run `AdminSetting.first.recache_settings`.
4. Do a rolling restart of the servers and workers with the new code.

features/step_definitions/admin_steps.rb
```diff
@@ -88,6 +88,20 @@
   click_button("Update")
 end
 
+Given /^guest comments are on$/ do
```
tickinginstant, line 91: Would you mind writing this step definition as a string instead of a regex? I think at this point, the goal is to use regexes only for complex steps that can't be handled by cucumber expressions.

features/step_definitions/comment_steps.rb
```diff
@@ -19,6 +19,11 @@
   FactoryBot.create(:comment, :by_guest, commentable: work.first_chapter)
 end
 
+Given "a guest comment on the admin post {string}" do |title|
+  admin_post = AdminPost.find_by(title: title)
+  FactoryBot.create(:comment, :by_guest, commentable: admin_post)
+end
```
tickinginstant, line 25: I think you can merge this with the previous step definition to get a single `"a guest comment on {commentable}"` step, like the ones defined below but for guests.

spec/controllers/comments_controller_spec.rb
sarken: So these shouldn't all be bunched here under `describe "POST #create"`: the `add_comment_reply` tests should be in the `describe "GET #add_comment_reply"` block and the `new` tests in `describe "POST #new"`.

features/admins/admin_settings.feature
```diff
@@ -36,3 +36,117 @@ Feature: Admin Settings Page
     When I am logged in as a random user
       And I go to the support page
     Then I should see "We can answer Support inquiries in"
+
+  Scenario: Turn on guest comments (on works)
+    Given guest comments are on
+      And I am logged out
+      And the work "Generic Work"
+      And a guest comment on the work "Generic Work"
```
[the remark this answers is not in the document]
EchoEkhi: This is to check if the "Reply" buttons on existing comments are hidden or not.

spec/controllers/comments_controller_spec.rb
```diff
@@ -104,6 +108,37 @@
         it_behaves_like "no one can add comment reply on a hidden comment"
       end
     end
+
+    context "guest comments are turned on in admin settings" do
```
sarken, line 112: This is a start, but we should test the redirect as well as the lack of error and be more specific about describing the behavior. 

We also need to extend the test, at least for when guest commenting is off:
>how this interacts with other settings on commentables (i.e., when the creator of a work disallows all comments) -- I'm okay with just checking works in this case, although someone else might have a good argument for also testing admin posts

Here's something to get you started. You can adapt it for the other actions.

```
    context "guest comments are turned on in admin settings" do
      let(:admin_setting) { AdminSetting.first || AdminSetting.create }

      before do
        admin_setting.update_attribute(:guest_comments_off, false)
      end

      it "redirects logged out user to the comment on the commentable without an error" do
        get :add_comment_reply, params: { comment_id: comment.id }
        expect(response).to redirect_to(chapter_path(comment.commentable, show_comments: true, anchor: "comment_#{comment.id}"))
      end
    end

    context "guest comments are turned off in admin settings" do
      let(:admin_setting) { AdminSetting.first || AdminSetting.create }
      let(:work) { comment.ultimate_parent }

      before do
        admin_setting.update_attribute(:guest_comments_off, true)
      end

      [:enable_all, :disable_anon].each do |permissions|
        context "when work comment permissions are #{permissions}" do
          before do
            work.update_attribute(:comment_permissions, permissions)
          end

          it "redirects logged out user with an error" do
            get :add_comment_reply, params: { comment_id: comment.id }
            it_redirects_to_with_error("/where_i_came_from", "Sorry, the Archive doesn't allow guests to comment right now.")
          end

          it "redirects logged in user to the comment on the commentable without an error" do
            fake_login
            get :add_comment_reply, params: { comment_id: comment.id }
            expect(response).to redirect_to(chapter_path(comment.commentable, show_comments: true, anchor: "comment_#{comment.id}"))
          end
        end
      end

      context "when work comment permissions are disable_all" do
        before do
          work.update_attribute(:comment_permissions, :disable_all)
        end

        it "redirects logged out user with an error" do
          get :add_comment_reply, params: { comment_id: comment.id }
          it_redirects_to_with_error("/where_i_came_from", "Sorry, the Archive doesn't allow guests to comment right now.")
        end

        it "redirects logged in user with an error" do
          fake_login
          get :add_comment_reply, params: { comment_id: comment.id }
          it_redirects_to_with_error(work_path(work), "Sorry, this work doesn't allow comments.")
        end
      end
    end
  end
```

spec/controllers/comments_controller_spec.rb
```diff
@@ -136,6 +199,10 @@
   end
 
   describe "POST #new" do
+    let(:anon_comment_attributes) do
```
sarken, line 202: So the existing tests are... mistaken, let's say. Much like the `add_comment_reply` action, the `new` actually is actually `GET`, not a `POST`, so you shouldn't need these attributes.

spec/controllers/comments_controller_spec.rb
```diff
@@ -196,6 +263,75 @@
...
+
+      before do
+        admin_setting.update_attribute(:guest_comments_off, false)
+      end
+
+      it "allows guest comments" do
+        post :new, params: { work_id: work.id, comment: anon_comment_attributes }
+
+        expect(flash[:error]).to be_nil
+      end
```
sarken, line 279: Here we should be checking that the correct template loads:
```
it "renders the :new template" do
  get :new, params: { work_id: work.id }
  expect(response).to render_template(:new)
end 
```

spec/controllers/comments_controller_spec.rb
```diff
@@ -196,6 +263,75 @@
...
+          before do
+            work.update_attribute(:comment_permissions, permissions)
+          end
+
+          it "redirects logged out user with an error" do
+            post :new, params: { work_id: work.id, comment: anon_comment_attributes }
+            it_redirects_to_with_error("/where_i_came_from", "Sorry, the Archive doesn't allow guests to comment right now.")
+          end
+
+          it "redirects logged in user to the comment on the commentable without an error" do
```
sarken, line 308: I think this description might be copied from elsewhere -- the behavior here is `it "renders the :new template for logged in user"`

sarken (approved):

PR #4495, opened 2023-04-15
AO3-6510 Handle hit counts with RedisHashJobs, a modified version of the RedisSetJob.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6510

## Purpose

- Create a subclass of the `RedisSetJob` called `RedisHashJob`, so that we can create batch jobs with Redis hashes as well as Redis sets.
- Rename the `RedisSetJobSpawner` to the `RedisJobSpawner`, to reflect its new role in spawning both `RedisSetJob`s and `RedisHashJob`s. Also modify the code so that the job spawning is handled by the job class, so that we don't need to know whether the job that we're processing is a hash or a set.
- Move the `retry_on ActiveRecord::Deadlocked` and `retry_on ActiveRecord::LockWaitTimeout` calls to the `RedisSetJob`, since it's used by almost all of the subclasses, and there's no harm in including it for the subclasses that don't need it. Also modifies the `retry_on` calls to change the wait to 5 minutes, so that if it's conflicting with a long transaction there'll be more time to wait for the transaction to end.
- Add a `HitCountUpdateJob` class to replace the `RedisHitCounter.save_recent_counts` method.
- Remove some (now unused) code from the `RedisHitCounter`, and use the `RedisScanning` module instead of having custom scanning code.
- Add `scan_hash_in_batches` to the `RedisScanning` module.


redsummernight (approved):
Looks good, thanks!

PR #4496, opened 2023-04-16
AO3-6512 Fix api works RSpec tests leaking state

Bilka2:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6512

## Purpose

Fixes that the tests in that `spec/controllers/api/v2/api_works_spec.rb` don't clean up all records created in `before(:all)`. This would leak records to other tests, see https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/BeforeAfterAll.

Before this PR, the following tests fail when run after api_works_spec.rb:
```
rspec ./spec/lib/tasks/after_tasks.rake_spec.rb:374 # rake After:remove_translation_admin_role remove translation admin role
rspec ./spec/helpers/works_helper_spec.rb:50 # WorksHelper#sorted_languages returns all languages sorted alphabetically
rspec ./spec/lib/tasks/work_tasks.rake_spec.rb:125 # rake work:reset_word_counts when there are multiple languages updates only works in the specified language
```

The first test failed due to the archivist role existing, the two other tests failed due to a language with the short "es" existing.

## Testing Instructions

Run the full RSpec test suite with `bundle exec rspec spec` or run `bundle exec rspec spec/controllers/api/v2/api_works_spec.rb spec/lib/tasks/after_tasks.rake_spec.rb spec/helpers/works_helper_spec.rb spec/lib/tasks/work_tasks.rake_spec.rb`. The CI does not reproduce/test this issue because the controller tests are run separately from the other tests.

## Credit

Bilka (he/him)

### DIFF
--- a/spec/controllers/api/v2/api_works_spec.rb
+++ b/spec/controllers/api/v2/api_works_spec.rb
@@ -144,14 +144,14 @@ describe "API v2 WorksController - Create works", type: :request do
 
     describe "Provided API metadata should be used if present" do
       before(:all) do
-        Language.find_or_create_by(short: "es", name: "Español")
+        Language.create(short: "es", name: "Español")
 
         mock_external
 
-        archivist = create(:archivist)
+        @archivist = create(:archivist)
 
         valid_params = {
-          archivist: archivist.login,
+          archivist: @archivist.login,
           works: [
             { id: "123",
               title: api_fields[:title],
@@ -181,6 +181,9 @@ describe "API v2 WorksController - Create works", type: :request do
       after(:all) do
         @work&.destroy
         WebMock.reset!
+        Language.find_by(short: "es").destroy
+        @archivist.destroy
+        Role.find_by(name: "archivist").destroy
       end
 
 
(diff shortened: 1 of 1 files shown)


spec/controllers/api/v2/api_works_spec.rb
```diff
@@ -464,6 +473,8 @@
       after(:all) do
         @work&.destroy
         WebMock.reset!
+        @archivist.destroy
```
redsummernight, line 476: Can you switch from `before(:all)`/`after(:all)` to `before`/`after`, which cleans up database records after each test, instead of manually destroying each of them at the end?

Bilka2: Done. I didn't switch it in my first commit because it means that the test time increases. On my local setup running `spec/controllers/api/v2/api_works_spec.rb` in a fresh container now takes 22s instead of 11s.

redsummernight: Thanks! Yeah, the WebMock setup can likely be optimized a little. But this should address the leaked state problem.

redsummernight (approved):

PR #4501, opened 2023-04-23
AO3-5556 Don't change collections on preview.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5556

## Purpose

- Generalize the function `assign_tags_of_type` to a function `assign_through_association` defined in `ApplicationRecord` that can be used for any type of association, not just tags.
- Use that function to assign collections, so that setting `collection_names` will only set up the changes to be performed on save instead of performing them immediately. Similarly with `collections_to_add` and `collections_to_remove`.
- Change the `approve_automatically` and `set_anonymous_and_unrevealed` callbacks in the `CollectionItem` class to run `before_validation` instead of `before_save`, so that we get the correct values for `anonymous`, `unrevealed`, `user_approval_status`, and `collection_approval_status` after validation.
- Make the `Collectible` module a concern and move it to the correct directory.
- Rewrite the `Collectible` function `set_anon_unrevealed` so that it uses the status of the collection items to compute whether the work should be anonymous/unrevealed, and make it run as an `after_validation` callback so that it has access to the collection item statuses set before validation.
- Rewrite `BookmarksController#update` to use the collection handling from `Collectible`, and add a function akin to `WorksController#in_moderated_collection` so that the message on adding a moderated collection stays the same as it was before the change.
- Stop calling `save(validate: false)` so much, and replace it with a `reindex_item` callback in the `CollectionItem` class. The `item` association on `CollectionItem`s has `touch: true`, so changes to collection items will already update the work/bookmark timestamp. And indexing is the only other real purpose that calling `save(validate: false)` serves.


app/models/application_record.rb
```diff
@@ -15,4 +15,35 @@ def update_sanitizer_version
   def self.random_order
     order(Arel.sql("RAND()"))
   end
+
+  def unmark_for_destruction
```
redsummernight, line 19: Can you add a simple test that calls `mark_for_destruction` then `unmark_for_destruction` on a record, and expects `marked_for_destruction?` to be false?

sarken: Just checking -- is this ready now that the tests have been added, or does it need a more thorough review?

app/models/work.rb
```diff
@@ -569,17 +520,8 @@ def visible?(user = User.current_user)
...
-    #!self.collection_items.unrevealed.empty?
-    in_unrevealed_collection?
-  end
-
-  def anonymous?(user = User.current_user)
-    # here we check if the story is in a currently-anonymous challenge
-    #!self.collection_items.anonymous.empty?
-    in_anon_collection?
-  end
+  alias_attribute :unrevealed, :in_unrevealed_collection
```
pmonfort, line 523: 👏

app/models/collection.rb
```diff
@@ -321,65 +324,30 @@ def gift_notification
...
   end
 
-  def moderated?() = self.collection_preference.moderated
-
-  def closed?() = self.collection_preference.closed
-
-  def unrevealed?() = self.collection_preference.unrevealed
-
-  def anonymous?() = self.collection_preference.anonymous
+  delegate :moderated?,
```
pmonfort, line 327: 💪

app/views/works/_meta.html.erb
```diff
@@ -49,12 +49,13 @@
       </dd>
     <% end %>
 
-    <% unless @work.approved_collections.empty? %>
+    <% collections_data = show_collections_data(@work.approved_collections_after_saving) %>
```
pmonfort, line 52: Non-blocking suggestion: I'm not a big fan of setting variables in views. One alternative would be to set this in the controller, but since this is a partial that could be rendered from different contexts, I think a helper would be cleaner

```
# app/helpers/works_helper.rb
def collections_meta_tag(work)
  collections_data = show_collections_data(work.approved_collections_after_saving)
  return if collections_data.blank?
  
  content_tag(:dt, ts("Collections:"), class: "collections") +
    content_tag(:dd, collections_data, class: "collections")
end
```

Then the view becomes:

<%= collections_meta_tag(@work) %>

app/models/application_record.rb
pmonfort, line 19: Both `unmark_for_destruction` and `assign_through_association` are only used by `Collectible` and `Taggable` should they live in ApplicationRecord? Maybe they can be extracted to a concern.

app/models/application_record.rb
```diff
@@ -15,4 +15,35 @@ def update_sanitizer_version
...
+
+  # Given an association of through records, the name of the "source" field,
+  # and a list of values, marks the through records as destroyed or not based
+  # on whether their "source" value is included in the list of desired values,
+  # and builds new through records when necessary.
+  #
+  # The optional parameter klass is used to filter the through records -- when
+  # it's set, this function won't mark or unmark any through records whose
+  # source doesn't have that klass.
+  def assign_through_association(through_association, source, values, klass: nil)
```
pmonfort, line 31: See comment on unmark_for_destruction above, same concern about scope in ApplicationRecord

pmonfort (commented):
Great work, this PR covers a lot of ground. I left some comments, let me know your thoughts.

pmonfort:
Hi! I reviewed this PR and left some comments. As discussed with the team, I'll be opening a follow-up PR based on your work to resolve the feedback and merge conflicts. Thanks!

PR #4502, opened 2023-04-27
AO3-6470 Tag wrangler report CSV

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6470

## Purpose

Add a button for tag wrangling admins to download a wrangler's wrangled tags as CSV.

## Testing Instructions

On Jira ticket

## References

## Credit

Brian Austin (they/he)

### DIFF
--- a/app/controllers/tag_wranglers_controller.rb
+++ b/app/controllers/tag_wranglers_controller.rb
@@ -1,6 +1,8 @@
 class TagWranglersController < ApplicationController
+  include ExportsHelper
+
   before_action :check_user_status
-	before_action :check_permission_to_wrangle
+	before_action :check_permission_to_wrangle, except: [:report_csv]
 
   def index
     @wranglers = Role.find_by(name: "tag_wrangler").users.alphabetical
@@ -48,6 +50,21 @@ class TagWranglersController < ApplicationController
     @counts[:UnsortedTag] = Rails.cache.fetch("/wrangler/counts/sidebar/UnsortedTag", race_condition_ttl: 10, expires_in: 1.hour) do
       UnsortedTag.count
     end
+  end
+
+  def report_csv
+    authorize :tag_wrangler, :report_csv?
+
+    wrangler = User.find_by!(login: params[:id])
+    wrangled_tags = Tag.where(last_wrangler: wrangler)
+    header = [%w[Name Last\ Updated Type Merger Fandoms Unwrangleable]]
+    results = wrangled_tags.map do |tag|
+      merger = tag.merger_id? ? tag.merger.name : ""
+      fandoms = tag.respond_to?(:fandoms) ? tag.fandoms.map(&:name).join(" + ") : ""
+      [tag.name, tag.updated_at, tag.type, merger, fandoms, tag.unwrangleable]
+    end
+    filename = "wrangled_tags_#{wrangler.login}_#{Time.now.utc.strftime('%Y-%m-%d-%H%M')}.csv"
+    send_csv_data(header + results, filename)
   end
 
   def create
--- a/app/views/tag_wranglers/show.html.erb
+++ b/app/views/tag_wranglers/show.html.erb
@@ -4,9 +4,14 @@
     <h2 class="heading"><%= ts("%{wrangler_login}'s Wrangling Page", :wrangler_login => @wrangler.login) %></h2>
     <p class="icon"><%= icon_display(@wrangler) %></p>
     <!--/descriptions-->
+    <!--subnav-->
+    <% if policy(:tag_wrangler).report_csv? %>
+      <ul class="navigation actions" role="navigation">
+        <li><%= link_to t(".tags_wrangled_csv"), report_csv_tag_wrangler_path %></li> <%# TODO #%>
+      </ul>
+    <% end %>
+    <!--/subnav-->
   </div>
-  <!--subnav-->
-  <!--/subnav-->
 
   <!--main content-->
   <!--TEST NOTE: these tds are titled to expose headings to Chromevox, which doesn't read scope
--- a/spec/controllers/tag_wranglers_controller_spec.rb
+++ b/spec/controllers/tag_wranglers_controller_spec.rb
@@ -6,14 +6,71 @@ describe TagWranglersController do
 
   let(:user) { create(:tag_wrangler) }
 
-  before { fake_login_known_user(user) }
-
   describe "#show" do
+    before { fake_login_known_user(user) }
+
     context "when the target user does not exist" do
       it "raises a 404 error" do
         expect do
           get :show, params: { id: -1 }
         end.to raise_exception(ActiveRecord::RecordNotFound)
+      end
+    end
+  end
+
+  describe "#report_csv" do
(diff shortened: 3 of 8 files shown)


app/controllers/tag_wranglers_controller.rb
```diff
@@ -1,6 +1,8 @@
 class TagWranglersController < ApplicationController
+  include ExportsHelper
+
   before_action :check_user_status
-	before_action :check_permission_to_wrangle
+  before_action :check_permission_to_wrangle, except: [:report_csv]
```
ceithir, line 5: What would be the use case in which a superadmin or tag wangler is allowed to access `report_csv` without validating at the same time `check_permission_to_wrangle`?

app/controllers/tag_wranglers_controller.rb
```diff
@@ -50,6 +52,21 @@ def show
...
   end
 
+  def report_csv
+    authorize :tag_wrangler, :report_csv?
+
+    wrangler = User.find_by!(login: params[:id])
+    wrangled_tags = Tag.where(last_wrangler: wrangler)
+    header = [%w[Name Last\ Updated Type Merger Fandoms Unwrangleable]]
+    results = wrangled_tags.map do |tag|
+      merger = tag.merger_id? ? tag.merger.name : ""
```
ceithir, line 62: `merger = tag.merger&.name || ""`?

app/views/tag_wranglers/show.html.erb
```diff
@@ -4,9 +4,14 @@
     <h2 class="heading"><%= ts("%{wrangler_login}'s Wrangling Page", :wrangler_login => @wrangler.login) %></h2>
     <p class="icon"><%= icon_display(@wrangler) %></p>
     <!--/descriptions-->
+    <!--subnav-->
+    <% if policy(:tag_wrangler).report_csv? %>
+      <ul class="navigation actions" role="navigation">
+        <li><%= link_to t(".tags_wrangled_csv"), report_csv_tag_wrangler_path %></li> <%# TODO #%>
```
ceithir, line 10: TODO?

Also, out of curiosity, by which Rails magic does `report_csv_tag_wrangler_path` manage to determine the id without having `@wrangler` as parameter?

brianjaustin: My bad, that's left over from my initial testing :smile: 

I'm actually not 100% sure about that, magic but I suspect it has something to do with [this bit in ActionDispatch::Routing::RouteSet](https://github.com/rails/rails/blob/main/actionpack/lib/action_dispatch/routing/route_set.rb#L324).

app/controllers/tag_wranglers_controller.rb
brianjaustin: Thanks!

app/controllers/tag_wranglers_controller.rb
brianjaustin: I don't think it would cause problems to do both, but it seems like unnecessary duplication to have two authorization checks

spec/controllers/tag_wranglers_controller_spec.rb
```diff
@@ -17,4 +17,61 @@
...
+    let!(:tag2) { create(:relationship, last_wrangler: user) }
+
+    context "when anonymous" do
+      it "prevents access" do
+        get :report_csv, params: { id: user.login }
+        it_redirects_to_simple(root_path)
+      end
+    end
+
+    context "when logged in as a superadmin" do
```
sarken, line 32: Can we test that admins with other roles (including no roles) can't access the report? Usually we just loop through the roles like in [the skins controller specs](https://github.com/otwcode/otwarchive/blob/dc8fac616e4841f5bc600327ea7cc1a75ee48d89/spec/controllers/skins_controller_spec.rb#L28-L48).

spec/controllers/tag_wranglers_controller_spec.rb
```diff
@@ -17,4 +17,61 @@
...
+      let!(:tag3) { create(:relationship, last_wrangler: user, merger: tag2) }
+
+      before { fake_login_admin(create(:superadmin)) }
+
+      it "renders the report" do
+        get :report_csv, params: { id: user.login }
+        result = CSV.parse(response.body.encode("utf-8")[1..], col_sep: "\t")
+        expect(result)
+          .to eq([%w[Name Last\ Updated Type Merger Fandoms Unwrangleable],
```
sarken, line 41: It might also be a good idea to separate the tests for "can download the CSV" from "the CSV contains the right info," so we can more easily tell what the failure is if we ever have one. 

I might be missing it, but I don't think there's a test to make sure the CSV doesn't include tags with a different last wrangler than the one whose CSV is being downloaded.

spec/controllers/tag_wranglers_controller_spec.rb
```diff
@@ -17,4 +17,61 @@
       end
     end
   end
+
+  describe "#report_csv" do
+    let!(:tag1) { create(:freeform, last_wrangler: user, unwrangleable: true) }
+    let!(:tag2) { create(:relationship, last_wrangler: user) }
+
+    context "when anonymous" do
```
sarken, line 25: fwiw, we'd usually say something more like "when logged out" here, since "anonymous" has a special meaning on the Archive.

app/controllers/tag_wranglers_controller.rb
sarken: Yeah, I think we can keep the `except: [:report_csv]` here for now, at least until we have a better idea what `check_permission_to_wrangle` will look like when we lock admin wrangling down more.

spec/controllers/tag_wranglers_controller_spec.rb
brianjaustin: Done!

spec/controllers/tag_wranglers_controller_spec.rb
brianjaustin: Good idea! I've done the separation and added a test for that case (there was not one to miss originally)

spec/controllers/tag_wranglers_controller_spec.rb
```diff
@@ -17,4 +17,121 @@
...
+      %w[superadmin tag_wrangling].each do |admin_role|
+        context "with role #{admin_role}" do
+          let(:admin) { create(:admin, roles: [admin_role]) }
+
+          it "allows access to the report" do
+            get :report_csv, params: { id: user.login }
+            expect(response).to have_http_status(:success)
+          end
+
+          it "only includes wranling activity for the specified user" do
```
sarken, line 69: Typo: should be "wrangling."

sarken (commented):
Two little things and I think we're set -- thank you!

spec/controllers/tag_wranglers_controller_spec.rb
brianjaustin: Hah, I misspelled that word so many times, and I guess this one slipped through. Thanks!

sarken (approved):

PR #4513, opened 2023-05-11
AO3-6470 Wrangling report performance improvements

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6470

## Purpose

Add improvements to mitigate the problem uncovered with wranglers who last wrangled a lot of tags

## Testing Instructions

See Jira

## References

Follow up from #4502 

## Credit

Brian Austin (they/he)

### DIFF
--- a/spec/controllers/tag_wranglers_controller_spec.rb
+++ b/spec/controllers/tag_wranglers_controller_spec.rb
@@ -133,6 +133,19 @@ describe TagWranglersController do
                       [tag.name, tag.updated_at.to_s, tag.type, "", "#{fandom1.name}, #{fandom2.name}", "false"]])
           end
 
+          it "only includes canonical fandoms" do
+            fandom1 = create(:canonical_fandom)
+            fandom2 = create(:fandom)
+            media = create(:media, last_wrangler: user)
+            expect(media.add_association(fandom1)).to be_truthy
+            expect(media.add_association(fandom2)).to be_truthy
+
+            get :report_csv, params: { id: user.login }
+            result = CSV.parse(response.body.encode("utf-8")[1..], col_sep: "\t")
+
+            expect(result[1][4]).to eq(fandom1.name)
+          end
+
           it "correctly reports a tag marked unwrangleable" do
             tag = create(:tag, last_wrangler: user, unwrangleable: true)
 
--- a/spec/controllers/tag_wranglers_controller_spec.rb
+++ b/spec/controllers/tag_wranglers_controller_spec.rb
@@ -82,14 +82,14 @@ describe TagWranglersController do
           it "limits the number of tags" do
             stub_const("ArchiveConfig", OpenStruct.new(ArchiveConfig))
             ArchiveConfig.WRANGLING_REPORT_LIMIT = 1
-            tag1 = create(:tag, last_wrangler: user)
             create(:tag, last_wrangler: user)
+            tag2 = create(:tag, last_wrangler: user)
 
             get :report_csv, params: { id: user.login }
             result = CSV.parse(response.body.encode("utf-8")[1..], col_sep: "\t")
 
             expect(result.length).to eq(2)
-            expect(result[1][0]).to eq(tag1.name)
+            expect(result[1][0]).to eq(tag2.name)
           end
 
           it "correctly reports mergers" do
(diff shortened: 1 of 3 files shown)


spec/controllers/tag_wranglers_controller_spec.rb
```diff
@@ -79,6 +79,19 @@
...
+            stub_const("ArchiveConfig", OpenStruct.new(ArchiveConfig))
+            ArchiveConfig.WRANGLING_REPORT_LIMIT = 1
+            tag1 = create(:tag, last_wrangler: user)
+            create(:tag, last_wrangler: user)
+
+            get :report_csv, params: { id: user.login }
+            result = CSV.parse(response.body.encode("utf-8")[1..], col_sep: "\t")
+
+            expect(result.length).to eq(2)
+            expect(result[1][0]).to eq(tag1.name)
```
brianjaustin, line 92: I was split on whether this should be the most recently wrangled (not indexed) or based on the tag's ID. With find_each, we get batching [but sacrifice the ability to order by something other than id](https://apidock.com/rails/ActiveRecord/Batches/find_each#:~:text=NOTE%3A%20It%E2%80%99s%20not%20possible%20to%20set%20the%20order.%20That%20is%20automatically%20set%20to%20ascending%20on%20the%20primary%20key%20(%E2%80%9Cid%20ASC%E2%80%9D)%20to%20make%20the%20batch%20ordering%20work.%20This%20also%20means%20that%20this%20method%20only%20works%20when%20the%20primary%20key%20is%20orderable%20(e.g.%20an%20integer%20or%20string).); I could change it to be descending though if that makes sense

redsummernight: [Order for batch queries](https://api.rubyonrails.org/v6.1/classes/ActiveRecord/Batches.html#method-i-find_in_batches) is supported in Rails 6.1+, so we should be OK.

As @sarken mentioned elsewhere, the output should be the x most recently wrangled tags.

tickinginstant: Just FYI, using `updated_at` for the "most recently wrangled" check will start breaking very badly once https://github.com/otwcode/otwarchive/pull/4230 is in. You'll just end up getting a list of the most popular tags. `updated_at` isn't an accurate measure of when the `last_wrangler` last touched the tag (and [never has been](https://otwarchive.atlassian.net/browse/AO3-3631)).

brianjaustin: @redsummernight do you mean the `:order` option? That's only for the primary key, so neither `:asc` or `:desc` would correlate with wrangling time frame

redsummernight: Ah, I missed that it's only on the primary key.

redsummernight (approved):

spec/controllers/tag_wranglers_controller_spec.rb
tickinginstant: `:desc` would, I think, have a better chance at showing *some* tags wrangled more recently (though not all), since it would show newer tags. But I've never been 100% clear on the purpose of this CSV — is it to check whether someone has been messing around with older tags that they shouldn't have been modifying, or to check whether someone has been doing their assigned duties of processing newer tags?

sarken:
The age of the tag isn't particularly relevant to whether a wrangler should be editing it (e.g., new guidelines might mean I need to decanonize a very old tag and syn it to a new canonical). Instead, this is generally used to check that a wrangler has been following guidelines -- so more like making sure you're putting stuff in the right fandoms and synning to things that make sense. So while it's preferable to be able to see newer decisions if we can't list them all, but as long as we have a good sample size, it will help wrangling supervisors determine if there's a problem or they need more data. We can probably cope with sorting by ID, since I'm most cases newer tags are more likely to have been wrangled more recently. Sent from my iPhoneOn May 12, 2023, at 8:44 AM, tickinginstant ***@***.***> wrote:﻿
@tickinginstant commented on this pull request.



In spec/controllers/tag_wranglers_controller_spec.rb:
> @@ -79,6 +79,19 @@
                       [tag1.name, tag1.updated_at.to_s, tag1.type, "", "", "false"]])
           end
 
+          it "limits the number of tags" do
+            stub_const("ArchiveConfig", OpenStruct.new(ArchiveConfig))
+            ArchiveConfig.WRANGLING_REPORT_LIMIT = 1
+            tag1 = create(:tag, last_wrangler: user)
+            create(:tag, last_wrangler: user)
+
+            get :report_csv, params: { id: user.login }
+            result = CSV.parse(response.body.encode("utf-8")[1..], col_sep: "\t")
+
+            expect(result.length).to eq(2)
+            expect(result[1][0]).to eq(tag1.name)

:desc would, I think, have a better chance at showing some tags wrangled more recently (though not all), since it would show newer tags. But I've never been 100% clear on the purpose of this CSV — is it to check whether someone has been messing around with older tags that they shouldn't have been modifying, or to check whether someone has been doing their assigned duties of processing newer tags?

tickinginstant (approved):
Thanks!

PR #4520, opened 2023-05-26
AO3-5728 Allow flexbox in skins

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5728

## Purpose

Allows flexbox in skins.

## Testing Instructions

Refer to Jira.

## Credit

Sarken, she/her


config/config.yml
```diff
@@ -323,48 +323,73 @@ BLACKLISTED_MULTIMEDIA_SRCS: []
...
+# subproperties or variants (ie, because "border" is in here, we allow through: 
+# border-right, border-bottom-left-radius, -moz-border-foo, etc)
+# It is slightly less secure so do NOT put any property in here if you aren't
+# sure all variations on it are okay!
+SUPPORTED_CSS_SHORTHAND_PROPERTIES: ["background", "border", "column", "cue",
+  "flex", "font", "layer-background", "layout-grid", "list-style", "margin",
+  "marker", "outline", "overflow", "padding", "page-break", "pause",
+  "scrollbar", "text", "transform", "transition"]
+
+SUPPORTED_CSS_PROPERTIES: ["-replace", "-use-link-source", "accelerator",
```
brianjaustin, line 346: Out of curiosity, any reason not to use the YAML block-style list?
```yaml
SUPPORTED_CSS_PROPERTIES:
- "-replace"
- "-use-link-source"
- "accelerator"
- etc
```
It might make reviewing future diffs a bit easier, since there's one element per line. I guess it could result in this file getting pretty large though... :thinking:

spec/models/skin_spec.rb
```diff
@@ -256,50 +279,50 @@
       create(:work_skin, :private, author: skin_owner, title: "Private Skin 2")
     end
 
-    context 'no user argument given' do
-      context 'User.current_user is nil' do
-        it 'returns approved skins' do
+    context "no user argument given" do
```
brianjaustin, line 282: Thanks! :heart:

brianjaustin (approved):
One minor question, otherwise looks good to me!

config/config.yml
sarken: No reason that I'm aware of! I like the idea of easier-to-read diffs, so I can change it.

sarken: Actually, let me check I won't be making more work for Systems first...

cjrecordvt:
I'm nosing around for other reasons, but is there supposed to be a comma after Pause on l. 357?

config/config.yml
zz9pzza: No I think it would be fine :)

PR #4521, opened 2023-05-28
AO3-6535 Update linters and cops

brianjaustin:
# Pull Request Checklist

* [ ] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)? (n/a)

## Issue

https://otwarchive.atlassian.net/browse/AO3-6535 (Please fill in issue number and remove this comment.)

## Purpose

* Upgrade rubocop and related gems to the latest versions (except for erb_lint, that was upgraded to the earliest compatible with newer rubocop; erb_lint was struggling a bit with errors even before, so I didn't want to mess with it too much)
* Add todo files to track/ignore existing errors
* Manually enable and disable a few cops, both pending and introduced since the last version

## Testing Instructions

On Jira ticket

## References

## Credit

Brian Austin (they/he)


Gemfile
```diff
@@ -154,10 +154,10 @@ group :development do
...
 
 group :linters do
-  gem "erb_lint", "0.0.29"
-  gem "rubocop", "0.83.0"
-  gem "rubocop-rails", "2.6.0"
-  gem "rubocop-rspec", "1.41.0"
+  gem "erb_lint", "0.4.0"
+  gem "rubocop", "1.22.1"
+  gem "rubocop-rails", "2.12.4"
+  gem "rubocop-rspec", "2.6.0"
```
brianjaustin, line 160: There are newer versions available of these, but Hound only supports these (hence the error messages earlier)

sarken:
Are the todo files something we want? I might be misunderstanding, but it seems like they'll prevent certain cops from running in the specified files -- so for example, if I had touched the line with the `not` in `app/controllers/pseuds_controller.rb` in #4542, or worse, added a new `not`, Hound wouldn't bug me to fix it. 

If I'm wrong and Hound would still bug me about those, and the todo files just prevent it from barking if you _didn't_ touch the line or add a new violation, then they're probably fine!

tickinginstant:
I think the TODO files would make sense if you wanted to start a project to resolve the various rubocop violations in a more systematic way. With the TODO files, you could have regular PRs to just re-enable one or more rules, correct or autocorrect all violations, and then continue. That would also let you pivot away from using Hound (which doesn't appear to be well-supported anymore, since the last time they added a new rubocop version was 2021), and instead use some basic GH actions steps to get style comments.

However, if you want to continue with the current plan of just having people fix the violations that they encounter in lines of code that they change, the TODO files aren't a good plan.

Personally, I think it would be good to start taking a more focused approach to resolving rubocop violations, since:
1) I think PRs that change the logic and simultaneously fix rubocop violations are harder to review, since it makes it more difficult to see what the relevant changes are.
2) Newbies sometimes try to learn the desired style from the existing code (e.g. studying or copying some similar code that works), and may have difficulty dealing with Hound's comments.
3) Fixing all 10,000+ existing violations with the incremental approach might literally take forever.

But a project like that would require more work.

brianjaustin:
Thanks for the detailed comment @tickinginstant! Another thing I'd add (in favor of the larger effort to fix) is that several of the longest ignores in the TODO file should be automated fixes. If we're OK expanding the scope of this PR, maybe we could even start now. It would still be some effort though, so a couple compromises I can think of:
1. Comment out https://github.com/otwcode/otwarchive/pull/4521/files#diff-4f894049af3375c2bd4e608f546f8d4a0eed95464efcdea850993200db9fef5cR1 so that we still have documentation of where want to fix things and the ability to easily ignore things locally
2. Change the auto-generated TODO output to only ignore violations of new cops (technically we could just ignore Hound, but that feels messy to me)

sarken:
Let's go with compromise #1. While a serious effort to fix style violations would be great, I don't think prioritizing that is in the cards right now. Maybe we can sneak a few things in here and there, though.

tickinginstant (approved):

sarken (approved):

PR #4528, opened 2023-06-10
AO3-6212 AO3-6573 Add indices to tag_set_associations, and add the departure gem to make future migrations easier.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6212
https://otwarchive.atlassian.net/browse/AO3-6573

## Purpose

- Adds the departure gem, so that we can simplify the code for migrations on large tables.
- Moves the PTOSC arguments into the config file, so that it's easier to customize them for different systems.
- Adds three indices to the `tag_set_associations` table:
  - One on the tag ID.
  - One on the parent tag ID.
  - One on the owned tag set ID, the parent tag ID, and the tag ID. This one is unique.

A couple notes:
- I removed the `--no-drop-old-table` flag, since that may not be necessary now that the databases are on MariaDB. Because the flags are now controlled through the config files (and can be overridden by specifying the `PERCONA_ARGS` environment variable), it shouldn't require a PR to add back the flag if MariaDB turns out to have similar problems to MySQL.
- I think this is a good migration to test things with, including departure. It's on a relatively large table, so it probably should use PTOSC, but if something goes catastrophically wrong, it should only affect a relatively small number of users.

## Testing Instructions

As on Jira, but since this introduces a new tool, there'll need to be some systems changes prior to the deploys on staging and production.


sarken:
(james indicated he wanted to look at this when he had time, so I'm tagging him in as an assignee.)

zz9pzza (approved):

PR #4529, opened 2023-06-13
AO3-6539 Easier admin access to user creations

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6539

## Purpose

* Exposes dashboard on user admin page
* Adds new user creations page for PAC to review a user's works and comments at a glance
* Updates comment link text and destination in abbreviated comment view, which is used on new user creations page and existing spammer creation deletion page
* Adds an anchor to the pagination on the work/comment list boxes so you don't have to scroll so much
* Updates the user header navigation (which is used on user and pseud dashboards, plus profiles) so admins can easily access the User Administration and Invitations pages

I was torn on whether to make this part of the user policy or the user creation policy. I went with the user policy because user creation feels more "things you can do to a single creation," but I can switch.

The shared examples for the specs are because I originally updated all the tests to more thoroughly test the roles, but then I was like, no, no one needs that in this pull request. So I just left the shared examples and I'll do the test updates in a separate issue/PR.

## Testing Instructions

Refer to Jira.

### DIFF
--- a/spec/controllers/admin/admin_users_controller_spec.rb
+++ b/spec/controllers/admin/admin_users_controller_spec.rb
@@ -10,7 +10,7 @@ describe Admin::AdminUsersController do
   let(:user) { create(:user) }
 
   manage_roles = %w[superadmin legal open_doors policy_and_abuse support tag_wrangling].freeze
-  search_roles = %w[superadmin legal open_doors policy_and_abuse support tag_wrangling ].freeze
+  search_roles = %w[superadmin legal open_doors policy_and_abuse support tag_wrangling].freeze
 
   shared_examples "an action unauthorized admins can't access" do |authorized_roles:|
     before { fake_login_admin(admin) }
@@ -240,13 +240,7 @@ describe Admin::AdminUsersController do
   end
 
   describe "POST #update_next_of_kin" do
-    subject { -> {
-        post :update_next_of_kin, params: {
-          user_login: user.login,
-          next_of_kin_name: kin.login,
-          next_of_kin_email: kin.email
-        }
-    } }
+    subject { -> { post :update_next_of_kin, params: { user_login: user.login, next_of_kin_name: kin.login, next_of_kin_email: kin.email } } }
     let(:kin) { create(:user) }
     authorized_roles = %w[superadmin policy_and_abuse support].freeze
 


app/controllers/admin/admin_users_controller.rb
```diff
@@ -1,6 +1,31 @@
...
+
+  def load_user
+    @user = User.find_by!(login: params[:id])
+  end
+
+  def user_is_banned
+    return if @user&.banned?
+
+    flash[:error] = ts("That user is not banned!")
+    redirect_to admin_users_path and return
```
Bilka2, line 21: The "and return" can be removed here.

app/views/users/_header.html.erb
```diff
@@ -7,7 +7,9 @@
 <!--/descriptions-->
 
   <!--subnav-->
-  <%= render "users/header_navigation" %>
+  <% if logged_in? || policy(User).can_manage_users? %>
```
Bilka2, line 10: I think this should check `logged_in_as_admin?` as well as the policy.

sarken: Is there a case where the lack of explicit admin check here is causing an issue, like the navigation not showing when it should?

Bilka2: Not as far as I'm aware. The motivation is mostly that admin_helper does it that way and that this code reads a bit confusing without it because the obvious not `logged_in?` case is a visitor, not an admin.

sarken: I don't think it's worth making it more verbose if it's not necessary, then. Neither the comments helper nor the various views include that additional check, so I have a feeling the oddball check in admin helper might be either a relic from the very early days when we didn't include the admin check in `user_has_roles?(roles)` in the application policy, or maybe just a little leftover copypasta from previous methods in that file, which took admin as an argument, e.g.,

```
def admin_can_destroy_creations?(current_admin)
    return unless logged_in_as_admin?

    UserCreationPolicy.can_destroy_creations?(current_admin)
 end
```

app/helpers/comments_helper.rb
```diff
@@ -32,6 +32,24 @@ def link_to_comment_ultimate_parent(comment)
...
+  def comment_link_with_commentable_name(comment)
+    ultimate_parent = comment.ultimate_parent
+    commentable_name = ultimate_parent&.commentable_name
+    text = case ultimate_parent.class.to_s
+           when "Work"
+             t("comments_helper.comment_link_with_commentable_name.on_work_html", title: commentable_name)
+           when "AdminPost"
+             t("comments_helper.comment_link_with_commentable_name.on_admin_post_html", title: commentable_name)
+           else
+             if ultimate_parent.is_a?(Tag)
```
weeklies, line 44: Could we combine these lines to `elsif ultimate_parent.is_a?(Tag)`?

app/views/admin/admin_users/_user_creations_summary.html.erb
```diff
@@ -0,0 +1,18 @@
+<% unless @works.size.zero? %>
```
weeklies, line 1: Is there a reason why `size.zero?` is used over `empty? `?

app/helpers/comments_helper.rb
sarken: It won't work if this is a `case` statement. I can make this a plain if/elsif/else if we don't mind it being different from the `link_to_comment_ultimate_parent` method above.

weeklies: I didn't know that, I think this way works just fine.

app/views/admin/admin_users/_user_creations_summary.html.erb
sarken: Nope, that's just what it was originally, so I didn't change it.

weeklies (approved):
Tested on my environment and the changes looks good to me.

app/views/admin/admin_users/_user_creations_summary.html.erb
```diff
@@ -0,0 +1,18 @@
+<% unless @works.empty? %>
+  <div class="work listbox group" id="works-summary">
+    <h3 class="heading"><%= t(".works", count: works.size) %></h3>
```
Bilka2, line 3: This `works.size` and the `comments.size` are always the pagination size (20 on prod, 25 in dev) when there is more than one page of works/comments. That is the case even when on a page with less than the page size works/comments.

It looks like using `works.count` would make it show the total count instead, which would be less confusing. Ideally I think it would be nice to list something similar to work search results, so "1 - 20 of 34 Works" etc, but I don't know if that is possible

Bilka2 (approved):
Agreed with weeklies, looks good. Just one minor issue related to the pagination of the creations. Also, there is a merge conflict with master.

sarken:
The merge conflicts ran right into those test improvements I mentioned in the original PR comment, so the changes to the spec file are a bit more involved than planned. Feel free to just boot this to some other release if you don't want to review it!

Bilka2:
The spec merge/improvements look good. The diff is a lot less intimidating when hiding whitespace changes.

So this is ready to merge!

PR #4530, opened 2023-06-15
AO3-6513 Allow archivists to directly add to collections

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6513

## Purpose

Allow Open Doors archivists to add works to collections, regardless of whether or not creators allow regular users to send collection invitations. Works added by archivists are auto-approved when they maintain the target collection.

## Testing Instructions

Refer to Jira

## References

## Credit

Brian Austin (they/he)


spec/controllers/collection_items_controller_spec.rb
```diff
@@ -321,6 +321,43 @@
...
+      end
+
+      before do
+        fake_login_known_user(archivist)
+      end
+
+      context "when the item's creator does not allow collection invitations" do
+        it "adds the item anyway" do
+          post :create, params: params
+          it_redirects_to_simple(work)
```
sarken, line 348: Does the archivist not get any sort of confirmation message here that we could check? If not, we might need an issue for that.

spec/controllers/collection_items_controller_spec.rb
```diff
@@ -321,6 +321,43 @@
         end
       end
     end
+
+    context "as an archivist" do
+      let(:archivist) { create(:archivist) }
+      let(:work) { create(:work) }
+
+      let(:collection) do
+        participant = create(:collection_participant, pseud: archivist.default_pseud)
```
sarken, line 330: I was like, wait, _what_, but it turns out the default role in the `collection_participant` factory is owner. 🤷 

https://github.com/otwcode/otwarchive/blob/f7cd4fc80c1482e12d216422759b3eee670b45fb/factories/collections.rb#L12-L15

That seems not particularly ideal (imo, we should rename that factory `collection_owner`), but it's out of scope for this. Maybe we could call the variable `owner` here, though, for a little clarity?

sarken (changes requested):
Open Doors is going to be very happy -- thank you!

spec/controllers/collection_items_controller_spec.rb
brianjaustin: They do get a flash message, and I've added a check for that. I went through a few iterations of the implementation, so maybe that got lost

app/controllers/collection_items_controller.rb
```diff
@@ -103,7 +103,8 @@ def create
         errors << ts("%{collection_title}, because you don't own this item and the item is anonymous.", collection_title: collection.title)
       # add the work to a collection, and try to save it
       elsif @item.add_to_collection(collection) && @item.save(validate: false)
-        # approved_by_user? and approved_by_collection? are both true
+        # approved_by_user? and approved_by_collection? are both true.
+        # This is now only true for imports by archivists.
```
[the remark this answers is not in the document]
brianjaustin: You are correct! I've updated the comment to reflect that as well

sarken (approved):

PR #4218, opened 2022-04-27
AO3-6156 Skip logging admin activity for empty edits

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6156

## Purpose

Also add more specific error messages for failed ticket validations.

## Testing Instructions

See issue.

PR #4542, opened 2023-06-21
AO3-4595 Allow certain admins to update pseud descriptions and icons

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4595

## Purpose

* Allows certain admins to update pseud descriptions and icons using [permitted_attributes for strong parameters](https://github.com/varvet/pundit#strong-parameters)
* Updates tests that were failing because they were trying to create users/pseuds while logged in as admin (the favorite tag tests in [tag_wrangling_spec.rb](https://github.com/otwcode/otwarchive/compare/master...sarken:otwarchive:AO3-4595?expand=1#diff-10d8c71d085ded0e36c8687c9a6856e7eb5c78224c0ba9a484420a506308134c) and anything that used a cucumber step relying on [features/support/user.rb](https://github.com/otwcode/otwarchive/compare/master...sarken:otwarchive:AO3-4595?expand=1#diff-5b8c8bfbf3d40c20ea5226972cf82bb952aefb21263d29d5b0a55a5c50289908))
* Tweaks the admin activity summary so it only uses `raw sanitize_field(activity, :summary)` when it includes a link (introduced in #4198), which should prevent other summaries from getting semi-eaten (lines [59-61 of admin_works.feature](https://github.com/otwcode/otwarchive/compare/master...sarken:otwarchive:AO3-4595?expand=1#diff-8228637bd145b7957bd3fb8182a276fc7fe9c5d933e81cecb833cfc4686d46b3))
* i18n. So much i18n.

## Testing Instructions

## References

## Credit

they/them pronouns.*


sarken:
Oh, hang on, there were more tests.

app/views/pseuds/_pseuds_form.html.erb
```diff
@@ -2,13 +2,13 @@
...
   <dd>
-    <% if @pseud.name && @user.login == @pseud.name %>
+    <% if @pseud&.name == @user.login %>
       <p class="important informational"><%= @pseud.name %></p>
       <p class="footnote"><%= t(".change_matching_pseud_html", change_username_link: link_to(t(".change_username"), change_username_user_path(@user))) %></p>
     <% else %>
       <%= f.text_field :name, class: "observe_textlength", disabled: logged_in_as_admin? %>
     <% end %>
-    <%= generate_countdown_html("pseud_name", Pseud::NAME_LENGTH_MAX) if @pseud.name && @user.login != @pseud.name %>
+    <%= generate_countdown_html("pseud_name", Pseud::NAME_LENGTH_MAX) if @pseud&.name == @user.login %>
```
neuroalien, line 11: This conditional seems to do the opposite of what it was doing before.

Could we move the countdown before the end, on the "else" branch, so it wouldn't need its own conditional?

spec/controllers/pseuds_controller_spec.rb
```diff
@@ -6,29 +6,230 @@
...
+        put :update, params: params
+        it_redirects_to_with_notice(user_pseud_path(user, pseud), "Pseud was successfully updated.")
+      end
 
-        post :destroy, params: { user_id: user, id: matching_pseud }
-        it_redirects_to_with_error(user_pseuds_path(user), "You cannot delete the pseud matching your user name, sorry!")
+      it "creates admin activity" do
+        expect do
+          put :update, params: params
+        end.to change { AdminActivity.count }.by(1)
```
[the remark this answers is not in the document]
sarken: Gross tbh

app/controllers/pseuds_controller.rb
```diff
@@ -104,8 +106,14 @@ def create
   # PUT /pseuds/1.xml
   def update
     @pseud = @user.pseuds.find_by(name: params[:id])
+    authorize @pseud if logged_in_as_admin?
     default = @user.default_pseud
     if @pseud.update(pseud_params)
+      if logged_in_as_admin? && @pseud.ticket_url.present?
```
neuroalien, line 112: Do we not log the admin's activity if there's no ticket? (I'm out of the loop and have a migraine, so I might be missing a lot of context.)

features/support/user.rb
```diff
@@ -2,6 +2,11 @@ module UserHelpers
   def find_or_create_new_user(login, password, activate: true)
     user = User.find_by(login: login)
     if user.blank?
+      # If we're logged in as an admin when we try to use a step that invokes
+      # this method to create a user, user creation will fail because only some
+      # admins can edit certain fields on pseuds. Setting current_user to nil
+      # will bypass that.
+      User.current_user = nil
```
neuroalien, line 9: Is this not going to clobber the existing logged-in-as-admin situation for the scenario or step that invoked the user creation?

(Alternatively, do we need to add these guardrails to all user-creating support methods? If they're not used now, might we encounter this in the future?)

app/controllers/pseuds_controller.rb
sarken: We had to introduce the `ticket_url.present?` check for profiles (#4218) due to [an edge case](https://otwarchive.atlassian.net/browse/AO3-6156?focusedCommentId=358635) where admin activity was logged for empty edits. The same edge case happens here -- without `@pseud.ticket_url.present?`, `RAILS_ENV=test bundle exec cucumber features/other_a/pseuds.feature:177` will fail.

app/models/admin_activity.rb
```diff
@@ -16,6 +16,10 @@ def self.log_action(admin, target, options={})
   end
 
   def target_name
-    "#{target_type} #{target_id}"
+    if target.is_a?(Pseud)
+      "Pseud #{target&.name} (#{target&.user&.login})"
```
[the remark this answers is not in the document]
sarken: Hm, I guess not, since we're checking the actual class and not just the string. I'll take it out!

features/support/user.rb
sarken: I was worried about that, too, but it doesn't seem to. For example, these tests still pass:

https://github.com/otwcode/otwarchive/blob/d99747574ed348866cdeed6cda507b7636612582/features/other_a/invite_request.feature#L125-L146

I'll add it to `ensure_user` as well, since if I swap the order of the steps in this scenario (just to pick a random scenario), I get the same error.

https://github.com/otwcode/otwarchive/blob/d99747574ed348866cdeed6cda507b7636612582/features/admins/users/admin_fnok.feature#L35-L41

app/controllers/pseuds_controller.rb
```diff
@@ -74,12 +75,13 @@ def new
   # GET /pseuds/1/edit
   def edit
     @pseud = @user.pseuds.find_by(name: params[:id])
+    authorize @pseud if logged_in_as_admin?
```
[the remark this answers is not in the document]
sarken: Right. This doesn't mean "authorize any admin to access this page." It means "if the person trying to access the page is an admin, check the pseud policy and decide if they're authorized to access this page." 

Here's the [general Pundit documentation](https://github.com/varvet/pundit#policies) and [the docs for the authorize method](https://www.rubydoc.info/gems/pundit/Pundit.authorize) for explanations and general examples. 

For a more us-specific example, another place we do this is when deleting comments: https://github.com/otwcode/otwarchive/blob/49581ccbc42a05805fdb697f50aad1fbfb4af9c0/app/controllers/comments_controller.rb#L395-L417

app/controllers/pseuds_controller.rb
sarken: We can change the behavior of both in the future if PAC wants it logged. For now, let's keep them consistent and log nothing.

zz9pzza (approved):

PR #4552, opened 2023-06-21
AO3-6540 Allow banned users to delete works, chapters and remove themselves as co-creators

EchoEkhi:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6540

## Purpose

Allow suspended and banned users to delete works, chapters and remove themselves as co-creators.

## Testing Instructions

See JIRA ticket.

## Credit

EchoEkhi


brianjaustin (changes requested):
Some tests for `works_controller` could be useful as well; they could be split between
* https://github.com/otwcode/otwarchive/blob/master/spec/controllers/works/default_rails_actions_spec.rb
* https://github.com/otwcode/otwarchive/blob/master/spec/controllers/works/multiple_actions_spec.rb

EchoEkhi:
I'll add the multiple_actions test later

sarken:
I'm terribly sorry about this, but the Policy & Abuse team has asked for a change to the issue: they'd like only permanently banned users to be able to perform these actions. Users who are temporarily suspended should still not be able to perform these actions.

EchoEkhi:
Um so hound doesn't like my workaround in my tests. What should I do? Just ignore it?

brianjaustin:
@EchoEkhi one fix would be to update https://github.com/otwcode/otwarchive/pull/4552/files#diff-61b0876b1387d92d07e44fe491c4eee7b960e4acc8667a056b2389279273dc8eR220 (and the one below it for `multiple_work2`) to use the non-lazy `let!`. Since works will have to exist anyways for all these tests, I don't think it's a significant performance concern.

For example:
```rb
let!(:multiple_work1) do
      create(:work,
             authors: [multiple_works_user.default_pseud],
             title: "Work 1",
             comment_permissions: :disable_anon,
             moderated_commenting_enabled: true)
end
```
(includes other fixes because Hound will inevitably complain about the spacing and curly braces...)

spec/controllers/orphans_controller_spec.rb
```diff
@@ -22,7 +65,7 @@
 
     context "when logged in as the owner" do
       before { fake_login_known_user(user.reload) }
-
+
```
brianjaustin, line 68: Nitpick: extra spaces?

spec/controllers/orphans_controller_spec.rb
```diff
@@ -49,6 +92,69 @@
       end
     end
 
+    context "when logged in as a suspended user" do
+      before { fake_login_known_user(suspended_user.reload) }
+      
+      it "shows the form for orphaning a work" do
```
brianjaustin, line 98: Should the tests in this `context` say something like "does not show the form for ..."? (similar question for the delete action below too)

spec/controllers/orphans_controller_spec.rb
```diff
@@ -8,12 +8,55 @@
...
 
   let!(:user) { create(:user) }
-
   let!(:pseud) { create(:pseud, user: user) }
   let!(:work) { create(:work, authors: [pseud]) }
   let(:second_work) { create(:work, authors: user.pseuds) }
   let(:series) { create(:series, works: [work], authors: [pseud]) }
 
+  let!(:suspended_user) { create(:user, suspended: true, suspended_until: 1.week.since) }
+  let!(:suspended_pseud) { create(:pseud, user: suspended_user) }
```
brianjaustin, line 17: Could this be the default pseud?

brianjaustin (commented):
Thanks for all those tests @EchoEkhi ! I've left some comments/questions on the new things. It also looks like your PR fell into the issues with Cucumber, but pulling in #4588 should make the tests happy again

spec/controllers/orphans_controller_spec.rb
EchoEkhi: Well the original tests have `let!(:pseud) { create(:pseud, user: user) }`, so I just replicated that for the suspended version. I don't know if there are any complications with not being able to orphan a default pseud or something like that, so I think that sort of refactoring is probably better left to another separate task.

spec/controllers/orphans_controller_spec.rb
```diff
@@ -8,12 +8,55 @@
...
   let!(:pseud) { create(:pseud, user: user) }
   let!(:work) { create(:work, authors: [pseud]) }
   let(:second_work) { create(:work, authors: user.pseuds) }
   let(:series) { create(:series, works: [work], authors: [pseud]) }
 
+  let!(:suspended_user) { create(:user, suspended: true, suspended_until: 1.week.from_now) }
+  let!(:suspended_pseud) { create(:pseud, user: suspended_user) }
+  let!(:suspended_second_pseud) { create(:pseud, user: suspended_user) }
+  let!(:suspended_users_work) do
+    suspended_user.update(suspended: false, suspended_until: nil)
```
brianjaustin, line 20: It looks like there are a few more `update`s in here that could be `update!`

app/controllers/application_controller.rb
```diff
@@ -419,14 +419,22 @@ def use_caching?
...
       if current_user.suspended?
-        flash[:error] = t('suspension_notice', default: "Your account has been suspended until %{suspended_until}. You may not add or edit content until your suspension has been resolved. Please <a href=\"#{new_abuse_report_path}\">contact Abuse</a> for more information.", suspended_until: localize(current_user.suspended_until)).html_safe
+        flash[:error] = t("suspension_notice_html", default: "Your account has been suspended until %{suspended_until}. You may not add or edit content until your suspension has been resolved. Please <a href=\"#{new_abuse_report_path}\">contact Abuse</a> for more information.", suspended_until: localize(current_user.suspended_until))
       else
-        flash[:error] = t('ban_notice', default: "Your account has been banned. You are not permitted to add or edit archive content. Please <a href=\"#{new_abuse_report_path}\">contact Abuse</a> for more information.").html_safe
+        flash[:error] = t("ban_notice_html", default: "Your account has been banned. You are not permitted to add or edit archive content. Please <a href=\"#{new_abuse_report_path}\">contact Abuse</a> for more information.").html_safe
```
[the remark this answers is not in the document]
tickinginstant: Using `_html` in keys to avoid calls to `html_safe` is good, but unfortunately in Rails 6.1 it only works in views, not controllers. Rails 7.0 [introduces it for controllers as well](https://github.com/rails/rails/commit/a5247bb90895f843c45fb095a4072e8ef30eaa4e), but it'll probably be at least 6 months before that upgrade happens (judging by when the Rails upgrades happened in previous years).

brianjaustin (approved):

PR #4555, opened 2023-06-24
AO3-2722 Allow parent skins to be added in wizard

weeklies:
+ Fix bug where error 500 is thrown when non-existent parent skin is added

# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-2722
## Purpose

Add a Parent Skin section to the Skin Wizard, as shown in the default editor.

Also, I've fixed a small bug where Error 500 is thrown for non-existent parent skins. I ran into this during testing and didn't find an open Jira issue, so I hope including it in here is OK.

## Testing Instructions

See Jira issue.

## References

There are open issues regarding Error 500s when editing skins, but this PR doesn't resolve them.

https://otwarchive.atlassian.net/browse/AO3-4820
https://otwarchive.atlassian.net/browse/AO3-3810

## Credit

weeklies (she/her)


sarken:
>Also, I've fixed a small bug where Error 500 is thrown for non-existent parent skins. I ran into this during testing and didn't find an open Jira issue, so I hope including it in here is OK.

That should be fine! Do you have step-by-step instructions for reproducing it that we can add to the issue, just to make sure it gets tested?

weeklies:
Error 500 when adding non-existent parent skin 
Instructions: 
- Create a site skin, with a unique name 
- Under Advanced > Parent Skins, add a keysmash site skin (e. g. "keysmashwashere11")
- Submit the form.
- Please test for default and Wizard forms.

weeklies:
If it is more convenient, I can leave a comment on Jira.

sarken:
Thank you! And you're fine, I copied it right over and into the issue body to make sure it won't be missed.

app/views/skins/_wizard_form.html.erb
```diff
@@ -108,4 +108,28 @@
   <%= hidden_field_tag 'wizard', true %>
 </fieldset>
 
+<fieldset>
```
sarken, line 111: I think the indenting in this new section got a little wonky. For example, it looks like `<legend>` is indented four spaces under `<fieldset>` instead of two, and the closing `</fieldset>` tag doesn't seem to line up with the opening one. 

If you could also update it to use double quotes around strings, that would be fantastic.

app/views/skins/_wizard_form.html.erb
```diff
@@ -108,4 +108,28 @@
...
+      <% if f.object.skin_parents.count == 0 %>
+        <li class="last_id" style="display:none;">0</li>
+      <% end %>
+      <li class="actions" role="button">
+        <%= link_to_add_section(ts('Add parent skin'), f, :skin_parents, "skin_parent_fields") %>
+      </li>
+    </ul>
+  </fieldset>
+
 <%= submit_fieldset(f) %>
```
sarken, line 135: Could you add the missing new line at the end of the file, please?

sarken (changes requested):
I think we need some tests for these changes.

Can you add a test for the bug you've fixed in the skin parent model to [spec/models/skin_parent_spec.rb](https://github.com/otwcode/otwarchive/blob/master/spec/models/skin_parent_spec.rb)?

It looks like we don't have any feature tests for adding parent skins via the interface (everything I saw at a glance looked like it relied on factories) in features/other_b/skin.feature, which I _suspect_ might be because it's excessively difficult due to the JavaScript involved. However, even if it's not feasible to test the entire process of adding a parent skin, I think it would be a good idea to at least add a test to [features/other_b/skin_wizard.feature](https://github.com/otwcode/otwarchive/blob/master/features/other_b/skin_wizard.feature) to confirm the parent skin section is there.

features/other_b/skin_wizard.feature
```diff
@@ -12,6 +12,32 @@ Feature: Skin wizard
   Then I should see "Create New Skin"
     And I should see "CSS"
 
+  @javascript
+  Scenario: User can add a parent skin to the wizard
```
sarken, line 16: For clarity, maybe "User can add a parent skin using the wizard" for this and "User can add a parent skin using the Custom CSS form" for the other test?

features/other_b/skin_wizard.feature
```diff
@@ -12,6 +12,32 @@ Feature: Skin wizard
...
+  Given I am logged in
+    And I create the skin "Dad"
+  When I go to the new skin page
+    And I follow "Use Wizard"
+  Then I should see "Site Skin Wizard"
+    And I should see "Parent Skins"
+  When I fill in "Title" with "Child"
+    And I follow "Add parent skin"
+    And I wait 1 second
```
sarken, line 25: I think we prefer to use `And it is currently 1 second from now` when we need to set a delay.

features/step_definitions/skin_steps.rb
```diff
@@ -182,6 +182,10 @@
 
 ### THEN
 
+Then /^I should see a parent skin text field$/ do
```
sarken, line 185: Can we update the syntax here to `Then "I should see a parent skin text field" do`?

spec/models/skin_parent_spec.rb
```diff
@@ -3,61 +3,63 @@
...
 
   describe "save" do
-
-    before(:each) do
-      @child_skin = Skin.new(title: "Child", css: "body {background: #fff;}")
-      @parent_skin = Skin.new(title: "Parent", css: "body {color: #000;}")
-      @child_skin.save; @parent_skin.save
-      @skin_parent = SkinParent.new(child_skin: @child_skin, parent_skin: @parent_skin, position: 1)
-    end
+    let(:child_skin) { Skin.new(title: "Child", css: "body {background: #fff;}") }
```
sarken, line 6: Yay, thanks for changing these to `let`! 

Would using the factory and doing `build(:skin, title: "Child", css: "body {background: #fff;}")` work here?

app/models/skin_parent.rb
```diff
@@ -10,9 +10,9 @@ class SkinParent < ApplicationRecord
 
   validate :no_site_parent
   def no_site_parent
-    if parent_skin.get_role == "site" && !%w(override site).include?(child_skin.get_role)
-      errors.add(:base, ts("^You can't use %{title} as a parent unless replacing the default archive skin.", title: parent_skin.title))
-    end
+    return unless parent_skin&.get_role == "site" && !%w[override site].include?(child_skin.get_role)
+
+    errors.add(:base, ts("^You can't use %{title} as a parent unless replacing the default archive skin.", title: parent_skin.title))
```
sarken, line 15: I can't remember, but maybe someone else can: do we still need `^` here?

features/other_b/skin_wizard.feature
```diff
@@ -12,6 +12,32 @@ Feature: Skin wizard
...
+  Then I should see a parent skin text field
+
+  @javascript
```
sarken, line 28: Since this isn't testing the wizard, it should probably be in features/other_b/skin.feature instead.

features/other_b/skin_wizard.feature
```diff
@@ -12,6 +12,32 @@ Feature: Skin wizard
...
+  Given I am logged in
+    And I create the skin "Dad"
+  When I go to the new skin page
+  Then I should see "Advanced"
+  When I follow "Show ↓"
+  Then I should see "Parent Skins"
+  When I fill in "Title" with "Child"
+    And I follow "Add parent skin"
+    And I wait 1 second
+  Then I should see a parent skin text field
```
sarken, line 39: I wanted to see if we could test the whole process (i.e., filling in a parent skin name and saving), and here's what I came up with: 

```
  @javascript
  Scenario: User can add a parent skin to the form
  Given I am logged in
    And I create the skin "Dad"
  When I go to the new skin page
    And I fill in "Title" with "Child" # I moved this just to make it a little clearer that it's not related to adding the parent skin.
  Then I should see "Advanced"
  When I follow "Show ↓"
  Then I should see "Parent Skins"
  When I follow "Add parent skin"
    And it is currently 1 second from now
  Then I should see a parent skin text field
  When I enter "Dad" in the "skin_skin_parents_attributes_1_parent_skin_title_autocomplete" autocomplete field
    And I press "Submit"
  Then I should see "Parent Skins"
    And I should see Dad"
```

This approach should work for the wizard, too; I just used the regular skin form because I hadn't pulled the new code. 

What I did have to do to make this work is change the redirect to use `skin_path(@skin)` on line 80, like @tickinginstant  mentioned to you when you were working on series positions:

https://github.com/otwcode/otwarchive/blob/c7b3a865a605ae8419f2a88622574c179fe48b2b/app/controllers/skins_controller.rb#L62-L80

sarken (changes requested):

app/models/skin_parent.rb
weeklies: I've tried removing it and everything looks OK —will remove it.
image

sarken: I think we do need it, judging from that screen shot -- if we take it out, "Skin parent base" is added before the desired text. I feel like there's another way to keep that from happening but I'm blanking on what it is.

tickinginstant: You should be able to suppress that prefix using the locales file by setting `format: "%{message}"` in the appropriate `activerecord.errors.models` translation scope. (Specifically, I think you need to set it for `skin_parent` and `skin/skin_parents` — the related work/parent work relationships stuff in `config/locales/models/en.yml` is probably a good model to follow in this case.)

It applies to all error messages, so if you want to suppress the prefix that way, it'd be a good idea to go through all of the validations in the `SkinParent` class to make sure that the error messages are still properly formatted. Alternatively, you may be able to apply it to just `base` messages.

weeklies: I've tried adding this to `en.activerecord.errors.models` and using  `errors.add(:base, :site_parent, title: parent_skin.title)`  but the prefix continues to show up:
```
        skin:
          attributes:
            skin_parents:
              base:
                site_parent: You can't use %{title} as a parent unless replacing the default archive skin.
          format: "%{message}"
        skin_parent:
          attributes:
            base:
              site_parent: You can't use %{title} as a parent unless replacing the default archive skin.
          format: "%{message}"
 ```
 
 Is there something I am doing incorrectly?

tickinginstant: You need to set `format: %{message}` for `skin/skin_parents`, not for `skin`. In the translation code, `/` isn't used to mean nesting; it's just another character that can be used in a key. Like with the `work/parent_work_relationships` key: https://github.com/otwcode/otwarchive/blob/ba94c04580de2bac4c7a932e33a11fad4de7e06d/config/locales/models/en.yml#L153-L154
The class of the parent object is `Work`. The name of the association is `parent_work_relationships`. So when the related work error is displayed on the work, it uses the `work/parent_work_relationships` key to determine the formatting. Similarly, when the skin parent error is displayed on the skin, it uses the `skin/skin_parents` key to determine the formatting.

weeklies: Thank you for expalining, I was trying to Google this but din't know how or where to look.

tickinginstant: I used to have trouble finding resources on how strings are generated with i18n (aside from what's covered in the [Rails guide](https://guides.rubyonrails.org/i18n.html)). So I ended up adding a monkeypatch that prints out all i18n key look-ups, inspired by [the instructions here](http://www.unixgods.org/Rails/where_is_Rails_trying_to_lookup_L10N_strings.html) but starting with [the current version of the lookup function](https://github.com/ruby-i18n/i18n/blob/7cf09474b77fd41e65d979134b0525f67cf371b0/lib/i18n/backend/simple.rb#L88-L107). It can produce a LOT of spam in the logs when it's enabled, but I can usually track down the information I'm looking for eventually.

Have you double-checked that the error messages look okay for the other validations? Switching to the `%{message}` format often breaks the formatting for errors produced by the built-in validators like `presence` and `uniqueness`, because those error messages are designed around the default `%{attribute} %{message}` error formatting.

weeklies: Thanks for the tip! It slipped my mind to check the other messages, I will do it soon.

weeklies:
While testing this again, I also found that this PR also fixes another Error 500. Steps to reproduce on production:

1. Create a new site skin with a unique name (e.g. newskin101)
2. Add it (newskin101) as a parent skin 
3. See an Error 500 instead of an error message shown when editing a created skin ("You can't make a skin its own parent")

This PR causes the `presence` validation to run and show an error message. I have changed the name of the attribute from `Skin parents parent skin` to `Parent skin`., like so:

image
image

brianjaustin (approved):

PR #4557, opened 2023-06-29
AO3-6480 Update text on adult content warning to improve accessibility

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6480

## Purpose

* Adds a page heading to make the page more accessible to screenreader users, and tweak CSS to prevent extra space above our first-ever h2 landmark
* i18n
* Update text to be more accessible to our Chinese users

## Testing Instructions

Refer to Jira

## Credit

Sarken, she/her

### DIFF
--- a/config/locales/views/en.yml
+++ b/config/locales/views/en.yml
@@ -632,6 +632,14 @@ en:
         link_tos: Terms of Service
         welcome_text: Hi! It looks like you've just logged into the Archive for the first time. %{help_link} or dismiss this message permanently.
   works:
+    adult:
+      caution: This work could have adult content. If you proceed you have agreed that you are willing to see such content.
+      footnote: If you accept cookies from our site and you choose "Proceed", you will not be asked again during this session (that is, until you close your browser). If you log in you can store your preference and never be asked again.
+      navigation:
+        back: Go Back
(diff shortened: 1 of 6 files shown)


config/locales/views/en.yml
```diff
@@ -632,6 +632,14 @@ en:
         link_tos: Terms of Service
         welcome_text: Hi! It looks like you've just logged into the Archive for the first time. %{help_link} or dismiss this message permanently.
   works:
+    adult:
+      caution: This work could have adult content. If you continue you have agreed that you are willing to see such content.
```
brianjaustin, line 636: **Nitpick:** I think there should be a comma after the initial clause in "if you continue..."
```suggestion
      caution: This work could have adult content. If you continue, you have agreed that you are willing to see such content.
```

brianjaustin (approved):
Looks good to me (just one grammatical nitpick)!

config/locales/views/en.yml
sarken: I fully agree and actually mentioned that when getting approval... no one explicitly agreed, though, so I intentionally left it as-is. I guess we can do it and then revert if anyone complains >.>

PR #4561, opened 2023-07-03
AO3-6488 Optimize Challenge queries to make orphaned works editable

weeklies:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6488

## Purpose

Optimize ChallengeAssignments related queries so that editing orphaned works doesn't run into 504 Gateway Timeout. 🤞 

I hope the pre-existing spec and features regarding challenges are sufficient. Though if it isn't, I would appreciate if someone could point me in the right direction as to where to create a necessary feature/spec. 

## Testing Instructions

See Jira issue,

## Credit
weeklies (she/her)

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -810,7 +810,7 @@ class WorksController < ApplicationController
     end
 
     if params[:claim_id]
-      @posting_claim = ChallengeClaim.find(params[:claim_id])
+      @posting_claim = ChallengeClaim.find_by(id: params[:claim_id])
     end
   end
 
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -810,7 +810,7 @@ class WorksController < ApplicationController
     end
 
     if params[:claim_id]
-      @posting_claim = ChallengeClaim.find_by(id: params[:claim_id])
+      @posting_claim = ChallengeClaim.find(params[:claim_id])
     end
   end
 
--- a/app/helpers/works_helper.rb
+++ b/app/helpers/works_helper.rb
@@ -191,4 +191,13 @@ module WorksHelper
       chapter_total_display(work)
     end
   end
+
+  def get_open_assignments(user)
+    if user == User.orphan_account
+      # orphan_account has too many pseuds to feasibly get. This lets admins access the work form without timing out.
+      return user.offer_assignments.undefaulted.unstarted.sent + user.pinch_hit_assignments.undefaulted.unstarted.sent
+    end
+
+    ChallengeAssignment.by_offering_user(user).undefaulted.unstarted.sent
+  end
 end
--- a/app/helpers/works_helper.rb
+++ b/app/helpers/works_helper.rb
@@ -195,7 +195,7 @@ module WorksHelper
   def get_open_assignments(user)
     if user == User.orphan_account
       # orphan_account has too many pseuds to feasibly get. This lets admins access the work form without timing out.
-      return user.offer_assignments.undefaulted.unstarted.sent + user.pinch_hit_assignments.undefaulted.unstarted.sent
+      return []
     end
 
     ChallengeAssignment.by_offering_user(user).undefaulted.unstarted.sent
--- a/app/helpers/works_helper.rb
+++ b/app/helpers/works_helper.rb
@@ -193,11 +193,6 @@ module WorksHelper
   end
 
   def get_open_assignments(user)
-    if user == User.orphan_account
-      # orphan_account has too many pseuds to feasibly get. This lets admins access the work form without timing out.
-      return []
-    end
-
     offer_signups = ChallengeAssignment.offer_signups_by(user).undefaulted.unstarted.sent
     pinch_hits = ChallengeAssignment.pinch_hits_by(user).undefaulted.unstarted.sent
 
--- a/app/helpers/works_helper.rb
+++ b/app/helpers/works_helper.rb
@@ -193,9 +193,9 @@ module WorksHelper
   end
 
   def get_open_assignments(user)
-    offer_signups = ChallengeAssignment.offer_signups_by(user).undefaulted.unstarted.sent
-    pinch_hits = ChallengeAssignment.pinch_hits_by(user).undefaulted.unstarted.sent
+    offer_signups = user.offer_assignments.undefaulted.unstarted.sent
+    pinch_hits = user.pinch_hit_assignments.undefaulted.unstarted.sent
 
-    (offer_signups + pinch_hits).uniq
+    (offer_signups + pinch_hits)
   end
 end
--- a/app/helpers/works_helper.rb
+++ b/app/helpers/works_helper.rb
@@ -198,6 +198,9 @@ module WorksHelper
       return []
     end
 
-    ChallengeAssignment.by_offering_user(user).undefaulted.unstarted.sent
+    offer_signups = ChallengeAssignment.offer_signups_by(user).undefaulted.unstarted.sent
+    pinch_hits = ChallengeAssignment.pinch_hits_by(user).undefaulted.unstarted.sent
+
+    (offer_signups + pinch_hits).uniq
   end
 end
--- a/app/models/challenge_assignment.rb
+++ b/app/models/challenge_assignment.rb
@@ -53,6 +53,15 @@ class ChallengeAssignment < ApplicationRecord
 
   scope :order_by_offering_pseud, -> { joins(OFFERING_PSEUD_JOIN).order("pseuds.name") }
 
+  scope :offer_signups_by, lambda { |user|
+    joins(offer_signup: { pseud: :user })
+      .where(users: { id: user.id })
+  }
+
+  scope :pinch_hits_by, lambda { |user|
+    joins(pinch_hitter: :user)
+      .where(users: { id: user.id })
+  }
 
   # Get all of a user's assignments
   scope :by_offering_user, lambda {|user|
--- a/app/models/challenge_assignment.rb
+++ b/app/models/challenge_assignment.rb
@@ -53,15 +53,6 @@ class ChallengeAssignment < ApplicationRecord
 
   scope :order_by_offering_pseud, -> { joins(OFFERING_PSEUD_JOIN).order("pseuds.name") }
 
-  scope :offer_signups_by, lambda { |user|
-    joins(offer_signup: { pseud: :user })
-      .where(users: { id: user.id })
-  }
-
-  scope :pinch_hits_by, lambda { |user|
-    joins(pinch_hitter: :user)
-      .where(users: { id: user.id })
-  }
 
   # Get all of a user's assignments
   scope :by_offering_user, lambda {|user|
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -459,6 +459,7 @@ class Work < ApplicationRecord
 
     self.challenge_assignments =
       ChallengeAssignment.where(id: valid_ids)
+        .compact
         .select { |assign| valid_users.include?(assign.offering_user) }
   end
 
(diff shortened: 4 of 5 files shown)


weeklies:
I am not confident of tackling (or optimizing) the query in [models/challenge_assignment.rb](https://github.com/otwcode/otwarchive/blob/master/app/models/challenge_assignment.rb#L58), but I am wondering in the case of `orphan_account` whether or not we can just not query for the open assignments its non-default pseuds have?

For example:
```ruby
if user == User.orphan_account
  return user.offer_assignments.undefaulted.unstarted.sent + user.pinch_hit_assignments.undefaulted.unstarted.sent
end

ChallengeAssignment.by_offering_user(current_user).undefaulted.unstarted.sent
```

weeklies:
I think the test failure is an intermittent one — setitng this as ready for review.

app/helpers/works_helper.rb
```diff
@@ -191,4 +191,13 @@ def chapter_total_display_with_link(work)
       chapter_total_display(work)
     end
   end
+
+  def get_open_assignments(user)
+    if user == User.orphan_account
+      # orphan_account has too many pseuds to feasibly get. This lets admins access the work form without timing out.
+      return user.offer_assignments.undefaulted.unstarted.sent + user.pinch_hit_assignments.undefaulted.unstarted.sent
```
ceithir, line 198: Not super on point with how assignments work, so might be asking the obvious here but... How can the orphan account have open assignments? Should it? What would happen if we were to just return an empty list there?

sarken: In theory, someone could assign orphan_account a pinch hit in an exchange.

In reality, though, I checked staging and production, and the only assignment orphan_account had was the one I just gave it to test whether it could be assigned a pinch hit:

```
user = User.orphan_account
 => #<User id:...
user.offer_assignments.undefaulted.unstarted.sent.size
 => 0 
user.offer_assignments.size
 => 0 
user.pinch_hit_assignments.undefaulted.unstarted.sent.size
 => 0 

user.pinch_hit_assignments.undefaulted.unstarted.sent.size
 => 1 
```

weeklies: I am wondering then whether ceithir's empty list could work.

sarken: I think it would be fine to do it that way, yeah.

app/helpers/works_helper.rb
```diff
@@ -191,4 +191,13 @@ def chapter_total_display_with_link(work)
...
     end
   end
+
+  def get_open_assignments(user)
+    if user == User.orphan_account
+      # orphan_account has too many pseuds to feasibly get. This lets admins access the work form without timing out.
+      return []
+    end
+
+    ChallengeAssignment.by_offering_user(user).undefaulted.unstarted.sent
```
tickinginstant, line 201: I think it'd be cleaner to try to improve the performance of the query more generally, rather than special casing `orphan_account`. If anyone ever feels like adding enough pseuds to rival `orphan_account`, there's nothing stopping them, and they'll still expect their account to work properly.

weeklies: Okay, I can give it a try when I have the time. 

(As an aside, maybe it is good to consider a maximum number of pseuds a non-orphan account user  can create?)

weeklies:
### Testing Notes
I've tested the optimization on the same feature step in `challenge_yuletide.feature`, In summary, it looks like the current implementation doesn't use `ChallengeAssignment` indexes, namely `assignments_on_offer_signup_id` and `assignments_on_pinch_hitter_id`. 

As @tickinginstant has suggested in past discussion, I tried to split the OR in the original query and created two separate queries, which has been tested to be using the `ChallengeAssignment` indexes. 

I think it's an improvement? Phew.

#### Original query 
- `by_offering_user`
```sql
SELECT DISTINCT challenge_assignments.* FROM `challenge_assignments`
LEFT JOIN challenge_signups ON challenge_assignments.offer_signup_id = challenge_signups.id
INNER JOIN pseuds ON (challenge_assignments.pinch_hitter_id = pseuds.id OR challenge_signups.pseud_id = pseuds.id)
INNER JOIN users ON pseuds.user_id = users.id WHERE (users.id = 399)
```

```
+----+-------------+-----------------------+--------+------------------------------------------+----------------------------------+---------+-------------------------------------------------------+------+------------------------------------+
| id | select_type | table                 | type   | possible_keys                            | key                              | key_len | ref                                                   | rows | Extra                              |
+----+-------------+-----------------------+--------+------------------------------------------+----------------------------------+---------+-------------------------------------------------------+------+------------------------------------+
|  1 | SIMPLE      | users                 | const  | PRIMARY                                  | PRIMARY                          | 4       | const                                                 | 1    | Using index; Using temporary       |
|  1 | SIMPLE      | pseuds                | ref    | PRIMARY,index_pseuds_on_user_id_and_name | index_pseuds_on_user_id_and_name | 5       | const                                                 | 1    | Using index                        |
|  1 | SIMPLE      | challenge_assignments | ALL    | assignments_on_pinch_hitter_id           | NULL                             | NULL    | NULL                                                  | 7    | Using join buffer (flat, BNL join) |
|  1 | SIMPLE      | challenge_signups     | eq_ref | PRIMARY                                  | PRIMARY                          | 4       | otwarchive_test.challenge_assignments.offer_signup_id | 1    | Using where; Distinct              |
+----+-------------+-----------------------+--------+------------------------------------------+----------------------------------+---------+-------------------------------------------------------+------+------------------------------------+
```

#### New queries
I've split it into two seperate queries 

- `offer_signups_by`
```sql
SELECT `challenge_assignments`.* FROM `challenge_assignments`
INNER JOIN `challenge_signups` ON `challenge_signups`.`id` = `challenge_assignments`.`offer_signup_id` 
INNER JOIN `pseuds` ON `pseuds`.`id` = `challenge_signups`.`pseud_id` 
INNER JOIN `users` ON `users`.`id` = `pseuds`.`user_id` WHERE `users`.`id` = 391
```
```
+----+-------------+-----------------------+-------+------------------------------------------+----------------------------------+---------+--------------------------------------+------+-------------+
| id | select_type | table                 | type  | possible_keys                            | key                              | key_len | ref                                  | rows | Extra       |
+----+-------------+-----------------------+-------+------------------------------------------+----------------------------------+---------+--------------------------------------+------+-------------+
|  1 | SIMPLE      | users                 | const | PRIMARY                                  | PRIMARY                          | 4       | const                                | 1    | Using index |
|  1 | SIMPLE      | pseuds                | ref   | PRIMARY,index_pseuds_on_user_id_and_name | index_pseuds_on_user_id_and_name | 5       | const                                | 1    | Using index |
|  1 | SIMPLE      | challenge_signups     | ref   | PRIMARY,signups_on_pseud_id              | signups_on_pseud_id              | 5       | otwarchive_test.pseuds.id            | 1    | Using index |
|  1 | SIMPLE      | challenge_assignments | ref   | assignments_on_offer_signup_id           | assignments_on_offer_signup_id   | 5       | otwarchive_test.challenge_signups.id | 1    |             |
+----+-------------+-----------------------+-------+------------------------------------------+----------------------------------+---------+--------------------------------------+------+-------------+
```
- `pinch_hits_by`


```sql
SELECT `challenge_assignments`.* FROM `challenge_assignments`
INNER JOIN `pseuds` ON `pseuds`.`id` = `challenge_assignments`.`pinch_hitter_id`
INNER JOIN `users` ON `users`.`id` = `pseuds`.`user_id` WHERE `users`.`id` = 391
```


```
+----+-------------+-----------------------+-------+------------------------------------------+----------------------------------+---------+---------------------------+------+-------------+
| id | select_type | table                 | type  | possible_keys                            | key                              | key_len | ref                       | rows | Extra       |
+----+-------------+-----------------------+-------+------------------------------------------+----------------------------------+---------+---------------------------+------+-------------+
|  1 | SIMPLE      | users                 | const | PRIMARY                                  | PRIMARY                          | 4       | const                     | 1    | Using index |
|  1 | SIMPLE      | pseuds                | ref   | PRIMARY,index_pseuds_on_user_id_and_name | index_pseuds_on_user_id_and_name | 5       | const                     | 1    | Using index |
|  1 | SIMPLE      | challenge_assignments | ref   | assignments_on_pinch_hitter_id           | assignments_on_pinch_hitter_id   | 5       | otwarchive_test.pseuds.id | 1    |             |
+----+-------------+-----------------------+-------+------------------------------------------+----------------------------------+---------+---------------------------+------+-------------+
```

app/models/challenge_assignment.rb
```diff
@@ -53,6 +53,15 @@ def signups_match
...
+      .where(users: { id: user.id })
+  }
+
+  scope :pinch_hits_by, lambda { |user|
+    joins(pinch_hitter: :user)
+      .where(users: { id: user.id })
+  }
 
   # Get all of a user's assignments
   scope :by_offering_user, lambda {|user|
```
weeklies, line 67: I kept this because two challenge related views use this, and I don't think the timeout issue will affect those cases.

app/models/challenge_assignment.rb
```diff
@@ -53,6 +53,15 @@ def signups_match
...
 
+  scope :offer_signups_by, lambda { |user|
+    joins(offer_signup: { pseud: :user })
+      .where(users: { id: user.id })
+  }
+
+  scope :pinch_hits_by, lambda { |user|
+    joins(pinch_hitter: :user)
+      .where(users: { id: user.id })
+  }
```
tickinginstant, line 64: Both of these scopes contain more joins than necessary, which may impact the efficiency. Was there something wrong with the previous pair of queries that you were using? https://github.com/otwcode/otwarchive/blob/b79c75b7e4859148885a745f63cd9762a3f9cbad/app/helpers/works_helper.rb#L198

weeklies: 🥹 I was very out of my element when I returned to look at the code for challenges.  Pardon my confusion.

app/models/work.rb
```diff
@@ -454,9 +454,13 @@ def challenge_claim_ids
 
   # Only allow a work to fulfill an assignment assigned to one of this work's authors
   def challenge_assignment_ids=(ids)
+    valid_ids = ids.reject(&:blank?)
```
tickinginstant, line 457: Since you're using `where` below, instead of calling `find`, you don't need to filter the IDs to remove the empty string here. You should just be able to use `where(id: ids)`.

(If you want to preserve the existing behavior of a 404 when passing an invalid assignment ID, you could keep this line and replace `ChallengeAssignment.where(id: valid_ids)` below with `ChallengeAssignment.find(valid_ids)`. But given that this function is already built around ignoring invalid input, `where` is probably the better approach.)

app/models/work.rb
```diff
@@ -454,9 +454,13 @@ def challenge_claim_ids
...
+    valid_users = (self.users + [User.current_user]).compact
+
     self.challenge_assignments =
-      ids.map { |id| id.blank? ? nil : ChallengeAssignment.find(id) }.compact.
-      select { |assign| (self.users + [User.current_user]).compact.include?(assign.offering_user) }
+      ChallengeAssignment.where(id: valid_ids)
+        .compact
```
tickinginstant, line 462: `where` will never return an array containing `nil`, so calling `compact` here is unnecessary.

tickinginstant (commented):
The changes to the code causing the timeout look good to me! I just have a few comments on the other changes you've included.

app/controllers/works_controller.rb
```diff
@@ -810,7 +810,7 @@ def set_work_form_fields
     end
 
     if params[:claim_id]
-      @posting_claim = ChallengeClaim.find_by(id: params[:claim_id])
+      @posting_claim = ChallengeClaim.find(params[:claim_id])
```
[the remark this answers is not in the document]
weeklies: I'll change it back then.

tickinginstant (approved):
Thanks!

PR #4565, opened 2023-07-06
AO3-3997 Fix for email address validation #4390

potpotkettle:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-3997

## Purpose

Fix additional issues found in AO3-3997, after #4390 

## Testing Instructions

- As a user with the ability to import works, try importing with an ill-formatted email address. ("Archiving for others") It should show an error message (where no error was shown and importing succeeded).
- As a user with the ability to send invitations, try sending an invitation to an ill-formatted email address. It should show an error message (where it was missing previously).

## References

Comments in the Jira link above, particularly https://otwarchive.atlassian.net/browse/AO3-3997?focusedCommentId=360179 .

## Credit

Potpotkettle (they/them)

### DIFF
--- a/app/models/story_parser.rb
+++ b/app/models/story_parser.rb
@@ -421,8 +421,12 @@ class StoryParser
       name = name.to_ascii.gsub(/[^\w[ \-@.]]/u, "")
       external_author = ExternalAuthor.find_or_create_by(email: email)
       external_author_name = external_author.default_name
+
+      # if the name and email don't exist in the DB tables, add it
       unless name.blank?
-        external_author_name = ExternalAuthorName.where(name: name, external_author_id: external_author.id).first ||
+        raise Error, external_author.errors.full_messages.join(" ") if external_author.invalid?
+
+        external_author_name = ExternalAuthorName.find_by(name: name, external_author_id: external_author.id) ||
                                ExternalAuthorName.new(name: name)
         external_author.external_author_names << external_author_name
         external_author.save
--- a/app/models/story_parser.rb
+++ b/app/models/story_parser.rb
@@ -418,33 +418,24 @@ class StoryParser
   def parse_author_common(email, name)
     errors = []
 
-    if name.present?
-      # convert to ASCII and strip out invalid characters (everything except alphanumeric characters, _, @ and -)
-      redacted_name = name.to_ascii.gsub(/[^\w[ \-@.]]/u, "")
-    else
-      errors << "No author name specified"
-    end
+    errors << "No author name specified" if name.blank?
 
     if email.present?
       external_author = ExternalAuthor.find_or_create_by(email: email)
-      external_author_name = external_author.default_name
-
-      errors += external_author.errors.full_messages if external_author.invalid?
+      errors += external_author.errors.full_messages
     else
       errors << "No author email specified"
     end
 
-    raise Error, errors.join("\n") unless errors.empty?
+    raise Error, errors.join("\n") if errors.present?
 
-    # if the provided name and email don't exist in the DB tables, add them
+    # convert to ASCII and strip out invalid characters (everything except alphanumeric characters, _, @ and -)
+    redacted_name = name.to_ascii.gsub(/[^\w[ \-@.]]/u, "")
     if redacted_name.present?
-      external_author_name = ExternalAuthorName.find_by(name: redacted_name, external_author_id: external_author.id) ||
-                             ExternalAuthorName.new(name: redacted_name)
-      external_author.external_author_names << external_author_name
-      external_author.save
+      external_author.names.find_or_create_by(name: redacted_name)
+    else
+      external_author.default_name
     end
-
-    external_author_name
   end
 
   def get_chapter_from_work_params(work_params)
--- a/app/models/story_parser.rb
+++ b/app/models/story_parser.rb
@@ -416,28 +416,35 @@ class StoryParser
   end
 
   def parse_author_common(email, name)
-    if name.present? && email.present?
+    errors = []
+
+    if name.present?
       # convert to ASCII and strip out invalid characters (everything except alphanumeric characters, _, @ and -)
-      name = name.to_ascii.gsub(/[^\w[ \-@.]]/u, "")
+      redacted_name = name.to_ascii.gsub(/[^\w[ \-@.]]/u, "")
+    else
+      errors << "No author name specified"
+    end
+
+    if email.present?
       external_author = ExternalAuthor.find_or_create_by(email: email)
       external_author_name = external_author.default_name
 
-      # if the name and email don't exist in the DB tables, add it
-      unless name.blank?
-        raise Error, external_author.errors.full_messages.join(" ") if external_author.invalid?
-
-        external_author_name = ExternalAuthorName.find_by(name: name, external_author_id: external_author.id) ||
-                               ExternalAuthorName.new(name: name)
-        external_author.external_author_names << external_author_name
-        external_author.save
-      end
-      external_author_name
+      errors += external_author.errors.full_messages if external_author.invalid?
     else
-      messages = []
-      messages << "No author name specified" if name.blank?
-      messages << "No author email specified" if email.blank?
-      raise Error, messages.join("\n")
+      errors << "No author email specified"
     end
+
+    raise Error, errors.join("\n") unless errors.empty?
+
+    # if the provided name and email don't exist in the DB tables, add them
+    if redacted_name.present?
+      external_author_name = ExternalAuthorName.find_by(name: redacted_name, external_author_id: external_author.id) ||
+                             ExternalAuthorName.new(name: redacted_name)
+      external_author.external_author_names << external_author_name
+      external_author.save
+    end
+
+    external_author_name
   end
 
   def get_chapter_from_work_params(work_params)
--- a/app/views/invitations/_invitation.html.erb
+++ b/app/views/invitations/_invitation.html.erb
@@ -10,8 +10,9 @@
     <dd><%= invitation.invitee_email %></dd>
     <% else %>
     <dd>
+      <%= error_messages_for @invitation %>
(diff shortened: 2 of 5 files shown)


app/models/story_parser.rb
```diff
@@ -421,8 +421,12 @@ def parse_author_common(email, name)
       name = name.to_ascii.gsub(/[^\w[ \-@.]]/u, "")
       external_author = ExternalAuthor.find_or_create_by(email: email)
       external_author_name = external_author.default_name
+
+      # if the name and email don't exist in the DB tables, add it
       unless name.blank?
-        external_author_name = ExternalAuthorName.where(name: name, external_author_id: external_author.id).first ||
+        raise Error, external_author.errors.full_messages.join(" ") if external_author.invalid?
+
+        external_author_name = ExternalAuthorName.find_by(name: name, external_author_id: external_author.id) ||
```
potpotkettle, line 429: Does this comment look accurate? I didn't understand the intention of this part of the code first, and the comment I inserted was my best guess.

app/models/story_parser.rb
```diff
@@ -421,8 +421,12 @@ def parse_author_common(email, name)
       name = name.to_ascii.gsub(/[^\w[ \-@.]]/u, "")
       external_author = ExternalAuthor.find_or_create_by(email: email)
       external_author_name = external_author.default_name
+
+      # if the name and email don't exist in the DB tables, add it
       unless name.blank?
-        external_author_name = ExternalAuthorName.where(name: name, external_author_id: external_author.id).first ||
+        raise Error, external_author.errors.full_messages.join(" ") if external_author.invalid?
```
tickinginstant, line 427: I think this line is in the wrong place. If someone enters `!!!!!` as the name, and also uses an invalid email address, then the exclamation points in the name will be stripped down so that `name` is blank. So even though it's an invalid email, I don't think it will reach this line and show the error. I'm pretty sure this line needs to be outside of the `unless name.blank?` block.

app/views/invitations/_invitation.html.erb
```diff
@@ -10,8 +10,9 @@
     <dd><%= invitation.invitee_email %></dd>
     <% else %>
     <dd>
+      <%= error_messages_for @invitation %>
```
sarken, line 13: I poked around a bit to see where we usually put this, and it looks like most of the time it's either after the page heading (which we don't have here) or the first or second thing inside the form. Can we move this inside the `form_for`? 

You might also be able to use `invitation` rather than `@invitation` here.

app/views/invitations/_invitation.html.erb
```diff
@@ -10,8 +10,9 @@
...
       <%= form_for(invitation) do |f| %>
-        <p><label for="invitee_email">Enter an email address:</label> <%= f.text_field :invitee_email %></p>
+        <p><%= label_tag "invitation[invitee_email]", t('.email_address_label') %>: <%= f.text_field :invitee_email %></p>
```
sarken, line 15: `f.label :invitee_email` should work here, and it would be more consistent since we're using `f.text_field :invitee_email`. If you could also use double quotes around the translation key name and remove the colon (we've been slowly eliminating them on our form labels, plus taking it out of the actual label means it doesn't get the red highlight when there's an error), that would be great!

features/other_a/invite_request.feature
```diff
@@ -62,6 +62,18 @@ Feature: Invite requests
...
 
+  Scenario: User can fail to send an invitation to an invalid email address
+
+    Given invitations are required
+      And I am logged in as "user1"
+      And I request some invites
+      And an admin grants the request
+      And I try to invite a friend from my user page
+    When all emails have been delivered
+      And I fill in "invitation[invitee_email]" with "test@"
```
sarken, line 73: When possible, we [prefer the user-facing label text](https://github.com/otwcode/otwarchive/wiki/Automated-Testing#our-cucumber-standards).

features/other_a/invite_request.feature
```diff
@@ -62,6 +62,18 @@ Feature: Invite requests
       And I should not see "Sorry, you have no unsent invitations right now."
       And I should see "You have 2 open invitations and 0 that have been sent but not yet used."
 
+  Scenario: User can fail to send an invitation to an invalid email address
+
+    Given invitations are required
```
sarken, line 67: Looking at `Scenario:  A user can manage their invitations` below, I think we can simplify this:

```
Given I am logged in as "user1"
  And "user1" has "1" invitation
  And I am on user1's manage invitations page
When I follow the link for "user1" first invite
  And I fill in etc
```

app/models/story_parser.rb
potpotkettle: Thank you for your comment. In that scenario (for the case where the name is blank and the email is invalid), should the error just say "invalid email"? Or should it alert about the name *and* the email?

app/views/invitations/_invitation.html.erb
potpotkettle: Just to make sure, do you want the "Enter an email address" label to be in red as well, like this?

sarken: Yup, that's fine!

features/other_a/invite_request.feature
```diff
@@ -62,16 +62,14 @@ Feature: Invite requests
       And I should not see "Sorry, you have no unsent invitations right now."
       And I should see "You have 2 open invitations and 0 that have been sent but not yet used."
 
-  Scenario: User can fail to send an invitation to an invalid email address
+  Scenario: User can see an error after trying to invide an invalid email address
```
potpotkettle, line 65: Oops. I think I have some more things to fix anyway, and will fix the typo along them later.

app/models/story_parser.rb
tickinginstant: It'd be more difficult to handle that case, since the `ExternalAuthor` currently doesn't get created if either the `name` or the `email` is blank, so it never runs through the validator. Since it's not producing a 500 error, I'd be tempted to let it slide for the moment, but someone who's more familiar with the Open Doors committee might have a better answer for you.

(Really, I think the entire function should be rewritten, so that there are proper error messages for everything instead of silently fixing invalid names, and it's easier to follow the code. But IDK if that's in scope, and IDK if that's what Open Doors wants.)

potpotkettle: I also reorganized ``parser_author_common`` in a way that makes sense to me. And to avoid breaking something, I added tests and confirmed the behavior didn't change (as far as I can see) except for the email address validation, too.

spec/models/story_parser_spec.rb
```diff
@@ -193,21 +193,67 @@ class StoryParser
...
     it "raises an exception when neither the external author name nor email is provided" do
-      expect {
+      expect do
         @sp.parse_author("", nil, nil)
-      }.to raise_exception(StoryParser::Error) { |e| expect(e.message).to eq("No author name specified\nNo author email specified") }
+      end.to raise_exception(StoryParser::Error, "No author name specified\nNo author email specified")
+    end
+
+    it "gives the same external author object for the same email" do
+      res1 = @sp.parse_author("", "Author Name", "author@example.com")
```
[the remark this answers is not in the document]
potpotkettle: I think it's possible to replace the instance variable with ``let``. Not sure if it's a good idea to do it now, because that would be an unrelated refactoring that would touch most of the methods here.

app/models/story_parser.rb
tickinginstant: If you're going to rewrite it for clarity reasons, I think it'd be cleaner to consolidate most of the name-related code. As is, `redacted_name` is set, and then there are a bunch of lines involving errors & emails, and then `redacted_name` is finally used. But there's no reason for it to be set that early, and it's easier to keep track of the variable's purpose if it's set right before it's used.

Similarly, the two different possible values of `external_author_name` are set in two different `if` blocks, so it's hard to see at a glance what the function is supposed to return. It'd be cleaner to set it equal to `external_author.default_name` in the `else` of the `if redacted_name.present?` block.

With some other cleanup, that could get you something like this:
```ruby
  def parse_author_common(email, name)
    errors = []

    errors << "No author name specified" if name.blank?

    if email.present?
      external_author = ExternalAuthor.find_or_create_by(email: email)
      errors += external_author.errors.full_messages
    else
      errors << "No author email specified"
    end

    raise Error, errors.join("\n") if errors.present?

    # convert to ASCII and strip out invalid characters (everything except alphanumeric characters, _, @ and -)
    redacted_name = name.to_ascii.gsub(/[^\w[ \-@.]]/u, "")
    if redacted_name.present?
      external_author.names.find_or_create_by(name: redacted_name)
    else
      external_author.default_name
    end
  end
```
Ideally, it'd also be nice to move some or all of the error-checking code here into the models, but that's almost certainly out of scope.

potpotkettle: Thanks a lot for the rewrite. Do you think  ``.save`` is unneccesary? I thought the method makes persistent changes to the databases so that author names can be retrieved by querying emails later. (Or maybe it's implicitly saved somehow?)

tickinginstant: `find_or_create_by` calls `create` [under the hood](https://github.com/rails/rails/blob/593893c901f87b4ed205751f72df41519b4d2da3/activerecord/lib/active_record/encryption/extended_deterministic_queries.rb#L94-L96), and `create` calls `save` [under the hood](https://github.com/rails/rails/blob/593893c901f87b4ed205751f72df41519b4d2da3/activerecord/lib/active_record/persistence.rb#L33-L41). So both the external author and the external author name are getting saved to the database if they don't already exist, because they're both fetched/created with `find_or_create_by` in the modified code.

potpotkettle: Understood. Thanks for explaining.

tickinginstant (approved):

ariana-paris (approved):

PR #4505, opened 2023-04-29
AO3-6119 Remove automatically_approve_collections from preferences

gmjimenezc:
# Pull Request Checklist

* [x] Have you read ["How to write the perfect pull request"](https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/)?
* [x] Have you read the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [ ] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)?
* [x] Have you added the [Jira](https://otwarchive.atlassian.net) issue number
  as the *first* thing in your pull request title (e.g. `AO3-1234 Fix thing`)
* [x] Do you have fewer than 5 pull requests already open? If not, please wait
  until they are reviewed and merged before creating new pull requests.

## Issue

https://otwarchive.atlassian.net/browse/AO3-6119

## Purpose

Remove the automatically_approve_collections column from the preferences table with a migration

## Credit

warlockmel (she/her)

PR #4568, opened 2023-07-07
AO3-6549 Remove ignore of auto collection invite column

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6549

## Purpose

Remove the ignore for the column that was deleted in #4505

## Testing Instructions

Reference Jira

## References

## Credit

Brian Austin (they/he)

### DIFF
--- a/spec/models/preference_spec.rb
+++ b/spec/models/preference_spec.rb
@@ -1,0 +1,52 @@
+require "spec_helper"
+
+describe Preference do
+  it { is_expected.to allow_value("Test_Title-1 .,").for(:work_title_format) }
+  it { is_expected.not_to allow_value("@; Test").for(:work_title_format) }
+
+  describe ".disable_work_skin?" do
+    it "returns false for creator" do
+      expect(Preference.disable_work_skin?("creator")).to be(false)
+    end
+
+    %w[light disable].each do |param|
+      it "returns true for #{param}" do
+        expect(Preference.disable_work_skin?(param)).to be(true)
+      end
+    end
+
+    context "when the current user is a guest" do
+      it "returns false" do
+        expect(Preference.disable_work_skin?("foo")).to be(false)
+      end
+    end
+
+    context "when the current user is registered" do
+      let(:user) { create(:user) }
+
+      before do
+        User.current_user = user
+      end
+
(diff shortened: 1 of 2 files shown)


brianjaustin:
Since the change was so small, I also went through the RuboCop issues for that file (and added tests). They should be addressed eventually either way, but if we prefer not to do all this at once I can revert some or all of them

ceithir (commented):
I fear you've opened Pandora's box: Now you have to update a regexp to a new syntax, and so must add a bunch more tests to make quite sure nothing was broken in the process.

What's currently in the PR seems good though. :+1:

brianjaustin:
@ceithir thanks for the review! Based on https://stackoverflow.com/a/3632051, I think it won't actually be too bad (just the extra test which would fail with the old regex)

spec/models/preference_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+
+      context "when the user's preference has skins enabled" do
+        before do
+          user.preference.update!(disable_work_skins: false)
+        end
+
+        it "returns false" do
+          expect(Preference.disable_work_skin?("foo")).to be(false)
+        end
+      end
```
ceithir, line 40: Nitpick: Splitting between context and test might be a bit overkill here:
```suggestion
      it "returns false when the user's preference has skins enabled" do
        user.preference.update!(disable_work_skins: false)
        expect(Preference.disable_work_skin?("foo")).to be(false)
      end
```

spec/models/preference_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+
+      context "when the user's preference has skins disabled" do
+        before do
+          user.preference.update!(disable_work_skins: true)
+        end
+
+        it "returns true" do
+          expect(Preference.disable_work_skin?("foo")).to be(true)
+        end
+      end
```
ceithir, line 50:
```suggestion
      it "returns true when the user's preference has skins disabled" do
        user.preference.update!(disable_work_skins: true)
        expect(Preference.disable_work_skin?("foo")).to be(true)
      end
```

ceithir (approved):

PR #4570, opened 2023-07-09
AO3-4867 & AO3-4868 Prevent pseud ambiguity issues.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4867
https://otwarchive.atlassian.net/browse/AO3-4868

## Purpose

Changes the way that the `parse_byline` and `parse_bylines` functions are defined:
- The original functionality of `parse_byline` has been split into two functions: `parse_byline`, which returns a single pseud, and behaves like the old `assume_matching_login` option was enabled (if you call `first` on the result); and `parse_byline_ambiguous`, which returns an array of pseuds, and behaves like the old `assume_matching_login` option wasn't enabled.
- The function `parse_bylines` has been modified so that instead of allowing `assume_matching_login` as an option, it always calls the new, unambiguous `parse_byline`, because after AO3-4867 there are no cases where it should behave like `assume_matching_login: false` (and in general, I don't think it's a good idea to introduce any cases like that — the user interface gets messy when you have to let users disambiguate multiple pseuds).
- Removed the scope `by_byline`, because it was being used as a shorter replacement for `parse_byline` with the `assume_matching_login` option enabled.
- The `request_signup_pseud=` and `offer_signup_pseud=` functions were modified to use `parse_byline`.

### DIFF
--- a/spec/models/challenge_assignment_spec.rb
+++ b/spec/models/challenge_assignment_spec.rb
@@ -83,4 +83,75 @@ describe ChallengeAssignment do
     end
   end
 
+  describe "request_signup_pseud=" do
+    let!(:collection) { create(:collection, challenge: create(:gift_exchange)) }
+
+    let(:assignment) { collection.assignments.build }
+
+    context "when a user has signed up with a pseud matching their login" do
+      let(:user) { create(:user) }
+
+      let!(:signup) do
+        create(:challenge_signup,
+               collection: collection,
+               pseud: user.default_pseud)
+      end
+
+      it "assigns the user's signup when entering the user's login" do
+        assignment.request_signup_pseud = user.login
+        expect(assignment.request_signup).to eq(signup)
+      end
+
+      context "when another user has signed up with the same pseud name" do
+        let(:ambiguous) { create(:pseud, name: user.login) }
+
+        let(:ambiguous_signup) do
+          create(:challenge_signup,
+                 collection: collection,
+                 pseud: ambiguous_signup)
+        end
+
+        it "assigns the first user's signup when entering the first user's login" do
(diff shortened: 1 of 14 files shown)


spec/models/challenge_assignment_spec.rb
```diff
@@ -83,4 +83,75 @@
...
+      context "when another user has signed up with the same pseud name" do
+        let(:ambiguous) { create(:pseud, name: user.login) }
+
+        let(:ambiguous_signup) do
+          create(:challenge_signup,
+                 collection: collection,
+                 pseud: ambiguous_signup)
+        end
+
+        it "assigns the first user's signup when entering the first user's login" do
```
sarken, line 114: Should we also test the flip side, i.e., it assigns the second user's signup when entering `ambiguous.name (ambiguous.user.login)`?

sarken (commented):
Yay, thank you! Just one tiny suggestion about the tests.

> after AO3-4867 there are no cases where it should behave like assume_matching_login: false (and in general, I don't think it's a good idea to introduce any cases like that — the user interface gets messy when you have to let users disambiguate multiple pseuds)

Agreed x 1000.

spec/models/challenge_assignment_spec.rb
tickinginstant: I added an extra test (and in doing so discovered and fixed a bug with the existing test).

Do you want me to do it for the `OwnedTagSet` tests, as well?

sarken: Yay! And yes please, sorry for not being clear.

sarken (approved):
Thank you!

PR #4576, opened 2023-07-12
AO3-5803 Set minimums for gift exchange requests/offers

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5803

## Purpose

Require at least 1 request and offer per gift exchange signup. This avoids weird behavior when the requirements are 0. Also fixes the problem where 'allowed' values below 'required' were permitted.

## Testing Instructions

Refer to Jira ticket

## References

## Credit

Brian Austin (they/he)

### DIFF
--- a/app/models/challenge_models/gift_exchange.rb
+++ b/app/models/challenge_models/gift_exchange.rb
@@ -21,13 +21,15 @@ class GiftExchange < ApplicationRecord
     allow_blank: true,
     maximum: ArchiveConfig.INFO_MAX, too_long: ts("must be less than %{max} letters long.", max: ArchiveConfig.INFO_MAX)
   }
-  validates :requests_num_required, numericality: { greater_than_or_equal_to: 1, only_integer: true }
-  validates :offers_num_required, numericality: { greater_than_or_equal_to: 1, only_integer: true }
 
   PROMPT_TYPES.each do |type|
     %w[required allowed].each do |setting|
       prompt_limit_field = "#{type}_num_#{setting}"
-      validates prompt_limit_field, numericality: { only_integer: true, less_than_or_equal_to: ArchiveConfig.PROMPTS_MAX, greater_than_or_equal_to: 0 }
+      validates prompt_limit_field, numericality: {
+        only_integer: true,
+        less_than_or_equal_to: ArchiveConfig.PROMPTS_MAX,
+        greater_than_or_equal_to: 1
+      }
     end
   end
 
--- a/app/models/challenge_models/gift_exchange.rb
+++ b/app/models/challenge_models/gift_exchange.rb
@@ -22,6 +22,9 @@ class GiftExchange < ApplicationRecord
     maximum: ArchiveConfig.INFO_MAX, too_long: ts("must be less than %{max} letters long.", max: ArchiveConfig.INFO_MAX)
   }
 
+  validates :requests_num_required, numericality: { greater_than_or_equal_to: 1, only_integer: true }
+  validates :offers_num_required, numericality: { greater_than_or_equal_to: 1, only_integer: true }
+
   PROMPT_TYPES.each do |type|
     %w(required allowed).each do |setting|
       prompt_limit_field = "#{type}_num_#{setting}"
--- a/app/models/challenge_models/gift_exchange.rb
+++ b/app/models/challenge_models/gift_exchange.rb
@@ -1,5 +1,5 @@
 class GiftExchange < ApplicationRecord
-  PROMPT_TYPES = %w(requests offers)
+  PROMPT_TYPES = %w[requests offers].freeze
   include ChallengeCore
 
   override_datetime_setters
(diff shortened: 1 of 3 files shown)


spec/models/gift_exchange_spec.rb
```diff
@@ -1,18 +1,45 @@
...
-      @collection = FactoryBot.create(:collection)
-      @collection.challenge = GiftExchange.new
-      @challenge = @collection.challenge
+    it "succeeds with a valid gift exchange" do
+      expect(challenge.save).to be_truthy
     end
 
-    it "should save" do
-      expect(@challenge.save).to be_truthy
+    context "when allowed offers is bellow required offers" do
```
EchoEkhi, line 23: Oops!

app/models/challenge_models/gift_exchange.rb
```diff
@@ -17,26 +17,26 @@ class GiftExchange < ApplicationRecord
   belongs_to :potential_match_settings, dependent: :destroy
   accepts_nested_attributes_for :potential_match_settings
 
-  validates_length_of :signup_instructions_general, :signup_instructions_requests, :signup_instructions_offers, {
+  validates :signup_instructions_general, :signup_instructions_requests, :signup_instructions_offers, length: {
     allow_blank: true,
     maximum: ArchiveConfig.INFO_MAX, too_long: ts("must be less than %{max} letters long.", max: ArchiveConfig.INFO_MAX)
   }
+  validates :requests_num_required, numericality: { greater_than_or_equal_to: 1, only_integer: true }
+  validates :offers_num_required, numericality: { greater_than_or_equal_to: 1, only_integer: true }
```
tickinginstant, line 25: What's the advantage of adding these validations separately, instead of editing the validation below to have `greater_than_or_equal_to: 1`?

brianjaustin: To be honest, I missed that bit even after changing the style things :facepalm: If we wanted to maintain the current behavior of `_allowed` as 0 being corrected to the `_required` value, then I think the loop there would need to be split up; I'm not convinced that single case is worth the code duplication though. Thoughts?

tickinginstant: I think it makes sense to enforce that `_allowed` has to be at least 1, too. Letting it be 0 will only really affect things when `_required` is also 0 (otherwise it'll just get overridden in the `before_validation` hook), and I think an additional "Number of allowed requests per sign-up must be greater than or equal to 1." message wouldn't go awry in that case.

config/locales/models/en.yml
```diff
@@ -41,6 +41,11 @@ en:
         name_with_colon:
           one: 'Additional Tag:'
           other: 'Additional Tags:'
+      gift_exchange:
+        offers_num_allowed: Number of offers allowed per sign-up
+        offers_num_required: Number of offers required per sign-up
+        requests_num_allowed: Number of requests allowed per sign-up
+        requests_num_required: Number of requests required per sign-up
```
ceithir, line 48: Does having those in the i18n file this allow to clean up a bit view files like [this one](https://github.com/otwcode/otwarchive/blob/8d5b1c97839ec1eb736c3387c1e52e2e30786181/app/views/challenge/gift_exchange/_gift_exchange_form.html.erb#L33)?

spec/models/gift_exchange_spec.rb
```diff
@@ -1,18 +1,45 @@
-require 'spec_helper'
+require "spec_helper"
 
 describe GiftExchange do
+  it do
```
ceithir, line 4: On purpose there's no description there?

brianjaustin: Yes; normally, I'd do something like `it { is_expected.to do_something }` (https://stackoverflow.com/a/25729231) with [Shoulda matchers](https://github.com/thoughtbot/shoulda-matchers). But in this case, it ended up being quite long. I can add a description if that would make things clearer, but Shoulda will output something fairly understandable without

config/locales/models/en.yml
brianjaustin: Hmm. I think it makes migrating those partials to use `I18n.t` a bit easier, but we would still need to include the `:`. So it might be a bit cleaner, but no quite as nice as
```erb
<dt><%= f.label :requests_num_required %></dt>
```
unfortunately. For the purposes of this PR, I put these here to make the field names nicer in error messages

sarken: You can get rid of the colons in any field labels where you see them, btw -- we actually do not want those.

brianjaustin: Neat! Another thing I thought of though: the way the allowed/required fields are split, we wouldn't get the correct naming without manually specifying an I18n key

two rows in a form; the first is labelled Number of requests required per sign-up followed by two text input boxes, labelled Required: and Allowed: respectively; the second is labelled Number of offers required per sign-up followed by two text input boxes, labelled Required: and Allowed: respectively

app/models/challenge_models/gift_exchange.rb
```diff
@@ -17,26 +17,28 @@ class GiftExchange < ApplicationRecord
...
-    %w(request offer).each do |prompt_type|
-      required = eval("#{prompt_type}s_num_required")
-      allowed = eval("#{prompt_type}s_num_allowed")
-      if required > allowed
-        eval("#{prompt_type}s_num_allowed = required")
-      end
+    %w[request offer].each do |prompt_type|
+      required = send("#{prompt_type}s_num_required")
+      allowed = send("#{prompt_type}s_num_allowed")
+      send("#{prompt_type}s_num_allowed=", required) if required > allowed
```
ceithir, line 41: So, the old code was assigning the _string_ `required`? And somehow nothing exploded?

Should similar code in [challenge_core.rb](https://github.com/otwcode/otwarchive/blob/317e3e840ce28f1d66e2ede43fdbd3e6bf0218bd/lib/challenge_core.rb#L7) and [prompt_restriction.rb](https://github.com/otwcode/otwarchive/blob/317e3e840ce28f1d66e2ede43fdbd3e6bf0218bd/app/models/prompt_restriction.rb#L33) be likewise updated? Though might be best to do so in a different ticket/PR.

ceithir (approved):

app/models/challenge_models/gift_exchange.rb
brianjaustin: The old code just doesn't change the value. I _think_ the `eval` was just setting the value for its own context. The second example -- prompt restriction -- does work because it uses `self.<method>`. We could fix this one as well by doing
```rb
eval("self.{prompt_type}s_num_allowed = required")
```
but `eval` is unnecessarily dangerous when `send` will do. At some point we'll want to change the other `eval`s too. That first link -- challenge core -- does seem to have the same problem, so I'll poke a bit more and write something up.

PR #4577, opened 2023-07-15
AO3-6545 Default FAQ to last valid locale instead of always English

ceithir:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6545

## Purpose

If accessing FAQ with no specified locale or an invalid locale, redirect to last locale in session, then user preferred locale, then default locale.

## Testing Instructions

As per JIRA. To be tested as guest, logged in standard user, and admin.

## Credit

Ceithir (he/him)

### DIFF
--- a/spec/controllers/archive_faqs_controller_spec.rb
+++ b/spec/controllers/archive_faqs_controller_spec.rb
@@ -11,7 +11,6 @@ describe ArchiveFaqsController do
   let(:user) do
     user = create(:user)
     user.preference.update!(preferred_locale: user_locale.id)
-    $rollout.activate_user(:set_locale_preference, user)
     user
   end
 
@@ -39,10 +38,20 @@ describe ArchiveFaqsController do
     context "when logged in as a regular user" do
       before { fake_login_known_user(user) }
 
-      it "redirects to the user preferred locale when the locale param is invalid" do
+      it "redirects to the default locale when the locale param is invalid" do
         expect(I18n).not_to receive(:with_locale)
         get :index, params: { language_id: "eldritch" }
-        it_redirects_to(archive_faqs_path(language_id: user_locale.iso))
+        it_redirects_to(archive_faqs_path(language_id: I18n.default_locale))
+      end
+
+      context "with set_locale_preference" do
+        before { $rollout.activate_user(:set_locale_preference, user) }
+
+        it "redirects to the user preferred locale when the locale param is invalid" do
+          expect(I18n).not_to receive(:with_locale)
+          get :index, params: { language_id: "eldritch" }
+          it_redirects_to(archive_faqs_path(language_id: user_locale.iso))
+        end
       end
     end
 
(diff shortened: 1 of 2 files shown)


spec/controllers/archive_faqs_controller_spec.rb
```diff
@@ -6,38 +6,43 @@
...
+        expect(I18n).not_to receive(:with_locale)
+        get :index, params: { language_id: "" }
+        it_redirects_to(archive_faqs_path(language_id: I18n.default_locale))
+      end
 
-    it "redirects to the default locale when the locale param and the session locale are empty" do
-      expect(I18n).not_to receive(:with_locale)
-      get :index, params: { language_id: "" }, session: { language_id: "" }
-      it_redirects_to(archive_faqs_path(language_id: "en"))
+      it "redirects to the default locale when the locale param and the session locale are empty" do
```
brianjaustin, line 32: Isn't this redundant with `redirects to the default locale when the locale param is empty`, since the context is that `session[:locale]` is unset?

spec/controllers/archive_faqs_controller_spec.rb
```diff
@@ -6,38 +6,43 @@
...
-      expect(I18n).to receive(:with_locale).with("eldritch").and_call_original
-      get :index, params: { language_id: "eldritch" }
-      expect(response).to render_template(:index)
-    end
+  let(:non_standard_locale) { create(:locale) }
+  let(:user_locale) { create(:locale) }
+  let(:user) do
+    user = create(:user)
+    user.preference.update!(preferred_locale: user_locale.id)
+    $rollout.activate_user(:set_locale_preference, user)
```
brianjaustin, line 14: I'm not sure if this is on for everyone in prod, but it might be nice to test when it's disabled as well

spec/controllers/archive_faqs_controller_spec.rb
ceithir: This test, as is, is useful to catch an edge case forcing to add a couple of `.presence`, the one where `session[:language_id]` is blank (nil actually), which isn't something possible with the new code, but used to be with the old one. So it's kinda a legacy failsafe.

Name should definitely be updated to reflect that though.

spec/controllers/archive_faqs_controller_spec.rb
```diff
@@ -39,10 +38,20 @@
     context "when logged in as a regular user" do
       before { fake_login_known_user(user) }
 
-      it "redirects to the user preferred locale when the locale param is invalid" do
+      it "redirects to the default locale when the locale param is invalid" do
```
brianjaustin, line 41: I think it might make understanding this easier if it's wrapped in a context as well. Something like
```rb
context "when the set locale preference feature flag is off" do
```

spec/controllers/archive_faqs_controller_spec.rb
```diff
@@ -39,10 +38,20 @@
...
         expect(I18n).not_to receive(:with_locale)
         get :index, params: { language_id: "eldritch" }
-        it_redirects_to(archive_faqs_path(language_id: user_locale.iso))
+        it_redirects_to(archive_faqs_path(language_id: I18n.default_locale))
+      end
+
+      context "with set_locale_preference" do
```
brianjaustin, line 47: Similar to above,
```suggestion
      context "when the set_locale_preference feature flag is on" do
```

spec/controllers/archive_faqs_controller_spec.rb
brianjaustin: Ah, gotcha. Feel free to mark this thread as resolved too

brianjaustin (commented):
Thanks for the added test! Just a couple more wording suggestions, otherwise looks good to me

brianjaustin (approved):

PR #4554, opened 2023-06-22
AO3-5901 Limit number of pseuds on profile and sidebar

neuroalien:
# Pull Request Checklist

* [x] Have you read ["How to write the perfect pull request"](https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/)?
* [x] Have you read the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [x] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)?
* [x] Have you added the [Jira](https://otwarchive.atlassian.net) issue number
  as the *first* thing in your pull request title (e.g. `AO3-1234 Fix thing`)
* [x] Do you have fewer than 5 pull requests already open? If not, please wait
  until they are reviewed and merged before creating new pull requests.

## Issue

https://otwarchive.atlassian.net/browse/AO3-5901

## Purpose

What does this PR do?

* Limits the number of pseuds shown on the user's profile and on the user sidebar to the configured value in ArchiveConfig.ITEMS_PER_PAGE (with cesy's contribution)
* Dynamically loads pseuds on profile page (tickinginstant, sarken)

## Testing Instructions

How can the Archive's QA team verify that this is working as you intended?

If you have a Jira account with access, please update or comment on the issue
with any new or missing testing instructions instead.

## References

Are there other relevant issues/pull requests/mailing list discussions?

## Credit

What name and pronouns should we use to credit you in the [Archive of Our Own's Release Notes](https://archiveofourown.org/admin_posts?tag=1)?

alien, they/them

However, more people contributed to this PR over the years (years!), who should also be credited.

PR #4579, opened 2023-07-15
AO3-5901 Pseud switcher text on the New Pseud page

neuroalien:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5901

## Purpose

What does this PR do?

Provides fallback text for the edge case of the New Pseud page, where `@pseud` is defined, but doesn't have a `name` yet.

## Testing Instructions

## References

## Credit

they/them pronouns.*

### DIFF
--- a/app/views/users/_sidebar.html.erb
+++ b/app/views/users/_sidebar.html.erb
@@ -5,7 +5,7 @@
   <li><%= span_if_current ts("Profile"), user_profile_path(@user) %></li>
   <% if @user.pseuds.size > 1 %>
     <li class="pseud" aria-haspopup="true">
-      <% pseud_link_text = current_page?(@user) ? ts("Pseuds") : (@pseud ? @pseud.name : @user.login) %>
+      <% pseud_link_text = current_page?(@user) ? ts("Pseuds") : (@pseud ? ( @pseud.name ? @pseud.name : ts("Pseuds") ) : @user.login) %>
       <a href="#" title="<%= ts("Pseud Switcher") %>"><%= pseud_link_text %></a>
       <ul class="expandable secondary">
         <%= print_pseud_selector(@user.pseuds.abbreviated_list) %>
--- a/app/views/users/_sidebar.html.erb
+++ b/app/views/users/_sidebar.html.erb
@@ -5,8 +5,7 @@
   <li><%= span_if_current ts("Profile"), user_profile_path(@user) %></li>
   <% if @user.pseuds.size > 1 %>
     <li class="pseud" aria-haspopup="true">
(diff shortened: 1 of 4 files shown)


app/views/users/_sidebar.html.erb
```diff
@@ -5,7 +5,7 @@
   <li><%= span_if_current ts("Profile"), user_profile_path(@user) %></li>
   <% if @user.pseuds.size > 1 %>
     <li class="pseud" aria-haspopup="true">
-      <% pseud_link_text = current_page?(@user) ? ts("Pseuds") : (@pseud ? @pseud.name : @user.login) %>
+      <% pseud_link_text = current_page?(@user) ? ts("Pseuds") : (@pseud ? ( @pseud.name ? @pseud.name : ts("Pseuds") ) : @user.login) %>
       <a href="#" title="<%= ts("Pseud Switcher") %>"><%= pseud_link_text %></a>
```
sarken, line 9:
```suggestion
      <a href="#" title="<%= ts("Pseud Switcher") %>"><%= ts("Pseuds") %></a>
```

The good news is I was right and it is supposed to say Pseuds!

The bad news is, while I was checking the history to see what the logic _used_ to be to make sure I was understanding the ternary correctly, I discovered we actually clobbered #4127/[AO3-6249](https://otwarchive.atlassian.net/browse/AO3-6249) in the initial merge of this (#4554). We want it to say "Pseuds" _all_ the time, not just on the New Pseud and user page.

To make you slightly less inclined to want to murder me, I have submitted it as a suggestion so you can hopefully press the button and not need to do more work, but I'm happy to do penance in the form of opening up a PR if you'd rather.

neuroalien: It sounds like the original-original intention of the code was to have the current pseud on pages that relate to that pseud, and the username was just a fallback for when the `@author` evaporated without us noticing.

I'm not sure the decision to have it say "Pseuds" at all times was reached with consideration for past designers' intentions and what would make the most sense for a user or visitor, or just as an "idk let's put something there, there are bigger fish and they're on fire" measure, but I'm reaching the point of not giving a flying fox anymore. (Which is bad and a sign of burnout, but whatever. The fish are bigger and so's the fire.)

Thanks for the code suggestion!

neuroalien:
@sarken The reason the [test suite fails](https://github.com/otwcode/otwarchive/actions/runs/5567939478/jobs/10170198866#step:13:106) is that the current pseud is no longer guaranteed to show up in the pseud switcher. Are we ok with that?

sarken (approved):
Thank you! At some point in the future, we can discuss changing the should-always-say-Pseuds behavior introduced in [AO3-6249](https://otwarchive.atlassian.net/browse/AO3-6249). This will get us through until then.

PR #4580, opened 2023-07-17
AO3-6552 Login link needs ARIA role

neuroalien:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6552

## Purpose

What does this PR do?

### Add menuitem ARIA role to login link

The containing paragraph has a `menu` role. This requires[1] its direct
children to have one of a few allowed roles, the most appropriate for
this case being `menuitem`.

[1] https://www.w3.org/TR/wai-aria-1.1/#menu

### Add i18n for the login link

Separate commit so it can be easily removed if we don't want to i18n
partially. (I'd prefer not to i18n the entire file in this PR.)

## Testing Instructions

## References

## Credit

they/them pronouns.*

### DIFF
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -31,7 +31,7 @@
   <% else %>
     <div id="login" class="dropdown">
       <p class="user actions" role="menu">
-        <%= link_to ts('Log In', key: 'header'), new_user_session_path, id: "login-dropdown", role: "menuitem" %>
+        <%= link_to t(".login"), new_user_session_path, id: "login-dropdown", role: "menuitem" %>
       </p>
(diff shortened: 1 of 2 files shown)


neuroalien:
> If you have a sec to change the single quotes around the translation key to double quotes, that would be great! If not, this is still 👍

I have lots of secs >.>

I can't even tell anymore what's our preferred style - in some places it's "single quotes unless you need interpolation", in others it's double quotes...

sarken:
Thank you!

Double is our preferred style, unless a) you need to use double inside, in which case you can either escape or use single, or b) you're in a locale file, in which case we are at the whims of the normalizer.

PR #4581, opened 2023-07-17
AO3-6553 Better semantics and a11y for the media list on the homepage

neuroalien:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6553

## Purpose

What does this PR do?

### Turn homepage media list into navigation

Semantically, the list of media (plus All Fandoms) is a navigation
element, not a menu.

The preferred element to use would be `nav`, but we don't use this
anywhere else (yet), so for consistency I'm adding a `div` with the role
navigation.

(Per https://www.w3.org/TR/html-aria/#docconformance, `ul`s are not
allowed to have the role `navigation`.)

### Bonus i18n for media on homepage

This is not strictly related to the a11y issue, but we're trying to add
i18n if we can when we touch a file. (Can remove if needed.)

## Testing Instructions

## References

## Credit

they/them pronouns.*

### DIFF
--- a/app/views/home/_fandoms.html.erb
+++ b/app/views/home/_fandoms.html.erb
@@ -1,0 +1,10 @@
+<ul>
+  <li><%= link_to t(".all_fandoms"), media_path %></li>
+  <% cache "homepage-fandoms-version1", skip_digest: true do %>
+    <% Media.for_menu.each do |medium| %>
+      <% unless medium.id.nil? %>
+        <li><%= link_to ts("#{medium.name}"), medium_fandoms_path(medium) %></li>
+      <% end %>
+    <% end %>
+  <% end %>
+</ul>
--- a/app/views/home/index.html.erb
+++ b/app/views/home/index.html.erb
@@ -5,7 +5,7 @@
 
   <% if logged_in? && @homepage.favorite_tags.present? %>
     <div class="favorite module">
-      <h3 class="heading"><%= ts('Find your favorites') %></h3>
+      <h3 class="heading"><%= t(".find_your_favorites") %></h3>
       <ul>
         <% @homepage.favorite_tags.each do |favorite_tag| %>
           <li>
@@ -15,13 +15,13 @@
       </ul>
     </div>
   <% else %>
-    <div class="browse module">
-      <h3 class="heading"><%= ts('Find your favorites') %></h3>
+    <nav aria-label="<%= t(".media_navigation_label") %>" class="browse module">
+      <h3 class="heading"><%= t(".find_your_favorites") %></h3>
       <% if logged_in? %>
-        <p class="note"><%= ts('Browse fandoms by media or favorite up to %{maximum} tags to have them listed here!', maximum: ArchiveConfig.MAX_FAVORITE_TAGS) %></p>
+        <p class="note"><%= t(".browse_or_favorite", count: ArchiveConfig.MAX_FAVORITE_TAGS) %></p>
       <% end %>
-      <%= render 'menu/menu_fandoms' %>
-    </div>
+      <%= render 'fandoms' %>
+    </nav>
   <% end %>
 
   <% if @homepage.admin_posts.present? %>
--- a/public/stylesheets/site/2.0/26-media-narrow.css
+++ b/public/stylesheets/site/2.0/26-media-narrow.css
@@ -182,7 +182,7 @@ body .narrow-shown {
   padding: 0;
 }
 
-.splash div.module, .logged-in .splash div.module {
+.splash div.module, .splash nav.module, .logged-in .splash div.module, .logged-in .splash nav.module {
   clear: both;
   margin-left: 0;
   margin-right: 0;
--- a/config/locales/views/en.yml
+++ b/config/locales/views/en.yml
@@ -429,6 +429,14 @@ en:
   home:
     donate:
       page_title: Donate or Volunteer
+    fandoms:
+      all_fandoms: All Fandoms
+    index:
+      browse_or_favorite:
+        one: Browse fandoms by media or favorite up to %{count} tag to have it listed here!
+        other: Browse fandoms by media or favorite up to %{count} tags to have them listed here!
+      find_your_favorites: Find your favorites
+      media_navigation_label: Media
   invitations:
     invitation:
       email_address_label: Enter an email address


app/views/menu/_menu_fandoms.html.erb
```diff
@@ -1,10 +1,12 @@
...
-    <% Media.for_menu.each do |medium| %>
-      <% unless medium.id.nil? %>
-        <li id="medium_<%= medium.id %>"><%= link_to ts("#{medium.name}", key: 'header.fandom'), medium_fandoms_path(medium) %></li>
+<div role="navigation" aria-label="<%= t(".media") %>">
+  <ul class="menu">
+    <li><%= link_to t(".all_fandoms"), media_path %></li>
+    <% cache "menu-fandoms-version4", skip_digest: true do %>
+      <% Media.for_menu.each do |medium| %>
+        <% unless medium.id.nil? %>
+          <li id="medium_<%= medium.id %>"><%= link_to t(".#{medium.translation_key}"), medium_fandoms_path(medium) %></li>
```
sarken, line 7: You can ignore this comment; it's observational rather than something you need to action.

This poses a thrilling question for future!us: media are tags. Should we be translating tag names?

It makes sense here because it's site navigation more than a tag, imo, but will we want to show the localized media name, e.g., in the Parent Tags section on a [fandom's tag landing page](https://archiveofourown.org/tags/Broadchurch)? 

I kind of feel like that page should list the tag as-is (because I doubt we'll translate fandoms in the Parent Tags section, e.g., Good Omens (TV) on [Crowley's landing page](https://archiveofourown.org/tags/Crowley%20(Good%20Omens))), how important is it that the site navigation and the parent tags match? 🤔

sarken (changes requested):

app/views/home/_fandoms.html.erb
```diff
@@ -0,0 +1,10 @@
+<ul class="menu">
```
sarken, line 1: I think you can take the class off in addition to the role -- it doesn't seem to be used for the styling and doesn't make much sense here anyway.

app/views/home/_fandoms.html.erb
```diff
@@ -0,0 +1,10 @@
...
+  <li><%= link_to t(".all_fandoms"), media_path %></li>
+  <% cache "homepage-fandoms-version1", skip_digest: true do %>
+    <% Media.for_menu.each do |medium| %>
+      <% unless medium.id.nil? %>
+        <li id="medium_<%= medium.id %>"><%= link_to t(".#{medium.translation_key}"), medium_fandoms_path(medium) %></li>
```
sarken, line 6: Can you take the `id` attribute off? I'm not really sure what we use it for, if anything, in the dropdown menus, but having the same `id` in both places makes for invalid HTML because IDs have to be unique. I just made a mess all over the place with my genius reuse of that partial. 🙃

app/views/home/index.html.erb
```diff
@@ -15,13 +15,13 @@
       </ul>
     </div>
   <% else %>
-    <div class="browse module">
-      <h3 class="heading"><%= ts('Find your favorites') %></h3>
+    <nav aria-label="<%= t(".media_navigation_label") %>" class="browse module">
+      <h3 class="heading"><%= t(".find_your_favorites") %></h3>
       <% if logged_in? %>
-        <p class="note"><%= ts('Browse fandoms by media or favorite up to %{maximum} tags to have them listed here!', maximum: ArchiveConfig.MAX_FAVORITE_TAGS) %></p>
+        <p class="note"><%= t(".browse_or_favorite", maximum: ArchiveConfig.MAX_FAVORITE_TAGS) %></p>
```
sarken, line 21: I think we need to pass `ArchiveConfig.MAX_FAVORITE_TAGS` as `count` so Translation can have multiple versions of this sentence ready in case we ever change `MAX_FAVORITE_TAGS`.

spec/models/media_spec.rb
```diff
@@ -0,0 +1,10 @@
+require "spec_helper"
+
+describe Media do
+  describe "#translation_key" do
+    it "returns a parameterized version of the media name using underscore as a separator" do
+      media = FactoryBot.create(:media, name: "Widgets; Tools & Doodads")
```
sarken, line 6: Will `create(:media, name: etc)` work without the `FactoryBot.` here?

public/stylesheets/site/2.0/26-media-narrow.css
```diff
@@ -182,7 +182,10 @@ body .narrow-shown {
   padding: 0;
 }
 
-.splash div.module, .logged-in .splash div.module {
```
sarken, line 185: Could you put this back on one line like the other selectors, please? (Yeah, it's easier to see a diff if it's already spread across lines, but let's go for consistency until we do a bulk change.)

sarken (commented):
Just nitpicks and stuff I screwed up 100 years ago, nothing actually wrong with the accessibility change itself!

app/views/home/index.html.erb
Bilka2: I think sarken means this: https://guides.rubyonrails.org/i18n.html#pluralization

neuroalien: Thank you! Hashtag today-I-learned :D

spec/models/media_spec.rb
neuroalien: It did!

app/views/home/index.html.erb
sarken: Yup, that's what I meant! 

When we use `count`, we generally define `one:` and `other:` English translations to go with it. It's _pretty_ unlikely we'll ever drop `MAX_FAVORITE_TAGS` down to one, but if you could update `en.yml` to have both just to be safe, this will be ready to merge:

```
    index:
      browse_or_favorite:
        one: Browse fandoms by media or favorite up to %{count} tag to have them listed here!
        other: Browse fandoms by media or favorite up to %{count} tags to have them listed here!
```

neuroalien: Ahaha, I was coming here to nitpick that very thing :D 

(You have to admit that "favourite up to 1 tag" sounds pretty sad though, and the "have it listed here" only adds to the desolation...)

sarken: It really does. Something like "or favorite a tag/choose a favorite tag" sounds a bit less sad, or at least less awkward, but it's probably not worth getting bogged down in wording that shouldn't see the light of day unless Catastrophe™  strikes. (Unless you want to! I don't mind workshopping wording; I just don't want you to feel obligated.)

sarken (approved):

app/views/home/_fandoms.html.erb
```diff
@@ -0,0 +1,10 @@
+<ul>
+  <li><%= link_to t(".all_fandoms"), media_path %></li>
+  <% cache "homepage-fandoms-version1", skip_digest: true do %>
+    <% Media.for_menu.each do |medium| %>
+      <% unless medium.id.nil? %>
+        <li><%= link_to t(".#{medium.translation_key}"), medium_fandoms_path(medium) %></li>
```
tickinginstant, line 6: For whatever it's worth, I'm still team "don't feed this through i18n this until the caching questions from https://github.com/otwcode/otwarchive/pull/4584#discussion_r1272656196 are settled." This is no longer an active bug, since you can't set your locale on the homepage yet, but it could become a bug later if we don't opt to override the `cache` function.

(Also, I think [mobility](https://github.com/shioyama/mobility) or [globalize](https://github.com/globalize/globalize) would be a better choice for internationalizing medium names. As is, this would make it more difficult for other sites using the `otwarchive` software to use their own custom media names.)

neuroalien: I don't have the spoons to look into alternate ways of translating, so I've removed the contentious commit. We can cross or burn that bridge when we come closer to it. ;)

Thank you all for reviewing!

tickinginstant (approved):

PR #4582, opened 2023-07-18
AO3-6467 Record Fannish Next of Kin changes in user history 

ceithir:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6467

## Purpose

Track and display changes to FNOK in admin, on both sides of the relationship

## Testing Instructions

See JIRA

## References

~~[Alterity](https://github.com/gumroad/alterity) should supposedly take care of all the `pt-online-schema-change` automagically.~~
Obsolete, see https://github.com/otwcode/otwarchive/pull/4582#discussion_r1266609404

## Credit

Ceithir (he/him)


Gemfile
```diff
@@ -176,6 +176,7 @@ gem 'unicorn', '~> 5.5', require: false
 gem 'god', '~> 0.13.7'
 
 group :staging, :production do
+  gem "alterity"
```
ceithir, line 179: Big :crossed_fingers: energy here. Not sure how to test that properly in close to real life conditions.

config/i18n-tasks.yml
```diff
@@ -166,6 +166,10 @@ ignore_missing:
   - users_helper.log_troubleshot
   - users_helper.log_validated
   - users_helper.log_warn
+  - users_helper.log_add_fnok
+  - users_helper.log_remove_fnok
+  - users_helper.log_added_as_fnok
+  - users_helper.log_removed_as_fnok
```
ceithir, line 172: Not really sure why we do that for these keys instead of adding them to `config/locales/helpers/en.yml`. Kept current behavior for now but maybe something to change?

sarken: For a while, we were considering using `default` when doing i18n, but we later changed our minds and just haven't fixed the places where we used that approach. It's fine to leave it for now, though.

Gemfile
sarken: We already have a pull request to add Departure instead (#4528), which I believe serves the same purpose, so it's probably a good idea to leave this out.

ceithir: Ok. Since this ticket is of a higher priority, I've added the `pt-online-schema-change` syntax to this PR so it isn't blocked until the gem is added.

app/helpers/users_helper.rb
```diff
@@ -119,31 +119,66 @@ def authored_items(pseud, work_counts = {}, rec_counts = {})
...
+          user_id: item.fnok_user_id
+        )
+      end
+
+      return t(
+        "users_helper.log_add_fnok",
+        default: "Fannish Next of Kin Added: %{user_id}",
+        user_id: item.fnok_user_id
+      )
+    end
```
sarken, line 155: I think `log_item_action_name` is getting a little long with this. Could you move the logic for figuring out how to log FNOK actions into a separate method? 

It would also be a good idea to use our regular, `default`-less method of i18n here, even if we're leaving the `default` for the existing code.

features/admins/users/admin_fnok.feature
```diff
@@ -15,6 +15,7 @@ Feature: Admin Fannish Next Of Kind actions
       And I fill in "Fannish next of kin's email" with "testy@foo.com"
       And I press "Update Fannish Next of Kin"
     Then I should see "Fannish next of kin was updated."
+    Then I should see "Fannish Next of Kin Added"
```
sarken, line 18: This is going to be slightly annoying to do, but I think we need to be testing the entire string here to make sure we're including the right user ID on the right page. Please forgive my pseudo code, but maybe something like:

```
# Usage on harrykim's user admin page: 
# - Then adding libby as next of kin should be logged in the history table
# - Then removing libby as next of kin should be logged in the history table
Then "{string} {string} as next of kin should be logged in the history table" do |action, fnok_username|
  fnok_id = User.find_by(login: fnok).id
  step %{I should see "Fannish Next of Kin #{action == "adding" ? "Added" : "Removed"}: #{fnok_id}" within "#id_of_history_table"}
end

# Usage on libby's user admin page: 
# - Then the history table should show they were added as harrykim's next of kin
# - Then the history table should show they were removed as harrykim's next of kin
Then "the history table should show they were {string} as {string}'s next of kin" do |action, username|
  user_id = User.find_by(login: username).id
  step %{I should see "#{action == "added" ? "Added" : "Removed"} as Fannish Next of Kin for: #{user_id}" within "#id_of_history_table"}
 end
 ```

It looks like we're also missing any sort of test for when the FNOK (in this case, libby) deletes their account. When that happens, the user's (harrykim's) page should say "Removed Fannish Next of Kin: libby's user ID" with a note that it was "System Generated" rather than being done by an admin.

**Review by sarken — CHANGES_REQUESTED** [review]

File: spec/models/user_spec.rb [hunk 5]
```diff
@@ -15,6 +15,23 @@
         end
       end
     end
+
+    context "when the user is set as someone else's fnok" do
+      let(:fnok) { create(:fannish_next_of_kin) }
+      let(:user) { fnok.kin }
+      let(:person) { fnok.user }
+
+      it "explictly removes the relationship" do
```
sarken, line 24: Typo -- there's another `i` in `explicitly` -- but let's be a little more descriptive and say something like `removes the relationship and creates a log item of the removal`.

app/helpers/users_helper.rb
```diff
@@ -119,34 +119,51 @@ def authored_items(pseud, work_counts = {}, rec_counts = {})
...
-    elsif action == ArchiveConfig.ACTION_TROUBLESHOOT
-      t('users_helper.log_troubleshot', default: 'Account Troubleshot')
-    elsif action == ArchiveConfig.ACTION_NOTE
-      t('users_helper.log_note', default: 'Note Added')
+  def log_item_action_name(item, user)
+    action = item.action
+    
+    return fnok_action_name(item, user) if [ArchiveConfig.ACTION_ADD_FNOK, ArchiveConfig.ACTION_REMOVE_FNOK].include?(action)
+
+    case action
```
sarken, line 127: Thanks for updating these!

sarken (changes requested):

sarken (approved):
Thank you!

PR #4584, opened 2023-07-24
AO3-4309 Internationalize batch subscription notification

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4309

## Purpose

Make it so we can translate the subscription email. This includes

* Move the code for generating its subject line out of the subscription model and into the mailer helper, since it did not seem like code responsible for generating text should be there.
* Internationale the text generated by the `chapter_header` method, which remains in the chapter model.
* Allow `anonymous?` to be called on a chapter so we can remove the `anonymous_creation?(creation)` method from the subscription model and simply call `anonymous?` on any creation.
* Add `valid_entry_for_subscription?(creation)` method to the subscription model, so I could both mov the existing "should we skip this entry?" logic out of the mailer and extend it. Example: the mailer specs were including a work entry in a notification for a work subscription, which should never happen. Because this should never happen, there was no subject translation to cover a subject line simply reading "Anonymous posted Work Title," and thus the tests failed. 
This test has been updated.
** I initially added `return unless valid_entry_for_subscription?(creation)` to the new subject generation method as well, since it's still possible to pass that method an impossible combination of subscription and creation (i.e., again, work subscription and work creation) in tests or by calling it somewhere outside its intended use. However, I took it back out because it seemed silly to run the same check twice (once in the mailer, once in the subject line). I could put it back in if we think that's smart, though.
* Add two `creation_attribution_` helper methods for generating the text "by Anonymous" or "by [creator or creators]" for the mailers, since Translation needs to be able to translate "by Anonymous" as one word. I removed the anonymous logic from the existing `creator_` helper methods and updated the views for the mailer where they were being used to handle anonymous (the gift notification).
* Update the `work_info` partial so you can optionally suppress the summary, allowing the existing code to be reused in the subscription mailer.
* Move the logic for the preface of the email (the part that tells you who posted what) into the helpers. It doesn't result in fewer lines of code, but it's easier to test that it produces the desired results.

The tests for `batch_subscription_subject` are pretty verbose, but that is for a reason. I wanted the output to be more readable than "akghlkjalk (pohgalhga) and alkghalteiu (athwtwapjg) posted My title is long enough in the Awesome Series 16 series and 2 more" so I wouldn't need to write descriptions for each one (which I think would ultimately involve even more lines of code). However, when I tried making the works and chapters and such more reusable, it quickly turned into a tangled mess of overwriting the work variable to means something subtly different in one context than another, too many before blocks and reloads, etc etc. 

If someone has an idea for how to make them shorter and easier to read while still hitting every possible subject variation, please go right ahead! (Note that I did not test any scenarios that do not have translations. If we want to do that, it's probably best to add the `return unless valid_entry_for_subscription?(creation)` guard clause and then check that the method produces `nil`.)

## Testing Instructions

Refer to Jira. (Which tbh needs a lot more instructions if we want to test every possible combination of subject line and email content.)

### DIFF
--- a/app/helpers/mailer_helper.rb
+++ b/app/helpers/mailer_helper.rb
@@ -133,26 +133,23 @@ module MailerHelper
       creation_title: title, word_count: creation_word_count(creation))
   end
 
-  # The bylines used in subscription emails to prevent exposing the name(s) of
-  # anonymous creator(s).
+  # TODO: Update all mailers to use creator_links or creator_text as
+  # appropriate to eliminate pseuds.map { etc }.to_sentence in the views. Note
+  # that these should never be used on anonymous creations. Text dealing with
+  # Anonymous creations should always have a separate translation, as in the
+  # creation_attribution_ and batch_subscription_subject methods below.
   def creator_links(creation)
-    if creation.anonymous?
-      "Anonymous"
-    else
-      creation.pseuds.map { |p| style_pseud_link(p) }.to_sentence.html_safe
-    end
+    creation.pseuds.map { |p| style_pseud_link(p) }.to_sentence.html_safe
   end
 
   def creator_text(creation)
-    if creation.anonymous?
-      "Anonymous"
-    else
-      creation.pseuds.map { |p| text_pseud(p) }.to_sentence.html_safe
-    end
+    creation.pseuds.map { |p| text_pseud(p) }.to_sentence.html_safe
   end
 
-  # by name, which appears under or after creation titles
-  # plain text if anonymous
+  # "by name" or "by name, name, and name", which appears under or after
+  # creation titles.
+  # It's always plain text if the creation is anonymous, so we share that
+  # translation between two methods.
   def creation_attribution_links(creation)
     if creation.anonymous?
       t("mailer.general.creation.attribution.anon")
--- a/app/helpers/mailer_helper.rb
+++ b/app/helpers/mailer_helper.rb
@@ -248,6 +248,28 @@ module MailerHelper
     t(computed_key, **variables)
   end
 
+  def batch_subscription_text_preface(creation)
+    work = creation.is_a?(Chapter) ? creation.work : creation
+
+    variables = {}
+    variables[:creators] = creator_text(work) unless creation.anonymous?
+    variables[:work_title_with_word_count] = creation_title_with_word_count(creation.work) unless creation.is_a?(Work)
+    variables[:count] = creation.pseuds.size unless creation.anonymous?
+
+    t(batch_subscription_preface_key(creation, email_format: "text"), **variables)
+  end
+
+  def batch_subscription_html_preface(creation)
+    work = creation.is_a?(Chapter) ? creation.work : creation
+
+    variables = {}
+    variables[:creator_links] = creator_links(work) unless creation.anonymous?
+    variables[:work_link_with_word_count] = creation_link_with_word_count(creation.work, work_url(creation.work)) unless creation.is_a?(Work)
+    variables[:count] = creation.pseuds.size unless creation.anonymous?
+
+    t(batch_subscription_preface_key(creation, email_format: "html"), **variables)
+  end
+
   private
 
   # e.g., 1 word or 50 words
@@ -286,5 +308,29 @@ module MailerHelper
     else
       work_tag_metadata_list(tags)
     end
+  end
+
+  def batch_subscription_preface_key(creation, email_format:)
+    work = creation.is_a?(Chapter) ? creation.work : creation
+
+    base_key = "user_mailer.batch_subscription_notification.preface"
+    creator_key = creation.anonymous? ? "anon" : "named"
+    creation_key = creation.is_a?(Chapter) ? "chapter" : "work"
+    dating_key = creation.backdate ? ".backdated" : ".new" if creation.is_a?(Work)
+    format_key = ".#{email_format}" unless creation.is_a?(Work) && creation.anonymous?
+
+    # Note the lack of . before dating and format keys, which are not always
+    # included.
+    # i18n-tasks-use t("user_mailer.batch_subscription_notification.preface.anon.chapter.html")
+    # i18n-tasks-use t("user_mailer.batch_subscription_notification.preface.anon.chapter.text")
+    # i18n-tasks-use t("user_mailer.batch_subscription_notification.preface.anon.work.backdated")
+    # i18n-tasks-use t("user_mailer.batch_subscription_notification.preface.anon.work.new")
+    # i18n-tasks-use t("user_mailer.batch_subscription_notification.preface.named.chapter.html")
+    # i18n-tasks-use t("user_mailer.batch_subscription_notification.preface.named.chapter.text")
+    # i18n-tasks-use t("user_mailer.batch_subscription_notification.preface.named.work.backdated.html")
+    # i18n-tasks-use t("user_mailer.batch_subscription_notification.preface.named.work.backdated.text")
+    # i18n-tasks-use t("user_mailer.batch_subscription_notification.preface.named.work.new.html")
+    # i18n-tasks-use t("user_mailer.batch_subscription_notification.preface.naemd.work.new.text")
+    computed_key = "#{base_key}.#{creator_key}.#{creation_key}#{dating_key}#{format_key}"
   end
 end # end of MailerHelper
--- a/app/helpers/mailer_helper.rb
+++ b/app/helpers/mailer_helper.rb
@@ -311,8 +311,6 @@ module MailerHelper
   end
 
   def batch_subscription_preface_key(creation, email_format:)
-    work = creation.is_a?(Chapter) ? creation.work : creation
-
     base_key = "user_mailer.batch_subscription_notification.preface"
     creator_key = creation.anonymous? ? "anon" : "named"
     creation_key = creation.is_a?(Chapter) ? "chapter" : "work"
@@ -331,6 +329,6 @@ module MailerHelper
     # i18n-tasks-use t("user_mailer.batch_subscription_notification.preface.named.work.backdated.text")
     # i18n-tasks-use t("user_mailer.batch_subscription_notification.preface.named.work.new.html")
     # i18n-tasks-use t("user_mailer.batch_subscription_notification.preface.naemd.work.new.text")
-    computed_key = "#{base_key}.#{creator_key}.#{creation_key}#{dating_key}#{format_key}"
+    "#{base_key}.#{creator_key}.#{creation_key}#{dating_key}#{format_key}"
   end
 end # end of MailerHelper
--- a/app/helpers/mailer_helper.rb
+++ b/app/helpers/mailer_helper.rb
@@ -207,6 +207,7 @@ module MailerHelper
     work = creation.is_a?(Chapter) ? creation.work : creation
     series = subscription.subscribable if subscribable_type == "series"
     creator_list = creation.pseuds.map(&:byline).to_sentence unless creation.anonymous?
+    # For pluralization: creator publicó, creator y creator2 publicaron.
     creators_count = creation.pseuds.size unless creation.anonymous?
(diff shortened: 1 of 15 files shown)


app/helpers/mailer_helper.rb
```diff
@@ -176,6 +191,63 @@ def style_work_tag_metadata(tags)
...
+  # otherwise pass invalid or anonymity-breaking combinations of data to this.
+  def batch_subscription_subject(subscription, creation, additional_creations_count)
+    subscribable_type = subscription.subscribable_type.downcase
+
+    creation_type = creation.is_a?(Chapter) ? "chapter" : "work"
+
+    work = creation.is_a?(Chapter) ? creation.work : creation
+    series = subscription.subscribable if subscribable_type == "series"
+    creator_list = creation.pseuds.map(&:byline).to_sentence unless creation.anonymous?
+    # For pluralization: creator publicó, creator y creator2 publicaron.
```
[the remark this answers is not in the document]
sarken: Let the record show I know it's wrong without the accent, even though I am changing it.

app/helpers/mailer_helper.rb
```diff
@@ -133,21 +133,36 @@ def creation_title_with_word_count(creation)
...
-  def creator_links(work)
-    if work.anonymous?
-      "Anonymous"
+  # TODO: Update all mailers to use creator_links or creator_text as
+  # appropriate to eliminate pseuds.map { etc }.to_sentence in the views. Note
+  # that these should never be used on anonymous creations. Text dealing with
+  # Anonymous creations should always have a separate translation, as in the
+  # creation_attribution_ and batch_subscription_subject methods below.
+  def creator_links(creation)
+    creation.pseuds.map { |p| style_pseud_link(p) }.to_sentence.html_safe
```
tickinginstant, line 142: If you use the `ActionView::Helpers::OutputSafetyHelper` version of `to_sentence` here, you get the `html_safe` for free:
```suggestion
    to_sentence(creation.pseuds.map { |p| style_pseud_link(p) })
```
(Side note: I think it'd be good to re-enable the `Rails/OutputSafety` cop, so that people are more aware of when they're calling an unsafe function like `html_safe`, and will hopefully look for alternatives. Ultimately, if you want to fix the big parser issues without introducing XSS exploits, someone is going to need to examine every single call to `html_safe` to make sure that it's actually safe once things aren't being sanitized on input, so the fewer of them there are, the better.)

app/helpers/mailer_helper.rb
```diff
@@ -133,21 +133,36 @@ def creation_title_with_word_count(creation)
...
+  end
+
+  def creator_text(creation)
+    creation.pseuds.map { |p| text_pseud(p) }.to_sentence.html_safe
```
tickinginstant, line 146: And here:
```suggestion
    to_sentence(creation.pseuds.map { |p| text_pseud(p) })
```

app/helpers/mailer_helper.rb
```diff
@@ -176,6 +191,85 @@ def style_work_tag_metadata(tags)
...
+  # to user X, who has co-created a work with user Y, and Y posts a chapter that
+  # is not co-created, the subject line will say just "Y posted" even though the
+  # subscription is for X.
+  # Note that we assume this is being used in batch_subscription_notification,
+  # after a subscription.valid_notification_entry?(creation) check. You can
+  # otherwise pass invalid or anonymity-breaking combinations of data to this.
+  def batch_subscription_subject(subscription, creation, additional_creations_count)
+    subscribable_type = subscription.subscribable_type.downcase
+
+    creation_type = creation.is_a?(Chapter) ? "chapter" : "work"
```
tickinginstant, line 205: You can simplify this by calling `creation.model_name.i18n_key`, here and elsewhere.

app/helpers/mailer_helper.rb
```diff
@@ -176,6 +191,85 @@ def style_work_tag_metadata(tags)
...
+    # i18n-tasks-use t("user_mailer.batch_subscription_notification.subject.named.series.work.one_entry")
+    # i18n-tasks-use t("user_mailer.batch_subscription_notification.subject.named.work.multiple_entries")
+    # i18n-tasks-use t("user_mailer.batch_subscription_notification.subject.named.work.one_entry")
+    computed_key = "#{base_key}.#{creator_key}.#{creation_key}.#{entries_key}"
+
+    # "and X more," translated separately so we can pluralize "more" based on X.
+    # i18n-tasks-use t("user_mailer.batch_subscription_notification.subject.more")
+    more_translation = t("#{base_key}.more", count: additional_creations_count) unless additional_creations_count.zero?
+
+    variables = {}
```
tickinginstant, line 239: Hmm, not a big fan of `variables` as a name here — it feels a little generic. `interpolations`, maybe? Or `translation_options`?

app/helpers/mailer_helper.rb
```diff
@@ -176,6 +191,85 @@ def style_work_tag_metadata(tags)
...
+    variables[:app_name] = ArchiveConfig.APP_SHORT_NAME
+    variables[:creators] = creator_list unless creation.anonymous?
+    variables[:count] = creators_count unless creation.anonymous?
+    variables[:chapter_header] = chapter_header if creation_type == "chapter"
```
tickinginstant, line 243: You're checking the exact same conditions here as the initial lines that set these three variables (`creator_list`, `creators_count`, `chapter_header`). Would it make sense to set `variables = {}` earlier, so that you can add the values to the hash instead of setting the variables? Or skip the conditions here, so that the option is set to `nil`?

app/helpers/mailer_helper.rb
```diff
@@ -215,4 +309,26 @@ def style_work_tag_metadata_list(tags)
...
+
+  def batch_subscription_preface_key(creation, email_format:)
+    base_key = "user_mailer.batch_subscription_notification.preface"
+    creator_key = creation.anonymous? ? "anon" : "named"
+    creation_key = creation.is_a?(Chapter) ? "chapter" : "work"
+    dating_key = creation.backdate ? ".backdated" : ".new" if creation.is_a?(Work)
+    format_key = ".#{email_format}" unless creation.is_a?(Work) && creation.anonymous?
+
+    # Note the lack of . before dating and format keys, which are not always
+    # included.
```
tickinginstant, line 321: Rather than worrying about when to insert a `.`, would it maybe make sense to construct the key as an array and join it at the end? e.g.
```ruby
key = [:user_mailer, :batch_subscription_notification, :preface]
key << (creation.anonymous? ? :anon : :named)
key << creation.model_name.i18n_key
key << (creation.backdate ? :backdated : :new) if creation.is_a?(Work)
key << email_format unless creation.is_a?(Work) && creation.anonymous?
key.join(".")

File: app/models/subscription.rb [hunk 8]
```diff
@@ -35,19 +35,27 @@ def name
...
+  # or chapters, or vice versa.
+  def valid_notification_entry?(creation)
+    return false unless creation.is_a?(Chapter) || creation.is_a?(Work)
+    return false unless creation.try(:posted)
+    return false if creation.is_a?(Chapter) && !creation.work.try(:posted)
+    return false if creation.pseuds.any? { |p| p.user == User.orphan_account }
+    return false if subscribable_type == "User" && creation.anonymous?
+    return false if subscribable_type == "Work" && !creation.is_a?(Chapter)
+    return false if subscribable.respond_to?(:anonymous?) &&
+                    subscribable.anonymous? != creation.anonymous?
```
tickinginstant, line 57: Are these checks new? If so, it'd be good to add some QA steps for them.

(I'm also not convinced that this is the right behavior for anonymous series. A series is marked as anonymous if any of the works in the series are anonymous, but until the behavior outlined [here](https://otwarchive.atlassian.net/browse/AO3-1250?focusedCommentId=333510) is implemented, users are still allowed to mix anonymous and non-anonymous works in a single series, so I think the notifications should reflect that. If you want to get a head start on AO3-1250, I suppose you could suppress all series notifications for changes to anonymous works, but this does the opposite.)

app/views/user_mailer/batch_subscription_notification.html.erb
```diff
@@ -4,62 +4,31 @@
...
+    <% creation_url = creation.is_a?(Work) ? work_url(this_work) : work_chapter_url(this_work, creation) %>
 
-    <% cache("subscription-email-preface-#{creation.cache_key}") do %>
-      <%= creator_links(this_work) %> posted a
-      <% if creation.is_a?(Work) %>
-        <%= this_work.backdate ? "backdated" : "new" %> work:
-      <% else %>
-        new chapter of <i><b><%= work_link %></b></i> (<%= this_work.word_count %> words):
-      <% end %>
+    <% cache("/v1/subscription-email-preface-#{creation.cache_key}") do %>
```
tickinginstant, line 9: As I've mentioned previously, caching and internationalization don't play well together. So far, the translations have primarily been limited to emails, and the other emails don't use caching anywhere. But obviously this does.

I think you need to make some general decisions about how you want to handle the combination of caching and localization. One way to handle it would be to override the `cache` function so that it always adds `I18n.locale` to the cache key. This would prevent (some) bugs like the ones you've introduced here, at the cost of going a little overboard — it means that even if there's a cache block that doesn't have any localized text in it, you won't be able to share the cache between locales. Also, if you do this it will expire all existing `cache` blocks, which means that the deploy for this change would be very rough on the database — might be good to schedule it for a low period.

The other alternative is to add `I18n.locale` to the cache key of each cache block that contains localized text. This allows for more control over the caching, but forces developers to be aware of whether the views that they're modifying are cached, and whether any of the functions they're using inside the cache block return localized text.

(Another note on internationalization and caching: enabling internationalized emails will generate a bunch of extra work when sending subscriptions, because the info for each work may have to be regenerated an extra time for each extra locale. So you may want to do some optimization on the `batch_subscription_notification` function before you enable locales for everyone. Actually, you should probably also do at least a little optimization before this is deployed, since the `work_info` partial uses 6 more database queries per work than the existing code.)

(Another side note: you don't need to add a version number to the cache key here, since you're changing the contents of the cache block, and this doesn't have `skip_digest: true`.)

app/models/subscription.rb
sarken: They are new, yup, and I'll add steps make sure the "you shouldn't get emails" scenarios are covered in testing as well as the "you should get emails." 

And oh my god, I thought we'd prevented mixed anonymous/non-anonymous series _years_ ago. (I remember a specific volunteer who did a lot of testing around it and they haven't been around for... maybe six years or more?) Thank you for telling me it is still to-do.

sarken:
Thanks for the review, ticking! I'm going to close this until we figure out what to do about caching, but what were you using to count the number of database queries in the mailer view files so I can figure out how to get that number down?

tickinginstant:
I wasn't using any tools, just looking at the file. I cooked up an `n_plus_one` test case to try to help, and discovered that switching to `work_info` actually uses ~20 more database queries, not 6. So I wasn't even estimating correctly before.

The thing that stood out to me when I read the code is that the various `*_string` methods use the `tags` association, so you only need a single database query for all of them (more or less). Whereas the `work_info` partial uses the individual `ratings`, `categories`, etc. associations, each of which is a separate ActiveRecord CollectionProxy that must be loaded with its own database query. I thought it was one database query per tag type, but it turns out it's three.

This is the `n_plus_one` test case that I made:
```ruby
describe UserMailer, type: :mailer do
  describe "batch_subscription_notification" do
    let!(:creator) { create(:user) }
    let!(:subscription) { create(:subscription, subscribable: creator) }

    context "when notifying about multiple works", :n_plus_one do
      populate { |n| create_list(:work, n, authors: [creator.default_pseud]) }

      it "performs ??? queries" do
        expect do
          creation_info = Work.ids.map { |id| "Work_#{id}" }.to_json
          UserMailer.batch_subscription_notification(subscription.id, creation_info).deliver_now
        end.to perform_linear_number_of_queries(slope: 0)
      end
    end
  end
end
```
If you tune the slope to match the existing number of queries per work, that should help catch changes that increase the number of database queries per work.

PR #4586, opened 2023-07-25
AO3-6567 Use % for translate_string instead of I18n.translate.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6567

## Purpose

Tries to improve the `translate_string`/`ts` function by making it use `%` instead of `I18n.translate`.

### DIFF
--- a/app/views/tag_set_nominations/_tag_nominations_by_fandom.html.erb
+++ b/app/views/tag_set_nominations/_tag_nominations_by_fandom.html.erb
@@ -1,14 +1,14 @@
 <% @tag_set_nomination.fandom_nominations.each_with_index do |nom, index| %>
   <%= f.fields_for :fandom_nominations, nom do |nom_form| %>
     <fieldset class="tagset">
-      <legend><%= ts("Fandom %{index}", :index => index+1) %></legend>
+      <legend><%= ts("Fandom %{index}", index: index + 1) %></legend>
       <dl>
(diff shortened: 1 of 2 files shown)


app/views/tag_set_nominations/_tag_nominations_by_fandom.html.erb
```diff
@@ -1,14 +1,14 @@
 <% @tag_set_nomination.fandom_nominations.each_with_index do |nom, index| %>
   <%= f.fields_for :fandom_nominations, nom do |nom_form| %>
     <fieldset class="tagset">
-      <legend><%= ts("Fandom %{index}", :index => index+1) %></legend>
+      <legend><%= ts("Fandom %{index}", index: index + 1) %></legend>
```
sarken, line 4: Question! Why did this need to be changed but not `:index => index+1` on line 6?

tickinginstant: Oh, it doesn't! I just changed the syntax here because I looked at this line to double-check that I had the right index number for line 11, and noticed the hash-rocket syntax (and spacing). I never noticed that line 6 had a similar issue.

I can revert this line or fix line 6, whichever you prefer.

sarken: If you could fix line 6, too, that'd be great!

sarken (approved):

zz9pzza:
image
A very nice improvement

PR #4590, opened 2023-08-02
AO3-6566 ARIA controls must contain the exact ID

neuroalien:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6566

## Purpose

What does this PR do?

Correctly declare the ID of the element controlled by the link to modal.

The ID of the modal controlled by the link to modal is `modal`. The hash sign must not be included in the reference to the element.

## Testing Instructions

## References

## Credit

they/them pronouns.*


app/helpers/application_helper.rb
```diff
@@ -150,7 +150,7 @@ def link_to_modal(content = "", options = {})
     options[:for] ||= ""
     options[:title] ||= options[:for]
 
-    html_options = { "class" => options[:class] + " modal", "title" => options[:title], "aria-controls" => "#modal" }
+    html_options = { "class" => "#{options[:class]} modal", "title" => options[:title], "aria-controls" => "modal" }
```
sarken, line 153: Does this still need the old hash rocket syntax or does `html_options = { class: "#{options[:class]} modal", title: options[:title], aria: { controls: "modal" } }` work?

And it's okay if you don't know the answer to this, but any idea why we set the `aria-controls` attribute here and in the JavaScript? It seems like the JavaScript is overwriting this: if I just change it here and I check the modals in, e.g., the symbols block on blurbs or anywhere on the search pages, the modal links still have whatever `aria-controls` is set in the JavaScript.

sarken (commented):
Question/suggestion, but otherwise looks good!

app/helpers/application_helper.rb
neuroalien: I'm pretty sure there's no need for the old syntax for the first two keys, but I didn't know you could nest aria keys like that. I also wasn't sure if we wanted to do piecemeal syntax updates, or if that would open a whole can of worms. 😅 

Is it meant to work without JavaScript? No clue. 🤷

sarken: Without JavaScript, the modal definitely won't work -- it should just fall back to being a plain link, but I haven't tested it to see if that's wishful linking -- so including the ARIA attribute is pretty much unnecessary.

neuroalien: Updated!

sarken (approved):

PR #4596, opened 2023-08-11
AO3-6548 Add "(Guest)" to guest comments

EchoEkhi:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6548

## Purpose

Add "(Guest)" to guest names in guest comments in the web UI and in emails

## Testing Instructions

See JIRA ticket

## Credit

EchoEkhi (he/him)


app/helpers/mailer_helper.rb
```diff
@@ -176,6 +180,30 @@ def style_work_tag_metadata(tags)
     "#{label}#{style_work_tag_metadata_list(tags)}".html_safe
   end
 
+  def commenter_pseud_or_name_link(comment)
+    if comment.comment_owner
```
weeklies, line 184: Can we reduce the nested ifs by using `if comment.comment_owner.nil?` as a guard clause?

app/views/comment_mailer/comment_reply_notification.text.erb
```diff
@@ -1,12 +1,5 @@
 <% content_for :message do %>
-<% responder = @comment.comment_owner ? 
-    (
-      @comment.by_anonymous_creator? ?
-      "Anonymous Creator" : 
-      @comment.comment_owner_name + " (" + polymorphic_url(@comment.comment_owner, :only_path => false) + ")"
-    ) :
-    @comment.comment_owner_name %>
-<%= responder %> replied to your comment on <% if @comment.ultimate_parent.is_a?(Tag) then %> the tag <%= @comment.ultimate_parent.commentable_name %> (<%= url_for(:controller => :tags, :action => :show, :id => @comment.ultimate_parent.to_param, :only_path => false) %>)<% else %>"<%= @comment.ultimate_parent.commentable_name.html_safe %>" (<%= polymorphic_url(@comment.ultimate_parent, :only_path => false) %>)<% end %>:
+<%= commenter_pseud_or_name_text(@comment) %> replied to your comment on <% if @comment.ultimate_parent.is_a?(Tag) then %> the tag <%= @comment.ultimate_parent.commentable_name %> (<%= url_for(:controller => :tags, :action => :show, :id => @comment.ultimate_parent.to_param, :only_path => false) %>)<% else %>"<%= @comment.ultimate_parent.commentable_name.html_safe %>" (<%= polymorphic_url(@comment.ultimate_parent, :only_path => false) %>)<% end %>:
```
weeklies, line 2: We should probably update the hash rockets to Ruby 1.9 hash syntax, e.g. `only_path: false`

app/helpers/mailer_helper.rb
```diff
@@ -176,6 +180,22 @@ def style_work_tag_metadata(tags)
     "#{label}#{style_work_tag_metadata_list(tags)}".html_safe
   end
 
+  def commenter_pseud_or_name_link(comment)
+    style_bold(comment.comment_owner_name) + style_role("Guest") unless comment.comment_owner
```
weeklies, line 184: This needs a `return` to complete the guard clause, as the subsequent methods are still being called for guest comments which is probably causing the test failures. 

I am fairly certain the next line for anonymous creators needs it as well.

app/helpers/mailer_helper.rb
```diff
@@ -9,6 +9,10 @@ def style_link(body, url, html_options = {})
     link_to(body.html_safe, url, html_options)
   end
 
+  def style_role(text)
+    "<em><strong>(#{text})</strong></em>".html_safe
```
tickinginstant, line 13: I know that this is just following the pattern of several other functions in `app/helpers/mailer_helper.rb`, but it would be good to avoid calls to `html_safe` whenever possible, to reduce the chance of future XSS exploits. This isn't currently a security risk, because the only text that's passed to this function is the word "Guest", but there are no guarantees about what someone might pass to this function in the future.

Can you replace the `html_safe` with some calls to `tag.em` and `tag.strong`? e.g.
```ruby
tag.em(tag.strong("(#{text})"))
```

app/helpers/mailer_helper.rb
```diff
@@ -176,6 +180,22 @@ def style_work_tag_metadata(tags)
     "#{label}#{style_work_tag_metadata_list(tags)}".html_safe
   end
 
+  def commenter_pseud_or_name_link(comment)
+    return style_bold(comment.comment_owner_name) + style_role("Guest") unless comment.comment_owner
+
+    return style_bold("Anonymous Creator") if comment.by_anonymous_creator?
+      
+    style_link(comment.comment_owner_name, polymorphic_url(comment.comment_owner, only_path: false))
```
tickinginstant, line 188: I think I'm going to have to disagree with @weeklies here — I think the cleanest way to approach this is a structure like this:
```ruby
if comment.comment_owner.nil?
  [...]
elsif comment.by_anonymous_creator?
  [...]
else
  [...]
end
```
I know there's a lot of differences of opinion over when and how to use guard clauses, but I think guard clauses tend to make the most sense when they involve handling edge cases, because they allow bailing out early while skipping the main computation being performed.

In this case, I think the name of the function (`commenter_pseud_or_name_link`) indicates what the main purpose of this function is (returning the pseud link or the guest name). So only one of the three values being returned is an edge case; the other two are the main values computed by the function. I think a structure that gives them equal weight would be easier to follow.

weeklies: I agree with an `if else` depth of 1, just wanted to avoid nested `if`s previously.

weeklies (changes requested):

app/views/comment_mailer/edited_comment_reply_notification.html.erb
```diff
@@ -24,7 +15,7 @@
   </p>
     
   <p>
-    <%= style_bold(responder) %> edited their response to:
+    <%= commenter_pseud_or_name_text(@comment) %> edited their response to:
```
weeklies, line 18: I think you missed one.

weeklies (approved):
Looks good to me.

app/helpers/mailer_helper.rb
```diff
@@ -9,6 +9,10 @@ def style_link(body, url, html_options = {})
     link_to(body.html_safe, url, html_options)
   end
 
+  def style_role(text)
+    tag.em(tag.strong("(#{text})"))
```
brianjaustin, line 13: Should the parentheses be included here? I know some languages use different parentheses, for example https://en.wikipedia.org/wiki/Chinese_punctuation#Marks_similar_to_European_punctuation, so extracting it to the locale files might be more translation-friendly

brianjaustin (commented):
Just one comment/question from me! (Also, if you do change things would you mind getting the Reviewdog warnings too? No worries on those either way though :smile: )

brianjaustin (approved):

PR #4597, opened 2023-08-11
AO3-6589 Don't call ts() on user-generated text.

tickinginstant:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6589

## Purpose

Rewrites some calls to `ts()` that used to pass user-generated text as part of the format string, potentially causing errors if that user-generated text contains a `%` sign.

### DIFF
--- a/app/controllers/tag_set_nominations_controller.rb
+++ b/app/controllers/tag_set_nominations_controller.rb
@@ -315,7 +315,7 @@ class TagSetNominationsController < ApplicationController
               @errors << ts("Couldn't find a #{type} nomination for %{name}", name: name)
               @force_expand[type] = true
             elsif !tagnom.change_tagname?(val)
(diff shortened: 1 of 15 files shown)


app/controllers/tag_set_nominations_controller.rb
```diff
@@ -312,10 +312,10 @@ def collect_update_multiple_results
             # this is the tricky one: make sure we can do this name change
             tagnom = TagNomination.for_tag_set(@tag_set).where(type: "#{type.classify}Nomination", tagname: name).first
             if !tagnom
-              @errors << ts("Couldn't find a #{type} nomination for #{name}")
+              @errors << ts("Couldn't find a #{type} nomination for %{name}", name: name)
```
tickinginstant, line 315: I changed the handling of `name` but not `type` here because `type` is the tag type, which should be translated once this message is moved over to `t()`.

zz9pzza (approved):

PR #4601, opened 2023-08-12
AO3-6040 Allow banning certain usernames

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6040

## Purpose

Add the ability to ban the use of certain usernames (although none are currently blocked) to avoid confusion with official terms

## Testing Instructions

Refer to Jira

## Credit

Brian Austin (they/he)


tickinginstant (commented):
Do you think it's worth pulling out the check into a separate validator like [this one](https://github.com/otwcode/otwarchive/blob/master/app/validators/email_blacklist_validator.rb)? Or is it better to keep it as one function in the `User` class and one function in the `Comment` class?

app/models/user.rb
```diff
@@ -185,13 +185,21 @@ def unread_inbox_comments_count
...
+            },
+            format: {
+              with: /\A[A-Za-z0-9]\w*[A-Za-z0-9]\Z/,
+              message: I18n.t("user.login.bad_format",
+                              min_login: ArchiveConfig.LOGIN_LENGTH_MIN,
+                              max_login: ArchiveConfig.LOGIN_LENGTH_MAX)
+            },
+            uniqueness: true
+  validate :login, :username_is_not_recently_changed, if: :will_save_change_to_login?
+  validate :login, :username_not_forbidden, if: :will_save_change_to_login?
```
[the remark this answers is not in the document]
brianjaustin: :facepalm: thanks! I always get those mixed up. I was reaching for something like a shared validator, but the closest I got was a concern. I think it's reasonable to have something shared for this

config/locales/validators/en.yml
```diff
@@ -6,3 +6,4 @@ en:
       format:
         allow_blank: should look like an email address. Please use a different address or leave blank.
         no_blank: should look like an email address.
+    forbidden_name: "%{value} is not allowed"
```
weeklies, line 9: Not AD&T but would appending "as a user name" add more clarity to the error message?

brianjaustin: When Rails displays the errors, it will prepend the field name (that's why I overrode `login` to be `User name` instead). Here's an example of how it looks on the sign-up page:
an error message with the header 'Sorry! We couldn't save this user because:' and text 'User name foobar is not allowed' inside a light red rectangle

and for the guest comment form:
an error message with the header 'Sorry! We couldn't save this comment because:' and the text 'Name foobar is not allowed' inside a light red rectangle

That being said, @sarken do we have any preference from PAC on the exact copy for this?

sarken: I'll check with PAC and Support!

sarken: Neither PAC nor Support had objections to this wording.

app/models/user.rb
```diff
@@ -185,13 +185,21 @@ def unread_inbox_comments_count
...
+              within: ArchiveConfig.LOGIN_LENGTH_MIN..ArchiveConfig.LOGIN_LENGTH_MAX,
+              too_short: I18n.t("user.login.too_short", min_login: ArchiveConfig.LOGIN_LENGTH_MIN),
+              too_long: I18n.t("user.login.too_long", max_login: ArchiveConfig.LOGIN_LENGTH_MAX)
+            },
+            format: {
+              with: /\A[A-Za-z0-9]\w*[A-Za-z0-9]\Z/,
+              message: I18n.t("user.login.bad_format",
+                              min_login: ArchiveConfig.LOGIN_LENGTH_MIN,
+                              max_login: ArchiveConfig.LOGIN_LENGTH_MAX)
+            },
```
tickinginstant, line 200: Oh, I forgot to flag this before: the calls to `I18n.t` here aren't inside a block or a method, so they'll be evaluated when the class is loaded, i.e. when the server first boots up. This means that the error messages will always be in the default locale, even if someone is browsing with a different locale set. If you want to make this translatable, you have to pass symbols for each of the messages, or just rely on the default symbols.

(This was also true when using `ts` here. But `ts` means the string isn't translatable yet; whereas `t` ought to mean that translation works.)

The default symbols used by the length validator are `too_short` and `too_long`, and they pass a `count` param so that you can customize the message with e.g. `is too long (maximum is %{count} characters)`. But it's worth noting that I think the default length messages are actually [already set to the desired text](https://github.com/svenfuchs/rails-i18n/blob/556eb65a4b701eec430cfc5954f103a102ca1ae5/rails/locale/en.yml#L134-L139). So it would probably just be best to delete the `too_short` and `too_long` params altogether.

The format error message is a little trickier, because the min and max length aren't normally a part of the format validator. But it looks like you should be able to pass custom parameters to the validator and have them [added as parameters to the error](https://github.com/rails/rails/blob/3779b5069dd1742d878cc0e322fc642e730efa42/activemodel/lib/active_model/validations/format.rb#L31).

app/validators/not_forbidden_name_validator.rb
```diff
@@ -0,0 +1,8 @@
+class NotForbiddenNameValidator < ActiveModel::EachValidator
+  def validate_each(record, attribute, value)
+    return if value.nil?
+    return unless ArchiveConfig.FORBIDDEN_USERNAMES.include?(value.downcase)
+
+    record.errors.add(attribute, options[:message] || I18n.t("validators.forbidden_name", value: value))
```
tickinginstant, line 6: Unlike the cases above, it's okay to use `t` instead of adding a symbolic error here, but I do think I'd prefer a symbolic error — it makes it easier to customize the message for different classes. (I used the `BlacklistedEmailValidator` as an example to follow not because I like the way it's written, but because it was the validator with the closest functionality.)

Still, this change is more aesthetic than anything, since currently we don't need to customize the error messages on a per-class basis. So we can leave the code as is if you like it this way.

config/locales/models/en.yml
```diff
@@ -96,6 +98,7 @@ en:
           format: "%{message}"
         comment:
           attributes:
+            name:
```
tickinginstant, line 101: Is this line necessary?

config/locales/models/en.yml
```diff
@@ -145,6 +148,7 @@ en:
               changed_too_recently:
                 one: can only be changed once per day. You last changed your user name on %{renamed_at}.
                 other: can only be changed once every %{count} days. You last changed your user name on %{renamed_at}.
+              taken: has already been taken
```
tickinginstant, line 151: This is already the [default text](https://github.com/svenfuchs/rails-i18n/blob/556eb65a4b701eec430cfc5954f103a102ca1ae5/rails/locale/en.yml#L133) for the `taken` error, so I don't think this is necessary.

config/locales/models/en.yml
brianjaustin: Nope! I bet it's from one of my earlier, failed experiments. I've cleaned it up

app/validators/not_forbidden_name_validator.rb
brianjaustin: I also like the symbol way better, and I probably should've done a more thorough survey of those validator classes first. Updated!

app/models/user.rb
brianjaustin: Ah right, thanks for catching this (and for linking all those classes)! I checked again, and the defaults do indeed seem to be right for everything except `format`, which I have now updated to be the custom message.

sarken:
Could I bother you for a test showing it's possible to log in and change your username if your current username is on the forbidden list? 

I'm also wondering... are we going to run into any issues freezing, hiding, or deleting existing guest comments where the commenter has used a forbidden name?

brianjaustin:
@sarken sure! I've added a feature test for that (seemed like the right place, but I can pull it down into a controller or model test if you prefer)
> I'm also wondering... are we going to run into any issues freezing, hiding, or deleting existing guest comments where the commenter has used a forbidden name?

Initially, I thought no because of how my original implementation worked. It turned out, the answer was actually yes. Along with some other fixes, I've made the `Comment` and `User` validations only fire if name/login is changed

features/users/user_rename.feature
```diff
@@ -163,3 +163,13 @@ Feature:
       And I should see "Hi, newusername"
     When I follow "Series"
     Then I should see "Best Series by newusername"
+
+    Scenario: Changing the username from a forbidden name to non-forbidden
+      Given I am logged in as "forbidden" with password "password"
+        And the user name "forbidden" is on the forbidden list
```
sarken, line 169: Can we tweak this a little so the name is on the forbidden list before we log in? (The spec that shows you can save changes if you're not editing the user means this scenario _should_ be fine, but I think it's better to explicitly cover it in the feature.)

```
Given the following activated user exists
  | login     |  etc |
  | forbidden |      |
  And the user name "forbidden" is on the forbidden list
When I log in as etc
```

sarken (commented):
Thanks for the changes! I have one question and one suggestion and then I think this will be ready.

app/models/comment.rb
```diff
@@ -14,7 +14,8 @@ class Comment < ApplicationRecord
 
   has_many :thread_comments, class_name: 'Comment', foreign_key: :thread
 
-  validates_presence_of :name, unless: :pseud_id
+  validates :name, presence: { unless: :pseud_id }
+  validates :name, not_forbidden_name: true, if: :will_save_change_to_name?
```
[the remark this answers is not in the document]
brianjaustin: Yes! I've also added a test for that, since it is an important use case

features/users/user_rename.feature
```diff
@@ -163,3 +163,17 @@ Feature:
...
     When I follow "Series"
     Then I should see "Best Series by newusername"
+
+    Scenario: Changing the username from a forbidden name to non-forbidden
+      Given I have no users
+        And the following activated user exists
+          | login     | password |
+          | forbidden | secret   |
+        And the user name "forbidden" is on the forbidden list
+      When I am logged in as "forbidden" with password "password"
```
sarken, line 173: I am... deeply sorry to be this person, by why on earth is it letting us log in with a password that doesn't match the one set two lines up? 😬 😭

brianjaustin: No worries, I am happy to have anyone pry when something looks weird versus the alternative :slightly_smiling_face: . It looks like
1. the step calls `find_or_create_new_user`: https://github.com/otwcode/otwarchive/blob/dba57ea17dbdef10c4cacb1fffdad72af92f4cd6/features/step_definitions/user_steps.rb#L97
2. that function finds the existing user: https://github.com/otwcode/otwarchive/blob/92d7fc675437c76b8f1e33751ac3957224b064f0/features/support/user.rb#L3
3. and "helpfully" (? on the quotation marks) changes the password to whatever we say it is at login: https://github.com/otwcode/otwarchive/blob/92d7fc675437c76b8f1e33751ac3957224b064f0/features/support/user.rb#L12

So I think that's intended behaviour. I've changed this particular test, to avoid confusion later. We might want to change how this works later, but I suspect it will involve some amount of pain to get all the Cucumber specs passing.

sarken: "Helpfully" indeed. Thanks for investigating! We should probably make a low priority issue to change that at some point.

sarken (approved):

app/models/user.rb
```diff
@@ -186,13 +186,16 @@ def unread_inbox_comments_count
...
-                                   max_login: ArchiveConfig.LOGIN_LENGTH_MAX)
+  validates :login,
+            length: { within: ArchiveConfig.LOGIN_LENGTH_MIN..ArchiveConfig.LOGIN_LENGTH_MAX },
+            format: {
+              with: /\A[A-Za-z0-9]\w*[A-Za-z0-9]\Z/,
+              min_login: ArchiveConfig.LOGIN_LENGTH_MIN,
+              max_login: ArchiveConfig.LOGIN_LENGTH_MAX
+            },
+            uniqueness: true
+  validates :login, not_forbidden_name: true, if: :will_save_change_to_login?
```
tickinginstant, line 197: I think you can do this to restrict the scope of the `if` argument:
```suggestion
            uniqueness: true,
            not_forbidden_name: { if: :will_save_change_to_login? }
```

tickinginstant (approved):
Looks good to me! One optional style thing, but I think this is good to go.

app/models/user.rb
brianjaustin: Nice, thank you! I've updated both places to use this for consistency

PR #4604, opened 2023-08-13
AO3-6590 Add reviewdog for linting ERB files

Bilka2:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6590

## Purpose

Use reviewdog to run ERB lint on pull requests because Hound does not run it (despite being configured for it).

Demo PR from same repo: https://github.com/Bilka2/otwarchive/pull/1
Demo PR from forked repo: https://github.com/Bilka2/otwarchive/pull/2 (see reviewdog's docs on [Graceful degradation for PRs from forks](https://github.com/reviewdog/reviewdog#graceful-degradation-for-pull-requests-from-forked-repositories))

## Testing Instructions

See Jira.

## Credit

Bilka (he/him)

### DIFF
--- a/.github/workflows/reviewdog.yml
+++ b/.github/workflows/reviewdog.yml
@@ -1,0 +1,27 @@
+# Based on https://github.com/tk0miya/action-erblint/blob/main/README.md#example-usage
+
+name: Reviewdog ERB Lint
+
+on: [pull_request]
+
+permissions:
+  checks: write
+
+jobs:
+  erb-lint:
+    name: ERB Lint runner
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
+
+      - name: Set up Ruby and run bundle install
+        uses: ruby/setup-ruby@v1
+        with:
+          bundler-cache: true
+
+      - name: erb-lint
+        uses: tk0miya/action-erblint@667687e73b44e7b7a710a1204b180f49f80ebb5e
+        with:
+          use_bundler: true
+          reporter: github-pr-check # default
--- a/.github/workflows/reviewdog.yml
+++ b/.github/workflows/reviewdog.yml
@@ -1,6 +1,6 @@
 # Based on https://github.com/tk0miya/action-erblint/blob/main/README.md#example-usage
 
-name: Reviewdog ERB Lint
+name: Reviewdog
 
 on: [pull_request]
 
--- a/.hound.yml
+++ b/.hound.yml
@@ -1,10 +1,6 @@
 # Available linter versions:
 # http://help.houndci.com/en/articles/2461415-supported-linters
 
-erblint:
-  enabled: true
-  config_file: .erb-lint.yml
-
 jshint:
   config_file: .jshintrc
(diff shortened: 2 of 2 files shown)


Bilka2:
As is, the workflow [fails in this repository](https://github.com/otwcode/otwarchive/actions/runs/5846802542):
> tk0miya/action-erblint@667687e73b44e7b7a710a1204b180f49f80ebb5e is not allowed to be used in otwcode/otwarchive. Actions in this workflow must be: within a repository owned by otwcode, created by GitHub, or verified in the GitHub Marketplace.

See GitHub Actions docs on [Allowing select actions and reusable workflows to run](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#allowing-select-actions-and-reusable-workflows-to-run).

.github/workflows/reviewdog.yml
```diff
@@ -0,0 +1,27 @@
+# Based on https://github.com/tk0miya/action-erblint/blob/main/README.md#example-usage
+
+name: Reviewdog ERB Lint
```
tickinginstant, line 3: We may eventually want to add a second job for running rubocop to this file. Can we keep the workflow name more general, like "Reviewdog"?

.github/workflows/reviewdog.yml
```diff
@@ -0,0 +1,27 @@
...
+      - name: Check out code
+        uses: actions/checkout@v3
+
+      - name: Set up Ruby and run bundle install
+        uses: ruby/setup-ruby@v1
+        with:
+          bundler-cache: true
+
+      - name: erb-lint
+        uses: tk0miya/action-erblint@667687e73b44e7b7a710a1204b180f49f80ebb5e
```
brianjaustin, line 24: Any reason not to use `@v1`? https://github.com/tk0miya/action-erblint/releases/tag/v1

tickinginstant: If you don't trust the creator of the action, it's usually best practices to use the hash version for actions, because it refers to a specific commit. That means that you can view the repository, read the source code of the action, and be sure that when the action runs, it's running only code that you've already reviewed.

Tags, on the other hand, can be switched to point to a different commit, which means that the owner of the action could choose to replace the existing code with some kind of exploit.

I'm not sure the extra security is necessary, in this case, but I can see both sides of the argument.

Bilka2: GitHub recommends using a SHA in [their docs](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions), especially if the action author is not a verified creator, like in this case.

However, this workflow has very few permissions (write on checks for PRs from the same repo, read for third party repos), so if it were compromised I don't think it could do much harm. But I'd rather be safe than sorry, because github actions and their permissions are quite new to me.

brianjaustin: Fair enough! Thanks for the link @Bilka2!

brianjaustin (approved):

PR #4607, opened 2023-08-15
AO3-6058 Hide byline of unrevealed work in notes of related work

weeklies:
Apologies for the numerous PRs, I promise this is the last for a while. 

# Pull Request Checklist

## Issue

## Purpose

- Do not reveal parent or child works in unrevealed collections
- i18n
- Add tests that demonstrate the downloads update properly when parent or child work changes in visibility.

## Testing Instructions

In addition to the steps on Jira, please make sure the downloaded work notes do not break anonymity. (I've verified on development: PDF, EPUB, HTML.)

## References

https://github.com/otwcode/otwarchive/pull/3911

## Credit
james_

weeklies (she/her)


weeklies:
I am not sure how to resolvei18n-task failures, because when testing, the downloaded files required the `downloads.download_preface.html.inspired_by.unrevealed `rather than `downloads.download_preface.inspired_by.unrevealed`. 
```
+--------+-----------------------------------------------------------+----------------------------------------------------+
| Locale | Key                                                       | Value                                              |
+--------+-----------------------------------------------------------+----------------------------------------------------+
|   en   | downloads.download_preface.html.external_related_work     | %{work_link} by %{creator_link}.                   |
|   en   | downloads.download_preface.html.inspired_by.revealed      | Inspired by %{work_link} by %{creator_link}.       |
|   en   | downloads.download_preface.html.inspired_by.unrevealed    | Inspired by a work in an unrevealed collection.    |
|   en   | downloads.download_preface.html.translation_of.revealed   | Translation of %{work_link} by %{creator_link}.    |
|   en   | downloads.download_preface.html.translation_of.unrevealed | Translation of a work in an unrevealed collection. |
+--------+-----------------------------------------------------------+----------------------------------------------------+
```

sarken:
Have you tried adding comments with [i18n-tasks-use hints](https://github.com/glebm/i18n-tasks#fine-tuning)?

weeklies:
Current test failure is related to [this](https://github.com/otwcode/otwarchive/pull/4605/files), not this PR. It looks like it happens intermittently.

app/views/downloads/_download_preface.html.erb
```diff
@@ -67,17 +67,24 @@
 
     <% related_works = @work.parent_work_relationships.reject { |wr| !wr.parent } %>
     <% if related_works.length > 0 %>
+      <%# i18n-tasks-use t("downloads.download_preface.html.inspired_by.unrevealed")
+          i18n-tasks-use t("downloads.download_preface.html.inspired_by.revealed_html")
+          i18n-tasks-use t("downloads.download_preface.html.translation_of.unrevealed")
+          i18n-tasks-use t("downloads.download_preface.html.translation_of.revealed_html") %>
```
tickinginstant, line 73: I'm assuming that the extra `.html.` here is necessary because of the `.html` in this line:
https://github.com/otwcode/otwarchive/blob/538664f7b1de73539a5c61b28c37c72ca7a0821d/app/views/downloads/show.html.erb#L1
But passing the suffix like that is deprecated, so that line's going to need changing in the 7.0 upgrade. Which means that all of the keys here would have to be renamed as part of the 7.0 upgrade, and this would need to be re-tested. I think it'd be cleaner to fix the deprecated line now, and use the correct key names so that the 7.0 upgrade goes smoother.

tickinginstant (changes requested):
This still needs some callbacks to expire the caches on all children when a work changes from unrevealed to revealed or vice versa.

(Either that, or it needs some nginx changes to make sure that downloads are only ever cached for ... IDK, a day? Some short amount of time. Download generation is pretty slow, though, so it's probably better to go with the callback.)

Also, there may be a similar issue with details about unrevealed child related works being listed in the parent, once the parent work has approved the relationship. Should that be addressed as part of this issue, or as a separate issue?

app/views/downloads/_download_preface.html.erb
weeklies: Happy to resolve it here, thanks for the info!

weeklies:
For the case of child related works in an unrevealed collection, I am going with not displaying them in the notes of parent works even if they are approved. Because I think this looks a bit awkward:
`Translation into %{language} available: A work in an unrevealed collection.`

config/locales/views/en.yml
```diff
@@ -700,5 +712,25 @@ en:
...
+      translated_to:
+        restricted_html: "[Restricted Work] by %{creator_link} (translation into %{language}) Log in to view."
+        revealed_html: "%{work_link} by %{creator_link} (translation into %{language})"
+    work_header_notes:
+      inspired_by:
+        restricted_html: Inspired by [Restricted Work] by %{creator_link}. Log in to view.
+        revealed_html: Inspired by %{work_link} by %{creator_link}.
+        unrevealed: Inspired by a work in an unrevealed collection.
+      translated_to:
+        restricted_html: 'Translation into %{language} available: [Restricted Work] by %{creator_link}. Log in to view.'
```
weeklies, line 729: Calling `i18n-tasks normalize` [converts ](https://github.com/glebm/i18n-tasks/issues/248) this to single quotes.

features/works/work_download.feature
```diff
@@ -202,3 +202,66 @@ Feature: Download a work
...
+    And I follow "HTML"
+  Then I should not see the related work listed on the original work
+    And I should not see the translation listed on the original work
+    And I should see "A work in an unrevealed collection."
+  # Going from unrevealed to revealed
+  When I reveal works for "Hidden"
+    And I log out
+    And I view the work "Worldbuilding"
+    And I follow "HTML"
+  Then I should see the related work listed on the original work
```
weeklies, line 265: @tickinginstant I'm not sure how to expire the caches of works such that downloads are up to date. I tried poking the work model but the cache of the work appears to be already updating appropriately but not the download.

tickinginstant: The caching issues aren't reproducible in tests because the caching is nginx-side, rather than Rails-side. The cache expiration is handled by making each download URL append a timestamp:
https://github.com/otwcode/otwarchive/blob/538664f7b1de73539a5c61b28c37c72ca7a0821d/app/helpers/works_helper.rb#L121-L124
So that when the work's `updated_at` is changed, the link to the download changes. nginx's cache keys are based on the path, so when `updated_at` changes, the cache key used by nginx also changes. (It's the same principle behind the way that Rails does non-recyclable cache keys — if the cache key includes the timestamp, then changes to the timestamp won't delete the old cache key, but they will cause the cache key to change, forcing the cache to be regenerated.)

weeklies: Thank you!

features/step_definitions/work_related_steps.rb
```diff
@@ -159,25 +159,40 @@
   step %{I should see "Followup by remixer" within ".afterword .children"}
 end
 
+Then /^I should see the related work listed on the original work$/ do
```
Bilka2, line 162: I think this and the other new/changed steps in the file should be using [cucumber expressions](https://github.com/cucumber/cucumber-expressions#readme) instead of regex.

features/step_definitions/work_related_steps.rb
```diff
@@ -159,25 +159,40 @@
...
+  step %{I should see "Translation of Worldbuilding by inspiration" within ".preface .notes"}
 end
 
-Then /^I should see the translation in the beginning notes$/ do
+Then "I should see the translation in the beginning notes" do
   step %{I should see "Translation into Deutsch available:" within ".preface .notes"}
   step %{I should see "Worldbuilding Translated by translator" within ".preface .notes"}
 end
 
+Then /^I should see the translation listed on the original work$/ do
```
Bilka2, line 191: This regular expression was missed when doing the conversion to cucumber expressions.

sarken (commented):
Regarding the inconsistencies you noted, yeah, let's definitely address those! 

I made a [two-sheet spreadsheet](https://docs.google.com/spreadsheets/d/1U5USqPEAd93Vl3O8vbacuyXoDS3Dj1FSVw3R5amZ-gY/edit#gid=0) comparing the current text (you can see it on [this test work](https://test.archiveofourown.org/works/1083235)) to the new versions, and left comments pointing out some things that need to change. (You can ignore the highlighting; that was just to make it easier to show PAC and Support chairs which text had actually changed.)

weeklies:
Sorry for throwing off the commit history, fixing it seems to always make it worse. 🤦

weeklies:
I stared at the i18n strings for too long and now I'm wondering if the presence and absence of a ending period should be standardized. (Not very important but to cover all the bases)

sarken:
I've been thinking about that period because these aren't really complete sentences with subjects and verbs and so on -- we just needed something to separate the log in instructions from the rest. 

In fact, we were going to ask you to remove it from places like `Translation into %{language} available: %{work_link} by %{creator_link}.` but then I overthought it at the last minute and took that request out. 😆 We definitely shouldn't be adding it to `%{work_link} by %{creator_link}` or `A work in an unrevealed collection` (in ` work_approved_children`), though. 

So what if we just left it out in _all_ cases and put the prompt to log in in parentheses, e.g.:

- `[Restricted Work] by %{creator_link} (Log in to access.)`
- `Inspired by [Restricted Work] by %{creator_link} (Log in to access.)`

app/helpers/works_helper.rb
```diff
@@ -112,6 +112,35 @@ def get_inspired_by(work)
     work.approved_related_works.where(translation: false)
   end
 
+  def related_work_note(related_work, relation, download: false)
```
sarken, line 115: Just one more thing! In the download view files and when `download` is true for this method, could you add `locale: :en` to the translation method, like we did in #4615 and #4618? We wanted to hard code the locales in downloads just to be safe.

app/helpers/works_helper.rb
```diff
@@ -112,6 +112,39 @@ def get_inspired_by(work)
...
+    default_locale = download ? :en : nil
+
+    creator_link = if download
+                     byline(related_work, visibility: "public", only_path: false)
+                   else
+                     byline(related_work)
+                   end
+
+    if related_work.respond_to?(:unrevealed?) && related_work.unrevealed?
+      if relation == "translated_to"
```
weeklies, line 127: This doesn't apply for downloads, so setting the locale is not needed.

app/helpers/works_helper.rb
```diff
@@ -112,6 +112,39 @@ def get_inspired_by(work)
     work.approved_related_works.where(translation: false)
   end
 
+  def related_work_note(related_work, relation, download: false)
+    work_link = link_to related_work.title, polymorphic_url(related_work)
+    language = tag.span(related_work.language.name, lang: related_work.language.short) if related_work.language
+    default_locale = download ? :en : nil
```
sarken, line 118: Just to double check, because I'm having zero Google luck right now: when we let users choose a locale for the interface, any call to the `t` method including `locale: nil` should use the user's locale, not the site's default, right?

weeklies: I believe setting the locale only applies to mailers right now, but it looks like it will when we configure it to work with views. [source](https://github.com/ruby-i18n/i18n/blob/master/lib/i18n.rb#L211)

sarken (approved):

PR #4243, opened 2022-05-30
AO3-6312 Add preference to prevent guests from replying to your comments on news posts and other users' works

giuxtaposition:
# Pull Request Checklist

* [x] Have you read ["How to write the perfect pull request"](https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/)?
* [x] Have you read the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [x] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)?
* [x] Have you added the [Jira](https://otwarchive.atlassian.net) issue number
  as the *first* thing in your pull request title (e.g. `AO3-1234 Fix thing`)
* [x] Do you have fewer than 5 pull requests already open? If not, please wait
  until they are reviewed and merged before creating new pull requests.

## Issue

https://otwarchive.atlassian.net/browse/AO3-6312 (Please fill in issue number and remove this comment.)

## Purpose

Add preference to prevent guests from replying to your comments on news posts and other users' works

## Testing Instructions

The instructions on the Jira issue are enough.

## References

Nothing to add.

## Credit

giuxtaposition, she/they pronouns

PR #4612, opened 2023-08-19
AO3-6312 Add preference to prevent guest replies to comments

Bilka2:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6312

## Purpose

Add a preference to prevent logged out users from replying to your comments on news posts and other users' works. Guests can still reply to you on works you created or co-created.

## Testing Instructions

See Jira.

## References

Partial implementation in #4243.
Migration with departure gem based on #4528.
Help text for the admin emails feature removed because of [AO3-5043](https://otwarchive.atlassian.net/browse/AO3-5043).

## Credit

Bilka (he/him)

[AO3-5043]: https://otwarchive.atlassian.net/browse/AO3-5043?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/helpers/comments_helper.rb
+++ b/app/helpers/comments_helper.rb
@@ -108,8 +108,8 @@ module CommentsHelper
       comment_parent_hidden?(comment) ||
       blocked_by_comment?(comment) ||
       blocked_by?(comment.ultimate_parent) ||
-      guest? && admin_settings.guest_comments_off? ||
-      guest? && comment.user.disallow_guest_replies?(comment.ultimate_parent) )
+      (guest? && admin_settings.guest_comments_off?) ||
+      (guest? && comment.user&.disallow_guest_replies?(comment.ultimate_parent)))
   end
 
   def can_edit_comment?(comment)
--- a/app/helpers/comments_helper.rb
+++ b/app/helpers/comments_helper.rb
@@ -100,16 +100,18 @@ module CommentsHelper
 
   def can_reply_to_comment?(comment)
     admin_settings = AdminSetting.current
-    
-    !(comment.unreviewed? ||
-      comment.iced? ||
-      comment.hidden_by_admin? ||
-      parent_disallows_comments?(comment) ||
-      comment_parent_hidden?(comment) ||
-      blocked_by_comment?(comment) ||
-      blocked_by?(comment.ultimate_parent) ||
-      (guest? && admin_settings.guest_comments_off?) ||
-      (guest? && comment.guest_replies_disallowed?))
+
+    return false if comment.unreviewed?
+    return false if comment.iced?
+    return false if comment.hidden_by_admin?
+    return false if parent_disallows_comments?(comment)
+    return false if comment_parent_hidden?(comment)
+    return false if blocked_by_comment?(comment)
+    return false if blocked_by?(comment.ultimate_parent)
+
+    return true unless guest?
+
+    !(admin_settings.guest_comments_off? || comment.guest_replies_disallowed?)
   end
 
   def can_edit_comment?(comment)
--- a/app/helpers/comments_helper.rb
+++ b/app/helpers/comments_helper.rb
@@ -109,7 +109,7 @@ module CommentsHelper
       blocked_by_comment?(comment) ||
       blocked_by?(comment.ultimate_parent) ||
       (guest? && admin_settings.guest_comments_off?) ||
-      (guest? && comment.user&.disallow_guest_replies?(comment.ultimate_parent)))
+      (guest? && comment.guest_replies_disallowed?))
   end
 
   def can_edit_comment?(comment)
--- a/app/helpers/comments_helper.rb
+++ b/app/helpers/comments_helper.rb
@@ -108,7 +108,8 @@ module CommentsHelper
       comment_parent_hidden?(comment) ||
       blocked_by_comment?(comment) ||
       blocked_by?(comment.ultimate_parent) ||
-      guest? && admin_settings.guest_comments_off?)
+      guest? && admin_settings.guest_comments_off? ||
+      guest? && comment.user.disallow_guest_replies?(comment.ultimate_parent) )
   end
 
   def can_edit_comment?(comment)
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -22,11 +22,12 @@ class Comment < ApplicationRecord
     maximum: ArchiveConfig.COMMENT_MAX,
     too_long: ts("must be less than %{count} characters long.", count: ArchiveConfig.COMMENT_MAX)
 
-  validates_each :commentable, if: :reply_comment?, unless: :pseud_id? do |record, attribute, value|
-    record.errors.add(attribute, :guest_replies_off) if value.user&.disallow_guest_replies?(value.ultimate_parent)
-  end
-
   delegate :user, to: :pseud, allow_nil: true
+
+  validate :reply_as_guest_allowed, if: :reply_comment?, unless: :pseud_id, on: :create
+  def reply_as_guest_allowed
+    errors.add(:commentable, :guest_replies_off) if commentable.user&.disallow_guest_replies?(ultimate_parent)
+  end
 
   # Check if the writer of this comment is blocked by the writer of the comment
   # they're replying to:
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -24,16 +24,16 @@ class Comment < ApplicationRecord
 
   delegate :user, to: :pseud, allow_nil: true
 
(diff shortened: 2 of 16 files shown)


app/helpers/comments_helper.rb
```diff
@@ -108,7 +108,8 @@ def can_reply_to_comment?(comment)
       comment_parent_hidden?(comment) ||
       blocked_by_comment?(comment) ||
       blocked_by?(comment.ultimate_parent) ||
-      guest? && admin_settings.guest_comments_off?)
+      (guest? && admin_settings.guest_comments_off?) ||
```
weeklies, line 111: Since the cases are getting lengthy, should we refactor this to use guard clauses?

EchoEkhi: I don't think so, it‘s still very readable to me.

sarken:
Just a quick comment! I think this is missing a validation in the comment model.

spec/models/comment_spec.rb
```diff
@@ -14,7 +14,7 @@
 
     it "should be invalid if exactly duplicated" do
       expect(second_comment.valid?).to be_falsy
-      expect(second_comment.errors.keys).to include(:comment_content)
+      expect(second_comment.errors.attribute_names).to include(:comment_content)
```
Bilka2, line 17: [ActiveModel::Errors::keys](https://api.rubyonrails.org/v6.1.7/classes/ActiveModel/Errors.html#method-i-keys) is deprecated, to be removed in Rails 7. I used the opportunity to switch to the suggested replacement.

app/models/user.rb
```diff
@@ -431,6 +431,12 @@ def no_resets?
     has_role?(:no_resets)
   end
 
+  # Whether logged out users can reply to this user's comments with the given
+  # ultimate parent
+  def disallow_guest_replies?(ultimate_parent)
```
sarken, line 436: This works, but I'm not sure if it's the right approach. (Others might want to jump in and tell me I'm totally off base here.)

It's true that the user, not the comment, is the thing that controls reply permissions, so I can definitely see the logic behind putting this check in the User model.

However, we're only going to use this when we already have a comment handy, so it feels overly complicated to always need to do something like `comment.user.disallow_guest_replies?(comment.ultimate_parent)` instead of something like `comment.guest_replies_disallowed?` imo, the way we handled the block helpers (`blocked_by_comment?(comment)` and `blocked_by?(comment)`) sets a precedent for using the comment rather than the user in cases where it might be more convenient.

I think there's also some potential for confusion with this approach -- the code comment helps, but when I first saw the method, it was in the comment helper, and I thought it was checking if the user had the work (ultimate parent) set to disallow guest comments.

app/helpers/comments_helper.rb
zz9pzza: I agree with weeklies, can we refactor

app/models/user.rb
zz9pzza: I think I am in agreement with sarken here.

public/help/comment-preferences.html
```diff
@@ -9,6 +9,6 @@ <h4>Comment Preferences</h4>
   <dd>Enable this option if you prefer not to receive emails for your own comments (for example, when you reply to comments on your own works).</dd>
   <dt>Turn off emails about kudos</dt>
   <dd>Enable this option if you would prefer not to receive email notifications when someone leaves kudos on your work.</dd>
-  <dt>Turn off admin emails</dt>
```
zz9pzza, line 12: Why is this deleted ?

spec/controllers/comments_controller_spec.rb
```diff
@@ -854,6 +1079,25 @@
...
+          context "when comment is a guest reply to user who turns off guest replies afterwards" do
+            let(:reply) do
+              reply = create(:comment, :by_guest, commentable: comment)
+              comment.user.preference.update!(guest_replies_off: true)
+              reply
+            end
+
+            it "freezes reply and redirects with success message" do
+              fake_login_admin(admin)
+              put :freeze, params: { id: reply.id }
```
zz9pzza, line 1092: Where is this implemented ?

zz9pzza (changes requested):

public/help/comment-preferences.html
Bilka2: The preference was removed 6 years ago in https://otwarchive.atlassian.net/browse/AO3-5043.

spec/controllers/comments_controller_spec.rb
Bilka2: It's checking that the validation on the model only runs on :create. If it always runs, freezing the guest reply after the user turns off guest replies will fail validation.

I could not figure out how do this as a model test, would appreciate pointers there if that is preferred. I was hung up on how to do the freezing. Just updating the `iced` attribute didn't seem correct.

app/models/user.rb
Bilka2: I was debating where to put it, I initially decided User because it's a user preference. Anyway, happy to move it, done.

Now the comment model has two methods with very similar names, one of them used only for validation. I'm not happy with the method names, they don't feel self-explanatory for what the difference is. I could switch back to using validate_each so the validation uses a block instead of a method. Or the validation method could be private, or they could have better names. Suggestions appreciated.

spec/controllers/comments_controller_spec.rb
zz9pzza: I am trying to understand the test and reconcile it with the code I can see.

I think the test says a comment which is from a guest becomes a frozen thread if the user of a parent comment disables commenting by guests. 

I don't think that is intended so I suspect I am missing something.

Bilka2: User makes comment. Guest replies. User turns off "allow guest replies". Admin tries to freeze guest reply.

Intended: Guest reply is frozen.
Possible problem this test would catch: Freezing fails because the validation of guest reply says that guest replies aren't allowed for the user.

app/models/comment.rb
```diff
@@ -24,6 +24,19 @@ class Comment < ApplicationRecord
 
   delegate :user, to: :pseud, allow_nil: true
 
+  validate :can_be_guest_reply, if: :reply_comment?, unless: :pseud_id, on: :create
+  # Whether the writer of the comment this is replying to allows guest replies
+  def can_be_guest_reply
```
sarken, line 29: For consistency, can you put the comment above the `validate` line, like the `not_blocked` validations do?

I don't think it's super important for the methods to have very distinct names, since they're pretty closely connected -- in fact, we have both a `check_for_spam?` method and a `check_for_spam` validation in this model. But if we wanted to tweak the name, `guest_can_reply` would be in line with the `guest_cannot_kudos_restricted_work` kudos validation.

spec/controllers/comments_controller_spec.rb
```diff
@@ -346,6 +429,74 @@
...
+        it_behaves_like "guest cannot reply to a user with guest replies disabled"
+      end
+
+      context "when comment is on user's work" do
+        let(:work) { create(:work, authors: [user.default_pseud]) }
+        let(:comment) { create(:comment, pseud: user.default_pseud, commentable: work.first_chapter) }
+
+        it_behaves_like "guest can reply to a user with guest replies disabled on user's work"
+      end
+      context "when commentable is user's co-creation" do
```
sarken, line 493: Should there be a blank line here?

spec/controllers/comments_controller_spec.rb
sarken: @zz9pzza It's basically trying to make sure we don't run into a problem like [AO3-6373](https://otwarchive.atlassian.net/browse/AO3-6373).

spec/models/comment_spec.rb
```diff
@@ -167,6 +167,92 @@
...
+                  comment_content: "Hmm.")
+    end
+
+    shared_examples "creating guest reply is allowed" do
+      describe "save" do
+        it "allows guest replies" do
+          expect(guest_reply.save).to be_truthy
+          expect(guest_reply.errors.full_messages).to be_blank
+        end
+        it "allows user replies" do
```
sarken, line 197: I think we need a blank line between `end` and the next example -- same on line 210.

sarken (approved):
Yay, thank you! I have two spacing nitpicks but if james_ feels his comments were addressed and wants to approve and merge, that's fine, they're not vital.

zz9pzza (approved):

PR #4613, opened 2023-08-21
AO3-6533 Validates provided rating string must be one of the allowed strings

ellieyhcheng:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6533

## Purpose

Prevent users from adding more than one ratings. 

## Testing Instructions

Same as in issue:
1. Log in
1. Post > New Work
1. With your browser’s web developer tools enabled, right click on the Rating select menu and choose Inspect Element or equivalent
1. Locate the selected option, which will look something like <option selected="selected" value="Not Rated">Not Rated</option>
1. Edit its value attribute to contain two ratings separated by a comma, e.g., value="Not Rated, General Audiences"
1. Fill in remaining required information in the form
1. Post

Now, it gives an error:
```
Sorry! We couldn't save this work because:
- Only canonical rating tags are allowed.
```

## References

N/A

## Credit

ellieyhc (she/her)

### DIFF
--- a/app/models/rating.rb
+++ b/app/models/rating.rb
@@ -7,13 +7,17 @@ class Rating < Tag
     to_s
   end
 
+  def self.rating_tags
+    Set[ArchiveConfig.RATING_DEFAULT_TAG_NAME,
+      ArchiveConfig.RATING_GENERAL_TAG_NAME,
+      ArchiveConfig.RATING_TEEN_TAG_NAME,
+      ArchiveConfig.RATING_MATURE_TAG_NAME,
+      ArchiveConfig.RATING_EXPLICIT_TAG_NAME]
+  end
+
   # Gives us the default ratings as Not Rated + low to high
   def self.defaults_by_severity
-    ratings = [ArchiveConfig.RATING_DEFAULT_TAG_NAME,
-               ArchiveConfig.RATING_GENERAL_TAG_NAME,
-               ArchiveConfig.RATING_TEEN_TAG_NAME,
-               ArchiveConfig.RATING_MATURE_TAG_NAME,
-               ArchiveConfig.RATING_EXPLICIT_TAG_NAME]
+    ratings = self.rating_tags
     ratings_by_id = Rating.where(name: ratings).inject({}) do |result, rating|
       result[rating.name] = rating
       result
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -144,7 +144,9 @@ class Work < ApplicationRecord
   validates :archive_warning_string,
             presence: { message: "^Please select at least one warning." }
   validates :rating_string,
-            presence: { message: "^Please choose a rating." }
+            presence: { message: "^Please choose a rating." }, 
+            inclusion: { in: Rating.rating_tags,
+                         message: "^Only canonical rating tags are allowed." }
 
   # rephrases the "chapters is invalid" message
   after_validation :check_for_invalid_chapters


ellieyhcheng:
This is kind of a weird bug, and this fix is (I believe) the easiest fix without changing a lot of things. 

This bug comes from the fact that the rating is first stored as `rating_string`. Then, because a `Work` is a `Filterable`, and so it is also a `Taggable`. Before saving, the function `split_tag_string` ([Link](https://github.com/otwcode/otwarchive/blob/master/app/models/concerns/taggable.rb#L103) is called on every tag string, including `rating_string`. Then, the resulting tag(s) are either found (if it exists) or created (if it doesn't). And only now we do the validation for each tag, such as that `Rating` tags have to be canonical. 

So by editing the `value` of the `option` to something with a comma like `Not Rated, General Audiences`, the work's `rating_string` is the entire `Not Rated, General Audiences` string. Then `split_tag_string` splits it into two separate tags `Not Rated` and `General Audiences`. Then, since they both pass the `Rating` model's validation, they're both added as tags. (Note: Before the fix, when the `value` of the `option` is set to something like `Not Rated, NotARating`, the page correctly gives an error saying `Only canonical rating tags are allowed.`. However, this only because when we reach the validation step of each supposed `Rating` tag, `NotARating` is found to be not a canonical rating tag.)

This PR fixes this bug by adding the validation for canonical tags directly to `rating_string`, since the root problem is that `rating_string` is allowed to take any string possible, when we know it can only be one canonical rating. But, it does render the `Rating` model's validation essentially useless.

Another method is to do a validation that ensures a work can only have one `Rating` tag. But this still allows saving any string to `rating_string`, which I'm not sure if there's a reason for.

ellieyhcheng:
The two `rspec` tests that are removed are
- `rake After:fix_teen_and_up_imported_rating`
- `rake After:clean_up_noncanonical_ratings` 
These tasks are [tasks written just over two years ago](https://github.com/otwcode/otwarchive/commit/ad0fbced98ee3679dd4e190c6c33beb361ff93b2) that are no longer necessary with the added rating validation.

ceithir:
> These tasks are tasks written just over two years ago that are no longer necessary with the added rating validation.

Probably better to run them one last time in production once the rest of this PR is merged and deployed, then delete them for good only after?

[need an extra ticket to not forget to do that eventually though]

ceithir (commented):
Controller test to check the fix work as expected?

ceithir:
Thanks for the quite clear explanation to a twisted issue.

A possible way to use the Rating model validation instead of duplicating it might be to override the `split_tag_string` in Work to not split anything:
```rb
def split_tag_string(tag_string)
  # Work only accepts one (rating) tag
  [tag_string]    
end
```

ceithir:
Thanks for the quite clear explanation to a twisted issue.

Dunno if actually better, but a possible way to use the model validation instead of duplicating would be to add a separate validation for rating uniqueness:

Something like (untested pseudo-code):
```rb
class Work < ApplicationRecord

# [...]

validate :only_one_rating
def only_one_rating
  return unless split_tag_string(rating_string).count > 1

  errors.add(:rating_string, ts("must be unique."))
end
```
Note to self: Don't comment on PR when hungry because breakfast was too long ago and lunch is too far away, as it apparently ends with very fragmented answers.

ellieyhcheng:
@ceithir Thanks for the suggestion! I like your solution better, since it doesn't make the `Rating` model validation useless, and consolidates the canonical check in one place. 

I'll go ahead and adapt it and roll back the changes that are not longer necessary to implement this, including the task deletion. I'll also add a controller test.

ellieyhcheng:
The failing rspec tests are the tests for the two tasks I added back. They rely on the multi-rating bug to set up the environment. Since the tasks will be deleted after one more final run after this PR, do we want to leave it as is for this PR?

ceithir:
I think you can make the test pass simply by replacing a couple calls to `create` with [create_invalid](https://github.com/otwcode/otwarchive/blob/49581ccbc42a05805fdb697f50aad1fbfb4af9c0/spec/spec_helper.rb#L186).

As we have no idea when exactly these tasks will be run hopefully for the last time and then removed, likely better to keep maintaining their tests for now.

ceithir (approved):
Seems good, thanks for working on this.

PR #4614, opened 2023-08-21
AO3-5629 Switch from wkhtmltopdf to Calibre for PDF downloads

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5629

## Purpose

Switches from wkhtmltopdf to Calibre for generating our PDFs while attempting to maintain a similar appearance. (We can't update our operating system until we make this change.)

Once we've tested on staging, we should post beforehand to let users know PDFs will be changing in appearance (and explain that this change is necessary). Here are the difference I noticed:
* the title and work notes are on a separate page, not right below the meta
* there is slightly more space in the top/bottom margins
* the font on the meta, work title, and "summary" and "notes" headings is, at least on my system, a serif rather than a sans serif, _but_, based on past experience with our downloads, I think this might be due to different default system fonts on my computer and staging, so it may not be an issue 

## Testing Instructions

We'll want to download some PDF works and compare them to similar works on production:
* works with images
* works with embedded audio or video (using embeds from e.g., YouTube and Soundcloud as well as using the `video` and `audio` HTML tags)
* multi-chapter works, with work and chapter notes and chapter summaries 
* single chapter works, also ideally with notes and summaries 
* works that use a different alphabet than English, e.g. works in Chinese, Russian, and Hebrew

## References

## Credit

Sarken, she/her

### DIFF
--- a/.github/workflows/automated-tests.yml
+++ b/.github/workflows/automated-tests.yml
@@ -104,13 +104,6 @@ jobs:
           sudo apt-get install -y redis-server
           ./script/gh-actions/multiple_redis.sh
 
-      - name: Cache wkhtmltopdf package
-        if: ${{ matrix.tests.ebook }}
-        uses: actions/cache@v3
-        with:
-          path: wkhtmltopdf
-          key: wkhtmltopdf-${{ hashFiles('script/gh-actions/ebook_converters.sh') }}
-
       - name: Install ebook converters
         if: ${{ matrix.tests.ebook }}
         run: ./script/gh-actions/ebook_converters.sh
--- a/app/models/download_writer.rb
+++ b/app/models/download_writer.rb
@@ -1,4 +1,4 @@
-require 'open3'
+require "open3"
 
 class DownloadWriter
   attr_reader :download, :work
@@ -125,7 +125,7 @@ class DownloadWriter
   # creating an empty stylesheets directory.
   def get_web2disk_command
     [
-      "web2disk',
+      "web2disk",
       "--base-dir", download.assets_path,
       "--max-recursions", "0",
       "--dont-download-stylesheets",
--- a/app/models/download_writer.rb
+++ b/app/models/download_writer.rb
@@ -59,21 +59,7 @@ class DownloadWriter
 
   # Get the version of the command we need to execute
   def get_commands
-    download.file_type == "pdf" ? [get_pdf_command] :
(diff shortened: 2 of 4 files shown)


ceithir:
Ok, for reference for the next person working with PDF, things I had to do in order for the generation to run on my computer (standard Docker install):

1. Replace:
```rb
unless exit_status
```
With:
```rb
unless exit_status&.success?
```
To get logs instead of silent failures. [likely a good change in absolute?]

2. Print `stderr.read.dump` in the logs for easier debugging [that one might not be a good idea in non dev contexts due to the length of the error message being an unknown]

3. Add the `{"QTWEBENGINE_CHROMIUM_FLAGS" => "--no-sandbox"}` env var to the `ebook-convert` command call to get around the fact our local setup run too many things as root (ref: https://www.mobileread.com/forums/showthread.php?t=341473) [definitely not something that should ever leave a dev machine as is]

ceithir (approved):
Code changes seem good, and the few examples I exported locally looked fine. Definitely need thorough testing on staging though.

sarken:
@zz9pzza Do you want me to change `unless exit_status` to `unless exit_status&.success` here, as ceithir mentioned? I'm not sure how it'll affect anything you might do with the logs, so your call.

https://github.com/otwcode/otwarchive/blob/801ce0931c0dbb144f9631eed5e9cf743a6c8d00/app/models/download_writer.rb#L42-L58

potpotkettle:
I tried adding ``fonts-takao-gothic fonts-takao-mincho fonts-takao`` into the Dockerfile (after adding `QTWEBENGINE_CHROMIUM_FLAGS` mentioned above), and Japanese characters seemed to be rendered fine. Without fonts like those, anything in kanji showed up as a generic tofu.

I think this means that it will work in the real environment, as long as the right fonts are installed there.

zz9pzza:
> @zz9pzza Do you want me to change `unless exit_status` to `unless exit_status&.success` here, as ceithir mentioned? I'm not sure how it'll affect anything you might do with the logs, so your call.
> 
> https://github.com/otwcode/otwarchive/blob/801ce0931c0dbb144f9631eed5e9cf743a6c8d00/app/models/download_writer.rb#L42-L58

I am happy enough either way, we generally are not looking at error logs and looking at our exception handling code.

zz9pzza (approved):

PR #4615, opened 2023-08-21
AO3-5776 Remove mention of author from download afterword

Bilka2:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5776

## Purpose

Replace "author" with "creator" in the download afterword that encourages commenting.

Did not do the bonus because counting the work creators looked quite complicated, at least when looking at the logic that bylines do. But maybe I missed some easy way to get all of a work's creators/external creators/anon status.

## Testing Instructions

See Jira.

## References

#4607 for the change to `downloads/show.html.erb`

## Credit

Bilka (he/him)

### DIFF
--- a/app/views/downloads/_download_afterword.html.erb
+++ b/app/views/downloads/_download_afterword.html.erb
@@ -23,5 +23,5 @@
     </div>
   <% end %>
 
-  <p class="message"><%= t(".please_comment_html", work_comment_link: link_to(t(".work_comment"), new_work_comment_url(@work))) %></p>
+  <p class="message"><%= t(".please_comment_html", work_comment_link: link_to(t(".work_comment", locale: :en), new_work_comment_url(@work)), locale: :en) %></p>
 </div>
--- a/app/views/downloads/_download_afterword.html.erb
+++ b/app/views/downloads/_download_afterword.html.erb
@@ -23,5 +23,5 @@
     </div>
   <% end %>
 
-  <p class="message"><%= ts("Please") %> <%= link_to ts("drop by the archive and comment"), new_work_comment_url(@work) %> <%= ts("to let the author know if you enjoyed their work!") %></p>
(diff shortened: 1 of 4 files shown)


config/locales/views/en.yml
```diff
@@ -375,6 +375,10 @@ en:
           disable_anon: Sorry, this work doesn't allow non-Archive users to comment.
           hidden: Sorry, you can't add or edit comments on a hidden work.
           unrevealed: Sorry, you can't add or edit comments on an unrevealed work.
+  downloads:
+    download_afterword:
+      please_comment_html: Please %{work_comment_link} to let the creator know if you enjoyed their work!
+      work_comment: drop by the archive and comment
```
weeklies, line 381: (not AD&T) Since we're editing it, I think it would be good to capitalise the "Archive" as a proper noun.

Bilka2: Good point, the newest texts indeed capitalize it. If AD&T says that it's wrong it can be changed back.

sarken (approved):
Thanks!

app/views/downloads/_download_afterword.html.erb
```diff
@@ -23,5 +23,5 @@
     </div>
   <% end %>
 
-  <p class="message"><%= ts("Please") %> <%= link_to ts("drop by the archive and comment"), new_work_comment_url(@work) %> <%= ts("to let the author know if you enjoyed their work!") %></p>
+  <p class="message"><%= t(".please_comment_html", work_comment_link: link_to(t(".work_comment"), new_work_comment_url(@work))) %></p>
```
sarken, line 26: So @brianjaustin and I were talking about i18n and download caching, and we think it might be a good idea to explicitly pass the locale for now. Could you add `locale: :en` here, [like we do with zh-CN in the proxy notice](https://github.com/otwcode/otwarchive/blob/801ce0931c0dbb144f9631eed5e9cf743a6c8d00/app/views/layouts/_proxy_notice.html.erb#L9)?

zz9pzza (approved):

PR #4616, opened 2023-08-22
Omit -b option on macOS

eliahhecht:
## Purpose

Fixes the init.sh script on macOS, which does not support `cp -b`

## Credit

Eliah Hecht, he/him

### DIFF
--- a/script/docker/init.sh
+++ b/script/docker/init.sh
@@ -5,14 +5,12 @@ set -ex
 # Change directory to root of the repo
 cd "$(dirname "$0")/../.."
 
-copy="cp -b"
-if [[ "$OSTYPE" == "darwin"* ]]; then
-    copy="cp" # macOS cp does not support the -b option
-fi
-
-$copy config/docker/database.yml config/database.yml
-$copy config/docker/redis.yml config/redis.yml
-$copy config/docker/local.yml config/local.yml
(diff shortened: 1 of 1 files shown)


script/docker/init.sh
```diff
@@ -5,9 +5,14 @@ set -ex
...
-cp -b config/docker/redis.yml config/redis.yml
-cp -b config/docker/local.yml config/local.yml
+copy="cp -b"
+if [[ "$OSTYPE" == "darwin"* ]]; then
+    copy="cp" # macOS cp does not support the -b option
+fi
+
+$copy config/docker/database.yml config/database.yml
+$copy config/docker/redis.yml config/redis.yml
+$copy config/docker/local.yml config/local.yml
```
ceithir, line 15: Might be better to keep the same behavior across all versions by going verbose?
Something like:
```sh
for file in 'database.yml' 'redis.yml' 'local.yml'
do
  # Manual backup as the --backup option is not available for all versions of cp
  test -f "config/$file" && cp "config/$file" "config/$file~"
  cp "config/docker/$file" "config/$file"
done
```

eliahhecht: Yeah that makes sense! Done.

ceithir (approved):

zz9pzza (approved):

PR #4618, opened 2023-08-23
AO3-5774 Pluralize tag categories in downloaded work meta

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5774

## Purpose

Pluralise the tag category names in downloaded work metadata, the same as they are for the online view.

NOTE: this is not very internationalisation-friendly (particularly the hardcoding of only 1 being singular) but neither is the existing meta code, and I don't know that we have finalised the plan for translating downloads

## Credit

Brian Austin (they/he)


app/views/downloads/_download_preface.html.erb
```diff
@@ -14,7 +14,7 @@
       <% Tag::VISIBLE.each do |type| %>
         <% tags = @work.tag_groups[type] %>
         <% unless tags.blank? %>
-          <dt><%= type.constantize::NAME %>:</dt>
+          <dt><%= tags.size == 1 ? type.constantize::NAME : type.constantize::NAME.pluralize %>:</dt>
```
Bilka2, line 17:
```suggestion
          <dt><%= type.constantize.human_attribute_name("name_with_colon", count: tags.size) %></dt>
```
This would be more i18n friendly, it's already used by the challenge_assignment_notification mailer.

brianjaustin: I'm resolving this for now; as we discussed elsewhere, there are a few issues with translating this:
* In general, translating downloads will not work currently. They are cached (without language key), so everything has to be forced to `en`
* `Additional Tags` needs to be hardcoded as plural (not a big issue on its own)
* `Archive Warnings` already has a `human` name of just `Warnings`

We will need to solve for these problems eventually, but they are out of scope for now. If I missed something there, please feel free to un-resolve or ping me to do it.

Bilka2 (approved):

zz9pzza (approved):

PR #4619, opened 2023-08-23
AO3-6596 Don't show an error on first login help

eliahhecht:
If logged out, rather than showing an error on the "first login help" page, show the usual message that the users needs to log in

## Issue

https://otwarchive.atlassian.net/browse/AO3-6596

## Purpose

When the user is logged out and tries to access /first_login_help, don't show them an error; instead redirect them to login with a message that they need to log in to access that page

## Testing Instructions

Attempt to access /first_login_help while logged out. You should see a message indicating that you should log in (rather than an error page)

## Credit

Eliah Hecht, he/him


eliahhecht:
> Test might be slightly shorter and simpler as a controller test ([example](https://github.com/otwcode/otwarchive/blob/801ce0931c0dbb144f9631eed5e9cf743a6c8d00/spec/controllers/chapters_controller_spec.rb#L49-L54)) but it's also okay to leave it as an end-to-end test.

When I tried this (in a TDD fashion by commenting out my change to the controller), the controller test passed and said it got a 200 response from the controller even if there was no user logged in. Maybe in an rspec environment, it doesn't render the view? When I try it in a browser it gets a 500 (if not logged in & without my change).

ceithir:
> Maybe in an rspec environment, it doesn't render the view?
Likely, though I'm not specialist enough to bet on it.

However, such a test, on this line:
```
```
Should still fail with the old code and pass with the new one, no?

eliahhecht:
Yes, but given that the test doesn't fail for the right reason in the previous code, I'm skeptical to write it -- that kinda feels like a "wrong test" to me. But I can do it if you feel strongly.

features/other_a/help.feature
```diff
@@ -30,4 +30,13 @@ Scenario: view the help popup for chapter title
     And I press "Post"
     And I follow "Add Chapter"
     And I follow "Chapter title"
-  Then I should see "You can add a chapter title"
\ No newline at end of file
+    Then I should see "You can add a chapter title"
+
+  Scenario: Scenario: Asked to log in if trying to access the first login page as guest
```
brianjaustin, line 35: Nitpick: it looks like the word `Scenario` crept in twice here

brianjaustin (approved):

PR #4620, opened 2023-08-23
AO3-6594 don't show location section if empty

eliahhecht:
If the user's location is set to empty-string, omit the "I live in" section on their profile

## Issue

https://otwarchive.atlassian.net/browse/AO3-6594

## Purpose

Hides the "I live in" section on the profile if the location is set to empty-string

## Testing Instructions

Update your location to empty, make sure the corresponding section does not display on the profile view.

## Credit

Eliah Hecht, he/him

### DIFF
--- a/app/views/profile/show.html.erb
+++ b/app/views/profile/show.html.erb
@@ -14,7 +14,7 @@
       <dd><%= l(@profile.created_at) %></dd>
       <dt><%= ts("My user ID is:") %></dt>
       <dd><%= @user.id %></dd>
-      <% if @profile.location and !@profile.location.empty? %>
+      <% if @profile.location.present? %>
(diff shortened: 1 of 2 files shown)


app/views/profile/show.html.erb
```diff
@@ -14,7 +14,7 @@
       <dd><%= l(@profile.created_at) %></dd>
       <dt><%= ts("My user ID is:") %></dt>
       <dd><%= @user.id %></dd>
-      <% if @profile.location %>
+      <% if @profile.location and !@profile.location.empty? %>
```
weeklies, line 17: I think `<% if @profile.location.present? %>` would be more concise.

brianjaustin (approved):

PR #4621, opened 2023-08-23
AO3-6265 Use text-align start in comments

eliahhecht:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6265

## Purpose

Fixes the issue where RTL text within comments is incorrectly left-aligned.

I was unable to think of a good way to automatically test this -- open to suggestions. Here's a screenshot from my dev environment though:

46bb8e187da1e021351e5c371eb5fbed

## Testing Instructions

Write a comment on work with text like `<p dir="rtl">Some text goes here</p>`, and make sure it appears right-aligned within the comment box.

## Credit

Eliah Hecht, he/him

### DIFF
--- a/public/stylesheets/site/2.0/15-group-comments.css
+++ b/public/stylesheets/site/2.0/15-group-comments.css
@@ -15,6 +15,7 @@ div.comment, li.comment {
   position: relative;
   border: 1px solid #ddd;
   overflow: visible;
+  text-align: start;
 }
 
 .comment:after,


sarken (approved):
Looks good to me! No need to worry about automated tests for style issues; we just do manual testing for them.

Just a heads up you have the wrong issue number in your title and text, though -- it looks like this is actually [AO3-6265](https://otwarchive.atlassian.net/browse/AO3-6265), like your branch name.

PR #4626, opened 2023-09-03
AO3-5522 Limit number of password resets

brianjaustin:
# Pull Request Checklist

* [X] Have you read ["How to write the perfect pull request"](https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/)?
* [X] Have you read the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [X] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)?
* [X] Have you added the [Jira](https://otwarchive.atlassian.net) issue number
  as the *first* thing in your pull request title (e.g. `AO3-1234 Fix thing`)
* [X] Do you have fewer than 5 pull requests already open? If not, please wait
  until they are reviewed and merged before creating new pull requests.

## Issue

https://otwarchive.atlassian.net/browse/AO3-5522

## Purpose

This adds a limit to the number of password resets that can be requested within a period of time. If the limit is reached, then reset requests will be prevented until the cooldown expires.

A new users table column is added:

    resets_requested: The number of resets that have been requested. Integer, non-nullable, indexed, defaults to 0.

Two new archive configuration settings are added:

    PASSWORD_RESET_LIMIT: The limit to the number of resets within the PASSWORD_RESET_COOLDOWN_HOURS period. Defaults to 3.
    PASSWORD_RESET_COOLDOWN_HOURS: The period of time in which the reset request limit will be enforced. Defaults to 12.

## References

This PR adopts the changes from #4265

## Credit

Hunter Ada Smith (she/her)
Brian Austin (they/he)

PR #4630, opened 2023-09-10
AO3-5522 Password reset limit fixes

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5522

## Purpose

This fixes a few issues from the original PR (#4626):
1. The message shown when a reset link is sent now includes the number of attempts remaining, plus the configured cool-down time
2. When a user requests the maximum number of reset emails, changes their password when logged in, then requests another reset, the process no longer returns a 400 error. This happened because Devise resets the `reset_password_sent_at` to `nil` when a user changes their password: https://github.com/heartcombo/devise/blob/9f80dc2562524f744e8633b8562f2a0114efb32b/lib/devise/models/recoverable.rb#L103
3. (a secret third bug we hadn't gotten to in QA yet!) if a user requests the maximum number of reset emails, waits, and then requests a new reset, they will be able to request more. Previously, they could only get 1 more due to the order of when we checked things.

## Credit

Brian Austin (they/he)


app/controllers/users/passwords_controller.rb
```diff
@@ -21,13 +21,35 @@ def create
...
-      else
-        user.update_password_resets_requested
-        user.save
+    if user.present? && !user.new_record?
+      user.update_password_resets_requested
+      user.save
+    end
+
+    super do |target_user|
+      if target_user.nil? || target_user.new_record?
```
weeklies, line 30: I wonder if we could do a return early on in the method if user is nil so that we no longer need to use `user&.` in the lines above this.

brianjaustin: It does change the behaviour ever so slightly (if a non-existent user, the previous method would show an error on the field in addition to flash) but I _think_ that should be OK?

app/controllers/users/passwords_controller.rb
```diff
@@ -8,11 +8,15 @@ class Users::PasswordsController < Devise::PasswordsController
 
   def create
     user = User.find_for_authentication(resource_params.permit(:login))
+    if user.nil? || user.new_record?
+      flash[:notice] = t(".user_not_found")
```
weeklies, line 12: We should make this notice an error instead,  since the field error is no longer in effect.

app/controllers/users/passwords_controller.rb
weeklies: I don't think it's a bad thing considering the text of the field error starts with 

> Sorry! We couldn't save this user because:

which is a little confusing.

brianjaustin: Agreed!

app/controllers/users/passwords_controller.rb
```diff
@@ -21,13 +25,26 @@ def create
...
+    user.update_password_resets_requested
+    user.save
+
+    super
+  end
+
+  protected
+
+  # We need to include information about the user (the remaining reset attempts)
+  # in addition to the Archive's configured reset cooldown in the success
```
sarken, line 37: Can we lowercase or remove "Archive" (e.g., Squidge) for the not-us folks who are using the code?

app/models/concerns/password_resets_limitable.rb
```diff
@@ -2,8 +2,15 @@ module PasswordResetsLimitable
...
+    def password_resets_remaining
+      return ArchiveConfig.PASSWORD_RESET_LIMIT unless self.last_reset_within_cooldown?
+
+      limit_delta = ArchiveConfig.PASSWORD_RESET_LIMIT - self.resets_requested
+      limit_delta.positive? ? limit_delta : 0
+    end
+
     def password_resets_limit_reached?
-      self.resets_requested >= ArchiveConfig.PASSWORD_RESET_LIMIT && self.last_reset_within_cooldown?
+      password_resets_remaining < 1
```
sarken, line 13: Are there circumstances this could be less than 1 but not equal 0?

app/models/concerns/password_resets_limitable.rb
```diff
@@ -17,12 +24,19 @@ def update_password_resets_requested
         self.resets_requested += 1
       end
     end
+
+    protected
+
+    def clear_reset_password_token
```
sarken, line 30: I think we could probably use a code comment here, explaining that this is a Devise method and why we need to change it.

sarken (approved):

PR #4631, opened 2023-09-10
AO3-6605 Add Reviewdog for Rubocop check

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6605

## Purpose

Add a Reviewdog action to run Rubocop. This will give us more flexibility to add custom cops and upgrade to the latest version of rubcop (and its plugins). Example of how Rubocop warnings will show up: https://github.com/brianjaustin/ao3-sandbox/pull/1/files (the specific message is from a custom cop).

Note: no changes in _what_ Rubocop flags are expected yet. Custom cops (and maybe other plugins?) will come in later PR(s)

## Follow-Up Instructions
1. Allow-list the action in this Github repository's settings (similar to #4604)
~2. Deactivate/remove this repository from our Hound account -- optional but recommended to avoid duplicate comments and errors when we upgrade or add custom cops~

## Credit

Brian Austin (they/he)

### DIFF
--- a/.github/workflows/reviewdog.yml
+++ b/.github/workflows/reviewdog.yml
@@ -25,8 +25,6 @@ jobs:
         with:
           use_bundler: true
           reporter: github-pr-check
-          rubocop_version: gemfile
-          rubocop_extensions: rubocop-rails:gemfile rubocop-rspec:gemfile
           skip_install: true
 
   erb-lint:
--- a/.github/workflows/reviewdog.yml
+++ b/.github/workflows/reviewdog.yml
@@ -27,6 +27,7 @@ jobs:
           reporter: github-pr-check
           rubocop_version: gemfile
           rubocop_extensions: rubocop-rails:gemfile rubocop-rspec:gemfile
+          skip_install: true
 
   erb-lint:
     name: ERB Lint runner
--- a/.github/workflows/reviewdog.yml
+++ b/.github/workflows/reviewdog.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Check out code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: Set up Ruby and run bundle install
         uses: ruby/setup-ruby@v1
--- a/.github/workflows/reviewdog.yml
+++ b/.github/workflows/reviewdog.yml
@@ -8,6 +8,26 @@ permissions:
   checks: write
 
 jobs:
+  rubocop:
+    name: Rubocop
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
+
+      - name: Set up Ruby and run bundle install
+        uses: ruby/setup-ruby@v1
+        with:
+          bundler-cache: true
+
+      - name: rubocop
+        uses: reviewdog/action-rubocop@e70b014b8062c447d6b515ee0209f834ea93e696
+        with:
+          use_bundler: true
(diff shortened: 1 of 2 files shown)


.github/workflows/reviewdog.yml
```diff
@@ -8,6 +8,26 @@ permissions:
   checks: write
 
 jobs:
+  rubocop:
+    name: Rubocop
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
```
weeklies, line 16: Can we use v4 like https://github.com/otwcode/otwarchive/pull/4627?

brianjaustin: :facepalm: that was on my mental list of things to do, and I totally forgot. Thanks for reminding me!

Bilka2:
> Deactivate/remove this repository from our Hound account -- optional but recommended

According to `.hound.yml`, Hound should be running jshint. This leaves the question whether it runs as configured or if, like ERB Lint, it doesn't actually run. But in any case, `.hound.yml` should be updated to disable rubocop.

brianjaustin:
@Bilka2 good point! I have gotten JShint checks on a PR in the past, so I think it still works (at least somewhat)? I will also update the instructions to not totally remove Hound yet

.github/workflows/reviewdog.yml
```diff
@@ -8,6 +8,26 @@ permissions:
...
+
+      - name: Set up Ruby and run bundle install
+        uses: ruby/setup-ruby@v1
+        with:
+          bundler-cache: true
+
+      - name: rubocop
+        uses: reviewdog/action-rubocop@e70b014b8062c447d6b515ee0209f834ea93e696
+        with:
+          use_bundler: true
```
weeklies, line 26: I am not sure about this, but will adding a `skip_install: true` be an improvement according to a suggestion [here](https://github.com/reviewdog/action-rubocop/issues/83)?

.hound.yml
```diff
@@ -4,7 +4,3 @@
 jshint:
   config_file: .jshintrc
   ignore_file: .jshintignore
-
-rubocop:
-  version: 1.22.1
-  config_file: .rubocop.yml
```
Bilka2, line 10: Sorry, I forgot to mention that to disable rubocup, it needs to be explicitly be disabled because it [defaults to being enabled](http://help.houndci.com/en/articles/2137566-rubocop):
```
rubocop:
  enabled: false
```

.github/workflows/reviewdog.yml
brianjaustin: I tried this out on my sandbox repo, and it looks like it worked fine with this enabled!

.github/workflows/reviewdog.yml
```diff
@@ -8,6 +8,27 @@ permissions:
...
+          reporter: github-pr-check
+          rubocop_version: gemfile
+          rubocop_extensions: rubocop-rails:gemfile rubocop-rspec:gemfile
```
Bilka2, line 29: With the skip install configuration, these lines are no longer needed.
```suggestion
```

Bilka2 (changes requested):
Looks good and works on my fork. The suggested change would be nice to have though.

redsummernight (approved):

redsummernight:
I've added `reviewdog/action-rubocop@e70b014b8062c447d6b515ee0209f834ea93e696,` to the repo setting "Allow specified actions and reusable workflows".

PR #4628, opened 2023-09-04
AO3-6467 Track deletion of previous fnok on update

ceithir:
# Pull Request Checklist

* [x] Have you read ["How to write the perfect pull request"](https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/)?
* [x] Have you read the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [x] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)?
* [x] Have you added the [Jira](https://otwarchive.atlassian.net) issue number as the *first* thing in your pull request title (e.g. `AO3-1234 Fix thing`)
* [x] Do you have fewer than 5 pull requests already open? If not, please wait until they are reviewed and merged before creating new pull requests.

## Issue

https://otwarchive.atlassian.net/browse/AO3-6467?focusedCommentId=360436

## Purpose

Add the missing log when a fnok is replaced (akin to one deletion and one addition) 

## Testing Instructions

1. Add AccountB as an FNOK for AccountA. Check it shows in both the logs of AccountA and AccountB
2. Update the FNOK for AccountA to be AccountC. Check the deletion of previous fnok shows in the logs of AccountA and AccountB, and the addition of the new fnok shows in the log of AccountA and AccountC.

## Credit

Ceithir (he/him)

PR #4633, opened 2023-09-20
AO3-6467 Log on kin side when a user with a fnok is deleted

ceithir:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6467?focusedCommentId=360484

## Purpose

The logging of fannish next of kin addition and removal as previously coded was trying to be too clever, only saving data on the forebear side then inferring what to display on their successor side from that. This caused not only convoluted code but also bugs in the context of user deletion.

This PR moved to a more straightforward if more verbose approach by explicitly logging any change on both sides.

## Testing Instructions

As per the JIRA

## References

Untouched (as the visible behavior didn't change save for one bugfix) but nonetheless relevant [end-to-end tests](https://github.com/otwcode/otwarchive/blob/317e3e840ce28f1d66e2ede43fdbd3e6bf0218bd/features/admins/users/admin_fnok.feature)

## Credit

Ceithir (he/him)
Note: weeklies (she/her) also [contributed code](https://github.com/otwcode/otwarchive/pull/4628/commits/64a4d2e41e03c3c2ff348268b92fb258a2d3a668) to this feature in a previous PR


lib/user_history.rb
```diff
@@ -0,0 +1,52 @@
+module UserHistory
+  def log_assignment_of_next_of_kin(user, kin, admin:)
```
weeklies, line 2: If admin is a required argument, is the `:` needed?

app/models/user.rb
```diff
@@ -173,15 +175,17 @@ def remove_user_from_kudos
...
-      fnok.user.create_log_item({
-                                  action: ArchiveConfig.ACTION_REMOVE_FNOK,
-                                  fnok_user_id: self.id
-                                })
+      log_removal_of_next_of_kin(fnok.user, self)
     end
   end
 
+  def log_removal_of_self_as_previous_of_kin
+    successor = fannish_next_of_kin&.kin
```
weeklies, line 185: Successor is a great name, thank you!

app/models/user.rb
```diff
@@ -173,15 +175,17 @@ def remove_user_from_kudos
...
     fannish_next_of_kins.each do |fnok|
-      fnok.user.create_log_item({
-                                  action: ArchiveConfig.ACTION_REMOVE_FNOK,
-                                  fnok_user_id: self.id
-                                })
+      log_removal_of_next_of_kin(fnok.user, self)
     end
   end
 
+  def log_removal_of_self_as_previous_of_kin
```
weeklies, line 184: I get why it's `previous_of_kin`, but I am not sure it is intuitively used in English. I would go with putting the two removal steps in one callback, because the named variables have helped to improve the clarity.

weeklies (commented):
Just some general comments.

lib/user_history.rb
brianjaustin: I think the `:` is to mark `admin` as a (required) keyword argument, rather than a positional one: https://docs.ruby-lang.org/en/master/syntax/methods_rdoc.html#label-Keyword+Arguments

weeklies: Oh! TIL.

lib/user_history.rb
```diff
@@ -0,0 +1,52 @@
+module UserHistory
```
weeklies, line 1: Just realized: We do have a readings history feature, so maybe calling it KinHistory/KinLogs will be better?

lib/kin_history.rb
```diff
@@ -0,0 +1,61 @@
+module KinHistory
```
sarken, line 1: Would it be better to make this module a concern named something like `Loggable`? (I truly don't know the answer; it was just my first thought upon seeing `self.included`.)

ceithir: <insert usual quote about naming and caching here>

Would likely be more consistent with existing modules indeed. Blanking at the moment on an adjective that isn't a mouthful like KinChangeTrackable` however.

sarken: I think just `Loggable` or `UserLoggable` is fine, since it's handling things logged in the `log_items` table. I don't think there's a reason it has to be reserved for FNOK items.

sarken: There's a bit more we should do when making this module into a concern:

- It should go in `app/models/concerns`, since the model is where we're including our concern. (There's also `app/controllers/concerns` for when you're including the concern module in a controller.)
- It should have `extend ActiveSupport::Concern` on the line after `module UserLoggable`.
- Now we can use an [`included` block](https://api.rubyonrails.org/classes/ActiveSupport/Concern.html#method-i-included) for the callback.

sarken (approved):

PR #4635, opened 2023-09-23
AO3-6487 More attempts at fixing autocompletes after a fandom is made a synonym

ceithir:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6487

## Purpose

Shots in the dark at fixing the issue described in the ticket.

Because, while it's not issue reproducing it on staging:

Manually fixed tag:
```
curl -u "LOGIN:PASSWORD" "https://test.archiveofourown.org/autocomplete/character_in_fandom?term=bob&fandom[]=Your%20Organic%20Love" -H "Accept: application/json"
```

Tag left as is after synonymizing:
```
curl -u "LOGIN:PASSWORD" "https://test.archiveofourown.org/autocomplete/character_in_fandom?term=lindsey&fandom[]=Your%20Organic%20Love" -H "Accept: application/json"
```

~~I haven't been able to recreate the conditions for it in tests.~~

~~So trying to:~~
~~- Force the autocomplete refresh each time the association is updated, not only on create. It makes sense and is consistent with how the reindex works.~~
~~- Add an explicit autocomplete refresh at the end of the merging operation ("belts and suspenders")~~

-> Change of approach, by going granular instead, with unit tests checking if the autocomplete is updated properly when creating/deleting an association. Did find and fix some bugs at that level, though still unsure that's the cause of what's happening in staging/production.

## Testing Instructions

1. Rename a fandom by making it a synonym of a new (canonical) fandom
2. When creating a work for that fandom, check the autocomplete is only limited to that fandom tags

## Credit

Ceithir (he/him)

### DIFF
--- a/app/models/common_tagging.rb
+++ b/app/models/common_tagging.rb
@@ -30,7 +30,7 @@ class CommonTagging < ApplicationRecord
   end
 
   def update_child_autocomplete
-    common_tag&.refresh_autocomplete
+    common_tag&.add_to_parent_fandoms_autocomplete
   end
 
   # A relationship should inherit its characters' fandoms
--- a/app/models/common_tagging.rb
+++ b/app/models/common_tagging.rb
@@ -36,7 +36,7 @@ class CommonTagging < ApplicationRecord
   end
 
   def remove_from_autocomplete
-    common_tag&.remove_from_fandom_autocomplete(filterable)
+    common_tag.remove_from_fandom_autocomplete(filterable)
   end
 
   # A relationship should inherit its characters' fandoms
--- a/app/models/common_tagging.rb
+++ b/app/models/common_tagging.rb
@@ -36,7 +36,7 @@ class CommonTagging < ApplicationRecord
   end
 
   def remove_from_autocomplete
-    common_tag.remove_from_fandom_autocomplete(filterable)
+    common_tag&.remove_from_fandom_autocomplete(filterable)
   end
 
   # A relationship should inherit its characters' fandoms
--- a/app/models/common_tagging.rb
+++ b/app/models/common_tagging.rb
@@ -30,7 +30,7 @@ class CommonTagging < ApplicationRecord
   end
 
   def update_child_autocomplete
-    common_tag.refresh_autocomplete
+    common_tag&.refresh_autocomplete
   end
 
   # A relationship should inherit its characters' fandoms
--- a/app/models/common_tagging.rb
+++ b/app/models/common_tagging.rb
@@ -21,9 +21,9 @@ class CommonTagging < ApplicationRecord
   after_create :remove_uncategorized_media
   after_create :add_to_autocomplete
 
-  after_commit :update_search
-
   before_destroy :remove_from_autocomplete
+
+  after_commit :update_search
 
   def update_wrangler
     unless User.current_user.nil?
--- a/app/models/common_tagging.rb
+++ b/app/models/common_tagging.rb
@@ -32,11 +32,15 @@ class CommonTagging < ApplicationRecord
   end
 
   def add_to_autocomplete
-    common_tag.add_to_fandom_autocomplete(filterable)
+    if filterable.is_a?(Fandom) && common_tag.eligible_for_fandom_autocomplete?
+      common_tag.add_to_fandom_autocomplete(filterable)
+    end
   end
 
   def remove_from_autocomplete
-    common_tag&.remove_from_fandom_autocomplete(filterable)
+    if filterable.is_a?(Fandom) && common_tag&.was_eligible_for_fandom_autocomplete?
+      common_tag.remove_from_fandom_autocomplete(filterable)
+    end
   end
 
   # A relationship should inherit its characters' fandoms
--- a/app/models/common_tagging.rb
+++ b/app/models/common_tagging.rb
@@ -32,15 +32,15 @@ class CommonTagging < ApplicationRecord
   end
 
   def add_to_autocomplete
-    if filterable.is_a?(Fandom) && common_tag.eligible_for_fandom_autocomplete?
-      common_tag.add_to_fandom_autocomplete(filterable)
-    end
+    return unless filterable.is_a?(Fandom) && common_tag.eligible_for_fandom_autocomplete?
+
+    common_tag.add_to_fandom_autocomplete(filterable)
   end
 
   def remove_from_autocomplete
-    if filterable.is_a?(Fandom) && common_tag&.was_eligible_for_fandom_autocomplete?
-      common_tag.remove_from_fandom_autocomplete(filterable)
-    end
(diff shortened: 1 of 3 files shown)


app/models/tag.rb
```diff
@@ -817,6 +817,8 @@ def transfer_or_remove_associations
       self.sub_taggings.destroy_all
       self.meta_taggings.destroy_all
     end
+
+    refresh_autocomplete
```
brianjaustin, line 821: Would this plus the above result in refreshing autocomplete for tags twice?

brianjaustin (commented):
Do you think it would be worth adding tests for the new behaviour?

app/models/tag.rb
ceithir: Oops, sorry, wrong code: This does absolutely nothing.

That's the one that was supposed to be in the PR: https://github.com/otwcode/otwarchive/pull/4635/commits/f190e82f3b27da02693c0d26caa7b9fa7b55f7f9

And yes, the idea is to refresh the autocomplete for all children at the end, in a separate async task for good measure, even if it means doing it twice potentially.

I tested it through the technique of "commenting the after_commit in common_tagging". Not sure how to write a clean test for that case as I'm actually not sure how it's even possible for the issue to arise in real life.

brianjaustin: Whoops! In theory, we could do something with mocks/spies, but at that point the argument in favour of a unit test starts to get a little bit weaker (maybe still nice for documentation purposes? but even then tests with reliance on mocking can be tricky).

To take a slightly different angle on my original question: do we need https://github.com/otwcode/otwarchive/pull/4635/files#diff-b361dece27e8fd54e644111ffe427e4bc584369b48b920c781c16803ed65cd4bR24 alongside the code from https://github.com/otwcode/otwarchive/commit/f190e82f3b27da02693c0d26caa7b9fa7b55f7f9? It seems to me that the latter would be the most comprehensive, but it's possible I forgot something about autocomplete

ceithir:
Okay, I think I got something. Sorry for the mess, PR should definitely have stayed a draft for longer.

So, what's happening?

Each time a Character (or Relationship) tag is updated, if it's canonical, there's a bit of code refreshing its autocomplete value in Redis for each of its parent fandoms. Nothing too complicated, just a loop on all its parents, removing the old value from the set and adding the new one.

This works well for things like a character being renamed. However, it does not really deal with things like a character being added to a fandom, as this technically does not update the character tag. It still kinda worked due to the [wrangler tracking](https://github.com/otwcode/otwarchive/blob/317e3e840ce28f1d66e2ede43fdbd3e6bf0218bd/app/models/common_tagging.rb#L28) luckily triggering the after_update, or more, recently, an [explicit call](https://github.com/otwcode/otwarchive/blob/317e3e840ce28f1d66e2ede43fdbd3e6bf0218bd/app/models/common_tagging.rb#L33) to that refresh routine.

The later however turned to not only be a bit dump (as we were refreshing for every potential fandom when really all we needed was for one key to be added to one) but also to sometimes simply not work due to the in memory Character tag having yet to register its new Fandom parent tag and thus unable to loop on it (no reload mid process).

This PR cleans up all that by having an explicit unitary add on creation, tested its own unit test (that's indeed very much not passing without the change). Back to basics in a way.

I'm sadly unsure this will fix the production bug as I haven't been able to design an end-to-end test reproducing it. But behavior should at least be more consistent.

brianjaustin (commented):
Thanks for doing such a thorough investigation! I know from experience that locally testing autocomplete is unfortunately finicky, so I think we may have to just test on staging. Over all looks reasonable to me, just 1 (hopefully quick) question

app/models/common_tagging.rb
```diff
@@ -29,8 +31,12 @@ def update_wrangler
...
   end
 
-  def update_child_autocomplete
-    common_tag.refresh_autocomplete
+  def add_to_autocomplete
+    common_tag.add_to_fandom_autocomplete(filterable)
+  end
+
+  def remove_from_autocomplete
+    common_tag&.remove_from_fandom_autocomplete(filterable)
```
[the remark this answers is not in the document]
ceithir: A CommonTag cannot be created in a way that triggers callbacks without a valid common_tag attribute (cf validates_presence_of a few lines above), so we don't have to worry about it not existing in the after_create.

On the other hand, [this test](https://github.com/otwcode/otwarchive/blob/317e3e840ce28f1d66e2ede43fdbd3e6bf0218bd/spec/lib/tasks/tag_tasks.rake_spec.rb#L44) fails without the `&`. And it suggests CommonTagging can become orphans, maybe as a consequence of some batch process deleting tags without triggering their cleanup callbacks, and the code should be ready for it. Likely, update_search relies on a & for the same reason.

brianjaustin (approved):

app/models/tag.rb
```diff
@@ -471,24 +471,35 @@ def autocomplete_prefixes
...
+    return unless fandom.is_a?(Fandom)
+    return unless self.is_a?(Character) || self.is_a?(Relationship)
+
+    score ||= autocomplete_score
+    REDIS_AUTOCOMPLETE.zadd(self.transliterate("autocomplete_fandom_#{fandom.name.downcase}_#{type.downcase}"), score, autocomplete_value)
+  end
+
   def remove_from_autocomplete
     super
-    if self.is_a?(Character) || self.is_a?(Relationship)
```
weeklies, line 485: If this is removed and shifted into       `remove_from_fandom_autocomplete` (and its add counterpart), this might lead to more iterations than necessary. Would an early return be a significant improvement?

app/models/tag.rb
```diff
@@ -471,24 +471,41 @@ def autocomplete_prefixes
...
+  def remove_from_fandom_autocomplete(fandom)
+    REDIS_AUTOCOMPLETE.zrem(self.transliterate("autocomplete_fandom_#{fandom.name.downcase}_#{type.downcase}"), autocomplete_value)
+  end
+
+  def eligible_for_fandom_autocomplete?
+    (self.is_a?(Character) || self.is_a?(Relationship)) && canonical
+  end
+
+  def was_eligible_for_fandom_autocomplete?
+    (self.is_a?(Character) || self.is_a?(Relationship))
```
weeklies, line 506: Nitpick but the outer parentheses aren't necessary.

app/models/tag.rb
```diff
@@ -471,24 +471,41 @@ def autocomplete_prefixes
...
+  end
+
   def remove_stale_from_autocomplete
     super
     if self.is_a?(Character) || self.is_a?(Relationship)
```
weeklies, line 511: Do you think we can use `return unless was_eligible_for_fandom_autocomplete?` here?

weeklies (commented):
Just minor nitpicks.

sarken:
@weeklies -- just checking, are you happy with the changes? If so, I'll merge this.

weeklies:
@sarken it looks good to me.

PR #4636, opened 2023-10-01
AO3-6616 Fix floating and displaced form elements in new webkit

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6616

## Purpose

webkit did something terrible that affected the position of various form elements, most notably:

- the bottom Sort and Filter button on the filters has relocated to the top
- the notes options on the posting form has migrated upward to interfere with other form options

Full details are in the issue.

## Testing Instructions

We'll want to test Safari 17 to make sure the issue is fixed _and_ at least one older version of Safari, plus Firefox and Chrome (any version), to make sure we didn't break anything new.

## References

I diagnose the issue in [this comment](https://otwarchive.atlassian.net/browse/AO3-6616?focusedCommentId=360502).

## Credit

Sarken, she/her

### DIFF
--- a/public/stylesheets/site/2.0/07-interactions.css
+++ b/public/stylesheets/site/2.0/07-interactions.css
@@ -126,6 +126,10 @@ form dt {
   border-bottom: 1px solid #f3efec;
 }
 
+form dt.landmark, form dt.notes, .dashboard form dt.landmark {
+  clear: both;
+}
+
 form dd, form dd.any {
   float: left;
   width: 72.5%;
@@ -142,6 +146,10 @@ form dd + dd {
 
 form dd.required {
   color: #2a2a2a;
+}
+
+form dd.notes {
+  clear: none;
 }
 
 form .footnote code {
@@ -682,6 +690,7 @@ form.single .autocomplete + input + span.submit input[type="submit"] {
 .dashboard > form {
   float: left;
   width: 100%;
+  clear: none;
 }
 
 .dashboard fieldset {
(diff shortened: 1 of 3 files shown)


public/stylesheets/site/2.0/07-interactions.css
```diff
@@ -682,10 +682,15 @@ form.single .autocomplete + input + span.submit input[type="submit"] {
 .dashboard > form {
   float: left;
   width: 100%;
+  clear: none;
```
sarken, line 685: This is for where we see the subnavigation hanging over the form.

public/stylesheets/site/2.0/22-system-messages.css
```diff
@@ -51,7 +51,7 @@ div.error {
     box-shadow: 1px 1px 2px;
 }
 
-.notes {
+p.notes {
```
sarken, line 54: The notes field on the posting form was playing up because of a clear meant for notes in system messages (which are [either paragraphs or unordered lists](https://github.com/search?q=repo%3Aotwcode%2Fotwarchive+notes+path%3A%2F%5Eapp%5C%2Fviews%5C%2F%2F&type=code)), but the notes elements _also_ had the notes class. You'll see `ul.notes` already has this clear above, so we just needed to make sure it also applied to paragraphs with the `notes` class.

Fixing this allowed me to remove the clear from `.blurb blockquote`, which had the notes class in the bookmark blurb. https://github.com/otwcode/otwarchive/blob/f21f823d4842860e921909f6043cec3d609271b8/app/views/bookmarks/_bookmark_user_module.html.erb#L50

brianjaustin (approved):

sarken:
Guess what's fixed in the new macOS and iOS! 🎉

sarken:
Alas, it is not fully fixed.

PR #4632, opened 2023-09-16
AO3-6328 Allow tags to be sorted by number of uses

weeklies:
# Pull Request Checklist

* [x] Have you read ["How to write the perfect pull request"](https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/)?
* [x] Have you read the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [x] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)?
* [x] Have you added the [Jira](https://otwarchive.atlassian.net) issue number
  as the *first* thing in your pull request title (e.g. `AO3-1234 Fix thing`)
* [x] Do you have fewer than 5 pull requests already open? If not, please wait
  until they are reviewed and merged before creating new pull requests.

## Issue

https://otwarchive.atlassian.net/browse/AO3-6328

## Purpose

> This pull request is to allow tags to be sorted by number of uses in the recently created tag search. To do this, we need to allow reindexing when the use counts change and to reduce redundant writes to the database when counts do not change.

## Testing Instructions

See Jira issue.

## References

Initial creation of the tag search: https://otwarchive.atlassian.net/browse/AO3-5830
Unmerged PR attempting to reduce redundant writes: https://github.com/otwcode/otwarchive/pull/2868/files

## Credit

Tara Rosenthal (she/her) - Most of the heavy lifting!
weeklies (she/her)

PR #4637, opened 2023-10-07
AO3-6328 Try to fix large tags not updating in uses sort

weeklies:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6328 (Second PR)

## Purpose

Tries to ensure that when the number of uses of a large tag increases (e.g. when a work is posted), a reindex is triggered. After some investigation, it looks like the cached tag count prevents `write_taggings_to_redis` from being called when the tag has at least 1000 uses (`TAGGINGS_COUNT_MIN_CACHE_COUNT`). 

I am not 100% sure of this approach, which is to remove the old cached count in `after_update`. Another way I believe to be possible is setting the `taggings_count_cache` range in [ScheduledTagJob](https://github.com/otwcode/otwarchive/blob/660a67aae71f2e0dbfc0316727ce7ce160ed607a/app/models/scheduled_tag_job.rb#L5) to be a lower value like `TAGGINGS_COUNT_MIN_CACHE_COUNT`. Not sure of the broader implications of these methods.

## Testing Instructions

See Jira issue.

## References

https://github.com/otwcode/otwarchive/pull/4632

## Credit

weeklies (she/her)

### DIFF
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -1133,6 +1133,11 @@ class Tag < ApplicationRecord
       end
     end
 
+    # Ensure tags with count cache are updated.
+    if tag.saved_change_to_taggings_count_cache?
+      Rails.cache.delete(taggings_count_cache_key)
+    end
+
     # Reindex immediately to update the unwrangled bin.
     if tag.saved_change_to_unwrangleable?
       tag.reindex_document
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -1133,7 +1133,7 @@ class Tag < ApplicationRecord
       end
     end
 
-    # Ensure tags with count cache are updated.
+    # Ensure tags with count cache can be reindexed.
     if tag.saved_change_to_taggings_count_cache?
       Rails.cache.delete(taggings_count_cache_key)
     end
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -1134,9 +1134,7 @@ class Tag < ApplicationRecord
     end
 
     # Ensure tags with count cache can be reindexed.
-    if tag.saved_change_to_taggings_count_cache?
-      Rails.cache.delete(taggings_count_cache_key)
-    end
+    Rails.cache.delete(taggings_count_cache_key) if tag.saved_change_to_taggings_count_cache?
 
     # Reindex immediately to update the unwrangled bin.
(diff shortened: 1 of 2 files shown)


ceithir:
https://github.com/otwcode/otwarchive/blob/660a67aae71f2e0dbfc0316727ce7ce160ed607a/app/models/tag.rb#L37-L46
Do we know what are the values of TAGGINGS_COUNT_CACHE_DIVISOR, TAGGINGS_COUNT_MIN_TIME and TAGGINGS_COUNT_MAX_TIME on staging? Just they are not set to extremes quite different from the defaults to test things, ending up with extremely high `taggings_count_expiry`.

sarken:
@ceithir I just checked and all three are `nil` on staging.

ceithir:
I don't think it's the sole source of the current issue, as it only affects tags with a count < 1000, but dropping it here to not forget:
[update_tag_cache](https://github.com/otwcode/otwarchive/blob/660a67aae71f2e0dbfc0316727ce7ce160ed607a/app/models/tag.rb#L84) tries to update the cache if the cached value is less than TAGGINGS_COUNT_MIN_CACHE_COUNT... But it seems to be doing nothing in that case, as there's an [early return](https://github.com/otwcode/otwarchive/blob/660a67aae71f2e0dbfc0316727ce7ce160ed607a/app/models/tag.rb#L76) for any non nil value of the cache, be it 1, 999 or 1001.

ceithir:
Out of the blue, I'm wondering if the call to [write_taggings_to_redis](https://github.com/otwcode/otwarchive/blob/660a67aae71f2e0dbfc0316727ce7ce160ed607a/app/models/tag.rb#L71), and the whole job that goes with it, couldn't be replaced with something straightforward like:
```rb
if value != taggings_count_cache
  # Skipping callback to avoid changing the current in memory object, and cascading side effects in general, as we're in a getter
  Tag.where(id: self.id).update_all(taggings_count_cache: value)
  self.enqueue_to_index
end
```
What lock/performance issue am I missing here?

weeklies:
@ceithir I didn't want to try to redesign the tags update code, because I worry about unintended consequences considering the complexity of the various async jobs. If AD&T decides to do so, I think it should be in a separate issue.

sarken (approved):

PR #4641, opened 2023-10-17
Revert "AO3-6328 Try to fix large tags not updating in uses sort"

sarken:
Reverts otwcode/otwarchive#4637

## Issue

https://otwarchive.atlassian.net/browse/AO3-6328

## Purpose

It turns out our problem was simply not restarting the Resque workers 🤡


ceithir:
https://tvtropes.org/pmwiki/pmwiki.php/Main/CryLaughing

PR #4643, opened 2023-10-21
AO3-6210 Use cached number of tags on mass wrangling

ceithir:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6210

## Purpose

Display the cached number of taggings on the Mass Wrangling page rather than triggering the whole logic to recompute the accurate one. This should improve performances as well as sort consistency as the [cached count is already used when sorting](https://github.com/otwcode/otwarchive/blob/53fc408f8590e585647f1b9d93a55c2a3c0f2164/app/controllers/tag_wranglings_controller.rb#L22).

Note sure this covers the whole issue, just did the quick and easy bit before the next release.

## Testing Instructions

Visible performance improvement on the Mass Wrangling page on staging/production?

## Credit

Ceithir (he/him)

### DIFF
--- a/app/views/tag_wranglings/index.html.erb
+++ b/app/views/tag_wranglings/index.html.erb
@@ -120,7 +120,7 @@
                     <% end %>
                   </td>
 
-                  <td title="<%= ts('taggings') %>"><%= tag.taggings_count %></td>
+                  <td title="<%= ts('taggings') %>"><%= tag.taggings_count_cache %></td>
 
                   <td>
                     <ul class="actions" role="navigation">
--- a/app/views/tag_wranglings/index.html.erb
+++ b/app/views/tag_wranglings/index.html.erb
@@ -120,7 +120,7 @@
                     <% end %>
                   </td>
 
-                  <td title="<%= ts('taggings') %>"><%= tag.taggings_count_cache %></td>
+                  <td title="<%= ts("taggings") %>"><%= tag.taggings_count_cache %></td>
 
                   <td>
                     <ul class="actions" role="navigation">
--- a/features/step_definitions/tag_steps.rb
+++ b/features/step_definitions/tag_steps.rb
@@ -443,3 +443,7 @@ Then(/^show me what the tag "([^"]*)" is like$/) do |tagname|
   tag = Tag.find_by(name: tagname)
   puts tag.inspect
 end
+
+Then "no tag should scheduled for count update from now on" do
+  Tag.any_instance.should_not_receive(:write_taggings_to_redis)
+end
--- a/features/tags_and_wrangling/tag_wrangling.feature
+++ b/features/tags_and_wrangling/tag_wrangling.feature
@@ -353,3 +353,13 @@ Feature: Tag wrangling
     Then I should see "Youngest"
       But I should not see "Oldest"
       And I should not see "Middle"
+
+  Scenario: No call to Redis when no action is taken
+    Given the tag wrangling setup
+      And I am logged in as a tag wrangler
+    Then no tag should scheduled for count update from now on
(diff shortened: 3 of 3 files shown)


sarken:
I don't think this is quite what the issue intended. (@tickinginstant, since you're the issue creator, please correct me if I'm wrong!)

The goal of the issue, I think, is to start getting the contents of the mass wrangling bins from Elasticsearch instead of MySQL so we can reduce the load on the database, the same way we currently get the contents of unwrangled bins from Elasticsearch ([AO3-5493](https://otwarchive.atlassian.net/browse/AO3-5493)):

https://github.com/otwcode/otwarchive/blob/53fc408f8590e585647f1b9d93a55c2a3c0f2164/app/controllers/tags_controller.rb#L294-L298

https://github.com/otwcode/otwarchive/blob/53fc408f8590e585647f1b9d93a55c2a3c0f2164/app/models/tag.rb#L1040-L1053

While switching to `taggings_count_cache` would mean we avoid ever performing counts, if I'm not mistaken, it also means we will always access the value stored in MySQL. Using `taggings_count`, we at least hit the cache instead of the database sometimes (i.e., for big tags):

https://github.com/otwcode/otwarchive/blob/53fc408f8590e585647f1b9d93a55c2a3c0f2164/app/models/tag.rb#L74-L80

And I believe the following part of the issue refers to the counts in the sidebar, which are terribly slow when there's a cache miss, not the taggings counts:

>It would also be good to modify the cached unwrangled tags counts, shown in TagWranglingsController#index, TagWranglersController#show, and UnsortedTagsController#index, so that they are loaded from ES.

https://github.com/otwcode/otwarchive/blob/53fc408f8590e585647f1b9d93a55c2a3c0f2164/app/controllers/tag_wranglers_controller.rb#L44-L52

https://github.com/otwcode/otwarchive/blob/53fc408f8590e585647f1b9d93a55c2a3c0f2164/app/controllers/tag_wranglings_controller.rb#L9-L17

https://github.com/otwcode/otwarchive/blob/53fc408f8590e585647f1b9d93a55c2a3c0f2164/app/controllers/unsorted_tags_controller.rb#L8-L16

ceithir:
> While switching to taggings_count_cache would mean we avoid ever performing counts, if I'm not mistaken, it also means we will always access the value stored in MySQL

Unless I'm missing something, in the current iteration of the code, tags should already be fully loaded from the DB by this point (they are call to .id, .canonical?, etc. just above, and I don't see any specific code to only load some fields at first glance). So accessing the value of taggings_count_cache should come at no extraneous cost, while still alleviating the current performance issue (no triggering the taggings_count for potentially each listed tag).

I'm definitely less confident on a bigger change making it in time for the next release. But will give it a try.

sarken:
I don't think it will necessarily hurt (unless there's some reason we need to update the taggings count from the bins), but it's not really addressing the count-related performance problem in AO3-6210. For example, line 45 in `tag_wranglers_controller.rb` resulted in a database query for Relationships that took a whopping 33.4 _seconds_ on Saturday night.

If we can't think of a reason we need to trigger a cache update from the bins, we can do this for a boost, but we should probably put it under a different issue. Should we also do it for tags' wrangle pages?  https://github.com/otwcode/otwarchive/blob/53fc408f8590e585647f1b9d93a55c2a3c0f2164/app/views/tags/wrangle.html.erb#L136

ceithir:
Moving the move to ES to a different PR https://github.com/otwcode/otwarchive/pull/4646

> Should we also do it for tags' wrangle pages?

Having a general issue about replacing all calls to `taggings_count` to `taggings_count_cache` when precision is not needed might be worth it?

PR #4646, opened 2023-10-23
AO3-6210 Load the tags in the mass wrangling bins from Elasticsearch

ceithir:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6210

## Purpose

Load tags from ES on the Mass Wrangling page (main table + sidebar), to hopefully increase performances


app/views/tag_wranglings/index.html.erb
```diff
@@ -104,32 +104,34 @@
...
-
-                  <td>
-                    <ul class="actions" role="navigation">
-                      <li><%= link_to ts('Edit'), {controller: :tags, action: :edit, id: tag} %></li>
-                      <li><%= link_to ts('Wrangle'), {controller: :tags, action: :wrangle, id: tag} %></li>
-                      <li><%= link_to ts('Works'), {controller: :works, action: :index, tag_id: tag} %></li>
-                    </ul>
-                  </td>
-                </tr>
+                <% if tag.unwrangled? %>
```
ceithir, line 107: A non-solution to an actual problem.

Since the tags are now loaded from ES, they aren't always up to date. Typically, after a wrangler does any operation, they're returned to the page, nothing have been reindexed yet, and so the tag they've just wrangled is still there as if they did nothing.

I'm wondering if, while computing the numbers in the sidebar in ES is definitely good (they don't need to be perfectly accurate and all, especially as they're already cached), having the main table, with operations happening all the time, not getting the freshest info from the DB is not a cure worse than the disease, gaining a bit of performance over the price of so many potential visual bugs and bug reports.

brianjaustin: Based on the story's description, I think we'd like to try at least on staging. However, I could see the scenario you describe being quite difficult... @sarken thoughts?

brianjaustin (commented):
In general looks good to me, but I'd like sarken to weigh in on the comment below

app/models/search/tag_indexer.rb
```diff
@@ -67,7 +68,8 @@ def document(object)
     ).merge(
       has_posted_works: object.has_posted_works?,
       tag_type: object.type,
-      uses: object.taggings_count_cache
+      uses: object.taggings_count_cache,
+      to_wrangle: object.wrangleable? && object.used? && object.unwrangled?
```
sarken, line 72: I'm not sure defining a `to_wrangle` field whose value is computed partially from information we already have indexed (`uses`, `unwrangleable`, `canonical`) is the best way to handle this. 

Why not add an `unwrangled` (or `wrangled` if it's too easy to confuse with `unwrangleable`) field so we build a query using the three fields to get the results we want? That way if, e.g., we only want to include tags with greater than or equal to 3 uses in the mass bins, we'd only need to update the query, not reindex everything to update its `to_wrangle` field.

app/views/tag_wranglings/index.html.erb
sarken: So fwiw, this basically happens already -- e.g., I just wrangled two tags to my fandom, and they were still in my fandom's unwrangled bin (`/tags/Fandom%20Name/wrangle?show=freeforms&status=unwrangled`) when the page reloaded. Wranglers are pretty used to changes not being immediate, so pulling the contents of the bins from Elasticsearch should be fine.

features/tags_and_wrangling/tag_wrangling.feature
```diff
@@ -335,7 +335,8 @@ Feature: Tag wrangling
       And I should see "Grandparent" within "#parent_MetaTag_associations_to_remove_checkboxes"
       But I should not see "Parent" within "#parent_MetaTag_associations_to_remove_checkboxes"
 
-    When I view the tag "Child"
+    When I clear Rails cache
```
ceithir, line 338: Attempt at fixing the "Grandparent" test passing on a coin flip. It appears to work on my machine (famous last words of many developers).

My theory is that the root cause of this is [this cache](https://github.com/otwcode/otwarchive/blob/f2de69b0f3de9b3ab3c01ece61b85ebc2c0c961c/app/views/tags/show.html.erb#L73) being stuck somehow. I didn't exactly choose the most subtle way to bust it however.

sarken: Would `When it is currently 25 hours from now` work instead? I tried it and didn't get a failure after a whole bunch of runs in GitHub Actions.

ceithir: Seems like a clever way to burst that cache, and also seems to repeatedly work on my machine. Sold.

weeklies (commented):
Just out of curiosity, how are `unwrangled` and `unwrangleable` different?

ceithir:
> Just out of curiosity, how are `unwrangled` and `unwrangleable` different?

If I got it correctly:
- `unwrangleable` is a boolean on the tag object. When true, it means the tag should be excluded from the wrangling process altogether. I think this is the fate that awaits a lot of [freeform tags](https://archiveofourown.org/wrangling_guidelines/11) like, for example, "idk how to explain it but trust me".
- `unwrangled` actually means different things depending on what kind of tag we are talking about (cf the `def unwrangled?` in this very PR), though it can be approximate to "orphan" i.e. not linked to anything else (fandom, character...) at the moment.

spec/models/search/tag_query_spec.rb
```diff
@@ -202,4 +202,43 @@
...
+        used_and_canonical: create(:canonical_character, taggings_count_cache: 5),
+        unwrangleable: create(:character, taggings_count_cache: 5, unwrangleable: true),
+        wrangled: create(:canonical_character, common_taggings: [create(:common_tagging)])
+      }
+      run_all_indexing_jobs
+      tags
+    end
+
+    it "returns unwrangled tags in use" do
+      TagQuery.new(to_wrangle: true).search_results.should include(tags[:used])
```
brianjaustin, line 221: Based on RSpec deprecation warnings and https://www.betterspecs.org/#expect, could you rewrite these assertions to use `expect` instead?
```suggestion
      expect(TagQuery.new(to_wrangle: true).search_results).to include(tags[:used])
```

spec/controllers/tags_controller_spec.rb
```diff
@@ -533,5 +533,18 @@
         end
       end
     end
+
+    it "can add and remove metatags at the same time" do
+      old_metatag =  create(:canonical_freeform)
+      new_metatag =  create(:canonical_freeform)
```
brianjaustin, line 539: Extra space here?
```suggestion
      old_metatag = create(:canonical_freeform)
      new_metatag = create(:canonical_freeform)
```

spec/controllers/tags_controller_spec.rb
```diff
@@ -533,5 +533,18 @@
...
+
+      tag = create(:canonical_freeform)
+      MetaTagging.create(meta_tag: old_metatag, sub_tag: tag, direct: true)
```
brianjaustin, line 542: Could this be replaced with the following?
```suggestion
      create(:meta_tagging, meta_tag: old_metatag, sub_tag: tag, direct: true)
``

File: app/models/search/tag_query.rb [hunk 7]
```diff
@@ -109,7 +109,13 @@ def suggested_character_filter
...
 
   def to_wrangle_filter
-    term_filter(:to_wrangle, bool_value(options[:to_wrangle])) unless options[:to_wrangle].nil?
+    return [] unless options[:to_wrangle]
+
+    [
+      { bool: { should: [{ range: { uses: { gt: 0 } } }, term_filter(:canonical, true)] } }, # Check if used OR canonical
+      term_filter(:unwrangleable, false),
+      term_filter(:unwrangled, true)
+    ]
```
brianjaustin, line 118: Wouldn't this still need a `unless options[:to_wrangle].nil?` at the end?

app/models/search/tag_query.rb
```diff
@@ -23,7 +23,7 @@ def filters
       character_filter,
       suggested_fandom_filter,
       suggested_character_filter,
-      to_wrangle_filter
+      *to_wrangle_filter
```
brianjaustin, line 26: Is the splat needed here, since we already `flatten` the list (and flatten is recursive)?

brianjaustin (changes requested):

app/models/search/tag_query.rb
ceithir: Isn't this case covered in the early return just above?

I note that past me didn't covered the case where someone wants to explicitly request tags which are _not_ to wrangle (the filter is similarly skipped on `nil` or `false`), probably both because there is no use case for it as of now and the request sounded complicated. Should I add it nonetheless to avoid some future contributor being surprised by the behavior? Maybe out of scope and just leave an explicit comment instead?

app/models/search/tag_query.rb
ceithir: > flatten is recursive

TIL
(my mind likely assumed it wasn't as a default, like [flat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat) in JS)

app/models/search/tag_query.rb
brianjaustin: Ah yes, I was looking at a version of the diff that hid that return. I'd say we should add the `.nil?` since it's not a large change and avoids the confusion with Ruby's falsiness

brianjaustin (approved):

app/models/fandom.rb
```diff
@@ -14,9 +14,8 @@ class Fandom < Tag
 
   scope :by_media, lambda {|media| where(media_id: media.id)}
 
-  def self.unwrangled
-    joins(:common_taggings).
-    where("unwrangleable = 0 AND common_taggings.filterable_id = ? AND common_taggings.filterable_type = 'Tag'", Media.uncategorized.try(:id))
+  def unwrangled?
+    Media.uncategorized.fandoms.include?(self)
```
redsummernight, line 18: Would `include?` eager-load all the fandoms under uncategorized? Should we use [`exists?`](https://api.rubyonrails.org/v6.1/classes/ActiveRecord/Associations/ClassMethods.html#method-i-has_many) instead?

app/models/tag.rb
```diff
@@ -1034,6 +1031,10 @@ def merger_string=(tag_string)
     end
   end
 
+  def unwrangled?
```
redsummernight, line 1034: Your comment on the difference between unwrangled and unwrangleable at https://github.com/otwcode/otwarchive/pull/4646#issuecomment-1840664875 is really good, you should include that on top of this method as well.

app/models/search/tag_query.rb
```diff
@@ -103,6 +108,19 @@ def suggested_character_filter
     terms_filter(:pre_character_ids, options[:pre_character_ids]) if options[:pre_character_ids]
   end
 
+  def to_wrangle_filter
```
redsummernight, line 111: I'd split this filter and generally try to keep filters mostly one-to-one with the columns we indexed. I think that makes queries easier to read.

We already have `unwrangleable_filter`, so you can add an `in_use_filter` for "used OR canonical", and a simple `unwrangled_filter` that's just a boolean check matching the indexed column. Then you can use all 3 when setting up the query in the controller.

Simpler filters means you can avoid hard `raise` errors for unexpected combinations as well.

app/views/tag_wranglings/index.html.erb
```diff
@@ -120,7 +120,7 @@
                     <% end %>
                   </td>
 
-                  <td title="<%= ts('taggings') %>"><%= tag.taggings_count %></td>
+                  <td title="<%= ts('taggings') %>"><%= tag.taggings_count_cache %></td>
```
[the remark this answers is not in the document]
redsummernight: Can you also change the one at https://github.com/otwcode/otwarchive/blob/cb58df0bc28bcddc6b07f485d9d981b043ae1397/app/views/tags/wrangle.html.erb#L136 too?

features/tags_and_wrangling/tag_wrangling.feature
```diff
@@ -337,7 +337,8 @@ Feature: Tag wrangling
       And I should see "Grandparent" within "#parent_MetaTag_associations_to_remove_checkboxes"
       But I should not see "Parent" within "#parent_MetaTag_associations_to_remove_checkboxes"
 
-    When I view the tag "Child"
+    When it is currently 25 hours from now
```
redsummernight, line 340: Can you add a comment that this is to expire the meta tags tree cached for 24 hours?

features/step_definitions/tag_steps.rb
```diff
@@ -443,3 +444,7 @@
   tag = Tag.find_by(name: tagname)
   puts tag.inspect
 end
+
+Then "no tag is scheduled for count update from now on" do
+  Tag.any_instance.should_not_receive(:write_taggings_to_redis)
```
redsummernight, line 449: You can use the expect syntax here, like: https://github.com/otwcode/otwarchive/blob/cb58df0bc28bcddc6b07f485d9d981b043ae1397/spec/controllers/troubleshooting_controller_spec.rb#L123

app/models/fandom.rb
ceithir: Rails seems to be smartly converting the `include?` to `exists?` on its own there:
```
irb(main):003:0> Media.uncategorized.fandoms.include?(test)
  Tag Load (0.4ms)  SELECT `tags`.* FROM `tags` WHERE (tags.name = 'Uncategorized Fandoms') ORDER BY `tags`.`id` ASC LIMIT 1
Query Trace:
      app/models/tag.rb:575:in `find_by_name'
      app/models/tag.rb:582:in `find_or_create_by_name'
      app/models/media.rb:13:in `uncategorized'
  Tag Exists? (0.2ms)  SELECT 1 AS one FROM `tags` INNER JOIN `common_taggings` ON `tags`.`id` = `common_taggings`.`common_tag_id` WHERE `common_taggings`.`filterable_id` = 1063753922 AND `common_taggings`.`filterable_type` = 'Tag' AND `tags`.`type` = 'Fandom' AND `tags`.`id` = 130714128 LIMIT 1
```

features/tags_and_wrangling/tag_wrangling.feature
ceithir: Change actually not needed anymore, as already fixed in https://github.com/otwcode/otwarchive/pull/4658

spec/models/search/tag_query_spec.rb
```diff
@@ -194,12 +194,62 @@
...
     end
 
     it "allows you to sort by Uses in ascending order" do
       q = TagQuery.new(sort_column: "uses", sort_direction: "asc")
-      expect(q.generated_query[:sort]).to eq([{ "uses" => { order: "asc" } }, { id: { order: "asc" } }])
+      expect(q.generated_query[:sort]).to eq([{ "uses" => { order: "asc" } }, { "name.keyword" => { order: "asc" } }, { id: { order: "asc" } }])
+    end
+  end
+
+  describe "to_wrangle" do
```
redsummernight, line 206: This test should have the tag `tag_search: true` on it, to reset the search index.

https://github.com/otwcode/otwarchive/blob/e211c6bb9ad792e344aacb629020c3bc92576b0a/spec/spec_helper.rb#L106-L112

spec/models/search/tag_query_spec.rb
```diff
@@ -194,12 +194,62 @@
...
+
+    it "does not return wrangled tags" do
+      expect(results).not_to include(tags[:wrangled])
+    end
+
+    it "can also list unused tags" do
+      results = TagQuery.new(in_use: false).search_results
+      expect(results).to include(tags[:unused_and_not_canonical])
+      expect(results).not_to include(tags[:used])
+      expect(results).not_to include(tags[:unused_but_canonical])
```
redsummernight, line 248: Can you also check that `:used_and_canonical` does not show up in results here?

redsummernight (approved):
Looks good, thank you! Just some minor comments on the tests.

redsummernight (approved):

PR #4652, opened 2023-11-04
AO3-6058 Add 'translation into' to download preface

brianjaustin:
## Issue

## Purpose

Add the translation(s) of a work to the links displayed at the beginning of a download.

## Testing Instructions

Refer to Jira story

## References

Follow-up to #4608 

## Credit

@weeklies for the majority of the work in the linked PR
Brian Austin (they/he)

### DIFF
--- a/app/views/downloads/_download_preface.html.erb
+++ b/app/views/downloads/_download_preface.html.erb
@@ -65,15 +65,24 @@
       <% end %>
     <% end %>
 
+    <%# i18n-tasks-use t("downloads.download_preface.translated_to.restricted_html") %>
+    <%# i18n-tasks-use t("downloads.download_preface.translated_to.revealed_html") %>
+    <%# i18n-tasks-use t("downloads.download_preface.translated_to.unrevealed_html") %>
     <%# i18n-tasks-use t("downloads.download_preface.inspired_by.restricted_html") %>
     <%# i18n-tasks-use t("downloads.download_preface.inspired_by.revealed_html") %>
     <%# i18n-tasks-use t("downloads.download_preface.inspired_by.unrevealed") %>
     <%# i18n-tasks-use t("downloads.download_preface.translation_of.restricted_html") %>
     <%# i18n-tasks-use t("downloads.download_preface.translation_of.revealed_html") %>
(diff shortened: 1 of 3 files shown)


config/locales/views/en.yml
```diff
@@ -410,6 +410,10 @@ en:
         restricted_html: Inspired by [Restricted Work] by %{creator_link}
         revealed_html: Inspired by %{work_link} by %{creator_link}
         unrevealed: Inspired by a work in an unrevealed collection
+      translated_to:
+        restricted_html: 'Translation into %{language} available: [Restricted Work] by %{creator_link}'
+        revealed_html: 'Translation into %{language} available: %{work_link} by %{creator_link}'
+        unrevealed_html: 'Translation into %{language} available: A work in an unrevealed collection.'
```
brianjaustin, line 416: The spreadsheet of new copy has a full-stop here, but none of the others do. Is that intentional, or should we remove it for consistency?

sarken: Remove, please!

app/views/downloads/_download_preface.html.erb
```diff
@@ -65,15 +65,24 @@
       <% end %>
     <% end %>
 
+    <%# i18n-tasks-use t("downloads.download_preface.translated_to.restricted_html") %>
+    <%# i18n-tasks-use t("downloads.download_preface.translated_to.revealed_html") %>
+    <%# i18n-tasks-use t("downloads.download_preface.translated_to.unrevealed_html") %>
```
ceithir, line 70: Any particular reason these hints are within the template instead of [in the helper](https://github.com/otwcode/otwarchive/blob/d87652399df73fa5e3b829b0089529501a381538/app/helpers/works_helper.rb#L115) where they're used?

brianjaustin: I had to think about it, but I think the reasoning is that the calls here "cause" the translation key to be used (i.e. without calling the helper here, the translation keys would be unused), and in contrast, the helper just "facilitates" the use

weeklies (approved):
Looks good to me. Thanks for working on this 🙂

PR #4659, opened 2023-11-12
AO3-3471 Filter works per language on language_works_path

ceithir:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-3471

## Purpose

Filter per language using ElasticSearch; same general behavior as filtering using a tag

## Testing Instructions

Go to `/languages/en/works`, `/languages/de/works`, `/languages/fr/works`, etc.
Check the works are in the s 

Note: These routes are currently disabled on staging and production, and I'm not exactly sure how. nginx configuration?

## Credit

Ceithir (he/him)


features/step_definitions/work_steps.rb
```diff
@@ -564,6 +564,14 @@
   step "the periodic tag count task is run"
 end
 
+When "I browse works in {string}" do |language_name|
```
brianjaustin, line 567: How about
```suggestion
When "I browse works in language {string}" do |language_name|
```
to make it 100% clear that it's a language, and not say a fandom or other tag?

brianjaustin:
Summarising conversation from elsewhere: we do hardcode 404 the path for efficiency reasons. In particular https://github.com/otwcode/otwarchive/blob/bba3552e90254ea003cd8e67191a0c207fc1ea40/app/controllers/languages_controller.rb#L9 is not super great. For now, let's have `show` redirect to the `/language/:short/works` path (we'll need to check admin edits work as normal though).

weeklies:
@brianjaustin Is this ready to merge or action is still needed?

ceithir:
> we'll need to check admin edits work as normal though

Updated it to return to the index /languages, since that was already what the create route was doing.

brianjaustin:
@weeklies you're right, I put the wrong label. Thanks for the poke! I'll try to do another pass this evening (EST) but for now I changed to `Coder Has Actioned Review`

config/routes.rb
```diff
@@ -509,10 +509,11 @@
   #### I18N ####
 
   # should stay below the main works mapping
-  resources :languages do
+  resources :languages, except: [:show] do
     resources :works
     resources :admin_posts
   end
+  get "/languages/:id", to: redirect("/languages/%{id}/works", status: 302)
```
brianjaustin, line 516: I think it would be more obvious that this route belongs to `languages` if we use the `member` block like
```suggestion
  resources :languages, except: [:show] do
    resources :works
    resources :admin_posts

    member do
      get :show, to: redirect("/languages/%{id}/works", status: 302)
    end
  end
```

brianjaustin (changes requested):

config/routes.rb
ceithir: Doesn't appear to work. Or more exactly, it works too well, changing all get routes on /languages to a redirect.

I would say the `get "/languages/:id", to:`is good enough as is? It's consistent with [official documentation(https://guides.rubyonrails.org/routing.html#redirection).

brianjaustin (approved):

PR #4664, opened 2023-11-17
AO3-3866 update revised_at when a work is revealed

eliahhecht:
## Issue

https://otwarchive.atlassian.net/browse/AO3-3866 (Please fill in issue number and remove this comment.)

## Purpose

When a work changes from unrevealed to revealed, update its revised_at date, so that it sorts according to when it was revealed, rather than when it was originally posted.

## Testing Instructions

Testing instructions in the issue should confirm the fix.

## Credit

Eliah Hecht, he/him

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -197,6 +197,11 @@ class Work < ApplicationRecord
 
   before_save :clean_and_validate_title, :validate_published_at, :ensure_revised_at
 
+  # If the work is becoming revealed, bump the revised-at date (AO3-3866)
+  before_save do
+    set_revised_at(Date.today) if will_save_change_to_in_unrevealed_collection?
+  end
+
   after_save :post_first_chapter
   before_save :set_word_count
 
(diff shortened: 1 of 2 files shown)


app/models/work.rb
```diff
@@ -197,6 +197,11 @@ def new_recipients_allow_gifts
 
   before_save :clean_and_validate_title, :validate_published_at, :ensure_revised_at
 
+  # If the work is becoming revealed, bump the revised-at date (AO3-3866)
+  before_save do
+    set_revised_at(Time.zone.today) if will_save_change_to_in_unrevealed_collection?
```
weeklies, line 202: To be precise, should we use `Time.current`?

EchoEkhi:
I thought the current plan is to [re-do the whole thing](https://otwarchive.atlassian.net/browse/AO3-4546) instead?

app/models/work.rb
eliahhecht: Good call, done. Somehow I got it into my head that these fields were date instead of datetime, but I see that is not the case.

brianjaustin:
@eliahhecht I think we'd prefer to collapse the timestamp fields as described in the [epic @EchoEkhi linked](https://otwarchive.atlassian.net/browse/AO3-4546). I'm going to close this PR for now so it doesn't take 1 of your 5 slots, but feel free to continue discussing here!

PR #4669, opened 2023-11-21
AO3-6639 Output current DB schema

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6639

## Purpose

* Output the latest DB schema (based on a Docker setup)
  * This should make new environment bootstrapping a bit quicker
  * Since Rails upgrades change schema formats, we will have a record of things now _before_ Rails 7
* Change the schema output process to update `schema.rb` when a new migration runs; this will make it so PRs with migrations update by default, saving us work later (we update the structure SQL separately, and that will stay manual as MySQL formats may change to not match real prod)

## Credit

Brian Austin (they/he)

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -67,7 +67,7 @@ module Otwarchive
     config.filter_parameters += [:content, :password, :terms_of_service_non_production]
 
     # Disable dumping schemas after migrations.
-    config.active_record.dump_schema_after_migration = false
+    config.active_record.dump_schema_after_migration = true
 
     # Allows belongs_to associations to be optional
     config.active_record.belongs_to_required_by_default = false
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -2,17 +2,17 @@
 # of editing this file, please use the migrations feature of Active Record to
 # incrementally modify your database, and then regenerate this schema definition.
 #
-# Note that this schema.rb definition is the authoritative source for your
-# database schema. If you need to create the application database on another
-# system, you should be using db:schema:load, not running all the migrations
-# from scratch. The latter is a flawed and unsustainable approach (the more migrations
-# you'll amass, the slower it'll run and the greater likelihood for issues).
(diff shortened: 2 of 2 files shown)


brianjaustin:
(This isn't quite tests, but that's the closest label.)

config/application.rb
```diff
@@ -67,7 +67,7 @@ class Application < Rails::Application
     config.filter_parameters += [:content, :password, :terms_of_service_non_production]
 
     # Disable dumping schemas after migrations.
```
weeklies, line 69: I think this comment should be updated.

Bilka2:
Things I noticed:
Some tables lost the ENGINE=InnoDB option and new tables don't have it. MariaDB defaults to innodb so that seems fine.
The limit option on text columns was converted to size. size is specific to the mysql adapter/schema (and doesnt show up in the docs). Since that's the adapter that's getting used, that also seems fine.

Based on https://github.com/otwcode/otwarchive/wiki/Migrations#why-not-dumping-from-development-environments it may be better to run this in production, not in a local docker setup.

config/application.rb
brianjaustin: Thanks for catching that! It looks like `true` is the default, so I've just removed both of these lines

Bilka2:
Agreed to the reasoning for keeping the Docker schema, good to see that there are no major differences.

Regarding the config, [the docs](https://github.com/rails/rails/blob/main/activerecord/lib/active_record.rb#L382-L383) suggest setting dump_schema_after_migration to false in production (and likely also staging).

Bilka2 (approved):

brianjaustin:
(I've added the SQL dump too per conversation with @Bilka2 elsewhere; I'd recommend skimming it at most since the diff is a bit larger.)

Bilka2:
I looked over the SQL dump and it looks consistent with the rb dump and good in general :+1:

PR #4475, opened 2023-03-07
AO3-6492 Update text for abuse report page and confirmation email

forceofcalm:
# Pull Request Checklist

* [x] Have you read ["How to write the perfect pull request"](https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/)?
* [x] Have you read the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [ ] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)?
* [x] Have you added the [Jira](https://otwarchive.atlassian.net/) issue number
  as the first thing in your pull request title (e.g. AO3-1234 Fix thing)
* [x] Do you have fewer than 5 pull requests already open? If not, please wait
  until they are reviewed and merged before creating new pull requests.

## Issue

This PR is for the [Jira ticket AO3-6492:](https://otwarchive.atlassian.net/browse/AO3-6492) Improvement - Update text for abuse report page and confirmation email.

## Purpose
This pull request features text updates to the /abuse_report/new page, which is where users can submit a report of actions that may violate the terms of service, harassment, DMCA claims, and so forth.

This update updates the description and success message text of that form, as well as adds clarification text to the summary field, and moves the link field higher in the form.

The HTML and text emails for confirmation of a successful form submission for this page have also been updated with new text.

## Testing Instructions

This is my first PR to this project, as well as my first experience with Ruby/Ruby on Rails. I can check the text updates locally, but I wasn't sure how to test the confirmation emails. Any guidance on how to do so would be appreciated.

Similarly, if any code of mine is not up to the standards, please point it out. RoR looks very foreign to me, but I tried my best to follow the Coding Standards and mimic the styles I noted in other files. (And apologies for the double PR -- I fudged my fork).

I have a Jira account (forceofcalm@gmail.com) but do not have permissions to update the ticket status, nor to comment on tickets.

## Credit

calm they/them

PR #4680, opened 2023-12-04
AO3-6386 Fix that email addresses in mailer previews were localized

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6386

## Purpose

This changes the generator for the emails of users and admins in tests and mailer previews to always produce valid email addresses.

This also fixes that the mailer previews [errored in local dev](https://github.com/otwcode/otwarchive/pull/4475#pullrequestreview-1329628736). The comment in the code is based on [the docs](https://thoughtbot.github.io/factory_bot/ref/find_definitions.html).

## Testing Instructions

The preview UI should be at https://test.archiveofourown.org/rails/mailers and the creatorship mailer previews should not error when selecting languages such as `ja`, `vi` or `zh-CN`.

## Credit

Bilka (he/him)

### DIFF
--- a/test/mailers/previews/application_mailer_preview.rb
+++ b/test/mailers/previews/application_mailer_preview.rb
@@ -1,6 +1,7 @@
 require "factory_bot"
 
-FactoryBot.find_definitions
+# In test and dev this is automatically called by factory_bot_rails after initialize, don't call it again
+FactoryBot.find_definitions unless Rails.env.test? || Rails.env.development?
 
 class ApplicationMailerPreview < ActionMailer::Preview
   include FactoryBot::Syntax::Methods
--- a/factories/admin.rb
+++ b/factories/admin.rb
@@ -5,7 +5,7 @@ FactoryBot.define do
     login { generate(:login) }
     password { "adminpassword" }
     password_confirmation { |u| u.password }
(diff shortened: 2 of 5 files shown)


test/mailers/previews/application_mailer_preview.rb
```diff
@@ -1,6 +1,7 @@
 require "factory_bot"
 
-FactoryBot.find_definitions
```
ceithir, line 3: What are the use cases where this will be called outside of dev/test? In other words, would it possible to just remove the line entirely?

ceithir: `if Rails.env.staging?`might be more explicit then?

Bilka2: I went for test and dev because it depends on the factory_bot_rails gem being loaded. And the gem is in the test and dev group: https://github.com/otwcode/otwarchive/blob/4c2621eedc1ddaf07d11bf3b5f9519e1575fb917/Gemfile#L143-L148

`if Rails.env.staging?` would indeed have the same effect. I have no strong opinions towards changing it.

ceithir (approved):

PR #4684, opened 2023-12-06
AO3-6650 Move speculation rules next chapter prefetch

domenic:
Previously, we were adding the `<script type=speculationrules>` block inside the `_chapter` include. However, this means it was included on the "Entire Work" page, which is undesirable. (On that page, there would be no `@next_chapter` variable, so nothing would be rendered... and then that lack of speculation rules would be cached.)

We only want to preload the next chapter when we're looking at a single chapter in isolation. For this, it's best to put the `<script type=speculationrules>` in the `chapters/show` template.

# Pull Request Checklist

* [ ] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)?: per discussion [on the ticket](https://otwarchive.atlassian.net/browse/AO3-6636), this is probably fine without tests.

## Issue

https://otwarchive.atlassian.net/browse/AO3-6650

## Purpose

To ensure speculation rules are included on all single-chapter pages with a next chapter, and not included on the Entire Work page, and cached correctly while we do it.

## Testing Instructions

See https://otwarchive.atlassian.net/browse/AO3-6650 for how to reproduce the bug; the idea is that it should no longer reproduce, after this change.

## References

- The parent issues https://otwarchive.atlassian.net/browse/AO3-6635 and https://otwarchive.atlassian.net/browse/AO3-6637 are also relevant.
- All credit to @sarken for understanding why this was broken.

## Credit

Domenic Denicola, he/him

### DIFF
--- a/app/views/chapters/_chapter.html.erb
+++ b/app/views/chapters/_chapter.html.erb
@@ -17,6 +17,7 @@
     <%= render "chapters/chapter_management", chapter: chapter, work: @work %>
   <% end %>
 
+<%# Both Entire Work and Chapter by Chapter mode write to and read from this cache. Only content needed for both modes should be included here. %>
 <% cache_if(!@preview_mode, "#{@work.cache_key}/#{chapter.cache_key}-show-v7", skip_digest: true) do %>
 
   <div class="chapter<%= ' preface' unless @preview_mode %> group">
--- a/app/views/chapters/_chapter.html.erb
+++ b/app/views/chapters/_chapter.html.erb
@@ -17,7 +17,7 @@
     <%= render "chapters/chapter_management", chapter: chapter, work: @work %>
   <% end %>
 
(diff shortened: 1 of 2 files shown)


app/views/chapters/_chapter.html.erb
```diff
@@ -17,7 +17,7 @@
     <%= render "chapters/chapter_management", chapter: chapter, work: @work %>
   <% end %>
 
-<% cache_if(!@preview_mode, "#{@work.cache_key}/#{chapter.cache_key}-show-v6", skip_digest: true) do %>
+<% cache_if(!@preview_mode, "#{@work.cache_key}/#{chapter.cache_key}-show-v7", skip_digest: true) do %>
```
sarken, line 20: Would you mind adding a code comment above this for future!us? Just like
```
<%# Both Entire Work and Chapter by Chapter mode write to and read from this cache. Only content needed for both modes should be included here. %>
``` 

That way we might not get bitten by this in the future. 🤞

sarken (approved):
🎉  Thank you!

PR #4687, opened 2023-12-10
AO3-6626 Add after task for removing invalid commas from tags

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6626

## Purpose

Spoke to wrangling chairs and what we’re going to do is

- remove the commas
- if the tag can’t be saved because the comma-less version already exists, remove the comma and append “ - AO3-6626” to the end of the tag name so it can be found and dealt with by a wrangling admin

I’ve written a task that will do that:

`bundle exec rake After:remove_invalid_commas_from_tags`

Because tags can only be renamed by admins, the task will prompt you for your admin login. This is the whole thing, e.g., `admin-sarken`. 

If the login you enter cannot be found, it will tell you and the task won’t run.

If the login is found, the task will run and give you the new tag names.

## Testing Instructions

Run the task, first trying an invalid admin login, and then trying a valid one.

### DIFF
--- a/lib/tasks/after_tasks.rake
+++ b/lib/tasks/after_tasks.rake
@@ -269,7 +269,7 @@ namespace :After do
 
   desc "Remove full-width and ideographic commas from tags"
   task(remove_invalid_commas_from_tags: :environment) do
-    puts("Tags can only be renamed by an admin. Enter your admin login:")
+    puts("Tags can only be renamed by an admin, who will be listed as the tag's last wrangler. Enter the admin login we should use:")
     login = $stdin.gets.chomp.strip
     admin = Admin.find_by(login: login)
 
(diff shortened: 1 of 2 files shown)


lib/tasks/after_tasks.rake
```diff
@@ -267,5 +267,30 @@ namespace :After do
     r&.destroy
   end
 
+  desc "Remove full-width and ideographic commas from tags"
+  task(remove_invalid_commas_from_tags: :environment) do
+    puts("Tags can only be renamed by an admin. Enter your admin login:")
+    login = $stdin.gets.chomp.strip
+    admin = Admin.find_by(login: login)
```
brianjaustin, line 274: Would it be worthwhile to check the password too, with `admin.valid_password?(...)`?

sarken: tbh I'm not really doing it to make sure that the admin is authorized, just to avoid the validation failing (I was actually thinking about using a wrangling chair's admin account to avoid confusion about why a non-wrangling admin was renaming tags)

brianjaustin (approved):

zz9pzza (approved):

PR #4689, opened 2023-12-12
AO3-6609 Fix status check of old invites without sent_at

weeklies:
Third PR to fix issues raised in QA. PR may need further tweaks. Currently does set sent_at and relies on updated_at as fallback. 

## Issue

https://otwarchive.atlassian.net/browse/AO3-6609

## Purpose

Support resending and displaying status info of old invites without sent_at 

## Testing Instructions

Try to check and resend an old invite

## References

https://github.com/otwcode/otwarchive/pull/4685

## Credit

weeklies (she/her)

### DIFF
--- a/app/models/invitation.rb
+++ b/app/models/invitation.rb
@@ -67,16 +67,21 @@ class Invitation < ApplicationRecord
       UserMailer.invitation(self.id).deliver_now
     end
 
-    date_column = resend ? :resent_at : :sent_at
     # Skip callbacks within after_save by using update_column to avoid a callback loop
-    self.update_column(date_column, Time.current)
+    if resend
+      self.update_column(:resent_at, Time.current)
+      # This applies to old invites when AO3-6094 wasn't fixed.
+      self.update_column(:sent_at, self.updated_at) if self.sent_at.nil?
+    else
+      self.update_column(:sent_at, Time.current)
+    end
   rescue StandardError => e
     errors.add(:base, :notification_could_not_be_sent, error: e.message)
   end
 
   def can_resend?
-    # created_at fallback is a vestige of the already fixed AO3-6094.
-    checked_date = self.resent_at || self.sent_at || self.created_at
+    # updated_at fallback is a vestige of the already fixed AO3-6094.
+    checked_date = self.resent_at || self.sent_at || self.updated_at
     checked_date < ArchiveConfig.HOURS_BEFORE_RESEND_INVITATION.hours.ago
   end
 
--- a/app/models/invitation.rb
+++ b/app/models/invitation.rb
@@ -75,7 +75,8 @@ class Invitation < ApplicationRecord
   end
 
   def can_resend?
-    checked_date = self.resent_at || self.sent_at
+    # created_at fallback is a vestige of the already fixed AO3-6094.
+    checked_date = self.resent_at || self.sent_at || self.created_at
     checked_date < ArchiveConfig.HOURS_BEFORE_RESEND_INVITATION.hours.ago
   end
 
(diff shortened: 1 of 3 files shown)


spec/models/invitation_spec.rb
```diff
@@ -75,4 +75,41 @@
       end
     end
   end
+
+  describe "can_resend?" do
+    # Support old invites when AO3-6094 wasn't fixed.
+    context "old invites without sent_at" do
```
sarken, line 81: [Context](https://www.betterspecs.org/#contexts) should start with `when` or `with/out`, so let's go with `without sent_at` here and `with sent_at` below. (I was going to say `without sent_at due to AO3-6094`, but there might be other reasons for a nil `sent_at`)

app/models/invitation.rb
```diff
@@ -67,15 +67,21 @@ def send_and_set_date(resend: false)
       UserMailer.invitation(self.id).deliver_now
     end
 
-    date_column = resend ? :resent_at : :sent_at
     # Skip callbacks within after_save by using update_column to avoid a callback loop
-    self.update_column(date_column, Time.current)
+    if resend
+      self.update_column(:resent_at, Time.current)
+      # This applies to old invites when AO3-6094 wasn't fixed.
+      self.update_column(:sent_at, self.updated_at) if self.sent_at.nil?
```
sarken, line 74: Does anyone have strong feelings about calling `update_column` twice? I'm not sure how I feel about it.

sarken (commented):
Couple quick comments before I go afk -- thanks for working on this and sorry it's turned into a monster!

app/models/invitation.rb
brianjaustin: I think I'm OK with it, the performance difference doesn't seem too bad. But if we want to avoid it, something like
```rb
attrs = { resent_at: Time.current }
attrs[:sent_at] = self.updated_at if self.sent_at.nil?
self.update_columns(attrs)
```
should work

spec/models/invitation_spec.rb
```diff
@@ -75,4 +75,41 @@
       end
     end
   end
+
+  describe "can_resend?" do
```
brianjaustin, line 79: Per https://www.betterspecs.org/ (under "Describe Your Methods"), the recommended naming for methods starts with `#`:
```suggestion
  describe "#can_resend?" do
```

app/models/invitation.rb
```diff
@@ -67,15 +67,22 @@ def send_and_set_date(resend: false)
...
     end
 
-    date_column = resend ? :resent_at : :sent_at
     # Skip callbacks within after_save by using update_column to avoid a callback loop
-    self.update_column(date_column, Time.current)
+    if resend
+      attrs = { resent_at: Time.current }
+      # This applies to old invites when AO3-6094 wasn't fixed.
+      attrs[:sent_at] = self.updated_at if self.sent_at.nil?
+      self.update_columns(attrs)
```
sarken, line 75: I'm checking with PAC and Support chairs to make sure it's okay if we add a sent_at date. If it is, we probably want a test to make sure it's set correctly.

Also, it just occurred to me -- is this going to be the updated_at date that's on the invitation before we resend it? Or the updated_at date that's on the invitation _after_ we resend it (which would basically match the resent_at date, I believe)? I feel like the previous commit did the latter and this one does the former, but I'm not 100% sure.

weeklies: Using `update_column` doesn't update `updated_at` so I don't think so.

weeklies:
I'm going to be busy until next week, so if someone wants to take over, please do.

app/models/invitation.rb
sarken: Oh, you're right, I didn't notice it was update_column! Now I need to totally revisit what I told everyone. 🤦 Thanks for pointing that out.

sarken (approved):

PR #4690, opened 2023-12-12
AO3-6651 Update config to Rails 6.1

brianjaustin:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6651

## Purpose

Update the config defaults to use Rails 6.1 instead of 6.0, so that we're not overriding in 2 places when the Rails 7 upgrade happens.

## Credit

Brian Austin (they/he)


config/environments/production.rb
```diff
@@ -1,59 +1,135 @@
...
+
+  # Enable serving of images, stylesheets, and JavaScripts from an asset server.
+  # config.asset_host = 'http://assets.example.com'
+
+  # Specifies the header that your server uses for sending files.
+  # config.action_dispatch.x_sendfile_header = 'X-Sendfile' # for Apache
+  # config.action_dispatch.x_sendfile_header = 'X-Accel-Redirect' # for NGINX
+
+  # Store uploaded files on the local file system (see config/storage.yml for options).
+  config.active_storage.service = :local
```
brianjaustin, line 46: This won't really do anything at the moment, since we use Paperclip (`has_attached_file`). We do want to migrate away though: https://otwarchive.atlassian.net/browse/AO3-5578

brianjaustin:
After looking at the new Rails 6.1 defaults, we should be fine to just enable all of them because:
1. We have been running Rails 6.1 for awhile, so the admonishments about backwards incompatibility aren't super relevant
2. We're not using ActiveStorage
3. Most of the changes are pretty low impact

https://lilyreile.medium.com/rails-6-1-new-framework-defaults-what-they-do-and-how-to-safely-uncomment-them-c546b70f0c5e gives a pretty good summary with the safety of each change.

(I did have to remove a few `inverse_of`s that were confusing Rails, but the tests are passing as before without them using the new default setting.)

config/environments/development.rb
```diff
@@ -1,47 +1,84 @@
...
+  # Debug mode disables concatenation and preprocessing of assets.
+  # This option may cause significant delays in view rendering with a large
+  # number of complex assets.
+  config.assets.debug = true
+
+  # Suppress logger output for asset requests.
+  config.assets.quiet = true
+
+  # Raises error for missing translations.
+  config.i18n.raise_on_missing_translations = true
```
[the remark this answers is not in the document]
sarken: I think this is a good idea -- it's easy to miss sometimes if the key matches the text!

config/environments/development.rb
```diff
@@ -1,47 +1,84 @@
...
+
+  # Annotate rendered view with file names.
+  # config.action_view.annotate_rendered_view_with_filenames = true
```
sarken, line 69: !!!

Oh, my god, this seems like it's going to be so useful. (But I agree it's a good idea to have it off by default.)

app/models/creatorship.rb
```diff
@@ -1,6 +1,6 @@
 class Creatorship < ApplicationRecord
-  belongs_to :pseud, inverse_of: :creatorships
-  belongs_to :creation, inverse_of: :creatorships, polymorphic: true, touch: true
+  belongs_to :pseud
```
sarken, line 2: I'm not having a lot of luck googling here -- do you have info about the change that allows us to remove `inverse_of`?

sarken (commented):
Before we call this ready to merge, I'd like @zz9pzza to take a look at [config/cable.yml](https://github.com/otwcode/otwarchive/pull/4690/files#diff-661620ae7be69a63f01b319f70bb6e3e0f6183b59258e9ed7aee623d75f17d90) and [config/environments/production.rb](https://github.com/otwcode/otwarchive/pull/4690/files#diff-da60b4e96eff2b132991226d308949e23f4ef3aad45ad59edd09cbc32cc6251e) when he has a chance.

bin/setup
```diff
@@ -1,17 +1,16 @@
 #!/usr/bin/env ruby
-require 'pathname'
-require 'fileutils'
-include FileUtils
+require "fileutils"
 
 # path to your application root.
-APP_ROOT = Pathname.new File.expand_path('../../', __FILE__)
+APP_ROOT = File.expand_path('..', __dir__)
```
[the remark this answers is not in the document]
brianjaustin: They don't have to, but for the most part I tried to stay close to what Rails's upgrade script generates, in hopes that it will make the 7.0 upgrade a bit easier

app/models/creatorship.rb
brianjaustin: It's a couple different things, so not the easiest to Google unfortunately :disappointed: The first bit is:
> Starting with Rails 4, the inverse_of setting is implicit between these [`has_many` and `belongs_to`] models, allowing you to traverse them bidirectionally.

However, that only worked in 1 direction prior to Rails 6.1 (I'd recommend the example in https://lilyreile.medium.com/rails-6-1-new-framework-defaults-what-they-do-and-how-to-safely-uncomment-them-c546b70f0c5e#4e5e or https://www.bigbinary.com/blog/rails-6-1-adds-support-for-belongs_to-to-has_many-inversing).

The next bit is that Rails 6.1 fixed the behavior to be traverse both directions.

The secret third thing is that we seem to run into https://github.com/rails/rails/issues/40867 with these (or something similar at least), so keeping the `inverse_of` will actually break things...

sarken: Ah, okay, thanks! I did read about `has_many_inversing` while I was looking for an explanation, but I dismissed that as not applicable based on a number of silly assumptions.

app/models/skin_parent.rb
```diff
@@ -1,6 +1,6 @@
 class SkinParent < ApplicationRecord
-  belongs_to :child_skin, class_name: "Skin", inverse_of: :skin_parents, touch: true
-  belongs_to :parent_skin, class_name: "Skin", inverse_of: :skin_children
+  belongs_to :child_skin, class_name: "Skin", touch: true
+  belongs_to :parent_skin, class_name: "Skin"
```
Bilka2, line 3: There is a comment about why `inverse_of:` is fine to remove on [creatorship.rb](https://github.com/otwcode/otwarchive/pull/4690/files#diff-174381727fd93f2148df7f93d54cad663246c8bc9fa72b3a31c258c76861de63).

However, I think it's not fine in this case. I'm re-reading the rails guide on associations for another PR and found [this](https://guides.rubyonrails.org/v6.1/association_basics.html#bi-directional-associations):
> However, Active Record will not automatically identify bi-directional associations that contain a scope or any of the following options:
>     :through
>     :foreign_key

Both `skin_parents` and `skin_children` use `:foreign_key`, so I think this removal breaks the bi-directionality.

brianjaustin: I think we're OK in this case, since `Skin` already has `inverse_of` as well: https://github.com/otwcode/otwarchive/blob/54e5f2640a129acd3a1e593362c62bc0a9b01957/app/models/skin.rb#L40-L47 I tried the experiment listed in the docs, and got the second (expected) result without `inverse_of` in `SkinParent`:

```
irb(main):001:0> skin = Skin.second
  Skin Load (0.5ms)  SELECT `skins`.* FROM `skins` ORDER BY `skins`.`id` ASC LIMIT 1 OFFSET 1
=> 
#<Skin:0x0000558c0be25210
...
irb(main):002:0> skin_parent = skin.skin_children.first
  SkinParent Load (0.3ms)  SELECT `skin_parents`.* FROM `skin_parents` WHERE `skin_parents`.`parent_skin_id` = 2 ORDER BY `skin_parents`.`id` ASC LIMIT 1
=> 
#<SkinParent:0x0000558c0bd6a028
...
irb(main):003:0> skin.title == skin_parent.parent_skin.title
=> true
irb(main):004:0> skin.title
=> "Archive 2.0: (1) core"
irb(main):005:0> skin.title = 'Foobar'
=> "Foobar"
irb(main):006:0> skin.title == skin_parent.parent_skin.title
=> true
```

Bilka2: Interesting to see the test of the example, thank you! Looks indeed fine then.

For some further reading, I found this discussion which argues that `inverse_of` on the `belongs_to` side of the `has_many` is never really helpful: https://github.com/rubocop/rubocop-rails/issues/632 (note that it is outdated regarding has_many_inversing).

config/environments/production.rb
zz9pzza: In the long term yes.

config/cable.yml
```diff
@@ -2,8 +2,9 @@ development:
   adapter: async
 
 test:
-  adapter: async
+  adapter: test
```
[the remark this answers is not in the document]
zz9pzza: Happy this is in test so fine.

zz9pzza (approved):

PR #4686, opened 2023-12-08
AO3-6548 Fix display of kudos image

EchoEkhi:
# Pull Request Checklist

* [x] Have you read ["How to write the perfect pull request"](https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/)?
* [x] Have you read the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [ ] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)?
* [x] Have you added the [Jira](https://otwarchive.atlassian.net) issue number
  as the *first* thing in your pull request title (e.g. `AO3-1234 Fix thing`)
* [x] Do you have fewer than 5 pull requests already open? If not, please wait
  until they are reviewed and merged before creating new pull requests.

## Issue

https://otwarchive.atlassian.net/browse/AO3-6548

## Purpose

Fix the CSS for the kudos icon and user icons in comments.

## Testing Instructions

See JIRA ticket.

## Credit

EchoEkhi (He/him)

PR #4691, opened 2023-12-12
AO3-6548 Use pseud helpers in comment emails and fix kudos image display

EchoEkhi:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6548

## Purpose

Fix email style in AO3-6548

## Testing Instructions

See JIRA ticket

## Credit

EchoEkhi (He/him)


Bilka2 (approved):
Small note: This sets the red color on both the `<b>` tag and the `<a>` tag. But I don't think this matters, `batch_subscription_notification.html.erb` does it too. If it does matter then this could use `tag.strong` instead of `style_bold`.

public/stylesheets/site/2.0/10-types-groups.css
```diff
@@ -190,12 +190,19 @@ span.series .position + .divider:after {
...
+  padding: 0.5em;
   margin: 2em 0;
   min-height: 40px;
 }
 
+p.kudos::before {
+  content: "";
+  background: url("/images/imageset.png") no-repeat -111px -580px;
+  width: 55px;
+  height: 54px;
```
Bilka2, line 202:
```suggestion
  background: url("/images/imageset.png") no-repeat -110px -580px;
  width: 55px;
  height: 55px;
```
This matches more what is in the png.

Bilka2:
I can confirm that the `::before` trick fixes the kudos image display. However, it's now positioned lower than before.
This PR on the left, staging in the middle and production on the right:
image


Furthermore, the text flows differently if it's long (= many kudos by users):
Screenshot_20231215_110334

I don't know whether that's okay. Also, the `min-height` change in #4686 moved the comment box up (visible in staging + dev in the first screenshot). Again, no idea if that is okay. But leaving my screenshots here in case it helps.

public/stylesheets/site/2.0/10-types-groups.css
EchoEkhi: image

It shows a bit of the adjacent images when viewed in Firefox if I change it to round numbers. Works fine on chrome though.

app/helpers/mailer_helper.rb
```diff
@@ -186,7 +186,7 @@ def commenter_pseud_or_name_link(comment)
     elsif comment.by_anonymous_creator?
       style_bold(t("roles.anonymous_creator"))
     else
-      style_link(comment.comment_owner_name, polymorphic_url(comment.comment_owner, only_path: false))
+      style_bold(style_link(comment.comment_owner_name, polymorphic_url(comment.comment_owner, only_path: false)))
```
brianjaustin, line 189: @sarken pointed out to me that we have a helper that can help here: https://github.com/otwcode/otwarchive/blob/93184cec8d7954bc6085153f375ae91de57b7e9e/app/helpers/mailer_helper.rb#L37-L44 (it does include the kudos icon, but we're OK with that as another way to distinguish registered users from guests). It seems to play nicely with `style_bold` but we're fine as long as one of the icon or bold is there

public/stylesheets/site/2.0/10-types-groups.css
```diff
@@ -190,12 +190,20 @@ span.series .position + .divider:after {
...
+  float: left;
+  margin-top: -0.5em;
+}
+
```
sarken, line 206: Since we've fixed the image, this change should be removed and we should go back to the original code without the float, which was:

https://github.com/otwcode/otwarchive/blob/f5d1fac3a815472ae5c3346450c46eb55fd3b46c/public/stylesheets/site/2.0/10-types-groups.css#L192-L197

sarken (changes requested):

sarken (changes requested):
Oh, wait, my mistake! `commenter_pseud_or_name_text` also needs to be updated to use `text_pseud` (the plain text equivalent of `style_pseud_link`).

sarken:
Bilka can't access GitHub right now, but he pointed out we forgot some mailers 🤦 

In `app/views/admin_mailer/`,  the following mailers need to be updated to use `commenter_pseud_or_name_link` and `commenter_pseud_or_name_text`:

* comment_notification.html.erb
* comment_notification.text.erb
* edited_comment_notification.html.erb
* edited_comment_notification.text.erb

Bilka2 (approved):

public/stylesheets/site/2.0/10-types-groups.css
```diff
@@ -189,11 +189,11 @@ span.series .position + .divider:after {
   content: "\25CF\0020";
 }
 
-p.kudos {
-  background: url("/images/imageset.png") no-repeat -111px -580px;
-  padding: 0.5em 0.5em 0.5em 60px;
-  margin: 2em 0;
-  min-height: 40px;
+p.kudos {
```
sarken, line 192: Could you get rid of the extra blank spaces at the ends of these lines? (I think it's possible ReviewDog doesn't run on CSS files, so it missed them.)

sarken (approved):

PR #4693, opened 2023-12-18
AO3-6652 Use --subset-embedded-fonts for PDFs

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6652

## Purpose

Our PDFs have embedded fonts, which means they can get quite large if we're embedding a font with a lot of characters (e.g., I created a work with a Japanese title; the required fandom, rating, and warning tags; and five words of text. The PDF download was 19 MB). 

This adds [Calibre's --subset-embedded-fonts option](https://manual.calibre-ebook.com/generated/en/ebook-convert.html#cmdoption-ebook-convert-subset-embedded-fonts) to the command for generating PDFs so we only include the characters actually used in the PDF.

I wasn't sure if adding a test for this was practical. I could probably write a spec to check the size of the PDF file was smaller than the size of the font we use, but that would of course depend on the test environment having the font installed in the same format. (There's [a Linux command for installing the font](http://wenq.org/wqy2/index.cgi?ZenHei%28en%29), so installing it might be feasible. However, I'd note that I couldn't manually download the font to install the font on my machine. I thought the problem was the site was insecure so my browser was blocking the download, but when I tried in a less fussy browser, the download link just gave me a sort of domain parking message. 🤷 )

## Testing Instructions

Refer to Jira.

## Credit

Sarken, she/her


sarken:
(Playing with the labels on this to test something; don't mind me.)

sarken:
Putting the wrong label on this to test labeler.

sarken:
Putting the right label on it now!

app/models/download_writer.rb
```diff
@@ -74,15 +74,18 @@ def get_calibre_command
     ### Format-specific options
     # epub: don't generate a cover image
     epub = download.file_type == "epub" ? ["--no-default-epub-cover"] : []
-    # pdf: decrease margins from 72pt default
+    # pdf:
+    # - decrease margins from 72pt default
+    # - only include necessary characters when embedding fonts
```
redsummernight, line 79: Can you move the comments to just above the flags they're about? Like the "ebook-convert" array below.

sarken: How do we want it?

Option A:
```
    ### Format-specific options
    # epub: don't generate a cover image
    epub = download.file_type == "epub" ? ["--no-default-epub-cover"] : []
    # pdf:
    pdf = []
    if download.file_type == "pdf"
      pdf = [
        # decrease margins from 72pt default
        "--pdf-page-margin-top", "36",
        "--pdf-page-margin-right", "36",
        "--pdf-page-margin-bottom", "36",
        "--pdf-page-margin-left", "36",
        "--pdf-default-font-size", "17",
        # only include necessary characters when embedding fonts
        "--subset-embedded-fonts"
      ]
    end
```

Option B:
```
    ### Format-specific options
    # epub: don't generate a cover image
    epub = download.file_type == "epub" ? ["--no-default-epub-cover"] : []
    pdf = []
    if download.file_type == "pdf"
      pdf = [
        # pdf: decrease margins from 72pt default
        "--pdf-page-margin-top", "36",
        "--pdf-page-margin-right", "36",
        "--pdf-page-margin-bottom", "36",
        "--pdf-page-margin-left", "36",
        "--pdf-default-font-size", "17",
        # pdf: only include necessary characters when embedding fonts
        "--subset-embedded-fonts"
      ]
    end
```

Or something else?

redsummernight: I think either B, or A without that "# pdf:".

sarken: I think I'll go with B -- with A without the `# pdf:` comment, it's not entirely clear we've switched from modifying just EPUBs to just PDFs

redsummernight (approved):

PR #4701, opened 2023-12-21
AO3-6653 Fix removal of Reviewed: Action Needed label

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6653

## Purpose

The "Reviewed: Action Needed" label is no longer getting removed when the coder pushes a commit after the label is replied. ("Coder Actioned Review" is getting applied.) This is a shot in the dark at fixing it. 

> It's the inverse of the condition to add "Coder Has Actioned Review." We have `sync-labels` set to true, which says, "Whether or not to remove labels when matching files are reverted or no longer changed by the PR"
>
>So I _think_ that means we'd be saying in the config file, "add the 'Reviewed: Action Needed' label if there are no changes in this commit," and then the sync setting would mean that since there are changes, the label would not apply, so it would get removed? Maybe? It's a whole lot of negating in here

## Testing Instructions

I temporarily changed `on` to `pull_request` to make sure there's no syntax error in the new config; however, the action will still error when attempting to add the label. I'm going to change the label on this PR to "Reviewed: Action Needed" and then push a random change to trigger the proper action. If that fails in the expected way, I'll remove the temporary changes. 

**Update:** Here's the [action with the expected failure](https://github.com/otwcode/otwarchive/actions/runs/7293658602/job/19877148771?pr=4701#step:4:9); I've now removed the temporary changes.

Once that temporary change is removed, we'll have to merge this, find a pull request with the "Reviewed: Action Needed" label, and push a change to it. If "Reviewed: Action Needed" is removed when "Coder Has Actioned Review" is added, it works.

### DIFF
--- a/.github/labeler/review.yml
+++ b/.github/labeler/review.yml
@@ -4,4 +4,6 @@
   - any-glob-to-any-file: "**/*"
 
 # Label to be removed when the PR is updated:
-"Reviewed: Action Needed": []
+"Reviewed: Action Needed":
+- changed-files:
+  - any-glob-to-any-file: "!**/*"
--- a/.github/labeler/review.yml
+++ b/.github/labeler/review.yml
@@ -3,7 +3,7 @@
 - changed-files:
   - any-glob-to-any-file: "**/*"
 
-# Label to be removed when the PR is updated:
+# Condition is always false so the label is removed when the PR is updated:
 "Reviewed: Action Needed":
 - changed-files:
   - any-glob-to-any-file: "!**/*"


.github/labeler/review.yml
```diff
@@ -4,4 +4,6 @@
   - any-glob-to-any-file: "**/*"
 
 # Label to be removed when the PR is updated:
-"Reviewed: Action Needed": []
+"Reviewed: Action Needed":
+- changed-files:
+  - any-glob-to-any-file: "!**/*"
```
redsummernight, line 9: Can you add a comment that this condition is always false because we want to always remove the label on updates?

redsummernight (approved):

sarken:
Yay, [here is the action where it worked](https://github.com/otwcode/otwarchive/pull/4693#event-11330397974) after being merged.

PR #4714, opened 2024-01-09
AO3-6667 Languages page performance improvements

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6667

## Purpose

* Fetch language counts from Elasticsearch instead of the primary DB
* Round counts a bit to make temporary discrepancies (in the ones or tens place) less impactful. The counts don't need to be exact anyway
* Cache counts for 1 day to further reduce load


config/locales/views/en.yml
```diff
@@ -528,6 +528,15 @@ en:
       more_link:
         one: "%{count} more user"
         other: "%{count} more users"
+  languages:
+    index:
+      add: Add a language
+      description: Work Languages
+      edit: Edit
+      suggest: Suggest a language
```
sarken, line 536: Could you keep the original capitalization from the navigation actions? Our navigation should always be title case. (You can, however, drop the "a" if you want.) It also wouldn't hurt to group them under `navigation` like we do elsewhere to give Translation a bit more context, e.g. https://github.com/otwcode/otwarchive/blob/653c19c250d9e35b2ea33e86a51b7b28c1e4c975/config/locales/views/en.yml#L356-L361

app/views/languages/index.html.erb
```diff
@@ -1,38 +1,37 @@
 <!--Descriptive page name, messages and instructions-->
-<h2 class="heading"><%= ts('The Archive supports these languages') %></h2>
+<h2 class="heading"><%= t(".description") %></h2>
```
sarken, line 2: Can we call this `page_heading`? That way if we ever have to add a description to the page (i.e., a paragraph under the heading), we don't need to rename. It's also not a bad idea to have a consistent naming scheme for bits that appear on all pages, and it looks like that one is used pretty frequently.

sarken (commented):
Just took a quick look and it's possible the failing tests need `work_search: true` to set up the indexes, like https://github.com/otwcode/otwarchive/blob/653c19c250d9e35b2ea33e86a51b7b28c1e4c975/spec/controllers/collection_participants_controller_spec.rb#L113

config/locales/views/en.yml
brianjaustin: Sounds good! I think keeping the `a` sounds better but I can remove it if that's the standard

sarken (approved):

PR #4715, opened 2024-01-11
AO3-6662 Catch errors in rake tasks deleting zero use tags

zz9pzza:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6662

## Purpose

The rake task has been erroring and not deleting all the tags 

## Testing Instructions

Follow instructions on jira

## References

## Credit

james_, he/him

### DIFF
--- a/lib/tasks/tag_tasks.rake
+++ b/lib/tasks/tag_tasks.rake
@@ -57,7 +57,7 @@ namespace :Tag do
       print "."
     end
     unless deleted_names.blank?
-      puts "The following #{deleted_names.length} unused tags were deleted:"
+      puts "The following unused tags were deleted:"
       puts deleted_names.join(", ")
       puts "Started #{start}"
       puts "Ended #{Time.current}"
--- a/lib/tasks/tag_tasks.rake
+++ b/lib/tasks/tag_tasks.rake
@@ -43,6 +43,7 @@ namespace :Tag do
         next unless t.taggings.count.zero?
         next unless t.child_taggings.count.zero?
         next unless t.set_taggings.count.zero?
+
         deleted_names << t.name
         begin
           deleted_names << t.name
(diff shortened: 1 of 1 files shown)


lib/tasks/tag_tasks.rake
```diff
@@ -36,16 +36,33 @@ namespace :Tag do
...
+    Tag.where(canonical: false, merger_id: nil, taggings_count_cache: 0).find_in_batches do |batch|
+      batch.each do |t|
+        next unless t.taggings.count.zero?
+        next unless t.child_taggings.count.zero?
+        next unless t.set_taggings.count.zero?
+
         deleted_names << t.name
-        t.destroy
+        begin
+          deleted_names << t.name
```
Bilka2, line 49: This pushes each tag into the array (`deleted_names << t.name`) at least twice. I think you'll want to remove the second statement here, otherwise it will also push it into the array again for each retry.

Bilka2:
From what I can tell this PR is fine and works, but I think it should also be reviewed by someone more senior than me.

lib/tasks/tag_tasks.rake
```diff
@@ -36,16 +36,32 @@ namespace :Tag do
...
+        end
       end
+      print "."
     end
     unless deleted_names.blank?
       puts "The following #{deleted_names.length} unused tags were deleted:"
       puts deleted_names.join(", ")
+      puts "Started #{start}"
+      puts "Ended #{Time.current}"
+      puts "Deleted #{deleted_names.length}"
```
sarken, line 64: I think we can remove this line. The length is already there a few lines up: `puts "The following #{deleted_names.length} unused tags were deleted:"`

zz9pzza: If it delete's thousands of tags it is really hard to find the start. The end of the log is really easy to find. I could delete the one at the top ?

sarken: Yeah, that's fine!

sarken (approved):

PR #4716, opened 2024-01-13
AO3-6570 Import Work error message overflows for long URLs

smclairecarden:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6570

## Purpose

Set word-wrap to break-word on the error element so that the text wraps properly and does not overflow.

## Testing Instructions

Log in. 
Head to Import Work and enter a long external URL
Import it more than once and make sure the element doesn't spill out of the box - make your browser window small. 

## References

## Credit

Claire C
Screen Shot 2024-01-12 at 8 30 38 PM

they/them pronouns.*

### DIFF
--- a/public/stylesheets/site/2.0/22-system-messages.css
+++ b/public/stylesheets/site/2.0/22-system-messages.css
@@ -37,6 +37,7 @@ div.error {
   clear: right;
     box-shadow: inset 1px 1px 2px;
     border-radius: 0.25em;
+  word-wrap: break-word;
 }
 
 .caution {
@@ -49,7 +50,6 @@ div.error {
   background: #efd1d1;
   border-color: #900;
     box-shadow: 1px 1px 2px;
-    word-wrap: break-word;
 }
 
 .notes {
--- a/public/stylesheets/site/2.0/22-system-messages.css
+++ b/public/stylesheets/site/2.0/22-system-messages.css
@@ -49,6 +49,7 @@ div.error {
   background: #efd1d1;
   border-color: #900;
     box-shadow: 1px 1px 2px;
+    word-wrap: break-word;
 }
 
 .notes {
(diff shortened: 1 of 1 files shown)


public/stylesheets/site/2.0/22-system-messages.css
```diff
@@ -49,6 +49,7 @@ div.error {
   background: #efd1d1;
   border-color: #900;
     box-shadow: 1px 1px 2px;
+    word-wrap: break-word;
```
sarken, line 52: Can we move this up to the style rule for `.notice, .comment_notice, .kudos_notice, ul.notes, .caution, .error, .comment_error, .kudos_error, .alert.flash` instead? I think it makes sense to ensure it will be applied to any of our blue success or yellow caution messages as well.

sarken (commented):
Just a small question/suggestion! If there's a reason it won't work, this is fine as-is.

public/stylesheets/site/2.0/22-system-messages.css
smclairecarden: Ah okay you were right about this, it works! Updated.

public/stylesheets/site/2.0/22-system-messages.css
```diff
@@ -37,6 +37,7 @@ div.error {
   clear: right;
     box-shadow: inset 1px 1px 2px;
     border-radius: 0.25em;
+  word-wrap: break-word;
```
sarken, line 40: If you could nudge this in 2 more spaces so it's aligned with `border-radius` above it, that would be great! (CSS3 was new and note widely-supported when we first wrote these stylesheets, so we decided to put the new CSS3 properties at the end of rules, with an extra indent, to make it obvious where we were using them. It's a dated practice now, but consistency doesn't hurt.)

smclairecarden: Apologies for that, done. I debated where to align it and chose incorrectly!! Thanks for the explanation.

sarken (approved):

PR #4717, opened 2024-01-14
AO3-6669 Search for noncanonical tags for autocomplete via Elasticsearch

zz9pzza:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6669

## Purpose

What does this PR do?

## Testing Instructions

## References

## Credit

they/them pronouns.*


app/controllers/autocomplete_controller.rb
```diff
@@ -91,16 +91,30 @@ def associated_tags
...
     raise "Redshirt: Attempted to constantize invalid class initialize noncanonical_tag #{params[:type].classify}" unless Tag::TYPES.include?(params[:type].classify)
+
     tag_class = params[:type].classify.constantize
-    render_output(tag_class.by_popularity
-                      .where(["canonical = 0 AND name LIKE ?",
-                              '%' + search_param + '%']).limit(10).map(&:name))
+    one_tag = tag_class.where(canonical: false, name: search_param)
+    # Is there a tag which is just right ( this is really for testing )
+    match = []
+    match = [one_tag.first.name] if one_tag.count == 1
```
ceithir, line 99: Is ES not systematically returning this tag with a high relevance?

app/controllers/autocomplete_controller.rb
```diff
@@ -91,16 +91,30 @@ def associated_tags
...
+
+    word_list = search_param.split
+    last_word = word_list.pop
+    search_list = word_list.map { |w| { term: { name: { value: w, case_insensitive: true } } } } + [{ prefix: { name: { value: last_word, case_insensitive: true } } }]
```
ceithir, line 103: Out of curiosity, what is this request inspired by? I couldn't find anything similar in the current code or the [documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-suggesters.html).

app/controllers/autocomplete_controller.rb
zz9pzza: Put simply it is not guaranteed to and the test we have which is "Stargates SG-1" is an example were it does not work.

app/controllers/autocomplete_controller.rb
zz9pzza: From staging:

```
3.0.5 :133 > example("kaeinui")
 => ["Kaeinui Chwihyang | Personal Preference"]
 =>
["The Amazing Adventures of Kavalier & Clay",
 "Kare Kano",
 "Kaiji (anime)",
 "Amazing Adventures of Kavalier & Clay - Chabon",
 "Kabuki",
 "Kal Ho Naa Ho",
 "Kate Shugak Series",
 "Fionavar Tapestry - Guy Gavriel Kay",
 "Amazing Adventures of Kavalier & Clay - Michael Chabon",
 "Karakuridôji Ultimo/機巧童子ULTIMO"]
3.0.5 :135 > example("kaein")
 => ["Kaeinui Chwihyang | Personal Preference"]
3.0.5 :136 > example("kaei")
 => ["Kaeinui Chwihyang | Personal Preference", "Aoi Shiro - Kaeishou"]
3.0.5 :137 > example("kae")
 =>
["Kaeinui Chwihyang | Personal Preference",
 "Kael'thas Sunstrider/Illidan Stormrage",
 "Kael'thas/Illidan",
 "Kaelthas - Fandom",
 "Albox - Eien o Kaetara",
 "Kael'thas Sunstrider (character)",
 "Kael'thas Sunstrider - Fandom",
 "Aoi Shiro - Kaeishou",
 "Illidan x Kael'thas",
 "Illidan/Kael'thas - relationship"]
3.0.5 :138 > example("Kaeinui Chwihyang | Personal Preference")
 => ["Kaeinui Chwihyang | Personal Preference"]
```

Am I holding suggestions badly ?

```
curl -X POST 'localhost:9200/ao3_staging_tags/_search?pretty' -H 'Content-Type: application/json' '-d
{
  "suggest": {
    "my-suggest-1" : {
      "text" : "Kae",
      "term" : {
        "field" : "name"
      }
    }
  }
}
'
{
  "took" : 2,
  "timed_out" : false,
  "_shards" : {
    "total" : 5,
    "successful" : 5,
    "skipped" : 0,
    "failed" : 0
  },
  "hits" : {
    "total" : {
      "value" : 0,
      "relation" : "eq"
    },
    "max_score" : null,
    "hits" : [ ]
  },
  "suggest" : {
    "my-suggest-1" : [
      {
        "text" : "kae",
        "offset" : 0,
        "length" : 3,
        "options" : [ ]
      }
    ]
  }
}
```

ceithir: Okay, I've tried using the [Completion Suggester](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-suggesters.html#completion-suggester): https://github.com/ceithir/otwarchive/commit/9cd6bbb228504278678d8a54e47da9768a6181dd

... Raised an eyebrow at the fact it [only supports](https://stackoverflow.com/questions/48304499/elasticsearch-completion-suggester-not-working-with-whitespace-analyzer#answer-54080114) completion from the left (there's possibly a way to work around that with regex, but haven't delved that deep)...

And had said eyebrow curling at what needed to be done to add the equivalent of a filter query.

It might indeed not be what's best for our current usage in spite of it theoretically be ES auto-complete feature.

app/controllers/autocomplete_controller.rb
```diff
@@ -88,18 +88,36 @@ def associated_tags
...
+    search_param = Query.new.escape_reserved_characters(params[:term])
     raise "Redshirt: Attempted to constantize invalid class initialize noncanonical_tag #{params[:type].classify}" unless Tag::TYPES.include?(params[:type].classify)
+
     tag_class = params[:type].classify.constantize
-    render_output(tag_class.by_popularity
-                      .where(["canonical = 0 AND name LIKE ?",
-                              '%' + search_param + '%']).limit(10).map(&:name))
+    one_tag = tag_class.find_by(canonical: false, name: search_param)
+    # Is there a tag which is just right ( this is really for testing )
+    match = if one_tag then [one_tag.name]  else [] end # rubocop:disable Style/OneLineConditional
```
sarken, line 97: We really don't do one-line conditionals like this, hence the cop. Either of these would be preferable, please:

```
    match = if one_tag
              [one_tag.name]
            else
              []
            end

    match = one_tag ? [one_tag.name] : []
 ```
 
 Or, since it sounds like Brian's original suggestion didn't work, perhaps even:
 
 ```
     match = []
     match << one_tag.name unless one_tag.nil?
 ```

File: app/controllers/autocomplete_controller.rb [hunk 4]
```diff
@@ -88,18 +88,36 @@ def associated_tags
...
-    search_param = params[:term]
+    search_param = Query.new.escape_reserved_characters(params[:term])
     raise "Redshirt: Attempted to constantize invalid class initialize noncanonical_tag #{params[:type].classify}" unless Tag::TYPES.include?(params[:type].classify)
+
     tag_class = params[:type].classify.constantize
-    render_output(tag_class.by_popularity
-                      .where(["canonical = 0 AND name LIKE ?",
-                              '%' + search_param + '%']).limit(10).map(&:name))
+    one_tag = tag_class.find_by(canonical: false, name: search_param)
+    # Is there a tag which is just right ( this is really for testing )
```
sarken, line 96:
```suggestion
    # Is there a tag which is just right (this is really for testing)
```

(The spaces inside are right for code! Just not so much comments 😆)

sarken (commented):
Two minor style things I noticed!

app/controllers/autocomplete_controller.rb
```diff
@@ -88,18 +88,40 @@ def associated_tags
...
+    # Is there a tag which is just right (this is really for testing)
+    match = if one_tag
+              [one_tag.name]
+            else
+              []
+            end
+
+    # As explained in https://stackoverflow.com/a/54080114, the Elasticsearch suggestion suggester does not support
+    # matches in the middle of a series of words. Therefore, we break the autocomplete query into its individual
+    # words – based on whitespace – except for the last word, which could be a incomplete, so a prefix match is
```
sarken, line 105: Is this a typo? `could be a incomplete`

app/controllers/autocomplete_controller.rb
```diff
@@ -88,18 +88,40 @@ def associated_tags
...
+    # appropriate. This emulates the behavior of SQL `LIKE '%text%'.
+
+    word_list = search_param.split
+    last_word = word_list.pop
+    search_list = word_list.map { |w| { term: { name: { value: w, case_insensitive: true } } } } + [{ prefix: { name: { value: last_word, case_insensitive: true } } }]
+    begin
+      # Size is chosen so we get enough search results from each shard.
+      search_results = $elasticsearch.search(
+        index: "#{ArchiveConfig.ELASTICSEARCH_PREFIX}_#{Rails.env}_tags",
```
sarken, line 114: Does `TagIndexer.index_name` work here in place of `#{ArchiveConfig.ELASTICSEARCH_PREFIX}_#{Rails.env}_tags`? I think it would be better to use `index_name` rather than assuming we'll remember to update this should the index name change.

app/controllers/autocomplete_controller.rb
```diff
@@ -88,18 +88,40 @@ def associated_tags
...
-    search_param = params[:term]
+    search_param = Query.new.escape_reserved_characters(params[:term])
     raise "Redshirt: Attempted to constantize invalid class initialize noncanonical_tag #{params[:type].classify}" unless Tag::TYPES.include?(params[:type].classify)
+
     tag_class = params[:type].classify.constantize
-    render_output(tag_class.by_popularity
-                      .where(["canonical = 0 AND name LIKE ?",
-                              '%' + search_param + '%']).limit(10).map(&:name))
+    one_tag = tag_class.find_by(canonical: false, name: params[:term])
+    # Is there a tag which is just right (this is really for testing)
```
sarken, line 96: Could you explain how this is just for testing? I see `match` is used in rendering the output so I am slightly confused.

app/controllers/autocomplete_controller.rb
```diff
@@ -88,18 +88,40 @@ def associated_tags
...
   def noncanonical_tag
-    search_param = params[:term]
+    search_param = Query.new.escape_reserved_characters(params[:term])
     raise "Redshirt: Attempted to constantize invalid class initialize noncanonical_tag #{params[:type].classify}" unless Tag::TYPES.include?(params[:type].classify)
+
     tag_class = params[:type].classify.constantize
-    render_output(tag_class.by_popularity
-                      .where(["canonical = 0 AND name LIKE ?",
-                              '%' + search_param + '%']).limit(10).map(&:name))
+    one_tag = tag_class.find_by(canonical: false, name: params[:term])
```
sarken, line 95: Is there any chance of us regretting this if someone uses a MySQL wildcard character like `%` or `_`?

brianjaustin: I don't think so; I tried running
```irb
3.1.4 :001 > Tag.find_by(canonical: false, name: '%Stargate%')
 => nil
```
and it returned `nil` pretty fast, so I think there's some escaping going on with `find_by`

app/controllers/autocomplete_controller.rb
brianjaustin: @zz9pzza please correct me if I'm wrong, but this is hit in if a user types the whole tag into autocomplete without choosing one of the suggestions first. The tests definitely do that, but in theory a real user could as well (although that seems potentially unlikely)

sarken: If this is used where I think it's used (the Synonyms field on canonical tags' edit pages), there's actually a pretty good choice the wrangler will be entering the name of the exact tag they're looking for, fwiw!

app/controllers/autocomplete_controller.rb
zz9pzza: This is protected by the normal rails protection. 

https://stackoverflow.com/questions/9249624/should-i-escape-the-condition-for-my-rails-find

app/controllers/autocomplete_controller.rb
zz9pzza: The code will not find exact matches naturally, adding the check for the exact match match makes this use case work and the elasticsearch code finds good matches for the autocomplete.

sarken (approved):

PR #4719, opened 2024-01-17
AO3-6659 Option to ban a user and delete all their creations includes collections the user doesn't own

smclairecarden:
# Pull Request Checklist

* [X ] Have you read ["How to write the perfect pull request"](https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/)?
* [X ] Have you read the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [X ] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)?
* [ X] Have you added the [Jira](https://otwarchive.atlassian.net) issue number
* [ X] Do you have fewer than 5 pull requests already open? If not, please wait

## Issue

https://otwarchive.atlassian.net/browse/AO3-6659

## Purpose

What does this PR do?

Changed the way user collections are counted. It is using the logic for  sole-owned collections.

## Testing Instructions

See Jira.

## References

## Credit

Claire C, she/her

they/them pronouns.*


app/controllers/admin/admin_users_controller.rb
```diff
@@ -133,7 +133,7 @@ def confirm_delete_user_creations
     @works = @user.works.paginate(page: params[:works_page])
     @comments = @user.comments.paginate(page: params[:comments_page])
     @bookmarks = @user.bookmarks
-    @collections = @user.collections
+    @collections = @user.collections.to_a.delete_if { |collection| !(collection.all_owners - @user.pseuds).empty? }
```
ceithir, line 136: Might be worth adding an explicit `def sole_owned_collections` to `user.rb`?

ceithir (commented):
Could you add some test(s)? Likely in [admin_users_controller_spec.rb](https://github.com/otwcode/otwarchive/blob/54e5f2640a129acd3a1e593362c62bc0a9b01957/spec/controllers/admin/admin_users_controller_spec.rb#L401)

ceithir (changes requested):
Shouldn't [destroy_user_creations](https://github.com/otwcode/otwarchive/blob/54e5f2640a129acd3a1e593362c62bc0a9b01957/app/controllers/admin/admin_users_controller.rb#L142) also be updated so the code doesn't actually delete all collections regardless of what the display shows?

app/controllers/admin/admin_users_controller.rb
smclairecarden: I updated destroy_user_creations, thank you so much for that catch. All is working as expected.

I can add it here, but not completely sure what it would do? Would this be creating a universal variable?

smclairecarden:
> Could you add some test(s)? Likely in [admin_users_controller_spec.rb](https://github.com/otwcode/otwarchive/blob/54e5f2640a129acd3a1e593362c62bc0a9b01957/spec/controllers/admin/admin_users_controller_spec.rb#L401)

There are tests starting on line 440 "POST #destroy_user_creations" that seems to go through these actions I edited. Would I be ensuring the count is correct?

app/controllers/admin/admin_users_controller.rb
ceithir: I meant adding a method in the User class like:
```rb
class User < ApplicationRecord
# ...
  def sole_owned_collections
    collections.to_a.delete_if { |collection| !(collection.all_owners - @user.pseuds).empty? }
  end
```
So we can just call `user.sole_owned_collections` in relevant locations instead of copy-pasting the full line every time.

ceithir:
> There are tests starting on line 440 "POST #destroy_user_creations" that seems to go through these actions I edited. Would I be ensuring the count is correct?

Tests in a context where two collections exist, one where the user is the sole owner, and one where they aren't, and checking only the first is effectively listed/destroyed would be great.

There are factories to setup such a context relatively painlessly ([example of use](https://github.com/otwcode/otwarchive/blob/54e5f2640a129acd3a1e593362c62bc0a9b01957/spec/controllers/collection_participants_controller_spec.rb#L327-L333)) and `assigns` ([example of use](https://github.com/otwcode/otwarchive/blob/54e5f2640a129acd3a1e593362c62bc0a9b01957/spec/controllers/bookmarks_controller_spec.rb#L325)) can be used to check the state of the `@collections` variable.

app/controllers/admin/admin_users_controller.rb
weeklies: Maybe using [@user.owned_collections](https://github.com/otwcode/otwarchive/blob/54e5f2640a129acd3a1e593362c62bc0a9b01957/app/models/user.rb#L102) and adding a new filter on it would be more efficient? 

Also, we should update [this line](https://github.com/otwcode/otwarchive/blob/54e5f2640a129acd3a1e593362c62bc0a9b01957/app/controllers/users_controller.rb#L180) as well.

smclairecarden: Yes thank you, I remembered that one, its where I got the code from! I added a def sole_owned_collections, but can change to the filter if ya'll think that's best.

weeklies:
The test is failing because `collection1` has two owners, one of them created for you by the factory. Overwriting it will fix this, but I'm not sure if it's the most elegant way to do this:

```
collection1.collection_participants = [create(:collection_participant, user: user, collection: collection1, participant_role: CollectionParticipant::OWNER)]
collection1.save
```

smclairecarden:
> The test is failing because `collection1` has two owners, one of them created for you by the factory. Overwriting it will fix this, but I'm not sure if it's the most elegant way to do this:
> 
> ```
> collection1.collection_participants = [create(:collection_participant, user: user, collection: collection1, participant_role: CollectionParticipant::OWNER)]
> collection1.save
> ```

Thank you so so much for that. As you can see from my commit history I've been on quite the journey, couldn't figure it out. This test is passing now, let me know if there are any additional edits I should make before approval. Really appreciate the tip!!!

spec/controllers/admin/admin_users_controller_spec.rb
```diff
@@ -463,14 +474,19 @@ def expect_changes_made_by(admin, log_items)
...
-
-          expect(response).to have_http_status(:success)
+          expect do
+            post :confirm_delete_user_creations, params: { id: user.login }  
+            post :destroy_user_creations, params: { id: user.login }
+          end.to change(Collection, :count).by(-1)
+          # Check that the first user's collection is deleted
+          expect(Collection.exists?(collection1.id)).to be_falsey
+          # Check that the second user's collection still exists
+          expect(Collection.exists?(other_owner.collections.last.id)).to be_truthy
```
ceithir, line 489: As we're effectively testing two different things, I feel like we want to keep the existing test as is, and make a fully new, additional test (even if it's 50% a copy-paste of the old one).

app/models/user.rb
```diff
@@ -470,6 +470,11 @@ def coauthored_works
     return @coauthored_works
   end
 
+  # checks that only collections a user has created are deleted.
```
ceithir, line 473: Comment might be superfluous? Method name and definition are already relatively explicit.

spec/controllers/admin/admin_users_controller_spec.rb
```diff
@@ -476,6 +476,37 @@ def expect_changes_made_by(admin, log_items)
     end
   end
 
+  describe "POST #destroy_user_creations" do
```
weeklies, line 479: You probably don't have to create a new `describe` block as the examples above are testing the same method. You can add the `it` block within the context "when user is banned` below [this line](https://github.com/otwcode/otwarchive/pull/4719/files#diff-0db7b6f87501fc226438013cfe7b2f9bc5fa7fd2ee0a387a91ddb591534c1af2R466).

spec/controllers/admin/admin_users_controller_spec.rb
```diff
@@ -476,6 +476,37 @@ def expect_changes_made_by(admin, log_items)
...
+    context "when deleting the first user" do
+      it "deletes the first user's collection but preserves the second user's collection" do
+        admin.update(roles: ["policy_and_abuse"])
+        fake_login_admin(admin)
+        expect do
+          post :confirm_delete_user_creations, params: { id: user.login }  
+          post :destroy_user_creations, params: { id: user.login }
+        end.to change(Collection, :count).by(-1)
+        # Check that the first user's collection is deleted
+        expect(Collection.exists?(collection1.id)).to be_falsey
```
weeklies, line 503: To be precise, you can use `to be false` and `to be true`.

app/models/user.rb
```diff
@@ -470,6 +470,11 @@ def coauthored_works
...
+  def sole_owned_collections
+    self.collections.to_a.delete_if { |collection| !(collection.all_owners - pseuds).empty? }
```
weeklies, line 475: I think using this would be more efficient: `self.owned_collections.select { |collection| collection.all_owners.count == 1 }`

spec/controllers/admin/admin_users_controller_spec.rb
```diff
@@ -476,6 +476,37 @@ def expect_changes_made_by(admin, log_items)
...
+    before do
+      collection1.collection_participants = [create(:collection_participant, user: user, collection: collection1, participant_role: CollectionParticipant::OWNER)]
+      collection1.save
+      # Create the second user and make them an owner of the collection
+      create(:collection_participant, user: other_owner, collection: collection2, participant_role: CollectionParticipant::OWNER)
+      create(:collection_participant, user: user, collection: collection2, participant_role: CollectionParticipant::MEMBER)
+    end
+
+    context "when deleting the first user" do
+      it "deletes the first user's collection but preserves the second user's collection" do
```
weeklies, line 495: How about `it "only deletes the user's sole owned collections"`?

spec/controllers/admin/admin_users_controller_spec.rb
```diff
@@ -476,6 +476,37 @@ def expect_changes_made_by(admin, log_items)
...
+        admin.update(roles: ["policy_and_abuse"])
+        fake_login_admin(admin)
+        expect do
+          post :confirm_delete_user_creations, params: { id: user.login }  
+          post :destroy_user_creations, params: { id: user.login }
+        end.to change(Collection, :count).by(-1)
```
weeklies, line 501: I think the assertions that follow are enough, so how about simplifying this to:
```suggestion
        post :destroy_user_creations, params: { id: user.login }
```

weeklies (changes requested):
Just a few suggestions.

spec/controllers/admin/admin_users_controller_spec.rb
smclairecarden: @ceithir wanted a completely separate test.

spec/controllers/admin/admin_users_controller_spec.rb
smclairecarden: Is this better than falsey and Truthy? Id rather keep it that way if it doesn't make much difference.

spec/controllers/admin/admin_users_controller_spec.rb
smclairecarden: I like having this extra check in there, personally.

spec/controllers/admin/admin_users_controller_spec.rb
smclairecarden: I suppose I could change it, I just wanted to make it clear in the "it" statement the second collection would remain.

app/models/user.rb
smclairecarden: I'm sure if your statement is checking for the same thing mine is. I'd rather keep it as is.

app/models/user.rb
sarken: I'm not actually sure the comment is clear -- we're not really checking the deletion of anything here, are we? We're returning an array of collections that are owned by this user and only this user, like `sole_authored_works` above is described as `# Returns array of works where the user is the sole author`. The actual deletion comes at some other point.

app/models/user.rb
sarken: Yeah, I think @smclairecarden is right and these are two different checks. It looks like `all_owners` returns a list of pseuds, and it's possible for two pseuds belonging to the same user to be owners of a collection.

spec/controllers/admin/admin_users_controller_spec.rb
ceithir: A distinct test i.e. `it`, with its own relevant context indeed. They can definitely share the same `describe` though.

smclairecarden: I misunderstood then. If I add it back in to the original test as its own it, would I need line 499? Since this is initiated in the last test of the first destroy_user_creations?

ceithir:
Okay, so... Turns out the already existing tests on the POST route [were wrong](https://github.com/otwcode/otwarchive/pull/4723). Which might explain some of the confusion regarding them.

ceithir:
> > Okay, so... Turns out the already existing tests on the POST route [were wrong](https://github.com/otwcode/otwarchive/pull/4723). Which might explain some of the confusion regarding them.
> 
> I see your pull request. Are we going to wait for yours to be merged to fix mine?

Since it's just one small commit, might be quicker to cherry-pick it into the present PR, close my own, and getting it all merged at once?

smclairecarden:
> > > Okay, so... Turns out the already existing tests on the POST route [were wrong](https://github.com/otwcode/otwarchive/pull/4723). Which might explain some of the confusion regarding them.
> > 
> > 
> > I see your pull request. Are we going to wait for yours to be merged to fix mine?
> 
> Since it's just one small commit, might be quicker to cherry-pick it into the present PR, close my own, and getting it all merged at once?

Okay, I can do this. I noticed in your test you are blocking the user though at the very end. What does this mean for my test, since you want me to move it into the same describe? Should I create another user for my test specifically? Does the test forget what has happened outside each individual "it"?

ceithir:
> Does the test forget what has happened outside each individual "it"?

The database is indeed [reset](https://github.com/otwcode/otwarchive/blob/fd7efe103579bab1ec6b05aa6cf3d469f9a08411/spec/spec_helper.rb#L57) before each and every test. When a test is at its first line, the only objects existing in it are the ones that have been created as a consequence of contextual `let!` and `before do`.

I think the user needs to be banned for the case you'll be testing though, so you can put the test within the `"when user is banned"` context.

smclairecarden:
> > Does the test forget what has happened outside each individual "it"?
> 
> The database is indeed [reset](https://github.com/otwcode/otwarchive/blob/fd7efe103579bab1ec6b05aa6cf3d469f9a08411/spec/spec_helper.rb#L57) before each and every test. When a test is at its first line, the only objects existing in it are the ones that have been created as a consequence of contextual `let!` and `before do`.
> 
> I think the user needs to be banned for the case you'll be testing though, so you can put the test within the `"when user is banned"` context.

So I am running `git cherry-pick 27d121623f3ddacf94617d8dce54dc680104e33a `but am getting this error:
`fatal: bad object 27d121623f3ddacf94617d8dce54dc680104e33a` I've done some googling but am not sure what to do.

smclairecarden:
@ceithir I figured out the cherry-pick situation, and tests are passing! Let me know what else you think this needs.

app/models/user.rb
```diff
@@ -470,6 +470,11 @@ def coauthored_works
     return @coauthored_works
   end
 
+  #  Returns array of works where the user is the sole author
```
brianjaustin, line 473: I think this should be
```suggestion
  #  Returns array of collections where the user is the sole author
```
(@sarken please correct me if I got the wrong spot!)

spec/controllers/admin/admin_users_controller_spec.rb
```diff
@@ -439,38 +439,57 @@ def expect_changes_made_by(admin, log_items)
...
-    let(:user) { create(:user, banned: true) }
+    let(:user) { create(:user) }
+    let!(:work) { create(:work, authors: [user.default_pseud]) }
+    let(:other_owner) { create(:user, banned: false) }
+    let!(:collection1) { create(:collection) }
+    let!(:collection2) { create(:collection) }
+
+    before do
+      # Banning user only after creating works for them
+      user.update(banned: true)
```
brianjaustin, line 450: We don't have the cop enabled for it right now, but particularly in tests it's better to use `update!` to avoid silent failures: https://github.com/rubocop/rubocop-rspec/issues/911
```suggestion
      user.update!(banned: true)
```

spec/controllers/admin/admin_users_controller_spec.rb
```diff
@@ -439,38 +439,57 @@ def expect_changes_made_by(admin, log_items)
...
-        post :confirm_delete_user_creations, params: { id: user.login }
+        post :destroy_user_creations, params: { id: user.login }
 
         it_redirects_to_with_error(root_url, "Sorry, only an authorized admin can access the page you were trying to reach.")
       end
     end
 
     context "when admin has correct authorization" do
+      before do
+        admin.update(roles: ["policy_and_abuse"])
```
brianjaustin, line 470:
```suggestion
        admin.update!(roles: ["policy_and_abuse"])
```

brianjaustin (changes requested):

app/models/user.rb
smclairecarden: Ah geez. No you're exactly right. Thanks, addressed.

spec/controllers/admin/admin_users_controller_spec.rb
smclairecarden: Updated.

spec/controllers/admin/admin_users_controller_spec.rb
smclairecarden: Updated, thanks!

sarken (approved):
(Brian didn't have time to look tonight, but all their comments appear to have been address, so LGTM!)

PR #4720, opened 2024-01-20
AO3-6598 Slow query in challenge_signups#update and challenge_signups#create

ceithir:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6598

## Purpose

Reduce number of lengthy DB requests when creating/updating a challenge
Went for the ~~lazy~~ minimalistic approach there, with behavior basically untouched: a request that was systemically done is just done earlier, and now triggers an explicit early return

## Testing Instructions

Try answering a challenge with a prompt having no tag, only character tags, a mix of different kind of tags, etc.

## Credit

Ceithir (he/him)


app/models/prompt.rb
```diff
@@ -120,18 +120,23 @@ def allowed_tags
...
               tag_label: tag_type_label_name(tag_type).downcase,
               prompt_type: self.class.name.downcase,
               taglist: disallowed_taglist.collect(&:name).join(ArchiveConfig.DELIMITER_FOR_OUTPUT)))
           end
         else
-          noncanonical_taglist = tag_set.send("#{tag_type}_taglist").reject {|t| t.canonical}
+          noncanonical_taglist = taglist.reject {|t| t.canonical}
           unless noncanonical_taglist.empty?
             errors.add(:base, ts("^These %{tag_label} tags in your %{prompt_type} are not canonical and cannot be used in this challenge: %{taglist}. To fix this, please ask your challenge moderator to set up a tag set for the challenge. New tags can be added to the tag set manually by the moderator or through open nominations.",
               tag_label: tag_type_label_name(tag_type).downcase,
```
ceithir, line 142: From how I understand it, if the challenge has no restriction at all, we don't check for canonicity, but if it has any sort of restriction, we make sure all tags are (at least) canonical? Is that intended behavior?

spec/controllers/prompts_controller_spec.rb
```diff
@@ -131,6 +131,74 @@
...
+          edit_collection_signup_path(open_signup.collection, open_signup),
+          "That prompt would make your overall sign-up invalid, sorry."
+        )
+        expect(assigns[:prompt].errors[:base]).to eq(
+          ["^These character tags in your offer are not canonical and cannot be used in this challenge: Sakura Typomoto. To fix this, please ask your challenge moderator to set up a tag set for the challenge. New tags can be added to the tag set manually by the moderator or through open nominations."]
+        )
+      end
+
+      it "shouldn't make needless checks on unused kind of tags" do
+        expect_any_instance_of(PromptRestriction).to receive(:has_tags?).with("character").at_least(:once)
```
ceithir, line 182: Not sure where the second call to this method happens actually, might be worth investigating.

app/models/prompt.rb
sarken: My knowledge of the challenge code is extremely rusty, but [the Tag Options part of the Gift Exchange FAQ](https://archiveofourown.org/faq/tutorial-running-a-gift-exchange-on-ao3?language_id=en#setgiftxtagopt) might help explain the intended/current behavior. The tl;dr is tags shouldn't need to be canonical if they're in one of the challenge's tag sets.

spec/controllers/prompts_controller_spec.rb
ceithir: Well, simply because the validations are called a first time through [valid?](https://github.com/otwcode/otwarchive/blob/4e56a347944a6fb831414d9cccdfc866ddfacce5/app/controllers/prompts_controller.rb#L133) then a second time on [save](https://github.com/otwcode/otwarchive/blob/4e56a347944a6fb831414d9cccdfc866ddfacce5/app/controllers/prompts_controller.rb#L136).

Let's try to see if we can run them all only once without breaking everything.

Bilka2: I was also looking at this. My in-progress comment:

> The validation is called twice in `prompts_controller#create`, once by [@challenge_signup.valid?](https://github.com/otwcode/otwarchive/blob/4e56a347944a6fb831414d9cccdfc866ddfacce5/app/controllers/prompts_controller.rb#L133) and once by [@prompt.save](https://github.com/otwcode/otwarchive/blob/4e56a347944a6fb831414d9cccdfc866ddfacce5/app/controllers/prompts_controller.rb#L136).

> This happens because challenge_signup uses `accepts_nested_attributes_for :offers, :prompts, :requests`.  [accepts_nested_attributes_for](https://api.rubyonrails.org/v6.1.7.4/classes/ActiveRecord/NestedAttributes/ClassMethods.html) means the associated record is autosaved and thus validated if the parent is validated: https://api.rubyonrails.org/v6.1.7.4/classes/ActiveRecord/AutosaveAssociation.html.

> So, [prompts_controller#create](https://github.com/otwcode/otwarchive/blob/4e56a347944a6fb831414d9cccdfc866ddfacce5/app/controllers/prompts_controller.rb#L136) can use `@prompt.save(validate: false)` instead.

Your solution of combining it also looks okay, but I'm a bit hesitant to call save on the signup when it's the prompt that is getting created.

Bilka2 (approved):
In the validation the only functional change is the `next if taglist.empty?` line, which does all the skipping mentioned on Jira.

Combined with no longer calling the validation twice, I think this is a large performance improvement already and we don't need to look at adding indices for now.

PR #4726, opened 2024-01-27
AO3-6640 Remove config overrides from Rails 5.1

brianjaustin:
I did a stupid and it resulted in #4673 perma-closing...

## Issue

https://otwarchive.atlassian.net/browse/AO3-6640

## Purpose

Removes the previous upgrade's (5.1 -> 6) configuration override file. Of the overrides, one is not even a valid configuration option anymore, two are better for security if flipped to the new defaults, one is set in `application.rb`. The only one that may be tricky is `to_time_preserves_timezone` although it looks like we only use `to_time` on internal timestamps, which should already be UTC.

## Credit

Brian Austin (they/he)


Bilka2 (approved):

zz9pzza:
Checked red had approved the other and looks fine to me.

zz9pzza (approved):

PR #4727, opened 2024-01-27
AO3-6674 Upgrade gems and config to Rails 7.0

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6674


app/controllers/external_works_controller.rb
```diff
@@ -10,13 +10,13 @@ def new
...
-   end
-   respond_to do |format|
-    format.json { render 'fetch.js.erb' }
-   end
+    if params[:external_work_url]
+      url = Addressable::URI.heuristic_parse(params[:external_work_url]).to_str
+      @external_work = ExternalWork.where(url: url).first
+    end
+    respond_to do |format|
+      format.js
```
brianjaustin, line 18: Something about the way Rails resolves templates has changed in version 7, and this was easier than fighting it. The client seems not to care, and the file is JS anyways...

app/helpers/skin_cache_helper.rb
```diff
@@ -1,6 +1,6 @@
 module SkinCacheHelper
   def cache_timestamp
-    Time.now.utc.to_s(:usec)
+    Time.now.utc.to_fs(:usec)
```
brianjaustin, line 3: I saw a deprecation warning for this during Cucumber tests, so I figured I'd change it to avoid pain later. (Similar change below.)

config/application.rb
```diff
@@ -128,5 +128,7 @@ class Application < Rails::Application
                                                   authentication: ArchiveConfig.SMTP_AUTHENTICATION
                                                 })
     end
+
+    config.active_support.disable_to_s_conversion = true
```
brianjaustin, line 132: New Rails 7 parameter, but per the docs it needs to be put here and not in the overrides file

config/environments/production.rb
```diff
@@ -82,23 +73,17 @@
   # the I18n.default_locale when a translation cannot be found).
   config.i18n.fallbacks = true
 
-  # Send deprecation notices to registered listeners.
-  config.active_support.deprecation = :notify
+  # Don't log any deprecations.
+  config.active_support.report_deprecations = false
```
brianjaustin, line 77: They seem to have change their minds since 6.1 :see_no_evil: I don't mind though since we aren't usually scrutinising the prod logs

config/initializers/assets.rb
```diff
@@ -1,7 +1,7 @@
 # Be sure to restart your server when you modify this file.
 
 # Version of your assets, change this if you want to expire all your assets.
-Rails.application.config.assets.version = '1.0'
+# Rails.application.config.assets.version = "1.0"
```
brianjaustin, line 4: We don't use the assets pipeline, and now that Spring (not the Java one) isn't included by default, this line causes errors on boot

config/initializers/new_framework_defaults_7_0.rb
```diff
@@ -0,0 +1,110 @@
...
+# Raise when running tests if fixtures contained foreign key violations
+Rails.application.config.active_record.verify_foreign_keys_for_fixtures = true
+
+# Disable partial inserts.
+# This default means that all columns will be referenced in INSERT queries
+# regardless of whether they have a default or not.
+Rails.application.config.active_record.partial_inserts = false
+
+# Protect from open redirect attacks in `redirect_back_or_to` and `redirect_to`.
+# Rails.application.config.action_controller.raise_on_open_redirects = true
```
brianjaustin, line 63: This breaks several Cucumber tests, so I'd rather flip it in another PR to keep the diff more contained...

config/initializers/new_framework_defaults_7_0.rb
```diff
@@ -0,0 +1,110 @@
+# Be sure to restart your server when you modify this file.
```
brianjaustin, line 1: I've turned on things that are relatively safe already, others left for future PRs

features/other_a/languages.feature
```diff
@@ -1,7 +1,7 @@
 @works
 Feature: Languages
     
-  Scenario: Browse works by language
+  # Scenario: Browse works by language
```
brianjaustin, line 4: Cucumber now fails on undefined steps (scenario, when, etc.) I think that's the right call, but it means that some TODO tests broke the pipeline

features/support/formatter.rb
```diff
@@ -1,13 +1,14 @@
 # Adapted from https://github.com/tpope/fivemat
 require "cucumber/formatter/progress"
+require_relative "./elapsed_time"
 
 module Ao3Cucumber
```
brianjaustin, line 5: Our Cucumber version was... quite old. Several internals have changed, so I had to update this to get the same information in a slightly different way

lib/tasks/after_tasks.rake
```diff
@@ -223,44 +223,6 @@ namespace :After do
     end
   end
 
-  desc "Convert remaining chapter kudos into work kudos"
```
brianjaustin, line 226: The tests for this broke, and rather than investigate I thought removing would be fine since I recently ran this

brianjaustin:
@zz9pzza do you want to do the Ruby 3.1 PR first? If not, I'll grab the commit from that removing the mail starttls patch

app/helpers/application_helper.rb
```diff
@@ -343,12 +343,27 @@ def name_to_id(name)
...
+  def field_id(form_or_object_name, attribute, index: nil, **_kwargs)
+    name_to_id(field_name(form_or_object_name, attribute, index: index))
   end
 
-  def field_name(form, attribute)
-    "#{form.object_name}[#{field_attribute(attribute)}]"
+  # This is a partial re-implementation of ActionView::Helpers::FormTagHelper#field_name.
+  # The method contract changed in Rails 7.0, but we can't use the default because it sometimes
+  # includes other information that -- at a minimum -- wrecks havoc on the Cucumber feature tests.
+  # It is used in when constructing forms, like in app/views/tags/new.html.erb.
```
sarken, line 353: I think I'm missing something... where in that file is it used?

config/initializers/cookie_rotator.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+# As part of the Rails 7 upgrade, we need to convert legacy (SHA1) cookies to SHA256.
+# This can be removed after it has been in production for a little bit.
```
sarken, line 4: Can you make an issue for doing this so we don't forget?

features/other_a/languages.feature
sarken: ~What are the odds we are ever going to do these and wouldn't be better off removing them entirely.~

...who said that?

spec/controllers/external_works_controller_spec.rb
```diff
@@ -12,11 +12,6 @@
     context "when the URL has an external work" do
       let!(:external_work) { create(:external_work, url: url) }
 
-      it "responds with json" do
```
sarken, line 15: Should we be updating these tests to look for `format: :js` instead now, given that's what the format changed to in the controller?

config/environments/development.rb
```diff
@@ -22,6 +25,9 @@
     memcached_servers = YAML.load_file(Rails.root.join("config/local.yml")).fetch("MEMCACHED_SERVERS", memcached_servers) if File.file?(Rails.root.join("config/local.yml"))
     config.cache_store = :mem_cache_store, memcached_servers,
                          { namespace: "ao3-v2-dev", compress: true, pool_size: 10 }
+    config.public_file_server.headers = {
```
sarken, line 28: Google isn't being my friend on this -- is this a default?

config/initializers/new_framework_defaults_7_0.rb
```diff
@@ -0,0 +1,110 @@
...
+
+# Cookie serializer: 2 options
+#
+# If you're upgrading and haven't set `cookies_serializer` previously, your cookie serializer
+# is `:marshal`. The default for new apps is `:json`.
+#
+# Rails.application.config.action_dispatch.cookies_serializer = :json
+#
+#
+# NOTE: as of https://github.com/otwcode/otwarchive/pull/4651, we are using :hybrid, which is the first step to migate.
```
sarken, line 101: Typo: "migrate" is missing its r

app/helpers/application_helper.rb
brianjaustin: It's not very direct, but calls like https://github.com/otwcode/otwarchive/blob/68dff8e5164b3b14b3b4b574a63b379d44ab2f57/app/views/tags/new.html.erb#L17 eventually make their way to this method

config/initializers/cookie_rotator.rb
brianjaustin: Done! 
* https://otwarchive.atlassian.net/browse/AO3-6687
* https://otwarchive.atlassian.net/browse/AO3-6686 (for the config options because we'll want to clean those up too)

config/environments/development.rb
brianjaustin: It is: https://github.com/railsdiff/rails-new-output/blob/cb6a93be67c30040485603cf5b51c7dae4144251/config/environments/development.rb#L27. (That repository powers https://railsdiff.org/ and both are very helpful for comparing config changes between different Rails versions.)

app/controllers/challenge_assignments_controller.rb
```diff
@@ -33,7 +33,7 @@ def load_assignment_from_id
   def owner_only
     return if current_user == @challenge_assignment.offering_pseud.user
 
-    flash[:error] = t("challenge_assignments.not_owner", default: "You aren't the owner of that assignment.")
+    flash[:error] = t(".validations.not_owner", default: "You aren't the owner of that assignment.")
```
Bilka2, line 36: For this and the other changed translations, the `default` argument should be removed because the translation is now in the locale file. Also, it would be nice to remove the keys from `ignore_missing` in `config/i18n-tasks.yml`.

lib/tasks/after_tasks.rake
Bilka2: Rails 7.0 removes `errors.keys`, it is replaced by `errors.attribute_names`.  If that's the only problem with this test then fixing it should be rather simple.

(Currently the test warns about this: `DEPRECATION WARNING: ActiveModel::Errors#keys is deprecated and will be removed in Rails 7.0.`)

brianjaustin: That does fix the issue, thanks! However, I still think we should remove this to avoid confusion, since it's really a 1-off task that won't need to be run again now

Bilka2: Usually I think it would be better to keep it, for the following usecase: Somebody else who is using the archive software might still need to run the task, and there would be no technical reason to force them to do so before upgrading to Rails 7.

But chapter kudos haven't been a thing since before 2011, so that does not apply here.

app/helpers/application_helper.rb
```diff
@@ -343,12 +343,27 @@ def name_to_id(name)
...
-    name_to_id(field_name(form, attribute))
+  def field_id(form_or_object_name, attribute, index: nil, **_kwargs)
+    name_to_id(field_name(form_or_object_name, attribute, index: index))
   end
 
-  def field_name(form, attribute)
-    "#{form.object_name}[#{field_attribute(attribute)}]"
+  # This is a partial re-implementation of ActionView::Helpers::FormTagHelper#field_name.
+  # The method contract changed in Rails 7.0, but we can't use the default because it sometimes
+  # includes other information that -- at a minimum -- wrecks havoc on the Cucumber feature tests.
```
sarken, line 352: "wreaks havoc" rather than "wrecks havoc," I believe? (I had to look it up because I knew it _sounded_ like "reek," but wasn't spelled that way. 😆)

sarken (approved):
Just a typo left, I think! Thank you so much for working on this.

app/helpers/application_helper.rb
brianjaustin: Thanks! I have no idea why I did that, I know it's not spelled that way :facepalm:

PR #4730, opened 2024-01-31
AO3-6637 Avoid recording prerendered chapter in history

brianjaustin:
Ideally the test would be in Cucumber, but Selenium with Chrome doesn't seem to do speculative loading (either yet or as how it's currently configured).

## Issue

https://otwarchive.atlassian.net/browse/AO3-6637

## Purpose

Avoid recording a reading history for when a browser opens a multi-chapter work in Chrome. Otherwise, with prefetch/prerender, we record 2 entries (1 for the actual view, 1 for the speculative loading).

## References

Follow up to #4724 based on discussion elsewhere

## Credit

Brian Austin (they/he)
Domenic Denicola (he/him)

### DIFF
--- a/app/controllers/chapters_controller.rb
+++ b/app/controllers/chapters_controller.rb
@@ -69,7 +69,7 @@ class ChaptersController < ApplicationController
                         current_user.subscriptions.build(subscribable: @work)
       end
       # update the history.
-      Reading.update_or_create(@work, current_user) if current_user
+      Reading.update_or_create(@work, current_user) if current_user && !speculative_loading?
 
       respond_to do |format|
         format.html
@@ -285,5 +285,12 @@ class ChaptersController < ApplicationController
                                     :notes, :endnotes, :content, :published_at,
                                     author_attributes: [:byline, ids: [], coauthors: []])
 
+  end
+
+  def speculative_loading?
+    sec_purpose = request.headers["HTTP_SPEC_PURPOSE"]
+    return false unless sec_purpose
+
+    sec_purpose.include?("prefetch") || sec_purpose.include?("prerender")
   end
 end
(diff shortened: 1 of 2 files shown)


app/controllers/chapters_controller.rb
```diff
@@ -286,4 +286,11 @@ def chapter_params
                                     author_attributes: [:byline, ids: [], coauthors: []])
 
   end
+
+  def speculative_loading?
+    sec_purpose = request.headers["HTTP_SPEC_PURPOSE"]
```
domenic, line 291: SPEC_PURPOSE -> SEC_PURPOSE, maybe?

domenic (commented):
This will prevent recording a chapter that the user doesn't visit. But it will also prevent recording a chapter that the user does eventually visit.

To fix that part, we'd need an additional endpoint which is triggered by JavaScript, probably around here: https://github.com/otwcode/otwarchive/blob/2484b9fbc094b6cae6e54c347b7eb6bd480c63af/app/views/hit_count/_include.html.erb#L9-L15

spec/controllers/chapters_controller_spec.rb
```diff
@@ -114,6 +114,22 @@
...
+      context "when viewing as a logged-in user" do
+        before do
+          fake_login
+        end
+
+        context "when speculative loading is enabled" do
+          it "does not record history" do
+            create(:chapter, work: work)
+            chapter1 = work.chapters.first
+            @request.headers["Spec-Purpose"] = "prerender"
```
domenic, line 127:
```suggestion
            @request.headers["Sec-Purpose"] = "prefetch;prerender"
```

app/controllers/chapters_controller.rb
brianjaustin: :facepalm: I kept messing this up, and I copied the wrong one... Thanks, fixed!

sarken:
Thanks for looking into this! I've checked with Support and this isn't causing issues on their side of things, plus they were (like me) mostly unpleasantly surprised to find out visit counting even works like this. So we're going to leave it as-is for now and maybe change it to be more like hit counting (or something) in the future.

PR #4740, opened 2024-02-13
AO3-6297: Stop using MimeMagic in downloads code

gmjimenezc:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6297

## Purpose

Replaces MimeMagic with Marcel

## Testing Instructions
Log in

Post > New Work (we'll want a fresh work so we're sure we're not getting cached files)

Fill in required information

Post

Expand the "Downloads" menu on the work

Verify you can download and open each type of file

## References

## Credit

warlockmel (she/her)


gmjimenezc:
@redsummernight @tickinginstant hey guys! a problem here. I think marcel doesn't have a way to identify solely the extension from a mime type, but i can't be sure because the error logs are not really clear, and the documentation is also very limited. Is there a way for me to access the screenshots artifact, or at least the actual log? Or to debug the issue? My current installation is using Docker, and I need to see the actual output of variables to check exactly where is the issue.

app/models/download.rb
```diff
@@ -40,16 +40,17 @@ def set_file_type(options)
 
   # Given a mime type, return a file extension
   def file_type_from_mime(mime)
-    ext = MimeMagic.new(mime.to_s).subtype
-    case ext
+    subtype = Marcel::MimeType.for(mime).to_s
```
brianjaustin, line 43: It looks like `Marcel::MimeType`'s `for` structures its arguments differently than `MimeMagic#new`: https://github.com/rails/marcel/blob/12fc8daae656ccb09441c4a3c376b5de3af05172/lib/marcel/mime_type.rb#L29

You can try the `Marcel::Magic` class instead:
```suggestion
    subtype = Marcel::Magic.new(mime.to_s).subtype
```

app/models/download.rb
```diff
@@ -6,7 +6,7 @@ def initialize(work, options = {})
     @file_type = set_file_type(options.slice(:mime_type, :format))
     # TODO: Our current version of the mime-types gem doesn't include azw3, but
     # the gem cannot be updated without updating rest-client
```
brianjaustin, line 8: This comment isn't true for Marcel, so we can delete it
```suggestion
```

app/models/download.rb
```diff
@@ -6,7 +6,7 @@ def initialize(work, options = {})
...
-    @mime_type = @file_type == "azw3" ? "application/x-mobi8-ebook" : MIME::Types.type_for(@file_type).first
+    @mime_type = @file_type == "azw3" ? "application/x-mobi8-ebook" : Marcel::MimeType.for(@file_type).to_s
```
brianjaustin, line 9: Following from above, this can be simplified to
```suggestion
    @mime_type = Marcel::MimeType.for(extension: @file_type).to_s
```

config/initializers/mime_types.rb
```diff
@@ -4,10 +4,12 @@
 # Mime::Type.register "text/richtext", :rtf
 
 # for azw3 files
-Mime::Type.register 'application/x-mobi8-ebook', :azw3
+#Mime::Type.extend 'application/x-mobi8-ebook', :azw3
```
brianjaustin, line 7: If needed, we can get the old version of this file from git, so no need to keep the old lines commented out

features/step_definitions/work_download_steps.rb
```diff
@@ -11,7 +11,7 @@
 end
 
 Then /^I should receive a file of type "(.*?)"$/ do |filetype|
-  mime_type = filetype == "azw3" ? "application/x-mobi8-ebook" : MIME::Types.type_for("foo.#{filetype}").first
+  mime_type = filetype == "azw3" ? "application/x-mobi8-ebook" : Marcel::MimeType.for("foo.#{filetype}").to_s
```
brianjaustin, line 14: I _think_ these can be simplified too

brianjaustin (changes requested):

features/step_definitions/work_download_steps.rb
gmjimenezc: This one I'm not sure, I had to use extend method (for registering the type) since it's not natively on the extensive list of Marcel. Buuut I'll check anyway, thanks for the help! I'll change it tonight

gmjimenezc:
@brianjaustin i give up, I checked the syntaxis in the marcel repo (tests and examples), and the rubydocs, and it's still failing with `"expected nil to match /filename="file_name.extension"/ (RSpec::Expectations::ExpectationNotMetError)"`. Please let me know if you know where is the error, why in `expect(page.response_headers['Content-Disposition']).to match(/filename="#{filename}"/)` the 1st argument which is the page.response_headers is returning a `nil`

app/models/download.rb
```diff
@@ -4,9 +4,7 @@ class Download
   def initialize(work, options = {})
     @work = work
     @file_type = set_file_type(options.slice(:mime_type, :format))
-    # TODO: Our current version of the mime-types gem doesn't include azw3, but
-    # the gem cannot be updated without updating rest-client
-    @mime_type = @file_type == "azw3" ? "application/x-mobi8-ebook" : MIME::Types.type_for(@file_type).first
+    @mime_type =  @file_type == "azw3" ? "application/x-mobi8-ebook" : Marcel::MimeType.for(extension: @file_type).to_s
```
brianjaustin, line 7: We can sidestep the Rubocop by just removing the first bit (I ran one of the Cucumber tests to make sure and it still works)
```suggestion
    @mime_type = Marcel::MimeType.for(extension: @file_type).to_s
```

config/initializers/mime_types.rb
```diff
@@ -1,13 +1,5 @@
-# Be sure to restart your server when you modify this file.
-
-# Add new mime types for use in respond_to blocks:
-# Mime::Type.register "text/richtext", :rtf
-
 # for azw3 files
-Mime::Type.register 'application/x-mobi8-ebook', :azw3
```
brianjaustin, line 7: Ok, so this is where things get a bit confusing. Even though the calls here look the same as in the files above, the casing difference is vital. The former, `MIME`, is what we want to remove. The one here actually belongs to Rails's internal plumbing to figure out how to label request types properly: https://github.com/rails/rails/blob/44db6bbfcbe66ac5926f01f3b693a378552b752f/actionpack/lib/action_dispatch/http/mime_type.rb#L183

What was happening was:
* Rails gets a request for extension `mobi` and doesn't know what it is, so labels it with the null mimetype as per https://github.com/rails/rails/blob/44db6bbfcbe66ac5926f01f3b693a378552b752f/actionpack/lib/action_dispatch/http/mime_negotiation.rb#L64
* The respond_to block in https://github.com/otwcode/otwarchive/blob/34bfb3c642a1b304412315df837050e397b9f97c/app/controllers/downloads_controller.rb#L10 doesn't have something for the null type, so the controller action bails with a 404 HTML page instead of the download

From my testing in Docker, adding these back in (but probably keeping the `Marcel` ones too) should fix the tests. (On a related note, if you want to run the tests in docker locally you'll need to add `- QTWEBENGINE_CHROMIUM_FLAGS=--no-sandbox` to https://github.com/otwcode/otwarchive/blob/34bfb3c642a1b304412315df837050e397b9f97c/docker-compose.yml#L99)

brianjaustin (commented):
Looks like the issue was multi-part, and I only caught the first one :facepalm: Sorry for the resulting geese chase! I've added some more suggestions inline

config/initializers/mime_types.rb
gmjimenezc: Thanks to you i'm finalllyyyy finishing with this hahaha, I've been unable to test locally due to an issue with the db, which apparently doesn't find the conf file in the path when running `docker-compose run --rm test bundle exec cucumber features`, but the normal `docker-compose up` does work, so idk, I'm working on fixing that to be able to test locally instead of doing multiple tries in the PRs

features/step_definitions/work_download_steps.rb
```diff
@@ -11,7 +11,7 @@
 end
 
 Then /^I should receive a file of type "(.*?)"$/ do |filetype|
-  mime_type = filetype == "azw3" ? "application/x-mobi8-ebook" : MIME::Types.type_for("foo.#{filetype}").first
+  mime_type = filetype == "azw3" ? "application/x-mobi8-ebook" : Marcel::MimeType.for(name: "foo.#{filetype}").to_s
```
ceithir, line 14: Unless I'm missing some CI shenanigans (i.e. I've only tried running the tests locally), we don't seem to be needing the exception on azw3 anymore. Maybe a happy consequence of the changes to the initializer?

ceithir:
Semi-related: Wondering if we shouldn't add the `- QTWEBENGINE_CHROMIUM_FLAGS=--no-sandbox` line to the Docker file, or mention this trick somewhere in the wiki, as it's [the second time](https://github.com/otwcode/otwarchive/pull/4614) we have need of it.

features/step_definitions/work_download_steps.rb
gmjimenezc: Eh, I'm not sure. x-mobi8-ebook is not by default in the list that comes with Marcel so I had to use the extend() method to add the custom type, I can certainly try removing it tho, so I'll modify that this week

ceithir: I meant we don't seem to need the `filetype == "azw3" ? "application/x-mobi8-ebook"` part in step definitions anymore (the tests run fine without it, at least locally). The initializer is good as is, and adding this non-standard format by hand indeed necessary. Sorry for the confusion.

gmjimenezc: Done

brianjaustin (approved):
@gmjimenezc if you have time, I think it would be nice to add the option to `docker-compose.yml` like @ceithir mentioned, but that's more of a nice to have than required for this PR

brianjaustin:
Hi @gmjimenezc ! We're planning to include this in the next release ([0.9.367](https://otwarchive.atlassian.net/projects/AO3/versions/13088/tab/release-report-all-issues)), but it looks like there's a small merge conflict with one of the other PRs that was merged. We can take care of that, but if you'd prefer to take a look please let us know within the net 48 hours or so. Thanks again for the PR!

PR #4743, opened 2024-02-18
AO3-5803 Make errors for challenge limits display inline

brianjaustin:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5803

## Purpose

When errors are displayed for fields like required requests per sign-up, the fields should be highlighted in red but not break the flow of the form. This fixes the issue where error highlighting takes up the whole line (refer to https://otwarchive.atlassian.net/browse/AO3-5803?focusedCommentId=360957).

New look:
Challenge "Requests and Offers" section with number of requests and offers highlighted in red. The number required fields are also highlighted, but do not cause the allowed boxes to go to the next line.

**NOTE:** This will require recaching skins to show up properly.

## Credit

Brian Austin (they/he)

### DIFF
--- a/app/helpers/prompt_restrictions_helper.rb
+++ b/app/helpers/prompt_restrictions_helper.rb
@@ -58,7 +58,7 @@ module PromptRestrictionsHelper
         h(ts("Must Be Unique?")) + form.check_box("require_unique_#{tag_type}".to_sym, disabled: (hasprompts ? false : true))
       end
     end
-    content_tag(:dd, fields.html_safe, title: ts("#{tag_type_label_name(tag_type).pluralize.downcase}")) + "\n".html_safe
+    content_tag(:dd, fields.html_safe, class: "inline-error", title: ts("#{tag_type_label_name(tag_type).pluralize.downcase}")) + "\n".html_safe
   end
 
   # generate the string to use for the labels on sign-up forms
--- a/app/helpers/prompt_restrictions_helper.rb
+++ b/app/helpers/prompt_restrictions_helper.rb
@@ -58,7 +58,7 @@ module PromptRestrictionsHelper
         h(ts("Must Be Unique?")) + form.check_box("require_unique_#{tag_type}".to_sym, disabled: (hasprompts ? false : true))
       end
     end
-    content_tag(:dd, fields.html_safe, class: "inline-error", title: ts("#{tag_type_label_name(tag_type).pluralize.downcase}")) + "\n".html_safe
+    content_tag(:dd, fields.html_safe, class: "inline-error", title: ts(tag_type_label_name(tag_type).pluralize.downcase.to_s)) + "\n".html_safe
   end
 
   # generate the string to use for the labels on sign-up forms
--- a/app/helpers/prompt_restrictions_helper.rb
+++ b/app/helpers/prompt_restrictions_helper.rb
@@ -58,7 +58,7 @@ module PromptRestrictionsHelper
         h(ts("Must Be Unique?")) + form.check_box("require_unique_#{tag_type}".to_sym, disabled: (hasprompts ? false : true))
       end
     end
-    content_tag(:dd, fields.html_safe, class: "inline-error", title: ts(tag_type_label_name(tag_type).pluralize.downcase.to_s)) + "\n".html_safe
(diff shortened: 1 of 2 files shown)


app/helpers/prompt_restrictions_helper.rb
```diff
@@ -58,7 +58,7 @@ def required_and_allowed(form, tag_type, hasprompts, allowany)
         h(ts("Must Be Unique?")) + form.check_box("require_unique_#{tag_type}".to_sym, disabled: (hasprompts ? false : true))
       end
     end
-    content_tag(:dd, fields.html_safe, title: ts("#{tag_type_label_name(tag_type).pluralize.downcase}")) + "\n".html_safe
+    content_tag(:dd, fields.html_safe, class: "inline-error", title: ts(tag_type_label_name(tag_type).pluralize.downcase.to_s)) + "\n".html_safe
```
sarken, line 61: So this issue we're trying to address is actually a problem that’s not unique to these fields.

You can see if it you leave a checkbox unchecked when creating an account (normally, the text and the checkbox would be side-by-side):

Errors on the account creation page. The password and email fields display correctly, but the age confirmation checkbox has the checkbox and its label separated.

Or when entering an email address to join the queue (again, normally the label and the field would be side-by-side):

Error on the invitation request page. The field for entering your email and the label for it are on separate lines.

You can get a significant chunk of the way to fixing it by changing `span.error { display: block; }` to `dt span.error, dd span.error`, and we should do that if we're going to try to address this problem. 

But it doesn’t fix the problem _here_, because we're inside a `dd` and we have a whole bunch of poorly marked-up fields crammed into it. Ideally, we'd rewrite these two methods to produce better HTML and adjust the form to be more accessible and consistent (fun fact: if you click the "Details/Description" label, it controls the "Allowed" field, but the "URL" label is tied to "Required" instead), and in doing so, we'd probably fix the problem.

That's going to be a bit of a project, though. For an interim fix, you're probably on the right track with giving the `dd` a class in this specific case, but `inline-error` isn't a good choice. First, it's against our class naming system (single words, no hyphens) and second, we should avoid the word `error` because it displays even when there's an error. How about something like `complex`, since it's a very complex `form dd`? 

(Will we ever see error spans in the `dd` produced by `required_and_allowed(form, tag_type, hasprompts, allowany)`? If so, we might want the class there as well.)

Then we should be able to change the existing `span.error` style to this, fixing the issue in the majority of places:

```
dt span.error, dd span.error { 
  display: block;
}

dd.complex span.error {
  display: inline-block;
}
```

sarken: I forgot to mention -- we can't do without `dd span.error ( display: block; }` because then our text inputs shrink and no longer take up the full width of the `dd` when there are errors. 😩

brianjaustin: Thanks! (This is exactly why I pinged you for review... I too often fall into the trap of "oh this CSS should be pretty simple, ezpz" :sob: ) I've pushed those changes, although I'm not sure about that method you mentioned since it's where this PR already adds the `complex` class? https://github.com/otwcode/otwarchive/blob/656587fedfc31442a26b1525bb21c9e99986336a/app/helpers/prompt_restrictions_helper.rb#L61

I'm also fine deferring this until the actual structure is fixed, especially if it becomes too distracting from the Calibre fonts issue

sarken: Sorry, left off a word! `required_and_allowed_boolean`, the method above this one.

brianjaustin: Ah ok, that's what I thought but wasn't sure. It looks like that doesn't run into the same situation because any combination (both boxes checked, neither checked, one checked) is valid for the 3 cases where this is used. IMO that could do with some improvement (what does it mean to have Title required but not allowed?), so I'd advocate just leave it for now and come back later; maybe radio buttons would be better here instead?

sarken: Leaving it works!

I don't think radio buttons are a good idea here, since you can't really deselect them if you accidentally choose one. Unless we add a third option for "Not allowed" or something...

sarken (approved):

PR #4748, opened 2024-02-23
AO3-6563 Attach HTML download copies of works to abuse tickets

weeklies:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6563

## Purpose

Add a send attachment async job whenever a work is reported.

I'm not familiar with async jobs  and how we generate downloads, so I might need to be pointed in the right direction.
 
## Credit

weeklies (she/her)

### DIFF
--- a/app/jobs/report_attachment_job.rb
+++ b/app/jobs/report_attachment_job.rb
@@ -1,0 +1,6 @@
+class ReportAttachmentJob < ApplicationJob
+  def perform(ticket_id, work)
+    download = Download.new(work, mime_type: "text/html").generate
+    reporter.send_attachment!(ticket_id, download)
+  end
+end
--- a/app/jobs/report_attachment_job.rb
+++ b/app/jobs/report_attachment_job.rb
@@ -1,6 +1,7 @@
 class ReportAttachmentJob < ApplicationJob
   def perform(ticket_id, work)
-    download = Download.new(work, mime_type: "text/html").generate
-    reporter.send_attachment!(ticket_id, download)
+    download = Download.new(work, mime_type: "text/html")
+    html = DownloadWriter.new(download).generate_html
+    reporter.send_attachment!(ticket_id, html)
   end
 end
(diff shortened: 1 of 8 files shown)


app/jobs/report_attachment_job.rb
```diff
@@ -0,0 +1,6 @@
+class ReportAttachmentJob < ApplicationJob
+  def perform(ticket_id, work)
+    download = Download.new(work, mime_type: "text/html").generate
+    reporter.send_attachment!(ticket_id, download)
```
brianjaustin, line 4: I'm not sure this will actually work; `generate` returns the `Download` object itself, not an HTML string or a file. I think you'd need to do something like this to get the generated file:
```suggestion
    download = Download.new(work, mime_type: "text/html")
    html = DownloadWriter.new(download).generate_html
    reporter.send_attachment!(ticket_id, html)
```
or call `.generate` and do `File.open` on the path it creates

spec/models/abuse_report_spec.rb
```diff
@@ -345,4 +345,22 @@
       expect(safe_report.save).to be_truthy
     end
   end
+
+  describe "#attach_work_download" do
+    include ActiveJob::TestHelper
+
+    let(:work) { create(:work) }
```
brianjaustin, line 352: It looks like this wouldn't be created for the first test; is that intended?

brianjaustin (changes requested):

spec/models/abuse_report_spec.rb
weeklies: Yes, it is.

app/jobs/report_attachment_job.rb
```diff
@@ -0,0 +1,7 @@
+class ReportAttachmentJob < ApplicationJob
```
weeklies, line 1: @brianjaustin Would there be performance problems with this and numerous/large downloads? I'm not very familiar with async jobs.

brianjaustin: My initial thought was that we'll be fine, especially since HTML downloads require the least resources to generate. I double checked with the rest of Systems to be sure, and we're good in terms of workers for nwo

brianjaustin (approved):

PR #4750, opened 2024-02-24
AO3-6685 Prefer update! in tests

sarken:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6685

## Purpose

Companion piece to #4749, which updates [admin_posts_controller_spec.rb](https://github.com/otwcode/otwarchive/compare/master...sarken:otwarchive:AO3-6685?expand=1#diff-85b9bca34c47b64b76b55326ca3d9a16fbe030b6ba2543f9246ae4e23ff3bf89) and [comments_controller_spec.rb](https://github.com/otwcode/otwarchive/compare/master...sarken:otwarchive:AO3-6685?expand=1#diff-ea481d5353c6a244aa91858a122ceae02a551eaebb9e8913e6bf653d050cfef4) to use `update!` instead of `update`. (There's a third spec file changed in that pull request, but it doesn't use `update`.)

This one updates all the other tests. The only things that weren't straightforward find and replace were:

1. [a change in the comment model spec to expect an error](https://github.com/otwcode/otwarchive/pull/4750/files#diff-337a407f4c7083b9f286f027c7ca593faa5469bac6a635a29a4c97ecf506830eR94-R95)
2. [a change in the user model spec to expect an error](https://github.com/otwcode/otwarchive/pull/4750/files#diff-bb8bcb6ef2e13ab9491f4f9ec48be47ee025d7f0752a25029d40810af2c65ee1R203-R204)
3. [a second change in the user model spec to fix an invalid email format](https://github.com/otwcode/otwarchive/pull/4750/files#diff-bb8bcb6ef2e13ab9491f4f9ec48be47ee025d7f0752a25029d40810af2c65ee1R212-R213)

## Testing Instructions

No manual testing.

## References

[Discussion](https://github.com/otwcode/otwarchive/pull/4749#discussion_r1500693832)

## Credit

Sarken, she/her

### DIFF
--- a/spec/models/comment_spec.rb
+++ b/spec/models/comment_spec.rb
@@ -72,7 +72,7 @@ describe Comment do
         before { comment.save(validate: false) }
 
         it "changes the comment" do
-          expect(comment.update(comment_content: "Why did you block me?")).to be_truthy
+          expect(comment.update!(comment_content: "Why did you block me?")).to be_truthy
           expect(comment.errors.full_messages).to be_blank
           expect(comment.reload.comment_content).to eq("Why did you block me?")
         end
@@ -91,7 +91,7 @@ describe Comment do
         before { comment.save(validate: false) }
 
         it "doesn't change the comment" do
-          expect(comment.update(comment_content: "Why did you block me?")).to be_falsey
+          expect { comment.update!(comment_content: "Why did you block me?") }.to raise_error(ActiveRecord::RecordInvalid)
           expect(comment.errors.full_messages).to include(message)
           expect(comment.reload.comment_content).to eq("Hmm.")
         end
@@ -356,7 +356,7 @@ describe Comment do
 
         it "does not update last wrangling activity" do
           expect do
-            comment.update(comment_content: Faker::Lorem.sentence(word_count: 25))
+            comment.update!(comment_content: Faker::Lorem.sentence(word_count: 25))
           end.not_to change { tag_wrangler.reload.last_wrangling_activity.updated_at }
         end
       end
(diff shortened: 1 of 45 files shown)


spec/models/comment_spec.rb
```diff
@@ -72,7 +72,7 @@
         before { comment.save(validate: false) }
 
         it "changes the comment" do
-          expect(comment.update(comment_content: "Why did you block me?")).to be_truthy
+          expect(comment.update!(comment_content: "Why did you block me?")).to be_truthy
```
brianjaustin, line 75: Since we're using `update!` instead, we can drop the `expect(...).to be_truthy` (that's the other way to make it obvious when updates fail)

spec/models/comment_spec.rb
```diff
@@ -91,7 +91,8 @@
         before { comment.save(validate: false) }
 
         it "doesn't change the comment" do
-          expect(comment.update(comment_content: "Why did you block me?")).to be_falsey
+          expect { comment.update!(comment_content: "Why did you block me?") }
+            .to raise_error(ActiveRecord::RecordInvalid)
```
brianjaustin, line 95: Sorry, I should have mentioned on the other PR that the `.to be_falsey` pattern is OK to test for things not saving. Although IMO this is a little bit clearer in the test's intent so still a good thing to do

sarken: Yeah, I prefer checking for the relevant error rather than falsiness -- there were two tests where the change to update! revealed we were using invalid data (e.g., random text where something that looked like an email was expected).

brianjaustin (approved):
Thank you! :heart:

PR #4751, opened 2024-02-25
AO3-6676 Remove participant memoization causing Leave button to appear on all collection blurbs

smclairecarden:
# Pull Request Checklist

* [x ] Have you read ["How to write the perfect pull request"](https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/)?
* [x ] Have you read the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [x ] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)?
* [ x] Have you added the [Jira](https://otwarchive.atlassian.net) issue number
* [x ] Do you have fewer than 5 pull requests already open? If not, please wait

## Issue

https://otwarchive.atlassian.net/browse/AO3-6676
## Purpose

What does this PR do?

Makes sure that the text "Leave" is present only on buttons the user is a member of on the collections page when using the "Closed" false and "Moderated" true filters.

## Testing Instructions

See Jira.

## References

## Credit

Claire C (she/her) and alien.

### DIFF
--- a/features/collections/collection_participants.feature
+++ b/features/collections/collection_participants.feature
@@ -68,12 +68,10 @@
   When I am in the default browser
 
 Scenario: Collection member should see correct button text
-Given basic collections
-    And I have a moderated collection "ModeratedCollection"
-    And I have a moderated collection "ModeratedCollectionTheSequel"
-  When I am on the "Collections in the Example Archive" page
-    And I choose "collection_filters_moderated_true"
-    And I choose "collection_filters_closed_false"
-    And I press "Sort and Filter"
+  Given I have the moderated collection "ModeratedCollection"
+    And I have the moderated collection "ModeratedCollectionTheSequel"
+    And I am logged in as "sam"
+    And I have joined the collection "ModeratedCollection" as "sam"
+  When I am on the collections page
   Then I should see "Leave" button exactly 1 time
     And I should see "Join" button 0 or more times
--- a/features/collections/collection_participants.feature
+++ b/features/collections/collection_participants.feature
@@ -68,9 +68,9 @@
   When I am in the default browser
 
 Scenario: Collection member should see correct button text
-Given I have a moderated collection "ModeratedCollection"
+Given I am logged in as "sam"
+    And I have a moderated collection "ModeratedCollection"
     And I have a moderated collection "ModeratedCollectionTheSequel"
-    And I am logged in as "sam"
   When I am on the "Collections in the Example Archive" page
     And I choose "collection_filters_moderated_true"
     And I choose "collection_filters_closed_false"
--- a/features/collections/collection_participants.feature
+++ b/features/collections/collection_participants.feature
@@ -68,7 +68,7 @@
   When I am in the default browser
 
 Scenario: Collection member should see correct button text
-Given I have joined the collection "ModeratedCollection" as "sam"
+Given I have the moderated collection "ModeratedCollection"
     And I have the moderated collection "ModeratedCollectionTheSequel"
     And I am logged in as "sam"
   When I am on the "Collections in the Example Archive" page
--- a/features/collections/collection_participants.feature
+++ b/features/collections/collection_participants.feature
@@ -68,7 +68,7 @@
   When I am in the default browser
 
 Scenario: Collection member should see correct button text
-  Given I have joined the collection "Such a nice collection"
+  Given I have joined the collection "Such a nice collection" as "sam"
     And I am logged in as "sam"
   When I am on the "Collections in the Example Archive" page
     And I choose "collection_filters_moderated_true"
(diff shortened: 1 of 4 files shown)


smclairecarden:
I would appreciate some insight into why my collections are not being created! From what I can tell I am not doing anything different than the other tests in the file. How do I create two collections if not like this?? Thank you to whoever takes a look!

features/collections/collection_participants.feature
```diff
@@ -66,3 +66,14 @@
     And I follow "Join"
   Then I should see "You are now a member of Such a nice collection"
   When I am in the default browser
+
+Scenario: Collection member should see correct button text
+Given I have joined the collection "ModeratedCollection" as "sam"
```
sarken, line 71: I looked at the step definition for this, and it looks like the step doesn't create the collection -- just looks for and joins an existing one. So you'd need to start with `Given I have the collection "ModeratedCollection"` to create the collection.

smclairecarden:
> I noticed you were still having some trouble getting the collections set up, so here's one way you can fix the start of your test:
> 
> ```
>   Scenario: Collection member should see correct button text
>   Given I have the moderated collection "ModeratedCollection"
>     And I have the moderated collection "ModeratedCollectionTheSequel"
>     And I am logged in as "sam"
>     And I have joined the collection "ModeratedCollection" as "sam"
>   When I am on the collections page
>   Then I should see "Leave" button exactly 1 time
>     And I should see "Join" button 0 or more times
> ```
> 
> And here are two options for defining the `And I have joined the collection "ModeratedCollection" as "sam"` step -- either is fine! (I've written it using Cucumber expressions rather than regex, which your previous version used.)
> 
> ```
> Given "I have joined the collection {string} as {string}" do |title, login|
>   # OPTION 1: The really lazy, fast way -- just create the data.
>   #collection = Collection.find_by(title: title)
>   #user = User.find_by(login: login)
>   #FactoryBot.create(:collection_participant, pseud: user.default_pseud, collection: collection, participant_role: "Member")
>   # OPTION 2: Actually visit the collection and press the Join button, which will probably catch more errors, but is a bit slower.
>   step %{I am logged in as "#{login}"}
>   visit collection_path(Collection.find_by(title: title))
>   click_link("Join")
> end
> ```

I am so confused as to why when using the regex I kept getting an "ambiguous match error, but this passes! Thank you so much. I could have articulated this issue sooner.

features/step_definitions/collection_steps.rb
```diff
@@ -203,6 +213,14 @@
   step %{I should see /We couldn't find the collection(?:.+and)? #{collection_name}/}
 end
 
+Then("I should see {string} button exactly {int} time") do |_string, _int|
+  step %{I should see "Leave"}
+end
```
sarken, line 218: Unfortunately, this step isn't doing what you want. It's not using either the `{string}` or `{int}` arguments it's given; it's just looking for the word "Leave" to appear anywhere on the page any number of times, no matter what it string you tell it to look for, or how many times you tell it the string should appear. 

If you want to see what I mean, try changing the feature file to say "Then I should see "Pineapple" button exactly 6 time" and then run the tests. The scenario will still pass, even though "Pineapple" doesn't appear on the collections index six times.

I'd recommend deleting lines 216-223 here and defining a new step in step_definitions/generic_steps.rb: 

```
Then "I should see {string} exactly {int} time(s)" do |string, int|
  expect(page).to have_content(string).exactly(int)
end
```

That will actually look for the specified text to appear the specified number of times. (The guts are basically an RSpec test -- if you search for `expect` in our step definitions directory, you'll find a lot of step definitions where we do that because it's just easier. I didn't know how to do it off the top of my head, either, but some determined Googling turned up [Capybara::RSpecMatchers::CountSugar](https://www.rubydoc.info/gems/capybara/Capybara/RSpecMatchers/CountSugar).)

sarken (changes requested):

features/step_definitions/collection_steps.rb
smclairecarden: You're right thank you. Addressed!

features/step_definitions/collection_steps.rb
```diff
@@ -117,6 +117,16 @@
...
 end
 
+Given "I have joined the collection {string} as {string}" do |title, login|
+  collection = Collection.find_by(title: title)
+  user = User.find_by(login: login)
+  FactoryBot.create(:collection_participant, pseud: user.default_pseud, collection: collection, participant_role: "Member")
+  visit collections_path
+  find("label[for='collection_filters_moderated_true']").click
+  find("label[for='collection_filters_closed_false']").click
+  step %{I press "Sort and Filter"}
```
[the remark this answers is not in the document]
smclairecarden: addressed!

sarken (approved):

PR #4754, opened 2024-03-01
AO3-6689 Improve accessibility of header and navigation

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6689

## Purpose

A user reports that the items in the “Hi, username!” menu are read only as “menu item” instead of the text, e.g. My Dashboard.

The ARIA roles menu and menuitem are not meant to be used for navigation. Instead, they are meant to be used for application-style menus and menu items, like in the Rich Text Editor.

So let’s fix some of the accessibility issues in the header. That means:
* The menu and menuitem roles should be removed.
* The div around the header should now be a <header> element instead.
* This gives the section the banner role, which might seem weird if you’re not used to screen readers, but it's the normal behavior. (Refer to mdn’s [ARIA: banner role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/banner_role) and [header element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/header) documentation for more info.)
* The navigation lists should now be wrapped inside nav elements, with aria-label attributes to distinguish the user nav (the one containing "Hi, name!" and so on) from the site nav and the admin nav.
* The names should not include the word “navigation,” because screen readers will say “navigation” shortly thereafter due to the presence of the nav element. (More [info on screen reader behavior with aria-labels on nav elements](https://russmaxdesign.github.io/accessible-forms/landmarks-nav04.html).) 
* The search form in the site navigation was missing the search role. The role appears to have better support than the search element, so we want to use that, and we should give it an aria-label because it appears on all pages, including those with other search fields (although I don’t think any of them have the search role yet.)

Also, i18n and code style fixes, whee. 

Note: I didn't apply the various `width:` and `float:` styles from the `<ul>` elements to the `<nav>` elements. I _can_, but I don't think it makes any difference.

## Testing Instructions

Log in as a user and an admin and use the navigation with VoiceOver. Make sure to test the skins in the chooser, too.

## References

[Browser compatibility for search element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/search#browser_compatibility) and [mdn's recommendation to use the role instead](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/search_role#prefer_html).

## Credit

Sarken, she/her


app/views/layouts/_header.html.erb
```diff
@@ -1,78 +1,79 @@
...
     <h2 class="collections">
       <% unless @collection.header_image_url.present? %>
-        <%= @collection.title ? @collection.title : (@collection_parent ? @collection_parent.title : ts('New Collection', key: 'header')) %>
+        <%= if @collection.title
+              @collection.title
+            elsif @collection_parent
+              @collection_parent.title
+            else
+              t(".collections.new")
+            end %>
```
brianjaustin, line 69: I much prefer this over the nested ternary, but if you want to keep it a 1-liner:
```suggestion
        <%= @collection.title || @collection_parent&.title ||  t(".collections.new") %>
```

brianjaustin (approved):

PR #4755, opened 2024-03-01
AO3-6518 Spam check runs on abuse reports if the email you enter has different capitalization than the email on your account

forceofcalm:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6518

## Purpose

Previously, the spam checker was flagging the report as spam if the casing of the emails for the account and submission did not match. We fixed this by downcasing both values before comparing.

## Testing Instructions
1. Log in
2. Follow “Policy Questions & Abuse Reports” in the site footer
3. Change the “Your name (optional)” field to say “akismet-guaranteed-spam,” which will guarantee your report is flagged as spam if the spam checker runs on it
4. Change the capitalization of the email in the “Your email (required)” field
5. Fill in rest of form
6. Submit

Form should submit successfully.

## Credit

calm (they/them)


sarken (changes requested):
The code change looks good! But we could use a new example in the abuse report spec to cover this. We currently only have [a test that covers emails that are an exact match](https://github.com/otwcode/otwarchive/blob/9858bcfd9d6e38aba49607a5279103b92737ca6e/spec/models/abuse_report_spec.rb#L343-L346).

Also, just a reminder to make sure you transition the Jira issue to In Review when you open a pull request. (If you set the issue to In Progress when you start work, the change should happen automatically.) If Jira won't let you do that, please let me know and I'll see if something's wrong with your account permissions.

Thanks!

Bilka2 (changes requested):
As pointed out in the other PR where you accidentally added this test, the test wasn't failing there. That means it's not testing the fix.

To fix the test, the `safe_report` needs to be guaranteed to be created before the user is modified. To achieve that, change its creation to `let!(:safe_report)` so the report is guaranteed to be created before the example: https://rspec.info/features/3-13/rspec-core/helper-methods/let/

Bilka2 (approved):
Thank you!

PR #4757, opened 2024-03-01
AO3-5937 "Save" button visible on bookmarks when logged out and on a tag's bookmarks page

forceofcalm:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5937 

## Purpose

Wraps the "save" and "saved" actions for bookmarks in a `logged_in?` check.

## Testing Instructions

I have left a comment for clarification on the JIRA ticket.

## Credit

calm (they/them)

### DIFF
--- a/app/views/bookmarks/_bookmarkable_blurb.html.erb
+++ b/app/views/bookmarks/_bookmarkable_blurb.html.erb
@@ -13,21 +13,23 @@
   <%= render "bookmarks/bookmark_item_module", bookmarkable: unwrapped %>
 
   <!--actions-->
-  <ul class="actions" role="navigation">
-    <li>
-      <% if current_user_bookmark ||= bookmark_if_exists(unwrapped) %>
-        <%= link_to ts("Saved"),
-                    edit_bookmark_path(current_user_bookmark),
-                    id: "bookmark_form_trigger_for_#{bookmarkable.id}",
-                    remote: true %>
-      <% else %>
-        <%= link_to ts("Save"),
-                    new_polymorphic_path([bookmarkable, Bookmark]),
-                    id: "bookmark_form_trigger_for_#{bookmarkable.id}",
-                    remote: true %>
-      <% end %>
-    </li>
-  </ul>
+  <% if logged_in? %>
+    <ul class="actions" role="navigation">
+      <li>
+          <% if current_user_bookmark ||= bookmark_if_exists(unwrapped) %>
+            <%= link_to ts("Saved"),
+                        edit_bookmark_path(current_user_bookmark),
+                        id: "bookmark_form_trigger_for_#{bookmarkable.id}",
+                        remote: true %>
+          <% else %>
+            <%= link_to ts("Save"),
+                        new_polymorphic_path([bookmarkable, Bookmark]),
+                        id: "bookmark_form_trigger_for_#{bookmarkable.id}",
+                        remote: true %>
+          <% end %>
+      </li>
+    </ul>
+  <% end %>
 
   <% if logged_in_as_admin? && bookmarkable.class == ExternalWork %>
     <%= render "admin/admin_options", item: bookmarkable %>


ceithir:
Could you add [an integration test](https://github.com/otwcode/otwarchive/wiki/Automated-Testing#integration-tests) covering this fix?

brianjaustin (commented):
I _think_ these should help with the ERB lint warning (I have not had a chance to test locally though, so some further nudges may be required)

forceofcalm:
I'm writing the integration test for this now (or at least, trying to), and I've noticed that the `Save` button is actually an `<li>` element styled like a button. For accessibility reasons, this should be an actual `<button>` element. A change like that is out of the scope of this ticket, but should be noted somewhere.

forceofcalm:
Hi! I have no fucking idea what RuboCop wants from me. I've tried a ton of different indents and it is still upset.

brianjaustin (commented):
If rubocop and/or erblint is giving me trouble, I usually will run
```sh
rubocop -A path/to/file.rb
# or
erblint -A path/to/file.html.erb
```

That will often do a _lot_ more than we need, so I also use `git add -p` to selectively add lines to my commit, and then run `git checkout path/to/file.rb` to avoid changing a bunch of things at once. I sympathise with fighting erblint/rubocop though, so I've pushed what ERB lint seems to want to https://github.com/brianjaustin/otwarchive/commit/6f6f854905edf42c9f1c978b654b8c4b5a505322

features/bookmarks/bookmark_browse.feature
```diff
@@ -16,3 +15,8 @@ Feature: Browse Bookmarks
     When I go to the bookmarks page for user "ethel" with pseud "aka"
     Then I should see "Bookmarked with Other Pseud"
       And I should not see "Bookmarked with Default Pseud"
+
+  Scenario: When logged out, the "save" button on bookmarks does not show
```
Bilka2, line 19: Please create a bookmark in the tag beforehand - otherwise there will never be a "Save" link.

Bilka2 (changes requested):
Looks like you got reviewdog mostly under control! However, the test needs some adjustments.

features/bookmarks/bookmark_browse.feature
```diff
@@ -16,3 +15,11 @@ Feature: Browse Bookmarks
...
+    Given I am logged in as "bookmarker"
+      And I bookmark the work "Test"
+    When I log out
+      And I go to the bookmarks page for the user "bookmarker"
```
Bilka2: This is the wrong page to check - it doesn't have the bug that this is fixing. The test should visit the bookmarks page for a tag instead. Looking at the previous test failure, I think the issue with visiting the tags bookmarks page/making the bookmark was that the tag didn't exist yet.

features/bookmarks/bookmark_browse.feature
```diff
@@ -16,3 +15,11 @@ Feature: Browse Bookmarks
     When I go to the bookmarks page for user "ethel" with pseud "aka"
     Then I should see "Bookmarked with Other Pseud"
       And I should not see "Bookmarked with Default Pseud"
+
+  Scenario: When logged out, the "save" button on bookmarks does not show
+    Given I am logged in as "bookmarker"
+      And I bookmark the work "Test" with the tags "testing"
```
Bilka2: This test doesn't fail when the change is reverted. This is an indication that it's not testing the right thing.

In particular, this creates a bookmark with the bookmarker tag "testing". To reproduce the issue, a bookmark in the testing fandom is needed. For that, the fandom needs to be canonical, there needs to be a work in the fandom and then that work should be bookmarked.

Bilka2 (changes requested):

features/bookmarks/bookmark_browse.feature
forceofcalm: Understood! Next time I have the chance to look at this, I will do so.

Bilka2:
I've flipped the label back to action needed since the comment about the test still needs addressing

features/step_definitions/bookmark_steps.rb
```diff
@@ -495,3 +495,7 @@ def submit_bookmark_form(pseud, note, tags, collection)
   bookmark.reload
   assert orig_cache_key == bookmark.cache_key, "Cache key #{orig_cache_key} does not match #{bookmark.cache_key}."
 end
+
+When "I go to the bookmarks for the tag {string}" do |tag|
```
Bilka2: The tests don't actually use this step, instead they use the generic "I go to ": https://github.com/otwcode/otwarchive/blob/7474fad9c49b4ebfbca2e500ebe95ec74e9adbe4/features/step_definitions/web_steps.rb#L27-L29

This means the tests go to the overall bookmarks index instead of the tag's index and don't fail even if the change in this PR is reverted. If you define this path in [paths.rb](https://github.com/otwcode/otwarchive/blob/master/features/support/paths.rb) instead then it should work correctly

forceofcalm: Okay! I *think* I managed that. Sorry, I am so completely new to Cucumber and have so much trouble with the close-to-English frameworks. ^^;

features/support/paths.rb
```diff
@@ -195,6 +195,9 @@ def path_to(page_name)
     when /^the bookmarks tagged "(.*)"$/i
       step %{all indexing jobs have been run}
       tag_bookmarks_path(Tag.find_by_name($1))
+    when /^the bookmarks for the tag "(.*)"$/i
```
Bilka2:
```suggestion
    when /^the bookmarks page for the tag "(.*)"$/i
```
to match the name you used in the test - and then it'll finally work!

(it would also be nice if you could merge master. but if you're tired of this pr at this point we can take care of it!)

app/views/bookmarks/_bookmarkable_blurb.html.erb
```diff
@@ -13,23 +13,22 @@
   <%= render "bookmarks/bookmark_item_module", bookmarkable: unwrapped %>
 
   <!--actions-->
-  <% if logged_in? %>
```
Bilka2: Looks like this change got lost in the merge conflict (but at least that means the test is working as expected)

The locale file like just needs another normalization pass

PR #4758, opened 2024-03-01
AO3-6004 show hidden bookmarks on a bookmarkable to admins

walshyb:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6004

## Purpose

Allows admins to show hidden bookmarks on a bookmarkable

## Testing Instructions

 - Go to a bookmark index page with bookmarks
 - Hide a bookmark
 - When page refreshes, see "Make Bookmark Visible"
 - Clicking "Make Bookmark Visible" should refresh page and bring back bookmark

## Credit

Brandon W (he/him/they/them)


app/controllers/bookmarks_controller.rb
```diff
@@ -61,7 +61,11 @@ def search
   def index
     if @bookmarkable
       access_denied unless is_admin? || @bookmarkable.visible?
-      @bookmarks = @bookmarkable.bookmarks.is_public.paginate(page: params[:page], per_page: ArchiveConfig.ITEMS_PER_PAGE)
+      @bookmarks = @bookmarkable.bookmarks.is_public
+      if is_admin?
+        @bookmarks += @bookmarkable.bookmarks.where(hidden_by_admin: true)
```
walshyb, line 66: I think this is a bit weird how I'm doing it, but I didn't see another way to include bookmarks that have been hidden by admin *and* the public bookmarks.

app/controllers/bookmarks_controller.rb
```diff
@@ -61,7 +61,11 @@ def search
   def index
     if @bookmarkable
       access_denied unless is_admin? || @bookmarkable.visible?
-      @bookmarks = @bookmarkable.bookmarks.is_public.paginate(page: params[:page], per_page: ArchiveConfig.ITEMS_PER_PAGE)
+      @bookmarks = @bookmarkable.bookmarks.is_public
+      if is_admin?
```
walshyb, line 65: Should I add a specific check for `policy_and_abuse` role?

app/controllers/bookmarks_controller.rb
```diff
@@ -61,7 +61,9 @@ def search
   def index
     if @bookmarkable
       access_denied unless is_admin? || @bookmarkable.visible?
-      @bookmarks = @bookmarkable.bookmarks.is_public.paginate(page: params[:page], per_page: ArchiveConfig.ITEMS_PER_PAGE)
+      @bookmarks = @bookmarkable.bookmarks.is_public
+      @bookmarks += @bookmarkable.bookmarks.where(hidden_by_admin: true) if is_admin?
```
walshyb, line 65: I think this is a bit weird how I'm doing it, but I didn't see another way to include bookmarks that have been hidden by admin and the public bookmarks. 

Also should I add check for `policy_and_abuse` role?

features/admins/admin_works.feature
```diff
@@ -82,6 +82,7 @@ Feature: Admin Actions for Works, Comments, Series, Bookmarks
     When I follow "Hide Bookmark"
       And all indexing jobs have been run
     Then I should see "Item has been hidden."
+      And I should see "Make Bookmark Visible"
```
walshyb, line 85: Should I add another check to follow this and confirm visible from regular user end?

app/controllers/bookmarks_controller.rb
sarken: Just a quick drive-by comment, not a thorough review:

1. Let's not check the admin role, since I'm pretty sure we don't currently check it for, e.g., accessing hidden works directly. But maybe we should changes this two instances of `is_admin?` to `logged_in_as_admin?` -- this seems to be the only file where we use `is_admin?` and that's weird, imo.
2. I'm guessing the problem with simply removing the `is_public` scope is you get private bookmarks as well as those hidden by admin? Would changing it to the `not_private` scope fix that? If not, I'd suggest defining a new scope in the bookmark model that will get public bookmarks, regardless of whether they're hidden by admin.

features/admins/admin_works.feature
Bilka2: A regular user should not be able to see a hidden bookmark. But it would be nice to have a check that an admin can see the text of the hidden bookmark.

app/controllers/bookmarks_controller.rb
walshyb: 1. I updated those `is_admin?` references to be `logged_in_as_admin?`
2. On a work's bookmarks page, users should only see public bookmarks, unless you're an admin, then they'll want to see public and hidden, right? Or should admins also be able to see private bookmarks on a work? Also, my changes here appear to work in that non-admins can only see public bookmarks and admins can see public and hidden; do you still think we should create a new scope? It's possible I'm misunderstanding the ask

Thank you!

features/admins/admin_works.feature
walshyb: Would you like me to write a test to make sure that a regular user cannot see a hidden bookmark?

app/controllers/bookmarks_controller.rb
Bilka2: > On a work's bookmarks page, users should only see public bookmarks, unless you're an admin, then they'll want to see public and hidden, right?

Yes and the current code does that.

The way I understand sarken's suggestion is to use the `not_private` scope instead of `is_public`, since it gets the public bookmarks regardless of the hidden_by_admin status. Then filter out the hidden bookmarks depending on `logged_in_as_admin?`. That way, there aren't two queries that deal with `hidden_by_admin` (the `is_public` scope and the extra query with the `where` added in this PR), but only the one query (`not_private` and conditional `where` for hidden_by_admin on the returned relation).

walshyb: Hello! Sorry for the delay on this, I know it's been some time.

Thanks for the suggestion. I think I made it accordingly. Tests should be rerunning now

Let me know if you want me to make that filter line more readable. i.e

from: 
```
@bookmarks = @bookmarks.filter { |bookmark| !bookmark.hidden_by_admin } unless logged_in_as_admin?
```

to:
```
if !logged_in_as_admin?
  @bookmarks = @bookmarks.filter { |bookmark| !bookmark.hidden_by_admin }
end
```

Bilka2: It would be good to keep using where instead of filter: Rails is clever enough that it combines
```
@bookmarks = @bookmarkable.bookmarks....
@bookmarks = @bookmarks.where { hidden_by_admin: false } unless logged_in_as_admin?
```
into one db query. With `filter` all bookmarks are retrieved from the database and filtered on the ruby side of things, which isn't as fast as doing the filtering in the database. The pagination just needs to be done after the conditional where, but that too is combined into the query by Rails.

Pretty sure rubocop would complain about your suggested formatting, so the formatting is fine as is!

walshyb: good catch!! i just tested it locally and it does appear to combine the queries to only make 1 db transaction!

Bilka2 (approved):
Thank you!

app/controllers/bookmarks_controller.rb
Bilka2: Yeah, it's so nice that Rails is clever like that :D

Rereading my comment shows that it had a few mistakes (wrong formatting for the where, you don't need to move the pagination), apologies for that. Your commit is just better than my suggestion :+1:

walshyb: No worries and thank you!! 

And just to clarify, the pagination is fine where it is right?

Bilka2: Yep! That's what I meant with my original comment being incorrect.

PR #4763, opened 2024-03-05
AO3-6691 Custom cops for house rules

brianjaustin:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6691

## Purpose

Adds custom Rubocop rules to automatically flag some things we have to do manually in PR review. For example, using departure for large tables.

### DIFF
--- a/rubocop/cop/cucumber/regex_step_name.rb
+++ b/rubocop/cop/cucumber/regex_step_name.rb
@@ -42,7 +42,7 @@ module RuboCop
 
         def on_send(node)
           regex_name(node) do |regex_node|
-            add_offense(regex_node)
+            add_offense(regex_node, severity: :refactor)
           end
         end
       end
--- a/rubocop/cop/cucumber/regex_step_name.rb
+++ b/rubocop/cop/cucumber/regex_step_name.rb
@@ -1,0 +1,51 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Cucumber
+      # Checks that Cucumber step definitions use Cucumber expressions
+      # instead of Regex. Note: this may not always be possible, and this
+      # cop is not smart enough to detect those cases.
+      #
+      # @example
+      #   # bad
+      #   Given /foobar/ do
+      #     ...
+      #   end
+      #   When /baz/ do
+      #     ...
+      #   end
+      #   Then /oops(\w+)/ do |it|
+      #     ...
+      #   end
+      #
+      # @example
+      #   # good
+      #   Given "foobar(s)" do
+      #     ...
+      #   end
+      #   When "baz" do
+      #     ...
+      #   end
+      #   Then "oops{str}" do |it|
+      #     ...
+      #   end
+      class RegexStepName < RuboCop::Cop::Base
+        MSG = "Prefer Cucumber expressions (https://github.com/cucumber/cucumber-expressions) over regex for step names"
+
+        RESTRICT_ON_SEND = %i[Given When Then].freeze
+
+        # @!method regex_name(node)
+        def_node_matcher :regex_name, <<~PATTERN
+          (send nil? _ $(:regexp ...) ...)
+        PATTERN
+
+        def on_send(node)
+          regex_name(node) do |regex_node|
+            add_offense(regex_node)
+          end
+        end
+      end
+    end
+  end
+end
--- a/rubocop/cop/cucumber/regex_step_name.rb
+++ b/rubocop/cop/cucumber/regex_step_name.rb
@@ -31,7 +31,7 @@ module RuboCop
       #     ...
       #   end
       class RegexStepName < RuboCop::Cop::Base
-        MSG = "Prefer Cucumber expressions (https://github.com/cucumber/cucumber-expressions) over regex for step names"
+        MSG = "Prefer Cucumber expressions (https://github.com/cucumber/cucumber-expressions) over regex for step names; refer to https://github.com/otwcode/otwarchive/wiki/Reviewdog-and-RuboCop if regex is still required"
 
         RESTRICT_ON_SEND = %i[Given When Then].freeze
 
--- a/rubocop/cop/i18n/deprecated_helper.rb
+++ b/rubocop/cop/i18n/deprecated_helper.rb
@@ -18,7 +18,7 @@ module RuboCop
       #  t(".relative.path.to.translation")
       #  t(".greeting", name: "world")
       class DeprecatedHelper < RuboCop::Cop::Base
-        MSG = "Prefer Rails built-in `I18n.t` function over `ts`: the latter is not actually translatable"
+        MSG = "Prefer Rails built-in `I18n.t` function over `ts`: the latter is not actually translatable. For more information, refer to https://github.com/otwcode/otwarchive/wiki/Internationalization-(i18n)-Standards."
 
         RESTRICT_ON_SEND = %i[ts].freeze
 
--- a/rubocop/cop/i18n/deprecated_helper.rb
+++ b/rubocop/cop/i18n/deprecated_helper.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module I18n
+      # Checks for uses of the deprecated helper function, `ts`.
+      # Strings passed to it cannot be translated, and all calls
+      # will need to be replaced with I18n.t to enable UI translations
+      # in the future.
+      #
+      # @example
+      #  # bad
+      #  ts("This will only be in English :(")
+      #  ts("Hello %{name}", name: "world")
+      #
+      # @example
+      #  # good
+      #  t(".relative.path.to.translation")
+      #  t(".greeting", name: "world")
+      class DeprecatedHelper < RuboCop::Cop::Base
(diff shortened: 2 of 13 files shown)


rubocop/cop/i18n/deprecated_helper.rb
```diff
@@ -0,0 +1,31 @@
...
+      #  # bad
+      #  ts("This will only be in English :(")
+      #  ts("Hello %{name}", name: "world")
+      #
+      # @example
+      #  # good
+      #  t(".relative.path.to.translation")
+      #  t(".greeting", name: "world")
+      class DeprecatedHelper < RuboCop::Cop::Base
+        MSG = "Prefer Rails built-in `I18n.t` function over `ts`: the latter is not actually translatable"
```
sarken, line 21: Can this link to our GitHub wiki page about i18n standards?

brianjaustin: Done! (I wanted to make this multi-line with a heredoc, but the rubocop tests get upset at that and I think it's OK to be all one line so :shrug: )

rubocop/cop/migration/large_table_schema_update.rb
```diff
@@ -0,0 +1,99 @@
...
+      #
+      # @example
+      #  # bad
+      #  class ExampleMigration < ActiveRecord::Migration[6.1]
+      #    add_column :users, :new_field, :integer, nullable: true
+      #  end
+      class LargeTableSchemaUpdate < RuboCop::Cop::Base
+        MSG = "Use departure to change the schema of large table `%s`"
+
+        LARGE_TABLES = %i[
```
Bilka2, line 26: Same note regarding moving this to a config file as for `DENIED_TRANSLATION_KEYS`. For this one especially I can see it being very valuable to be able to link easily to this in the wiki, since the wiki already [mentions departure](https://github.com/otwcode/otwarchive/wiki/Migrations#user-content-Using_ptonlineschemachange).

Moving this to the config file would also make it easier for any other users of the archive software to adjust this cop for their database table sizes.

rubocop/cop/i18n/static_translation_definition.rb
```diff
@@ -0,0 +1,116 @@
...
+module RuboCop
+  module Cop
+    module I18n
+      # This cop flags translation definitions in static scopes because changing
+      # locales has no effect and won't translate this text again. The rule here is based upon
+      # https://gitlab.com/gitlab-org/gitlab/-/blob/master/rubocop/cop/static_translation_definition.rb?ref_type=heads.
+      #
+      # @example
+      #   # bad
+      #   class MyExample
```
Bilka2, line 12: To go with the point about `t` being preferred over `I18n.t` in controllers and views, maybe this class should be `ExampleController` so that `t` would actually work here.

Generally this example class is not great because it uses methods in the good example that are never used by us and have better alternatives. For example, see the comment on `validates` and consider that we usually do extra error messages via something like `self.errors.add(:base, :blocked_gifts, byline: gift.pseud.byline)`, which (like the validate example) doesn't call `t` at all but instead works with Rails magic to look up translations via a symbol.

Are translations in static scopes an issue for us at all? If not, it may be better to just not add this cop because it uses so much code in the good example that I would consider quite questionable.

Or maybe these examples could show only the `validates` case, since that is what the rspec test checks for. The rest of the checks could stay, but I would not include them in the example, especially not the "good" case.

rubocop/cop/i18n/static_translation_definition.rb
```diff
@@ -0,0 +1,116 @@
...
+      #     # Constant
+      #     Translation = I18n.t('A translation.')
```
Bilka2, line 14: The demo calls to `t` in this file should be using translation keys like:
```suggestion
      #     Translation = I18n.t('some_translation_key')
```
or maybe even relative keys (if the class is meant to be a controller).

rubocop/cop/i18n/static_translation_definition.rb
```diff
@@ -0,0 +1,116 @@
...
+      #     # Keep translations dynamic.
+      #     Translation = -> { I18n.t('A translation.') }
+      #     # OR
+      #     def translation
+      #       I18n.t('A translation.')
+      #     end
+      #
+      #     field :foo, title: -> { I18n.t('A title') }
+      #
+      #     validates :title, :presence, message: -> { I18n.t('is missing') }
```
Bilka2, line 42: Validations can take a symbol for the message which will be looked up relative to the model in the same way as the [default error messages](https://guides.rubyonrails.org/i18n.html#error-message-scopes). I cannot find docs for this right now, but it is used by the [blocking code](https://github.com/otwcode/otwarchive/blob/acebcaddee50ce6d6d1503f22666ea47b6891f91/app/models/comment.rb#L54), [en.yml](https://github.com/otwcode/otwarchive/blob/acebcaddee50ce6d6d1503f22666ea47b6891f91/config/locales/models/en.yml#L114).

```suggestion
      #     validates :title, :presence, message: :very_custom_missing
```

Bilka2 (commented):
I didn't look at the code itself much, my focus was more on the resulting messages and docs of the cops.

Bilka2:
Is it possible to check the arguments of the method call? In that case it would be nice to have a cop that checks for calls to `t` that include the default option. For example, `t('.email_address', :default => "Email address")` should have the translation in the `.yml` file and be just `t(".email_address")`.

rubocop/cop/i18n/deprecated_helper.rb
```diff
@@ -0,0 +1,31 @@
...
+      #  # bad
+      #  ts("This will only be in English :(")
+      #  ts("Hello %{name}", name: "world")
+      #
+      # @example
+      #  # good
+      #  t(".relative.path.to.translation")
+      #  t(".greeting", name: "world")
+      class DeprecatedHelper < RuboCop::Cop::Base
+        MSG = "Prefer Rails built-in `I18n.t` function over `ts`: the latter is not actually translatable. For more information, refer to https://github.com/otwcode/otwarchive/wiki/Internationalization-(i18n)-Standards."
```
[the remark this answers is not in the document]
brianjaustin: My initial reasoning behind the full `I18n.t` was that looking up just `t` for someone unfamiliar is nigh impossible. But I think with the examples in the now-linked "Internationalization Standards" page, that issue becomes a bit less problematic. Do you think it would be a good idea to include the a link to the Rails i18n docs on that page for completeness?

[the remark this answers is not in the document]
Bilka2: Link to the i18n guide on the wiki page? May be good, may be confusing because the guide walks through a setup that is already done here (and different). The page already links to subsections like the explanation on lazy lookup, but it is a bit hidden in the bullet points. So I could see a link to the https://guides.rubyonrails.org/i18n.html#overview-of-the-i18n-api-features being helpful.

One consideration to make the error easier to search could be to mention the module without using it:
```suggestion
        MSG = "Prefer Rails built-in `t` function from the I18n Module over `ts`: the latter is not actually translatable. For more information, refer to https://github.com/otwcode/otwarchive/wiki/Internationalization-(i18n)-Standards."
```

But I don't really like it.

rubocop/cop/i18n/static_translation_definition.rb
Bilka2: Looking at this example class again - it would be great to use `"string"` instead of `'string'`, so that rubocop won't immediately complain about that when someone copies the example.

rubocop/cop/cucumber/regex_step_name.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Cucumber
+      # Checks that Cucumber step definitions use Cucumber expressions
+      # instead of Regex. Note: this may not always be possible, and this
+      # cop is not smart enough to detect those cases.
```
Bilka2, line 8: This means that the cop will have false failures in those cases, leading to false positive failing tests on PRs. That's not great. Perhaps that would be helped if the offense could be considered a low severity. Though I don't know if that will keep the test from failing while still showing up on GitHub, I'd have to investigate how reviewdog handles it.

But did find that [add_offense](https://www.rubydoc.info/gems/rubocop/RuboCop%2FCop%2FBase:add_offense) indeed supports a [severity level](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Severity).

brianjaustin: Here's what it looks like with `severity: :convention` https://github.com/brianjaustin/ao3-sandbox/pull/9/files

The action is still marked as failed, but at least the annotation looks a little bit less angry (the higher severities are either yellow or red with an :x:)

Bilka2: As tested [here](https://github.com/brianjaustin/ao3-sandbox/pull/11/files) that unfortunately doesn't apply for commits from forks where reviewdog [degrades to annotations](https://github.com/reviewdog/reviewdog?tab=readme-ov-file#graceful-degradation-for-pull-requests-from-forked-repositories). Oh well, was worth the try.

Bilka2 (approved):
All the cops look good and I had a quick look over the code and tests as well, which looks fine (though I didn't look too deep).

sarken (approved):

PR #4765, opened 2024-03-17
AO3-6608 Delete Alt Text With Pseud Icon

smclairecarden:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6608 (Please fill in issue number and remove this comment.)

## Purpose

What does this PR do?

Delete alt text associated with a pseuds Icon when the icon is deleted.

## Testing Instructions

see Jira!

## References

## Credit

Claire C she/her

### DIFF
--- a/app/models/pseud.rb
+++ b/app/models/pseud.rb
@@ -419,6 +419,7 @@ class Pseud < ApplicationRecord
 
   def clear_icon
     return unless delete_icon?
+    
     self.icon = nil if delete_icon?
     self.icon_alt_text = nil
   end
--- a/app/models/pseud.rb
+++ b/app/models/pseud.rb
@@ -419,6 +419,7 @@ class Pseud < ApplicationRecord
 
   def clear_icon
     self.icon = nil if delete_icon? && !icon.dirty?
+    self.icon_alt_text = nil if delete_icon?
   end
 
   #################################
--- a/app/models/pseud.rb
+++ b/app/models/pseud.rb
@@ -418,8 +418,9 @@ class Pseud < ApplicationRecord
   alias_method :delete_icon?, :delete_icon
 
   def clear_icon
-    self.icon = nil if delete_icon? && !icon.dirty?
-    self.icon_alt_text = nil if delete_icon?
+    return unless delete_icon?
+    self.icon = nil if delete_icon?
+    self.icon_alt_text = nil
   end
 
   #################################
--- a/app/models/pseud.rb
+++ b/app/models/pseud.rb
@@ -420,7 +420,7 @@ class Pseud < ApplicationRecord
   def clear_icon
     return unless delete_icon?
     
-    self.icon = nil if delete_icon?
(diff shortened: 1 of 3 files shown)


features/step_definitions/icon_steps.rb
```diff
@@ -34,6 +34,12 @@
   step %{I press "Update"}
 end
 
+When /^I delete the icon from my psued$/ do
```
brianjaustin, line 37: For new steps, we lean towards Cucumber expressions: https://github.com/otwcode/otwarchive/wiki/Automated-Testing#our-cucumber-standards (at least where they're supported). I have a Rubocop rule in the works, but it's not merged just yet.

(Also it looks like there's a typo in pseud)

```suggestion
When "I delete the icon from my pseud" do
```

features/step_definitions/icon_steps.rb
```diff
@@ -45,3 +51,8 @@
 end
 
 ### THEN
+
+Then /^I should see the icon and alt text boxes are blank$/ do
```
brianjaustin, line 55:
```suggestion
Then "I should see the icon and alt text boxes are blank" do
```

app/models/pseud.rb
```diff
@@ -419,6 +419,7 @@ def delete_icon
 
   def clear_icon
     self.icon = nil if delete_icon? && !icon.dirty?
+    self.icon_alt_text = nil if delete_icon?
```
brianjaustin, line 422: Would it make sense to extract `delete_icon?` to a guard instead, since it's required for both of these?
```suggestion
    return unless delete_icon?

    self.icon = nil if delete_icon?
    self.icon_alt_text = nil
```

brianjaustin (changes requested):

features/step_definitions/icon_steps.rb
smclairecarden: Oh thank you, I'll remember this for future tests, and good catch on that misspelling. Changed here and in the feature file.

features/step_definitions/icon_steps.rb
smclairecarden: Changed.

app/models/pseud.rb
smclairecarden: You're right, this does make more sense. Updated.

app/models/pseud.rb
```diff
@@ -418,7 +418,10 @@ def delete_icon
   alias_method :delete_icon?, :delete_icon
 
   def clear_icon
-    self.icon = nil if delete_icon? && !icon.dirty?
+    return unless delete_icon?
+    
+    self.icon = nil if delete_icon?
```
brianjaustin, line 423: 🤦 looks like I carried over the wrong condition in my last review, sorry about that!
```suggestion
    self.icon = nil unless icon.dirty?
```

brianjaustin (approved):
Thanks!

PR #4771, opened 2024-03-28
AO3-4820 AO3-3810 Validate skins with "archive" in title

weeklies:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4820 

https://otwarchive.atlassian.net/browse/AO3-3810 
## Purpose

* Resolve AO3-3810 and AO3-4820 
* Remove an unnecessary step in skin_steps
* Fix failing tests
 
Note: I am not sure which admins should be allowed to edit a skin with “Archive”.

## Testing Instructions

See both issues.

### DIFF
--- a/app/models/skin.rb
+++ b/app/models/skin.rb
@@ -108,8 +108,13 @@ class Skin < ApplicationRecord
     errors.add(:base, ts("You need to upload a screencap if you want to share your skin."))
   end
 
-  validates_presence_of :title
-  validates :title, uniqueness: { message: ts("must be unique"), case_sensitive: true }
+  validates :title, presence: true, uniqueness: { message: ts("must be unique"), case_sensitive: true }
+  validate :allowed_title
+  def allowed_title
+    return true if (!self.title.match(/archive/i) || User.current_user.is_a?(Admin))
+
+    errors.add(:base, ts("You can't use the word 'Archive' in your skin title, sorry! (We have to reserve it for official skins.)"))
+  end
 
   validates_numericality_of :margin, :base_em, allow_nil: true
   validate :valid_font
@@ -479,7 +484,7 @@ class Skin < ApplicationRecord
           skin.unusable = true
           skin.official = true
           File.open(version_dir + 'preview.png', 'rb') {|preview_file| skin.icon = preview_file}
-          skin.save!
+          skin.save!(validate: false)
           skins << skin
         end
 
@@ -494,7 +499,7 @@ class Skin < ApplicationRecord
         end
         File.open(version_dir + 'preview.png', 'rb') {|preview_file| top_skin.icon = preview_file}
         top_skin.official = true
-        top_skin.save!
+        top_skin.save!(validate: false)
         skins.each_with_index do |skin, index|
           skin_parent = top_skin.skin_parents.build(child_skin: top_skin, parent_skin: skin, position: index+1)
           skin_parent.save!
--- a/features/step_definitions/skin_steps.rb
+++ b/features/step_definitions/skin_steps.rb
@@ -82,13 +82,20 @@ end
 
 Given /^the approved public skin "([^"]*)" with css "([^"]*)"$/ do |skin_name, css|
   step "the unapproved public skin \"#{skin_name}\" with css \"#{css}\""
-  step "I am logged in as an admin"
   step "I approve the skin \"#{skin_name}\""
   step "I am logged out"
 end
 
 Given /^the approved public skin "([^"]*)"$/ do |skin_name|
   step "the approved public skin \"#{skin_name}\" with css #{DEFAULT_CSS}"
+end
+
+Given /^the approved public skin "(.*?)" that has reserved words in the title$/ do |skin_name|
+  # Admins can't create skins, so we have to create it this way.
+  FactoryBot.build(:skin, title: skin_name, public: true).save!(validate: false)
+
+  step "I approve the skin \"#{skin_name}\""
+  step "I am logged out"
 end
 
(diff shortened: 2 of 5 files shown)


sarken:
> Note: I am not sure which admins should be allowed to edit a skin with “Archive”.

It should depend on the type of skin -- superadmins can edit site or work skins, and support admins can edit work skins only.

https://github.com/otwcode/otwarchive/blob/acebcaddee50ce6d6d1503f22666ea47b6891f91/app/policies/skin_policy.rb#L3-L4

app/models/skin.rb
```diff
@@ -108,8 +108,13 @@ def valid_public_preview
     errors.add(:base, ts("You need to upload a screencap if you want to share your skin."))
   end
 
-  validates_presence_of :title
-  validates :title, uniqueness: { message: ts("must be unique"), case_sensitive: true }
+  validates :title, presence: true, uniqueness: { message: ts("must be unique"), case_sensitive: true }
```
Bilka2, line 111: Since you're touching this, it would be great to i18n this by using the default locale for the error:

```suggestion
  validates :title, presence: true, uniqueness: { case_sensitive: true }
```
and [overriding it](https://guides.rubyonrails.org/i18n.html#error-message-scopes) in the `models/en.yml` file:

```
  activerecord:
    errors:
      models:
        skin:
          attributes:
            title:
              taken: must be unique
```

features/step_definitions/skin_steps.rb
```diff
@@ -91,6 +90,14 @@
   step "the approved public skin \"#{skin_name}\" with css #{DEFAULT_CSS}"
 end
 
+Given /^the approved public skin "(.*?)" that has reserved words in the title$/ do |skin_name|
```
Bilka2, line 93: Please use a [cucumber expression](https://github.com/cucumber/cucumber-expressions).

Bilka2 (commented):
I pointed out a few code style things that would be nice to change.

It looks like reviewdog got a bit confused with the lines in this file for some reason (pointing out issues for code that didn't change), which might get fixed if you merge `master` into this branch. But the `Style/ParenthesesAroundCondition` violation that hound points out should indeed be fixed.

config/locales/models/en.yml
```diff
@@ -155,6 +155,15 @@ en:
               not_work: Only a link to a work can be listed as an inspiration.
               protected: You can't use the related works function to cite works by the protected user %{login}.
           format: "%{message}"
+        skin:
+          attributes:
+            base:
+              archive_in_title: You can't use the word 'Archive' in your skin title, sorry! (We have to reserve it for official skins.)
+              format: "%{message}"
```
Bilka2, line 162: This is not needed when the error is added onto `:base`. Base means the overall model, so there is no attribute that gets added to the error message: https://guides.rubyonrails.org/active_record_validations.html#errors-base

weeklies:
I added a migration to make the skin title index unique as Rubocop suggested, as it seems that would be beneficial for efficiency.

Bilka2:
Just dropping a note here that I have run out of knowledge to continue reviewing this PR in particular, someone else will have to review the functional code. But the locale looks good :)

config/locales/models/en.yml
```diff
@@ -155,6 +155,13 @@ en:
               not_work: Only a link to a work can be listed as an inspiration.
               protected: You can't use the related works function to cite works by the protected user %{login}.
           format: "%{message}"
+        skin:
+          archive_in_title: You can't use the word 'Archive' in your skin title, sorry! (We have to reserve it for official skins.)
```
brianjaustin, line 159: I talked with AD&T and Support about updating the wording for some of these, since it's been a bit and they don't quite match how we word things today. First suggestion:
```suggestion
          archive_in_title: Sorry, titles including the word 'Archive' are reserved for official skins.
```

config/locales/models/en.yml
```diff
@@ -155,6 +155,13 @@ en:
...
+          attributes:
+            title:
+              taken: must be unique
+          invalid_media: We don't currently support the media type %{media}, sorry! If we should, please let Support know.
```
brianjaustin, line 163: Second suggestion: update this to link "let Support know" to the feedback/support form

config/locales/models/en.yml
```diff
@@ -155,6 +155,13 @@ en:
...
+          no_public_preview: You need to upload a screencap if you want to share your skin.
```
brianjaustin, line 164: Noting that we would update the wording here, but at some point this feature is going away anyways so it's not really worth updating only to delete later

brianjaustin (changes requested):

weeklies:
I can't seem to get the support link to work in the error  message. Help?

app/models/skin.rb
```diff
@@ -93,7 +94,12 @@ def clean_media
   validate :valid_media
   def valid_media
     if media && media.is_a?(Array) && media.any? {|m| !MEDIA.include?(m)}
-      errors.add(:base, ts("We don't currently support the media type %{media}, sorry! If we should, please let Support know.", media: media.join(', ')))
+      errors.add(
+        :base,
+        :invalid_media_html,
+        media: media.join(", "),
+        support_link: "<a href=\"#{new_feedback_report_path}\">#{I18n.t("activerecord.errors.models.skin.contact_support")}</a>".html_safe,
```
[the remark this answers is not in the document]
brianjaustin: It's unfortunate that `link_to` doesn't work, but all the workarounds to make it work were even less nice :cry: I'm really sorry for the flip-flopping, but I talked it over some more with the rest of the committee, and we can live without the link given it the hackery to make it work

brianjaustin (approved):
Thanks!

PR #4782, opened 2024-04-04
AO3-6704 Use URL from current host in abuse report response mailer preview

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6704

## Purpose

Use `ArchiveConfig.APP_HOST` to create the URL for the abuse report in the abuse_report mailer preview. That way it uses a staging URL on staging.

## Testing Instructions

See Jira.

## Credit

Bilka (he/him)

### DIFF
--- a/test/mailers/previews/user_mailer_preview.rb
+++ b/test/mailers/previews/user_mailer_preview.rb
@@ -1,6 +1,6 @@
 class UserMailerPreview < ApplicationMailerPreview
   # Sent to a user when they submit an abuse report
-  def abuse_report_response
+  def abuse_report
     abuse_report = create(:abuse_report, url: "https://#{ArchiveConfig.APP_HOST}/tags/1984%20-%20George%20Orwell")
     UserMailer.abuse_report(abuse_report.id)
   end
--- a/test/mailers/previews/user_mailer_preview.rb
+++ b/test/mailers/previews/user_mailer_preview.rb
@@ -1,7 +1,7 @@
 class UserMailerPreview < ApplicationMailerPreview
   # Sent to a user when they submit an abuse report
   def abuse_report_response
-    abuse_report = create(:abuse_report)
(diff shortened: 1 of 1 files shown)


test/mailers/previews/user_mailer_preview.rb
```diff
@@ -1,7 +1,7 @@
 class UserMailerPreview < ApplicationMailerPreview
   # Sent to a user when they submit an abuse report
   def abuse_report_response
-    abuse_report = create(:abuse_report)
+    abuse_report = create(:abuse_report, url: "https://#{ArchiveConfig.APP_HOST}/tags/1984%20-%20George%20Orwell")
```
ceithir, line 4: Might be more robust and more consistent with  https://github.com/otwcode/otwarchive/pull/4488 to instead have the factory create an always valid for the environment url like:
`"https://#{ArchiveConfig.PERMITTED_HOSTS.first}/tags/2000%20AD%20(Comics)/works"`
?

Bilka2: The IP that's currently first in the [config file in this repo](https://github.com/brianjaustin/otwarchive/blob/139b76b284aac4bae5f959e5e9fd6155d0cb30f2/config/config.yml#L674) is not a valid host, so that wouldn't work for development environments unless people manually override the config. Currently nothing in the dev setup tutorials instructs to override the config file via `local.yml`, so I'm hesitant to start requiring. Especially because `APP_HOST` works just fine without further configuration. I also don't know how `PERMITTED_HOSTS` is configured for staging and prod, since there is nothing that requires the first (or last or whatever) entry in the `PERMITTED_HOSTS` array to be the host that's used for most of the archive.

Though I do get your argument, since things would start breaking when someone overrides `APP_HOST` to a host not included in `PERMITTED_HOSTS`. However, `APP_HOST` is also used for the [other URLs in mailers](https://github.com/brianjaustin/otwarchive/blob/139b76b284aac4bae5f959e5e9fd6155d0cb30f2/config/application.rb#L111 ) by default, so I think at that point that hypothetical someone would likely have larger problems than this email preview.

PR #4785, opened 2024-04-07
AO3-6297 Revert accidental gem downgrades in MimeMagic removal

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6297

## Purpose

The MimeMagic removal commit d40d751 had some extra changes in the Gemfile.lock that led to various gems getting downgraded. Most likely this was caused by a merge conflict. This commit reverts the extra changes.

## Testing Instructions

No extra testing.

## References

I noticed this because nokogiri got downgraded to a vulnerable version and dependabot opened #4784 to upgrade it again.

Comparison between this branch and master before d40d751: https://github.com/otwcode/otwarchive/compare/test-0.9.367.5...Bilka2:AO3-6297-bot. Shows that after this PR, only following things are changed in the Gemfile.lock:
* The originally intended removal of MimeMagic.
* Update of net-smtp from 0.4.0 to 0.5.0. This was also introduced with d40d751. I decided not to revert it because it's a version upgrade and based on [the changelog](https://github.com/ruby/net-smtp/blob/master/NEWS.md#version-050-2024-03-27) there are no changes that affect us.
* Minor nokogiri version bump (1.16.2 -> 1.16.3). Shouldn't matter and it was easier than forcing bundler to pick 1.16.2.
* Update of bundler to 2.5.6, which is the version used in the [Dockerfile](https://github.com/otwcode/otwarchive/blob/9894e31f32da7f61240f4489d8454aa06334fd79/config/docker/Dockerfile#L19).

## Credit

Bilka (he/him)

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -352,7 +352,7 @@ GEM
       mime-types-data (~> 3.2015)
     mime-types-data (3.2022.0105)
     mini_mime (1.1.2)
-    mini_portile2 (2.8.2)
+    mini_portile2 (2.8.5)
     minitest (5.17.0)
     mono_logger (1.1.2)
     multi_json (1.15.0)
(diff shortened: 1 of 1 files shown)


brianjaustin (approved):
Thank you! Very much not a fan of Gemfile.lock conflicts, I always seem to run into something nasty when trying to fix them :facepalm:

PR #4519, opened 2023-05-25
AO3-6240 Use i18n-tasks to rename numbered mailer translation keys

Bilka2:
# Pull Request Checklist

* [X] Have you read ["How to write the perfect pull request"](https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/)?
* [X] Have you read the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [ ] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)?
* [X] Have you added the [Jira](https://otwarchive.atlassian.net) issue number
  as the *first* thing in your pull request title (e.g. `AO3-1234 Fix thing`)
* [X] Do you have fewer than 5 pull requests already open? If not, please wait
  until they are reviewed and merged before creating new pull requests.

## Issue

https://otwarchive.atlassian.net/browse/AO3-6240

## Purpose

Rename numbered locale keys used by the mailers.

## Testing Instructions

See comment on Jira issue.

## References

#4497 for moving instructions, will add them to the wiki once this PR is reviewed.

## Credit

Bilka (he/him)

PR #4796, opened 2024-04-22
AO3-4308 Update claim email text with t's suggestion

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-4308

### DIFF
--- a/app/views/user_mailer/claim_notification.html.erb
+++ b/app/views/user_mailer/claim_notification.html.erb
@@ -25,7 +25,7 @@
     <% end %>
   </ul>
 
-  <p><%= t(".redirects.html") %></p>
+  <p><%= t(".redirects.html", negation: style_bold(t(".redirects.negation"))) %></p>
 
   <p><%= t(".update_redirect.html", contact_open_doors_link: opendoors_link(t(".update_redirect.contact_open_doors"))) %></p>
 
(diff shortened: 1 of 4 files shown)


app/views/user_mailer/claim_notification.html.erb
```diff
@@ -25,7 +25,7 @@
     <% end %>
   </ul>
 
-  <p><%= t(".redirects") %></p>
+  <p><%= t(".redirects_html") %></p>
```
Bilka2, line 28: Would be nice to have the locale keys be `redirects.html` and `redirects.text`, see https://github.com/otwcode/otwarchive/pull/4519#discussion_r1212423492.

Bilka2: Note: This is a nitpick. If it's urgent, the rest of the commit looks good and this could be merged as is.

Bilka2 (approved):
Thank you!

brianjaustin:
(marking as action needed again after some discussion with @sarken elsewhere)

Bilka2:
It's not clear to me whether this still applies, but in [a past PR](https://github.com/otwcode/otwarchive/pull/4519#discussion_r1212423492) it was preferred to have a html and text version of a html safe key even when they use the same English text.

sarken:
All good, thank you!

sarken (approved):

PR #4772, opened 2024-03-29
AO3-6564 Strip images from more fields

brianjaustin:
# Pull Request Checklist

* [X] Have you read ["How to write the perfect pull request"](https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/)?
* [X] Have you read the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [X] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)?
* [X] Have you added the [Jira](https://otwarchive.atlassian.net) issue number
  as the *first* thing in your pull request title (e.g. `AO3-1234 Fix thing`)
* [X] Do you have fewer than 5 pull requests already open? If not, please wait
  until they are reviewed and merged before creating new pull requests.

## Issue

https://otwarchive.atlassian.net/browse/AO3-6564

## Purpose

Strip images in a few funky places, in addition to the original PR (#4729)
* automatic replies to abuse reports (the Zoho side already should strip)
* automatic replies to feedback/support requests (same re Zoho)
* inbox comments for replies to comments on admin/news posts

PR #4729, opened 2024-01-30
AO3-6564 Reduce fields where images are shown

weeklies:
# Pull Request Checklist

* [x] Have you read ["How to write the perfect pull request"](https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/)?
* [x] Have you read the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [x] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)?
* [x] Have you added the [Jira](https://otwarchive.atlassian.net) issue number
  as the *first* thing in your pull request title (e.g. `AO3-1234 Fix thing`)
* [x] Do you have fewer than 5 pull requests already open? If not, please wait
  until they are reviewed and merged before creating new pull requests.

## Issue

https://otwarchive.atlassian.net/browse/AO3-6564

## Purpose

Strip images in the following:
- bookmarker notes
- pseud descriptions
- admin banners
- comments on news posts
- comment field on PAC and Support forms

## Testing Instructions

See Jira issue.

## Credit

weeklies (she/her)

PR #4799, opened 2024-04-25
AO3-6719 Leave src URLs instead of emptiness for images stripped in AO3-6564

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6719

## Purpose

Adds a `keep_src` keyword argument to the `strip_images` method that lets us have the option of replacing the stripped `img` tag with its `src` URL instead of empty space. I'm happy to give it a different name or make it a separate method. I very much hate the hack to make it work with tests, but I also felt like I'd get side-eyed for changing the tests. 😆 I can do that instead, though!

Replaces the `strip_images` keyword argument on `sanitize_field` with a new argument called `image_safety_mode`. We use it in the same places we were using `strip_images`, but it's a more flexible name in case we decide to switch to placeholders or go back to nothingness or something else entirely -- right now it uses `strip_images` with `keep_src: true` on the sanitized content. Again, delighted to give it a different name if you have one. 

I assumed that we wouldn't mind including the image URL in the main content of support tickets and abuse reports, as well as the emailed copies of same. They were kind of a weird hybrid of "affected by AO3-6564" and "not affected by AO3-6564." Happy to go back to no images at all, though.

ETA: And the big one, make it so we can use the config to temporarily strip images in work comments without a deploy. (You can do tags, too, but that seems unlikely to be an issue.) We'll need to set `PARENTS_WITH_IMAGE_SAFETY_MODE: ["AdminPost"]` when this is deployed to keep the current behavior of stripping images from news post comments.

## Testing Instructions

## References

* [AO3-6564](https://otwarchive.atlassian.net/browse/AO3-6564)
* #4729 
* #4772

[AO3-6564]: https://otwarchive.atlassian.net/browse/AO3-6564?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/features/comments_and_kudos/image_safety_mode.feature
+++ b/features/comments_and_kudos/image_safety_mode.feature
@@ -8,12 +8,20 @@ Feature: Image safety mode
       And image safety mode is disabled for comments
     When I view <commentable> with comments
     Then I should see the image "src" text "https://example.com/image.jpg"
+      And I should not see "OMG! https://example.com/image.jpg"
     When I go to the homepage
     Then I should see the image "src" text "https://example.com/image.jpg"
+      And I should not see "OMG! https://example.com/image.jpg"
     When I go to my inbox page
     Then I should see the image "src" text "https://example.com/image.jpg"
     When image safety mode is enabled for comments on a "<parent_type>"
       And I view <commentable> with comments
+    Then I should not see the image "src" text "https://example.com/image.jpg"
+      But I should see "OMG! https://example.com/image.jpg"
+    When I go to the homepage
+    Then I should not see the image "src" text "https://example.com/image.jpg"
+      But I should see "OMG! https://example.com/image.jpg"
+    When I go to my inbox page
     Then I should not see the image "src" text "https://example.com/image.jpg"
       But I should see "OMG! https://example.com/image.jpg"
 
@@ -38,7 +46,13 @@ Feature: Image safety mode
     When image safety mode is disabled for comments
       And I view <commentable> with comments
     Then I should see the image "src" text "https://example.com/image.jpg"
-      But I should not see "OMG! https://example.com/image.jpg"
+      And I should not see "OMG! https://example.com/image.jpg"
+    When I go to the homepage
+    Then I should see the image "src" text "https://example.com/image.jpg"
+      And I should not see "OMG! https://example.com/image.jpg"
+    When I go to my inbox page
+    Then I should see the image "src" text "https://example.com/image.jpg"
+      And I should not see "OMG! https://example.com/image.jpg"
 
     Examples:
       | parent_type | commentable                 |
--- a/features/comments_and_kudos/image_safety_mode.feature
+++ b/features/comments_and_kudos/image_safety_mode.feature
@@ -1,0 +1,48 @@
+Feature: Image safety mode
+  In order to protect users
+  As a site owner
+  I'd like to control which comments can include images
+
+  Scenario Outline: Images are embedded in comments when image safety mode is off.
+    Given the setup for testing image safety mode on <commentable>
+      And image safety mode is disabled for comments
+    When I view <commentable> with comments
+    Then I should see the image "src" text "https://example.com/image.jpg"
+    When I go to the homepage
+    Then I should see the image "src" text "https://example.com/image.jpg"
+    When I go to my inbox page
+    Then I should see the image "src" text "https://example.com/image.jpg"
+    When image safety mode is enabled for comments on a "<parent_type>"
+      And I view <commentable> with comments
+    Then I should not see the image "src" text "https://example.com/image.jpg"
+      But I should see "OMG! https://example.com/image.jpg"
+
+    Examples:
+      | commentable                 | parent_type |
+      | the admin post "Change Log" | AdminPost   |
+      | the work "My Opus"          | Chapter     |
+      | the tag "No Fandom"         | Tag         |
+
+  Scenario Outline: Embedded images in comments are replaced with their URLs when image safety mode is enabled.
+    Given the setup for testing image safety mode on <commentable>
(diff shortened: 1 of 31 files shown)


Bilka2:
I think we've always used the default configuration, which is slashes with AllowInnerSlashes false. That makes it want `r%{}` for expressions with slashes in them like on this [old PR](https://github.com/otwcode/otwarchive/pull/3383#discussion_r202239646) and `//` otherwise, like in this PR.

lib/html_cleaner.rb
```diff
@@ -144,9 +146,18 @@ def add_paragraphs_to_text(text)
...
-  # strip img tags
-  def strip_images(value)
-    value.gsub(/<img .*?>/, '')
+  # strip img tags, optionally leaving the src URL exposed
+  def strip_images(value, keep_src: false)
+    value.gsub(%r{(<img .*?>)}) do |img_tag|
+      if keep_src
+        # The cleaned field will always use double quotes (") but the fields
+        # in tests aren't always cleaned, so we're allowing double or single
+        # quotes here.
```
Bilka2, line 155: Oof, how sure is it that it's only the tests where the fields aren't cleaned? An uncleaned field could have unexpected formatting beyond just different quotes, which would break/bypass the two regular expressions here.

A lot of the time, this function is called from `sanitize_field`. I think it would be more reliable to use Sanitize or nokogiri in `sanitize_value` to do the stripping. There we're already dealing with those libs and I'd expect them to be more reliable than making our own regexp (don't reinvent the wheel and all that).

Doing the image stripping in `sanitize_value` would also make things a bit nicer because we could use a config similar to `ArchiveConfig.FIELDS_ALLOWING_VIDEO_EMBEDS` to designate fields to clean (or rather, not clean), instead of calling strip_images during display. That also makes it less likely that some display spots will get missed like it already happened in AO3-6564.

With the source urls getting preserved, doing this in the sanitisation step also wouldn't lose any data, if that was a concern previously.

sarken:
Hold pls.

spec/lib/html_cleaner_spec.rb
```diff
@@ -1104,4 +1104,90 @@
...
+
+    context "with keep_src: true" do
+      it "keeps the img src URL when the src uses double quotes" do
+        string = 'Hi! <img src="http://example.org/image.png" /> Bye'
+        result = "Hi! http://example.org/image.png Bye"
+        expect(strip_images(string, keep_src: true)).to eq(result)
+      end
+
+      it "removes the img tag entirely when the src uses single quotes" do
+        string = "Hi! <img src='http://example.org/image.png'> Bye"
```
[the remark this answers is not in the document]
sarken: Yeah, idk. I originally treated them the same for consistency, even though I don't really like writing code just to make tests pass, but after Bilka's comment, I got more skittish about that approach. 

I think maybe we should have different behavior so it forces us to think about what's going on, e.g., if we're seeing the wrong behavior on production, that's a big yikes because it means the text isn't getting sanitized properly, and if we're seeing the wrong behavior in tests, is it because we're making test data in a way that's bypassing the sanitizer, or because there's actually a problem.

That said, I'm happy to take the path of least resistance and do `['"]` in the regex instead.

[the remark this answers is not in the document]
Bilka2: I can think of ways to break this new regexp, from more spaces to newlines. But I don't think rewriting this regexp until I run out of obvious ways to break it will get us anywhere actually safe, that's why I suggested using a lib for it. The regexp will never quite protect from the case of some place accidentally doing `sanitize_field(strip_images())` (perhaps hidden behind or separated by steps of abstraction) instead of using the arg or the correct order. Or one of the myriad of other ways unsanitized content could end up in strip_images (like in tests) while still appearing safe at first glance.

But I see that the dynamic config option means it can't happen during the sanitize step and tossing the whole thing into a new nokogiri document for a second time in strip_images isn't a great solution either. Furthermore, this PR is not the right place to rework comments to store unsanitized content and only sanitize (and strip images during sanitization) for display.

So, I'd say to make a regexp that's readable/feels good to you and only secondary that it handles unexpected input. And then in the integration tests where things _should_ be sanitized beforehand, use an img tag that specifically would break the regexp if things run in the wrong order. That should be slightly more protection against the "big yikes" than the current test setup. Not good, but better 😭

spec/mailers/admin_mailer_spec.rb
```diff
@@ -10,15 +10,33 @@
...
+        let(:image_tag) { "<img src=\"#{image_url}\" />" }
 
         before do
           comment.comment_content += image_tag
           comment.save!
         end
 
-        it "strips the image from the email message" do
-          expect(email).not_to have_html_part_content(image_tag)
+        context "when image safety mode is enabled for admin post comments" do
```
Bilka2, line 21: This context should check text content, like its counterpart for `#edited_comment_notification` below.

features/comments_and_kudos/image_safety_mode.feature
```diff
@@ -0,0 +1,62 @@
...
+  In order to protect users
+  As a site owner
+  I'd like to control which comments can include images
+
+  Scenario Outline: Images are embedded in comments when image safety mode is off.
+    Given the setup for testing image safety mode on <commentable>
+      And image safety mode is disabled for comments
+    When I view <commentable> with comments
+    Then I should see the image "src" text "https://example.com/image.jpg"
+      And I should not see "OMG! https://example.com/image.jpg"
```
Bilka2, line 11: The test added to add_comment.feature in #4729 can be removed, because this tests the same.

Bilka2 (approved):
Three minor missed things in the tests, rest looks good!

brianjaustin (approved):

PR #4801, opened 2024-04-28
AO3-6565 Move admin user search to Elasticsearch, add past emails and username search 

weeklies:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6565
## Purpose

Move admin user search to Elasticsearch, add past emails and username search 

## Testing Instructions

Test admin user search in a variety of ways, including past emails and usernames. 

## References

https://otwarchive.atlassian.net/browse/AO3-6493


app/views/admin/admin_users/show.html.erb
```diff
@@ -35,6 +35,26 @@
       <dd><%= @user.invitation ? link_to(@user.invitation.id, invitation_path(@user.invitation.id)) : t(".info.no_invitation") %></dd>
       <dt><%= t(".info.role", count: @user.roles.size) %></dt>
       <dd><%= @user.roles.any? ? @user.roles.map { |role| t("activerecord.attributes.role.#{role.name}") }.to_sentence : t(".info.no_role") %></dd>
+      <% if policy(User).can_view_past? %>
+        <% past_logins = @user.historic_values("login") %>
+        <% if past_logins.present? %>
+          <dt><%= t(".info.past_logins", count: past_logins.size) %></dt>
+          <dd>
+            <ul>
+              <li><%= past_logins.join(", ") %></li>
```
sarken, line 44: Can we use `to_sentence` for this and emails instead? iirc, it's easier to localize than hard coding the comma with `join`. I'm also not sure the `<ul>` and `<li>` are necessary, since it's not like each login/email is its own item.

app/views/admin/admin_users/index.html.erb
```diff
@@ -22,18 +22,24 @@
...
-      </dd>
+      <% if policy(User).can_view_past? %>
+        <dt><%= label_tag "settings", t(".settings.label") %></dt>
+        <dd>
+          <ul>
+              <li>
+                <%= check_box_tag "search_past", "1", params[:search_past].present? %>
+                <%= label_tag "search_past", t(".settings.past") %>
+              </li>
+          </ul>
```
sarken, line 40: I'm not sure we need a `<ul>` here -- there's no list.

sarken (commented):
Just two small front end comments -- feel free to lave them until further review is done!

features/admins/users/admin_manage_users.feature
```diff
@@ -85,6 +85,19 @@ Feature: Admin Actions to manage users
...
 
+  Scenario: Admins can view past emails and usernames
+    Given the following activated user exists
+      | login | email      |
+      | cats  | d@fake.com |
+      And I am logged in as "cats"
+      And I want to edit my profile
+      And I change my email
+      And I change my username to "new_user"
+    When I am logged in as a super admin
```
brianjaustin, line 96: Could you add a test for role(s) that don't see this?

features/admins/users/admin_find_users.feature
```diff
@@ -11,69 +11,98 @@ Feature: Admin Find Users page
...
       And I submit
     Then I should see "0 users found"
     When I fill in "Name" with "userA"
       And I submit
     Then the field labeled "user_email" should contain "a@ao3.org"
       But I should not see "UserB"
 
-  Scenario: The Find Users page should perform a partial match by email
-    When I fill in "Email" with "bo3"
+  Scenario: The Find Users page searches past logins only if the option is selected
```
brianjaustin, line 30: Could you add a test (not necessarily here, but I think it needs to be a feature test) for admins without the right role(s) not having the checkbox? And a test that if someone passes it anyway, the option is ignored? (I think the second would be easier as a controller test.)

brianjaustin (changes requested):

features/admins/users/admin_find_users.feature
weeklies: The first suggestion has been actioned. I took a look at `admin_users_controller_spec.rb` but couldn't see a way to implement your second suggestion.

Bilka2: First component is to add a config.before and config.after for the UserIndexer in `spec/spec_helper.rb`.

Then, for the controller test, here is an ugly implementation of the test. It should have less repetition than this and probably test more roles, but it works as an example:

 ```rb
     context "when passing search_past", user_search: true do
      let(:user) { create(:user, login: "old") }

      context "as an authorized admin" do
        let(:admin) { create(:superadmin) }
        it "is happy" do
          fake_login_admin(admin)
          user.update!(login: "new")
          run_all_indexing_jobs
          get :index, params: {name: "old", commit: "Find", search_past:"1"}
          expect(assigns(:users)).to include(user)
        end
      end

      context "as an unauthorized admin" do
        let(:admin) { create(:admin, roles: ["legal"] ) }
        it "is unhappy" do
          fake_login_admin(admin)
          user.update!(login: "new")
          run_all_indexing_jobs
          expect do
            get :index, params: {name: "old", commit: "Find", search_past:"1"}
          end.to raise_exception(ActionController::UnpermittedParameters)
        end
      end
    end
```

Bilka2:
The checkbox escaped containment

Screenshot_20250124_215315

features/admins/users/admin_find_users.feature
weeklies: Thanks!

brianjaustin (approved):
Thanks! When this is merged, could you update https://github.com/otwcode/otwarchive/wiki/Development-Data#resque---elasticsearch-indexes-reading-history-and-more to add the new index task?

Bilka2:
Speaking of index task, `index_admin_users` also needs to be added to https://github.com/otwcode/otwarchive/blob/master/script/reset_database.sh

Bilka2:
Noting here that merging this is blocked by getting [new servers](https://archiveofourown.org/admin_posts/31204), due to this querying the very large audit table, especially when initially creating the index.

PR #4802, opened 2024-04-28
AO3-6493 Notify wrangling supervisors via email when a wrangler changes their username

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6493

## Purpose

Create a Tag Wrangling Supervisor mailer and add a mail to the Tag Wrangler Supervisors when a tag wrangler changes their username.

## Testing Instructions

See Jira.

## Credit
Bilka (he/him)


config/config.yml
```diff
@@ -34,6 +34,7 @@ RETURN_ADDRESS: 'do-not-reply@example.org'
 SPAM_ALERT_ADDRESS: 'abuse-discuss@example.org'
 SPAM_THRESHOLD: 15
 ADMIN_ADDRESS: 'admin@example.org'
+TAG_WRANGLER_SUPERVISORS_ADDRESS: 'tagwranglers-personnel@transformativeworks.org'
```
Bilka2, line 37: Should this also have `@example.org` like the other default addresses?

sarken: Yes please! We prefer placeholder values in the config file.

test/mailers/previews/tag_wrangling_supervisor_mailer_preview.rb
```diff
@@ -0,0 +1,6 @@
+class TagWranglingSupervisorMailerPreview < ApplicationMailerPreview
+  # Sent to tag wrangling supervisors when a tag wrangler changes their username
+  def wrangler_username_change_notification
+    TagWranglingSupervisorMailer.wrangler_username_change_notification("stack inserter", "bulk inserter")
```
weeklies, line 4: Not a big deal, but spaces aren't allowed in usernames.

config/locales/mailers/en.yml
```diff
@@ -74,6 +74,9 @@ en:
           addressed: Hi, %{name}!
           unaddressed: Hi!
         introductory: Hello from the Archive of Our Own!
+        tag_wrangling_supervisor:
+          html: Hello %{tag_wrangler_supervisors},
```
Bilka2, line 78: Would it be okay to use the standard informal greeting instead? That would be `Hi, Tag Wrangler Supervisors!`. I didn't want to make the email different from the Jira issue, but it seems a bit awkward to have two so similar greetings.

sarken: I checked with TW and the standard one is indeed fine! They also send their thanks for working on this; they've very excited.

brianjaustin (approved):

PR #4805, opened 2024-05-02
AO3-6700 Have the New Orphan form more accurately describe the information that Orphaning removes

smclairecarden:
# Pull Request Checklist

* [ X] Have you read ["How to write the perfect pull request"](https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/)?
* [ X] Have you read the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [ ] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)? - I dont think this one needs tests? Let me know.
* [ X] Have you added the [Jira](https://otwarchive.atlassian.net) issue number
* [ X] Do you have fewer than 5 pull requests already open? If not, please wait

## Issue

https://otwarchive.atlassian.net/browse/AO3-6700

## Purpose

What does this PR do?

changes the text on the orphan form.

## Testing Instructions

see Jira!

## References

## Credit

Claire C she/her

they/them pronouns.*


config/locales/views/en.yml
```diff
@@ -952,4 +956,4 @@ en:
         revealed_html: A translation of %{work_link} by %{creator_link}
         unrevealed: A translation of a work in an unrevealed collection
     work_module:
-      draft_deletion_notice_html: This draft will be <strong>scheduled for deletion</strong> on %{deletion_date}.
+      draft_deletion_notice_html: This draft will be <strong>scheduled for deletion</strong> on %{deletion_date}
```
smclairecarden, line 959: Just now noticing this... I ran `docker compose exec web bundle exec i18n-tasks normalize -l en #` and this probably happened during that. I can remove this change if you'd like.

Bilka2: That's strange. I think you should remove the change, i18n-tasks should hopefully still accept the file as normalized.

config/locales/views/en.yml
```diff
@@ -678,6 +678,10 @@ en:
...
+  orphans:
+    orphan_user:
+      orphaning_bylines_only_message_html: Orphaning will automatically remove your personal information from <strong>bylines only.</strong> It will <strong>not</strong> automatically remove your personal information from anywhere else in the work(s). Social media accounts, contact information, email addresses, names, and any other personal information posted in the title, summary, notes, tags, work body, or comment text will <strong>not</strong> be automatically removed. Comments posted by other users will also <strong>not</strong> be affected by Orphaning. If you want information removed from these places, you should edit or delete it prior to Orphaning. <strong>Once you Orphan the work(s), you will no longer be able to delete information in these places yourself.</strong>
+      orphaning_works_message_html: 'Orphaning all of your works will <strong>permanently</strong> remove your username and/or pseud from the bylines of: the following work(s), their chapters, associated series, and any comments you may have left on them while logged into this account.'
```
Bilka2, line 684: Per Jira this should be:
 ```suggestion
       orphaning_works_message_html: 'Orphaning will <strong>permanently</strong> remove your username and/or pseud from the bylines of: the following work(s), their chapters, associated series, and any comments you may have left on them while logged into this account.'
```

File: config/locales/views/en.yml (hunk 1 above)
Comment (smclairecarden) [reply config/locales/views/en.yml]: Great, I added the period back in.

File: config/locales/views/en.yml (hunk 2 above)
Comment (smclairecarden) [reply config/locales/views/en.yml]: Good catch, thank you. Addressed.

**Comment by Bilka2** [comment]
The text looks good. However, maybe all of the HTML strong tags should be pulled out of the locale file as it's preferred not to have HTML in the locale. But that would pull apart the text a lot and naming three different locale keys for three "not"s (etc) would be difficult. Maybe @sarken knows better.

**Comment by sarken** [comment]
Hmm. So I think in this case, leaving the formatting in the locale file is the least bad choice, especially since it's simple `strong` tags, unlike the emails which have `style` attributes on those tags. 

Relatedly, I suspect Translation will ask us to revise this text when the time comes for it to be translated. I think they prefer us to avoid things like "work(s)" because not all languages can do that sort of ambiguity about whether there's one or many of something.

File: config/locales/views/en.yml [hunk 3]
```diff
@@ -678,6 +678,10 @@ en:
           hide_content_for_others: hide their works, series, bookmarks, and comments from anyone else
           intro: 'Muting a user will not:'
           prevent_emails: prevent you from receiving comment or subscription emails from this user
+  orphans:
+    orphan_user:
+      orphaning_bylines_only_message_html: Orphaning will automatically remove your personal information from <strong>bylines only.</strong> It will <strong>not</strong> automatically remove your personal information from anywhere else in the work(s). Social media accounts, contact information, email addresses, names, and any other personal information posted in the title, summary, notes, tags, work body, or comment text will <strong>not</strong> be automatically removed. Comments posted by other users will also <strong>not</strong> be affected by Orphaning. If you want information removed from these places, you should edit or delete it prior to Orphaning. <strong>Once you Orphan the work(s), you will no longer be able to delete information in these places yourself.</strong>
```
sarken, line 683: Could you move the period in the first sentence outside the `strong` tag, i.e., `<strong>bylines only</strong>.`?

smclairecarden: Yes for sure. What do you recommend for work(s)? We could cut it off entirely, end the sentence after "anywhere else.". Or just make works permanently plural? Whatever you think!

sarken: We can keep `work(s)` as is for now -- it'll be a while before this page is actually translated, and we might want other revisions to the text by then anyway!

smclairecarden: Awesome thank you, just wanted to make sure. Feedback addressed!

sarken (approved):

PR #4806, opened 2024-05-05
AO3-6612 AO3-6710 Extend creatorship mailer previews and update creatorship_notification i18n

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6612
https://otwarchive.atlassian.net/browse/AO3-6710

## Purpose

Adds email previews for all allowed creation types for `creatorship_notification`, `creatorship_notification_archivist` and `creatorship_request`.

Remove the work related English locale for `creatorship_notification` because it is never used in normal operation. I did not touch the non-English locale files, however if it is desired I could also remove the locale there.

The previews are separate for each creation type. This results in the following list of previews:
* creatorship_notification_archivist_chapter
* creatorship_notification_archivist_series
* creatorship_notification_archivist_work
* creatorship_request_chapter
* creatorship_request_series
* creatorship_request_work
* creatorship_notification_chapter
* creatorship_notification_series

## Testing Instructions

See Jira.

## Credit

Bilka (he/him)


test/mailers/previews/user_mailer_preview.rb
```diff
@@ -37,11 +40,12 @@ def claim_notification_registered
...
 
-  def creatorship_notification_data
+  def creatorship_notification_data(allowed_creations, creation_type)
     first_creator = create(:user, login: "JayceHexmaster")
     second_creator = create(:user, login: "viktor_the_machine")
-    work = create(:work, authors: [first_creator.default_pseud, second_creator.default_pseud])
-    second_creatorship = Creatorship.find_by(creation: work, pseud: second_creator.default_pseud)
-    [second_creatorship, first_creator]
+
+    creation_factory = allowed_creations.include?(creation_type) ? creation_type : :series
```
Bilka2, line 47: I am not super happy with silently ignoring invalid creation types like this, especially because it will be confusing for the `?creation=work` case, where it shows a series for `creatorship_notification`, but a work for the other mailers.

An option would be to raise an error instead. However, if an error is raised here it will show up as an Error 500, and on staging this gives no further information on what is wrong. They would look the same as the ActiveRecord::Deadlocked errors that pop up when many previews are opened at once. So, also confusing.

Furthermore, the URL parameters and their options are not shown on the `/rails/mailers` index. So the Translation team will need extra documentation for the available parameters.

Due the above, I am leaning towards making separate methods for each mailer and not using URL parameters.

sarken: I talked to Translation about this, and they're generally happy to use parameters. The previews will be used by a script they already document heavily, not individual translators, so having things listed on `/rails/mailers` isn't a major concern, and neither is "a translator enters invalid data for the params." Parameters will be particularly useful for things like letting them set the number of guest and/or user kudos to include in previews of kudos emails. 

That being said, these particular emails have a finite number of variations, so separate methods sounds like a good approach here! 

We talked a little bit about naming them, and we'd like to ensure the mailer previews match the mailer key as closely as possible (e.g., the mailer preview that is currently called `abuse_report_response` should be `abuse_report`), with any extra information about the variation tacked on to the end.

So here we'd want:
- creatorship_notification_archivist_chapter
- creatorship_notification_archivist_series
- creatorship_notification_archivist_work
- creatorship_notification_chapter
- creatorship_notification_series
- creatorship_request_chapter
- creatorship_request_series
- creatorship_request_work

test/mailers/previews/user_mailer_preview.rb
```diff
@@ -5,22 +5,26 @@ def abuse_report_response
...
-    second_creatorship, first_creator = creatorship_notification_data
-    UserMailer.creatorship_notification(second_creatorship.id, first_creator.id)
-  end
+    # Sends email when a user is added as an unapproved/pending co-creator
+    define_method :"creatorship_request_#{creation_type}" do
+      second_creatorship, first_creator = creatorship_notification_data(creation_type)
+      UserMailer.creatorship_request(second_creatorship.id, first_creator.id)
+    end
+
+    next if creation_type == :work
```
ceithir, line 21: Probably worth adding a comment clarifying the why of this next. Can be a simple link to https://otwarchive.atlassian.net/browse/AO3-6710

ceithir (approved):

test/mailers/previews/user_mailer_preview.rb
Bilka2: Good point, added.

PR #4807, opened 2024-05-06
AO3-5578 Use ActiveStorage for existing image uploads

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5578

## Purpose

Migrate existing places that use kt-paperclip to ActiveStorage (pseud, collection, and skin icons)

## Credit

Brian Austin (they/he)


brianjaustin:
Because ActiveStorage doesn't support prefixes within buckets well, we'll need to update the Ansible local config to separate things out IMO that's better anyways, since we currently co-mingle some production and non-production things... :grimacing: There's probably a way to do this with less friction, but I think this should be OK since we're copying things over

app/views/challenge_claims/_unposted_claim_blurb.html.erb
```diff
@@ -35,7 +35,7 @@
     </p>
 
     <div class="icon">
-      <%= image_tag(claim.collection.icon.url(:standard), :size => "100x100", :alt => claim.collection.icon_alt_text, :class => "icon", skip_pipeline: true) %>
+      <%= image_tag(standard_icon_url(claim.collection), size: "100x100", alt: claim.collection.icon_alt_text, class: "icon", skip_pipeline: true) %>
```
Bilka2, line 38: Because this is always the same image tag, would it make sense pull that into the method? Similar to the other image_tag helper methods that this PR modifies.

Okay well, the two in the form and header don't have the icon class. But that looks more like a bug.

features/step_definitions/icon_steps.rb
```diff
@@ -42,7 +42,7 @@
 
 Then /^the "([^"]*)" collection should have an icon$/ do |title|
   collection = Collection.find_by(title: title)
-  assert !collection.icon_file_name.blank?
+  assert collection.icon.attached?
```
Bilka2, line 45: Same change needed in line 50.

Bilka2 (commented):
I had a look at all the code except for the rake task.

The fixtures in `test/fixtures/pseuds.yml` have (empty) lines for `icon_file_name`, `icon_content_type`, `icon_updated_at` and `icon_file_size` that should probably be removed.

I think there should be a followup Jira issue to remove the `icon_file_name`, `icon_content_type`, `icon_updated_at` and `icon_file_size` columns from the database for the affected classes.

app/helpers/users_helper.rb
```diff
@@ -21,21 +21,17 @@ def print_pseuds(user)
...
-  def standard_icon(user = nil, pseud = nil)
-    if pseud && pseud.icon
-      pseud.icon.url(:standard).gsub(/^http:/, "https:")
-    elsif user && user.default_pseud && user.default_pseud.icon
-      user.default_pseud.icon.url(:standard).gsub(/^http:/, "https:")
-    else
-      '/images/skins/iconsets/default/icon_user.png'
-    end
+  def standard_icon(pseud = nil)
+    return pseud.icon.variant(:standard).processed.url if pseud&.icon&.attached?
```
Bilka2, line 25: Oops, missed this comment from the review.

The second safe navigation looks unneeded.
```suggestion
    return pseud.icon.variant(:standard).processed.url if pseud&.icon.attached?
```

Also, this method is arranged differently than the one in collections_helper. It will probably make our lives easier in the future if they're both arranged in the same order.

app/helpers/skins_helper.rb
```diff
@@ -9,9 +9,13 @@ def skin_author_link(skin)
 
   # we only actually display an image if there's a file
   def skin_preview_display(skin)
-    if skin && skin.icon_file_name
-      link_to image_tag(skin.icon.url(:standard), alt: skin.icon_alt_text, class: "icon", skip_pipeline: true), skin.icon.url(:original)
-    end
+    return unless skin&.icon&.attached?
```
Bilka2, line 12:
```suggestion
    return unless skin&.icon.attached?
```

app/models/collection.rb
```diff
@@ -2,16 +2,16 @@ class Collection < ApplicationRecord
...
 
-  has_attached_file :icon,
-  styles: { standard: "100x100>" },
-  url: "/system/:class/:attachment/:id/:style/:basename.:extension",
-  path: %w(staging production).include?(Rails.env) ? ":class/:attachment/:id/:style.:extension" : ":rails_root/public:url",
-  storage: %w(staging production).include?(Rails.env) ? :s3 : :filesystem,
-  s3_protocol: "https",
-  default_url: "/images/skins/iconsets/default/icon_collection.png"
+  has_one_attached :icon do |attachable|
+    attachable.variant(:standard, resize_to_fill: [100, nil])
```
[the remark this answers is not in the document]
brianjaustin: It's been a bit since I wrote this particular bit, but I vaguely remember an SO answer that was related to the trailing `>`. But according to https://stackoverflow.com/a/26272192 (and a couple other answers) whatever I interpreted wasn't right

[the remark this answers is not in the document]
Bilka2: Thanks for the link!

That doc raises another question: Whether `resize_to_fill` is the correct method. Based on the [description in ImageMagick](http://www.imagemagick.org/script/command-line-processing.php#geometry):
>  Use > to shrink an image only if its dimension(s) are larger than the corresponding width and/or height arguments.

I think the method to keep this behaviour would be [resize_to_limit](https://github.com/janko/image_processing/blob/master/doc/vips.md#resize_to_limit):

> Downsizes the image to fit within the specified dimensions while retaining the original aspect ratio. Will only resize the image if it's larger than the specified dimensions.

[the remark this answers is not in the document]
brianjaustin: Good call! (And thanks for the docs link, I've had trouble tracking down robust documentation for the new image resizing implementation)

Bilka2 (changes requested):
A few final comments. The rake task looks good, as does the rest of the code.

I also tried a few things in local dev, including looking at various icons in various places, uploading a pseud icon and using that to test the validator and resizing. Resizing acts the same as on production.

app/controllers/collections_controller.rb
```diff
@@ -44,7 +56,7 @@ def index
       params[:sort_column] = "collections.created_at" if !valid_sort_column(params[:sort_column], 'collection')
       params[:sort_direction] = 'DESC' if !valid_sort_direction(params[:sort_direction])
       sort = params[:sort_column] + " " + params[:sort_direction]
-      @collections = Collection.sorted_and_filtered(sort, params[:collection_filters], params[:page])
+      @collections = Collection.sorted_and_filtered(sort, params[:collection_filters], params[:page]).with_attached_icon
```
Bilka2, line 59: The `with_attached_icon` needs to be inside the `sorted_and_filtered` method.

spec/requests/blocked_users_n_plus_one_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+        proc do
+          get user_blocked_users_path(user_id: blocker)
+        end
+      end
+
+      warmup { subject.call }
+
+      it "produces a constant number of queries" do
+        expect { subject.call }
+          .to perform_constant_number_of_queries
```
Bilka2, line 33: This needs the `with_attached_*` here: https://github.com/otwcode/otwarchive/blob/1588f6c552b1c85adeb1b30e31411cb2cb6f37d7/app/controllers/blocked/users_controller.rb#L14

spec/requests/muted_users_n_plus_one_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+        proc do
+          get user_muted_users_path(user_id: muter)
+        end
+      end
+
+      warmup { subject.call }
+
+      it "produces a constant number of queries" do
+        expect { subject.call }
+          .to perform_constant_number_of_queries
```
Bilka2, line 33: This needs the `with_attached_*` here: https://github.com/otwcode/otwarchive/blob/1588f6c552b1c85adeb1b30e31411cb2cb6f37d7/app/controllers/muted/users_controller.rb#L14

spec/requests/people_n_plus_one_spec.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+describe "n+1 queries in the people controller" do
+  describe "#search", n_plus_one: true do
```
Bilka2, line 6: This is testing `#index`, not `#search`. It would be nice to have both.

And then search needs the `with_attached_*` here: https://github.com/otwcode/otwarchive/blob/1588f6c552b1c85adeb1b30e31411cb2cb6f37d7/app/controllers/people_controller.rb#L11

(Not quite sure about the exact line number.)

Bilka2 (commented):
I think I found how to fix the tests.

app/controllers/collections_controller.rb
```diff
@@ -25,11 +25,23 @@ def load_collection_from_id
 
   def index
     if params[:work_id] && (@work = Work.find_by(id: params[:work_id]))
-      @collections = @work.approved_collections.by_title.includes(:parent, :moderators, :children, :collection_preference, owners: [:user]).paginate(page: params[:page])
+      @collections = @work.approved_collections
+        .by_title
+        .includes(:parent, :moderators, :children, :collection_preference, owners: [:user])
```
Bilka2, line 30: Considering that all three includes are the same here, maybe a `for_blurb` scope would be nice?

app/controllers/collections_controller.rb
```diff
@@ -25,11 +25,23 @@ def load_collection_from_id
 
   def index
```
brianjaustin, line 26: I ended up deleting the N+1 spec for this method, as it was flagging something unrelated and we want to move collections to Elasticsearch anyways (also I'd like to keep the PR from getting too much out of hand :see_no_evil:)

app/models/comment.rb
```diff
@@ -85,7 +85,7 @@ def check_for_spam
     includes(
       pseud: { user: [:roles, :block_of_current_user, :block_by_current_user, :preference] },
       parent: { work: [:pseuds, :users] }
-    )
+    ).merge(Pseud.with_attached_icon)
```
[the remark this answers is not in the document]
brianjaustin: I've fixed some of these, but to keep the PR manageable (it's already quite large), I'm going to skip
* Anything related to collections (we're moving them to Elasticsearch, so things involving them and challenges will likely change quite a bit)
* Inbox comments (since there's already a ticket to handle them)

spec/requests/people_n_plus_one_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+        expect { subject.call }
+          .to perform_constant_number_of_queries
+      end
+    end
+  end
+
+  describe "#search", n_plus_one: true, pseud_search: true do
+    context "when there are search results" do
+      populate do |n|
+        create_list(:pseud, n, name: "nplusone")
```
Bilka2, line 32: Based on `works_n_plus_one_spec`, the indexer needs to be prepared for testing:
```suggestion
        PseudIndexer.prepare_for_testing
        create_list(:pseud, n, name: "nplusone")
```

spec/requests/people_n_plus_one_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+
+      warmup { subject.call }
+
+      # TODO: AO3-6743, ideally tis would be a constant number of queries too.
+      # However, I'm only testing that we didn't add more with ActiveStorage
+      # now, as those changes are already very involved.
+      # - Brian Austin, June 2024
+      it "produces about 1 query per pseud" do
+        expect { subject.call }
+          .to perform_linear_number_of_queries(slope: 1)
```
Bilka2, line 50: Locally this passes for me even with `perform_constant_number_of_queries`.

Bilka2 (commented):
I looked over everything again and mostly have some comments on the scopes! Thank you for the work on the N+1 query prevention and specs, those were more annoying than I had expected.

spec/requests/people_n_plus_one_spec.rb
brianjaustin: That should already be happening because `pseud_search: true` results in https://github.com/otwcode/otwarchive/blob/1d2d228bfa5e5eb8803b38d4b33d56834467d44b/spec/spec_helper.rb#L98

app/controllers/blocked/users_controller.rb
```diff
@@ -12,7 +12,8 @@ class UsersController < ApplicationController
     # GET /users/:user_id/blocked/users
     def index
       @blocks = @user.blocks_as_blocker
-        .joins(:blocked).includes(blocked: :default_pseud)
+        .joins(:blocked)
+        .includes(blocked: { default_pseud: { icon_attachment: :blob }, pseuds: { icon_attachment: :blob } })
```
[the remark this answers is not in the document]
brianjaustin: Looks like we can lose the attachment, but _something_ is hitting pseuds there
```
       Expected to make the same number of queries, but got:
         31 for N=2
         32 for N=3
       Unmatched query numbers by tables:
         pseuds (SELECT): 7 != 8
```

spec/requests/pseuds_n_plus_one_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+    subject do
+      proc do
+        get user_pseuds_path(user_id: user)
+      end
+    end
+
+    warmup { subject.call }
+
+    # TODO: https://otwarchive.atlassian.net/browse/AO3-6738
+    xit "produces a constant number of queries" do
```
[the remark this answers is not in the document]
brianjaustin: Jira is the wrong one, I've removed that line so it's more accurate (the filter was wrong anyways, because it should've been `active_storage` not `active_record` -- that's why I removed it entirely, too much room for silliness IMO)

spec/requests/people_n_plus_one_spec.rb
brianjaustin: :thinking: it consistently fails for me... seeing what happens in CI, if it passes there (and maybe in Gitpod?) I'll leave it on and close that story

brianjaustin: Looks like Github actions also runs into what I'm seeing locally: https://github.com/otwcode/otwarchive/actions/runs/9726260721/job/26844510622

Bilka2: Ah sorry, I meant with the `pseud_search: true` removed and the `PseudIndexer.prepare_for_testing` added to exactly mimic what `works_n_plus_one_spec.rb` does. If I don't do those additional changes it also doesn't pass for me.

app/controllers/collections_controller.rb
```diff
@@ -44,7 +53,7 @@ def index
       params[:sort_column] = "collections.created_at" if !valid_sort_column(params[:sort_column], 'collection')
       params[:sort_direction] = 'DESC' if !valid_sort_direction(params[:sort_direction])
       sort = params[:sort_column] + " " + params[:sort_direction]
-      @collections = Collection.sorted_and_filtered(sort, params[:collection_filters], params[:page])
+      @collections = Collection.for_blurb.sorted_and_filtered(sort, params[:collection_filters], params[:page])
```
Bilka2, line 56: This scope shouldn't be here, it's already in `sorted_and_filtered`.

```suggestion
      @collections = Collection.sorted_and_filtered(sort, params[:collection_filters], params[:page])
```

app/controllers/muted/users_controller.rb
```diff
@@ -12,7 +12,8 @@ class UsersController < ApplicationController
     # GET /users/:user_id/muted/users
     def index
       @mutes = @user.mutes_as_muter
-        .joins(:muted).includes(muted: :default_pseud)
+        .joins(:muted)
+        .includes(muted: { default_pseud: { icon_attachment: :blob }, pseuds: { icon_attachment: :blob } })
```
Bilka2, line 16: To be the same as the blocked users:
```suggestion
        .includes(muted: [:pseuds, default_pseud: { icon_attachment: :blob }])
```

spec/requests/people_n_plus_one_spec.rb
brianjaustin: Ah got it, thanks! Looks like that was it!

app/controllers/blocked/users_controller.rb
```diff
@@ -12,7 +12,8 @@ class UsersController < ApplicationController
     # GET /users/:user_id/blocked/users
     def index
       @blocks = @user.blocks_as_blocker
-        .joins(:blocked).includes(blocked: :default_pseud)
+        .joins(:blocked)
+        .includes(blocked: { default_pseud: { icon_attachment: :blob }, pseuds: {} })
```
Bilka2, line 16: Slightly prettier syntax:
```suggestion
        .includes(blocked: [:pseuds, default_pseud: { icon_attachment: :blob }])
```

Bilka2 (commented):
Just some more nitpicks, then this is finally done!

app/controllers/blocked/users_controller.rb
brianjaustin: Ooh thanks, I was wondering if there was a good option for the mixed preloading

Bilka2 (approved):
:tada: :tada: :tada:

Bilka2:
You can remove the `imagemagick` installation from the automated tests, ActiveStorage replaces it with libvips.

Bilka2:
Also, regarding proxy requests, https://guides.rubyonrails.org/v7.0/active_storage_overview.html#putting-a-cdn-in-front-of-active-storage has a section on making the URL is the CDN host instead of the app host. Is that something we would want to add as well? I would assume no, since cloudflare is in front of the app, not the CDN.

brianjaustin:
> Also, regarding proxy requests, https://guides.rubyonrails.org/v7.0/active_storage_overview.html#putting-a-cdn-in-front-of-active-storage has a section on making the URL is the CDN host instead of the app host. Is that something we would want to add as well? I would assume no, since cloudflare is in front of the app, not the CDN.

I asked @AlbertSPedersen about that, and he said we can have CloudFlare reroute the request to either cache/hit S3 directly as long as it's using an ao3.org URL. So yeah, we don't need to do the extra CDN step in this case

Bilka2 (approved):

PR #4812, opened 2024-05-17
AO3-6724 Update volunteering/about information

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6724

## Purpose

I18n and update the text on the donate and about pages.

~~I resisted the urge to update all links to be https instead of http.~~ All links on the two pages have been updated to https.

## Testing Instructions

See Jira.

## Credit

Bilka (he/him)

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -52,6 +52,7 @@ class HomeController < ApplicationController
 
   # about
   def about
+    @page_subtitle = t(".page_title")
     render action: "about", layout: "application"
   end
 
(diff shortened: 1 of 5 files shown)


weeklies:
I personally think standardizing https and http, and the use or absence of `www` in transformariveworks.org links would be nice.

ceithir:
> I personally think standardizing https and http, and the use or absence of `www` in transformativeworks.org links would be nice.

:+1: 
Doubly so as HTTP requests to fanlore.org and transformativeworks.org appear to be 301 redirected to HTTPS.

Bilka2:
Okay, I updated all links to be https because I also found AO3-6070 which did that for the footer. For `transformativeworks.org`, I standardized to using the `www` because `transformativeworks.org` redirects to `www.transformativeworks.org`.

weeklies (approved):

PR #4813, opened 2024-05-17
AO3-5860 Error message when commenting as admin

ceithir:
Do the admin check first

# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5860

## Purpose

Follow-up of https://github.com/otwcode/otwarchive/pull/4378
Fix conflict with https://github.com/otwcode/otwarchive/pull/4556

~~I'm actually not sure this is the best solution. Moving first the admin restriction is far more consistent as admins can never comment whatever the circumstances. But I suspect some admins might have a workflow where they check for messages like "you can't add or edit comments on a hidden work".~~ -> Yeah, too many side effects.

For now, I've just updated the tests to match the current behavior. If that behavior is not the intended one, I'm not against a ticket detailing what is expected especially as it has been a long time since my last foray into this.

### DIFF
--- a/app/views/comments/_commentable.html.erb
+++ b/app/views/comments/_commentable.html.erb
@@ -79,7 +79,11 @@
   <%= flash_div :comment_error, :comment_notice %>
 
   <% commentable_parent = find_parent(commentable) %>
-  <% if AdminSetting.current.guest_comments_off? && guest? %>
+  <% if logged_in_as_admin? %>
+    <p class="notice">
+      <%= t(".logged_as_admin") %>
+    </p>
+  <% elsif AdminSetting.current.guest_comments_off? && guest? %>
     <p class="notice">
       <%= t(".guest_comments_disabled") %>
     </p>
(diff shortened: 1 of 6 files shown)


app/views/comments/_commentable.html.erb
```diff
@@ -79,7 +79,11 @@
   <%= flash_div :comment_error, :comment_notice %>
 
   <% commentable_parent = find_parent(commentable) %>
-  <% if AdminSetting.current.guest_comments_off? && guest? %>
+  <% if logged_in_as_admin? %>
```
Bilka2, line 82: Instead of moving this above the other checks, do you think it would make sense to adjust the anon commenting checks to check for admin login instead? Specifically I am thinking that the two `!logged_in?` in this elsif chain could be `guest?` instead, because that also checks the admin login.

The reasoning for my idea is that I imagine that admins would like to be able to see the effect of the comment setting on the work without always getting the "please log out" message. But this is just an idea, I am not sure what the best solution is.

ceithir: After sleeping on it, I would be in favor of a quick `Scope: Tests Only` PR right now to fix the build, and then an addendum to AO3-5860 or a new ticket altogether to rework the behavior (if need be).

Bilka2 (approved):

PR #4814, opened 2024-05-23
AO3-6729 Fix searching for otp: true

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6729

## Purpose

Separate the fandoms and relationships `includes` in the WorkIndexer into two separate method calls. If they are combined, the second set of tags ends up with tags of the first type in their return value. So, in this case the Fandom tags ended up in `relationships`. `relationships` is used in `#otp` to determine OTP status, so that gets messed up by the extra Fandom tag(s). 

I adjusted the existing N+1 spec that was added together with the `includes` in #4291 to make sure `#otp` hits the line that queries the [merger](https://github.com/otwcode/otwarchive/blob/47cc03880769f4256456598e94eed99a8d6e0faf/app/models/work.rb#L1251).

This preload going wrong is most likely a Rails bug, I found a few Rails 7.0+ issues related to preloading leading to wrongly returned records. However, I have not yet checked if this particular case was reported and/or fixed for a newer version of Rails. That would likely need a more isolated reproduction case. If anyone wants to have a go at it, the different tag types are accessed via the scoped `has_many` in [taggable.rb](https://github.com/otwcode/otwarchive/blob/47cc03880769f4256456598e94eed99a8d6e0faf/app/models/concerns/taggable.rb#L8-L15).

There are no other spots where `:merger` is preloaded.

## Testing Instructions

See Jira.

## Credit

Bilka
Sarken (for the otp search feature test)

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -1248,7 +1248,8 @@ class Work < ApplicationRecord
   def otp
     return true if relationships.size == 1
 
-    all_without_syns = relationships.map { |r| r.merger_id ? r.merger_id : r.id }.uniq.compact
+    all_without_syns = relationships.map { |r| r.merger_id || r.id }
+      .uniq
     all_without_syns.count == 1
   end
 
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -1248,7 +1248,7 @@ class Work < ApplicationRecord
   def otp
     return true if relationships.size == 1
 
-    all_without_syns = relationships.map { |r| r.merger ? r.merger : r }.uniq.compact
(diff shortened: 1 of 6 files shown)


app/models/work.rb
```diff
@@ -1248,7 +1248,8 @@ def crossover
   def otp
     return true if relationships.size == 1
 
-    all_without_syns = relationships.map { |r| r.merger ? r.merger : r }.uniq.compact
+    all_without_syns = relationships.map { |r| r.merger_id || r.id }
+      .uniq
```
Bilka2, line 1252: Removed the `.compact` here because neither `r` nor `r.id` can be nil.

brianjaustin (approved):

PR #4815, opened 2024-05-28
AO3-6017 Use unique names for users in mailer previews

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6017

## Purpose

Fix Error 422 "Validation failed: User name has already been taken" in some mailer previews on staging. This is done by using generated and thus hopefully unique user names when creating users for the previews.

## Testing Instructions

Visit https://test.archiveofourown.org/rails/mailers/admin_mailer/comment_notification_official and https://test.archiveofourown.org/rails/mailers/comment_mailer/comment_reply_sent_notification and make sure they display working emails, not "Error 422".

## Credit

Bilka (he/him)

### DIFF
--- a/factories/users.rb
+++ b/factories/users.rb
@@ -25,6 +25,11 @@ FactoryBot.define do
       confirmed_at { nil }
     end
 
+    # User names used in mailer preview should be unique but recognizable as user names
+    trait :for_mailer_preview do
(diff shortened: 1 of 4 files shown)


weeklies (commented):
I suggest creating a helper function with a comment on why this is necessary?

Bilka2:
Good point, I added a trait with a comment.

PR #4816, opened 2024-05-28
AO3-4728 Fix Error 500 on subscriptions page for invalid items

weeklies:
## Issue

## Purpose

Updates old #4150 and apply i18n on subscription index page. 

## Testing Instructions

See Jira.


app/views/subscriptions/index.html.erb
```diff
@@ -1,61 +1,67 @@
 <!--Descriptive page name, messages and instructions-->
 <h2 class="heading">
-  <%= params[:type].blank? ? ts('My Subscriptions') : 
-  ts('My %{type} Subscriptions', type: params[:type].to_s.singularize.titleize)
-  %>
+  <% if params[:type].blank? %>
+    <%= t(".page_heading.general") %>
+  <% else %>
+    <%= t(".page_heading.type", type: params[:type].to_s.singularize.titleize) %>
```
sarken, line 6: Oh, ouch, I can't believe we've been doing it like this. It's just taking the text from the parameters and using it on the page heading, so `https://archiveofourown.org/users/sarken/subscriptions?type=fakey%20fake` will create a page that says "My Fakey Fake Subscriptions." That's not great, but it really doesn't work once we factor in translations, because even if the parameters are valid, you're going to end up with nonsense that mixes English and the desired language, e.g., "Mes Work abonnements."

That's obviously a sign of a bigger issue -- we should be checking the parameters against a list in the controller, so entering something other than `?type=series`, `?type=works`, or `type=users` will just take you to the plain subscriptions page -- that's probably out of scope.

However, since we're adding i18n, we should at least fix this heading. That means we need to stop using the `type` in the string and use it in the key instead. However, before we do that, it also means making sure the `type` is one of our valid options. Maybe something like this:

```
<% type_leaf = if Subscription::VALID_SUBSCRIBABLES.include?(params[:type].to_s.singularize.titleize)
                params[:type].to_s
              else
                "all"
              end %>
<%= t(".page_heading.#{type_leaf}") %>
```

And then the locale file would be like
```
page_heading:
  all: My Subscriptions
  works: My Work Subscriptions
  etc
```

If it's neater/less work to start the fix at the controller level, though, that's fine!

sarken (changes requested):
Thanks for working on this! I'm sorry you inherited the issue I just noticed. 😬

app/views/subscriptions/index.html.erb
weeklies: I've tried to fix it in the controller and added some tests.

app/controllers/subscriptions_controller.rb
```diff
@@ -15,9 +15,12 @@ def load_user
   # GET /subscriptions.xml
   def index
     @subscriptions = @user.subscriptions.includes(:subscribable)
-    if params[:type].present?
+
+    if params[:type] && Subscription::VALID_SUBSCRIBABLES.include?(params[:type].singularize.titleize)
+      @subscribable_type = params[:type]
       @subscriptions = @subscriptions.where(subscribable_type: params[:type].classify)
```
sarken, line 21: Now that we have `@subscribable_type`, it would probably make sense to use that in place of `params[:type]` wherever possible (so here and in various places in the view file).

sarken (commented):
Yay, thank you, that's much better! Just one more comment that might make things a bit easier to follow and then I think we're set.

sarken (approved):
Thank you!

PR #4781, opened 2024-04-04
AO3-5506 Don't include hidden works in Readings

de3sw2aq1:
# Pull Request Checklist

* [x] Have you read ["How to write the perfect pull request"](https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/)?
* [x] Have you read the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [x] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)?
* [x] Have you added the [Jira](https://otwarchive.atlassian.net) issue number
  as the *first* thing in your pull request title (e.g. `AO3-1234 Fix thing`)
* [x] Do you have fewer than 5 pull requests already open? If not, please wait
  until they are reviewed and merged before creating new pull requests.

## Issue

https://otwarchive.atlassian.net/browse/AO3-5506

## Purpose

Don't include hidden works in reading history

* Require works to be visible to mark for later

* Hide draft works and hidden by admin works from the reading history

* Test for visibility of hidden by admin works in reading history

## Testing Instructions

An automated test for hidden by admin works not being visible in the reading list is included. This may be tested manually by adding a work to a users reading list, an admin hiding it, and then viewing the users reading list and confirming it's absence.

Accessing a URL for a draft work like http://localhost:3000/works/110/mark_for_later (manually replace 110 with the id of a draft work owned by another user) should now return the error "Sorry, you don't have permission to access the page you were trying to reach." This URL was never linked in the UI for draft works, but could be guessed and draft works could previously be added to the reading list.

I don't know how to test the fact that the reading list no longer shows draft works if they somehow get added to the reading list (the above `.../mark_for_later` action is now blocked), but they should no longer be visible in the reading list now.

## Credit

@de3sw2aq1

Thanks @sarken for initial review of the security aspect of this issue and the pointer to `visible_to_registered_user` in Jira was helpful.

PR #4851, opened 2024-06-02
AO3-5506 Don't include hidden works in Readings: fixed query

de3sw2aq1:
Move `.or()` inside `.merge()` to prevent showing deleted works from other user's reading lists showing on all users' reading lists.

# Pull Request Checklist

## Issue

Fix for bug in AO3-5506 implementation: https://otwarchive.atlassian.net/browse/AO3-5506?focusedCommentId=361327

## Purpose

https://github.com/otwcode/otwarchive/pull/4781/ excluded admin works and never published works from the reading list. 

However there was a bug. `.or()` needs to be moved inside `.merge()` to ensure the generated SQL is correct. Previously deleted works from other users' reading lists were being shown for all users.

Bad SQL:

```
SELECT `readings`.* FROM `readings` LEFT OUTER JOIN `works` ON `works`.`id` = `readings`.`work_id` WHERE (`readings`.`user_id` = 1 AND `works`.`posted` = TRUE AND `works`.`hidden_by_admin` = FALSE OR `works`.`id` IS NULL) ORDER BY last_viewed DESC LIMIT 25 OFFSET 0
```

Fixed SQL:

```
SELECT `readings`.* FROM `readings` LEFT OUTER JOIN `works` ON `works`.`id` = `readings`.`work_id` WHERE `readings`.`user_id` = 37 AND (`works`.`posted` = TRUE AND `works`.`hidden_by_admin` = FALSE OR `works`.`id` IS NULL) ORDER BY last_viewed DESC LIMIT 25 OFFSET 0
```

## Testing Instructions

1. Add a work to the reading list for user 1
2. Check that it shows "Deleted work" on the reading list for user 1
3. Log into a second new user 2, the reading list should be empty (no "Deleted work" should show)

## References

https://github.com/otwcode/otwarchive/pull/4781/

## Credit

@de3sw2aq1

### DIFF
--- a/app/controllers/readings_controller.rb
+++ b/app/controllers/readings_controller.rb
@@ -10,13 +10,13 @@ class ReadingsController < ApplicationController
   end
 
   def index
-    @readings = @user.readings
+    @readings = @user.readings.visible
     @page_subtitle = ts("History")
     if params[:show] == 'to-read'
       @readings = @readings.where(toread: true)
       @page_subtitle = ts("Marked For Later")
     end
-    @readings = @readings.left_joins(:work).merge(Work.visible_to_registered_user).or(Work.where(id: nil)).order("last_viewed DESC").page(params[:page])
+    @readings = @readings.last_viewed.page(params[:page])
   end
 
   def destroy
--- a/app/controllers/readings_controller.rb
+++ b/app/controllers/readings_controller.rb
@@ -16,7 +16,7 @@ class ReadingsController < ApplicationController
       @readings = @readings.where(toread: true)
       @page_subtitle = ts("Marked For Later")
     end
-    @readings = @readings.last_viewed.page(params[:page])
+    @readings = @readings.order("last_viewed DESC").page(params[:page])
   end
 
   def destroy
--- a/app/decorators/homepage.rb
+++ b/app/decorators/homepage.rb
@@ -46,13 +46,13 @@ class Homepage
   def readings
     return unless logged_in? && @user.preference.try(:history_enabled?)
     if Rails.env.development?
-      @readings ||= @user.readings.random_order.
+      @readings ||= @user.readings.visible.random_order.
           limit(ArchiveConfig.NUMBER_OF_ITEMS_VISIBLE_ON_HOMEPAGE).
           where(toread: true).
           all
     else
       @readings ||= Rails.cache.fetch("home/index/#{@user.id}/home_marked_for_later") do
-        @user.readings.random_order.
+        @user.readings.visible.random_order.
           limit(ArchiveConfig.NUMBER_OF_ITEMS_VISIBLE_ON_HOMEPAGE).
           where(toread: true).
           to_a
--- a/app/decorators/homepage.rb
+++ b/app/decorators/homepage.rb
@@ -15,7 +15,7 @@ class Homepage
     @fandom_count = Rails.cache.fetch("/v1/home/counts/fandom", expires_in: 40.minutes) do
       estimate_number(Fandom.canonical.count)
     end
-    return @user_count, @work_count, @fandom_count
+    [@user_count, @work_count, @fandom_count]
   end
 
   def logged_in?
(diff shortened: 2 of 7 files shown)


otwtranslation (approved):
Thanks!

sarken:
Whoops, sorry, that was me, and I was a bit premature. I think it would be worth adding a test for this, and probably the easiest way to handle this scenario is a controller spec like so:

```ruby
# Disclaimer: I can't run specs on dev at the moment, so this is just an approximation.
require "spec_helper"

describe ReadingsController do
  include LoginMacros
  let(:user) { create(:user) }

  describe "GET #index" do
    context "with user parameters" do
      context "when logged in as the user" do
        let(:my_reading) { create(:reading, user: user, work: create(:work)) }
        let(:my_reading_deleted) { create(:reader, user: user, work_id: 000) }
        let(:other_reading) { create(:reading, user: create(:user), work: create(:work)) }
        let(:other_reading_deleted) { create(:reading, user: create(:user), work_id: 000) }

        it "contains readings for that user" do
          fake_login_known_user(user)
          get :index, params: { user_id: user }
          expect(assigns(:readings)).to eq([my_reading, my_reading_deleted])
        end
      end
    end
  end
end
```

It might be necessary to define a factory for readings.

de3sw2aq1:
I've never actually written an RSpec test. I tried but `create(:work)` is somehow failing with an error:
```
ActiveRecord::RecordInvalid:
  Validation failed: Invalid creator: chpi8ue6 (cl30sfdn4) does not allow others to invite them to be a co-creator., Work must have at least one creator., Invalid chapter: Chapter must have at least one creator.
```

If you don't mind writing the test, I'd appreciate it.

de3sw2aq1:
There may actually be an issue with this whole approach of putting the filtering in `index`. I just remembered readings are on the home page too, and that accesses `@user.readings` directly:
https://github.com/otwcode/otwarchive/blob/5d710e6883c26d7782d14680d73e5307421a28e5/app/decorators/homepage.rb#L55

Is there some kind of rails feature that can be used to add required `.where()` conditions to `.readings`? Or should we add a helper somewhere and avoid calling `.readings`?

sarken:
>If you don't mind writing the test, I'd appreciate it.

Sure! I found the Redis issue that was preventing my tests from running, so I can finish the tests tomorrow. I see exactly what I did wrong. 😅 

>There may actually be an issue with this whole approach of putting the filtering in index. I just remembered readings are on the home page too, and that accesses @user.readings directly:

Good catch!

>Is there some kind of rails feature that can be used to add required .where() conditions to .readings? Or should we add a helper somewhere and avoid calling .readings?

The answer to this is "yes, but." Basically, yes, it's possible to set a [default_scope](https://apidock.com/rails/ActiveRecord/Scoping/Default/ClassMethods/default_scope). However, we very recently ran into an issue involving [using default_scope on bookmarks](https://otwarchive.atlassian.net/issues/AO3-6336), and it can cause confusion for anyone troubleshooting from the console, so I wouldn't really recommend it. 

I think we might be better off taking the same approach we use for inbox comments:

https://github.com/otwcode/otwarchive/blob/5d710e6883c26d7782d14680d73e5307421a28e5/app/models/inbox_comment.rb#L38-L42
https://github.com/otwcode/otwarchive/blob/5d710e6883c26d7782d14680d73e5307421a28e5/app/decorators/homepage.rb#L65
https://github.com/otwcode/otwarchive/blob/5d710e6883c26d7782d14680d73e5307421a28e5/app/controllers/inbox_controller.rb#L19

sarken:
[Here are some tests](https://github.com/otwcode/otwarchive/compare/master...sarken:otwarchive:AO3-5506)! You only need `spec/models/reading_spec.rb` if you define a method or scope as we discussed above. I haven't run these through a linter, so I apologize for whatever horrors await when you do.

de3sw2aq1:
Oh, you fixed... everything. Thanks. I applied rubocop auto corrections to `app/decorators/homepage.rb`, everything else passed the linter.

The new `readings` method looks good. I tested it all seems to work right. I can't find any other usages of `readings` other than the homepage and the reading list itself. I think that's everything then?

I'm unsure if you wanted to finish this in this PR or different PR? But for now I pulled your commits into this branch, I hope that's okay. I have "Allow edits and access to secrets by maintainers" enabled if you want to edit this PR.

sarken:
> Oh, you fixed... everything. Thanks.

Haha, yeah, no problem! I realized if I wanted to be sure the tests passed, I also needed to write the code, and it seemed silly to write it and then remove it just so you could redo it. 😆 Thanks for the code style fixes!

> I think that's everything then?

I hope so! I will ask someone who didn't stick their hands all over this code to finish reviewing this PR for us. If it needs more changes, whoever is available can push them to this branch. (I probably won't be around much until around midnight UTC on Monday.)

zz9pzza (approved):

PR #4854, opened 2024-06-07
AO3-5351 Extract feedback email to locales too

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5351 (Please fill in issue number and remove this comment.)

## Purpose

* Fix the subject of the email so it-s I18n-able
~* Wrap the send with a `I18n.with_locale` that uses the language in the ticket (if feature flags are enabled)~

~tagging @sarken to make sure we actually want to do the second thing~ per the below discussion, not yet at least


sarken:
> Wrap the send with a I18n.with_locale that uses the language in the ticket (if feature flags are enabled)

I'm checking with Support and PAC, since presumably both forms should function the same.

sarken:
> Wrap the send with a I18n.with_locale that uses the language in the ticket (if feature flags are enabled)

Let's not do this right now. There are differing opinions on the desired behavior, so it would be better to get everyone on the same page first, and I think it's going to involve tweaking the forms so they're clearer about what language users should select. (The `email_and_send` option for the PAC form doesn't have locales enabled yet, either.)

brianjaustin:
Fair enough! I've reverted the change so this is just the subject change; do we want a Jira ticket for the other things yet (even if just a placeholder until discussions are finalised)?

Bilka2 (approved):

sarken:
Nah, the placeholder doesn't get us anywhere, it's just something we have to hold on to to keep someone from taking and making up their own design.

PR #4338, opened 2022-09-14
AO3-6308 fix user suspension end time

Cesium-Ice:
# Pull Request Checklist

* [x] Have you read ["How to write the perfect pull request"](https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/)?
* [x] Have you read the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [x] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)?
* [x] Have you added the [Jira](https://otwarchive.atlassian.net) issue number
  as the *first* thing in your pull request title (e.g. `AO3-1234 Fix thing`)
* [x] Do you have fewer than 5 pull requests already open? If not, please wait
  until they are reviewed and merged before creating new pull requests.

## Issue

https://otwarchive.atlassian.net/browse/AO3-6308

## Purpose

What does this PR do?
Fix incorrect time for the end of a user's suspension

## Testing Instructions

How can the Archive's QA team verify that this is working as you intended?

If you have a Jira account with access, please update or comment on the issue
with any new or missing testing instructions instead.

## References

Are there other relevant issues/pull requests/mailing list discussions?

## Credit

What name and pronouns should we use to credit you in the [Archive of Our Own's Release Notes](https://archiveofourown.org/admin_posts?tag=1)?
Cesium-Ice, they/them

If you have a Jira account, please include the same name in the "Full name"
field on your Jira profile, so we can assign you the issues you're working on.

*Please note that if you do not fill in this section, we will use your GitHub account name and
they/them pronouns.*

PR #4855, opened 2024-06-09
AO3-6728 Delete comments properly for spammer bans

EchoEkhi:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6728

## Purpose

Fix "Spammer: ban and delete all creations" option for PAC admins and delete the comments properly

## Testing Instructions

See JIRA ticket.

## Credit

EchoEkhi (He/him)

### DIFF
--- a/app/controllers/admin/admin_users_controller.rb
+++ b/app/controllers/admin/admin_users_controller.rb
@@ -139,12 +139,23 @@ class Admin::AdminUsersController < Admin::BaseController
 
   def destroy_user_creations
     authorize @user
-    creations = @user.works + @user.bookmarks + @user.sole_owned_collections + @user.comments
+
+    creations = @user.works + @user.bookmarks + @user.sole_owned_collections
     creations.each do |creation|
       AdminActivity.log_action(current_admin, creation, action: "destroy spam", summary: creation.inspect)
       creation.mark_as_spam! if creation.respond_to?(:mark_as_spam!)
       creation.destroy
     end
+
+    # comments are special and needs to be handled separately
+    @user.comments.each do |comment|
+      AdminActivity.log_action(current_admin, comment, action: "destroy spam", summary: comment.inspect)
+      # Submit spam sample to Akismet if in production mode
+      # comment.mark_as_spam cannot be used here because it also sets :approved to false, which would hide the whole thread
+      Akismetor.submit_spam(akismet_attributes) if Rails.env.production?
+      comment.destroy_or_mark_deleted # comments with replies cannot be destroyed, mark deleted instead
+    end
+
     flash[:notice] = ts("All creations by user %{login} have been deleted.", login: @user.login)
     redirect_to(admin_users_path)
   end
--- a/app/controllers/admin/admin_users_controller.rb
+++ b/app/controllers/admin/admin_users_controller.rb
@@ -150,13 +150,11 @@ class Admin::AdminUsersController < Admin::BaseController
     # comments are special and needs to be handled separately
     @user.comments.each do |comment|
       AdminActivity.log_action(current_admin, comment, action: "destroy spam", summary: comment.inspect)
-      # Submit spam sample to Akismet if in production mode
-      # comment.mark_as_spam cannot be used here because it also sets :approved to false, which would hide the whole thread
-      Akismetor.submit_spam(akismet_attributes) if Rails.env.production?
+      # Akismet spam procedures are skipped, since logged-in comments aren't spam-checked anyways
       comment.destroy_or_mark_deleted # comments with replies cannot be destroyed, mark deleted instead
     end
 
-    flash[:notice] = ts("All creations by user %{login} have been deleted.", login: @user.login)
+    flash[:notice] = t(".success", login: @user.login)
     redirect_to(admin_users_path)
   end
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -389,11 +389,11 @@ public
   # Prevents banned and suspended users from adding/editing content
   def check_user_status
     if current_user.is_a?(User) && (current_user.suspended? || current_user.banned?)
-      if current_user.suspended?
-        flash[:error] = t("users.suspension_notice_html", contact_abuse_link: view_context.link_to(t("users.contact_abuse"), new_abuse_report_path), suspended_until: localize(current_user.suspended_until))
-      else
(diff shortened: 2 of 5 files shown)


app/controllers/application_controller.rb
EchoEkhi, line 1: The html_safe functions were placed at the wrong places, and the anchor tags were being unnecessarily escaped. Should I also add the _html suffix to the keys?

Bilka2: Yes and please move the locale into the yml file like Rubocop suggests and use the link_to helper to create the links, see https://github.com/otwcode/otwarchive/wiki/Internationalization-%28i18n%29-Standards#hyperlinks.

config/locales/controllers/en.yml
```diff
@@ -101,3 +101,6 @@ en:
           one: You may reset your password %{count} more time.
           other: You may reset your password %{count} more times.
         user_not_found: We couldn't find an account with that email address or username. Please try again.
+    suspension_notice_html: Your account has been suspended until %{suspended_until}. You may not add or edit content until your suspension has been resolved. Please <a href=\"#{new_abuse_report_path}\">contact Abuse</a> for more information.
```
Bilka2, line 104: Creating the link like this doesn't work:
Screenshot_20240609_153234

Please use the link_to helper:

```rb
flash[:error] = t("users.suspension_notice_html", suspended_until: localize(current_user.suspended_until), contact_abuse_link: view_context.link_to(t("users.contact_abuse"), new_abuse_report_path))
```

```suggestion
    suspension_notice_html: Your account has been suspended until %{suspended_until}. You may not add or edit content until your suspension has been resolved. Please %{contact_abuse_link} for more information.
    contact_abuse: contact Abuse
```

Same for ban_notice_html, and please remove the duplicate suspension_notice_html entry.

It would be great if you could also remove the following lines from i18n-tasks.yml because they will no longer be needed:
https://github.com/otwcode/otwarchive/blob/dd6ec55a93d549cef4e9a43269dc1cebb975dd91/config/i18n-tasks.yml#L126-L128

You can ignore the "should be" comments, the translation keys you chose are better. However, you could move the two `*_notice_html` keys to `users.status` to align with #4338.

EchoEkhi: Thanks, I was wondering how to use the link_to function in a controller.

app/controllers/admin/admin_users_controller.rb
```diff
@@ -139,12 +139,23 @@ def confirm_delete_user_creations
...
+    # comments are special and needs to be handled separately
+    @user.comments.each do |comment|
+      AdminActivity.log_action(current_admin, comment, action: "destroy spam", summary: comment.inspect)
+      # Submit spam sample to Akismet if in production mode
+      # comment.mark_as_spam cannot be used here because it also sets :approved to false, which would hide the whole thread
+      Akismetor.submit_spam(akismet_attributes) if Rails.env.production?
+      comment.destroy_or_mark_deleted # comments with replies cannot be destroyed, mark deleted instead
+    end
+
     flash[:notice] = ts("All creations by user %{login} have been deleted.", login: @user.login)
```
Bilka2, line 159: Entirely optional, but this issue is a good opportunity to get this converted to `t(".success", .....)` as well. The relative locale key should work here because the method has the same name as the action.

features/admins/users/admin_abuse_users.feature
```diff
@@ -183,6 +183,24 @@ Feature: Admin Abuse actions
...
+      And I press "Update"
+    Then I should see "permanently suspended"
+      And the user "Spamster" should be permanently banned
+      And I should see "I like spam"
+    When I press "Yes, Delete All Spammer Creations"
+    Then I should see "All creations by user Spamster have been deleted."
+    When I go to the work comments page for "Generic Work"
+    Then I should not see "I like spam"
+      And I should see "(Previous comment deleted.)"
+      And I should see "I don't :("
```
Bilka2, line 202: If you follow my suggestion regarding `mark_as_spam!` then these lines won't work. However, the reply should still be viewable with `When I view the latest comment`. The name of the Scenario should probably be updated to something like "accessible" since it won't be visible on the work itself.

features/admins/users/admin_abuse_users.feature
```diff
@@ -183,6 +183,24 @@ Feature: Admin Abuse actions
       And there should be no bookmarks on the work "Not Spam"
       And there should be no comments on the work "Not Spam"
 
+  Scenario: A permabanned spammer's comments' replies from others should stay visible
+    Given I have a work "Generic Work"
+      And a comment "I like spam" by "Spamster" on the work "Generic Work"
+      And a reply "I don't :(" by "NotSpamster" on the work "Generic Work"
```
Bilka2, line 189: Could you make a second comment thread with a comment from Spamster and a reply from Spamster to test the Error 500 from the Jira issue? It occurs when deleting the creations, so it seems convenient to cover it in the same scenario.

Bilka2 (commented):
Thanks for working on the locale! I made a note on the behaviour of marking the comment as spam and then more review comments cascaded from there.

features/admins/users/admin_abuse_users.feature
Bilka2: Note: To make this work with the "I view the latest comment" step, the spammers comment thread would have to be the _first_ comment thread on the work.

app/controllers/admin/admin_users_controller.rb
```diff
@@ -139,12 +139,23 @@ def confirm_delete_user_creations
...
       creation.mark_as_spam! if creation.respond_to?(:mark_as_spam!)
       creation.destroy
     end
+
+    # comments are special and needs to be handled separately
+    @user.comments.each do |comment|
+      AdminActivity.log_action(current_admin, comment, action: "destroy spam", summary: comment.inspect)
+      # Submit spam sample to Akismet if in production mode
+      # comment.mark_as_spam cannot be used here because it also sets :approved to false, which would hide the whole thread
+      Akismetor.submit_spam(akismet_attributes) if Rails.env.production?
```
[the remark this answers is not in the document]
sarken: I would honestly suggest skipping `mark_as_spam!` for comments. We don't run spam checks on logged-in comments, so it's weird and not necessarily helpful (logged-in spam and logged-out spam don't generally look the same) that we were ever doing it in this case.

PR #4790, opened 2024-04-15
AO3-6713 Enable comment moderation for admin posts

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6713

## Purpose

Adds comment moderation for admin posts. 

Dreadfully sorry about not having the issue number in any of the commits; I merged with master before I thought to fix it. At least we squash now?

## Testing Instructions

Refer to Jira.

## Credit

Sarken, she/her

PR #4859, opened 2024-06-18
AO3-6739 Display moderated commenting notice on individual chapters

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6739

## Purpose

If you accessed at a work with moderated commenting enabled in chapter-by-chapter mode, it said "Chapter" (the i18n key, which was wrong -- there's not a dedicated notice for chapters) instead of "This work's creator has chosen to moderate comments on the work. Your comment will not appear until it has been approved by the creator." 

Now it uses the right key and says the big long thing.

## Testing Instructions

Refer to Jira.

## References

https://github.com/otwcode/otwarchive/pull/4790

### DIFF
--- a/app/views/comments/_comment_form.html.erb
+++ b/app/views/comments/_comment_form.html.erb
@@ -28,10 +28,11 @@
 
       <% if comments_are_moderated(commentable) && !current_user_is_work_creator(commentable) %>
         <p class="notice">
-          <%# i18n-tasks-use
-              t("comments.commentable.permissions.moderated_commenting.notice.admin_post")
-              t("comments.commentable.permissions.moderated_commenting.notice.work") %>
-          <%= t("comments.commentable.permissions.moderated_commenting.notice.#{commentable.model_name.i18n_key}") %>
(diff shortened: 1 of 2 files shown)


Bilka2:
Are the other spots that use `commentable.model_name.i18n_key` all fine or can they also be per chapter? https://github.com/otwcode/otwarchive/blob/231c34e0537f6a75bda99860693ed1780f77392d/app/views/comments/unreviewed.html.erb#L15 https://github.com/otwcode/otwarchive/blob/231c34e0537f6a75bda99860693ed1780f77392d/app/controllers/comments_controller.rb#L349

sarken:
It looks like they're fine. There's no unreviewed comments page for individual chapters (it 404s if you do, e.g., `https://test.archiveofourown.org/works/1089709/chapters/2182327/comments/unreviewed`) and the `ultimate_parent` is always either a work, admin post, or tag (I initially wanted to use `ultimate_parent.model_name.i18n_key` for this, but that didn't work quite as nicely as I'd hoped).

Bilka2 (approved):

PR #4863, opened 2024-06-26
AO3-6714 Prevent tag sets becoming ownerless

tommilligan:
# Pull Request Checklist

* [x] Have you read ["How to write the perfect pull request"](https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/)?
* [x] Have you read the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [x] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)?
* [x] Have you added the [Jira](https://otwarchive.atlassian.net) issue number
  as the *first* thing in your pull request title (e.g. `AO3-1234 Fix thing`)
* [x] Do you have fewer than 5 pull requests already open? If not, please wait
  until they are reviewed and merged before creating new pull requests.

## Issue

https://otwarchive.atlassian.net/browse/AO3-6714

## Purpose

This PR implements point 2. of the "What should happen" section of AO3-6714. For that specific case, if you:

- Are already an owner of a tagset more than once
- Edit the tagset
- Remove yourself as an owner

You are now only removed as a tagset owner once. You remain a tagset owner at least once. This means the tagset cannot be orphaned.

In more general terms, this PR enforces that when removing owners from a tagset:
- The number of `TagSetOwnership` records removed is at most one
- At least one `TagSetOwnership` records exist for an `OwnedTagSet`

## Testing Instructions

Follow the reproduction instructions in AO3-6714. You will still be able to create a tagset where a single user owns it twice, but you can no longer remove all owners from the tagset.

## References

N/A

## Credit

> What name and pronouns should we use to credit you in the [Archive of Our Own's Release Notes](https://archiveofourown.org/admin_posts?tag=1)?

Tom Milligan (they/them)

PR #4864, opened 2024-06-26
AO3-6714 Fix duplicate tag set ownership on create

tommilligan:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6714

## Purpose

This PR is based on https://github.com/otwcode/otwarchive/pull/4863

This PR implements point 1. of the "What should happen" section of AO3-6714. 

This PR prevents a `OwnedTagSet` being created with duplicate `TagSetOwnership` relationships. This happens when the currently logged in user explicitly adds themselves as an owner via the UI.

### Notes

As this PR blocks creating tag sets with duplicate owners in the first place, the test case added in #4863 cannot be maintained without directly constructing an invalid state, with which to continue testing. I didn't do this; please let me know if you think it's required.

## Testing Instructions

Manually follow the recreation steps in the issue. Verify that when adding your username as an explicit owner, the created tag set only lists the currently logged in username as a maintainer once.

## References

N/A

## Credit

> What name and pronouns should we use to credit you in the [Archive of Our Own's Release Notes](https://archiveofourown.org/admin_posts?tag=1)?

Tom Milligan (they/them)

### DIFF
--- a/app/controllers/owned_tag_sets_controller.rb
+++ b/app/controllers/owned_tag_sets_controller.rb
@@ -144,7 +144,17 @@ class OwnedTagSetsController < ApplicationController
 
   def create
     @tag_set = OwnedTagSet.new(owned_tag_set_params)
-    @tag_set.add_owner(current_user.default_pseud)
+
+    # Check if the current user already owns the tag set, and only add them if they don't.
+    #
+    # This happens when the user adds themselves explicitly as an owner.
+    # This check is done by direct Enumerable.find rather than ActiveRecord `find_by`
+    # as we cannot use `find_by` before the record is stored in the database with `.save`
+    current_user_ownership = @tag_set.tag_set_ownerships.find do |ownership|
+      ownership.owner == true && ownership.pseud_id == current_user.default_pseud.id
+    end
+    @tag_set.add_owner(current_user.default_pseud) unless current_user_ownership
+
     if @tag_set.save
       flash[:notice] = ts('Tag Set was successfully created.')
       redirect_to tag_set_path(@tag_set)
--- a/app/controllers/owned_tag_sets_controller.rb
+++ b/app/controllers/owned_tag_sets_controller.rb
@@ -144,17 +144,7 @@ class OwnedTagSetsController < ApplicationController
 
   def create
     @tag_set = OwnedTagSet.new(owned_tag_set_params)
-
-    # Check if the current user already owns the tag set, and only add them if they don't.
-    #
-    # This happens when the user adds themselves explicitly as an owner.
-    # This check is done by direct Enumerable.find rather than ActiveRecord `find_by`
-    # as we cannot use `find_by` before the record is stored in the database with `.save`
-    current_user_ownership = @tag_set.tag_set_ownerships.find do |ownership|
-      ownership.owner == true && ownership.pseud_id == current_user.default_pseud.id
-    end
-    @tag_set.add_owner(current_user.default_pseud) unless current_user_ownership
-
+    @tag_set.add_owner(current_user.default_pseud)
     if @tag_set.save
       flash[:notice] = ts('Tag Set was successfully created.')
       redirect_to tag_set_path(@tag_set)
--- a/app/models/tagset_models/owned_tag_set.rb
+++ b/app/models/tagset_models/owned_tag_set.rb
@@ -154,8 +154,28 @@ class OwnedTagSet < ApplicationRecord
 
   def owner_changes=(pseud_list)
     Pseud.parse_bylines(pseud_list)[:pseuds].each do |pseud|
-      if self.owners.include?(pseud)
-        self.owners -= [pseud] if self.owners.count > 1
+      # NOTE: We load the underlying TagSetOwnership record here
+      # as if we need to delete an owner later, we cannot do it via `self.owners`
+      # without potentially deleting all owners and leaving the tag set orphaned.
+      #
+      # See additional comment below.
+      #
+      # ref: https://otwarchive.atlassian.net/browse/AO3-6714
+      first_ownership_record = self.tag_set_ownerships.find_by(owner: true, pseud_id: pseud.id)
+
+      if first_ownership_record
+        # If this pseud is marked as an owner, delete the first ownership record
+        # linking the psued to this tag set, IF there is more than one ownership record.
+        #
+        # WARNING: Calling `Array#delete` or `array -= [pseud]` on self.owners would
+        # remove the pseud anywhere it appears, which could orphan the tag set and
+        # is not what we want.
+        #
+        # This edge case should only impact existing OwnedTagSets as newer ones will
+        # be properly de-duplicated on creation.
+        #
+        # ref: https://otwarchive.atlassian.net/browse/AO3-6714
+        self.tag_set_ownerships.delete(first_ownership_record) if self.owners.count > 1
       else
         self.moderators -= [pseud] if self.moderators.include?(pseud)
         add_owner(pseud)
--- a/app/models/tagset_models/owned_tag_set.rb
+++ b/app/models/tagset_models/owned_tag_set.rb
@@ -154,28 +154,8 @@ class OwnedTagSet < ApplicationRecord
 
   def owner_changes=(pseud_list)
     Pseud.parse_bylines(pseud_list)[:pseuds].each do |pseud|
-      # NOTE: We load the underlying TagSetOwnership record here
-      # as if we need to delete an owner later, we cannot do it via `self.owners`
-      # without potentially deleting all owners and leaving the tag set orphaned.
-      #
-      # See additional comment below.
-      #
-      # ref: https://otwarchive.atlassian.net/browse/AO3-6714
-      first_ownership_record = self.tag_set_ownerships.find_by(owner: true, pseud_id: pseud.id)
-
-      if first_ownership_record
-        # If this pseud is marked as an owner, delete the first ownership record
-        # linking the psued to this tag set, IF there is more than one ownership record.
-        #
-        # WARNING: Calling `Array#delete` or `array -= [pseud]` on self.owners would
-        # remove the pseud anywhere it appears, which could orphan the tag set and
-        # is not what we want.
-        #
-        # This edge case should only impact existing OwnedTagSets as newer ones will
-        # be properly de-duplicated on creation.
-        #
-        # ref: https://otwarchive.atlassian.net/browse/AO3-6714
-        self.tag_set_ownerships.delete(first_ownership_record) if self.owners.count > 1
+      if self.owners.include?(pseud)
+        self.owners -= [pseud] if self.owners.count > 1
       else
         self.moderators -= [pseud] if self.moderators.include?(pseud)
         add_owner(pseud)
--- a/app/models/tagset_models/tag_set_ownership.rb
+++ b/app/models/tagset_models/tag_set_ownership.rb
@@ -1,4 +1,6 @@
 class TagSetOwnership < ApplicationRecord
   belongs_to :pseud
   belongs_to :owned_tag_set
+
+  validates :pseud, uniqueness: { scope: :owned_tag_set_id }
 end
--- a/features/step_definitions/tag_set_steps.rb
+++ b/features/step_definitions/tag_set_steps.rb
@@ -33,7 +33,7 @@ end
 # ...with the fandom tags "x, y, z" and the character tags "a, b, c"
 # ...with an invisible tag list and the freeform tags "m, n, o"
 # ...with owners "a, b, c" and the freeform tags "x"
-When /^I set up the tag ?set "([^\"]*)" with(?: (?:an? )(visible|invisible) tag list and)?(:? additional owners "([^\"]*)" and)? (.*)$/ do |title, visibility, owners, tags|
+When /^I set up the( invalid)? tag ?set "([^\"]*)" with(?: (?:an? )(visible|invisible) tag list and)?(:? additional owners "([^\"]*)" and)? (.*)$/ do |invalid, title, visibility, owners, tags|
   unless OwnedTagSet.find_by(title: title).present?
     visit new_tag_set_path
     fill_in("owned_tag_set_title", with: title)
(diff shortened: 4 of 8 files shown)


app/models/tagset_models/owned_tag_set.rb
```diff
@@ -155,8 +155,16 @@ def add_moderator(pseud)
 
   def owner_changes=(pseud_list)
     Pseud.parse_bylines(pseud_list)[:pseuds].each do |pseud|
-      if self.owners.include?(pseud)
-        self.owners -= [pseud] if self.owners.count > 1
+      # We load the underlying TagSetOwnership record here
+      # as if we need to delete an owner later, we cannot do it via `self.owners`
+      # without potentially deleting all owners and leaving the tag set orphaned.
```
brianjaustin, line 160: I would add a little extra context here. Specifically, we can't use `Array#delete` or `array -= [...]` because both will remove ALL instances of the pseud, not just the first one. For example (but there's probably a cleaner way to word it)
```rb
# Find the first ownership record for the pseud. If there is one, we delete it
# but ONLY the first one to avoid removing the only owner who happens to
# be assigned multiple times. Calling `Array#delete` on self.owners will remove
# the pseud anywhere it appears, which isn't desirable.
# (NOTE: this should only impact existing OwnedTagSets as newer ones will
# be properly de-duplicated.)
```

brianjaustin (changes requested):
Thank you! A couple of comments (mostly about tests) for you, but overall your approach looks reasonable to me.

On a more general note: I talked over the owners field with some of the other devs, and at some point we'd like to rework it so it's more like the tags field (where you can click an "X" to remove things instead of, somewhat confusingly, adding the pseud to the list to mean it should be removed). If you'd like to tackle that now, great! But that definitely would add a lot more scope, so we are happy to put a Jira ticket in for that to be done later.

features/tag_sets/tag_set.feature
```diff
@@ -10,7 +10,27 @@ Feature: Creating and editing tag sets
...
+    And I set up the tag set "Additional Owner" with owners "anotherowner" and the freeform tags "Shared Responsibilities"
+  Then I should see a create confirmation message
+    And Maintainers should be "anotherowner tagsetter"
+
+  Scenario: A user should not be able to dupicate or remove themselves as an owner (AO3-6714)
+  Given I am logged in as "tagsetter"
+    And I set up the tag set "Duplicate Ownership" with owners "tagsetter" and the freeform tags "Clones"
+  Then I should see a create confirmation message
     And I should see "tagsetter" within ".meta"
+  When I go to the "Duplicate Ownership" tag set edit page
```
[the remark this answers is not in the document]
tommilligan: I'm not sure what you mean here - the steps to reproduce the duplication are here, where we specify an additional owner on creation who is the same as the logged in user: https://github.com/otwcode/otwarchive/pull/4864/files#diff-7aa40c6f60753ed081b2a3c186f775734fc427fdc9adcb714b424ff5007e8e40R26

Possibly the step should be phrased "additional owners" to make this clear?
```cucumber
    And I set up the tag set "Duplicate Ownership" with additional owners "tagsetter" and the freeform tags "Clones"
```

So I think this test is already reproducing the bug ticket.

You can verify that these setup steps would have reproduced the failure, by looking at this test from the first PR I submitted: https://github.com/otwcode/otwarchive/pull/4863/commits/64092e89e3d66831b8144b1cc2cb3711a00921d5#diff-7aa40c6f60753ed081b2a3c186f775734fc427fdc9adcb714b424ff5007e8e40R28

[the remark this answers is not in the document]
brianjaustin: Ah, got it. Yes, I like that proposed wording, I think it makes things a bit clearer (especially given the current UX is Weird :tm:)

app/controllers/owned_tag_sets_controller.rb
```diff
@@ -144,7 +144,17 @@ def new
...
+
+    # Check if the current user already owns the tag set, and only add them if they don't.
+    #
+    # This happens when the user adds themselves explicitly as an owner.
+    # This check is done by direct Enumerable.find rather than ActiveRecord `find_by`
+    # as we cannot use `find_by` before the record is stored in the database with `.save`
+    current_user_ownership = @tag_set.tag_set_ownerships.find do |ownership|
+      ownership.owner == true && ownership.pseud_id == current_user.default_pseud.id
+    end
+    @tag_set.add_owner(current_user.default_pseud) unless current_user_ownership
```
brianjaustin, line 156: I'm thinking I'd prefer this to live in the model somewhere, but bouncing some ideas around right now (with a few clarifications I think the Jira story could use) so feel free to wait on any other changes until I have an answer for you on that

brianjaustin: Ok, I think we can do this with a model validation in https://github.com/otwcode/otwarchive/blob/1bdb4441e193ca269bb31f8c4a78ab5db7d1b663/app/models/tagset_models/tag_set_ownership.rb:
```rb
class TagSetOwnership < ApplicationRecord
  belongs_to :pseud
  belongs_to :owned_tag_set

  validates :pseud, uniqueness: { scope: :owned_tag_set_id }
end
```
(credit to @Bilka2 for the above). This will validate that the pseud is unique and only appears for moderator or owner, not both. The message can be taken care of with Rails I18n, similar to https://github.com/otwcode/otwarchive/blob/dd6ec55a93d549cef4e9a43269dc1cebb975dd91/config/locales/models/en.yml#L109 (docs: https://github.com/otwcode/otwarchive/wiki/Internationalization-%28i18n%29-Standards#overriding-a-default-error-message); please let me know if additional examples would be helpful there.

With that validation in place, we won't need to worry about removing owners multiple times. So, instead of special-casing in the model, could you write a one-time Rake task to find and fix any duplicate owners?

Sorry for the wall of text, please don't hesitate to let us know if you have any questions!

tommilligan: Thanks, this is now done. I've reverted the changes at the `OwnedTagSet` model and have implemented the validation at the `TagSetOwnership` model as you recommended.

My main question here is the performance of the oneoff rake task - I'll a comment where relevant.

lib/tasks/owned_tag_set_tasks.rake
```diff
@@ -0,0 +1,30 @@
+namespace :OwnedTagSet do
+  desc "Deduplicate owners for owned tag sets"
+  task(deduplicate_owners: :environment) do
+    # It's only possible to have duplicate ownership with more than one ownership record
+    OwnedTagSet.find_in_batches do |batch|
```
tommilligan, line 5: I had a think about whether we we could write a query here to target a smaller section of `OwnedTagSet`s, or handle this purely as a db update. This might look something like
```sql
SELECT data.owned_tag_set_id FROM (
  SELECT owned_tag_set_id, count(*)
  AS ownership_count
  FROM tag_set_ownerships
  GROUP BY owned_tag_set_id
) AS data WHERE data.ownership_count > 1;
```
but I suspect this would end up materializing the whole `tag_set_ownerships` table at the same time, which is the opposite of what we want

lib/tasks/owned_tag_set_tasks.rake
```diff
@@ -0,0 +1,30 @@
...
+          # by construction, all ownerships are inside a single OwnedTagSet
+          # so just track which pairs of (psued, owner/maintainer) exist
+          pseud_owner = [ownership.pseud_id, ownership.owner]
+          # Keep the first assignment for each pseud and role, remove others
+          if unique_ownerships.include?(pseud_owner)
+            ownership.destroy
+          else
+            unique_ownerships.add(pseud_owner)
+          end
+        end
```
tommilligan, line 22: Likewise this feels a bit heavy handed, but if it's being run as a one-time cleanup job asynchronously then I'm guessing performance isn't such a concern.

features/step_definitions/tag_set_steps.rb
```diff
@@ -32,23 +32,25 @@
 # ...with a visible tag list
 # ...with the fandom tags "x, y, z" and the character tags "a, b, c"
 # ...with an invisible tag list and the freeform tags "m, n, o"
-When /^I set up the tag ?set "([^\"]*)" with(?: (?:an? )(visible|invisible) tag list and)? (.*)$/ do |title, visibility, tags|
-  unless OwnedTagSet.find_by(title: title).present?
+# ...with owners "a, b, c" and the freeform tags "x"
+When /^I set up the( invalid)? tag ?set "([^"]*)" with(?: (?:an? )(visible|invisible) tag list and)?(:? additional owners "([^"]*)" and)? (.*)$/ do |invalid, title, visibility, owners, tags|
```
brianjaustin, line 36: Because we have some advanced Regex usage here that's not quite replicable using Cucumber Expressions, please follow the instructions on https://github.com/otwcode/otwarchive/wiki/Reviewdog-and-RuboCop to explicitly suppress the RuboCop check for this and the following steps

features/step_definitions/tag_set_steps.rb
```diff
@@ -72,8 +74,14 @@
   step %{I should see an update confirmation message}
 end
 
-When /^I set up the nominated tag ?set "([^\"]*)" with (\d*) fandom noms? and (\d*) (character|relationship) noms?$/ do |title, fandom_count, nested_count, nested_type|
-  unless OwnedTagSet.find_by(title: "#{title}").present?
+# Takes things like When I toggle the owners "tagsetter"
+# Don't forget the extra s, even if it's singular.
```
brianjaustin, line 78: Which extra s? Because of the `()` surrounding it below, the s is optional so
```gherkin
When I toggle the owner "tagsetter"
```
The step name is also a bit vague; can you rename it to something like `I toggle the tagset owner(s) {string}`? That way, we can tell the relation to tagsets from the test output easily

features/step_definitions/tag_set_steps.rb
```diff
@@ -222,20 +230,26 @@
...
   type = fandom_type.tr(" ", "_")
   step %{I should see "#{tag}" within "ol#list_for_fandom_#{name}_in_#{type}_Fandoms li"}
 end
 
-Then /^"([^\"]*)" should be an unassociated tag$/ do |tag|
+Then /^"([^"]*)" should be an unassociated tag$/ do |tag|
   step %{I should see "#{tag}" within "ol#list_for_unassociated_char_and_rel"}
 end
+
+Then "the maintainers should be {string}" do |maintainers|
```
brianjaustin, line 251: Similar note about the naming of this step as above

spec/lib/tasks/owned_tag_set_tasks.rake_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+describe "rake OwnedTagSet:deduplicate_owners" do
+  it "does not update owned tag sets with only one owner" do
+    owned_tag_set = create(:owned_tag_set)
+    # OwnedTagSet is implicitly created with a single owner
+    expect(owned_tag_set.owners.count).to eq(1)
+    expected_owners = owned_tag_set.owners
+
+    subject.invoke
+
+    expect(owned_tag_set.reload.owners).to eq(expected_owners)
```
brianjaustin, line 12: Could this be simplified to the following?
```suggestion
    expected_owners = owned_tag_set.owners

    expect { subject.invoke }.not_to change { owned_tag_set.reload.owners }
```

(formatting likely not 100% correct)

spec/lib/tasks/owned_tag_set_tasks.rake_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+        .transform_values { :length }
+    end
+
+    expect(owner_count(owned_tag_set.owners)).to eq({ pseud1 => 2, pseud2 => 1, pseud3 => 3 })
+
+    subject.invoke
+
+    owned_tag_set = owned_tag_set.reload
+    expect(owned_tag_set.owners.count).to eq(3)
+    expect(owner_count(owned_tag_set.owners)).to eq({ pseud1 => 1, pseud2 => 1, pseud3 => 1 })
```
brianjaustin, line 70: Can this (and maybe the above) be simplified a bit via what's described in https://stackoverflow.com/a/11796957?

app/models/tagset_models/tag_set_ownership.rb
```diff
@@ -1,4 +1,6 @@
 class TagSetOwnership < ApplicationRecord
   belongs_to :pseud
   belongs_to :owned_tag_set
+
+  validates :pseud, uniqueness: { scope: :owned_tag_set_id }
```
brianjaustin: Could you add the index rubocop is mentioning? It will need to be unique based on both `pseud_id` and `owned_tag_set_id`

lib/tasks/owned_tag_set_tasks.rake
brianjaustin: Will take a look at this in the database, but fair warning, it may take some time as I have a fair amount on my plate at the minute

app/models/tagset_models/tag_set_ownership.rb
brianjaustin, line 5: This will end up with an error on duplicates. While that prevents the problem, the ticket asks for
> If you follow Step 3, you should not be added as a tag set owner twice. (It should just add you once; no need for an error.)

The other point, about duplicates, is addressed by the post-deploy rake task to just clean everything up. It might be nice to prevent the weird removal behaviour regardless, but not as high a priority

lib/tasks/owned_tag_set_tasks.rake
brianjaustin:
```
MariaDB [otwarchive_production]> explain SELECT data.owned_tag_set_id FROM (
    ->   SELECT owned_tag_set_id, count(*)
    ->   AS ownership_count
    ->   FROM tag_set_ownerships
    ->   GROUP BY owned_tag_set_id
    -> ) AS data WHERE data.ownership_count > 1;
+------+-------------+--------------------+------+---------------+------+---------+------+------+---------------------------------+
| id   | select_type | table              | type | possible_keys | key  | key_len | ref  | rows | Extra                           |
+------+-------------+--------------------+------+---------------+------+---------+------+------+---------------------------------+
|    1 | PRIMARY     | <derived2>         | ALL  | NULL          | NULL | NULL    | NULL | 5830 | Using where                     |
|    2 | DERIVED     | tag_set_ownerships | ALL  | NULL          | NULL | NULL    | NULL | 5830 | Using temporary; Using filesort |
+------+-------------+--------------------+------+---------------+------+---------+------+------+---------------------------------+
```

versus
```
MariaDB [otwarchive_production]> select count(*) from owned_tag_sets;
+----------+
| count(*) |
+----------+
|     7398 |
+----------+
```
I don't think we get a huge optimization from the first one, so the batching approach (the one you have already) is fine IMO

PR #4869, opened 2024-07-01
AO3-6232 Include IP, referer, skin in support tickets

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6232

## Purpose

Adds custom field to Zoho tickets for the submitter's IP address, referer, and the skin they're using at the time of submitting a support ticket

## References

The names in this PR will need a final approval from support before merging, to make sure they've created the fields in Zoho as required

## Credit

Brian Austin (they/he)

### DIFF
--- a/app/controllers/feedbacks_controller.rb
+++ b/app/controllers/feedbacks_controller.rb
@@ -19,6 +19,8 @@ class FeedbacksController < ApplicationController
     @feedback.rollout = @feedback.rollout_string
     @feedback.user_agent = request.env["HTTP_USER_AGENT"]
     @feedback.ip_address = request.remote_ip
+    @feedback.referer = request.referer if request.referer && ArchiveConfig.PERMITTED_HOSTS.include?(URI(request.referer).host)
+    @feedback.site_skin = current_skin
     if @feedback.save
       @feedback.email_and_send
       flash[:notice] = t("successfully_sent",
@@ -37,10 +39,16 @@ class FeedbacksController < ApplicationController
     @support_languages = Language.where(support_available: true).default_order
   end
 
+  def current_skin
+    skin = Skin.approved_or_owned_by.usable.find_by(id: session[:site_skin]) if session[:site_skin]
+    skin ||= current_user&.preference&.skin
+    skin ||= Skin.default
+    skin
+  end
+
   def feedback_params
     params.require(:feedback).permit(
       :comment, :email, :summary, :username, :language
     )
   end
-
 end
--- a/app/controllers/feedbacks_controller.rb
+++ b/app/controllers/feedbacks_controller.rb
@@ -20,7 +20,7 @@ class FeedbacksController < ApplicationController
     @feedback.user_agent = request.env["HTTP_USER_AGENT"]
     @feedback.ip_address = request.remote_ip
     @feedback.referer = request.referer if request.referer && ArchiveConfig.PERMITTED_HOSTS.include?(URI(request.referer).host)
-    @feedback.site_skin = current_skin
+    @feedback.site_skin = helpers.current_skin
     if @feedback.save
       @feedback.email_and_send
       flash[:notice] = t("successfully_sent",
@@ -37,13 +37,6 @@ class FeedbacksController < ApplicationController
 
   def load_support_languages
     @support_languages = Language.where(support_available: true).default_order
-  end
-
-  def current_skin
-    skin = Skin.approved_or_owned_by.usable.find_by(id: session[:site_skin]) if session[:site_skin]
-    skin ||= current_user&.preference&.skin
-    skin ||= Skin.default
-    skin
   end
 
   def feedback_params
--- a/app/helpers/skins_helper.rb
+++ b/app/helpers/skins_helper.rb
@@ -36,7 +36,7 @@ module SkinsHelper
     skin = current_skin
     return "" unless skin
 
-    # We include the version information for both the skin_id and the
+    # We include the version information for both the skin's id and the
     # AdminSetting.default_skin_id because the default skin is used in skins of
     # type "user", so we need to regenerate the cache block when it's modified.
     #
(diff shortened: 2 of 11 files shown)


app/controllers/feedbacks_controller.rb
```diff
@@ -37,10 +39,16 @@ def load_support_languages
     @support_languages = Language.where(support_available: true).default_order
   end
 
+  def current_skin
```
Bilka2, line 42: In `SkinsHelper#skin_tag`, there is also a check for `params[:site_skin]`, e.g. https://test.archiveofourown.org/?site_skin=928.

Because you're essentially replicating the skin setting logic from the `#skin_tag` method, I think it would be good to separate that out into its own method that can be used both for the support tickets and for `#skin_tag`.

app/helpers/skins_helper.rb
```diff
@@ -47,12 +44,11 @@ def skin_tag
     # regenerate the cache block when an admin updates the current default
     # skin.
     Rails.cache.fetch(
-      [:v1, :site_skin, skin_id, logged_in_as_admin?],
-      version: [skin_cache_version(skin_id),
+      [:v1, :site_skin, skin.id, logged_in_as_admin?],
+      version: [skin_cache_version(skin.id),
```
sarken, line 48: The note above references `skin_id` -- should that be updated?

app/models/feedback.rb
```diff
@@ -69,8 +70,17 @@ def send_report
       username: username,
       user_agent: user_agent,
       site_revision: ArchiveConfig.REVISION.to_s,
-      rollout: rollout
+      rollout: rollout,
+      ip_address: ip_address,
+      referer: referer,
```
sarken, line 75: Did we get a definite answer from Support on how many characters were allowed in this field to make sure we avoid a repeat of [AO3-6692](https://otwarchive.atlassian.net/issues/AO3-6692)?

app/models/feedback_reporters/support_reporter.rb
```diff
@@ -16,7 +16,10 @@ def custom_zoho_fields
     {
       "cf_archive_version" => site_revision.presence || "Unknown site revision",
       "cf_rollout" => rollout.presence || "Unknown",
-      "cf_user_agent" => user_agent.presence || "Unknown user agent"
+      "cf_user_agent" => user_agent.presence || "Unknown user agent",
+      "cf_ip" => ip_address.presence || "Unknown",
```
sarken, line 20: Would it make sense to do "Unknown IP" and "Unknown URL" to match the abuse reporter? Although "URL" could be "Referer," too.

...actually, I wonder if we should check with Support/PAC and see if they should both use `cf_url` as the field name? I know they sometimes transfer tickets between their Zoho departments, and I don't know if a) the field names need to match for information to transfer, or b) if they'd want the URL/referer to transfer. 🤔

spec/controllers/feedbacks_controller_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+    end
+
+    context "when accessed by a logged-in user" do
+      let(:user) { create(:user) }
+
+      before do
+        fake_login_known_user(user)
+      end
+
+      context "and the user has no skin set" do
```
sarken, line 35: Can you swap the "and" in the context descriptions for "when"? It's awkward grammar if we're writing it out as a sentence, but it's consistent with what we do in other specs.

spec/helpers/skins_helper_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+      admin_setting.default_skin = Skin.default
+      admin_setting.save(validate: false)
+    end
+
+    context "when the parameters include a skin id" do
+      before do
+        params[:site_skin] = skin.id
+      end
+
+      context "and the skin is usable" do
```
sarken, line 21: I'm not sure "usable" is a good word here, since we have a `usable` method for skins that checks if `unusable: false`. "The skin is applied" or something?

spec/models/feedback_reporters/support_reporter_spec.rb
```diff
@@ -97,5 +103,45 @@
         expect(subject.report_attributes.fetch("description")).to eq("Hi!http://example.com/Camera-icon.svgBye!")
       end
     end
+
+    context "if the report has an empty ip address" do
```
sarken, line 107: "IP" for consistent capitalization? Or "ip_address" if you'd rather the field name.

sarken (commented):
I am so excited about this -- thank you for working on it! I just have some nitpicks and something we should probably check with Support/PAC about.

app/models/feedback_reporters/support_reporter.rb
sarken: I'm checking with Support and PAC chairs about the field names, btw.

brianjaustin: (I'm happy to change it back, but since part of the ticket is about handing tickets to PAC I think it makes sense to use the same name to make that as easy as possible.)

app/models/feedback.rb
brianjaustin: They asked what PAC uses, and PAC said 255. I've added something to truncate that just in case, since we technically support something silly like 20k.

sarken: Makes sense! If Support has something longer than 255, passing it to PAC will probably not work super great.

app/models/feedback_reporters/support_reporter.rb
sarken: We should change it back! Support has had a cf_url field for a while now, for the sole purpose to storing URLs from transferred PAC tickets, and it has been working fine.

sarken (changes requested):
Marking changes needed for [this comment](https://github.com/otwcode/otwarchive/pull/4869#discussion_r1722568405)

sarken (approved):

PR #4872, opened 2024-07-13
AO3-6756 Add i18n-tasks workflow for Phrase export PRs

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6756

## Purpose

Adds a GitHub Action workflow. The workflow lists all redundant English keys that should be removed from Phrase. The check will show as failed if there are any and the list will be available as workflow artifact. Furthermore, the workflow runs the consistent interpolations check for all locales and uploads its result. If there are no redundant keys on Phrase then it also runs the unused keys check for all locales. The check will show as failed and the unused keys will be available as workflow artifact if there are any unused keys.

The workflow runs automatically when making a pull request from a branch named `phrase-translations`. Phrase does this when exporting strings to GitHub, so it should be fully automatic.

[Demo PR](https://github.com/Bilka2/otwarchive/pull/10):
English keys that should removed from Phrase: https://github.com/Bilka2/otwarchive/actions/runs/9919350109

## Testing Instructions

See Jira.

## Credit

Bilka (he/him)

### DIFF
--- a/.github/workflows/phrase-export-checks.yml
+++ b/.github/workflows/phrase-export-checks.yml
@@ -1,0 +1,84 @@
+name: Check Phrase translations
+on:
+  pull_request:
+    branches:
+      - master
+      - main
+  workflow_dispatch:
+
+jobs:
+  en-keys-to-remove:
+    if: github.head_ref == 'phrase-translations'
+    runs-on: ubuntu-latest
+    name: Check for outdated English keys in Phrase
+
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v4
+
+      - name: Set up Ruby and run bundle install
+        uses: ruby/setup-ruby@v1
+        with:
+          bundler-cache: true
+
+      - name: Run i18n-tasks
+        run: bundle exec i18n-tasks missing -t diff -f keys en | bundle exec i18n-tasks tree-mv -f keys "en.{*}" "\1" > en-keys-to-remove.txt
+
+      - name: Check that there are zero keys in other locales that don't exist in English
+        run: "! test -s en-keys-to-remove.txt"
+
+      - name: Upload list of English keys that should be removed from Phrase
+        if: ${{ failure() }}
+        uses: actions/upload-artifact@v4
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/phrase-export-checks.yml
```diff
@@ -0,0 +1,84 @@
+name: Check Phrase translations
```
sarken, line 1: Maybe "Check Phrase exports" or "Check exported Phrase translations" just to make it a tiny bit clearer to new folks that it shouldn't be running on all translations they might've added?

.github/workflows/phrase-export-checks.yml
```diff
@@ -0,0 +1,84 @@
...
+
+      - name: Set up Ruby and run bundle install
+        uses: ruby/setup-ruby@v1
+        with:
+          bundler-cache: true
+
+      - name: Run i18n-tasks
+        run: bundle exec i18n-tasks missing -t diff -f keys en | bundle exec i18n-tasks tree-mv -f keys "en.{*}" "\1" > en-keys-to-remove.txt
+
+      - name: Check that there are zero keys in other locales that don't exist in English
```
sarken, line 27: I had some trouble parsing this. At first I thought this meant "zero keys" as in "keys for count: 0". Maybe:
- "Check that there are no keys in..."
- "Check that all keys exported from Phrase exist in English"

.github/workflows/phrase-export-checks.yml
```diff
@@ -0,0 +1,84 @@
...
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v4
+
+      - name: Set up Ruby and run bundle install
+        uses: ruby/setup-ruby@v1
+        with:
+          bundler-cache: true
+
+      - name: Run i18n-tasks
```
sarken, line 24: `Run i18n-tasks missing`, since the other ones mention the  specific task?

sarken (commented):
Just a couple suggestions around phrasing, actual workflow looks good!

sarken (approved):

PR #4875, opened 2024-07-16
AO3-6764 Add kudo mailer preview

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6764

## Purpose

Add a preview for the kudo mailer. The amount of guest and user kudos is configurable with URL params.

Move the `I18n.with_locale` call so that the locale dropdown in the preview works (see #4292 for another PR that did this). Update the locale keys to get rid of the `html_safe` calls.

## Testing Instructions

See Jira.

## References

PR Limit: 6 out of 5
Using reviewer bonus:
https://github.com/otwcode/otwarchive/pull/4720  
https://github.com/otwcode/otwarchive/pull/4736

## Credit

Bilka (he/him)


test/mailers/previews/kudo_mailer_preview.rb
```diff
@@ -0,0 +1,12 @@
+class KudoMailerPreview < ApplicationMailerPreview
+  # Sends a kudos notification
+  def batch_kudo_notification
+    user = create(:user)
+    work = create(:work)
+    guest_count = params[:guest_count] || 1
+    user_count = params[:user_count].to_i || 1
+    names = Array.new(user_count) { "User#{Faker::Alphanumeric.alpha(number: 8)}" }
+    hash = { "Work_#{work.id}": { guest_count:, names: } }
```
Bilka2, line 9: This syntax is new to Ruby 3.1 and the rubocop annotation is a false positive. Adding a space would make a different cop complain. This cop accepts the formatting starting with [rubocop 1.24.0](https://github.com/rubocop/rubocop/releases/tag/v1.24.0), we are on 1.22.3.

spec/mailers/user_mailer_spec.rb
```diff
@@ -1205,10 +1205,25 @@
...
+      expect(I18n).to receive(:with_locale).with("en").and_call_original
+      email = UserMailer.admin_hidden_work_notification(work.id, user.id)
+      expect(email).to have_html_part_content("fan-run and fan-supported archive")
+      expect(email).not_to have_html_part_content("translated footer")
+      expect(email).to have_text_part_content("fan-run and fan-supported archive")
+      expect(email).not_to have_text_part_content("translated footer")
+    end
+  end
+
   describe "obeys the set locale preference feature flag" do
```
Bilka2, line 1219: When a mailer preview is added for admin_hidden_work_notification, its `I18n.with_locale` call will have to be moved outside of `UserMailer.admin_hidden_work_notification` and this test will break.

Proposal: I could remove this test and expand the cucumber test added in this PR to turn the feature flag off again (with the preference still set), send another email and check that it's not translated.

features/step_definitions/email_custom_steps.rb
```diff
@@ -1,40 +1,59 @@
-Given /^(?:a clear email queue|no emails have been sent|the email queue is clear)$/ do
+Given "the email queue is clear" do
   reset_mailer
 end
 
-Then /^"([^\"]*)" should be emailed$/ do |user|
+Given "a locale with translated emails" do
+  language = Language.find_or_create_by(short: "new", name: "New")
```
Bilka2, line 6: This is a completely new language to prevent the existing translations from interfering with the tests. Otherwise a test might break when a Phrase export changes a translation.

brianjaustin (approved):
Thanks!

PR #4880, opened 2024-07-22
AO3-6709 Allow certain admins to access inboxes

nish-shai-scacap:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6709

## Purpose

- Allows admins with the role `superadmin` or `policy_and_abuse` to access user inboxes
- Admins cannot approve or delete comments or mark them as read/unread. They will be redirected to homepage with error if they perform these actions.

## Credit

Nisha (she/her)


spec/controllers/inbox_controller_spec.rb
```diff
@@ -19,6 +19,111 @@
...
+              expect(assigns(:inbox_total)).to eq(2)
+              expect(assigns(:unread)).to eq(2)
+            end
+          end
+
+          context "with a deleted comment" do
+            let(:inbox_comment) { create(:inbox_comment, user: user) }
+
+            it "excludes deleted comments" do
+              inbox_comment.feedback_comment.destroy
```
brianjaustin, line 117: The version of `destroy` that throws an error on failure is usually nicer for tests, since the non-error-throwing one can fail "silently" and make test failures harder to diagnose
```suggestion
              inbox_comment.feedback_comment.destroy!
```

app/policies/dashboard_policy.rb
```diff
@@ -0,0 +1,7 @@
+class DashboardPolicy < ApplicationPolicy
+  VIEW_INBOX_ROLES = %w[superadmin policy_and_abuse].freeze
+
+  def can_view_inbox_link?
```
brianjaustin, line 4: Since this is the link and also viewing the inbox in general, I'd make this method name a little bit more generic
```suggestion
  def can_view_inbox?
```

app/controllers/application_controller.rb
```diff
@@ -269,13 +269,17 @@ def access_denied(options ={})
...
       end
       format.json do
-        errors = [ts("Sorry, only an authorized admin can do that.")]
+        errors = [t("admin.access.action_access_denied")]
         render json: { errors: errors }, status: :forbidden
       end
+      format.js do
+        flash[:error] = t("admin.access.page_access_denied") 
+        render js: "window.location.href = '#{root_path}';" 
+      end
```
brianjaustin, line 282: Might be a silly question, but to make sure I understand, what requires adding this format?

brianjaustin (changes requested):
Thanks! (and double thanks for making things translatable!) One question and some minor suggestions for you when you have some time

app/policies/dashboard_policy.rb
Bilka2: I was wondering whether it would make sense to combine the two policies into one by removing this one and checking `policy(:inbox_comment).show?` in the view instead. Other similar policies appear to be combined like this.

sarken: Bilka's suggestion should be fine! But just to jump on here, we'd generally prefer to use terms like `access` over `view` -- we try to avoided using sight-centered language whenever possible, even in the code.

app/controllers/application_controller.rb
nisha-shaikh: This is for redirecting user and showing the error when handling an ajax request. In this case, when the admin tries to approve an inbox comment, they will be redirected to the root page with the error flashing above it. I tried to show the error on the same inbox page, but I wasn't able to do that. If there's a better way of doing this, please let me know

brianjaustin (approved):
Thank you!

Bilka2:
This branch has a merge conflict, could you merge master and solve the conflict? We'd like to include this PR in a release soon.

PR #4888, opened 2024-08-03
AO3-6783 Fix that removing a tag nomination would result in a blank tag nomination

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6783

## Purpose

If a tag nomination is blank (empty string for tagname) because it is meant to be destroyed, don't modify it in `before_save` and exclude it from certain validations.

`Tag.find_by(name: "")` where name is an empty string returns the broken blank tag. Tag nominations that are meant to be removed have an empty string for the tag name. So,  `Tag.find_by_name(tagname)` in TagNomination's `before_save :set_tag_status` returns the broken blank tag and its name is used to set the `TagNomination.tagname`. This means the nomination that was meant to be destroyed by  `after_save :destroy_if_blank` is no longer blank, instead it contains all the information from the broken blank tag so it doesn't get destroyed. There is no way to get rid of this tag nomination, since any attempt to clear the tagname field will repeat this process. The only way to escape the broken blank tag is by renaming the nomination to a completely different tag name, but that still keeps the unwanted tag nomination around.

The problem of the DB returning the broken tag also affects the autocomplete, so I've also added the bypass there.

While investigating which callbacks to skip, I noticed that `validate :known_fandom` on CastNomination would run for nominations about to be destroyed, so you needed to keep Fandom? filled out if you wanted to remove a character or relationship nomination in a tag set that doesn't allow fandom nominations. That's also fixed by skipping it for blank tagnames, so I went ahead and changed it.

## Testing Instructions

See Jira.

## References

The database returning the zero width space tag when querying for an empty string is potentially caused by the database collation. However, the comments on [AO3-6366](https://otwarchive.atlassian.net/issues/AO3-6366) indicates that changing this collation would be a complicated endeavor. So this PR can serve as a workaround to get rid of the incredibly annoying behaviour currently happening on production until it's clear whether changing the collation fixes the underlying issue.

Note that all tests I added here (with the exception of `Scenario: Zero width space tag appears in the autocomplete for zero width space`) fail before the changes in this PR. So users on production are likely hitting the confusing "Someone else has already nominated the tag for this set but in fandom ...." error in `Scenario: A tag nomination with for the zero width space tag doesn't prevent removing other tag nominations`.

## Credit

Bilka (he/him)

[AO3-6366]: https://otwarchive.atlassian.net/browse/AO3-6366?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/autocomplete_controller.rb
+++ b/app/controllers/autocomplete_controller.rb
@@ -92,9 +92,9 @@ class AutocompleteController < ApplicationController
     raise "Redshirt: Attempted to constantize invalid class initialize noncanonical_tag #{params[:type].classify}" unless Tag::TYPES.include?(params[:type].classify)
 
     tag_class = params[:type].classify.constantize
-    one_tag = tag_class.find_by(canonical: false, name: params[:term])
+    one_tag = tag_class.find_by(canonical: false, name: params[:term]) if params[:term].present?
     # If there is an exact match in the database, ensure it is the first thing suggested.
-    match = if one_tag && params[:term].present?
+    match = if one_tag
               [one_tag.name]
             else
               []
--- a/app/controllers/autocomplete_controller.rb
+++ b/app/controllers/autocomplete_controller.rb
@@ -94,7 +94,7 @@ class AutocompleteController < ApplicationController
     tag_class = params[:type].classify.constantize
     one_tag = tag_class.find_by(canonical: false, name: params[:term])
     # If there is an exact match in the database, ensure it is the first thing suggested.
-    match = if one_tag
+    match = if one_tag && params[:term].present?
               [one_tag.name]
             else
               []
--- a/features/tag_sets/tag_set_nominations.feature
+++ b/features/tag_sets/tag_set_nominations.feature
@@ -114,6 +114,25 @@ Feature: Nominating and reviewing nominations for a tag set
     And I should see "My Favorite Character/Their Worst Enemy"
     But I should not see "Their Best Friend"
 
+  Scenario: You should be able to delete a nominated character and its fandom at once
+  when the tagset doesn't allow fandom nominations
+    Given a canonical character "Common Character" in fandom "Canon"
+    And I am logged in as "tagsetter"
+    And I set up the nominated tag set "Nominated Tags" with 0 fandom noms and 3 character noms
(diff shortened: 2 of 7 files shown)


features/tag_sets/tag_set_nominations.feature
```diff
@@ -260,6 +279,46 @@ Feature: Nominating and reviewing nominations for a tag set
...
+    Then I should see "Not Yet Reviewed (may be edited or deleted)"
+    When I follow "Edit"
+      And I fill in "Character" with ""
+      And I press "Submit"
+    Then I should see "Your nominations were successfully updated."
+      And I should see "Treasure Chest"
+      But I should not see "Gold"
+      And I should see "None nominated in this fandom."
+
+  Scenario: A tag nomination with for the zero width space tag doesn't prevent removing other tag nominations
```
brianjaustin, line 303: Should this be something like the following?
```suggestion
  Scenario: A tag nomination with the zero width space tag doesn't prevent removing other tag nominations
```

app/controllers/autocomplete_controller.rb
```diff
@@ -94,7 +94,7 @@ def noncanonical_tag
     tag_class = params[:type].classify.constantize
     one_tag = tag_class.find_by(canonical: false, name: params[:term])
     # If there is an exact match in the database, ensure it is the first thing suggested.
-    match = if one_tag
+    match = if one_tag && params[:term].present?
```
adt-chair, line 97: Just one question! Why does this need to be here and not on line 95, like

```
one_tag = tag_class.find_by(canonical: false, name: params[:term]) if params[:term].present?
```

sarken, line 97: Just one question! (This time, from the right account.) Why does this need to be here and not on line 95, like

```
one_tag = tag_class.find_by(canonical: false, name: params[:term]) if params[:term].present?
```

Bilka2: Either spot should work, but your suggestion is likely slightly better because it skips the database query. Changed!

sarken (approved):

PR #4889, opened 2024-08-04
AO3-6010 Add role for Legal

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6010

## Purpose

Add a role for our Legal folks, along with some permissions for that role

## Credit

Brian Austin (they/he)

### DIFF
--- a/app/policies/comment_policy.rb
+++ b/app/policies/comment_policy.rb
@@ -1,11 +1,11 @@
 class CommentPolicy < ApplicationPolicy
-  DESTROY_COMMENT_ROLES = %w[superadmin board policy_and_abuse support].freeze
-  DESTROY_ADMIN_POST_COMMENT_ROLES = %w[superadmin board board_assistants_team communications elections policy_and_abuse support].freeze
+  DESTROY_COMMENT_ROLES = %w[superadmin board legal policy_and_abuse support].freeze
+  DESTROY_ADMIN_POST_COMMENT_ROLES = %w[superadmin board board_assistants_team communications elections legal policy_and_abuse support].freeze
   FREEZE_TAG_COMMENT_ROLES = %w[superadmin tag_wrangling].freeze
   FREEZE_WORK_COMMENT_ROLES = %w[superadmin policy_and_abuse].freeze
-  HIDE_TAG_COMMENT_ROLES = %w[superadmin tag_wrangling].freeze
-  HIDE_WORK_COMMENT_ROLES = %w[superadmin policy_and_abuse].freeze
-  SPAM_ADMIN_POST_COMMENT_ROLES = %w[superadmin board board_assistants_team communications elections policy_and_abuse support].freeze
+  HIDE_TAG_COMMENT_ROLES = %w[superadmin legal tag_wrangling].freeze
+  HIDE_WORK_COMMENT_ROLES = %w[superadmin legal policy_and_abuse].freeze
+  SPAM_ADMIN_POST_COMMENT_ROLES = %w[superadmin board board_assistants_team communications elections legal policy_and_abuse support].freeze
   SPAM_COMMENT_ROLES = %w[superadmin board policy_and_abuse support].freeze
 
   def can_destroy_comment?
@@ -63,6 +63,6 @@ class CommentPolicy < ApplicationPolicy
   alias review_all? can_review_all?
 
   def show_email?
-    user_has_roles?(%w[policy_and_abuse support superadmin])
+    user_has_roles?(%w[legal policy_and_abuse support superadmin])
   end
 end
--- a/app/policies/comment_policy.rb
+++ b/app/policies/comment_policy.rb
@@ -5,7 +5,7 @@ class CommentPolicy < ApplicationPolicy
   FREEZE_WORK_COMMENT_ROLES = %w[superadmin policy_and_abuse].freeze
   HIDE_TAG_COMMENT_ROLES = %w[superadmin legal tag_wrangling].freeze
   HIDE_WORK_COMMENT_ROLES = %w[superadmin legal policy_and_abuse].freeze
-  SPAM_ADMIN_POST_COMMENT_ROLES = %w[superadmin board board_assistants_team communications elections legal policy_and_abuse support].freeze
+  SPAM_ADMIN_POST_COMMENT_ROLES = %w[superadmin board board_assistants_team communications elections policy_and_abuse support].freeze
   SPAM_COMMENT_ROLES = %w[superadmin board policy_and_abuse support].freeze
 
   def can_destroy_comment?
--- a/app/policies/user_policy.rb
+++ b/app/policies/user_policy.rb
@@ -4,7 +4,7 @@ class UserPolicy < ApplicationPolicy
   # - managing a user's invitations
   # - updating a user's email and roles (e.g. wranglers, archivists, not admin roles)
   # This is further restricted using ALLOWED_ATTRIBUTES_BY_ROLES.
-  MANAGE_ROLES = %w[superadmin policy_and_abuse open_doors support tag_wrangling].freeze
+  MANAGE_ROLES = %w[superadmin legal policy_and_abuse open_doors support tag_wrangling].freeze
 
   # Roles that allow updating the Fannish Next Of Kin of a user.
   MANAGE_NEXT_OF_KIN_ROLES = %w[superadmin policy_and_abuse support].freeze
--- a/features/admins/admin_works.feature
+++ b/features/admins/admin_works.feature
@@ -13,10 +13,10 @@ Feature: Admin Actions for Works, Comments, Series, Bookmarks
       And I press "Troubleshoot"
     Then I should see "Work sent to be reindexed."
 
-  Scenario: Can hide works
+  Scenario Outline: Can hide works
     Given I am logged in as "regular_user"
       And I post the work "ToS Violation"
-    When I am logged in as a "policy_and_abuse" admin
+    When I am logged in as a "<role>" admin
       And all emails have been delivered
       And I view the work "ToS Violation"
       And I follow "Hide Work"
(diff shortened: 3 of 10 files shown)


features/admins/admin_works.feature
```diff
@@ -86,6 +116,35 @@ Feature: Admin Actions for Works, Comments, Series, Bookmarks
       And I am on bad_user's bookmarks page
     Then I should not see "Rude comment"
 
+    Examples:
+      | role             |
+      | legal            |
+      | policy_and_abuse |
+
+  Scenario Outline: Deleting bookmarks
+    Given basic tags
```
Bilka2, line 125: Is this step needed?

features/admins/admin_works.feature
```diff
@@ -86,6 +116,35 @@ Feature: Admin Actions for Works, Comments, Series, Bookmarks
...
+    When I follow "Bookmark"
+      And I fill in "bookmark_notes" with "Rude comment"
+      And I press "Create"
+      And all indexing jobs have been run
+    Then I should see "Bookmark was successfully created"
+    When I am logged in as a "<role>" admin
+      And I am on bad_user's bookmarks page
+      And I follow "Delete Bookmark"
+    Then I should see "Item was successfully deleted."
+    When I am logged in as "regular_user" with password "password1"
```
Bilka2, line 139: I think logging in as bad_user would be better, to distinguish from hiding the bookmark.

features/admins/admin_works.feature
```diff
@@ -155,15 +214,37 @@ Feature: Admin Actions for Works, Comments, Series, Bookmarks
...
+    When I follow "Make External Work Visible"
+    Then I should see "Item is no longer hidden."
+
+    Examples:
+      | role             |
+      | legal            |
+      | policy_and_abuse |
+
+  Scenario Outline: Deleting external works
+    Given basic tags
```
Bilka2, line 235: Is this step needed?

spec/controllers/comments_controller_spec.rb
```diff
@@ -2387,7 +2387,7 @@
             end
           end
 
-          %w[superadmin tag_wrangling].each do |admin_role|
+          %w[superadmin legal tag_wrangling].each do |admin_role|
```
Bilka2, line 2390: In line 2456 it looks like legal should also be added.

spec/controllers/comments_controller_spec.rb
```diff
@@ -2557,7 +2557,7 @@
...
```
Bilka2, line 2560: In line 2626 it looks like legal should also be added.

spec/controllers/comments_controller_spec.rb
```diff
@@ -2729,7 +2729,7 @@
...
```
Bilka2, line 2732: In line 2798 it looks like legal should also be added.

spec/controllers/comments_controller_spec.rb
```diff
@@ -2899,7 +2899,7 @@
...
```
Bilka2, line 2902: In line 2968 it looks like legal should also be added.

spec/controllers/comments_controller_spec.rb
```diff
@@ -3283,7 +3283,7 @@
             end
           end
 
-          (Admin::VALID_ROLES - %w[superadmin board policy_and_abuse support]).each do |admin_role|
+          (Admin::VALID_ROLES - %w[superadmin board legal policy_and_abuse support]).each do |admin_role|
```
Bilka2, line 3286: This array and the one in line 3299 could be consolidated into a variable.

spec/controllers/comments_controller_spec.rb
```diff
@@ -3401,7 +3401,7 @@
...
```
Bilka2, line 3404: This array and the one in line 3417 could be consolidated into a variable.

spec/controllers/comments_controller_spec.rb
```diff
@@ -3414,7 +3414,7 @@
             end
           end
 
-          %w[superadmin board policy_and_abuse support].each do |admin_role|
+          %w[superadmin board legal policy_and_abuse support].each do |admin_role|
```
Bilka2, line 3417: In lines 3851 it looks like legal should also be added.

spec/controllers/admin/admin_users_controller_spec.rb
```diff
@@ -6,83 +6,87 @@
...
-        admin.update!(roles: [])
-        fake_login_admin(admin)
-        get :bulk_search
+  shared_examples "permits access to authorized admins" do
+    search_roles.each do |role|
+      context "with role #{role}" do
+        let(:admin) { create(:admin, roles: [role]) }
 
-        it_redirects_to_with_error(root_url, "Sorry, only an authorized admin can access the page you were trying to reach.")
+        it "allows access to index" do
```
Bilka2, line 36: This text looks like it needs an update.

spec/controllers/admin/admin_users_controller_spec.rb
```diff
@@ -6,83 +6,87 @@
...
+    before do
+      fake_login_admin(admin)
+      get :show, params: { id: user.login }
     end
 
+    it_behaves_like "denies access to unauthorized admins"
+    it_behaves_like "permits access to authorized admins"
+
     context "when admin has correct authorization" do
+      let(:admin) { create(:admin, roles: ["policy_and_abuse"]) }
```
Bilka2, line 75: There is also this option:
```suggestion
      let(:admin) { create(:policy_and_abuse_admin) }
```

Bilka2 (changes requested):
Two questions regarding mismatches between Jira and what was implemented and a few notes for the tests.

`admin_users_controller_spec.rb` is missing a few tests:
* test that update doesn't allow legal to change email or roles
* test that update_next_of_kin, update_status, confirm_delete_user_creations, destroy_user_creations doesn't allow the unauthorized admin roles to do changes/access the actions
* activate and troubleshoot aren't testing any roles besides support

app/policies/comment_policy.rb
```diff
@@ -1,11 +1,11 @@
...
+  DESTROY_COMMENT_ROLES = %w[superadmin board legal policy_and_abuse support].freeze
+  DESTROY_ADMIN_POST_COMMENT_ROLES = %w[superadmin board board_assistants_team communications elections legal policy_and_abuse support].freeze
   FREEZE_TAG_COMMENT_ROLES = %w[superadmin tag_wrangling].freeze
   FREEZE_WORK_COMMENT_ROLES = %w[superadmin policy_and_abuse].freeze
-  HIDE_TAG_COMMENT_ROLES = %w[superadmin tag_wrangling].freeze
-  HIDE_WORK_COMMENT_ROLES = %w[superadmin policy_and_abuse].freeze
-  SPAM_ADMIN_POST_COMMENT_ROLES = %w[superadmin board board_assistants_team communications elections policy_and_abuse support].freeze
+  HIDE_TAG_COMMENT_ROLES = %w[superadmin legal tag_wrangling].freeze
+  HIDE_WORK_COMMENT_ROLES = %w[superadmin legal policy_and_abuse].freeze
+  SPAM_ADMIN_POST_COMMENT_ROLES = %w[superadmin board board_assistants_team communications elections legal policy_and_abuse support].freeze
```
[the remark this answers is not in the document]
brianjaustin: This is for approve/reject, so if not spam then this would need to be broken apart further. To me, though, those imply spam as well. Additionally, both support and PAC have this permission, so I don't see a reason to block Legal from marking things as spam/not spam

app/policies/user_policy.rb
```diff
@@ -4,7 +4,7 @@ class UserPolicy < ApplicationPolicy
   # - managing a user's invitations
   # - updating a user's email and roles (e.g. wranglers, archivists, not admin roles)
   # This is further restricted using ALLOWED_ATTRIBUTES_BY_ROLES.
-  MANAGE_ROLES = %w[superadmin policy_and_abuse open_doors support tag_wrangling].freeze
+  MANAGE_ROLES = %w[superadmin legal policy_and_abuse open_doors support tag_wrangling].freeze
```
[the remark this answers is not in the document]
brianjaustin: I think Legal having access to those is fine, but I'll defer to @sarken on that one

app/policies/comment_policy.rb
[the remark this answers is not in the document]
Bilka2: approve/reject seems to be for spam too. Are you maybe thinking of `can_review_comment?` for "approve moderated admin post comments"? That one is independent of role and always allowed.

But I agree that it sound off to allow legal to delete admin post comments, but not spam/ham admin post comments. So this also sounds like @sarken should take a look.

app/policies/user_policy.rb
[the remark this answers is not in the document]
sarken: Yeah, it's fine for Legal to have access to those, but we should update the issue to mention it.

app/policies/comment_policy.rb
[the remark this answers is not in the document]
sarken: Bilka is right that the issue is referring to moderated comments, meaning this: https://github.com/otwcode/otwarchive/blob/cd20f8e66c426944136f4673c35c70c5a6333548/app/policies/comment_policy.rb#L51-L57 
And it's not something Legal explicitly asked for; it just comes as a little side benefit of being an admin (for now -- we'll eventually limit which admins can do that, but BAT/Comms/NPM were still figuring out workflow and purview when we implemented moderated news post comments).

While there's no _harm_ in giving Legal the ability to spam comments, it's not something they're going to use, so let's leave them out of `SPAM_ADMIN_POST_COMMENT_ROLES`.

features/admins/admin_works.feature
brianjaustin: Appears not, :wastebasket:

features/admins/admin_works.feature
brianjaustin: Appears not, 🗑️

app/policies/user_policy.rb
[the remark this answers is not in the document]
brianjaustin: :+1: Added a note to the Jira bullet point about user admin pages

Bilka2 (approved):
Thank you!

Bilka2:
Wait, sorry, I missed my own comment regarding tests in `admin_users_controller_spec`. I think the "test that update doesn't allow legal to change email or roles" is needed, since no other admin role has this behaviour currently. The other two points aren't as important, since legal is treading any new ground there.

Bilka2 (approved):
Thank you!

PR #4892, opened 2024-08-05
AO3-6777 Fix that tag names could be blank

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6777

## Purpose

Forbid tag names that consist exclusively of Unicode characters from the [format category](https://www.compart.com/en/unicode/category/Cf) because they end up looking blank.

## Testing Instructions

See Jira.

## References

This validation is based on the fact that AO3-6087 created this bug by no longer stripping these characters from tag names.

## Credit

Bilka (he/him)

### DIFF
--- a/spec/models/tag_spec.rb
+++ b/spec/models/tag_spec.rb
@@ -174,7 +174,7 @@ describe Tag do
       ["200D".hex].pack("U"), # zero width joiner
       ["2060".hex].pack("U"), # word joiner
       (["200B".hex].pack("U") * 3),
-      (["200D".hex, "200D".hex, "2060".hex, "200B".hex].pack("U")),
+      ["200D".hex, "200D".hex, "2060".hex, "200B".hex].pack("U")
     ].each do |tag_name|
       tag = Tag.new
       tag.name = tag_name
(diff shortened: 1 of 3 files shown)


spec/models/tag_spec.rb
```diff
@@ -168,6 +168,39 @@ def expect_tag_update_flag_in_redis_to_be(flag)
...
+  context "tags with blank names are invalid" do
+    [
+      ["200B".hex].pack("U"), # zero width space
+      ["200D".hex].pack("U"), # zero width joiner
+      ["2060".hex].pack("U"), # word joiner
+      (["200B".hex].pack("U") * 3),
+      ["200D".hex, "200D".hex, "2060".hex, "200B".hex].pack("U")
+    ].each do |tag_name|
+      tag = Tag.new
+      tag.name = tag_name
```
brianjaustin, line 180: Could the following work instead?
```suggestion
      let(:tag) { build(:tag, name: tag_name) }
```

Bilka2: The emoji test starts failing with this and I can't tell why. The `expect(tag.name).to eq(tag_name)` line fails for all but the _last_ tag (it always returns the scientist for the tag.name), even if I attempt something like destroying the tag in the it block.

Bilka2: It always fails on the first two tags for the emojis. I guess something is going wrong with the `let` memoization. So I would rather not change this at all.

brianjaustin: Fair enough!

brianjaustin (approved):

PR #4897, opened 2024-08-19
AO3-6539 Update user creation summary h3s and pagination behavior

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6539

## Purpose

If an admin manually entered a URL with incorrect pagination params for user creations, they would see a page saying the user had no works or comments. Now it will not say that, but it will display empty list boxes with pagination links (if applicable) to be more consistent with other pages on the site.

I also:
* updated the heading on the user creation summary list boxes to say, "X - Y of Total Items"
* removed the `role="navigation"` from two lists, because that's not a valid way to use that attribute
* switched to `span_if_current` in the navigation, which is a little unnecessary, but never a bad idea for consistency
* added a "Creations" navigation item as a breadcrumb on the User Creations page

## Testing Instructions

[What needs testing](https://otwarchive.atlassian.net/browse/AO3-6539?focusedCommentId=361466)

### DIFF
--- a/app/views/admin/admin_users/_user_creations_summary.html.erb
+++ b/app/views/admin/admin_users/_user_creations_summary.html.erb
@@ -1,17 +1,20 @@
-<% unless @works.empty? %>
+<% unless @user.works.empty? %>
+  <%# Checking @user.items.empty? rather than @items.empty? allows us to display
+# an empty listbox with pagination if the admin manually enters the wrong URL.
+# This is consistent with pagination on other site pages. %>
   <div class="work listbox group" id="works-summary">
-    <h3 class="heading"><%= t(".works", count: works.size) %></h3>
+    <h3 class="heading"><%= search_header(@works, nil, "Work") %></h3>
     <%= render "works/work_abbreviated_list", works: works %>
     <%= will_paginate(works, param_name: "works_page", params: { anchor: "works-summary" }) %>
   </div>
 <% end %>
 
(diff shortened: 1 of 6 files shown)


app/views/admin/admin_users/_user_creations_summary.html.erb
```diff
@@ -1,17 +1,20 @@
-<% unless @works.empty? %>
+<% unless @user.works.empty? %>
+  <%# Checking @user.items.empty? rather than @items.empty? allows us to display
+# an empty listbox with pagination if the admin manually enters the wrong URL.
+# This is consistent with pagination on other site pages. %>
   <div class="work listbox group" id="works-summary">
-    <h3 class="heading"><%= t(".works", count: works.size) %></h3>
+    <h3 class="heading"><%= search_header(@works, nil, "Work") %></h3>
```
sarken, line 6: We just call `pluralize` on the string passed here, which is... oof. A problem for future us.

https://github.com/otwcode/otwarchive/blob/3f1a3f96c02d9424378f98e8cfc7b460d50769d1/app/helpers/search_helper.rb#L4-L37

Bilka2 (approved):
Very nice.

app/views/admin/admin_users/_user_creations_summary.html.erb
Bilka2: :scream:

PR #4898, opened 2024-08-19
AO3-6796 fix 500 error when trying to access user's works for invalid fandom_id

AliceLoeser:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6796

## Purpose

This PR fixes a 500 error when `@fandom` variable in `WorksController#index` was `nil`. 
Now the `@fandom` is retrieved using `find` which will raise an `ActiveRecord::ErrorNotFound` and redirect to 404 if the id is invalid.

## Testing Instructions

Access the following URL: `users/testuser/works?fandom_id=29292929`
It will raise an `ActiveRecord:: RecordNotFound` and be displayed as a 404 in production.

## Credit

AliceLsr (feminine pronouns)
You can find me on Jira and assign it to AliceLsr.

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -62,6 +62,9 @@ class WorksController < ApplicationController
       end
 
       tag = @fandom || @tag
+
+      return raise_not_found unless tag
+
       options[:filter_ids] ||= []
       options[:filter_ids] << tag.id
     end
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -58,12 +58,10 @@ class WorksController < ApplicationController
 
     if params[:fandom_id] || (@collection.present? && @tag.present?)
       if params[:fandom_id].present?
-        @fandom = Fandom.find_by(id: params[:fandom_id])
+        @fandom = Fandom.find(params[:fandom_id])
       end
 
       tag = @fandom || @tag
-
-      return raise_not_found unless tag
 
       options[:filter_ids] ||= []
       options[:filter_ids] << tag.id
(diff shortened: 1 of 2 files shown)


app/controllers/works_controller.rb
```diff
@@ -62,6 +62,9 @@ def index
       end
 
       tag = @fandom || @tag
+
+      return raise_not_found unless tag
```
Bilka2, line 66: It'd be preferred to raise RecordNotFound by using `Fandom.find_by!` in line 61. That will be converted to a 404 error in production and is a common pattern in this codebase.

spec/controllers/works/default_rails_actions_spec.rb
```diff
@@ -350,6 +351,14 @@ def call_with_params(params)
       expect(assigns(:fandom)).to eq(@fandom)
     end
 
+    describe "when the fandom id is invalid" do
+      it "causes a 404 error" do
+        params = { fandom_id: @invalid_fandom_id }
```
Bilka2, line 356: Please use `0` for the id here, it's a bit simpler.

Bilka2 (changes requested):
Thank you for this pull request!

It'd be appreciated if you could fill out the pull request template. I also left some review comments.

sarken:
Thank you for this pull request, AliceLoeser! I see Bilka has already left a review, but I wanted to let you know I've assigned this Jira issue to you and given your account permission to comment on, assign, and transition issues in the future.

app/controllers/works_controller.rb
AliceLoeser: @Bilka2 I fixed it by using `find` instead of `find_by!` since `find` will also raise `RecordNotFound` in the same usecase. Does that work for you?

spec/controllers/works/default_rails_actions_spec.rb
AliceLoeser: Done in 0854987af09831c12415170578c1e1f050c1a4ef

AliceLoeser:
@Bilka2 I filled the PR template! 
@sarken thanks for the jira add

Bilka2 (approved):
Thank you very much!

PR #4903, opened 2024-08-23
AO3-6760 Check admin roles in UnsortedTagsController

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6760

## Purpose

Restricts access to the Unsorted Tags bin to `superadmin`, `policy_and_abuse`, and `tag_wrangling` admins. For the PAC admins, view-only access is allowed; the other two roles have full (view/edit) access.


app/controllers/unsorted_tags_controller.rb
```diff
@@ -5,13 +5,17 @@ class UnsortedTagsController < ApplicationController
   before_action :check_permission_to_wrangle
 
   def index
+    authorize :wrangling if logged_in_as_admin?
```
Bilka2, line 8: I think it would be good to use `view_access?` directly here instead of going via the action because `index` is one of the actions where the various wrangling related controllers have different access levels.

That way the `index?` alias can be removed and we won't have someone accidentally converting e.g. TagWranglersController to using the `index?` permission and giving more roles access than intended.

Bilka2 (commented):
Just one thing!

Bilka2 (approved):
Thanks!

PR #4904, opened 2024-08-24
AO3-5421 Adds line-height to userstuff list items

anechol:
# Pull Request Checklist

## Issue

[https://otwarchive.atlassian.net/browse/AO3-5421](https://otwarchive.atlassian.net/jira/software/c/projects/AO3/issues/AO3-5421?)

## Purpose

Adds `line-height` to list item tags within `.userstuff` class.

## Testing Instructions

See Jira ticket

## Credit

anechol (she/her)

### DIFF
--- a/public/stylesheets/site/2.0/21-userstuff.css
+++ b/public/stylesheets/site/2.0/21-userstuff.css
@@ -2,12 +2,12 @@
 
 .userstuff {
     word-wrap: break-word;
+    line-height: 1.5;
 }
 
 .userstuff p, .userstuff details {
   margin: 1.286em auto;
   padding: 0;
-  line-height: 1.5;
 }
 
 /*lists */
@@ -29,7 +29,6 @@
 
 .userstuff dd {
   display: block;
-  line-height: 1.5;
   margin-block-start: 1.25em;
   margin-block-end: 1.25em;
   margin-inline-start: 3em;
@@ -48,7 +47,6 @@
 .userstuff li, .userstuff ol ul li {
   font-weight: normal;
   display: list-item;
-  line-height: 1.5;
   list-style-type: disc;
   margin-block-start: 0.75em;
   margin-inline-end: 0;
@@ -132,6 +130,7 @@
 /* quotes and stresses */
 
 .userstuff blockquote {
+  line-height: 1.5;
   margin-inline-start: 1.5em;
   padding: 0.75em;
   border-inline-start: 2px solid #999;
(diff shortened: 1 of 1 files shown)


sarken:
Hi, anechol!

I've updated the Jira issue status to In Review, so no one will mistakenly create a duplicate pull request. If you'd like the ability to comment on, assign, and transition issues in the future, you're welcome to create a Jira account! You can just reply here with the account name and we'll set up the permissions for you. (It makes things a bit easier for us on the organizational side if the Full Name on your Jira account either closely matches the name you'd like us to credit in the release notes or includes it in parentheses, e.g. "Nickname (PREFERRED NAME).")

sarken (changes requested):
Thanks for looking into this!

I noticed this doesn't address the issue with text following block quotes, but when I tried to reproduce it using the steps outlined on Jira, the text was wrapped in paragraph tags 🎉   So that's probably not a huge issue going forward, although it will probably still be present in older works.

That being said, there's at least one other place this issue can occur: in `dt` tags. (I can update the issue to mention this and remove the block quote information.)

I'm wondering if, rather than trying to fix all the individual problem areas, it would make sense to apply the fix to `.userstuff` itself. I was a little leery that it might affect the various headings, which have their own line heights, but it doesn't seem to when I give it a quick try using my browser's dev tools. 

Do you think that would work without having major side effects?

anechol:
> Thanks for looking into this!
> 
> I noticed this doesn't address the issue with text following block quotes, but when I tried to reproduce it using the steps outlined on Jira, the text was wrapped in paragraph tags 🎉 So that's probably not a huge issue going forward, although it will probably still be present in older works.
> 
> That being said, there's at least one other place this issue can occur: in `dt` tags. (I can update the issue to mention this and remove the block quote information.)
> 
> I'm wondering if, rather than trying to fix all the individual problem areas, it would make sense to apply the fix to `.userstuff` itself. I was a little leery that it might affect the various headings, which have their own line heights, but it doesn't seem to when I give it a quick try using my browser's dev tools.
> 
> Do you think that would work without having major side effects?

@sarken Thanks for clarifying all this. I think the easiest, albeit "riskier" solution is to just add the line-height to the `userstuff` class as you mentioned, and yeah, as far as I can tell in the browser, it doesn't seem to be affecting any of the headings. I've gone ahead and committed a change reflecting that. Let me know what you think or if there are any other tags to be mindful of.

sarken (approved):
This looks good to me now, both in terms of code and in terms of what I see in the browser -- thanks again! (And apologies for not getting back to you sooner; I was under the weather for a few weeks.)

PR #4907, opened 2024-08-30
AO3-6010 Show admin post actions to the correct roles

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6010

## Purpose

Change the permissions check for admin nav on the show admin post (news post) page to make sure admins who can moderate comments actually see that button. (I also updated edit/delete to check the exact permissions, instead of the generic `can_post?` in case those become different in the future.)

## References

We discussed this in chat elsewhere, and while it's not really related to the issue at hand, it will reduce confusion in the long run so we can do it now.

### DIFF
--- a/app/views/admin_posts/show.html.erb
+++ b/app/views/admin_posts/show.html.erb
@@ -6,16 +6,20 @@
     <!--/descriptions-->
     <!--subnav-->
     <div class="navigation actions module" role="navigation">
-      <% if policy(@admin_post).can_post? %>
+      <% if logged_in_as_admin? %>
         <h4 class="landmark heading"><%= t(".navigation.admin.landmark") %></h4>
         <ul class="management actions">
-          <li><%= link_to t(".navigation.admin.edit"), edit_admin_post_path(@admin_post) %></li>
-          <li>
-            <%= link_to t(".navigation.admin.delete"),
-                  @admin_post,
-                  data: { confirm: t(".navigation.admin.confirm_delete") },
-                  method: :delete %>
-          </li>
+          <% if policy(@admin_post).edit? %>
+            <li><%= link_to t(".navigation.admin.edit"), edit_admin_post_path(@admin_post) %></li>
+          <% end %>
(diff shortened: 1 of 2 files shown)


app/views/admin_posts/show.html.erb
```diff
@@ -6,16 +6,20 @@
     <!--/descriptions-->
     <!--subnav-->
     <div class="navigation actions module" role="navigation">
-      <% if policy(@admin_post).can_post? %>
+      <% if logged_in_as_admin? %>
```
sarken, line 9: Could we update [this test](https://github.com/otwcode/otwarchive/blob/2055928e5d26ae0271fe121aa3f32f0c3c0d0a8e/features/admins/admin_post_news.feature#L275) to use a role that isn't able to access the edit and delete links but should be able to access the unreviewed comment link?

sarken (approved):

app/views/admin_posts/show.html.erb
sarken, line 9: Oh, dammit, _now_ that I merged this, I realized what was bugging me about this /o\ 

This should be `<% if policy(@admin_post).can_post? || (logged_in_as_admin? && @admin_post.find_all_comments.unreviewed_only.exists?) %>` to ensure we're not including an unnecessary heading and empty `ul` in cases where there are no unreviewed comments and you're logged in as an admin who can't post.

Do you want to do a pull request or should I?

brianjaustin: Visually there's no difference, but I guess there could be issues with screen readers. I have time to do another PR, but I think it should be
```rb
      <% if policy(@admin_post).edit? || policy(@admin_post).destroy? || (logged_in_as_admin? && @admin_post.find_all_comments.unreviewed_only.exists?) %>
```
just in case we ever have a role that can _post_ but not edit or delete. Either that, or I'd change the if checks for edit/destroy back to check `can_post?` too

PR #4912, opened 2024-09-16
AO3-6268 Fix All challenge owners/moderators get CCed in "Assignments sent" notification

Cesium-Ice:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6268

## Purpose

What does this PR do?

## Testing Instructions

## References

## Credit

Cesium-Ice, they/them

they/them pronouns.*

### DIFF
--- a/app/controllers/challenge_assignments_controller.rb
+++ b/app/controllers/challenge_assignments_controller.rb
@@ -219,11 +219,10 @@ class ChallengeAssignmentsController < ApplicationController
   def default
     @challenge_assignment.defaulted_at = Time.now
     @challenge_assignment.save
-    offer_byline = @challenge_assignment.offer_byline
-    request_byline = @challenge_assignment.request_byline
+    
     assignments_page_url = collection_assignments_url(@challenge_assignment.collection)
     
-    @challenge_assignment.collection.notify_maintainers_challenge_default(offer_byline, request_byline, assignments_page_url)
+    @challenge_assignment.collection.notify_maintainers_challenge_default(@challenge_assignment, assignments_page_url)
 
     flash[:notice] = "We have notified the collection maintainers that you had to default on your assignment."
     redirect_to user_assignments_path(current_user)
--- a/app/controllers/challenge_assignments_controller.rb
+++ b/app/controllers/challenge_assignments_controller.rb
@@ -219,9 +219,10 @@ class ChallengeAssignmentsController < ApplicationController
   def default
     @challenge_assignment.defaulted_at = Time.now
     @challenge_assignment.save
-    @challenge_assignment.collection.notify_maintainers("Challenge default by #{@challenge_assignment.offer_byline}",
-        "Signed-up participant #{@challenge_assignment.offer_byline} has defaulted on their assignment for #{@challenge_assignment.request_byline}. " +
-        "You may want to assign a pinch hitter on the collection assignments page: #{collection_assignments_url(@challenge_assignment.collection)}")
+    subject = t("user_mailer.collection_notification.challenge_default.subject", offer_byline: @challenge_assignment.offer_byline )
+    message = t("user_mailer.collection_notification.challenge_default.complete", offer_byline: @challenge_assignment.offer_byline, request_byline: @challenge_assignment.request_byline, assignments_page_link: collection_assignments_url(@challenge_assignment.collection))
+    @challenge_assignment.collection.notify_maintainers(subject, message)
+
     flash[:notice] = "We have notified the collection maintainers that you had to default on your assignment."
     redirect_to user_assignments_path(current_user)
   end
--- a/app/controllers/challenge_assignments_controller.rb
+++ b/app/controllers/challenge_assignments_controller.rb
@@ -219,9 +219,11 @@ class ChallengeAssignmentsController < ApplicationController
   def default
     @challenge_assignment.defaulted_at = Time.now
     @challenge_assignment.save
-    subject = t("user_mailer.collection_notification.challenge_default.subject", offer_byline: @challenge_assignment.offer_byline )
-    message = t("user_mailer.collection_notification.challenge_default.complete", offer_byline: @challenge_assignment.offer_byline, request_byline: @challenge_assignment.request_byline, assignments_page_link: collection_assignments_url(@challenge_assignment.collection))
-    @challenge_assignment.collection.notify_maintainers(subject, message)
+    offer_byline = @challenge_assignment.offer_byline
+    request_byline = @challenge_assignment.request_byline
+    assignments_page_url = collection_assignments_url(@challenge_assignment.collection)
+    
+    @challenge_assignment.collection.notify_maintainers_challenge_default(offer_byline, request_byline, assignments_page_url)
 
     flash[:notice] = "We have notified the collection maintainers that you had to default on your assignment."
     redirect_to user_assignments_path(current_user)
--- a/app/models/challenge_assignment.rb
+++ b/app/models/challenge_assignment.rb
@@ -374,7 +374,7 @@ class ChallengeAssignment < ApplicationRecord
       end
     end
     REDIS_GENERAL.del(progress_key(collection))
-    @maintainers = collection.get_maintainers_list
+    @maintainers = collection.maintainers_list
     @maintainers.each do |i|
       UserMailer.potential_match_generation_notification(collection.id, i.email).deliver_later
     end 
--- a/app/models/challenge_assignment.rb
+++ b/app/models/challenge_assignment.rb
@@ -298,7 +298,7 @@ class ChallengeAssignment < ApplicationRecord
     collection.assignments.each do |assignment|
       assignment.send_out
     end
-    collection.notify_maintainers_challenge_sent
+    collection.notify_maintainers_assignments_sent
 
     # purge the potential matches! we don't want bazillions of them in our db
     PotentialMatch.clear!(collection)
@@ -386,9 +386,16 @@ class ChallengeAssignment < ApplicationRecord
       end
     end
     REDIS_GENERAL.del(progress_key(collection))
-    collection.maintainers_list.each do |user|
-      I18n.with_locale(user.preference.locale.iso) do
-        UserMailer.potential_match_generation_notification(collection.id, user.email).deliver_later
+
+    if !collection.email.blank?
+      UserMailer.potential_match_generation_notification(collection.id, collection.email).deliver_later
+    elsif collection.parent && !collection.parent.email.blank?
+      UserMailer.potential_match_generation_notification(collection.id, collection.parent.email).deliver_later
+    else
+      collection.maintainers_list.each do |user|
+        I18n.with_locale(user.preference.locale.iso) do
+          UserMailer.potential_match_generation_notification(collection.id, user.email).deliver_later
+        end
       end 
     end 
   end
--- a/app/models/challenge_assignment.rb
+++ b/app/models/challenge_assignment.rb
@@ -387,9 +387,9 @@ class ChallengeAssignment < ApplicationRecord
     end
     REDIS_GENERAL.del(progress_key(collection))
 
-    if !collection.email.blank?
+    if collection.email.present?
       UserMailer.potential_match_generation_notification(collection.id, collection.email).deliver_later
-    elsif collection.parent && !collection.parent.email.blank?
+    elsif collection.parent && collection.parent.email.present?
       UserMailer.potential_match_generation_notification(collection.id, collection.parent.email).deliver_later
     else
       collection.maintainers_list.each do |user|
--- a/app/models/challenge_assignment.rb
+++ b/app/models/challenge_assignment.rb
@@ -14,29 +14,29 @@ class ChallengeAssignment < ApplicationRecord
   validate :signups_match, on: :update
   def signups_match
     if self.sent_at.nil? &&
-      self.request_signup.present? &&
-      self.offer_signup.present? &&
-      !self.request_signup.request_potential_matches.pluck(:offer_signup_id).include?(self.offer_signup_id)
+       self.request_signup.present? &&
+       self.offer_signup.present? &&
+       !self.request_signup.request_potential_matches.pluck(:offer_signup_id).include?(self.offer_signup_id)
       errors.add(:base, ts("does not match. Did you mean to write-in a giver?"))
     end
   end
 
-  scope :for_request_signup, lambda {|signup| where('request_signup_id = ?', signup.id)}
+  scope :for_request_signup, ->(signup) { where("request_signup_id = ?", signup.id) }
 
-  scope :for_offer_signup, lambda {|signup| where('offer_signup_id = ?', signup.id)}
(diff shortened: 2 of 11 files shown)


Cesium-Ice:
Running rubocop locally on the files changed in this PR brings up a lot of changes that the rubocop check doesn't seem to catch, should I commit the changes made by the rubocop auto-corrector as well?

Bilka2 (changes requested):
I left some notes. Also, could you write a cucumber scenario where the email notification is sent to two maintainers and translated for one and untranslated for the other? The cucumber steps for translated emails defined in `email_custom_steps.rb` should be helpful for checking the translation status.

> should I commit the changes made by the rubocop auto-corrector as well?

It makes reviewing a lot easier to not have style corrections for unrelated lines. But on the other hand, it would be nice to have a consistent style. If you're up for it, you could make it a separate commit from the rest of the changes.

app/models/collection.rb
```diff
@@ -334,15 +334,16 @@ def prompt_meme?
     return self.challenge_type == "PromptMeme"
   end
 
-  def get_maintainers_email
-    return self.email if !self.email.blank?
-    return parent.email if parent && !parent.email.blank?
-    "#{self.maintainers.collect(&:user).flatten.uniq.collect(&:email).join(',')}"
+  def maintainers_list
+    @maintainers_list = self.maintainers.collect(&:user).flatten.uniq
```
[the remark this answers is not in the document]
Cesium-Ice: This both sets it as an instance variable and returns it. Rubocop complains about an unnecessary return when I return without setting the instance variable.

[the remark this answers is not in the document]
Bilka2: I think rubocop might complain about an extra `return` keyword, since Ruby returns the last statement by default. Hopefully it should be happy with this though:
```suggestion
    self.maintainers.collect(&:user).flatten.uniq
```

Cesium-Ice:
> I left some notes. Also, could you write a cucumber scenario where the email notification is sent to two maintainers and translated for one and untranslated for the other? The cucumber steps for translated emails defined in `email_custom_steps.rb` should be helpful for checking the translation status.

Do you have suggestions on how to go about setting up a collection with two maintainers in cucumber? 

```
When I am logged in as "mod1"
      And I go to "Holiday Swap" collection's page
      And I follow "Profile"
      And I follow "Membership"
      And I fill in "Add new members" with "participant1"
      And I press "Submit"
```
This gets as far as adding participant1 as a member but I don't know how to interact with the drop-down selector to change them into a collection owner

Bilka2:
> Do you have suggestions on how to go about setting up a collection with two maintainers in cucumber?

There is a `I have added a co-moderator "..." to collection "..."` step that should hopefully do the job.

Cesium-Ice:
I split off the notify_maintainers function because it was being used for different messages and  I couldn't move the message translation into the locale block if I didn't know what message was actually going to be received, unsure if this is the best way to go about things

app/models/collection.rb
```diff
@@ -111,49 +105,49 @@
...
-    too_long: ts("must be less than %{max} characters long.", max: ArchiveConfig.SUMMARY_MAX)
+  validates :description,
+            length: { allow_blank: true,
+                      maximum: ArchiveConfig.SUMMARY_MAX,
+                      too_long: ts("must be less than %{max} characters long.", max: ArchiveConfig.SUMMARY_MAX) }
 
-  validates_format_of :header_image_url, allow_blank: true, with: URI::regexp(%w(http https)), message: ts("is not a valid URL.")
-  validates_format_of :header_image_url, allow_blank: true, with: /\.(png|gif|jpg)$/, message: ts("can only point to a gif, jpg, or png file."), multiline: true
+  validates :header_image_url, format: { allow_blank: true, with: URI::DEFAULT_PARSER.make_regexp(%w[http https]), message: ts("is not a valid URL.") }
+  validates :header_image_url, format: { allow_blank: true, with: /\.(png|gif|jpg)$/, message: ts("can only point to a gif, jpg, or png file."), multiline: true }
```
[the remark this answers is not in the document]
Cesium-Ice: I don't understand what this is asking for...

Cesium-Ice:
Do you know why rubocop seems to be inconsistent about the things it flags as errors vs the things it flags as warnings?

app/models/collection.rb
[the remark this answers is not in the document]
Bilka2: Pulled from a [different PR](https://github.com/otwcode/otwarchive/pull/4692) that also came across this warning:
> If you haven't had a chance to dig in, https://stackoverflow.com/a/577675 has some good information (I had it handy because we made a similar change in #4601).
> 
> TLDR - if someone changes the input to be multiline (like a textarea) they could, in theory enter something like
> 
> ```
> sneaky.jpeg
> haha, another line
> ```
> 
> IIRC in most languages, `$` and `\z` end up being functionally equivalent, but not in Ruby

So it's asking for the `with: /\.(png|gif|jpg)$/` to be changed to `with: /\.(png|gif|jpg)\z/`

Bilka2:
Generally rubocop flags most things as warnings, but I have found that the error level doesn't matter much, usually all the things should be fixed. There might be some strangeness going on with the reporting of the warnings in the github actions due to some reviewdog changes, but rubocop itself should be working fine, especially locally.

app/models/collection.rb
```diff
@@ -321,28 +315,44 @@ def gift_notification
...
   end
+
   def prompt_meme?
-    return self.challenge_type == "PromptMeme"
+    self.challenge_type == "PromptMeme"
   end
 
-  def get_maintainers_email
-    return self.email if !self.email.blank?
-    return parent.email if parent && !parent.email.blank?
```
Bilka2, line 339: More annoying things, sorry.

These two lines of code are also needed for all the changed collection email sending. So if the collection email or parent collection email are set, send an untranslated email there. If not, send individual emails to all maintainers translated to the language they have selected.

app/models/collection.rb
```diff
@@ -321,28 +315,44 @@ def gift_notification
...
-    "#{self.maintainers.collect(&:user).flatten.uniq.collect(&:email).join(',')}"
+  def maintainers_list
+    self.maintainers.collect(&:user).flatten.uniq
+  end
+
+  def notify_maintainers_challenge_sent
+    # loop through maintainers and send each a notice via email
+    self.maintainers_list.each do |user|
+      I18n.with_locale(user.preference.locale.iso) do
+        UserMailer.collection_notification(self.id, I18n.t("user_mailer.collection_notification.assignments_sent.subject"), I18n.t("user_mailer.collection_notification.assignments_sent.complete"), user.email).deliver_later
```
Bilka2, line 344: Would be nice if you could pull the subject and message out into variables so this method call doesn't get so long.

app/models/collection.rb
```diff
@@ -321,28 +315,44 @@ def gift_notification
...
 
-  def get_maintainers_email
-    return self.email if !self.email.blank?
-    return parent.email if parent && !parent.email.blank?
-    "#{self.maintainers.collect(&:user).flatten.uniq.collect(&:email).join(',')}"
+  def maintainers_list
+    self.maintainers.collect(&:user).flatten.uniq
+  end
+
+  def notify_maintainers_challenge_sent
```
Bilka2, line 340:
```suggestion
  def notify_maintainers_assignments_sent
```
to use the same wording as the locale key.

app/controllers/challenge_assignments_controller.rb
```diff
@@ -219,9 +219,10 @@ def default_all
   def default
     @challenge_assignment.defaulted_at = Time.now
     @challenge_assignment.save
-    @challenge_assignment.collection.notify_maintainers("Challenge default by #{@challenge_assignment.offer_byline}",
-        "Signed-up participant #{@challenge_assignment.offer_byline} has defaulted on their assignment for #{@challenge_assignment.request_byline}. " +
-        "You may want to assign a pinch hitter on the collection assignments page: #{collection_assignments_url(@challenge_assignment.collection)}")
+    subject = t("user_mailer.collection_notification.challenge_default.subject", offer_byline: @challenge_assignment.offer_byline )
+    message = t("user_mailer.collection_notification.challenge_default.complete", offer_byline: @challenge_assignment.offer_byline, request_byline: @challenge_assignment.request_byline, assignments_page_link: collection_assignments_url(@challenge_assignment.collection))
+    @challenge_assignment.collection.notify_maintainers(subject, message)
```
Bilka2, line 224: The `t()` calls here should also be pulled into `I18n.with_locale` block in the `notify_maintainers` method, which could maybe be renamed to `notify_maintainers.challenge_default` to signify that specialization/split of the method.

app/controllers/challenge_assignments_controller.rb
```diff
@@ -219,9 +219,10 @@ def default_all
   def default
     @challenge_assignment.defaulted_at = Time.now
     @challenge_assignment.save
-    @challenge_assignment.collection.notify_maintainers("Challenge default by #{@challenge_assignment.offer_byline}",
-        "Signed-up participant #{@challenge_assignment.offer_byline} has defaulted on their assignment for #{@challenge_assignment.request_byline}. " +
-        "You may want to assign a pinch hitter on the collection assignments page: #{collection_assignments_url(@challenge_assignment.collection)}")
+    subject = t("user_mailer.collection_notification.challenge_default.subject", offer_byline: @challenge_assignment.offer_byline )
+    message = t("user_mailer.collection_notification.challenge_default.complete", offer_byline: @challenge_assignment.offer_byline, request_byline: @challenge_assignment.request_byline, assignments_page_link: collection_assignments_url(@challenge_assignment.collection))
```
Bilka2, line 223: Please use `assignments_page_url` as the interpolation variable name for `collection_assignments_url`.

features/gift_exchanges/notification_emails.feature
```diff
@@ -1,6 +1,35 @@
...
+      And I start signing up for "Holiday Swap"
+      And I press "Submit"
+    Then I should see "Sign-up was successfully created."
+
+    When I am logged in as "participant2"
+      And I start signing up for "Holiday Swap"
+      And I press "Submit"
+    Then I should see "Sign-up was successfully created."
+
+    Given I have added a co-moderator "mod2" to collection "Holiday Swap"
```
Bilka2, line 18:
```suggestion
    When I have added a co-moderator "mod2" to collection "Holiday Swap"
```

features/gift_exchanges/notification_emails.feature
```diff
@@ -1,6 +1,35 @@
...
+      And a locale with translated emails
+      And the user "mod1" enables translated emails
+    When I close signups for "Holiday Swap"
```
Bilka2, line 21:
```suggestion
      And I close signups for "Holiday Swap"
```

features/gift_exchanges/notification_emails.feature
```diff
@@ -1,6 +1,35 @@
...
+      And I have generated matches for "Holiday Swap"
+      And I have sent assignments for "Holiday Swap"
+
+    Then 4 emails should be delivered
+      And "mod1" should receive 1 email
+      And the email to "mod1" should be translated
```
Bilka2, line 27: For the mods it would be nice to check that it's also the right email, e.g. by checking the subject. The subject will be in English for both the translated and non-translated email, because the fake locale used by mod1 doesn't have translations for the subject and falls back to English.

Bilka2 (changes requested):
The approach to split `notify_maintainers` is a good idea! I left comments regarding that and some other things, which includes some more annoyances with the target email addresses.

Cesium-Ice:
Sorry for all the commits but it seems like rubocop really wants me to internationalize collection.rb

app/models/collection.rb
```diff
@@ -52,170 +53,163 @@ def clean_up_challenge
...
-    too_long: ts("must be less than %{max} characters long.", max: ArchiveConfig.TITLE_MAX)
-  validates_format_of :name,
-    message: ts('must begin and end with a letter or number; it may also contain underscores. It may not contain any other characters, including spaces.'),
-    with: /\A[A-Za-z0-9]\w*[A-Za-z0-9]\Z/
-  validates_length_of :icon_alt_text, allow_blank: true, maximum: ArchiveConfig.ICON_ALT_MAX,
-    too_long: ts("must be less than %{max} characters long.", max: ArchiveConfig.ICON_ALT_MAX)
-  validates_length_of :icon_comment_text, allow_blank: true, maximum: ArchiveConfig.ICON_COMMENT_MAX,
-    too_long: ts("must be less than %{max} characters long.", max: ArchiveConfig.ICON_COMMENT_MAX)
+  validates :name, presence: { message: I18n.t("collection.validation.name.presence") }
+  validates :name, uniqueness: { message: I18n.t("collection.validation.name.uniqueness") }
```
Cesium-Ice, line 109: I can add write a migration script to add a database constraint so rubocop stops complaining about this, but it feels kinda weird to do that just to please the linter... should I do that or suppress the warning in some way?

Bilka2: Let's not, that would go completely out of the scope of the original PR. Furthermore, please revert 450dbeb5b9d068841d91fc5f143b9234a22390bf and ec338b0cf2b7c00da62ceec451504a15736aa105. While the dedication to i18n is appreciated, it would take a lot of extra work to [correctly localise](https://github.com/otwcode/otwarchive/wiki/Internationalization-%28i18n%29-Standards#user-content-Models) the errors and this PR is already quite large. We also risk a bug slipping through because manual QA wont test this i18n.

Reviewdog/Rubocop will complain about the ts, but that's fine since we're aware that we wont be handling those in this PR.

Cesium-Ice: Ok, I've reverted the requested commits

Bilka2 (commented):
Just a few more comments.

Regarding tests, because this has quite a few changes to the "challenge default" email I think it'd be good to test that as well, with a setup similar to the other test you added.

Unfortunately I don't see a test for the fallback email behaviour (preferring the collection email, then the parent collection email) in any of the existing files. I'd appreciate if you could add a test for that as well, just to make sure the emails are getting sent to the right addresses. No need to check for translation behaviour there.

app/models/challenge_assignment.rb
```diff
@@ -209,16 +203,27 @@ def offering_pseud
...
+    end
   end
 
-
   def offer_byline
-    offer_signup && offer_signup.pseud ? offer_signup.pseud.byline : (pinch_hitter ? (pinch_hitter.byline + "* (pinch hitter)") : "- none -")
+    if offer_signup && offer_signup.pseud
+      offer_signup.pseud.byline
+    else
+      (pinch_hitter ? (pinch_hitter.byline + "* (pinch hitter)") : "- none -")
```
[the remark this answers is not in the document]
Cesium-Ice: Is there a particular reason why None is capitalized request byline but not offer byline?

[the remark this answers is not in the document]
Bilka2: Probably not, the previous code is quite old. From what I could find out from talking to people, this basically never shows up in emails, since the user would have to be deleted for that.

Bilka2 (approved):
Thank you very much for wrangling this annoying bit of code!

PR #4915, opened 2024-09-17
AO3-6768 update browser page title in drafts page

mystyrust:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6768 

## Purpose

fixes the browser title to be consistent with other pages in the users profile

## Testing Instructions

follow instructions as listed in jira ticket

## References

similar to https://github.com/otwcode/otwarchive/pull/4916 (may have merge conflict with this pr since both edit the same en.yml file)

## Credit

mystyrust, she/her


brianjaustin:
Thank you for the PR @mystyrust! Someone should be along to review it soon. In the meantime, I have assigned the Jira ticket to your user and marked it as "In Review" so no one else picks it up by accident. I've also given your Jira user some basic permissions, including the ability to assign yourself to tickets in the future.

app/controllers/works_controller.rb
```diff
@@ -161,6 +161,8 @@ def drafts
       redirect_to logged_in? ? user_path(current_user) : new_user_session_path
       return
     end
+    
+    @page_subtitle = ts("%{username} - Drafts", username: @user.login)
```
Bilka2, line 165: Please follow the suggestion by rubocop regarding i18n. The wiki page on [i18n](https://github.com/otwcode/otwarchive/wiki/Internationalization-(i18n)-Standards) should help explain how it works, but the gist is that this should be a `t()` call with a locale key and the text itself should be in the locale file (`/config/locales/controllers/en.yml`).

Bilka2 (changes requested):
Thank you for contributing!

The code works, but it would be nice to get it internationalized (i18n). Furthermore, it would be nice to have a test for this change - the `I should see the page title "..."` cucumber step should fit into one of the scenarios in `/features/works/work_drafts.feature`.

app/controllers/works_controller.rb
mystyrust: Hi, updated based on the i18n format. 
I think it's worth noting that all the errors and other page subtitles on the work_controller file do not follow this format.

Bilka2: Indeed, there is a lot of old code that doesn't follow the latest style guidelines and i18n guidelines. That's why we have tools like rubocop to automatically notify us when new or modified code needs to be adjusted to follow the new guidelines, so we can convert the code over as we go.

app/controllers/works_controller.rb
```diff
@@ -159,6 +159,8 @@ def drafts
       redirect_to logged_in? ? user_path(current_user) : new_user_session_path
       return
     end
+    
+    @page_subtitle = t('users.page_subtitle.drafts_page', username: @user.login)
```
Bilka2, line 163: Speaking of style guidelines, please follow rubocop and use double quotes here. Generally it's a good idea to follow the linter warnings unless something looks very off about them.

Furthermore, prefer relative keys for localisation. Other spots that set the page subtitle use `t(".page_title", ...)`. Rails will deduce the path to the key automatically based on the controller and action name, so that it in the yml file it should be placed in `works.drafts.page_title`.

features/works/work_drafts.feature
```diff
@@ -67,6 +67,7 @@ Feature: Work Drafts
       And the draft "draft to post"
     When I am on drafter's works page
     Then I should see "Drafts (1)"
+      And the page title should include "drafter - Drafts"
```
Bilka2, line 70: As indicated by the [failing test](https://github.com/otwcode/otwarchive/actions/runs/10926698849/job/30331184518#step:12:108), this is in the wrong spot. It should be one line lower, after the test navigates to the Drafts page.

config/locales/controllers/en.yml
```diff
@@ -108,3 +108,6 @@ en:
       ban_notice_html: Your account has been banned. You are not permitted to add or edit archive content. Please %{contact_abuse_link} for more information.
       contact_abuse: contact Abuse
       suspension_notice_html: Your account has been suspended until %{suspended_until}. You may not add or edit content until your suspension has been resolved. Please %{contact_abuse_link} for more information.
+  works:
+    drafts:
+      page_title: "%{username} - Drafts"
```
Bilka2, line 113: The test fails because of the missing newline at the end of the file.

Bilka2 (changes requested):

features/works/work_drafts.feature
mystyrust: oh thanks, I've been having trouble running cucumber tests locally

Bilka2 (approved):
Thank you!

PR #4926, opened 2024-09-20
AO3-6812 Modify cookie rotator to accept SHA256 cookies

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6812

## Purpose

Change the cookie rotator to force everyone to be backwards compatible (when before it didn't actually do anything :sob:)

## Testing Instructions

## Credit

Bilka (he/him)

### DIFF
--- a/config/initializers/cookie_rotator.rb
+++ b/config/initializers/cookie_rotator.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
-# As part of the Rails 7 upgrade, we need to convert legacy (SHA1) cookies to SHA256.
-# This can be removed after it has been in production for a little bit.
+# Due to the rolled back deploy, some users have SHA256 cookies.
+# Read them, but write back SHA1 cookies (writing is based on current setting of config.active_support.key_generator_hash_digest_class)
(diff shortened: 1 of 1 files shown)


config/initializers/cookie_rotator.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
-# As part of the Rails 7 upgrade, we need to convert legacy (SHA1) cookies to SHA256.
-# This can be removed after it has been in production for a little bit.
+# Due to the rolled back deploy, some users have SHA256 cookies.
+# Read them, but write back SHA1 cookies (writing is based on current setting of config.active_support.key_generator_hash_digest_class)
```
brianjaustin, line 4: Since we ended up not pushing this out, could you update the comment to something along the lines of
```suggestion
# To support rolling deploys, force all session cookies to be compatible with older versions.
# Read the cookies, but write back SHA1 cookies (writing is based on current setting of config.active_support.key_generator_hash_digest_class)
```

Bilka2: I took this suggestion and added the steps we need to take in the future and a short explanation.

brianjaustin (approved):
Thank you!

PR #4934, opened 2024-09-28
AO3-6765 Improve Skin Preview Message

BoldestDungeon:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6765
## Purpose

* Adds the skin ID to the preview tip (rather than making the user copy-paste it themselves)
* Wraps the "Return" button in a paragraph tag so that it is not a bulleted list item
* Adjusts wording in message

## Testing Instructions

Updated preview message following the changes in the ticket should look like this:
image

## References

## Credit

David Bilsky/Ironskink (He/Him)

### DIFF
--- a/app/controllers/skins_controller.rb
+++ b/app/controllers/skins_controller.rb
@@ -1,5 +1,4 @@
 class SkinsController < ApplicationController
-
   before_action :users_only, only: [:new, :create, :destroy]
   before_action :load_skin, except: [:index, :new, :create, :unset]
   before_action :check_title, only: [:create, :update]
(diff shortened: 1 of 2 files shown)


app/controllers/skins_controller.rb
```diff
@@ -120,9 +114,10 @@ def update
   def preview
     flash[:notice] = []
     flash[:notice] << ts("You are previewing the skin %{title}. This is a randomly chosen page.", title: @skin.title)
-    flash[:notice] << ts("Go back or click any link to remove the skin.")
-    flash[:notice] << ts("Tip: You can preview any archive page you want by tacking on '?site_skin=[skin_id]' like you can see in the url above.")
-    flash[:notice] << "<a href='#{skin_path(@skin)}' class='action' role='button'>".html_safe + ts("Return To Skin To Use") + "</a>".html_safe
+    flash[:notice] << ts("Go back or follow any link to remove the skin.")
+    flash[:notice] << (ts("Tip: You can preview any archive page you want by adding '?site_skin=%{skin_id}' to the end of the URL.", skin_id: @skin.id) +
+                      "<p><a href='#{skin_path(@skin)}' class='action'>".html_safe + ts("Return to Skin to Use") + "</a></p>".html_safe)
```
BoldestDungeon, line 119: Apparently the linter had a field day with this file when I saved it - the three lines above are the only three deliberately edited in this file, the rest was the auto-linting, sorry.

features/other_b/skin_public.feature
```diff
@@ -120,7 +120,8 @@ Feature: Public skins
     And I follow "Preview"
   Then I should be on the works tagged "Major Crimes"
     And I should see "You are previewing the skin Usable Skin. This is a randomly chosen page."
-    And I should see "Go back or click any link to remove the skin"
-    And I should see "Tip: You can preview any archive page you want by tacking on '?site_skin=[skin_id]' like you can see in the url above."
-  When I follow "Return To Skin To Use"
+    And I should see "Go back or follow any link to remove the skin"
+    And I should see "Tip: You can preview any archive page you want by adding '?site_skin="
+    And I should see "' to the end of the URL."
```
BoldestDungeon, line 125: I'm not familiar enough with Ruby test suites to have been able to figure out how to put the actual test-skin ID into this string, so split it up into the before and after parts. If there's a better way to do that I would love to learn :)

app/controllers/skins_controller.rb
brianjaustin, line 119: Same thing about the translation helper, but also noting that this does not follow our current [internationalisation standards](https://github.com/otwcode/otwarchive/wiki/Internationalization-%28i18n%29-Standards). Generally, we'd now prefer something like
```suggestion
    flash[:notice] << t(".preview_tip_html",
      skin_id: @skin.id,
      return_to_skin_link: link_to(t(".return_to_skin"), skin_path(@skin)))
```
with the corresponding keys in https://github.com/otwcode/otwarchive/blob/master/config/locales/controllers/en.yml. (I haven't run the above through a linter, so some tweaks will probably be needed for it to work perfectly.)

Additionally, make sure to swap out the single quotes around `'site_skin?=%{skin_id}'` to be double quotes `"site_skin?=%{skin_id}"` (this should require less escaping when things are moved into the YAML file, so I'd recommend doing that move first).

features/other_b/skin_public.feature
brianjaustin: I think this is fine; if we wanted to check the actual ID, we would need to add a custom step to https://github.com/otwcode/otwarchive/blob/master/features/step_definitions/skin_steps.rb (which IMO is overkill here)

brianjaustin (changes requested):

Bilka2:
Since this PR was stalled for over a year, the Jira issue has been adopted at #5744

PR #4917, opened 2024-09-17
AO3-6762: Restrict access to actions on TagsController

Cesium-Ice:
# Pull Request Checklist

* [x] Have you read ["How to write the perfect pull request"](https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/)?
* [x] Have you read the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [x] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)?
* [x] Have you added the [Jira](https://otwarchive.atlassian.net) issue number
  as the *first* thing in your pull request title (e.g. `AO3-1234 Fix thing`)
* [x] Do you have fewer than 5 pull requests already open? If not, please wait
  until they are reviewed and merged before creating new pull requests.

## Issue

https://otwarchive.atlassian.net/browse/AO3-6762

## Purpose

What does this PR do?

## Testing Instructions

How can the Archive's QA team verify that this is working as you intended?

If you have a Jira account with access, please update or comment on the issue
with any new or missing testing instructions instead.

## References

Are there other relevant issues/pull requests/mailing list discussions?

## Credit

What name and pronouns should we use to credit you in the [Archive of Our Own's Release Notes](https://archiveofourown.org/admin_posts?tag=1)?

Cesium-Ice, they/them

If you have a Jira account, please include the same name in the "Full name"
field on your Jira profile, so we can assign you the issues you're working on.

*Please note that if you do not fill in this section, we will use your GitHub account name and
they/them pronouns.*

PR #4937, opened 2024-10-04
AO3-6761 Check admin roles in TagWranglingsController

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6758
https://otwarchive.atlassian.net/browse/AO3-6761

## Purpose

Restrict access to PAC (tag bins only) and wrangling/super admins (wrangling tools, mass update)

## References

This is probably going to cause conflicts with https://github.com/otwcode/otwarchive/pull/4917; please merge that one first if this is in the same release, then I will fix the conflicts.


brianjaustin:
I noticed that "Wrangling Tools" in the left-hand navigation menu is selected, even when on mass wranglings page:
wrangling navigation menu with "Wrangling Tools" and "Fandoms by media (0)" selected

Do we want to hide the 'Wrangling Tools" item (which is not selectable) for any admin who does not have access to the general wrangling tools page?

sarken:
Huh, that's weird. Just to clarify because I can't pull the code right now: if you're on, e.g., `https://test.archiveofourown.org/tag_wranglings?show=fandoms` with this new code, "Wrangling Tools" is is a span instead of a link? Is this for all admins or just admins who can't access Wrangling Tools?

Regardless, I think the fix for that might be changing the link in the sidebar to explicitly check that the params are blank.

https://github.com/otwcode/otwarchive/blob/f607097eacad333ee18fc92d3578026802f801e0/app/views/tag_wranglings/_wrangler_dashboard.html.erb#L6


```
 <li><%= span_if_current(ts('Wrangling Tools'), tag_wranglings_path, params[:show].blank?) %></li> 
```

(It's possible there's more than one version of this sidebar.)

brianjaustin:
@sarken that's actually the case _now_, not just with the new code... :grimacing: I don't object to fixing it here, but the main question I had was if we should hide that part of the menu for non-write admins (iirc the ticket doesn't say either way)

sarken:
Ah, okay! I thought it was a new bug, and you were asking if the way to deal with it was to hide that particular link while leaving the others visible.

I think we can leave the full sidebar visible for anyone with either type of access and just stick to this plan from AO3-6758 for everyone else:

> The final pull request will need to hide the major navigational elements: 
> * the “Tag Wrangling” item in the admin navigation from all admin roles except superadmin, policy_and_abuse, and tag_wrangling 
> * the sidebar on https://test.archiveofourown.org/tags/search and https://test.archiveofourown.org/tags/testing from all admin roles except superadmin and tag_wrangling

sarken:
(Feel free to fix or make a new issue for the `span_if_current` thing as desired!)

spec/controllers/tag_wranglings_controller_spec.rb
```diff
@@ -4,60 +4,146 @@
...
-  before do
-    fake_login
-    controller.current_user.roles << Role.new(name: "tag_wrangler")
+  full_access_roles = %w[superadmin tag_wrangling].freeze
+  read_access_roles = %w[superadmin policy_and_abuse tag_wrangling].freeze
+
+  shared_examples "an action only authorized admins can access" do |authorized_roles:|
+    before do
+      fake_login_admin(admin)
+      subject
```
Bilka2, line 13: I'd prefer the subject call to be in the `it` blocks, since that is what is being tested, but it doesn't really matter.

spec/controllers/tag_wranglings_controller_spec.rb
```diff
@@ -4,60 +4,146 @@
...
       let(:tag1) { create(:character) }
       let(:tag2) { create(:character) }
+      let(:success) do
+        expect(tag1.reload.canonical?).to be(true)
+        expect(tag2.reload.canonical?).to be(true)
+      end
 
       before do
         post :wrangle, params: { canonicals: [tag1.id, tag2.id] }
       end
```
Bilka2, line 115: With the subject calls in the two shared examples, I think this is no longer needed

app/views/tag_wranglings/_wrangler_dashboard.html.erb
```diff
@@ -1,39 +1,43 @@
-<div id="dashboard" role="navigation region" class="tag wrangling region">
-  <ul class="navigation actions">
-    <% if current_user.is_a?(User) %>
-      <li><%= span_if_current(ts('Wrangling Home'), tag_wrangler_path(current_user)) %></li>
+<% if (logged_in_as_admin? && policy(:wrangling).full_access?) || current_user.try(:is_tag_wrangler?) || @counts %>
+  <div id="dashboard" role="navigation region" class="tag wrangling region">
+    <% if (logged_in_as_admin? && policy(:wrangling).full_access?) || current_user.try(:is_tag_wrangler?) %>
```
Bilka2, line 3: Based on the comments on this PR and on Jira, the sidebar should be visible to read_access admins:

> I think we can leave the full sidebar visible for anyone with either type of access

app/controllers/tag_wranglings_controller.rb
```diff
@@ -8,7 +8,10 @@ class TagWranglingsController < ApplicationController
 
   def index
     @counts = tag_counts_per_category
-    unless params[:show].blank?
+    if params[:show].blank?
+      authorize :wrangling, :full_access? if logged_in_as_admin?
+    else
+      authorize :wrangling, :read_access? if logged_in_as_admin?
```
Bilka2, line 14: And then based on hiding the option from the navbar but not the sidebar, I think wrangling tools should be visible to read_access admins. PAC doesnt care either way and it's the only way they'd be able navigate to the mass bins.

Bilka2 (commented):
Should there be a test for the navbar visibility?

I think it would be nice if you could i18n the wrangler_dashboard, none of the other pull requests touched it so hopefully it shouldn't lead to merge conflicts.

app/views/tag_wranglings/_wrangler_dashboard.html.erb
Bilka2: Might be good to still hide `Wranglers` and `New Tag` from PAC though, since PAC can't access those.

Bilka2 (approved):
Thank you!

app/views/tag_wranglings/_wrangler_dashboard.html.erb
```diff
@@ -1,39 +1,62 @@
-<div id="dashboard" role="navigation region" class="tag wrangling region">
-  <ul class="navigation actions">
-    <% if current_user.is_a?(User) %>
-      <li><%= span_if_current(ts('Wrangling Home'), tag_wrangler_path(current_user)) %></li>
+<% if (logged_in_as_admin? && policy(:wrangling).full_access?) || current_user.try(:is_tag_wrangler?) || @counts %>
```
Bilka2, line 1: :thinking: I thought this needed to be `read_access?` as well but the tests pass without it?

Bilka2: It doesn't show on the tags index page, because that one doesn't have `@counts` set. So this should be changed and the tests should all start out in the tags index, not the tag wrangling tools

PR #4940, opened 2024-10-12
AO3-6211 Exclude tags without posted works from unwrangled bins

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6211

## Purpose

Exclude tags that are only used on draft works from the tag wrangling bins for unwrangled tags.

## Testing Instructions

See Jira.

## Credit

Bilka (he/him)


sarken:
I'll talk to the wrangling chairs who might want to consult with their team!

sarken:
TW chairs have confirmed not showing external work fandoms would be great, thanks! Do you happen to know if there is also a test that shows bookmark-only tags don't show up in bins even after being sorted (i.e., marked a Relationship tag from the Unsorted bin) and assigned to a fandom? If not, it might be good to add a test for that as well.

features/step_definitions/tag_steps.rb
```diff
@@ -374,10 +374,21 @@
   click_button("Save changes")
 end
 
+When "I assign the tag {string} to the fandom {string}" do |tag, fandom|
```
sarken, line 377: Could we update/repurpose the existing step that adds the fandom to character tags instead, or is it important we visit the page instead of using `add_association`? 

https://github.com/otwcode/otwarchive/blob/6443729393955d8a75a4f4e86772428aabdc9191/features/step_definitions/tag_steps.rb#L87-L91

features/step_definitions/tag_steps.rb
```diff
@@ -374,10 +374,21 @@
...
+  click_button("Save changes")
+end
+
 When /^I view the (canonical|synonymous|unfilterable|unwrangled|unwrangleable) (character|relationship|freeform) bin for "(.*?)"$/ do |status, type, tag|
   visit wrangle_tag_path(Tag.find_by(name: tag), show: type.pluralize, status: status)
 end
 
+When /^I view the (character|relationship|freeform|fandom) mass bin$/ do |type| # rubocop:disable Cucumber/RegexStepName
+  visit tag_wranglings_path(show: type.pluralize)
+end
```
sarken, line 390: Could we define these in support/paths.rb instead? The main benefit of that is it gives us "Given I am on the x mass bin page" and "When I go to the x mass bin page" and "Then I should be on the x mass bin page," which can all be useful.

sarken (approved):
This is fine as-is, just two little test-related suggestions since it'll be a while before we can get this merged. (Tag Wrangling is going to be so happy when we do!)

PR #4945, opened 2024-10-16
AO3-6823 <added validation for invalid tags and update of autocomplete…

chanthakoun2002:
## Issue

This PR addresses [AO3-6823](https://otwarchive.atlassian.net/browse/AO3-6823).  

## Purpose

This pull request:

Adds validation for invalid tags to ensure that:
Tags are less than 100 characters.
Tags do not contain restricted characters: , ^ * < > { } = \ %.
Ensures that invalid tags do not appear in the autocomplete feature.
Removes existing invalid tags from the Redis autocomplete cache.

## Testing Instructions

Try to add tags with more than 100 characters.
adding restricted characters i.e   ^ * < > { } = \ % to tags
Verify that Tags with these invalid characteristics trigger an error message.
Valid tags should be successfully added.
It should be successful if invalid tags are not visible in autocomplete.
Existing invalid tags should have been removed from Redis.

## Credit

GitHub: chanthakoun2002
pronouns: He/Him

they/them pronouns.*

### DIFF
--- a/app/models/tagset_models/tag_set.rb
+++ b/app/models/tagset_models/tag_set.rb
@@ -7,6 +7,9 @@ class TagSet < ApplicationRecord
   TAG_TYPES_RESTRICTED_TO_FANDOM = %w(character relationship)
   TAGS_AS_CHECKBOXES = %w(category rating archive_warning)
 
+  INVALID_CHARACTERS = /[,\^*<>{}=`\\%]/
+  MAX_LENGTH = 100
+
   attr_accessor :from_owned_tag_set
 
   has_many :set_taggings, dependent: :destroy
@@ -72,9 +75,15 @@ class TagSet < ApplicationRecord
   # [type]_tagnames
   # [type]_tagnames_to_add/_remove
   after_save :assign_tags
+
+  def self.invalid_tag?(tag_name)
+    tag_name.length > MAX_LENGTH || tag_name.match?(INVALID_CHARACTERS)
+  end
+
   def assign_tags
     tags_to_add = []
     tags_to_remove = []
+    invalid_tags = [] # collect invalid tags
 
     TAG_TYPES.each do |type|
       if self.instance_variable_get("@#{type}_tagnames")
@@ -83,6 +92,12 @@ class TagSet < ApplicationRecord
         old_tags = self.with_type(type.classify)
         tags_to_add += (new_tags - old_tags)
         tags_to_remove += (old_tags - new_tags)
+        if TagSet.invalid_tag?(tag.name)
+          invalid_tags << tag.name
+        else
+          tags_to_add << tag
+        end
+      end
       else
         #
         unless self.instance_variable_get("@#{type}_tagnames_to_add").blank?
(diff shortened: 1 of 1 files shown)


app/models/tagset_models/tag_set.rb
```diff
@@ -83,6 +92,12 @@ def assign_tags
         old_tags = self.with_type(type.classify)
         tags_to_add += (new_tags - old_tags)
         tags_to_remove += (old_tags - new_tags)
+        if TagSet.invalid_tag?(tag.name)
```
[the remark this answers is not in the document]
Bilka2: Furthermore, this `if self.instance_variable_get("@#{type}_tagnames")` code path is only taken when using tags for prompts/gift exchanges, which cannot be non-canonical. The problematic code for manually adding code to owned tag sets goes via `unless self.instance_variable_get("@#{type}_tagnames_to_add").blank?`.

app/models/tagset_models/tag_set.rb
```diff
@@ -250,17 +271,21 @@ def remove_from_autocomplete
 
   def add_tags_to_autocomplete(tags_to_add)
     tags_to_add.each do |tag|
+      next if TagSet.invalid_tag?(tag.name) # skips invalid tags in autocomplete
```
Bilka2, line 274: Due to your changes above, the code never gets to adding tags to the tag set if any tag is invalid - so this extra check is not necessary.

app/models/tagset_models/tag_set.rb
```diff
@@ -250,17 +271,21 @@ def remove_from_autocomplete
...
     end
   end
 
-  def remove_tags_from_autocomplete(tags_to_remove)
-    tags_to_remove.each do |tag|
-      value = tag.autocomplete_value
-      REDIS_AUTOCOMPLETE.zrem("autocomplete_tagset_all_#{self.id}", value)
-      REDIS_AUTOCOMPLETE.zrem("autocomplete_tagset_#{tag.type.downcase}_#{self.id}", value)
+  # cleanup of invalid tags for autocomplete
+  def remove_invalid_tags_from_autocomplete
```
Bilka2, line 282: This method can't just be repurposed like this, it's used for removing tags from the autocomplete when removing them from the tag set.

Instead, to do the cleanup suggested in the Jira issue, you'd have to write a rake task that goes through all tag sets and removes the invalid tags from the tag set's autocomplete. This is optional according to the Jira issue, so feel free to leave it out.

If you do want to create the cleanup task, there is even more to be done. The current code adds the invalid tags to the `set_taggings` of the tag set, with a nil tag on the set_tagging. So the cleanup code should remove those set_taggings as well.

Bilka2 (changes requested):
Thanks for taking a look at this tag set code! I suggested a different approach for the validation and cleanup of invalid tags. Furthermore, this PR is missing an automated test for the issue and the interface text change to inform tag set moderators about the restrictions for the tag formats. Please add these two things in addition to changing the validation strategy.

Bilka2:
Hi @chanthakoun2002! It's been a year now without progress on this, so could you let us know within the next two weeks if you have the time to update this PR? If you don’t, or if we don’t hear from you in that time, I'll close this PR and unassign the Jira issue so that someone else can pick this up. Thank you!

chanthakoun2002:
My apologies i completely forgot about this. I don’t currently have the time to update this PR due to other commitments, so I understand if it needs to be closed and unassigned for someone else to take over.

PR #4947, opened 2024-10-19
AO3-6810 Don't send "Automatically flagged as spam" email when hiding works manually

EchoEkhi:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6810

## Purpose

When manually hiding a work that has been marked as spam by Akismet or otherwise, don't send the "Automatically flagged as spam" email, send the "Your work has been hidden by the Policy & Abuse team" email.

## Testing Instructions

See JIRA ticket.

## Credit

EchoEkhi (He/him)


app/models/work.rb
```diff
@@ -1146,7 +1151,7 @@ def mark_as_ham!
   def notify_of_hiding
     return unless hidden_by_admin? && saved_change_to_hidden_by_admin?
     users.each do |user|
-      if spam?
+      if spam? && hidden_for_spam
```
Bilka2, line 1154:
```suggestion
      if spam? && saved_change_to_spam?
```
This should achieve the same thing as the extra attribute, but it's not as fragile as an attribute that can be forgotten about.

features/admins/admin_works.feature
```diff
@@ -32,6 +32,27 @@ Feature: Admin Actions for Works, Comments, Series, Bookmarks
       | superadmin       |
       | legal            |
       | policy_and_abuse |
+  
+    Scenario Outline: Can hide works already marked as spam
```
Bilka2, line 36: This scenario title should also mention something regarding the correct email (so not the spam one) getting sent.

Bilka2 (changes requested):
Please action the comments by reviewdog, thank you!

I removed the rest of my review comments - my alternative approach didn't work, but I think the current approach is also problematic because it's prone to someone forgetting to set the attribute.

sarken:
Would it help if we put the `hide_spam` and `bulk_review(ids)` methods in charge of sending the `admin_spam_work_notification` email instead? 

I thought `notify_of_hiding` would also need `return if spam? && saved_change_to_spam?` to go with that change, but the test seems to pass without it, so I'm not sure.

Speaking of tests, though, I think we should probably also test which emails are sent when using the bulk hiding option on the Moderated Works page. It looks like the current tests only cover the option on the work itself.

EchoEkhi:
I don't think the `return if spam? && saved_change_to_spam?` works, since it's effectively a rising edge detector, and the decision to send which email is not dependent on whether if there was a rising edge (spam changed from false to true), instead it only depends on which button the admin clicks.

I initially chose the approach of a virtual attribute because the alternative of changing the `notify_of_hiding` function to not be a `after_save` hook and instead have individual controllers call either `notify_of_hiding` or `notify_of_hiding_for_spam` could result in a situation where if a coder makes a mistake and forgets to call either of these functions when hiding a work, _no emails would be sent at all_ and the user would not be aware that anything had happened. Whereas if the coder makes the mistake of forgetting to set the virtual attribute, the wrong email would be sent, but at least there would be an email.

I have changed the virtual attribute to be internal to the model. Instead of setting the attribute, the coder calls the `notify_of_hiding_for_spam` to send the hidden-for-spam emails, and that function would also set the attribute to prevent the normal hidden-by-admin emails from being sent. This way it's not possible for a coder to make a mistake in the future and cause a situation where no emails are sent when the work is hidden since the hidden-by-admin email is sent by default.

EchoEkhi:
ps. I think that test might be flakey

EchoEkhi:
nevermind it was my fault, fixed it now

app/controllers/admin/user_creations_controller.rb
```diff
@@ -37,8 +37,11 @@ def set_spam
     action = "mark as " + (params[:spam] == "true" ? "spam" : "not spam")
     AdminActivity.log_action(current_admin, @creation, action: action, summary: @creation.inspect)    
     if params[:spam] == "true"
+      unless @creation.hidden_by_admin
+        @creation.notify_of_hiding_for_spam if @creation_class == Work
+        @creation.hidden_by_admin = true
+      end
       @creation.mark_as_spam!
-      @creation.update_attribute(:hidden_by_admin, true)
```
Bilka2, line 41: This change is suspicious to me. update_attribute does the change and saves it. The `hidden_by_admin = true` makes the change, but doesn't save it and #save isn't called anywhere later.

This might explain the test that was failing

EchoEkhi: I think the `update_attribute(:spam, true)` inside the `@creation.mark_as_spam!` function saves it. This was a deliberate change because otherwise it would have been saved twice, so the after_save hook would have been called twice, and two emails would have been sent, the hidden-for-spam email first then the tos-violation email second.

The test was failing because I changed the test steps and it caused an unrelated test elsewhere to fail

Bilka2: The `Given /^the spam work "([^\"]*)"$/` step? I am confused why that needed modification, the `before_save :hide_spam` should set `hidden_by_admin` even with your changes

EchoEkhi: The `Given /^the spam work "([^\"]*)"$/` step needed to be changed because I wanted to test cases where spam works aren't hidden automatically

> the before_save :hide_spam should set hidden_by_admin even with your changes

Only if the `admin_settings.hide_spam?` is set to true. The 'mark as spam' button is supposed to hide the work regardless of the admin setting.

EchoEkhi: Oh sorry I just realized I misunderstood you there, yeah you're right, I'll try rolling back the test step definition changes when I get a chance

brianjaustin (changes requested):
A couple suggestions in addition to the Rubocop

brianjaustin (approved):
Thanks!

PR #4948, opened 2024-10-27
AO3-6775 Add delete work mailer preview

weeklies:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6775

Not sure how to test this with the feature test changes, and I would appreciate some direction.

## Purpose

Add delete work mailer preview. 
## Testing Instructions

See Jira issue.


Bilka2:
The simplest example for a test for localised emails is likely the feature test included in #4921.

test/mailers/previews/user_mailer_preview.rb
```diff
@@ -40,6 +40,21 @@ def claim_notification_registered
...
+    UserMailer.delete_work_notification(work.pseuds.first.user, work)
+  end
+
+  def delete_work_by_co_creator
+    first_creator = create(:user, :for_mailer_preview)
+    second_creator = create(:user, :for_mailer_preview)
+
+    work = create(:work, authors: [first_creator.default_pseud, second_creator.default_pseud])
+
+    User.current_user = second_creator
```
Bilka2, line 54: This is a possible way to do it, but it feels a bit hacky. What do you think about instead pulling the self vs other check out of the mailer erb files into work.rb and only passing the result into the mailer method?

test/mailers/previews/user_mailer_preview.rb
```diff
@@ -40,6 +40,21 @@ def claim_notification_registered
     UserMailer.claim_notification(creator_id, [work.id], true)
   end
 
+  def delete_work
```
Bilka2, line 43:
```suggestion
  def delete_work_notification_self
```

app/mailers/user_mailer.rb
```diff
@@ -348,12 +348,10 @@ def delete_work_notification(user, work)
...
 
-    I18n.with_locale(@user.preference.locale.iso) do
-      mail(
-        to: user.email,
-        subject: t('user_mailer.delete_work_notification.subject', app_name: ArchiveConfig.APP_SHORT_NAME)
-      )
-    end
+    mail(
+      to: user.email,
+      subject: t('user_mailer.delete_work_notification.subject', app_name: ArchiveConfig.APP_SHORT_NAME)
```
Bilka2, line 353: It'd be nice if you replace this with `default_i18n_subject`

Bilka2:
Another example for a cucumber test for the email translations is now available with #4959. It shows something that would be nice to have: Generating emails for two users at once, one translated and one non-translated. For work deletions that should be achievable by deleting a work with multiple creators.

weeklies:
Putting this up for adoption, since I've been busy.

PR #4960, opened 2024-11-14
AO3-5792 Add missing class to additional tag groups in Tag Sets to fix skins

slavalamp:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5792

The issue only mentions Reversi but Happy 17th! and Low Vision Default also have the wrong border color there.

## Purpose

Adds a missing "group" class to additional tag groups in Tag Sets so that they're targeted by Reversi, Happy 17th! and Low Vision Default site skins correctly.

This does not affect the default skin or most other public skins. Out of the three that are affected: Fixie (911) seems to get fixed by this as well, The Hustings (889) also becomes consistent (can't say "fixed" because the contrast there is terrible), and Panda Madness (890) doesn't account for these tag groups and has them look a bit out of place either way.

## Testing Instructions

1. Create a tag set with **31+** Additional tags _(the number in the Jira issue is incorrect)_
2. Switch skin to Reversi, or use ?site_skin=929 switch
3. The tag group headers in the Additional tags section should have dark backgrounds

## References

n/a

## Credit

slavalamp (they/them)


sarken (changes requested):
Thanks for working on this, and for noting the issue with the number of tags mentioned in the Jira issue! I've updated the issue with the new information.

sarken (approved):
Looks great now -- thank you!

PR #4966, opened 2024-11-21
AO3-5977 Hide "Approve All Unreviewed Comments" button from admins on works

WelpThatWorked:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5977

## Purpose

Hides the "Approve All Unreviewed Comments" button if an admin is viewing a work with comment moderation turned on.

## Credit

Jake Faulkner

### DIFF
--- a/app/views/comments/unreviewed.html.erb
+++ b/app/views/comments/unreviewed.html.erb
@@ -3,7 +3,7 @@
 <!--/descriptions-->
 
 <!--Subnavigation, sorting and actions-->
-<% if @comments[0] && can_review_comment?(@comments[0]) %>
+<% if can_review_all_comments?(@commentable) %>
   <ul class="navigation actions">
     <% path = @commentable.is_a?(AdminPost) ? review_all_admin_post_comments_path(@commentable) : review_all_work_comments_path(@commentable) %>
     <li><%= button_to(t(".approve_all"), path, method: :put) %></li>
--- a/app/views/comments/unreviewed.html.erb
+++ b/app/views/comments/unreviewed.html.erb
@@ -3,10 +3,12 @@
 <!--/descriptions-->
 
 <!--Subnavigation, sorting and actions-->
-<ul class="navigation actions">
-  <% path = @commentable.is_a?(AdminPost) ? review_all_admin_post_comments_path(@commentable) : review_all_work_comments_path(@commentable) %>
-  <li><%= button_to(t(".approve_all"), path, method: :put) %></li>
-</ul>
+<% if @comments[0] && can_review_comment?(@comments[0]) %>
+  <ul class="navigation actions">
+    <% path = @commentable.is_a?(AdminPost) ? review_all_admin_post_comments_path(@commentable) : review_all_work_comments_path(@commentable) %>
+    <li><%= button_to(t(".approve_all"), path, method: :put) %></li>
+  </ul>
+<% end %>
 <!--/subnav-->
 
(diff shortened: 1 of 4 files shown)


app/controllers/comments_controller.rb
```diff
@@ -283,6 +283,7 @@ def unreviewed
       .unreviewed_only
       .for_display
       .page(params[:page])
+    @reviewable = (@commentable.is_a?(AdminPost) || current_user_owns?(@commentable))
```
sarken, line 286: I think a helper method might make more sense than a variable here. We have a bunch of methods for checking which comment-related buttons to display in the comments helper -- you might even be able to reuse this one: https://github.com/otwcode/otwarchive/blob/5ea268b75c40d40b7fc1f6094bfeaacaed712257/app/helpers/comments_helper.rb#L207-L211

sarken (changes requested):
Hi, Jake, and thanks for the pull request! 

I've left a review comment, and while there's already [a test in features/admins/admin_post_news.feature](https://github.com/otwcode/otwarchive/blob/5ea268b75c40d40b7fc1f6094bfeaacaed712257/features/admins/admin_post_news.feature#L289-L295) that checks that the button _is_ visible on news posts, we'd appreciate it if you could add a test to [features/admins/admin_works.feature](https://github.com/otwcode/otwarchive/blob/5ea268b75c40d40b7fc1f6094bfeaacaed712257/features/admins/admin_works.feature) that ensures the button isn't visible on works.

I've assigned the Jira issue to you and set its status to In Review so no one will mistakenly create a duplicate pull request. I've also given your account permission to comment on, assign, and transition issues in the future.

WelpThatWorked:
Test added, threw in a check for the per-comment "Approve" button as well, since I don't see it anywhere else. 

Used the helper, and decided to let it hide the button on works with no unreviewed comments as well, since the button would do nothing in that case.

app/views/comments/unreviewed.html.erb
```diff
@@ -3,10 +3,12 @@
 <!--/descriptions-->
 
 <!--Subnavigation, sorting and actions-->
-<ul class="navigation actions">
-  <% path = @commentable.is_a?(AdminPost) ? review_all_admin_post_comments_path(@commentable) : review_all_work_comments_path(@commentable) %>
-  <li><%= button_to(t(".approve_all"), path, method: :put) %></li>
-</ul>
+<% if @comments[0] && can_review_comment?(@comments[0]) %>
```
sarken, line 6: I think `@comments.first` would be the more typical way to get the first comment in the array. 

That being said, if we have to deal with the possibilities of empty arrays here, it's probably tidier to define a new helper method that takes the commentable, e.g. 

```
def can_review_all_comments?(commentable)
  commentable.is_a?(AdminPost) || is_author_of?(commentable)
end
```

features/step_definitions/web_steps.rb
```diff
@@ -257,6 +257,12 @@ def with_scope(locator)
   end
 end
 
+Then /^(?:|I )should not see button "([^"]*)"(?: within "([^"]*)")?$/ do |text, selector|
+  with_scope(selector) do
+    page.body.should_not have_button(text)
+  end
+end
+
```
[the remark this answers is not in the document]
WelpThatWorked: Thanks, not sure how I overlooked these.

sarken (approved):
Thank you!

PR #4968, opened 2024-11-22
AO3-6841 Remove TOS FAQ updated date

lydia-theda:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6841 

## Purpose

Removes the Updated Date from the TOS FAQ page

## Credit

lydia-theda


sarken (approved):

sarken:
It's not a dealbreaker and I'm going to merge this, but just a heads up that when you're submitting a second pull request for the same issue, it's a little cleaner if you either
- delete the branch and create a new one with the same name, or
- create a new branch with a name like AO3-1234_BOT
That way, you don't end up dragging the old commits (in this case, everything before [559600b](https://github.com/otwcode/otwarchive/pull/4968/commits/559600bf785d01d42598d003ca1650bf2e27abe6)) along with you.

PR #4972, opened 2024-11-30
AO3-6832 Don't attach work download to abuse report if provided comment URL

WelpThatWorked:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6832

## Purpose

If the URI indicates a comment or a work's comment thread, don't attach an HTML download of the work to the ticket.

## Testing Instructions

Bypass/Comment out initial report send, fill in dummy `ticket_id`  ([link](https://github.com/otwcode/otwarchive/blob/dfaa140ed3b2d221d03d298040a94392fd6673b0/app/models/abuse_report.rb#L109C5-L120C31)), then verify `attach_work_download` returns early for non-work URLs

## Credit

Jake Faulkner

### DIFF
--- a/app/models/abuse_report.rb
+++ b/app/models/abuse_report.rb
@@ -124,9 +124,9 @@ class AbuseReport < ApplicationRecord
   end
 
   def attach_work_download(ticket_id)
-    comment_id = url[%r{/comments/(\d*)}, 1]
+    is_not_comments = url[%r{/comments/}, 0].nil?
     work_id = url[%r{/works/(\d+)}, 1]
-    return unless work_id and not comment_id
+    return unless work_id && is_not_comments
 
     work = Work.find_by(id: work_id)
     ReportAttachmentJob.perform_later(ticket_id, work) if work
--- a/app/models/abuse_report.rb
+++ b/app/models/abuse_report.rb
@@ -124,8 +124,9 @@ class AbuseReport < ApplicationRecord
   end
 
   def attach_work_download(ticket_id)
+    comment_id = url[%r{/comments/(\d*)}, 1]
     work_id = url[%r{/works/(\d+)}, 1]
-    return unless work_id
+    return unless work_id and not comment_id
 
     work = Work.find_by(id: work_id)
     ReportAttachmentJob.perform_later(ticket_id, work) if work
--- a/spec/models/abuse_report_spec.rb
+++ b/spec/models/abuse_report_spec.rb
@@ -359,6 +359,13 @@ describe AbuseReport do
         .not_to have_enqueued_job
     end
 
+    it "does not attach a download for comment sub-URLs asynchronously" do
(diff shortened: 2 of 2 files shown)


app/models/abuse_report.rb
```diff
@@ -124,8 +124,9 @@ def send_report
   end
 
   def attach_work_download(ticket_id)
+    comment_id = url[%r{/comments/(\d*)}, 1]
     work_id = url[%r{/works/(\d+)}, 1]
-    return unless work_id
+    return unless work_id and not comment_id
```
brianjaustin, line 129: In addition to the Rubocop change, I think the condition would read a little more easily with the `.blank?` method – IMO `not` can get a little bit unwieldy with `unless`
```suggestion
    return unless work_id && comment_id.blank?
```

brianjaustin (changes requested):
Please add a unit test for this case

app/models/abuse_report.rb
WelpThatWorked: `.blank?` wouldn't work here, but I changed it to avoid the `not`, is that better?

spec/models/abuse_report_spec.rb
```diff
@@ -359,6 +359,13 @@
         .not_to have_enqueued_job
     end
 
+    it "does not attach a download for comment sub-URLs asynchronously" do
+      allow(subject).to receive(:url).and_return("http://archiveofourown.org/works/#{work.id}/comments/")
```
brianjaustin, line 363: Thanks! I notice on the Jira there's no trailing slash, and I think (assuming I read the regex right) that case would still fail. Would you mind updating this test case, or perhaps adding another that matches what's described in the linked Jira ticket?

WelpThatWorked: When sent through the web form, URLs have the trailing slash added in `abuse_report.rb#clean_url`, so since this test skips that step this should be accurate? I've tested the full form with all the examples in the ticket and it works as expected.

Bilka2: You're right, thanks for pointing that out!

app/models/abuse_report.rb
```diff
@@ -124,8 +124,9 @@ def send_report
   end
 
   def attach_work_download(ticket_id)
+    is_not_comments = url[%r{/comments/}, 0].nil?
     work_id = url[%r{/works/(\d+)}, 1]
-    return unless work_id
+    return unless work_id && is_not_comments
```
Bilka2, line 129: I think this may be easier to read if the variable is simply `comments` and you do the nil check in this line instead of line 127:
```suggestion
    return unless work_id && comments.nil?
```
But this is a subjective style thing so feel free not to do this, I'm already approving the PR.

Bilka2 (approved):

brianjaustin (approved):

PR #4860, opened 2024-06-18
AO3-6686 AO3-6687 Update configs to Rails 7.0 defaults

brianjaustin:
# Pull Request Checklist

* [X] Have you read ["How to write the perfect pull request"](https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/)?
* [X] Have you read the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [X] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)?
* [X] Have you added the [Jira](https://otwarchive.atlassian.net) issue number
  as the *first* thing in your pull request title (e.g. `AO3-1234 Fix thing`)
* [X] Do you have fewer than 5 pull requests already open? If not, please wait
  until they are reviewed and merged before creating new pull requests.

## Issue

https://otwarchive.atlassian.net/browse/AO3-6686
https://otwarchive.atlassian.net/browse/AO3-6687

## Purpose

Update configs to use the Rails 7.0 defaults and fix things that this breaks (namely, a few buttons and tests that redirect to example.com)

Also removes the cookie rotator, since that's related to the new config defaults

## Credit
Brian Austin (they/he)
tickinginstant for the open redirect fix

PR #4973, opened 2024-11-30
AO3-6686 Reintroduce Rails 7 defaults (without dangerous cookie stuff)

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6686

## Purpose

Resurrects https://github.com/otwcode/otwarchive/pull/4860 to upgrade us to Rails 7 defaults, except without touching cookies for the most part. ~The default format is (by default) updated to SHA256, but because of https://github.com/otwcode/otwarchive/pull/4926 it will really be a no-op until the rotator is removed in a month or so.~ Better (correct-er) explanation in the comments.


Bilka2:
> The default format is (by default) updated to SHA256, but because of https://github.com/otwcode/otwarchive/pull/4926 it will really be a no-op until the rotator is removed in a month or so.

This is not correct. The rotator will indeed rotate the cookies from SHA1 to SHA256 once it is deployed. So before this is deployed, it needs to be certain that all our other non-Rails code can deal with SHA256 cookies. (During the original issue Sep 20th this was indicated to be the case: "The worker doesn't care about the hash.")

The point of #4926 is to make sure that those rotated cookies can be read by the application code that is running on the servers right now. So if we have a situation where one application server is running this new code here and one application server is running the old code and a user gets pingponged between them, the following happens:
1. User connects to application server running new code.
2. The new rotator on that server rotates the old SHA1 cookie to SHA256 (yay, upgrade!). The user now has a SHA256 cookie and stays logged in.
3. User connects to application server running old code.
4. Due to #4926, the old rotator can read the upgraded SHA256 cookie (aw, downgrade). The user now has a SHA1 cookie and stays logged in.
5. Go to step 1.

This pingponging between code versions will happen during the rolling deploy. Once the rolling deploy is over and all servers are running the new application code, steps 3 and 4 will no longer happen as there are no more servers running the old application code. That means that over time, all users will have their cookies rotate to SHA256, which is the intended upgrade. Once that has happened then the new rotator will be a no-op because all cookies will have the new format. Since the cookies last 14 days, that should be 14 days after deploy. However, since it is a no-op there is no time pressure or complication to keeping this new rotator around for longer and I would recommend to do so. See step 3 in the comment in the cookie rotator file.

This means this PR should go on a release on its own and needs throughout testing on staging, similar to #4926. This PR here is the risky one and will explode in our faces if we misunderstood something.

Bilka2 (approved):
Just to make sure we have the right cookie setting I started Rails with this code and read `Rails.application.config.active_support.key_generator_hash_digest_class` from the console. It is SHA256 as expected.

Noting that we will need a strategy and Jira issue for the cookie serialisation change, since it is excluded here to reduce the risk of this change and do less cookie things at once.

PR #4974, opened 2024-12-01
AO3-6850 Remove pinch_request_signup from challenge assignments

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6850

## Purpose

Removes all code that referenced pinch_request_signup and adds a rake task to migrate the less than 200 uses of it on production.

## Credit

Bilka


spec/lib/tasks/after_tasks.rake_spec.rb
```diff
@@ -503,3 +503,51 @@
...
+
+    it "does nothing" do
+      expect do
+        subject.invoke
+      end.to avoid_changing { assignment.reload.request_signup_id }
+        .and output("Migrated pinch_request_signup for 0 challenge assignments.\n").to_stdout
+    end
+  end
+
+  context "for an assignment without a request_signup_id and a pinch_request_signup_id" do
```
brianjaustin, line 535: Minor thing, to me this sounds like "an assignment with neither a request_signup_id nor a pinch_request_signup_id". Maybe something like the following?
```suggestion
  context "for an assignment with a pinch_request_signup_id but no request_signup_id" do
```

brianjaustin (approved):

PR #4978, opened 2024-12-04
AO3-6792 Admins with the support and policy_and_abuse roles should be able to edit languages

Cesium-Ice:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6792 

## Purpose

Adds support and policy_and_abuse roles to language policy + validation to disallow support and policy_and_abuse roles to edit certain fields in the form

## Testing Instructions

## References

## Credit

they/them pronouns.*

### DIFF
--- a/app/controllers/languages_controller.rb
+++ b/app/controllers/languages_controller.rb
@@ -1,5 +1,4 @@
 class LanguagesController < ApplicationController
-
   def index
     @languages = Language.default_order
     @works_counts = Rails.cache.fetch("/v1/languages/work_counts/#{current_user.present?}", expires_in: 1.day) do
@@ -16,7 +15,7 @@ class LanguagesController < ApplicationController
     @language = Language.new(language_params)
     authorize @language
     if @language.save
-      flash[:notice] = t('successfully_added', default: 'Language was successfully added.')
+      flash[:notice] = t("languages.successfully_added")
       redirect_to languages_path
     else
       render action: "new"
@@ -31,8 +30,24 @@ class LanguagesController < ApplicationController
   def update
     @language = Language.find_by(short: params[:id])
     authorize @language
+    
+    unless User.current_user.roles.compact.flatten.include?("superadmin") || User.current_user.roles.compact.flatten.include?("translation")
+      
+      if User.current_user.roles.compact.flatten.include?("policy_and_abuse") && (@language.name != language_params[:name] || @language.short != language_params[:short] || @language.sortable_name != language_params[:sortable_name] || @language.support_available != (language_params[:support_available] == "1"))
+        flash[:error] = t("languages.update.policy_and_abuse_admin_error")
+        redirect_to languages_path
+        return
+      end
+
+      if User.current_user.roles.compact.flatten.include?("support") && (@language.abuse_support_available != (language_params[:abuse_support_available] == "1"))
+        flash[:error] = t("languages.update.support_admin_error")
+        redirect_to languages_path
+        return
+      end
+    end 
+
     if @language.update(language_params)
-      flash[:notice] = t('successfully_updated', default: 'Language was successfully updated.')
+      flash[:notice] = t("languages.successfully_updated")
       redirect_to languages_path
     else
       render action: "new"
@@ -40,6 +55,7 @@ class LanguagesController < ApplicationController
   end
 
   private
+
   def language_params
     params.require(:language).permit(
       :name, :short, :support_available, :abuse_support_available, :sortable_name
--- a/app/controllers/languages_controller.rb
+++ b/app/controllers/languages_controller.rb
@@ -30,21 +30,18 @@ class LanguagesController < ApplicationController
   def update
     @language = Language.find_by(short: params[:id])
     authorize @language
-    
-    unless User.current_user.roles.compact.flatten.include?("superadmin") || User.current_user.roles.compact.flatten.include?("translation")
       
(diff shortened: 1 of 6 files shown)


spec/controllers/languages_controller_spec.rb
```diff
@@ -211,5 +237,69 @@
...
+      before do
+        fake_login_admin(admin)
+        put :update, params: language_params
+      end
+      it "redirects with error" do
+        it_redirects_to_with_error(languages_path, "Support Admin cannot update the 'Abuse support available' field.")
+      end
+    end 
+
+    context "when logged in as an admin with support role and attempt to edit non-abuse fields" do
```
Cesium-Ice, line 285: This test case fails, but when I log in as a support admin on my local environment and update the support available field the change goes through, I have no idea why this is happening

Bilka2: The `false` and `true` in `language_params`, `language_params_abuse` and `language_params_support` need to be `"0"` and `"1"` to match what the form sends.

app/controllers/languages_controller.rb
```diff
@@ -31,15 +30,32 @@ def edit
   def update
     @language = Language.find_by(short: params[:id])
     authorize @language
+    
+    unless User.current_user.roles.compact.flatten.include?("superadmin") || User.current_user.roles.compact.flatten.include?("translation")
```
Bilka2, line 34: I would leave out this separate check.

app/controllers/languages_controller.rb
```diff
@@ -31,15 +30,32 @@ def edit
...
+      
+      if User.current_user.roles.compact.flatten.include?("policy_and_abuse") && (@language.name != language_params[:name] || @language.short != language_params[:short] || @language.sortable_name != language_params[:sortable_name] || @language.support_available != (language_params[:support_available] == "1"))
```
Bilka2, line 36:
```suggestion
      if !policy(@language).can_edit_other_fields? && (@language.name != language_params[:name] || @language.short != language_params[:short] || @language.sortable_name != language_params[:sortable_name] || @language.support_available != (language_params[:support_available] == "1"))
```

app/controllers/languages_controller.rb
```diff
@@ -31,15 +30,32 @@ def edit
...
+        flash[:error] = t("languages.update.policy_and_abuse_admin_error")
+        redirect_to languages_path
+        return
+      end
+
+      if User.current_user.roles.compact.flatten.include?("support") && (@language.abuse_support_available != (language_params[:abuse_support_available] == "1"))
```
Bilka2, line 42:
```suggestion
      if !policy(@language).can_edit_abuse_support_available? && (@language.abuse_support_available != (language_params[:abuse_support_available] == "1"))
```

app/policies/language_policy.rb
```diff
@@ -1,11 +1,23 @@
...
+
+  def edit?
+    user_has_roles?(LANGUAGE_EDIT_ACCESS)
+  end
+
+  def can_edit_abuse_support_available?
+    user_has_roles?(%w[superadmin translation policy_and_abuse])
+  end
+
+  def can_edit_other_fields?
```
Bilka2, line 17: I think this name makes more sense when you don't have the context of the other check:
```suggestion
  def can_edit_non_abuse_fields?
```

app/views/languages/_form.html.erb
```diff
@@ -2,51 +2,50 @@
...
 <%= form_for(@language, html: { class: "post" }) do |f| %>
 
-  <p class="required notice">* <%= t('.required_notice', default: "Required information") %></p>
+  <p class="required notice">* <%= t(".required_notice") %></p>
 
   <fieldset>
     <dl>
       <dt class="required">
-        <%= f.label :name, t('.name', default: "Name") + " *" %>
+        <%= f.label :name, "#{t('.fieldset.name')} *" %>
```
Bilka2, line 10:
```suggestion
        <%= f.label :name, "#{t(".fieldset.name")} *" %>
```

I would also remove the `.fieldset` prefix from the fieldset locale keys because they're already prefixed with `form`. But thank you for the i18n work!

app/views/languages/_form.html.erb
```diff
@@ -2,51 +2,50 @@
...
       </dt>
       <dd class="required">
-        <%= f.text_field :name %>
+        <%= f.text_field :name, disabled: !policy(@language).can_edit_other_fields? %>
       </dd>
 
       <dt class="required">
-        <%= f.label :short, t('.short', default: "Abbreviation") + " *" %>
+        <%= f.label :short, "#{t('.fieldset.short')} *" %>
```
Bilka2, line 17:
```suggestion
        <%= f.label :short, "#{t(".fieldset.short")} *" %>
```

Bilka2 (changes requested):
I left a few notes. Also, please adjust the error messages to not specifically mention the admin roles but instead mention the permissions themselves, so something like "Sorry, only an authorized admin can do update the 'Abuse support available' field."

app/controllers/languages_controller.rb
Cesium-Ice: Can there be a case where a user is both a translation admin and a support admin? If so I feel like the check is needed.

app/views/languages/_form.html.erb
Cesium-Ice: it seems like reviewdog wants the quotes to be single quotes

Bilka2 (approved):

app/views/languages/_form.html.erb
Bilka2: I see, thank you!

PR #4676, opened 2023-12-01
AO3-6647 Upgrade Ruby to 3.1.4

zz9pzza:
# Pull Request Checklist

* [x] Have you read ["How to write the perfect pull request"](https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/)?
* [x] Have you read the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [x] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)?
* [x] Have you added the [Jira](https://otwarchive.atlassian.net) issue number
  as the *first* thing in your pull request title (e.g. `AO3-1234 Fix thing`)
* [x] Do you have fewer than 5 pull requests already open? If not, please wait
  until they are reviewed and merged before creating new pull requests.

## Issue

https://otwarchive.atlassian.net/browse/AO3-6647

## Purpose

Change the version of ruby that the Archive uses.

## Testing Instructions

I would do a smoke test

## References

Are there other relevant issues/pull requests/mailing list discussions?

## Credit

He/Him

PR #4981, opened 2024-12-07
AO3-6529 Prevent work imports from AO3

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6529

## Purpose

Adds an error when trying to import works from URLs that are in the permitted hosts.

## References

PR Limit: 6 out of 5
Using reviewer bonus:
https://github.com/otwcode/otwarchive/pull/4676
https://github.com/otwcode/otwarchive/pull/4763

## Credit

Bilka


app/controllers/works_controller.rb
```diff
@@ -476,6 +476,19 @@ def import
...
+      uri = UrlFormatter.new(url).standardized
+      next unless ArchiveConfig.PERMITTED_HOSTS.include?(uri.host)
+
+      work_id = uri.path[%r{/works/(\d+)}, 1]
+      flash.now[:error] = if work_id
+                            t(".on_archive", bookmark_it_link: view_context.link_to(t(".bookmark_it"), new_work_bookmark_path(work_id)))
+                          else
+                            t(".on_archive", bookmark_it_link: t(".bookmark_it"))
+                          end
+      render(:new_import) and return
```
Bilka2, line 489: If someone has an idea how to fix the rubocop error I'd appreciate it. I couldn't come up with anything besides just suppressing it.

brianjaustin: Based on the Rubocop docs, it sounds like
```suggestion
      render(:new_import) and return nil
```
might suppress it, but IMO that's less clear than just adding a comment to tell RuboCop we do indeed want that behaviour here

Bilka2:
I had another look at this and I would really like to remove the link from the error in all cases. That would make it possible to move the code and error message into StoryParser, which would automatically handle multiple URLs.

brianjaustin:
I think that sounds reasonable, @sarken any concerns changing that part of the Jira story?

sarken:
I do not personally care about the link, _but_ I think this was someone else's pet issue, so I will ask around!

sarken:
It's PAC's issue and dropping the link is fine!

brianjaustin (approved):

PR #4989, opened 2024-12-14
AO3-6232 Fix support ticket referer field

brianjaustin:

PR #4990, opened 2024-12-14
AO3-6232 Fix referer for support tickets

brianjaustin:
Fixes an issue where referrer is always the support ticket form, plus a test I missed in #4989.

### DIFF
--- a/features/other_b/support.feature
+++ b/features/other_b/support.feature
@@ -74,7 +74,7 @@ Feature: Filing a support request
     And I press "Send"
   Then a Zoho ticket should be created with referer "http://www.example.com/works"
 
-  Scenario: Submit a request with an on-Archive referer
+  Scenario: Submit a request with a referer that is not on-Archive
 
   Given I am logged in as "puzzled"
     And basic languages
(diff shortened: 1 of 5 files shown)


features/other_b/support.feature
```diff
@@ -60,3 +60,28 @@ Feature: Filing a support request
...
+
+  Scenario: Submit a request with an on-Archive referer
+
+  Given I am logged in as "puzzled"
+    And basic languages
+    And Zoho ticket creation is enabled
+  When I go to the works page
+    And I follow "Support & Feedback"
+    And I fill in "Brief summary" with "Just a brief note"
+    And I fill in "Your question or problem" with "Hi, I came from the Archive"
```
Bilka2, line 85:
```suggestion
    And I fill in "Your question or problem" with "Hi, I didn't come from the Archive"
```

Bilka2 (commented):
The Jira issue specifically says that the referer shouldn't be a hidden field. But the hidden field is the best way to pass info from `new` to `create`. So I think this is okay but it needs a comment from @sarken whether the hidden field is acceptable.

Bilka2 (approved):
I heard back from sarken and the hidden field is okay :tada:

While the test changes would be nice to make them a bit clearer, I'm already tentatively approving since the code itself is fine.

PR #4991, opened 2024-12-15
AO3-6587 Allow # as initial character in Ticket ID Field when editing a profile or pseud

scottsds:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6587

## Purpose

This pull request will allow AO3 Support staffers to copy/paste Ticket ID numbers from their internal tools. This copied data often includes an "#" character, which staffers have to manually erase. 

Previously, a live validation check would prevent staffers from hitting the "Update" button. Now, live validation is removed and the initial "#" is automagically stripped from the copied Ticket ID before validation.

Validation will allow an "#" in the first character position. If an "#" or any other character is in any other position, validation will fail.

## Testing Instructions

Automated tests were updated to check for the above validation conditions, they can be manually checked as follows:

### Instructions for Passing Validation
1. Copy the ticket ID from the ticketing platform and note that it comes with an "#" prefix.
2. Log in to the Archive as PAC admin or superadmin.
3. Edit a user profile (https://test.archiveofourown.org/users/USERNAME/edit) OR pseud (https://test.archiveofourown.org/users/USERNAME/pseuds/PSEUDNAME.
4. Paste the ticket ID with (or without) the "#" prefix. 
5. Click 'Update'.
6. Be joyous that the update happens without error.

### Instructions for Failing Validation
1. Copy the ticket ID from the ticketing platform and note that it comes with an "#" prefix.
2. Log in to the Archive as PAC admin or superadmin.
3. Edit a user profile (https://test.archiveofourown.org/users/USERNAME/edit) OR pseud (https://test.archiveofourown.org/users/USERNAME/pseuds/PSEUDNAME.
4. Paste the ticket ID, change initial # to any other character, or add an "#" or any other character to any other position.
5. Click 'Update'.
7. Validation fails.

## References

https://otwarchive.atlassian.net/browse/AO3-6156
https://otwarchive.atlassian.net/browse/AO3-4595

### DIFF
--- a/app/models/concerns/justifiable.rb
+++ b/app/models/concerns/justifiable.rb
@@ -5,15 +5,22 @@ module Justifiable
     attr_accessor :ticket_number
     attr_reader :ticket_url
 
+    before_validation :strip_octothorpe
     validates :ticket_number,
               presence: true,
-              numericality: { only_integer: true },
+              numericality: { only_integer: true,
+                              message: "may begin with an # and otherwise contain only numbers."
+              },
               if: :enabled?
 
     validate :ticket_number_exists_in_tracker, if: :enabled?
   end
 
   private
+
+  def strip_octothorpe
+    self.ticket_number = self.ticket_number.delete_prefix("#") unless self.ticket_number.nil?
+  end
 
   def enabled?
     # Only require a ticket if the record has been changed by an admin.
--- a/app/models/concerns/justifiable.rb
+++ b/app/models/concerns/justifiable.rb
@@ -9,8 +9,7 @@ module Justifiable
     validates :ticket_number,
               presence: true,
               numericality: { only_integer: true,
-                              message: "may begin with an # and otherwise contain only numbers."
-              },
+                              message: "may begin with an # and otherwise contain only numbers." },
               if: :enabled?
 
     validate :ticket_number_exists_in_tracker, if: :enabled?
--- a/app/models/concerns/justifiable.rb
+++ b/app/models/concerns/justifiable.rb
@@ -18,7 +18,9 @@ module Justifiable
   private
 
   def strip_octothorpe
-    self.ticket_number = self.ticket_number.delete_prefix("#") unless self.ticket_number.nil?
+    unless ticket_number.is_a?(Integer)
+      self.ticket_number = self.ticket_number.delete_prefix("#") unless self.ticket_number.nil?
+    end
   end
 
   def enabled?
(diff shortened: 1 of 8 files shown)


app/views/pseuds/_pseuds_form.html.erb
```diff
@@ -62,8 +62,6 @@
     <dt><%= f.label :ticket_number, class: "required" %></dt>
     <dd>
       <%= f.text_field :ticket_number, class: "required" %>
-      <%= live_validation_for_field("pseud_ticket_number", numericality: true) %>
-      <p class="footnote"><%= t(".ticket_footnote") %></p>
```
Bilka2, line 66: To solve the test failure, the locale string associated with this key has to be removed from `config/locales/views/en.yml`. Same for the other removed `t()` call.

app/models/concerns/justifiable.rb
```diff
@@ -5,16 +5,24 @@ module Justifiable
     attr_accessor :ticket_number
     attr_reader :ticket_url
 
+    before_validation :strip_octothorpe
     validates :ticket_number,
               presence: true,
-              numericality: { only_integer: true },
+              numericality: { only_integer: true,
+                              message: "may begin with an # and otherwise contain only numbers." },
```
Bilka2, line 12: It'd be nice to [i18n](https://github.com/otwcode/otwarchive/wiki/Internationalization-%28i18n%29-Standards#user-content-Validation_errors) this error message by making it use a custom error message i18n key:

```suggestion
                              message: :numeric_with_optional_hash },
```

And then defining the translation for that key in `config/locales/models/en.yml` under `activerecord`:
```
  errors:
    messages:
      numeric_with_optional_hash: 'may begin with an # and otherwise contain only numbers.'
```

Bilka2 (changes requested):
The code looks good, but there some things to improve for i18n.

app/models/concerns/justifiable.rb
```diff
@@ -5,16 +5,24 @@ module Justifiable
     attr_accessor :ticket_number
     attr_reader :ticket_url
 
+    before_validation :strip_octothorpe
     validates :ticket_number,
               presence: true,
-              numericality: { only_integer: true },
+              numericality: { only_integer: true,
+                              message: :numeric_with_optional_hash },
```
Bilka2, line 12: i18n-tasks doesn't recognize the Active Record error message keys, so this will make it stop complaining.
```suggestion
              # i18n-tasks-use t("activerecord.errors.messages.numeric_with_optional_hash")
              numericality: { only_integer: true, message: :numeric_with_optional_hash },
```

Bilka2 (approved):
Thank you!

PR #4997, opened 2024-12-20
AO3-6015 Script to get data for guests

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6015

## Purpose

Adds a script to get data for guests/deleted users by email (i.e. for GDPR reasons)

## Credit

sarken, for the majority of the script
Brian Austin, for some YAML stuff to handle audit entries properly

### DIFF
--- a/script/get_email_data.rb
+++ b/script/get_email_data.rb
@@ -60,7 +60,7 @@ audits = ActiveRecord::Base.connection.exec_query("SELECT * FROM audits WHERE au
 audits&.rows&.each do |audit|
   action = audit[8]
   changes = YAML.safe_load(audit[9],
-                           permitted_classes: [ActiveSupport::TimeWithZone, Time, ActiveSupport::TimeZone, BCrypt::Password],
+                           permitted_classes: Rails.application.config.active_record.yaml_column_permitted_classes,
                            aliases: true)
   ip = audit[12]
   # Created or deleted account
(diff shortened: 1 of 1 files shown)


script/get_email_data.rb
```diff
@@ -0,0 +1,168 @@
...
+# the email address -- the audits with the email address are only a subset of
+# the audits we need.
+previous_user_ids = ActiveRecord::Base.connection
+  .exec_query("SELECT DISTINCT(auditable_id) FROM audits WHERE audited_changes LIKE '%#{email}%'")
+  .map { |row| row["auditable_id"] }
+audits = ActiveRecord::Base.connection.exec_query("SELECT * FROM audits WHERE auditable_id IN (#{previous_user_ids.join(',')})") if previous_user_ids.present?
+audits&.rows&.each do |audit|
+  action = audit[8]
+  changes = YAML.safe_load(audit[9],
+                           permitted_classes: [ActiveSupport::TimeWithZone, Time, ActiveSupport::TimeZone, BCrypt::Password],
```
redsummernight, line 63: You can replace this list with `Rails.application.config.active_record.yaml_column_permitted_classes`.

https://github.com/otwcode/otwarchive/blob/e785293b798883df9ac8d9840dd32ecf761c323c/config/application.rb#L82-L87

script/get_email_data.rb
```diff
@@ -0,0 +1,168 @@
...
+  if %w[create destroy].include?(action)
+    ips << ip if ip.present?
+  elsif action == "update"
+    changes.each do |k, v|
+      # Silence rubocop, we want the separate branches for documentation purposes.
+      # rubocop:disable Lint/DuplicateBranch
+      case k
+      when "accepted_tos_version"
+        ips << ip if ip.present?
+        # Changed email address
```
redsummernight, line 76: If the comments are indented like this, can you move them into the next block, e.g. `# Changed email address` under `when "email"`?

redsummernight: Yeah that works!

redsummernight (approved):

brianjaustin:
This has a query starting with a wildcard, which is not OK for the prod audits table. @zz9pzza says once we have a way to ensure that runs on secondaries it will be OK, but for now we can't guarantee that

Nil-Cipher:
> This has a query starting with a wildcard, which is not OK for the prod audits table. @zz9pzza says once we have a way to ensure that runs on secondaries it will be OK, but for now we can't guarantee that

New to the codebase and ruby so this might be a dumb question, why can't we replace the `select *` with the relevant columns if the wild card is the issue?

brianjaustin:
Hi @Nil-Cipher!

The wildcard in question is the `%`, not the `*` -- the former means we're querying for a column where it contains something not necessarily at the beginning. That's a problem because MariaDB/MySQL can't really do much to make the query efficient when the condition starts with a wildcard (`%`). Unfortunately, the gem that generates the entries in that table shoves everything into 1 single column, so we can't filter by something better for the moment

brianjaustin:
That's correct; we have some discussions going internally about how we can handle queries that are needed but expensive, but they likely won't conclude for a bit with other things going on right now as well

PR #4998, opened 2024-12-20
AO3-6866 Add no_resets and protected_user user roles to test fixtures

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6866

## Purpose

Make the role setup for new local dev environments and tests match production.

## Testing Instructions

Automated tests should pass. I also recreated my local dev database with `./script/reset_database.sh` and confirmed that the roles are created.

## Credit

Bilka

### DIFF
--- a/test/fixtures/roles.yml
+++ b/test/fixtures/roles.yml
@@ -48,3 +48,17 @@ role_00008:
   authorizable_type:
   id: 8
   authorizable_id:
+role_00009:
+  name: protected_user
+  created_at: 2024-12-20 21:45:02 Z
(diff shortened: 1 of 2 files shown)


sarken (commented):
The additions are good! Would you also want to take out the extra `translator` roles (`role_00004` and `role_00005`) and set the `authorizable_type` and `authorizable_id` on the remaining one to empty? 

You might have to update `test/fixtures/user_roles.yml` so users are only assigned the remaining translator role.

sarken (approved):

PR #5000, opened 2024-12-21
AO3-5498 AO3-931 Show placeholder for deleted invitee

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-931 (yes, that's a real ticket number :sweat_smile: )

## Purpose

When a user sends an invitation, someone accepts it, and then the invitee deletes their account, show the old invitee as a placeholder (suggested by PAC/Support elsewhere):

invitations list view with deleted user placeholder

invitation detail with deleted user placeholder

Side note: the invitation detail page is... a bit confusing as-is. I'm OK using this story to update that if we want, but I haven't done that on my own for a number of reasons.


config/locales/views/en.yml
```diff
@@ -1358,7 +1358,18 @@ en:
         what_you_cant_do: What you can't do
   invitations:
     invitation:
+      copy_and_use: copy and use
+      copy_link: Copy link
+      created_at: Created at
+      deleted_user: deleted user (%{user_id})
```
sarken, line 1364: Can you capitalize "Deleted" here so the invitee and sender will match when we do [AO3-6095](https://otwarchive.atlassian.net/browse/AO3-6095)? I was going to say I'd add the user ID to that issue, but I think it would show the ID to users, not just admins, so we probably don't want that. 

Relatedly, are we sure this only shows the user ID if you're an admin?

brianjaustin: Oh at the moment it definitely shows to both, my thought was it's not a big concern because user ID isn't PII. But it's also not particularly _useful_ to a non-admin so I am also not opposed to adding another check in the helper

sarken: Let's go with an additional check, then! I'll update AO3-6095 to request the same behavior.

app/helpers/invitations_helper.rb
```diff
@@ -1,18 +1,20 @@
 module InvitationsHelper
-
   def creator_link(invitation)
-    if invitation.creator.is_a?(User)
+    case invitation.creator
+    when User
       link_to(invitation.creator.login, invitation.creator)
-    elsif invitation.creator.is_a?(Admin)
+    when Admin
```
sarken, line 6: I might be wandering out of scope here, but while I was updating [AO3-6095](https://otwarchive.atlassian.net/browse/AO3-6095), I noticed it specifies using `Deleted admin` when the admin is deleted. We don't cover that case here, so is it possible this page would either 500 or incorrectly specify the queue if the admin who issued an invitation was now deleted?

config/locales/views/en.yml
```diff
@@ -1358,7 +1358,19 @@ en:
         what_you_cant_do: What you can't do
   invitations:
     invitation:
+      copy_and_use: copy and use
+      copy_link: Copy link
+      created_at: Created at
+      deleted_user: deleted user
+      deleted_user_with_id: deleted user (%{user_id})
```
sarken, line 1365: Could you fix the capitalization here? *shakes fist at lowercase*

sarken (commented):
One of these days, I will hit the right button on my review comments. One of them! (And no other.)

sarken:
Oh, hey, this is actually the updated version of the issue you worked on: https://otwarchive.atlassian.net/issues/AO3-5498

Would you like to grab that and finish it off?

brianjaustin:
Sure!

app/helpers/invitations_helper.rb
brianjaustin: This refactor wouldn't change the behaviour mentioned in AO3-6095. I'd say it's out of scope for this PR, since this change here is just a rubocop thing

features/other_a/invite_use.feature
```diff
@@ -29,3 +29,17 @@ I want to use an invitation to create an account
...
+      And an invitation created by "creator" and used by "invitee"
+      And I am logged in as "creator"
+    When I view the most recent invitation for "creator"
+    Then I should see "invitee"
+    When I am logged in as "invitee"
+      And "invitee" deletes their account
+      And I am logged in as "creator"
+      And I view the most recent invitation for "creator"
+    Then I should see "Deleted User"
+      But I should not see "invitee"
```
sarken, line 45: I know it should be fine, but can we also check that "invitee" doesn't appear in the table on your manage invitations page (`/users/USERNAME/invitations/manage)`?

features/other_a/invite_use.feature
```diff
@@ -29,3 +29,22 @@ I want to use an invitation to create an account
     And I go to SOME_USER2's invitations page
   Then I should be on Scott's user page
     And I should see "Sorry, you don't have permission to access the page you were trying to reach."
+
+  Scenario: Viewing invitation details when the invitee has deleted their account
+    Given the user "creator" exists and is activated
+      And the user "invitee" exists and is activated
+      And an invitation created by "creator" and used by "invitee"
+      And I am logged in as "creator"
+    When I manage invitations belonging to "creator"
```
sarken, line 38: `When I go to creator's manage invitations page` should work without adding any new step definitions:

https://github.com/otwcode/otwarchive/blob/092aa07d6b57ee5cb603a3628b1f0a89b0153783/features/support/paths.rb#L171-L172

sarken (changes requested):

sarken (approved):
Merge conflicts, but otherwise looks good!

PR #5001, opened 2024-12-23
AO3-6859 Fix admins being disallowed from tags pages

WelpThatWorked:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6859

## Purpose

Only check auth if the tag is banned, so non-wrangler admin roles can view normal tags.
Drop comment that doesn't describe the action anymore.
Tweak check so wrangler tools are fully hidden for non-wrangler admins.
Fix test that was calling the wrong (?) action.

I'm not fully certain why the auth check was added to #show, so I'm assuming that it was meant to only apply to Banned tags.

## Credit

Jake Faulkner

### DIFF
--- a/app/controllers/tags_controller.rb
+++ b/app/controllers/tags_controller.rb
@@ -60,22 +60,22 @@ class TagsController < ApplicationController
     @page_subtitle = @tag.name
     if @tag.is_a?(Banned) 
       if !logged_in_as_admin?
-        flash[:error] = ts('Please log in as admin')
+        flash[:error] = t("admin.access.not_admin_denied")
         redirect_to(tag_wranglings_path) && return
       elsif !policy(:wrangling).read_access?
-        flash[:error] = ts('Sorry, only an authorized admin can access the page you were trying to reach.')
+        flash[:error] = t("admin.access.page_access_denied")
         redirect_to(root_path) && return
       end
     end
     # if tag is NOT wrangled, prepare to show works and bookmarks that are using it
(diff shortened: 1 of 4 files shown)


Bilka2:
There should not have been a restriction per role on the show action at all since it's not mentioned in the original Jira issue. So I think it would be best to simply remove that from the controller and to adjust the tests accordingly.

For `can_wrangle?`, I think that needs to stay as is as well, without your change. It also controls the visibility of the comments button on the tags show page, which needs to stay because any admin can access those.

WelpThatWorked:
> For `can_wrangle?`, I think that needs to stay as is as well, without your change. It also controls the visibility of the comments button on the tags show page, which needs to stay because any admin can access those.

I don't think this is the intended/correct behavior. d00f3de, the commit where the permission check for viewing tag-comments (`check_tag_wrangler_access` in comments controller) was added specifies that it should be tag wranglers only. Additionally, the other two buttons in the same nav header deny access to non-wrangler admins, so I suspect that the comments controller was just overlooked in the original issue, and `check_tag_wrangler_access` should be updated in a similar way.

> There should not have been a restriction per role on the show action at all since it's not mentioned in the original Jira issue. So I think it would be best to simply remove the authorize call from the action and to adjust the tests accordingly.

Done

Bilka2:
The linked commit is from before admins had roles. I double-checked, it's intended that all admins have access to comments on tags. Let's keep the scope of this PR to only the show action permissions and test and leave the rest as is.

PR #5004, opened 2024-12-25
AO3-6726 separate languages used for works from other uses of languages

Cesium-Ice:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6726 

## Purpose

What does this PR do?

## Testing Instructions

## References

## Credit

they/them pronouns.*

### DIFF
--- a/app/controllers/abuse_reports_controller.rb
+++ b/app/controllers/abuse_reports_controller.rb
@@ -27,12 +27,12 @@ class AbuseReportsController < ApplicationController
   private
 
   def load_abuse_languages
-    @abuse_languages = Language.where(abuse_support_available: true).default_order
+    @abuse_languages = LocaleLanguage.where(abuse_support_available: true).default_order
   end
 
   def abuse_report_params
     params.require(:abuse_report).permit(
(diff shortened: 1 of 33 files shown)


Cesium-Ice:
Not sure why the tests for features/other_a/abuse_report.feature break with a "missing locale_language" error, I can submit abuse reports successfully in my dev environment

Bilka2:
From discussion elsewhere (from January), so the info doesn't get lost:
redsummernight:
> > do we want to force the new LocaleLanguage records to have the same database ids as the (Work)Language they're duplicating?
>
> I think so in this case because it saves us from having to write data migrations
it's possible to basically clone a table (so structure, indices, and data) using [raw SQL](https://stackoverflow.com/questions/3280006/duplicating-a-mysql-table-indices-and-data), via an `execute` call in a migration
> when you define `belongs_to`, I think you can give it the option `foreign_key` to use `language_id` instead of `locale_language_id`, that way we don't need migrations for all the models currently using `belongs_to :language`

Bilka2:

PR #5005, opened 2024-12-27
AO3-6868 Add language form validation

connie-feng:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6868

## Purpose

This adds two form validations to the language form on the `/languages/<language abbreviation/edit` page:
* validates that the display name is not a duplicate
* validates that the abbreviation is 4 characters or fewer 

Also:
* added rake task to check for duplicate language names before running migrations
    * `rake db:check_language_name_duplicates`
* fixed some existing tests that were failing on the new validations and rubocop

## Testing Instructions

1. Login as an admin
2. Go to `/languages`
3. Click `Edit` on any of the languages
4. Change the display name to one that already exists
5. Change the abbreviation to something with more than 5 characters
6. Click `Update Language` and form validations should appear

Screenshot 2024-12-27 at 10 37 58 AM

## Credit

Connie Feng, she/her

### DIFF
--- a/app/models/language.rb
+++ b/app/models/language.rb
@@ -1,9 +1,7 @@
 class Language < ApplicationRecord
   include WorksOwner
-  validates :short, presence: true
-  validates :short, uniqueness: true, length: { maximum: 5 }
-  validates :name, uniqueness: true
-  validates :name, presence: true
+  validates :short, presence: true, uniqueness: true, length: { maximum: 5 }
+  validates :name, presence: true, uniqueness: true
 
   has_many :works
   has_many :locales
--- a/app/models/language.rb
+++ b/app/models/language.rb
@@ -1,7 +1,8 @@
 class Language < ApplicationRecord
   include WorksOwner
   validates_presence_of :short
-  validates :short, uniqueness: true
+  validates :short, uniqueness: true, length: { maximum: 5 }
+  validates :name, uniqueness: true
   validates_presence_of :name
 
   has_many :works
--- a/app/models/language.rb
+++ b/app/models/language.rb
@@ -1,9 +1,9 @@
 class Language < ApplicationRecord
   include WorksOwner
-  validates_presence_of :short
+  validates :short, presence: true
   validates :short, uniqueness: true, length: { maximum: 5 }
   validates :name, uniqueness: true
-  validates_presence_of :name
+  validates :name, presence: true
 
   has_many :works
   has_many :locales
@@ -25,6 +25,6 @@ class Language < ApplicationRecord
   end
 
   def fandom_count
-    Fandom.joins(:works).where(works: {id: self.works.posted.collect(&:id)}).distinct.select('tags.id').count
+    Fandom.joins(:works).where(works: { id: self.works.posted.collect(&:id) }).distinct.select("tags.id").count
   end
 end
--- a/app/models/language.rb
+++ b/app/models/language.rb
@@ -1,6 +1,6 @@
 class Language < ApplicationRecord
   include WorksOwner
-  validates :short, presence: true, uniqueness: true, length: { maximum: 5 }
+  validates :short, presence: true, uniqueness: true, length: { maximum: 4 }
   validates :name, presence: true, uniqueness: true
 
   has_many :works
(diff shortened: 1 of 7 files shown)


db/migrate/20241227165942_add_unique_index_to_languages.rb
```diff
@@ -0,0 +1,10 @@
+class AddUniqueIndexToLanguages < ActiveRecord::Migration[6.1]
```
brianjaustin, line 1: It looks like this migration was generated with Rails 6.1 rather than Rails 7. (Here's one from a recent PR that uses Rails 7: https://github.com/otwcode/otwarchive/pull/4999/files#diff-c8c18752ab7910e1481946e61b4e79c6ea31a75cf374308db8ced589f489cdb4R3). Would you mind updating this so it also uses Rails 7? I think that should be as simple as changing this to the following
```suggestion
class AddUniqueIndexToLanguages < ActiveRecord::Migration[7.0]
```

brianjaustin (changes requested):
Hi, Connie, thanks for the pull request!

A couple small comments for you, and noting for myself that we'll want to run a query before the migration is run to check for duplicate language names. (If you feel confident in writing a Before rake task to do that, it would be helpful but the query should be easy by hand as well so don't worry about it either way!)

I've assigned the Jira issue to you and set its status to In Review so no one will mistakenly create a duplicate pull request. I've also given your account permission to comment on, assign, and transition issues in the future.

app/models/language.rb
```diff
@@ -1,8 +1,9 @@
 class Language < ApplicationRecord
   include WorksOwner
-  validates_presence_of :short
-  validates :short, uniqueness: true
-  validates_presence_of :name
+  validates :short, presence: true
+  validates :short, uniqueness: true, length: { maximum: 5 }
+  validates :name, uniqueness: true
+  validates :name, presence: true
```
brianjaustin, line 6: I think these can be combined into 1 `validates` per column
```suggestion
  validates :short, presence: true, uniqueness: true, length: { maximum: 5 }
  validates :name, uniqueness: true, presence: true
```

Bilka2:
Please revert the update to schema.rb, we prefer not to update the schema dump in unrelated PRs (see #4677).

connie-feng:
> Please revert the update to schema.rb, we prefer not to update the schema dump in unrelated PRs (refer to #4677).

Hi @Bilka2! I've reverted the update to schema.rb, but it does cause rubocop to fail (example: https://github.com/otwcode/otwarchive/actions/runs/12519083042) -- should I disable rubocop for those specific lines?

connie-feng:
> Hi, Connie, thanks for the pull request!
> 
> A couple small comments for you, and noting for myself that we'll want to run a query before the migration is run to check for duplicate language names. (If you feel confident in writing a Before rake task to do that, it would be helpful but the query should be easy by hand as well so don't worry about it either way!)
> 

Hi @brianjaustin -- thanks! I will add the Before rake task :)

Bilka2:
> Hi @Bilka2! I've reverted the update to schema.rb, but it does cause rubocop to fail (example: https://github.com/otwcode/otwarchive/actions/runs/12519083042) -- should I disable rubocop for those specific lines?

You technically fixed the failure with the migration so it's fine to just keep it enabled. We'll ignore the failure when reviewing this PR since it's clear that it was addressed.

app/models/language.rb
```diff
@@ -1,8 +1,7 @@
 class Language < ApplicationRecord
   include WorksOwner
-  validates_presence_of :short
-  validates :short, uniqueness: true
-  validates_presence_of :name
+  validates :short, presence: true, uniqueness: true, length: { maximum: 5 }
```
brianjaustin, line 3: Sorry I missed this last time, but per https://github.com/otwcode/otwarchive/blob/master/db/schema.rb#L638 the maximum should be 4 (the ticket has 5 or more characters as something that will cause an error)

brianjaustin (changes requested):
Thanks for the rake task! Just 1 thing I missed on my first review, otherwise looks good!

brianjaustin (approved):
Thanks!

PR #5006, opened 2024-12-28
AO3-6450 Exclude hidden/unrevealed tags from bins

brianjaustin:
# Pull Request Checklist

Bonus from reviews on 4967, 4972

## Issue

https://otwarchive.atlassian.net/browse/AO3-6450

## Purpose

Exclude tags that are only used on unrevealed or admin-hidden works

**Includes a post-deploy rake task**


app/models/work.rb
```diff
@@ -310,10 +310,11 @@ def should_reindex_pseuds?
     destroyed? || (saved_changes.keys & pertinent_attributes).present?
   end
 
-  # If the work gets posted, we should (potentially) reindex the tags,
-  # so they get the correct draft-only status.
+  # If the work gets posted, hidden, or (un)revealed, we should (potentially) reindex the tags,
```
sarken, line 313: `(un)hidden`, too?

sarken (commented):
Minor thing, otherwise looks good!

sarken (approved):

Bilka2:
This is already approved, but I think it would still be nice to have a feature test that tests a specific tag on a work that is unrevealed and then re-revealed and hidden and then unhidden.

PR #5010, opened 2025-01-05
Docs: Clarify workflow for contributions and Jira

Bilka2:
## Issue

None, doc change.

## Purpose

Improve the documentation of the code contribution workflow by turning it into an ordered list. This makes the order of the steps more clear, especially in regards to Jira account permissions. Mention the Jira account permissions in the readme, together with the mailing list.

## References

Internal discussion from Dec 27 and Dec 30.

## Credit

Bilka


CONTRIBUTING.md
```diff
@@ -35,30 +35,21 @@ with the exception of spelling corrections and documentation improvements
...
-and our QA team will check that everything is working as intended. If not, we may
-set the issue to ["Broken on Test"](https://github.com/otwcode/otwarchive/wiki/Issue-Tracking-with-Jira)
-and ask you to make further changes in new pull requests.
+### Workflow
 
-If all is well, your contribution will be deployed to the [Archive of Our Own](https://archiveofourown.org)
-and you will be credited in the [release notes](https://archiveofourown.org/admin_posts?tag=1)!
+1. If you're a new contributor, find a task on the [issues reserved for first timers](https://otwarchive.atlassian.net/issues/?filter=13119). Otherwise, or if you're up for a challenge, pick a task from the general [open and unassigned issues](https://otwarchive.atlassian.net/issues/?filter=10800).
+2. Write code to address the issue.
+3. Optional: Create a Jira account if you'd like the ability to comment on, assign, and transition issues. Please make sure the Full Name on your Jira account either closely matches the name you'd like us to credit in the release notes or includes it in parentheses, e.g. "Nickname (CREDIT NAME)."
```
sarken, line 47: I like the list format! That's definitely nice and readable.

I was going to ask if we could make this Step 1a, since most people are probably thinking about creating an account when they choose an issue and want to assign it to themselves... but then I realized its position is meant to convey "if you make a Jira account, the earliest you should do it is right before you submit a PR because we won't give you permissions any sooner than Step 5." 

I think my failure to pick up on the implication might mean we should make that information more explicit. Maybe Step 1 could include a variation on the line from the README like, "(Don't worry about claiming the issue for now. If you choose to create a Jira account, we'll grant it permissions for commenting on, assigning, and transitioning issues after you create your first pull request.)"?

README.md
```diff
@@ -22,7 +22,8 @@ We welcome pull requests for bugs described in our issue tracker. Please see our
 * [Developer Documentation](https://github.com/otwcode/otwarchive/wiki)
 * [Commit Policy](https://github.com/otwcode/otwarchive/wiki/Commit-policy)
 
-We do not have a public chat, but you are welcome to contact us at otw-coders@transformativeworks.org if you have any questions.
+We do not have a public chat, but you are welcome to contact us at otw-coders@transformativeworks.org if you have any questions.  
+We grant your Jira account permissions for commenting on, assigning, and transitioning issues [after you create your first pull request](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md#workflow).
```
sarken, line 26: It looks like this is meant to be a different paragraph than the part about the chat, so I think we need a blank line between the two lines.

Bilka2: It's meant to be a linebreak but not a different paragraph, like this.  
Second sentence.

So my question is whether we want that formatting or if a new paragraph would be more standard. I'd be fine with either.

redsummernight: Let's make it a new paragraph.

redsummernight (approved):
LGTM!

PR #5012, opened 2025-01-06
AO3-5502 Fix page title of the adult content warning for chapters

slavalamp:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5502

## Purpose

Fixes the page title on the adult content warning page for chapters of a work so that it's informative (has work title, chapter number, author and fandom) and consistent with the warning page for the whole work

## Testing Instructions

See Jira ticket

## References

none

## Credit

slavalamp

### DIFF
--- a/features/other_a/page_title.feature
+++ b/features/other_a/page_title.feature
@@ -31,3 +31,13 @@ Scenario: Page title should change when tags are edited
   When I view the work "New Story"
   Then the page title should include "Harry Potter"
   And the page title should not include "Stargate"
+
+Scenario: Page title should be informative on the adult content notice page
+
+  Given I am logged in as "author"
+  And I post the 2 chapter work "New Story" with fandom "Stargate" with rating "Mature"
+  When I am logged out
+  And I view the work "New Story"
+  Then the page title should include "New Story - author - Stargate"
+  When I follow the recent chapter link for the work "New Story"
+  Then the page title should include "New Story - Chapter 2 - author - Stargate"
--- a/features/other_a/page_title.feature
+++ b/features/other_a/page_title.feature
@@ -12,32 +12,32 @@ Scenario: user reads a TOS or FAQ page
 Scenario: Page title should respect user preference
 
   Given I am logged in as "author"
-  And I go to my preferences page
-  And I fill in "Browser page title format" with "FANDOM - AUTHOR - TITLE"
-  And I press "Update"
-  And I post the work "New Story" with fandom "Stargate"
+    And I go to my preferences page
+    And I fill in "Browser page title format" with "FANDOM - AUTHOR - TITLE"
+    And I press "Update"
+    And I post the work "New Story" with fandom "Stargate"
   When I view the work "New Story"
   Then the page title should include "Stargate - author - New Story"
 
 Scenario: Page title should change when tags are edited
 
   Given I am logged in as "author"
-  And I post the work "New Story" with fandom "Stargate"
+    And I post the work "New Story" with fandom "Stargate"
   When I view the work "New Story"
   Then the page title should include "Stargate"
   When I edit the work "New Story"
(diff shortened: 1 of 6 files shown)


Bilka2:
Thank you, the code looks good! Could you add a test in `features/other_a/page_title.feature` that checks for the browser page title? 

> (I think it could be satisfied by turning the negated if condition into an unless. Also it could return after access_denied and else wouldn't be needed and that whole block after it would have one less indentation level...)

What about going even further and turning it into this?

`access_denied and return unless @chapters.include?(@chapter)`

Bilka2:
> This doesn't work, I think it's because access_denied always returns false.

You're right! There's one spot in the code that does this anyway, which mislead me.

> I wasn't sure if it was okay for me to touch those lines because they aren't directly related to the issue, but this sounds like an approval, so I commited my idea, let me know if I need to undo/change that then

Yep, you're good! We try to keep changes to unrelated lines low because it makes reviewing harder, merge conflicts are more likely and so on. That goes especially for rubocop because fixing all the violations can become a bit of an endless endavour. But that's just a guideline, so it's okay to have exceptions sometimes, like here.

I'd stop at this state in regards to rubocop and just tolerate the remaining errors instead of going down the rest of the rabbit hole.

Regarding the commits, I think this may be a result of merge commit that was then rebased. There's likely some git magic to undo this, but the current state should be fine too. For the future, we prefer merge commits (or the occasional commit with a missing issue number) over rebasing/force pushing. Because we squash on merge those messups become pretty much invisible in the final git history anyway.

features/other_a/page_title.feature
```diff
@@ -31,3 +31,13 @@ Scenario: Page title should change when tags are edited
...
+
+Scenario: Page title should be informative on the adult content notice page
+
+  Given I am logged in as "author"
+  And I post the 2 chapter work "New Story" with fandom "Stargate" with rating "Mature"
+  When I am logged out
+  And I view the work "New Story"
+  Then the page title should include "New Story - author - Stargate"
+  When I follow the recent chapter link for the work "New Story"
+  Then the page title should include "New Story - Chapter 2 - author - Stargate"
```
Bilka2, line 43: It's nice to have the final newline in files, like GitHub points out.

features/other_a/page_title.feature
```diff
@@ -31,3 +31,13 @@ Scenario: Page title should change when tags are edited
   When I view the work "New Story"
   Then the page title should include "Harry Potter"
   And the page title should not include "Stargate"
+
+Scenario: Page title should be informative on the adult content notice page
+
+  Given I am logged in as "author"
+  And I post the 2 chapter work "New Story" with fandom "Stargate" with rating "Mature"
```
Bilka2, line 38: Usually we indent `And` and `But` by one step. Not sure why the rest of the file doesn't do this, it might be older than that style guideline.

features/other_a/page_title.feature
```diff
@@ -31,3 +31,13 @@ Scenario: Page title should change when tags are edited
...
+  When I am logged out
+  And I view the work "New Story"
+  Then the page title should include "New Story - author - Stargate"
```
Bilka2, line 41: Here it would be nice to make sure that we're actually on the "adult content notice" page, because it's only implied in the testing steps. Currently, with just reading the test and knowing that it passed, you couldn't be sure that the intended case is getting tested. So you could add a `I should see "This work could have adult content"` before or after this step and the other page title step to make it clearer

However, I locally checked that the test indeed ends up on the adult content notice page, so this is fine as is. So these comments are optional to address, like I said.

Bilka2 (approved):
I'm approving this because the test and code works as-is.

However, I left some extra comments on the test since you said you're new to the tests and might want to learn. You don't have to address these if you'd rather not, they are nitpicks and not important in the grand scheme of things!

PR #5015, opened 2025-01-07
AO3-5578 ActiveStorage copy script performance fixes

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5578

## Purpose

Update the icon copy tasks to try to be nicer to the prod database (in particular the pseuds table).

### DIFF
--- a/lib/tasks/after_tasks.rake
+++ b/lib/tasks/after_tasks.rake
@@ -348,7 +348,7 @@ namespace :After do
         end
       end
 
-      puts "Finished batch #{index + 1}"
+      puts "Finished batch #{index + 1}" && $stdout.flush
       sleep 10
     end
   end
@@ -381,7 +381,7 @@ namespace :After do
         end
       end
 
-      puts "Finished batch #{index + 1}"
+      puts "Finished batch #{index + 1}" && $stdout.flush
       sleep 10
     end
   end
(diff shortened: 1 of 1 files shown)


lib/tasks/after_tasks.rake
```diff
@@ -329,22 +329,27 @@ namespace :After do
...
+          collection.icon.attach(io: URI.parse(icon_url).open,
+                                 filename: image_original,
+                                 content_type: collection.icon_content_type)
+        rescue StandardError => e
+          puts "Error '#{e}' copying #{icon_url}"
+        end
       end
 
-      puts "Finished up to ID #{collection.id}" if collection.id.modulo(100).zero?
+      puts "Finished batch #{index + 1}"
```
Bilka2, line 351: I think because you sleep right after, it would good to have a `$stdout.flush` so that this definitely shows up in the output immediately.

And if you want to be fancy you can do the math for the total batch number that the other `find_in_batches` tasks do, but it's definitely not necessary.

brianjaustin: I would not want to do the count for Pseud, it's large enough that I want to touch the table as little as possible (even things that are relatively safe)

Bilka2 (approved):
:crossed_fingers:

redsummernight (approved):

PR #4885, opened 2024-07-31
AO3-6759 Check admin roles for actions in TagWranglersController

brianjaustin:
# Pull Request Checklist

* [X] Have you read ["How to write the perfect pull request"](https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/)?
* [X] Have you read the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [X] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)?
* [X] Have you added the [Jira](https://otwarchive.atlassian.net) issue number
  as the *first* thing in your pull request title (e.g. `AO3-1234 Fix thing`)
* [X] Do you have fewer than 5 pull requests already open? If not, please wait
  until they are reviewed and merged before creating new pull requests.

## Issue

https://otwarchive.atlassian.net/browse/AO3-6759

## Purpose

Update the `TagWranglersController` to verify the current admin has the right roles to access each action

## Credit

Brian Austin (they/he)

PR #5022, opened 2025-01-09
AO3-6204 Allow certain admins to remove non-default orphan pseuds from orphaned works

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6204

## Purpose

Adds a button and a confirmation page to allow some admins to remove non-default orphan_account pseuds from orphaned works.

## References

PR Limit: 8 out of 5
Using reviewer bonus:
#4378
#4885

## Credit

Bilka


app/controllers/admin/user_creations_controller.rb
```diff
@@ -59,4 +59,38 @@ def destroy
...
+      pseuds = @work.orphan_pseuds
+      if pseuds.length > 1
+        flash[:error] = t(".must_select_pseud")
+        redirect_to work_path(@work) and return
+      end
+    else
+      pseuds = Pseud.find(pseuds).select { |p| p.user_id == orphan_account.id }
+    end
+
+    orphan_pseud = orphan_account.default_pseud
```
redsummernight, line 88: You can inline this variable since it's only used once.

spec/controllers/admin/user_creations_controller_spec.rb
```diff
@@ -164,4 +164,151 @@
       it_behaves_like "unauthorized admin cannot delete bookmarks"
     end
   end
+
+  shared_examples "an action only authorized admins can access" do |authorized_roles:|
```
redsummernight, line 168: Looks like this block is defined in a bunch of places. Can you move it into `spec/support/shared_examples/admin_shared_examples.rb`?

spec/controllers/admin/user_creations_controller_spec.rb
```diff
@@ -164,4 +164,151 @@
...
+
+      context "with a orphan_account pseuds parameter" do
+        subject { put :remove_pseud, params: { id: work.id, pseuds: [orphan_pseud.id] } }
+
+        it "redirects removes only that pseud" do
+          subject
+          it_redirects_to_with_notice(work_path(work), "Successfully removed pseud Leaver (orphan_account) from this work.")
+          expect(work.reload.pseuds).to include(orphan_account.default_pseud)
+          expect(work.pseuds).not_to include(orphan_pseud)
+          expect(work.reload.pseuds).to include(orphaneer_orphan_pseud)
```
redsummernight, line 297: Is a second reload necessary?

app/controllers/admin/user_creations_controller.rb
```diff
@@ -59,4 +59,38 @@ def destroy
...
+    pseuds.each do |pseud|
+      pseud.change_ownership(@work, orphan_pseud)
+    end
+    AdminActivity.log_action(current_admin, @work, action: "remove orphan_account pseuds")
```
redsummernight, line 92: If `pseuds` is empty we can skip the admin log and the notice.

app/controllers/admin/user_creations_controller.rb
Bilka2: It's used multiple times in the loop if multiple pseuds are removed. I want avoid the DB lookup/cache hit in the loop.

redsummernight: Nice!

redsummernight (approved):

PR #5028, opened 2025-01-15
AO3-6878 Disable ActiveStorage analyzers

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6878

## Purpose

Keep the database happier by turning off this very noisy job (that we probably won't use anyway)

### DIFF
--- a/config/initializers/disable_active_storage_analyzer.rb
+++ b/config/initializers/disable_active_storage_analyzer.rb
@@ -1,5 +1,0 @@
-# We don't want to fill up the queue with jobs that hammer the database
-# (and extract metadata we don't use). Ref https://stackoverflow.com/a/74879360.
-Rails.application.config.to_prepare do
-  ActiveStorage::Attachment.skip_callback(:commit, :after, :analyze_blob_later)
-end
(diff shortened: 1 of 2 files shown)


config/initializers/disable_active_storage_analyzer.rb
```diff
@@ -0,0 +1,5 @@
+# We don't want to fill up the queue with jobs that hammer the database
+# (and extract metadata we don't use). Ref https://stackoverflow.com/a/74879360.
+Rails.application.config.to_prepare do
+  ActiveStorage::Attachment.skip_callback(:commit, :after, :analyze_blob_later)
```
redsummernight, line 4: Should we set empty analyzers/previewers in the Rails config instead? As of https://github.com/rails/rails/pull/37773 that should no longer create no-op queued jobs.

brianjaustin: Nice! I was mistaking the `PurgeJob` for `AnalyzeJob` when I thought that didn't work still. Much better!

redsummernight (approved):

PR #5029, opened 2025-01-16
AO3-6875 Add additional wrangling status filtering options to tag search

metallicity:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6875

## Purpose

This PR expands the existing options for filtering based on wrangling status on the Tag Search page. Previously, users could only filter on canonical tags, non-canonical tags, or both. Now the options also allow distinguishing between tags that are synonyms of canonical tags, in every combination of these wrangling statuses.

(Notably, this includes a "Canonical or synonymous" option that was omitted from the original ticket description. It seemed like an obviously useful addition, and required no additional tracking within the tag index data model to support, but it would be easy enough to remove it from the page if there is some concern over showing too many options to the user.)

## Testing Instructions

1. Identify or create a set of **canonical**, **synonymous**, and **non-canonical and non-synonymous** tags.
2. Search for each tag by name (or all tags at once if they share one or more distinct words in their names)
3. Any **canonical** tag should appear only when selecting the following wrangling status options: "Canonical", "Canonical or synonymous", "Any status"
4. Any **synonymous** tag should appear only when selecting the following wrangling status options: "Non-canonical", "Synonymous", "Canonical or synonymous", "Any status"
5. Any **non-canonical and non-synonymous** tag should appear only when selecting the following wrangling status options: "Non-canonical", "Non-canonical and non-synonymous", "Any status"

## References

#4144 - Original PR introducing the "wrangling status" section of the Tag Search page

## Credit

Metallicity (they/them)

### DIFF
--- a/app/models/search/tag_indexer.rb
+++ b/app/models/search/tag_indexer.rb
@@ -70,7 +70,7 @@ class TagIndexer < Indexer
       tag_type: object.type,
       uses: object.taggings_count_cache,
       unwrangled: object.unwrangled?,
-      filterable: object.synonymous? || object.canonical
+      canonical_or_synonymous: object.synonymous? || object.canonical
     ).merge(parent_data(object))
   end
 
--- a/app/models/search/tag_indexer.rb
+++ b/app/models/search/tag_indexer.rb
@@ -69,7 +69,8 @@ class TagIndexer < Indexer
       has_posted_works: object.has_posted_works?,
       tag_type: object.type,
       uses: object.taggings_count_cache,
-      unwrangled: object.unwrangled?
+      unwrangled: object.unwrangled?,
+      filterable: object.synonymous? || object.canonical
     ).merge(parent_data(object))
   end
 
--- a/app/models/search/tag_indexer.rb
+++ b/app/models/search/tag_indexer.rb
@@ -69,8 +69,7 @@ class TagIndexer < Indexer
       has_posted_works: object.has_posted_works?,
       tag_type: object.type,
       uses: object.taggings_count_cache,
-      unwrangled: object.unwrangled?,
-      canonical_or_synonymous: object.synonymous? || object.canonical
+      unwrangled: object.unwrangled?
     ).merge(parent_data(object))
   end
 
--- a/app/models/search/tag_query.rb
+++ b/app/models/search/tag_query.rb
@@ -31,7 +31,6 @@ class TagQuery < Query
   def exclusion_filters
     [
       wrangled_filter
-      # wrangling_status_exclude_filter
     ].compact
   end
 
@@ -85,11 +84,11 @@ class TagQuery < Query
     when "noncanonical"
       term_filter(:canonical, false)
     when "synonymous"
-      [{ exists: { field: "merger_id" } }, term_filter(:canonical, false)]
+      [exists_filter("merger_id"), term_filter(:canonical, false)]
     when "canonical_synonymous"
-      { bool: { should: [{ exists: { field: "merger_id" } }, term_filter(:canonical, true)] } }
+      { bool: { should: [exists_filter("merger_id"), term_filter(:canonical, true)] } }
     when "noncanonical_nonsynonymous"
-      [{ bool: { must_not: { exists: { field: "merger_id" } } } }, term_filter(:canonical, false)]
+      [{ bool: { must_not: exists_filter("merger_id") } }, term_filter(:canonical, false)]
     end
   end
 
@@ -142,7 +141,7 @@ class TagQuery < Query
   # the fandom exists, because this particular filter is included in the
   # exclusion_filters section.
   def wrangled_filter
-    { exists: { field: "fandom_ids" } } unless options[:wrangled].nil?
+    exists_filter("fandom_ids") unless options[:wrangled].nil?
   end
 
   ################
--- a/app/models/search/tag_query.rb
+++ b/app/models/search/tag_query.rb
@@ -16,6 +16,7 @@ class TagQuery < Query
     [
       type_filter,
       canonical_filter,
+      filterable_filter,
       unwrangleable_filter,
       posted_works_filter,
       media_filter,
@@ -79,6 +80,10 @@ class TagQuery < Query
 
   def canonical_filter
     term_filter(:canonical, bool_value(options[:canonical])) if options[:canonical].present?
+  end
+
+  def filterable_filter
+    term_filter(:filterable, bool_value(options[:filterable])) if options[:filterable].present?
   end
 
   def unwrangleable_filter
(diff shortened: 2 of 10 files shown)


sarken:
Hi, Metallicity!

Thank you so much for this pull request. I'm checking in with the Tag Wrangling team about whether they'd like the "Canonical or synonymous" option added, and someone will be along to review the code soon.

In the meantime, I've updated the Jira issue status to In Review, so no one will mistakenly create a duplicate pull request. If you'd like the ability to comment on, assign, and transition issues in the future, you're welcome to create a Jira account! You can just reply here with the account name and we'll set up the permissions for you. (It makes things a bit easier for us on the organizational side if the Full Name on your Jira account either closely matches the name you'd like us to credit in the release notes or includes it in parentheses, e.g. "Nickname (CREDIT NAME).")

metallicity:
@sarken I just joined the otwarchive Jira with an account named "Metallicity".

sarken:
@metallicity Great! I've given your account the permissions I mentioned and updated the issue to assign it to you and add your testing instructions. The Tag Wrangling team has also approved the inclusion of the "Canonical or synonymous" option, so this is ready for review!

I did notice some Reviewdog comments about using the old `ts()` method when I glanced at the code while updating the Jira issue. If you'd like some info on how to deal with those comments, you can check out [our Internationalization (i18n) Standards wiki page](https://github.com/otwcode/otwarchive/wiki/Internationalization-%28i18n%29-Standards).

metallicity:
I moved all the "wrangling status" option labels into `en.yml`, as well as the more low hanging fruit from the rest of the page. But there are several non-obvious sections on the Tag Search page, particularly the tag types radio button set and the sort by/sort direction drop downs, which I'm not sure have an established i18n convention within the codebase. Regardless, all of those sections are unrelated to the PR itself.

app/models/search/tag_indexer.rb
```diff
@@ -69,7 +69,8 @@ def document(object)
       has_posted_works: object.has_posted_works?,
       tag_type: object.type,
       uses: object.taggings_count_cache,
-      unwrangled: object.unwrangled?
+      unwrangled: object.unwrangled?,
+      filterable: object.synonymous? || object.canonical
```
redsummernight, line 73: We shouldn't call this field `filterable`, because there exists a definition of `unfilterable` that takes into account other factors:

https://github.com/otwcode/otwarchive/blob/775ed1341c628f91fb345e12b12e5f62e08810b8/app/models/tag.rb#L632-L634

Let's simply call it `canonical_or_synonymous`.

app/models/search/tag_search_form.rb
```diff
@@ -36,6 +38,22 @@ def search_results
     @searcher.search_results
   end
 
+  def set_canonical
```
redsummernight, line 41: Splitting the handling of `wrangling_status` into 2 methods makes it harder to follow what option is set in which case. Can you combine them into `set_wrangling_status` and use a case statement for `wrangling_status`?

metallicity:
Addressed both comments. Nothing that I wasn't half expecting, I had already gone back and forth on which way `set_wrangling_status` would be more clear.

Also re-requested you as a review, @redsummernight, but feel free to take yourself off if it's better to just leave it with a general "Awaiting Review" tag or something.

redsummernight:
Because we're adding a new field to tag index, we'll need to reindex all tags on production when deploying this change by running:

```rb
TagIndexer.index_from_db
```

But I think it may be possible to do wrangling status filtering not by adding a new field, but by checking the existing field `merger_id`. I'll get back to you on that.

metallicity:
@redsummernight Intuitively, using just the value of `canonical` and whether `merger_id` is present seems like it should work fine. It does make the interaction with the original `canonical_filter` a bit more awkward though, so it struck me as simpler to instead combine things into a single `wrangling_status_filter`, and then move the `wrangling_status` case block directly into `tag_query.rb`.

metallicity:
It should be safe to remove the [Needs Reindex](https://github.com/otwcode/otwarchive/labels/Needs%20Reindex) label on this PR now, since I reverted all the changes that using `merger_id` made unnecessary.

app/models/search/tag_query.rb
```diff
@@ -31,6 +31,7 @@ def filters
   def exclusion_filters
     [
       wrangled_filter
+      # wrangling_status_exclude_filter
```
redsummernight, line 34: This can be removed.

app/models/search/tag_search_form.rb
```diff
@@ -7,14 +7,14 @@ class TagSearchForm
   ATTRIBUTES = [
     :query,
     :name,
-    :canonical,
```
redsummernight, line 10: We should make sure if people keep URLs of current tag searches, those still work. Let's allow `canonical` here, and in `initialize` call a new method that converts `options[:canonical]` into the corresponding `options[:wrangling_status]`.

app/models/search/tag_query.rb
```diff
@@ -77,8 +78,19 @@ def type_filter
...
-  def canonical_filter
-    term_filter(:canonical, bool_value(options[:canonical])) if options[:canonical].present?
+  def wrangling_status_filter
+    case options[:wrangling_status]
+    when "canonical"
+      term_filter(:canonical, true)
+    when "noncanonical"
+      term_filter(:canonical, false)
+    when "synonymous"
+      [{ exists: { field: "merger_id" } }, term_filter(:canonical, false)]
```
redsummernight, line 88: Can you add a helper method `exists_filter(field, value)` in the Query class, similar to `term_filter`?

You can use the new helper here as well:

https://github.com/otwcode/otwarchive/blob/25865ac2f122e8027994be6809a54bacdfdf9d6e/app/models/search/tag_query.rb#L133

redsummernight (commented):
The approach looks good! Just some minor comments.

app/models/search/tag_search_form.rb
```diff
@@ -43,6 +45,12 @@ def set_fandoms
     @options[:fandom_ids] = Tag.where(name: names).pluck(:id)
   end
 
+  def set_wrangling_status
+    return unless @options[:canonical]
+    
+    @options[:wrangling_status] = "canonical"
```
redsummernight, line 51: https://github.com/otwcode/otwarchive/blob/5748b65c170c976e33f7a71b5903da5c9797bdc9/app/views/tags/_search_form.html.erb#L42-L49

`@options[:wrangling_status] = "canonical"` only if `@options[:canonical]` is `T`. You should handle the `F` to `"noncanonical"` case as well.

redsummernight (approved):
Looks good, thank you!

brianjaustin:
@metallicity the wranglers say thank you for this improvement!

sea-otw:
@metallicity hi, just popping by from wranglerland to confirm that we wranglers are indeed overjoyed with the improvement! thank you for all the effort you've put into it <3

PR #5038, opened 2025-01-28
AO3-6890 Add puma worker killer gem

brianjaustin:
https://otwarchive.atlassian.net/browse/AO3-6890

### DIFF
--- a/config/config.yml
+++ b/config/config.yml
@@ -747,6 +747,7 @@ PERCONA_ARGS: >
 BOOKMARKABLE_SHARDS: 5
 WORKS_SHARDS: 5
 
-# Puma killer config (automatically restart Puma when resource constrained)
+# Puma killer config (automatically restart Puma when resource constrained);
+# both are measured in seconds.
 PUMA_ROLLING_RESTART_FREQUENCY: 172800 # 2 days
 PUMA_ROLLING_RESTART_SPLAY: 702
--- a/config/config.yml
+++ b/config/config.yml
@@ -746,3 +746,7 @@ PERCONA_ARGS: >
 # Production has more than 5 for these
 BOOKMARKABLE_SHARDS: 5
 WORKS_SHARDS: 5
+
+# Puma killer config (automatically restart Puma when resource constrained)
+PUMA_ROLLING_RESTART_FREQUENCY: 172800 # 2 days
+PUMA_ROLLING_RESTART_SPLAY: 702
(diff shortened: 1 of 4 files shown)


Bilka2:
Based on the docs I think you might have to call `PumaWorkerKiller.enable_rolling_restart` somewhere. https://github.com/zombocom/puma_worker_killer/issues/4 / https://github.com/zombocom/puma_worker_killer?tab=readme-ov-file#attention also suggest not putting it in an initializer for rails (and putting into before_fork), not sure if that still applies these days (maybe you know better how Puma is started in staging/production)

config/initializers/puma_worker_killer.rb
```diff
@@ -0,0 +1,3 @@
+PumaWorkerKiller.config do |config|
```
redsummernight, line 1: Do we need a `before_fork` hook, or a call to `start`? Would the gem be activated this way?

config/config.yml
```diff
@@ -746,3 +746,7 @@ PERCONA_ARGS: >
 # Production has more than 5 for these
 BOOKMARKABLE_SHARDS: 5
 WORKS_SHARDS: 5
+
+# Puma killer config (automatically restart Puma when resource constrained)
+PUMA_ROLLING_RESTART_FREQUENCY: 172800 # 2 days
```
redsummernight, line 751: Can you comment that these configs are in seconds?

config/initializers/puma_worker_killer.rb
brianjaustin: Yeah I'm pretty sure Bilka is right. Based on the docs (which I am having a little trouble following since there seem to be multiple paths to configuring it), we need to do this in Ansible instead. The linked PR (visible only to folks with read+ access to that repo) needs to do it instead, since we don't have `config/puma.rb` here

zz9pzza: I was expecting to do it in the config/puma.rb

config/config.yml
```diff
@@ -746,3 +746,8 @@ PERCONA_ARGS: >
 # Production has more than 5 for these
 BOOKMARKABLE_SHARDS: 5
 WORKS_SHARDS: 5
+
+# Puma killer config (automatically restart Puma when resource constrained);
+# both are measured in seconds.
+PUMA_ROLLING_RESTART_FREQUENCY: 172800 # 2 days
```
zz9pzza, line 752: In ansible, I am expecting this to be a time with an offset based on say the ip address so we are unlikely to get a thundering herd.

redsummernight (approved):

PR #5039, opened 2025-01-29
AO3-6893 Upgrade gems and configs to Rails 7.1.x

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6893

## Purpose

Upgrade to Rails 7.1. Should go in after Ruby 3.2 and AO3-6892.  

## Testing Instructions

Refer to Jira.

## Credit

Bilka


config/environments/production.rb
```diff
@@ -40,29 +39,43 @@
...
   # config.action_cable.mount_path = nil
   # config.action_cable.url = "wss://example.com/cable"
   # config.action_cable.allowed_request_origins = [ "http://example.com", /http:\/\/example.*/ ]
 
+  # Assume all access to the app is happening through a SSL-terminating reverse proxy.
+  # Can be used together with config.force_ssl for Strict-Transport-Security and secure cookies.
+  # config.assume_ssl = true
+
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
   # config.force_ssl = true
```
Bilka2, line 54: Do we want to set these to true? The upgrade script wants at least force_ssl to be true.

app/models/search/bookmark_indexer.rb
```diff
@@ -84,7 +84,7 @@ def document(object)
     json_object = object.as_json(
       root: false,
       only: [
-        :id, :created_at, :bookmarkable_type, :bookmarkable_id, :user_id,
+        :id, :created_at, :bookmarkable_type, :bookmarkable_id,
```
Bilka2, line 87: This `user_id` was duplicate with the one in the merge call later (as_json uses string keys, the merge uses symbols, so it wasn't getting overwritten by the merge). Something changed in the json serialization for Elasticsearch so it wasn't getting rid of this duplicate `"user_id": nil` anymore, so deleting it here is the fix.

config/application.rb
```diff
@@ -141,5 +142,8 @@ class Application < Rails::Application
     config.active_storage.queues.preview_image = :active_storage
     config.active_storage.queues.purge = :active_storage
     config.active_storage.queues.transform = :active_storage
+
+    # Use secret from archive config
+    config.secret_key_base = ArchiveConfig.SESSION_SECRET
```
Bilka2, line 147: According to the Rails documentation, it's preferred to do this in `credentials.yml.enc` instead: https://guides.rubyonrails.org/v7.1/configuring.html#config-secret-key-base / https://api.rubyonrails.org/v7.1.5.1/classes/Rails/Application.html#method-i-secret_key_base

config/application.rb
```diff
@@ -99,7 +103,6 @@ class Application < Rails::Application
       "X-Frame-Options" => "SAMEORIGIN",
       "X-XSS-Protection" => "1; mode=block",
       "X-Content-Type-Options" => "nosniff",
-      "X-Download-Options" => "noopen",
```
Bilka2, line 102: This is suggested by the update script. It worsens Internet Explorer support.

config/environments/production.rb
brianjaustin: I think not, since we support http://insecure.archiveofourown.org and friends and do not plan to drop that support

brianjaustin (commented):
Still some tests failing so not my final review, but here are my initial thoughts. (And if you have time, it might be nice to convert the failing tests to RSpec instead of minitest, but that's out of scope so not a big deal either way)

Bilka2:
The tests seem to be flaky, they didn't fail on the commit with the active storage changes, just the one with the robocop changes. I also had them fail on `master` on my fork, and then pass after a re-run: https://github.com/Bilka2/otwarchive/actions/runs/13028412099/attempts/1 I won't have time to investigate why they fail before Sunday, if someone else wants to have a look.

app/controllers/blocked/users_controller.rb
```diff
@@ -13,7 +13,10 @@ class UsersController < ApplicationController
     def index
       @blocks = @user.blocks_as_blocker
         .joins(:blocked)
-        .includes(blocked: [:pseuds, { default_pseud: { icon_attachment: :blob } }])
+        .includes(blocked: [:pseuds, { default_pseud: { icon_attachment: { blob: {
+                    variant_records: { image_attachment: :blob },
+                    preview_image_attachment: { blob: { variant_records: { image_attachment: :blob } } }
+                  } } } }])
```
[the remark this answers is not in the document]
redsummernight: It's 7.1 so we should be able to re-enable these lines:

https://github.com/otwcode/otwarchive/blob/603cb16ce5590f96de1dc60e0b707936ffa9eeda/spec/requests/blocked_users_n_plus_one_spec.rb#L15-L17

and

https://github.com/otwcode/otwarchive/blob/603cb16ce5590f96de1dc60e0b707936ffa9eeda/spec/requests/muted_users_n_plus_one_spec.rb#L15-L17

However, with these lines restored, I didn't get any test failures on the current main branch / Rails 7.0. Not sure why yet.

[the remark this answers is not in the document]
Bilka2: Maybe something related to these comments got lost in the whole active storage PR, it was quite large and there were changes to the related scope after these comments were added. I'll just uncomment the tests and we can be happy that they pass now, I guess?

db/migrate/20250129101815_add_service_name_to_active_storage_blobs.active_storage.rb
```diff
@@ -0,0 +1,22 @@
+# This migration comes from active_storage (originally 20190112182829)
+class AddServiceNameToActiveStorageBlobs < ActiveRecord::Migration[6.0]
+  def up
+    return unless table_exists?(:active_storage_blobs)
```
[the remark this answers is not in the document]
Bilka2: Our current DB schema has this:
```rb
add_foreign_key "active_storage_attachments", "active_storage_blobs", column: "blob_id"
add_foreign_key "active_storage_variant_records", "active_storage_blobs", column: "blob_id"
```
Meaning that we have foreign keys pointing to `active_storage_blobs` (which is getting modified here), so if we use pt-online-schema-change / departure here, we run into [issues with foreign keys](https://docs.percona.com/percona-toolkit/pt-online-schema-change.html#cmdoption-pt-online-schema-change-alter-foreign-keys-method):
> When the tool renames the original table to let the new one take its place, the foreign keys “follow” the renamed table, and must be changed to reference the new table instead.

Departure currently sets `--alter-foreign-keys-method auto`.

app/controllers/blocked/users_controller.rb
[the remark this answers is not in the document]
Bilka2: Good? news:

The test changes behaviour based on whether we show the images via the proxy url or with `processed.url`, it fails only when using `processed.url`. This matches [the documentation](https://guides.rubyonrails.org/v7.1/active_storage_overview.html#lazy-vs-immediate-loading). So AO3-6898 temporarily broke it, then AO3-6900 fixed it again.

I reverted AO3-6900 locally to try to make the test work, but sadly it doesn't seem to want to behave, no matter what includes I use. According to bullet, the `includes` are effective when viewing the blocked user index in a browser, but the test still fails.

I recall that we also had problems with this in the initial Active Storage PR. So I'll go ahead and disable the test again and note that using `processed.url` may result in N+1 queries. Hopefully we can just stick with the proxy url.

spec/requests/blocked_users_n_plus_one_spec.rb
```diff
@@ -12,8 +12,7 @@
       populate do |n|
         blocked_users = create_list(:user, n)
         blocked_users.each do |blocked|
-          # Rails doesn't seem to want to include variants, so this won't work right now.
-          # We can revisit when https://github.com/rails/rails/pull/49231 is released OR we upgrade to Rails 7.1
+          # Rails doesn't seem to want to include all variants for default_pseud, so this won't work right now.
```
redsummernight, line 15: Can you add a TODO to start this comment? Same with the muted spec.

config/config.yml
```diff
@@ -739,7 +739,7 @@ PERCONA_ARGS: >
   --max-load Threads_running=15
   --critical-load Threads_running=100
   --set-vars innodb_lock_wait_timeout=2
-  --alter-foreign-keys-method=auto
+  --alter-foreign-keys-method=drop_swap
```
redsummernight, line 742: I'd like @zz9pzza or @brianjaustin to confirm that this option is fine.

redsummernight (approved):

config/application.rb
```diff
@@ -48,8 +52,7 @@ class Application < Rails::Application
     # Run "rake -D time" for a list of tasks for finding time zone names. Default is UTC.
     config.time_zone = "UTC"
 
-    # The default locale is :en and all translations from config/locales/**/*.rb,yml are auto loaded.
-    config.i18n.load_path += Dir[Rails.root.join("config/locales/**/*.{rb,yml}")]
+    # The default locale is :en.
```
Bilka2, line 55: The load_path configuration defaults to this since Rails 7.0: https://guides.rubyonrails.org/v7.1/configuring.html#config-i18n-load-path

config/config.yml
brianjaustin: From the pt-online-schema-change manpage:
```
           drop_swap
               Disable foreign key checks (FOREIGN_KEY_CHECKS=0), then drop the original table before renaming the new
               table into its place. This is different from the normal method of swapping the old and new table, which
               uses an atomic "RENAME" that is undetectable to client applications.

               This method is faster and does not block, but it is riskier for two reasons.  First, for a short time
               between dropping the original table and renaming the temporary table, the table to be altered simply
               does not exist, and queries against it will result in an error.  Secondly, if there is an error and the
               new table cannot be renamed into the place of the old one, then it is too late to abort, because the old
               table is gone permanently.

               This method forces "--no-swap-tables" and "--no-drop-old-table".
```

The last line seems a bit contradictory to the initial paragraph, but we'll want to be aware that this might leave us with tables to clean up (although we haven't always done that in the past so...) Also noting that this setting is overridden by Ansible, so we'll need to change that as well.

Overall, I think this is OK. If we're particularly worried about a table going wonky, we have backups and can also disable replication for a bit to have a "good" copy to fall back to.

redsummernight: We ended up using `rebuild_constraints` instead (#5066).

PR #5041, opened 2025-01-30
AO3-6895 Reduce the number of badly closed database connections

zz9pzza:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6895

## Purpose

What does this PR do?

## Testing Instructions

This will need a smoke test on staging if test pass.

I think we will only know if it helps when deployed to production, I a can test with tcpdump to see if the connections are being dropped.

## References

## Credit

This is inspired by Ben Sheldon and special thanks needs to be given to him in the release notes


app/models/concerns/async_with_resque.rb
```diff
@@ -22,14 +22,16 @@ def async(method, *args)
...
+        if method.is_a?(Integer)
+          # TODO: For backwards compatibility, if the "method" is an integer, we
+          # treat it like an ID and use perform_on_instance instead. But once all
+          # of the jobs in the queue have been processed (or deleted), we should
+          # be able to remove this check.
+          perform_on_instance(method, *args)
+        else
+          send(method, *args)
+        end
       end
```
brianjaustin, line 35: Based on this TODO, it sounds like we can delete the `if ` and just do the following?
```suggestion
      Rails.application.executor.wrap { send(method, *args) }
```

brianjaustin (changes requested):
Could the `extend` happen at initialization instead, so we don't forget to do it for future jobs? It seems not-very-Rails to do it by hand everywhere. Please take a look at the Rubocop comments as well, at least 1 is valid. I think some of the others may be silly, in which case we should suppress via comment rather than just ignore them.

(Linking to https://resque.github.io/docs/HOOKS.html as well because I had a hard time figuring out where that was in the docs)

PR #5045, opened 2025-02-02
AO3-6892 Update schema and structure dumps

brianjaustin:
https://otwarchive.atlassian.net/browse/AO3-6892

Preparation for the Rails 7.1 upgrade, no testing needed.


db/schema.rb
```diff
@@ -10,14 +10,125 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[6.1].define(version: 2023_09_20_094945) do
+ActiveRecord::Schema[7.0].define(version: 2024_04_04_183910) do
+  create_table "_comments_old", id: :integer, charset: "utf8mb4", collation: "utf8mb4_unicode_ci", options: "ENGINE=InnoDB ROW_FORMAT=DYNAMIC", force: :cascade do |t|
```
redsummernight, line 14: Can you remove the `_old` tables? These are left over from old pt-osc migrations, no reason to track them here.

Same for the other file.

brianjaustin: Done, but these really need to be cleaned up

redsummernight (approved):

PR #5053, opened 2025-02-06
AO3-6884 Update to Ruby 3.2

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6884

## Purpose

Update the running Ruby version to 3.2(.7) with room to update the patch version later if needed.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -403,7 +403,7 @@ GEM
       racc (~> 1.4)
     orm_adapter (0.5.0)
     parallel (1.25.1)
-    parser (3.3.0.5)
+    parser (3.3.7.1)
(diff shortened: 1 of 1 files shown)


Bilka2 (approved):
All good, thank you!

In the future we should update the parser gem once it supports Ruby 3.2.7 so that it won't complain about the ruby version, but that's a minor thing.

Bilka2:
Oops, I was on the wrong branch for the changelog, the parser gem does already support Ruby 3.2.7. Please update it to [v3.3.7.1](https://github.com/whitequark/parser/blob/master/CHANGELOG.md#v3371-2025-02-05).

PR #5054, opened 2025-02-06
AO3-6880 Use pagy gem for pagination on select pages

ceithir:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6880

## Purpose

As per ticket.

Visible (frontend) differences from will_paginate:
- `rel="prev"` and `rel="next"` have disappeared but seem to be obsolete anyway (cf https://developers.google.com/search/docs/specialty/ecommerce/pagination-and-incremental-page-loading)
- Weird `title` attributes (all equal to an untranslated "pagination" string) have not made the cut
- Blank space characters between `<li>` tags have been removed. In other words, the HTML was previously `</li> <li>` and is now `</li><li>`. This means the pagination is slightly more compact looking. Maybe need to add back some padding in CSS?
- will_paginate was always showing the first two and the last two items in the list (`1 2 ... <> ... 312 313`) while pagy only shows the very first and the last (`1 ... <> ... 313`). There doesn't seem to be an option to modify that behavior without going into the code (cf [Pagy#series](https://github.com/ddnexus/pagy/blob/d70e443872a5b18da4e482454b3c8b4a1c86cb6b/gem/lib/pagy.rb#L72-L102)). Unsure if worth it.

Before/after pictures:
will_paginate
pagy

## Testing Instructions

In theory, everything should be working as before on the three modified pages, except for the minor changes listed above.

## Credit

Ceithir (he/him)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -44,7 +44,12 @@ class ApplicationController < ActionController::Base
 
   include Pagy::Backend
   def pagy_query_result(query_result, vars = {})
-    Pagy.new(count: query_result.total_entries, page: params[:page], **vars)
+    Pagy.new(
+      count: query_result.total_entries,
+      page: query_result.current_page,
+      limit: query_result.per_page,
+      **vars
+    )
   end
 
   def display_auth_error
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -49,7 +49,7 @@ class ApplicationController < ActionController::Base
     end
   end
 
-  def pagy_query_result(query_result, vars = {})
+  def pagy_query_result(query_result, **vars)
     pagy_overflow_handler do
       Pagy.new(
         count: query_result.total_entries,
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -43,13 +43,29 @@ class ApplicationController < ActionController::Base
   end
 
   include Pagy::Backend
+  def pagy(collection, **vars)
+    pagy_overflow_handler do
+      super
+    end
+  end
+
   def pagy_query_result(query_result, vars = {})
-    Pagy.new(
-      count: query_result.total_entries,
-      page: query_result.current_page,
-      limit: query_result.per_page,
-      **vars
-    )
+    pagy_overflow_handler do
+      Pagy.new(
+        count: query_result.total_entries,
+        page: query_result.current_page,
+        limit: query_result.per_page,
+        **vars
+      )
+    end
+  end
+
+  def pagy_overflow_handler(&block)
+    begin
+      yield
+    rescue Pagy::OverflowError
+      nil
+    end
   end
 
   def display_auth_error
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -60,12 +60,10 @@ class ApplicationController < ActionController::Base
     end
   end
 
-  def pagy_overflow_handler(&block)
-    begin
-      yield
-    rescue Pagy::OverflowError
-      nil
-    end
+  def pagy_overflow_handler(*)
+    yield
+  rescue Pagy::OverflowError
+    nil
   end
 
   def display_auth_error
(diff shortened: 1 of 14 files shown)


ceithir:
[Supposedly](https://ddnexus.github.io/pagy/docs/how-to/#ignore-brakeman-unescapedoutputs-false-positives-warnings), the Brakeman warnings should be ignored. Not sure if/how I can tell Github Actions to ignore them on my side ([our doc](https://github.com/otwcode/otwarchive/wiki/Brakeman) only refers to the local ignore file)

app/controllers/application_controller.rb
```diff
@@ -42,6 +42,32 @@ def sanitize_ac_params
...
+    end
+  end
+
+  def pagy_overflow_handler(&block)
+    begin
+      yield
+    rescue Pagy::OverflowError
+      nil
+    end
+  end
```
ceithir, line 69: Keeping the same behavior as now by not showing the pagination in case of overflow ("page 66 of 10").

pagy's [alternative options](https://ddnexus.github.io/pagy/docs/extras/overflow/) for reference.

redsummernight:
> [Supposedly](https://ddnexus.github.io/pagy/docs/how-to/#ignore-brakeman-unescapedoutputs-false-positives-warnings), the Brakeman warnings should be ignored. Not sure if/how I can tell Github Actions to ignore them on my side ([our doc](https://github.com/otwcode/otwarchive/wiki/Brakeman) only refers to the local ignore file)

You can add `pagy_nav` as a safe method in [config/brakeman.yml](https://github.com/otwcode/otwarchive/blob/d5cdd4506874a376c6e7989a757b214671567fa7/config/brakeman.yml).

We stopped keeping the ignore file in the repo since #4158. I'll update the wiki.

app/helpers/application_helper.rb
```diff
@@ -528,6 +528,57 @@ def will_paginate(collection_or_options = nil, options = {})
...
+    return nil if pagy.series.length <= 1
+
+    id = %( id="#{id}") if id
+    a  = pagy_anchor(pagy, **vars)
+
+    html = %(<h4 class="landmark heading">#{t('pagination.title')}</h4>)
+
+    html << %(<ol#{id} class="pagination actions pagy" role="navigation" aria-label="#{aria_label || t('pagination.aria-label')}">)
+
+    prev_text = t("pagination.previous")
```
redsummernight, line 546: Is it possible to make `pagy_t('pagy.prev')` output what we want by defining `pagy.prev` in our own view locales file?

app/helpers/application_helper.rb
```diff
@@ -528,6 +528,57 @@ def will_paginate(collection_or_options = nil, options = {})
     super(*[collection_or_options, options].compact)
   end
 
+  include Pagy::Frontend
```
redsummernight, line 531: Let's move this to a separate pagination helper module.

app/controllers/application_controller.rb
ceithir: Without the `require 'pagy/extras/overflow'` line, Pagy always raises an exception in case of overflow ([source](https://ddnexus.github.io/pagy/docs/how-to/#handle-pagyoverflowerror-exceptions))

app/helpers/application_helper.rb
ceithir: With an extra line, yes: https://ddnexus.github.io/pagy/docs/extras/i18n/ Code updated

app/helpers/pagination_helper.rb
```diff
@@ -0,0 +1,65 @@
...
+  def pagy_nav(pagy, id: nil, aria_label: nil, **vars)
+    return nil unless pagy
+
+    # Keep will_paginate behavior of showing nothing if only one page
+    return nil if pagy.series.length <= 1
+
+    id = %( id="#{id}") if id
+    a  = pagy_anchor(pagy, **vars)
+
+    html = %(<h4 class="landmark heading">#{t('pagy.h4')}</h4>)
```
ceithir, line 26: Are invisible h* for navigation purpose still a thing nowadays? Or should we get rid of it?

redsummernight: They are for screen readers, we can keep it. You can use the existing key `t("a11y.navigation")` instead of adding a new `pagy` one.

redsummernight (approved):
Looks good, thank you!

PR #5055, opened 2025-02-08
AO3-6901 Use ActiveJob for asynchronous indexing

redsummernight:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6901

## Purpose

Move the asynchronous part of [AsyncIndexer](https://github.com/otwcode/otwarchive/blob/671cdb9273be76984a14957bfe7e44f0c7bfd841/app/models/search/async_indexer.rb#L56) to Active Job.

I thought about setting `config.active_job.queue_adapter = :test` for all of the test environment, but that broke too many things, including all Cucumber tests expecting emails sent with `deliver_later`. On top of that, pre-7.2 Rails does not consistently follow that setting which complicates things (https://github.com/rails/rails/pull/48585). I'm keeping test updates simple for now.

## Testing Instructions

See issue.

### DIFF
--- a/app/models/search/async_indexer.rb
+++ b/app/models/search/async_indexer.rb
@@ -9,7 +9,7 @@ class AsyncIndexer
     # TODO: Keep the method so we can still run queued jobs from previous
     # versions. However, tests should no longer depend on it.
     #
-    # Remove in a future versions, once all old jobs have been retried or
+    # Remove in a future version, once all old jobs have been retried or
     # cleared.
     raise "Avoid using AsyncIndexer.perform in tests" if Rails.env.test?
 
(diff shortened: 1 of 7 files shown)


brianjaustin (approved):

app/models/search/async_indexer.rb
```diff
@@ -1,12 +1,18 @@
...
 
   ####################
   # CLASS METHODS
   ####################
 
   def self.perform(name)
+    # TODO: Keep the method so we can still run queued jobs from previous
+    # versions. However, tests should no longer depend on it.
+    #
+    # Remove in a future versions, once all old jobs have been retried or
```
sarken, line 12: Little typo if you have a second! Should be version, without the s

PR #5056, opened 2025-02-08
AO3-6903 Remove extra spaces after colons in mailers

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6903

## Purpose

Remove the extra spaces after the labels in the `challenge_assignment_notification` and `abuse_report` mailers. Convert the `challenge_assignment_notification` to use the existing methods for adding the `metadata_label_indicator`. Address a todo to remove the word "work" from those methods, separated into an extra commit for easier reviewing.

Also, make the rating label in challenge_assignment_notification pluralize because I realized that I forgot that in the initial preview changes.

## Testing Instructions

Refer to Jira.

## Credit

Bilka

### DIFF
--- a/app/views/user_mailer/challenge_assignment_notification.text.erb
+++ b/app/views/user_mailer/challenge_assignment_notification.text.erb
@@ -1,7 +1,7 @@
 <% content_for :message do %>
 <%= t(".assignment.text", collection_title: @collection.title, collection_url: collection_url(@collection)) %>
 
-<%= work_metadata_label(t(".recipient")) %><%= @request.nil? ? t(".recipient_missing") : text_pseud(@request.pseud) %>
+<%= metadata_label(t(".recipient")) %><%= @request.nil? ? t(".recipient_missing") : text_pseud(@request.pseud) %>
 
 <%= t(".prompts") %>
 <% @request.requests.each_with_index do |prompt, index| %>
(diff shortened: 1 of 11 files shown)


app/views/user_mailer/challenge_assignment_notification.text.erb
```diff
@@ -1,9 +1,9 @@
 <% content_for :message do %>
-<%= t ".assignment.text", collection_title: @collection.title, collection_url: collection_url(@collection) %>
+<%= t(".assignment.text", collection_title: @collection.title, collection_url: collection_url(@collection)) %>
 
-<%= t ".recipient" %> <%= @request.nil? ? t(".recipient_missing") : text_pseud(@request.pseud) %>
+<%= metadata_label(t(".recipient")) %><%= @request.nil? ? t(".recipient_missing") : text_pseud(@request.pseud) %>
 
-<%= t ".prompts" %>
+<%= t(".prompts") %>
```
redsummernight, line 6: Do we need to use `style_metadata_label` for `t(".prompts")`?

Bilka2: That's a good question. I left it out because it's not really a metadata label, more of a headline. Because this is on its own line there is no worry with whitespace after the colon. But it could still be switched for consistency, I don't have a strong opinion here.

redsummernight: It should be fine for now. We can change it later if necessary.

redsummernight (approved):

PR #5063, opened 2025-02-13
AO3-4250 Made gift recipients update on blurbs on pseud / username changes

melinath:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-4250

## Purpose

Made it so that changing a pseud or username, deleting a pseud, or deleting a user account will expire the cache for all gift works (so that the byline will no longer display the old pseud or username.)

For deleting a pseud, this is done by expiring the cache on all gifts for the default pseud (since gifts are reassigned to the default pseud.) This will clear more caches than necessary - technically we only need to expire the cache for the gifts that were previously assigned to pseud being deleted. I did it this way for now because it's easier & keeps all the code in one place, which I think is easier to understand / less likely to break; however, I can definitely go back and change this if desired.

I also tweaked some of the tests related to series blurb updates (to ensure that they behave properly if only the username is updated and no pseuds) and fixed some minor whitespace issues on adjacent tests.

## Credit

Stephen Lewis, he/him

### DIFF
--- a/app/models/pseud.rb
+++ b/app/models/pseud.rb
@@ -392,6 +392,7 @@ class Pseud < ApplicationRecord
 
   def expire_caches
     return unless saved_change_to_name? || (destroyed? && !user.nil? && !user.default_pseud.nil?)
+
     pseud = destroyed? ? user.default_pseud : self
     pseud.series.each(&:expire_byline_cache)
     pseud.works.each do |work|
--- a/app/models/pseud.rb
+++ b/app/models/pseud.rb
@@ -391,9 +391,11 @@ class Pseud < ApplicationRecord
   end
 
   def expire_caches
-    return unless saved_change_to_name? || (destroyed? && !user.nil? && !user.default_pseud.nil?)
+    return unless saved_change_to_name? || (destroyed? && !user.nil?)
 
     pseud = destroyed? ? user.default_pseud : self
+    return unless !pseud.nil?
+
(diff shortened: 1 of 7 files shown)


app/models/pseud.rb
```diff
@@ -390,9 +391,19 @@ def check_default_pseud
...
+
+    pseud.series.each(&:expire_byline_cache)
+    pseud.works.each do |work|
+      work.touch
+      work.expire_caches
+    end
+    pseud.gift_works.each do |work|
+      work.touch
+      work.expire_caches
     end
```
Bilka2, line 407: The discussion whether touching the works is the right cache busting strategy hasn't quite concluded, but it looks like we were leaning towards it. So, this can be optimized further.

For the gift works, the information is in the blurb. The blurb cache key depends on the work cache key, which depends on the work's updated_at. So it's enough to touch the work to bust the cache, no need to manually do so via `expire_caches`. That means that this can use the more optimised touch_all instead of looping through the gift works.

Your PR doesn't say why you updated `pseud.works` to use `expire_caches` as well, but I think it should be fine to go back to `touch_all` in this case too.

brianjaustin (changes requested):

Bilka2:
Adopted at #5967 based on the [policy for stalled pull requests](https://github.com/otwcode/otwarchive/wiki/Commit-Policy#user-content-Stalled_Pull_Requests_and_Jira_Issues)

PR #5065, opened 2025-02-15
AO3-6909 Bump tk0miya/action-erblint to 44c5fe3

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6909

## Purpose

It gets ahead of the dependabot PR to update the action so that we can bundle the input change in the same PR. This is the same change as #4975 but for erblint instead of rubocop, now that the erblint action also supports the new flag.

## Credit

Bilka

### DIFF
--- a/.github/workflows/reviewdog.yml
+++ b/.github/workflows/reviewdog.yml
@@ -45,8 +45,8 @@ jobs:
           bundler-cache: true
 
       - name: erb-lint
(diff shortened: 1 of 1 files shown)


brianjaustin (approved):

brianjaustin:
Allow-listed `tk0miya/action-erblint@44c5fe3552356fe8bff23f30d534aa4258aa3f7b` and removed the others (besides the previous one, for fork compatibility)

PR #5072, opened 2025-02-17
AO3-6893 Remove patch of departure.rb

scottsds:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6893

## Purpose

Removing the monkeypatch


brianjaustin (approved):

redsummernight:
Looks like this patch just stops departure from using pt-osc, which is not what we want on production. We're removing it and temporarily [disabling advisory locks in `config/database.yml`](https://guides.rubyonrails.org/configuring.html#configuring-a-mysql-or-mariadb-database) just for the migrations in #5039.

PR #5074, opened 2025-02-20
AO3-6911 Add error page for Rack::Timeout::RequestTimeoutException

boyer-victor:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6911 

## Purpose

1. Add a page under `/timeout` that is returned should the app controller encounter `Rack::Timeout::RequestTimeoutException`. 
2. Added a test that the page is rendered on that route.
3. Added a test that the `auth_error` page is rendered on its route to `errors_controller_spec.rb`

## Testing Instructions
*Needs Systems change*, see story for details

I added a unit test that this page can be routed too, but I do not think there is a way to test the actual timeout rescue except from an integration test. [See the docs here.](https://github.com/zombocom/rack-timeout/blob/main/doc/exceptions.md) If I'm missing something or if we have a method for setting up such tests already that someone could point me too it would be much appreciated.

## Credit

unsafe-deref, he/him

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -28,7 +28,7 @@ class ApplicationController < ActionController::Base
   rescue_from Rack::Timeout::RequestTimeoutException, with: :raise_timeout
   
   def raise_timeout
-    redirect_to '/timeout'
+    redirect_to "/timeout"
   end
 
   helper :all # include all helpers, all the time
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -25,6 +25,12 @@ class ApplicationController < ActionController::Base
     redirect_to '/404'
   end
 
+  rescue_from Rack::Timeout::RequestTimeoutException, with: :raise_timeout
+  
+  def raise_timeout
+    redirect_to '/timeout'
+  end
+
   helper :all # include all helpers, all the time
 
   include HtmlCleaner
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -28,7 +28,7 @@ class ApplicationController < ActionController::Base
   rescue_from Rack::Timeout::RequestTimeoutException, with: :raise_timeout
   
   def raise_timeout
-    redirect_to "/timeout"
+    redirect_to timeout_error_path
   end
 
   helper :all # include all helpers, all the time
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -6,10 +6,11 @@ class ErrorsController < ApplicationController
   end
 
   def auth_error
-    @page_subtitle = "Auth Error"
+    @page_subtitle = t(".subtitle")
   end
 
   def timeout_error
-    @page_subtitle = "Timeout Error"
+    @page_subtitle = t(".subtitle")
+    render "timeout_error", status: 200, formats: :html
   end
 end
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -8,4 +8,8 @@ class ErrorsController < ApplicationController
   def auth_error
     @page_subtitle = "Auth Error"
   end
+
+  def timeout_error
+    @page_subtitle = "Timeout Error"
+  end
 end
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -6,11 +6,11 @@ class ErrorsController < ApplicationController
   end
 
   def auth_error
-    @page_subtitle = t(".subtitle")
+    @page_subtitle = t(".browser_title")
   end
 
   def timeout_error
-    @page_subtitle = t(".subtitle")
-    render "timeout_error", status: 504, formats: :html
+    @page_subtitle = t(".browser_title")
+    render "timeout_error", status: :gateway_timeout
   end
 end
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -11,6 +11,6 @@ class ErrorsController < ApplicationController
 
   def timeout_error
     @page_subtitle = t(".subtitle")
-    render "timeout_error", status: 200, formats: :html
+    render "timeout_error", status: 504, formats: :html
   end
 end
--- a/app/views/errors/timeout.html.erb
+++ b/app/views/errors/timeout.html.erb
@@ -1,3 +1,0 @@
-<h2 class="heading"><%= t("errors.timeout.title") %></h2>
-<h3 class="heading"><%= t("errors.timeout.subtitle") %></h3>
-<p><%= t("errors.timeout.html", contact_support_link: link_to(ts("contact Support"), new_feedback_report_path)) %></p>
--- a/app/views/errors/timeout.html.erb
+++ b/app/views/errors/timeout.html.erb
@@ -1,2 +1,3 @@
-<h3 class="heading">The page was responding too slowly. Please try again after a few minutes.</h3>
-<p>If you still get this error after a few tries, you can  <%= link_to ts('contact Support'), new_feedback_report_path %>. In the form, please include a link to the page you're trying to reach and what action you are taking.</p>
+<h2 class="heading"><%= t("errors.timeout.title") %></h2>
+<h3 class="heading"><%= t("errors.timeout.subtitle") %></h3>
+<p><%= t("errors.timeout.html", contact_support_link: link_to(ts("contact Support"), new_feedback_report_path)) %></p>
--- a/app/views/errors/timeout.html.erb
+++ b/app/views/errors/timeout.html.erb
@@ -1,0 +1,2 @@
+<h3 class="heading">The page was responding too slowly. Please try again after a few minutes.</h3>
+<p>If you still get this error after a few tries, you can  <%= link_to ts('contact Support'), new_feedback_report_path %>. In the form, please include a link to the page you're trying to reach and what action you are taking.</p>
--- a/app/views/errors/timeout_error.html.erb
+++ b/app/views/errors/timeout_error.html.erb
@@ -1,0 +1,3 @@
+<h2 class="heading"><%= t(".title") %></h2>
+<h3 class="heading"><%= t(".subtitle") %></h3>
+<p><%= t(".html", contact_support_link: link_to(t(".contact_support_link"), new_feedback_report_path)) %></p>
--- a/app/views/errors/timeout_error.html.erb
+++ b/app/views/errors/timeout_error.html.erb
@@ -1,3 +1,3 @@
-<h2 class="heading"><%= t(".title") %></h2>
+<h2 class="heading"><%= t(".page_heading") %></h2>
 <h3 class="heading"><%= t(".subtitle") %></h3>
-<p><%= t(".html", contact_support_link: link_to(t(".contact_support_link"), new_feedback_report_path)) %></p>
+<p><%= t(".html", contact_support_link: link_to(t(".contact_support"), new_feedback_report_path)) %></p>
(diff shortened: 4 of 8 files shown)


boyer-victor:
I am ignoring the reviewdog comments on single v double quotes as in all instances using single quotes matches the existing page styling or the recommended page to adapt from (re: 500 error page).

redsummernight:
Please do address the reviewdog comments on the parts that you've added or changed. It's fine to deviate from the existing 500 error page in terms of code style.

You don't have to fix any of the existing lint violations, however.

app/views/errors/timeout.html.erb
```diff
@@ -0,0 +1,2 @@
+<h3 class="heading">The page was responding too slowly. Please try again after a few minutes.</h3>
+<p>If you still get this error after a few tries, you can  <%= link_to ts('contact Support'), new_feedback_report_path %>. In the form, please include a link to the page you're trying to reach and what action you are taking.</p>
```
Bilka2, line 2: While you address the reviewdog annotations, please also set up this page (and the page subtitle) for translation. You can find the i18n standards on the github wiki: https://github.com/otwcode/otwarchive/wiki/Internationalization-%28i18n%29-Standards

boyer-victor: Thanks! See changes in latest commit.

app/views/errors/timeout.html.erb
```diff
@@ -0,0 +1,3 @@
+<h2 class="heading"><%= t("errors.timeout.title") %></h2>
+<h3 class="heading"><%= t("errors.timeout.subtitle") %></h3>
+<p><%= t("errors.timeout.html", contact_support_link: link_to(ts("contact Support"), new_feedback_report_path)) %></p>
```
Bilka2, line 3: Please prefer [lazy lookup](https://guides.rubyonrails.org/i18n.html#lazy-lookup), so for example `<h2 class="heading"><%= t(".page_heading") %></h2>`.

For the link, like ERB Lint says, the link caption should also be localised. We have a [specific standard](https://github.com/otwcode/otwarchive/wiki/Internationalization-%28i18n%29-Standards#hyperlinks) for hyperlinks that should result in something like `<p><%= t(".description_html", contact_support_link: link_to(t(".contact_support"), new_feedback_report_path)) %></p>`, with the added string in the locale file

config/routes.rb
```diff
@@ -53,6 +53,7 @@
   get '/422', to: 'errors#422'
   get '/500', to: 'errors#500'
   get '/auth_error', to: 'errors#auth_error'
+  get "/timeout", to: "errors#timeout"
```
Bilka2, line 56: This needs to have the name of the controller action in the `to` option, and then lets match the auth_error for the `get`

```suggestion
  get "/timeout_error", to: "errors#timeout_error"
```

The reason the test didn't fail is because by default, even if there is no method in the controller, Rails renders the HTML template with the action name, which is `timeout.html.erb`. The html.erb should also be renamed to `timeout_error.html.erb` to match the controller (and the auth error).

app/controllers/application_controller.rb
```diff
@@ -25,6 +25,12 @@ def raise_not_found
     redirect_to '/404'
   end
 
+  rescue_from Rack::Timeout::RequestTimeoutException, with: :raise_timeout
+  
+  def raise_timeout
+    redirect_to "/timeout"
```
Bilka2, line 31: This is more of the Rails way:

```suggestion
    redirect_to timeout_error_path
```

config/locales/views/en.yml
```diff
@@ -611,6 +611,11 @@ en:
         restricted_html: A translation of [Restricted Work] by %{creator_link}
         revealed_html: A translation of %{work_link} by %{creator_link}
         unrevealed: A translation of a work in an unrevealed collection
+  errors:
+    timeout:
+      title: Timeout
```
Bilka2, line 616: Please change the heading to "Timeout Error"

spec/controllers/errors_controller_spec.rb
```diff
@@ -28,4 +28,20 @@
       expect(response.header["Content-Type"]).to eq("text/html; charset=utf-8")
     end
   end
+
+  describe "auth_error" do
```
Bilka2, line 32:
```suggestion
  describe "GET #auth_error" do
```

spec/controllers/errors_controller_spec.rb
```diff
@@ -28,4 +28,20 @@
...
+    it "returns an HTML auth error page" do
+      get :auth_error
+      expect(response.status).to eq(200)
+      expect(response.header["Content-Type"]).to eq("text/html; charset=utf-8")
+    end
+  end
+
+  describe "timeout" do
+    it "returns an HTML timeout error page" do
+      get :timeout
```
Bilka2, line 42:
```suggestion
  describe "GET #timeout_error" do
    it "returns an HTML timeout error page" do
      get :timeout_error
```

Bilka2 (changes requested):

app/controllers/errors_controller.rb
```diff
@@ -8,4 +8,8 @@ class ErrorsController < ApplicationController
   def auth_error
     @page_subtitle = "Auth Error"
   end
+
+  def timeout_error
+    @page_subtitle = "Timeout Error"
```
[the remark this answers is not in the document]
boyer-victor: Thanks! Apologies, I am used to a different pattern and did not look carefully enough at the locale specs for this project. I updated `auth_error` as well.

config/routes.rb
boyer-victor: Tricky, I would prefer just a failure than a default loading that is non-obvious. Thanks for pointing it out, resolved.

spec/controllers/errors_controller_spec.rb
```diff
@@ -28,4 +28,20 @@
       expect(response.header["Content-Type"]).to eq("text/html; charset=utf-8")
     end
   end
+
+  describe "auth_error" do
+    it "returns an HTML auth error page" do
+      get :auth_error
+      expect(response.status).to eq(200)
```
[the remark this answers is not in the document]
boyer-victor: Updated to 504. I did **not** update the auth_error page to a 401 as I am not aware if it would have downstream impact.

app/controllers/errors_controller.rb
```diff
@@ -6,6 +6,11 @@ class ErrorsController < ApplicationController
   end
 
   def auth_error
-    @page_subtitle = "Auth Error"
+    @page_subtitle = t(".subtitle")
+  end
+
+  def timeout_error
+    @page_subtitle = t(".subtitle")
```
Bilka2, line 13: This should use different locale keys, like "browser_title" to distinguish between this and the subtitle on the page itself

config/locales/controllers/en.yml
```diff
@@ -90,6 +90,11 @@ en:
       error: Sorry, that comment could not be unhidden.
       permission_denied: Sorry, you don't have permission to unhide that comment.
       success: Comment successfully unhidden!
+  errors:
+    timeout_error:
+      subtitle: "Timeout Error"
+    auth_error:
+      subtitle: "Authentication Error"
```
Bilka2, line 97:
```suggestion
      subtitle: "Auth Error"
```

app/controllers/errors_controller.rb
```diff
@@ -6,6 +6,11 @@ class ErrorsController < ApplicationController
...
+    render "timeout_error", status: 504, formats: :html
```
Bilka2, line 14: The format defaults to html, so it can be left out:
```suggestion
    render "timeout_error", status: :gateway_timeout
```

config/locales/views/en.yml
```diff
@@ -611,6 +611,12 @@ en:
         restricted_html: A translation of [Restricted Work] by %{creator_link}
         revealed_html: A translation of %{work_link} by %{creator_link}
         unrevealed: A translation of a work in an unrevealed collection
+  errors:
+    timeout_error:
+      title: Timeout
```
Bilka2, line 616:
```suggestion
      title: Timeout Error
```

app/views/errors/timeout_error.html.erb
```diff
@@ -0,0 +1,3 @@
+<h2 class="heading"><%= t(".title") %></h2>
+<h3 class="heading"><%= t(".subtitle") %></h3>
+<p><%= t(".html", contact_support_link: link_to(t(".contact_support_link"), new_feedback_report_path)) %></p>
```
Bilka2, line 3: Please follow the i18n standard:
```suggestion
<p><%= t(".html", contact_support_link: link_to(t(".contact_support"), new_feedback_report_path)) %></p>
```

config/locales/views/en.yml
```diff
@@ -611,6 +611,12 @@ en:
...
+      subtitle: The page was responding too slowly. Please try again after a few minutes.
+      html: If you still get this error after a few tries, please %{contact_support_link}.
+      contact_support_link: contact Support
```
Bilka2, line 619:
```suggestion
      contact_support: contact Support
```

app/views/errors/timeout_error.html.erb
```diff
@@ -0,0 +1,3 @@
+<h2 class="heading"><%= t(".title") %></h2>
```
Bilka2, line 1: This is a style thing, but we usually use this locale key for h2 headings: 
```suggestion
<h2 class="heading"><%= t(".page_heading") %></h2>
```

config/locales/views/en.yml
```diff
@@ -611,6 +611,12 @@ en:
         restricted_html: A translation of [Restricted Work] by %{creator_link}
         revealed_html: A translation of %{work_link} by %{creator_link}
         unrevealed: A translation of a work in an unrevealed collection
+  errors:
+    timeout_error:
+      title: Timeout
+      subtitle: The page was responding too slowly. Please try again after a few minutes.
+      html: If you still get this error after a few tries, please %{contact_support_link}.
```
Bilka2, line 618:
```suggestion
      html: If you still get this error after a few tries, you can %{contact_support_link}.
```

Bilka2 (commented):
As redsummernight said in their comment, please address the reviewdog comments for code that you added or modified. Please also fix the test by running the suggested command

Bilka2 (changes requested):
Thanks for the locale fixes!

Please fix the failing test. If you're using Docker or gitpod, you can run the suggested command in the testing container by prefixing it with `docker compose run --rm test `.

boyer-victor:
> Thanks for the locale fixes!
> 
> Please fix the failing test. If you're using Docker or gitpod, you can run the suggested command in the testing container by prefixing it with `docker compose run --rm test `.

Sorry! I had missed that that had even failed. 

Seems like a potential improvement to flag that somehow for changes to locale files. I've used pre-commit-hooks in the past for things like rustfmt or eslint that would likely fit that need (though be one more thing that people would have to configure).

Bilka2 (approved):

Bilka2:
Thank you for the fix!

> I've used pre-commit-hooks in the past for things like rustfmt or eslint that would likely fit that need (though be one more thing that people would have to configure).

Yeah, the configuration is the issue since some people develop without Docker, some with Docker.  So beyond having to set up the hook, the hook itself would also need to support a variety of systems. Any issues with that would be hard to debug for the same people who would have trouble with running the command manually :/

boyer-victor:
> Thank you for the fix!
> 
> > I've used pre-commit-hooks in the past for things like rustfmt or eslint that would likely fit that need (though be one more thing that people would have to configure).
> 
> Yeah, the configuration is the issue since some people develop without Docker, some with Docker. So beyond having to set up the hook, the hook itself would also need to support a variety of systems. Any issues with that would be hard to debug for the same people who would have trouble with running the command manually :/

It's hard enough for my team that I can talk to daily... Regardless I might try some things out and open a documentation request if I get something that works well at least with docker in my use case. Thanks for being patient!

PR #5082, opened 2025-02-26
AO3-6401 Prevent official user accounts from leaving kudos

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6401

## Purpose

Adds an error message if you're logged in with your official account and try to leave kudos. Also adds a task to clean up any kudos from existing official accounts.

Note: There are only 22 official accounts and I'm sure very few, if any, kudos from them, so I wasn't overly concerned with batching or performance for the task.

## Testing Instructions

Refer to Jira.

### DIFF
--- a/lib/tasks/after_tasks.rake
+++ b/lib/tasks/after_tasks.rake
@@ -521,13 +521,14 @@ namespace :After do
   desc "Convert user kudos from users with the official role to guest kudos"
   task(convert_official_kudos: :environment) do
     official_role = Role.find_by(name: "official")
-    official_users = RolesUser.where(role_id: official_role.id).map(&:user) unless official_role.blank?
+    official_users = RolesUser.where(role_id: official_role.id).map(&:user) if official_role.present?
     if official_role.blank? || official_users.blank?
       puts "No official users found"
     else
       official_users.each do |user|
         kudos = user.kudos
-        next unless kudos.present?
+        next if kudos.blank?
+
         puts "Updating #{kudos.size} kudos from #{user.login}"
         user.kudos.each do |kudo|
           kudo.update_attribute(:user_id, nil)
--- a/lib/tasks/after_tasks.rake
+++ b/lib/tasks/after_tasks.rake
@@ -530,7 +530,6 @@ namespace :After do
 
         puts "Updating #{kudos.size} kudos from #{user.login}"
         user.remove_user_from_kudos
-        end
       end
 
       puts "Finished converting kudos from official users to guest kudos"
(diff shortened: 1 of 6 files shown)


lib/tasks/after_tasks.rake
```diff
@@ -517,5 +517,26 @@ namespace :After do
 
     puts "Finished reindexing tags on hidden and unrevealed works"
   end
+
+  desc "Convert user kudos from users with the official role to guest kudos"
+  task(convert_official_kudos: :environment) do
+    official_role = Role.find_by(name: "official")
+    official_users = RolesUser.where(role_id: official_role.id).map(&:user) if official_role.present?
```
redsummernight, line 524: You can use `official_role.users`.

lib/tasks/after_tasks.rake
```diff
@@ -517,5 +517,26 @@ namespace :After do
...
+    if official_role.blank? || official_users.blank?
+      puts "No official users found"
+    else
+      official_users.each do |user|
+        kudos = user.kudos
+        next if kudos.blank?
+
+        puts "Updating #{kudos.size} kudos from #{user.login}"
+        user.kudos.each do |kudo|
```
redsummernight, line 533: You can use `user.remove_user_from_kudos`.

lib/tasks/after_tasks.rake
sarken: I was going to say "TIL!" but then I realized what you were saying and wow, last night me was a dumbass.

lib/tasks/after_tasks.rake
sarken: ...I might have forgotten #4264 was done. (Why did I think that was the issue we were just talking about whether we could use `touch` with?)

lib/tasks/after_tasks.rake
```diff
@@ -517,5 +517,24 @@ namespace :After do
...
+  task(convert_official_kudos: :environment) do
+    official_users = Role.find_by(name: "official")&.users
+    if official_users.blank?
+      puts "No official users found"
+    else
+      official_users.each do |user|
+        kudos = user.kudos
+        next if kudos.blank?
+
+        puts "Updating #{kudos.size} kudos from #{user.login}"
```
sarken, line 531: If we don't care about 100% accurate or informative output, I can also just do this even if the user has no kudos:
```
        puts "Updating kudos from #{user.login}"
        user.remove_user_from_kudos
      end
```

redsummernight: Let's keep the count. It'll be fun to see which official user has the most stray clicks.

lib/tasks/after_tasks.rake
redsummernight: We _were_ heavily referencing the discussions on that PR.

redsummernight (approved):

lib/tasks/after_tasks.rake
sarken: Oh, yay, I'm not _totally_ losing it 🎉

PR #5084, opened 2025-02-27
AO3-6544 Send comment_type and user_role to spam checker for tickets and comments

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6544

## Purpose

Lets the spam checker know if it's checking 
- content from a guest or (on the Suport and PAC contact forms) someone who is logged in but has entered an email address that isn't the one on their account
- content from a contact form, blog (admin post) comment, or fanwork comment (as opposed to a fanwork)

## Testing Instructions

Refer to Jira

## References

* [Akismet comment check docs](https://akismet.com/developers/detailed-docs/comment-check/)
* A June 21, 2023 email discussion with the Akismet devs saying this would be nice to eventually have

### DIFF
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -93,7 +93,12 @@ class Comment < ApplicationRecord
   has_comment_methods
 
   def akismet_attributes
+    # While we do have tag comments, those are from logged-in users with special
+    # access granted by admins, so we never spam check them, unlike comments on
+    # works or admin posts.
+    comment_type = ultimate_parent.is_a?(Work) ? "fanwork-comment" : "comment"
     {
+      comment_type: comment_type,
       key: ArchiveConfig.AKISMET_KEY,
       blog: ArchiveConfig.AKISMET_NAME,
       user_ip: ip_address,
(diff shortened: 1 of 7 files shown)


app/models/comment.rb
redsummernight, line 1: Do we need to set `comment_type` on comments, or is `comment` assumed by Akismet to be the default?

sarken: It's not something they suggested we do, but we can! However, I think it might be worth getting a little fancier. The [docs](https://akismet.com/blog/pro-tip-tell-us-your-comment_type/) say:

> Akismet works with almost any kind of user‑submitted web content: blog comments, forum posts, blog posts, contact forms and so on. The characteristics of spam tend to vary across those type – comment spam is quite different from forum spam. So it’s important to give us some context by telling us what type of messages you’re asking Akismet to check.
> [...]
>It’s especially important not to default to comment for messages that are fundamentally different from blog comments – if you do that, you can expect to see mixed results. It’s better to be too specific than too ambiguous.

So I think the best option for us would be to use:
* `comment` for comments on admin posts, which are basically blog posts 
* `fanwork-comment` for comments on works

redsummernight (approved):

PR #5088, opened 2025-03-01
AO3-6830 Add creator IDs to abuse reports

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6830

## Purpose

Adds creator IDs to abuse reports for works. Similar to the attached works, no info is added if the reported URL is a comment on a work.

## Testing Instructions

Refer to Jira.

## Credit

Bilka


app/models/abuse_report.rb
```diff
@@ -123,10 +125,27 @@ def send_report
...
   end
 
+  def creator_ids
+    work_id = reported_work_id
+    return unless work_id
+
+    work = Work.find_by(id: work_id)
+    return "deletedwork" unless work
+
+    ids = work.pseuds.pluck(:user_id).push(*work.original_creators.pluck(:user_id)).uniq.sort!
```
brianjaustin, line 135: Could we use `sort` here instead of `sort!`, since the return value is being used (and we don't necessarily need to mutate the original)?

brianjaustin (approved):

PR #4871, opened 2024-07-10
AO3-6754 Restrict access to API tokens

nisha-shaikh:
# Pull Request Checklist

* [x] Have you read ["How to write the perfect pull request"](https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/)?
* [x] Have you read the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [x] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)?
* [x] Have you added the [Jira](https://otwarchive.atlassian.net) issue number
  as the *first* thing in your pull request title (e.g. `AO3-1234 Fix thing`)
* [x] Do you have fewer than 5 pull requests already open? If not, please wait
  until they are reviewed and merged before creating new pull requests.

## Issue

https://otwarchive.atlassian.net/browse/AO3-6754

## Purpose

- Restrict access to management of API tokens to only super admins
- Rename browser titles for API token pages

## Credit

Nisha (she/her)

PR #5089, opened 2025-03-01
AO3-5513 Allow admins to create, recategorize and rename media tags

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5513

## Purpose

Fix various errors and forms to allow admins to create media tags and recategorize tags as media tags. Add cache busting for the fandoms menu in the header and the "Find your favorites" fandom module on the homepage, when creating, destroying, renaming and recategorizing media tags.

## Testing Instructions

Create media tags, make sure they show up everywhere, rename them, make sure that shows up everywhere, recategorize them to other types, make sure they disappear from all media listings, recategorize them as media tags again and make sure they show up everywhere.

## References

https://github.com/sarken/otwarchive/tree/AO3-5513

PR Limit: 6 out of 5
Using reviewer bonus:
#4871
#4855

## Credit

Sarken for the code linked above
Bilka


app/controllers/tags_controller.rb
```diff
@@ -181,10 +181,10 @@ def create
       return
     end
 
-    raise "Redshirt: Attempted to constantize invalid class initialize create #{type.classify}" unless Tag::TYPES.include?(type.classify)
+    raise "Redshirt: Attempted to constantize invalid class initialize create #{type.classify}" unless Tag::TYPES.include?(type.classify) || type == "Media"
```
redsummernight, line 184: Instead of special-casing everywhere we use `.classify` on tag types, you could add an irregular inflection rule that "media" is both singular and plural:

https://github.com/otwcode/otwarchive/blob/fd5046c72ca2b2f6874aaaf839bd54a5410e2677/config/initializers/inflections.rb#L9

features/step_definitions/tag_steps.rb
```diff
@@ -102,6 +102,10 @@
   rel.add_association(fand)
 end
 
+Given "the non(-)canonical media {string}" do |tag_name|
+  Media.find_or_create_by_name(tag_name)
+end
+
 Given /^a (non-?canonical|canonical) (\w+) "([^\"]*)"$/ do |canonical_status, tag_type, tag_name|
```
redsummernight, line 109: With inflection rules, you can reuse this step instead of creating a separate one for media.

features/step_definitions/tag_steps.rb
```diff
@@ -226,6 +230,22 @@
   blank_tag.save!(validate: false)
 end
 
+Given "I have just created the canonical media tag {string}" do |name|
```
redsummernight, line 233: Let's use "I create" / "I recategorize", most of our step names use present tense.

app/controllers/tags_controller.rb
```diff
@@ -248,9 +248,8 @@ def update
     new_tag_type = params[:tag].delete(:type)
 
     # Limiting the conditions under which you can update the tag type
-    if @tag.can_change_type? && %w(Fandom Character Relationship Freeform UnsortedTag).include?(new_tag_type)
-      @tag = @tag.recategorize(new_tag_type)
-    end
+    types = logged_in_as_admin? ? Tag::TYPES : Tag::USER_DEFINED
```
redsummernight, line 251: This will also allow rating, warning, and category tags to be renamed by admins. Let's hold off on that.

redsummernight (changes requested):

app/controllers/fandoms_controller.rb
```diff
@@ -5,22 +5,22 @@ def index
...
       @media = Media.canonical.by_name - [Media.find_by(name: ArchiveConfig.MEDIA_NO_TAG_NAME)] - [Media.find_by(name: ArchiveConfig.MEDIA_UNCATEGORIZED_NAME)]
       @page_subtitle = @collection.title
-      @medium = Media.find_by_name(params[:medium_id]) if params[:medium_id]
+      @medium = Media.find_by_name(params[:media_id]) if params[:media_id]
       @counts = SearchCounts.fandom_ids_for_collection(@collection)
       @fandoms = (@medium ? @medium.fandoms : Fandom.all).where(id: @counts.keys).by_name
-    elsif params[:medium_id]
-      if @medium = Media.find_by_name(params[:medium_id])
+    elsif params[:media_id]
+      if @medium = Media.find_by_name(params[:media_id])
```
redsummernight, line 12: Let's do a `find_by!` and avoid the if-else-raise `ActiveRecord::RecordNotFound` setup entirely.

app/views/home/_fandoms.html.erb
```diff
@@ -1,9 +1,10 @@
 <ul>
-  <li><%= link_to t(".all_fandoms"), media_path %></li>
+  <li><%= link_to t(".all_fandoms"), media_index_path %></li>
+  <%# When changing the cache key, also update Media#expire_caches %>
   <% cache "homepage-fandoms-version1", skip_digest: true do %>
     <% Media.for_menu.each do |medium| %>
       <% unless medium.id.nil? %>
-        <li><%= link_to ts("#{medium.name}"), medium_fandoms_path(medium) %></li>
+        <li><%= link_to ts("#{medium.name}"), media_fandoms_path(medium) %></li>
```
redsummernight, line 7: Let's skip `ts` here. It's a tag name so we're likely not localizing it.

redsummernight (commented):
Sorry, the inflection rule changed more things than I expect. You can sidestep all the reviewdog comments about using `t` over `ts` by removing the helper entirely, you've had to do way more than the original issue scope as it is.

Bilka2:
> You can sidestep all the reviewdog comments about using t over ts by removing the helper entirely, you've had to do way more than the original issue scope as it is.

I knew what I was getting into, no worries! I'd rather do it properly, so I fixed the lines reviewdog pointed out but didn't go and i18n much beyond that so that the review doesn't get too large.

app/controllers/fandoms_controller.rb
```diff
@@ -5,32 +5,29 @@ def index
...
-        if @medium == Media.uncategorized
-          @fandoms = @medium.fandoms.in_use.by_name
-        else
-          @fandoms = @medium.fandoms.canonical.by_name.with_count
-        end
-      else
-        raise ActiveRecord::RecordNotFound, "Couldn't find media category named '#{params[:medium_id]}'"
-      end
+    elsif params[:media_id]
+      @medium = Media.find_by_name!(params[:media_id]) # rubocop:disable Rails/DynamicFindBy
```
redsummernight, line 12: Since it's a method we add, you can put it on the list here:
https://github.com/otwcode/otwarchive/blob/fd5046c72ca2b2f6874aaaf839bd54a5410e2677/.rubocop.yml#L134-L137

redsummernight (approved):

PR #5077, opened 2025-02-21
AO3-6799 update ticket url fields

lydia-theda:
# Pull Request Checklist

* [x] Have you read ["How to write the perfect pull request"](https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/)?
* [x] Have you read the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [x] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)?
* [x] Have you added the [Jira](https://otwarchive.atlassian.net) issue number
  as the *first* thing in your pull request title (e.g. `AO3-1234 Fix thing`)
* [x] Do you have fewer than 5 pull requests already open? If not, please wait
  until they are reviewed and merged before creating new pull requests.

## Issue

https://otwarchive.atlassian.net/browse/AO3-6799

## Purpose

updates the cf_url field (limited to 255 characters) to the new cf_ticket_url field (limit of 2080 characters)
removes "Unknown URL" entry for blank referers

## References

https://otwarchive.atlassian.net/browse/AO3-6232

## Credit

lydia-theda

PR #5092, opened 2025-03-03
AO3-6799 Update Zoho URL field names and length limit

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6799

## Purpose

Change the name of the field sent to Zoho from `cf_url` to `cf_ticket_url` and updates the limit to 2080 to reflect the new field limit

## References

Adopts the work from https://github.com/otwcode/otwarchive/pull/5077

## Credit

lydia-theda
Brian Austin


db/migrate/20250303154718_increase_report_ticket_url_limit.rb
```diff
@@ -0,0 +1,7 @@
+class IncreaseReportTicketUrlLimit < ActiveRecord::Migration[7.1]
+  uses_departure! if Rails.env.staging? || Rails.env.production?
+
+  def change
```
Bilka2, line 4: According to https://guides.rubyonrails.org/active_record_migrations.html#changing-columns we need to use #up and #down for a reversible change_column migration.

Bilka2 (changes requested):

db/migrate/20250303154718_increase_report_ticket_url_limit.rb
brianjaustin: Sorry, I forgot to add a note: I don't think we should make a down migration, because there is a risk of data loss if limits are changed to a lower length

Bilka2 (approved):

db/migrate/20250303154718_increase_report_ticket_url_limit.rb
redsummernight: Let's make that clear by moving `change_column` to `up` and have a `down` method that is either empty or throwing [an irreversible error](https://guides.rubyonrails.org/active_record_migrations.html#throwing-an-error-to-prevent-reverts).

redsummernight (approved):

PR #5093, opened 2025-03-04
AO3-5573 Fix sorting on user collected works page

metallicity:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5573

## Purpose

The "Works in Collections" page (`/users/<name>/works/collected`), where all of a user's works in collections can be filtered by those collections, had a broken "Sort by" field. Regardless of what was selected, it always sorted by "Date Updated". This PR corrects the issue by changing the improperly named `work_search_form[sort_column]` to just `work_search[sort_column]`, as expected.

## Testing Instructions

1. Visit the "Works" page of a user who has published multiple works in collections
2. Navigate to "Works in Collections" (url: "/users/[name]/works/collected")
3. Verify the collected works appear with the latest updated work first
4. Sort by "Word Count" and verify the works appear with the largest work first
5. Sort by "Title" and verify the works appear alphabetically by title

## Credit

Metallicity (they/them)


features/search/works_collected.feature
```diff
@@ -0,0 +1,24 @@
...
+
+  Scenario: Works that are collected show up in the creator's
+  collected works page
+    Given I am logged in as "author"
+      And I create the collection "Author Collection"
+      And I create the collection "Other Collection"
+      And I post the work "Old Title" to the collection "Author Collection"
+      And I post the work "Revised Title" to the collection "Author Collection"
+      And I post the work "New Title" to the collection "Other Collection"
+      And I post a chapter for the work "Revised Title"
```
Bilka2, line 13: Please add a `And I wait 1 second` before both of these two steps. Otherwise, if the test runs especially quickly, the works can end up with the same timestamp which messes with the sorting.

features/search/works_collected.feature
```diff
@@ -0,0 +1,24 @@
...
+    When I go to author's user page
+      And I follow "Works (3)"
+      And I follow "Works in Collections"
+    Then I should see "Works in Challenges/Collections"
```
Bilka2, line 17: Can you add a comment that this is sorting by date updated by default?

spec/controllers/works/default_rails_actions_spec.rb
```diff
@@ -815,6 +815,51 @@ def call_with_params(params)
         expect(assigns(:works)).to include(work, unrevealed_work)
       end
     end
+
+    context "with sorting options" do
```
Bilka2, line 819: This test also passes on `master`, since the bug is with the form itself, not the search. I don't think we need to test the search separately like this because it's already covered by the feature test. Could you remove the test?

features/search/works_collected.feature
Bilka2, line 1: Could you add a test for filtering with the collection name facet? Just to make sure it's not broken by the change.

Bilka2 (commented):
The change looks good, thank you! I just left some notes on the tests.

spec/controllers/works/default_rails_actions_spec.rb
metallicity: I wrote this test first, then realized it was already passing, which led me to find the actual issue. Figured I'd leave it in the PR since I already wrote it, but happy to remove it. Technically this is also checking that the `sort_direction` param works correctly here, which UI tests are not, due to not being accessible through the UI in the first place. If that seems important, I can restore this test again.

metallicity:
I decided to have the sorting tests check the actual value of "Sort by" on each page, rather than just leaving a comment claiming this was true. And on top of adding tests for filtering by collection name, I also included one additional sorting test for "Date Posted". Half the reason for setting up this trio of Old/New/Revised works was to ensure they'd have different ordering for Updated vs. Posted, and removing the rspec test meant this was no longer covered elsewhere.

spec/controllers/works/default_rails_actions_spec.rb
Bilka2: Oh, the sort direction is a good point. Since they're already written and you're offering, please do restore them. Thank you!

Bilka2 (approved):
Thank you very much!

PR #5094, opened 2025-03-04
AO3-6591 Move imported_from_url to a new table

ceithir:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6591

## Purpose

First step of migrating `imported_from_url` into its own table:
- Create new table
- Update existing place where the url is imported to fill it properly
- Update search to use it
- Provide a task to migrate existing import urls

What this PR does not do and that will have to be done in a second time, after the new task is successfully run, is deleting all the code still using imported_from_url (albeit now as a fallback) and dropping the column from the table.

## Testing Instructions

No visible change to any behavior related to importing urls/searching by import urls. Lines should be created/updated in the new table however.

## Credit

Ceithir (he/him)

### DIFF
--- a/app/controllers/api/v2/works_controller.rb
+++ b/app/controllers/api/v2/works_controller.rb
@@ -127,6 +127,7 @@ class Api::V2::WorksController < Api::V2::BaseController
       message = "Please provide the original URL for the work."
     else
       # We know the url will be identical no need for a call to find_by_url
+      # TODO AO3-6591
       works = Work.where(imported_from_url: original_url)
       message = if works.empty?
                   "No work has been imported from \"" + original_url + "\"."
--- a/app/controllers/opendoors/tools_controller.rb
+++ b/app/controllers/opendoors/tools_controller.rb
@@ -51,7 +51,8 @@ class Opendoors::ToolsController < ApplicationController
         flash[:error] = ts("There is already a work imported from the url %{url}.", url: @imported_from_url)
       else
         # ok let's try to update
-        @work.update_attribute(:imported_from_url, @imported_from_url)
+        @work.update_attribute(:imported_from_url, @imported_from_url) # TODO AO3-6591
+        WorkUrl.prepare(@work, @imported_from_url).save(validate: false)
         flash[:notice] = "Updated imported-from url for #{@work.title} to #{@imported_from_url}"
       end
     end    
--- a/app/models/work_url.rb
+++ b/app/models/work_url.rb
@@ -1,0 +1,73 @@
+class WorkUrl < ApplicationRecord
+  belongs_to :work
+
+  VARIANTS = %w[original minimal no_www with_www encoded decoded minimal_no_protocol_no_www].freeze
+
+  def self.build_url_attributes(url)
+    url = UrlFormatter.new(url)
+
+    VARIANTS.to_h { |s| [s.to_sym, url.send(s)] }
+  end
+
+  def self.prepare(work, url)
+    work_url = WorkUrl.find_or_initialize_by(work_id: work.id)
+
+    work_url.assign_attributes(self.build_url_attributes(url))
+
+    work.work_url = work_url
+
+    work_url
+  end
+
+  def self.find_by_url_generation_key
+    "/v1/find_by_url_generation_key"
+  end
+
+  def self.find_by_url_generation
+    Rails.cache.fetch(WorkUrl.find_by_url_generation_key, raw: true) { rand(1..1000) }
+  end
+
+  def self.flush_find_by_url_cache
+    Rails.cache.increment(WorkUrl.find_by_url_generation_key)
+  end
+
+  def self.find_by_url_cache_key(url)
+    url = UrlFormatter.new(url)
+    "/v1/find_by_url/#{WorkUrl.find_by_url_generation}/#{url.encoded}"
+  end
+
+  # Match `url` to a work's imported_from_url field using progressively fuzzier matching:
+  # 1. first exact match
+  # 2. first exact match with variants of the provided url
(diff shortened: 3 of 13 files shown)


app/controllers/opendoors/tools_controller.rb
```diff
@@ -51,7 +51,8 @@ def url_update
         flash[:error] = ts("There is already a work imported from the url %{url}.", url: @imported_from_url)
       else
         # ok let's try to update
-        @work.update_attribute(:imported_from_url, @imported_from_url)
+        @work.update_attribute(:imported_from_url, @imported_from_url) # TODO AO3-6591
+        WorkUrl.prepare(@work, @imported_from_url).save(validate: false)
```
ceithir, line 56: Disabling validation to be consistent with the `update_attribute` just above... but might be a bad idea actually.

app/models/work_url.rb
```diff
@@ -0,0 +1,73 @@
...
+
+  def self.prepare(work, url)
+    work_url = WorkUrl.find_or_initialize_by(work_id: work.id)
+
+    work_url.assign_attributes(self.build_url_attributes(url))
+
+    work.work_url = work_url
+
+    work_url
+  end
```
ceithir, line 20: I'm sure there's a more idiosyncratic Rails syntax for that but I'm drawing a blank at the moment.

app/models/opendoors.rb
```diff
@@ -2,4 +2,19 @@ module Opendoors
   def self.table_name_prefix
     'opendoors_'
   end
+
+  def self.update_work_from_csv(row)
```
ceithir, line 6: Rubocop really wanted to have it out of the task file.

brianjaustin (changes requested):
Some of my thoughts, but someone from Open Doors might come around to review this in the next couple weeks, so feel free to wait until then to take any action on them

app/controllers/api/v2/works_controller.rb
```diff
@@ -127,6 +127,7 @@ def find_work_by_import_url(original_url)
       message = "Please provide the original URL for the work."
     else
       # We know the url will be identical no need for a call to find_by_url
+      # TODO: AO3-6591
```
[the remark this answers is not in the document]
ceithir: Possibly. I'm sure this is not the first time a feature on AO3 had to be split into two PRs with a task run in the middle but I haven't found anything in the wiki about what the standard procedure in that case. Create two sub-issues on JIRA maybe?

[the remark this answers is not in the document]
brianjaustin: Yes, we should split this into two different Jira tickets (that makes accounting for things in the releases easier). If you don't have access to do that please let me know, and I can split the ticket (once that happens you should have edit access to the newly created one as usual)

spec/lib/tasks/opendoors.rake_spec.rb
```diff
@@ -15,16 +15,19 @@
...
-      expect(Work.find(work_with_no_url.id).imported_from_url).to eq(original_url2)
     end
     it "fails if the CSV path is invalid" do
       expect { subject.invoke("not a path") }.to raise_error Errno::ENOENT
     end
   end
 
-  context "UrlUpdater" do
-    let(:url_updater) { UrlUpdater.new }
+  context "update_work_from_csv" do
```
brianjaustin, line 30: Should this be moved to `spec/lib/models/opendoors_spec.rb` (new file)?

app/models/work_url.rb
```diff
@@ -0,0 +1,74 @@
...
+
+  def self.find_by_url_generation_key
```
brianjaustin, line 22: Can we make most of these `private` to avoid other classes calling them that don't need to? (Similar question from for the method `build_url_attributes` above.)

spec/models/work_spec.rb
```diff
@@ -463,38 +467,99 @@
...
+        work.destroy
+      end
+
+      it "finds works imported with HTTPS protocol and irrelevant query parameters" do
+        work = create(:work, imported_from_url: "https://lj-site.com/thing1?style=mine")
+        expect(Work.find_by_url("http://lj-site.com/thing1?style=other")).to eq(work)
+        work.destroy
+      end
+
+      it "gets the work from cache when searching for an imported work by URL" do
```
brianjaustin, line 554: Since these are now testing a method on `WorkUrl`, could you move them to a new spec, `spec/models/work_url_spec.rb`?

spec/models/work_spec.rb
```diff
@@ -463,38 +467,99 @@
...
+      work.import_url!(url)
+      expect(Rails.cache.read(WorkUrl.find_by_url_cache_key(url))).to be_nil
       expect(Work.find_by_url(url)).to eq(work)
-      expect(Rails.cache.read(Work.find_by_url_cache_key(url))).to eq(work)
+      expect(Rails.cache.read(WorkUrl.find_by_url_cache_key(url))).to eq(work)
       work.destroy
     end
+
+    describe "legacy – to be removed eventually – cf AO3-6979" do
+      it "should find imported works with various URL formats" do
```
brianjaustin, line 509: For this (and the other modified cases), can you reword them to avoid should? https://www.betterspecs.org/#should

app/models/work_url.rb
brianjaustin: Actually, I am wondering if we will need them long term (e.g. can the caching be removed once we get everything converted to the new table?)

Bilka2:

ceithir:
I am clearly not actively working on this, so closing. Has been a long time, not sure if there's anything worth scavenging there for anyone willing to take on that issue.

PR #5095, opened 2025-03-04
Revert "AO3-6712 Allow invites to be resent and index invitee_email (#4791)"

brianjaustin:
Just in case we need to revert due to performance issues when the original code is put into production

This reverts commit 129e6e954e097bb986d0ead807638e8f41b03f3e.


redsummernight:
Deploy went well, we don't need to revert #4791.

PR #5099, opened 2025-03-13
AO3-6933 Add spam checking to user comments

EchoEkhi:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6933

## Purpose

- Adds a new admin setting called account_age_threshold_for_comment_spam_check (self-explanatory)
- When an account is too new, its comments are spam-checked
- When an account gets spam-banned by an admin, all of its comments are marked as spam and sent to Akismet for training

## Testing Instructions

See JIRA ticket

## Credit
EchoEkhi

### DIFF
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -65,7 +65,7 @@ class Comment < ApplicationRecord
   validate :check_for_spam, on: :create
 
   def check_for_spam
-    errors.add(:base, ts("This comment looks like spam to our system, sorry! Please try again, or create an account to comment.")) unless check_for_spam?
+    errors.add(:base, ts("This comment looks like spam to our system, sorry! Please try again.")) unless check_for_spam?
   end
 
   validates :comment_content, uniqueness: {
@@ -97,15 +97,26 @@ class Comment < ApplicationRecord
     # access granted by admins, so we never spam check them, unlike comments on
     # works or admin posts.
     comment_type = ultimate_parent.is_a?(Work) ? "fanwork-comment" : "comment"
+
+    if self.pseud_id.nil?
+      user_role = "guest"
+      comment_author = name
+      comment_author_email = email
+    else
+      user_role = "user"
+      comment_author = self.pseud.user.login
+      comment_author_email = self.pseud.user.email
+    end
+    
     {
       comment_type: comment_type,
       key: ArchiveConfig.AKISMET_KEY,
       blog: ArchiveConfig.AKISMET_NAME,
       user_ip: ip_address,
       user_agent: user_agent,
-      user_role: "guest",
-      comment_author: name,
-      comment_author_email: email,
+      user_role: user_role,
+      comment_author: comment_author,
+      comment_author_email: comment_author_email,
       comment_content: comment_content
     }
   end
@@ -466,8 +477,17 @@ class Comment < ApplicationRecord
   end
 
   def check_for_spam?
-    #don't check for spam while running tests or if the comment is 'signed'
-    self.approved = Rails.env.test? || !self.pseud_id.nil? || !Akismetor.spam?(akismet_attributes)
+    if !(['staging', 'production'].include?(Rails.env))
+      # don't check for spam while in a dev or test environment
+      self.approved = true
+    elsif (!self.pseud_id.nil? && !self.pseud.user.should_spam_check_comments?)
+      # or if the comment is 'signed' by an account over a certain age
+      self.approved = true
+    else
+      self.approved = !Akismetor.spam?(akismet_attributes)
+    end
+
+    self.approved
   end
 
   def mark_as_spam!
--- a/app/views/admin/settings/index.html.erb
+++ b/app/views/admin/settings/index.html.erb
@@ -64,12 +64,23 @@
 
       <dt><%= f.label :cache_expiration, t(".fields.cache_expiration") %></dt>
       <dd><%= admin_setting_text_field(f, :cache_expiration, size: "3") %></dd>
+    </dl>
+  </fieldset>
 
+  <fieldset>
+    <legend><%= t(".legend.content") %></legend>
+    <dl>
       <dt><%= admin_setting_checkbox(f, :hide_spam) %></dt>
       <dd><%= f.label :hide_spam, t(".fields.hide_spam") %></dd>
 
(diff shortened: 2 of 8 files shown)


app/models/comment.rb
```diff
@@ -65,7 +65,7 @@ def by_anonymous_creator?
   validate :check_for_spam, on: :create
 
   def check_for_spam
-    errors.add(:base, ts("This comment looks like spam to our system, sorry! Please try again, or create an account to comment.")) unless check_for_spam?
+    errors.add(:base, ts("This comment looks like spam to our system, sorry! Please try again.")) unless check_for_spam?
```
EchoEkhi, line 68: I think I'll leave the i18n to a later ticket

app/models/comment.rb
```diff
@@ -466,8 +477,15 @@ def count_visible_comments
   end
 
   def check_for_spam?
-    #don't check for spam while running tests or if the comment is 'signed'
-    self.approved = Rails.env.test? || !self.pseud_id.nil? || !Akismetor.spam?(akismet_attributes)
+    self.approved = if %w[staging production].exclude?(Rails.env)
+                      # don't check for spam while in a dev or test environment
+                      true
+                    elsif !self.pseud_id.nil? && !self.pseud.user.should_spam_check_comments?
```
EchoEkhi, line 483: Rubocop is trying to make me convert the whole thing to a one-liner, but I'm not convinced that's good for readability

app/models/comment.rb
```diff
@@ -466,20 +477,31 @@ def count_visible_comments
...
+    is_logged_in = !self.pseud_id.nil? 
+    is_account_old_enough = is_logged_in && !self.pseud.user.should_spam_check_comments?
+    is_on_tag = self.on_tag?
+    
+    should_skip_checking = is_logged_in && (is_account_old_enough || is_on_tag)
+
+    self.approved = should_skip_checking || !Akismet.spam?(akismet_attributes)
+  end
+
+  def submit_spam
```
Bilka2, line 489: Would it be doable to override these methods in the tests instead of the ones on Akismet? I'm a bit worried about adding an extra abstraction layer (the Akismet module) just to be able to stub methods in tests .

app/views/admin/settings/index.html.erb
```diff
@@ -64,12 +64,23 @@
...
       <dt><%= admin_setting_checkbox(f, :hide_spam) %></dt>
       <dd><%= f.label :hide_spam, t(".fields.hide_spam") %></dd>
 
       <dt><%= admin_setting_checkbox(f, :guest_comments_off) %></dt>
       <dd><%= f.label :guest_comments_off, t(".fields.guest_comments_off") %></dd>
+
+      <dt><%= f.label :account_age_threshold_for_comment_spam_check, t(".fields.account_age_threshold_for_comment_spam_check") %></dt>
+      <dd>
+        <%= admin_setting_text_field(f, :account_age_threshold_for_comment_spam_check, size: "3", "aria-labelledby" => "account_age_threshold_for_comment_spam_check_note") %>
+        <p id="account_age_threshold_for_comment_spam_check_note" role="label" class="footnote"><%= t(".fields.account_age_threshold_for_comment_spam_check_note") %></p>
```
sarken, line 82: This should use `aria-describedby` because it's intended to add information rather than override the label. ["aria-label, aria-labelledby, and aria-describedby: What's the Difference?"](https://benmyers.dev/blog/aria-labels-and-descriptions/) gives a good overview of when each attribute should be used.

We also use `kebab-case` rather than `snake_case` when naming these. (The main exception is the site header, which should probably be updated to match the others at some point.)

If you'd like to see some examples of how we usually do it, you can check out [these search results for "aria-describedby"](https://github.com/search?q=repo%3Aotwcode%2Fotwarchive%20aria-describedby&type=code).

app/views/admin/settings/index.html.erb
```diff
@@ -64,12 +64,23 @@
...
       <dt><%= admin_setting_checkbox(f, :hide_spam) %></dt>
       <dd><%= f.label :hide_spam, t(".fields.hide_spam") %></dd>
 
       <dt><%= admin_setting_checkbox(f, :guest_comments_off) %></dt>
       <dd><%= f.label :guest_comments_off, t(".fields.guest_comments_off") %></dd>
+
+      <dt><%= f.label :account_age_threshold_for_comment_spam_check, t(".fields.account_age_threshold_for_comment_spam_check") %></dt>
+      <dd>
+        <%= admin_setting_text_field(f, :account_age_threshold_for_comment_spam_check, size: "3", "aria-describedby": "account-age-threshold-for-comment-spam-check-note") %>
+        <p id="account-age-threshold-for-comment-spam-check-note" role="label" class="footnote"><%= t(".fields.account_age_threshold_for_comment_spam_check_note") %></p>
```
sarken, line 82: Could you remove `role="label"` since this isn't acting as a label? (I'm actually not sure label is one of the [allowed values for the role attribute](https://www.w3.org/TR/wai-aria-1.2/#role_definitions). It sounds like something that should be, but it's not on that list or any others I can find.)

sarken (commented):
The removal of the `role` attribute looks good, thank you! I think that just leaves Bilka's comment.

features/step_definitions/user_steps.rb
```diff
@@ -111,6 +111,14 @@
   step(%{I am logged in as "#{login}" with password "#{DEFAULT_PASSWORD}"})
 end
 
+Given "I am logged in as a new user {string}" do |login|
+  step(%{I am logged in as "#{login}"})
+  user = User.find_by(login: login)
+  user.created_at = Time.now.utc
+  user.confirmed_at = Time.now.utc
+  user.save
```
Bilka2, line 119:
```suggestion
  user.save!
```
It'll throw if something is invalid.

Bilka2 (changes requested):
Thank you for working on this! In addition to addressing my comments, please also update the `PUT #update` test in `settings_controller_spec.rb` for the new setting.

As a sidenote, you can use `comment.user` instead of `comment.pseud.user` if you prefer; `user` is delegated to `pseud`.

Bilka2 (approved):
Thank you!

brianjaustin:
Looks like the setting got cached with the wrong (old) value on staging. I will make a note that the migration needs to be applied _before_ the production deploy to prevent this

PR #5100, opened 2025-03-14
AO3-6942 Update page title in admin for spam ban deletion 

laurasimmons30:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6942

## Purpose

Updates the page title for `confirm_delete_user_creation` page in admin view to be more dynamic and consistent with other similar page titles

## Credit

Laura (she/her)

### DIFF
--- a/config/locales/views/en.yml
+++ b/config/locales/views/en.yml
@@ -148,7 +148,6 @@ en:
         caution_html: Are you sure you want to <strong><em>delete</em></strong> all the works and comments created by this user, along with their %{bookmarks} bookmarks, %{series} series, and %{collections} collections? This <strong>cannot be undone</strong>.
         confirm: Are you sure? Remember this will destroy ALL these objects!
         page_heading: Delete Spammer Creations
-        page_title: "%{login} - Confirm Deletion of User Creations"
         submit: Yes, Delete All Spammer Creations
       creations:
         navigation:
--- a/config/locales/views/en.yml
+++ b/config/locales/views/en.yml
@@ -148,6 +148,7 @@ en:
         caution_html: Are you sure you want to <strong><em>delete</em></strong> all the works and comments created by this user, along with their %{bookmarks} bookmarks, %{series} series, and %{collections} collections? This <strong>cannot be undone</strong>.
         confirm: Are you sure? Remember this will destroy ALL these objects!
         page_heading: Delete Spammer Creations
+        page_title: "%{login} - Confirm Deletion of User Creations"
         submit: Yes, Delete All Spammer Creations
       creations:
         navigation:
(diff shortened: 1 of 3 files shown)


EchoEkhi:
You might want to pick a shorter title for your pull request, because when the commits get squashed during merge, that becomes the new commit message

features/search/filters.feature
```diff
@@ -12,6 +12,7 @@ Feature: Filters
       And I post the work "A Hobbit's Meandering" with fandom "The Hobbit"
       And I post the work "Bilbo Does the Thing" with fandom "The Hobbit, Legend of Korra"
       And I post the work "Roonal Woozlib and the Ferrets of Nimh" with fandom "Harry Potter"
+      And it is currently 1 second from now
```
laurasimmons30, line 15: There seems to be some flakyness around the works count and bookmarks count links getting populated with the correct number of created works before they're clicked in some tests. I'm not sure this totally fixed it, but worth flagging.

zz9pzza: Have you looked at https://github.com/otwcode/otwarchive/blob/master/features/step_definitions/search_steps.rb#L1C6-L1C37

all indexing jobs have been run ?

Bilka2: I pulled this text fix out into a separate Jira issue and pull request: #5101

config/locales/views/en.yml
```diff
@@ -148,6 +148,7 @@ en:
         caution_html: Are you sure you want to <strong><em>delete</em></strong> all the works and comments created by this user, along with their %{bookmarks} bookmarks, %{series} series, and %{collections} collections? This <strong>cannot be undone</strong>.
         confirm: Are you sure? Remember this will destroy ALL these objects!
         page_heading: Delete Spammer Creations
+        page_title: "%{login} - Confirm Deletion of User Creations"
```
Bilka2, line 151: Please move this to `config/locales/controllers/en.yml` instead.

Bilka2 (changes requested):
Thank you for contributing! This just needs one tiny change.

Please remove the fix for the flaky test because I pulled it out into its own Jira issue and pull request to avoid expanding the scope of this PR. But thank you for getting started on that!

config/locales/views/en.yml
laurasimmons30: Is that correct? This change is for the text in the browser tab title to be updated, not a controller action message (to my understanding).

example of a different page per the ticket: https://github.com/otwcode/otwarchive/blob/451f0b80bef64d96831dcf8742b37a365954ad3d/config/locales/views/en.yml#L159

Bilka2: Yes, because it's getting set in the controller, not in the view. A few of the locales for the page titles ended up in the wrong file, we have https://otwarchive.atlassian.net/browse/AO3-6924 to fix some of them.

laurasimmons30: Ok sounds good! Thanks for the clarification, i will make that update 👍

Bilka2 (approved):

config/locales/views/en.yml
Bilka2: Thank you!

PR #4964, opened 2024-11-19
AO3-6840 Add TOS FAQ to site map

lydia-theda:
## Issue

[https://otwarchive.atlassian.net/browse/AO3-6840](https://otwarchive.atlassian.net/browse/AO3-6840) 

## Purpose

Adds entry for TOS FAQ to Site Map > About the Archive of Our Own

## Credit

lydia-theda

PR #4963, opened 2024-11-19
AO3-6841 Further clarify the definition of the Underage Sex warning

lydia-theda:
## Issue

[https://otwarchive.atlassian.net/browse/AO3-6841](https://otwarchive.atlassian.net/browse/AO3-6841)

## Purpose

Updates the explanation of the Underage Sex warning from "sexual activity **by** characters under the age of eighteen" to "sexual activity **involving** characters under the age of eighteen".

## Credit

lydia-theda

PR #5105, opened 2025-03-20
AO3-6028 Automatically email inactive tag wranglers and alert wrangling supervisors of wrangler inactivity

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6028

## Purpose

Adds automatic sending of emails to inactive wranglers and then if they stay inactive also alerts wrangling supervisors.

## Testing Instructions

Pending.

## References

PR Limit: 8 out of 5
Using reviewer bonus:
https://github.com/otwcode/otwarchive/pull/4963
https://github.com/otwcode/otwarchive/pull/4964

## Credit

Bilka


app/models/user.rb
```diff
@@ -439,7 +439,10 @@ def update_last_wrangling_activity
     return unless is_tag_wrangler?
 
     last_activity = LastWranglingActivity.find_or_create_by(user: self)
+    last_activity.notified_inactive_wrangler = false
+    last_activity.notified_inactive_supervisors = false
     last_activity.touch
```
EchoEkhi, line 444: Is `touch` still necessary here?

spec/models/concerns/wrangleable_spec.rb
```diff
@@ -10,23 +10,30 @@
...
+      end
 
       context "a wrangling activity has happened" do
         before { User.should_update_wrangling_activity = true }
 
         it "sets a last wrangling time" do
           freeze_time do
             expect(wrangleable.save).to be_truthy
-            expect(User.current_user.last_wrangling_activity.updated_at).to eq(Time.now.utc)
+            expect(User.current_user.reload.last_wrangling_activity.updated_at).to eq(Time.now.utc)
```
EchoEkhi, line 25: Should this be `Time.current`?

app/models/user.rb
Bilka2: Yes, without it if the two attributes are already false Rails will skip setting saving the record, including the new `updated_at`. This solution feels inelegant though, because it results in two database queries (the touch and the save!), but I couldn't find anything better.

spec/models/concerns/wrangleable_spec.rb
Bilka2: Yes. I'll batch it with other review changes if there are any.

config/config.yml
```diff
@@ -259,6 +259,22 @@ FREEFORM_CATEGORY_NAME: 'Additional Tags'
...
+ASSIGNED_FANDOMS_POLICY_URL: "https://example.com/Assigned_Fandoms_Policy_placeholder"
+USERS_EXCLUDED_FROM_WRANGLING_INACTIVITY:
+- AO3_Wrangling_Project
+- zz9pzza
+- Qem_Chibati
+- AO3_Support
+- AO3_Tag_Wranglers
+- testy_wrangler
+- elz
+- astolat
```
sarken, line 276: I talked to TW and we have some changes here! Please remove the last two names and add TWSupervisor_Fandom_Hold. I'll update the list on the issue. 

(Although now that I think about it, @brianjaustin and/or @zz9pzza, should we leave this blank and just have Systems add the names in when this is deployed? These names won't be relevant to other instances.)

brianjaustin: Yes, let's configure this via Ansible instead of here

config/locales/mailers/en.yml
```diff
@@ -81,6 +81,7 @@ en:
         open_doors: The Open Doors team
         parent_org: Organization for Transformative Works
         support: The AO3 Support team
+        tag_wrangling_supervisors: Tag Wrangling Supervisors
```
lydia-theda, line 84: I'm not sure if this is the right place to raise the issue but:

line 76 (and a few other places, like the config name?) calls them "Tag Wrangler Supervisors" (ER ending); however, since the committee name is Tag Wrangling and the official role name is Tag Wrangling Supervisor, I'm pretty sure how you have it here (ING ending) is more correct. Though I don't know if that only applies to the text portion or if the tag_wrangler_supervisors here and/or in the config address can/should also be updated.

In any case, it's probably a little weird for the greeting and signature to be different, even if they're never used in the same email?

Bilka2: The greeting is only used in emails to the supervisors themselves so it doesn't really matter, and the config option would be annoying to change. For anything new in this PR, I took care to only use the -ing ending when talking about the role.

config/locales/mailers/en.yml
```diff
@@ -321,6 +330,38 @@ en:
...
+        one: This is an automated email from %{app_name}. You are receiving this email because the Archive hasn't detected any wrangling from you in the past %{weeks} week.
+        other: This is an automated email from %{app_name}. You are receiving this email because the Archive hasn't detected any wrangling from you in the past %{weeks} weeks.
+      must_wrangle:
+        one: Wrangling policies require that everyone wrangle at least once every week.
+        other: Wrangling policies require that everyone wrangle at least once every %{weeks} weeks.
+      please_reply:
+        another_issue: You have another issue.
+        mentor: You would like a mentor or have questions.
+        more_work:
+          one: You need more work in order to have tags to wrangle every %{weeks} week.
```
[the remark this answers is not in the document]
Bilka2: In English yes. In other locales, `one` is used for plurals besides `1`, such as numbers ending with 1. So if translators copy how English does it, it might go wrong, but if they are aware of the plural rules and always use the count then it's fine.

I'll just change this.

brianjaustin (approved):

brianjaustin (changes requested):
TW has provided us with updated text for the guidelines around this policy, which I have sent to you elsewhere. Could you update them as part of this so we don't have to time two PRs around each other?

Bilka2:

Bilka2:
TODO:
* [x] (merge latest master)
* [x] update migration timestamp (and update Jira with it)
* [x] update email text for the email to wranglers based on new text from TW received 16 November (note the greeting and signature change!)
* [x] update default times (2/3 weeks -> 3/4 weeks)
* [x] inactive_wrangler_notification should go to chairs, not supervisors. figure out how to name the mailer class after this change because supervisor is no longer correct (tagwranglingcommittee?). add a config variable for the email and document it in the deploy notes on Jira
* [x] setting the last wrangled times to 0 should turn off the emails, so we can turn them off on staging

~Note to self: AsyncWithActiveJob~

Bilka2:
The required code updates are done.

PR #5115, opened 2025-03-26
AO3-6952 Skip submitting deleted comments to spam checker when spam-banning

EchoEkhi:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6952

## Purpose

Skip submitting marked-as-deleted comments to spam checker when banning a user for spam

## Credit

EchoEkhi

### DIFF
--- a/spec/controllers/admin/admin_users_controller_spec.rb
+++ b/spec/controllers/admin/admin_users_controller_spec.rb
@@ -483,6 +483,7 @@ describe Admin::AdminUsersController do
     let!(:collection1) { create(:collection) }
     let!(:collection2) { create(:collection) }
     let!(:comment) { create(:comment, pseud: user.default_pseud) }
+    let!(:deleted_comment) { create(:comment, pseud: user.default_pseud, is_deleted: true) }
 
     authorized_roles = %w[superadmin policy_and_abuse].freeze
 
@@ -536,6 +537,12 @@ describe Admin::AdminUsersController do
 
             it "sends all of the user's comments as spam reports to Akismet" do
               expect_any_instance_of(Comment).to receive(:submit_spam)
+
+              subject.call
+            end
+
+            it "does not send deleted comments to Akismet" do
+              expect(deleted_comment).to_not receive(:submit_spam)
 
               subject.call
             end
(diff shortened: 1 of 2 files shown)


spec/controllers/admin/admin_users_controller_spec.rb
```diff
@@ -539,6 +540,12 @@ def expect_changes_made_by(admin, log_items)
 
               subject.call
             end
+
+            it "does not send deleted comments to Akismet" do
+              expect(deleted_comment).to_not receive(:submit_spam)
```
Bilka2, line 545: This passes even when the change is reverted. Something like this works better:
```suggestion
              comment.is_deleted = true
              comment.save!(validate: false)
              expect_any_instance_of(Comment).to_not receive(:submit_spam)
```

Bilka2 (changes requested):

Bilka2:
Since the `is_deleted` comments aren't getting `destroy_or_mark_deleted` called on them now, could you please add a test that they're deleted anyway (which should happen via [check_can_destroy_parent](https://github.com/otwcode/otwarchive/blob/e383d3baf20608c208fc11cc6ad49df7f58f9e7e/lib/acts_as_commentable/comment_methods.rb#L133-L148))? A realistic way to test that is probably to give the current comment a reply and mark it deleted via the method instead of setting the status directly.

spec/controllers/admin/admin_users_controller_spec.rb
```diff
@@ -547,6 +547,17 @@ def expect_changes_made_by(admin, log_items)
...
+            it "deletes marked-as-deleted comments when all its replies are deleted" do
+              reply = create(:comment, commentable: comment, pseud: user.default_pseud)
+              comment.is_deleted = true
+              comment.save!(validate: false)
+
+              subject.call
+
+              expect(Comment.exists?(comment.id)).to be_falsey
+              expect(Comment.exists?(reply.id)).to be_falsey
+            end
```
EchoEkhi, line 560: Do you mean something like this?

Bilka2: Yep! I think it would be a bit cleaner to call `destroy_or_mark_deleted` on `comment` instead of manually doing the setting and saving (and check that the comment is marked as deleted as expected).

Bilka2 (approved):

PR #5117, opened 2025-03-26
AO3-6750 Add chapter number to inbox comments

EchoEkhi:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6750

## Purpose

Add chapter number to inbox comments.

## Credit

EchoEkhi

### DIFF
--- a/app/helpers/comments_helper.rb
+++ b/app/helpers/comments_helper.rb
@@ -65,6 +65,10 @@ module CommentsHelper
     end
   end
 
+  def chapter_description_link(comment)
+    link_to ts('Chapter %{position}', position: comment.parent.position), work_chapter_path(comment.parent.work, comment.parent)
+  end
+
   def image_safety_mode_cache_key(comment)
     "image-safety-mode" if comment.use_image_safety_mode?
   end
--- a/app/views/comments/_single_comment.html.erb
+++ b/app/views/comments/_single_comment.html.erb
@@ -24,7 +24,8 @@
           <% end %>
           <% if single_comment.parent.is_a?(Chapter) && single_comment.ultimate_parent.expected_number_of_chapters != 1 %>
             <span class='parent'>
-              <%= ts('on Chapter %{position}', :position => single_comment.parent.position) %>
+              <%= ts('on') %>
+              <%= chapter_description_link(single_comment) %>
             </span>
           <% end %>
           <% if single_comment.unreviewed? %>
--- a/app/views/inbox/_inbox_comment_contents.html.erb
+++ b/app/views/inbox/_inbox_comment_contents.html.erb
@@ -4,6 +4,12 @@
   <% else %>
     <%= get_commenter_pseud_or_name(feedback_comment) %> <%= ts('on') %> <%= commentable_description_link(feedback_comment) %>
   <% end %>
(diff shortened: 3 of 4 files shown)


weeklies:
Not staff but I wonder if the verbiage could be improved? 
Screenshot 2025-03-29 203138

For example:
Screenshot 2025-03-29 203229

sarken:
@weeklies is right! It should match the wording in [AO3-3071](https://otwarchive.atlassian.net/issues/AO3-3071), which is "Chapter # of Work Title." (As with email subject lines, there's not enough space to include the full chapter title on the inbox subject lines.)  I'm not sure why the issue didn't provide the desired phrasing; I'll update it now.

sarken:
Actually, hang on, one more change: in the inbox, "Chapter # of Work Title" should be one single link that points to the comment. It doesn't make sense for the "Chapter #" to link to the chapter while the work title links to the comment; it just makes it so you have a 50% chance of clicking the wrong link when you want to access the comment. 

If we want "Chapter #" to link to the chapter, it would be better to make "Chapter #" a link to the chapter, the work title a link to the work, and add some other way to access the comment directly (e.g., a "Thread" link by "Reply," similar to what comments themselves have). 

However, that's a bit out of scope here, so let's just update the inbox's link text and add a link to the single comment's chapter info.

app/helpers/inbox_helper.rb
```diff
@@ -8,6 +8,8 @@ def commentable_description_link(comment)
       link_to commentable.name, tag_comment_path(commentable, comment)
     elsif commentable.is_a?(AdminPost)
       link_to commentable.title, admin_post_comment_path(commentable, comment)
+    elsif commentable.expected_number_of_chapters != 1
```
Bilka2, line 11: Please use `chaptered?` instead

app/views/comments/show.html.erb
```diff
@@ -1,5 +1,8 @@
 <%= flash_div :comment_error, :comment_notice %>
 
-<h3 class="heading"><%=h t('.comment_on', :default => "Comment on") %>  <%= link_to_comment_ultimate_parent(@comment) %></h3>
+<h3 class="heading">
+  <%= t(".comment_on") %>
+  <%= link_to_comment_ultimate_parent(@comment) %>
+</h3>
```
Bilka2, line 6: From how I understand the Jira issue, this page heading should not be changed. The change you had to add the chapter link in `_single_comment.html.erb` was in the right place.

Bilka2: Also, please add a test for this change.

Bilka2 (changes requested):

app/views/comments/show.html.erb
```diff
@@ -1,7 +1,7 @@
 <%= flash_div :comment_error, :comment_notice %>
 
 <h3 class="heading">
-  <%= t(".comment_on", default: "Comment on") %>
+  <%= t(".comment_on") %>
   <%= link_to_comment_ultimate_parent(@comment) %>
```
Hamham6, line 5: I think the link should be interpolated in the .comment_on so it can be put before the text.

Hamham6 (changes requested):

Bilka2 (approved):
Thank you!

PR #5118, opened 2025-03-27
AO3-6439 Fix pseud autocomplete using data from previous users

Bilka2:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6439

## Purpose

Fix pseud autocomplete using data from previous users when searching for a space. **I could not reproduce the issue in tests or locally, so this is a blind fix based on how the `site_skins` method works!**

This also fixes that the autocomplete would load forever if you searched for a space, which is also an issue on production.

## Testing Instructions

Log in as UserA in one browser, search for a space in the people search. Log in as UserB in a different browser/incognito window. Search for a space in the people search. You should get the results for UserB, not UserA.

Log out. Search for a space in the people search. You should get "(No suggestions found)".

## Credit

Bilka


app/controllers/autocomplete_controller.rb
```diff
@@ -21,9 +21,9 @@ class AutocompleteController < ApplicationController
 
   # PSEUDS
   def pseud
-    if params[:term].blank?
+    if params[:term].blank? && logged_in?
       # get the user's own pseuds
-      set_current_user { render_output(current_user.pseuds.collect(&:byline)) }
```
Bilka2, line 26: The `set_current_user` is used to set `User.current_user` so that models can access the current user via that. This is global state that ended up shared when it shouldn't be.

From what I can tell in the tests (when running the set_current_user like this or in the around_action), the autocomplete_controller would mess with User.current_user while the works_controller was in the middle of doing things. So then when this action here finished, User.current_user wasn't set in the works_controller anymore. This controller here doesn't do anything that uses User.current_user so it doesn't need to be changed. It's fine to just use devises current_user directly.

brianjaustin (approved):

PR #5120, opened 2025-03-30
AO3-6940: Refine subnavigation buttons for Admin actions

lou-codes:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6940

## Purpose
This PR serves to streamline the visible navigation options for admin posts, wrangling guidelines, FAQs, and known issues. This includes limiting nav options to their own controller actions as well as adding nav options to extra action pages (ie new, manage, edit, etc)

## Testing Instructions

See JIRA for testing instructions, and a full list pages updated with the nav options they should have

## Credit

lou, she/her


app/views/admin/_admin_nav.html.erb
```diff
@@ -1,41 +1,56 @@
 <h3 class="landmark heading"><%= t(".landmark") %></h3>
-<ul class="navigation actions" role="navigation">
-  <li>
-    <%= span_if_current t(".ao3_news"), admin_posts_path %>
-  </li>
-  <% if policy(AdminPost).can_post? %>
+<ul class="navigation actions">
+  <% if params[:controller] == "admin_posts" %>
```
sarken, line 3: I think rather than checking which controller we're in, it would be a little cleaner and simpler to break these out into separate partials for each area, like we do in other parts of the code, so you'd have:
- admin_posts/_navigation.html.erb
- known_issues/_navigation.html.erb
- wrangling_guidelines/_navigation.html.erb
- archive_faqs/_navigation.html.erb

app/views/archive_faqs/_admin_index.html.erb
```diff
@@ -5,7 +5,7 @@
 
 <!--subnav-->
 <%= render "archive_faqs/filters" %>
-<%= render "admin/admin_nav" %>
+<%= render :partial => 'admin/admin_nav' %>
```
sarken, line 8: Is there a reason `partial` is necessary here now? `<%= render "admin/admin_nav" %>` _should_ still work fine.

app/views/archive_faqs/manage.html.erb
```diff
@@ -4,6 +4,12 @@
 
 <!--subnav-->
 <%= render :partial => 'admin/admin_nav' %>
+<ul class="navigation actions">
+  <% if Globalize.locale.to_s == "en" && policy(ArchiveFaq).full_access? %>
+    <li><%= link_to t(".new_faq_category"), new_archive_faq_path %></li>
+    <li><%= link_to t(".reorder_faqs"), manage_archive_faqs_path(language_id: params[:language_id]) %></li>
+  <% end %>
+</ul>
```
sarken, line 12: All of the navigation items should be in the same `<ul>`. (The existing code in the wrangling guidelines admin index that puts them in separate lists is wrong.)

Also, as a general rule, we don't want to include empty elements on a page. So logic like `<% if Globalize.locale.to_s == "en" && policy(ArchiveFaq).full_access? %>` would go outside the `<ul>` in this code, since there are no `<li>` elements that would be included if that logic were `false`.

sarken (commented):
Thanks for working on this! I realize it's still a draft, but I've left a few comments that will hopefully help. If you were already planning on making the changes I've described, my apologies for telling you what you already know!

app/views/admin/_admin_nav.html.erb
lou-codes: Agreed- I'll get to work on the cleanup!

In the meantime, is it a hindrance to have draft PRs open in the repo? I'm used to having one open while I'm working on an issue but I can 100% keep my work in my own fork until its ready if that makes the review process easier!

sarken: It depends! If you're stuck and need help or want to double check your approach before you get in too deep, we're more than happy to have draft PRs (especially if you tell us what parts you'd like us to take a look at). If you're just working and don't need anything from us, though, it's better to hold off on opening a PR until you're done and ready for review.

app/views/admin_posts/_navigation.html.erb
```diff
@@ -0,0 +1,15 @@
+<h3 class="landmark heading"><%= t(".landmark") %></h3>
+<% if policy(AdminPost).can_post? %>
```
sarken, line 2: Only admins who `can_post?` are able to access the pages where this navigation appears, so this check probably isn't necessary -- especially since we know the logic for this navigation will need reworking when we add news post drafts in [AO3-6960](https://otwarchive.atlassian.net/issues/AO3-6960).

app/views/archive_faqs/_navigation.html.erb
```diff
@@ -0,0 +1,22 @@
+<h3 class="landmark heading"><%= t(".landmark") %></h3>
+<ul class="navigation actions">
+  <% if policy(ArchiveFaq).translation_access? %>
+    <li>
+      <%= span_if_current t(".archive_faq"), archive_faqs_path %>
+    </li>
+  <% end %>
```
sarken, line 7: I'm pretty sure you can leave this check out. If you can see this navigation partial, you should be able to access archive_faqs_path.

features/admins/admin_post_news.feature
```diff
@@ -8,6 +8,10 @@ Feature: Admin Actions to Post News
     Given I am logged in as a "tag_wrangling" admin
     When I go to the admin-posts page
     Then I should not see "Post AO3 News"
+    Given I am logged in as a "communications" admin
+    When I go to the admin-posts page
+    Then I should see "AO3 News"
+      And I should see "Post AO3 News"
```
sarken, line 14: I think we can update "Scenario: Log in as an admin and create an admin post with tags" instead, like so:
```
    Given I am logged in as a "communications" admin
    When I follow "Admin Posts"
    Then I should see "AO3 News" within "#main ul.navigation"
      And I should see "Post AO3 News" within "#main ul.navigation"
    When I follow "Post AO3 News" within "#main ul.navigation"
    Then I should see "New AO3 News Post"
```

(As with the Known Issues test, I didn't check that the selectors I used here are correct; they might need some work. But using them is a good idea because the text we're looking for appears in multiple locations.)

features/admins/admin_post_issues.feature
```diff
@@ -14,6 +14,10 @@ Feature: Admin Actions to Post Known Issues
       And I press "Post"
     Then I should see "Known issue was successfully created"
       And I should see "First known problem"
+    When I follow "Admin Posts"
+      And I follow "Known Issues" within "#header"
```
sarken, line 18: I think you can skip `When I follow "Admin Posts"`: it's not expanding the admin posts menu, just taking you to /admin_posts and then following the "Known Issues" link in the header.

features/admins/admin_post_issues.feature
```diff
@@ -14,6 +14,10 @@ Feature: Admin Actions to Post Known Issues
...
+    Then I should see "Known Issues"
```
sarken, line 19: It would be a good idea to specify a selector here (something like `#main ul.navigation`, but you'll have to double check that it's specific enough), since the text we're looking for appears in multiple places on the page. If our "Known Issues" navigation link wasn't appearing, this test might still pass because the words "Known Issues" also appear in the admin header navigation, the page's h2, and the site footer.

features/tags_and_wrangling/wrangling_guidelines.feature
```diff
@@ -8,10 +8,13 @@ Feature: Wrangling Guidelines
 
   Given I am logged in as a "tag_wrangling" admin
     And I am on the wrangling guidelines page
-  And I follow "New Wrangling Guideline"
+  Then I should see "Wrangling Guidelines"
```
sarken, line 11: Could you make the line above this into "When I go to the wrangling guidelines page"? Having a "Then" right after a "Given" isn't the end of the world, but it is slightly nonstandard, and fixing it is an easy tweak.

I'd also mention it would be a good idea to add a selector when checking for the "Wrangling Guidelines" link to avoid any false test results due to the text appearing in multiple places on the page.

app/views/archive_faqs/_navigation.html.erb
```diff
@@ -0,0 +1,22 @@
...
+  <% if policy(ArchiveFaq).full_access? %>
+    <% if Globalize.locale.to_s == "en" %>
+      <li><%= link_to t(".new_faq_category"), new_archive_faq_path %></li>
+      <% if params[:action] == "edit" %>
+        <li>
+          <%= link_to t(".faq.reorder_questions"), manage_archive_faq_questions_path(@archive_faq) %>
```
sarken, line 13: I think it would be fine to just use `reorder_questions` as the key here.

sarken (changes requested):
Thank you for working on this! It's a lot neater and easier to understand than the old admin_nav partial.

app/views/archive_faqs/_navigation.html.erb
```diff
@@ -0,0 +1,22 @@
+<h3 class="landmark heading"><%= t(".landmark") %></h3>
+<ul class="navigation actions">
+  <% if policy(ArchiveFaq).translation_access? %>
+    <li>
+      <%= span_if_current t(".archive_faq"), archive_faqs_path %>
+    </li>
+  <% end %>
+  <% if policy(ArchiveFaq).full_access? %>
+    <% if Globalize.locale.to_s == "en" %>
```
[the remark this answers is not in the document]
Bilka2: > it looks like the New and Reorder options should only be available if the admin has full access _and_ the locale is English

That's correct per my PR that added these checks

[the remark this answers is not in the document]
Hamham6: Is it "en" for technical reasons or is it "en" because that's the default language for AO3? If it's because "en" is the default language, can ArchiveConfig be used instead?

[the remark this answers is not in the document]
sarken: I'm not sure if we'd want to use ArchiveConfig for that or something else, but there are a number of places in the FAQ where we've coded it to assume English is the default language. We'll need to review that at some point, but that's outside the scope of this issue.

PR #5122, opened 2025-03-30
AO3-6578 Move external work notice

IESNunes:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6578

## Purpose

The external work notice was overlapping the 4 symbols because there were no fandoms. By moving the notice, it will not be obscured. 

## Credit

Isabel Nunes (she/her)


sarken (changes requested):
Thank you--I gave the code a try and it looks much better! It also seems to fix an issue where there's a little too much empty space above the notice in certain circumstances:

A new external work blurb contrasted with an old one. The old blurb has more whitespace above the blue notice banner.

I think this means you can also delete a selector from the CSS: [.blurb .header p.notice](https://github.com/otwcode/otwarchive/blob/e383d3baf20608c208fc11cc6ad49df7f58f9e7e/public/stylesheets/site/2.0/13-group-blurb.css#L62) is what was giving the notice its extra top margin, and I'm pretty sure there aren't any other blurbs with notices in the header (but you might want to double check). 

We'd also appreciate it if you could address the style comment about the translation helper, but since it was an existing issue, we understand if you'd rather not.

config/locales/views/en.yml
```diff
@@ -2170,3 +2170,8 @@ en:
     show:
       edit: Edit
       heading: Wrangling Guidelines
+  external_works:
+    work_module:
+      notice: This work isn't hosted on the Archive so this blurb might not be complete or accurate.
+    blurb:
+      notice: This work isn't hosted on the Archive so this blurb might not be complete or accurate.
```
sarken, line 2177: Since the text is the same, we can create one translation:

```
external_works:
  notice:  This work isn't hosted on the Archive so this blurb might not be complete or accurate.
```

And then use it in both partials using basic lookup instead of lazy lookup: `t("external_works.notice")` 
  
You'll also need to run the [normalize i18n-task](https://github.com/otwcode/otwarchive/wiki/Internationalization-(i18n)-Standards#i18n-tasks) to get these keys sorted into the proper alphabetical order within the file -- that's what's causing the test failure.

sarken (changes requested):
Thanks for doing the i18n!

config/locales/views/en.yml
IESNunes: Thank you for the comment. It looks much better with only one translation :smile: I also ran the normalize i18n task, and the test ran successfully.

sarken (approved):
Thank you!

PR #5123, opened 2025-03-31
AO3-6961 Fix incorrect classes on caution text and update Reversi styling of caution text

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6961

## Purpose

Updates some HTML classes to ensure we have consistent styling of caution text, and updates Reversi so the caution text isn't the same color as regular text.

## Testing Instructions

Refer to Jira.

### DIFF
--- a/app/views/orphans/_orphan_user.html.erb
+++ b/app/views/orphans/_orphan_user.html.erb
@@ -6,7 +6,7 @@
 
 <p class="caution"><%= t(".orphaning_bylines_only_message_html") %></p>
 
-<p class="caution">
+<p class="caution notice">
   <%= ts("Orphaning a work removes it from your account and re-attaches it to the specially created orphan_account. Please note that this is")%>
   <strong><%= ts("permanent and irreversible.")%></strong>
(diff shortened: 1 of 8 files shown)


app/views/orphans/_orphan_user.html.erb
```diff
@@ -6,7 +6,7 @@
 
 <p class="caution"><%= t(".orphaning_bylines_only_message_html") %></p>
 
-<p class="caution">
+<p class="caution notice">
```
Bilka2, line 9: The two `<p>` tags above this one are missing the extra class

Bilka2 (changes requested):

app/views/orphans/_orphan_user.html.erb
sarken: Whoops, that would be the file I tried to wrap everything in a single `<div class="caution notice">` on... I apparently misjudged how much Cmd+Z-ing to do after that didn't work out 😆

Bilka2 (approved):

PR #5131, opened 2025-04-07
AO3-5919 Use intermediate set to avoid KEYS in kudos delivery

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5919

## Purpose

Remove the call to `REDIS_KUDOS.keys`, which scans the whole Redis database, in favour of using an intermediate set to keep track of which things an author has gotten kudos on.

### DIFF
--- a/app/models/redis_mail_queue.rb
+++ b/app/models/redis_mail_queue.rb
@@ -5,6 +5,7 @@ class RedisMailQueue
   def self.queue_kudo(author, kudo)
     key = "kudos_#{author.id}_#{kudo.commentable_type}_#{kudo.commentable_id}"
     REDIS_KUDOS.rpush(key, kudo.name)
+    REDIS_KUDOS.sadd("kudos_#{author.id}", key)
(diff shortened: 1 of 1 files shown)


sarken:
YAAAAAY

Bilka2:
I poked around moving this to ActiveJob earlier and wondered whether we want to use RedisSetJob for this and the subscription emails. Does this additional Redis set interfere with whether we can use RedisSetJob for kudos in the future?

brianjaustin:
@Bilka2 I think it should be fine, the tag stats job (which already uses `RedisSetJob`) contains a separate read from Redis (https://github.com/otwcode/otwarchive/blob/a0775cd551a588ee8e65bed69319a3b51ef069d6/app/jobs/tag_count_update_job.rb#L19). Plus, we would probably be scanning the ultimate parent set in this case, which is author IDs. Further processing would be up to the newly implemented job itself

zz9pzza (approved):

PR #5134, opened 2025-04-08
AO3-6031 Divide comments_controller_spec.rb into several smaller files

Hamham6:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6031

## Purpose

Refactor comments_controller_spec.rb into smaller files.

## Testing Instructions

Only touch automated tests, no changed functionality.
## References

The conversation included on the Jira issue.

## Credit

Hamham6, it/its


Hamham6:
Preliminary division by action, as I didn't want to get too in-depth without first checking for direction. I did not edit any of the tests themselves, only moved them around. I wasn't sure if I could put the show/hide_comments and cancel actions specs in the default_rails_action_spec or not, since they aren't actually default actions.

Hamham6:
I'm up to trying to refactor the tests by commentable/ultimate parent as suggested in the Jira issue, if that way still fits the desired path.

Hamham6:
I am on the Jira under the same username, Hamham6.

Bilka2:
Hi, Hamham6!

spec/controllers/comments/reply_spec.rb
```diff
@@ -0,0 +1,289 @@
+require "spec_helper"
+
+describe CommentsController do
+  include LoginMacros
+  include RedirectExpectationHelper
+
+  let(:comment) { create(:comment) }
+  let(:unreviewed_comment) { create(:comment, :unreviewed) }
```
Bilka2, line 8: I think these two are not needed

spec/controllers/comments/hide_spec.rb
```diff
@@ -0,0 +1,697 @@
+require "spec_helper"
+
+describe CommentsController do
+  include LoginMacros
+  include RedirectExpectationHelper
+
+  let(:comment) { create(:comment) }
+  let(:unreviewed_comment) { create(:comment, :unreviewed) }
```
Bilka2, line 8: I think these two aren't needed here

Bilka2 (changes requested):
I like the approach to divide by action and the divide you arrived at. Since that's also how we tend to add tests, because we add actions usually for all parents/commentables/users at once since we can't just skip defining the action behaviour for one of those.

`hidden_and_unrevealed_spec.rb` is by parent though :thinking: I think you could merge those back into `comments_controller_spec.rb`.

I agree with not changing the tests or correcting the rubocop violations. So this looks pretty good already!

Hamham6:
I now realize that the tests for hidden and unrevealed works are the only ones grouped by commentable in their actual implementation. Would it be desired to split those up such that each respective action includes tests for hidden and unrevealed works? If it is, could a new Issue be made for that?

spec/controllers/comments/hide_spec.rb
Hamham6: cancel_comment_reply uses that comment variable, so I left it in.

Hamham6: That response was supposed to be on the reply spec, whoops.

Bilka2:
I think it's fine that the hidden an unrevealed tests are arranged by commentable, since it allows reusing the examples. This controller is so complex that I don't think we can ever achieve a perfectly clean divide of the tests - but this is close enough. Thank you!

Bilka2 (approved):

PR #5135, opened 2025-04-08
AO3-6956 Corrected error message for duplicate comments

Nick-Raal:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6956 

## Purpose

This pull request changes the error message upon posting a duplicate comment from "This comment has already been left on this work. (It may not appear right away for performance reasons." to the more descriptive (and accurate) "You’ve already left this comment here. (It may not appear right away for performance reasons.)"

## Credit

niic (he/him)

### DIFF
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -71,7 +71,7 @@ class Comment < ApplicationRecord
   validates :comment_content, uniqueness: {
     scope: [:commentable_id, :commentable_type, :name, :email, :pseud_id],
     unless: :is_deleted?,
-    message: ts("^This comment has already been left on this work. (It may not appear right away for performance reasons.)")
+    message: ts("^You’ve already left this comment here. (It may not appear right away for performance reasons.")
   }
 
   scope :ordered_by_date, -> { order('created_at DESC') }
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -71,7 +71,7 @@ class Comment < ApplicationRecord
   validates :comment_content, uniqueness: {
     scope: [:commentable_id, :commentable_type, :name, :email, :pseud_id],
     unless: :is_deleted?,
-    message: ts("^You’ve already left this comment here. (It may not appear right away for performance reasons.")
+    message: :duplicate_comment
   }
 
   scope :ordered_by_date, -> { order('created_at DESC') }
--- a/config/locales/models/en.yml
+++ b/config/locales/models/en.yml
@@ -91,7 +91,7 @@ en:
         comment:
           attributes:
             comment_content:
-              duplicate_comment: ^You've already left this comment here. (It may not appear right away for performance reasons.)
+              duplicate_comment: "^You've already left this comment here. (It may not appear right away for performance reasons.)"
             commentable:
               format: "%{message}"
               guest_replies_off: Sorry, this user doesn't allow non-Archive users to reply to their comments.
--- a/config/locales/models/en.yml
+++ b/config/locales/models/en.yml
@@ -91,8 +91,7 @@ en:
         comment:
           attributes:
             comment_content:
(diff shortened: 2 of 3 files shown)


app/models/comment.rb
```diff
@@ -71,7 +71,7 @@ def check_for_spam
   validates :comment_content, uniqueness: {
     scope: [:commentable_id, :commentable_type, :name, :email, :pseud_id],
     unless: :is_deleted?,
-    message: ts("^This comment has already been left on this work. (It may not appear right away for performance reasons.)")
+    message: ts("^You’ve already left this comment here. (It may not appear right away for performance reasons.")
```
Bilka2, line 74: It would be nice to internationalize this, by [overriding the default error message key](https://github.com/otwcode/otwarchive/wiki/Internationalization-%28i18n%29-Standards#user-content-Overriding_a_default_error_message_key):

```suggestion
    message: :duplicate_comment
```
and then putting the translation, without the leading `^`, into [config/locales/models/en.yml](https://github.com/otwcode/otwarchive/blob/a0775cd551a588ee8e65bed69319a3b51ef069d6/config/locales/models/en.yml#L92)
```yml
  activerecord:
    errors:
      models:
        comment:
          attributes:
            comment_content:
              duplicate_comment: You've already ....
              format: "%{message}"
```
Could you also fix your typos? It should be a plain `'` in `You've` instead of a curly one and the closing parenthesis at the end should be there.

app/models/comment.rb
```diff
@@ -71,7 +71,7 @@ def check_for_spam
   validates :comment_content, uniqueness: {
```
Bilka2, line 71: You can ignore rubocop's `Rails/UniqueValidationWithoutIndex` complaint, we don't need the unique index for now.

Bilka2 (changes requested):
Hi, Nick!

Thank you so much for this pull request. I reviewed it with a suggestion for how to i18n the error message. Could you please also add a rspec test for the error message?

I've updated the Jira issue status to In Review, so no one will mistakenly create a duplicate pull request. If you'd like the ability to comment on, assign, and transition issues in the future, you're welcome to create a Jira account! You can just reply here with the account name and we'll set up the permissions for you. (It makes things a bit easier for us on the organizational side if the Full Name on your Jira account either closely matches the name you'd like us to credit in the release notes or includes it in parentheses, e.g. "Nickname (CREDIT NAME).")

Nick-Raal:
Still need to fix the test. Don't worry about this right now.

Nick-Raal:
I hard coded the test case in the spec, which is obviously not ideal. A similar method seems to have been done in the spec for kudos though, which is what I'm basing my decision to do the same in my spec on. I can't for the life of me get I18n to recognize the translation in the context of rspec, despite pouring over ample documentation. Either way, thank you so much for your suggestions and my username on Jira is Nick Raal.

config/locales/models/en.yml
```diff
@@ -90,6 +90,9 @@ en:
               required: can't be blank
         comment:
           attributes:
+            comment_content:
+              duplicate_comment: You've already left this comment here. (It may not appear right away for performance reasons.)
+              format: "%{message}"
```
Bilka2, line 95: This was also my initial thought. But it turns out we can't use this error format here because there are other error messages that need to have "Comment content" at the start. So, this should not override the general format and instead use the `^` at the start of the message itself.
```suggestion
              duplicate_comment: ^You've already left this comment here. (It may not appear right away for performance reasons.)
```

Bilka2 (changes requested):
The hardcoded string is what we always use in tests, that's fine! This way we know that the translation is working.

I've given your Jira account permission to comment on, assign, and transition issues in the future.

Bilka2 (approved):

PR #5137, opened 2025-04-09
AO3-6845 Always display invitation results in table format

michtnt:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6845

## Purpose

Display invitations regardless of whether the result returned is one or multiple. Before, when we were displaying one, it doesn't allow the Admin to delete the invitation as delete option only is shown on the table so Admin had to fuzzy search to pop up the table. It's better to keep the UX consistent as well this way.

**Note:** This is my first time learning Ruby and picking up on an improvement instead of a small fix, please correct me if there's something I miss. I'm trying to start moving to bigger task slowly, while learning Rails at the same time.

**Misc:**
- UI Improvements for Invitation page 

## Testing Instructions

**Note:** log in as Admin
**Search with username**
- Search with exact username, should return a user and invitations (if present)
- Search with username that doesn't exist, should not return a user

**Search with invite token**
- Search with exact token, should return an invitation (if present)
- Search with exact token that doesn't exist, should not return an invitation

**Search with email address**
- Search with exact email, should return invitations (if present)
- Search with partial email, should return invitations (if present)

**Combining searching with username and invite token**
if user is found, display the `Invitations for [user]` and if token is present, the table should display the invitation that match that token

## Credit

Michelle Tanoto


app/views/admin/admin_invitations/find.html.erb
```diff
@@ -19,11 +19,9 @@
 <% end %>
 
 <% if @user %>
-<h3 class="heading"><%h= 'Invitations for' %> <%= link_to @user.login, user_invitations_path(@user) %></h3>
-<%= render :partial => 'invitations/user_invitations', :locals => {:invitations => @invitations} %>
-<% elsif @invitation %>
-<%= render :partial => 'invitations/invitation', :locals => {:invitation => @invitation} %>
-<% elsif @invitations %>
+<h3 class="heading"><%= 'Invitations for' %> <%= link_to @user.login, user_invitations_path(@user) %></h3>
```
michtnt, line 22: typo `<%=h`, showing properly now:
Screenshot 2025-04-09 at 9 43 09 pm

config/locales/views/en.yml
```diff
@@ -1580,7 +1580,7 @@ en:
         what_you_cant_do: What you can't do
   invitations:
     invitation:
-      copy_and_use: copy and use
+      copy_and_use: Copy and use
```
michtnt, line 1583: I just capitalise it for consistency because we have `Copy and use` and `copy and use`, altho I'm tempted to reuse it instead of having duplicate
Screenshot 2025-04-09 at 10 31 45 pm

features/other_a/invite_queue.feature
```diff
@@ -106,8 +106,7 @@ Feature: Invite queue management
       And I follow "Invitations"
       And I fill in "track_invitation_invitee_email" with "test@archiveofourown.org"
       And I press "Search" within "form.invitation.simple.search"
-    Then I should see "Sender queue"
-    When I follow "copy and use"
+    When I follow "Copy and use"
```
brianjaustin, line 109: For steps with the same type (i.e. `When`), we use `And` and indent them by two spaces:
```suggestion
      And I follow "Copy and use"
```
You can see an example of this in the list of steps above

features/admins/admin_invitations.feature
```diff
@@ -309,10 +309,12 @@ Feature: Admin Actions to Manage Invitations
       And I fill in "Username" with "dax"
       And I press "Search" within "form.invitation.simple.search"
     Then I should see "Copy and use"
+    Then I should see "Delete"
```
brianjaustin, line 312: Similar to the other feature test, please update the repeated step types to use `And`, for example
```suggestion
      And I should see "Delete"
```

features/admins/admin_invitations.feature
```diff
@@ -363,6 +365,9 @@ Feature: Admin Actions to Manage Invitations
     When I press "Search" within "form.invitation.simple.search"
       And I fill in "All or part of an email address" with "test@example.com"
       And I press "Search" within "form.invitation.simple.search"
+    Then I should see "Copy and use"
+    Then I should see "Delete"
```
brianjaustin, line 369:
```suggestion
    Then I should see "Copy and use"
      And I should see "Delete"
```

features/admins/admin_invitations.feature
```diff
@@ -373,13 +378,18 @@ Feature: Admin Actions to Manage Invitations
       And I fill in "Username" with "dax"
       And I press "Search" within "form.invitation.simple.search"
     Then I should see "Copy and use"
+    Then I should see "Delete"
```
brianjaustin, line 381:
```suggestion
      And I should see "Delete"
```

features/admins/admin_invitations.feature
```diff
@@ -373,13 +378,18 @@ Feature: Admin Actions to Manage Invitations
...
     When I follow "Invite New Users"
       And I fill in "Invite token" with "dax's" invite code
       And I press "Search" within "form.invitation.simple.search"
-    Then I should see "copy and use"
+    Then I should see "Copy and use"
+    Then I should see "Delete"
```
brianjaustin, line 386:
```suggestion
      And I should see "Delete"
```

features/admins/admin_invitations.feature
```diff
@@ -373,13 +378,18 @@ Feature: Admin Actions to Manage Invitations
...
+    When I find and press the invitation token url
+    Then I should see "Redeemed at -"
     When I fill in "Enter an email address" with "oldman@ds9.com"
       And I press "Update Invitation"
-    Then I should see "oldman@ds9.com" in the "Enter an email address" input
+    Then I should see "oldman@ds9.com"
+    Then I should see "Invitation was successfully sent."
```
brianjaustin, line 392:
```suggestion
    Then I should see "oldman@ds9.com"
      And I should see "Invitation was successfully sent."
```

features/admins/admin_invitations.feature
```diff
@@ -309,10 +309,12 @@ Feature: Admin Actions to Manage Invitations
       And I fill in "Username" with "dax"
       And I press "Search" within "form.invitation.simple.search"
     Then I should see "Copy and use"
+    Then I should see "Delete"
     When I follow "Invite New Users"
       And I fill in "Invite token" with "dax's" invite code
       And I press "Search" within "form.invitation.simple.search"
-    Then I should see "copy and use"
+    When I find and press the invitation token url
+    Then I should see "Please log out of your admin account first!"
```
brianjaustin, line 317: With my proposed change to the step below, this won't work anymore. I'm not 100% sure we need these two lines in the test, but if we do keep them, it would require a different step definition

brianjaustin (changes requested):

features/step_definitions/admin_steps.rb
```diff
@@ -360,6 +360,10 @@
   expect(page.accept_alert).to eq("Are you sure you want to remove the creator's pseud from this work?") if @javascript
 end
 
+When("I find and press the invitation token url") do
```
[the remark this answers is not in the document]
michtnt: sorry! missed this thanks🙏🏼

features/admins/admin_invitations.feature
michtnt: Hey Brian, I'm having trouble understanding what you mean by this won't work, can you explain it to me? Thanks!

app/controllers/admin/admin_invitations_controller.rb
```diff
@@ -42,13 +42,12 @@ def find
       @invitations = @user.invitations if @user
     end
     if !invitation_params[:token].blank?
-      @invitation = Invitation.find_by(token: invitation_params[:token])
+      @invitations = Invitation.where(token: invitation_params[:token]).limit(1)
```
brianjaustin, line 45: Actually, now that we always treat this as a list, I don't think we even need the `.limit(1)` -- just the `where` should do it

features/admins/admin_invitations.feature
brianjaustin: Ah, I didn't realise that the page under this link is restricted to particular roles (as well as the other link that this would have found before the change to `admin_steps.rb` to make sure the invite link is the one inside the table). Could you add a comment before this step that explains this? Something like
```rb
 # Only certain admin roles have access to this page; other admins will see the following error message
```

brianjaustin (approved):
Thank you!

PR #5139, opened 2025-04-10
AO3-6388 Add confirmation page and email when a user changes the email on their account

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6388

## Purpose

Enables the reconfirmable setting of devise, so that users need to confirm the new email address they are trying to change to via a link sent to that email. Adds a confirmation page for users changing their email.

## Credit

Bilka

### DIFF
--- a/config/locales/controllers/en.yml
+++ b/config/locales/controllers/en.yml
@@ -25,6 +25,10 @@ en:
         success:
           one: Successfully removed pseud %{pseuds} from this work.
           other: Successfully removed pseuds %{pseuds} from this work.
+  application:
+    access_denied:
+      access_denied:
+        logged_in: Sorry, you don't have permission to access the page you were trying to reach.
   archive_faqs:
     create:
       success: Archive FAQ was successfully created.
@@ -213,6 +217,8 @@ en:
           other: You may reset your password %{count} more times.
         user_not_found: We couldn't find an account with that email address or username. Please try again.
     reconfirm_email:
+      access_denied:
+        logged_in: You are not logged in to the account whose email you are trying to change. Please log out and try again.
       invalid_token: This email confirmation link is invalid or expired. Please check your email for the correct link or submit the email change form again.
       success: Your email has been successfully updated.
     status:
--- a/config/locales/controllers/en.yml
+++ b/config/locales/controllers/en.yml
@@ -202,6 +202,9 @@ en:
           one: You may reset your password %{count} more time.
           other: You may reset your password %{count} more times.
         user_not_found: We couldn't find an account with that email address or username. Please try again.
+    reconfirm_email:
+      invalid_token: This email confirmation link is invalid or expired. Please check your email for the correct link or submit the email change form again.
+      success: Your email has been successfully updated.
     status:
       ban_notice_html: Your account has been banned. You are not permitted to post or edit content on AO3. Please check your email or %{contact_abuse_link} for more information.
       suspension_notice_html: Your account has been suspended until %{suspended_until}. You cannot post, edit, or delete content until your suspension has ended. Please check your email or %{contact_abuse_link} for more information.
--- a/config/locales/controllers/en.yml
+++ b/config/locales/controllers/en.yml
@@ -182,12 +182,22 @@ en:
     index:
       collection_page_title: "%{collection_title} - Tags"
   users:
+    changed_password:
+      blank_password: You must enter your old password.
+      wrong_password: Your old password was incorrect.
     changed_username:
       admin:
         successfully_updated: Username has been successfully updated.
       user:
         incorrect_password: Your password was incorrect
         successfully_updated: Your username has been successfully updated.
+    confirm_change_email:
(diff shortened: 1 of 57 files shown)


config/locales/mailers/en.yml
```diff
@@ -199,10 +203,25 @@ en:
...
+        text:
+          one: 'If you made this request, check your email at %{unconfirmed_email} within %{count} day to confirm your email change. If you do not receive a confirmation email, or if you have other issues, please contact Support: %{contact_support_url}.'
+          other: 'If you made this request, check your email at %{unconfirmed_email} within %{count} days to confirm your email change. If you do not receive a confirmation email, or if you have other issues, please contact Support: %{contact_support_url}.'
+      contact_policy_abuse: contact Policy & Abuse
+      contact_support: contact Support
+      email_change_form: submit the email change form
+      made_request: Someone has made a request to change the email address associated with your %{app_name} account.
+      not_made_request:
+        html: If you did not make this request, someone else may have access to your AO3 account. Please %{reset_password_link}. If you are unable to reset your password, %{contact_policy_abuse_link}.
```
lydia-theda, line 218: If line 216 uses %{app_name} instead of AO3, then line 218 should too?

config/locales/controllers/en.yml
```diff
@@ -25,6 +25,10 @@ en:
         success:
           one: Successfully removed pseud %{pseuds} from this work.
           other: Successfully removed pseuds %{pseuds} from this work.
+  application:
+    access_denied:
+      access_denied:
```
lydia-theda, line 30: (sorry for using this as a learning experience, should I do that over slack) does this need to be nested twice? if so why?

Bilka2: (It's fine here!) So, the way I did this locale is a workaround that might be considered a bit of a hack:
https://github.com/otwcode/otwarchive/blob/e20d3cb1032714112c59a6a6bb9756b3c1e02220/app/controllers/application_controller.rb#L285

The `.` at the start of the first key means that the locale key is relative, in this case to the controller and action (`application_controller` and `access_denied`). That's the first two levels of nesting. The other two levels are the ones defined right in the key.  
I'm making use of the key being relative because when we get to showing this error from the `users_controller` and `reconfirm_email` action, Rails will use that instead - so the full key becomes `users.reconfirm_email.access_denied.logged_in` in that case! That allows the special case of the error for this one action. This key is what my goal was, since it's very clear where it comes from with the `access_denied` part clearly identifying the underlying method. It means we get the duplication for the default locale that you point out, but that feels worth it for making it very clear where and how we override it with the special case for the specific controller and action.

This is not a set thing - a reviewer could come along and ask for the key to be something that incorperates both the lack of access and the logged in status into one level (`user_access_denied` or something) but I couldn't come up with a good wording for it.

features/other_a/orphan_account.feature
```diff
@@ -52,7 +52,7 @@ Given I have an orphan account
     And I should see "Shenanigans 2"
     And I should see "Shenanigans - the early years"
   When I follow "Preferences"
-  When I follow "Orphan My Works"
+  When I go to the orphan all works page
```
lydia-theda, line 55: second question because I don't understand tests: Since the button is gone from the preferences page what's the purpose of going to that page in line 54? can you even get to the orphan all works page via a direct url link (since I think now the interface makes it so can only do it via edit multiple works > select all > orphan)?

Bilka2: The page is just https://archiveofourown.org/orphans/new. The version you get to from the edit multiple works page is slightly different.

Since we have the code for this page still around, I didn't want to remove the tests for it. I did not want to make this pull request larger by discussing whether we want to force users to select specific works/pseuds/whatever to orphan and then potentially removing/changing that code.

lydia-theda: That makes sense, but I think I was asking why "When I follow "Preferences"" was needed if the page can be accessed from anywhere.

Bilka2: Oh. I should remove that, thank you for spotting it.

config/locales/controllers/en.yml
```diff
@@ -25,6 +25,10 @@ en:
         success:
           one: Successfully removed pseud %{pseuds} from this work.
           other: Successfully removed pseuds %{pseuds} from this work.
+  application:
+    access_denied:
```
brianjaustin, line 29: To avoid the duplicate key name, could this be `auth` or something similar? That way, we could put other authentication-related messages under here if we ever have those

Bilka2: It would be nicer to change the inner two keys instead of the outer ones, because i18n-tasks will by default grab the controller and method name for the relative key. Would `application.access_denied.auth.logged_in` work?

brianjaustin: Hmm, in that case I think it's fine to be honest

brianjaustin (approved):

PR #5141, opened 2025-04-11
AO3-6964 Correctly Display Reply Notice on Admin Posts

Nick-Raal:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6964 

## Purpose

This pull request changes the way the notice for moderated comments is displayed by checking the parent of the commentable object rather than the object itself.

## References

## Credit

niic (he/him)

### DIFF
--- a/features/comments_and_kudos/comments_adminposts.feature
+++ b/features/comments_and_kudos/comments_adminposts.feature
@@ -113,6 +113,27 @@ Feature: Commenting on admin posts
     Then I should see "Comment created!"
       And I should see "zug zug"
 
+  Scenario: Admin post with comment moderation
+    Given I have posted an admin post
+      And I am logged in as a "communications" admin
+    When I go to the admin-posts page
+      And I follow "Edit"
+      And I check "Enable comment moderation"
+    When I am logged out
+      And I go to the admin-posts page
+      And I follow "Default Admin Post"
+      And I fill in "comment[name]" with "tester"
+      And I fill in "comment[email]" with "tester@example.com"
+      And I fill in "comment[comment_content]" with "guz guz"
+      And I press "Comment"
+    When I follow "Reply"
+      And I fill in "comment[name]" with "tester"
+      And I fill in "comment[email]" with "tester@example.com"
+      And I fill in "comment[comment_content]" with "guz guz"
+      And I press "Comment"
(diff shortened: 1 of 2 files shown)


features/comments_and_kudos/comments_adminposts.feature
```diff
@@ -113,6 +113,34 @@ Feature: Commenting on admin posts
     Then I should see "Comment created!"
       And I should see "zug zug"
 
+  Scenario: Admin post with comment moderation
```
sarken, line 116: This should be more descriptive, like: "Scenario: Replying to a comment on an admin post with comment moderation shows a notice saying comments are moderated"

I haven't tested this, so it might need some tweaks, but I think we can make this whole test a lot shorter by creating an admin post with a comment and then enabling moderation:
```cucumber
Given the admin post "Generic Admin Post" with guest comments enabled
  And a guest comment on the admin post "Generic Admin Post"
When I am logged in as a "communications" admin
  And I go to the admin-posts page
  And I follow "Edit"
  And I check "Enable comment moderation"
  And I log out
  And I view the admin post "Generic Admin Post" with comments
  And I follow "Reply" # I looked at the issue and we don't need to actually leave a comment. The text we're testing should appear as soon as we open the reply form.
Then I should see "Comments on this news post are moderated." within #some-selector # I think we need to specify a selector here that's unique to the reply so we can be sure the test isn't just passing because the correct text appears on the form for leaving a top level comment. I'm pretty sure just checking for the text without the selector will pass even with the old, broken code.
  And I should not see "Comments on this work are moderated." # This isn't strictly necessary, but might not be a bad idea given the selector issue and the potential for the test to pass with the old code.
```

It would also avoid the issue with the test failing when looking for the Unreviewed Comments link.

sarken (commented):
I realize this is a draft and so you might not want review yet, but I left a suggestion about the test that might help with the failure.

However, it looks like `features/comments_and_kudos/kudos.feature:70` is consistently failing even when I rerun the tests, so I think that might be a real issue. My first suggestion would be merging with master to see if that helps at all before spending too much time digging into it.

Bilka2:
Since this PR has been stalled for over a year, I've adopted the code at #5930

PR #5142, opened 2025-04-12
AO3-6963 Remove excess whitespace from page titles and description meta tag

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6963

## Purpose

Removes excess whitespace from page titles and description meta tag so they display nicer link previews on sites like Bluesky. ~It might be nice to have this before our next batch of release notes.~

I stopped checking `defined?` because I couldn't imagine why we'd want to return the variable's type, only whether it was nil or not. If we're leery of changing, or if it was better for performance, I'm happy to put it back.

Reviewdog is probably going to complain about the instance variable in the helper, but this is far from the only place we do that. I feel like we could probably do `rubocop:disable` here, but if we'd rather some other solution, I'm all ears.

## References

We had quite a few tests for sensible `@page_title` titles, so I added some tests for a couple we intend to replace, just to make sure they're broken in the same way:

* [AO3-5528: Tag set pages have weird or uninformative browser page titles](https://otwarchive.atlassian.net/issues/AO3-5528)
* [AO3-6937: Bad browser page title on inbox](https://otwarchive.atlassian.net/issues/AO3-6937)


app/helpers/application_helper.rb
```diff
@@ -631,4 +631,20 @@ def tos_exempt_page?
       %w[index show].include?(params[:action])
     end
   end
+
+  def browser_page_title
```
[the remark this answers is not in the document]
sarken: It occurs to me removing the instance variables will be two lines shorter than disabling/renabling the cop at the start/end of the method, so I have switched sides and now favor this approach 😆

Bilka2:
What do you think about adding a test for the `"#{process_title(controller.action_name)} #{process_title(controller.controller_name.singularize)}"` line with a page where we're not going to end up replacing it with something custom? Best I can quickly come up with is "New Work" for works/new

features/comments_and_kudos/inbox.feature
```diff
@@ -159,3 +159,8 @@ Feature: Get messages in the inbox
       And I go to the homepage
     Then I should see "sewwiththeflo on Cat Thor's Bizarre Adventure"
       And I should see "Thank you! Please go to bed."
+
+   Scenario: Inbox has the expected browser page title
+    Given I am logged in as "boxer"
+    When I go to boxer's inbox page
+    Then I should see the page title "Show Inbox | Example Archive"
```
Bilka2, line 166: I'd remove the "Example Archive" because it will cause the tests to fail if you change it locally (which I've done).

sarken: I'd really like to check the entire title, not just the part before the pipe, so I'll update the tests to set the app name. Do you get the failure in [this challenge test](https://github.com/otwcode/otwarchive/blob/4418571dfa0dc5e00438a005937ae1e7dc365e9b/features/gift_exchanges/challenge_yuletide.feature#L591), too? I don't think I've ever run into it even though my local.yml sets `APP_NAME: Development Archive`, but it's possible I've never run that test locally.

Bilka2: Yes, the challenge test failed for me before, that's why I knew it was a potential issue.

Nice solution!

Bilka2:
Could you merge master so the tests can run through? The code looks good!

Bilka2 (approved):

PR #5143, opened 2025-04-13
AO3-5714 Do not treat non-canonical tags as fandom-associated in TagSets

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5714

## Purpose

Non-canonical characters associated with canonical fandoms _look_ like they're associated when added to a TagSet, but in prompts, they aren't treated that way (either for autocomplete or, after https://github.com/otwcode/otwarchive/pull/5140 is merged, for situations when fandom-only is selected). This PR updates the categorization to show these tags under the association-needed section instead to reduce confusion.

## Review Notes

I cleaned up a few Rubocop complaints as well, which changed indentation. I'd recommend using the "hide whitespace" option when reviewing, to make the diff a bit more comprehensible

### DIFF
--- a/app/controllers/owned_tag_sets_controller.rb
+++ b/app/controllers/owned_tag_sets_controller.rb
@@ -74,31 +74,31 @@ class OwnedTagSetsController < ApplicationController
       # we use this to store the tag name results
       @tag_hash = HashWithIndifferentAccess.new
 
-      %w(character relationship).each do |tag_type|
-        if @tag_set.has_type?(tag_type)
-          ## names_by_parent returns a hash of arrays like so:
-          ## hash[parent_name] => [child name, child name, child name]
+      %w[character relationship].each do |tag_type|
+        next unless @tag_set.has_type?(tag_type)
 
-          # get the manually associated fandoms
-          assoc_hash = TagSetAssociation.names_by_parent(TagSetAssociation.for_tag_set(@tag_set), tag_type)
+        ## names_by_parent returns a hash of arrays like so:
+        ## hash[parent_name] => [child name, child name, child name]
 
-          # get canonically associated fandoms
-          # Safe for constantize as tag_type restricted to character relationship
-          canonical_hash = Tag.names_by_parent(tag_type.classify.constantize.in_tag_set(@tag_set), "fandom")
+        # get the manually associated fandoms
+        assoc_hash = TagSetAssociation.names_by_parent(TagSetAssociation.for_tag_set(@tag_set), tag_type)
 
-          # merge the values of the two hashes (each value is an array) as a set (ie remove duplicates)
-          @tag_hash[tag_type] = assoc_hash.merge(canonical_hash) {|key, oldval, newval| (oldval | newval) }
+        # get canonically associated fandoms
+        # Safe for constantize as tag_type restricted to character relationship
+        canonical_hash = Tag.names_by_parent(tag_type.classify.constantize.in_tag_set(@tag_set).canonical, "fandom")
 
-          # get any tags without a fandom
-          remaining = @tag_set.with_type(tag_type).with_no_parents
-          if remaining.count > 0
-            @tag_hash[tag_type]["(No linked fandom - might need association)"] ||= []
-            @tag_hash[tag_type]["(No linked fandom - might need association)"] += remaining.pluck(:name)
-          end
+        # merge the values of the two hashes (each value is an array) as a set (ie remove duplicates)
+        @tag_hash[tag_type] = assoc_hash.merge(canonical_hash) { |_key, oldval, newval| (oldval | newval) }
 
-          # store the parents
-          @fandom_keys_from_other_tags += @tag_hash[tag_type].keys
+        # get any tags without a fandom
+        remaining = @tag_set.with_type(tag_type).where.not(name: @tag_hash[tag_type].values.flatten)
+        if remaining.any?
+          @tag_hash[tag_type]["(No linked fandom - might need association)"] ||= []
+          @tag_hash[tag_type]["(No linked fandom - might need association)"] += remaining.pluck(:name)
         end
+
+        # store the parents
+        @fandom_keys_from_other_tags += @tag_hash[tag_type].keys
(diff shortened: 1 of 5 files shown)


Bilka2:
Just a quick comment - should the backfill task also check that the fandom is part of the tagset?

brianjaustin:
@Bilka2 I tried removing the Fandom tag from a tagset locally, and based on the way things look, I would still create that assocation:
image

app/controllers/owned_tag_sets_controller.rb
```diff
@@ -74,31 +74,31 @@ def show
...
-          @fandom_keys_from_other_tags += @tag_hash[tag_type].keys
+        # get canonically associated fandoms
+        # Safe for constantize as tag_type restricted to character relationship
+        canonical_hash = Tag.names_by_parent(tag_type.classify.constantize.in_tag_set(@tag_set).canonical, "fandom")
+
+        # merge the values of the two hashes (each value is an array) as a set (ie remove duplicates)
+        @tag_hash[tag_type] = assoc_hash.merge(canonical_hash) { |_key, oldval, newval| (oldval | newval) }
+
+        # get any tags without a fandom
+        remaining = @tag_set.with_type(tag_type).where.not(name: @tag_hash[tag_type].values.flatten)
```
Bilka2, line 94: This new code make sense to me but help me a bit - how did the old code avoid grabbing tags that have manual associations? In the new one it's clear, they are included in the merged hash. But in the old one..?

lib/tasks/after_tasks.rake
```diff
@@ -535,5 +535,25 @@ namespace :After do
...
     end
   end
+
+  desc "Create TagSetAssociations for non-canonical tags belonging to canonical fandoms in TagSets"
+  task(create_non_canonical_tagset_associations: :environment) do
+    # We want to get all set taggings where the tag is not canonical, but has a parent fandom that _is_ canonical.
+    # This might be possible with pure Ruby, but unfortunately the parent tag in common_taggings is polymorphic
+    # (even though it doesn't need to be), which makes that trickier.
+    non_canonicals = SetTagging
+      .joins("INNER JOIN `tag_sets` `tag_set` ON `tag_set`.`id` = `set_taggings`.`tag_set_id` INNER JOIN `owned_tag_sets` ON `owned_tag_sets`.`tag_set_id` = `tag_set`.`id` INNER JOIN `tags` `tag` ON `tag`.`id` = `set_taggings`.`tag_id` JOIN `common_taggings` `common_tagging` ON `tag`.`id` = `common_tagging`.`common_tag_id` JOIN `tags` `tag2` ON `common_tagging`.`filterable_id` = `tag2`.`id`")
```
Bilka2, line 545:
```suggestion
      .joins("INNER JOIN `tag_sets` `tag_set` ON `tag_set`.`id` = `set_taggings`.`tag_set_id` INNER JOIN `owned_tag_sets` ON `owned_tag_sets`.`tag_set_id` = `tag_set`.`id` INNER JOIN `tags` `tag` ON `tag`.`id` = `set_taggings`.`tag_id` JOIN `common_taggings` `common_tagging` ON `common_tagging`.`common_tag_id` = `tag`.`id` JOIN `tags` `tag2` ON `common_tagging`.`filterable_id` = `tag2`.`id`")
```
for consistency to make it slightly more readable. It'd probably also be good to decide on one of join and inner join. Might also be good to name tag2 something like parent_tag.

lib/tasks/after_tasks.rake
```diff
@@ -535,5 +535,25 @@ namespace :After do
...
+      .where("`tag`.`canonical` = FALSE AND `tag`.`type` IN ('Character', 'Relationship') AND `tag_set`.`id` IS NOT NULL AND `tag2`.`type` = 'Fandom' AND `tag2`.`canonical` = TRUE")
+
+    non_canonicals.find_in_batches.with_index do |batch, index|
+      puts "Creating TagSetAssociations for batch #{index + 1}"
+      batch.each do |set_tagging|
+        set_tagging.tag.fandoms.where(canonical: true).find_each do |fandom|
+          TagSetAssociation.create!(owned_tag_set: set_tagging.tag_set.owned_tag_set,
+                                    tag: set_tagging.tag, parent_tag: fandom)
```
Bilka2, line 553: Does it change anything in the query logs to pull the two `set_tagging...` out of the fandoms loop? Since technically we shouldn't need to go to the database for every iteration, but maybe Rails is clever enough to cache it.

app/controllers/owned_tag_sets_controller.rb
brianjaustin: I don't think the old code _would_ have done that properly. The only reason it _seemed_ to work (in this case) was that we're dealing with non-canonical tags that did have a parent, so regardless of whether there was a manual association, they wouldn't show up in `with_no_parents`. If someone had an association for a fandom that didn't have a `CommonTagging` yet to that relationship/character, I think the old code would not have done the right thing. I don't know how common that is, though

lib/tasks/after_tasks.rake
```diff
@@ -535,5 +535,28 @@ namespace :After do
       puts "Finished converting kudos from official users to guest kudos"
     end
   end
+
+  desc "Create TagSetAssociations for non-canonical tags belonging to canonical fandoms in TagSets"
+  task(create_non_canonical_tagset_associations: :environment) do
+    # We want to get all set taggings where the tag is not canonical, but has a parent fandom that _is_ canonical.
+    # This might be possible with pure Ruby, but unfortunately the parent tag in common_taggings is polymorphic
+    # (even though it doesn't need to be), which makes that trickier.
+    non_canonicals = SetTagging
```
Bilka2, line 544:
```
3.2.7 :021 > non_canonicals.size
 => 3813 
3.2.7 :022 > non_canonicals.uniq.size
 => 2898 
```
looks like this is getting duplicates somehow (numbers from staging) :/

Bilka2 (approved):
Massive kudos for dealing with this code!

PR #4773, opened 2024-03-29
AO3-5053 Update bookmarker byline when pseud or username changes

irrationalpie7:
# Pull Request Checklist

* [x] Have you read ["How to write the perfect pull request"](https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/)?
* [x] Have you read the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [x] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)?
* [x] Have you added the [Jira](https://otwarchive.atlassian.net) issue number
  as the *first* thing in your pull request title (e.g. `AO3-1234 Fix thing`)
* [x] Do you have fewer than 5 pull requests already open? If not, please wait
  until they are reviewed and merged before creating new pull requests.

## Issue

https://otwarchive.atlassian.net/browse/AO3-5053

## Purpose

Updates bookmark blurbs when a bookmarker changes their username or pseud

## Testing Instructions

See Jira.

## References

I modeled some of my approach on https://otwarchive.atlassian.net/browse/AO3-4120

## Credit

irrationalpie, they/them

PR #5144, opened 2025-04-18
AO3-5053 Update bookmarker byline when pseud or username changes

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5053

## Purpose

Updates bookmark blurbs when a bookmarker changes their username or pseud.

## References

Adopts #4773 

## Credit

irrationalpie
Bilka


spec/requests/bookmarks_n_plus_one_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  describe "#index" do
+    context "when viewing recent bookmarks" do
+      subject do
+        proc do
+          get bookmarks_path
+        end
+      end
+
+      context "when logged out" do
```
Bilka2, line 48: The logged in version has an extra query per bookmark because of the `is_author_of?` check so I decided to just not test that because it's really not the point here.

app/views/bookmarks/_bookmark_user_module.html.erb
```diff
@@ -1,12 +1,13 @@
 <% # expects "bookmark" %>
 <div class="<% if is_author_of?(bookmark) %>own <% end %>user module group">
+  <!--bookmarker-->
```
sarken, line 3: If you're in the mood, please feel free to remove these HTML comments! They really don't offer any particular benefit.

brianjaustin (approved):

PR #5152, opened 2025-04-21
AO3-5862 Use Departure to really delete users.recently_reset

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5862

## Purpose

Update the old migration (that was never actually applied per the ticket above) to use Departure and have a more recent timestamp.


Bilka2 (approved):
Agreed with redating this so we don't need to do a schema dump while we have a bunch of open big PRs with migrations

PR #5157, opened 2025-04-24
AO3-6977 Spam check edited comments, exclude comments on own work

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6977

## Purpose

Make the spam check run on edited comments. Exclude comments on the commenter's on work from the spam check.

## Credit

Bilka


EchoEkhi:
image
It looks like the error messages aren't showing up properly. I think this issue predates this ticket

EchoEkhi:
I think there is still a loophole in this implementation. I can gradually build up a spam comment by making small edits each time just under the character threshold

Bilka2:
> I think this issue predates this ticket
Indeed this is existing behaviour, so not relevant here.

We're handling the other issue.

EchoEkhi:
> not relevant here

I think we should ideally try to deploy a fix for this bug in parallel with this patch, since this is the first use-case that would cause this problem to manifest for a non-trivial proportion of users - the other two realistically encounterable pre-existing cases are editing a comment to zero length and editing it past the character limit, both have client-side js validation error messages, so they appear fine even though the underlying problem is still there.

sarken:
I agree we should make sure the error message shows up. It doesn't matter a ton for the spammers, but if a real user's edits get marked as spam, it would be good to let them know that's the problem instead of having to deal with tickets that say, "I can't edit my comment," and nothing else. (We'll probably still get some tickets that say, "It says my edits are spam," but that's easier to deal with than trying to figure out what they mean when they say they can't edit the comment: is the button missing because of an extension, are they suspended, are they logged out, are they getting this error, etc.)

Bilka2:
With javascript:
Screenshot_20250428_142002

Without javascript:
Screenshot_20250428_142026

brianjaustin (approved):

EchoEkhi:
> I think there is still a loophole in this implementation. I can gradually build up a spam comment by making small edits each time just under the character threshold

@brianjaustin I don't think this issue is fixed yet?

brianjaustin:
This has been discussed with the relevant committees, and the decision was to move forward without addressing that at this time. PAC will continue to monitor the situation and discussion will be held internally as needed

PR #5160, opened 2025-04-26
AO3-6982 Fix flaky cucumber features using javascript

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6982

## Purpose

No more flaky tests.

## Testing Instructions

No manual QA required.

**search/filters.feature:**
Without the fix [40 test runs](https://github.com/Bilka2/otwarchive/actions/runs/14663051259) resulted in 6 failures.
With the fixes [120 test runs](https://github.com/Bilka2/otwarchive/actions/runs/14664324231) result in no failures.

**bookmarks:**
Without the fix [40 test runs](https://github.com/Bilka2/otwarchive/actions/runs/14665264729) resulted in 9 failures.
With the fixes [120 test runs](https://github.com/Bilka2/otwarchive/actions/runs/14665498093) result in no failures.

**other_b:**
Without the fix [40 test runs](https://github.com/Bilka2/otwarchive/actions/runs/14681505380) resulted in 5 failures.
With the fixes [120 test runs](https://github.com/Bilka2/otwarchive/actions/runs/14682012423) result in no failures.

**works:**
Without the fix [40 test runs](https://github.com/Bilka2/otwarchive/actions/runs/14683081113) resulted in 7 failures.
With the fixes [120 test runs](https://github.com/Bilka2/otwarchive/actions/runs/14683842096) result in no failures.

## Credit

Bilka

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -19,7 +19,7 @@ class User < ApplicationRecord
   include BackwardsCompatiblePasswordDecryptor
 
   # Allows other models to get the current user with User.current_user
-  cattr_accessor :current_user
+  thread_cattr_accessor :current_user
 
   # Authorization plugin
   acts_as_authorized_user
--- a/features/search/filters.feature
+++ b/features/search/filters.feature
@@ -8,11 +8,12 @@ Feature: Filters
     Given a canonical fandom "The Hobbit"
       And a canonical fandom "Harry Potter"
       And a canonical fandom "Legend of Korra"
-      And I am logged in as "meatloaf"
-      And I post the work "A Hobbit's Meandering" with fandom "The Hobbit"
-      And I post the work "Bilbo Does the Thing" with fandom "The Hobbit, Legend of Korra"
-      And I post the work "Roonal Woozlib and the Ferrets of Nimh" with fandom "Harry Potter"
+      And the work "A Hobbit's Meandering" by "meatloaf" with fandom "The Hobbit"
+      And the work "Bilbo Does the Thing" by "meatloaf" with fandom "The Hobbit, Legend of Korra"
+      And the work "Roonal Woozlib and the Ferrets of Nimh" by "meatloaf" with fandom "Harry Potter"
+      And all indexing jobs have been run
       And the dashboard counts have expired
+      And I am logged in as "meatloaf"
 
   @javascript
   Scenario: You can filter through a user's works using inclusion filters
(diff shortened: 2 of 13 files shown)


app/models/user.rb
```diff
@@ -19,7 +19,7 @@ class User < ApplicationRecord
   include BackwardsCompatiblePasswordDecryptor
 
   # Allows other models to get the current user with User.current_user
-  cattr_accessor :current_user
+  thread_cattr_accessor :current_user
```
Bilka2, line 22: This has no effect in production because we have puma set to only use on thread. However, I think it may help in tests. Maybe. Possibly. It shouldn't hurt, at least.

brianjaustin (approved):
Thank you!!

features/search/filters.feature
```diff
@@ -84,6 +85,7 @@ Feature: Filters
   @javascript
   Scenario: Filter through a user's works with non-existent tags
     Given the tag "legend korra" does not exist
+      And all indexing jobs have been run
```
sarken, line 88: I realize this has been merged, but I have a question! Why isn't the `all indexing jobs have been run` in the path definition for `when /^([^ ]*?)(?:'s)? works page$/i` sufficient here?

Bilka2: Good question! I see that it should run with `user's bookmarks page` as well. But these tests were flaky anyway... The `I post the work` step runs the indexing as well, yet we had indexing issues with the works.

A lot of the failures I fixed here I can't really explain. Why would posting a work sometimes fail? Why did creating one of the collections sometimes fail for the work test? No idea.

The only common pattern seems to be that it's logged in actions, so we might still have issues around `User.current_user`. My suspicion is that autocomplete (which is enabled by javascript) messes with it somehow, but that's just a guess. Beyond the change to the accessor I did here, we could try https://api.rubyonrails.org/classes/ActiveSupport/CurrentAttributes.html. But that's a pretty large change to make based on my complete guess as to what's going on.

The fixes mostly aim at using FactoryBot because if we don't use the browser then its behaviour can't be flaky. And we get the minor speed benefit on top. Just have to be careful not to replace too much, since some things need to be tested via the controller/GUI.

The only other thing that stood out to me with our test setup is https://github.com/otwcode/otwarchive/blob/7da89d39ecc5aa70fb8281b54ab9dc6492ccae4a/features/support/capybara.rb#L22 We use puma in prod and dev now, we could also use it in tests. But once again that's a significant change. There's also https://github.com/teamcapybara/capybara/commit/e2f9f6cc60d622baf27624e5d224d38cc85f551f which we could adopt.

Sorry that I don't have a satisfying answer for you :/

PR #5162, opened 2025-04-26
AO3-6757 Remove image embeds when hiding comments

scottsds:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6757

## Purpose

When a comment is hidden by an administrator, the contents are replaced with a placeholder and are no longer visible to users (other than the user who posted the comment, if it was by a registered user). However, they are still visible to admins. This PR will strip the IMG tag from comment content, and leave a plain URL in its place.

### DIFF
--- a/features/comments_and_kudos/hidden_comments.feature
+++ b/features/comments_and_kudos/hidden_comments.feature
@@ -66,13 +66,10 @@ Feature: Comment hiding
     Then I should see "A suspicious comment"
       And I should not see "This comment has been hidden by an admin."
 
-  Scenario Outline: Embedded images in hidden comments are replaced with their URLs.
-    Given I am logged in as "author"
-    And I post the work "Popular Fic"
-    And I am logged out
-    And I am logged in as "commenter"
-    And I post the comment "OMG! <img src= 'https://example.com/image.jpg'>" on the work "Popular Fic"
-    And I am logged out
+  Scenario: Embedded images in hidden comments are replaced with their URLs.
+    Given the work "Popular Fic"
+      And I am logged in as "commenter"
+      And I post the comment "OMG! <img src= 'https://example.com/image.jpg'>" on the work "Popular Fic"
 
     # Delay to make sure the cache is expired when the comment is hidden:
     When it is currently 1 second from now
--- a/features/comments_and_kudos/hidden_comments.feature
+++ b/features/comments_and_kudos/hidden_comments.feature
@@ -65,3 +65,26 @@ Feature: Comment hiding
       And I view the work "Popular Fic" with comments
     Then I should see "A suspicious comment"
       And I should not see "This comment has been hidden by an admin."
+
+  Scenario Outline: Embedded images in hidden comments are replaced with their URLs.
+    Given I am logged in as "author"
+    And I post the work "Popular Fic"
+    And I am logged out
+    And I am logged in as "commenter"
+    And I post the comment "OMG! <img src= 'https://example.com/image.jpg'>" on the work "Popular Fic"
+    And I am logged out
+
+    # Delay to make sure the cache is expired when the comment is hidden:
+    When it is currently 1 second from now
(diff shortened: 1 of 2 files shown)


Bilka2:
I think it would be better to modify `use_image_safety_mode?` instead to check for whether the comment is hidden: https://github.com/otwcode/otwarchive/blob/3bcfaf96b7ad6546ee43a31fa03fd2b4c5f33647/app/models/comment.rb#L532-L534

When I test locally, this also allows the image to show up again when you unhide the comment.

scottsds:
> I think it would be better to modify `use_image_safety_mode?` instead to check for whether the comment is hidden:
> 
> https://github.com/otwcode/otwarchive/blob/3bcfaf96b7ad6546ee43a31fa03fd2b4c5f33647/app/models/comment.rb#L532-L534
> 
> When I test locally, this also allows the image to show up again when you unhide the comment.

Wouldn't that only prevent comments from initially being posted with embedded images? I don't see how to use if in one-off situations.

Bilka2:
The method is used when displaying the comment content. From what I can see, it doesn't affect posting the comment at all.

features/comments_and_kudos/hidden_comments.feature
```diff
@@ -65,3 +65,26 @@ Feature: Comment hiding
       And I view the work "Popular Fic" with comments
     Then I should see "A suspicious comment"
       And I should not see "This comment has been hidden by an admin."
+
+  Scenario Outline: Embedded images in hidden comments are replaced with their URLs.
+    Given I am logged in as "author"
+    And I post the work "Popular Fic"
+    And I am logged out
```
Bilka2, line 72: Can you convert this into a single `Given the work "Popular Fic"` step? It runs a bit faster than logging in and posting manually

features/comments_and_kudos/hidden_comments.feature
```diff
@@ -65,3 +65,26 @@ Feature: Comment hiding
...
+    And I am logged in as "commenter"
+    And I post the comment "OMG! <img src= 'https://example.com/image.jpg'>" on the work "Popular Fic"
+    And I am logged out
```
Bilka2, line 75: Could you indent these one step, like the other And steps? Also, I think you can leave out the logging out step, logging in as super admin should also run that

features/comments_and_kudos/hidden_comments.feature
```diff
@@ -65,3 +65,26 @@ Feature: Comment hiding
...
+    Then I should see the image "src" text "https://example.com/image.jpg"
+      And I should not see "OMG! https://example.com/image.jpg"
+      And I press "Hide Comment"
+    Then I should see "Comment successfully hidden!"
+    Then I should not see the image "src" text "https://example.com/image.jpg"
+      And I should see "OMG! https://example.com/image.jpg"
+    Then I press "Make Comment Visible"
+      And I should see "Comment successfully unhidden!"
+    Then I should see the image "src" text "https://example.com/image.jpg"
+      And I should not see "OMG! https://example.com/image.jpg"
```
Bilka2, line 90: One newline to end the file please

features/comments_and_kudos/hidden_comments.feature
```diff
@@ -65,3 +65,26 @@ Feature: Comment hiding
       And I view the work "Popular Fic" with comments
     Then I should see "A suspicious comment"
       And I should not see "This comment has been hidden by an admin."
+
+  Scenario Outline: Embedded images in hidden comments are replaced with their URLs.
```
Bilka2, line 69:
```suggestion
  Scenario: Embedded images in hidden comments are replaced with their URLs.
```
Scenario Outlines are for use with examples

Bilka2 (commented):
Looks good! I left some comments about the test

Bilka2 (approved):
Thank you!

PR #5164, opened 2025-04-27
AO3-6005 Convert Cucumber tests in other_b/errors.feature to RSpec tests

Hamham6:
# Pull Request Checklist

* [ ] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)? N/A only tests

## Issue

[AO3-6005](https://otwarchive.atlassian.net/browse/AO3-6005): Convert Cucumber tests in other_b/errors.feature to RSpec tests

## Purpose

Cover tests in rspec instead of cucumber.

## Testing Instructions

N/A

## Credit
Hamham6, it/its

[AO3-6005]: https://otwarchive.atlassian.net/browse/AO3-6005?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Hamham6:
I am missing something here in regards to the '^' error spec test. I added the test 1. via a fandom tag and 2. in works/updating_tags_spec, because the original cucumber test 1. added a fandom tag and 2. used the edit work page. The expects is copyed from the series controller spec, specifically "update", with the thinking of it having the same 'collect all the errors and display them all and don't redirect anywhere' behavior. The rest of that test is copyed from the shared examples "cannot update work tags and language" on the same file, with the thinking of trying to copy the  'don't save a change if it errors' behavior. But the error message is showing up as `Taggings tagger name ^Tag name 'fandom^tag' cannot include the following restricted characters: , &#94; * < > { } = ` ， 、 \\ %`, and I can't seem to follow the code well enough to find where the "Taggings tagger name ^" is coming from.

app/helpers/validation_helper.rb
```diff
@@ -28,7 +28,7 @@ module ValidationHelper
   }
   
   # much simplified and html-safed version of error_messages_for
-  # error messages containing a "^" will have everything before the "^" wiped out
+  # error messages containing a "^" will have everything before the "^" wiped out
```
Hamham6, line 31: (I only editted this so I could ask the question on the code, I'll revert the change after.) Was this fixed in AO3-6320?

spec/controllers/works/updating_tags_spec.rb
```diff
@@ -36,6 +36,18 @@
       before { fake_login_known_user(work.users.first) }
 
       it_behaves_like "can update work tags and language"
+
+      context "when adding a tag that contains '^'" do
```
Hamham6, line 40: This is the broken test.

Hamham6:
Sorry about the extra commits, I think I accidentally merged the branch I used for AO3-6031 over this one.

app/helpers/validation_helper.rb
Bilka2: This is intentional, we use it to remove attribute name prefix from the error message. So basically what you found in your broken test - the `Taggings tagger name ^` is removed from the error display.

Bilka2:
> I am missing something here in regards to the '^' error spec test. 

Yeah, this issue was made before that test existed. It is testing the display code, which we can't test with rspec. So this test needs to stay as a cucumber test. Only the other tests can be moved to rspec.

spec/controllers/pseuds_controller_spec.rb
```diff
@@ -251,4 +251,13 @@
       it_behaves_like "an action admins can't access"
     end
   end
+
+  describe "index" do
```
Hamham6, line 267: While coding the pseud test I got confused between index and show and coded this while figuring it out, and since it works I added it. It wasn't covered in the cucumber test, so I can remove it if it's out of scope.

Bilka2: It's fine to have! However, we'd like to change this behaviour, so please add a todo comment pointing out that this should be changed with [AO3-6989](https://otwarchive.atlassian.net/browse/AO3-6989).

spec/controllers/tags_controller_spec.rb
Bilka2, line 1: Could you remove the "Scenario: Tags that don't exist cause errors" in tag_wrangling_more.feature as well?

Once you've done that, you should be able to remove the `fail with a not found error` step from /home/`features/step_definitions/web_steps.rb`.

Bilka2 (changes requested):

spec/controllers/collections_controller_spec.rb
```diff
@@ -9,4 +9,14 @@
       expect(assigns[:page_subtitle]).to eq("#{collection.title} - Subcollections")
     end
   end
+
+  describe "GET #load_collection_from_id" do
+    context "when id does not exist" do
```
Bilka2, line 14: It'd be better to make this about the action, since that is what's getting tested:
```suggestion
  describe "GET #show" do
    context "when collection does not exist" do
```

Bilka2 (approved):
Thank you!

PR #5167, opened 2025-04-28
AO3-6273 Word count blank when posting from draft

ceithir:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6273

## Purpose

Tentative fix of the work word_count sometimes being nil when posting a draft through the post_draft route.

From my understanding, the issue is that `set_word_count`, when doing a read on the database [here](https://github.com/otwcode/otwarchive/blob/master/app/models/work.rb#L831), can sometimes query a version of the db where the first chapter hasn't been updated to `posted=TRUE`, thus counting nil.

My proposed solution is to get the data directly from the object, who I think should be up to date in all cases (all Ruby code is called synchronously). Might be a placebo however, as I'm not sure how to reproduce the issue locally or in tests.

## Testing Instructions

In theory, as per JIRA. However, I'm not sure how often the bug appears.

## Credit

Ceithir (he/him)

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -637,7 +637,9 @@ class WorksController < ApplicationController
 
     # AO3-3498: since a work's word count is calculated in a before_save and the chapter is posted in an after_save,
     # work's word count needs to be updated with the chapter's word count after the chapter is posted
-    @work.set_word_count
+    # AO3-6273 Cannot rely on set_word_count here in a production environment, as it might query an older version of the database
+    # Instead, as the the work in this context is reduced its first chapter, we copy the value directly
+    @work.word_count = @work.first_chapter.word_count
     @work.save
 
     if !@collection.nil? && @collection.moderated?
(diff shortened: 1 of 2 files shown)


Bilka2:
I locally reproduced the issue by placing a breakpoint before this code and then when hitting that breakpoint I directly modified the database to mark the chapter as not posted. The previous code indeed would then result in a `nil` word_count, while this code correctly calculates the word count :tada:

app/controllers/works_controller.rb
```diff
@@ -637,7 +637,9 @@ def post_draft
 
     # AO3-3498: since a work's word count is calculated in a before_save and the chapter is posted in an after_save,
     # work's word count needs to be updated with the chapter's word count after the chapter is posted
-    @work.set_word_count
+    # AO3-6273 Cannot rely on set_word_count here in a production environment, as it might query an older version of the database
+    # Instead, as the the work in this context is reduced its first chapter, we copy the value directly
```
Bilka2:
```suggestion
    # Instead, as the work in this context is reduced its first chapter, we copy the value directly
```

PR #5169, opened 2025-04-28
AO3-6973 Add email when user changes or resets password

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6973

## Purpose

When a user changes or resets their password they will be emailed.

Do the same for admin password resets.

## Credit

Bilka

### DIFF
--- a/app/mailers/archive_devise_mailer.rb
+++ b/app/mailers/archive_devise_mailer.rb
@@ -27,4 +27,10 @@ class ArchiveDeviseMailer < Devise::Mailer
     subject = t("users.mailer.confirmation_instructions.subject", app_name: ArchiveConfig.APP_SHORT_NAME)
     devise_mail(record, :confirmation_instructions, opts.merge(subject: subject))
   end
+
+  def password_change(record, opts = {})
+    @pac_footer = true
+    subject = t("users.mailer.password_change.subject", app_name: ArchiveConfig.APP_SHORT_NAME)
+    devise_mail(record, :password_change, opts.merge(subject: subject))
+  end
 end
--- a/app/mailers/archive_devise_mailer.rb
+++ b/app/mailers/archive_devise_mailer.rb
@@ -30,7 +30,13 @@ class ArchiveDeviseMailer < Devise::Mailer
 
   def password_change(record, opts = {})
     @pac_footer = true
-    subject = t("users.mailer.password_change.subject", app_name: ArchiveConfig.APP_SHORT_NAME)
+    subject = if record.is_a?(Admin)
+                t("admin.mailer.password_change.subject",
+                  app_name: ArchiveConfig.APP_SHORT_NAME)
+              else
+                t("users.mailer.password_change.subject",
+                  app_name: ArchiveConfig.APP_SHORT_NAME)
+              end
     devise_mail(record, :password_change, opts.merge(subject: subject))
   end
 end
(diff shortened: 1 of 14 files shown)


app/mailers/archive_devise_mailer.rb
```diff
@@ -27,4 +27,16 @@ def confirmation_instructions(record, token, opts = {})
     subject = t("users.mailer.confirmation_instructions.subject", app_name: ArchiveConfig.APP_SHORT_NAME)
     devise_mail(record, :confirmation_instructions, opts.merge(subject: subject))
   end
+
+  def password_change(record, opts = {})
+    @pac_footer = true
```
sarken, line 32: Do we want the PAC footer on the admin email? I don't know if we should be telling admins to contact PAC if they're getting unexpected password change emails.

config/locales/mailers/en.yml
```diff
@@ -2,6 +2,19 @@
 en:
   admin:
     mailer:
+      password_change:
+        changed: The password for your AO3 admin account was changed on %{date}.
+        did_not_make_change: If you did not make this change, someone else may have access to your %{app_name} admin account or your OTW email. Please let AD&T and VolCom know immediately so they can take steps to secure your accounts.
```
sarken, line 7: We have the app_name as a variable here, but not in the above text. We should probably pick one (iirc, not passing it works better for Translation).

Bilka2: And passing it works better for other people using otwarchive. Since T doesn't translate `AO3`, I think it should be fine to have it as a parameter - they can move it around etc and generally treat it exactly as if it says AO3.

features/users/authenticate_users.feature
```diff
@@ -105,6 +105,16 @@ Feature: User Authentication
       And 1 email should be delivered to "notsam@example.com"
       And the email should have "Reset your password" in the subject
       And the email to "notsam" should be non-translated
+    When 1 email should be delivered to "sam@example.com"
```
brianjaustin, line 108: Shouldn't this step be a `Then` (well technically an `And` because of the above steps)

spec/mailers/archive_devise_mailer_spec.rb
```diff
@@ -104,4 +104,72 @@
       end
     end
   end
+
+  describe "#password_change" do
+    subject(:email) { ArchiveDeviseMailer.password_change(record) }
+
+    context "when record is user" do
```
brianjaustin, line 111:
```suggestion
    context "when the record is a user" do
```

spec/mailers/archive_devise_mailer_spec.rb
```diff
@@ -104,4 +104,72 @@
...
+
+      describe "text version" do
+        it "has the correct content" do
+          expect(email).to have_text_part_content("Hi #{record.login},")
+          expect(email).to have_text_part_content("The password for your AO3 account was changed")
+        end
+      end
+    end
+
+    context "when record is admin" do
```
brianjaustin, line 143:
```suggestion
    context "when the record is an admin" do
```

config/locales/mailers/en.yml
brianjaustin: Should we generalise "OTW email" to something like "admin account email" or "organization email", since we generalise the app name?

lydia-theda: I'm not sure whether most volunteers know which email their admin account is linked to/what phrase is most likely for them to recognize as meaning "OTW email" (does VolCom ever call it anything else?)

brianjaustin (approved):

PR #5171, opened 2025-04-30
AO3-5953 Actions that aren't no-ops should not be able to be accessed through GET requests

weeklies:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5953

## Purpose

Aside from the issue details:
- Use `button_to` where possible for non-GET actions. This is to provide support for non-JS environments.
- To preserve UI of the footer skin chooser, style the buttons as links. This also fixes some weirdness in the footer links in Low Vision Default on small screens.

## Testing Instructions

See issue.


app/views/chapters/manage.html.erb
```diff
@@ -37,10 +37,11 @@
             <% if @work.pseuds.size > 1 && chapter.pseuds.size > 1 && current_user.is_author_of?(chapter) %>
               <li>
                 <%= link_to ts("Remove Me As Chapter Co-Creator"),
```
weeklies, line 39: This has to be a `link_to`, because this element is already within a form for `update_positions`. There is already a `link_to` for Delete Chapter here so I left it be.

app/views/potential_matches/_in_progress.html.erb
```diff
@@ -5,12 +5,12 @@
...
+  <%= ts("If the progress value below doesn't change when you reload the page
           after about 10 minutes, there may have been a problem. Please check again after an hour, and then contact archive support.") %>
 </p>
 
 <p>
   <%= ts("Potential match generation progress:") %> <strong><%= @progress %>%</strong>
 </p>
 
-<p class="actions" role="button"><%= link_to ts("Cancel Potential Match Generation"), cancel_generate_collection_potential_matches_path(@collection) %></p>
+<span class="actions" role="button"><%= button_to ts("Cancel Potential Match Generation"), cancel_generate_collection_potential_matches_path(@collection), method: :post %></span>
```
sarken, line 16: I'm a little torn on this. The W3 HTML validator flags this because `form` isn't a valid child of `span`. However, it at least displays correctly, which is more than I can say for the `p` tag -- my browser closes before the `form` elements starts, and the HTML validator tells me it can't find the opening `p` tag. 

I guess we can leave it like this for now and rethink the page a bit later. However, we should take the `role` off the span.

app/helpers/works_helper.rb
```diff
@@ -82,12 +82,12 @@ def marked_for_later?(work)
     reading && reading.toread?
   end
 
-  def mark_as_read_link(work)
-    link_to ts("Mark as Read"), mark_as_read_work_path(work)
+  def mark_as_read_button(work)
+    button_to t(".mark_as_read"), mark_as_read_work_path(work), method: :delete
```
sarken, line 86: Because we don't change the text of these buttons based on context, it's probably better to skip lazy lookup here and use `works_helper.mark_as_read_button.mark_as_read` and `works_helper.mark_for_later_button.mark_for_later` instead. That way, the button will always use the text provided in `helpers/en.yml` and you won't need to provide text and `i18n-use` comments every time it's used in a view file.

(I also think you'd be fine with shorter keys, like `works_helper.mark_for_later_button.text` or even just `works_helper.mark_for_later_button`.)

public/stylesheets/site/2.0/08-actions.css
```diff
@@ -50,6 +50,39 @@ button {
   box-sizing: content-box;
 }
 
+.actions form.button_to {
+  line-height: 1.5
+}
```
sarken, line 55: Could you tell me what browser or page this is needed for? It wasn't immediately clear to me from the code, so I took it out while I was checking over all the pages and couldn't find any obvious issues.

sarken (changes requested):
Thank you for working on this! Since it's a really big pull request, I just took the view, helper, and CSS changes -- someone else will need to cover the models, controllers, routes, and tests. You're welcome to wait for the full review before actioning any comments.

public/stylesheets/site/2.0/08-actions.css
weeklies: Removed.

spec/controllers/works/mark_for_later_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+    before do
+      fake_login_known_user(user)
+    end
+
+    describe "POST #mark_for_later" do
+      it "marks the work for later" do
+        post :mark_for_later, params: { id: work.id }
+        it_redirects_to_simple(root_path)
+        expect(flash[:notice]).to include("This work was added to your ")
+        expect(flash[:notice]).to include("Marked for Later list")
```
brianjaustin, line 22: Could you add expectations that the `Reading` record is created to this test, and the reverse for the test below?

spec/controllers/skins_controller_spec.rb
```diff
@@ -143,4 +143,60 @@
...
+        context "when admin has #{role} role" do
+          let(:admin) { create(:admin, roles: [role]) }
+
+          it_behaves_like "user cannot set it"
+        end
+      end
+
+      context "when admin has superadmin role" do
+        it_behaves_like "user cannot set it"
+      end
```
brianjaustin: I think you would need to set the role here as well, similar to the above contexts. However, since `superadmin`is a valid role name, any reason not to just remove the `- %w[superadmin]` bit above? Same observation and question for the one below

features/other_b/skin_public.feature
```diff
@@ -13,19 +13,13 @@ Feature: Public skins
...
-    And I follow "public skin"
+    And I press "public skin"
   Then I should see "The skin public skin has been set. This will last for your current session."
     And the page should have the cached skin "public skin"
-  When I follow "Default"
+  When I press "Default"
   Then I should see "You are now using the default Archive skin again!"
     And the page should not have the cached skin "public skin"
 
-  Scenario: A user can't set an uncached public skin for a session
```
brianjaustin, line 23: Isn't this still needed, since the controller spec only covers admin cases?

config/routes.rb
```diff
@@ -503,9 +510,9 @@
     end
     resources :potential_matches do
       collection do
-        get :generate
-        get :cancel_generate
-        get :regenerate_for_signup
+        post :generate
+        post :cancel_generate
+        post :regenerate_for_signup
```
brianjaustin, line 514: What do you think about the following?
```suggestion
        delete :cancel_generate
        put :regenerate_for_signup
```

These are a bit more RPC-y than REST-y, so I think either is fine to be honest

features/other_b/skin_public.feature
weeklies: This was removed because the [underlying logic](https://github.com/otwcode/otwarchive/pull/5171/files/5514694c83a3e3f1f8fe6d94cfaa15620cfaa384#diff-2e88cbd2b9eeded170fdbe8b3720324ae7740b922c24a6f4c4c1e69218eecd23L190-L193) for setting the public skin no longer works with `visit`. The controller spec was the only way I found that worked in sending a POST request. I've added a case to the spec.

spec/controllers/works/mark_for_later_spec.rb
weeklies: Thanks, this comment made me realise that readings do not get deleted from `mark_as_read` and should actually use <s>POST</s> PATCH instead. I've added a line to make the tests more specific.

config/routes.rb
weeklies: I think it makes more sense to keep with the REST syntax, after looking at how these routes work?

spec/controllers/skins_controller_spec.rb
weeklies: You're right, thank you!

config/routes.rb
brianjaustin: Oh sorry, I meant the concept of issuing a command via POST is RPC-y, my suggestion is to make it more REST-y. But overall the endpoints don't fit well within a REST framework IMO, which is why I said either is fine 🤷

public/stylesheets/masters/low_vision_default/low_vision_default_site_screen_.css
```diff
@@ -112,13 +112,36 @@ pre {
...
+
+#footer button:hover,
+#footer button:focus {
+  box-shadow: 0 0 0 2px #eee;
+}
+
+#footer button:active {
+  outline: 1px dotted;
+}
+
```
sarken: This doesn't quite match the other footer links. For example, the gradient is missing. 

An easier way to make the buttons match the other links would be to use similar code to what we added in the default stylesheet and then just let it inherit the other styles:
```
#footer button {
  cursor: pointer;
  height: unset;
  width: 100%;
  white-space: normal;
    box-sizing: border-box;
}
```

I'd also be inclined to put it after the last of the header-specific styles (so `#header .dropdown .menu a:hover,
#header .dropdown .menu a:focus`) just to reflect the order of the default stylesheets.

public/stylesheets/masters/low_vision_default/low_vision_default_site_screen_.css
```diff
@@ -153,7 +176,9 @@ pre {
 #header .user .dropdown a:focus,
 #footer,
 .actions a:hover,
-.actions a:focus {
+.actions a:focus,
+#footer button:hover,
+#footer button:focus {
```
sarken: Since we're changing a lot of links to buttons, let's make all of our buttons consistent with our link styles and do `input[type="submit"]:hover, input[type="submit"]:focus, .actions button:hover, .actions button:focus` here instead of just applying the style to the footer.

sarken (changes requested):
Two more comments about Low Vision Default! 

If you'd like, you could also add the other changes to `#footer` and `#footer *` from e0cdf34351668f4c959d3a9ebd172ff3212a3f73 to fix the issue with the footer links being a lot wider than the links in the header. If not, we can make an issue to do that some other time.

public/stylesheets/masters/low_vision_default/low_vision_default_site_screen_.css
Bilka2: This no longer catches the footer buttons for the skin chooser, so they do not have the background change on hover. Maybe because they're a button instead of input?

public/stylesheets/masters/low_vision_default/low_vision_default_site_screen_.css
```diff
@@ -142,18 +163,17 @@ pre {
...
-#footer {
-  float: left;
-  padding: 0;
-}
-
 #header,
 #header .open a,
 #header .primary .dropdown a:focus,
 #header .user .dropdown a:focus,
 #footer,
```
weeklies, line 148:
```suggestion
#footer,
#footer button:hover,
#footer button:focus,
```

public/stylesheets/masters/low_vision_default/low_vision_default_site_screen_.css
sarken: Whoops -- these old lines need to be removed. The new ones were supposed to replace them, and leaving them in means the gradient is still missing. Sorry about that!

public/stylesheets/masters/low_vision_default/low_vision_default_site_screen_.css
```diff
@@ -142,18 +163,19 @@ pre {
...
+#footer button:hover,
+#footer button:focus,
 .actions a:hover,
-.actions a:focus {
+.actions a:focus,
+input[type="submit"]:hover,
+input[type="submit"]:focus,
```
sarken, line 152: Can you remove `#footer button:hover, #footer button:focus` above and add `button:hover, button:focus` here, so we style all button elements, not just the ones in the footer or in actions blocks?

PR #5179, opened 2025-05-08
AO3-6304 Prevent searching series by restricted tags when logged out

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6304

## Purpose

Split the tag fields in Elasticsearch into _restricted and _public fields; the former is queried when a logged in user searches/filters, the latter when a guest does the same. The query also replaces `tags_<visibility>` with `tag`, which is then mapped to the correct `tags_<visibility>` to allow existing searches like "tag:Character" but prevent searches that directly reference the visibility-sensitive fields.

Due to the internal ES changes, bookmarkables will need to be reindexed.

Note: I have not added tests yet. A feature test is probably worthwhile, regardless of implementation details, but other tests depend on if we like this approach. Once folks have had a chance to offer feedback, I will go through and fix/add unit tests.

### DIFF
--- a/app/models/concerns/filterable.rb
+++ b/app/models/concerns/filterable.rb
@@ -66,41 +66,30 @@ module Filterable
     (tags.pluck(:name) + filters.pluck(:name)).uniq
   end
 
-  # Index all the filters for pulling works
-  def filter_ids
-    (tags.pluck(:id) + filters.pluck(:id)).uniq
-  end
+  # Restricted tags only really apply to series, as works are either fully
+  # restricted or fully public. We define the various visibility-based methods
+  # to be the same here, and they are overridden in the Series class.
+  %w[restricted public].each do |visibility|
+    alias_method :"tags_#{visibility}", :tag
 
-  # Index only direct filters (non meta-tags) for facets
-  def filters_for_facets
-    @filters_for_facets ||= direct_filters.to_a
-  end
+    # Index all the filters for pulling works
+    define_method("filter_ids_#{visibility}") do
+      (tags.pluck(:id) + filters.pluck(:id)).uniq
+    end
 
-  def rating_ids
-    filters_for_facets.select { |t| t.type == "Rating" }.map(&:id)
-  end
+    # Index only direct filters (non meta-tags) for facets
+    define_method("filters_for_facets_#{visibility}") do
+      cache_variable = "@filters_for_facets_#{visibility}"
+      instance_variable_set(cache_variable, direct_filters.to_a) unless instance_variable_defined?(cache_variable)
+      instance_variable_get(cache_variable)
+    end
 
-  def archive_warning_ids
-    filters_for_facets.select { |t| t.type == "ArchiveWarning" }.map(&:id)
-  end
-
-  def category_ids
-    filters_for_facets.select { |t| t.type == "Category" }.map(&:id)
-  end
-
-  def fandom_ids
-    filters_for_facets.select { |t| t.type == "Fandom" }.map(&:id)
-  end
-
-  def character_ids
-    filters_for_facets.select { |t| t.type == "Character" }.map(&:id)
-  end
-
-  def relationship_ids
-    filters_for_facets.select { |t| t.type == "Relationship" }.map(&:id)
-  end
-
-  def freeform_ids
-    filters_for_facets.select { |t| t.type == "Freeform" }.map(&:id)
+    %w[archive_warning category character fandom freeform rating relationship].each do |tag_type|
+      define_method("#{tag_type}_ids_#{visibility}") do
+        send("filters_for_facets_#{visibility}")
+          .select { |tag| tag.type.to_s == tag_type.camelcase }
+          .map(&:id)
+      end
+    end
   end
 end
--- a/app/models/external_work.rb
+++ b/app/models/external_work.rb
@@ -86,16 +86,21 @@ class ExternalWork < ApplicationRecord
   ######################
 
   def bookmarkable_json
+    methods = %i[creators posted restricted revised_at]
+    %w[restricted public].each do |visibility|
+      methods << :"tags_#{visibility}"
+
+      %w[archive_warning category character fandom filter freeform rating relationship].each do |tag_type|
+        methods << :"#{tag_type}_ids_#{visibility}"
+      end
+    end
+
     as_json(
       root: false,
       only: [
         :title, :summary, :hidden_by_admin, :created_at
       ],
-      methods: [
-        :posted, :restricted, :tag, :filter_ids, :rating_ids,
-        :archive_warning_ids, :category_ids, :fandom_ids, :character_ids,
-        :relationship_ids, :freeform_ids, :creators, :revised_at
-      ]
+      methods: methods
     ).merge(
       language_id: language&.short,
       bookmarkable_type: "ExternalWork",
--- a/app/models/search/bookmarkable_query.rb
+++ b/app/models/search/bookmarkable_query.rb
@@ -13,7 +13,7 @@ class BookmarkableQuery < Query
   # return objects from more than one table, so we need to use a special class
   # that can handle IDs of multiple types.
   def klass
-    'BookmarkableDecorator'
+    "BookmarkableDecorator"
   end
 
   def index_name
@@ -122,7 +122,7 @@ class BookmarkableQuery < Query
       %w[rating archive_warning category fandom character relationship freeform].each do |facet_type|
         aggs[facet_type] = {
           terms: {
-            field: "#{facet_type}_ids"
+            field: restrictable_field_name("#{facet_type}_ids").to_s
           }
         }
       end
@@ -198,7 +198,7 @@ class BookmarkableQuery < Query
   ####################
 
   def complete_filter
-    term_filter(:complete, 'true') if options[:complete].present?
+    term_filter(:complete, "true") if options[:complete].present?
   end
 
   def language_filter
@@ -206,9 +206,9 @@ class BookmarkableQuery < Query
   end
 
   def filter_id_filter
-    if filter_ids.present?
-      filter_ids.map { |filter_id| term_filter(:filter_ids, filter_id) }
-    end
+    return if filter_ids.blank?
+
+    filter_ids.map { |filter_id| term_filter(restrictable_field_name(:filter_ids), filter_id) }
   end
 
   # The date filter on the bookmarkable (i.e. when the bookmarkable was last
@@ -223,42 +223,43 @@ class BookmarkableQuery < Query
   # Note that this is used as an exclusion filter, not an inclusion filter, so
   # the boolean is flipped from the way you might expect.
   def unposted_filter
-    term_filter(:posted, 'false')
+    term_filter(:posted, "false")
   end
 
   # Exclude items hidden by admin from bookmarkable search results.
   # Note that this is used as an exclusion filter, not an inclusion filter, so
   # the boolean is flipped from the way you might expect.
   def hidden_filter
-    term_filter(:hidden_by_admin, 'true')
+    term_filter(:hidden_by_admin, "true")
   end
 
   # Exclude restricted works/series when the user isn't logged in.
   # Note that this is used as an exclusion filter, not an inclusion filter, so
   # the boolean is flipped from the way you might expect.
   def restricted_filter
-    term_filter(:restricted, 'true') unless include_restricted?
+    term_filter(:restricted, "true") unless include_restricted?
   end
 
   def tag_exclusion_filter
-    if exclusion_ids.present?
-      terms_filter(:filter_ids, exclusion_ids)
-    end
+    return if exclusion_ids.blank?
+
+    terms_filter(restrictable_field_name(:filter_ids), exclusion_ids)
   end
 
   # This filter is used to restrict our results to only include bookmarkables
-  # whose "tag" text matches all of the tag names in included_tag_names. This
-  # is useful when the user enters a non-existent tag, which would be discarded
+  # whose "tags_(restricted|public)" text matches all of the tag names in included_tag_names.
+  # This is useful when the user enters a non-existent tag, which would be discarded
   # by the TaggableQuery.filter_ids function.
   def named_tag_inclusion_filter
     return if included_tag_names.blank?
-    match_filter(:tag, included_tag_names.join(" "))
+
+    match_filter(restrictable_field_name(:tags), included_tag_names.join(" "))
   end
 
   # This set of filters is used to prevent us from matching any bookmarkables
-  # whose "tag" text matches one of the passed-in tag names. This is useful
-  # when the user enters a non-existent tag, which would be discarded by the
-  # TaggableQuery.exclusion_ids function.
+  # whose "tags_(restricted|public)" text matches one of the passed-in tag names.
+  # This is useful when the user enters a non-existent tag, which would be discarded
+  # by the TaggableQuery.exclusion_ids function.
   #
   # Note that we separate these into different filters to get the logic of tag
   # exclusion right: if we're excluding "A B" and "C D", we want the query to
@@ -266,7 +267,7 @@ class BookmarkableQuery < Query
   # match query.
   def named_tag_exclusion_filter
     excluded_tag_names.map do |tag_name|
-      match_filter(:tag, tag_name)
+      match_filter(restrictable_field_name(:tags), tag_name)
     end
   end
 
@@ -283,5 +284,11 @@ class BookmarkableQuery < Query
     # Use fetch instead of || here to make sure that we don't accidentally
     # override a deliberate choice not to show restricted bookmarks.
     options.fetch(:show_restricted, User.current_user.present?)
+  end
+
+  private
+
+  def restrictable_field_name(field_name)
+    :"#{field_name}_#{include_restricted? ? "restricted" : "public"}"
   end
 end
--- a/app/models/search/bookmarkable_query.rb
+++ b/app/models/search/bookmarkable_query.rb
@@ -92,7 +92,25 @@ class BookmarkableQuery < Query
 
   def bookmarkable_query_text
     query_text = (options[:bookmarkable_query] || "").dup
-    escape_slashes(query_text.strip)
+    escape_tags_field(escape_slashes(query_text.strip))
+  end
+
+  # Internally, we use tags_(public|restricted) to denote which tags are visible to everyone
+  # (public) and which are only available to logged in users (restricted). We do not, however,
+  # want users to be able to query those fields directly. To avoid that, we replace tags_<visibility>
+  # with the "tag" query. Then, we replace "tag" with the appropriate visibility-based field.
+  #
+  # Examples:
+  # When searching as a guest,
+  # escape_tags_field("tags_public:1234") => tags_public:1234
+  # escape_tags_field("tags_restricted:1234") => tags_public:1234
+  # escape_tags_field("tag:1234") => tags_public:1234
+  # When searching as a registered user,
+  # escape_tags_field("tags_public:1234") => tags_restricted:1234
+  # escape_tags_field("tags_restricted:1234") => tags_restricted:1234
+  # escape_tags_field("tag:1234") => tags_restricted:1234
+  def escape_tags_field(query)
+    query.gsub(/tags_\w+/, "tag").gsub("tag", restrictable_field_name(:tags).to_s)
   end
 
   ####################
--- a/app/models/series.rb
+++ b/app/models/series.rb
@@ -69,18 +69,20 @@ class Series < ApplicationRecord
   end
 
   # Get the filters for the works in this series
-  def filters
+
+  def filters_restricted
     Tag.joins("JOIN filter_taggings ON tags.id = filter_taggings.filter_id
                JOIN works ON works.id = filter_taggings.filterable_id
-               JOIN serial_works ON serial_works.work_id = works.id").
-        where("serial_works.series_id = #{self.id} AND
-               works.posted = 1 AND
-               filter_taggings.filterable_type = 'Work'").
-        group("tags.id")
+               JOIN serial_works ON serial_works.work_id = works.id")
+       .where(serial_works: { series_id: self.id }, works: { posted: true }, filter_taggings: { filterable_type: "Work" })
+       .group("tags.id")
   end
 
-  def direct_filters
-    filters.where("filter_taggings.inherited = 0")
+  def filters_public
+    Tag.joins("JOIN filter_taggings ON tags.id = filter_taggings.filter_id
+               JOIN works ON works.id = filter_taggings.filterable_id
+               JOIN serial_works ON serial_works.work_id = works.id")
+       .where(serial_works: { series_id: self.id }, works: { posted: true, restricted: false }, filter_taggings: { filterable_type: "Work" })
   end
 
   # visibility aped from the work model
@@ -194,10 +196,6 @@ class Series < ApplicationRecord
     works.collect(&:fandoms).flatten.compact.uniq.sort
   end
 
-  def author_tags
-    self.work_tags.select{|t| t.type == "Relationship"}.sort + self.work_tags.select{|t| t.type == "Character"}.sort + self.work_tags.select{|t| t.type == "Freeform"}.sort
-  end
-
   def tag_groups
     self.work_tags.group_by { |t| t.type.to_s }
   end
@@ -224,17 +222,22 @@ class Series < ApplicationRecord
   ######################
 
   def bookmarkable_json
+    methods = %i[creators posted revised_at word_count work_types]
+    %w[restricted public].each do |visibility|
+      methods << :"tags_#{visibility}"
+
+      %w[archive_warning category character fandom filter freeform rating relationship].each do |tag_type|
+        methods << :"#{tag_type}_ids_#{visibility}"
+      end
+    end
+
     as_json(
       root: false,
       only: [
         :title, :summary, :hidden_by_admin, :restricted, :created_at,
         :complete
       ],
-      methods: [
-        :revised_at, :posted, :tag, :filter_ids, :rating_ids,
-        :archive_warning_ids, :category_ids, :fandom_ids, :character_ids,
-        :relationship_ids, :freeform_ids, :creators,
-        :word_count, :work_types]
+      methods: methods
     ).merge(
       language_id: language&.short,
       anonymous: anonymous?,
@@ -265,39 +268,40 @@ class Series < ApplicationRecord
   alias_method :posted?, :posted
 
   # Simple name to make it easier for people to use in full-text search
-  def tag
-    (work_tags + filters).uniq.map{ |t| t.name }
+  def tags_restricted
+    (work_tags.pluck(:name) + filters_restricted.pluck(:name)).uniq
+  end
+
+  def tags_public
+    (work_tags.where(works: { restricted: false }).pluck(:name) + filters_public.pluck(:name)).uniq
   end
 
   # Index all the filters for pulling works
-  def filter_ids
-    (work_tags.pluck(:id) + filters.pluck(:id)).uniq
+  def filter_ids_public
+    (work_tags.pluck(:id) + filters_public.pluck(:id)).uniq
   end
 
-  # Index only direct filters (non meta-tags) for facets
-  def filters_for_facets
-    @filters_for_facets ||= direct_filters
+  def filter_ids_restricted
+    (work_tags.where(works: { restricted: false }).pluck(:id) + filters_restricted.pluck(:id)).uniq
   end
-  def rating_ids
-    filters_for_facets.select{ |t| t.type.to_s == 'Rating' }.map{ |t| t.id }
-  end
-  def archive_warning_ids
-    filters_for_facets.select{ |t| t.type.to_s == 'ArchiveWarning' }.map{ |t| t.id }
-  end
-  def category_ids
-    filters_for_facets.select{ |t| t.type.to_s == 'Category' }.map{ |t| t.id }
-  end
-  def fandom_ids
-    filters_for_facets.select{ |t| t.type.to_s == 'Fandom' }.map{ |t| t.id }
-  end
-  def character_ids
-    filters_for_facets.select{ |t| t.type.to_s == 'Character' }.map{ |t| t.id }
-  end
-  def relationship_ids
-    filters_for_facets.select{ |t| t.type.to_s == 'Relationship' }.map{ |t| t.id }
-  end
-  def freeform_ids
-    filters_for_facets.select{ |t| t.type.to_s == 'Freeform' }.map{ |t| t.id }
+
+  %w[restricted public].each do |tag_visibility|
+    # Index only direct filters (non meta-tags) for facets
+    define_method("direct_filters_#{tag_visibility}") do
+      send("filters_#{tag_visibility}").where(filter_taggings: { inherited: false })
+    end
+
+    define_method("filters_for_facets_#{tag_visibility}") do
+      cache_variable = "@filters_for_facets_#{tag_visibility}"
+      instance_variable_set(cache_variable, send("direct_filters_#{tag_visibility}")) unless instance_variable_defined?(cache_variable)
+      instance_variable_get(cache_variable)
+    end
+
+    %w[archive_warning category character fandom freeform rating relationship].each do |tag_type|
+      define_method("#{tag_type}_ids_#{tag_visibility}") do
+        send("filters_for_facets_#{tag_visibility}").select { |tag| tag.type.to_s == tag_type.camelcase }.map(&:id)
+      end
+    end
   end
 
   def creators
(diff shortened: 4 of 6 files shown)


lydia-theda:
Not sure if this is covered already or if you'll have to account for it specifically, but since tags on a work hidden by admin should not be searchable by users or guests, searching/indexing series tags also shouldn't include tags only used on hidden works in the series.

app/models/search/bookmarkable_query.rb
```diff
@@ -92,7 +92,25 @@ def general_query
...
+  # Examples:
+  # When searching as a guest,
+  # escape_tags_field("tags_public:1234") => tags_public:1234
+  # escape_tags_field("tags_restricted:1234") => tags_public:1234
+  # escape_tags_field("tag:1234") => tags_public:1234
+  # When searching as a registered user,
+  # escape_tags_field("tags_public:1234") => tags_restricted:1234
+  # escape_tags_field("tags_restricted:1234") => tags_restricted:1234
+  # escape_tags_field("tag:1234") => tags_restricted:1234
+  def escape_tags_field(query)
```
Bilka2: I'm a bit lost in the code here still but this method feels like the crux of my questions: What about all queries that don't start with "tags", like doing "fandom_ids_restricted: 0000" in the same way you can currently do "fandom_ids: 0000"? [Example query](https://test.archiveofourown.org/bookmarks?bookmark_search%5Bsort_column%5D=created_at&bookmark_search%5Bother_tag_names%5D=&bookmark_search%5Bother_bookmark_tag_names%5D=&bookmark_search%5Bexcluded_tag_names%5D=&bookmark_search%5Bexcluded_bookmark_tag_names%5D=&bookmark_search%5Bbookmarkable_query%5D=fandom_ids%3A6670&bookmark_search%5Bbookmark_query%5D=&bookmark_search%5Blanguage_id%5D=&bookmark_search%5Brec%5D=0&bookmark_search%5Bwith_notes%5D=0&commit=Sort+and+Filter&user_id=testy) of what I mean, it filters to only bookmarks in the Testing fandom.

Like, how can we actually safely protect the fields from getting accessed when they shouldnt be? And is there any other solution than trying to sanitize the query like here, e.g. a completely different way to map those fields, or something like [field level security](https://www.elastic.co/docs/deploy-manage/users-roles/cluster-or-deployment-auth/controlling-access-at-document-field-level)?

brianjaustin: Yep, you're right that the other fields need to be sanitized too. In terms of other ways of doing this, I haven't been able to find really any other options besides this or the field-level security you linked. I like the second approach, _but_ it would require some pretty deep changes to how we query Elasticsearch (both on the infrastructure side -- adding the roles, creating a user for the app to query as those roles, and provisioning the user's API key onto the app servers) as well as the code (sending the API key). So I'm not sure it's good to increase the scope here by that much.

(Noting for self: there's also a need to pass the `es-security-runas-user` header to switch roles, I think. That's a little less clean than idea, but we can only get so nice when the server is running queries on behalf of end users :shrug:)

There's also the likelihood that field-level security would result in an error if someone uses a field they shouldn't, but I think that's not a big problem because at that point you should know what you're getting into. Maybe a follow-up ticket to investigate/implement that here and for series?

app/models/search/bookmarkable_query.rb
```diff
@@ -284,4 +306,10 @@ def include_restricted?
     # override a deliberate choice not to show restricted bookmarks.
     options.fetch(:show_restricted, User.current_user.present?)
   end
+
+  private
+
+  def restrictable_field_name(field_name)
+    :"#{field_name}_#{include_restricted? ? "general" : "public"}"
+  end
```
Bilka2, line 346: In collections and pseuds we prefix the general/public, instead of suffix. Do we want to do the same here or is awkward because of the field names in use here?

app/models/concerns/filterable.rb
```diff
@@ -66,41 +66,37 @@ def tag
...
-
-  def fandom_ids
-    filters_for_facets.select { |t| t.type == "Fandom" }.map(&:id)
-  end
+  # Restricted tags (which are in the general list but not the public one) only
+  # really apply to series, as works are either fully restricted or fully public.
+  # We define the various visibility-based methods to be the same here, and they are
+  # overridden in the Series class.
+  %w[general public].each do |visibility|
+    alias_method :"tags_#{visibility}", :tag
```
Bilka2: How come the original `tag` method is still available above, instead of removed? Since the comment on the method definition says it's for making it easier to use in full tag search, I would kind of expect it to no longer be needed, similar to fandom_ids etc

app/models/concerns/filterable.rb
```diff
@@ -66,41 +66,37 @@ def tag
...
+    # Index only direct filters (non meta-tags) for facets
+    define_method("filters_for_facets_#{visibility}") do
+      cache_variable = "@filters_for_facets_#{visibility}"
+      instance_variable_set(cache_variable, direct_filters.to_a) unless instance_variable_defined?(cache_variable)
+      instance_variable_get(cache_variable)
+    end
 
-  def relationship_ids
-    filters_for_facets.select { |t| t.type == "Relationship" }.map(&:id)
+    %w[archive_warning category character fandom freeform rating relationship].each do |tag_type|
```
Bilka2: Could this use `Tag::FILTERS` instead (with `underscore` in the right spot)? It's what's used in the spec, and the types and comment about what it does matches this list here

app/models/concerns/filterable.rb
```diff
@@ -66,41 +66,37 @@ def tag
...
+          .select { |tag| tag.type.to_s == tag_type.camelcase }
+          .map(&:id)
+      end
+    end
   end
 
-  def freeform_ids
-    filters_for_facets.select { |t| t.type == "Freeform" }.map(&:id)
+  # For filterables like Work, tags only on restricted works are filtered out by virtue of the filterable having
+  # restricted set to true. Therefore, we just want a generic "<tag_type>_ids" method as well.
```
Bilka2: These could be removed if we were willing to update the WorkIndexer mapping to work the same as the bookmarkable_json mapping, right? But because we don't want to do that, we keep them here?

Would it theoretically work to do something like this in the work mapping and would it be helpful to avoid these methods here?
```
.merge(
  archive_warning_ids: archive_warning_ids_general,
  category_ids: category_ids_general,
  ....
)
```

That way the mapping would look the same to ES but we wouldn't need the methods here (maybe?)

Same for CollectionIndexer, I suppose..?

app/models/external_work.rb
```diff
@@ -86,16 +86,21 @@ def should_reindex_pseuds?
   ######################
 
   def bookmarkable_json
+    methods = %i[creators posted restricted revised_at]
+    %w[general public].each do |visibility|
+      methods << :"tags_#{visibility}"
+
+      %w[archive_warning category character fandom filter freeform rating relationship].each do |tag_type|
```
Bilka2: I think here we could also use Tag::FILTERS instead, with filter in a separate line because I think it's important to make clear it's not only tag types but also the filter_ids, which are a different thing (they have metatags etc)

app/models/concerns/filterable.rb
```diff
@@ -66,41 +66,37 @@ def tag
...
-    filters_for_facets.select { |t| t.type == "Category" }.map(&:id)
-  end
-
-  def fandom_ids
-    filters_for_facets.select { |t| t.type == "Fandom" }.map(&:id)
-  end
+  # Restricted tags (which are in the general list but not the public one) only
+  # really apply to series, as works are either fully restricted or fully public.
+  # We define the various visibility-based methods to be the same here, and they are
+  # overridden in the Series class.
```
Bilka2:
```suggestion
  # overridden in the Series class to account for tags on restricted works in the series.
```

docker-compose.yml
```diff
@@ -88,6 +88,18 @@ services:
     # Make `docker compose attach web` work for debugging
     stdin_open: true
     tty: true
+  workers:
```
Bilka2: What do you think about making this a separate PR/Jira issue combined with making it a dependency of the web container? I think it's quite confusing to people when workers aren't started. But maybe it hurts something to start them here? I think if we manage the containers correctly, it should hopefully work out so with the init.sh script it also fills the indexes (because it runs reindexing and the workers are running)

app/models/search/bookmarkable_query.rb
```diff
@@ -92,7 +92,28 @@ def general_query
 
   def bookmarkable_query_text
     query_text = (options[:bookmarkable_query] || "").dup
-    escape_slashes(query_text.strip)
+    escape_restrictable_fields(escape_slashes(query_text.strip))
+  end
+
+  # Internally, we use <fieldname>_(public|general) to denote certain fields with data from all works (general)
+  # versus fields with data only from works visible to guests (public). We do not, however,  want users to be able
```
Bilka2:
```suggestion
  # versus fields with data only from works visible to guests (public). We do not, however, want users to be able
```

app/models/search/bookmarkable_query.rb
```diff
@@ -92,7 +92,28 @@ def general_query
...
+  # Examples:
+  # When searching as a guest,
+  # escape_tags_field("tags_public:1234") => tags_public:1234
+  # escape_tags_field("fandom_ids_general:1234") => fandom_ids_public:1234
+  # escape_tags_field("tag:1234") => tags_public:1234
+  # When searching as a registered user,
+  # escape_tags_field("tags_public:1234") => tags_general:1234
+  # escape_tags_field("character_ids_general:1234") => character_ids_general:1234
+  # escape_tags_field("tag:1234") => tags_general:1234
+  def escape_restrictable_fields(query)
```
Bilka2, line 113: So these are nice to have as a comment, but would it be doable to also have the examples as a test? That way we know it definitely works like this

Bilka2, line 113: Doesn't this also have to do `freeform_ids` -> `freeform_ids_public` (for example)? Since users are currently used to searching by the former. From what I see in the regex, it currently only converts field names that already have general/public suffixes

app/models/search/bookmarkable_query.rb
```diff
@@ -223,50 +244,51 @@ def date_filter
...
-    end
+    return if exclusion_ids.blank?
+
+    terms_filter(restrictable_field_name(:filter_ids), exclusion_ids)
   end
 
   # This filter is used to restrict our results to only include bookmarkables
-  # whose "tag" text matches all of the tag names in included_tag_names. This
-  # is useful when the user enters a non-existent tag, which would be discarded
+  # whose "tags_(restricted|public)" text matches all of the tag names in included_tag_names.
```
Bilka2:
```suggestion
  # whose "tags_(general|public)" text matches all of the tag names in included_tag_names.
```

app/models/search/bookmarkable_query.rb
```diff
@@ -223,50 +244,51 @@ def date_filter
...
-    match_filter(:tag, included_tag_names.join(" "))
+
+    match_filter(restrictable_field_name(:tags), included_tag_names.join(" "))
   end
 
   # This set of filters is used to prevent us from matching any bookmarkables
-  # whose "tag" text matches one of the passed-in tag names. This is useful
-  # when the user enters a non-existent tag, which would be discarded by the
-  # TaggableQuery.exclusion_ids function.
+  # whose "tags_(restricted|public)" text matches one of the passed-in tag names.
```
Bilka2:
```suggestion
  # whose "tags_(general|public)" text matches one of the passed-in tag names.
```

app/models/series.rb
```diff
@@ -6,6 +6,7 @@ class Series < ApplicationRecord
   has_many :serial_works, dependent: :destroy
   has_many :works, through: :serial_works
   has_many :work_tags, -> { distinct }, through: :works, source: :tags
+  has_many :unhidden_work_tags, -> { where(works: { hidden_by_admin: false }) }, through: :works, source: :tags
```
Bilka2: Couldn't this be a scope that reuses `work_tags`?

app/models/series.rb
```diff
@@ -69,18 +70,25 @@ def works_in_order
...
+  def filters_general
     Tag.joins("JOIN filter_taggings ON tags.id = filter_taggings.filter_id
                JOIN works ON works.id = filter_taggings.filterable_id
-               JOIN serial_works ON serial_works.work_id = works.id").
-        where("serial_works.series_id = #{self.id} AND
-               works.posted = 1 AND
-               filter_taggings.filterable_type = 'Work'").
-        group("tags.id")
+               JOIN serial_works ON serial_works.work_id = works.id")
+      .where(serial_works: { series_id: self.id },
```
Bilka2, line 77:
```suggestion
      .where(serial_works: { series: self },
```
?

app/models/series.rb
Bilka2: Also this will likely be updated for drafts with https://otwarchive.atlassian.net/browse/AO3-6302 so maybe you could pick a different name here that would be applicable for both (visible_to_registered_user or something?)

... which overlaps with the `where` used in `tag_groups` in #5184. I think it would make sense to just combine these into the same scope across both PRs? But then you'd be stealing AO3-6302 away from ömer too :/

app/models/series.rb
```diff
@@ -265,39 +274,41 @@ def posted
...
+    define_method("filters_for_facets_#{tag_visibility}") do
+      cache_variable = "@filters_for_facets_#{tag_visibility}"
+      instance_variable_set(cache_variable, send("direct_filters_#{tag_visibility}")) unless instance_variable_defined?(cache_variable)
+      instance_variable_get(cache_variable)
+    end
+
+    %w[archive_warning category character fandom freeform rating relationship].each do |tag_type|
+      define_method("#{tag_type}_ids_#{tag_visibility}") do
+        send("filters_for_facets_#{tag_visibility}").select { |tag| tag.type.to_s == tag_type.camelcase }
+          .map(&:id)
```
Bilka2:
```suggestion
          .pluck(:id)
```

features/other_a/preferences_edit.feature
```diff
@@ -301,7 +301,7 @@ Feature: Edit preferences
     # Check hidden tags on the blurb
     Then I should see "Asteroid Blues"
       And I should not see "No Archive Warnings Apply" within "li.warnings"
-      And I should not see "Ed is a sweetie"
+      And I should not see "Ed is a sweetie" within "li.freeforms"
```
Bilka2: How did this change end up here?

spec/models/search/bookmark_query_spec.rb
```diff
@@ -114,11 +114,26 @@ def find_parent_filter(query_list)
...
+    context "when querying as a guest" do
+      it "allows filtering by unrestricted bookmarkable tags" do
+        q = BookmarkQuery.new(parent: tag)
+        parent = find_parent_filter(q.generated_query.dig(:query, :bool, :must))
+        expect(parent.dig(:has_parent, :query, :bool, :filter)).to \
+          include({ term: { filter_ids_public: tag.id } })
+      end
+    end
+
+    context "when querying as a registered user" do
```
Bilka2, line 169: Worth it to test admin here too?

spec/models/series_spec.rb
```diff
@@ -110,4 +152,46 @@
...
+    it "includes tags on unrestricted works" do
+      series.works = [unrestricted_work]
+      series.reload
+      expect(series.tags_general).to include(*unrestricted_work.tags.pluck(:name))
+    end
+
+    it "includes tags on restricted works" do
+      series.works = [restricted_work]
+      series.reload
+      expect(series.tags_general).to include(*unrestricted_work.tags.pluck(:name))
```
Bilka2:
```suggestion
      expect(series.tags_general).to include(*restricted_work.tags.pluck(:name))
```

Bad sign though. Could you give the three works unique (canonical for filters_*?) tags like you did in #5184?

spec/models/series_spec.rb
Bilka2, line 1: Would be good to have tests for the filter_ids too (they're currently where the draft tag bug comes from, since they have work_tags while the facets dont)

and it might be good to also test the facets / ids and direct_filters. or at least direct_filters

Bilka2 (changes requested):
Hey, it's me, the reviewer. I have reviewed

app/models/search/bookmarkable_query.rb
brianjaustin: Not more awkward, just that I did this one first based on what I preferred at the time. Consistency is fine, though :+1:

docker-compose.yml
brianjaustin: Sounds good! https://github.com/otwcode/otwarchive/pull/5395

app/models/series.rb
brianjaustin: I don't think that will work...
```
Mysql2::Error: Unknown column 'serial_works.series' in 'where clause' (ActiveRecord::StatementInvalid)
      ./app/models/series.rb:281:in 'Series#general_tags'
      ./app/models/series.rb:243:in 'Series#bookmarkable_json'
      ./app/models/search/bookmarkable_indexer.rb:28:in 'BookmarkableIndexer#document'
      ./app/models/search/indexer.rb:188:in 'block in Indexer#batch'
      ./app/models/search/indexer.rb:184:in 'Array#each'
      ./app/models/search/indexer.rb:184:in 'Indexer#batch'
      ./app/models/search/indexer.rb:197:in 'Indexer#index_documents'
```

app/models/concerns/filterable.rb
brianjaustin: It looks like we don't need to do that for collections, just works/series. Reasonable enough, updated :heavy_check_mark:

app/models/concerns/filterable.rb
brianjaustin: Similar to the other thing you mentioned, this was used by the work indexer. I've removed it as part of addressing the other comment.

app/models/search/bookmarkable_query.rb
```diff
@@ -1,4 +1,6 @@
 class BookmarkableQuery < Query
+  RESTRICABLE_FIELDS = Regexp.union(Tag::FILTERS.map(&:underscore) + %w[filter tags])
```
Bilka2:
```suggestion
  RESTRICTABLE_FIELDS = Regexp.union(Tag::FILTERS.map(&:underscore) + %w[filter tags])
```

app/models/search/bookmarkable_query.rb
```diff
@@ -284,4 +308,12 @@ def include_restricted?
     # override a deliberate choice not to show restricted bookmarks.
     options.fetch(:show_restricted, User.current_user.present?)
   end
+
+  private
+
+  def restrictable_field_name(field_name)
+    return field_name.to_sym unless field_name.match?(RESTRICABLE_FIELDS)
```
Bilka2:
```suggestion
    return field_name.to_sym unless field_name.match?(RESTRICTABLE_FIELDS)
```

app/models/search/bookmarkable_query.rb
```diff
@@ -92,7 +94,28 @@ def general_query
...
+  # escape_tags_field("public_tags:1234") => public_tags:1234
+  # escape_tags_field("general_fandom_ids:1234") => public_fandom_ids:1234
+  # escape_tags_field("tag:1234") => public_tags:1234
+  # When searching as a registered user,
+  # escape_tags_field("public_tags:1234") => general_tags:1234
+  # escape_tags_field("general_character_ids:1234") => general_character_ids:1234
+  # escape_tags_field("tag:1234") => general_tags:1234
+  def escape_restrictable_fields(query)
+    # Special-case for the "tag" convenience field name first, then sanitize visibility level.
+    query.gsub("tag:", "public_tags:").gsub(/(?:(?:public|general)_)?([^\s]+):/) do
```
Bilka2: First thought:
```suggestion
    query.gsub("tag:", "public_tags:").gsub(/(?:(?:public|general)_)?(\w+):/) do
```
Second thought:
This could use `RESTRICTABLE_FIELDS` instead of `\w` to match less fields. E.g. right now this also matches `foo:123` which we don't really need. (And then we don't need to run that regex in restrictable_field_name)

app/models/search/bookmarkable_query.rb
```diff
@@ -92,7 +94,28 @@ def general_query
 
   def bookmarkable_query_text
     query_text = (options[:bookmarkable_query] || "").dup
-    escape_slashes(query_text.strip)
+    escape_restrictable_fields(escape_slashes(query_text.strip))
+  end
+
+  # Internally, we use (public|general)_<fieldname> to denote certain fields with data from all works (general)
+  # versus fields with data only from works visible to guests (public). We do not, however, want users to be able
+  # to query those fields directly. To avoid that, we replace <fieldname>_<visibility> with the appropriate visibility
```
Bilka2:
```suggestion
  # to query those fields directly. To avoid that, we replace <visibility>_<fieldname> or <fieldname> with the appropriate visibility
```

spec/models/search/bookmark_query_spec.rb
```diff
@@ -114,11 +114,46 @@ def find_parent_filter(query_list)
...
+      end
+    end
+
+    it "allows you to filter by bookmarkable language" do
+      q = BookmarkQuery.new(language_id: "ig")
       parent = find_parent_filter(q.generated_query.dig(:query, :bool, :must))
       expect(parent.dig(:has_parent, :query, :bool, :filter)).to \
-        include({ term: { filter_ids: tag.id } })
+        include({ term: { "language_id.keyword": "ig" } })
+    end
```
Bilka2: Duplicated from the spec below "when querying as a registered user" > "allows filtering by restricted bookmarkable tags"

spec/models/search/bookmarkable_query_spec.rb
```diff
@@ -1,6 +1,70 @@
...
+      Tag::FILTERS.map(&:underscore).each do |tag_type|
+        it "does not change 'general_#{tag_type}_ids'" do
+          expect(query.escape_restrictable_fields("general_#{tag_type}_ids:foo")).to eq("general_#{tag_type}_ids:foo")
+        end
+
+        it "replaces 'public_#{tag_type}_ids' with 'general_#{tag_type}'" do
+          expect(query.escape_restrictable_fields("public_#{tag_type}_ids:foo")).to eq("general_#{tag_type}_ids:foo")
+        end
+
+        it "replaces '#{tag_type}_ids' with 'general_#{tag_type}'" do
```
Bilka2:
```suggestion
        it "replaces '#{tag_type}_ids' with 'general_#{tag_type}_ids'" do
```

spec/models/search/bookmarkable_query_spec.rb
```diff
@@ -1,6 +1,70 @@
...
+      it "does not change 'general_tags'" do
+        expect(query.escape_restrictable_fields("general_tags:foo")).to eq("general_tags:foo")
+      end
+
+      Tag::FILTERS.map(&:underscore).each do |tag_type|
+        it "does not change 'general_#{tag_type}_ids'" do
+          expect(query.escape_restrictable_fields("general_#{tag_type}_ids:foo")).to eq("general_#{tag_type}_ids:foo")
+        end
+
+        it "replaces 'public_#{tag_type}_ids' with 'general_#{tag_type}'" do
```
Bilka2:
```suggestion
        it "replaces 'public_#{tag_type}_ids' with 'general_#{tag_type}_ids'" do
```

spec/models/search/bookmarkable_query_spec.rb
```diff
@@ -1,6 +1,70 @@
...
 
 describe BookmarkableQuery do
+  describe "#escape_restrictable_fields" do
+    context "when restricted tags are shown" do
+      let(:bookmark_query) { BookmarkQuery.new(show_restricted: true) }
+      let(:query) { bookmark_query.bookmarkable_query }
+
+      it "replaces 'tag' with 'general_tags'" do
+        expect(query.escape_restrictable_fields("tag:foo")).to eq("general_tags:foo")
+      end
```
Bilka2, line 11: Can you show `tags` being converted as well?

spec/models/search/bookmarkable_query_spec.rb
```diff
@@ -1,6 +1,70 @@
...
+      it "does not change 'public_tags'" do
+        expect(query.escape_restrictable_fields("public_tags:foo")).to eq("public_tags:foo")
+      end
+
+      Tag::FILTERS.map(&:underscore).each do |tag_type|
+        it "does not change 'public_#{tag_type}_ids'" do
+          expect(query.escape_restrictable_fields("public_#{tag_type}_ids:foo")).to eq("public_#{tag_type}_ids:foo")
+        end
+
+        it "replaces 'general_#{tag_type}_ids' with 'public_#{tag_type}'" do
```
Bilka2:
```suggestion
        it "replaces 'general_#{tag_type}_ids' with 'public_#{tag_type}_ids'" do
```

spec/models/search/bookmarkable_query_spec.rb
```diff
@@ -1,6 +1,70 @@
...
+          expect(query.escape_restrictable_fields("general_#{tag_type}_ids:foo")).to eq("public_#{tag_type}_ids:foo")
+        end
+
+        it "replaces '#{tag_type}_ids' with 'public_#{tag_type}'" do
```
Bilka2:
```suggestion
        it "replaces '#{tag_type}_ids' with 'public_#{tag_type}_ids'" do
```

spec/models/search/bookmarkable_query_spec.rb
```diff
@@ -70,5 +162,67 @@
       expect(child_filter.dig(:has_child, :query, :bool, :filter)).to include(term: { user_id: 5 })
       expect(child_filter.dig(:has_child, :query, :bool, :must_not)).to include(terms: { tag_ids: [666] })
     end
+
+    context "when querying as a user" do
```
Bilka2, line 179: Could these three querying context blocks also have a `fieldname_ids` -> `visibility_fieldname_ids` example please

spec/models/series_spec.rb
```diff
@@ -110,4 +154,147 @@
...
+    end
+
+    it "includes tags on restricted works" do
+      series.works = [restricted_work]
+      series.reload
+      expect(series.general_filter_ids).to include(*restricted_work.tags.canonical.pluck(:id))
+    end
+
+    it "does not include tags on works hidden by an admin" do
+      hidden_work = create(:work, hidden_by_admin: true)
```
Bilka2: This needs a canonical tag as well

spec/models/series_spec.rb
```diff
@@ -80,6 +82,48 @@
...
+    end
+
+    it "includes tags on restricted works" do
+      series.works = [restricted_work]
+      series.reload
+      expect(series.general_filters).to include(*restricted_work.tags.canonical)
+    end
+
+    it "does not include tags on works hidden by an admin" do
+      hidden_work = create(:work, hidden_by_admin: true)
```
Bilka2: Just to be safe, can you also give this a canonical tag?

spec/models/series_spec.rb
```diff
@@ -110,4 +154,147 @@
...
+
+    it "includes direct tags on restricted works" do
+      series.works = [restricted_work]
+      series.reload
+      expect(series.general_direct_filters).to include(*restricted_work.tags.canonical)
+      expect(series.general_direct_filters).not_to include(meta)
+    end
+
+    it "does not include tags on works hidden by an admin" do
+      hidden_work = create(:work, hidden_by_admin: true)
```
Bilka2: And another canonical tag on here too please

spec/models/series_spec.rb
```diff
@@ -110,4 +154,147 @@
...
+    before do
+      unrestricted_character.meta_tags << meta
+      restricted_character.meta_tags << meta
+    end
+
+    it "includes direct tags on unrestricted works" do
+      series.works = [unrestricted_work]
+      series.reload
+      expect(series.public_direct_filters).to include(*unrestricted_work.tags.canonical)
+      expect(series.general_direct_filters).not_to include(meta)
```
Bilka2:
```suggestion
      expect(series.public_direct_filters).not_to include(meta)
```

Bilka2 (changes requested):
Just one implementation question, rest is minor things and tests!

app/models/search/bookmarkable_query.rb
brianjaustin: That works... ish. I have to do a bit of annoying hackery because there doesn't seem to be an "easy" way to inject one regex into another (probably a good thing usually). I have something that makes this second regex ever so slightly more complicated, but I think it's OK?

spec/models/series_spec.rb
```diff
@@ -110,4 +154,147 @@
...
+    end
+
+    it "does not include tags on restricted works" do
+      series.works = [restricted_work]
+      series.reload
+      expect(series.public_filter_ids).to be_empty
+    end
+
+    it "does not include tags on works hidden by an admin" do
+      hidden_work = create(:work, hidden_by_admin: true)
```
Bilka2: canonical tag pls

spec/models/series_spec.rb
```diff
@@ -80,6 +82,48 @@
...
+    end
+
+    it "does not include tags on restricted works" do
+      series.works = [restricted_work]
+      series.reload
+      expect(series.public_filters).to be_empty
+    end
+
+    it "does not include tags on works hidden by an admin" do
+      hidden_work = create(:work, hidden_by_admin: true)
```
Bilka2: canonical tag pls

spec/models/search/bookmarkable_query_spec.rb
```diff
@@ -175,13 +175,41 @@
...
 
+      it "converts the 'tags' field to 'general_tags'" do
+        bookmark_query = BookmarkQuery.new(bookmarkable_query: "tags:foo")
+        q = bookmark_query.bookmarkable_query.generated_query
+        filter_string = q[:query][:bool][:filter][0][:query_string][:query]
+        expect(filter_string).to eq("general_tags:foo")
+      end
+
       %w[archive_warning category character fandom filter freeform rating relationship tags].each do |field|
+        it "converts the '#{field}' field to 'general_#{field}'" do
```
Bilka2: These aren't valid fields for the 7 tag types or "filter", so it _shouldn't_ do this... and same goes for the next `it` block

spec/models/search/bookmarkable_query_spec.rb
```diff
@@ -175,13 +175,41 @@
...
+        end
+
         it "converts the 'public_#{field}' field to 'general_#{field}'" do
           bookmark_query = BookmarkQuery.new(bookmarkable_query: "public_#{field}:foo")
           q = bookmark_query.bookmarkable_query.generated_query
           filter_string = q[:query][:bool][:filter][0][:query_string][:query]
           expect(filter_string).to eq("general_#{field}:foo")
         end
+
+        it "converts the '#{field}_ids' field to 'general_#{field}_ids'" do
```
Bilka2, line 218: and this one checks tags_ids which shouldn't be converted

app/models/search/bookmarkable_query.rb
```diff
@@ -92,7 +94,28 @@ def general_query
...
+  # escape_tags_field("public_tags:1234") => public_tags:1234
+  # escape_tags_field("general_fandom_ids:1234") => public_fandom_ids:1234
+  # escape_tags_field("tag:1234") => public_tags:1234
+  # When searching as a registered user,
+  # escape_tags_field("public_tags:1234") => general_tags:1234
+  # escape_tags_field("general_character_ids:1234") => general_character_ids:1234
+  # escape_tags_field("tag:1234") => general_tags:1234
+  def escape_restrictable_fields(query)
+    # Special-case for the "tag" convenience field name first, then sanitize visibility level.
+    query.gsub("tag:", "public_tags:").gsub(/(?:(?:public|general)_)?((?:#{RESTRICTABLE_FIELDS})(?:_ids)?):/) do
```
Bilka2: Okay so I think the tests are messed up because this code here is overly permissive. I think in this case it might be easier to give up using `Tag::FILTERS` and hardcode it. That would also make brakeman happy. Here is an attempt:

```suggestion
    field_names = "rating_ids|archive_warning_ids|category_ids|fandom_ids|relationship_ids|character_ids|freeform_ids|filter_ids|tags"
    query.gsub("tag:", "public_tags:").gsub(/(?:(?:public|general)_)?(#{field_names}):/) do
```

I also kind of like this because it gives a very direct overview which fields in the index this affects - something we'd otherwise would have to hunt for in the tests or other files, if the coder even knows where to look

brianjaustin: Yeah, I was thinking about that as well. I think it makes sense (and bonus, less being clever with regex... which is usually good to avoid :laughing: )

spec/models/search/bookmark_query_spec.rb
```diff
@@ -138,13 +166,47 @@ def find_parent_filter(query_list)
...
+          expect(aggregations[:bookmarkable]).to \
+            include({ parent: { type: "bookmark" } })
+
+          expect(aggregations.dig(:bookmarkable, :aggs, type.underscore)).to \
+            include({ terms: { field: "general_#{type.underscore}_ids" } })
+        end
+      end
+    end
+
+    context "when run by a guest admin" do
```
Bilka2:
```suggestion
    context "when run by a guest" do
```

PR #5180, opened 2025-05-08
AO3-6166 Prevent unrevealed/hidden works title exposure

Bilka2:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6166

## Purpose

Adopted previous PR with apologies to the previous coder that nobody reviewed it for nearly 3 years. Reduced scope of changes and tests to cover only the actions listed in the issue.

## References

#4285

## Credit

korrien
Bilka


app/controllers/application_controller.rb
```diff
@@ -507,6 +507,15 @@ def check_permission_to_wrangle
     end
   end
 
+  # Checks if user is allowed to see related page if parent item is hidden or in unrevealed collection
+  def check_parent_visibility(parent)
+    # Only admins and the owner can see related pages on something hidden by an admin.
+    logged_in_as_admin? || current_user_owns?(parent) || access_denied(redirect: root_path) if parent.respond_to?(:hidden_by_admin) && parent.hidden_by_admin
+
+    # Only admins and the owner can see related pages on unrevealed works.
+    logged_in_as_admin? || current_user_owns?(parent) || access_denied(redirect: root_path) if parent.respond_to?(:in_unrevealed_collection) && parent.in_unrevealed_collection
```
brianjaustin, line 516: Can we simplify the guards using `try`?
```suggestion
    logged_in_as_admin? || current_user_owns?(parent) || access_denied(redirect: root_path) if parent.try(:hidden_by_admin)

    # Only admins and the owner can see related pages on unrevealed works.
    logged_in_as_admin? || current_user_owns?(parent) || access_denied(redirect: root_path) if parent.try(:in_unrevealed_collection)
```

app/controllers/application_controller.rb
```diff
@@ -507,6 +507,15 @@ def check_permission_to_wrangle
     end
   end
 
+  # Checks if user is allowed to see related page if parent item is hidden or in unrevealed collection
+  def check_parent_visibility(parent)
```
brianjaustin, line 511: Should this be just "check_visibility"? Otherwise, I would expect this to call a variant of `#parent` on the object passed here

Bilka2: There's no unified way to get the parent for all these. It's not named `check_visibility` because that's already a thing, it's two methods above this one. But I don't like this name either.

brianjaustin: Maybe either `check_visibility_of` or `check_visibility_for`? The other one is generically for the current page, where this is for the passed in object/record

brianjaustin (approved):

PR #5184, opened 2025-05-10
Ao3-6303 Hide restricted work tags on series blurbs

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6303

## Purpose

Updates series' blurbs (e.g. when viewing a bookmark) to hide tags that are only on restricted works when the blurb is viewed as a guest.

## References

This doesn't _have_ to be merged at the same time as https://github.com/otwcode/otwarchive/pull/5179, although there is potential for confusion if not (either bookmarks are returned in search where the blurb doesn't have a tag, or bookmarks with a tag in their blurb aren't returned in search)

### DIFF
--- a/app/models/series.rb
+++ b/app/models/series.rb
@@ -190,8 +190,8 @@ class Series < ApplicationRecord
   end
 
   # returns list of fandoms on this series
-  def allfandoms
-    works.collect(&:fandoms).flatten.compact.uniq.sort
+  def fandoms
+    tag_groups["Fandom"].sort
   end
 
   def author_tags
@@ -199,7 +199,11 @@ class Series < ApplicationRecord
   end
 
   def tag_groups
-    self.work_tags.group_by { |t| t.type.to_s }
+    @tag_groups ||= if User.current_user.present?
+                      self.work_tags.group_by { |t| t.type.to_s }
+                    else
+                      self.work_tags.where(works: { restricted: false }).group_by { |t| t.type.to_s }
+                    end
   end
 
   # Grabs the earliest published_at date of the visible works in the series
--- a/app/views/series/_series_module.html.erb
+++ b/app/views/series/_series_module.html.erb
@@ -1,27 +1,23 @@
-<% # expects "series" %>
+<%# expects "series" %>
 
 <!--title etc-->
 <div class="header module">
   <h4 class="heading">
-    <%= link_to series.title, series %>
-    <%= ts('by') %>
-    <%= byline(series) %>
+    <%= t(".series_by_html", series_title_link: link_to(series.title, series), byline: byline(series)) %>
     <% if series.restricted %>
-      <%= image_tag("lockblue.png", size: "15x15", alt: ts("(Restricted)"),
-                    title: ts("Restricted"),
+      <%= image_tag("lockblue.png", size: "15x15", alt: t(".restricted_alt"),
+                    title: t(".restricted"),
                     skip_pipeline: true) %>
     <% end %>
     <% if series.hidden_by_admin %>
-      <%= image_tag("lockred.png", size: "15x15", alt: ts("(Hidden by Admin)"),
-                    title: ts("Hidden by Administrator"),
+      <%= image_tag("lockred.png", size: "15x15", alt: t(".hidden_by_admin_alt"),
+                    title: t(".hidden_by_administrator"),
                     skip_pipeline: true) %>
     <% end %>
   </h4>
   <h5 class="fandoms heading">
-    <span class="landmark"><%= ts('Fandom') %>:</span>
-    <%= series.work_tags.select{ |tag| tag.type == "Fandom" }.
-                         sort.collect{ |tag| link_to_tag_works(tag) }.
-                         join(', ').html_safe %>
+    <span class="landmark"><%= t(".landmark.fandom") %></span>
+    <%= series.fandoms.map { |tag| link_to_tag_works(tag) }.join(t("support.array.words_connector")).html_safe %>
   </h5>
   <%= get_symbols_for(series) %>
   <p class="datetime"><%= set_format_for_date(series.revised_at) %></p>
@@ -34,7 +30,7 @@
 
 <!--summary-->
 <% unless series.summary.blank? %>
-  <h6 class="landmark heading"><%= ts("Summary") %></h6>
+  <h6 class="landmark heading"><%= t(".landmark.summary") %></h6>
   <blockquote class="userstuff summary">
     <%=raw strip_images(sanitize_field(series, :summary)) %>
   </blockquote>
--- a/spec/models/series_spec.rb
+++ b/spec/models/series_spec.rb
@@ -1,28 +1,45 @@
 # frozen_string_literal: true
 
-require 'spec_helper'
+require "spec_helper"
 
 describe Series do
   let(:unrestricted_work) { create(:work, restricted: false) }
   let(:restricted_work) { create(:work, restricted: true) }
   let(:series) { create(:series) }
 
-  it "should be unrestricted when it has unrestricted works" do
-    series.works = [unrestricted_work]
-    series.reload
-    expect(series.restricted).not_to be_truthy
-  end
+  describe "#restricted" do
+    context "when the series has only unrestricted works" do
+      before do
+        series.works = [unrestricted_work]
+        series.reload
+      end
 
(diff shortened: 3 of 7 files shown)


app/models/series.rb
```diff
@@ -190,16 +190,20 @@ def remove_author(author_to_remove)
...
   def author_tags
     self.work_tags.select{|t| t.type == "Relationship"}.sort + self.work_tags.select{|t| t.type == "Character"}.sort + self.work_tags.select{|t| t.type == "Freeform"}.sort
   end
 
   def tag_groups
-    self.work_tags.group_by { |t| t.type.to_s }
+    @tag_groups ||= if User.current_user.present?
+                      self.work_tags.group_by { |t| t.type.to_s }
+                    else
+                      self.work_tags.where(works: { restricted: false }).group_by { |t| t.type.to_s }
```
sarken, line 205: I think this came up elsewhere, but should this also have `hidden_by_admin: false`?

zz9pzza:
Has conflicts.

brianjaustin:
> Has conflicts.

Yes; I plan to resolve things when https://github.com/otwcode/otwarchive/pull/5234 is merged.

app/models/series.rb
```diff
@@ -190,16 +190,19 @@ def remove_author(author_to_remove)
...
 
   def author_tags
     self.work_tags.select{|t| t.type == "Relationship"}.sort + self.work_tags.select{|t| t.type == "Character"}.sort + self.work_tags.select{|t| t.type == "Freeform"}.sort
   end
 
   def tag_groups
-    self.work_tags.group_by { |t| t.type.to_s }
+    tags = self.work_tags
+    tags = tags.where(works: { restricted: false }) unless User.current_user
+    tags = tags.where(works: { hidden_by_admin: false }) unless User.current_user.is_a?(Admin)
```
Bilka2: Would we want to allow the author to see the tags of hidden works in the series?

On the series page the hidden work itself is visible to the creator

irrationalpie7: I can see reasons to have it either way--if they can see the tags, they are a preview of what others will see once the work is shown, and might help with finding the series (either by visual scan of their own series page, or if we also support filtering or searching by these hidden tags as long as you're the author). On the other hand, if the tags are hidden, it's a nice confirmation that the work is hidden and others won't see those tags. I kind of lean towards the second option, partly for that reassurance of "this is what others see, yes your work is actually hidden", partly because I think it'll make searching/filtering bookmarked series easier to manage from a technical standpoint if we don't have to have a whole separate index(?) to handle the case of "author's own work" in addition to logged in/logged out users

(I know that's not what the current pr is about, but the reason I'm following this pr is because it's tangentially related to bookmark filtering by word count, which has a related problem of "what do we show"/"what do we filter by")

lydia-theda: after thinking on it, I lean towards going the other way: hiding the hidden work tags on the series blurb for everyone. (PAC would prefer to only hide a series if the series itself is in violation (title, description, etc) so as to reduce confusion regarding what does and doesn't need to be edited. I also think it'd be confusing for the admin and/or creator to wonder why the work is hidden but not the series if they're able to see those inherited tags on the series blurb/user's dashboard/etc.)

irrationalpie7: Oh, I'm not sure I was using terminology correctly--I was thinking of tags belonging to unrevealed works, not works hidden due to a violation. I think my preference and reasoning applies to both situations though

lydia-theda: Oh, I commented about admin hidden works because that's what was visible in the preview. I also tried testing on staging and it appears as though adding an unrevealed work to a series removes the series from the user's series page(? I knew this happened for anonymous works but didn't know it also affected unrevealed), so I can't tell what tags are currently shown on the blurb. I'm not able to search Jira right now to see if that's a known bug or how it's intended to be handled…

irrationalpie7: Does the series only have one work? I know that series with only anonymous or unrevealed works get hidden, but I thought if you had a public non-anonymous fic in a series then the series stays on your dashboard even if you add a new work to it that's anonymous or unrevealed 🤔

lydia-theda: It was a series with multiple public works and one admin hidden work. I added an unrevealed work to it, belatedly realized it was also anonymous (adding one anonymous work to a series currently turns the whole series anonymous instead of listing both Anonymous and Creator; an anonymous series no longer shows up on user's dash – the latter was changed in recent years), took it out of the anonymous collection and put it in a collection that was only unrevealed, waited a while to no avail, removed the unrevealed work from the series and instantly saw the series on my dash again, added the work back to the series and saw Series (0) again. I agree this doesn't sound like intended behavior, but…

Bilka2: Based on this discussion, AD&T has decided that the tags from the hidden work should not be visible on the series blurb, neither for the creator nor for admins

app/models/series.rb
```diff
@@ -194,16 +194,18 @@ def remove_author(author_to_remove)
...
   end
 
   def author_tags
     self.work_tags.select{|t| t.type == "Relationship"}.sort + self.work_tags.select{|t| t.type == "Character"}.sort + self.work_tags.select{|t| t.type == "Freeform"}.sort
   end
 
   def tag_groups
-    self.work_tags.group_by { |t| t.type.to_s }
+    self.work_tags
+      .where(works: { restricted: false, hidden_by_admin: false })
```
omerfaruk-pseud: In this case, with series with only restricted works, blurb doesn't show any tags or fandoms to registered users. Guests can't see the series at all. Is this intended behavior?

Bilka2: Agreed with this comment, restricted tags should still be included for users. It's only hidden and draft tags that should be removed for everyone

app/views/series/_series_module.html.erb
```diff
@@ -1,27 +1,24 @@
...
     <% end %>
   </h4>
   <h5 class="fandoms heading">
-    <span class="landmark"><%= ts('Fandom') %>:</span>
-    <%= series.work_tags.select{ |tag| tag.type == "Fandom" }.
-                         sort.collect{ |tag| link_to_tag_works(tag) }.
-                         join(', ').html_safe %>
+    <span class="landmark"><%= t(".landmark.fandom") %></span>
+    <%= series.fandoms.map { |tag| link_to_tag_works(tag) }
+          .join(t("support.array.words_connector")).html_safe %>
```
Bilka2: Could you use `safe_join` instead of `join` with `html_safe` please

spec/models/series_spec.rb
```diff
@@ -110,4 +127,92 @@
...
+      it "does not return fandoms on restricted works" do
+        expect(series.fandoms).not_to include(*restricted_work.fandoms)
+      end
+
+      it "does not return fandoms on hidden works" do
+        expect(series.fandoms).not_to include(*hidden_work.fandoms)
+      end
+    end
+
+    context "when no user is logged in" do
```
Bilka2:
```suggestion
    context "when logged out" do
```

spec/models/series_spec.rb
```diff
@@ -110,4 +127,92 @@
...
+        User.current_user = create(:admin)
+      end
+
+      it_behaves_like "only returns fandoms on unrestricted, unhidden works"
+    end
+  end
+
+  describe "#tag_groups" do
+    let(:restricted_work) { create(:work, restricted: true, fandom_string: "Testing2") }
+    let(:hidden_work) { create(:work, hidden_by_admin: true, fandom_string: "Testing3") }
```
Bilka2: Could you test this with a different tag type, since Fandom is already mostly covered by the `#fandoms` test?

Bilka2 (approved):
Besides the functionality change with restricted tags, could you also add a test for a normal series blurb not showing the tags, e.g. on the user dashboard

app/models/series.rb
```diff
@@ -195,17 +195,17 @@ def remove_author(author_to_remove)
 
   # returns list of fandoms on this series
   def fandoms
-    tag_groups["Fandom"]&.sort || []
+    tag_groups["Fandom"].sort
```
Bilka2: How sure are you that there are always Fandoms? I seem to recall a bug where some works didn't have fandoms

brianjaustin: Most of the time it should be fine (fandoms are required on works, and series won't be shown if they don't have any unrestricted works when a guest lists), but no harm in programming defensively :shrug:

PR #5188, opened 2025-05-14
AO3-6967 add save button to top of tag edit page

Gray-V:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6967

## Purpose

Adding a save button to the top of the tag edit page for tag wranglers.

## Testing Instructions

Check Jira issue

## Credit

Grayson von Goetz (they/them)

### DIFF
--- a/app/views/tags/edit.html.erb
+++ b/app/views/tags/edit.html.erb
@@ -17,9 +17,9 @@
 
 <%= form_for @tag, as: :tag, url: { action: "update", id: @tag}, html: { method: :put } do |f| %>
 <fieldset>
-    <legend><%= ts("Submit") %></legend>
+    <legend><%= t(".submit_legend") %></legend>
     <p class="submit actions">
-      <%= submit_tag ts("Save changes") %>
+      <%= submit_tag t(".save_changes") %>
     </p>
   </fieldset>
   <fieldset>
@@ -223,9 +223,9 @@
   <% end %>
 
   <fieldset>
-    <legend><%= ts("Submit") %></legend>
+    <legend><%= t(".submit_legend") %></legend>
     <p class="submit actions">
-      <%= submit_tag ts("Save changes") %>
+      <%= submit_tag t(".save_changes") %>
     </p>
   </fieldset>
 <% end %>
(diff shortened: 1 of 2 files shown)


app/views/tags/edit.html.erb
```diff
@@ -16,7 +16,12 @@
 <% end %>
 
 <%= form_for @tag, as: :tag, url: { action: "update", id: @tag}, html: { method: :put } do |f| %>
-
+<fieldset>
+    <legend><%= ts("Submit") %></legend>
+    <p class="submit actions">
+      <%= submit_tag ts("Save changes") %>
+    </p>
```
brianjaustin, line 23: Since this is a new button (and the page isn't cached), we should use the Rails t-helper so that the buttons are actually translatable. You can add the following to https://github.com/otwcode/otwarchive/blob/979eb8d2ca117b367ae2a8c66d120b1fa6ffd622/config/locales/views/en.yml#L2050
```yml
    edit:
      save_changes: Save changes
      submit_legend: Submit
```

Then, you would be able to use them here with
```suggestion
    <legend><%= t.(".submit_legend") %></legend>
    <p class="submit actions">
      <%= submit_tag t(".save_changes") %>
    </p>
```

If you want to change the save changes button at the bottom to use those translation keys as well, that would be nice, but we can do it later if you would rather keep this PR simple.

More information about our internationalization standards is [available on our wiki](https://github.com/otwcode/otwarchive/wiki/Internationalization-(i18n)-Standards).

brianjaustin (changes requested):
Hi, Grayson, and thanks for the PR! I have one bit of feedback, but in general it looks like you're headed in the right direction.

I've assigned the Jira issue to you and set its status to In Review, so no one will mistakenly create a duplicate pull request. I've also given your account permission to comment on, assign, and transition issues in the future.

Thanks again, and if you have any questions, please don't hesitate to contact us at otw-coders@transformativeworks.org!

app/views/tags/edit.html.erb
Gray-V: Sounds good! Working on it now. I will use it to update the bottom as well. I can also work on the rest of the file but I will to hear back from that lol. Thanks for getting me set up on Jira! Also, I put it under tag like you said, but would it potentially be the play to put it in under tag wrangling over on 2022?

Gray-V:
Ahh ok makes sense thank you.

PR #5189, opened 2025-05-14
AO3-6987 AO3-6991 Guest comment image safety with alt text

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6987
https://otwarchive.atlassian.net/browse/AO3-6991

## Purpose

1. Update the image stripping function to include `img` tag attributes (e.g. src and alt)
2. Set guest comments to always use image safety mode

### DIFF
--- a/lib/html_cleaner.rb
+++ b/lib/html_cleaner.rb
@@ -141,13 +141,9 @@ module HtmlCleaner
   # These assume they are running on well-formed XHTML, which we can do
   # because they will only be used on already-cleaned fields.
 
-  # strip img tags, optionally leaving the src URL exposed
+  # strip img tags, optionally leaving the HTML attributes (e.g. src and alt) exposed
   def strip_images(value, keep_src: false)
-    value.gsub(/(<img .*?>)/) do |img_tag|
-      match_data = img_tag.match(/src="([^"]+)"/) if keep_src
-      src = match_data[1] if match_data
-      src || ""
-    end
+    value.gsub(%r{(?:<(img .*?) ?/?>)}, keep_src ? "\\1" : "")
   end
 
   def strip_html_breaks_simple(value)
--- a/spec/lib/html_cleaner_spec.rb
+++ b/spec/lib/html_cleaner_spec.rb
@@ -1,6 +1,5 @@
-# -*- coding: utf-8 -*-
-require 'spec_helper'
-require 'nokogiri'
+require "spec_helper"
+require "nokogiri"
 
 describe HtmlCleaner do
   include HtmlCleaner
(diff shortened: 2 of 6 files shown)


spec/lib/html_cleaner_spec.rb
```diff
@@ -1147,56 +1146,22 @@
     end
 
     context "with keep_src: false" do
```
Bilka2, line 1148: The new regex definitely doesn't care about the quote formatting anymore, but maybe it'd be good to keep these tests to show that?

lib/html_cleaner.rb
```diff
@@ -141,13 +141,9 @@ def add_paragraphs_to_text(text)
...
 
-  # strip img tags, optionally leaving the src URL exposed
+  # strip img tags, optionally leaving the HTML attributes (e.g. src and alt) exposed
   def strip_images(value, keep_src: false)
-    value.gsub(/(<img .*?>)/) do |img_tag|
-      match_data = img_tag.match(/src="([^"]+)"/) if keep_src
-      src = match_data[1] if match_data
-      src || ""
-    end
+    value.gsub(%r{(?:<(img .*?) ?/?>)}, keep_src ? "\\1" : "")
```
Bilka2, line 146: The way I understand the ` ?` before the trailing slash is that you're trying to remove the trailing space from the source. I think it would be clearer to instead `.strip` the `"\\1"`, since regex tends to be harder to read

Bilka2 (commented):
Code looks good, just two minor notes

features/comments_and_kudos/image_safety_mode.feature
Bilka2, line 1: Ah, one thing - a super simple feature test for the images being hidden in guest comment would put me more at ease

lib/html_cleaner.rb
brianjaustin: Not quite, and the difference means `.strip` won't work. This regex is intended to match
```html
<img src="foo" />
<img src="foo"/>
<img src="foo">
```
where both the space and trailing slash are optional. (Technically this also supports the case where `.strip` _would_ work -- `` -- but that's not really the primary goal.) They're specified now to avoid grabbing them in the inner (capture group (alongside the HTML attributes)

Bilka2: I see, thank you for explaining!

Bilka2 (approved):

lib/html_cleaner.rb
sarken: Won't this also make it so any allowed attributes that are basically just clutter when the image isn't being embedded (e.g., width or border) will still be left behind? Or am I missing something that makes it only apply to src and alt?

brianjaustin: It will. I decided that was an OK tradeoff to make the code much simpler than it was otherwise

PR #5191, opened 2025-05-15
AO3-6995 Add recently_reset to ignored_columns

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6995

## Purpose

Try to prevent errors when dropping the column by adding it to ignore_columns.

The migration should be run in the next deploy after the one that this PR is included in, and then it should be run shortly before that deploy.

## Testing Instructions

None.

## Credit

Bilka


brianjaustin (approved):
Thanks!

PR #5197, opened 2025-05-17
AO3-6807 Fix error 500 for invalid URLs in prompts and external works

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6807

## Purpose

Fix that entering an invalid URL in the "inspired by" field when posting a work or in the prompt URL field when signing up to a challenge would result in an error 500.

I decided to fix the external work one as well because it has the most occurrences of `InvalidURIError` in Sentry: https://organization-for-transformativ.sentry.io/issues/6477975500 But I'd be happy to split this into a separate issue + PR instead. Collection header URLs and URLs in the external works controller likely have the same issue (all caused by #4600) but they don't really show up in Sentry so I didn't want to combine them into this as well.

## Testing Instructions

Inspired by instructions to be added to Jira based on the cucumber test if this is fine to combine.

## Credit

Bilka

### DIFF
--- a/app/models/external_work.rb
+++ b/app/models/external_work.rb
@@ -42,6 +42,8 @@ class ExternalWork < ApplicationRecord
   validates :url, presence: true, url_format: true, url_active: true
   def cleanup_url
     self.url = Addressable::URI.heuristic_parse(self.url) if self.url
+  rescue Addressable::URI::InvalidURIError
+    # url_format validation creates the error message
   end
 
   # Allow encoded characters to display correctly in titles
--- a/app/models/prompt.rb
+++ b/app/models/prompt.rb
@@ -65,6 +65,8 @@ class Prompt < ApplicationRecord
   before_validation :cleanup_url
   def cleanup_url
     self.url = Addressable::URI.heuristic_parse(self.url) if self.url
+  rescue Addressable::URI::InvalidURIError
+    # url_format validation creates the error message
   end
 
   validate :correct_number_of_tags
--- a/config/locales/models/en.yml
+++ b/config/locales/models/en.yml
@@ -16,6 +16,8 @@ en:
         support: Support
         tag_wrangling: Tag Wrangling
         translation: Translation
+      challenge_signup/requests:
+        url: Request URL
       chapters/creatorships:
         base: 'Invalid creator:'
         pseud_id: Pseud
@@ -35,6 +37,8 @@ en:
         meta_tag_id: Metatag
         sub_tag: Subtag
         sub_tag_id: Subtag
+      request:
+        url: URL
       role:
         archivist: Archivist
         no_resets: No Resets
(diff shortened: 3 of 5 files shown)


app/models/prompt.rb
```diff
@@ -65,6 +65,8 @@ def inherit_from_signup
   before_validation :cleanup_url
```
ceithir, line 65: On paper, this seems like the perfect place to implement the [shiny new thing](https://www.shakacode.com/blog/rails-add-normalize-to-activerecord/) of Rails 7.1, with a nice UrlNormalizer and all.

In practice, I'm always suspicious of shiny new things.

app/models/external_work.rb
```diff
@@ -42,6 +42,8 @@ class ExternalWork < ApplicationRecord
   validates :url, presence: true, url_format: true, url_active: true
```
ceithir, line 42: Should AO3 [custom url validator](https://github.com/otwcode/otwarchive/blob/7d27b34b8aad7fc980e155217805cde004a5d696/app/validators/url_format_validator.rb) also call `Addressable::URI.heuristic_parse` instead of checking against a regex? Would avoid weird edge cases where the regex is matched but the parsing failed.

app/models/prompt.rb
Bilka2: Thank you! I knew there was something like this, but had forgotten it and didn't find it when searching. Unfortunately [the docs](https://api.rubyonrails.org/classes/ActiveRecord/Normalization/ClassMethods.html#method-i-normalizes) say:
> Be aware that if your app was created before [Rails](https://api.rubyonrails.org/classes/Rails.html) 7.1, and your app marshals instances of the targeted model (for example, when caching), then you should set [ActiveRecord.marshalling_format_version](https://api.rubyonrails.org/classes/ActiveRecord.html#method-c-marshalling_format_version) to 7.1 or higher via either config.load_defaults 7.1 or config.active_record.marshalling_format_version = 7.1. Otherwise, Marshal may attempt to serialize the normalization Proc and raise TypeError.

And we do not have that set to 7.1 yet. So this will have to stay as is for now.

app/validators/url_format_validator.rb
```diff
@@ -8,7 +8,13 @@ def validate_each(record,attribute,value)
...
     regexp = /^https?:\/\/[_a-z\d\-]+\.[._a-z\d\-]+(:\d+)?\/?.+/i
     unless value.match regexp
-      record.errors.add(attribute, options[:message] || "does not appear to be a valid URL.")
+      record.errors.add(attribute, options[:message] || :invalid)
+    end
+
+    begin
+      Addressable::URI.heuristic_parse(value)
+    rescue Addressable::URI::InvalidURIError
+      record.errors.add(attribute, options[:message] || :invalid)
```
Bilka2, line 17: I don't like this control flow with the repeated errors.add but could not come up with anything better since it's using a `rescue` :/

Bilka2: Hm, this feels like a risky change since it changes when the parsing runs and it doesn't solve the double `errors.add` :/

And the regex is needed unless we'd want to implement restrictions like the checking the scheme on top of the parse.

PR #5198, opened 2025-05-18
AO3-5352 i18n deleted work notification attachment

Hamham6:
# Pull Request Checklist

* [x] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)?&ast;

&ast;No changed behaviour; tests added for existing behaviour.
## Issue

https://otwarchive.atlassian.net/browse/AO3-5352

## Purpose

i18n download views

## References

See Jira

## Credit

Hamham6, it/its


Hamham6:
I wasn't sure wether/how to add tests for this.

features/works/work_download.feature
```diff
@@ -88,12 +88,12 @@ Feature: Download a work
     And I should see "Archive Warning: No Archive Warnings Apply"
     And I should see "Category: Gen"
     And I should see "Fandom: Cool Fandom"
-    And I should see "Relationships: Character 1/Character 2, Character 1 & Character 3"
-    And I should see "Characters: Character 1, Character 2, Character 3"
+    And I should see "Relationships: Character 1/Character 2 and Character 1 & Character 3"
```
weeklies, line 91: Not staff but I don't know if using `to_sentence` for tags, collections and characters is desired behavior?

Hamham6: I thought I saw somewhere that was how it was supposed to be handled but now I don't remember how exactly I got that idea lol. I see join() just being used with .support.array.words_connector in other places so if behavior should be maintained I can switch to that and revert the test, along with any other review changes.

sarken: weeklies is correct; we use `join` instead of `to_sentence` in the work meta

app/views/downloads/_download_preface.html.erb
```diff
@@ -1,66 +1,65 @@
...
   <div class="meta">
     <dl class="tags">
       <% Tag::VISIBLE.each do |type| %>
         <% tags = @work.tag_groups[type] %>
         <% unless tags.blank? %>
-          <dt><%= tags.size == 1 ? type.constantize::NAME : type.constantize::NAME.pluralize(locale: :en) %>:</dt>
-          <dd><%= tags.map {|t| link_to(t.display_name, tag_url(t))}.join(", ").html_safe %></dd>
+          <dt><%= t(".tag_type_html", tag_type: tags.size == 1 ? type.constantize::NAME : type.constantize::NAME.pluralize) %></dt>
+          <dd><%= tags.map { |t| link_to(t.display_name, tag_url(t)) }
+                    .join(t(".support.array.words_connector")).html_safe %></dd>
```
weeklies: If you use [safe_join](https://api.rubyonrails.org/classes/ActionView/Helpers/OutputSafetyHelper.html#method-i-safe_join), you should be able to get rid of the `html_safe` and the security warnings.

weeklies: To fix the missing key, you should remove the first period as that is for [lazy lookup](
https://guides.rubyonrails.org/i18n.html#lazy-lookup)
: `t("support.array.words_connector")`.

app/views/downloads/_download_afterword.html.erb
```diff
@@ -1,12 +1,12 @@
 <div id="afterword">
-  <h2 class="toc-heading"><%= ts("Afterword") %></h2>
+  <h2 class="toc-heading"><%= t(".toc_heading") %></h2>
```
weeklies: As toc-heading is a class name rather than content, `afterword` may be more helpful as a translation key.

app/views/downloads/_download_afterword.html.erb
```diff
@@ -1,12 +1,12 @@
...
 
   <% unless @work.endnotes.blank? && @work.approved_children.blank? %>
     <div class="meta">
       <% unless @work.endnotes.blank? %>
         <div id="endnotes">
-          <p><%= ts("End Notes") %></p>
-          <blockquote class="userstuff"><%=raw sanitize_field(@work, :endnotes) %></blockquote>
+          <p><%= t(".endnotes") %></p>
```
weeklies:
```suggestion 
          <p><%= t(".end_notes") %></p>
```

app/views/downloads/_download_chapter.html.erb
```diff
@@ -4,36 +4,36 @@
   <h2 class="heading"><%= @chapter.chapter_title.html_safe %></h2>
   <% if (!@chapter.pseuds.blank? && (@chapter.pseuds != @work.pseuds) && (!@work.anonymous?)) %>
     <%# only display byline if different from the main byline %>
-    <p class="byline"><%= ts('Chapter by') %> <%= byline(@chapter, only_path: false) %></p>
+    <p class="byline"><%= t(".chapter_byline_html", byline_names: byline(@chapter, only_path: false)) %></p>
```
weeklies: The [i18n standards](https://github.com/otwcode/otwarchive/wiki/Internationalization-%28i18n%29-Standards#user-content-Hyperlinks) recommends a `_link` suffix.

app/views/downloads/_download_chapter.html.erb
```diff
@@ -4,36 +4,36 @@
...
 <div class="userstuff">
-  <%=raw (defined?(@content) ? @content : sanitize_field(@chapter, :content)) %>
+  <%= raw defined?(@content) ? @content : sanitize_field(@chapter, :content) %>
 </div>
 <!--/chapter content-->
 
 <% unless @chapter.endnotes.blank? %>
   <div class="meta" id="endnotes<%= @chapter.position %>">
-    <p><%= ts('Chapter End Notes') %></p>
+    <p><%= t(".chapter_endnotes") %></p>
```
weeklies:
```suggestion 
    <p><%= t(".chapter_end_notes") %></p>
```

app/views/downloads/_download_preface.html.erb
```diff
@@ -1,66 +1,65 @@
 <div id="preface">
-  <h2 class="toc-heading"><%= ts("Preface") %></h2>
+  <h2 class="toc-heading"><%= t(".toc_heading") %></h2>
```
weeklies:
```suggestion 
  <h2 class="toc-heading"><%= t(".preface") %></h2>
```

app/views/downloads/_download_preface.html.erb
```diff
@@ -1,66 +1,65 @@
...
-#{link_to(s.series.title, series_url(s.series))}"}.join(", ").html_safe %></dd>
+        <dt><%= t(".series", count: series_list.size) %></dt>
+        <dd><%= series_list.map { |s| t(".series_list_html", position: s.position, series_link: link_to(s.series.title, series_url(s.series))) }
+                  .join(t(".support.array.words_connector")).html_safe %></dd>
       <% end %>
       <% unless @work.approved_collections.empty? %>
-        <dt><%= ts("Collections") %>:</dt>
-        <dd><%= @work.approved_collections.map {|c| link_to c.title, collection_url(c)}.join(",
-").html_safe %></dd>
+        <dt><%= t(".collections", count: @work.approved_collections.size) %></dt>
```
weeklies: I don't think the pluralization is desired and would be a change from the existing work behaviour.

config/locales/views/en.yml
```diff
@@ -687,18 +687,48 @@ en:
...
+      endnotes_with_worknotes_html: See the end of the work for more %{notes_link}
       inspired_by:
         restricted_html: Inspired by [Restricted Work] by %{creator_link}
         revealed_html: Inspired by %{work_link} by %{creator_link}
         unrevealed: Inspired by a work in an unrevealed collection
+      language: 'Language:'
+      notes: notes
+      originally_posted_html: Posted originally on the %{archive_link} at %{work_url}.
+      published_html: 'Published: %{date}'
+      series:
```
weeklies: Like Collections, I don't think the pluralization is part of the spec.

app/views/downloads/_download_preface.html.erb
```diff
@@ -1,66 +1,65 @@
...
+        <dd><%= @work.approved_collections.map { |c| link_to(c.title, collection_url(c)) }
+                  .join(t(".support.array.words_connector")).html_safe %></dd>
       <% end %>
-      <dt><%= ts("Stats") %>:</dt>
+      <dt><%= t(".stats") %></dt>
       <dd>
-        <%= ts("Published") %>: <%= l(@work.first_chapter.published_at) %>
+        <%= t(".published_html", date: l(@work.first_chapter.published_at)) %>
```
weeklies: The html suffix is not required here.

app/views/downloads/_download_preface.html.erb
```diff
@@ -1,66 +1,65 @@
...
         <% if @work.first_chapter.published_at < @work.revised_at.to_date %>
-          <%= @work.is_wip ? ts("Updated") : ts("Completed") %>: <%= l(@work.revised_at.to_date) %>
+          <%= @work.is_wip ? t(".updated_html", date: l(@work.revised_at.to_date)) : t(".completed_html", date: l(@work.revised_at.to_date)) %>
```
weeklies: The html suffix can be removed.

app/views/downloads/_download_preface.html.erb
```diff
@@ -1,66 +1,65 @@
...
+      <p><%= t(".work_notes") %></p>
       <% unless @work.notes.blank? %>
-        <blockquote class="userstuff"><%=raw sanitize_field(@work, :notes) %></blockquote>
+        <blockquote class="userstuff"><%= raw sanitize_field(@work, :notes) %></blockquote>
       <% end %>
       <% unless @work.endnotes.blank? %>
         <div class="endnote-link">
-          <%= ts("See the end of the work for") %> <% unless @work.notes.blank? %><%= ts("more") %><% end
-%> <%= link_to ts("notes"), "#endnotes" %>
+          <% @work.notes.present? ? t(".endnotes_with_worknotes_html", notes_link: link_to(t(".notes"), "#endnotes")) : t(".endnotes_no_worknotes_html", notes_link: link_to(t(".notes"), "#endnotes")) %>
```
weeklies: I wonder about moving the ternary logic into an notes_link variable? This is what I encountered in the `app/views/works/_work_header_notes.html.erb` 

``` 
  <% endnotes_link = link_to (@work.notes.blank? ? ts("notes") : ts("more notes")), get_endnotes_link %>
```

Then something like: 
```
          <% t(".endnotes_html", endnotes_link: endnotes_link)) %>
```

Although, this means that the "more" in "more notes" will be then be included in the hyperlink as work pages do. I wonder if this is desirable @sarken?

app/views/downloads/_download_preface.html.erb
```diff
@@ -1,66 +1,65 @@
...
       <dd>
-        <%= ts("Published") %>: <%= l(@work.first_chapter.published_at) %>
+        <%= t(".published_html", date: l(@work.first_chapter.published_at)) %>
         <% if @work.first_chapter.published_at < @work.revised_at.to_date %>
-          <%= @work.is_wip ? ts("Updated") : ts("Completed") %>: <%= l(@work.revised_at.to_date) %>
+          <%= @work.is_wip ? t(".updated_html", date: l(@work.revised_at.to_date)) : t(".completed_html", date: l(@work.revised_at.to_date)) %>
         <% end %>
-        <%= ts("Words") %>: <%= number_with_delimiter(@work.word_count) %>
-        <%= ts("Chapters") %>: <%= chapter_total_display(@work) %>
+        <%= t(".words_html", count_with_delimiters: number_with_delimiter(@work.word_count)) %>
```
weeklies: The suffix can be removed here too.

app/views/downloads/_download_preface.html.erb
```diff
@@ -1,66 +1,65 @@
...
   </p>
 
   <div class="meta">
     <dl class="tags">
       <% Tag::VISIBLE.each do |type| %>
         <% tags = @work.tag_groups[type] %>
         <% unless tags.blank? %>
-          <dt><%= tags.size == 1 ? type.constantize::NAME : type.constantize::NAME.pluralize(locale: :en) %>:</dt>
-          <dd><%= tags.map {|t| link_to(t.display_name, tag_url(t))}.join(", ").html_safe %></dd>
+          <dt><%= t(".tag_type_html", tag_type: tags.size == 1 ? type.constantize::NAME : type.constantize::NAME.pluralize) %></dt>
```
weeklies: The html suffix is not needed here.

weeklies (changes requested):
Thanks for working on this, I just have some suggestions.

app/views/downloads/_download_preface.html.erb
sarken: I think it makes sense to have "more" in the hyperlink!

app/views/downloads/_download_preface.html.erb
Hamham6: Thank you for the tip!

app/views/downloads/_download_preface.html.erb
Hamham6: I put the ternary logic inside the link_to because I thought it was easy enough to read, but I can pull it out into a variable if that makes it more understandable.

app/views/downloads/_download_afterword.html.erb
```diff
@@ -28,5 +28,5 @@
     </div>
   <% end %>
 
-  <p class="message"><%= t(".please_comment_html", work_comment_link: link_to(t(".work_comment", locale: :en), new_work_comment_url(@work)), locale: :en) %></p>
+  <p class="message"><%= t(".please_comment_html", comment_link: link_to(t(".comment"), new_work_comment_url(@work))) %></p>
```
Hamham6, line 31: I changed this because I couldn't guess what `work_comment_link` would read as from `.please_comment_html` and linking to a new comment form. I think this change makes it more self-explanatory, but I can change it back if it was fine before.

app/views/downloads/_download_preface.html.erb
```diff
@@ -1,66 +1,65 @@
...
     <% unless @work.summary.blank? %>
-      <p><%= ts("Summary") %></p>
-      <blockquote class="userstuff"><%=raw sanitize_field(@work, :summary) %></blockquote>
+      <p><%= t(".summary") %></p>
+      <blockquote class="userstuff"><%= raw sanitize_field(@work, :summary) %></blockquote>
     <% end %>
 
     <% unless @work.notes.blank? && @work.endnotes.blank? %>
-      <p><%= ts("Notes") %></p>
+      <p><%= t(".notes") %></p>
```
Hamham6, line 53: Is it alright to have `.notes` and `.notes_link` be unrelated keys in the same view file or should I name `.notes` as `.work_notes` instead to clarify that they're different?

app/views/downloads/_download_preface.html.erb
Hamham6: Actually, does applying the ternary logic to the link work for every language? I don't know any specific examples, but I imagine the difference between "click to see notes" and "click to see more notes" wouldn't always be in one place that can double as the link. Eg the "more" goes at the front but the "notes" you'd click on goes at the end.

weeklies: Setting "more notes" as an i18n string and shared hyperlink is consistent with the behaviour of work pages. I don't think a separation of "more" and "notes" will occur as that would mean breaking the hyperlink in work pages and a separate issue.

app/views/downloads/_download_preface.html.erb
```diff
@@ -1,66 +1,62 @@
...
       <% unless @work.notes.blank? %>
-        <blockquote class="userstuff"><%=raw sanitize_field(@work, :notes) %></blockquote>
+        <blockquote class="userstuff"><%= raw sanitize_field(@work, :notes) %></blockquote>
       <% end %>
       <% unless @work.endnotes.blank? %>
         <div class="endnote-link">
-          <%= ts("See the end of the work for") %> <% unless @work.notes.blank? %><%= ts("more") %><% end
-%> <%= link_to ts("notes"), "#endnotes" %>
+          <% t(".see_end_notes_html", end_notes_link: link_to(@work.notes.present? ? t(".more_notes_link") : t(".notes_link"), "#endnotes")) %>
```
weeklies: This is missing an `=` sign to display the text, i.e. `<%= t(".see_end_notes_html"`. The end notes for `download_chapter` also has the [same issue](https://github.com/otwcode/otwarchive/pull/5198/files#diff-70e04a9524cc7a5f2cd08154551ba236a97ddcf1d1e4de81e15fe013ea477c61R22).

app/views/downloads/_download_chapter.html.erb
```diff
@@ -4,36 +4,36 @@
...
-    <p><%= ts('Chapter Notes') %></p>
+    <p><%= t(".chapter_notes") %></p>
     <% unless @chapter.notes.blank? %>
-      <blockquote class="userstuff"><%=raw sanitize_field(@chapter, :notes) %></blockquote>
+      <blockquote class="userstuff"><%= raw sanitize_field(@chapter, :notes) %></blockquote>
     <% end %>
     <% unless @chapter.endnotes.blank? %>
       <div class="endnote-link">
-        <%= ts("See the end of the chapter for") %> <% unless @chapter.notes.blank? %><%= ts("more") %><% end %> <%= link_to ts("notes"), "#endnotes#{@chapter.position}" %>
+        <% t(".see_chapter_end_notes_html", chapter_end_notes_link: link_to(@chapter.notes.present? ? t(".more_notes_link") : t(".notes_link"), "#endnotes##{@chapter.position}")) %>
```
weeklies: The additional # in `#endnotes##{@chapter.position}` should be removed, it's breaking the anchor.

app/views/downloads/_download_preface.html.erb
weeklies: It seems alright to me.

weeklies (changes requested):

config/locales/views/en.yml
```diff
@@ -687,18 +687,45 @@ en:
...
       inspired_by:
         restricted_html: "[Restricted Work] by %{creator_link}"
         revealed_html: "%{work_link} by %{creator_link}"
         title: Works inspired by this one
         unrevealed: A work in an unrevealed collection
-      please_comment_html: Please %{work_comment_link} to let the creator know if you enjoyed their work!
-      work_comment: drop by the Archive and comment
+      please_comment_html: Please %{comment_link} to let the creator know if you enjoyed their work!
+    download_chapter:
+      chapter_byline_html: 'Chapter by: %{byline_names_link}'
```
weeklies: The `:` seems to be incorrectly added
```suggestion
      chapter_byline_html: 'Chapter by %{byline_names_link}'
```

app/views/downloads/_download_chapter.html.erb
```diff
@@ -4,36 +4,36 @@
...
-    <p><%= ts('Chapter Notes') %></p>
+    <p><%= t(".chapter_notes") %></p>
     <% unless @chapter.notes.blank? %>
-      <blockquote class="userstuff"><%=raw sanitize_field(@chapter, :notes) %></blockquote>
+      <blockquote class="userstuff"><%= raw sanitize_field(@chapter, :notes) %></blockquote>
     <% end %>
     <% unless @chapter.endnotes.blank? %>
       <div class="endnote-link">
-        <%= ts("See the end of the chapter for") %> <% unless @chapter.notes.blank? %><%= ts("more") %><% end %> <%= link_to ts("notes"), "#endnotes#{@chapter.position}" %>
+        <%= t(".see_chapter_end_notes_html", chapter_end_notes_link: link_to(@chapter.notes.present? ? t(".more_notes_link") : t(".notes_link"), "#endnotes#{@chapter.position}")) %>
```
Bilka2: The locale keys for the link captions should not end in `_link`. [Ideally](https://github.com/otwcode/otwarchive/wiki/Internationalization-%28i18n%29-Standards#user-content-Hyperlinks) they'd be `chapter_end_notes` but that won't work since it's two possible captions... Maybe it could be `chapter_end_notes` with a suffix or prefix to distinguish the two captions

Bilka2: Also, could you add a feature test for the condition here? So that we see that it generates the correct text in both cases

Bilka2 (changes requested):

config/locales/views/en.yml
```diff
@@ -691,18 +691,45 @@ en:
...
         restricted_html: "[Restricted Work] by %{creator_link}"
         revealed_html: "%{work_link} by %{creator_link}"
         title: Works inspired by this one
         unrevealed: A work in an unrevealed collection
-      please_comment_html: Please %{work_comment_link} to let the creator know if you enjoyed their work!
-      work_comment: drop by the Archive and comment
+      please_comment_html: Please %{comment_link} to let the creator know if you enjoyed their work!
+    download_chapter:
+      chapter_byline_html: Chapter by %{byline_names_link}
+      chapter_end_notes: Chapter End Notes
```
Hamham6: This key should probably be renamed to something that doesn't start with `chapter_end_notes` to avoid confusion with the key for the link text (or the interpolation variable for the link should be renamed) but I cannot settle on anything. What can this kind of thing be called? A section? A title? A label? A header???? In the HTML it's just a `<p>` tag without any attributes TT

Bilka2: Good point. It's all going to be awkward since it's all referring to each other. I'd call something like `section_chapter_end_notes` for this key good enough

config/locales/views/en.yml
```diff
@@ -691,18 +691,45 @@ en:
...
+      chapter_end_notes_with_chapter_notes: more notes
+      chapter_end_notes_without_chapter_notes: notes
+      chapter_notes: Chapter Notes
+      chapter_summary: Chapter Summary
+      see_chapter_end_notes_html: See the end of the chapter for %{chapter_end_notes_link}
     download_preface:
+      byline_html: by %{byline_names_link}
+      chapters: 'Chapters: %{chapter_total_display}'
+      collections: 'Collections: '
```
Bilka2: What's the reason for the space here and at the end of tag_type, series, language and stats? The labels are all in a dt so the space shouldn't do anything from what I can see?

Hamham6: Oh, I didn't think of that! You're right, it wouldn't do anything; I'll remove them.

config/locales/views/en.yml
Hamham6: 🙏🙏🙏 Thank you for naming!

Bilka2 (approved):
Thank you very much!

PR #5200, opened 2025-05-20
AO3-6550 Remove unnecessary html_safe calls

weeklies:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6550

## Purpose

Remove unnecessary html_safe calls on work.title (both ExternalWork and Work)

## Testing Instructions

See issue.

### DIFF
--- a/app/views/collections/index.html.erb
+++ b/app/views/collections/index.html.erb
@@ -5,7 +5,7 @@
   <% elsif @collection %>
     <%= ts("Challenges/Subcollections in %{collection_name}", :collection_name => @collection.title) %>
   <% elsif @work %>
-    <%= ts("Collections including ") + @work.title.html_safe %>
+    <%= ts("Collections including ") + @work.title %>
   <% else %>
     <%= ts("Collections in the %{archive_name}", :archive_name => ArchiveConfig.APP_NAME) %>
   <% end %>
@@ -64,7 +64,7 @@
   <!--filters subnav-->
   <%= render 'collections/filters' %>
   <!---/subnav-->
-<% end %> 
+<% end %>
 
 <% unless @collections.blank? %>
   <%= will_paginate @collections %>
--- a/app/views/downloads/show.html.erb
+++ b/app/views/downloads/show.html.erb
@@ -7,7 +7,7 @@
     <% end %>
   <% # Use elsif so we don't error when there are no chapters. %>
   <% elsif @chapters.size == 1 %>
-    <h2 class="toc-heading"><%= @work.title.html_safe %></h2>
+    <h2 class="toc-heading"><%= @work.title %></h2>
     <div class="userstuff">
       <%=raw sanitize_field(@chapters.first, :content) %>
     </div>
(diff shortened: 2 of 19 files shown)


app/views/downloads/show.html.erb
```diff
@@ -7,7 +7,7 @@
     <% end %>
   <% # Use elsif so we don't error when there are no chapters. %>
```
brianjaustin, line 8: While you're here, could you clean up the ERB comment here so GitHub doesn't get confused?
```suggestion
  <%# Use elsif so we don't error when there are no chapters. %>
```

app/views/series/_series_order.html.erb
```diff
@@ -5,7 +5,7 @@
       <li id="serial_<%= serial.id %>" class="serial-position-list">
         <%= text_field_tag 'serial_works[]', nil, :size => 3, :maxlength => 3, :class => 'number serial-position-field', :id => "serial_#{i}" %>
         <span id='position-for-<%= serial.id %>'><%= serial.position %></span>.
-        <h3 class="heading"><%=serial.work.title.html_safe + (serial.work.posted? ? "" : " (DRAFT)") %></h3>
+        <h3 class="heading"><%= serial.work.title + (serial.work.posted? ? "" : " (DRAFT)") %></h3>
```
brianjaustin, line 8: Since you're here (and this page isn't cached), could you update this to use the I18n `t` helper as well? One idea I had (it may not be the best way to do things though)
```suggestion
        <h3 class="heading"><%= serial.work.posted? ? serial.work.title : t(".draft_work_title", title: serial.work.title)) %></h3>
```

Another way I thought of, if we don't mind having a "useless" translation key:
```suggestion
        <h3 class="heading"><%= t(serial.work.posted? ? ".published_work_title" : ".draft_work_title", title: serial.work.title)) %></h3>
```
(where the `"published_work_title" key is just `published_work_title: '%{title}'`)

app/views/collections/index.html.erb
```diff
@@ -1,13 +1,13 @@
 <!--Descriptive page name and system messages, descriptions, and instructions.-->
 <h2 class="heading">
   <% if @user %>
-    <%= ts("%{user_name}'s Collections", :user_name => @user.login) %>
+    <%= t(".page_heading.user", user: @user.login) %>
```
brianjaustin, line 4: Based on
>  Furthermore, Translation finds it helpful for variable names in keys to closely correspond with the key for the text the variable represents.

from https://github.com/otwcode/otwarchive/wiki/Internationalization-(i18n)-Standards#user-content-Locale_key_names, I wonder if it would be better to call these things like `users_collections` and `challenges_subcollections_in`?

app/views/series/_series_order.html.erb
```diff
@@ -5,7 +5,9 @@
       <li id="serial_<%= serial.id %>" class="serial-position-list">
         <%= text_field_tag 'serial_works[]', nil, :size => 3, :maxlength => 3, :class => 'number serial-position-field', :id => "serial_#{i}" %>
         <span id='position-for-<%= serial.id %>'><%= serial.position %></span>.
-        <h3 class="heading"><%=serial.work.title.html_safe + (serial.work.posted? ? "" : " (DRAFT)") %></h3>
+        <h3 class="heading">
+          <%= serial.work.posted? ? serial.work.title : t(".draft_work_title", title: serial.work.title)) %>
```
weeklies, line 9:
```suggestion
          <%= serial.work.posted? ? serial.work.title : t(".draft_work_title", title: serial.work.title) %>
```

PR #5202, opened 2025-05-22
AO3-6060 Tag comments emailed to no longer wranglers

ceithir:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6060

## Purpose

Disable email notification on a reply to a tag comment from someone who is a tag wrangler no more

## Credit

Ceithir (he/him)


sarken (changes requested):
Thanks for working on this!

I'm not sure the check in the mailer is sufficient, however. It will stop the email from being sent, but the issue mentions that the former wrangler _also_ receives a notification in their AO3 inbox when they shouldn't. I think we might need to add something in `notify_parent_comment_owner` to stop that.

https://github.com/otwcode/otwarchive/blob/7d27b34b8aad7fc980e155217805cde004a5d696/app/models/comment.rb#L382-L388

(I suspect we can put the check somewhere that will stop both the inbox comment and the email without having to touch the mailer, although I can see an argument for including it in the mailer and the `update_feedback_in_inbox` and `add_feedback_to_inbox` methods instead of `notify_parent_comment_owner`.)

However, I'll also happily accept just a test showing the mailer change is enough and the inbox comment is no longer an issue!

app/models/comment.rb
```diff
@@ -272,128 +272,128 @@ def record_wrangling_activity
...
+  end
+
+  def not_user_commenter?(parent_comment)
+    (!parent_comment.comment_owner && parent_comment.comment_owner_email && parent_comment.comment_owner_name)
+  end
+
+  def different_owner?(parent_comment)
+    not_user_commenter?(parent_comment) || (parent_comment.comment_owner != self.comment_owner)
+  end
+
```
ceithir, line 360: No change above this point except for indentation being fixed and one function renamed to please the hound.

ceithir:
> I suspect we can put the check somewhere that will stop both the inbox comment and the email without having to touch the mailer

Clearing the forest of if of `notify_parent_comment_owner` seems to have revealed a spot where a well-positioned early return covers all cases indeed.

app/models/comment.rb
```diff
@@ -272,128 +272,128 @@ def record_wrangling_activity
...
+    # if I'm replying to a comment you left for me, mark your comment as replied to in my inbox
+    if self.comment_owner && (inbox_comment = self.comment_owner.inbox_comments.find_by(feedback_comment_id: parent_comment.id))
+      inbox_comment.update(replied_to: true, read: true)
     end
 
-    def notify_parent_comment_owner
-      if self.reply_comment? && !self.unreviewed?
-        parent_comment = self.commentable
-        parent_comment_owner = parent_comment.comment_owner # will be nil if not a user, including if an admin
+    return unless different_owner?(parent_comment)
```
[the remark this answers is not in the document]
ceithir: Allegedly, this bit of code shouldn't be in that function at all, as it does not notify the parent comment's owner the slightest.

But I'm really not familiar enough with comments and inboxes to know if moving it (two lines below or in another function) would be a matter of minutes or the door to a rabbit hole of issues.

spec/models/comment_spec.rb
```diff
@@ -500,9 +500,28 @@
       end
 
       context "direct parent is a tag" do
-        before { create(:comment, :on_tag, pseud: tag_wrangler.default_pseud) }
+        let!(:comment) { create(:comment, :on_tag, pseud: tag_wrangler.default_pseud) }
 
         include_examples "updates last wrangling activity"
+
+        it "notifies the tag wrangler when their comment on a tag is replied to" do
```
sarken: I don't think these tests are in the right place. We're testing replies, so the direct parent is the comment we're replying to, not the tag. The tag is the ultimate parent.

I also think we can chain our matchers and avoid nesting an `expect` inside an `expect` block and checking `ActionMailer::Base` directly to make things a little tidier. Here's what I came up with: https://github.com/otwcode/otwarchive/commit/8deb1b19845c44585e8c1cb6d36895d82edd1e38

PR #5206, opened 2025-05-31
AO3-5949 Remove cookie-based redirection to the last visited page

marcus8448:
(Blocked by #5171)

# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5949

## Purpose

Replaces cookie-based redirection (`store_location` & `redirect_back_or_default`) with other forms of redirects.

Most redirects have been updated to either point to specific paths or redirect to the referrer (if its a non-GET request).

The user login form now takes a `redirect_to` parameter (e.g. `?return_to=/works/12345`).
To ensure that it doesn't redirect anywhere outside the archive the path is checked to start with exactly one `/` and be recognized as a relative URI.

Chapter and work edit/preview/manage cancel buttons are special as there are many ways to navigate to them. It feels bad to press cancel and go to some page you weren’t on before, so these pages keep track of where they generally came from via either a hidden form tag or a url parameter (and redirect appropriately). I’ve hardcoded the sources (e.g. `manage` or `edit-tags`) rather than storing the full paths, but that can be changed if needed.

### Notable redirect changes

**Viewing draft work > Edit > Cancel | Viewing draft work > Edit > Preview > Cancel**
Redirects back to the draft work, instead of user drafts. Creating a draft and cancelling the preview is unchanged.

**Edit Work Page > Edit Chapter X > Cancel**
Redirects back to the edit work page, rather than the work page.

**Manage Chapters Page > Edit > Cancel**
Redirects back to the manage chapters page, rather than the edited chapter's page.

**Edit Work Page > Add Chapter > Cancel**
Redirects back to the edit work page, rather than the work page.

**? > New Orphan > Cancel**
When orphaning…
* One work → Edit work page
* Multiple works → Show multiple works (user page)
* Series → Edit series page
* Pseud → User pseuds page
* User → User page

### Miscellaneous other changes
These can be reverted or split off if necessary.

* The cancel button on the delete user confirmation page did not work previously (it would lead to a 204).
  * Now it redirects to the user page (unchanged from intended functionality).
* Closed an unclosed \<p\> tag in app/views/chapters/manage.html.erb

## Testing Instructions

TBD. Will be posted on JIRA.

## Credit

marcus8448 (he/him)

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -1,5 +1,4 @@
 class CommentsController < ApplicationController
-  skip_before_action :store_location, except: [:show, :index, :new]
   before_action :load_commentable,
                 only: [:index, :new, :create, :edit, :update, :show_comments,
                        :hide_comments, :add_comment_reply,
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -449,7 +449,7 @@ class CommentsController < ApplicationController
     authorize @commentable, policy_class: CommentPolicy if logged_in_as_admin?
     unless (@commentable && current_user_owns?(@commentable)) || (@commentable && logged_in_as_admin? && @commentable.is_a?(AdminPost))
       flash[:error] = ts("What did you want to review comments on?")
-      redirect_back_or_to(root_path)
+      redirect_to(root_path)
       return
     end
 
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -100,7 +100,8 @@ class CommentsController < ApplicationController
     parent = find_parent
 
     return unless parent.respond_to?(:restricted) && parent.restricted? && !(logged_in? || logged_in_as_admin?)
-    redirect_to new_user_session_path(restricted_commenting: true)
+
+    redirect_to new_user_session_path(restricted_commenting: true, return_to: request.fullpath)
   end
 
   # Check to see if the ultimate_parent is a Work or AdminPost, and if so, if it allows
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -650,10 +650,10 @@ class CommentsController < ApplicationController
 
     if commentable.is_a?(Tag)
       redirect_to comments_path(tag_id: commentable.to_param,
-                                add_comment_reply_id: options[:add_comment_reply_id],
-                                delete_comment_id: options[:delete_comment_id],
-                                page: options[:page],
-                                anchor: options[:anchor])
+                  add_comment_reply_id: options[:add_comment_reply_id],
+                  delete_comment_id: options[:delete_comment_id],
+                  page: options[:page],
+                  anchor: options[:anchor])
     else
       if commentable.is_a?(Chapter) && (options[:view_full_work] || current_user.try(:preference).try(:view_full_works))
         commentable = commentable.work
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -129,7 +129,7 @@ class CommentsController < ApplicationController
     admin_settings = AdminSetting.current
 
     return unless admin_settings.guest_comments_off? && guest?
-    
+
     flash[:error] = t("comments.commentable.guest_comments_disabled")
     redirect_back(fallback_location: root_path)
   end
@@ -290,7 +290,7 @@ class CommentsController < ApplicationController
   def new
     if @commentable.nil?
       flash[:error] = ts("What did you want to comment on?")
-      redirect_back_or_default(root_path)
+      redirect_to root_path
     else
       @comment = Comment.new
       @controller_name = params[:controller_name] if params[:controller_name]
@@ -325,7 +325,7 @@ class CommentsController < ApplicationController
   def create
     if @commentable.nil?
       flash[:error] = ts("What did you want to comment on?")
-      redirect_back_or_default(root_path)
+      redirect_back_or_to root_path
     else
       @comment = Comment.new(comment_params)
       @comment.ip_address = request.remote_ip
@@ -405,7 +405,7 @@ class CommentsController < ApplicationController
     elsif unreviewed
       # go back to the rest of the unreviewed comments
       flash[:notice] = ts("Comment deleted.")
-      redirect_back(fallback_location: unreviewed_work_comments_path(@comment.commentable))
+      redirect_back_or_to(unreviewed_work_comments_path(@comment.commentable))
     elsif parent_comment
       flash[:comment_notice] = ts("Comment deleted.")
       redirect_to_comment(parent_comment)
@@ -449,7 +449,7 @@ class CommentsController < ApplicationController
     authorize @commentable, policy_class: CommentPolicy if logged_in_as_admin?
     unless (@commentable && current_user_owns?(@commentable)) || (@commentable && logged_in_as_admin? && @commentable.is_a?(AdminPost))
       flash[:error] = ts("What did you want to review comments on?")
-      redirect_back_or_default(root_path)
+      redirect_back_or_to(root_path)
       return
     end
 
@@ -650,10 +650,10 @@ class CommentsController < ApplicationController
 
     if commentable.is_a?(Tag)
       redirect_to comments_path(tag_id: commentable.to_param,
-                  add_comment_reply_id: options[:add_comment_reply_id],
-                  delete_comment_id: options[:delete_comment_id],
-                  page: options[:page],
-                  anchor: options[:anchor])
+                                add_comment_reply_id: options[:add_comment_reply_id],
+                                delete_comment_id: options[:delete_comment_id],
+                                page: options[:page],
+                                anchor: options[:anchor])
     else
       if commentable.is_a?(Chapter) && (options[:view_full_work] || current_user.try(:preference).try(:view_full_works))
         commentable = commentable.work
--- a/app/controllers/feedbacks_controller.rb
+++ b/app/controllers/feedbacks_controller.rb
@@ -26,7 +26,7 @@ class FeedbacksController < ApplicationController
       @feedback.email_and_send
       flash[:notice] = t("successfully_sent",
         default: "Your message was sent to the Archive team - thank you!")
-      redirect_back_or_default(root_path)
+      redirect_to(@feedback.referer || root_path)
     else
       flash[:error] = t("failure_send",
         default: "Sorry, your message could not be saved - please try again!")
--- a/app/controllers/feedbacks_controller.rb
+++ b/app/controllers/feedbacks_controller.rb
@@ -1,5 +1,4 @@
 class FeedbacksController < ApplicationController
-  skip_before_action :store_location
   before_action :load_support_languages
 
   def new
--- a/app/controllers/kudos_controller.rb
+++ b/app/controllers/kudos_controller.rb
@@ -1,5 +1,4 @@
 class KudosController < ApplicationController
-  skip_before_action :store_location
   before_action :load_parent, only: [:index]
   before_action :check_parent_visible, only: [:index]
   before_action :admin_logout_required, only: [:create]
--- a/app/controllers/orphans_controller.rb
+++ b/app/controllers/orphans_controller.rb
@@ -14,18 +14,23 @@ class OrphansController < ApplicationController
   def new
     if params[:work_id]
       @to_be_orphaned = Work.find(params[:work_id])
+      @return_to = edit_work_path(@to_be_orphaned)
       check_one_owned(@to_be_orphaned, current_user.works)
     elsif params[:work_ids]
       @to_be_orphaned = Work.where(id: params[:work_ids]).to_a
+      @return_to = show_multiple_user_works_path(current_user)
       check_all_owned(@to_be_orphaned, current_user.works)
     elsif params[:series_id]
       @to_be_orphaned = Series.find(params[:series_id])
+      @return_to = edit_series_path(@to_be_orphaned)
       check_one_owned(@to_be_orphaned, current_user.series)
     elsif params[:pseud_id]
       @to_be_orphaned = Pseud.find(params[:pseud_id])
+      @return_to = user_pseuds_path(@to_be_orphaned)
       check_one_owned(@to_be_orphaned, current_user.pseuds)
     else
       @to_be_orphaned = current_user
+      @return_to = user_path(@to_be_orphaned)
     end
   end
 
--- a/app/controllers/related_works_controller.rb
+++ b/app/controllers/related_works_controller.rb
@@ -31,11 +31,11 @@ class RelatedWorksController < ApplicationController
     unless @user
       if current_user_owns?(@child)
         flash[:error] = ts("Sorry, but you don't have permission to do that. Try removing the link from your own work.")
-        redirect_back_or_to user_related_works_path(current_user)
       else
         flash[:error] = ts("Sorry, but you don't have permission to do that.")
-        redirect_back_or_to root_path
       end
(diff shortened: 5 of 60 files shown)


app/controllers/feedbacks_controller.rb
```diff
@@ -26,7 +25,7 @@ def create
       @feedback.email_and_send
       flash[:notice] = t("successfully_sent",
         default: "Your message was sent to the Archive team - thank you!")
-      redirect_back_or_default(root_path)
+      redirect_to(@feedback.referer || root_path)
```
marcus8448: (This referer redirect is ok as the host is checked at line 22)

app/controllers/admin/admin_users_controller.rb
```diff
@@ -103,7 +103,7 @@ def update
     else
       flash[:error] = ts("The user %{name} could not be updated: %{errors}", name: params[:id], errors: @user.errors.full_messages.join(" "))
     end
-    redirect_to request.referer || root_path
+    redirect_back_or_to admin_users_path
```
Bilka2: Would it be viable to add the user as param here, similar to what the Manage Roles button does on the admin user dashboard?

app/controllers/feedbacks_controller.rb
Bilka2: We could still run into the host header <-> URL mismatch (where the host header is archiveofourown.org because that's what we rewrite it to in nginx, but the URL is a proxy), which [would throw ](https://github.com/rails/rails/blob/9204eb520c2784ca7a1da9a4884aad21c59088fd/actionpack/lib/action_controller/metal/redirecting.rb#L233-L234)ActionController::Redirecting::UnsafeRedirectError. We could avoid that by passing `allow_other_host` here if we are sure that this is a safe host or alternatively wrap it in [url_from](https://api.rubyonrails.org/classes/ActionController/Redirecting.html#method-i-url_from) to silence the error if it would occur

app/controllers/kudos_controller.rb
```diff
@@ -42,7 +41,7 @@ def create
       respond_to do |format|
         format.html do
           flash[:kudos_notice] = t(".success")
-          redirect_to request.referer and return
+          redirect_back_or_to @kudo.commentable || root_path and return
```
Bilka2: It succeeded, in which case would commentable be nil?

app/controllers/orphans_controller.rb
```diff
@@ -14,18 +14,23 @@ def index
   def new
     if params[:work_id]
       @to_be_orphaned = Work.find(params[:work_id])
+      @return_to = edit_work_path(@to_be_orphaned)
       check_one_owned(@to_be_orphaned, current_user.works)
     elsif params[:work_ids]
       @to_be_orphaned = Work.where(id: params[:work_ids]).to_a
+      @return_to = show_multiple_user_works_path(current_user)
```
Bilka2: Would it make sense to include the work ids here?

app/controllers/orphans_controller.rb
```diff
@@ -14,18 +14,23 @@ def index
...
       @to_be_orphaned = Series.find(params[:series_id])
+      @return_to = series_path(@to_be_orphaned)
       check_one_owned(@to_be_orphaned, current_user.series)
     elsif params[:pseud_id]
       @to_be_orphaned = Pseud.find(params[:pseud_id])
+      @return_to = user_pseuds_path(@to_be_orphaned)
       check_one_owned(@to_be_orphaned, current_user.pseuds)
     else
       @to_be_orphaned = current_user
+      @return_to = show_multiple_user_works_path(@to_be_orphaned)
```
Bilka2, line 33: Orphaning the entire user used to be a button the preferences page, but we removed it with https://otwarchive.atlassian.net/browse/AO3-6388 so technically people shouldnt be able to reach this anymore. Since it would indeed be reachable by not having any works selected on the show multiple page this seems like a sensible fallback though :+1:

app/controllers/comments_controller.rb
```diff
@@ -101,7 +100,8 @@ def check_if_restricted
     parent = find_parent
 
     return unless parent.respond_to?(:restricted) && parent.restricted? && !(logged_in? || logged_in_as_admin?)
-    redirect_to new_user_session_path(restricted_commenting: true)
+
+    redirect_to new_user_session_path(restricted_commenting: true, return_to: request.fullpath)
```
Bilka2, line 104: Should return_to also be set for the other `new_user_session_path` redirects here in the controller? (And also like, generally everywhere we link to or redirect to the page? E.g. where we make the "You can only see your own drafts, sorry!" error in the works controller etc) Because if the user just forgot to log in, they would want to return to the page they tried to access after logging in

app/controllers/comments_controller.rb
```diff
@@ -131,14 +131,14 @@ def check_guest_comment_admin_setting
     return unless admin_settings.guest_comments_off? && guest?
 
     flash[:error] = t("comments.commentable.guest_comments_disabled")
-    redirect_back(fallback_location: root_path)
+    redirect_back_or_to @commentable
```
Bilka2: Should this maybe be `find_parent`, similar to check_blocked? Same for check_guest_replies_preference, check_frozen, check_hidden_by_admin and check_not_replying_to_spam.

Don't change this if this would mean modifying tests though

app/views/users/sessions/_passwd_small.html.erb
```diff
@@ -1,6 +1,6 @@
 <%# We need to override the ids to avoid accessibility issues on the new user session page,
 which has a second user session form %>
-<%= form_for(User.new, url: user_session_path, html: { id: "new_user_session_small" }) do |f| %>
+<%= form_for(User.new, url: user_session_path(return_to: request.path.start_with?(new_user_session_path) ? params[:return_to] : request.fullpath), html: { id: "new_user_session_small" }) do |f| %>
```
Bilka2: I'm a bit confused by `request.path`, I can't find evidence of that existing in Rails docs?

Comment also applies to other usage of it in the various files here

features/support/capybara.rb
```diff
@@ -30,6 +30,10 @@
   config.server_port = CAPYBARA_PORT
 end
 
+Capybara.register_driver :rack_test_no_data_method do |app|
+  Capybara::RackTest::Driver.new(app, respect_data_method: false)
+end
+
```
Bilka2, line 36: I'm confused as to why this showing up in this diff, it's already there in master/the commit this branch is based on...?

features/support/hooks.rb
```diff
@@ -58,6 +58,14 @@
...
 end
 
+Before "@no-js-emulation" do
+  Capybara.current_driver = :rack_test_no_data_method
+end
+
+After "@no-js-emulation" do
+  Capybara.use_default_driver
+end
+
```
Bilka2, line 68: Same confused diff as the driver...

app/controllers/users_controller.rb
```diff
@@ -266,6 +265,7 @@ def destroy
   end
 
   def delete_confirmation
+    redirect_to user_path(current_user) if logged_in?
```
Bilka2, line 248: How did this come up to change here? Since the user can only end up here after a `@user.destroy` how would they get here if logged in? Is it related to some other change you did?

spec/controllers/collection_items_controller_spec.rb
```diff
@@ -275,12 +275,7 @@
       subject { get :new, params: { work_id: work.id } }
       let(:work) { create(:work) }
       let(:success) { expect(response).to render_template("new") }
-      let(:redirects_to_login) do
-        it_redirects_to_with_error(
-          new_user_session_path,
-          "Sorry, you don't have permission to access the page you were trying to reach. Please log in."
-        )
-      end
+      let(:redirects_to_login) { it_redirects_to_user_login_with_error }
```
Bilka2: Could you use `it_redirects_to_user_login_with_error` directly instead of keeping this variable?

app/controllers/related_works_controller.rb
```diff
@@ -27,17 +27,16 @@ def show
...
       if current_user_owns?(@child)
         flash[:error] = ts("Sorry, but you don't have permission to do that. Try removing the link from your own work.")
-        redirect_back_or_default(user_related_works_path(current_user))
-        return
+        redirect_to user_related_works_path(current_user)
       else
         flash[:error] = ts("Sorry, but you don't have permission to do that.")
-        redirect_back_or_default(root_path)
-        return
+        redirect_to related_work_path(@related_work)
```
Bilka2: Why redirect to @related_work here but not below in `destroy`? In both cases, the user doesn't own the related work, so I think redirecting to it would be confusing

app/controllers/related_works_controller.rb
```diff
@@ -57,29 +56,28 @@ def destroy
...
     if params[:user_id].blank?
       flash[:error] = ts("Whose related works were you looking for?")
-      redirect_back_or_default(search_people_path)
+      redirect_back_or_to search_people_path
     else
       @user = User.find_by(login: params[:user_id])
       if @user.blank?
         flash[:error] = ts("Sorry, we couldn't find that user")
-        redirect_back_or_default(root_path)
+        redirect_back_or_to search_people_path
```
Bilka2: Our latest convention would be to 404 here (refer to e.g. https://otwarchive.atlassian.net/browse/AO3-6989) but when we used to redirect to `search_people_path` for not found users we would go there directly, without the referer. I think that's better in this case too - `load_user` is only called from index so it'd be strange to be thrown back onto your previous page when you're just following a link to something, not trying to execute an action

spec/controllers/path_cleaner_spec.rb
Bilka2, line 1: This should be in the spec/lib subfolder

spec/controllers/path_cleaner_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+describe PathCleaner do
+  include PathCleaner
+
+  describe "#relative_uri" do
+    it "allows valid relative paths" do
+      expect(relative_path("/")).to_not eq(nil)
+      expect(relative_path("/path")).to_not eq(nil)
+      expect(relative_path("/foo/bar/")).to_not eq(nil)
+      expect(relative_path("/path?query=true")).to_not eq(nil)
+      expect(relative_path("/path#anchor")).to_not eq(nil)
```
Bilka2: Could you check the actual return values here, just in case we ever decide to adjust the implementation (e.g. to return parsed.to_s)

features/users/authenticate_users.feature
```diff
@@ -335,3 +335,77 @@ Feature: User Authentication
...
+
+  @no-js-emulation
+  Scenario: Log out with javascript disabled redirects to page prior to log out
+    Given I am logged in
+      And I am on the works page
+    When I follow "Log Out"
+    Then I should see "Are you sure you want to log out?"
+    When I press "Yes, Log Out"
+    Then I should see "Successfully logged out."
+      And I should see "Recent Works"
```
Bilka2: This is a bit more explicit
```suggestion
      And I should be on the works page
```
also applies to the other scenarios

features/users/authenticate_users.feature
```diff
@@ -335,3 +335,77 @@ Feature: User Authentication
...
+      And I should see "This work is only available to registered users of the Archive."
+    When I fill in "Username or email:" with "test" within "#main"
+      And I fill in "Password:" with "password" within "#main"
+      And I press "Log in" within "#main"
+    Then I should see "Secret"
+
+  Scenario: User is redirected to previous page after using the small login
+    Given I am logged in as "test" with password "password"
+      And I post the locked work "Secret"
+      And I log out
```
Bilka2: This seems redundant?

lib/path_cleaner.rb
```diff
@@ -0,0 +1,8 @@
+module PathCleaner
+  def relative_path(uri)
```
Bilka2, line 2: Comment from an older PR: https://github.com/otwcode/otwarchive/pull/2825/files#r464128883

> maybe also port/username/password

I think we should be able to check user and password directly on the URI - not sure what to do about port if anything

app/controllers/admin/admin_users_controller.rb
Bilka2, line 1: Could you also replace the HTTP_REFERER redirect in line 200 with `redirect_back_or_to` with the user's admin dashboard as the fallback (or maybe just that directly)

app/controllers/works_controller.rb
Bilka2, line 1: Could you replace the redirects to `HTTP_REFERER` in this file with `redirect_back_or_to` as well? The one in the share action can act the same what you did for bookmarks

app/controllers/users_controller.rb
Bilka2, line 1: Could you replace the referer redirect in end_banner with `redirect_back_or_to` as well please

app/controllers/users_controller.rb
marcus8448: It's for if a user deletes their account and then immediately logs into a different account. Without this patch the user will be brought back to this "account deleted" page (which could be suprising).

features/support/capybara.rb
marcus8448: This branch is ~a month behind (based on cd8fae0f9d92464233ee5f336b1cd64659d4f595). I think git isn't complaining about conflicts as it's probably a 1-to-1 match to the changes in 0d401f80e409d4ffc6fa980252d92208ede3d260 (I developed the patch for this PR, but it was also useful for #5332 which got merged first). I think it should go away when I pull in master.

app/controllers/users_controller.rb
Bilka2: I see! Is this something that could be tested in cucumber tests?

sarken: There's a separate issue for this, so we should add it to the issue description/title: https://otwarchive.atlassian.net/browse/AO3-6402

Bilka2: Extra note that it would be nice to have automated tests for the "other issues this fixes" in general, e.g. also the banner one, if possible

app/controllers/related_works_controller.rb
marcus8448: I've updated it to redirect directly to people search. Would you like it to 404 to follow the new convention?

app/controllers/feedbacks_controller.rb
marcus8448: :facepalm: I totally misunderstood the purpose of `PERMITTED_HOSTS`. I've wrapped it with `url_from`.

lib/path_cleaner.rb
marcus8448: I've added checks for username/password/port, but I couldn't come up with any valid URI strings to test that would match without also failing the host and scheme checks.

app/controllers/related_works_controller.rb
```diff
@@ -57,29 +56,28 @@ def destroy
     unless current_user_owns?(@child)
       if @user
         flash[:error] = ts("Sorry, but you don't have permission to do that. You can only approve or remove the link from your own work.")
-        redirect_back_or_default(user_related_works_path(current_user))
-        return
+        redirect_to related_work_path(@related_work)
```
Bilka2: In this case I think this should be `redirect_to user_related_works_path(current_user)` - that way the user can get to the right related work to remove link. So same as in `update`

app/controllers/related_works_controller.rb
Bilka2: Let's keep it as the people search for this PR

lib/path_cleaner.rb
Bilka2: I think that's fine, in this case I'd rather have the redundant checks than not

Bilka2 (approved):
Thank you so much!

PR #5208, opened 2025-06-02
AO3-4676 Extend tests in invitations_controller

Hamham6:
# Pull Request Checklist

* [ ] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)? N/A only tests

## Issue

https://otwarchive.atlassian.net/browse/AO3-4676

## Purpose

Covers lines of invitations_controller that aren't covered yet.

## Credit

Hamham6, it/its


Hamham6:
I can't figure out how to get delete/save to be false to test that logic branch.

ceithir:
> I can't figure out how to get delete/save to be false to test that logic branch.

Did you try:
```rb
allow_any_instance_of(Invitation).to receive(:save).and_return(false)
```
?

Hamham6:
I can role back the last commit if the dif is too big.

spec/controllers/invitations_controller_spec.rb
```diff
@@ -30,23 +30,53 @@
   end
 
   describe "GET #manage" do
-    context "when admin does not have correct authorization" do
+    subject { get :manage, params: { user_id: user.login } }
+
+    authorized_roles = InvitationPolicy::EXTRA_INFO_ROLES + %w[legal]
```
Hamham6: Is `InvitationPolicy::EXTRA_INFO_ROLES` supposed to be the policy for manage/invite_friend/update/destroy invitations? The current behavior lets all those roles and `legal`, even tho previously the spec only tested `policy_and_abuse` versus no role.

spec/controllers/invitations_controller_spec.rb
```diff
@@ -30,23 +30,53 @@
...
+
+          expect(response).to render_template("manage")
+        end
+
+        %i[unsent unredeemed redeemed].each do |status|
+          it "includes invitation status when invitation is #{status}" do
+            get :manage, params: { user_id: user.login, status: status }
+
+            expect(response).to render_template("manage")
+            expect(user.invitations).to eq(user.invitations.send(status))
```
Hamham6: I'm not sure what invitation status means so this test is just copyed from the controller code. I don't know what behavior it's actually testing.

spec/controllers/invitations_controller_spec.rb
```diff
@@ -30,23 +30,53 @@
...
+    end
+
+    context "when logged in as an admin with no role" do
       it "redirects with error" do
         admin.update!(roles: [])
         fake_login_admin(admin)
-        get :manage, params: { user_id: user.login }
+        subject
 
         it_redirects_to_with_error(new_user_session_path, "Sorry, you don't have permission to access the page you were trying to reach. Please log in.")
```
Hamham6: Should the error messsages for admins without roles and without authorized roles be changed to `"Sorry, only an authorized admin can access the page you were trying to reach."`? Also, should the redirect be changed to `root_url` or is it supposed to be `new_user_session_path`? These differences are keeping me from being able to use `access_shared_examples.rb`.

spec/controllers/invitations_controller_spec.rb
```diff
@@ -194,39 +261,191 @@
...
+          admin.update!(roles: [admin_role])
+          fake_login_admin(admin)
+        end
+
+        it "updates invitation and redirects to find admin invitations path" do
+          new_email = invitee.email
+          put :update, params: { id: invitation.id, invitation: { invitee_email: new_email } }
+
+          it_redirects_to_with_notice(find_admin_invitations_path("invitation[token]" => invitation.token), "Invitation was successfully sent.")
+          expect(invitation.invitee_email).to eq(new_email)
```
Hamham6: I don't know why this doesn't work TT

spec/controllers/invitations_controller_spec.rb
```diff
@@ -194,39 +261,191 @@
...
+        end
+
+        it "errors if email is missing" do
+          allow_any_instance_of(Invitation).to receive(:update).and_return(false)
```
Hamham6: Even if the email is blank, the error and show template aren't loaded unless the invitation failed to save or the email didn't change. Or possibly I'm misunderstanding the code, but without this line this test fails because nothing renders.

spec/controllers/invitations_controller_spec.rb
```diff
@@ -194,39 +261,191 @@
...
+    authorized_roles = InvitationPolicy::EXTRA_INFO_ROLES + %w[legal]
+
+    authorized_roles.each do |admin_role|
+      context "when logged in as an admin with role #{admin_role}" do
+        before do
+          admin.update!(roles: [admin_role])
+          fake_login_admin(admin)
+        end
+
+        context "when invitation creator is a user" do
```
Hamham6: I am confused over what
`if @user.is_a?(User)
      redirect_to user_invitations_path(@user)
    else
      redirect_to admin_invitations_path
    end`
(lines 81-85 of invitations_controller.rb) means. Does `@user` here refer to the creator of the invitation or to who is deleting it? Can users delete invitations?

spec/controllers/invitations_controller_spec.rb
```diff
@@ -194,39 +261,191 @@
...
     end
 
-    it "allows admin to delete invitations" do
-      invitation = create(:invitation)
-      admin.update!(roles: ["policy_and_abuse"])
-      fake_login_admin(admin)
- 
-      delete :destroy, params: { id: invitation.id }
-      it_redirects_to_with_notice(admin_invitations_path, "Invitation successfully destroyed")
+    context "when logged in as an admin with no role" do
```
Hamham6: Admins without roles can apparently delete invitations? As can admins without authorized roles? An error pops up `"Sorry, you don't have permission to access the page you were trying to reach. Please log in."` and redirects to `new_user_session_path`, but the invitation is still deleted? The same thing seems to be happening for users and guests.

spec/controllers/invitations_controller_spec.rb
Bilka2: No, this is restricted by `InvitationsController#check_permission` which uses `UserPolicy#can_manage_users?` because changing/viewing a user's invitations in considered managing that user

spec/controllers/invitations_controller_spec.rb
Bilka2: Users can have invitations that they can send to other users so they can skip the invite queue. status goes unsent (user has invite available) -> unredeemed (user has sent invite to friend, but friend hasn't used it yet) -> redeemed (friend has used invite). These status options filter the large table on the page to only show invites with that status.

I think it would be better to test the filtering/status with a cucumber test - making invites with the right status and making sure they show up in the correct filters

spec/controllers/invitations_controller_spec.rb
Bilka2: Yes please. It should be possible to do that by using `authorize` in `InvitationsController#check_permission`. [authorize docs](https://github.com/varvet/pundit?tab=readme-ov-file#policies), in case it's new for you (or ask here or via email and I can explain more!)

spec/controllers/invitations_controller_spec.rb
```diff
@@ -152,22 +182,59 @@
...
+    authorized_roles = InvitationPolicy::EXTRA_INFO_ROLES + %w[legal]
+
+    authorized_roles.each do |admin_role|
+      context "when logged in as an admin with role #{admin_role}" do
+        before do
+          admin.update!(roles:[admin_role])
+          fake_login_admin(admin)
+        end
+
+        it "sends invitation and redirects to invitation path" do
```
Bilka2: It would be nice if this also checked that the invitation was created

spec/controllers/invitations_controller_spec.rb
```diff
@@ -152,22 +182,59 @@
...
+        it "errors if email is missing" do
+          post :invite_friend, params: { user_id: user.id, id: invitation.id, invitation: { invitee_email: nil, number_of_invites: 1 } }
+
+          expect(response).to render_template("show")
+          expect(flash[:error]).to match("Please enter an email address.")
+        end
 
-      expect(response).to render_template("show")
-      expect(flash[:error]).to match("Please enter an email address.")
+        it "renders #show if the invitation fails to save" do
```
Bilka2: This should check that no flash is set

spec/controllers/invitations_controller_spec.rb
Bilka2: Rails caches records. Try `invitation.reload.invitee_email`

spec/controllers/invitations_controller_spec.rb
```diff
@@ -194,39 +261,191 @@
...
   describe "PUT #update" do
-    it "updates and resends invitations" do
-      invitee = create(:user)
-      invitation = create(:invitation)
-      admin.update!(roles: ["policy_and_abuse"])
-      fake_login_admin(admin)
-      put :update, params: { id: invitation.id, invitation: { invitee_email: invitee.email } }
+    let(:invitee) { create(:user) }
+    let(:invitation) { create(:invitation) }
+    subject { put :update, params: { id: invitation.id, invitation: { invitee_email: invitee.email } } }
```
Bilka2: It's a bit strange that this test uses the email address of an existing user to send them an invitation. It'd be better to use an unused email address

spec/controllers/invitations_controller_spec.rb
```diff
@@ -152,22 +182,59 @@
 
   describe "POST #invite_friend" do
     let(:invitation) { create(:invitation) }
+    subject { post :invite_friend, params: { user_id: user.id, id: invitation.id, invitation: { invitee_email: user.email, number_of_invites: 1 } } }
```
Bilka2: It's a bit strange that this test uses the email address of an existing user to send them an invitation. It'd be better to use an unused email address

Bilka2: Also, number_of_invites is not a valid param here

spec/controllers/invitations_controller_spec.rb
```diff
@@ -194,39 +261,191 @@
...
+    authorized_roles.each do |admin_role|
+      context "when logged in as an admin with role #{admin_role}" do
+        before do
+          admin.update!(roles: [admin_role])
+          fake_login_admin(admin)
+        end
+
+        it "updates invitation and redirects to find admin invitations path" do
+          new_email = invitee.email
+          put :update, params: { id: invitation.id, invitation: { invitee_email: new_email } }
```
Bilka2: Why not `subject`?

spec/controllers/invitations_controller_spec.rb
Bilka2: Sounds like the behaviour should be changed, it should always error with an empty email.

spec/controllers/invitations_controller_spec.rb
```diff
@@ -194,39 +261,191 @@
...
+          expect(response).to render_template("show")
+          expect(flash[:notice]).to be(nil)
+        end
+
+        it "renders #show if the update did not change invitee_email" do
+          put :update, params: { id: invitation.id, invitation: { invitee_email: invitation.invitee_email } }
+
+          expect(response).to render_template("show")
+          expect(flash[:notice]).to be(nil)
+        end
```
Bilka2: I'm a bit wary about cementing a behaviour in a test that is hard to reason about/explain. I think the `invitee_email_changed?` is just a workaround to get the "Please enter an email address." error more reliably. At least it looks like that in the commit that introduced it, cd80943d

spec/controllers/invitations_controller_spec.rb
```diff
@@ -194,39 +261,191 @@
...
+      end
+    end
+
+    context "when logged in as a user" do
+      it "updates invitation and redirects to user path" do
```
Bilka2: This is not the user path, it's the invitation (scoped by user)

spec/controllers/invitations_controller_spec.rb
Bilka2: Users can't delete invitations. `@user` here is the creator of the invite.

The creator of an invitation can be a user, an admin or the queue, refer to `InvitationsHelper#creator_link`. So that piece of code only redirects to the users invitations page if it was a user who created the invite

spec/controllers/invitations_controller_spec.rb
```diff
@@ -194,39 +261,191 @@
...
+          it "deletes invitation and redirects to admin invitations path" do
+            invitation_creator = create(:admin)
+            invitation.creator = invitation_creator
+            subject
+
+            it_redirects_to_with_notice(admin_invitations_path(), "Invitation successfully destroyed")
+          end
+        end
+
+        it "errors if invitation fails to destroy" do
```
Bilka2: This should also redirect

spec/controllers/invitations_controller_spec.rb
```diff
@@ -194,39 +261,191 @@
...
+        fake_login_admin(admin)
+      end
+
+      it "redirects with error and does not delete the invitation" do
+        admin.update!(roles: [])
+        fake_login_admin(admin)
+        subject
+
+        it_redirects_to_with_error(new_user_session_path, "Sorry, you don't have permission to access the page you were trying to reach. Please log in.")
+        expect(invitation).to exist
```
Bilka2: This check tries to call `exist?` or `exists?` on the object which I don't think does anything sensible (it's not defined for the record in Rails). Try making a `find` for the invitation instead, then the exist check will hit [a Rails method](https://guides.rubyonrails.org/active_record_querying.html#existence-of-objects).

Bilka2 (changes requested):

spec/controllers/invitations_controller_spec.rb
Bilka2: Found a Jira issue for this behavior change: https://otwarchive.atlassian.net/browse/AO3-5903

Hamham6: Can I fold AO3-5903 into this PR or would it be more convenient for yall if I submitted a blocking PR and handle AO3-5903 separately? I'm fine with either!

spec/controllers/invitations_controller_spec.rb
```diff
@@ -7,226 +7,333 @@
...
-        fake_login_admin(admin)
-        get :index, params: { user_id: user.login }
+      it "succeeds when logged in as the invitation owner" do
+        invitation_owner = create(:user)
+        user_invitation = create(:invitation, creator: invitation_owner)
+        fake_login_known_user(invitation_owner)
+        get :show, params: { user_id: invitation_owner.login, id: user_invitation.id }
 
-        expect(response).to render_template("index")
+        expect(response).to render_template("show")
```
Hamham6, line 71: This test, and some other tests with the invitation owner succeeding, are reporting the render_template is "[]" and failing. I've tryed rephrasing the code, and looking it up, but I can't figure out what's happenning.

spec/controllers/invitations_controller_spec.rb
```diff
@@ -7,226 +7,333 @@
...
-    context "with id parameters" do
-      context "when logged out" do
-        it "redirects with error" do
-          get :show, params: { id: invitation.id }
+    subject { post :create, params: { user_id: user.login, invitation: { invitee_email: user.email } } }
 
-          it_redirects_to_with_error(new_user_session_path, "Sorry, you don't have permission to access the page you were trying to reach. Please log in.")
-        end
-      end
+    it_behaves_like "an action only authorized admins can access", authorized_admin_roles # fsr success_admin isn't getting to "an action authorized admins can access"
```
Hamham6, line 258: `success_admin` is just not staying defined in the `shared_example`. I suspect it's something to do with rspec contexts, but I'm not getting it vv

spec/controllers/invitations_controller_spec.rb
```diff
@@ -7,226 +7,333 @@
...
-          it_redirects_to_with_error(new_user_session_path, "Sorry, you don't have permission to access the page you were trying to reach. Please log in.")
-        end
+    context "when logged in as the invitation owner" do
+      it "succeeds" do
+        owner = user
+        invite = create(:invitation, creator: owner)
+        fake_login_known_user(owner)
+        post :invite_friend, params: { user_id: owner.id, id: invite.id, invitation: { invitee_email: "not_a_user@example.com" } }
+
+        it_redirects_to_with_notice(user_path(owner), "Invitation was successfully sent.") # TODO notice isn't showing up
```
Hamham6, line 190: and this notice is getting `nil`, but only when it's the invitation owner for some reason! Do I need to reload anything to get the flash?

app/controllers/invitations_controller.rb
```diff
@@ -12,7 +12,11 @@ def load_invitation
 
   def check_permission
     @user = User.find_by(login: params[:user_id])
-    access_denied unless policy(User).can_manage_users? || @user.present? && @user == current_user
+    if logged_in_as_admin?
+      authorize @invitations, :can_manage_users?, policy_class: UserPolicy
+    else
+      access_denied unless @user&.equal?(current_user)
```
Hamham6, line 18: I...disagree that this is redundent?_?

Bilka2: You can call `equal?` on nil, so it's indeed redundant: nothing will error if `@user` is nil and `equal?` is called on it, so we don't need to avoid that with the safe navigation

However, this is the wrong method to use here. We want to know if it's the same user, not whether it is the exact same user object, [which is what `equal?` checks](https://stackoverflow.com/questions/7156955/whats-the-difference-between-equal-eql-and/7157051#7157051). The previous code with `@user.present? && @user == current_user` had the correct behaviour. This will also fix the test failures below, since they're caused by getting redirected by `access_denied`

spec/support/shared_examples/access_shared_examples.rb
Bilka2, line 1: Please leave these shared examples as is, if we use the shared example then we always want to test the "wrong role", "right role" and "no role" case together in one go.

PR #5212, opened 2025-06-04
AO3-6515 Fix 500 error for work end notes without posted chapters

weeklies:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6515

## Purpose

- Fix 500 error for work end notes without posted chapters. 
- Do not use instance variable `@work` for helper function. 
- Fix issue in view not having right syntax highlighting in GitHub.
- i18n

## Testing Instructions

See issue.

### DIFF
--- a/app/helpers/works_helper.rb
+++ b/app/helpers/works_helper.rb
@@ -91,14 +91,12 @@ module WorksHelper
   end
 
   def get_endnotes_link
-    if current_page?({ controller: "chapters", action: "show" })
-      if @work.posted?
-        chapter_path(@work.last_posted_chapter.id, anchor: 'work_endnotes')
-      else
-        chapter_path(@work.last_chapter.id, anchor: 'work_endnotes')
-      end
+    return "#work_endnotes" unless current_page?({ controller: "chapters", action: "show" })
+
+    if @work.posted? && @work.last_posted_chapter
+      chapter_path(@work.last_posted_chapter.id, anchor: "work_endnotes")
     else
-      "#work_endnotes"
+      chapter_path(@work.last_chapter.id, anchor: "work_endnotes")
     end
   end
 
--- a/app/helpers/works_helper.rb
+++ b/app/helpers/works_helper.rb
@@ -90,13 +90,13 @@ module WorksHelper
     link_to ts("Mark for Later"), mark_for_later_work_path(work)
   end
 
-  def get_endnotes_link
+  def get_endnotes_link(work)
     return "#work_endnotes" unless current_page?({ controller: "chapters", action: "show" })
 
-    if @work.posted? && @work.last_posted_chapter
-      chapter_path(@work.last_posted_chapter.id, anchor: "work_endnotes")
+    if work.posted? && work.last_posted_chapter
+      chapter_path(work.last_posted_chapter.id, anchor: "work_endnotes")
     else
-      chapter_path(@work.last_chapter.id, anchor: "work_endnotes")
+      chapter_path(work.last_chapter.id, anchor: "work_endnotes")
     end
   end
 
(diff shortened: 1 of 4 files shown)


app/views/works/_work_header_notes.html.erb
```diff
@@ -57,19 +57,23 @@
...
+    <p class="jump">
+      <% if @work.endnotes.present? && related_works_link %>
+        <%= t(".jump.endnotes_and_related_works_html", endnotes_link: endnotes_link, related_works_link: related_works_link) %>
+      <% elsif !@work.endnotes.blank? %>
+        <%= t(".jump.endnotes_html", endnotes_link: endnotes_link) %>
+      <% elsif related_works_link %>
+        <%= t(".jump.related_works_html", related_works_link: related_works_link) %>
+      <% end %>
+    </p>
   <% end %>
```
ceithir, line 78: Assuming I haven't mixed anything up, can be slightly simplified as:
```erb
<% if @work.endnotes.present? %>
  <% endnotes_link = link_to (@work.notes.blank? ? t(".jump.notes") : t(".jump.more_notes")), get_endnotes_link(@work) %>
  <p class="jump">
    <% if related_works_link %>
      <%= t(".jump.endnotes_and_related_works_html", endnotes_link: endnotes_link, related_works_link: related_works_link) %>
    <% else %>
      <%= t(".jump.endnotes_html", endnotes_link: endnotes_link) %>
    <% end %>
  </p>
<% elsif related_works_link %>
  <p class="jump"><%= t(".jump.related_works_html", related_works_link: related_works_link) %></p>
<% end %>
```

app/helpers/works_helper.rb
```diff
@@ -90,15 +90,13 @@ def mark_for_later_link(work)
...
-        chapter_path(@work.last_chapter.id, anchor: 'work_endnotes')
-      end
+  def get_endnotes_link(work)
+    return "#work_endnotes" unless current_page?({ controller: "chapters", action: "show" })
+
+    if work.posted? && work.last_posted_chapter
+      chapter_path(work.last_posted_chapter.id, anchor: "work_endnotes")
     else
-      "#work_endnotes"
+      chapter_path(work.last_chapter.id, anchor: "work_endnotes")
```
ceithir, line 99: I guess there's no edge case where `last_chapter`can be nil?

app/views/works/_work_header_notes.html.erb
weeklies: I didn't really like the duplicated `<p>` tags, so I went with this.

app/helpers/works_helper.rb
weeklies: I think there should be a draft chapter to fall back on.

app/views/works/_work_header_notes.html.erb
brianjaustin: I found the current version more confusing when in diff mode, but I think it's OK after looking at the code in place. I would ask that you change `!@work.endnotes.blank?` to `@work.endnotes.present?` to be more consistent with the previous condition

app/views/works/_work_header_notes.html.erb
```diff
@@ -57,19 +57,23 @@
...
-  <% elsif !@work.endnotes.blank? %>
-    <p class="jump">(<%= ts("See the end of the work for ") %><%= endnotes_link %>.)</p>
-  <% elsif related_works_link %>
-    <p class="jump">(<%= ts("See the end of the work for ") %><%= related_works_link %>.)</p>
+  <% if @work.endnotes.present? || related_works_link %>
+    <% endnotes_link = link_to (@work.notes.blank? ? t(".jump.notes") : t(".jump.more_notes")), get_endnotes_link(@work) %>
+    <p class="jump">
+      <% if @work.endnotes.present? && related_works_link %>
+        <%= t(".jump.endnotes_and_related_works_html", endnotes_link: endnotes_link, related_works_link: related_works_link) %>
+      <% elsif !@work.endnotes.blank? %>
```
weeklies:
```suggestion 
      <% elsif @work.endnotes.present? %>
```

PR #5213, opened 2025-06-04
AO3-6484 Try to use loaded pseuds in css classes helper

weeklies:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6484

## Purpose

Try bypassing the cache for CSS classes when the pseuds are already loaded

## Testing Instructions

@tickinginstant noted that this should be in a release without any other performance changes, so that it’s easier to check if it’s an improvement.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -557,7 +557,7 @@ module ApplicationHelper
     # series.
     return [] if creation.is_a?(Work) && creation.unrevealed?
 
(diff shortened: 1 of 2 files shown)


brianjaustin (approved):
LGTM, but one question: do you know which ticket the caching the issue refers to was removed under?

weeklies:
Sorry @brianjaustin, I'm not sure either.

PR #5217, opened 2025-06-10
AO3-7009 Dark Mode skin

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-7009

## Purpose

Adds the CSS for the new Dark Mode skin to the repo. Refer to the issue and the included README for more information on the reason and set up.

Some notable differences from Reversi:
- a lighter shade of blue that meets accessibility guidelines for contrast when used against the skin's dark grey background
- visited links that are a lighter shade of grey, also to meet guidelines for contrast
- different shadows and borders on various elements, mostly due to rearranging the cascade, but there was a deliberate change to the border on the filters so they'd stand out a bit more, especially on mobile 
- color-coded text and borders for error messages and other notices, similar to the default site skin (in previous versions of the skin that have been in circulation for the past several years, I used low contrast colors, but I have revised them to meet accessibility standards given the other changes)
- form elements (e.g., checkboxes and dropdown menus) that use your browser's default dark styling

## Testing Instructions

Apply the skin and go to your favorite obscure Archive pages and see if anything is more broken than it is in Default and/or Reversi.

## Credit

Sarken, she/her


public/stylesheets/masters/dark_mode/dark_mode_site_screen_.css
```diff
@@ -0,0 +1,750 @@
...
+#header .actions a:hover,
+#header .actions a:focus,
+#header .dropdown:hover a,
+#header .open a,
+#header .menu {
+  background: #2a2a2a;
+}
+
+#header .menu {
+  box-shadow: 1px 1px 3px -1px #000;
```
brianjaustin: CSS3 property, so should be indented 4 instead of 2 spaces
```suggestion
    box-shadow: 1px 1px 3px -1px #000;
```

public/stylesheets/masters/dark_mode/dark_mode_site_screen_.css
```diff
@@ -0,0 +1,750 @@
...
+a:link,
+a:visited:hover,
+.heading {
+  color: #fff;
+}
+
+a:visited {
+  color: #c2c2c2;
+}
+
```
brianjaustin, line 22: The color probably still works, but just to make sure, you don't need to override https://github.com/otwcode/otwarchive/blob/900c8e833ced9a1b027fa0e5ab5701396e19eaab/public/stylesheets/site/2.0/02-elements.css#L37-L39 right?

brianjaustin: Similar question for https://github.com/otwcode/otwarchive/blob/900c8e833ced9a1b027fa0e5ab5701396e19eaab/public/stylesheets/site/2.0/03-region-header.css#L241-L243

https://github.com/otwcode/otwarchive/blob/900c8e833ced9a1b027fa0e5ab5701396e19eaab/public/stylesheets/site/2.0/03-region-header.css#L137

public/stylesheets/masters/dark_mode/dark_mode_site_screen_.css
```diff
@@ -0,0 +1,750 @@
...
+#header .user a:hover,
+#header .user a:focus {
+  color: #95c6f4;
+}
+
+#header .actions a:hover,
+#header .actions a:focus,
+#header .dropdown:hover a,
+#header .open a,
+#header .menu {
```
brianjaustin, line 82: Does this also need `#small_login`? Based on https://github.com/otwcode/otwarchive/blob/900c8e833ced9a1b027fa0e5ab5701396e19eaab/public/stylesheets/site/2.0/03-region-header.css#L75

public/stylesheets/masters/dark_mode/dark_mode_site_screen_.css
```diff
@@ -0,0 +1,750 @@
...
+  background: #333;
+}
+
+span.offered.requested {
+  color: #ffe6a8;
+}
+
+/* 10-TYPES-GROUPS */
+
+a.tag:hover,
```
brianjaustin, line 465: Is https://github.com/otwcode/otwarchive/blob/900c8e833ced9a1b027fa0e5ab5701396e19eaab/public/stylesheets/site/2.0/10-types-groups.css#L7-L8 missing?

public/stylesheets/masters/dark_mode/dark_mode_site_screen_.css
```diff
@@ -0,0 +1,750 @@
...
+}
+
+.listbox li.blurb {
+  background: #333;
+}
+
+.bookmark .user {
+  border-color: #555;
+}
+
```
brianjaustin, line 555: Are the following needed too?
https://github.com/otwcode/otwarchive/blob/900c8e833ced9a1b027fa0e5ab5701396e19eaab/public/stylesheets/site/2.0/13-group-blurb.css#L321-L323

https://github.com/otwcode/otwarchive/blob/900c8e833ced9a1b027fa0e5ab5701396e19eaab/public/stylesheets/site/2.0/13-group-blurb.css#L326-L328

public/stylesheets/masters/dark_mode/dark_mode_site_screen_.css
```diff
@@ -0,0 +1,750 @@
...
+.splash .favorite li:nth-of-type(odd) a:focus {
+  background: #95c6f4;
+}
+
+/* module: news */
+
+.splash .news li {
+  border-bottom-color: #2a2a2a;
+}
+
```
brianjaustin, line 617: Is this also needed?
https://github.com/otwcode/otwarchive/blob/900c8e833ced9a1b027fa0e5ab5701396e19eaab/public/stylesheets/site/2.0/16-zone-system.css#L120

public/stylesheets/masters/dark_mode/dark_mode_site_screen_.css
sarken: The main navigation (`#header .primary a`) in Reversi also uses white, so we're good inheriting from the default! `#login` isn't included because it's only relevant for logged out users. That's also why `#small_login` and the error page styles aren't included -- they're unused in skins at this point in time.

The #999 link hover color actually wouldn't work! But Reversi doesn't have a separate hover color for links, and I've kept that here for Dark Mode. But we don't have to add `a:hover { #fff }` to explicitly make it look like non-hover links because hover styling only works if it comes after a:link and a:visited styling. The default skin CSS with `a:hover { color: #999; }` is loaded before Dark Mode, so the cascade wipes out the hover style for us.

public/stylesheets/masters/dark_mode/dark_mode_site_screen_.css
sarken: ...depends 😆  

Unlike the default sheet, Reversi styles visited tag links differently from unvisited ones. I doubt it was truly intentional as opposed to a consequence of the cascade (a.tag is defined before a:visited), but given the usefulness for visited link styling, there are going to be at least a few Reversi users who are very attached to it. 

But anyway, if I add `a.tag` instead of letting it inherit the default Dark Mode link styling, it will remove the visited link styling on tags. We could accept that, or we could restore the visited link styling by also adding `a.tag:visited` styling. 

Leaving off `a.tag` seemed like the path of least resistance here -- less code, same behavior as Reversi -- but it's not something I feel incredibly strongly about. I could go with changing the behavior or with the "more code for the same behavior" approach.

public/stylesheets/masters/dark_mode/dark_mode_site_screen_.css
sarken: Nope, those are for the text in the white icon in the upper right corner of bookmark blurbs. We want that to stay dark blue (to match the bookmark/rec icon) unless or until we also change the icon colors for Dark Mode.

public/stylesheets/masters/dark_mode/dark_mode_site_screen_.css
sarken: Good catch! It's the big section on the homepage that tells you the benefits of having an account, so I was thinking it wasn't necessary (it's not defined in Reversi). However, even though we don't talk about it a lot, admins *can* use skins in the skin chooser and access the logged out homepage, so they could theoretically see that section.

Because I wasn't thinking about admins, there are a few other things I need to change, too.

public/stylesheets/masters/dark_mode/dark_mode_site_screen_.css
brianjaustin: I didn't even realise the behaviour was different (I have been using Reversi almost exclusively for years)! I like what we have in current Reversi, so less code to achieve that is fine by me.

brianjaustin: I didn't even realise the behaviour was different (I have been using Reversi almost exclusively for years)! I like what we have in current Reversi, so less code to achieve that is fine by me.

Bilka2:
The live error message highlighting when the field is focused feels too colorful, there's 3 different colors going on in those borders I think? This is on Firefox, windows.
firefox_ZTcu8G07x0

sarken:
@Bilka2 The red/pink border to highlight the error is us, but the blue and white double outline is the focus style for Firefox. A bunch of browsers adopted bicolor outlines to make sure the focus indicator is visible regardless of the colors it's used on. You can see it on pretty much any form field, although it's definitely more intense on dark backgrounds than light, especially when combined with our live validation error styling.

I can make our border color less saturated when the live validation field has focus, which might make the overall effect a little less intense.

weeklies:
How about [styling the Firefox border](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible)?

sarken:
@weeklies That's a possibility for the future -- I really don't want to touch the default skin in this pull request, and if we're changing how we apply focus styles on form elements, it should be changed in the default as well. It's also often safer to stick with the default focus styling and simply add to it (e.g., if you change a color on hover, also change the color on focus while maintaining the default) because the accessibility criteria for the size and contrast of the focus indicator can be a little tricky to get right. Our current link focus styling definitely fails to meet the requirements, for example, but the Firefox focus indicator we find garish most likely does (even though it's over 2px wide when you take both the white and blue into account, the part that offers sufficient contrast isn't, so I'm not 100% convinced).

Bilka2:
On current Reversi, the focus border on an error field is only blue, that's why it threw me off as a Reversi user. But it looks like with the default skin it has the same tri-color as the new Dark Mode, so technically this is likely an improvement for the reasons you mention, Sarken.

public/stylesheets/masters/dark_mode/dark_mode_site_screen_.css
```diff
@@ -0,0 +1,750 @@
...
+/* 11-GROUP-LISTBOX */
+
+.listbox,
+fieldset fieldset.listbox {
+  background: #2a2a2a;
+  border-color: #555;
+    box-shadow: 1px 1px 3px #000;
+}
+
+/* this should have a direct child selector but cannot, so be careful */
```
sarken: This comment is an amazing testament to how long I have been working on this. I'll take it out and add the direct child selector. This was fixed in PR #2015 almost ten years ago.

PR #5219, opened 2025-06-15
AO3-7010 AO3-7017 Fix error when trying to access subpage of hidden unrevealed work

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-7010

## Purpose

Fix double render error when trying to access a subpage of a hidden and unrevealed work.

## Credit

Bilka


app/controllers/application_controller.rb
```diff
@@ -509,11 +509,9 @@ def check_permission_to_wrangle
 
   # Checks if user is allowed to see related page if parent item is hidden or in unrevealed collection
   def check_visibility_for(parent)
-    # Only admins and the owner can see related pages on something hidden by an admin.
-    logged_in_as_admin? || current_user_owns?(parent) || access_denied(redirect: root_path) if parent.try(:hidden_by_admin)
+    return if logged_in_as_admin? || current_user_owns?(parent) # Admins and the owner can see all related pages
 
-    # Only admins and the owner can see related pages on unrevealed works.
-    logged_in_as_admin? || current_user_owns?(parent) || access_denied(redirect: root_path) if parent.try(:in_unrevealed_collection)
+    access_denied(redirect: root_path) if parent.try(:hidden_by_admin) || parent.try(:in_unrevealed_collection)
```
brianjaustin: What if the person going to the page is an admin or the parent's owner?

Bilka2: Then the code should never reach this line due to the `return if` in line 512

brianjaustin: Ah, somehow my eyes just passed right over that line… thanks for pointing it out!

brianjaustin (approved):

PR #5222, opened 2025-06-19
AO3-6925 Remove Open Challenges link from Subcollections page

gelphcore:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6925

## Purpose

- Remove the Open Challenges link from Subcollections page. (It still appears in the main collections index.)
- Have also removed the `role="navigation"` attribute from the ul

**Before:**
image

**After:**
image

## Credit

gelphcore, they/them
Same on [Jira](https://otwarchive.atlassian.net/jira/people/712020:c09d3147-f70f-47f7-acf8-0943e01197c4)

### DIFF
--- a/app/views/collections/index.html.erb
+++ b/app/views/collections/index.html.erb
@@ -31,22 +31,13 @@
   <% unless @user || @collection %>
     <li><%= link_to ts("Open Challenges"), list_challenges_collections_path %></li>
   <% end %>
-  <% if logged_in? %>
-    <%# Logged in user on own collections index gets links for user Collections and Manage Collection Items %>
-    <% if @user && @user == current_user %>
-      <li><%= span_if_current ts("Collections"), user_collections_path(@user) %></li>
+  <% if @user %>
+    <%# User collections index gets link for user Collections %>
+    <li><%= span_if_current ts("Collections"), user_collections_path(@user) %></li>
+    <%# Logged in user on own collections index gets link for Manage Collections Item %>
+    <% if logged_in? && @user && @user == current_user %>
       <li><%= link_to ts("Manage Collection Items"), user_collection_items_path(@user) %></li>
     <% end %>
-    <%# Logged in collection maintainer on own collection gets link for New Subcollection and all other logged in users get New Collection link %>
-    <% if @collection && !@collection.parent && @collection.user_is_maintainer?(current_user) %>
-      <li><%= link_to ts("New Subcollection"), new_collection_collection_path(@collection) %></li>
-    <% else %>
-      <li><%= link_to ts("New Collection"), new_collection_path %></li>
-    <% end %>
-  <% end %>
-  <% if @sort_and_filter %>
-    <%# Filters button for narrow screens jumps to filters when JavaScript is disabled and opens filters when JavaScript is enabled %>
-    <li class="narrow-shown hidden"><a href="#collection-filters" id="go_to_filters"><%= ts("Filters") %></a></li>
   <% end %>
 </ul>
 
--- a/app/views/collections/index.html.erb
+++ b/app/views/collections/index.html.erb
@@ -23,21 +23,21 @@
 <!--Subnavigation, sorting and actions-->
 <h3 class="landmark heading"><%= ts("Navigation") %></h3>
 <ul class="navigation actions">
-  <% # Collections and link unless a logged in user viewing own collections page %>
-  <% unless logged_in? && @user && @user == current_user  %>
+# Collections and link unless a logged in user viewing own collections page %>
+  <% unless logged_in? && @user && @user == current_user %>
     <li><%= span_if_current ts("Collections"), collections_path %></li>
   <% end %>
-  <% # Open Challenges link unless a logged in user viewing own collections page or on an individual collection page %>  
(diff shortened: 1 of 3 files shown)


app/views/collections/index.html.erb
```diff
@@ -55,14 +58,14 @@
   <h3 class="landmark heading"><%= ts("List of Collections") %></h3>
   <ul class="collection picture index group">
     <% @collections.each do |collection| %>
-      <%= render :partial => "collection_blurb", :locals => {:collection => collection} %>
+      <%= render partial: "collection_blurb", locals: { collection: collection } %>
```
sarken: The syntax here is pretty old. We actually don't need `partial` or `locals` here anymore -- it can just be `<%= render "collection_blurb", collection: collection %>`

app/views/collections/index.html.erb
```diff
@@ -22,27 +22,30 @@
 
 <!--Subnavigation, sorting and actions-->
 <h3 class="landmark heading"><%= ts("Navigation") %></h3>
-<ul class="navigation actions" role="navigation">
-  <% # Collections and Open Challenges links unless a logged in user viewing own collections page %>
+<ul class="navigation actions">
+# Collections and link unless a logged in user viewing own collections page %>
```
sarken: This should say, "Collections link unless a logged in user viewing own collections page," without the "and." It also looks like the opening `<%` on all of the commented lines might have gotten eaten by a find + replace.

However, I just realized this comment is also indicative of a bug. This link points to `ao3.org/collections` and should never be used on a user's collections page. The link on a user's collections page should always use the code on line 37, regardless of whether the user is logged in.

So what we need here is:
```
<%# Collections link unless on a user's collections page %>
<% unless @user %>
  <li><%= span_if_current ts("Collections"), collections_path %></li>
<% end %>
```

And to replace lines 34-59:
```
<% if @user %>
  <%# User collections index gets link for user Collections %>
  <li><%= span_if_current ts("Collections"), user_collections_path(@user) %></li>
  <%# Logged in user on own collections index gets link for Manage Collections Item %>
  <% if logged_in? && @user && @user == current_user %>
    <li><%= link_to ts("Manage Collection Items"), user_collection_items_path(@user) %></li>
  <% end %>
<% end %>
```

However, that's a bit outside the issue you signed up for, so please let me know if you'd rather not work on that. I'm happy to create a new Jira issue for it instead.

app/views/collections/index.html.erb
```diff
@@ -22,27 +22,30 @@
...
   <% unless logged_in? && @user && @user == current_user %>
     <li><%= span_if_current ts("Collections"), collections_path %></li>
-    <li><%= link_to ts("Open Challenges"), list_challenges_collections_path %></li>
+  <% end %>
+# Open Challenges link unless a logged in user viewing own collections page or on an individual collection page %>
+  <li><%= link_to ts("Open Challenges"), list_challenges_collections_path %></li>
+  <% unless (logged_in? && @user && @user == current_user) || @collection %>
```
sarken: I think these two lines got switched. The link should be inside the conditional.

The conditional should also be `<% unless @user || @collection %>`, since we never want the Open Challenges link on a user's collections page. I think when I wrote the issue, I only checked my own collections page and foolishly assumed the link would also be omitted if I looked at another user's page.

sarken (changes requested):
Hi, gelphcore!

Thank you so much for this pull request. I've left some comments on it, plus it looks like it could use some tests to make sure the correct links are present. The tests would probably be a good fit for `features/collections/collections_browse.feature`, which already has tests like that for the main collection index.

Since you've already created a Jira account, I've gone ahead and assigned the Jira issue to you and set its status to In Review. I've also given your account permission to comment on, assign, and transition issues in the future.

app/views/collections/index.html.erb
gelphcore: Thanks for the review @sarken! Looks like the linter autocompleter was eating the `<%` so I added this to `.erb-lint.yml`:
```    
Layout/CommentIndentation:
        Enabled: false
```

After implementing your code above, the Collections link on the Subcollections page still goes to the main collections index. Should this go to the user collections path instead? Or is this what you meant for it to do?

image

sarken: Thanks for catching that! We don't want to link to the main collections index, so that link should be under the same `<% unless @user || @collection %>` conditional as the Open Challenges link. 

It wouldn't hurt to add a new "Subcollections" item:
```
<% if @collection %>
  <li><%= span_if_current ts("Subcollections"), collection_collections_path(@collection) %></li>
<% end %>
```
Much like "Collections" on the user's collections index, it will never actually appear as a link, but it will serve as a breadcrumb to remind us where we are. 

The best place for it would probably be above the `<% if @user %>` conditional that's currently on line 34.

Bilka2:
Could you please reenable `Layout/CommentIndentation`? I think it may have been misbehaving due to the extra space between `<%` and `#`, it should work fine generally.

Bilka2:
Adopted at #5957 based on the [policy for stalled pull requests](https://github.com/otwcode/otwarchive/wiki/Commit-Policy#user-content-Stalled_Pull_Requests_and_Jira_Issues)

PR #5223, opened 2025-06-19
AO3-6373 Allow comments by blocked user to be frozen, unfrozen

weeklies:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6373

## Purpose

Remove `@comment.save` and use a transaction for freezing/unfreezing comments
Also tidy up existing spec tests.

### DIFF
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -115,7 +115,7 @@ class Comment < ApplicationRecord
       user_role = "user"
       comment_author = user.login
     end
-    
+
     {
       comment_type: comment_type,
       key: ArchiveConfig.AKISMET_KEY,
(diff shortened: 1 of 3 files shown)


app/models/comment.rb
```diff
@@ -519,11 +519,25 @@ def mark_frozen!
     update_attribute(:iced, true)
   end
 
+  # Freeze all comments.
+  def mark_all_frozen!(comments)
```
brianjaustin: Since this and the following method work on the given collection, not a specific comment, how about making them static?
```suggestion
  def self.mark_all_frozen!(comments)
```

brianjaustin (approved):
Thanks!

PR #5230, opened 2025-06-22
AO3-4590 Titles in RTL languages cause display issues in emails

Hamham6:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-4590

## Purpose

Handling of RTL/LTR work or chapter title in a LTR/RTL localized email in the specific case of putting the title next to a word count that's in parenthesis using the `creation_link_with_word_count`/`creation_title_with_word_count` mailer helpers. So far this helper is only used in `recipient_notification` mailers. As part of this change tho, I changed `full_chapter_title` in the `chapter` model, and that is also used in `batch_subscription_notification` mailers.

## Testing Instructions

How can the Archive's QA team verify that this is working as you intended? wip

You can remove this section if there are already full testing instructions in the Jira issue.

## References

Considerations taken from discussion on a previous declined PR linked in the Jira issue. (https://github.com/otwcode/otwarchive/pull/4102)

## Credit

Hamham6, it/its


app/models/chapter.rb
```diff
@@ -123,13 +123,10 @@ def chapter_title
...
-  # Header plus title, used in subscriptions
+  # Header plus title, used in subscriptions and creation_link/title_with_word_count mailer helpers
   def full_chapter_title
-    str = chapter_header
-    if title.present?
-      str += ": #{title}"
-    end
-    str
+    # FIRST-STRONG ISOLATE and POP DIRECTIONAL ISOLATE respectively, so bidi titles won't affect the parenthesis around word counts, see AO3-4590
+    self.title.present? ? t(".full_chapter_title", title: "&#x2068;#{self.title}&#x2069;", position: self.position) : t(".chapter_header", position: self.position)
```
Hamham6, line 137: I haven't fully checked whether this works yet, but the idea is that anything like a work/chapter title variable in the `.yml` file that should be bidi isolated will be formatted as such in the `.t` call or earlyer, so that by the time the variable is being handled by the translation team it's already been isolated and won't affect anything around it.

Bilka2 (commented):
We had a discussion about the approach here:
* For HTML emails we'd like to go with `dir = "auto"` on the HTML elements that contain the title, and also add `unicode-bidi:plaintext` in the CSS for those elements because gmail appears to [sometimes discard the dir attribute](https://medium.com/@python-javascript-php-html-css/resolving-rtl-text-alignment-issues-in-gmail-html-emails-8e84a99eb38a) but definitely [supports the CSS](https://developers.google.com/workspace/gmail/design/css)
* For plaintext emails FSI and PDI are indeed the correct approach, but the characters should not be used anywhere where we can use markup instead (like HTML emails or the interface)

This is based on the [unicode standard annex 9 section 2.7](https://www.unicode.org/reports/tr9/tr9-46.html#Markup_And_Formatting).

So it looks like confining these changes to the mailer helpers and then consistently using only the mailer helpers for work/chapter titles in emails is the right move. Note that with the recently merged subscriptions email i18n the use of the methods you modified here has changed a bit, so it'd be good to merge master and adjust for that.

PR #5231, opened 2025-06-23
AO3-3976 Style series link in subscription email

marcus8448:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-3976

## Purpose

Changes series links in subscription emails to be red, to match the other email links.

## Credit

marcus8448 (he/him)

### DIFF
--- a/app/helpers/series_helper.rb
+++ b/app/helpers/series_helper.rb
@@ -59,13 +59,12 @@ module SeriesHelper
   end
 
   def series_list_with_work_position(work, email_styling: false)
-    series = work.series
-    return t("series_helper.none") if series.empty?
-
-    list = series.map do |s|
-      t("series_helper.series_description_html", index: s.serial_works.where(work_id: work.id).select(:position).first.position, series_link: email_styling ? style_link(s.title, series_url(s)) : link_to(s.title, series_url(s)))
-    end
-    list.join(t("support.array.words_connector"))
+    safe_join(work.series.map do |s|
+                t("series_helper.series_description_html",
+                  index: s.serial_works.where(work_id: work.id).pick(:position),
+                  series_link: email_styling ? style_link(s.title, series_url(s)) : link_to(s.title, series_url(s)))
+              end,
+              t("support.array.words_connector"))
   end
 
   # Generates confirmation message for "Remove Me As Co-Creator"
--- a/app/views/user_mailer/_work_info.html.erb
+++ b/app/views/user_mailer/_work_info.html.erb
@@ -15,7 +15,7 @@
   <% end %>
   <% series_count = work.series.count %>
   <% unless series_count.zero? %>
-    <br /><%= style_metadata_label(Series.model_name.human(count: series_count)) %><%= series_list_with_work_position(work, email_styling: true).html_safe %>
+    <br /><%= style_metadata_label(Series.model_name.human(count: series_count)) %><%= series_list_with_work_position(work, email_styling: true) %>
   <% end %>
 </p>
 
--- a/spec/mailers/user_mailer_spec.rb
+++ b/spec/mailers/user_mailer_spec.rb
@@ -717,6 +717,10 @@ describe UserMailer do
         expect(email).to have_text_part_content("Paragraph 2.")
       end
 
+      it "includes the series list" do
+        expect(email).to have_text_part_content("Series: Part 1 of #{series.title}")
+      end
+
       it "reformats HTML from the chapter summary" do
         expect(email).to have_text_part_content("*Another* HTML summary.")
       end
--- a/app/views/share/_embed_meta.html.erb
+++ b/app/views/share/_embed_meta.html.erb
@@ -11,7 +11,7 @@
 <% end %>
 <% if work.series.exists? %>
 <%= ts("Series: ") %>
-<%= series_list_with_work_position(work).html_safe %><br />
+<%= series_list_with_work_position(work) %><br />
 <% end %>
 <% if work.summary.present? %>
 <%= ts("Summary: ") %>
--- a/config/locales/helpers/en.yml
+++ b/config/locales/helpers/en.yml
@@ -37,7 +37,6 @@ en:
   inbox_helper:
     comment_link_with_chapter_number: Chapter %{position} of %{title}
   series_helper:
-    none: None
     series_description_html: Part %{index} of %{series_link}
   users_helper:
     collections_link: Collections (%{coll_number})


ceithir (commented):
Seems fine, just a matter of pleasing the dog before we're good.

Note that you can call the cop locally if easier. See https://github.com/otwcode/otwarchive/wiki/Coding-Standards#previewing-reviewdog-comments for instructions

ceithir (approved):
Failing test seems to be a [known intermittent error](https://github.com/otwcode/otwarchive/wiki/Automated-Testing#known-test-failures) though I wouldn't be against re-running them to be sure of that.

app/helpers/series_helper.rb
```diff
@@ -60,15 +60,13 @@ def work_series_description(work, series)
...
     series = work.series
-    if series.empty?
-      return "None"
-    else
-      list = []
-      for s in series
-        list << ts("Part %{serial_index} of %{link_to_series}", serial_index: s.serial_works.where(work_id: work.id).select(:position).first.position, link_to_series: style_link(s.title, series_url(s)))
-      end
-      return list.join(', ')
+    return "None" if series.empty?
```
Hamham6: `"None"` should also be i18nned, I think.

app/helpers/series_helper.rb
```diff
@@ -60,15 +60,13 @@ def work_series_description(work, series)
...
-      list = []
-      for s in series
-        list << ts("Part %{serial_index} of %{link_to_series}", serial_index: s.serial_works.where(work_id: work.id).select(:position).first.position, link_to_series: link_to(s.title, series_url(s)))
-      end
-      return list.join(', ')
+    return "None" if series.empty?
+
+    list = []
+    series.each do |s|
+      list << t("series_helper.feed_description", serial_index: s.serial_works.where(work_id: work.id).select(:position).first.position, link_to_series: style_link(s.title, series_url(s)))
```
Hamham6:
```suggestion
      list << t(".list_for_feeds_html", index: s.serial_works.where(work_id: work.id).select(:position).first.position, series_link: style_link(s.title, series_url(s)))
```
I am assuming that lazy lookup doesn't cause problems here, if it does then ignore that, my bad 🙏 (You can never really be sure with helpers lol)

Otherwize just nitpicks to names: the variable name for the link ending in `_link` and the key name ending in `_html` because it includes html (for the link). (Taken from here: https://github.com/otwcode/otwarchive/wiki/Internationalization-%28i18n%29-Standards#hyperlinks)

Changing `serial_index` to `index` is just because I personally think it's easyer to understand, feel free to leave the name as-is if you think the change takes out necessary context. `.feed_description` to `.list_for_feeds` is because I found it confusing to call it "description", but that's also entirely just my opinion and not a necessary change at all. 

Also `en.yml` needs to be updated with any changes accordingly, I am just too lazy to go grab it as suggestions xP

app/helpers/series_helper.rb
```diff
@@ -60,15 +60,13 @@ def work_series_description(work, series)
...
     end
+    list.join(", ")
```
Hamham6:
```suggestion
    list.join(t("support.array.words_connector"))
```

Hamham6 (changes requested):

app/helpers/series_helper.rb
marcus8448: `series_list_for_feeds` is called from multiple views (gift emails use it too) so I think it's best to keep it non-lazy.

I'm not sure about the name `list_for_feeds` as the translation itself is not a list - instead it gets concatenated into one (I hope that makes sense).

marcus8448: Okay, it turns out `series_list_for_feeds` was used for more than just email notifications (it's used for the work sharing feature and RSS feeds). I've reverted back to the name `series_description`, but I'm open to other naming ideas.

Hamham6: `series_description` works, I think. I certainly can't think of anything better 😅

app/helpers/series_helper.rb
```diff
@@ -58,17 +58,15 @@ def work_series_description(work, series)
     ts("Part <strong>%{position}</strong> of %{title}".html_safe, position: serial.position, title: link_to(series.title, series))
   end
 
-  def series_list_for_feeds(work)
+  def series_list(work, red_link: false)
```
Hamham6:
```suggestion
  def series_list(work, email_styling: false)
```
(Because I hold a hope in my heart that one day emails can also have dark mode/reversi styling lol)

app/helpers/series_helper.rb
```diff
@@ -58,17 +58,15 @@ def work_series_description(work, series)
     ts("Part <strong>%{position}</strong> of %{title}".html_safe, position: serial.position, title: link_to(series.title, series))
   end
 
-  def series_list_for_feeds(work)
+  def series_list(work, email_styling: false)
```
brianjaustin: Nitpick (non blocking): I'm not sure `list` best describes what this does anymore (or maybe ever). I don't have something much better, but maybe the following could work?
```suggestion
  def series_position_description(work, email_styling: false)
```

app/views/user_mailer/_work_info.html.erb
```diff
@@ -14,7 +14,7 @@
     <br /><%= style_work_tag_metadata(work.tag_groups["Freeform"]) %>
   <% end %>
   <% unless work.series.count.zero? %>
-    <br /><%= style_metadata_label(Series.model_name.human(count: work.series.count)) %><%= series_list_for_feeds(work).html_safe %>
+    <br /><%= style_metadata_label(Series.model_name.human(count: work.series.count)) %><%= series_list(work, red_link: true).html_safe %>
```
brianjaustin: This, as well as the txt version and both HTML/txt for app/views/user_mailer/batch_subscription_notification need to be updated with the new keyword argument name

Hamham6 (approved):
LGTM! Also, agreed, naming things is the worst.

app/helpers/series_helper.rb
sarken: I think `_list` is actually a necessary part of the name, since this method produces a comma-separated list of all the series a work is in. Without an English plural form of series, the new name risks sounding like it's just the information for a single series. Something like `list_of_series_with_position` or `series_list_with_work_position`?

brianjaustin: I see your point; I think there's potential for confusion either way, because something with "list" in it sounds like it would return a list, not a single string. But I think the version with "list" in the middle is least likely to result in that confusion, so I would go with that

Bilka2:
Localises something inside a cache that is not per locale and there is already a translation available (the word connector from active support), so can't be merged until this caching problem is solved (which is part of separate Jira issues).

Bilka2:
No longer deferred, yay! But it now has a whole pile of merge conflicts. And I have a suggestion, could you use `map` instead of `each` in the method?

app/helpers/series_helper.rb
```diff
@@ -58,13 +58,14 @@ def work_series_description(work, series)
     ts("Part <strong>%{position}</strong> of %{title}".html_safe, position: serial.position, title: link_to(series.title, series))
   end
 
-  def series_list_for_feeds(work)
+  def series_list_with_work_position(work, email_styling: false)
     series = work.series
-    list = []
-    series.each do |s|
-      list << ts("Part %{serial_index} of %{link_to_series}", serial_index: s.serial_works.where(work_id: work.id).select(:position).first.position, link_to_series: link_to(s.title, series_url(s)))
+    return t("series_helper.none") if series.empty?
```
Bilka2: Could you remove this as well, to match the optimisation done in the subscriptions i18n PR? All call sites already check that the series isn't empty

app/helpers/series_helper.rb
```diff
@@ -58,13 +58,14 @@ def work_series_description(work, series)
...
+
+    list = series.map do |s|
+      t("series_helper.series_description_html", index: s.serial_works.where(work_id: work.id).select(:position).first.position, series_link: email_styling ? style_link(s.title, series_url(s)) : link_to(s.title, series_url(s)))
     end
-    list.join(", ")
+    list.join(t("support.array.words_connector"))
```
Bilka2: Does this work with just a `join`? I would have expected it to need a [safe_join](https://api.rubyonrails.org/classes/ActionView/Helpers/OutputSafetyHelper.html#method-i-safe_join) instead to preserve the HTML safe status. Maybe we could get rid of the html_safe calls at the call sites if we switch to `safe_join`?

app/views/user_mailer/_work_info.text.erb
```diff
@@ -14,7 +14,7 @@
 <% end %>
 <% series_count = work.series.count %>
 <% unless series_count.zero? %>
-<%= metadata_label(Series.model_name.human(count: series_count)) %><%= raw to_plain_text(series_list_for_feeds(work)) %>
+<%= metadata_label(Series.model_name.human(count: series_count)) %><%= raw to_plain_text(series_list_with_work_position(work, email_styling: true)) %>
```
Bilka2: Does this need the styling here? Since it's a text email. But maybe it should match anyway?

spec/mailers/user_mailer_spec.rb
Bilka2, line 1: While this change isn't about the text email, I think it might be good to check that as well, to make sure we don't end up with something garbled after the style is fed through `to_plain_text`

Bilka2 (changes requested):

PR #5236, opened 2025-06-28
AO3-7019 Disallow changing username to current username

kitbur:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-7019

## Purpose

Displays an error when a user attempts to change their username to their current username, but allows users to change the case of characters in their username.

## Credit

kitbur

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -81,11 +81,14 @@ class UsersController < ApplicationController
       render(:change_username) && return
     end
 
+    if @new_login == @user.login
+      flash.now[:error] = t(".user.not_different") 
+      render :change_username and return
+    end
+
     old_login = @user.login
     @user.login = @new_login
     @user.ticket_number = params[:ticket_number]
-
-    @user.username_change_attempt = true
 
     if @user.save
       if logged_in_as_admin?
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -82,7 +82,7 @@ class UsersController < ApplicationController
     end
 
     if @new_login == @user.login
-      flash.now[:error] = t(".user.not_different") 
+      flash.now[:error] = t(".new_username_must_be_different") 
       render :change_username and return
     end
 
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -85,6 +85,8 @@ class UsersController < ApplicationController
     @user.login = @new_login
     @user.ticket_number = params[:ticket_number]
 
+    @user.username_change_attempt = true
+
     if @user.save
       if logged_in_as_admin?
         flash[:notice] = t(".admin.successfully_updated")
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -200,7 +200,6 @@ class User < ApplicationRecord
             not_forbidden_name: { if: :will_save_change_to_login? }
   validate :username_is_not_recently_changed, if: :will_save_change_to_login?
   validate :admin_username_generic, if: :will_save_change_to_login?
-  validate :username_must_be_different_from_current, on: :update, if: :username_change_attempt
 
   # allow nil so can save existing users
   validates_length_of :password,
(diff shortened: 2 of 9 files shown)


app/models/user.rb
```diff
@@ -200,6 +200,7 @@ def unread_inbox_comments_count
             not_forbidden_name: { if: :will_save_change_to_login? }
   validate :username_is_not_recently_changed, if: :will_save_change_to_login?
   validate :admin_username_generic, if: :will_save_change_to_login?
+  validate :username_must_be_different_from_current, on: :update, if: :username_change_attempt
```
brianjaustin: Could we use the built-in `will_save_change_to_login?` here, similar to the `username_is_not_recently_changed` validation?

app/models/user.rb
```diff
@@ -637,6 +638,12 @@ def admin_username_generic
     errors.add(:login, :admin_must_use_default) unless login == "user#{id}"
   end
 
+  def username_must_be_different_from_current
+    return if User.current_user != self
+
+    errors.add(:login, :username_not_different) if login == login_was
```
brianjaustin: Nitpick: this can just be
```suggestion
    errors.add(:login, :not_different) if login == login_was
```
since the error is already scoped to `:login`

kitbur: 0fdc97748

app/models/user.rb
kitbur: Thank you so much for the review!

I initially tried an approach with `will_save_change_to_login?` to keep things consistent, but discovered two issues during testing:

- It allows the original bug to happen (the validation doesn't fire if an identical username is submitted)
- Simpler validations without that condition incorrectly blocked users from changing their email or password.

The `:username_change_attempt` flag set by the controller was the solution I found to solve both of these issues by providing the necessary context to the model.

I'm definitely open to any other ideas!

brianjaustin: :facepalm: yep I should have realised that by the name... After taking another look and rubber ducking with @Bilka2, the solution here is to lift the whole validation into the controller

kitbur: OH!! 😭 That makes sense. Thank you, Brian and Bilka!

I moved the validation over to the controller and took out the old model test and code!

app/controllers/users_controller.rb
```diff
@@ -81,6 +81,11 @@ def changed_username
       render(:change_username) && return
     end
 
+    if @new_login == @user.login
+      flash.now[:error] = t(".user.not_different")
```
brianjaustin: Sorry for the extra nitpick on top of the pain of moving the validation, but I think we'll want a different key now that it's here
```suggestion
      flash.now[:error] = t(".new_username_must_be_different") 
```
or possibly
```suggestion
      flash.now[:error] = t(".username.must_be_different") 
```
(It's a bit easier for the translators if the key name is close to the English text when it makes sense/is possible for us to do.)

config/locales/views/en.yml
```diff
@@ -2182,6 +2182,9 @@ en:
       submit: Change Username
       submit_landmark: Submit
       username_requirements: "%{minimum} to %{maximum} characters (A-Z, a-z, _, 0-9 only), no spaces, cannot begin or end with underscore (_)"
+    changed_username:
+      user:
+        not_different: Your new username must be different from your current username.
```
brianjaustin: In addition to the changes required for my other comment, this should go under https://github.com/otwcode/otwarchive/blob/ba22585f6d177a2ec2f9bc16e07a2e6f242807dd/config/locales/controllers/en.yml#L217

since it's in a controller, not a view.

brianjaustin (changes requested):

app/controllers/users_controller.rb
kitbur: Noted; thank you! I appreciate the guidance, since I am not used to working with internationalization. ❤️

01253726b3612bef21ffa0e4219dca5d6b32af77

config/locales/views/en.yml
kitbur: Thank you! 01253726b3612bef21ffa0e4219dca5d6b32af77

I ran the failed `features/search/filters.feature` test locally and it passed, so it seems unrelated to this validation change.

brianjaustin (approved):

PR #5238, opened 2025-06-28
AO3-5349 i18n prompt response notification

weeklies:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5349 

## Purpose

i18n prompt response notification and implement three types of mailer previews:
- prompter_notification with the optional URL param: creator_count
- prompter_notification_collection with the optional URL param: creator_count
- prompter_notification_collection_anon 

Also, correct a typo in a step definition filename


lib/creation_notifier.rb
```diff
@@ -84,11 +84,19 @@ def notify_subscribers_on_reveal
...
   # notify prompters of response to their prompt
   def notify_prompters
-    if !self.challenge_claims.empty? && !self.unrevealed?
-      if self.collections.first.nil?
-        UserMailer.prompter_notification(self.id,).deliver_after_commit
-      else
-        UserMailer.prompter_notification(self.id, self.collections.first.id).deliver_after_commit
+    return if self.challenge_claims.empty? || self.unrevealed?
+
+    user_ids = self.challenge_claims.map { |claim| claim.request_signup.pseud.user.id }.uniq
```
weeklies:
```suggestion
    user_ids = self.challenge_claims.map { |claim| claim.request_signup.pseud.user.id }
                                    .uniq
```

lib/creation_notifier.rb
```diff
@@ -84,11 +84,20 @@ def notify_subscribers_on_reveal
...
   def notify_prompters
-    if !self.challenge_claims.empty? && !self.unrevealed?
-      if self.collections.first.nil?
-        UserMailer.prompter_notification(self.id,).deliver_after_commit
-      else
-        UserMailer.prompter_notification(self.id, self.collections.first.id).deliver_after_commit
+    return if self.challenge_claims.empty? || self.unrevealed?
+
+    user_ids = self.challenge_claims.map { |claim| claim.request_signup.pseud.user.id }
+                                    .uniq
```
weeklies:
```suggestion

    user_ids = self.challenge_claims.map { |claim| claim.request_signup.pseud.user.id } 
      .uniq
```

app/mailers/user_mailer.rb
```diff
@@ -309,18 +309,15 @@ def recipient_notification(user_id, work_id, collection_id = nil)
...
-        mail(
-          to: user.email,
-          subject: "[#{ArchiveConfig.APP_SHORT_NAME}] A response to your prompt"
-        )
-      end
-    end
+    mail(
+      to: @user.email,
+      subject: t("user_mailer.prompter_notification.subject",
+                 app_name: ArchiveConfig.APP_SHORT_NAME)
```
slavalamp:
```suggestion
      subject: default_i18n_subject(app_name: ArchiveConfig.APP_SHORT_NAME)
```
[Translating subjects](https://github.com/otwcode/otwarchive/wiki/Internationalization-%28i18n%29-Standards#translating-subjects)

lib/creation_notifier.rb
```diff
@@ -84,11 +84,21 @@ def notify_subscribers_on_reveal
...
   # notify prompters of response to their prompt
   def notify_prompters
-    if !self.challenge_claims.empty? && !self.unrevealed?
-      if self.collections.first.nil?
-        UserMailer.prompter_notification(self.id,).deliver_after_commit
-      else
-        UserMailer.prompter_notification(self.id, self.collections.first.id).deliver_after_commit
+    return if self.challenge_claims.empty? || self.unrevealed?
+
+
```
weeklies:
```suggestion

```

lib/creation_notifier.rb
```diff
@@ -84,11 +84,19 @@ def notify_subscribers_on_reveal
 
   # notify prompters of response to their prompt
   def notify_prompters
-    if !self.challenge_claims.empty? && !self.unrevealed?
-      if self.collections.first.nil?
-        UserMailer.prompter_notification(self.id,).deliver_after_commit
-      else
-        UserMailer.prompter_notification(self.id, self.collections.first.id).deliver_after_commit
+    return if self.challenge_claims.empty? || self.unrevealed?
```
weeklies, line 87:
```suggestion
    return if self.challenge_claims.empty? || self.unrevealed?

```

app/views/user_mailer/prompter_notification.html.erb
```diff
@@ -1,29 +1,30 @@
...
+      <%= style_bold(t(".label.fandom")) %> <%= @work.fandom_string %>
     <% end %>
   </p>
-  
-  <% unless @work.summary.blank? %>
-    <%= style_bold("Summary:") %>
+
+  <% if @work.summary.present? %>
+    <%= style_bold(t(".label.summary")) %>
     <%= style_quote(raw sanitize_field(@work, :summary)) %>
```
sarken, line 24: I don't know if we'd want to change it in this pull request, or make an issue to change it in all the emails at once, but we should modify this to use the work_info partial, which contains more information and lets us reuse translations/code.

sarken: Oh, sorry, forgot to finish my thought: even if we don't do that, we should be using `style_work_tag_metadata`/`style_metadata_label` for these sorts of labels.

config/locales/mailers/en.yml
```diff
@@ -492,6 +492,19 @@ en:
         other: We regret to inform you that your request for %{count} new invitations cannot be fulfilled at this time.
       reason: 'Your request was:'
       subject: "[%{app_name}] Additional invitation request declined"
+    prompter_notification:
+      a_response_to_your_prompt:
+        collection:
+          html: A response to your prompt has been posted in the %{collection_link} collection at the Archive of Our Own!
+          text: A response to your prompt has been posted in the "%{collection_title}" collection (%{collection_url}) at the Archive of Our Own!
```
brianjaustin: Perhaps it's out of scope for now, but the the preposition for referencing AO3 doesn't sound right
```suggestion
          html: A response to your prompt has been posted in the %{collection_link} collection on the Archive of Our Own!
          text: A response to your prompt has been posted in the "%{collection_title}" collection (%{collection_url}) on the Archive of Our Own!
```

app/views/user_mailer/prompter_notification.html.erb
weeklies: I have made the incremental change. A new issue to change all emails sounds like a better idea.

config/locales/mailers/en.yml
weeklies: Not sure about editing it, so I left it for now

sarken: I think we should change it now. It's a lot more work for everyone to make an issue for this, which may also require Translation to edit any translations they've done in the meantime.

app/views/user_mailer/prompter_notification.text.erb
```diff
@@ -1,14 +1,20 @@
...
 
-"<%= @work.title %>" (<%= @work.word_count %> words)
+<%= creation_title_with_word_count(@work) %>
 <%= @collection ? collection_work_url(@collection, @work) : work_url(@work) %>
 
-by <%= @work.anonymous? ? "an anonymous responder" : (@work.pseuds.map{|p| text_pseud(p)}.to_sentence) %>
+<%= t(".byline.text", work_creators: creator_text(@work)) %>
 
-<% unless @work.fandom_string.blank? %>Fandom: <%= @work.fandom_string %><% end %>
+<% if @work.fandom_string.present? %><%= work_tag_metadata(work.tag_groups["Fandom"]) %><% end %>
```
weeklies:
```suggestion
<% if @work.fandom_string.present? %><%= work_tag_metadata(@work.tag_groups["Fandom"]) %><% end %>
```

app/views/user_mailer/prompter_notification.html.erb
sarken: Just following up to say it looks like we already have one for this email! [AO3-5764](https://otwarchive.atlassian.net/browse/AO3-5764)

app/models/collection_item.rb
```diff
@@ -232,7 +232,13 @@ def notify_of_reveal
 
       # also notify prompters of responses to their prompt
       if item_type == "Work" && !item.challenge_claims.blank?
-        UserMailer.prompter_notification(self.item.id, self.collection.id).deliver_after_commit
+        self.item.challenge_claims.each do |claim|
+          user = User.find(claim.request_signup.pseud.user.id)
+
+          I18n.with_locale(user.preference.locale_for_mails) do
+            UserMailer.prompter_notification(user.id, self.item.id, self.collection.id).deliver_after_commit
```
Bilka2: You can pass the `claim.request_signup.pseud.user.id` into the mailer/preference, no need to refind the user

lib/creation_notifier.rb
```diff
@@ -84,11 +84,20 @@ def notify_subscribers_on_reveal
...
+
+    user_ids = self.challenge_claims.map { |claim| claim.request_signup.pseud.user.id } 
+      .uniq
+
+    user_ids.each do |user_id|
+      user = User.find(user_id)
```
Bilka2: You can handle the users directly, no need to refind via the ids

lib/creation_notifier.rb
```diff
@@ -84,11 +84,20 @@ def notify_subscribers_on_reveal
...
+
+      I18n.with_locale(user.preference.locale_for_mails) do
+        if self.collections.first.nil?
+          UserMailer.prompter_notification(user_id, self.id).deliver_after_commit
+        else
+          UserMailer.prompter_notification(user_id, self.id, self.collections.first.id).deliver_after_commit
+        end
```
Bilka2: Could this use safe navigation for the collection id to remove the if here?

config/locales/mailers/en.yml
```diff
@@ -492,6 +492,16 @@ en:
         other: We regret to inform you that your request for %{count} new invitations cannot be fulfilled at this time.
       reason: 'Your request was:'
       subject: "[%{app_name}] Additional invitation request declined"
+    prompter_notification:
+      a_response_to_your_prompt:
+        collection:
+          html: A response to your prompt has been posted in the %{collection_link} collection on the Archive of Our Own!
+          text: A response to your prompt has been posted in the "%{collection_title}" collection (%{collection_url}) on the Archive of Our Own!
+        text: A response to your prompt has been posted at the Archive of Our Own!
```
Bilka2: Could you name this key something like `no_collection` instead? Using `text` will make the translators wonder if there is a html version of the locale

lib/creation_notifier.rb
```diff
@@ -84,11 +84,20 @@ def notify_subscribers_on_reveal
...
   def notify_prompters
-    if !self.challenge_claims.empty? && !self.unrevealed?
-      if self.collections.first.nil?
-        UserMailer.prompter_notification(self.id,).deliver_after_commit
-      else
-        UserMailer.prompter_notification(self.id, self.collections.first.id).deliver_after_commit
+    return if self.challenge_claims.empty? || self.unrevealed?
+
+    user_ids = self.challenge_claims.map { |claim| claim.request_signup.pseud.user.id } 
+      .uniq
```
Bilka2, line 90: Can this intermediate array and the uniq be removed, to have an each loop similar to the other spot?

config/locales/mailers/en.yml
Bilka2:
```suggestion
        text: A response to your prompt has been posted on the Archive of Our Own!
```

app/views/user_mailer/prompter_notification.html.erb
```diff
@@ -1,29 +1,30 @@
...
+    <% url = @collection ? collection_work_url(@collection, @work) : work_url(@work) %>
+    <%= creation_link_with_word_count(@work, url) %>
     <br>
-    by <%= @work.anonymous? ? style_bold("an anonymous responder") : (@work.pseuds.map{|p| style_pseud_link(p)}.to_sentence.html_safe) %>
+    <%= t(".byline.html", work_creator_links: creator_links(@work)) %>
     <br>
-    <% unless @work.fandom_string.blank? %>
-      <%= style_bold("Fandom:") %> <%= @work.fandom_string %>
+    <% if @work.fandom_string.present? %>
+      <%= style_work_tag_metadata(@work.tag_groups["Fandom"]) %>
```
Bilka2:
```suggestion
      <%= style_work_tag_metadata(@work.fandoms) %>
```
tag_groups is better when we're using multiple tag types (because they all get retrieved with a single query). But this only shows one tag type, so we load less from the db if we only get the fandoms.

(same for text version)

features/prompt_memes_b/challenge_promptmeme_posting_fills.feature
```diff
@@ -46,6 +46,39 @@ Feature: Prompt Meme Challenge
...
+    And I follow "Fulfill"
+    And I check "No Archive Warnings Apply"
+    And I select "English" from "Choose a language"
+    And I fill in "Fandoms" with "GhostSoup"
+    And I fill in "Work Title" with "Kinky Story"
+    And I fill in "content" with "Story written for your kinks, oh mystery reader!"
+  When all emails have been delivered
+    And I press "Post"
+  Then 1 email should be delivered to "my1@e.org"
+    And the email to "myname1" should be translated
```
Bilka2, line 81: Can you check that this is the right email, so subject/or some part of the content?

Bilka2 (changes requested):
Please also merge the most recent master branch to resolve the merge conflicts, thank you!

weeklies:
I suspect the test failure relates to the removal of the `uniq` to users. Should I update the test or am I missing something?

lib/creation_notifier.rb
```diff
@@ -84,11 +84,13 @@ def notify_subscribers_on_reveal
...
-        UserMailer.prompter_notification(self.id,).deliver_after_commit
-      else
-        UserMailer.prompter_notification(self.id, self.collections.first.id).deliver_after_commit
+    return if self.challenge_claims.empty? || self.unrevealed?
+
+    self.challenge_claims.each do |claim|
+      user = claim.request_signup.pseud.user
+
+      I18n.with_locale(user.preference.locale_for_mails) do
+        UserMailer.prompter_notification(user.id, self.id, self.collections&.first&.id).deliver_after_commit
```
Bilka2:
```suggestion
        UserMailer.prompter_notification(user.id, self.id, self.collections.first&.id).deliver_after_commit
```

Bilka2:
I am quite confused by this failure, my suggestions were aimed at matching the sending code as it was before. So, you loop through `work.challenge_claims` and send to every `claim.request_signup.pseud.user`, same as what used to happen in `UserMailer#prompter_notification`.

It looks like the mailer simply doesnt queue two emails if `mail` is called twice with the same arguments, so moving the loop out of the mailer is what makes the difference.

I suppose that means we need the uniq again to keep existing behaviour the same, yes (in both spots where the mailer is called, not just creation notifier)

app/models/collection_item.rb
```diff
@@ -232,7 +232,13 @@ def notify_of_reveal
 
       # also notify prompters of responses to their prompt
       if item_type == "Work" && !item.challenge_claims.blank?
-        UserMailer.prompter_notification(self.item.id, self.collection.id).deliver_after_commit
+        users = self.challenge_claims.map { |claim| claim.request_signup.pseud.user }
```
weeklies:
```suggestion
        users = item.challenge_claims.map { |claim| claim.request_signup.pseud.user }
```

PR #5225, opened 2025-06-20
AO3-6746 Expire chapter byline cache when pseud is renamed

weeklies:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6746

## Purpose

Make sure that chapter bylines update when a co-creator changes their username or pseud

PR #5242, opened 2025-07-01
AO3-7020 i18n creation bylines

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-7020

## Purpose

Prepare creation bylines for translation. This includes making the cache locale independent by not including the translations in the cache and instead caching the raw data.

The first commit in this PR shows a intermediate solution which had two layers of caching, one of which included the locale in the cache key. However, I found a way that avoided the extra db query for the user outside of `byline_data`, so there are no more db queries happening outside of the `byline_data`, as shown by the N+1 specs.

This does not make the format for `pseud#byline` translatable because it's used in untranslatable parts of the interface, such as the co-creator field on the work form where you can enter `pseud (user)`. If users got used to other formats (e.g. using different parens), these fields would become confusing.

This fixes that the text byline (used in RSS feeds and the share function) was not cached.

This PR has a conflict with #5225.

## Testing Instructions

TBD. I can't come up with anything good to test manually, since we don't have translations for this.

## Credit

Bilka

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -91,11 +91,9 @@ module ApplicationHelper
   end
 
   def byline_text_cached(creation, only_path)
-    # Update Series#expire_byline_cache when changing cache keys here
-    Rails.cache.fetch("byline_text/#{creation.cache_key}_#{only_path}_#{I18n.locale}") do
-      creators = Rails.cache.fetch("byline_internal/#{creation.cache_key}") { byline_data(creation) }
-      byline_text_internal(creators, only_path)
-    end
+    # Update Series#expire_byline_cache when changing cache key here
+    creators = Rails.cache.fetch(["byline_internal", creation.cache_key]) { byline_data(creation) }
+    byline_text_internal(creators, only_path)
   end
 
   def byline_text_uncached(creation, only_path, text_only = false)
@@ -107,7 +105,7 @@ module ApplicationHelper
     return creators if creators.is_a?(String)
 
     safe_join(creators.map do |creator|
-      pseud_byline = text_only ? creator[:pseud].byline : pseud_link(creator[:pseud], only_path)
+      pseud_byline = text_only ? creator[:pseud].byline(creator[:user].login) : pseud_link(creator[:pseud], creator[:user], only_path)
       if creator[:archivists].empty?
         pseud_byline
       else
@@ -133,14 +131,14 @@ module ApplicationHelper
       end
     end
 
-    pseuds.map { |pseud| { pseud: pseud, archivists: archivists[pseud] } }
+    pseuds.map { |pseud| { pseud: pseud, user: pseud.user, archivists: archivists[pseud] } }
   end
 
-  def pseud_link(pseud, only_path = true)
+  def pseud_link(pseud, user, only_path = true)
     if only_path
-      link_to(pseud.byline, user_pseud_path(pseud.user, pseud), rel: "author")
+      link_to(pseud.byline(user.login), user_pseud_path(user, pseud), rel: "author")
     else
-      link_to(pseud.byline, user_pseud_url(pseud.user, pseud), rel: "author")
+      link_to(pseud.byline(user.login), user_pseud_url(user, pseud), rel: "author")
     end
   end
 
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -70,9 +70,9 @@ module ApplicationHelper
   # Byline helpers
   def byline(creation, options={})
     if creation.respond_to?(:anonymous?) && creation.anonymous?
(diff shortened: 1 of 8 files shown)


app/decorators/pseud_decorator.rb
```diff
@@ -115,6 +115,7 @@ def authored_items_links(options = {})
     end
   end
 
+  # Do NOT internationalize this
```
redsummernight: You should mention here that we don't want to i18n this because it's also an input format. Same goes for the comments in pseud.

app/helpers/application_helper.rb
```diff
@@ -70,77 +70,96 @@ def allowed_html_instructions(show_text = true)
...
-  end
 
-  def pseud_link(pseud, only_path = true)
-    if only_path
-      link_to(pseud.byline, user_pseud_path(pseud.user, pseud), rel: "author")
-    else
-      link_to(pseud.byline, user_pseud_url(pseud.user, pseud), rel: "author")
+    pseuds.map do |pseud|
+      {
+        byline: pseud.byline,
```
redsummernight, line 143: Let's add some comments here on each piece of data returned: `byline` is for the plain-text "pseud (username)", `pseud` and `user` for when we need to linkify the byline, and `archivists` is the list of unclaimed author names (meaning the pseud is an archivist), IIUC.

app/helpers/application_helper.rb
```diff
@@ -70,77 +70,96 @@ def allowed_html_instructions(show_text = true)
...
+        pseud: pseud.to_param,
```
redsummernight, line 146: `to_param` should be called by `link_to`/`url_for`, but do you call it here so they're more cacheable strings? Worth a comment as well.

spec/helpers/application_helper_spec.rb
```diff
@@ -379,4 +379,25 @@
...
+    let(:user1) { create(:user, login: "Beetle") }
+    let(:user2) { create(:user, login: "Muppet") }
+    let(:work) { create(:work, authors: [user1.default_pseud, user2.default_pseud]) }
+
+    before do
+      create(:locale, iso: "new")
+      I18n.backend.store_translations(:new, { support: { array: { words_connector: "，" } } })
+    end
+
+    it "is locale independent" do
```
redsummernight: It's not super clear to me how, `helper.byline` still depends on `words_connector`.

Bilka2: Perhaps a bad way to word it? The cache is locale independent, the resulting text is not and changes based on locale despite the cache staying the same. I've attempted a reword

app/helpers/application_helper.rb
```diff
@@ -70,77 +70,100 @@ def allowed_html_instructions(show_text = true)
...
+    pseuds.map do |pseud|
+      {
+        # Cache the plain-text pseud (username)
+        byline: pseud.byline,
+        # Cache the parameters that we need for generating the pseud URL later
+        # We can't cache the record itself (for later URL generation) since it could change or be deleted
+        pseud: pseud.to_param,
+        user: pseud.user.to_param,
+        # Cache the array of plain-text names of the unclaimed external creators
+        archivists: archivists[pseud]
```
redsummernight: This should be renamed to `external_creators`.

app/helpers/application_helper.rb
```diff
@@ -70,77 +70,100 @@ def allowed_html_instructions(show_text = true)
...
+                     else
+                       url_options = { controller: "pseuds", action: "show", user_id: creator[:user], id: creator[:pseud], only_path: only_path }
+                       link_to(creator[:byline], url_options, rel: "author")
+                     end
+
+      if creator[:archivists].empty?
+        pseud_byline
+      else
+        safe_join(creator[:archivists].map do |ext_author|
+          t("application_helper.archivist_byline_html", external_author: ext_author, pseud_byline: pseud_byline)
```
redsummernight: We can use `external_creator` for both the i18n argument and the local variable too. We're moving away from "author" terms (e.g. [AO3-5776](https://otwarchive.atlassian.net/browse/AO3-5776)) so let's be consistent where we can.

spec/helpers/application_helper_spec.rb
redsummernight: Yeah, the problem is the test doesn't make any assertions about what's cached and not cached. So the test description should just be "results in different bylines per locale" (removing "despite caching") to avoid confusion. You can make this behavior even clearer by having the `new` locale use a connector other than the comma.

If we had to test caching behavior explicitly (and I don't think we need to since the n+1 spec should cover it), we should have assertions against database reads, like:

https://github.com/otwcode/otwarchive/blob/253c3edfc9a428798a5bbecf95ee9741399df414/spec/models/admin_setting_spec.rb#L9-L14

redsummernight (approved):

PR #5246, opened 2025-07-03
AO3-7024 - Fixed inbox success message being incorrectly displayed

talvalin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-7024

## Purpose

Fixes a display issue where the success message "Inbox successfully updated.” is displayed when Mark Read/Mark Unread/Delete is used without any inbox items being selected.

## Credit

talvalin
he/him

### DIFF
--- a/app/controllers/inbox_controller.rb
+++ b/app/controllers/inbox_controller.rb
@@ -42,10 +42,10 @@ class InboxController < ApplicationController
       elsif params[:delete]
         @inbox_comments.each { |i| i.destroy }
       end
+      success_message = ts('Inbox successfully updated.')
     rescue
-      flash[:caution] = ts("Please select something first")
+      flash[:caution] = ts("Please select something first.")
     end
-    success_message = ts('Inbox successfully updated.')
     respond_to do |format|
       format.html { redirect_to request.referer || user_inbox_path(@user, page: params[:page], filters: params[:filters]), notice: success_message }
       format.json { render json: { item_success_message: success_message }, status: :ok }
--- a/spec/controllers/inbox_controller_spec.rb
+++ b/spec/controllers/inbox_controller_spec.rb
@@ -271,19 +271,15 @@ describe InboxController do
       before { fake_login_known_user(user) }
 
       context "with no comments selected" do
-        it "redirects to inbox with caution and a notice" do
+        it "redirects to the user's inbox with a caution" do
           put :update, params: { user_id: user.login, read: "yeah" }
-          it_redirects_to_with_caution_and_notice(user_inbox_path(user),
-                                                  "Please select something first",
-                                                  "Inbox successfully updated.")
+          it_redirects_to_with_caution(user_inbox_path(user), "Please select something first.")
         end
 
-        it "redirects to the previously viewed page if HTTP_REFERER is set, with a caution and a notice" do
+        it "redirects to the previously viewed page if HTTP_REFERER is set, with a caution" do
           @request.env["HTTP_REFERER"] = root_path
           put :update, params: { user_id: user.login, read: "yeah" }
-          it_redirects_to_with_caution_and_notice(root_path,
-                                                  "Please select something first",
-                                                  "Inbox successfully updated.")
+          it_redirects_to_with_caution(root_path, "Please select something first.")
         end
       end
 


talvalin:
I would fix the warnings that Rubocop is complaining about, but I've nuked my branch already... 🫠 
Are those warnings a blocker to the PR being approved?

In future I will see if I can run Rubocop locally to check for these things so to avoid CI/CD failures like this one.

ceithir:
> I would fix the warnings that Rubocop is complaining about, but I've nuked my branch already... 🫠

Since your code is still on Github, you should be able to retrieve it easily even from a fully empty directory?

```sh
git clone git@github.com:talvalin/otwarchive.git
cd otwarchive/
git fetch origin
git checkout -b AO3-7024 origin/AO3-7024
```
> In future I will see if I can run Rubocop locally to check for these things so to avoid CI/CD failures like this one.

Relevant instructions available here: https://github.com/otwcode/otwarchive/wiki/Coding-Standards#previewing-reviewdog-comments

talvalin:
@ceithir Thank you! It's been an age since I last contributed to AO3 and my mediocre git skills have only atrophied in the intervening years.

I made another commit to address Reviewdog/Rubocop's comments, so that check seems to have passed this time around so a cautious woo (or should that be "Whee!"?).

spec/controllers/inbox_controller_spec.rb
```diff
@@ -271,19 +271,15 @@
       before { fake_login_known_user(user) }
 
       context "with no comments selected" do
-        it "redirects to inbox with caution and a notice" do
+        it "redirects to the user's inbox with a caution" do
           put :update, params: { user_id: user.login, read: "yeah" }
-          it_redirects_to_with_caution_and_notice(user_inbox_path(user),
-                                                  "Please select something first",
-                                                  "Inbox successfully updated.")
+          it_redirects_to_with_caution(user_inbox_path(user), "Please select something first.")
```
ceithir: Oh, nice. Didn't know `it_redirects_to_with_caution` already included  `expect(flash[:notice]).to be_blank` as a default.

ceithir (approved):

config/locales/controllers/en.yml
```diff
@@ -132,7 +132,9 @@ en:
     tos_faq:
       page_title: Terms of Service FAQ
   inbox:
+    must_select_item: Please select something first.
     page_title: "%{user} - Inbox"
+    update_success: Inbox successfully updated.
```
Bilka2: Not required, but it would be nice to scope these by action instead, like with the other controllers. And then the success message key can just be `success`.

sarken:
Hi, @talvalin -- it's great to see you again! I've updated your Jira account so you'll have permission to comment on, assign, and transition issues in the future. 

As always, if you have any questions, you can contact us at otw-coders@transformativeworks.org.

config/locales/controllers/en.yml
talvalin: For the success message, that's easy enough. 

For `must_select_item` I'm not sure what to use as a better key. The only similar key that I could find in the `locales/controllers/en.yml` file was for removing pseuds and it was my original inspiration (`must_select_pseud`).

lydia-theda: I think Bilka means something like this:
```suggestion
inbox:
    page_title: "%{user} - Inbox"
    update:
        must_select_item: Please select something first.
        success: Inbox successfully updated.
```
(Updating the text is probably out of scope, but: 

must_select_comment: You must select a comment to update.
(or "Please select a comment first.")

either of those would better match must_select_pseud and specify the actual functionality being suggested to the user, imo "something" and "item" are both really vague.)

Bilka2: Yeah, that's what I mean. (Ideally also for the page_title with `show`.)

For the message, `Please select at least one comment first.` would be good.

talvalin: Thank you @lydia-theda and @Bilka2 for the clarification. I've updated as per your feedback, and I updated the text too!

talvalin:
Okay, I have no idea what is going on with the test failures. Any help clearing these up would be greatly appreciated. 🙇 

> Failed examples:
>
> rspec ./spec/lib/i18n/i18n_tasks_spec.rb:10 # I18n en does not have missing keys
> rspec ./spec/lib/i18n/i18n_tasks_spec.rb:15 # I18n en does not have unused keys

app/controllers/inbox_controller.rb
```diff
@@ -14,7 +14,7 @@ def load_user
 
   def show
     authorize InboxComment if logged_in_as_admin?
-    @page_subtitle = t("inbox.page_title", user: @user.login)
+    @page_subtitle = t(".show.page_title", user: @user.login)
```
Bilka2: Just the 
```suggestion
    @page_subtitle = t(".page_title", user: @user.login)
```
here (and in the other keys in this ruby file) will work! The scoping by action works automatically with the [lazy lookup](https://guides.rubyonrails.org/i18n.html#lazy-lookup). That should also fix the tests

config/locales/controllers/en.yml
```diff
@@ -132,7 +132,11 @@ en:
     tos_faq:
       page_title: Terms of Service FAQ
   inbox:
-    page_title: "%{user} - Inbox"
+    show:
+      page_title: "%{user} - Inbox"
+    update:
+      must_select_item: Please select at least one comment first.
```
lydia-theda, line 138: Maybe change the key name to must_select_comment to more closely match the new text?

app/controllers/inbox_controller.rb
talvalin: Got it. Thank you!

PR #5247, opened 2025-07-03
AO3-7022 Default to only registered commentors on news post

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-7022

### DIFF
--- a/features/admins/admin_post_news.feature
+++ b/features/admins/admin_post_news.feature
@@ -37,27 +37,6 @@ Feature: Admin Actions to Post News
     Then 0 emails should be delivered to "testadmin-communications@example.org"
       But 1 email should be delivered to "admin@example.org"
 
-  Scenario: Evil user can impersonate admin in comments
-  # However, they can't use an icon, so the admin's icon is the guarantee that they're real
-  # also their username will be plain text and not a link
-
-    Given I have posted an admin post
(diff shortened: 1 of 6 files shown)


features/admins/admin_post_news.feature
```diff
@@ -37,27 +37,6 @@ Feature: Admin Actions to Post News
     Then 0 emails should be delivered to "testadmin-communications@example.org"
       But 1 email should be delivered to "admin@example.org"
 
-  Scenario: Evil user can impersonate admin in comments
```
brianjaustin, line 40: I don't think this was adding any value, so I deleted it instead of fixing it

Bilka2: I think this was from before we started displaying (Guest) on guest comments and also before we stopped admins from commenting with their admin accounts, so doubly no-value....

Bilka2 (changes requested):
It would nice to have a test that checks that the default permission is indeed the right one. Right now we could revert the change and keep the tests the same and everything would still pass.

I think you can co-opt the "Modifying the comment permissions of an admin post with translations" scenario in comments_adminposts for this, since it doesn't do its intended testing now with the changed default anyway.

Bilka2 (approved):

PR #5248, opened 2025-07-04
AO3-7031 Migrate foreign ids for bookmarks to bigint as well

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-7031

## Purpose

Update the type of columns containing bookmark ids to bigint.
 
## Credit

Bilka


sarken (approved):
...let me actually hit the right button.

PR #5249, opened 2025-07-04
AO3-7031 Add certainty in bookmark ID column data type in test cases

TemperedPetals:
A value of 5_294_967_295 (greater than max 32 bit unsigned integer) will add confidence that bigint types are used for bookmark ID fields.

## Issue

https://otwarchive.atlassian.net/browse/AO3-7031

## Purpose
This change will increase certainty across code base and systems, e.g., redis, elasticsearch, that the Bookmarks ID field is of a size big int (64 bit int), and not merely unsigned int (32 bit int). 
Note: It would be a good idea to review the associated Bookmarks ID field data types inside any redis and elasticsearch systems across environments.

## Testing Instructions

Run the tests for the associated models spec file through automation.

## Credit

Credit however you feel like.

### DIFF
--- a/spec/models/bookmark_spec.rb
+++ b/spec/models/bookmark_spec.rb
@@ -21,24 +21,24 @@ describe Bookmark do
     expect(bookmark.errors[:pseud].first).to eq("can't be blank")
   end
 
-  it "can be tagged if has an id larger than int" do
-    bookmark = build(:bookmark, tag_string: "Huge", id: 2_247_740_375)
+  it "can be tagged if has an id larger than unsigned int" do
+    bookmark = build(:bookmark, tag_string: "Huge", id: 5_294_967_295)
     expect(bookmark).to be_valid
     expect(bookmark.save).to be_truthy
     expect(bookmark.reload.taggings.last.tagger.name).to eq("Huge")
   end
 
-  it "can be collected if has an id larger than int" do
+  it "can be collected if has an id larger than unsigned int" do
     collection = create(:collection)
-    bookmark = build(:bookmark, collection_names: collection.name, id: 2_247_740_375)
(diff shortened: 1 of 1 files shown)


TemperedPetals:
Not sure if this is intended, but there was a test failure for `features/bookmarks/bookmark_share.feature` that did not cause that step to fail. Not sure how wide spread this false positive test behavior is in the code base since I am a new contributor.

Here is where the failure occurs:
https://github.com/otwcode/otwarchive/actions/runs/16073710438/job/45363899808?pr=5249#step:13:39

spec/models/bookmark_spec.rb
```diff
@@ -21,24 +21,24 @@
...
+    bookmark = build(:bookmark, collection_names: collection.name, id: 5_294_967_295)
     expect(bookmark).to be_valid
     expect(bookmark.save).to be_truthy
     expect(bookmark.collections).to include(collection)
     expect(collection.bookmarks).to include(bookmark)
   end
 
-  it "can be hidden if has an id larger than int" do
-    bookmark = create(:bookmark, id: 2_247_740_375)
+  it "can be hidden if has an id larger than unsigned int" do
```
Bilka2, line 40: Can you change this to something like "can be target of admin activity ..."?

Bilka2 (commented):
Thanks for the nudge towards checking on Redis and Elasticsearch! Thankfully it looks like we're good on that side (Redis is used for indexing and stores the ids as strings, ES uses long in the mapping).

Regarding that test, it's flaky and nothing to worry about. You can see that in the `62 scenarios (1 flaky, 61 passed)` line at the end. If a test fails, it's automatically rerun once and only considered failed if it fails again. Unfortunately, our tests with javascript enabled are currently known to be flaky, but we have yet to figure out why.

PR #5252, opened 2025-07-05
AO3-7021 Add language short codes to language names

AbhinavGupta-de:
Took 21 minutes

# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-7021

## Purpose

Added the language short codes to the languages names on the language page. 

## Testing Instructions

As mentioned in the issue, head over to `/language` page and check that language codes are mentioned along side the languages.

## Credit

Abhinav Gupta, he/him

### DIFF
--- a/spec/controllers/languages_controller_spec.rb
+++ b/spec/controllers/languages_controller_spec.rb
@@ -240,7 +240,6 @@ describe LanguagesController do
       end
     end
 
-
     context "when logged in as an admin with policy_and_abuse role and I attempt to edit a non-abuse field" do
       let(:admin) { create(:admin, roles: ["policy_and_abuse"]) }
       before do


app/views/languages/index.html.erb
```diff
@@ -17,12 +17,14 @@
   <% @languages.each do |language| %>
     <% works_count = estimate_number(@works_counts[language.short] || 0) %>
     <% if language == Language.default %>
-      <dt><span lang="<%= language.short %>"><%= language.name %></span></dt>
+      <dt><span lang="<%= language.short %>"><%= "#{language.name} (#{language.short})" %></span></dt>
```
sarken: The short code will need to go outside the `span`, since the code itself is not in the language that the span is specifying. It wouldn't be a bad idea to put the short name inside an `<abbr>` tag, either.

And since we're using parentheses now, which can vary by language, this text also needs i18n. You can check out our [Internationalization Guidelines wiki page](https://github.com/otwcode/otwarchive/wiki/Internationalization-%28i18n%29-Standards) for some info on our i18n standards.

app/views/languages/index.html.erb
```diff
@@ -17,12 +17,14 @@
...
       <dd>
         <%= link_to t(".works_count", count: works_count, formatted_count: number_with_delimiter(works_count)), works_path %>
       </dd>
     <% else %>
-      <dt><%= link_to language.name, language, lang: language.short %></dt>
+      <dt>
+        <%= link_to "#{language.name} (#{language.short})", language, lang: language.short %>
```
sarken: Judging from the issue, I think we want the short code to be separate from the link.

sarken (changes requested):
Thanks for working on this!

app/views/languages/index.html.erb
```diff
@@ -17,12 +17,17 @@
   <% @languages.each do |language| %>
     <% works_count = estimate_number(@works_counts[language.short] || 0) %>
     <% if language == Language.default %>
-      <dt><span lang="<%= language.short %>"><%= language.name %></span></dt>
+      <dt>
+        <span lang="<%= language.short %>"><%= language.name %></span>
+        <abbr title="<%= t(".language_code_abbr_title") %>"><%= t(".language_code_format", code: language.short) %></abbr>
```
sarken: Some screen readers will read the `title` attributes on `abbr` elements in place of the element's text, so they should contain the expanded version of the abbreviation or acronym rather than an explanation of its purpose. However, we _want_ the screen reader to read the abbreviation, not the full version, so we can leave the `title` attribute off completely. 

We also only want the abbreviation itself inside `abbr`, like `(<abbr>en</abbr>)`. The tidiest way to accomplish that is probably by using the `content_tag` helper and adding an `_html` suffix to the key to mark our translation HTML safe: 

```ruby
<%= t(".language_code_format_html", code: content_tag(:abbr, language.short)) %>
```

spec/controllers/languages_controller_spec.rb
```diff
@@ -10,6 +10,13 @@
         get :index
         expect(response).to render_template("index")
       end
+
+      it "displays language names and codes" do
+        create(:language, name: "Deutsch", short: "de")
+        get :index
+        expect(response.body).to include("Deutsch")
+        expect(response.body).to include("(de)")
+      end
```
sarken: We generally prefer to check the content of view files in feature tests. A better place for this check might be at the end of this feature test: https://github.com/otwcode/otwarchive/blob/b6c964443004cac7ebf04664ba08b1618d1a107d/features/admins/admin_languages.feature#L6-L17

config/locales/views/en.yml
```diff
@@ -1731,6 +1731,9 @@ en:
         update: Update Language
       support_available: Support available
     index:
+      language_code_abbr_title: language code
+      language_code_format: "(%{code})"
```
sarken: These can be removed since they're no longer used. That should clear up the unused keys failure in `rspec ./spec/lib/i18n/i18n_tasks_spec.rb:15`.

sarken (commented):
The code itself is looking good!

sarken (approved):
Thank you!

PR #5256, opened 2025-07-07
AI3-6918 Implement 2fa for admins via TOTP

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6918

## Credit

This is a continuation of #5136, so EchoEkhi gets credit as well


app/controllers/admin/preferences_controller.rb
```diff
@@ -0,0 +1,92 @@
+class Admin::PreferencesController < Admin::BaseController
+  before_action :check_ownership
+  before_action :check_totp_disabled, only: [:totp_setup, :totp_setup_form]
+  before_action :check_totp_enabled, only: [:totp_disable, :totp_disable_form]
+
+  def check_ownership
```
Bilka2: Not sure whether I like overriding `check_ownership` from application controller here. Probably fine though

app/controllers/admin/preferences_controller.rb
```diff
@@ -0,0 +1,92 @@
...
+    admin_only_access_denied unless params[:admin_id] == current_admin.login
```
Bilka2: This error message is the generic "only an authorized admin" that we give for admin pages if you're not logged in as admin at all. I think that's confusing here because you are logged in, you're just not the right admin

config/config.yml
```diff
@@ -122,6 +122,8 @@ PASSWORD_LENGTH_MIN: 6
 PASSWORD_LENGTH_MAX: 40
 ADMIN_PASSWORD_LENGTH_MIN: 10
 ADMIN_PASSWORD_LENGTH_MAX: 40
+ADMIN_TOTP_BACKUP_CODE_LENGTH: 10
+ADMIN_TOTP_BACKUP_CODES: 4
```
Bilka2:
```suggestion
ADMIN_TOTP_BACKUP_CODE_COUNT: 4
```

config/application.rb
```diff
@@ -89,6 +89,11 @@ class Application < Rails::Application
       BCrypt::Password
     ]
 
+    # Set admin two-factor authentication keys
+    config.active_record.encryption.primary_key = ENV["ACTIVE_RECORD_ENCRYPTION_PRIMARY_KEY"]
+    config.active_record.encryption.deterministic_key = ENV["ACTIVE_RECORD_ENCRYPTION_DETERMINISTIC_KEY"]
+    config.active_record.encryption.key_derivation_salt = ENV["ACTIVE_RECORD_ENCRYPTION_KEY_DERIVATION_SALT"]
```
Bilka2, line 92: Should this maybe use archive config variables instead, like the secret key base does? What's easiest to use/set in production?

Though there's also https://github.com/otwcode/otwarchive/pull/5039#discussion_r1933676637 ...

app/views/admin/passwords/edit.html.erb
```diff
@@ -19,6 +19,10 @@
     </dd>
     <dt><%= label_tag :edit_password_confirmation, t(".label.confirmation") %></dt>
     <dd><%= f.password_field :password_confirmation, id: :edit_password_confirmation %></dd>
+    <% if @totp_required %>
+      <dt><%= f.label :otp_attempt, t(".label.two_factor_authentication_code") %></dt>
+      <dd><%= f.text_field :otp_attempt, "aria-describedby": "otp-field-description" %></dd>
```
Bilka2:
```suggestion
      <dd><%= f.text_field :otp_attempt %></dd>
```

config/routes.rb
```diff
@@ -227,7 +227,15 @@
...
     resources :api
   end
-  resources :admins, only: [:index]
+  resources :admins, only: [:index] do
+    resource :preferences, controller: "admin/preferences", only: [:show] do
+      get :totp_setup
+      post :totp_setup_form
+      get :totp_setup_backup_codes
+      get :totp_disable
+      post :totp_disable_form
```
Bilka2: These route names aren't Rails-like. If we make them similar to users changing password and username it'd be something like `disable_totp` and `disabled_totp`, though that doesn't work well for the setup one.

Maybe we could instead go for something similar to new and create, so `new_totp`, `create_totp`, similar to how it's named for editing multiple works. For disabling it... maybe `confirm_disable_totp` and `disable_totp`, similar to how deleting and destroying routes are usually setup, or like my recently added remove pseud routes.

At the very least the verb should be at the start of the action name.

(Also for consideration: Should it be scoped to preferences at all, since there is no model backing that)

app/views/admin/sessions/totp.html.erb
```diff
@@ -0,0 +1,17 @@
...
+<h2 class="heading"><%= t(".page_heading") %></h2>
+<!--/descriptions-->
+
+<!--main content-->
+<%= form_for(resource, as: resource_name, url: session_path(resource_name), method: :post) do |form| %>
+  <fieldset>
+  <dl>
+    <dt><%= label_tag "admin[otp_attempt]", t(".two_factor_authentication_code") %></dt>
+    <dd>
+      <%= form.text_field :otp_attempt, "aria-describedby": "otp-field-description" %>
```
Bilka2:
```suggestion
      <%= form.text_field :otp_attempt %>
```

app/views/admin/preferences/totp_disable.erb
```diff
@@ -0,0 +1,21 @@
...
+<!--/descriptions-->
+
+<!--main content-->
+<fieldset>
+  <%= form_for(current_admin, url: totp_disable_form_admin_preferences_path, method: :post) do |form| %>
+    <dl>
+      <dt><%= form.label :password, t(".account_password") %></dt>
+      <dd><%= form.password_field :password %></dd>
+      <dt><%= form.label :otp_attempt, t(".two_factor_authentication_code") %></dt>
+      <dd><%= form.text_field :otp_attempt, "aria-describedby": "otp-field-description" %></dd>
```
Bilka2:
```suggestion
      <dd><%= form.text_field :otp_attempt %></dd>
```

app/views/admin/preferences/totp_disable.erb
```diff
@@ -0,0 +1,21 @@
...
+    </dl>
+    
+    <dt class="landmark"><%= label_tag :submit, t(".submit") %></dt>
```
Bilka2:
```suggestion
    <dt class="landmark"><%= t(".landmark.submit") %></dt>
```

app/views/admin/preferences/totp_setup.erb
```diff
@@ -0,0 +1,33 @@
...
+  <p><%= t(".step_2.instructions") %></p>
+  <%= form_for(current_admin, url: totp_setup_form_admin_preferences_path, method: :post) do |form| %>
+    <dl>
+      <dt><%= form.label :password, t(".step_2.account_password") %></dt>
+      <dd><%= form.password_field :password %></dd>
+      <dt><%= form.label :otp_attempt, t(".step_2.otp_code") %></dt>
+      <dd><%= form.text_field :otp_attempt %></dd>
+    </dl>
+    
+    <dt class="landmark"><%= label_tag :submit, t(".submit") %></dt>
```
Bilka2:
```suggestion
    <dt class="landmark"><%= t(".landmark.submit") %></dt>
```

app/views/admin/preferences/totp_disable.erb
```diff
@@ -0,0 +1,21 @@
+<!--Descriptive page name, messages and instructions-->
+<h2 class="heading"><%= t(".heading") %></h2>
+<!--/descriptions-->
+
+<!--main content-->
+<fieldset>
+  <%= form_for(current_admin, url: totp_disable_form_admin_preferences_path, method: :post) do |form| %>
```
Bilka2: Here and for the other forms: `form_with` is [more modern and recommended](https://guides.rubyonrails.org/form_helpers.html#using-form-tag-and-form-for).

app/views/admin/preferences/totp_disable.erb
```diff
@@ -0,0 +1,21 @@
...
+    <dl>
+      <dt><%= form.label :password, t(".account_password") %></dt>
+      <dd><%= form.password_field :password %></dd>
+      <dt><%= form.label :otp_attempt, t(".two_factor_authentication_code") %></dt>
+      <dd><%= form.text_field :otp_attempt, "aria-describedby": "otp-field-description" %></dd>
+    </dl>
+    
+    <dt class="landmark"><%= label_tag :submit, t(".submit") %></dt>
+    <dd class="submit actions">
+      <%= submit_tag t(".disable") %>
```
Bilka2:
```suggestion
      <%= form.submit t(".disable") %>
```

app/views/admin/preferences/totp_setup.erb
```diff
@@ -0,0 +1,33 @@
...
+    <dd class="submit actions">
+      <%= submit_tag t(".confirm_and_enable") %>
```
Bilka2:
```suggestion
      <%= form.submit t(".confirm_and_enable") %>
```

app/views/admin/preferences/totp_disable.erb
```diff
@@ -0,0 +1,21 @@
+<!--Descriptive page name, messages and instructions-->
+<h2 class="heading"><%= t(".heading") %></h2>
+<!--/descriptions-->
+
+<!--main content-->
+<fieldset>
```
Bilka2: Looking at recent things that got front-end design, we didn't use a fieldset, like https://otwarchive.atlassian.net/browse/AO3-6204. https://sarken.github.io/docs/front_end_coding/patterns-interactions also only suggests using a fieldset when there are multiple forms per page, so I think we shouldn't use one here (same for the other files except totp_setup)

app/views/admin/preferences/totp_setup_backup_codes.html.erb
```diff
@@ -0,0 +1,21 @@
+<!--Descriptive page name, messages and instructions-->
+<h2 class="heading"><%= t("admin.preferences.totp_setup.two_factor_authentication_set_up") %></h2>
```
Bilka2: Probably shouldn't be reusing the locale from totp_setup.erb here, it's a different page - I find it confusing that it has the same heading

app/views/admin/preferences/totp_disable.erb
Bilka2, line 1: This file and totp_setup should have the `html.erb` extension, not `.erb`

app/mailers/admin_mailer.rb
```diff
@@ -29,7 +29,19 @@ def set_password_notification(admin, token)
...
+  end
+
+  # Sends an email to the admin who lost their 2FA device when a sysadmin runs
+  # the script in script/send_backup_codes_to_admin.rb
+  def totp_2fa_backup_codes(admin, codes)
+    @admin = admin
+    @codes = codes
+
+    mail(
+      to: admin.email,
```
Bilka2:
```suggestion
      to: @admin.email,
```

app/views/admin/passwords/edit.html.erb
```diff
@@ -19,6 +19,10 @@
     </dd>
     <dt><%= label_tag :edit_password_confirmation, t(".label.confirmation") %></dt>
     <dd><%= f.password_field :password_confirmation, id: :edit_password_confirmation %></dd>
+    <% if @totp_required %>
```
Bilka2: Noting that this page is used for resetting your admin password (https://otwarchive.atlassian.net/browse/AO3-6440), for my own understanding of why 2fa is here

app/views/admin/preferences/show.html.erb
Bilka2, line 1: This page is in need of some front-end design, but I don't know what besides it looking off right now

app/controllers/admin/preferences_controller.rb
Bilka2, line 1: A lot of the pages here could use better page titles, though the change in routes might already improve the auto-generated page titles a bit

config/config.yml
Bilka2: and I think we should have more than 4? Most places give me 10 backup codes for 2fa from what I can tell. We could go lower than that since we can regenerate them if person can prove their identity to us, but 4 seems very low

config/config.yml
```diff
@@ -122,6 +122,8 @@ PASSWORD_LENGTH_MIN: 6
 PASSWORD_LENGTH_MAX: 40
 ADMIN_PASSWORD_LENGTH_MIN: 10
 ADMIN_PASSWORD_LENGTH_MAX: 40
+ADMIN_TOTP_BACKUP_CODE_LENGTH: 10
```
Bilka2: Considering https://github.com/devise-two-factor/devise-two-factor/discussions/281 I think this should be at least 16

config/initializers/filter_parameter_logging.rb
```diff
@@ -4,5 +4,5 @@
 # Use this to limit dissemination of sensitive information.
 # See the ActiveSupport::ParameterFilter documentation for supported notations and behaviors.
 Rails.application.config.filter_parameters += [
-  :passw, :secret, :token, :_key, :crypt, :salt, :certificate, :otp, :ssn
+  :passw, :secret, :token, :_key, :crypt, :salt, :certificate, :otp, :ssn, :otp_attempt
```
Bilka2: These are partially matched so this addition isn't needed - but if it were then it should go into application.rb instead

app/views/admin/preferences/totp_setup_backup_codes.html.erb
```diff
@@ -0,0 +1,21 @@
...
+    <div>
+      <code class="big"><%= code %></code>
+    </div>
+  <% end %>
+  <p><%= t(".step_3.you_wont_see_them_again") %></p>
+</fieldset>
+
+<dt class="landmark"><%= t(".finish") %></dt>
+<dd class="submit actions">
+  <%= link_to t(".finish"), admins_path %>
```
Bilka2: This button should be inside the fieldset (if we use a fieldset)

public/stylesheets/site/2.0/02-elements.css
```diff
@@ -129,9 +129,15 @@ blockquote p:last-child {
 
 kbd, tt, code, var, pre, samp {
   font: normal 0.857em 'Monaco', 'Consolas', Courier, monospace;
+  font-variant-ligatures: none;
   margin: auto;
 }
 
+code.big {
```
Bilka2: the `big` class clashes with a native HTML tag and I don't think we name classes this way. I think this might need the eyes of sarken? The text also feels too big

config/locales/mailers/en.yml
```diff
@@ -39,6 +39,9 @@ en:
       subject: "[%{app_name}] Your AO3 admin account"
       url: Admin login URL
       username: Admin username
+    totp_2fa_backup_codes:
+      created: A set of new backup codes has been requested for your account. Please use one of the following to log in, and then set up two-factor authentication again.
```
Bilka2: For me this implies that entering a backup code will disable 2FA, but it doesn't. It just enables you to log in and then disable/re-enable it (if you want to) TODO Bilka see this convo https://github.com/otwcode/otwarchive/pull/5136#discussion_r2083320481, probably fine?

See the other TODO Bilka comment, maybe it does disable 2FA but doesn't update the state I can see from my preferences when I try to disable it there?

config/locales/mailers/en.yml
```diff
@@ -39,6 +39,9 @@ en:
...
+      subject: "[%{app_name}] Your admin account recovery codes"
```
Bilka2: `backup codes` to be consistent?

app/controllers/admin/preferences_controller.rb
```diff
@@ -0,0 +1,92 @@
...
+  def totp_disable
+  end
+
+  def totp_disable_form
+    unless current_admin.valid_password?(enable_2fa_params[:password])
+      flash[:error] = t("devise.failure.admin.invalid")
+      return redirect_to totp_disable_admin_preferences_path
+    end
+
+    unless current_admin.validate_and_consume_otp!(enable_2fa_params[:otp_attempt])
```
Bilka2: I logged in with a backup code, went to go disable 2FA and it keeps giving me this error, no matter what backup code or 2FA I try. Normal logging in with 2FA worked beforehand. TODO Bilka investigate what the state is for me right now?

Bilka2 (commented):
Submitting already with partial review, as requested. Barely looked at the controller code, didn't look at wording besides the email, didn't look at tests.

Bilka2:
The migration in this PR needs to be updated to be have a timestamp after July 4, as defined in the schema file: https://github.com/otwcode/otwarchive/blob/10c2842aac0776aa809ce5ca4a2d10dc8472d4a3/db/schema.rb#L13

config/config.yml
brianjaustin: I think we don't want to encourage people to use these rather than figuring out a more permanent solution (like 1Password TOTP codes)

config/application.rb
brianjaustin: Yeah, we currently do secrets via environment variables (for the most part). That's also preferable for containerisation. In terms of what's easier to set with our current deployment mechanism, either requires a file change (it just may be a different one) so neither is particularly easy compared to the other

config/routes.rb
brianjaustin: Agreed about the scope. I ended up splitting it into its own controller (sorry about the larger diff)

app/controllers/admin/preferences_controller.rb
brianjaustin: Weird... when I use a backup code, the behaviour is to keep 2fa enabled as normal :thinking:

config/locales/mailers/en.yml
brianjaustin: Added a small clarification; what do you think?

docker-compose.yml
```diff
@@ -77,6 +77,9 @@ services:
       - RACK_TIMEOUT_SERVICE_TIMEOUT=0
       - GITPOD_WORKSPACE_ID=${GITPOD_WORKSPACE_ID:-}
       - GITPOD_WORKSPACE_CLUSTER_HOST=${GITPOD_WORKSPACE_CLUSTER_HOST:-}
+      - ACTIVE_RECORD_ENCRYPTION_PRIMARY_KEY=bEZjYLY9tCYGh6WlcMtEJpIi7GO2plZC
+      - ACTIVE_RECORD_ENCRYPTION_DETERMINISTIC_KEY=PpLWizzsQHIWnIihtECw8nDHZQd0amzf
+      - ACTIVE_RECORD_ENCRYPTION_KEY_DERIVATION_SALT=3S99KdpdEWLnCYudBgUfdCFDBWePWCud
```
Bilka2: we probably also need these in the workers container now? if yes, it'd probably be nice to use yml anchors (?) for these values so we don't have to duplicate them

app/controllers/admin/base_controller.rb
```diff
@@ -1,3 +1,18 @@
...
   before_action :admin_only
+
+  protected
+
+  # This is the equivalent of the check_ownership method, but for admins.
+  # It's mostly used for admin-specific account configuration, like TOTP setup.
+  def require_admin_owner
+    return if params[:admin_id] == current_admin.login
+
+    respond_to do |format|
```
Bilka2: Do we expect to hit this with different formats where we don't want to redirect? `check_ownership` redirects for all formats

app/models/admin.rb
```diff
@@ -40,4 +43,43 @@ def send_set_password_notification
     token = set_reset_password_token
     AdminMailer.set_password_notification(self, token).deliver
   end
+
+  def to_param
+    login
+  end
```
Bilka2, line 49: Noting that this will have a dash in it (e.g. ao3.org/admin-Bilka/foo), but that's likely fine

app/controllers/admin/base_controller.rb
```diff
@@ -1,3 +1,18 @@
 class Admin::BaseController < ApplicationController
   before_action :admin_only
+
+  protected
+
+  # This is the equivalent of the check_ownership method, but for admins.
+  # It's mostly used for admin-specific account configuration, like TOTP setup.
+  def require_admin_owner
```
Bilka2: If we want to follow the comment and match `check_ownership`, then this should not hardcode the url param in this method. Instead we'd be setting `@check_ownership_of` and checking whether that's the `current_admin` here. (And also rename this to something like `check_admin_ownership`)

But I'm not convinced that we _need_ to match the other method. My biggest issue with this method is that it uses `params[:admin_id]` here when the route itself is defined for only the descending controllers, so it feels like we're violating principles of information encapsulation/decoupling

app/models/admin.rb
```diff
@@ -40,4 +43,43 @@ def send_set_password_notification
...
+  def generate_otp_secret_if_missing!
+    return unless otp_secret.nil?
+
+    update!(otp_secret: Admin.generate_otp_secret)
+  end
+
+  # Ensure that the user is prompted for their OTP when they login
+  def enable_otp!
+    update!(otp_required_for_login: true)
+  end
```
Bilka2: I know this is from the tutorial but it's a bit strange to have this method but then everywhere just reads directly from `otp_required_for_login` to check whether it's enabled. I don't know if this is anything actionable, though. Maybe it'd be good to have a method for checking whether it's enabled too?

app/controllers/admin/preferences_controller.rb
```diff
@@ -0,0 +1,7 @@
+class Admin::PreferencesController < Admin::BaseController
+  before_action :require_admin_owner
+
+  def show
+    @totp_enabled = current_admin.otp_required_for_login
```
Bilka2: Wondering whether we can't just get this directly from the admin in the view since it's loaded and this is just a bool. No idea what's better though, since I suppose this is clearer in the controller

config/routes.rb
```diff
@@ -231,7 +231,14 @@
     end
     resources :api
   end
-  resources :admins, only: [:index]
+  resources :admins, only: [:index] do
+    resource :preferences, controller: "admin/preferences", only: [:show]
+    resource :totp, controller: "admin/totp", only: [:create, :new] do
+      get :show_backup_codes
+      get :confirm_disable
+      post :disable_totp
```
Bilka2: Should these two routes match for whether they have `totp` in the name? I suppose the concern here is how the routes work out when we add another 2fa method: one controller per method or one controller for 2fa so different actions per method or ...?

app/controllers/admin/totp_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+
+    flash[:error] = t("admin.totp.already_disabled")
+    redirect_to admins_path
+  end
+
+  def check_totp_disabled
+    return unless current_admin.otp_required_for_login
+
+    flash[:error] = t("admin.totp.already_enabled")
+    redirect_to admins_path
```
Bilka2, line 82: Could these maybe redirect to `admin_preferences_path` instead? Since that is where the admin is likely coming from

app/controllers/admin/totp_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+  end
+
+  def enable_2fa_params
```
Bilka2:
```suggestion
  def totp_params
```
to match the name of the controller

app/controllers/admin/totp_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+    @backup_codes = current_admin.generate_otp_backup_codes!
+    current_admin.save!
+  end
+
+  def confirm_disable
+    @page_subtitle = t(".page_title")
+  end
+
+  def disable_totp
+    unless current_admin.valid_password?(enable_2fa_params[:password])
```
Bilka2: Suggestion to rename `:password` to `:password_check` to match `reauthenticate` used for users

(same in `create`)

and this can probably also be done with a separate, reusable method since it's the same code in `create`, also to match the users controller

This sounds a bit silly, but should this check 2fa again...?

app/controllers/admin/totp_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+      flash[:error] = t("devise.failure.admin.invalid")
+      return redirect_to confirm_disable_admin_totp_path
```
Bilka2: I think this should `render` instead of redirect, since we have a form fill failure and likely want to keep the input. If yes, then the flash should be a `flash.now`

app/controllers/admin/totp_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+    end
+
+    if current_admin.validate_and_consume_otp!(enable_2fa_params[:otp_attempt])
+      current_admin.enable_otp!
+
+      flash[:notice] = t(".success")
+      redirect_to show_backup_codes_admin_totp_path
+    else
+      flash[:error] = t(".incorrect_code")
+      redirect_to new_admin_totp_path
```
Bilka2: I think this should `render` instead of redirect, since we have a form fill failure and likely want to keep the input that was entered in the form. If yes, then the flash should be a `flash.now`

app/controllers/admin/totp_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+    end
+  end
+
+  def show_backup_codes
+    unless current_admin.otp_required_for_login
+      flash[:error] = t(".not_enabled")
+      return redirect_to new_admin_totp_path
```
Bilka2: This is copied from the tutorial, but we should do `redirect_to *_path and return`, for code style consistency (here and everywhere)

app/models/admin.rb
Bilka2, line 1: Should the methods here use `totp` instead of `otp`, for consistency with the controller etc?

app/controllers/admin/totp_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+    current_admin.save!
+  end
+
+  def confirm_disable
+    @page_subtitle = t(".page_title")
+  end
+
+  def disable_totp
+    unless current_admin.valid_password?(enable_2fa_params[:password])
+      flash[:error] = t("devise.failure.admin.invalid")
```
Bilka2: We should use a custom error message, like in UsersController#reauthenticate

app/controllers/admin/totp_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+      return redirect_to confirm_disable_admin_totp_path
+    end
+
+    if current_admin.disable_otp!
+      flash[:notice] = t(".success")
+    else
+      flash[:error] = t(".failure")
+    end
+
+    redirect_to admin_preferences_path
```
Bilka2: What do you think about redirecting to this on success, but going back to `confirm_disable` on failure?

app/controllers/application_controller.rb
```diff
@@ -164,7 +164,14 @@ def ensure_user_credentials
...
   end
 
-protected
+  # Allow otp_attempt parameter in the :sign_in controller for admin two-factor authentication
+  before_action :configure_permitted_parameters, if: :devise_controller?
+
+  protected
+
+  def configure_permitted_parameters
+    devise_parameter_sanitizer.permit(:sign_in, keys: [:otp_attempt])
```
Bilka2: Why is this here instead of the sessions (and passwords?) controller?

app/controllers/admin/sessions_controller.rb
```diff
@@ -4,10 +4,74 @@ class Admin
...
+      if valid_otp_attempt?(admin)
+        # Remove any lingering admin data from login
+        session.delete(:otp_admin_id)
+  
+        admin.save!
+
+        flash[:notice] = t("devise.sessions.signed_in")
+        sign_in(admin, event: :authentication)
+        
+        redirect_to admins_path
```
Bilka2, line 54: Is this custom redirect required? It's not in the tutorial

app/controllers/admin/sessions_controller.rb
```diff
@@ -4,10 +4,74 @@ class Admin
...
+      else
+        flash.now[:alert] = t("admin.sessions.invalid_totp")
```
Bilka2:
```suggestion
        flash.now[:error] = t("admin.sessions.invalid_totp")
```

app/controllers/admin/passwords_controller.rb
```diff
@@ -3,4 +3,44 @@
...
+
+    flash[:error] = t("admin.sessions.invalid_totp")
+
+    redirect_to edit_admin_password_path(reset_password_token: admin_params[:reset_password_token])
+  end
+
+  private
+
+  def find_admin_by_reset_password_token(original_token)
+    reset_password_token = Devise.token_generator.digest(self, :reset_password_token, original_token)
```
Bilka2:
```suggestion
  def find_admin_by_reset_password_token(token_from_url)
    reset_password_token = Devise.token_generator.digest(self, :reset_password_token, token_from_url)
```
based on the feedback I got on #5350

app/controllers/admin/passwords_controller.rb
```diff
@@ -3,4 +3,44 @@
...
+
+    Admin.find_by(reset_password_token: reset_password_token)
```
Bilka2:
```suggestion
    Admin.find_for_authentication(reset_password_token: reset_password_token)
```

matching https://github.com/otwcode/otwarchive/pull/5350

Also needs `admin.nil? || admin.new_record?` in the methods then, instead of just checking nil

app/controllers/admin/passwords_controller.rb
```diff
@@ -3,4 +3,44 @@
...
   before_action :user_logout_required
   layout "session"
+
+  before_action :check_if_totp_required, only: [:edit]
+  before_action :verify_otp_code, only: [:update]
+
+  def check_if_totp_required
+    admin = find_admin_by_reset_password_token(params[:reset_password_token])
+
+    return unless admin
```
Bilka2: I would like an error here instead of just returning, just in case shit is funky and `edit` does later find an admin. Better safe than sorry when it's about security

app/controllers/admin/passwords_controller.rb
```diff
@@ -3,4 +3,44 @@
 class Admin::PasswordsController < Devise::PasswordsController
   before_action :user_logout_required
   layout "session"
+
+  before_action :check_if_totp_required, only: [:edit]
+  before_action :verify_otp_code, only: [:update]
```
Bilka2: Should these be before actions, or should we override the methods like we do in Users::PasswordsController?

app/controllers/admin/passwords_controller.rb
```diff
@@ -3,4 +3,44 @@
...
+
+    return unless admin
+
+    @totp_required = admin.otp_required_for_login
+  end
+
+  def verify_otp_code
+    admin = find_admin_by_reset_password_token(admin_params[:reset_password_token])
+
+    return unless admin
```
Bilka2: I would like an error here instead of just returning, just in case shit is funky and `update` does later find an admin. Better safe than sorry when it's about security

app/controllers/admin/passwords_controller.rb
```diff
@@ -3,4 +3,44 @@
...
+  end
+
+  def valid_otp_attempt?(admin)
+    admin.validate_and_consume_otp!(admin_params[:otp_attempt]) ||
+      admin.invalidate_otp_backup_code!(admin_params[:otp_attempt])
+  end
+    
+  def admin_params
+    params.require(:admin).permit(:reset_password_token, :otp_attempt, :password, :password_confirmation)
+  end
```
Bilka2: The whole param situation just seems weird, why do we have this here and in sessions + totp controller but also the devise one in base controller... it's copied from the tutorial but I am very confused. Do you know more? Do we need this at all? Afaik the permit isn't needed if we're directly accessing all the params and not mass assigning

app/controllers/admin/passwords_controller.rb
Bilka2, line 1: By having the OTP in the form you get to from the email, we technically expose whether the admin has totp enable to whoever has access to the link. This may be bad? Would we maybe want to make it a separate page... but it's not like the password you enter in the "change your password" form proves in any way that you own the account, so making it another page afterwards doesn't really change much

Personally I find it a bit strange though when the field is on the same form since the password fields are not time limited, but the OTP is, so I can't fill that one first, which isn't very obvious

script/send_backup_codes_to_admin.rb
```diff
@@ -0,0 +1,28 @@
...
+
+unless admin.otp_required_for_login
+  puts "This admin does not have TOTP two-factor authentication enabled."
+  return
+end
+
+codes = admin.generate_otp_backup_codes!
+admin.save!
+
+AdminMailer.totp_2fa_backup_codes(admin, codes).deliver
```
Bilka2:
```suggestion
AdminMailer.totp_2fa_backup_codes(admin, codes).deliver_now
```

script/send_backup_codes_to_admin.rb
```diff
@@ -0,0 +1,28 @@
...
+  Enter the login of the admin to generate and send recovery codes to:
+PROMPT
+
+login = gets.chomp
+
+login = login.strip
+admin = Admin.find_by(login: login)
+
+unless admin
+  puts "Admin not found."
```
Bilka2:
```suggestion
  puts "Admin #{login} not found."
```

script/send_backup_codes_to_admin.rb
```diff
@@ -0,0 +1,28 @@
...
+  puts "This admin does not have TOTP two-factor authentication enabled."
+  return
+end
+
+codes = admin.generate_otp_backup_codes!
+admin.save!
+
+AdminMailer.totp_2fa_backup_codes(admin, codes).deliver
+
+puts "Backup codes successfully sent to the admin."
```
Bilka2:
```suggestion
puts "Backup codes successfully sent to the admin #{login}."
```

script/send_backup_codes_to_admin.rb
```diff
@@ -0,0 +1,28 @@
...
+login = login.strip
+admin = Admin.find_by(login: login)
+
+unless admin
+  puts "Admin not found."
+  return
+end
+
+unless admin.otp_required_for_login
+  puts "This admin does not have TOTP two-factor authentication enabled."
```
Bilka2:
```suggestion
  puts "The admin #{login} does not have TOTP two-factor authentication enabled."
```

config/locales/mailers/en.yml
Bilka2: Better, thank you! Though we should probably be clearer that like "please dont rely on backup codes. if your 2fa doesnt work right now then disable and set up again ..."

app/controllers/admin/passwords_controller.rb
```diff
@@ -3,4 +3,44 @@
...
+  def verify_otp_code
+    admin = find_admin_by_reset_password_token(admin_params[:reset_password_token])
+
+    return unless admin
+
+    return unless admin.otp_required_for_login && !valid_otp_attempt?(admin)
+
+    flash[:error] = t("admin.sessions.invalid_totp")
+
+    redirect_to edit_admin_password_path(reset_password_token: admin_params[:reset_password_token])
```
Bilka2: I think this should `render` instead of redirect, since we have a form fill failure and likely want to keep the input. If yes, then the flash should be a `flash.now`

Bilka2 (changes requested):
Not looked at: Tests, view files, locale

Mostly looked at whether it followed the tutorial/made sensible modifications, not whether the tutorial itself is correct

config/routes.rb
brianjaustin: I think I would do 1 controller per mechanism, especially because something like u2fa we could just replace with passkeys at this point :shrug:

app/controllers/admin/preferences_controller.rb
brianjaustin: Based on one of your earlier comments, I think we can get rid of this and do something like `current_admin.totp_enabled?`

app/controllers/admin/sessions_controller.rb
brianjaustin: Yes; it's because we have a second screen for the TOTP code, which isn't the case in the tutorial. If it's not there, the redirect is back to the login page, which results in a flash error (correctly saying you can't do that because you're already logged in as an admin).

app/controllers/application_controller.rb
brianjaustin: This is how [the library's getting started instructions](https://github.com/devise-two-factor/devise-two-factor?tab=readme-ov-file#getting-started) recommend doing it. I think it also makes sense since we plan to extend this to users later on, and I would want to avoid duplicating this in all the sign in places (easier to miss one)

Bilka2:

app/views/admin/passwords/edit.html.erb
sarken: Do we still want this code in the PR? We moved password resets with 2FA to its own issue and as far as I could tell while working on the front end, this doesn't do anything currently.

sarken:
I would appreciate also having some coding credit on this, please and thank you! (Front end is done; just left one question for Brian.)

docker-compose.yml
```diff
@@ -71,12 +71,15 @@ services:
...
-      - RAILS_ENV=development
+    environment: &environment
+      RAILS_ENV: development
       # Disable rack-timeout
-      - RACK_TIMEOUT_SERVICE_TIMEOUT=0
-      - GITPOD_WORKSPACE_ID=${GITPOD_WORKSPACE_ID:-}
-      - GITPOD_WORKSPACE_CLUSTER_HOST=${GITPOD_WORKSPACE_CLUSTER_HOST:-}
+      RACK_TIMEOUT_SERVICE_TIMEOUT: 0
+      GITPOD_WORKSPACE_ID: ${GITPOD_WORKSPACE_ID:-}
+      GITPOD_WORKSPACE_CLUSTER_HOST: ${GITPOD_WORKSPACE_CLUSTER_HOST:-}
```
Bilka2, line 79: I think it would be better to not share these, but only the active record encryption and rails env

config/locales/mailers/en.yml
```diff
@@ -39,6 +39,9 @@ en:
       subject: "[%{app_name}] Your AO3 admin account"
       url: Admin login URL
       username: Admin username
+    totp_2fa_backup_codes:
+      created: A set of new backup codes has been requested for your account. Please use one of the following to log in, then disable and set up two-factor authentication again.
```
Bilka2: Suggested wording update, sourced from lydia, based on my thought that we should tell people why we want them to set it up again, and lydia's thought from the old pr that we should tell them that the old codes were invalidated:

```
A set of new backup codes has been requested for your account, and your old backup codes have been invalidated. Please use one of the following codes to log in:
[list of codes]
We strongly recommend that you do not rely on the use of backup codes to access your account. Once you have logged in, please disable two-factor authentication and set it up again.
```

app/views/admins/index.html.erb
```diff
@@ -9,21 +9,48 @@
...
+          </ul>
+        <% else %>
+          <p><%= t(".account.roles.none") %></p>
+        <% end %>
+      </dd>
+    </div>
+
+    <div class="verification wrapper">
+      <dt><%= t(".account.totp.label") %></dt>
+      <% if current_admin.otp_required_for_login %>
```
Bilka2:
```suggestion
      <% if current_admin.totp_enabled? %>
```

app/controllers/admin/totp_controller.rb
```diff
@@ -0,0 +1,85 @@
...
+  end
+
+  def create
+    if current_admin.validate_and_consume_otp!(params[:totp_attempt])
+      current_admin.enable_totp!
+
+      flash[:notice] = t(".success")
+      redirect_to show_backup_codes_admin_totp_path
+    else
+      flash[:error] = t(".incorrect_code")
```
Bilka2:
```suggestion
      flash.now[:error] = t(".incorrect_code")
```

app/models/admin.rb
```diff
@@ -40,4 +43,48 @@ def send_set_password_notification
...
+
+  # URI for TOTP two-factor QR code
+  def totp_qr_code_uri
+    otp_provisioning_uri(login, issuer: ArchiveConfig.APP_NAME)
+  end
+
+  # Determine if backup codes have been generated
+  def totp_backup_codes_generated?
+    otp_backup_codes.present?
+  end
```
Bilka2: Unused
```suggestion
```

app/views/admin/totp/reauthenticate_create.erb
```diff
@@ -0,0 +1,36 @@
...
+  <p><%= t(".about.general_info") %></p>
+  <p><%= t(".about.effects_of_enabling") %></p>
+</section>
+<!--/descriptions-->
+
+<!--main content-->
+<section class="step group">
+  <h3 class="heading"><%= t(".app_setup.heading") %></h3>
+  <p><%= t(".app_setup.instructions") %></p>
+  <div class="annotation module">
```
Bilka2, line 15: This appears to be the only centered thing on a left-justified page, is that intended?
Screenshot_20260109_165953

app/controllers/admin/totp_controller.rb
```diff
@@ -0,0 +1,85 @@
...
+
+  def reauthenticate_create
+    unless current_admin.valid_password?(params[:password_check])
+      flash[:error] = t(".incorrect_password")
+      redirect_to new_admin_totp_path and return
+    end
+
+    current_admin.generate_totp_secret_if_missing!
+    @page_subtitle = t(".page_title")
+    render "confirm_enable"
```
Bilka2: You don't have to action this if it's too annoying, but it kept throwing me off that this view is named differently than the action, thinking that you had introduced an extra step in here. You haven't, it's reauthenticate_create -> create -> show_backup_codes as it should be, just with a different view name in there

config/locales/views/en.yml
```diff
@@ -418,6 +423,43 @@ en:
...
+    totp:
+      confirm_disable:
+        page_heading: Disable Two-Step Verification
+        page_title: Disable Two-Step Verification
+        password_label: Admin password
+        required_before_disabling: Before disabling two-step verification, you need to verify your password.
+        submit: Disable Two-Step Verification
+      confirm_enable:
+        about:
+          effects_of_enabling: When two-step verification is enabled, logging in to your AO3 admin account will require both your admin password and a verification code from your authenticator app.
```
Bilka2: Can we use the app_name variable for AO3 here?

config/locales/views/en.yml
```diff
@@ -418,6 +423,43 @@ en:
         update: Update
       update:
         success: Archive settings were successfully updated.
+    totp:
+      confirm_disable:
+        page_heading: Disable Two-Step Verification
+        page_title: Disable Two-Step Verification
```
Bilka2: Please move the page_title's to controllers/en.yml (if you don't want to bother I am open to making it a first timer issue instead)

features/admins/authenticate_admins_with_totp.feature
Bilka2, line 1: I would like to see a test for regenerating the backup codes

features/admins/authenticate_admins_with_totp.feature
```diff
@@ -0,0 +1,202 @@
...
+      And I am logged in as admin "admin" with password "testpassword"
+    When I follow "Set up two-step verification with an authenticator app"
+    Then I should see "Confirm Password"
+    When I fill in "Admin password" with "testpassword"
+      And I press "Continue"
+    Then I should see "Set Up Two-Step Verification"
+    When I fill in a valid TOTP two-step verification code for admin "admin"
+      And I press "Enable Two-Step Verification"
+    Then I should see "Successfully enabled two-step verification; please make note of your backup codes."
+      And I should see "Finish"
```
Bilka2, line 16: Could you follow this and check that you end up on the right page

spec/controllers/admin/totp_controller_spec.rb
Bilka2, line 1: Would it be possible to have a test that makes sure that regenerating the backup codes invalidates the existing ones? (I tested it manually, and it does, but I would like to make sure). Maybe here or as a model test, pretty sure you can't do it as an cucumber test for test setup reasons.

Bilka2 (changes requested):
Just a bunch of tiny things and some requests for more tests!

docker-compose.yml
brianjaustin: Discussed elsewhere, and we'll leave this for now (limitation of YAML anchors)

app/views/admin/totp/reauthenticate_create.erb
sarken: Yup! It's sort of the main attraction, so we want to draw attention to it

PR #5257, opened 2025-07-08
AO3-7008 Fix invitations with blank emails being retrieved

talvalin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-7008

## Purpose

Fixes the problem in production where if you click on "Look me up" on the Invitation Request Status page without entering anything in the email field, an old invitation with a blank email is retrieved that can be resent.

## Credit

talvalin
he/him


app/controllers/invite_requests_controller.rb
```diff
@@ -10,17 +10,20 @@ def index
 
   # GET /invite_requests/1
   def show
-    @invite_request = InviteRequest.find_by(email: params[:email])
-
-    if @invite_request.present?
-      @position_in_queue = @invite_request.position
+    if params[:email].blank?
+      @email_blank = true
```
ceithir, line 14: Definitely not sure it's better. But, in theory, it should be possible to use an early return here, something like (quick example, please don't trust me on the exact syntax):
```rb
if params[:email].blank?
  return respond_to do |format|
    format.html { render "invite_requests/_no_email" }
    format.js { render "invite_requests/_no_email" }
  end
end
```
Pointing directly at the right "final" erb/js templates, without the need to pass around that email_blank variable or modify existing templates.

ceithir (commented):
Two questions about the exact scope of the change:

Should we also add a `live_validation_for_field` alongside https://github.com/otwcode/otwarchive/blob/4aaf5eac5af2e5a9c16648886edbd4fc35336a0b/app/views/invite_requests/status.html.erb#L21 ?

Does the resend also need a non blankness check https://github.com/otwcode/otwarchive/blob/4aaf5eac5af2e5a9c16648886edbd4fc35336a0b/app/controllers/invite_requests_controller.rb#L28 ?

app/controllers/invite_requests_controller.rb
Bilka2: I think this might not work for the javascript version :thinking:

Bilka2:
> Should we also add a live_validation_for_field

No need!

> Does the resend also need a non blankness check 

Yes please!

PR #5258, opened 2025-07-09
AO3-7033 Include skin title in browser page title for skin show page

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-7033

## Purpose

Add the skin title to the browser page title for the skin show page. We should at some future point move the `hmtl_safe` call to the title attribute on the model instead of sprinkling it into views, similar to how it's done with works. I suspect Brakeman will complain but it's fine because it's sanitized.

## Credit

Bilka

### DIFF
--- a/app/controllers/skins_controller.rb
+++ b/app/controllers/skins_controller.rb
@@ -44,6 +44,7 @@ class SkinsController < ApplicationController
 
   # GET /skins/1
   def show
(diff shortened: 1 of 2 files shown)


sarken (approved):
Grazie! I'll leave this for @brianjaustin to merge once the tests have run. I look forward to having this on my dev environment later :D

PR #5262, opened 2025-07-11
AO3-7026 Add recheck_reason to Akismet attributes for comment edits

EchoEkhi:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/issues/AO3-7026

## Purpose

Add recheck_reason to Akismet attributes for comment edits

## Credit

EchoEkhi

### DIFF
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -116,6 +116,8 @@ class Comment < ApplicationRecord
       comment_author = user.login
     end
 
+    recheck_reason = comment_content_changed? ? "edit" : nil
+
     {
       comment_type: comment_type,
       key: ArchiveConfig.AKISMET_KEY,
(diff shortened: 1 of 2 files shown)


app/models/comment.rb
```diff
@@ -127,6 +127,10 @@ def akismet_attributes
       comment_author_email: comment_owner_email,
       comment_content: comment_content
     }
+
+    attributes[:recheck_reason] = "edit" if edited_at_changed? && comment_content_changed?
```
Bilka2: This will be used in the `edited_spam` validation, which uses `will_save_change_to_*` to check whether the attributes are dirty. So I think it would be better to match that, but it is hard to automatically test for what needs to be here...

PR #5263, opened 2025-07-11
AO3-7015 Add data-updated-at attribute to work blurb HTML for crawlers

EchoEkhi:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/issues/AO3-7015

## Purpose

Add data-updated-at attribute to work blurb HTML for crawlers.

## Testing Instructions

Open a page with work blurbs, inspect the blurb's element, and see  if the data-updated-at attribute is correct.

## Credit

EchoEkhi

### DIFF
--- a/app/views/works/_work_blurb.html.erb
+++ b/app/views/works/_work_blurb.html.erb
@@ -1,5 +1,5 @@
 <% # expects "work" %>
-<li id="work_<%= work.id %>" class="<% if is_author_of?(work) %>own <% end %><% if work.unrevealed? %>mystery <% end %>work <%= css_classes_for_creation_blurb(work) %>" role="article">
+<li id="work_<%= work.id %>" class="<% if is_author_of?(work) %>own <% end %><% if work.unrevealed? %>mystery <% end %>work <%= css_classes_for_creation_blurb(work) %>" role="article" data-updated-at="<%= work.updated_at.to_i %>">
   <%= render "works/work_module", work: work %>
   <% if is_author_of?(work) %>
     <h6 class="landmark heading"><%= ts("Author Actions") %></h6>
--- a/features/works/work_browse.feature
+++ b/features/works/work_browse.feature
@@ -178,3 +178,9 @@ Scenario: Can also browse work indexed by language
       Then I should see "1 Work in Deutsch"
     When I browse works in language "Persian"
       Then I should see "0 Works in Persian"
+
+Scenario: Work blurb includes correct data-updated-at attribute
+  Given it is currently 2025-04-12 17:00 UTC
+    And the work "Test"
+  When I go to the works page
+  Then I should see the data-updated-at attribute to be around 1744477200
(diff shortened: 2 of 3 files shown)


app/views/works/_work_blurb.html.erb
```diff
@@ -1,5 +1,5 @@
 <% # expects "work" %>
-<li id="work_<%= work.id %>" class="<% if is_author_of?(work) %>own <% end %><% if work.unrevealed? %>mystery <% end %>work <%= css_classes_for_creation_blurb(work) %>" role="article">
+<li id="work_<%= work.id %>" class="<% if is_author_of?(work) %>own <% end %><% if work.unrevealed? %>mystery <% end %>work <%= css_classes_for_creation_blurb(work) %>" role="article" data-updated-at="<%= work.updated_at.to_i %>">
```
Bilka2: The Jira issue is for making this a comment, not an attribute

EchoEkhi: Yeah but I think an attribute would be easier for crawlers to parse. @zz9pzza what do you think?

brianjaustin: For now, let's keep to the description in Jira as currently written. It's possible to reassess later if we get feedback from 3rd party devs :slightly_smiling_face:

EchoEkhi: OK but I'm a bit worried that if we ship this we can't really change it later since it will break any existing scripts

app/views/works/_work_blurb.html.erb
```diff
@@ -1,5 +1,6 @@
 <% # expects "work" %>
 <li id="work_<%= work.id %>" class="<% if is_author_of?(work) %>own <% end %><% if work.unrevealed? %>mystery <% end %>work <%= css_classes_for_creation_blurb(work) %>" role="article">
+  <!-- updated_at=<%= work.updated_at.to_i.to_s %> -->
```
Bilka2: Could you move this into work_module? It should also be available in work blurbs in bookmark listings etc

features/works/work_browse.feature
```diff
@@ -178,3 +178,9 @@ Scenario: Can also browse work indexed by language
       Then I should see "1 Work in Deutsch"
     When I browse works in language "Persian"
       Then I should see "0 Works in Persian"
+
+Scenario: Work blurb includes an HTML comment containing the unix epoch of the updated time
+  Given it is currently 2025-04-12 17:00 UTC
```
Bilka2: You can freeze the time instead (`time is frozen at  ...`) so that you don't have to use `be_within` for the number. That way we can be sure that the test won't end up flaky if it ever ends up taking longer than 5 seconds for some reason

app/views/works/_work_module.html.erb
```diff
@@ -3,7 +3,7 @@
 <% if work.unrevealed? && !is_author_of?(work) %>
   <%= render "works/mystery_blurb", item: work %>
 <% else %>
-
+  <!-- updated_at=<%= work.updated_at.to_i.to_s %> -->
```
Bilka2: Could you move this into the header module div?

Bilka2 (changes requested):
Please add tests for your previous commit, e.g. for seeing it in readings or bookmarks.

Bilka2 (approved):
Thank you!

PR #5265, opened 2025-07-12
AO3-7032 Duplicate error message if email address is already part of invitation queue

riyakannapareddy:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-7032

## Purpose

This PR fixes the issue where the "is already part of our queue" error message shows up twice by including a
check in invite_request.rb to see if the raw email error message is already present and exiting the check to validate
the simplified email.

## Testing Instructions

The team can run the test "should not return error for duplicate simplified email if duplicate raw email error already present" in invite_request.rb for the duplicate email tests case. They can further validate this by recreating the steps in the bug.

## Credit

Riya K (She/Her)


Bilka2:
Welcome and thank you for contributing!

Someone will be along to review this pull request soon.

In the meantime, I've given your Jira account permission to comment on, assign, and transition issues in the future. While I was there, I also assigned this issue to you and set its status to In Review, so you don't have to do anything there. However, it would things a bit easier for us on the organizational side if the Full Name on your Jira account either closely matched the name you'd like us to credit in the release notes or included it in parentheses, e.g. "Nickname (CREDIT NAME)."

app/models/invite_request.rb
```diff
@@ -14,6 +14,9 @@ def set_simplified_email
 
   # Doing this with a method so the error message makes more sense
   def simplified_email_uniqueness
+    # Exit if raw email uniqueness error already exists
+    return if errors[:email].include?("is already part of our queue.")
```
sarken: Would `return if errors.of_kind? :email, :taken` work here instead? That way, we'd have one less thing to update if we change the text of the error message.

riyakannapareddy: That makes sense! I just pushed the changes to reflect that.

Bilka2 (approved):
Thank you!

PR #5266, opened 2025-07-13
AO3-7052 Current schema dump

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-7052


Bilka2:
Comments change is missing from structure.sql (it is present in the other one)

brianjaustin:
Lovely, it looks like something is off between the secondaries (both regions) and primaries... I copied a new version of the structure from one of the new database servers, so there are a few more changes, but it has the important stuff now

Bilka2:
> All other currently open PRs with migrations will need the date of their migration updated to be after the date in the schema dump file

I have commented on the relevant PRs.

PR #5272, opened 2025-07-14
AO3-6542 Allow admins to view users' preferences

EchoEkhi:
# Pull Request Checklist

# TODOs

* [x] RSpecs
* [x] i18n

## Issue

https://otwarchive.atlassian.net/browse/AO3-6542

## Purpose

Allow superadmins, PAC and support admins to view users' preferences

Allow superadmins and PAC admins to change users' email visibility

## Credit

EchoEkhi

### DIFF
--- a/app/views/preferences/index.html.erb
+++ b/app/views/preferences/index.html.erb
@@ -74,10 +74,8 @@
       <dd><%= f.time_zone_select :time_zone, nil, default: Time.zone.name %></dd>
       <% if $rollout.active?(:set_locale_preference, @user) %>
         <dt><%= f.label :preferred_locale, t(".your_locale") %> <%= link_to_help "locale-preferences" %></dt>
-        <dd>
-          <%= f.select :preferred_locale, locale_options_for_select(@available_locales, "id"),
-                { default: @preference.preferred_locale } %>
-        </dd>
+        <dd><%= f.select :preferred_locale, locale_options_for_select(@available_locales, "id"),
+                  default: @preference.preferred_locale %></dd>
       <% end %>
       <dt><%= f.label :work_title_format, t(".browser_page_title_format") %> <%= link_to_help "work_title_format" %></dt>
       <dd><%= f.text_field :work_title_format %></dd>
--- a/features/support/paths.rb
+++ b/features/support/paths.rb
@@ -151,8 +151,6 @@ module NavigationHelpers
       user_subscriptions_path(user_id: $1)
     when /^(.*?)(?:'s)? profile page$/i
       user_profile_path(user_id: $1)
-    when /^(.*?)(?:'s)? profile edit page$/i
-      edit_user_path(id: Regexp.last_match(1))
     when /^(.*)'s skins page/
       user_skins_path(user_id: $1)
     when /^"(.*)" skin page/
--- a/spec/controllers/preferences_controller_spec.rb
+++ b/spec/controllers/preferences_controller_spec.rb
@@ -51,7 +51,7 @@ describe PreferencesController do
     end
   end
 
-  describe "POST #update" do
+  describe "PUT #update" do
     let(:skin) { create(:skin, :public) }
 
     let(:preference_params) do
--- a/spec/controllers/profile_controller_spec.rb
+++ b/spec/controllers/profile_controller_spec.rb
@@ -28,54 +28,22 @@ describe ProfileController do
     let(:user) { create(:user) }
 
     context "as admin" do
-      before { fake_login_admin(admin) }
+      subject { get :edit, params: { user_id: user.login } }
 
-      read_roles = %w[superadmin policy_and_abuse]
-
-      context "with no role" do
-        let(:admin) { create(:admin, roles: []) }
-        
-        it "redirects with an error" do
-          get :edit, params: { user_id: user.login }
-
-          it_redirects_to_with_error(root_url, "Sorry, only an authorized admin can access the page you were trying to reach.")
-        end
+      let(:success) do
+        expect(assigns(:user)).to eq(user)
+        expect(response).to be_successful    
       end
 
-      (Admin::VALID_ROLES - read_roles).each do |role|
-        context "with role #{role}" do
-          let(:admin) { create(:admin, roles: [role]) }
-
-          it "redirects with an error" do
-            get :edit, params: { user_id: user.login }
-
-            it_redirects_to_with_error(root_url, "Sorry, only an authorized admin can access the page you were trying to reach.")
-          end
-        end
-      end
-
-      read_roles.each do |role|
-        context "with role #{role}" do
-          let(:admin) { create(:admin, roles: [role]) }
-
-          it "assigns user and profile" do
-            get :edit, params: { user_id: user.login }
-          
-            expect(assigns(:user)).to eq(user)
-            expect(response).to be_successful            
-          end
-        end
-      end
+      it_behaves_like "an action only authorized admins can access", authorized_roles: %w[superadmin policy_and_abuse]
     end
   end
 
-  describe "POST #update" do
+  describe "PATCH #update" do
     let(:user) { create(:user) }
 
     context "as admin" do
       before do 
-        fake_login_admin(admin)
-
         ticket = {
           "departmentId" => ArchiveConfig.ABUSE_ZOHO_DEPARTMENT_ID,
           "status" => "Open",
@@ -83,8 +51,6 @@ describe ProfileController do
         }
         allow_any_instance_of(ZohoResourceClient).to receive(:find_ticket).and_return(ticket)
       end
-
-      update_roles = %w[superadmin policy_and_abuse]
 
       let(:params) do
         {
@@ -97,39 +63,13 @@ describe ProfileController do
         }
       end
 
-      context "with no role" do
-        let(:admin) { create(:admin, roles: []) }
-        
-        it "redirects with an error" do
-          patch :update, params: params
+      subject { patch :update, params: params }
 
-          it_redirects_to_with_error(root_url, "Sorry, only an authorized admin can access the page you were trying to reach.")
-        end
+      let(:success) do
+        it_redirects_to_with_notice(user_profile_path(user), "Your profile has been successfully updated")
       end
 
-      (Admin::VALID_ROLES - update_roles).each do |role|
-        context "with role #{role}" do
-          let(:admin) { create(:admin, roles: [role]) }
-
-          it "redirects with an error" do
-            patch :update, params: params
-
-            it_redirects_to_with_error(root_url, "Sorry, only an authorized admin can access the page you were trying to reach.")
-          end
-        end
-      end
-
-      update_roles.each do |role|
-        context "with role #{role}" do
-          let(:admin) { create(:admin, roles: [role]) }
-
-          it "redirects with a success message" do
-            patch :update, params: params
-          
-            it_redirects_to_with_notice(user_profile_path(user), "Your profile has been successfully updated")            
-          end
-        end
-      end
+      it_behaves_like "an action only authorized admins can access", authorized_roles: %w[superadmin policy_and_abuse]
     end
   end
 end


app/controllers/preferences_controller.rb
```diff
@@ -12,27 +12,45 @@ def load_user
   def index
     @user = User.find_by(login: params[:user_id])
```
weeklies: You can remove this line for `index` and `update` as `load_user` is already called.

app/controllers/preferences_controller.rb
```diff
@@ -12,27 +12,45 @@ def load_user
...
 
+    if logged_in_as_admin?
+      authorize @preference
+
+      return render action: "index" unless @preference.update(permitted_attributes(@preference))
+      
+      if @preference.ticket_url.present?
+        link = view_context.link_to("Ticket ##{@preference.ticket_number}", @preference.ticket_url)
+        summary = @preference.email_visible ? "Enable" : "Disable"
+        summary += " \"Show my email address to other people\" for #{link}"
```
weeklies: You will need to update [admin_activity_summary](https://github.com/otwcode/otwarchive/blob/master/app/helpers/admin_helper.rb#L21) for this link to work.

app/policies/profile_policy.rb
```diff
@@ -1,10 +1,18 @@
...
 
+  def can_read_profile?
+    user_has_roles?(READ_ROLES)
+  end
+
   def can_edit_profile?
     user_has_roles?(EDIT_ROLES)
   end
 
+  alias edit? can_read_profile?
```
weeklies: This seems like an incorrect alias.

app/views/preferences/index.html.erb
```diff
@@ -76,41 +76,43 @@
...
       <dt><%= f.label :time_zone, t(".your_time_zone") %></dt>
-      <dd><%= f.time_zone_select :time_zone, nil, default: Time.zone.name %></dd>
+      <dd><%= f.time_zone_select(:time_zone, nil, { default: Time.zone.name }, disabled: logged_in_as_admin?) %></dd>
       <% if $rollout.active?(:set_locale_preference, @user) %>
         <dt><%= f.label :preferred_locale, t(".your_locale") %> <%= link_to_help "locale-preferences" %></dt>
-        <dd><%= f.select :preferred_locale, locale_options_for_select(@available_locales, "id"),
-                  default: @preference.preferred_locale %></dd>
+        <dd>
+          <%= f.select :preferred_locale, locale_options_for_select(@available_locales, "id"),
+            { default: @preference.preferred_locale }, disabled: logged_in_as_admin? %>
```
weeklies:
```suggestion
                { default: @preference.preferred_locale },
                disabled: logged_in_as_admin? %>
```

app/views/preferences/index.html.erb
```diff
@@ -146,19 +148,31 @@
...
+  <% if policy(@preference).update? %>
+    <fieldset>
+      <dt><%= f.label :ticket_number, class: "required" %></dt>
+      <dd>
+        <%= f.text_field :ticket_number, class: "required" %>
+      </dd>
+    </fieldset>
+  <% end %>
+
+  <% if !logged_in_as_admin? or policy(@preference).update? %>
```
weeklies: For boolean logic, it is recommended to use || [instead](https://rubystyle.guide/#and-or-flow).

weeklies (commented):
I've left some suggestions. However, with the merging of https://github.com/otwcode/otwarchive/pull/5293, the Jira issue needs to be updated before this PR can be updated as the sole editable preference has since been removed.

sarken:
Issue has been updated! Basically, all three roles should be able to access the Preferences page but not change anything now.

app/policies/profile_policy.rb
EchoEkhi: I think it's correct, `edit` is the controller method to view the edit form page, and update is the method for the form's POST request

weeklies: `policy(@user.profile).can_edit_profile?` can be easily confused with `policy(@user.profile).edit?`, especially for newer contributors. In this case, adding the alias may be unnecessary at this point in time.

app/controllers/users_controller.rb
```diff
@@ -400,16 +398,14 @@ def destroy_author
...
+      flash[:notice] = ts("You have successfully deleted your account.")
       redirect_to(delete_confirmation_path)
     else
-      flash[:error] = ts('Sorry, something went wrong! Please try again.')
+      flash[:error] = ts("Sorry, something went wrong! Please try again.")
       redirect_to(@user)
     end
   end
 
-  private
```
weeklies: Keeping the params private is important for encapsulation. I'm not sure if this file needs to be modified at all for this issue.

app/views/users/edit.html.erb
```diff
@@ -19,28 +19,30 @@
...
         <dt><%= p.label :ticket_number, class: "required" %></dt>
         <dd>
-          <%= p.text_field :ticket_number, class: "required" %>
+          <%= p.text_field :ticket_number, class: "required", disabled: logged_in_as_admin? && !policy(@user.profile).update? %>
         </dd>
       <% end %>
 
-      <dt class="landmark"><%= p.label :update, t(".update") %></dt>
-      <dd class="submit actions"><%= f.submit t(".update") %></dd>
+      <% if !logged_in_as_admin? or policy(@user.profile).update? %>
```
weeklies: For boolean logic, it is recommended to use || [instead](https://rubystyle.guide/#and-or-flow).

app/views/users/edit.html.erb
```diff
@@ -19,28 +19,30 @@
...
+        <%= p.text_area :about_me, class: "observe_textlength", disabled: logged_in_as_admin? && !policy(@user.profile).update? %>
         <%= live_validation_for_field("profile_attributes_about_me", presence: false, maximum_length: Profile::ABOUT_ME_MAX) %>
         <%= generate_countdown_html("profile_attributes_about_me", Profile::ABOUT_ME_MAX) %>
       </dd>
 
       <% if policy(@user.profile).can_edit_profile? %>
         <dt><%= p.label :ticket_number, class: "required" %></dt>
         <dd>
-          <%= p.text_field :ticket_number, class: "required" %>
+          <%= p.text_field :ticket_number, class: "required", disabled: logged_in_as_admin? && !policy(@user.profile).update? %>
```
weeklies: As only admins that can edit profiles will have this field displayed, adding this logic is unnecessary.

app/views/users/edit.html.erb
```diff
@@ -19,28 +19,30 @@
...
       <dt><%= p.label :about_me, t(".about_me") %></dt>
       <dd class="description">
         <p><%= allowed_html_instructions %></p>
-        <%= p.text_area :about_me, class: "observe_textlength" %>
+        <%= p.text_area :about_me, class: "observe_textlength", disabled: logged_in_as_admin? && !policy(@user.profile).update? %>
         <%= live_validation_for_field("profile_attributes_about_me", presence: false, maximum_length: Profile::ABOUT_ME_MAX) %>
         <%= generate_countdown_html("profile_attributes_about_me", Profile::ABOUT_ME_MAX) %>
       </dd>
 
       <% if policy(@user.profile).can_edit_profile? %>
```
weeklies: As this is a small file, can you change this to `policy(@user.profile).update?` for consistency with your additions?

spec/controllers/users_controller_spec.rb
```diff
@@ -338,4 +338,113 @@
...
+          
+            expect(assigns(:user)).to eq(user)
+            expect(response).to be_successful            
+          end
+        end
+      end
+    end
+  end
+
+  describe "PATCH #update" do
```
weeklies: In practice, the method used is POST.

spec/controllers/users_controller_spec.rb
```diff
@@ -338,4 +338,113 @@
       end
     end
   end
+
+  describe "GET #edit (edit profile)" do
```
weeklies: How about moving both of the following specs to `profile_controller_spec.rb`?

features/other_a/preferences_edit.feature
```diff
@@ -335,3 +335,104 @@ Feature: Edit preferences
...
+      | communications             |
+      | development_and_membership |
+      | docs                       |
+      | elections                  |
+      | legal                      |
+      | translation                |
+      | tag_wrangling              |
+      | open_doors                 |
+  
+  Scenario Outline: Admins can navigate to some pages but not others
```
weeklies: While useful, I don't think the scenarios below belong in `preferences_edit.feature`.

spec/controllers/preferences_controller_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+            expect(assigns(:user)).to eq(user)
+            expect(assigns(:preference)).to eq(user.preference)
+            expect(response).to be_successful            
+          end
+        end
+      end
+    end
+  end
+
+  describe "PUT #update" do
```
weeklies, line 54: In practice, the method used is `POST`.

spec/controllers/users_controller_spec.rb
EchoEkhi: I did think about doing that at first, but `profile_controller_spec.rb` corresponds to `profile_controller.rb` and the method I want to test is in `users_controller.rb`

features/other_a/preferences_edit.feature
EchoEkhi: It does test the buttons on the preferences page though. Where do you suggest I should move it to?

app/controllers/users_controller.rb
EchoEkhi: sorry that's my editor doing auto-formatting. I've disabled it

app/policies/profile_policy.rb
EchoEkhi: I think it is necessary - it controls access to the edit method in users_controller. I'll rename the other methods

spec/controllers/users_controller_spec.rb
weeklies: Thanks for the info, I've created an issue.

app/policies/profile_policy.rb
weeklies: I'm hesitant about this, so I'll leave this to someone else.

app/models/preference.rb
```diff
@@ -8,7 +10,7 @@ class Preference < ApplicationRecord
 
   validates :work_title_format,
             format: {
-              with: /\A[a-zA-Z0-9_\-,\. ]+\z/,
+              with: /\A[a-zA-Z0-9_\-,. ]+\z/,
```
sarken: This looks unrelated to the issue. Can you either remove it or provide testing instructions?

app/policies/preference_policy.rb
```diff
@@ -0,0 +1,9 @@
+class PreferencePolicy < ApplicationPolicy
+  READ_ROLES = %w[superadmin policy_and_abuse support].freeze
```
sarken: Let's follow the wrangling policy's example and go with the names `READ_ACCESS_ROLES` and `read_access?`

app/policies/profile_policy.rb
```diff
@@ -1,10 +1,18 @@
 class ProfilePolicy < ApplicationPolicy
+  # Roles that allow viewing a user's profile
+  READ_ROLES = %w[superadmin policy_and_abuse support].freeze
```
sarken: Again, let's follow the wrangling policy's example: `READ_ACCESS_ROLES`, `FULL_ACCESS_ROLES`, `read_access?`, `full_access?` 

I think we can also define `READ_ACCESS_ROLES` like so: `(FULL_ACCESS_ROLES + %w[support])`

app/policies/profile_policy.rb
sarken: I think it will be fine with the new `read_access?` name and a brief comment explaining it's necessary because the edit page is the only way to see preferences.

spec/controllers/preferences_controller_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+      before { fake_login_known_user(user) }
+
+      it "assigns user and preference" do
+        get :index, params: { user_id: user.login }
+        expect(assigns(:user)).to eq(user)
+        expect(assigns(:preference)).to eq(user.preference)
+        expect(response).to be_successful
+      end
+
+      it "disallows viewing others' preferences" do
```
sarken:
```suggestion
      it "disallows accessing others' preferences" do
```

spec/controllers/preferences_controller_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+        end
+      end
+    end
+  end
+
+  describe "POST #update" do
+    context "as regular user" do
+      before { fake_login_known_user(user) }
+
+      let(:skin) { create(:work_skin, :public) }
```
sarken: Can we use a site skin here, please?

I think it would also be a good idea to also cover a user updating someone else's preferences and a logged-out user trying to update someone's preferences.

app/models/preference.rb
```diff
@@ -1,4 +1,6 @@
 class Preference < ApplicationRecord
+  include Justifiable
```
sarken: Admins can no longer edit preferences, so this should no longer be necessary.

app/controllers/preferences_controller.rb
```diff
@@ -1,6 +1,6 @@
 class PreferencesController < ApplicationController
   before_action :load_user
-  before_action :check_ownership
+  before_action :check_ownership_or_admin
```
sarken, line 3: This is going to apply to both the `index` and `update` action, but we no longer want admins to be able to access `update`.

app/views/preferences/index.html.erb
```diff
@@ -24,15 +24,15 @@
     <h4 class="heading"><%= t(".privacy.heading") %> <%= link_to_help "privacy-preferences" %></h4>
     <ul>
       <li>
-        <%= f.check_box :minimize_search_engines %>
+        <%= f.check_box :minimize_search_engines, disabled: logged_in_as_admin? %>
```
sarken: It's fine to leave these enabled. If doing so causes a problem, `readonly` would be preferable -- it's much easier to see what is checked, and it will better resemble what the user sees if Support or PAC need to share a screenshot with the user.

features/other_a/preferences_edit.feature
sarken: I don't think we need to test these buttons specifically, just access to the pages, which doesn't change based on what link you clicked on to get there. 

We already have tests covering some, if not all, of these (e.g., access to the muted users page is covered in features/users/muting.feature and spec/controllers/muted/user_controller_spec.rb). If there are pages that don't have tests, specs will be faster than feature tests.

spec/controllers/preferences_controller_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+      end
+
+      it "disallows viewing others' preferences" do
+        get :index, params: { user_id: other_user.login }
+
+        it_redirects_to_with_error(user_path(other_user), "Sorry, you don't have permission to access the page you were trying to reach.")
+      end
+    end
+
+    context "as admin" do
```
sarken, line 38: I think we'd prefer if you could use the shared examples: 
https://github.com/otwcode/otwarchive/blob/48abf3174c5d2c5caf0e7ca9419f7674081eab12/spec/support/shared_examples/access_shared_examples.rb#L1-L34

app/views/preferences/index.html.erb
```diff
@@ -138,19 +141,22 @@
...
       <li>
-        <%= f.check_box :banner_seen %>
+        <%= f.check_box :banner_seen, disabled: logged_in_as_admin? %>
         <%= f.label :banner_seen, t(".misc.turn_off_banner_every_page") %>
       </li>
     </ul>
   </fieldset>
-  <%= submit_fieldset(f) %>
+
+  <% if !logged_in_as_admin? %>
```
sarken: Please leave the submit button on the page.

app/controllers/preferences_controller.rb
```diff
@@ -10,29 +10,29 @@ def load_user
   end
 
   def index
-    @user = User.find_by(login: params[:user_id])
     @preference = @user.preference
-    @available_skins = (current_user.skins.site_skins + Skin.approved_skins.site_skins).uniq
+    authorize @preference if logged_in_as_admin?
+    @available_skins = (@user.skins.site_skins + Skin.approved_skins.site_skins).uniq
```
sarken, line 14: Good catch! It would be a good idea to add a test for this that shows the user's skins are listed even if an admin is the one accessing the page.

app/models/preference.rb
EchoEkhi: Sorry that was probably my code editor's format-on-save. I've disabled it now

app/controllers/preferences_controller.rb
```diff
@@ -1,6 +1,7 @@
 class PreferencesController < ApplicationController
   before_action :load_user
-  before_action :check_ownership
+  before_action :check_ownership, only: [:update]
+  before_action :check_ownership_or_admin, except: [:update]
```
sarken: While this _will_ keep admins from accessing the `update` action, the solution we actually want here is `authorize @preference if logged_in_as_admin?` in the `update` action. That will give us the standard admin authorization error from Pundit (which is slightly misleading in terms of text, but that's okay) rather than the "please log in" error from `check_ownership`.

app/views/users/edit.html.erb
```diff
@@ -19,28 +19,30 @@
     <%= fields_for :profile_attributes, @user.profile do |p| %>
       <dt><%= p.label :title, t(".title") %></dt>
       <dd>
-        <%= p.text_field :title, class: "observe_textlength" %>
+        <%= p.text_field :title, class: "observe_textlength", disabled: logged_in_as_admin? && !policy(@user.profile).update? %>
```
sarken: This view should also have the `disabled` attribute removed and the submit button restored.

app/controllers/users_controller.rb
Bilka2, line 1: Drive-by comment: Could you revert the changes in this file since it's just rubocop autocorrection and there are no functional changes, so it just adds noise?

app/views/preferences/index.html.erb
```diff
@@ -71,11 +71,13 @@
         <%= f.select :skin_id, (@available_skins.collect { |s| [s.title, s.id] }) %>
       </dd>
       <dt><%= f.label :time_zone, t(".your_time_zone") %></dt>
-      <dd><%= f.time_zone_select :time_zone, nil, default: Time.zone.name %></dd>
+      <dd><%= f.time_zone_select(:time_zone, nil, { default: Time.zone.name }) %></dd>
```
sarken: Unless the old style interferes with admins accessing this page, these style changes should probably also be reverted, given Bilka's previous comment.

app/views/profile/show.html.erb
```diff
@@ -35,7 +35,7 @@
     <li><%= link_to ts("Manage My Pseuds"), user_pseuds_path(@user) %></li>
     <li><%= link_to ts("Delete My Account"), @user, data: {confirm: ts('This will permanently delete your account and cannot be undone. Are you sure?')}, :method => :delete %></li>
   </ul>
-  <% elsif policy(@user.profile).can_edit_profile? %>
+  <% elsif policy(@user.profile).read_access? %>
```
sarken: I double checked with Support and this was just a mistake in the testing instructions, not a request for a change. Support should not be able to access the Edit Profile page/see the links to it. I'll update the testing instructions.

sarken: Actually, let me clarify: there should not be any changes to _this_ link. The issue is only concerned with the links on the Preferences page.

app/views/preferences/index.html.erb
Bilka2: Speaking of style, please also action the rubocop warning in the new code

sarken:
It looks like there are some real test failures going on here. At the very least, I think you need to merge with master and make sure you're using the right paths for edit profile now that #5339 has been merged.

features/support/paths.rb
```diff
@@ -151,6 +151,8 @@ def path_to(page_name)
       user_subscriptions_path(user_id: $1)
     when /^(.*?)(?:'s)? profile page$/i
       user_profile_path(user_id: $1)
+    when /^(.*?)(?:'s)? profile edit page$/i
+      edit_user_path(id: Regexp.last_match(1))
```
sarken: I think this should've been removed. This doesn't seem to be used in any of the tests you've added, and the path is now `edit_user_profile_path` anyway.

spec/controllers/profile_controller_spec.rb
```diff
@@ -20,4 +23,113 @@
       expect(user.profile).not_to be_nil
     end
   end
+
+  describe "GET #edit" do
```
sarken, line 27: I think I'm missing something. Why can't we use the shared example like the preferences controller? 
https://github.com/otwcode/otwarchive/blob/48abf3174c5d2c5caf0e7ca9419f7674081eab12/spec/support/shared_examples/access_shared_examples.rb#L1-L34

spec/controllers/preferences_controller_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+        expect(response).to be_successful     
+      end
+
+      read_roles = %w[superadmin policy_and_abuse support]
+
+      it_behaves_like "an action only authorized admins can access", authorized_roles: read_roles
+    end
+  end
+
+  describe "POST #update" do
```
sarken: I know weeklies say this is a `POST` request based on the method specified in the `form` element, but we should say `PUT` here. It's the method we're using in the tests. 

However, neither `POST` nor `PUT` is technically correct. If you check the logs, it's actually a `PATCH` request:
```
method=PATCH path=/users/USERNAME/preferences/444434 format=html controller=PreferencesController action=update status=302 allocations=54427 duration=139.79 view=0.00 db=89.21 location=http://localhost:3000/users/USERNAME
```

That's because there's a hidden `input` that specifies the method:

```html
<input type="hidden" name="_method" value="patch" autocomplete="off">
```

Rails uses that method rather than the one in the `form` tag. It was implemented as a workaround for spotty browser support for `PATCH`, `PUT`, and `DELETE` requests (so you won't see the hidden input on forms that are true `POST` requests, like the form for creating a new work). It only gets added if the resource is persisted. 

[Rails began preferring `PATCH` over `PUT` for persisted records in 2012](https://rubyonrails.org/2012/2/26/edge-rails-patch-is-the-new-primary-http-method-for-updates), but both routes are created and `PUT` is still supported. So it's fine to keep using `put` in the tests -- preferable, even, since it's what we usually use and it means making fewer changes to this PR -- but please update the description.

spec/controllers/profile_controller_spec.rb
```diff
@@ -20,4 +23,113 @@
...
+          
+            expect(assigns(:user)).to eq(user)
+            expect(response).to be_successful            
+          end
+        end
+      end
+    end
+  end
+
+  describe "POST #update" do
```
sarken: It looks like the tests are using the `PATCH` method, not `POST`, so please update this.

app/views/preferences/index.html.erb
```diff
@@ -74,8 +74,10 @@
       <dd><%= f.time_zone_select :time_zone, nil, default: Time.zone.name %></dd>
       <% if $rollout.active?(:set_locale_preference, @user) %>
         <dt><%= f.label :preferred_locale, t(".your_locale") %> <%= link_to_help "locale-preferences" %></dt>
-        <dd><%= f.select :preferred_locale, locale_options_for_select(@available_locales, "id"),
-                  default: @preference.preferred_locale %></dd>
+        <dd>
+          <%= f.select :preferred_locale, locale_options_for_select(@available_locales, "id"),
```
sarken: This looks like another style autocorrect that should've been reverted.

PR #5273, opened 2025-07-14
AO3-7057 Add error message display to admin invitation page

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-7057

## Purpose

Show validation errors when sending invitations directly to an email as admin.

## Credit

Bilka

### DIFF
--- a/features/admins/admin_invitations.feature
+++ b/features/admins/admin_invitations.feature
@@ -263,6 +263,15 @@ Feature: Admin Actions to Manage Invitations
     Then I should see "An invitation was sent to fred@bedrock.com"
       And 1 email should be delivered
 
+  Scenario: An admin can't create an invite with invalid email
+    Given I am logged in as an admin
+      And all emails have been delivered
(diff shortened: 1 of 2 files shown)


features/admins/admin_invitations.feature
```diff
@@ -263,6 +263,15 @@ Feature: Admin Actions to Manage Invitations
...
       And 1 email should be delivered
 
+  Scenario: An admin can't create an invite with invalid email
+    Given I am logged in as an admin
+      And all emails have been delivered
+    When I follow "Invite New Users"
+      And I fill in "invitation[invitee_email]" with "abcdefgh"
+      And I press "Invite user"
+    Then I should see "Invitee email should look like an email address. Please use a different address or leave blank."
+      And 0 email should be delivered
```
sarken: If you had a second to change this to `0 emails`, that would be great, but it's definitely not a blocker!

PR #5275, opened 2025-07-16
AO3-6817 Prerender works links on hover

domenic:
# Pull Request Checklist

* [ ] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)? Like [previous speculative loading PRs](https://github.com/otwcode/otwarchive/pulls?q=is%3Apr+author%3Adomenic+is%3Aclosed), we think it's OK not to have tests for these

## Issue

https://otwarchive.atlassian.net/browse/AO3-6817

## Purpose

Adds prerendering of any link to a URL starting with `/works/`, which will trigger on hover. This is addition to (and does not conflict with) the idle-time prerendering of the next chapter. This makes clicking around Ao3 pretty snappy in general!

## Testing Instructions

See https://otwarchive.atlassian.net/browse/AO3-6817.

## References

See https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/script/type/speculationrules for more on the feature being used here.

## Credit

Domenic Denicola, he/him

### DIFF
--- a/app/views/layouts/_javascripts.html.erb
+++ b/app/views/layouts/_javascripts.html.erb
@@ -111,4 +111,18 @@
   <% end %>
 <% end %>
 
+<script type="speculationrules">
+  {
+    "prerender": [
+      {
+        "source": "document",
+        "eagerness": "moderate",
+        "where": {
+          "href_matches": "/works/*"
+        }
+      }
+    ]
+  }
+</script>
+
 <%= yield :footer_js %>


domenic:
Back to draft while I investigate the test failures...

domenic:
It seems the new prerendering reproducibly screws up some tests.

```
docker compose run --rm test bundle exec cucumber features/works/work_create.feature:53
```

is one case. Without the new code, it works fine; with it, it breaks.

A distinguishing feature of this test is the `@javascript` annotation.

I believe it gets stuck on the step that starts a new session, which bottoms out at `page.driver.reset!`.

It's _possible_ this is due to some of the problems with WebDriver and prerendering that we've seen in the past. It's a bit surprising though, especially since these prerenders are hover-triggered, so I'm surprised they would cause this.

Will keep digging...

noahbravo:
I think this is a great direction for performance, but I have a few concerns about scope and safety with the current implementation:

- `source: "document" `+ `href_matches: "/works/*"` is broader than the issue goal ("work links in work blurbs").
- It may also include `/works/...` GET endpoints with side effects (e.g. mark-for-later/read), which could be risky for speculative loading.

Would you be open to a narrower approach?

- Triggering only from explicit work-blurb title links (e.g. via a `data-*` hook),
- Applying prerender on hover/focus intent,
- Keeping strict URL guards (same-origin + canonical `/works/:id`, no querystring).

This might stay closer to the ticket scope and help reduce the WebDriver/prerender interference seen in the JS tests.

PR #5276, opened 2025-07-16
AO3-7037 AO3-7061 Only permit email for password resets

EchoEkhi:
# Pull Request Checklist

# TODOs

- [x] Enumeration attack mitigation
- [x] Timing attack mitigation

## Issue

https://otwarchive.atlassian.net/browse/AO3-7061
https://otwarchive.atlassian.net/browse/AO3-7037

## Purpose

Remove the ability to request password resets with usernames

## Credit

EchoEkhi

### DIFF
--- a/features/users/authenticate_users.feature
+++ b/features/users/authenticate_users.feature
@@ -5,7 +5,7 @@ Feature: User Authentication
   Scenario: Forgot password
     Given I have no users
       And the following activated user exists
-      | email       | login | password |
+      | email           | login | password |
       | sam@example.com | sam   | secret   |
       And all emails have been delivered
     When I am on the home page
@@ -86,7 +86,7 @@ Feature: User Authentication
   Scenario: Users should not be able to request password resets with their username
     Given I have no users
       And the following activated user exists
-      | email       | login | password |
+      | email           | login | password |
       | sam@example.com | sam   | secret   |
       And all emails have been delivered
     When I request a password reset for "sam"
@@ -97,7 +97,7 @@ Feature: User Authentication
   Scenario: Attackers should see a fake success message when requesting password resets with a non-existant email
     Given I have no users
       And the following activated user exists
-      | email       | login | password |
+      | email           | login | password |
       | sam@example.com | sam   | secret   |
       And all emails have been delivered
     When I request a password reset for "1@example.com"
@@ -281,7 +281,7 @@ Feature: User Authentication
 
   Scenario: Login case (in)sensitivity
     Given the following activated user exists
-      | login      | password |
+      | login      | password  |
       | TheMadUser | password1 |
     When I am on the home page
       And I fill in "Username or email:" with "themaduser"


EchoEkhi:
Should I implement the timing attack mitigation in the app code or would it be easier if it's somewhere higher up in the stack?

brianjaustin:
What do you mean timing attack?

EchoEkhi:
Because the fake success message branch skips a lot of processing, the server answers faster than the case where the real success message is rendered. An attacker can observe the difference in response times and deduce whether the email exists

Bilka2:
We should be protecting against timing attacks on the application side by using [find_or_initialize_with_errors](https://github.com/heartcombo/devise/blob/cf93de390a29654620fdf7ac07b4794eb95171d0/lib/devise/models/authenticatable.rb#L277) instead of `find_for_authentication`, similar to how [devise does by it default](https://github.com/heartcombo/devise/blob/cf93de390a29654620fdf7ac07b4794eb95171d0/lib/devise/models/recoverable.rb#L123-L127). The method protects against the attack in much the same way as Rails own [authenticate_by](https://api.rubyonrails.org/classes/ActiveRecord/SecurePassword/ClassMethods.html#method-i-authenticate_by).

As for the timing attack protection on the middleware side, we're already rate-limiting the page.

Bilka2:
Also, could you please make all reset request failures as well as the reset request success redirect to the reset password page? That way we could integrate https://otwarchive.atlassian.net/issues/AO3-7037 into this.

zz9pzza:
I would add about timing the database response is about 1/4 of a millisecond and the application jitter is much bigger as we have a wide set of generations of application hardware ( p50 about 145ms )

app/controllers/users/passwords_controller.rb
```diff
@@ -6,23 +6,24 @@ class Users::PasswordsController < Devise::PasswordsController
   skip_before_action :store_location
   layout "session"
 
+  def new
+    @page_title = t(".reset_password")
```
Bilka2:
```suggestion
    @page_subtitle = t(".reset_password")
```

page_title removes the `| Archive of Our Own` at the end

app/controllers/users/passwords_controller.rb
```diff
@@ -31,20 +32,17 @@ def create
...
   def successfully_sent?(resource)
     return super if Devise.paranoid
     return unless resource.errors.empty?
 
-    flash[:notice] = t("users.passwords.create.send_instructions",
-                       send_times_remaining: t("users.passwords.create.send_times_remaining",
-                                               count: resource.password_resets_remaining),
-                       send_cooldown_period: t("users.passwords.create.send_cooldown_period",
-                                               count: ArchiveConfig.PASSWORD_RESET_COOLDOWN_HOURS))
+    flash[:notice] = t("users.passwords.create.send_instructions")
```
Bilka2: This method override should not be needed now that we no longer include the extra info because devise defaults to this locale key https://github.com/heartcombo/devise/blob/cf93de390a29654620fdf7ac07b4794eb95171d0/app/controllers/devise_controller.rb#L142

app/controllers/users/passwords_controller.rb
```diff
@@ -6,23 +6,24 @@ class Users::PasswordsController < Devise::PasswordsController
...
+    super
+  end
+
   def create
-    user = User.find_for_authentication(resource_params.permit(:login))
-    if user.nil? || user.new_record?
-      flash[:error] = t(".user_not_found")
+    user = User.find_or_initialize_with_errors([:email], resource_params, :not_found)
+
+    unless params[:user][:email].to_s.match?(/\A[^@]+@[^@]+\z/)
```
Bilka2: We have a regex in the email format validator that we could use here. Could you see if you extract that from the method there so it can be reused here?

app/views/users/passwords/new.html.erb
```diff
@@ -1,14 +1,15 @@
 <!--Descriptive page name, messages and instructions-->
-<h2 class="heading"><%= t(".page_heading") %></h2>
+<h2 class="heading"><%= t(".reset_password") %></h2>
```
Bilka2: Please keep the same locale key here, meaning a separate key from the button

spec/controllers/users/passwords_controller_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+  describe "create" do
+    let(:user) { create(:user) }
+
+    context "when resetting password with a correct email address" do
+      it "sends reset instructions and shows a success message" do
+        expect do
+          post :create, params: { user: { email: user.email } }
+        end.to change { 
+          ActionMailer::Base.deliveries.count
+        }.by(1)
```
Bilka2: Could you use rspecs email matchers like `send_email`?

app/controllers/users/passwords_controller.rb
Bilka2: Also, this should use `browser_title` as the locale key please (and be in the controller's yml file)

sarken: I think we do `page_title` usually -- there's ~15 of those and ~5 of `browser_title` in the controllers locale file.

features/users/authenticate_users.feature
```diff
@@ -5,8 +5,8 @@ Feature: User Authentication
   Scenario: Forgot password
     Given I have no users
       And the following activated user exists
-      | login    | password |
-      | sam      | secret   |
+      | email       | login | password |
+      | sam@otw.org | sam   | secret   |
```
sarken: Please use `@example.com` for example email addresses.

app/controllers/users/passwords_controller.rb
```diff
@@ -6,23 +6,31 @@ class Users::PasswordsController < Devise::PasswordsController
...
-    if user.nil? || user.new_record?
-      flash[:error] = t(".user_not_found")
-      redirect_to new_user_password_path and return
+    user = User.find_or_initialize_with_errors([:email], resource_params, :not_found)
+
+    email_regex ||= begin
+      email_name_regex = '[A-Z0-9_\.&%\+\-\']+'
+      domain_head_regex = "(?:[A-Z0-9\-]+\.)+"
+      domain_tld_regex = "(?:[A-Z]{2,25})"
+      /\A#{email_name_regex}@#{domain_head_regex}#{domain_tld_regex}\z/i
```
Bilka2: By extracting the regex from the validator I meant putting it into a reusable method/class/something, not copy-pasting it

app/controllers/users/passwords_controller.rb
```diff
@@ -6,23 +6,31 @@ class Users::PasswordsController < Devise::PasswordsController
   skip_before_action :store_location
   layout "session"
 
+  def new
+    @page_title = t(".browser_title")
```
Bilka2: Please do follow sarken's comment about using `.page_title` here

config/locales/views/en.yml
```diff
@@ -2263,10 +2263,11 @@ en:
       new_work: Post New
     passwords:
       new:
-        email_or_username_html: Email address %{or} username
-        instructions: If you've forgotten your password, we can send instructions that will allow you to reset it. Please tell us the username or email address you used when you signed up for your Archive account.
-        or: or
-        page_heading: Forgotten your password?
+        browser_title: Reset Password
```
Bilka2: As I said in my previous review, this should be in controllers/en.yml

app/controllers/users/passwords_controller.rb
```diff
@@ -31,22 +39,12 @@ def create
...
-    return unless resource.errors.empty?
-
-    flash[:notice] = t("users.passwords.create.send_instructions",
-                       send_times_remaining: t("users.passwords.create.send_times_remaining",
-                                               count: resource.password_resets_remaining),
-                       send_cooldown_period: t("users.passwords.create.send_cooldown_period",
-                                               count: ArchiveConfig.PASSWORD_RESET_COOLDOWN_HOURS))
+  def after_sending_reset_password_instructions_path_for(*)
+    new_user_password_path
   end
```
Bilka2: This should go inside `protected`

spec/controllers/users/passwords_controller_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  describe "create" do
+    let(:user) { create(:user) }
+
+    context "when resetting password with a correct email address" do
+      it "sends reset instructions and shows a success message" do
+        expect do        
+          post :create, params: { user: { email: user.email } }
+        end.to send_email(
+          from: "do-not-reply@example.org",
```
Bilka2: No need to check this

spec/controllers/users/passwords_controller_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+          to: user.email,
+          subject: "[AO3] Reset your password"
```
Bilka2: Please use `ArchiveConfig.APP_SHORT_NAME` instead of hardcoding AO3

config/locales/devise/en.yml
```diff
@@ -40,7 +40,7 @@ en:
       success: Successfully authenticated from %{kind} account.
     passwords:
       no_token: You can't access this page without coming from a password reset email. If you do come from a password reset email, please make sure you used the full URL provided.
-      send_instructions: Check your email for instructions on how to reset your password.
+      send_instructions: If the email address you entered is currently associated with an AO3 account, you should receive an email with instructions to reset your password.
```
Bilka2: We'd like to keep the message for admins as is. You can scope this by user by putting it into `passwords.user.send_instructions` (and adding another locale for admins), similar to how it's done for the failure messages at the top of this file

Bilka2 (changes requested):

Bilka2 (approved):
Looks good, thank you!

Bilka2:
This is now blocked by https://otwarchive.atlassian.net/browse/AO3-7112 and will need updates to remove " username or" from the email once the email is coded and merged into master.

Bilka2:
Please merge master and update the email text, thank you!

config/locales/devise/en.yml
```diff
@@ -39,11 +39,15 @@ en:
       failure: Could not authenticate you from %{kind} because "%{reason}".
       success: Successfully authenticated from %{kind} account.
     passwords:
+      admin:
+        send_instructions: Check your email for instructions on how to reset your password.
       no_token: This password reset link is invalid or incomplete. Please check your email for the correct link or request a new password reset.
       send_instructions: Check your email for instructions on how to reset your password.
```
Bilka2: This line should be removed

spec/controllers/users/passwords_controller_spec.rb
```diff
@@ -1,12 +1,49 @@
...
+    context "when resetting password with an incorrect email address" do
+      it "does not send reset instructions and shows a fake success message" do
+        expect do
+          post :create, params: { user: { email: "incorrect-email@example.com" } }
+        end.to_not send_email
+
+        it_redirects_to_with_notice(new_user_password_path, "If the email address you entered is currently associated with an AO3 account, you should receive an email with instructions to reset your password.")
+      end
+    end
+  end
```
Bilka2, line 46: Add a newline after this please

Bilka2 (approved):
Thank you!

PR #5278, opened 2025-07-18
AO3-7046 Migrate subscription table to bigint

connie-feng:
## Issue

https://otwarchive.atlassian.net/browse/AO3-7026

## Purpose

This PR migrates the subscription table IDs from INT to BIGINT.

## Credit

Connie Feng, she/her


db/migrate/20250718220533_change_subscriptions_id_to_bigint.rb
```diff
@@ -0,0 +1,11 @@
+class ChangeSubscriptionsIdToBigint < ActiveRecord::Migration[7.1]
+  uses_departure! if Rails.env.staging? || Rails.env.production?
+
+  def up
+    change_column :subscriptions, :id, "bigint NOT NULL AUTO_INCREMENT"
+  end
+
+  def down
+    change_column :subscriptions, :id, "int NOT NULL AUTO_INCREMENT"
```
Bilka2: Please remove this and replace it with `raise ActiveRecord::IrreversibleMigration` similar to how `20250515095633_change_user_agent_limit.rb` does it. We'd like this since we can't safely migrate to a smaller id (once there records with a `> int` id)

Bilka2 (changes requested):
Could you add an rspec test that creates a subscription with a very high id to make sure it's possible? If you'd like an example, https://github.com/otwcode/otwarchive/blob/0d71fcde8b2fa5bde02abf593ac685a5edb5fc51/spec/models/bookmark_spec.rb#L25 is similar

PR #5199, opened 2025-05-19
AO3-6937 Fix inbox page title to show username

kitbur:
# Pull Request Checklist

* [X] Have you read ["How to write the perfect pull request"](https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/)?
* [X] Have you read the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [X] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)?
* [X] Have you added the [Jira](https://otwarchive.atlassian.net) issue number
  as the *first* thing in your pull request title (e.g. `AO3-1234 Fix thing`)
* [X] Do you have fewer than 5 pull requests already open? If not, please wait
  until they are reviewed and merged before creating new pull requests.

## Issue

https://otwarchive.atlassian.net/browse/AO3-6937

## Purpose

Fix the browser tab title on the user's inbox page to: `user - Inbox | Archive of Our Own`

## Testing Instructions

1. Visit a user’s inbox page
2. Confirm the page title in the browser tab displays as `user - Inbox | Archive of Our Own`
3. Confirm that other page titles are not affected

## Credit

kitbur

PR #5280, opened 2025-07-21
AO3-6580 Sets confirmation page title to "Account Created"

anmazz:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6580

## Purpose

Sets confirmation page title to "Account Created" from "Create Registration"

## Testing Instructions

1. With signups open, create a new account
2. On the "Almost Done!" page, confirm tab title displays "Account Created"

## Credit

anna (she/her)

### DIFF
--- a/app/controllers/users/registrations_controller.rb
+++ b/app/controllers/users/registrations_controller.rb
@@ -3,7 +3,7 @@ class Users::RegistrationsController < Devise::RegistrationsController
   before_action :configure_permitted_parameters
 
   def new
-    @page_subtitle = t("registrations.page_title") # Displays "New Registration" otherwise
+    @page_subtitle = t(".page_title")
     super do |resource|
       if params[:invitation_token]
         @invitation = Invitation.find_by(token: params[:invitation_token])
@@ -16,7 +16,7 @@ class Users::RegistrationsController < Devise::RegistrationsController
   end
 
   def create
-    @page_subtitle = t("registrations.account_created_page_title")
+    @page_subtitle = t(".page_title")
     @hide_dashboard = true
     build_resource(sign_up_params)
 
--- a/features/users/user_create.feature
+++ b/features/users/user_create.feature
@@ -77,10 +77,10 @@ Feature: Sign Up for a new account
 
   Scenario: The user should be able to create a new account with a valid email and password
     When I fill in the sign up form with valid data
-    Then I should see the page title "Create Account | Example Archive"
-      And all emails have been delivered
+    Then I should see the page title "Create Account"
+    When all emails have been delivered
       And I press "Create Account"
-    Then I should see the page title "Account Created | Example Archive"
-    Then I should see "Almost Done!"
+    Then I should see the page title "Account Created"
+      And I should see "Almost Done!"
       And I should get a new user activation email
       And a new user account should exist
--- a/config/locales/controllers/en.yml
+++ b/config/locales/controllers/en.yml
@@ -186,9 +186,6 @@ en:
     clear:
       error: There were problems deleting your history. Please try again later.
       success: Your history is now cleared.
-  registrations:
-    account_created_page_title: Account Created
-    page_title: Create Account
   related_works:
     index:
       page_title: "%{login} - Related Works"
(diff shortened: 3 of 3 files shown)


sarken:
Hi, anna!

I've updated the Jira issue status to In Review, so no one will mistakenly create a duplicate pull request. If you'd like the ability to comment on, assign, and transition issues in the future, you're welcome to create a Jira account! You can just reply here with the account name and we'll set up the permissions for you. (It makes things a bit easier for us on the organizational side if the Full Name on your Jira account either closely matches the name you'd like us to credit in the release notes or includes it in parentheses, e.g. "Nickname (CREDIT NAME).")

anmazz:
> If you'd like the ability to comment on, assign, and transition issues in the future, you're welcome to create a Jira account! You can just reply here with the account name and we'll set up the permissions for you.

hi @sarken! I’ve created a jira account, the full name is anna. let me know if you’re able to see it!

sarken:
I actually can't find you, but Jira is like that sometimes! The fix is usually to send an invitation to the user's email address, which forces the admin panel to update.

 Would I be correct to assume your Jira account uses the same address you used to contacted otw-coders the other day? (We'll get back to you soon about that email, btw! We're just trying to get consensus on a design.) If that's not the right address, could you send us another email from the address associated with your Jira account?

app/controllers/users/registrations_controller.rb
```diff
@@ -16,6 +16,7 @@ def new
   end
 
   def create
+    @page_title = "Account Created"
```
sarken: We actually want to use `@page_subtitle` here, which will change the current `Create Registration` part of the page title while leaving the ` | Archive of Our Own` part. The issue was a little unclear about that, and I realize it gets even more confusing because the `new` action above uses the wrong one  -- my apologies for that! I'll get the issue updated.

If you could make the text translatable and add a test as well, we'd appreciate it. We have wiki pages with [our internationalization standards](https://github.com/otwcode/otwarchive/wiki/Internationalization-%28i18n%29-Standards) and [information on automated tests](https://github.com/otwcode/otwarchive/wiki/Automated-Testing), but you can also check out how we handled i18n and tests for a similar issue in pull request #5199.

anmazz:
> I actually can't find you, but Jira is like that sometimes! The fix is usually to send an invitation to the user's email address, which forces the admin panel to update.
> 
> Would I be correct to assume your Jira account uses the same address you used to contacted otw-coders the other day? (We'll get back to you soon about that email, btw! We're just trying to get consensus on a design.) If that's not the right address, could you send us another email from the address associated with your Jira account?

Yep same email!

app/controllers/users/registrations_controller.rb
anmazz: @sarken updated! I was trying to follow the linked PR. Not sure if the added test lines are good (just added to the user_create, didn't add a separate test in page_title).

app/controllers/users/registrations_controller.rb
```diff
@@ -3,7 +3,7 @@ class Users::RegistrationsController < Devise::RegistrationsController
   before_action :configure_permitted_parameters
 
   def new
-    @page_title = "Create Account" # Displays "New Registration" otherwise
+    @page_subtitle = t("registrations.page_title") # Displays "New Registration" otherwise
```
sarken: We can take advantage of Rails' [lazy lookup](https://guides.rubyonrails.org/i18n.html#lazy-lookup) here and just do `t(".page_title")` for both page subtitles. Then, in the locale file, we can put the following under `users:`:

```
    registrations:
      create:
        page_title: Account Created
      new:
        page_title: Create Account
```

You can also delete the existing code comment on this line -- it was probably useful in the original pull request, but it's not anymore.

features/users/user_create.feature
```diff
@@ -77,8 +77,10 @@ Feature: Sign Up for a new account
 
   Scenario: The user should be able to create a new account with a valid email and password
     When I fill in the sign up form with valid data
+    Then I should see the page title "Create Account | Example Archive"
```
sarken: We prefer to leave off the " | Example Archive" part for the tests so people who've changed the name of their dev environment don't run into test failures.

features/users/user_create.feature
```diff
@@ -77,8 +77,10 @@ Feature: Sign Up for a new account
...
       And all emails have been delivered
```
sarken: Since this `And` step was under a `When` until we added the new `Then` step, it should get changed to a `When`.

features/users/user_create.feature
```diff
@@ -77,8 +77,10 @@ Feature: Sign Up for a new account
...
       And I press "Create Account"
+    Then I should see the page title "Account Created | Example Archive"
     Then I should see "Almost Done!"
```
sarken: This line should get changed to `And` and indented now

sarken (commented):
Your Jira account should be all set now! 

If you'd like to try out your new permissions, you can comment on the issue and mention that testers should also verify that the browser page title on the Create Account page now includes the site name

app/controllers/users/registrations_controller.rb
anmazz: Ah ok, I see the lazy loading. Pretty neat

anmazz:
@sarken is it possible to force a rerun of these tests?

sarken (approved):
I reran the test (only devs with merge access can do that) and everything looks good now -- thank you so much!

PR #5285, opened 2025-07-24
AO3-6818 Check invite request for banned email

anmazz:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6818

## Purpose

Blocks an email on the banned list from requesting an invite. Removes email from invite queue once email has been banned (after commit).

Should this functionality also be extended to sign-ups?

## References

Added question in Jira about potentially doing a migration if needing to clear out existing list. Was also trying to think through if there would be any reason to do anything if the user checks the email's place in the queue.

## Credit

anna (she/her)

### DIFF
--- a/features/admins/admin_email_blacklist.feature
+++ b/features/admins/admin_email_blacklist.feature
@@ -83,11 +83,11 @@ Scenario: User's email banned after joining invite queue should remove their ema
   Given I am a visitor
   When I am on the homepage
     And I follow "Get an Invitation"
-  When I fill in "Email" with "foo+bar@baz.com"
+    And I fill in "Email" with "foo+bar@baz.com"
     And I press "Add me to the list"
   Then I should see "You've been added to our queue"
-  Given I am logged in as a "policy_and_abuse" admin
-  When I go to the manage invite queue page
+  When I am logged in as a "policy_and_abuse" admin
+    And I go to the manage invite queue page
   Then I should see "foo+bar@baz.com"
     And I should not see "foo@baz.com"
   When I have banned the address "foo+qux@baz.com"


Bilka2:
> Should this functionality also be extended to sign-ups?

Yes but it's low priority to make a design, so it should not be part of this PR.

app/models/admin_blacklisted_email.rb
```diff
@@ -1,12 +1,17 @@
 class AdminBlacklistedEmail < ApplicationRecord
   before_validation :canonicalize_email
+  after_create_commit :remove_invite_requests
```
Bilka2: I think this should instead be `after_create` because removing the email from the invite requests is a db operation, so it should be part of the general create object db transaction. `after_commit` is more for things that aren't part of the db, like sending an email or updating elasticsearch, as they can't be rolled back if they are done with `after_create` (https://guides.rubyonrails.org/active_record_callbacks.html#after-commit-and-after-rollback)

app/models/admin_blacklisted_email.rb
```diff
@@ -1,12 +1,17 @@
...
 
   validates :email, presence: true, uniqueness: { case_sensitive: false }, email_format: true
 
   def canonicalize_email
     self.email = AdminBlacklistedEmail.canonical_email(self.email) if self.email
   end
 
+  def remove_invite_requests
+    InviteRequest.where(email: self.email).destroy_all
```
Bilka2: This should instead find the invite request by the `simplified_email`.

`email` is canonized/simplified here before this callback, so blocking something like `foo+bar@example.com` turns email into `foo@example.com`. InviteRequest on the other hand does that [simplification only on `simplified_email`](https://github.com/otwcode/otwarchive/blob/7e4fdcbcaa1d874b902320c60e9d864198164b70/app/models/invite_request.rb#L9) and leaves `email` the same. So, if you requested an invite for `foo+bar@example.com`, that would be set as the `email` for InviteRequest, so creating a block of `foo+bar@example.com`, canonized to `foo@example.com`, wouldn't find that request to destroy here. (It'd probably be good to have a test for this.)

app/models/invite_request.rb
```diff
@@ -30,15 +31,24 @@ def position
     InviteRequest.where("id <= ?", id).count
   end
 
-  #Ensure that invite request is for a new user
+  # Ensure that email is not banned
+  def check_admin_banned_list
+    return unless AdminBlacklistedEmail.exists?(email: self.email)
```
Bilka2: `is_blacklisted?` would be better here to match the email canonization/simplification that the banlist does.  (It'd probably be good to have a test for the blocking of the variants. The general variants that should work are the [same as for guest comments](https://otwarchive.atlassian.net/browse/AO3-4536).)

app/models/invite_request.rb
```diff
@@ -30,15 +31,24 @@ def position
...
+
+    errors.add(:email, "has been blocked at the owner's request. That means it can't be used for invitations.
```
Bilka2: Could you i18n this by using a symbol and putting the string into the locale file? There's some docs here: https://github.com/otwcode/otwarchive/wiki/Internationalization-%28i18n%29-Standards#user-content-errorsadd

app/models/invite_request.rb
```diff
@@ -1,8 +1,9 @@
 class InviteRequest < ApplicationRecord
   validates :email, presence: true, email_format: true
   validates :email, uniqueness: { message: "is already part of our queue." }
-  before_validation :compare_with_users, on: :create
   before_validation :set_simplified_email, on: :create
+  before_validation :check_admin_banned_list, on: :create
```
Bilka2: It'd be better to make this a `validate` (I guess you copied compare_with_users, but that shouldn't be `before_validation` either...)

features/admins/admin_email_blacklist.feature
```diff
@@ -59,3 +59,30 @@ Scenario: Banning a user's email should not affect their ability to post comment
...
+
+Scenario: Banning a user's email should prevent requesting invites
+  Given I am logged in as a "policy_and_abuse" admin
+  When I have banned the address "foo@bar.com"
+    And I log out
+    And account creation requires an invitation
+    And the invitation queue is enabled
+  When I am on the homepage
+    And I follow "Get an Invitation"
+  When I fill in "invite_request_email" with "foo@bar.com"
```
Bilka2: Could you make this into an `And` step since there is directly after another `When`? Also, we prefer to use the user-facing field names ("Email" in this case) over the input names (invite_request_email). Same thing for the other test

Bilka2 (changes requested):

features/admins/admin_email_blacklist.feature
anmazz: Yep. I just frankensteined this one from the invite_queue.feature file

app/models/admin_blacklisted_email.rb
anmazz: I originally used after_create_commit just to be (probably too) safe in case future side effects were added, but I'm fine switching to after_create if that’s preferred. I ran a rollback test and didn’t see any difference in behavior between the two.

anmazz:
The `features/importing/work_import_lj.feature` has failed twice now, I'm actually not sure if anything I've done has affected that or if it's being flaky

features/admins/admin_email_blacklist.feature
```diff
@@ -59,3 +59,36 @@ Scenario: Banning a user's email should not affect their ability to post comment
...
+  Then I should not be able to add the email "foo@gmail.com" to the invite queue
+    And I should not be able to add the email "fo.o@gmail.com" to the invite queue
+    And I should not be able to add the email "foo+baz@gmail.com" to the invite queue
+    And I should not be able to add the email "foo@googlemail.com" to the invite queue
+
+Scenario: User's email banned after joining invite queue should remove their email
+  Given I am a visitor
+  When I am on the homepage
+    And I follow "Get an Invitation"
+  When I fill in "invite_request_email" with "foo@bar.com"
```
Bilka2:
```suggestion
  When I fill in "invite_request_email" with "foo+bar@baz.com"
```
To test the variant destroying (needs further changes in the test down below)

app/models/invite_request.rb
```diff
@@ -30,15 +32,23 @@ def position
...
+
+  # Ensure that invite request is for a new user
   def compare_with_users
-    if User.find_by(email: self.email)
-      errors.add(:email, "is already being used by an account holder.")
-      throw :abort
-    end
+    return unless User.exists?(email: self.email)
+
+    errors.add(:email, "is already being used by an account holder.")
```
Bilka2: i18n for this would also be welcome, but no worries if you'd rather not

Bilka2 (commented):
Yep, the livejournal test is flaky. I kicked the test, but if it doesn't want to behave I'll make a fix for it in another PR (I have it floating around somewhere already)

app/models/invite_request.rb
anmazz: yeah I can do that. there's the uniqueness error as well, can I add that too?

anmazz: rubocop didn't like that one lol

app/models/invite_request.rb
```diff
@@ -1,13 +1,15 @@
 class InviteRequest < ApplicationRecord
   validates :email, presence: true, email_format: true
-  validates :email, uniqueness: { message: "is already part of our queue." }
-  before_validation :compare_with_users, on: :create
+  validates :email, uniqueness: { message: :already_in_queue }
```
anmazz: Should there be a uniqueness check on the simplified_email column here instead? I see in the schema it's got the unique key

anmazz: I reverted this change but wondering about this for a possible future fix ^

Bilka2: No, if we use the standard Rails validation then we can't skip the error when `email` already was not unique (https://otwarchive.atlassian.net/browse/AO3-7032)

features/admins/admin_email_blacklist.feature
```diff
@@ -59,3 +59,37 @@ Scenario: Banning a user's email should not affect their ability to post comment
...
+  Then I should not be able to add the email "foo@gmail.com" to the invite queue
+    And I should not be able to add the email "fo.o@gmail.com" to the invite queue
+    And I should not be able to add the email "foo+baz@gmail.com" to the invite queue
+    And I should not be able to add the email "foo@googlemail.com" to the invite queue
+
+Scenario: User's email banned after joining invite queue should remove their email
+  Given I am a visitor
+  When I am on the homepage
+    And I follow "Get an Invitation"
+  When I fill in "Email" with "foo+bar@baz.com"
```
Bilka2: One And please (we don't want to repeat When without a Then in-between)

features/admins/admin_email_blacklist.feature
```diff
@@ -59,3 +59,37 @@ Scenario: Banning a user's email should not affect their ability to post comment
...
+    And I press "Add me to the list"
+  Then I should see "You've been added to our queue"
+  Given I am logged in as a "policy_and_abuse" admin
+  When I go to the manage invite queue page
```
Bilka2: And then this one should be a When -> And. Given should only be used at the start of scenarios

Bilka2 (approved):
Thank you! I left two nitpicks on the tests, but you are free to ignore them

app/models/invite_request.rb
anmazz: hm ok. Sorry if this is a dumb question (and I wouldn't change this in this PR), but does the `validates :email, uniqueness: { message: "is already part of our queue." }` do anything that isn't being done already via the `simplified_email_uniqueness`? Does it need to be there?

PR #5286, opened 2025-07-25
AO3-3071 Add chapter number to comment notifications for multi-chaptered works

marcus8448:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-3071

## Purpose

Updates comment emails so that they contain chapter number information on works with multiple chapters.

**Note**: I have not translated the comment emails as it seems bad to bundle 5 issues into one PR. (But, if it's required I can do it)

## Credit

marcus8448 (he/him)


app/mailers/comment_mailer.rb
```diff
@@ -75,7 +75,20 @@ def comment_reply_sent_notification(comment)
...
     mail(
       to: @comment.comment_owner_email,
-      subject: "[#{ArchiveConfig.APP_SHORT_NAME}] Reply you left to a comment on " + (@comment.ultimate_parent.is_a?(Tag) ? "the tag " : "") + @comment.ultimate_parent.commentable_name.gsub("&gt;", ">").gsub("&lt;", "<")
+      subject: "[#{ArchiveConfig.APP_SHORT_NAME}] Reply you left to a comment on #{commentable_title(@comment)}"
     )
   end
+  
+  private
+
+  def commentable_title(comment)
```
Bilka2: This is a good idea generally. However, I think we need to be able to vary the translation of the first part of the subject ("Comment on" etc) based on what the comment is left on. So each whole subject should have three total variants for the locale (tag/chapter/rest), not just the commentable tile itself.

Unfortunately, it looks like the same applies for the sentences in the emails themselves. The `commenter_pseud_or_name_*` should be fine to keep as one variable, but for the different commentables we need different full sentence variations.

I would limit the i18n to those sentences and not worry about the rest of the emails (especially the footer), since as you say it wouldn't be great to combine all that into one PR.

test/mailers/previews/comment_mailer_preview.rb
Bilka2, line 1: A preview for a comment on a tag would be nice

spec/mailers/comment_mailer_spec.rb
Bilka2, line 1: Could you add a check that for a comment on a single-chapter work it *doesn't* say the chapter?

app/views/comment_mailer/_comment_notification_footer_for_other.text.erb
```diff
@@ -7,4 +7,4 @@ Review comments on "<%= @comment.ultimate_parent.commentable_name %>": <%= polym
 Reply to this comment: <%= comment_url(@comment, :add_comment_reply_id => @comment.id, :only_path => false) %><% end %>
 Go to the thread starting from this comment: <%= comment_url(@comment, :only_path => false) %><% unless @comment.id == @comment.thread %>
 Go to the thread to which this comment belongs: <%= comment_url(:id => @comment.thread, :only_path => false) %><% end %>
-Read all comments on "<%= @comment.ultimate_parent.commentable_name %>": <%= polymorphic_url(@comment.ultimate_parent, :view_full_work => true, :show_comments => true, :anchor => :comments, :only_path => false) %>
+<% if @comment.original_ultimate_parent.kind_of?(Chapter) && @comment.ultimate_parent.chaptered? %>Read all comments on Chapter <%= @comment.original_ultimate_parent.position %> of "<%= @comment.ultimate_parent.commentable_name %>": <%= work_chapter_url(@comment.original_ultimate_parent.work, @comment.original_ultimate_parent, show_comments: true, anchor: :comments) %><% else %>Read all comments on "<%= @comment.ultimate_parent.commentable_name %>": <%= polymorphic_url(@comment.ultimate_parent, view_full_work: true, show_comments: true, anchor: :comments, only_path: false) %><% end %>
```
Bilka2: You can add some whitespace in the code and then use the `<%-` and `-%>` erb tags, they strip whitespace before/after them.

config/i18n-tasks.yml
```diff
@@ -149,6 +149,17 @@ ignore_missing:
...
+  - comment_mailer.subject_for_commentable.subject.chapter
+  - comment_mailer.subject_for_commentable.subject.tag
+  - comment_mailer.subject_for_commentable.subject.work
+  # File: app/mailers/comment_mailer.rb
+  - mailer_helper.content_for_commentable_text.content.chapter
+  - mailer_helper.content_for_commentable_html.content.chapter_html
+  - mailer_helper.content_for_commentable_text.content.tag
+  - mailer_helper.content_for_commentable_html.content.tag_html
+  - mailer_helper.content_for_commentable_text.content.work
+  - mailer_helper.content_for_commentable_html.content.work_html
```
Bilka2: Could you move these above the `All of the following keys are using default values defined in the respective .rb files` comment since that doesn't apply here?

Also, you should be able to work with `{chapter,tag,work}` for the last part of the string, that way you don't need so many lines. But that's a matter of taste, so if you'd rather not that's fine too!

app/mailers/comment_mailer.rb
```diff
@@ -75,7 +90,23 @@ def comment_reply_sent_notification(comment)
...
+  private
+
+  def subject_for_commentable(comment)
+    name = comment.ultimate_parent.commentable_name.gsub("&gt;", ">").gsub("&lt;", "<").html_safe
+    if comment.ultimate_parent.is_a?(Tag)
+      t(".subject.tag", app_name: ArchiveConfig.APP_SHORT_NAME, name: name)
+    elsif comment.original_ultimate_parent.is_a?(Chapter) && comment.ultimate_parent.chaptered?
+      t(".subject.chapter", app_name: ArchiveConfig.APP_SHORT_NAME, position: comment.original_ultimate_parent.position, title: name)
+    else
+      t(".subject.work", app_name: ArchiveConfig.APP_SHORT_NAME, title: name)
```
Bilka2: Technically this is also used for admin posts, not just works, so it'd be better to use a more generic key here. Same for the key used in  `content_for_commentable_*`

app/mailers/comment_mailer.rb
```diff
@@ -75,7 +90,23 @@ def comment_reply_sent_notification(comment)
...
   end
+
+  private
+
+  def subject_for_commentable(comment)
+    name = comment.ultimate_parent.commentable_name.gsub("&gt;", ">").gsub("&lt;", "<").html_safe
+    if comment.ultimate_parent.is_a?(Tag)
+      t(".subject.tag", app_name: ArchiveConfig.APP_SHORT_NAME, name: name)
+    elsif comment.original_ultimate_parent.is_a?(Chapter) && comment.ultimate_parent.chaptered?
+      t(".subject.chapter", app_name: ArchiveConfig.APP_SHORT_NAME, position: comment.original_ultimate_parent.position, title: name)
```
Bilka2:
```suggestion
    elsif comment.parent.is_a?(Chapter) && comment.ultimate_parent.chaptered?
      t(".subject.chapter", app_name: ArchiveConfig.APP_SHORT_NAME, position: comment.parent.position, title: name)
```
`original_ultimate_parent` is just `parent` (refer to comment_methods.rb), and it's clearer to use that instead of this alias

Same applies for the other uses of `original_ultimate_parent` in various files

app/views/comment_mailer/_comment_notification_footer_for_other.html.erb
```diff
@@ -30,10 +30,11 @@
...
-<%= style_link("Read all comments on <i>" + @comment.ultimate_parent.commentable_name + "</i>", 
-  polymorphic_url(@comment.ultimate_parent, 
-    :view_full_work => true, 
-    :show_comments => true, 
-    :anchor => :comments, 
-    :only_path => false
-  )) %>
+<% if @comment.original_ultimate_parent.kind_of?(Chapter) && @comment.ultimate_parent.chaptered? %>
+  <%= style_link("Read all comments on Chapter #{@comment.original_ultimate_parent.position} of <i>#{@comment.ultimate_parent.commentable_name}</i>",
+        work_chapter_url(@comment.original_ultimate_parent.work, @comment.original_ultimate_parent, show_comments: true, anchor: :comments)) %>
```
Bilka2:
```suggestion
        work_chapter_url(@comment.ultimate_parent, @comment.parent, show_comments: true, anchor: :comments)) %>
```
or
```suggestion
        work_chapter_url(@comment.parent.work, @comment.parent, show_comments: true, anchor: :comments)) %>
```

would be better (same for text version and the methods in mailer_helper.rb)

app/views/comment_mailer/_comment_notification_footer_for_other.html.erb
```diff
@@ -30,10 +30,11 @@
...
+  <br />
+<% else %>
+  <%= style_link("Read all comments on <i>#{@comment.ultimate_parent.commentable_name}</i>",
+        polymorphic_url(@comment.ultimate_parent, view_full_work: true, show_comments: true, anchor: :comments, only_path: false)) %>
```
Bilka2: It would be nice to remove the only_path here as well (same for text version)

app/helpers/mailer_helper.rb
```diff
@@ -190,6 +190,43 @@ def commenter_pseud_or_name_link(comment)
     end
   end
 
+  def content_for_commentable_text(comment)
```
Bilka2, line 204: Can you move these two methods either one method up or down in this file? So that it's not inbetween the two versions of commenter_pseud_or_name

app/helpers/mailer_helper.rb
```diff
@@ -190,6 +190,43 @@ def commenter_pseud_or_name_link(comment)
...
+      t(".content.chapter",
+        pseud: commenter_pseud_or_name_text(comment),
+        chapter: t("mailer.general.creation.chapter_position", position: comment.original_ultimate_parent.position),
+        work: comment.ultimate_parent.commentable_name,
+        chapter_url: work_chapter_url(comment.original_ultimate_parent.work, comment.original_ultimate_parent))
+    else
+      t(".content.work",
+        pseud: commenter_pseud_or_name_text(comment),
+        work: comment.ultimate_parent.commentable_name,
+        work_url: work_url(comment.ultimate_parent))
```
Bilka2:
```suggestion
        commentable_url: polymorphic_url(comment.ultimate_parent))
```
Because this can be a news post as well, as mentioned in my comment in comment_mailer

app/helpers/mailer_helper.rb
```diff
@@ -190,6 +190,43 @@ def commenter_pseud_or_name_link(comment)
...
+        tag_link: style_link(comment.ultimate_parent.commentable_name, tag_url(comment.ultimate_parent)))
+    elsif comment.original_ultimate_parent.is_a?(Chapter) && comment.ultimate_parent.chaptered?
+      t(".content.chapter_html",
+        pseud_link: commenter_pseud_or_name_link(comment),
+        chapter_link: style_link(t("mailer.general.creation.chapter_position", position: comment.original_ultimate_parent.position), work_chapter_url(comment.original_ultimate_parent.work, comment.original_ultimate_parent)),
+        work_link: style_creation_link(comment.ultimate_parent.commentable_name, work_url(comment.original_ultimate_parent.work)))
+    else
+      t(".content.work_html",
+        pseud_link: commenter_pseud_or_name_link(comment), 
+        work_link: style_creation_link(comment.ultimate_parent.commentable_name, polymorphic_url(comment.ultimate_parent, only_path: false)))
```
Bilka2: accordingly like the text version, this should be style_link and polymorphic_url as well (and the adjusted locale key)

also the only_path here can make a short journey into the trash bin

app/helpers/mailer_helper.rb
```diff
@@ -190,6 +190,43 @@ def commenter_pseud_or_name_link(comment)
...
+  def content_for_commentable_text(comment)
+    if comment.ultimate_parent.is_a?(Tag)
+      t(".content.tag",
+        pseud: commenter_pseud_or_name_text(comment),
+        tag: comment.ultimate_parent.commentable_name,
+        tag_url: tag_url(comment.ultimate_parent))
+    elsif comment.original_ultimate_parent.is_a?(Chapter) && comment.ultimate_parent.chaptered?
+      t(".content.chapter",
+        pseud: commenter_pseud_or_name_text(comment),
+        chapter: t("mailer.general.creation.chapter_position", position: comment.original_ultimate_parent.position),
```
Bilka2: Since this isn't a link, I think you can combine this locale key with the general locale key for the sentence.

Also, this should contain the chapter title

app/helpers/mailer_helper.rb
```diff
@@ -190,6 +190,43 @@ def commenter_pseud_or_name_link(comment)
...
+
+  def content_for_commentable_html(comment)
+    if comment.ultimate_parent.is_a?(Tag)
+      t(".content.tag_html",
+        pseud_link: commenter_pseud_or_name_link(comment),
+        tag_link: style_link(comment.ultimate_parent.commentable_name, tag_url(comment.ultimate_parent)))
+    elsif comment.original_ultimate_parent.is_a?(Chapter) && comment.ultimate_parent.chaptered?
+      t(".content.chapter_html",
+        pseud_link: commenter_pseud_or_name_link(comment),
+        chapter_link: style_link(t("mailer.general.creation.chapter_position", position: comment.original_ultimate_parent.position), work_chapter_url(comment.original_ultimate_parent.work, comment.original_ultimate_parent)),
```
Bilka2: I think this link caption should be per mailer (so a relative key). And also should have the chapter title

app/helpers/mailer_helper.rb
```diff
@@ -190,6 +190,43 @@ def commenter_pseud_or_name_link(comment)
...
+        work_link: style_creation_link(comment.ultimate_parent.commentable_name, work_url(comment.original_ultimate_parent.work)))
```
Bilka2: I believe this should be a normal link, not a creation link

app/helpers/mailer_helper.rb
```diff
@@ -190,6 +190,43 @@ def commenter_pseud_or_name_link(comment)
...
+      t(".content.work",
+        pseud: commenter_pseud_or_name_text(comment),
+        work: comment.ultimate_parent.commentable_name,
+        work_url: work_url(comment.ultimate_parent))
+    end
+  end
+
+  def content_for_commentable_html(comment)
+    if comment.ultimate_parent.is_a?(Tag)
+      t(".content.tag_html",
```
Bilka2: Locale key wise, for keys that have a HTML and text variant, we prefer going with `.content.tag.text` and `content.tag.html` (the HTML safe thing will also work with this html suffix)

config/locales/mailers/en.yml
```diff
@@ -39,6 +39,79 @@ en:
...
       username: Admin username
+  comment_mailer:
+    comment_notification:
+      content:
+        chapter: "%{pseud} left the following comment on %{chapter} of %{work} (%{chapter_url}):"
+        chapter_html: "%{pseud_link} left the following comment on %{chapter_link} of %{work_link}"
+        tag: "%{pseud} left the following comment on the tag %{tag} (%{tag_url}):"
+        tag_html: "%{pseud_link} left the following comment on the tag %{tag_link}"
+        work: "%{pseud} left the following comment on %{work} (%{work_url}):"
+        work_html: "%{pseud_link} left the following comment on %{work_link}"
```
Bilka2: the html versions of the strings should also have the colon at the end. same for the other emails

spec/mailers/comment_mailer_spec.rb
```diff
@@ -245,6 +245,89 @@
...
+                               comment.original_ultimate_parent,
+                               show_comments: true,
+                               anchor: :comments)
+        expect(subject).to have_text_part_content("Read all comments on Chapter #{comment.original_ultimate_parent.position} of \"#{comment.ultimate_parent.commentable_name}\": #{url}")
+      end
+    end
+  end
+
+  shared_examples "a notification without a chapter reference" do
+    it "has the chapter in the subject line" do
```
Bilka2: needs a wording update to be negated

app/helpers/mailer_helper.rb
marcus8448: I think it should be a creation link - looking at my inbox the work titles are styled this way (italicized and bold).

marcus8448: Previously it was bolded and italicized in the mailer itself

https://github.com/otwcode/otwarchive/blob/882ce4c239358a285d09717ddd6012f8c3a1f5e0/app/views/comment_mailer/comment_notification.html.erb#L9

Bilka2: oops, missed that - you're right!

app/helpers/mailer_helper.rb
Bilka2: (style_creation_link is correct here as well then)

test/mailers/previews/comment_mailer_preview.rb
Bilka2, line 1: And one chaptered without a chapter title, please

app/helpers/mailer_helper.rb
```diff
@@ -201,6 +201,52 @@ def commenter_pseud_or_name_text(comment)
...
+
+  def content_for_commentable_html(comment)
+    if comment.ultimate_parent.is_a?(Tag)
+      t(".content.tag.html",
+        pseud_link: commenter_pseud_or_name_link(comment),
+        tag_link: style_link(comment.ultimate_parent.commentable_name, tag_url(comment.ultimate_parent)))
+    elsif comment.parent.is_a?(Chapter) && comment.ultimate_parent.chaptered?
+      t(".content.chapter.html",
+        pseud_link: commenter_pseud_or_name_link(comment),
+        chapter_link: style_link(comment.parent.title.blank? ? t(".chapter_position.untitled", position: comment.parent.position) : t(".chapter_position.titled", position: comment.parent.position, title: comment.parent.title), work_chapter_url(comment.parent.work, comment.parent)),
```
Bilka2: This is getting into maybe not worth it territory, but we try to match the variable for the link and the key for the link caption. Maybe you can just drop the `position` from the caption keys so we can get closer to that? (so `chapter_link` and `.chapter.untitled` + `.chapter.titled`)

Bilka2 (changes requested):
Two more things, sorry

test/mailers/previews/comment_mailer_preview.rb
marcus8448: Should there be more previews for each email variant (sent/reply/etc)? Or is just one sufficient?

Bilka2: Just one is fine for now, we'll have to do a bit of an inventory of the comment mailer previews with doing i18n for the emails  anyway

Bilka2: E.g. I'm wondering if it'd be better to use URL params for all the different commentables. But like I said, that's for another Jira issue/PR

PR #5288, opened 2025-07-26
AO3-3730 series stats

anmazz:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-3730 

## Purpose

I'm opening this as a draft so I can get feedback on my general approach/design decisions. This will hopefully have a lot of discussion if/when people can review.

Added:
- Series stats that appear/sort in the same listing as work stats
- Series stats include Work Count, Bookmarks, and Subscriptions
- Bookmarks and Subscriptions are sortable among both Series and Works
- Anything specific to works (kudos, hits, word count) gets set to 0  (NULL when pulled from the db but set to 0 when mapped to the StatItem object -> open to making this get 0 directly from the query)
- Adds new "Types" filter to stats page which groups by Work and Series (retains Fandoms and Flat view)

Design approach
I began trying to use ActiveRecord but realized that was probably not the best way to go about this. This SQL is pretty big, but I've tried to evolve it so that it's easier to read with the CTEs and comments. That being said, it's still a lot, so sorry 😬.

My main goal was to do a union between the series and work stats and project them onto a single, standard object for ease of sorting. I felt the raw SQL was the best and most performant option for this since it gives me fine-grain control over selects/aggregates, doesn't hydrate a bunch of ActiveRecord entries, and avoids N+1 queries altogether. I've also tried as much as possible to limit Cartesian explosion.

Things to consider
- I've left a few comments in the code itself. One big question I had was whether or not to use stat_counters for Works as my table for kudos/bookmarks, or retrieve them from kudos/bookmarks directly?
- I'm not sure what db prod uses. This syntax might not work if it's not the same as local (MariaDB's... interesting)
- I've tested locally, but not incredibly thoroughly. I've made a bunch of rspec tests but I think I should continue adding more. Wanted to get more eyes on this in case design direction was changed.

There's a bunch of commented stuff/puts statements/linting errors/failed tests galore that I'll clean up later so apologies for that. Open to feedback on any and all parts of this PR!

## Testing Instructions

Will fill out later.

## References

## Credit

anna (she/her)


app/helpers/stats_helper.rb
```diff
@@ -0,0 +1,176 @@
...
+        WHERE work_stats.published_in_range = TRUE
+	        AND uw.user_id = :user_id
+          -- Only account for posted works
+          AND uw.work_posted = TRUE
+	      GROUP BY series.id
+      ),
+      -- Get the concatenated fandom string
+      fandom_string AS (
+        SELECT taggings.taggable_id AS work_id,
+        GROUP_CONCAT(DISTINCT tags.name ORDER BY tags.name SEPARATOR ', ') AS fandom_string
```
anmazz, line 91: This is MariaDB specific, so might need to be changed if the SQL solution is used and is meant to be universal

Bilka2:
We would prefer if this used ActiveRecord instead of such a large amount of handwritten SQL.

anmazz:
https://github.com/otwcode/otwarchive/pull/5304

PR #5293, opened 2025-07-31
AO3-7073 Remove location, email, dob from profile & tests

anmazz:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-7073

## Purpose

Removes DOB, location, and email from profile. Removes "Show my date of birth/email to other people" preferences. Removes from tests. Returns nil from the ProfilePresenter.

I think I got everything but of course let me know if not!

## References
Depending on merge order will need to resolve with:
https://github.com/otwcode/otwarchive/pull/5272

## Credit

anna (she/her)

### DIFF
--- a/features/other_a/profile_edit.feature
+++ b/features/other_a/profile_edit.feature
@@ -87,8 +87,11 @@ Scenario: Changing email address shows a confirmation page and sends a confirmat
   When I am a visitor
     And I follow "confirm your email change" in the email
   Then I should see "Sorry, you don't have permission to access the page you were trying to reach. Please log in."
-  When I go to editname's profile page
-  Then I should not see "My email address: bar@ao3.org"
+  When I am logged in as "editname"
+    And I go to editname's profile page
+    And I follow "Edit My Profile"
+    And I follow "Change Email"
+  Then I should see "bar@ao3.org"
 
   When I am logged in as "editname"
     And I follow "confirm your email change" in the email
@@ -97,10 +100,9 @@ Scenario: Changing email address shows a confirmation page and sends a confirmat
     But I should not see "bar@ao3.org"
     But I should not see "You have requested to change your email address"
   When I go to editname's profile page
-  Then I should not see "My email address: valid2@archiveofourown.org"
-  When I log out
-    And I go to editname's profile page
-  Then I should not see "My email address: valid2@archiveofourown.org"
+    And I follow "Edit My Profile"
+    And I follow "Change Email"
+  Then I should see "valid2@archiveofourown.org"
 
 Scenario: Changing email address -- canceling in confirmation step
 


features/other_a/profile_edit.feature
```diff
@@ -12,22 +12,19 @@ Background:
...
-    And I should see "I live in"
+    And I should not see "I live in"
 
 Scenario: Change details
-
   When I change the details in my profile
   Then I should see "Your profile has been successfully updated"
     And 0 emails should be delivered
-    And I should see "I live in"
+    And I should not see "I live in"
```
marcus8448: Not seeing "I live in" doesn't confirm whether the profile update actually worked. Maybe check for the custom Title or About Me instead?

app/decorators/profile_presenter.rb
```diff
@@ -4,10 +4,14 @@ def created_at
...
   end
 
   def email
-    user.email if user.preference.try(:email_visible)
+    nil
+  end
+
+  def location
+    nil
   end
```
marcus8448: These stubs seem unnecessary to me as they're no longer accessed/used on the profile page

marcus8448 (commented):
I'm not an official volunteer, so take all of this with a grain of salt.

From my interpretation of the JIRA ticket, the location and DOB columns (plus the preferences) should also be dropped from the database since they're unused.

I also don't think testing for the lack of a feature is necessary since the code/text that's being tested for is gone (but idk if there's specific rules regarding that here).

features/other_a/profile_edit.feature
anmazz: It confirms the profile labels aren’t there anymore since they’ve been removed

app/decorators/profile_presenter.rb
anmazz: I’m assuming here the columns aren’t being dropped. So I’m just ensuring that the profile returns nil for email/location/dob. Whether or not that’s overkill is up for debate.

anmazz:
> I'm not an official volunteer, so take all of this with a grain of salt.
> 
> From my interpretation of the JIRA ticket, the location and DOB columns (plus the preferences) should also be dropped from the database since they're unused.
> 
> I also don't think testing for the lack of a feature is necessary since the code/text that's being tested for is gone (but idk if there's specific rules regarding that here).

I could be wrong as well but I interpreted it as keeping the columns since a migration seems like a lot and wasn’t explicitly mentioned. I’ll let someone clarify, maybe I should’ve asked

features/other_a/profile_edit.feature
anmazz: But haha yeah I’m trying to prove a negative here so I can see these lines being removed. Half was just proving to myself it was working in the tests

app/decorators/profile_presenter.rb
sarken: The columns should indeed be dropped -- we don't want to hold on to personal data we're not using.

anmazz: Alright cool, thanks! I’ll update this PR

anmazz:
Updated to drop columns, sorry about that. The cucumber tests - I can definitely remove the "not" checks if we want, let me know!

features/other_a/profile_edit.feature
sarken: marcus8448 is right here: we need to update this so we're still confirming that the "I change the details in my profile" step worked and the profile has been updated. It would be a good idea to replace this line with a check for the text "Test title thingy", which is some of the new text added in "I change the details in my profile."

app/decorators/profile_presenter.rb
```diff
@@ -2,12 +2,4 @@ class ProfilePresenter < SimpleDelegator
   def created_at
     user.created_at.to_date
   end
```
sarken: It's probably not worth keeping `ProfilePresenter` around for this method; we can update the profile show view to use `@user.created_at.to_date` instead.

Bilka2:
The dropped columns must be added to [ignored_columns](https://api.rubyonrails.org/classes/ActiveRecord/ModelSchema/ClassMethods.html#method-i-ignored_columns-3D) to guarantee a smooth migration.

sarken:
Ah, thanks for mentioning the ignored columns, @Bilka2 ! I've added it to the migrations wiki page now so
I don't forget next time, but i did it from memory so please tweak as needed: https://github.com/otwcode/otwarchive/wiki/Migrations#user-content-Dropping_columns

features/other_a/profile_edit.feature
```diff
@@ -99,10 +97,10 @@ Scenario: Changing email address shows a confirmation page and sends a confirmat
     But I should not see "bar@ao3.org"
     But I should not see "You have requested to change your email address"
   When I go to editname's profile page
-  Then I should see "My email address: valid2@archiveofourown.org"
+  Then I should not see "My email address: valid2@archiveofourown.org"
   When I log out
     And I go to editname's profile page
-  Then I should see "My email address: valid2@archiveofourown.org"
+  Then I should not see "My email address: valid2@archiveofourown.org"
```
anmazz: Realizing I need to fix this to test whether the email was updated

sarken (approved):
Thank you!

PR #5294, opened 2025-08-02
AO3-7043 Migrate filter tagging table to bigint

connie-feng:
## Issue

https://otwarchive.atlassian.net/browse/AO3-7043

## Purpose

This PR migrates the filter taggings table IDs from INT to BIGINT.

## Credit

Connie Feng, she/her

### DIFF
--- a/factories/tags.rb
+++ b/factories/tags.rb
@@ -140,4 +140,9 @@ FactoryBot.define do
     f.sequence(:name) { |n| "Archive Warning #{n}" }
     canonical { true }
   end
+
+  factory :filter_tagging do
+    association :filter, factory: :canonical_fandom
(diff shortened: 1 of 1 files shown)


Bilka2:
Similar to your subscriptions PR, could you add a spec that creates a filter tagging with a very large id?

connie-feng:
> Similar to your subscriptions PR, could you add a spec that creates a filter tagging with a very large id?

Oh yeah, I wanted to ask about that! I couldn't seem to find a filter tagging spec file, and wanted to check that there isn't already one before I create one?

Bilka2:
Yup, please create one! It looks like you'll also have to define a factory for filer tagging, that one can go in factories/tags.rb though

PR #5295, opened 2025-08-02
AO3-7091 Stop using fixtures in feature tests

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-7091

## Purpose

Stop using the fixtures in feature tests.

## Credit

Bilka

PR #5297, opened 2025-08-03
AO3-7093 Add test for invalid records in otwseed fixtures

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-7093

## Purpose

Make sure the fixtures used to seed the database during development are valid records. Remove the invalid ones so the test passes. Also, replace some user emails that looked suspiciously valid.

Note: This is a different set of fixtures than the ones removed in #5295. These are in test/fixtures, the other ones were in features/fixtures.

## Credit

Bilka


sarken (approved):
Thank you for the email changes especially!

PR #5298, opened 2025-08-03
AO3-7023 Fix ExternalWorkscontroller#fetch when the URL is invalid

connie-feng:
## Issue

https://otwarchive.atlassian.net/browse/AO3-7023

## Purpose

This PR fixes the 500 error when ExternalWorksController#fetch gets an invalid URL for `external_work_url`. 

## Credit

Connie Feng, she/her

### DIFF
--- a/app/controllers/external_works_controller.rb
+++ b/app/controllers/external_works_controller.rb
@@ -11,8 +11,12 @@ class ExternalWorksController < ApplicationController
   # Used with bookmark form to get an existing external work and return it via ajax
   def fetch
     if params[:external_work_url]
-      url = Addressable::URI.heuristic_parse(params[:external_work_url]).to_str
-      @external_work = ExternalWork.where(url: url).first
+      begin
+        url = Addressable::URI.heuristic_parse(params[:external_work_url]).to_str
+        @external_work = ExternalWork.where(url: url).first
+      rescue Addressable::URI::InvalidURIError => e
+        Rails.logger.info("Invalid external_work_url: #{e.message}")
+      end
     end
     respond_to do |format|
       format.js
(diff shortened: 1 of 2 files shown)


app/controllers/external_works_controller.rb
```diff
@@ -11,8 +11,12 @@ def new
   # Used with bookmark form to get an existing external work and return it via ajax
   def fetch
     if params[:external_work_url]
-      url = Addressable::URI.heuristic_parse(params[:external_work_url]).to_str
-      @external_work = ExternalWork.where(url: url).first
+      begin
+        url = Addressable::URI.heuristic_parse(params[:external_work_url]).to_str
+        @external_work = ExternalWork.where(url: url).first
+      rescue Addressable::URI::InvalidURIError => e
+        Rails.logger.info("Invalid external_work_url: #{e.message}")
```
Bilka2: We don't care about this error, it's just users typing in something that's not a URL, so we shouldn't log anything here

spec/controllers/external_works_controller_spec.rb
```diff
@@ -44,6 +44,13 @@
         expect(assigns(:external_work)).to be_nil
       end
     end
+
+    context "when the URL is invalid" do
+      it "does not set an external work" do
+        get :fetch, params: { external_work_url: "Invalid URL.", format: :json }
```
Bilka2: Please check the status of the response here to make sure it's not 500

spec/controllers/external_works_controller_spec.rb
```diff
@@ -44,6 +44,15 @@
         expect(assigns(:external_work)).to be_nil
       end
     end
+
+    context "when the URL is invalid" do
+      it "does not error and does not set an external work" do
+        get :fetch, params: { external_work_url: "Invalid URL.", format: :json }
```
Bilka2: I had another look at our monitoring and the request format this is erroring on is the js request, not the json, so that's the one we should check

Bilka2 (approved):
Thank you!

PR #5299, opened 2025-08-04
AO3-7090 Update action mailer default protocol

anmazz:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-7090

## Purpose

Updates action mailer default protocol to https.

I don't believe this needs tests since the test.rb overrides the config, but not 100% sure.

## Credit

anna (she/her)


Bilka2:
> I don't believe this needs tests since the test.rb overrides the config, but not 100% sure.

Could you override the host in the mailer config in tests as well? That should make a few tests fail that currently check the protocol, so fixing those should be enough automated testing for this PR.

anmazz:
> > I don't believe this needs tests since the test.rb overrides the config, but not 100% sure.
> 
> Could you override the host in the mailer config in tests as well? That should make a few tests fail that currently check the protocol, so fixing those should be enough automated testing for this PR.

@Bilka2 
change both:
```
config.action_controller.default_url_options = { host: "www.example.com", port: nil, protocol: "https" }
config.action_mailer.default_url_options = config.action_controller.default_url_options
```

or just
```
config.action_controller.default_url_options = { host: "www.example.com", port: nil }
config.action_mailer.default_url_options = config.action_controller.default_url_options.merge(protocol: "https")
```
keeping action_controller the same?

Bilka2:
Just the action mailer

anmazz:
Need suggestions here on how best to fix:
```
Failing Scenarios:
cucumber features/works/work_download.feature:172 # Scenario: Download work shows inspiring work link
cucumber features/works/work_download.feature:180 # Scenario: Download work shows inspiring external inspiring work link
```

One possible solution is to relax the checks in `work_download_steps.rb` where it's matching the exact url. Since there are only 2 tests failing, this seems more ideal. Updating the action_controller default options to https in test.rb is another solution but I don't think we wanna deal with that.

Bilka2:
The URLs in the downloads tests should not change because they aren't mailer URLs.

I suspect they're changing because of the change to `Rails.application.routes.default_url_options` to fix the other tests. Instead it would be better to change the ... email should link to ... testing step to expect HTTPS links.

anmazz:
> The URLs in the downloads tests should not change because they aren't mailer URLs.
> 
> I suspect they're changing because of the change to `Rails.application.routes.default_url_options` to fix the other tests. Instead it would be better to change the ... email should link to ... testing step to expect HTTPS links.

true that. fixed

Bilka2 (approved):
Thank you!

weeklies (commented):
May I suggest that that the HTTP links to Open Doors be fixed as well? They can be found in 
```
admin_deleted_work_notification.text.erb 
claim_notification.html.erb 
claim_notification.text.erb 
invitation_to_claim.html.erb 
invitation_to_claim.text.erb
opendoors_link in mailer_helper.rb
```

Bilka2:
Agreed with weeklies if you're up for it, anna, but it's fine if you'd rather not

weeklies (approved):
Thank you!

PR #5301, opened 2025-08-05
AO3-7094 Allow basic use of custom properties and var() function in site skins

sarken:
## Issue

## Purpose

Allow users to define custom properties (aka variables) and use them with the `var()` function in site skins, with some restrictions:
* The values for custom properties can't use `url()` or single or double quotation marks. We only allow values with `url()`, `'`, or `"` for certain properties, but we don't know which property a variable will be paired with when we're sanitizing it.
* The `var()` function can't be used as the value for `content` or `font-family`. Their values currently get different sanitization than the values for every other property and there's no reasonable way to apply that to the value of the custom property being accessed with `var()`.

### DIFF
--- a/app/models/work_skin.rb
+++ b/app/models/work_skin.rb
@@ -18,7 +18,7 @@ class WorkSkin < Skin
         return false
       end
       if property == "position" && value == "fixed"
-        # Do not internationalize the , used as a join in this erorr -- it's reflective of the comma used in the list of selectors, which does not change based on locale.
+        # Do not internationalize the , used as a join in this error -- it's reflective of the comma used in the list of selectors, which does not change based on locale.
         errors.add(:base, :work_skin_banned_value_for_property, property: property, selectors: ruleset.selectors.join(", "), value: value)
         return false
       end
--- a/lib/css_cleaner.rb
+++ b/lib/css_cleaner.rb
@@ -76,7 +76,7 @@ module CssCleaner
         (prefix.blank? || sel.start_with?(prefix)) ? sel : "#{prefix} #{sel}"
       end
       clean_declarations = ""
-      # Do not internationalize the , used as a join in these erorrs -- it's reflective of the comma used in the list of selectors, which does not change based on locale.
+      # Do not internationalize the , used as a join in these errors -- it's reflective of the comma used in the list of selectors, which does not change based on locale.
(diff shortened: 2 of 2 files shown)


app/models/work_skin.rb
```diff
@@ -8,8 +8,18 @@ class WorkSkin < Skin
...
+        errors.add(:base, :work_skin_custom_properties)
+        return false
+      end
+      if value.match(/\A(#{CssCleaner::VAR_FUNCTION_REGEX})\z/)
+        errors.add(:base, :work_skin_var)
+        return false
+      end
       if property == "position" && value == "fixed"
-        errors.add(:base, ts("The %{property} property in %{selectors} cannot have the value %{value} in Work skins, sorry!", property: property, selectors: ruleset.selectors.join(", "), value: value))
+        # Do not internationalize the , used as a join in this erorr -- it's reflective of the comma used in the list of selectors, which does not change based on locale.
```
brianjaustin: Nitpick:
```suggestion
        # Do not internationalize the , used as a join in this error -- it's reflective of the comma used in the list of selectors, which does not change based on locale.
```

lib/css_cleaner.rb
```diff
@@ -56,55 +63,63 @@ def clean_css_code(css_code, options = {})
...
-          elsif (cleanval = sanitize_css_declaration_value(property, value)).blank?
-            errors.add(:base, ts("The %{property} property in %{selectors} cannot have the value %{value}, sorry!", property: property, selectors: rs.selectors.join(", "), value: value))
-          elsif (!caller_check || caller_check.call(rs, property, value))
-            clean_declarations += "  #{property}: #{cleanval}#{is_important ? ' !important' : ''};\n"
+        # remove whitespace and convert &gt; entities back to the > direct child selector
+        sel = selector.gsub(/\n/, "").gsub("&gt;", ">").strip
+        (prefix.blank? || sel.start_with?(prefix)) ? sel : "#{prefix} #{sel}"
+      end
+      clean_declarations = ""
+      # Do not internationalize the , used as a join in these erorrs -- it's reflective of the comma used in the list of selectors, which does not change based on locale.
```
brianjaustin: Same as above
```suggestion
      # Do not internationalize the , used as a join in these errors -- it's reflective of the comma used in the list of selectors, which does not change based on locale.
```

PR #5303, opened 2025-08-06
AO3-7058 Upgrade to Rails 7.2

Bilka2:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-7058

## Purpose

Upgrade to Rails 7.2.

There were 404 and 422 error html's (in addition to the 406 one) generated by the upgrade script that I removed, since the errors_controller redirects those to the custom views.

## Testing Instructions

The only big change is the test-related change, so I think we can just do only a smoke test for this.

A volunteer with Staging db access should test that calling `inspect` on a comment in the console displays the email field as filtered.

## Credit

Bilka


config/puma.rb
Bilka2, line 1: We have a custom version of this file in ansible, so do we need this one?

config/environments/production.rb
```diff
@@ -92,6 +101,9 @@
   # Do not dump schema after migrations.
   config.active_record.dump_schema_after_migration = false
 
+  # Only use :id for inspections in production.
+  config.active_record.attributes_for_inspect = [ :id ]
```
Bilka2: Do we want this change? It will affect the default output in the console

Bilka2:
Test failures related to https://github.com/rails/rails/pull/48585. We might have to override the adapter in certain tests so we can test enqueuing there while keeping the inline adapter for the other tests. https://api.rubyonrails.org/classes/ActiveJob/TestHelper.html#method-i-queue_adapter_for_test might help.

config/initializers/new_framework_defaults_7_2.rb
```diff
@@ -0,0 +1,70 @@
...
+# topic in the database.
+#
+# If `enqueue_after_transaction_commit` is set to `:default`, the queue adapter
+# will define the behaviour.
+#
+# Note: Active Job backends can disable this feature. This is generally done by
+# backends that use the same database as Active Record as a queue, hence they
+# don't need this feature.
+#++
+# Rails.application.config.active_job.enqueue_after_transaction_commit = :default
```
Bilka2: Based on this option being deprecated again in Rails 8.0 and https://github.com/rails/rails/pull/52659#issuecomment-2302824156, I think we should set this to `:always`. Will try in another commit after fixing the other test issue

config/environments/production.rb
Bilka2: I use inspect for getting all the fields quite often so I decided to comment this out for now

config/puma.rb
```diff
@@ -0,0 +1,34 @@
...
+# Global VM Lock (GVL) it has diminishing returns and will degrade the
+# response time (latency) of the application.
+#
+# The default is set to 3 threads as it's deemed a decent compromise between
+# throughput and latency for the average Rails application.
+#
+# Any libraries that use a connection pool or another resource pool should
+# be configured to provide at least as many connections as the number of
+# threads. This includes Active Record's `pool` parameter in `database.yml`.
+threads_count = ENV.fetch("RAILS_MAX_THREADS", 3)
```
zz9pzza: We do not do threading....

config/puma.rb
```diff
@@ -0,0 +1,34 @@
...
+threads threads_count, threads_count
+
+# Specifies the `port` that Puma will listen on to receive requests; default is 3000.
+port ENV.fetch("PORT", 3000)
```
zz9pzza: And it doesn't run on that port either :)

config/puma.rb
zz9pzza: No we don't

config/initializers/filter_parameter_logging.rb
```diff
@@ -4,5 +4,5 @@
 # Use this to limit dissemination of sensitive information.
 # See the ActiveSupport::ParameterFilter documentation for supported notations and behaviors.
 Rails.application.config.filter_parameters += [
-  :passw, :secret, :token, :_key, :crypt, :salt, :certificate, :otp, :ssn
+  :passw, :email, :secret, :token, :_key, :crypt, :salt, :certificate, :otp, :ssn
```
Bilka2: This will filter fields containing "email" in the inspect output which is relevant for the admin activity log which calls inspect on comments (which can have guest emails). Do we want to revert this?

spec/spec_helper.rb
```diff
@@ -189,14 +189,6 @@ def run_all_indexing_jobs
     ScheduledReindexJob.perform(reindex_type)
   end
 
-  # In Rails pre-7.2, "config.active_job.queue_adapter" is respected by some
-  # test cases but not others. In request specs, the queue adapter will be
-  # overridden to ":test", so we need to call "perform_enqueued_jobs" to
-  # process jobs.
-  #
-  # Refer to https://github.com/rails/rails/pull/48585.
-  perform_enqueued_jobs if ActiveJob::Base.queue_adapter.instance_of? ActiveJob::QueueAdapters::TestAdapter
```
redsummernight, line 198: I was gonna suggest setting `config.active_job.queue_adapter = :test` on the test environment config overall to avoid needing to override `queue_adapter_for_test` in places, but I remember now that it broke all Cucumber tests expecting emails sent with `deliver_later` (#5055). Is it still the case with the upgrade?

Either way we can do this part separately.

config/application.rb
```diff
@@ -108,6 +108,9 @@ class Application < Rails::Application
     # Use Resque to run ActiveJobs (including sending delayed mail):
     config.active_job.queue_adapter = :resque
 
+    # TODO: Remove with Rails 8.0 where this option will be deprecated
+    config.active_job.enqueue_after_transaction_commit = :always
```
redsummernight, line 109: Is `:always` the default with 7.2?

config/initializers/filter_parameter_logging.rb
```diff
@@ -4,5 +4,9 @@
...
 ]
+
+# WARNING: Due to https://github.com/rails/rails/issues/48704, the filter parameters defined here do not get applied to
```
redsummernight: Let's avoid setting this list in 2 places. Should we remove `config.filter_parameters` from application.rb, and keep the list here with the addition from https://github.com/rails/rails/issues/48704#issuecomment-1873236944?

config/application.rb
```diff
@@ -65,7 +65,7 @@ class Application < Rails::Application
     config.action_view.automatically_disable_submit_tag = false
 
     # Configure sensitive parameters which will be filtered from the log file.
-    config.filter_parameters += [:content, :password, :terms_of_service_non_production]
```
redsummernight, line 68: Should we keep `:password` still?

config/initializers/filter_parameter_logging.rb
redsummernight: If we do need the info and access the field directly, e.g. `comment.email`, does it get filtered? If not, then filtering it out from `.inspect` output should be fine.

Bilka2: It doesn't get filtered if accessed directly, but the admin activity log specifically uses `inspect` which would filter it: https://github.com/otwcode/otwarchive/blob/81cad7454af9c10d5fdb269c9ae5f13ac6f599f0/app/controllers/admin/admin_users_controller.rb#L182

It's only for user comments currently, so these won't have emails set, so maybe we still don't care?

config/application.rb
Bilka2: No, 7.2 defaults to `:default`, which means the job adapter "chooses". However, Rails 8.0 deprecates this option and defaults it to `:always`, refer to this discussion here: https://github.com/rails/rails/pull/52659#issuecomment-2302824156

Considering that we would never get the advantage of a rollback if we did enqueue within the transaction (because we don't use the db as our job backend), we can already use the Rails 8.0 default

config/application.rb
Bilka2: from the other file:
> partially matched (e.g. passw matches password) 

so no, that would be redundant

spec/spec_helper.rb
Bilka2: I can't find a way to set it only for specs, and setting it in `test.rb` still breaks the cucumber tests since they rely on jobs running immediately

sarken:
fwiw, inspect really shouldn't be used that way, so we can and should change that (we just haven't gotten to it yet)Sent from my iPhoneOn Aug 10, 2025, at 3:12 AM, Bilka ***@***.***> wrote:﻿
@Bilka2 commented on this pull request.



In config/initializers/filter_parameter_logging.rb:
> @@ -4,5 +4,5 @@
 # Use this to limit dissemination of sensitive information.
 # See the ActiveSupport::ParameterFilter documentation for supported notations and behaviors.
 Rails.application.config.filter_parameters += [
-  :passw, :secret, :token, :_key, :crypt, :salt, :certificate, :otp, :ssn
+  :passw, :email, :secret, :token, :_key, :crypt, :salt, :certificate, :otp, :ssn

It doesn't get filtered if accessed directly, but the admin activity log specifically uses inspect which would filter it: https://github.com/otwcode/otwarchive/blob/81cad7454af9c10d5fdb269c9ae5f13ac6f599f0/app/controllers/admin/admin_users_controller.rb#L182
It's only for user comments currently, so these won't have emails set, so maybe we still don't care?

spec/spec_helper.rb
Bilka2: If I set the adapter only in specs (by doing `ActiveJob::Base.queue_adapter = :test` in spec_helper) a large number of other specs start failing because they currently expect jobs to be executed immediately (most obvious are ones that call `RedisJobSpawner.perform_now`). So I think overriding the queue adapter to `:test` selectively is the best solution for now

redsummernight: Thanks for checking this!

config/initializers/filter_parameter_logging.rb
redsummernight: With https://github.com/otwcode/otwarchive/pull/5303#issuecomment-3172475309, let's leave a TODO here to remove `email` once we've cleaned up the use of `inspect` in `AdminActivity.log_action`.

Bilka2: I think I may be misunderstanding - why would we want to allow emails to be logged after we clean up our usage of inspect?

redsummernight: Sorry, I misspoke... I meant leave `email` out for now and re-add it later.

But since the logs are only for user comments, you're right that we don't care. This is fine as is.

Bilka2: :tada:

redsummernight:
Can you bump the version to [7.2.2.2](https://rubyonrails.org/2025/8/13/Rails-Versions-8-0-2-1-7-2-2-2-and-7-1-5-2-have-been-released)? Then we can skip #5317 and #5318.

PR #5304, opened 2025-08-07
AO3-3730 series stats

anmazz:
## Issue

https://otwarchive.atlassian.net/browse/AO3-3730
https://otwarchive.atlassian.net/browse/AO3-6301

## Purpose

Refactored with ActiveRecord.

- Series stats that appear/sort in the same listing as work stats
- Series stats include Work Count, Bookmarks, and Subscriptions
- Bookmarks and Subscriptions are sortable among both Series and Works
- Adds new "Types" filter to stats page which groups by Work and Series (retains Fandoms and Flat view) 

Also fixes fandom tags on series overview page where fandoms attached only to draft works would appear in listing.
image

## Credit

anna


app/models/chapter.rb
```diff
@@ -49,7 +49,14 @@ def inherit_creatorships
 
   scope :in_order, -> { order(:position) }
   scope :posted, -> { where(posted: true) }
-
+  
+  # filtered subset, not intended for deletion
+  # rubocop:disable Rails/HasManyOrHasOneDependent, Rails/InverseOf
+  has_many :approved_root_comments, lambda {
+    where(commentable_type: "Chapter", depth: 0, spam: false, approved: true)
```
Hamham6, line 58: From https://otwarchive.atlassian.net/browse/AO3-5994, 'spam' isn't actually being used yet, and 'approved' is the column for checking spam. Does this need to account for comments on moderated works that aren't approved by the creator ('unreviewed', from the Jira issue) and/or comments hidden by admins that aren't spam?

app/views/series/_series_module.html.erb
```diff
@@ -19,9 +19,11 @@
   </h4>
   <h5 class="fandoms heading">
     <span class="landmark"><%= ts('Fandom') %>:</span>
-    <%= series.work_tags.select{ |tag| tag.type == "Fandom" }.
-                         sort.collect{ |tag| link_to_tag_works(tag) }.
-                         join(', ').html_safe %>
+    <%= series.posted_works_fandom_tags
+          .sort
+          .collect { |tag| link_to_tag_works(tag) }
+          .join(", ")
```
Hamham6, line 25: You can use `support.array.words_connector` in the `.join()`  so the commas and spaces are i18nned. Or `to_sentence`, but idk whether the "sentence" behavior is desired or not in this context so go with what official volunteers say on that.

app/views/stats/index.html.erb
```diff
@@ -23,17 +23,19 @@
...
+         kudos: t(".kudos"),
+         comment_thread_count: t(".comment_threads"),
+         work_bookmarks: t(".work_bookmarks"),
+         work_subscriptions: t(".work_subscriptions"),
+         series_bookmarks: t(".series_bookmarks"),
+         series_subscriptions: t(".series_subscriptions"),
+         word_count: t(".word_count") }.each do |stat, stat_text| %>
       <% stat == :word_count ? html_class = "words" : html_class = stat.to_s.humanize.downcase %>
-      <dt class="<%= html_class %>"><%= stat_text %></dt>
+      <dt class="<%= html_class %>"><%= stat_text %>:</dt>
```
Hamham6: `:` is included in the translation of the label, if the label is translatable, and is put in as `mailer.general.metadata_label_indicator` (altho I'm not sure whether non-mailers have a different key?) if the label is not translatable. So the hardcoded colons here and in all the other places should be removed. 

(very sorry for spamming comments, I don't know much about datatables or the like, and I feel embarrassed to show up as a confident "approved these changes" on the whole PR when I understood less than a quarter of it... (^^;)

app/views/stats/index.html.erb
```diff
@@ -74,25 +77,36 @@
           <li>
             <dl>
               <dt>
-                <%= link_to work.title, work_path(:id => work.id) %> 
-                <% if params[:flat_view] %><span class="fandom">(<%= work.fandom_string %>)</span><% end %>
-                <span class="words">(<%= ts("%{wordcount} words", wordcount: number_with_delimiter(work.word_count)) %>)</span>
+                <%= link_to work.title, work.item_path %>
+                <% if work.series? %>(Series)<% end %>
```
Hamham6: I think `(Series)` should be i18nned? Is there already a key for it in helpers?, it feels like the kind of thing that should.

app/views/stats/index.html.erb
```diff
@@ -74,25 +77,36 @@
...
+                <% if (params[:view_type] != nil && params[:view_type] != "fandom") %><span class="fandom">(<%= work.fandom_string %>)</span><% end %>
+                <% if work.work? %><span class="words">
+                  (<%= t(".word_count_with_input", wordcount: number_with_delimiter(work.word_count)) %>)</span>
```
Hamham6: The parenthesises here, and also the ones inside `<span>` on line 82, should be translatable. For this line, just deleting the hardcoded ones and putting them in the yml file works, but for line 82 you'll have to make a new key and use interpolation.

app/views/stats/index.html.erb
```diff
@@ -74,25 +77,36 @@
...
+                <% end %>
               </dt>
               <dd>
                 <dl class="stats">
-                  <% if (work_subscriber_count = Subscription.where(:subscribable_id => work.id, :subscribable_type => 'Work').count) > 0 %>
-                    <dt class="subscriptions"><%= ts("Subscriptions:") %></dt>
-                    <dd class="subscriptions"><%= number_with_delimiter(work_subscriber_count) %></dd>
+                  <% if work.series? %>
+                    <dt class="work_count"><%= t(".work_count") %>:</dt>
+                    <dd class="work_count"><%= number_with_delimiter(work.work_count) %> </dd>
```
Hamham6: It's not that important, because it doesn't actually display, but there shouldn't be a space between the embedded ruby and the closing html tag here, and on lines 101 and 104.

app/models/chapter.rb
anmazz: (Just dumping a bunch of thoughts here so apologies!) I need to double check some of this/have someone else confirm but when I was testing just locally, I set the hidden_by_admin flag = true manually in the database. So I'm not sure if in the process of doing it with an admin account ALSO sets the approved flag to false in that instance. In that scenario (just setting the hidden_by_admin flag to be true, but approved = false) it wasn't filtering out that from the count. 

I can check/test that locally. All that to say, if it doesn't set approved = false when hidden_by_admin = true, then the current version of the code doesn't seem to filter out those comments from the count. Seems like if approved will be renamed to spam then it won't... but this is me saying that without checking.

Anyways, I do think that it should filter those out from the count (hidden_by_admin = true). I'm not super familiar with the comments on moderated works so if anyone could add some more info that'd be great (I'll also dig back into the code/start testing)

app/views/stats/index.html.erb
anmazz: Ah ok cool I'll check that out. Yeah if there's a specific way to go about the label indicator here let me know. I can do it the mailer way if that's preferred

anmazz:
Wanted to bring up a few things for discussion:
- I've basically done a comment thread count for the posted chapters in that particular time range, which is different than current implementation (based off total comment threads for the entire work). So I can revert that to the current implementation.
- As to point 1, it _would_ make more sense if the other stats for the works were for that particular range of time (bookmarks in X year, kudos in X year, so on so forth). Hits wouldn't be able to display accurate counts for the year though so that's the issue, along with the potential hit to the db counting the kudos. 
- I've had the work count for series reflects total posted in time frame just like word count for the work stats (inclusive only of posted works)

app/models/chapter.rb
sarken: You're right: since `approved` is just for spam, setting `hidden_by_admin` to true does *not* automatically set `approved = false`.

We generally do not include `reviewed = false` comments in counts. Here's what we usually include when we count comments: https://github.com/otwcode/otwarchive/blob/81cad7454af9c10d5fdb269c9ae5f13ac6f599f0/lib/acts_as_commentable/commentable_entity.rb#L26-L54

app/views/series/_series_module.html.erb
sarken: Just the `join()` is right -- we don't use `to_sentence` in tag lists. (We actually use CSS to insert commas for the non-fandom tags to allow users to customize blurbs more to their liking. I'm not sure why we didn't do that with fandoms.)

app/views/stats/index.html.erb
sarken: I don't think we have a way to handle colons for non-mailers yet, so it's okay to skip i18n on this for now.

app/views/stats/index.html.erb
sarken: We actually don't have anything like that yet, even though we have the same "(Series)" text on the subscriptions page. 

https://github.com/otwcode/otwarchive/blob/81cad7454af9c10d5fdb269c9ae5f13ac6f599f0/app/views/subscriptions/index.html.erb#L60
https://github.com/otwcode/otwarchive/blob/81cad7454af9c10d5fdb269c9ae5f13ac6f599f0/config/locales/views/en.yml#L2071

app/views/stats/index.html.erb
```diff
@@ -74,25 +77,36 @@
...
+                  (<%= t(".word_count_with_input", wordcount: number_with_delimiter(work.word_count)) %>)</span>
+                <% end %>
               </dt>
               <dd>
                 <dl class="stats">
-                  <% if (work_subscriber_count = Subscription.where(:subscribable_id => work.id, :subscribable_type => 'Work').count) > 0 %>
-                    <dt class="subscriptions"><%= ts("Subscriptions:") %></dt>
-                    <dd class="subscriptions"><%= number_with_delimiter(work_subscriber_count) %></dd>
+                  <% if work.series? %>
+                    <dt class="work_count"><%= t(".work_count") %>:</dt>
```
sarken: The class should be plain `works` to correspond with what we use on series blurbs.

app/views/stats/index.html.erb
```diff
@@ -74,25 +77,36 @@
...
+                  <% if work.work? %>
+                    <dt class="hits"><%= t(".hits") %>:</dt>
+                    <dd class="hits"><%= number_with_delimiter(work.hits) %> </dd>
+                    <%# dt ts("Downloads:") /dt dd work.downloads /dd %>
+                    <dt class="kudos"><%= t(".kudos") %>:</dt>
+                    <dd class="kudos"><%= number_with_delimiter(work.kudos_count) %></dd>
+                    <dt class="comments"><%= t(".comment_threads") %>:</dt>
+                    <dd class="comments"><%= number_with_delimiter(work.comment_thread_count) %></dd>
+                  <% end %>
+                  <dt class="bookmark-count"><%= t(".bookmarks") %>:</dt>
```
sarken: The class here should be `bookmarks` to correspond with what we use on works and series blurbs/meta. (We also as a general rule only use single-word class names, although there are a few exceptions, like third-party JavaScript.)

anmazz: Yeah I changed it because it was getting caught in rubocop as a name I couldn’t use

sarken: That's odd. Rubocop shouldn't care about HTML class names, especially not one we use in many places across the site. You can just ignore it.

anmazz: The deprecated class names were applying on the HTML which was a bit funny:
https://github.com/otwcode/otwarchive/blob/81cad7454af9c10d5fdb269c9ae5f13ac6f599f0/.erb-lint.yml#L8C1-L12C62

anmazz: I can exclude .html.erb files from that rule if that's ok

sarken: Oh, okay, I remember that rule now! That rule is meant for things like divs with the `listbox` class and lists with the `index` class (e.g., [AO3-4283](https://otwarchive.atlassian.net/browse/AO3-4283), not stats. So we do want that check, but you can ignore it in this case.

anmazz:
@redsummernight I’ll definitely fix the internationalization stuff, I just want to make sure there’s discussion about the backend Ruby that I’ve done here as well. Figured that might be easier to fix all in one go if there are general issues/comments

anmazz:
@redsummernight  Sorry for my real poor phrasing. Not trying to call you out or request review from you specifically. If the “Awaiting Review” label COULD be re-added (looking for backend, overall design review) that would be great. If I should address the above comments FIRST, that’s totally fine as well.

really sorry for the bother 😭 I’m getting used to  the activerecord so just want to make sure I’m doing things right.

redsummernight:
No worries, I set the label back to "Awaiting Review".

Little-Pink-Cloud:
Thank you all for your contributions to AO3.

Bilka2 (changes requested):
An overall impression without looking deep yet:

Was the blurb draft tag change the only change outside of the original scope of the stats page? If no, please either list the other changes (so we can decide if we want to keep them in this PR) or revert them. For the blurb change, that has since then been changed in another PR, so please merge `master` to account for that.

Helpers are the wrong place for methods for a controller, they're meant for methods used in views. Methods used only by the controller can just be private methods on the controller itself.

StatItem seems like it may do better being a delegator instead, e.g. similar to CollectionDecorator.

I am confused by the changes in unrelated files like the works and chapter model, they suggest scope creep beyond series. Generally this whole thing seems to scope creep towards a total rework of the stats controller. A rework of the controller isn't a bad idea generally, it does have a lot of cruft. But I think it would be better to do that separately from functional changes and with an overall discussed plan of both what problems it's trying to solve and what design patterns we should use to address those problems, based on what's already in use in this project or other Rails projects. As is, I'm getting lost in the weeds of the implementation details while trying to separate what is a functional change and what is refactoring instead of being able to properly review the functional change or come up with a good design for refactoring. I suspect that this is why this PR lingered in awaiting review for so long: It's trying to do both at once while there's only a plan for one thing. I'd be happy to work with you to figure out that refactoring (e.g. splitting out the param parsing into methods) and organize that via separate PRs/Jira issues, once we have an overall plan.

PR #5305, opened 2025-08-07
AO3-7095 Add user and admin accounts for more roles to fixtures

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-7095

## Purpose

Add more admin and users accounts in the initial seed data to make development with gitpod easier. 

## References

Internal discussion regarding what data to add, reflected in the Jira issue,

## Credit

Bilka

### DIFF
--- a/test/fixtures/pseuds.yml
+++ b/test/fixtures/pseuds.yml
@@ -5,7 +5,7 @@ pseud_00027:
   created_at: 2009-12-12 21:07:09 Z
   delta: true
   updated_at: 2009-12-12 21:07:09 Z
-  is_default: false
+  is_default: true
   id: 27
   user_id: 21
   description: 
@@ -35,7 +35,7 @@ pseud_00038:
   created_at: 2009-12-12 22:39:41 Z
   delta: true
   updated_at: 2009-12-12 22:39:41 Z
-  is_default: false
+  is_default: true
   id: 38
   user_id: 31
   description: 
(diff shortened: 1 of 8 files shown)


test/fixtures/pseuds.yml
```diff
@@ -5,7 +5,7 @@ pseud_00027:
   created_at: 2009-12-12 21:07:09 Z
   delta: true
   updated_at: 2009-12-12 21:07:09 Z
-  is_default: false
+  is_default: true
```
Bilka2, line 8: These changes are done because a number of users did not have default pseuds which lead to errors e.g. on the blocked users page

test/fixtures/roles_users.yml
```diff
@@ -34,3 +34,6 @@ join_00001:
   updated_at: 2009-06-27 18:59:55
   role_id: "6"
   user_id: "1"
+testofficial_official:
+  role_id: 8
+  user_id: 24
```
sarken, line 39: Just a drive-by comment: I think we should also have someone with the opendoors role because (afaik) they have access to tools no other role has.

I don't think the translator role does anything at the moment, so maybe we could drop it from testuser or testuser2 and replace it with opendoors. My initial thought was to give the opendoors role to testuser2 to reduce confusion between the opendoors role and the archivist role, but otoh, it could be convenient to have both roles on one account... 🤔

Bilka2: I added it to testuser since the role is only ever used together with archivist

redsummernight (approved):

PR #5306, opened 2025-08-08
AO3-7038 Restrict ability to grant invitations to all current users

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-7038

## Purpose

Add an AdminInvitationPolicy and restrict access to `grant_invites_to_users` to superadmins.

## Credit

Bilka

### DIFF
--- a/app/controllers/admin/admin_invitations_controller.rb
+++ b/app/controllers/admin/admin_invitations_controller.rb
@@ -26,6 +26,8 @@ class Admin::AdminInvitationsController < Admin::BaseController
   end
 
   def grant_invites_to_users
+    authorize :admin_invitation
+
     if invitation_params[:user_group] == "All"
       Invitation.grant_all(invitation_params[:number_of_invites].to_i)
     else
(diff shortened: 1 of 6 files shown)


app/controllers/admin/admin_invitations_controller.rb
```diff
@@ -26,6 +26,8 @@ def invite_from_queue
   end
 
   def grant_invites_to_users
+    authorize :admin_invitation
+
```
zz9pzza, line 30: Is it out of scope to say this should be an asynchronous task ?  Its going to timeout.

sarken: Yeah, definitely out of scope.

app/policies/admin_invitation_policy.rb
```diff
@@ -0,0 +1,7 @@
+class AdminInvitationPolicy < ApplicationPolicy
```
redsummernight: I think if it's relevant to a specific model, we should use model policies instead of headless ones. You can combine this with the existing `InvitationPolicy` and use `authorize(Invitation)`/`policy(Invitation)`.

redsummernight (approved):

PR #5307, opened 2025-08-09
AO3-5996 record spam column in comments

dismayonnaise:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5996

## Purpose
The `comments.approved` column is used to flag a comment as spam. This is not a very intuitive column name, so a `spam` column was added to the `comments` table in a previous PR. This PR starts writing to the `spam` column on create & update, and backfills that column with a rake task.

Note that `comments.approved` and `comments.spam` are inverse to one another. When a comment's `approved: true`, then then `spam` should be `false`.

I also made some small changes to the rspec tests:
- In `comment_spec.rb`, I put the `account_age_threshold_for_comment_spam_check` tests into its own context so I could add spam flagging on comment creation tests under the `when any comment is considered spam` context. This is purely organizational-- `account_age_threshold_for_comment_spam_check` tests were not updated
- In `comment_moderation_spec.rb`, I moved a `when logged-in as admin with no role` test context outside of a loop testing various roles. I think it was accidentally included in the `.each` loop looping through roles.

## Testing Instructions

See JIRA Ticket

## Credit

dismayonnaise (she/her)


lib/tasks/comment_tasks.rake
```diff
@@ -0,0 +1,41 @@
+namespace :comment do
+  desc "Syncs the Comments.approved column to the new Comments.spam column"
+  task(:sync_approved_to_spam, [:limit] => :environment) do |_t, args|
```
dismayonnaise: Check out this JIRA comment for more documentation on this rake task: https://otwarchive.atlassian.net/browse/AO3-5996?focusedCommentId=369411

The `limit` arg, when provided, will allow an admin to only run this rake task across a few records as a test run before running it across the entire Comment collection.

app/models/comment.rb
```diff
@@ -66,6 +66,7 @@ def by_anonymous_creator?
 
   def check_for_spam
     self.approved = skip_spamcheck? || !spam?
+    self.spam = !skip_spamcheck? && spam? # Logically equivalent to !(skip_spamcheck? || !spam?)
```
dismayonnaise: `!skip_spamcheck? && spam?` should be logically equivalent to `!(skip_spamcheck? || !spam?)`. Here's the truth table I used to validate. Please double check my work.

| skip_spamcheck? | spam? | !skip_spamcheck? | !spam? | skip_spamcheck? OR !spam? | !skip_spamcheck? AND spam? |
| --- | --- | --- | --- | --- | --- |
| T | T | F | F | T | F |
| T | F | F |T | T | F |
| F | T | T | F | F | T |
| F | F | T | T | T | F |

app/models/comment.rb
```diff
@@ -498,12 +499,16 @@ def submit_ham
   end
 
   def mark_as_spam!
-    update_attribute(:approved, false)
+    self.approved = false
+    self.spam = true
+    save(validate: false)
```
dismayonnaise: `update_attribute` skips validation but still executes callbacks & updates `updated_at`. The only way to get equivalent functionality while updating two columns is to use `save(validate: false)`. When `approved` is deprecated from the code, `spam` should swap to using `update_attribute`.

sarken:
Thank you so much for working on this -- I'm really looking forward to being able to use these clearer names! We're a little shorthanded at the moment and are juggling some high priority issues, so it may take us a while to review this, but please know we really appreciate the work you've put into it.

dismayonnaise:
@sarken No worries!! This can go in whenevers 😄 Thank you for the update ❤️Good luck with the high priority issues and thank you for your hard work!!

app/models/comment.rb
Bilka2: This means the spam check is re-run here (as part of calling `spam?`), which we'd like to avoid. So for now it'd be better to make `self.spam = !self.approved` or to set `self.spam` via the spamcheck and then make `self.approved = !self.spam` so the spamcheck is only run once

Bilka2: You have the some mistakes in the second row for the two ! columns, but your logic is still correct

app/models/comment.rb
Bilka2: It should be fine to just make this two separate `update_attribute` calls for now, so it's easier to update to one call later

lib/tasks/comment_tasks.rake
Bilka2: This task is only run once and it's tested by running it on Staging (and with the automated tests), so this argument is not needed. I think it makes things more complicated here, so please remove it

lib/tasks/comment_tasks.rake
```diff
@@ -0,0 +1,40 @@
+namespace :comment do
```
Bilka2: Please move this task into the after tasks instead

lib/tasks/comment_tasks.rake
```diff
@@ -0,0 +1,40 @@
...
+
+    # success_count is used to provide status updates as the rake task is running.
+    success_count = 0
+
+    # Count is only used when a limit is provided. Otherwise, it is ignored.
+    count = 0 if limit.present?
+
+    # The approved and spam columns are inverse values of each other. If approved is true, then spam
+    # should be false, and vice versa. Thus, only records where approved and spam values are the same need to be synced.
+    Comment.where("(approved IS TRUE AND spam IS TRUE) OR (approved IS FALSE AND spam IS FALSE)").find_in_batches do |batch|
```
Bilka2: This query will kill the database because there is no index on either of these columns, so it needs to retrieve all rows to check the where condition.

Instead, it would be better to simply get all comments and then just update all of them with the find_in_batches

lib/tasks/comment_tasks.rake
```diff
@@ -0,0 +1,40 @@
...
+      batch.each do |comment|
+        if comment.approved == comment.spam # Guard clause: guards against
```
Bilka2: Rails is clever enough to not actually do a database write if the new value is the same as the current value, so the update can be pretty naive and just say `comment.update_attribute(:spam, !comment.approved)` without this precheck

lib/tasks/comment_tasks.rake
```diff
@@ -0,0 +1,40 @@
...
+        next if limit.blank?
+
+        # Provide verbose logging if limit is set
+        puts "Updated comment #{comment.id} spam attribute to #{comment.spam}."
+        # Break out of loop if limit is met or exceeded
+        count += 1
+        break if limit <= count
+      end
+      # Provide periodic logging to monitor rake task progress.
+      puts "#{success_count} comments synced."
```
Bilka2: You can find examples in the after tasks for how we usually do progress updates like this, look for `find_in_batches.with_index` (though note that we do not want to get the total count of comment batches to update, because that count query would also kill the db)

spec/controllers/comments/comment_moderation_spec.rb
```diff
@@ -463,12 +483,12 @@
 
           it_behaves_like "a comment that can only be approved by an authorized admin"
         end
+      end
```
Bilka2, line 486: Nicely spotted

spec/models/comment_spec.rb
```diff
@@ -31,143 +31,207 @@
...
     end
 
-    shared_examples "disallows editing the comment if it's changed significantly" do
-      it "prevents editing the comment if it's changed significantly" do
-        subject.edited_at = Time.current
-        subject.comment_content = "Spam content" * 12
-        expect(subject.save).to be_falsey
-        expect(subject.errors[:base]).to include("This comment looks like spam to our system, sorry! Please try again.")
-      end
+    context "when any comment is considered spam" do
```
Bilka2: this seems like a misnomer, later down below you have `receive(:spam?).and_return(false)` so not all comments are considered spam. I think it'd be better to have the spam flagging on creation in its own context and the keep `"when any comment is considered spam"` context the same

spec/models/comment_spec.rb
```diff
@@ -31,143 +31,207 @@
...
+            allow_any_instance_of(Comment).to receive(:skip_spamcheck?).and_return(false)
+            allow_any_instance_of(Comment).to receive(:spam?).and_return(true)
+          end
 
-      context "when account_age_threshold_for_comment_spam_check is set" do
-        before do
-          admin_setting.update_attribute(:account_age_threshold_for_comment_spam_check, 10)
+          it "is flagged as spam" do
+            expect { subject.save! }
+              .to raise_error(ActiveRecord::RecordInvalid)
```
Bilka2: I think this test will be slightly faster/cleaner if you call `save` without the bang and then check the return value instead

spec/models/comment_spec.rb
```diff
@@ -31,143 +31,207 @@
...
+            expect(subject.errors[:base][0]).to include "spam"
```
Bilka2:
```suggestion
            expect(subject.errors[:base].first).to include "spam"
```

spec/models/comment_spec.rb
```diff
@@ -31,143 +31,207 @@
...
+            expect(subject.approved).to be_falsey
+            expect(subject.spam).to be_truthy
+          end
         end
 
-        context "for a new user" do
-          let(:user) { create(:user, created_at: Time.current) }
+        context "when spamcheck is not skipped and content is spammy" do
```
Bilka2:
```suggestion
        context "when spamcheck is not skipped and content is not spammy" do
```

spec/models/comment_spec.rb
```diff
@@ -765,4 +829,36 @@
       end
     end
   end
+
+  describe "#mark_as_spam!" do
+    let(:comment) { create(:comment, approved: 1, spam: 0) }
+
+    before do
+      # Guard clause; emails shouldn't send in test env, but including just to be safe
```
Bilka2: emails...? this would at most go to akismet and there it's fine to rely on the environment check in the model. Even if that was removed the config doesn't have a valid AKISMET_KEY so nothing can actually happen. So this guard clause is redundant

Bilka2 (changes requested):
Thank you for working on this, I left some feedback. I hope I addressed all your questions

app/models/comment.rb
dismayonnaise: Fantastic catch!! I'll go ahead and edit the second row :D Thank you!!

spec/models/comment_spec.rb
dismayonnaise: You're totally right--thank you for this catch. Will be shifting the spam flagging on creation into its own context.

spec/models/comment_spec.rb
dismayonnaise: I love this enhancement. Thank you.

spec/models/comment_spec.rb
dismayonnaise: Gotcha--that makes sense. I'll go ahead and remove the guard clauses.

dismayonnaise:
Apologies for the long turnaround on the feedback. JIRA comment has been update to reflect the rake task changes. Please let me know if there's anything else I should update/tweak. Thank you very much!!

app/models/comment.rb
```diff
@@ -65,7 +65,8 @@ def by_anonymous_creator?
   validate :check_for_spam, on: :create
 
   def check_for_spam
-    self.approved = skip_spamcheck? || !spam?
+    self.spam = !skip_spamcheck? && spam?
```
Bilka2, line 68: Sidenote, this diff makes it easy to reason that this logic has the same result with [De Morgan's laws](https://en.wikipedia.org/wiki/De_Morgan%27s_laws), it's "The negation of `A or B` is the same as `not A and not B`"

spec/controllers/comments/default_rails_actions_spec.rb
```diff
@@ -460,7 +460,7 @@
         context "when the commentable is spam" do
           let(:spam_comment) { create(:comment, commentable: work_with_guest_comment_on) }
 
-          before { spam_comment.update_attribute(:approved, false) }
+          before { spam_comment.update_columns(approved: false, spam: true) }
```
Bilka2: [update_columns](https://api.rubyonrails.org/classes/ActiveRecord/Persistence.html#method-i-update_columns) skips callbacks, unlike `update_attribute`. We should continue using update_attribute in this test as well as the freeze spec

spec/models/comment_spec.rb
```diff
@@ -768,4 +829,26 @@ def queue_adapter_for_test
       end
     end
   end
+
+  describe "#mark_as_spam!" do
+    let(:comment) { create(:comment, approved: 1, spam: 0) }
```
Bilka2: Can you make this let and the other one below use `true` and `false`?

spec/models/comment_spec.rb
```diff
@@ -176,6 +176,67 @@ def queue_adapter_for_test
       end
     end
 
+    context "spam flagging on creation" do
+      subject { build(:comment, pseud: user.default_pseud) }
+      let(:user) { create(:user) }
```
Bilka2:
```suggestion
      subject { build(:comment) }
```

dismayonnaise:
Thank you so much for the notes :D Let me know if there's anything I need to tweak. I really appreciate the review!!

PR #5312, opened 2025-08-10
AO3-6644 Add jpeg support for collection header image URL

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6644

## Purpose

Allow jpeg image links for the collection header URL.

## References

Adopted #4692

PR Limit: 6 out of 5
Using reviewer bonus:
https://github.com/otwcode/otwarchive/pull/4937
https://github.com/otwcode/otwarchive/pull/4903

## Credit

Ivedonestranger (code for previous PR)
Brian (code review for previous PR)
Bilka

### DIFF
--- a/spec/models/collection_spec.rb
+++ b/spec/models/collection_spec.rb
@@ -82,6 +82,22 @@ describe Collection do
       expect(collection.errors.full_messages).to \
         include /Sorry, a collection can only have 10 tags./
     end
+
+    it "disallows invalid header image urls" do
+      collection.header_image_url = "https://example.com/image.webp"
+      expect(collection.save).to be_falsey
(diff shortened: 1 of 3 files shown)


spec/models/collection_spec.rb
```diff
@@ -82,6 +82,22 @@
       expect(collection.errors.full_messages).to \
         include /Sorry, a collection can only have 10 tags./
     end
+
+    it "disallows invalid header image urls" do
+      collection.header_image_url = "https://example.com/image.webp"
+      expect(collection.save).to be_falsey
```
redsummernight: Since we're just testing validations, you could use `.valid?`, which is faster than saving the record.

PR #5322, opened 2025-08-15
AO3-7094 Decrease case-sensitivity for custom properties/variables

sarken:
## Issue

## Purpose

**Please do review the code! This is a draft because I'm waiting on some additional text changes from Docs, not because the code is unfinished.**

- The text about dashes being allowed got eaten in some drafting back and forth with Docs; this restores it.
- The css_parser gem downcases all properties to normalize them. This was a problem because CSS variables are case-sensitive: if you entered `--RED: #900`, it would get converted to `--red: #900` and would no longer work when trying to access it with `var(--RED)`. To handle this, I've chosen to downcase the `var()` function and remove any reference to uppercase letters being allowed in variable names. (more info) 
  - Help text explaining this to come.
  - I removed a line of code downcasing properties on our side since it is not necessary (css_parser downcases as soon as we `add_block!`). 
  - I also added a test so we'll be aware if a css_parser update ever changes this behavior.
- It was possible to use the `var()` function in places we didn't want it (e.g., in work skins or as a value for `content`) by using uppercase letters (e.g., `VAR(--newcolor)` or `var(--NEWcolor)`), which would evade our regexes for giving errors when the `var()` function was found. I've made the regexes case-insensitive. (Downcasing the function as described in the previous point didn't fix this because the value gets downcased when we sanitize it, which happens after we check for its presence.)
  - We had the same problem when checking for the `url()` function, so I fixed that as well.
 
## Testing Instructions

1. Repeat the existing Step 1 and confirm the error message now mentions dashes and does not mention uppercase letters.
2. Check the help text and confirm it now mentions dashes and does not mention uppercase letters. It should also mention that variables will be downcased.
3. Use the `var()` function with uppercase letters, either in the function name itself or the name of the custom property. Confirm it is converted to lowercase, and that it gets stripped if you use it in work skins or with the `font-family` or `content` property. Make sure to test using it in shorthand values as well, e.g. `margin: var(--skinny) auto`.
4. Use the `url()` function with uppercase letters in the function name itself, i.e. `URL()`. Confirm it gets stripped if you use it with a property other than background, background-image, border, border-image, list-style, list-style-image. Confirm it's accepted as the value for at least one of those properties.

### DIFF
--- a/spec/lib/css_cleaner_spec.rb
+++ b/spec/lib/css_cleaner_spec.rb
@@ -37,7 +37,7 @@ describe CssCleaner do
           expect(skin.reload.css).to eq(":root {\n  --rainbow: #fff;\n}\n\n")
         end
 
-        ["url", "URL"].each do |function_name|
+        %w[url URL].each do |function_name|
           it "strips custom property and returns error when value uses #{function_name}() function" do
             skin = build(:skin, css: ":root { --art: #{function_name}(\"https://example.com/img.png\"); }")
             expect(skin.save).to be_falsey
@@ -133,7 +133,7 @@ describe CssCleaner do
           expect(skin.css).to eq("div {\n  border: var(--thick);\n  margin: var(--thin);\n}\n\n")
         end
 
-        ["var", "VAR"].each do |function_name|
+        %w[var VAR].each do |function_name|
           it "strips #{function_name}() function from content property and returns error" do
             skin = build(:skin, css: "p:before { content: #{function_name}(--text) }")
             expect(skin.save).to be_falsey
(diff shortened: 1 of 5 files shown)


spec/lib/css_cleaner_spec.rb
```diff
@@ -137,11 +173,19 @@
...
 
-      it "strips variable functions and returns error" do
-        skin = build(:work_skin, css: "p { color: var(--yellow) }")
-        expect(skin.save).to be_falsey
-        expect(skin.css).to eq("")
-        expect(skin.errors[:base]).to include("The var() function is not allowed in work skins.")
+      context "with var() function as a value" do
+        {
+          "strips lowercase function and returns error" => "p { color: var(--puce) ; }",
+          "strips uppercase Vfunction and returns error" => ".class { width: var(--narrow) }",
```
redsummernight: Typo on function. Also did you mean to test `VAR(--narrow)` here?

redsummernight (approved):

redsummernight:
Let's switch the labels when it's out of draft state.

PR #5323, opened 2025-08-16
AO3-6581 Update heading + page title for "Marked for Later" page

gelvinp:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6581

## Purpose

This PR updates the heading and page title for the "Marked for Later" page, as well as registers both the "History" and "Marked for Later" strings with the i18n system. 

## Credit

Jamis Gelvin (they/them)

### DIFF
--- a/app/controllers/readings_controller.rb
+++ b/app/controllers/readings_controller.rb
@@ -11,10 +11,10 @@ class ReadingsController < ApplicationController
 
   def index
     @readings = @user.readings.visible
-    @page_subtitle = t(".history_page_title", username: @user.login)
+    @page_subtitle = t(".history_page_title")
     if params[:show] == 'to-read'
       @readings = @readings.where(toread: true)
-      @page_subtitle = t(".marked_for_later_page_title", username: @user.login)
+      @page_subtitle = t(".marked_for_later_page_title")
     end
     @readings = @readings.order("last_viewed DESC")
     @pagy, @readings = pagy(@readings)
--- a/app/controllers/readings_controller.rb
+++ b/app/controllers/readings_controller.rb
@@ -11,10 +11,10 @@ class ReadingsController < ApplicationController
 
   def index
     @readings = @user.readings.visible
-    @page_subtitle = ts("History")
+    @page_subtitle = t(".history")
     if params[:show] == 'to-read'
       @readings = @readings.where(toread: true)
-      @page_subtitle = ts("Marked For Later")
+      @page_subtitle = t(".marked_for_later")
     end
     @readings = @readings.order("last_viewed DESC")
     @pagy, @readings = pagy(@readings)
--- a/app/controllers/readings_controller.rb
+++ b/app/controllers/readings_controller.rb
@@ -11,10 +11,10 @@ class ReadingsController < ApplicationController
 
   def index
     @readings = @user.readings.visible
-    @page_subtitle = t(".history")
+    @page_subtitle = t(".history_page_title", username: @user.login)
     if params[:show] == 'to-read'
       @readings = @readings.where(toread: true)
-      @page_subtitle = t(".marked_for_later")
+      @page_subtitle = t(".marked_for_later_page_title", username: @user.login)
     end
     @readings = @readings.order("last_viewed DESC")
     @pagy, @readings = pagy(@readings)
--- a/app/views/readings/index.html.erb
+++ b/app/views/readings/index.html.erb
@@ -1,5 +1,11 @@
 <!--Descriptive page name, messages and instructions-->
-<h2 class="heading"><%= @page_subtitle %></h2>
+<h2 class="heading">
+  <% if params[:show] == "to-read" %>
+    <%= t(".marked_for_later") %>
+  <% else %>
+    <%= t(".history") %>
+  <% end %>
+</h2>
 <!--/descriptions-->
 
 <!--subnav-->
@@ -19,7 +25,13 @@
 <!--/subnav-->
 
 <!--main content-->
-<h3 class="landmark heading"><%= ts('List of History Items') %></h3>
+<h3 class="landmark heading">
+  <% if params[:show] == "to-read" %>
+    <%= t(".marked_for_later_landmark") %>
+  <% else %>
+    <%= t(".history_landmark") %>
+  <% end %>
+</h3>
 <% if logged_in? && @readings.present? %>
   <ol class="reading work index group">
     <% @readings.each do |reading| %>
--- a/app/views/readings/index.html.erb
+++ b/app/views/readings/index.html.erb
@@ -1,10 +1,8 @@
+<% mode = params[:show] == "to-read" ? "marked_for_later" : "history" %>
+
 <!--Descriptive page name, messages and instructions-->
 <h2 class="heading">
-  <% if params[:show] == "to-read" %>
-    <%= t(".marked_for_later") %>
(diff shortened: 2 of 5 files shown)


Bilka2:
Hi!

I've updated the Jira issue status to In Review so no one will mistakenly create a duplicate pull request and I've also given your Jira account the ability to comment on, assign, and transition issues in the future.

config/locales/views/en.yml
```diff
@@ -2025,6 +2025,10 @@ en:
       make_default: Make this name default
       name: Name
       submit: Submit
+  readings:
```
Hamham6, line 2028: Since page titles are set in the controller, these should go in the controllers en.yml instead of the views en.yml, altho if it also gets used as the header then I don't know if that changes that...😵‍💫

app/controllers/readings_controller.rb
```diff
@@ -11,10 +11,10 @@ def load_user
 
   def index
     @readings = @user.readings.visible
-    @page_subtitle = ts("History")
+    @page_subtitle = t(".history")
```
Hamham6: Should these keys both have `_page_title` tacked onto the ends of them? That's what was done for the key for the Subscription page type titles: https://github.com/otwcode/otwarchive/commit/063fd1c71553182bee4698277d536ea41706e4dc.

app/views/readings/index.html.erb
```diff
@@ -1,5 +1,5 @@
 <!--Descriptive page name, messages and instructions-->
-<h2 class="heading"><%= ts('History') %></h2>
+<h2 class="heading"><%= @page_subtitle %></h2>
```
Hamham6: Headers are usually distinct i18n strings from page titles. Altho I don't know if there's a header equivalent of `span_if_current` or something to make the header switch?

features/other_a/reading.feature
Hamham6, line 1: A test can also be made for the page title in `other_a/page_title.feature`, similar past PR for reference: https://github.com/otwcode/otwarchive/commit/4aaf5eac5af2e5a9c16648886edbd4fc35336a0b. (I've also seen page title tests put in spec tests for their controller, and I'm not sure what the preference is for one over the other, but if a spec test is preferred this PR does that, for reference: https://github.com/otwcode/otwarchive/commit/063fd1c71553182bee4698277d536ea41706e4dc.)

app/views/readings/index.html.erb
Hamham6, line 1: For whatever reason, I can't comment on line 22 so it's going here instead. 
```
<h3 class="landmark heading"><%= ts('List of History Items') %></h3>
```
Can this also change to "List of Marked for Later Items"? Or would that cause problems because it's a landmark? Or just out of scope?

app/controllers/readings_controller.rb
Hamham6: Also, should the username be added to the front of the subtitle itself, like `username - Readings` etc for consistency with other user pages?

features/other_a/reading.feature
gelvinp: It looks like it's also fairly common to test for page title outside of the page_title.feature (it shows up in nine other features), so in the interest of keeping the number of cucumber steps down, I think it makes sense to put those checks in the scenario's I already wrote? Let me know if there is a specific preference for it to be somewhere else though!

gelvinp:
You brought up a lot of good points, thanks! I've split it up so that the page title translations are in the controllers en.yml (and have the username), meanwhile the text for the header and landmark are in the views en.yml.

config/locales/controllers/en.yml
```diff
@@ -193,6 +193,9 @@ en:
     clear:
       error: There were problems deleting your history. Please try again later.
       success: Your history is now cleared.
+    index:
+      history_page_title: "%{username} - History"
+      marked_for_later_page_title: "%{username} - Marked for Later"
```
weeklies: I'm not a volunteer, but I don't think adding the username is part of this issue. Users can only access their own history and marked for later pages (admins also cannot view these pages), so adding it is unnecessary.

sarken: weeklies is correct; the username is unnecessary here.

config/locales/views/en.yml
```diff
@@ -2025,6 +2025,12 @@ en:
...
+    index:
+      history: History
+      history_landmark: List of History Items
+      marked_for_later: Marked for Later
+      marked_for_later_landmark: List of Marked for Later Items
```
sarken: We can reduce the repetition a smidge by doing this instead:

```suggestion
      history: 
        page_heading: History
        landmark: List of History Items
      marked_for_later: 
        page_heading: Marked for Later
        landmark: List of Marked for Later Items
```

Then you'd access these translations in the view using, e.g., `.marked_for_later.landmark` and `.history.landmark`. 

If you do this, you could probably reduce the logic a bit in the view as well by setting the first part of our key in a variable, which would allow you to only write one line of code for each translation:

```ruby
<% mode = params[:show] == "to-read" ? "marked_for_later" : "history" >

<%# This would go in the h2, without needing the surrounding logic %>
<%= t(".#{mode}.page_heading") %>

<%# And this would likewise go in the h3 %>
<%= t(".#{mode}.landmark") %>
```

features/other_a/reading.feature
sarken: It's definitely fair to put the page title checks in existing feature tests! (We generally don't like to use specs tests for that.) 

I'm looking at the existing tests in reading.feture and I think it would make a lot of sense to add the history page title test to the first scenario:

```cucumber
  Given the following activated user exists
    | login        |
    | first_reader |
  When I am logged in as "second_reader"
    And I go to first_reader's reading page
  Then I should see "Sorry, you don't have permission"
    And I should not see "History" within "div#dashboard"
  When I go to second_reader's reading page
  # Changes after this comment
  Then I should see the page title "History"
    And I should see "History" within "h2.heading"
    And I should see "History" within "div#dashboard"
  ```

And the marked for later one to this scenario:

```cucumber
  # If you have a second to change "story" to "work," we'd appreciate it! We try not to assume works are stories (or even text).
  Scenario: Mark a story to read later

  Given I am logged in as "writer"
  When I post the work "Testy"
  Then I should see "Work was successfully posted"
  When I am logged out
    And I am logged in as "reader"
    And I view the work "Testy"
  Then I should see "Mark for Later"
  When I follow "Mark for Later"
  Then I should see "This work was added to your Marked for Later list."
  # Changes after this comment
  When I go to reader's reading page
  Then I should see the page title "Marked for Later"
    And I should see "Marked for Later" within "h2.heading"
    And I should see "Testy"
    And I should see "(Marked for Later.)"
```

sarken (changes requested):

gelvinp:
Thanks for your feedback!

features/other_a/reading.feature
```diff
@@ -11,7 +11,9 @@ Feature: Reading count
   Then I should see "Sorry, you don't have permission"
     And I should not see "History" within "div#dashboard"
   When I go to second_reader's reading page
-  Then I should see "History" within "div#dashboard"
+  Then I should see the page title "History | Example Archive"
```
sarken: Just one more quick note! We prefer to leave the " | Example Archive" portion off so the tests don't fail for folks who have changed their dev environment's name.

gelvinp: That makes sense, thanks :+1:

PR #5324, opened 2025-08-17
AO3-7096 Track when a collection's anonymous/unrevealed settings were…

connie-feng:
## Issue

https://otwarchive.atlassian.net/browse/AO3-7096

## Purpose
This PR adds audit tracking for a collection’s anonymous and unrevealed settings, recording when they were last updated to aid database administration and troubleshooting.

- Adds a migration to add `unrevealed_updated_at` and `anonymous_updated_at` columns to the CollectionPreferenceTable
- New columns are always initialized to the same timestamp as `created_at`
- When the anonymous or unrevealed setting is changed, the corresponding timestamps are updated to match the `updated_at` timestamp

## Credit

Connie Feng, she/her

### DIFF
--- a/app/models/collection_preference.rb
+++ b/app/models/collection_preference.rb
@@ -1,15 +1,31 @@
 class CollectionPreference < ApplicationRecord
   belongs_to :collection
+
+  after_create :set_initial_audit_timestamps
+  after_update :set_audit_timestamps
   after_update :after_update
 
+  private
+
+  def set_initial_audit_timestamps
+    update_columns(
+      unrevealed_updated_at: created_at,
+      anonymous_updated_at: created_at
+    )
+  end
+
+  def set_audit_timestamps
+    updates = {}
+    updates[:unrevealed_updated_at] = updated_at if saved_change_to_unrevealed?
+    updates[:anonymous_updated_at]  = updated_at if saved_change_to_anonymous?
+
+    update_columns(updates) if updates.any?
+  end
+
   def after_update
-    if self.collection.valid? && self.valid?
-      if self.saved_change_to_unrevealed? && !self.unrevealed?
-        self.collection.reveal!
-      end
-      if self.saved_change_to_anonymous? && !self.anonymous?
-        collection.reveal_authors!
-      end
-    end
+    return unless collection.valid? && valid?
+
+    collection.reveal! if saved_change_to_unrevealed? && !unrevealed?
+    collection.reveal_authors! if saved_change_to_anonymous? && !anonymous?
   end
 end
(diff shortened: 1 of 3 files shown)


app/models/collection_preference.rb
```diff
@@ -1,15 +1,31 @@
...
   after_update :after_update
 
+  private
+
+  def set_initial_audit_timestamps
+    update_columns(
+      unrevealed_updated_at: created_at,
+      anonymous_updated_at: created_at
+    )
+  end
```
Bilka2: I think we would prefer if these default to `nil` if possible, because it takes us one less comparison to know that the setting was not changed since creation. It should be fine to just not set them at all on creation

app/models/collection_preference.rb
```diff
@@ -1,15 +1,31 @@
...
+
+  def set_audit_timestamps
```
Bilka2: audit means something specific for us internally, so I'd prefer to not use that word here. Maybe `set_updated_at_timestamps` or something similar?

db/migrate/20250816213210_add_audit_timestamps_to_collection_preferences.rb
```diff
@@ -0,0 +1,20 @@
...
+
+    reversible do |dir|
+      dir.up do
+        execute <<~SQL.squish
+          UPDATE collection_preferences
+          SET unrevealed_updated_at = created_at,
+              anonymous_updated_at  = created_at
+        SQL
+      end
+    end
```
Bilka2: same as my comment on `set_initial_audit_timestamps` though here with the added note that I think this would confuse us in the future, since it would be set to created_at although the setting was changed at some point before this PR was merged

db/migrate/20250816213210_add_audit_timestamps_to_collection_preferences.rb
```diff
@@ -0,0 +1,20 @@
+class AddAuditTimestampsToCollectionPreferences < ActiveRecord::Migration[7.1]
+  uses_departure! if Rails.env.staging? || Rails.env.production?
+
+  def change
+    change_table :collection_preferences, bulk: true do |t|
```
Bilka2: Are you sure this interacts well with departure? We usually use `add_column` instead of bulk changing the table

Bilka2 (changes requested):

app/models/collection_preference.rb
connie-feng: Okay, got it! I got confused because the ticket said "It’s fine if the new columns are initially set to the created_at date rather than nil"

Bilka2 (approved):
Thank you!

PR #5326, opened 2025-08-18
AO3-7094 Downcase var() function when used as a value for any property

sarken:
## Issue

## Purpose

- Makes sure we downcase the `var()` function when it's used as a value for non-shorthand properties like `color`. I originally had it in `sanitize_css_token` which is only used when tokenizing the values for shorthand properties like `border`.
- Adds a few seemingly random tests. I was initially exploring another approach that necessitated the tests and it seemed a shame to waste them when I found an easier fix.
- Adds a lot of code comments for things I learned on the way.

## Testing Instructions

Make sure the `var()` function gets converted to lowercase when used as a value for both shorthand and non-shorthand properties.

### DIFF
--- a/lib/css_cleaner.rb
+++ b/lib/css_cleaner.rb
@@ -202,7 +202,7 @@ module CssCleaner
   end
 
   def sanitize_css_token(token)
-    if token.match(/gradient/)
+    if token.match?(/gradient/)
       sanitize_css_gradient(token)
     else
       sanitize_css_value(token)
@@ -237,7 +237,7 @@ module CssCleaner
     # If it's a comma-separated set of valid values, it's fine. However, we need
     # to downcase any var() functions to match the css_parser gem's downcasing
     # of property names.
-    if value_stripped.match?(/^(#{VALUE_REGEX}\,?\s*)+$/i)
+    if value_stripped.match?(/^(#{VALUE_REGEX},?\s*)+$/i)
       return value unless value.match?(/#{VAR_FUNCTION_REGEX}/)
 
       return value.gsub(/#{VAR_FUNCTION_REGEX}/, &:downcase)


lib/css_cleaner.rb
```diff
@@ -220,22 +224,36 @@ def sanitize_css_gradient(value)
...
+
+    # If it's a comma-separated set of valid values, it's fine. However, we need
+    # to downcase any var() functions to match the css_parser gem's downcasing
+    # of property names.
+    if value_stripped.match?(/^(#{VALUE_REGEX}\,?\s*)+$/i)
+      return value unless value.match?(/#{VAR_FUNCTION_REGEX}/)
 
-    # if it's a comma-separated set of valid values it's fine
-    return value if value_stripped =~ /^(#{VALUE_REGEX}\,?\s*)+$/i
+      return value.gsub(/#{VAR_FUNCTION_REGEX}/) { |match| match.downcase }
```
sarken: I'm not sure if this is worth doing or if I should just do `return value.gsub(/#{VAR_FUNCTION_REGEX}/) { |match| match.downcase }` without first checking that the value contains a `var()`.

lib/css_cleaner.rb
```diff
@@ -194,12 +202,8 @@ def tokenize_and_sanitize_css_value(value)
   end
 
   def sanitize_css_token(token)
-    case token
-    when /gradient/
+    if token.match(/gradient/)
```
Bilka2:
```suggestion
    if token.match?(/gradient/)
```

lib/css_cleaner.rb
Bilka2: I prefer this because it means the change has a smaller scope

PR #5327, opened 2025-08-18
AO3-7084 Disclose when image stripping occurs

WelpThatWorked:
## Issue

https://otwarchive.atlassian.net/browse/AO3-7084

## Purpose

Disclose when image stripping occurs. Applied to locations listed in issue plus site banner form. Text on Logged in, multi-pseud comment form now positioned/styled in the same way as the single-pseud and logged out versions. Other cases where new text does not appear should look identical to before.
CSS tweaks
\- swap `line-height` for negative margin in `.toggled form p.footnote, .dynamic form p.footnote`. Should have identical effect, but without overlapping on multiline text.

<details>
<summary>New text screenshots</summary>

Logged out commenter: 
image

Logged in commenter:
image

Bookmark creation:
image

Banner edit:
image
</details>

## Credit

WelpThatWorked

### DIFF
--- a/app/views/bookmarks/_bookmark_form.html.erb
+++ b/app/views/bookmarks/_bookmark_form.html.erb
@@ -61,15 +61,16 @@
         <dl>
           <dt><%= f.label :bookmarker_notes, ts("Notes"), for: notes_id %></dt>
           <dd>
-            <p class="footnote" id="notes-field-description">
-              <% if bookmarkable.class != ExternalWork %>
+            <% if bookmarkable.class != ExternalWork %>
+              <p class="footnote" id="notes-field-description-summary">
                 <%= ts("The creator's summary is added automatically.") %>
-                <br />
-              <% end %>
+              </p>
+            <% end %>
+            <p class="footnote" id="notes-field-description-html">
               <%= allowed_html_instructions(strip_images: true) %>
             </p>
             <%= f.text_area :bookmarker_notes, rows: 4, id: notes_id, class: "observe_textlength",
-                "aria-describedby" => "notes-field-description" %>
+                aria: { describedby: "notes-field-description-summary notes-field-description-html" } %>
             <%= generate_countdown_html(notes_id, ArchiveConfig.NOTES_MAX) %>
           </dd>
 
--- a/public/stylesheets/site/2.0/07-interactions.css
+++ b/public/stylesheets/site/2.0/07-interactions.css
@@ -653,7 +653,7 @@ form.single .autocomplete + input + span.submit input[type="submit"] {
 
 .toggled form p.footnote, .dynamic form p.footnote {
   float: none;
-  margin: -0.25em auto;
+  margin: auto;
   padding: 0;
 }
 
@@ -672,7 +672,7 @@ form.single .autocomplete + input + span.submit input[type="submit"] {
     box-shadow: none;
 }
 
-.toggled .bookmark form p.footnote {
+.toggled .bookmark form p.footnote, .dynamic .bookmark form p.footnote {
   padding-bottom: 0.643em;
 }
 
@@ -701,6 +701,13 @@ form.single .autocomplete + input + span.submit input[type="submit"] {
 
 .dashboard form dt.landmark {
   clear: both;
+}
+
+/* CONTEXT: blurb */
+
+.blurb form li.input {
+  display: list-item;
+  padding-left: 0;
 }
 
 /* END == */


app/helpers/application_helper.rb
```diff
@@ -62,9 +62,10 @@ def link_to_rss(link_to_feed)
 
   # 1: default shows just the link to help
   # 2: show_text = true: shows "plain text with limited html" and link to help
-  def allowed_html_instructions(show_text = true)
-    (show_text ? h(ts("Plain text with limited HTML")) : "".html_safe) +
-      link_to_help("html-help")
+  def allowed_html_instructions(show_text: true, strip_images: false)
+    (show_text ? t("application_helper.text_limited_html") : "".html_safe) +
+      link_to_help("html-help") +
+      (strip_images ? "<br />".html_safe + t("application_helper.strip_images") : "".html_safe)
```
Bilka2: It looks like this never called with `show_text` set to false, so you can remove that parameter.

For when this should have the extra sentence it should just be the sentence directly after the link, no linebreak. Ideally, it would be nice to have one big locale string for both sentences, and switch between that and a different locale strings based on the condition, instead of breaking it up per sentence. And the link should be a variable: https://github.com/otwcode/otwarchive/wiki/Internationalization-%28i18n%29-Standards#user-content-Variable_interpolation

app/views/comments/_comment_form.html.erb
```diff
@@ -54,7 +54,7 @@
                           commentable_creator: commentable.by_anonymous_creator? ? t(".anonymous_creator") : get_commenter_pseud_or_name(commentable),
                           commentable_link: commentable_description_link(commentable)) %></span>
             <% end %>
-            (<%= allowed_html_instructions %>)
+            <%= allowed_html_instructions(strip_images: comment.hidden_by_admin || Comment.commentable_object(commentable).class.name.in?(ArchiveConfig.PARENTS_WITH_IMAGE_SAFETY_MODE)) %>
```
Bilka2: Is it possible to use `use_image_safety_mode?` from the comment model here instead?

Bilka2 (changes requested):
From what I can see in the issue, there should be no CSS changes made here? The intended look on the comment form is visible here: https://otwarchive.atlassian.net/browse/AO3-7083

Please add some simple tests that check that the text shows up for guest comments and shows up for user comments in the right circumstances

app/views/comments/_comment_form.html.erb
WelpThatWorked: Not as it stands. `use_image_safety_mode?` checks through `Comment.parent` and `Comment.pseud`, which aren't set until the comment is saved. I considered rewriting it, but that would require something like
```ruby
def use_image_safety_mode?(commentable = nil)
  (persisted? && pseud_id.nil?) || hidden_by_admin || (persisted? ? parent_type : Comment.commentable_object(commentable)).in?(ArchiveConfig.PARENTS_WITH_IMAGE_SAFETY_MODE)
end
  ```
  so since this is _probably_ the only place that will have non-persisted comments, I figured it'd be cleaner just to leave it as-is. Should it be made into it's own function instead?

File: app/helpers/application_helper.rb (hunk 1 above)
Comment (WelpThatWorked) [reply app/helpers/application_helper.rb]: Did not see the screenshot as it wasn't linked from the issue. Guess I should have looked around a bit more, sorry.

For the locale bit, do you mean something like:
```yaml
text_limited_html: Plain text with limited HTML %{help_link}
text_limited_html_strip_images: Plain text with limited HTML %{help_link} Embedded images (<img> tags) will be displayed as HTML, including the image's source link and any alt text.
```

Bilka2: Yes!

app/views/comments/_comment_form.html.erb
Bilka2: I'm a bit worried we may end up updating the conditions in one place but not the other. Maybe it could be a method in the comments helper that's used in both places?

As another note, I think the current condition is incorrect because `commentable` here can be another comment. What we're looking for is the `parent`, which would be only chapter, tag or admin post. (notably not Work or Comment!)

app/helpers/application_helper.rb
WelpThatWorked: I've run into a bit of difficulty with this, I'm not seeing a neat way to have it escape the text of the string (e.g. where it says "\ tags") but _not_ the help link HTML. I could try extending the translate/interpolate function to work better here, or do the escaping/interpolation manually like:
```ruby
html_escape(t(strip_images ? "application_helper.text_limited_html_strip_images" : "application_helper.text_limited_html")).sub("%{help_link}", link_to_help("html-help")).html_safe
```
but otherwise, I think they'd have to be separate strings like they are now. Which would be preferred?

app/views/comments/_comment_form.html.erb
WelpThatWorked: You are correct on the condition being incorrect. On review, I think I can just set the needed fields on the `comment`, since it gets discarded anyway and the other needed models are already there, allowing it to use `use_image_safety_mode?` as-is...

app/helpers/application_helper.rb
Bilka2: It should hopefully work to use html entities for the lesser than and greater than, so `foo bar &lt;img&gt;` together with the suffix for the `_html` locale key so that the link works

WelpThatWorked: Will do, I wasn't sure if that'd cause difficulties for translators.

lydia-theda:
re: styling

I didn't put the image on this issue because I didn't want to formally commit to a final style design based on a mockup of a different issue that I threw together from playing around with Inspect Element during the initial concept phase. 

That said:
* agree the logged in comment form needs style tweaking. perhaps both sentences should be placed below the dropdown, with the smaller font?
* slight preference for no line break between existing HTML sentence and img tag sentence; if anything, there can/should be a line break on the bookmark form between the "automatic summary" sentence and the html sentence(s). 
* agree with your margin change to prevent overlapping, especially on the bookmark form
* unsure/no opinion on right vs left alignment. I'm not sure what "consistent" is primarily referring to – consistent with other fields on the PAC form? on the comment/banner/bookmark forms? footer/footnote alignment elsewhere on the site? something else?

config/locales/helpers/en.yml
```diff
@@ -1,5 +1,8 @@
 ---
 en:
+  application_helper:
+    text_limited_html: Plain text with limited HTML %{help_link}
+    text_limited_html_strip_images_html: Plain text with limited HTML %{help_link} Embedded images (&lt;img&gt; tags) will be displayed as HTML, including the image&#39;s source link and any alt text.
```
Bilka2:
```suggestion
    text_limited_html_strip_images_html: Plain text with limited HTML %{help_link} Embedded images (&lt;img&gt; tags) will be displayed as HTML, including the image's source link and any alt text.
```
It doesn't need to be escaped

public/stylesheets/site/2.0/07-interactions.css
```diff
@@ -648,9 +648,8 @@ form.single .autocomplete + input + span.submit input[type="submit"] {
 
 .toggled form p.footnote, .dynamic form p.footnote {
   float: none;
-  margin: auto;
+  margin: -0.25em auto;
   padding: 0;
-  line-height: 0.5;
 }
```
Bilka2, line 658: This CSS change looks good, but I'm not sure whether it might affect another spot where we don't want it to change. @sarken could you take a look this change is okay or if the class selection is too broad and it should be on `.toggled .bookmark form p.footnote` instead?

Bilka2 (commented):
This looks very good, thank you!

I have just two notes, one of which needs input from someone more experienced with front-end.

Bilka2:
Oh, a note for the future: Forcepushes make subsequent reviews harder since the diff is harder to find and/or subdivide. Since we squash on merge you can just push commits (including merges) normally and it will still only be a single commit in the history on `master`. That will make it easier for future reviewers looking at your PRs in this repo!

app/views/bookmarks/_bookmark_form.html.erb
```diff
@@ -64,8 +64,9 @@
             <p class="footnote" id="notes-field-description">
               <% if bookmarkable.class != ExternalWork %>
                 <%= ts("The creator's summary is added automatically.") %>
+                <br />
               <% end %>
-              <%= allowed_html_instructions %>
+              <%= allowed_html_instructions(strip_images: true) %>
             </p>
             <%= f.text_area :bookmarker_notes, rows: 4, id: notes_id, class: "observe_textlength",
                 "aria-describedby" => "notes-field-description" %>
```
sarken: 👋 Hi, I've been summoned!

Because these are two different topics, let's split this into two different paragraphs. That will also mean juggling the logic a bit so we don't include an empty `<p>` when the bookmarkable is a work or series and updating the `describedby` attribute on the text area to reference both paragraphs:

```suggestion
            <% if bookmarkable.class != ExternalWork %>
              <p class="footnote" id="notes-field-description-summary">
                <%= ts("The creator's summary is added automatically.") %>
              </p>
            <% end %>
            <p class="footnote" id="notes-field-description-html">
              <%= allowed_html_instructions(strip_images: true) %>
            </p>
            <%= f.text_area :bookmarker_notes, rows: 4, id: notes_id, class: "observe_textlength",
                aria: { describedby: "notes-field-description-summary notes-field-description-html" } %>
```

public/stylesheets/site/2.0/07-interactions.css
sarken: First, let me say the `line-height` removal is a definite improvement -- the text is currently very squished. That's especially true on the dynamic bookmark form (that's the class we use on the form when we display it in the bookmark blurb, while toggled is the class we use on the work page).

If we're going to modify the CSS, we actually want to fix our squished text issues, not keep them by replacing the too-short line height with a negative margin. So let's put the `margin` back to its previous value here.

Since we see that the toggled form has way better spacing than the dynamic form, we can just give the dynamic form the same styling by changing `.toggled .bookmark form p.footnote` to `.toggled .bookmark form p.footnote, .dynamic .bookmark form p.note`. The bottom padding fixes our squished text very nicely.

But doing that made me realize _why_ we only had that bottom margin on the toggled form: there was more space between the footnotes and the inputs on the dynamic form than the toggled form, and adding the bottom padding makes the space really noticeable.

Image alt="The toggled form (left) beside the dynamic form (right). The dynamic form has more vertical space above the autocomplete inputs than the toggled form."

That's because the `<li class="input">` on the dynamic form's autocomplete inputs is getting `display: inline` from this rule set, which is mainly there to make things like tags display horizontally: https://github.com/otwcode/otwarchive/blob/b1e112618faa943471ee1dd5b70737638765fa4d/public/stylesheets/site/2.0/13-group-blurb.css#L16-L18

That basically shuffles the input down a bit because it no longer fits inside its surrounding `<li>`.

Luckily, it's an easy fix: right above the `END` comment in this stylesheet, we just need to add this:

```css
/* CONTEXT: blurb */

.blurb form li.input {
  display: list-item;
}
```

sarken (changes requested):
Very long comments, but relatively easy changes, I promise! We do need to remember to check both bookmark form locations when testing.

WelpThatWorked:
Apologies for the delay. 
I _think_ that's everything listed. I noticed the input `li`s pick up a bit of padding when switched to `list-item` display on the `.dynamic` version, so I cancelled that as well.
> We do need to remember to check both bookmark form locations when testing.

Clarify, is this requesting automated testing for the presence of the text, or just something to note when manually testing?

PR #5331, opened 2025-08-22
AO3-6484 Bypass CSS blurb cache if pseuds loaded

weeklies:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6484

## Purpose

Bypass CSS blurb cache for Work and Series with loaded pseuds.

## References

#5213, which didn't make a significant difference in performance

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -563,11 +563,16 @@ module ApplicationHelper
   def css_classes_for_creation_blurb(creation)
     return if creation.nil?
 
-    Rails.cache.fetch("#{creation.cache_key_with_version}/blurb_css_classes-v2") do
-      creation_id = creation_id_for_css_classes(creation)
-      creator_ids = creator_ids_for_css_classes(creation).join(" ")
-      "blurb group #{creation_id} #{creator_ids}".strip
-    end
+    creation_id = creation_id_for_css_classes(creation)
+    creator_ids = if creation.try(:pseuds)&.loaded?
+                    creator_ids_for_css_classes(creation).join(" ")
+                  else
+                    Rails.cache.fetch("#{creation.cache_key_with_version}/blurb_css_classes-v2") do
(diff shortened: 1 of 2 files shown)


app/helpers/application_helper.rb
```diff
@@ -563,11 +563,16 @@ def creator_ids_for_css_classes(creation)
   def css_classes_for_creation_blurb(creation)
     return if creation.nil?
 
-    Rails.cache.fetch("#{creation.cache_key_with_version}/blurb_css_classes-v2") do
-      creation_id = creation_id_for_css_classes(creation)
-      creator_ids = creator_ids_for_css_classes(creation).join(" ")
-      "blurb group #{creation_id} #{creator_ids}".strip
-    end
+    creation_id = creation_id_for_css_classes(creation)
```
Bilka2, line 566: Nice work spotting that there is no reason to cache this call

app/helpers/application_helper.rb
```diff
@@ -563,11 +563,16 @@ def creator_ids_for_css_classes(creation)
...
+    creator_ids = if creation.try(:pseuds)&.loaded?
+                    creator_ids_for_css_classes(creation).join(" ")
+                  else
+                    Rails.cache.fetch("#{creation.cache_key_with_version}/blurb_css_classes-v2") do
```
Bilka2: The cache format changed because you pulled the last line out of this cache call, so the cache key needs to change. Though now I'm wondering whether this will affect how accurately we can measure the performance impact. I hope it should be okay though because the cache will be filled again relatively quickly, likely within the same timeframe that inaccurate due to the release anyway?

Bilka2 (changes requested):

app/helpers/application_helper.rb
weeklies: I think observing the longer term performance should be okay.

Bilka2 (approved):
Thank you!

PR #5332, opened 2025-08-22
AO3-2520 Make comment deletion work on noscript

weeklies:
## Issue

https://otwarchive.atlassian.net/browse/AO3-2520

## Purpose

- Use `button_to` for delete comment, which supports noscript.
- i18n and tidy up comments in affected views.
- Also, fix a visual bug for noscript comment deletion.

Before

After 

As for automated tests, I wasn't sure why `comments_delete.feature` isn't failing before this fix and didn't know of a automated way to ensure the bug is fixed.

### DIFF
--- a/app/helpers/comments_helper.rb
+++ b/app/helpers/comments_helper.rb
@@ -284,7 +284,7 @@ module CommentsHelper
   end
 
   def do_cancel_delete_comment_link(comment)
-    if params[:delete_comment_id] && params[:delete_comment_id] == comment.id.to_s
+    if params[:delete_comment_id] == comment.id.to_s
       cancel_delete_comment_link(comment)
     else
       delete_comment_link(comment)
@@ -315,7 +315,7 @@ module CommentsHelper
   # return html link to delete comments
   def delete_comment_link(comment)
     link_to(
-      ts("Delete"),
+      t("comments.confirm_delete.delete"),
       url_for(controller: :comments,
               action: :delete_comment,
               id: comment,
@@ -326,7 +326,7 @@ module CommentsHelper
   # return link to cancel new reply to a comment
   def cancel_delete_comment_link(comment)
     link_to(
-      ts("Cancel"),
+      t("comments.confirm_delete.cancel"),
       url_for(controller: :comments,
               action: :cancel_comment_delete,
               id: comment,
--- a/app/views/comments/_confirm_delete.html.erb
+++ b/app/views/comments/_confirm_delete.html.erb
@@ -1,5 +1,5 @@
-<h3 class="heading"><%= ts("Are you sure you want to delete this comment?") %></h3>
+<h3 class="heading"><%= t(".delete_confirmation") %></h3>
 <ul class="actions" role="menu">
-	<li><%= link_to ts('Yes, delete!'), comment, :method => :delete %></li>
+	<li><%= button_to t(".button"), comment, method: :delete %></li>
 	<li><%= cancel_delete_comment_link(comment) %></li>
 </ul>
--- a/features/comments_and_kudos/comments_delete.feature
+++ b/features/comments_and_kudos/comments_delete.feature
@@ -5,7 +5,7 @@ Feature: Delete a comment
   I want to be able to delete a comment I added
   As an author
   I want to be able to delete a comment a reader added to my work
-  
+
   Scenario: User deletes a comment they added to a work
     When I am logged in as "author"
       And I post the work "Awesome story"
@@ -28,7 +28,7 @@ Feature: Delete a comment
       And I should see "I didn't mean that"
       And I should see "Comments:1"
       And I should see a link "Hide Comments (1)"
-      
+
   Scenario: Author deletes a comment another user added to their work
     When I am logged in as "author"
       And I post the work "Awesome story"
@@ -40,7 +40,7 @@ Feature: Delete a comment
     Then I should see "Comment deleted."
       And I should not see "Comments:"
       And I should not see a link "Hide Comments (1)"
-    
+
   Scenario: Author deletes a parent comment that another user added to their work
     When I am logged in as "author"
       And I post the work "Awesome story"
--- a/features/step_definitions/comment_steps.rb
+++ b/features/step_definitions/comment_steps.rb
@@ -257,14 +257,14 @@ When /^I compose an invalid comment(?: within "([^"]*)")?$/ do |selector|
   end
 end
 
-When /^I delete the comment$/ do
+When "I delete the comment" do
   step %{I follow "Delete" within ".odd"}
-  step %{I follow "Yes, delete!"}
+  step %{I press "Yes, delete!"}
 end
 
-When /^I delete the reply comment$/ do
+When "I delete the reply comment" do
   step %{I follow "Delete" within ".even"}
-  step %{I follow "Yes, delete!"}
+  step %{I press "Yes, delete!"}
 end
 
 When /^I view the latest comment$/ do
@@ -377,7 +377,7 @@ When /^I delete all visible comments on "([^\"]*?)"$/ do |work|
     visit work_url(work, show_comments: true)
     break unless page.has_content? "Delete"
     click_link("Delete")
-    click_link("Yes, delete!") # TODO: Fix along with comment deletion.
+    click_button("Yes, delete!")
   end
 end
 
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -607,7 +607,7 @@ class CommentsController < ApplicationController
         options = {}
         options[:show_comments] = params[:show_comments] if params[:show_comments]
         options[:delete_comment_id] = params[:id] if params[:id]
-        redirect_to_comment(@comment, options) # TO DO: deleting without javascript doesn't work and it never has!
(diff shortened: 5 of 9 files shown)


app/views/comments/_confirm_delete.html.erb
```diff
@@ -1,5 +1,5 @@
-<h3 class="heading"><%= ts("Are you sure you want to delete this comment?") %></h3>
+<h3 class="heading"><%= t(".delete_confirmation") %></h3>
 <ul class="actions" role="menu">
-	<li><%= link_to ts('Yes, delete!'), comment, :method => :delete %></li>
+	<li><%= button_to t(".button"), comment, method: :delete %></li>
```
weeklies:
```suggestion
    <li><%= button_to t(".button"), comment, method: :delete %></li>
```

marcus8448:
The tests in `comments_delete.feature` previously passed because Capybara uses `data-method` to decide what type of request to use when clicking a link, even with JS disabled.

The driver option `respect_data_method` should change this behaviour, but I haven't been able to get it to work myself.

weeklies:
[It didn't work for me](https://github.com/weeklies/otwarchive/actions/runs/17171910863) either.

app/views/comments/_confirm_delete.html.erb
```diff
@@ -1,5 +1,5 @@
-<h3 class="heading"><%= ts("Are you sure you want to delete this comment?") %></h3>
+<h3 class="heading"><%= t(".delete_confirmation") %></h3>
 <ul class="actions" role="menu">
```
sarken: Could you remove the menu role while you're here?

Edit: I do not know why, but this comment seems to be appearing on the wrong line in "Files change" view but the right line on the main "Conversation" page.

app/views/comments/_confirm_delete.html.erb
```diff
@@ -1,5 +1,5 @@
...
-	<li><%= link_to ts('Yes, delete!'), comment, :method => :delete %></li>
+    <li><%= button_to t(".button"), comment, method: :delete %></li>
```
sarken: Since we have two buttons (or at least two things that look like buttons) here, let's use a name like `confirm_button` or something for a little extra clarity.

app/helpers/comments_helper.rb
```diff
@@ -326,7 +326,7 @@ def delete_comment_link(comment)
   # return link to cancel new reply to a comment
```
sarken: Could you update this comment? It's wrong -- or at least I hope it is and we don't also use this for canceling replies.

app/helpers/comments_helper.rb
```diff
@@ -315,7 +315,7 @@ def frozen_comment_indicator
   # return html link to delete comments
   def delete_comment_link(comment)
     link_to(
-      ts("Delete"),
+      t("comments.confirm_delete.delete"),
```
sarken: Should this be just `comments.delete`? It doesn't seem to be used for deletion confirmation, only for the initial request.

app/views/comments/_confirm_delete.html.erb
```diff
@@ -1,5 +1,5 @@
-<h3 class="heading"><%= ts("Are you sure you want to delete this comment?") %></h3>
+<h3 class="heading"><%= t(".delete_confirmation") %></h3>
```
sarken, line 2: Could you add a comment above this line? Something like `TODO: Make this an h5`. It will need some style adjustments as well, so I don't want to get involved in that right now, but it definitely shouldn't be a higher heading level than then h4 we use for the comment itself.

sarken (changes requested):
I can't get `respect_data_method` override to work either, but I'm not sure adding it in this pull request is a good idea anyway. We might end up with a whole bunch of other failures to tackle because this is far from the only place we use this approach. 

It seems like what we want here is a version of the test that would fail if we were unknowingly using JavaScript/making a JavaScript request. 

It looks like the major difference between the two versions of deletion is that following the comment's "Delete" link with JavaScript disabled reloads the work with params like `?delete_comment_id=123&show_comments=true#comment_123` in order to add the confirmation form. 

What if we add a step to confirm we are on that page before we proceed with deletion and then made a separate `@javascript` test that checked we are _not_ taken to that page? The steps here are clunky and I might not have needed so many changes to make the JavaScript test work, but something like this should work:

```
Then "it should take me to the non-JavaScript delete page" do
  expect(current_url).to include("?delete_comment_id")
end

Then "it should not take me to the non-JavaScript delete page" do
  expect(current_url).not_to include("?delete_comment_id")
end

  Scenario: User deletes a comment they added to a work
    When I am logged in as "author"
      And I post the work "Awesome story"
    When I am logged in as "commenter"
      And I post the comment "Fail comment" on the work "Awesome story"
      And I follow "Delete"
    Then it should take me to the non-JavaScript delete page
    When I press "Yes, delete!"
      And I should not see "Comments:"
      And I should not see a link "Hide Comments (1)"

  @javascript
  Scenario: User deletes a comment they added to a work
    Given the work "Aftermath" by "creator" with guest comments enabled
      And a comment "Fail comment" by "pest" on the work "Aftermath"
      And I am logged in as "pest"
    When I view the work "Aftermath"
    Then I should not see "Fail comment"
    When I display comments
    Then I should see "Fail comment"
    When I follow "Delete"
    Then it should not take me to the non-JavaScript delete page
    When I press "Yes, delete!"
      And I should not see "Comments:"
      And I should not see a link "Hide Comments (1)"
```

app/helpers/comments_helper.rb
weeklies: I've amended it to use the `comments_helper` locale file instead

weeklies:
Not sure why the automated check appears to be passing in previous commits, when the logs show [a failure](https://github.com/otwcode/otwarchive/actions/runs/17511719263/job/49744159101#step:13:42). It's also odd because the failing test is passing on my Docker environment.

features/comments_and_kudos/comments_delete.feature
```diff
@@ -5,13 +5,31 @@ Feature: Delete a comment
...
+      And I should not see a link "Hide Comments (1)"
+
+  @javascript
+  Scenario: User deletes a comment they added to a work with JavaScript
+    Given the work "Awesome story" by "author"
+      And a comment "Fail comment" by "commenter" on the work "Awesome story"
+    When I am logged in as "commenter"
+      And I view the work "Awesome story"
+    Then I should not see "Fail comment"
+    When I display comments
```
weeklies:
```suggestion
    When I follow "Comments (1)"
```

features/comments_and_kudos/comments_delete.feature
```diff
@@ -5,13 +5,31 @@ Feature: Delete a comment
...
+      And I should not see a link "Hide Comments (1)"
+
+  @javascript
+  Scenario: User deletes a comment they added to a work with JavaScript
+    Given the work "Awesome story" by "author"
+      And a comment "Fail comment" by "commenter" on the work "Awesome story"
+    When I am logged in as "commenter"
+      And I view the work "Awesome story"
+    Then I should not see "Fail comment"
+    When I follow "Comments (1)"
```
weeklies:
```suggestion
    When I follow "Comments (1)"
      And I wait 1 second
```

features/comments_and_kudos/comments_delete.feature
```diff
@@ -5,13 +5,32 @@ Feature: Delete a comment
...
+      And I wait 1 second
```
weeklies:
```suggestion
      And I view the work "Awesome story" with comments
```

features/step_definitions/comment_steps.rb
```diff
@@ -390,7 +398,7 @@
 end
 
 When "I display comments" do
-  click_link("Comments")
+  click_link("Comments ")
```
weeklies:
```suggestion
  click_link("Comments")
```

marcus8448:
I did some more digging into why Capybara was using `data-method` in non-JS tests, and it turns out that Cucumber does its own [JS emulation](https://github.com/cucumber/cucumber-rails/blob/8c88d398f2626c7ba477bf5b5941393ea6a7649c/lib/cucumber/rails/capybara/javascript_emulation.rb#L112) in addition to Capybara's `respect_data_method`  parameter. 

[Sample patch](https://github.com/marcus8448/otwarchive/commit/599607b8ffe15284a9cdc3508b975576c3d51631) for disabling emulation per-test (+[associated test run](https://github.com/marcus8448/otwarchive/actions/runs/17521560054/job/49766090652)).

sarken:
Oh, that was some great sleuthing, @marcus8448 -- thank you! I think we should definitely switch to that approach.

weeklies:
The added test is now passing (thank you @marcus8448), but I think it's worth pointing out that a [kudos test](https://github.com/otwcode/otwarchive/actions/runs/17740570381/job/50413260834?pr=5332#step:13:76) failed even when the job shows a success.

sarken:
@weeklies Nothing to worry about! That's the flaky test being referenced a bit lower down in the output. You can see [another recent case where that same test was flaky](https://github.com/otwcode/otwarchive/actions/runs/17778789233/job/50533058056)

sarken (approved):

PR #5334, opened 2025-08-23
AO3-6415 Reorganize the My Assignments page

marcus8448:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6415

## Purpose

This PR splits the My Assignments page to either list unposted or completed assignments and lists assignments from newest to oldest, so that it's easier to find assignments that need work.
The assignment count in the user sidebar has also been updated to only include unposted (actionable) assignments.

It also adds pagination to aid users with many assignments.

## Testing Instructions

TBD

## Credit

marcus8448 (he/him)


app/views/challenge_assignments/_user_index.html.erb
```diff
@@ -1,20 +1,41 @@
...
+          user_assignments_path(@user, collection_id: params[:collection_id], posted: true),
+          params[:posted] == "true") %>
+  </li>
+</ul>
 <!--/subnav-->
 
 <!--main content-->
 <% unless @collection %>
-  <p class="note"><%= ts('Looking for prompts you claimed in a prompt meme? Try') %> <%= link_to ts("My Claims"), user_claims_path(@user) %></p>
+  <p class="notes"><%= t(".claims_note_html", claims_link: link_to(t(".claims_link"), user_claims_path(@user))) %></p>
```
sarken: Just a quick [i18n house style](https://github.com/otwcode/otwarchive/wiki/Internationalization-%28i18n%29-Standards#hyperlinks) note, not a full review! (But I do like what I'm seeing so far; thank you for working on this!) 

When we use links in variable interpolation, we use the `_link` suffix for the variable, but drop it for the translated text, e.g. `claims_link: link_to(t(".claims")`

app/models/challenge_assignment.rb
```diff
@@ -79,6 +79,7 @@ def signups_match
   }
 
   scope :posted, -> { joins(WORKS_JOIN).where("challenge_assignments.creation_id IS NOT NULL AND works.posted = 1") }
+  scope :unposted, -> { joins(WORKS_JOIN).where("challenge_assignments.creation_id IS NULL OR works.posted = 0") }
```
sarken: It looks like we already have an `unposted` scope: https://github.com/otwcode/otwarchive/blob/48abf3174c5d2c5caf0e7ca9419f7674081eab12/app/models/challenge_assignment.rb#L113

The placement is a little odd so it was easy to miss, but I think we want to keep it there so `challenge_assignment` and `challenge_claim` stay organized roughly the same.

app/views/challenge_assignments/_user_index.html.erb
```diff
@@ -1,20 +1,41 @@
...
+</ul>
 <!--/subnav-->
 
 <!--main content-->
 <% unless @collection %>
-  <p class="note"><%= ts('Looking for prompts you claimed in a prompt meme? Try') %> <%= link_to ts("My Claims"), user_claims_path(@user) %></p>
+  <p class="notes"><%= t(".claims_note_html", claims_link: link_to(t(".claims"), user_claims_path(@user))) %></p>
 <% end %>
 
+<%== pagy_nav @pagy %>
```
sarken, line 32: Could you add a blank line after this and another before the bottom pagination? We don't have any hard and fast rules and blank lines between sections, but I think it generally makes one-line bits like pagination easier to spot when skimming the view files. (I actually missed the top pagination when I was first looking for it!)

app/models/user.rb
```diff
@@ -328,6 +328,13 @@ def destroy_empty_series
       destroy_all
   end
 
+  def assignments
+    offers = ChallengeAssignment.joins(offer_signup: :pseud).where(pseuds: { user_id: self.id })
+    pinch_hits = ChallengeAssignment.joins(:pinch_hitter).where(pseuds: { user_id: self.id })
```
sarken: Is there a reason we want to avoid using `offer_assignments` and `pinch_hit_assignments` here? They already only include sent assignments.

config/locales/views/en.yml
```diff
@@ -559,6 +559,22 @@ en:
         no_sign_ups_yet: No sign-ups yet!
         offers_html: Offers &#8595;
         requests_html: Requests &#8595;
+  challenge_assignments:
+    index:
+      access_denied_user: You aren't allowed to see that user's assignments.
```
sarken: This works, but we usually put error messages from the controller in `locales/controllers/en.yml`.

features/gift_exchanges/challenge_giftexchange.feature
```diff
@@ -382,15 +382,23 @@ Feature: Gift Exchange Challenge
...
 
-  Scenario: User fulfills their assignment and it shows on their assigments page as fulfilled
+  Scenario: User fulfills their assignment and it shows on their assignments page as fulfilled
 
     Given everyone has their assignments for "Awesome Gift Exchange"
     When I am logged in as "myname1"
-      And I fulfill my assignment
-    When I go to myname1's user page
+      And I go to myname1's user page
       And I follow "Assignments"
```
sarken: We can tighten these tests up a bit by using `I go to the assignments page for "user"` instead of going to the user page and following the Assignments link all the time.

features/gift_exchanges/challenge_giftexchange.feature
```diff
@@ -716,3 +724,40 @@ Feature: Gift Exchange Challenge
...
+
+  Scenario: User sidebar assignments count only includes unfulfilled assignments.
+    Given the user "recip" exists and is activated
+      And I am logged in as "gifter"
+      And "gifter" has an assignment for the user "recip" in the collection "exchange_collection"
+    When I go to the assignments page for "gifter"
+    Then I should see "Assignments (1)" within "#dashboard"
+    When I fulfill my assignment
+      And I go to the assignments page for "gifter"
+      Then I should see "Assignments (0)" within "#dashboard"
```
sarken: I wouldn't object to incorporating this with the "User fulfills their assignment and it shows on their assignments page as fulfilled" scenario. They're very interconnected and it saves us some set up.

features/gift_exchanges/challenge_giftexchange.feature
```diff
@@ -716,3 +724,40 @@ Feature: Gift Exchange Challenge
...
+
+  Scenario: Many Assignments.
+    Given the following activated users exist
+      | login  |
+      | recip1 |
+      | recip2 |
+      | recip3 |
+      | recip4 |
+      | recip5 |
+      And there are 3 assignments per page
```
sarken: It would be fine to cut this down to one per page to make this a little shorter.

features/gift_exchanges/challenge_giftexchange.feature
```diff
@@ -716,3 +724,40 @@ Feature: Gift Exchange Challenge
...
+    Given the user "recip" exists and is activated
+      And I am logged in as "gifter"
+      And "gifter" has an assignment for the user "recip" in the collection "exchange_collection"
+    When I go to the assignments page for "gifter"
+    Then I should see "Assignments (1)" within "#dashboard"
+    When I fulfill my assignment
+      And I go to the assignments page for "gifter"
+      Then I should see "Assignments (0)" within "#dashboard"
+
+  Scenario: Many Assignments.
```
sarken: "Scenario: When there are many assignments, they're paginated" or something similar?

app/controllers/challenge_assignments_controller.rb
```diff
@@ -67,21 +67,22 @@ def assignments_not_sent
...
-          @challenge_assignments = @user.offer_assignments.undefaulted + @user.pinch_hit_assignments.undefaulted
-        end
+        @challenge_assignments = @user.assignments.undefaulted
+        @challenge_assignments = @challenge_assignments.in_collection(@collection) if params[:collection_id] && (@collection = Collection.find_by(name: params[:collection_id]))
+        @challenge_assignments = if params[:posted] == "true"
+                                   @challenge_assignments.posted
+                                 else
+                                   @challenge_assignments.unposted
+                                 end
+        @pagy, @challenge_assignments = pagy(@challenge_assignments.order(Arel.sql("COALESCE(sent_at, covered_at) ASC")))
```
sarken: I think it's possible for multiple assignments to have the same `sent_at` or `covered_at`, especially if they're in the same challenge. We should probably add something more deterministic to the sort order as a fallback to avoid any issues with the pagination (e.g., the same item appears on two pages). Descending id would probably be fine.

sarken (changes requested):
Thanks for working on this!

app/models/user.rb
marcus8448: No reason - it's leftover from attempts to get all the assignments without a union.

app/controllers/challenge_assignments_controller.rb
marcus8448: I've sorted by ascending id, since we're sorting by oldest first (per slack). The Jira issue description probably should be updated to match.

sarken: Updated!

sarken (approved):
Thank you!

PR #5335, opened 2025-08-24
AO3-7018 Added rescue to avoid error 500 when entering invalid URL in collection custom header field

katieyang:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-7018

## Purpose

It makes it so that Error 500 gets rescued and it goes straight to the next step, so for example if something like "This will error." is entered, it would pass it to validation, where it would be caught and an error would be displayed to the user on the collections page.

## Testing Instructions

I added this test in collection_spec.rb:

`it "does not raise an 500 error when header_image_url is not a valid URL" do`
`collection.header_image_url = "This will error."`
`expect { collection.save }.not_to raise_error`
`end`

## Credit

katieyang

### DIFF
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -169,6 +169,7 @@ class Collection < ApplicationRecord
 
   def cleanup_url
     self.header_image_url = Addressable::URI.heuristic_parse(self.header_image_url) if self.header_image_url
+  rescue Addressable::URI::InvalidURIError
   end
 
   # Get only collections with running challenges
--- a/spec/models/collection_spec.rb
+++ b/spec/models/collection_spec.rb
@@ -82,6 +82,12 @@ describe Collection do
       expect(collection.errors.full_messages).to \
         include /Sorry, a collection can only have 10 tags./
     end
+
+    it "does not raise an 500 error when header_image_url is not a valid URL" do
+      collection.header_image_url = "This will error."
+      expect { collection.save }.not_to raise_error
+    end
+    
   end
 
   describe "#clear_icon" do


spec/models/collection_spec.rb
```diff
@@ -82,6 +82,12 @@
       expect(collection.errors.full_messages).to \
         include /Sorry, a collection can only have 10 tags./
     end
+
+    it "does not raise an 500 error when header_image_url is not a valid URL" do
```
Bilka2: This test does not test the desired behaviour, which is that this will result in a validation error. Generally this test is a bit in the wrong place because saving a model can never result in an error 500 - error codes are generated int the controller, not the model.

app/models/collection.rb
```diff
@@ -168,6 +168,9 @@ def title
   scope :for_blurb, -> { includes(:parent, :moderators, :children, :collection_preference, owners: [:user]).with_attached_icon }
 
   def cleanup_url
+    plausible_url_regex = %r{\A(?:https?://|www\.|\S+\.\S+)}i
+    return if header_image_url !~ plausible_url_regex
+
```
Bilka2: Thank you for working on this! We'd like to deal with this by rescuing the exception instead of running another regex, e.g. how it was done in #5298

Bilka2 (changes requested):

Bilka2:
Please make sure the invalid URL generates a validation error as described in the Jira issue

app/models/collection.rb
```diff
@@ -167,8 +167,12 @@ def title
   scope :by_title, -> { order(:title) }
   scope :for_blurb, -> { includes(:parent, :moderators, :children, :collection_preference, owners: [:user]).with_attached_icon }
 
-  def cleanup_url
-    self.header_image_url = Addressable::URI.heuristic_parse(self.header_image_url) if self.header_image_url
+  def cleanup_url   
+    self.header_image_url = begin 
+      Addressable::URI.heuristic_parse(self.header_image_url) if self.header_image_url
+    rescue Addressable::URI::InvalidURIError
+      self.header_image_url
```
Bilka2, line 175: This is a good start, however the URL could now pass the other validation because it's not guaranteed to cover everything that `Addressable::URI.heuristic_parse` errors on. That can be solved by using the `url_format` validator for the URL instead of the current validation, which checks the heuristic parse as well as a regex. This is what we do for example in `prompt.rb`.

spec/controllers/collections_controller_spec.rb
```diff
@@ -6,6 +6,15 @@
 
   let(:collection) { create(:collection) }
 
+  describe "POST #create" do
+    context "when the header_image_url is invalid" do
+      it "does not error" do
+        post :create, params: { collection: attributes_for(:collection).merge(header_image_url: "This will error.") }
+        expect(response).to have_http_status(:found)
```
Bilka2: Please also check for the validation error here (it might be good to also adjust the test name to be clearer for what error is and isn't expected)

Bilka2 (changes requested):

PR #5341, opened 2025-08-30
AO3-7110 Fix flaky database seed spec

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-7110

## Purpose

Use WebMock to stub out requests to livejournal/dreamwidth that keep failing, e.g. https://github.com/otwcode/otwarchive/actions/runs/17314839517/job/49155766858. Also flip the `account_creation_enabled` setting as discussed in Slack.

## Testing Instructions

Automated tests should pass.

## Credit

Bilka

### DIFF
--- a/spec/lib/tasks/database_seed.rake_spec.rb
+++ b/spec/lib/tasks/database_seed.rake_spec.rb
@@ -1,6 +1,14 @@
 require "spec_helper"
 
 describe "rake db:fixtures:load" do
+  before do
+    WebMock.stub_request(:any, /example/)
+  end
+
+  after do
+    WebMock.reset!
+  end
+
   it "should result in valid records" do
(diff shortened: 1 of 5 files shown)


spec/lib/tasks/database_seed.rake_spec.rb
```diff
@@ -1,6 +1,14 @@
 require "spec_helper"
 
 describe "rake db:fixtures:load" do
+  before do
+    WebMock.stub_request(:any, /example/)
```
redsummernight: You can update this line in the helper instead:

https://github.com/otwcode/otwarchive/blob/def268313012f6fee9b5c0a0a7a70b8866e3c9c3/spec/spec_helper.rb#L71-L72

PR #5342, opened 2025-08-31
AO3-6167 Fixed that restricted series on user's series page weren't visible to admins

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6167

## Purpose

Change user dashboard sidebar series count and user series page to include restricted series when logged in as admin. 

## Testing Instructions

Refer to Jira.

## Credit

Bilka

### DIFF
--- a/features/other_b/series_lock.feature
+++ b/features/other_b/series_lock.feature
@@ -65,3 +65,22 @@ Feature: Locked and partially locked series
       And I should see "Part 1 of Antiholidays"
     When I press "Update"
     Then I should see "Work was successfully updated"
+
+  Scenario: Admins can see locked series on user's series page
+    Given I am logged in as "Accumulator"
+      And I post the work "Restricted work" as part of a series "So Restrictive"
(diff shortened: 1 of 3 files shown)


features/other_b/series_lock.feature
```diff
@@ -65,3 +65,22 @@ Feature: Locked and partially locked series
       And I should see "Part 1 of Antiholidays"
     When I press "Update"
     Then I should see "Work was successfully updated"
+
+  Scenario: Admins can see locked series on user's series page
+    Given I am logged in as "Accumulator"
+      And I post the work "Restricted work" as part of a series "So Restrictive"
+      And I lock the work "Restricted work"
```
sarken: If we want to avoid the double save, we could do this to check the restricted box before posting:

```
I set up the draft etc etc
I lock the work
I post the work
```

But the phrasing on your version is nicer, so I don't mind either way!

sarken (approved):
I have wanted this for so looooong. Thanks!

PR #5350, opened 2025-09-08
AO3-7114 Update password reset error messages and redirects when logged out

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-7114

## Purpose

Update the errors shown when accessing the change password form a reset your password link with the link being malformed in some way.

## Credit

Bilka

### DIFF
--- a/app/controllers/users/passwords_controller.rb
+++ b/app/controllers/users/passwords_controller.rb
@@ -31,6 +31,20 @@ class Users::PasswordsController < Devise::PasswordsController
     super
   end
 
+  def edit
+    original_token = params[:reset_password_token]
+    reset_password_token = Devise.token_generator.digest(self, :reset_password_token, original_token)
+
+    user = User.find_for_authentication(reset_password_token: reset_password_token)
+
+    if user.nil? || user.new_record?
+      flash[:error] = t(".invalid_link") + helpers.tag.br + t(".request_again", count: ArchiveConfig.PASSWORD_RESET_COOLDOWN_HOURS)
+      redirect_to new_user_password_path and return
+    end
+
+    super
+  end
+
   protected
 
   # We need to include information about the user (the remaining reset attempts)
@@ -50,5 +64,12 @@ class Users::PasswordsController < Devise::PasswordsController
   def after_resetting_password_path_for(resource)
     resource.create_log_item(action: ArchiveConfig.ACTION_PASSWORD_RESET)
     super
+  end
+
+  def assert_reset_token_passed
+    return if params[:reset_password_token].present?
+
+    set_flash_message(:error, :no_token)
+    redirect_to new_user_password_path
   end
 end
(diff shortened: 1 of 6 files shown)


features/users/authenticate_users.feature
```diff
@@ -58,11 +58,8 @@ Feature: User Authentication
     # password reset link should no longer work
     When I log out
       And I follow "Change my password." in the email
-      And I fill in "New password" with "override"
-      And I fill in "Confirm new password" with "override"
-      And I press "Change Password"
-    Then I should see "We couldn't save this user because:"
-      And I should see "Reset password token is invalid"
+    Then I should see "This password reset link is invalid or expired. Please check your email for the most recent password reset link."
+      And I should see "Forgotten your password?"
```
sarken: Could we replace this with `I should be on the new user password page`? (If you wanted to rename it to something like "the forgot password page" for clarity, this is the only file it's used in.)

app/controllers/users/passwords_controller.rb
```diff
@@ -31,6 +31,20 @@ def create
     super
   end
 
+  def edit
+    original_token = params[:reset_password_token]
```
sarken: This is clear to me right now because I just read the issue and know the context, but in general, I think "original_token" would make me expect that this was always going to correspond with the original, correct token we sent the user. It wouldn't immediately occur to me the token in the URL might have been mangled.

Could we add an explanatory comment? Just something above the variable like, "The token from the URL, which might be expired or incorrect (e.g., due to copy/paste error)." Or even something above the whole method explaining we want to give a friendlier error in case the token in the URL has a typo/is expired.

config/locales/models/en.yml
```diff
@@ -212,6 +212,8 @@ en:
               invalid: must be %{min_login} to %{max_login} characters (A-Z, a-z, _, 0-9 only), no spaces, cannot begin or end with underscore (_).
             password_confirmation:
               confirmation: doesn't match new password.
+            reset_password_token:
+              invalid: This password reset link is invalid or expired. Please request a new password reset instead.
```
sarken, line 216: Just checking -- is this a giant pain to test?

lydia-theda: You should be able to get this message by entering a fake keysmash (or just clicking the link in the issue since KEYSMASH isn't a valid token), rather than waiting 7 days. At least, I'm pretty sure there's no reason to test actual expiry since that behavior isn't changing?

sarken: I meant an automated test -- there isn't one for this message, which is the error that's given when the token expires after you've successfully used it to access `/users/password/edit` but before you submit the form. (The fake keysmash scenario you mentioned gets the `no_token` message above and is covered in the test on line 29 of the specs.)

Bilka2: Yes, it is a pain. What devise does is redirect to the edit page, without any url params. So trying to check for the error in the test just gives back the blank token error (due to the lack of url params), instead of the error we're trying to check for. Theoretically I could set up a cucumber test where the token expires at the right time but that doesn't feel worth it for something that is just a locale change

sarken: Okay, cool. The cuke was the only solution I could think of too and it does seem like overkill, but I just wanted to double check in case it was an oversight.

sarken (approved):

PR #5354, opened 2025-09-11
AO3-7002 Set up skin chooser skins in development 

weeklies:
## Issue

https://otwarchive.atlassian.net/browse/AO3-7002 

## Purpose

Rewrite and use <s>load_user_skins</s>`load_official_skins` during the initialization process.

* Use File.exist? over removed File.exists?
* Update README
* Add support for parent_only skins through parent and top level skin subdirectories 
* Add support for `/* MEDIA: */` and `/* MEDIA: ENDMEDIA */`
* Set up Basic Formatting work skin as well

To reduce complexity, the following unused features are removed:
* Remove support for `/* REPLACE: */`
* Remove support for multiple skin declarations using `/* END SKIN */`

This does not change production behavior.

### DIFF
--- a/spec/lib/tasks/skin_tasks.rake_spec.rb
+++ b/spec/lib/tasks/skin_tasks.rake_spec.rb
@@ -37,15 +37,12 @@ end
 describe "rake skins:load_official_skins" do
   before do
     allow($stdin).to receive(:gets).and_return("n")
-
-    allow(Skin).to receive(:skin_dir_entries).with(user_skin_path, anything).and_return(["test_skin.css", "child_skin.css"])
-    allow(File).to receive(:read).with(/.*css/).and_return(
+    allow(File).to receive(:read).with(/.*css/).and_return("")
+    allow(File).to receive(:read).with(/top_level/).and_return(
       "/* SKIN: Test Skin */",
       "/* SKIN: Child */\n/* PARENTS: Parent */"
     )
-
-    allow(Skin).to receive(:skin_dir_entries).with(parent_skin_path, anything).and_return(["parent_skin.css"])
-    allow(File).to receive(:read).with(/.*parent_skin\.css/).and_return("/* SKIN: Parent */\n#unused-selector { content: none; },")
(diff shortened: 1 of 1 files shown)


app/helpers/application_helper.rb
```diff
@@ -160,7 +160,7 @@ def link_to_help(help_entry, link = '<span class="symbol question"><span>?</span
     #  help_file = "#{ArchiveConfig.HELP_DIRECTORY}/#{Locale.active.language.code}/#{help_entry}.html"
     #end
 
-    unless !help_file.blank? && File.exists?("#{Rails.root}/public/#{help_file}")
+    unless help_file.present? && File.exist?("#{Rails.root}/public/#{help_file}")
```
weeklies, line 163: This wasn't part of the issue, but I thought fixing the broken `File.exists?` here too would be helpful.

public/stylesheets/README.md
```diff
@@ -21,3 +21,5 @@ process as all other code changes.
 Remember, though: changes to a skin's files in the masters directory will not
 affect the skin on production! An admin will still need to edit the skin via
 the admin interface and paste in the updated master version.
+
+**Note**: To keep development data in sync with the master copies, please ensure any skin changes are mirrored in `site/user_skins_to_load`.
```
sarken: This isn't a good idea -- it's simply too easy for things to get out of sync. We need an approach where we will only have one copy of each skin and then load them into the database from there. It's okay if that means adding code comments to the master copies, renaming the files, and/or moving them to a different directory, as long as we also keep the subdirectory structure. 

(I don't think we need to stick with the `user_skins_to_load` name, either. It's an unintuitive name that makes it sound like we're loading skins for individual users rather than loading the officially supported skins for the chooser.)

sarken:
We discussed this a little and we'd like to take a different approach to this problem. Instead of putting information about the parents, media, and so on in the stylesheets and trying to parse it out with regular expressions, we'd like to move all of that information to the fixtures. Then the task can just be concerned with updating the skins' `css` fields to contain the CSS from the files.

weeklies:
Due to extenuating factors, I'm going to leave this PR for adoption. I'll unassign myself from the issue.

PR #5234, opened 2025-06-26
AO3-6026 Move collections to Elasticsearch

brianjaustin:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6026

## References

Continuation of #3912 (there were a lot of comments and a lot of merge issues, so we decided it was easier to do a new PR)

## Credit

Littlelines (contractor)
Brian Austin (they/he)

PR #5355, opened 2025-09-15
AO3-7122 Change nil to false for collections' open_doors and multifandom columns

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-7122

## Purpose

Change any `nil` values for collections' `open_doors` and `multifandom` columns to `false` and then prompt us to reindex collections.

When we run this on production, all collections will need to be updated because they'll all be `open_doors: nil`. Therefore, we can just run the task to reindex all collections at the end. If we want to make this more robust in case we accidentally rerun it at a future time, I can undo the last two commits and enqueue collections individually instead.

## References

This should go in after #5234, which has the referenced reindexing task.


db/migrate/20250906065547_change_collection_multifandom_and_open_doors_default.rb
```diff
@@ -0,0 +1,8 @@
+class ChangeCollectionMultifandomAndOpenDoorsDefault < ActiveRecord::Migration[7.1]
+  uses_departure! if Rails.env.staging? || Rails.env.production?
+
+  def change
+    change_column_default :collections, :multifandom, from: nil, to: false
+    change_column_default :collections, :open_doors, from: nil, to: false
```
brianjaustin: Should we make these both not nullable with default of false? That would require running the rake task first, but I think it would be nicer to not let weird values come back in

lib/tasks/after_tasks.rake
```diff
@@ -579,5 +579,25 @@ namespace :After do
       end
     end
   end
+
+  desc "Change nil to false for multifandom and open_doors collection columns"
+  task(update_collections_multifandom_and_open_doors: :environment) do
+    total_collections = Collection.where("multifandom IS NULL OR open_doors IS NULL").size
```
Bilka2: The db will not like this (the explain shows that it's over 380k rows and took long enough to make me worried about actually running the query). I think we can just skip the total count

lib/tasks/after_tasks.rake
```diff
@@ -579,5 +579,25 @@ namespace :After do
...
+    total_batches = (total_collections + 999) / 1000
+    puts "Updating #{total_collections} collections in #{total_batches} batches"
+
+    Collection.where("multifandom IS NULL OR open_doors IS NULL").find_in_batches.with_index do |batch, index|
```
Bilka2: Same with this query. I think it will be easy on the db to get all in batches and filter in Ruby than to make the db look so many rows for the where condition

db/migrate/20250906065547_change_collection_multifandom_and_open_doors_default.rb
Bilka2: Noting this [from the docs](https://api.rubyonrails.org/v8.0.2.1/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-change_column_null) that I think we should use: 
> The method accepts an optional fourth argument to replace existing NULLs with some other value. Use that one when enabling the constraint if needed, since otherwise those rows would not be valid.

lib/tasks/after_tasks.rake
sarken: Gladly! I kept going back and forth on which to do because, in our case, we'll be updating every collection. I just decided to commit the more involved version because I figured it would be easier for me to switch to the simpler version if asked than vice versa.

Bilka2 (approved):

PR #5356, opened 2025-09-16
AO3-7123 Add filter for multifandom collections

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-7123

## Purpose

Adds a Multifandom section to the collection filters with yes/no/either options.

### DIFF
--- a/config/locales/views/en.yml
+++ b/config/locales/views/en.yml
@@ -645,10 +645,10 @@ en:
       title: Filter by title
       type:
         any: Any
-        gift_exchange: Gift Exchange
+        gift_exchange: Gift Exchange Challenge
         label: Collection Type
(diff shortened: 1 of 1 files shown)


config/locales/views/en.yml
```diff
@@ -612,6 +612,43 @@ en:
   collections:
     collection_blurb:
       collection_tags: 'Collection Tags:'
+    filters:
+      clear_filters: Clear Filters
+      closed:
+        exclude: 'No'
+        include: Either
+        label: Closed
+        only: 'Yes'
```
sarken, line 621: I thought about making one reusable set of Yes/No/Either options, but we do need to switch to better, unique labels at some point, and this will make doing so easier.

PR #5363, opened 2025-09-18
AO3-6026 Fix collection tag backfill issues

brianjaustin:
Fixes the following problems with backfilling collection tags:
* series tags are missing
* public bookmark of hidden work leads to the work tags being included
* public bookmark of restricted work leads to the work tags being included
* tags of unrevealed works are included

### DIFF
--- a/app/models/bookmark.rb
+++ b/app/models/bookmark.rb
@@ -83,7 +83,7 @@ class Bookmark < ApplicationRecord
   # users (regardless of the bookmark's hidden_by_admin/private status).
   scope :with_bookmarkable_visible_to_all, -> {
     join_bookmarkable.where(
-      "(works.posted = 1 AND works.restricted = 0 AND works.hidden_by_admin = 0) OR
+      "(works.posted = 1 AND works.restricted = 0 AND works.hidden_by_admin = 0 AND works.in_unrevealed_collection = 0) OR
       (series.restricted = 0 AND series.hidden_by_admin = 0) OR
       (external_works.hidden_by_admin = 0)"
     )
--- a/app/models/bookmark.rb
+++ b/app/models/bookmark.rb
@@ -83,7 +83,7 @@ class Bookmark < ApplicationRecord
   # users (regardless of the bookmark's hidden_by_admin/private status).
   scope :with_bookmarkable_visible_to_all, -> {
     join_bookmarkable.where(
-      "(works.posted = 1 AND works.restricted = 0 AND works.hidden_by_admin = 0 AND works.in_unrevealed_collection = 0) OR
+      "(works.posted = 1 AND works.restricted = 0 AND works.hidden_by_admin = 0) OR
       (series.restricted = 0 AND series.hidden_by_admin = 0) OR
       (external_works.hidden_by_admin = 0)"
     )
--- a/lib/tasks/after_tasks.rake
+++ b/lib/tasks/after_tasks.rake
@@ -586,7 +586,7 @@ namespace :After do
     total_batches = (collections.count + 999) / 1000
 
     def approved_taggables(collection)
-      bookmark_visible_revealed = Bookmark.join_bookmarkable.where(
+      bookmark_visible_revealed = Bookmark.is_public.join_bookmarkable.where(
         "(works.posted = 1 AND works.restricted = 0 AND works.hidden_by_admin = 0 AND works.in_unrevealed_collection = 0) OR
         (series.restricted = 0 AND series.hidden_by_admin = 0) OR
         (external_works.hidden_by_admin = 0)"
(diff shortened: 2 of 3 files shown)


lib/tasks/after_tasks.rake
```diff
@@ -586,16 +586,16 @@ namespace :After do
...
+        Bookmark.visible_to_all.in_collection(collection).includes(:fandoms) + \
+        Bookmark.visible_to_all.in_collection(collection).includes(bookmarkable: :fandoms).filter_map { |bookmark| bookmark.bookmarkable unless bookmark.bookmarkable.unrevealed? }
     end
 
     Collection.no_touching do
       collections.find_in_batches.with_index do |batch, index|
         batch.each do |collection|
           tags = approved_taggables(collection)
-            .flat_map { |taggable| taggable.try(:fandoms) || [] }
+            .flat_map { |taggable| taggable.try(:fandoms) || taggable.try(&:work_tags)&.where(type: "Fandom") || [] }
```
Bilka2: both work (apparently!) but this is cleaner:

```suggestion
            .flat_map { |taggable| taggable.try(:fandoms) || taggable.try(:work_tags)&.where(type: "Fandom") || [] }
```

lib/tasks/after_tasks.rake
```diff
@@ -586,16 +586,16 @@ namespace :After do
     total_batches = (collections.count + 999) / 1000
 
     def approved_taggables(collection)
-      Work.visible_to_all.in_collection(collection).includes(:fandoms) + \
-        Bookmark.is_public.in_collection(collection).includes(:fandoms) + \
-        Bookmark.is_public.in_collection(collection).includes(bookmarkable: :fandoms).map(&:bookmarkable)
+      Work.visible_to_all.revealed.in_collection(collection).includes(:fandoms) + \
+        Bookmark.visible_to_all.in_collection(collection).includes(:fandoms) + \
+        Bookmark.visible_to_all.in_collection(collection).includes(bookmarkable: :fandoms).filter_map { |bookmark| bookmark.bookmarkable unless bookmark.bookmarkable.unrevealed? }
```
Bilka2: I think `unrevealed?` isn't available on external works

app/models/bookmark.rb
```diff
@@ -83,7 +83,7 @@ def check_new_external_work
   # users (regardless of the bookmark's hidden_by_admin/private status).
   scope :with_bookmarkable_visible_to_all, -> {
     join_bookmarkable.where(
-      "(works.posted = 1 AND works.restricted = 0 AND works.hidden_by_admin = 0) OR
+      "(works.posted = 1 AND works.restricted = 0 AND works.hidden_by_admin = 0 AND works.in_unrevealed_collection = 0) OR
```
Bilka2: I think this change will affect too many things around the archive. Could you just copy paste the modified visible_to_all scope for the task and revert this here?

spec/lib/tasks/after_tasks.rake_spec.rb
```diff
@@ -708,27 +708,97 @@
...
+      end
+
+      it_behaves_like "does not tag the collection"
+    end
   end
 
   context "when a collection has work bookmarks" do
     let(:fandom) { create(:canonical_fandom) }
-    let(:items) { [create(:bookmark, tag_string: fandom.name)] }
+    let(:items) { [create(:bookmark, fandom_string: fandom.name)] }
```
Bilka2: in the interface bookmarks dont have fandom tags, only general unsorted tags. the previous code was correct

spec/lib/tasks/after_tasks.rake_spec.rb
```diff
@@ -708,27 +708,97 @@
...
+    end
   end
 
   context "when a collection has a series bookmark" do
     let(:fandom) { create(:canonical_fandom) }
     let(:bookmark) { create(:series_bookmark, tag_string: fandom.name) }
     let(:items) { [bookmark] }
 
-    it "only includes the bookmark's fandoms" do
+    it "includes the bookmark's and serie's fandoms" do
```
Bilka2:
```suggestion
    it "includes the bookmark's and series's fandoms" do
```

app/models/bookmark.rb
brianjaustin: Fair. I think this version is wrong, but we need more time to figure out the other consequences (which I do not think we should spend now)

lib/tasks/after_tasks.rake
```diff
@@ -586,16 +586,21 @@ namespace :After do
     total_batches = (collections.count + 999) / 1000
 
     def approved_taggables(collection)
-      Work.visible_to_all.in_collection(collection).includes(:fandoms) + \
-        Bookmark.is_public.in_collection(collection).includes(:fandoms) + \
-        Bookmark.is_public.in_collection(collection).includes(bookmarkable: :fandoms).map(&:bookmarkable)
+      bookmark_visible_revealed = Bookmark.join_bookmarkable.where(
```
Bilka2:
```suggestion
      bookmark_visible_revealed = Bookmark.is_public.join_bookmarkable.where(
```

Bilka2 (changes requested):

PR #5366, opened 2025-09-20
AO3-6026 Cover all Collection blurb count steps by automated tests

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6026

## Purpose

Cover the testing steps for Collection blurb Works count and Collection blurb Bookmarked Items count with automatic tests instead because the blurb caching interacts so badly with updates that manual testing is not feasible. (Problem is that the blurb cache updates with info from ES before ES updates for the changes, so the stale data is cached for the whole blurb cache expiry time which is up to 2 hours).

Works:
Counts general vs public: public work, restricted work, hidden work, rejected work all in collection_spec.rb
Updating counts:
* adding work: created in collection_item_spec.rb and work_spec.rb
* restricting, hiding: work_spec.rb. unhiding code is same as hiding so no need to test
* approving, rejecting: collection_item_spec.rb
* removing work: destroyed collection item in collection_item_spec.rb
* destroying work: destroyed work in work_spec.rb

Count display for user/guest/admin: collection_browse.feature:240

Bookmarks:
Counts general vs public: public bookmark, private bookmark, hidden bookmark, rejected bookmark, bookmark of restricted work, bookmark of hidden work, bookmark of public work all in collection_spec.rb
Updating counts:
* adding bookmark: created in collection_item_spec.rb and bookmark_spec.rb
* privating, hiding: bookmark_spec.rb. unhiding code is same as hiding so no need to test
* approving, rejecting: collection_item_spec.rb
* hiding bookmarked work, restricting bookmarked work, deleting bookmarked work: bookmarkable.spec
* removing bookmark: destroyed collection item in collection_item_spec.rb
* destroying bookmark: destroyed work in bookmark_spec.rb

Count display for user/guest/admin: collection_browse.feature:240

## Credit

Bilka

### DIFF
--- a/spec/models/bookmark_spec.rb
+++ b/spec/models/bookmark_spec.rb
@@ -46,4 +46,58 @@ describe Bookmark do
     expect(activity.save).to be_truthy
     expect(activity.target_name).to eq("Bookmark #{bookmark.id}")
   end
+
+  context "reindexing" do
+    let!(:parent_collection) { create(:collection) }
+    let!(:collection) { create_invalid(:collection, parent: parent_collection) }
+
+    context "when bookmark is created in collection" do
+      it "enqueues the collection for reindex" do
+        expect do
+          create(:bookmark, collections: [collection])
+        end.to add_to_reindex_queue(collection, :background) &
+               add_to_reindex_queue(parent_collection, :background)
+      end
+    end
+
+    context "bookmark already exists in the collection" do
+      let!(:bookmark) { create(:bookmark, collections: [collection]) }
+
+      context "when bookmark is hidden by an admin" do
+        it "enqueues its collection for reindex" do
+          expect do
+            bookmark.update!(hidden_by_admin: true)
+          end.to add_to_reindex_queue(collection, :background) &
+                 add_to_reindex_queue(parent_collection, :background)
+        end
+      end
+
+      context "when bookmark is privated" do
+        it "enqueues its collection for reindex" do
+          expect do
+            bookmark.update!(private: true)
+          end.to add_to_reindex_queue(collection, :background) &
+                 add_to_reindex_queue(parent_collection, :background)
+        end
+      end
+
+      context "when bookmark is not significantly changed" do
+        it "doesn't enqueue its collection for reindex" do
+          expect do
+            bookmark.touch
+          end.to not_add_to_reindex_queue(collection, :background) &
+                 not_add_to_reindex_queue(parent_collection, :background)
+        end
+      end
+
+      context "when bookmark is destroyed" do
+        it "enqueues its collection for reindex" do
+          expect do
+            bookmark.destroy!
+          end.to add_to_reindex_queue(collection, :background) &
+                 add_to_reindex_queue(parent_collection, :background)
+        end
+      end
+    end
+  end
 end
--- a/spec/models/bookmark_spec.rb
+++ b/spec/models/bookmark_spec.rb
@@ -60,7 +60,7 @@ describe Bookmark do
       end
     end
 
-    context "bookmark already exists in the collection" do
+    context "when bookmark already exists in the collection" do
       let!(:bookmark) { create(:bookmark, collections: [collection]) }
 
       context "when bookmark is hidden by an admin" do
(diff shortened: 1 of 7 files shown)


spec/models/bookmark_spec.rb
```diff
@@ -46,4 +46,58 @@
...
+    context "when bookmark is created in collection" do
+      it "enqueues the collection for reindex" do
+        expect do
+          create(:bookmark, collections: [collection])
+        end.to add_to_reindex_queue(collection, :background) &
+               add_to_reindex_queue(parent_collection, :background)
+      end
+    end
+
+    context "bookmark already exists in the collection" do
```
sarken:
```suggestion
    context "when bookmark already exists in the collection" do
```

spec/models/collection_item_spec.rb
```diff
@@ -110,23 +112,37 @@
         it "enqueues the collection for reindex" do
           expect do
             item.update!(collection_approval_status: :rejected)
-          end.to add_to_reindex_queue(collection, :background)
+          end.to add_to_reindex_queue(collection, :background) &
+                 add_to_reindex_queue(parent_collection, :background)
+        end
+      end
+
+      context "when collection item is accepted" do
```
sarken: I think we're missing tests for changing the `user_approval_status`. We could maybe also use one for setting `collection_approval_status` to `:unreviewed`, but that's probably not as important as a test that confirms unreviewed items aren't included in the counts.

spec/models/collection_spec.rb
```diff
@@ -165,6 +165,17 @@
       it_behaves_like "does not count the work"
     end
 
+    context "when the collection includes a rejected public work" do
+      let(:work) { create(:work) }
+
+      before do
+        work.collections << collection
+        work.collection_items.update!(collection_approval_status: :rejected)
```
sarken, line 173: Should we also test an item where `user_approval_status: :rejected`? It would probably also be a good idea to test the `:unreviewed` status for one or both approval types, since unreviewed items should not be counted either.

spec/models/collection_spec.rb
```diff
@@ -315,6 +381,45 @@
...
+
+      before do
+        bookmark.collections << collection
+        bookmark.collection_items.update!(collection_approval_status: :approved, user_approval_status: :unreviewed)
+      end
+
+      it "does not count the bookmark" do
+        expect(collection.general_bookmarked_items_count).to eq(0)
+      end
+    end
```
sarken: This is definitely possible to do if the bookmarker changes the bookmark's status via their /collection_items page after adding it to an open collection, but since there's no direct way to invite a bookmark to a collection, maybe we should test the opposite, more common case: `collection_approval_status: :unreviewed, user_approval_status: :approved`, which occurs when a user adds a bookmark to a moderated collection and it's awaiting approval.

spec/models/search/collection_indexer_spec.rb
```diff
@@ -84,6 +84,23 @@
...
+      let(:work) { create(:work) }
+      let(:bookmark) { create(:bookmark, bookmarkable: work) }
+      let(:bookmark2) { create(:bookmark, bookmarkable: work, private: true) }
+
+      before do
+        bookmark.collections << collection
+        bookmark2.collections << collection
+      end
+
+      it "is counts as one item" do
```
sarken:
```suggestion
      it "counts as one item" do
```

I think this was a copy-pasted typo from the example above if you want to fix that one too.

spec/models/collection_spec.rb
```diff
@@ -315,6 +381,45 @@
...
+        bookmark.collections << collection
+        bookmark.collection_items.update!(user_approval_status: :rejected)
+      end
+
+      it "does not count the bookmark" do
+        expect(collection.general_bookmarked_items_count).to eq(0)
+      end
+    end
+
+    context "when the collection contains an invited public bookmark" do
```
sarken:
```suggestion
    context "when the collection contains a public bookmark awaiting mod approval" do
```
Or some other, better phrasing that reflects the changes to the bookmark tests. 

Once these are updated, anyone can hit the big green merge button. I've got to go to sleep.

PR #5367, opened 2025-09-21
AO3-7128 Collection maintainers listed twice in subcollection blurbs and profiles if they also have a different role in the parent collection

Cubostar:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-7128

## Purpose

This PR removes duplicate maintainers listed in subcollection blurbs and profiles if they have a different role in the parent collection.

## Testing Instructions

Testing is explained in the Jira ticket under the section "Steps to reproduce". Following these steps and checking that there are no duplicates and the classes of the maintainers are correct as described in the section "What should happen" should suffice.

## Credit

Cubostar, he/him

### DIFF
--- a/app/views/collection_profile/show.html.erb
+++ b/app/views/collection_profile/show.html.erb
@@ -27,7 +27,7 @@
       <dt class="maintainers"><%= ts('Maintainers:') %></dt>
       <dd class="maintainers">
         <ul class="commas">
-          <%= safe_join(@collection.maintainers.map { |maintainer| content_tag(:li, link_to(maintainer.byline, maintainer.user), class: @collection.user_is_owner?(maintainer.user) ? "owner" : "mod") }, "\n") %>
+          <%= safe_join(@collection.maintainers.uniq.map { |maintainer| content_tag(:li, link_to(maintainer.byline, maintainer.user), class: @collection.user_is_owner?(maintainer.user) ? "owner" : "mod") }, "\n") %>
         </ul>
       </dd>
       <% unless @collection.email.blank? %>
--- a/app/views/collections/_collection_blurb.html.erb
+++ b/app/views/collections/_collection_blurb.html.erb
@@ -8,7 +8,7 @@
         <span class="name">(<%= collection.name %>)</span>
         <%= ts('by') %>
         <%= safe_join(
-          collection.maintainers.map { |maintainer| link_to(maintainer.byline, maintainer.user, class: collection.user_is_owner?(maintainer.user) ? "owner" : "mod")}, t("support.array.words_connector")) %>
+          collection.maintainers.uniq.map { |maintainer| link_to(maintainer.byline, maintainer.user, class: collection.user_is_owner?(maintainer.user) ? "owner" : "mod")}, t("support.array.words_connector")) %>
       </h4>
       <!--collections header iconised header image-->
       <div class="icon">


features/collections/collection_participants.feature
```diff
@@ -67,6 +67,20 @@
...
+      And I have the collection "Such a nice collection" with name "Collection"
+      And I am logged in as the owner of "Such a nice collection"
+      And I have added the co-owner "sam" to collection "Such a nice collection"
+      And I add the subcollection "Such a nice subcollection" to the parent collection named "Collection"
+      And I have added the co-moderator "sam" to collection "Such a nice subcollection"
+    When I view the collection "Such a nice subcollection"
+      And I follow "Profile"
+    Then I should see "sam" exactly 1 time 
+    When I view the collection "Such a nice collection"
+      And I follow "Subcollections"
```
sarken: Just like you suspected, collections aren't being indexed before we access this page, which prevents us from finding the blurb. 

We _could_ just add the step "And all indexing tasks have been run" in this scenario and call it a day, but even better would be a solution that will work for anyone else who wants to test the Subcollections page in the filter.

If you check out the paths we use in Cucumber tests, you'll notice we added an indexing step for pages that come from Elasticsearch: https://github.com/otwcode/otwarchive/blob/48abf3174c5d2c5caf0e7ca9419f7674081eab12/features/support/paths.rb#L168-L170

That guarantees that whenever we use the step "I go to {page}", the indexing task will run first and the information will be updated.

Let's do that for a collection's subcollections page. We can also add one (without the indexing step) for the collection profile page so we can use that instead of accessing the collection and then following the "Profile" link in the sidebar. It will make things just a tiny bit shorter and more direct.

features/collections/collection_participants.feature
```diff
@@ -67,6 +67,20 @@
   Then I should see "You are now a member of Such a nice collection"
   When I am in the default browser
 
+  Scenario: A subcollection profile and blurb does not show duplicates when a moderator is also an owner of the parent collection
+    Given a user exists with login: "sam"
+      And I have the collection "Such a nice collection" with name "Collection"
+      And I am logged in as the owner of "Such a nice collection"
+      And I have added the co-owner "sam" to collection "Such a nice collection"
+      And I add the subcollection "Such a nice subcollection" to the parent collection named "Collection"
```
sarken: This is very nitpicky, but I think it would be easier to follow what is going on if the collection titles weren't so long and so similar. Just "Subcollection" and "Collection" might work. (If you do that, you can also leave off the optional "with name" part of the first step. The way that step is set up, it will automatically have the name "Collection.")

features/step_definitions/collection_steps.rb
```diff
@@ -115,6 +115,21 @@
   step %{I should see "Updated #{name}"}
 end
 
+Given /^I have added (?:a|the) co\-owner "([^\"]*)" to collection "([^\"]*)"$/ do |name, title|
```
sarken: `Given "I have added (a)(the) co-owner {string} to collection {string}" do |name, title|` will probably work here and make Reviewdog happy.

(If you want to get really ambitious, you could check if the test suite still passes when making the same change to the co-moderator step definition this is based on. That's totally optional, though.)

features/collections/collection_participants.feature
Cubostar: Ah, I didn't notice those steps, thanks!

features/collections/collection_participants.feature
Cubostar: Sounds good!

features/step_definitions/collection_steps.rb
Cubostar: I changed the step syntax for both the co-owner and co-moderator case, thanks for the help!

features/step_definitions/collection_steps.rb
```diff
@@ -100,10 +100,10 @@
   step %{I am logged out}
 end
 
-Given /^I have added (?:a|the) co\-moderator "([^\"]*)" to collection "([^\"]*)"$/ do |name, title|
+Given "I have added (a)(the) co-moderator {string} to collection {string}" do |name, title|
```
Bilka2: Just a quick note, this should work for these step names too and is a bit nicer to read:
```suggestion
Given "I have added a/the co-moderator {string} to collection {string}" do |name, title|
```

Cubostar: Got it!

features/collections/collection_participants.feature
```diff
@@ -67,6 +67,20 @@
   Then I should see "You are now a member of Such a nice collection"
   When I am in the default browser
 
+  Scenario: A subcollection profile and blurb does not show duplicates when a moderator is also an owner of the parent collection
```
sarken: Little grammatical nitpick: 
```suggestion
  Scenario: A subcollection profile and blurb do not show duplicates when a moderator is also an owner of the parent collection
```

Cubostar: Added, thanks!

PR #5289, opened 2025-07-27
AO3-7086 Update password form, headers, error messages

AbhinavGupta-de:
Took 23 minutes

# Pull Request Checklist

* [x] Have you read ["How to write the perfect pull request"](https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/)?
* [x] Have you read the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [x] Have you added the [Jira](https://otwarchive.atlassian.net) issue number
  as the *first* thing in your pull request title (e.g. `AO3-1234 Fix thing`)
* [x] Do you have fewer than 5 pull requests already open? If not, please wait
  until they are reviewed and merged before creating new pull requests.

## Issue

https://otwarchive.atlassian.net/browse/AO3-7086

## Purpose

- Updated the headers for username and password page to remove "my"
- moved the old password field at the top to change password
- updated the error messages on changing username and password

## Credit

Abhinav, he/him

PR #5368, opened 2025-09-21
[AO3-7113]: Update new password form

Tani-Shah:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-7113

## Purpose

What does this PR do?
Updates the edit password form to more closely match the logged-in version of the form.

## Credit

Tani
She/Her

### DIFF
--- a/app/views/users/change_password.html.erb
+++ b/app/views/users/change_password.html.erb
@@ -1,5 +1,5 @@
 <!--Descriptive page name, messages and instructions-->
-<h2 class="heading"><%= ts("Change My Password") %></h2>
+<h2 class="heading"><%= ts("Change Password") %></h2>
 <%= error_messages_for :user %>
 <!--/descriptions-->
 
--- a/app/views/users/passwords/edit.html.erb
+++ b/app/views/users/passwords/edit.html.erb
@@ -1,5 +1,5 @@
 <!--Descriptive page name, messages and instructions-->
-<h2 class="heading"><%= ts("Change My Password") %></h2>
+<h2 class="heading"><%= ts("Change Password") %></h2>
 <%= error_messages_for resource %>
 <!--/descriptions-->
 
--- a/config/locales/models/en.yml
+++ b/config/locales/models/en.yml
@@ -211,7 +211,7 @@ en:
                 other: can only be changed once every %{count} days. You last changed your username on %{renamed_at}.
               invalid: must be %{min_login} to %{max_login} characters (A-Z, a-z, _, 0-9 only), no spaces, cannot begin or end with underscore (_).
             password_confirmation:
-              confirmation: doesn't match new password.
+              confirmation: The passwords you entered do not match. Please try again.
             terms_of_service:
               accepted: Sorry, you need to accept the Terms of Service in order to sign up.
               format: "%{message}"
--- a/config/locales/devise/en.yml
+++ b/config/locales/devise/en.yml
@@ -42,7 +42,7 @@ en:
       no_token: You can't access this page without coming from a password reset email. If you do come from a password reset email, please make sure you used the full URL provided.
       send_instructions: Check your email for instructions on how to reset your password.
       send_paranoid_instructions: If your email address exists in our database, you will receive a password recovery link at your email address in a few minutes.
-      updated: Your password has been changed successfully. You are now signed in.
+      updated: Your password has been changed. You are now logged in.
       updated_not_active: Your password has been changed successfully.
     registrations:
       destroyed: Bye! Your account has been successfully cancelled. We hope to see you again soon.


sarken:
Hi, Tani!

Thanks for working on this! When it's ready for review, or if you run into issues and want some help, just take it out of draft status and someone will come along to look it over. 

In the meantime, I've updated the Jira issue status to In Review to help make sure so no one mistakenly creates a duplicate pull request. If you'd like the ability to comment on, assign, and transition issues in the future, you're welcome to create a Jira account! It makes things a bit easier for us on the organizational side if the Full Name on your Jira account either closely matches the name you'd like us to credit in the release notes or includes it in parentheses, e.g. "Nickname (CREDIT NAME)."

Tani-Shah:
Hi! I need some help with this. 
The search test is failing even with nothing changed in that file or a related file.
For the linting - I can make the changes to use t instead of ts but that feels like its out of scope for this ticket & should be another PR of its own.
Thoughts?

sarken:
Hi again!

The search test failures aren't your fault -- they're just flaky tests. I hit the rerun button and they passed.

Unless a pull request is fairly complex, we try to make the change from `ts()` to t`()` whenever we touch a string. (Sometimes we even do it even if we're not editing strings and just happen to be in the neighborhood.) Since this is a pretty small pull request, we'd appreciate it if you could make the change. We usually use the key `page_heading` for `<h2>` page names like these.

If you'd like any more information, our wiki has a page on our [i18n Standards](https://github.com/otwcode/otwarchive/wiki/Internationalization-(i18n)-Standards).

Tani-Shah:
Thank you so much for all your help!!

Tani-Shah:
I've also created a JIRA account!!
email: its.tanishah@gmail.com
name: Tani Shah

app/views/users/passwords/edit.html.erb
```diff
@@ -1,27 +1,27 @@
...
   <%= f.hidden_field :reset_password_token %>
   <dl>
-    <dt><%= label_tag :edit_password, ts("New password") %></dt>
+    <dt><%= label_tag :edit_password, t(".new_password") %></dt>
     <dd>
       <%= f.password_field(:password, id: :edit_password,
                            "aria-describedby" => "password-field-description") %>
       <p class="footnote" id="password-field-description">
-        <%= ts("%{minimum} to %{maximum} characters",
+        <%= t(".password_requirements",
```
Hamham6: General question to everyone, not directed towards you specifically (altho of course if you have any thoughts feel free to chime in!): Should this have a `count` variable for pluralization? My first thought is yes, but I couldn't actually find how to pluralize ranges of numbers, ex whether to pluralize according to the larger number/last number/difference between numbers/something else. I found that  the unicode localization data technically collects this information (https://unicode.org/reports/tr35/tr35-numbers.html#Plural_Ranges), but in practice it looks like most languages don't have anything actually filled in for it (https://www.unicode.org/cldr/charts/47/supplemental/language_plural_rules.html). Not to mention this is just unicode data anyway and it doesn't seem integrated into neither rails's built-in i18n gem or the `i18n-rails` gem. 

Unicode's default if nothing is specifyed is to pluralize according to the larger number in the range, so in this case `ArchiveConfig.PASSWORD_LENGTH_MAX`. If we were to pluralize that's what I think the `count` variable should be set to. But considering that the gems we're using don't seem to have a way to take into account separate range pluralization rules to handle the language-dependent rules automatically, and that most languages don't have range pluralization rules defined to begin with in the unicode localization data, it might not be worth the added complication to treat ranges different from single numbers? 

I know in English it doesn't really make sense to think about ranges independent of plurals cause if something's not plural it has to be one and 1 thing isn't really a range. But like for those languages that also pluralize numbers ending in 1 as "singular" no matter how big the number is are ranges still considered "plural" by default? I couldn't find an anser about it off the internet ):

app/views/users/passwords/edit.html.erb
```diff
@@ -1,27 +1,27 @@
...
       </p>
     </dd>
-    <dt><%= label_tag :edit_password_confirmation, ts("Confirm new password") %></dt>
+    <dt><%= label_tag :edit_password_confirmation, t(".confirm_password") %></dt>
     <dd><%= f.password_field :password_confirmation, id: :edit_password_confirmation %></dd>
-    <dt class="landmark"><%= ts("Submit") %></dt>
+    <dt class="landmark"><%= t(".submit") %></dt>
     <dd class="submit actions">
-      <%= f.submit ts("Change Password") %>
+      <%= f.submit t(".page_heading") %>
```
Hamham6: You wanna use different keys in different contexts (in this case a descriptive page heading vs what a button does) even if it's the same words in English, cause in other languages it might be different words and the translations folks can't do that without separate keys. A special case for forms in particular, there's usually a "submit" landmark distinct from the "submit" button itself, and `.submit` is usually used for the *button* while `.landmark.submit` is usually used for the landmark instead.

config/locales/models/en.yml
```diff
@@ -211,7 +211,7 @@ en:
                 other: can only be changed once every %{count} days. You last changed your username on %{renamed_at}.
               invalid: must be %{min_login} to %{max_login} characters (A-Z, a-z, _, 0-9 only), no spaces, cannot begin or end with underscore (_).
             password_confirmation:
-              confirmation: doesn't match new password.
+              confirmation: The passwords you entered do not match. Please try again.
```
Hamham6, line 220:
```suggestion
              confirmation: The passwords you entered do not match. Please try again.
              format: "%{message}"
```
Or else "Password" gets automatically tacked on the front and it becomes "Password The passwords you entered do not match. Please try again." 😭

app/views/users/change_password.html.erb
```diff
@@ -10,22 +10,22 @@
...
       </p>
     </dd>
-    <dt><%= label_tag :password_confirmation, ts("Confirm new password") %></dt>
+    <dt><%= label_tag :password_confirmation, t(".confirm_password") %></dt>
     <dd><%= password_field_tag :password_confirmation %></dd>
-    <dt><%= label_tag :password_check, ts("Old password") %></dt>
+    <dt><%= label_tag :password_check, t(".old_password") %></dt>
     <dd><%= password_field_tag :password_check%></dd>
-    <dt class="landmark"><%= label_tag :submit, ts("Submit") %></dt>
+    <dt class="landmark"><%= label_tag :submit, t(".submit") %></dt>
```
Hamham6, line 26: Same thing here as in `app/views/users/password/edit.html.erb` line 24 regarding the key names.

Bilka2:
Hey, we really appreciate the effort you're putting into this! However, as marked on Jira, this issue is blocked by #5289, so further changes made here will likely lead to more merge conflicts with that PR. For now, it would be best to pause coding here until we can get the other PR merged and then you can update your branch based on it.

If you'd like to work on something in the meantime, there are always other issue available on Jira, for example the issues marked as [Easy difficulty](https://otwarchive.atlassian.net/issues/?filter=10800&jql=project%20%3D%20AO3%20AND%20status%20%3D%20%22To%20Do%22%20AND%20assignee%20IN%20%28empty%29%20AND%20%22difficulty%5Bdropdown%5D%22%20%3D%20Easy).

app/views/users/change_password.html.erb
Bilka2, line 1: This form is changed by the other PR I linked and not related to the password reset form (passwords/edit) we should be changing here. When resolving the merge conflicts from master, you can just take all the changes from master and make none here in this PR

app/views/users/passwords/edit.html.erb
Bilka2: To action Hamham's review here, you can take a look at the (now updated) change_password form and mimic how it localises the error message: https://github.com/otwcode/otwarchive/blob/b1e112618faa943471ee1dd5b70737638765fa4d/app/views/users/change_password.html.erb#L19-L21

Please also match the key name (password_length instead of password_requirements), for consistency

app/views/users/passwords/edit.html.erb
```diff
@@ -1,27 +1,27 @@
...
                            "aria-describedby" => "password-field-description") %>
       <p class="footnote" id="password-field-description">
-        <%= ts("%{minimum} to %{maximum} characters",
+        <%= t(".password_requirements",
                                  minimum: ArchiveConfig.PASSWORD_LENGTH_MIN,
                                  maximum: ArchiveConfig.PASSWORD_LENGTH_MAX) %>
       </p>
     </dd>
-    <dt><%= label_tag :edit_password_confirmation, ts("Confirm new password") %></dt>
+    <dt><%= label_tag :edit_password_confirmation, t(".confirm_password") %></dt>
```
Bilka2: Would be nice to match the now updated change_password here as well, so
 ```suggestion
    <dt><%= label_tag :edit_password_confirmation, t(".confirm_new_password") %></dt>
```

File: app/views/users/passwords/edit.html.erb (hunk 2 above)
Comment (Bilka2) [reply app/views/users/passwords/edit.html.erb]: While Hamham is correct about the landmark key, the (now updated) change_password uses the plain `.submit` for it, so it's fine to keep here. But the button here should indeed have a different key than the page heading, and the now updated change_password offers us a good key to match here as well with this:
```suggestion
      <%= f.submit t(".change_password") %>
```

config/locales/models/en.yml
Bilka2: Agreed, though it's important to double-check that no other message on the attribute is using this format when changing it. In this case, https://github.com/heartcombo/devise/blob/cf93de390a29654620fdf7ac07b4794eb95171d0/lib/devise/models/validatable.rb#L38 is indeed the only place running a validation on `password_confirmation`, so this change is fine to make to fix the message formatting

Bilka2 (changes requested):
We've merged the PR that was blocking this, so you should be able to merge master and hopefully be able to finish this off without any new merge conflicts. If the merge is too intimidating as is, please let us know and we can help.

From the Jira issue, please also address point 2, the browser page title. It's set in the controller using `@page_subtitle` - if you search for that in the codebase you should be able to find plenty of examples.

Your Jira account should now also have permissions!

Tani-Shah:
Thank you for all the help and feedback @Hamham6 @Bilka2!!

PR #5378, opened 2025-10-01
AO3-5668 Change fandom tag check logic

anmazz:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5668

## Purpose

Query the db instead of loading all prompts as to whether to show the fandom sort button or not.

## Credit

anna

### DIFF
--- a/app/views/challenge/shared/_challenge_requests.html.erb
+++ b/app/views/challenge/shared/_challenge_requests.html.erb
@@ -12,7 +12,7 @@
   <% end %>
   <li><%= ts("Sort By") %></li>
   <% if @show_request_fandom_tags && @collection.prompts.joins(:tags).where(tags: { type: "Fandom" }).exists? %>
-    <li><%= sort_link t(".fandom"), :fandom %></li>
+    <li><%= sort_link t(".fandom", allowed_fandom_count: @collection.challenge.request_restriction.allowed("fandom")), :fandom %></li>
   <% end %>
   <% unless @collection.prompts.where(:anonymous => true).exists? %>
     <li><%= sort_link ts("Prompter"), :prompter %></li>
--- a/config/locales/views/en.yml
+++ b/config/locales/views/en.yml
@@ -561,7 +561,7 @@ en:
         requests_html: Requests &#8595;
     shared:
       challenge_requests:
-        fandom: Fandom
+        fandom: Fandom %{allowed_fandom_count}
   challenge_assignments:
     to_default:
       button: Default


app/views/challenge/shared/_challenge_requests.html.erb
```diff
@@ -11,13 +11,8 @@
     </li>
   <% end %>
   <li><%= ts("Sort By") %></li>
-  <% if @show_request_fandom_tags %>
-    <% @collection.prompts.each do |f| %>
-      <% if f.tag_groups.include?("Fandom") %>
-        <li><%= sort_link ts("Fandom #{@collection.challenge.request_restriction.allowed("fandom")}"), :fandom %></li>
```
anmazz, line 17: I removed `#{@collection.challenge.request_restriction.allowed("fandom")}` - seemed like a mistake that was left in

anmazz: ok haha nevermind, that's in the tests. can that be removed? it's not clear to me why that's there. it's always 1

sarken: No, it should stay. It's the number of fandom tags participants are allowed to include in a prompt, so it can vary. For example, it's 6 in [this list of requests](https://archiveofourown.org/collections/Third_Monday_13/requests). This information can be a useful reminder that things may not be sorted as you'd expect.

anmazz: @sarken ah 🤦 my bad I kept testing with different fandoms per prompt (thought it was a count of fandoms across actual posted prompts) + then was completely reading that value as a flag (either 1 or 0), so I was wrong on both accounts lol.

This is just my 2 cents so please ignore me but the number after "Fandom" has never been super clear to me which is why I thought it was a mistake (I feel like IRL I've only seen the "Fandom 1" sort button and I've had no clue what the 1 meant.)

I'll put it back for this PR though!

anmazz:
No gem updates here 😭 just an initial bad rebase on my part

anmazz:
> This code looks good! Could you add a feature test to check that the button still shows up? You can likely do the same testing steps as what you added on Jira.

@Bilka2 As I was looking around it seems like `challenge_promptmeme_setup.feature:210` and `238` check for the sort button. Do you think those good to cover this case?

Bilka2 (approved):
Yup that's fine, totally missed it in my search! Thanks for pointing it out

anmazz:
> Yup that's fine, totally missed it in my search! Thanks for pointing it out

Perfect thanks!

PR #5379, opened 2025-10-01
AO3-7085 AO3-4181 AO3-6217 Update buttons when posting, previewing, or editing a work

weeklies:
## Issue

https://otwarchive.atlassian.net/browse/AO3-7085
https://otwarchive.atlassian.net/browse/AO3-4181
https://otwarchive.atlassian.net/browse/AO3-6217
## Purpose

- Update the displayed text and order of buttons when posting/previewing/editing a work/chapter/work tags.
- Try to use links for Cancel buttons where appropriate, rather than going through the controller 
- I also noticed a minor bug, and fixed a page subtitle from 'Update Tags Work' to 'Preview Work Tags'
- Make Cancel button on edit work redirect to work itself (AO3-4181)
- Fix 500 error when canceling edit work tags for admins (AO3-6217)
- Per [97b652f](https://github.com/otwcode/otwarchive/pull/5379/commits/97b652fe4956d218a123da439739df0eb1baa45a), make the posting fieldsets more consistent. See details at https://github.com/otwcode/otwarchive/pull/5379#discussion_r2409271403

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -315,13 +315,14 @@ class WorksController < ApplicationController
 
       render :new and return unless @work.save
 
-      in_moderated_collection
       if @work.posted
         # We check here to see if we are attempting to post to moderated collection
         flash[:notice] = t(".posted_notice")
+        in_moderated_collection
         redirect_to work_path(@work)
       else
         flash[:notice] = t(".draft_notice_html", scheduled_for_deletion_bold: helpers.tag.strong(t(".scheduled_for_deletion")), deletion_date: view_context.date_in_zone(@work.created_at + 29.days))
+        in_moderated_collection
         redirect_to preview_work_path(@work)
       end
     end


app/controllers/works_controller.rb
```diff
@@ -409,11 +395,9 @@ def update
     end
   end
 
+  # POST /works/1/edit_tags
```
sarken: I think this is a PATCH  according to our routes.rb, although we do tend to call those PUT.

config/locales/helpers/en.yml
```diff
@@ -53,5 +53,15 @@ en:
...
+    actions:
+      cancel: Cancel
+      edit: Edit
+      please_wait: Please wait...
+      post: Post
+      post_chapter: Post Chapter
+      post_work: Post Work
+      preview: Preview
+      save_draft: Save Draft
+      update: Update
```
sarken: I definitely see your intention here, but since these aren't used in the works helper, they shouldn't be here. 

When we want to reuse text across a lot of files, we usually put it in the appropriate locale folder, but with a name that isn't tied to a particular view/controller/etc. For example, a landmark we use in multiple views or a closing we use in multiple mailers:

https://github.com/otwcode/otwarchive/blob/39732571ee6614e85027b258c45afaffe2942dd5/config/locales/views/en.yml#L3-L4

https://github.com/otwcode/otwarchive/blob/39732571ee6614e85027b258c45afaffe2942dd5/config/locales/mailers/en.yml#L211-L214

That being said, I think we should take a different approach here.

There's a ton of duplication both within and across the six views whose buttons this applies to, so let's create two partials: one we can use for works and one we can use for chapters. Each can have its own group of translations to make it easier to change the buttons to say, e.g., Post Work or Post Chapter in the future. (It's usually a good idea to use [Verb][Object] phrasing on a button for clarity.)

I've got a quick version in 9d163a401771d4624f7610d54d096987c6ffba11, but the names of my partials are somewhat questionable, so you might want to come up with something that fits in a bit better with the existing work and chapter partials. 

There are a couple of changes that come with that approach, but they're neutral or good:

* The buttons on the tag preview are currently a paragraph; this approach will convert them into a list, like the other views.
* On the Edit Tags page for draft works, the name of the Post button is currently update_button. It will be changed to post_button, like it is on the other forms. This doesn't seem to cause problems.
* The tag preview is currently missing the message about works needing to conform to the TOS; this will add it. (Policy & Abuse confirmed this was an oversight and not an intentional decision.)
* It will remove the class `create` from the field set on the standard work form, but we don't use that.

features/works/work_edit_tags.feature
```diff
@@ -81,6 +81,14 @@ Feature: Edit tags on a work
     And I view the work "Work 1"
     And I follow "Edit Tags"
     And I fill in "Fandoms" with ""
-    And I press "Cancel"
-  When I view the work "Work 1"
+    And I follow "Cancel"
+  Then I should see the page title "Work 1 -"
```
sarken, line 85: The `Then` on the next line needs to become an `And`

features/works/work_edit.feature
```diff
@@ -208,16 +207,15 @@ Feature: Edit Works
       And I post the work "Work 1" with fandom "testing"
       And I edit the work "Work 1"
       And I fill in "Fandoms" with ""
-      And I press "Cancel"
-    When I view the work "Work 1"
+      And I follow "Cancel"
```
sarken: Let's make this our `When`

features/works/chapter_edit.feature
```diff
@@ -618,3 +618,11 @@ Feature: Edit chapters
     When I view the work "Over the Limit"
     Then I should see "1/1"
       And I should not see "Next Chapter"
+
+  Scenario: Canceling edit chapter redirects to work page
```
sarken: I think the description here is wrong. We're redirecting to the chapter. (We should also test cancelling when adding a new chapter, which is when we redirect to the work.)

features/works/work_drafts.feature
```diff
@@ -4,8 +4,8 @@ Feature: Work Drafts
   Scenario: Creating a work draft
   Given I am logged in as "Scott" with password "password"
   When the draft "scotts draft"
-    And I press "Cancel"
-  Then I should see "The work was not posted. It will be saved here in your drafts for one month, then deleted from the Archive."
+    And I follow "Cancel"
+  Then I should see "This work is a draft and has not been posted. The draft will be scheduled for deletion on"
```
sarken, line 8: I think we also need tests for canceling a new record and canceling edits on a posted work, but GitHub's new review mode is misbehaving in some fun ways, so I might have missed them (and the latter wouldn't go in this file anyway).

sarken (changes requested):
This is a good start! The controllers look so much better without all of that button-specific logic. I think we can do some tidying up on the views, too.

config/locales/helpers/en.yml
weeklies: Thanks for this!

app/controllers/works_controller.rb
```diff
@@ -422,12 +406,13 @@ def update_tags
       render :edit_tags
     elsif params[:preview_button]
       @preview_mode = true
+      @page_subtitle = t(".page_title.preview")
```
sarken: I might be missing it in the collapsed lines -- are there other places `@page_subtitle` is being set for this action? If not, I think this can just be `page_title` instead of `page_title.preview`.

features/works/work_create.feature
```diff
@@ -27,6 +27,13 @@ Feature: Create Works
     When I go to the works page
     Then I should see "All Hell Breaks Loose 🤬💩"
 
+  Scenario: Canceling a new work redirects to the user's works page
+    Given I am logged in as "newbie"
+    When I set up the draft "All Hell Breaks Loose"
+      And I fill in "content" with "Bad things happen, etc."
```
sarken: We should be okay without this step; content gets added by the `I fill in the basic work information` part of the `I set up the draft` step.

features/works/work_edit_tags.feature
```diff
@@ -81,6 +81,14 @@ Feature: Edit tags on a work
...
+    And I should see "Fandom: testing"
+
+  Scenario: Preview edit tags
+  Given I am logged in as "regularuser"
+    And I post the work "Some Work"
+    And I follow "Edit Tags"
+    And I press "Preview"
+  Then I should see the page title "Preview Work Tags"
```
sarken, line 94: This should probably either be described as something like "Preview page has the right browser page title" or else it should get filled out a bit more so it's actually testing that previewing works (i.e., we should change a tag and make sure the changes shows up in the preview).

app/controllers/works_controller.rb
```diff
@@ -319,19 +313,17 @@ def create
...
+
+      if @work.posted
+        # We check here to see if we are attempting to post to moderated collection
+        flash[:notice] = t(".posted_notice")
+        in_moderated_collection
+        redirect_to work_path(@work)
       else
-        render :new
+        flash[:notice] = t(".draft_notice_html", deletion_date: view_context.date_in_zone(@work.created_at + 29.days))
+        in_moderated_collection
```
brianjaustin, line 325: Since this appears in both branches of the `if`, could it be raised to above `if @work.posted` (currently line 318)?

config/locales/controllers/en.yml
```diff
@@ -282,11 +282,21 @@ en:
       ban_notice_html: Your account has been banned. You are not permitted to post or edit content on AO3. Please check your email or %{contact_abuse_link} for more information.
       suspension_notice_html: Your account has been suspended until %{suspended_until}. You cannot post, edit, or delete content until your suspension has ended. Please check your email or %{contact_abuse_link} for more information.
   works:
+    create:
+      draft_notice_html: Draft was successfully created. It will be <strong>scheduled for deletion</strong> on %{deletion_date}.
+      posted_notice: Work was successfully posted. It should appear in work listings within the next few minutes.
```
brianjaustin: We try to avoid raw HTML in translation strings because it can be more difficult for the translators. Instead, you could do
```suggestion
      draft_notice_html: Draft was successfully created. It will be %{scheduled_for_deletion_bold} on %{deletion_date}.
      posted_notice: Work was successfully posted. It should appear in work listings within the next few minutes.
      scheduled_for_deletion: scheduled for deletion
```

and update the code that references this to
```rb
t(".draft_notice_html", scheduled_for_deletion_bold: style_bold(t(".scheduled_for_deletion")), ...)
```

Bilka2:
I think the test issues are cause by moving around `in_moderated_collection` since it appends to the current flash (and that is now getting replaced). I'd say to just move it back to where it was

weeklies:
Thanks!

app/controllers/works_controller.rb
```diff
@@ -319,19 +313,16 @@ def create
...
-          in_moderated_collection
-          redirect_to work_path(@work)
-        end
+      render :new and return unless @work.save
+
+      in_moderated_collection
+      if @work.posted
+        # We check here to see if we are attempting to post to moderated collection
+        flash[:notice] = t(".posted_notice")
+        redirect_to work_path(@work)
```
weeklies:
```suggestion
      if @work.posted
        # We check here to see if we are attempting to post to moderated collection
        flash[:notice] = t(".posted_notice")
        in_moderated_collection
        redirect_to work_path(@work)
```

app/controllers/works_controller.rb
```diff
@@ -319,19 +313,16 @@ def create
...
       else
-        render :new
+        flash[:notice] = t(".draft_notice_html", scheduled_for_deletion_bold: helpers.tag.strong(t(".scheduled_for_deletion")), deletion_date: view_context.date_in_zone(@work.created_at + 29.days))
```
weeklies, line 324:
```suggestion
        flash[:notice] = t(".draft_notice_html", scheduled_for_deletion_bold: helpers.tag.strong(t(".scheduled_for_deletion")), deletion_date: view_context.date_in_zone(@work.created_at + 29.days))
        in_moderated_collection
```

PR #5382, opened 2025-10-02
AO3-7148 Remove UnicodeUtils gem

brianjaustin:


anmazz:
worked on this yesterday and this is what my change set looks like exactly. can confirm that this is working 👍

Bilka2 (approved):

PR #5384, opened 2025-10-02
AO3-7045 Migrate audits table to BIGINT

alyssapowell:
## Issue

https://otwarchive.atlassian.net/browse/AO3-7045

## Purpose

Migrates the audits table primary key from integer to BIGINT to prevent integer overflow as the audit log grows. This is a preventive infrastructure improvement to handle the increasing volume of audit records in the system.

## Testing Instructions

1. Run the migration in a development environment: `rails db:migrate`
2. Verify the schema change: Check that `audits.id` column is now BIGINT type
3. Test audit functionality: Ensure audit records are still created for tracked changes
4. Run existing test suite to verify no functionality breaks

## References

This follows the same pattern as recent similar migrations:
- `20250704060231_change_bookmark_foreign_keys_type.rb` 
- `20250718220533_change_subscriptions_id_to_bigint.rb`

## Credit

Alyssa Powell (she/her)

### DIFF
--- a/db/migrate/20251002104708_change_audits_id_to_bigint.rb
+++ b/db/migrate/20251002104708_change_audits_id_to_bigint.rb
@@ -1,0 +1,11 @@
+class ChangeAuditsIdToBigint < ActiveRecord::Migration[7.1]
+  uses_departure! if Rails.env.staging? || Rails.env.production?
+
+  def up
+    change_column :audits, :id, "bigint NOT NULL AUTO_INCREMENT"
+  end
+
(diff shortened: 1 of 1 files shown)


Bilka2:

Please note that we may take over the code in this PR (with credit) because we are hoping to finish the bigint migrations soon and this PR has been stalled for a few months now.

brianjaustin:
Adopted in https://github.com/otwcode/otwarchive/pull/5516

PR #5388, opened 2025-10-04
AO3-7027 Update error pages and other status page links, add status page to footer

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-7027

## Purpose

- Update the error page text, including adding missing links.
- Add Status Page link to footer.
- Replace links to our Twitter status account with Bluesky.
- Use Font Awesome icons for Twitter, Bluesky, and Tumblr.

I entertained the idea of setting the various URLs in the config file, but the vast majority of uses are on static pages, so it didn't seem particularly helpful, but I can do it. 

[https://sharingbuttons.io](https://github.com/mxstbr/sharingbuttons.io) hasn't been updated in 5 years, so we had to roll our own Bluesky share button:
- The non-hover color is the blue from the butterfly in the bluesky_media_kit_logo_transparent_1.png of [the Bluesky media kit](https://drive.google.com/drive/folders/16mlJPfWNnc6jj-3vGZ88SFysIY13GBp0).
- The hover color is the color the "New Post" button on Bluesky turns when you hover over it (`rgb(16, 131, 254)`, but converted to hex). 
- The sharingbuttons.io site says their icons are by [Streamline](https://www.streamlinehq.com), but they don't appear to offer their own free Bluesky icon, so I turned to Bluesky's [documentation for share button branding](https://docs.bsky.app/docs/advanced-guides/intent-links#button-styling). It points to [Font Awesome](https://fontawesome.com), so to have all of our icons match in terms of appearance and code needs (note the attribution comment), I switched everything to Font Awesome. Legal has confirmed we're okay to use these icons.

## Testing Instructions

Refer to Jira.


config/locales/views/en.yml
```diff
@@ -1303,11 +1304,11 @@ en:
           title: Is it later already?
         note: Some works you've marked for later.
       social:
+        bluesky: "@status.archiveofourown.org on Bluesky"
         heading: Follow us
-        note_html: Follow the Archive on Twitter or Tumblr for status updates, and don't forget to check out the %{other_outlets_link} for updates on our other projects!
+        note_html: Follow the Archive on Bluesky or Tumblr for status updates, and don't forget to check out the %{other_outlets_link} for updates on our other projects!
```
Bilka2: Should we maybe use AO3 instead "the Archive" here since we're generally moving towards that wording?

public/nomaintenance.html
```diff
@@ -99,20 +99,26 @@ <h1 class="heading">
...
       <div id="inner" class="wrapper">
         <div id="main" class="system errors error-503-maintenance region" role="main">
-          <h2 class="heading">Error 503 - Service unavailable</h2>
-          <h3 class="heading">The Archive is down for maintenance.</h3>
-          <p>If <a href="https://twitter.com/ao3_status/">@AO3_Status</a> and <a href="https://ao3org.tumblr.com">ao3org</a> say the site is up, but you still see this page, try <a href="https://kb.iu.edu/data/ahic.html">clearing your browser cache</a> and refreshing the page.</p>
+          <h2 class="heading">Error 503</h2>
+          <h3 class="heading">AO3 is temporarily down for maintenance.</h3>
+          <p>Check our <a href="https://www.otwstatus.org">status page</a> and follow <a href="https://bsky.app/profile/status.archiveofourown.org">@status.archiveofourown.org</a> on Bluesky or <a href="https://ao3org.tumblr.com">ao3org</a> on Tumblr for updates.</p>
+          <p>If our <a href="https://www.otwstatus.org">status page</a> says the site is up, but you still see this page, try <a href="https://kb.iu.edu/data/ahic.html">clearing your browser cache</a> and refreshing the page.</p>
```
Bilka2:
```suggestion
          <p>If our <a href="https://www.otwstatus.org">status page</a> says the site is up, but you still see this page, try <a href="https://en.wikipedia.org/wiki/Wikipedia:Bypass_your_cache">clearing your browser cache</a> and refreshing the page.</p>
```

app/helpers/share_helper.rb
```diff
@@ -59,20 +59,27 @@ def remove_newlines(html)
     html.delete("\n")
   end
 
+  # We have replaced the svgs from https://sharingbuttons.io/ with Font Awesome
```
Bilka2: Since we're no longer using sharingbuttons, I think this comment can go as well? Since it's not very useful to know what we used to use

sarken: I can rephrase it -- it's basically a case of we're using everything about sharingbuttons (the comment on `sharing_button` above notes the specific style we're using) _except_ the SVGs, which is an important change to note.

config/locales/views/en.yml
sarken: Words for me! I'll also ask Support if they'd mind me changing it on their page.

Bilka2:
Just noticed: It would be good to update 406-unsupported-browser.html as well, it was added with #5303 (after this issue was originally created)

config/locales/views/en.yml
sarken: (They said it was fine.)

PR #5389, opened 2025-10-04
AO3-6983 Add support form notices

marcus8448:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6983

## Purpose

Adds support form notices, similar to banners, but displayed just above the support form.

## Credit

marcus8448 (he/him)

### DIFF
--- a/features/other_b/support_notice.feature
+++ b/features/other_b/support_notice.feature
@@ -69,3 +69,11 @@ Feature: Notices on support page
     Then I should see the "caution" support notice
     When I follow "2"
     Then I should see the "error" support notice
+
+  Scenario: Support notices are still displayed after form submission is rejected
+    Given an active support notice
+    When I go to the support page
+    Then I should see the support notice
+    When I press "Send"
+    Then I should see "Sorry! We couldn't save this feedback"
+      And I should see the support notice
--- a/features/step_definitions/support_notice_steps.rb
+++ b/features/step_definitions/support_notice_steps.rb
@@ -1,3 +1,9 @@
+### GIVEN
+
+Given "an active support notice" do
+  FactoryBot.create(:support_notice, :active, notice_content: "This is some support notice text")
+end
+
 ### WHEN
 
 When /^I create an?( active)?(?: "([^"]*)")? support notice$/ do |active, notice_type| # rubocop:disable Cucumber/RegexStepName


Bilka2:
Here with some pre-review feedback: Bugs of banners that support form notices should already account for: https://otwarchive.atlassian.net/browse/AO3-5932, https://otwarchive.atlassian.net/browse/AO3-6604

And we'd also like to log admin activities for creating/updating/deleting support form notices. You can base the design around the planned logging for banners please, [AO3-6395](https://otwarchive.atlassian.net/browse/AO3-6395).

[AO3-6395]: https://otwarchive.atlassian.net/browse/AO3-6395?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

app/views/feedbacks/new.html.erb
```diff
@@ -41,6 +41,11 @@
     </ul>
     <p><%= t(".do_not_spam_html") %></p>
     <p><%= t(".languages_html", list_html: to_sentence(@support_languages.map { |language| tag.span(language.name, lang: language.short) })) %></p>
+    <% if @support_notice %>
+      <div class="notice<%= @support_notice.css_class %>">
```
Bilka2: How about making the method return all the classes (so notice + anything custom), similar to how `css_classes_for_comment` and similar work?

app/views/admin/support_notices/index.html.erb
```diff
@@ -0,0 +1,22 @@
...
+  <%= render "navigation" %>
+  <!--/subnav-->
+
+  <%== pagy_nav(@pagy) %>
+
+  <!--main content-->
+  <ul class="support-notices index group">
+    <% @support_notices.each { |support_notice| %>
+      <%= render "support_notice", support_notice: support_notice %>
+    <% } %>
```
Bilka2: Please use do/end for multiline blocks instead of `{}`

db/migrate/20251006194643_create_support_notices.rb
```diff
@@ -0,0 +1,12 @@
+class CreateSupportNotices < ActiveRecord::Migration[7.1]
+  def change
+    create_table :support_notices do |t|
+      t.text :notice
+      t.integer :notice_sanitizer_version
```
Bilka2: Should this have `limit: 2, default: 0, null: false` like the other sanitizer versions?

db/migrate/20251006194643_create_support_notices.rb
```diff
@@ -0,0 +1,12 @@
...
+      t.string :support_notice_type
+      t.boolean :active
```
Bilka2: I think it would be nice to set a default here and make it non-null

app/controllers/admin/support_notices_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+  def confirm_delete
+  end
+
+  # DELETE /admin/notices/support/1
+  def destroy
+    AdminActivity.log_action(current_admin, @support_notice, action: "destroy", summary: @support_notice.notice)
+    @support_notice.destroy
+
+    flash[:notice] = t(".successfully_deleted")
+    redirect_to admin_support_notices_path
```
Bilka2: TODO Bilka

app/policies/support_notice_policy.rb
```diff
@@ -0,0 +1,12 @@
...
+  SUPPORT_NOTICE_ROLES = %w[superadmin support].freeze
+
+  def index?
+    user_has_roles?(SUPPORT_NOTICE_ROLES)
+  end
+
+  alias show? index?
+  alias update? index?
+  alias create? index?
+  alias destroy? index?
```
Bilka2, line 11: this is missing update and a few other actions... now I'm wondering how the tests passed for those actions

spec/controllers/admin/support_notices_controller_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+
+describe Admin::SupportNoticesController do
+  include LoginMacros
+  include RedirectExpectationHelper
+
+  let(:admin) { create(:admin, roles: ["superadmin"]) }
+  let(:support_notice) { create(:support_notice) }
+  let(:support_notice_params) { attributes_for(:support_notice, notice: "New notice content") }
+
+  shared_examples "only authorized admins are allowed" do |authorized_roles:|
```
Bilka2: We have a `an action only authorized admins can access` shared example (for all specs) that should already cover most of this. I think it would be better to use that and make a separate example for what it doesn't cover

app/views/admin/support_notices/_navigation.html.erb
```diff
@@ -0,0 +1,6 @@
+<ul class="navigation actions">
+  <li><%= span_if_current t(".index"), admin_support_notices_path %></li>
+  <% if policy(SupportNotice).new? %>
+    <li><%= span_if_current t(".new"), new_admin_support_notice_path %></li>
+  <% end %>
```
Bilka2, line 12: Aren't these always accessible to the same set of admin? So I think we don't need to check role here at all since only admins that have access to everything support notice can see the navigation

app/views/admin/support_notices/_support_notice_form.html.erb
```diff
@@ -0,0 +1,43 @@
+<%= form_for @support_notice, url: @support_notice.new_record? ? admin_support_notices_path : admin_support_notice_path(@support_notice), html: { class: "post" } do |f| %>
```
Bilka2: I guess the polymorphic routing problems thing is why this can't auto deduce the URL? Does it work if you try the more modern `form_with`?

config/locales/views/en.yml
```diff
@@ -418,6 +421,46 @@ en:
...
+          content: Support Notice
+        legend:
+          actions: Actions
+          content: Support Notice
+        required_information: Required Information
+        text: Support Notice Text
+        type:
+          caution: Caution
+          error: Error
+          label: Support Notice Type
```
Bilka2: Here and for Support Notice Text I think we do sentence case for these form labels now?

app/controllers/admin/support_notices_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+  # GET /admin/notices/support/1/edit
+  def edit
+  end
+
+  # POST /admin/notices/support
+  def create
+    @support_notice = authorize SupportNotice.new(support_notice_params)
+
+    if @support_notice.save
+      AdminActivity.log_action(current_admin, @support_notice, action: "create", summary: @support_notice.notice)
```
Bilka2: What do you think about making this and the other admin actions "create support notice" and similar instead? So it's easier to see at a glance which type of thing was modified. If I'm reading the issue correctly then that would also match the design we have for banners

app/controllers/admin/support_notices_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+
+      flash[:notice] = t(".created")
```
Bilka2:
```suggestion
      flash[:notice] = t(".success")
```
is a bit more standard. Same goes for the other two actions

app/views/admin/support_notices/show.html.erb
```diff
@@ -0,0 +1,15 @@
...
+  <!--/descriptions-->
+
+  <!--subnav-->
+  <%= render "navigation" %>
+  <!--/subnav-->
+
+  <!--main content-->
+  <ul class="support-notices index group">
+    <%= render "support_notice", support_notice: @support_notice %>
+  </ul>
```
Bilka2: The `support-notices` class name reads strange, could you poke either the front end standards or sarken about it? I think it should at least be singular but maybe the dash is also wrong. Might be good to get a sarken front end review in general since this has a bunch of differences to the banner code (but good difference from what i see)

app/views/admin/support_notices/_support_notice.html.erb
```diff
@@ -0,0 +1,24 @@
...
+    <% end %>
+    <% if support_notice.active? %>
+      <li>
+        <span class="active current"><%= t(".active") %></span>
+      </li>
+    <% elsif policy(SupportNotice).destroy? %>
+      <li>
+        <%= link_to t(".delete"), confirm_delete_admin_support_notice_path(support_notice), data: { confirm: t(".confirm_delete") } %>
+      </li>
+    <% end %>
```
Bilka2: I really like this solution

app/models/support_notice.rb
```diff
@@ -0,0 +1,19 @@
...
+  def css_class
+    self.support_notice_type == "notice" ? "" : " #{self.support_notice_type}"
+  end
+
+  private
+
+  def ensure_single_active_notice
+    return unless self.active?
+
+    SupportNotice.where(active: true).where.not(id: self.id).update_all(active: false)
```
Bilka2: Could this not use `update_all` but instead something that triggers callbacks so we have accurate `updated_at` timestamps? I think it might throw us off if we ever need to debug stuff if a notice is "silently" deactivated

app/policies/support_notice_policy.rb
marcus8448: I'm not sure if I understand what you mean by this. Unless I'm reading the pundit docs all wrong, update should be covered by aliasing to index and the rest should be inherited from [`ApplicationPolicy`](https://github.com/otwcode/otwarchive/blob/a4968c02067e8680df39ea33e122ec23f3a13150/app/policies/application_policy.rb) (which then defers to the 5 policies defined in this file).

Do you want all of them explicitly defined in this file?

Bilka2: That answers my question how the tests passed! It's fine to keep it that way, I completely missed the ApplicationPolicy code (and `update` in the list here), so thank you for linking that and explaining! It's fine to keep this file as is

app/controllers/admin/support_notices_controller.rb
marcus8448: I was thinking that the "target" column already covers this, but I suppose there's no harm in having the information in the action column as well.

app/views/admin/support_notices/_navigation.html.erb
marcus8448: I was thinking more for future-proofing? I don't think there's really any harm in having the extra checks.

app/views/admin/support_notices/_support_notice_form.html.erb
marcus8448: Yeah... the routing is weird. Unfortunately `form_with` has the same issue.

app/models/support_notice.rb
```diff
@@ -0,0 +1,19 @@
+class SupportNotice < ApplicationRecord
+  SUPPORT_NOTICE_TYPES = %w[notice caution error].freeze
```
marcus8448: Should these be capitalized? (Not as in the CSS class names, but for what goes into the database).

app/views/admin/support_notices/confirm_delete.html.erb
```diff
@@ -0,0 +1,16 @@
+<div class="admin">
+  <!--Descriptive page name, messages and instructions-->
+  <h2 class="heading"><%= t(".delete") %></h2>
+  <!--/descriptions-->
+
+  <!--main content-->
+  <%= form_with model: @support_notice, url: admin_support_notice_path(@support_notice), method: :delete, html: { class: "simple destroy" } do |f| %>
+    <p class="caution notice">
+      <%= t(".warning_html") %>
```
marcus8448, line 9: The warning doesn't have any information on *which* support notice is being deleted, which doesn't seem helpful.
Should the entire support notice be displayed on the page? I can't come up with anything small to stick into the warning (e.g. a title).

app/views/admin/_header.html.erb
```diff
@@ -54,6 +54,15 @@
       <li><%= link_to t(".nav.banners"), admin_banners_path %></li>
     <% end %>
 
+    <% if policy(SupportNotice).index? %>
+      <li class="dropdown">
+        <%= link_to t(".nav.notices.notices"), admin_support_notices_path %>
+        <ul class="menu">
+          <li><%= link_to t(".nav.notices.support"), admin_support_notices_path %></li>
+        </ul>
```
sarken: Since both of these links point to `admin_support_notices_path` it's okay to skip the `ul` menu and the `dropdown` class until we actually need a menu with additional options. (Settings and Banners on the lines above are also just a top-level items.) We can change the name to "Support Notices" if just plain "Notices" is unclear.

app/views/admin/support_notices/index.html.erb
```diff
@@ -0,0 +1,22 @@
...
+  <%= render "navigation" %>
+  <!--/subnav-->
+
+  <%== pagy_nav(@pagy) %>
+
+  <!--main content-->
+  <ul class="support-notice index group">
+    <% @support_notices.each do |support_notice| %>
+      <%= render "support_notice", support_notice: support_notice %>
+    <% end %>
```
sarken: Instead of the `each...end`, would the single-line version `<%= render partial: "support_notice", collection: @support_notices, as: support_notice %>` work here?

db/migrate/20251206194643_create_support_notices.rb
```diff
@@ -0,0 +1,12 @@
+class CreateSupportNotices < ActiveRecord::Migration[7.2]
+  def change
+    create_table :support_notices do |t|
+      t.text :notice, null: false
```
sarken: Can we call this `content` so the name and behavior (i.e., what the config file allows in the field) match admin banners? 

It's possible we'll want to rename the `content` field for this and admin banners one day (e.g., so we can disallow media embeds while allowing the class attribute), but I think we'd probably go with `[something]_content` like we did when we changed comments to `comment_content`.

config/locales/views/en.yml
```diff
@@ -418,6 +421,46 @@ en:
...
+          create: Create Support Notice
+          update: Update Support Notice
+        active: Active
+        landmark:
+          actions: Actions
+          content: Support Notice
+        legend:
+          actions: Actions
+          content: Support Notice
+        required_information: Required Information
```
sarken: Could you make this "information" with a lowercase "i", please?

app/controllers/admin/support_notices_controller.rb
```diff
@@ -0,0 +1,80 @@
+class Admin::SupportNoticesController < Admin::BaseController
+  before_action :load_support_notice, only: [:confirm_delete, :destroy, :edit, :show, :update]
+  before_action :check_inactive, only: [:confirm_delete, :destroy]
+  
+  # GET /admin/notices/support
+  def index
```
sarken, line 6: We should probably set translatable `@page_subtitle`s for the new pages now to avoid needing to make issues for it later.

app/models/support_notice.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  validates :notice, presence: true
+  validates :support_notice_type, inclusion: { in: SUPPORT_NOTICE_TYPES }
+  after_save_commit :ensure_single_active_notice
+
+  def css_classes
+    ["notice", self.support_notice_type].uniq.join(" ")
```
sarken: Can you change the order here? We always go from [most specific class to least](https://otwcode.github.io/docs/classes.html#rules), and `caution` and `error` are acting as [state modifiers](https://otwcode.github.io/docs/classes-taxonomy.html#modifiers) for `notice`. 

I'm also wondering if this would be better as a helper, since it's just creating classes for the views...

config/locales/controllers/en.yml
```diff
@@ -17,6 +17,14 @@ en:
...
       destroy_user_creations:
         success: All creations by user %{login} have been deleted.
+    support_notices:
+      cannot_delete_active: Active support notices cannot be deleted.
+      create:
+        success: Support Notice successfully created.
+      destroy:
+        success: Support Notice successfully deleted.
+      update:
+        success: Support Notice successfully updated.
```
sarken: Can you lowercase "notice" in these? We don't tend to treat feature names as proper nouns -- when we capitalize them, it's usually just because we prefer title case for navigational items -- and it's correctly written as "support notice" in the message above and "Support notice" in the form labels.

app/views/admin/support_notices/_support_notice_form.html.erb
```diff
@@ -0,0 +1,43 @@
...
+    </dl>
+  </fieldset>
+
+  <fieldset>
+    <legend><%= t(".legend.actions") %></legend>
+    <h3 class="landmark heading"><%= t(".landmark.actions") %></h3>
+    <p class="submit cancel actions">
+      <%= f.submit @support_notice.new_record? ? t(".actions.create") : t(".actions.update") %>
+      <%= link_to t(".actions.cancel"), @support_notice.new_record? ? admin_support_notices_path : admin_support_notice_path(@support_notice) %>
+    </p>
```
sarken: Could you remove the cancel button here? We generally don't use those on most forms, since the back button is a viable alternative.

app/views/admin/support_notices/index.html.erb
```diff
@@ -0,0 +1,22 @@
...
+  <!--/descriptions-->
+
+  <!--subnav-->
+  <%= render "navigation" %>
+  <!--/subnav-->
+
+  <%== pagy_nav(@pagy) %>
+
+  <!--main content-->
+  <ul class="support-notice index group">
```
sarken: For class names, we use [unhyphenated English words in common usage](https://otwcode.github.io/docs/classes.html#rules), so in an ideal world, this would be `support notice index group`. 

However, `.notice` is a well-established class, and as you no doubt noticed (heh), using it here adds a whole bunch of unwanted styling. 

Luckily, we ran into this same problem on the admin banners page: `.announcement` was already an established class with styles specific to the single banner that displays on the top of the page. Using it in the `ul` wouldn't have worked without a lot of CSS changes. So we added the rather generic `banners` class to the `ul.`

While admin banners and support notices aren't exactly the same thing, they're close enough we can use `support banners` here. We don't currently have any styles that use `.banners` as a selector, but if we add any in the future, including `support` here gives us the option to make additional modifications that only apply to this page -- and we can always use the [automatically added Rails class](https://otwcode.github.io/docs/classes.html#rails-classes) to do `.support_notices-index .banners`.

(Note: We typically prefer to use singular class names, but I'd rather use the existing, incorrect one than a mix of the two. Changing it to `banner` the admin banners page is also an option, though!)

app/views/admin/support_notices/_support_notice.html.erb
```diff
@@ -0,0 +1,24 @@
+<li class="blurb">
```
sarken: It's fine to include `<li class="blurb">` in the partial, but it should be broken up into two partials, like we do in other places (e.g., the [work blurb](https://github.com/otwcode/otwarchive/blob/master/app/views/works/_work_blurb.html.erb)):

1. `support_notice_blurb` for the file with the blurb container and actions
2. `support_notice` for the partial with the notice `div`
  
That approach lets us reuse the same notice `div` code in three places: the blurb on the index, the show page, and the support form itself. The current single-file approach forces us to

1. use an index `ul` on the show page, even though it's not an index and will only ever display one blurb
2. reproduce the div on the support form itself

The other option is to copy the approach from the admin banner: put the blurb `li` and actions in `index.html.erb` and just include the notice `div` in this `support_notice` partial.

app/views/admin/support_notices/_support_notice.html.erb
```diff
@@ -0,0 +1,24 @@
...
+  <div class="<%= support_notice.css_classes %>">
+    <blockquote class="userstuff">
+      <%= raw sanitize_field(support_notice, :notice, image_safety_mode: true) %>
+    </blockquote>
+  </div>
+
+  <ul class="actions">
```
sarken: To Bilka's point, I don't think we need these extra permissions checks for each button. It doesn't seem very likely that we'll let people who can't edit, delete, or create support notices access this page.

If we do end up keeping those checks, though, we should have an extra check outside the `ul` to make sure we never include an empty `ul` element, which could happen if the notice isn't active and the admin can't edit or delete the notices.

app/views/admin/support_notices/confirm_delete.html.erb
sarken: The confirmation pop-up also doesn't tell you which notice you're deleting, and this page is just a non-JavaScript fallback that will be largely unused , so I'm inclined to say we can live with it.

db/migrate/20251206194643_create_support_notices.rb
Bilka2: I would much prefer not using a column name that we'll have to migrate in the future because column rename migrations are quite painful, they either require a lot of steps like https://otwarchive.atlassian.net/browse/AO3-5994 or very careful timing around the deploy. We already have https://otwarchive.atlassian.net/browse/AO3-7158 to update the admin banner field name. So I think we should go with the desired name here (like marcus did), not an old, incorrect name that we'll have to painfully migrate away from

If you are set on matching the field name to the admin banner field name here, I would prefer if you could adjust the admin banner Jira issue to the desired field name and we'd already use the desired field name here. However, I think it's a bad idea to use the same field names here and in admin banners, since it would limit us again if we ever want to have different behaviours between them

sarken:
I didn't see that issue -- as long as we plan to move away from the plain content name for admin banners, I'm happy doing that here too! But I'd still like to see this as notice_content and banners as banner_content. (I didn't mean they would both be the same thing after the rename, but that they would both follow the same naming scheme we'd been using for renaming columns we wanted to allow HTML.)Sent from my iPhoneOn Dec 11, 2025, at 7:55 AM, Bilka ***@***.***> wrote:﻿
@Bilka2 commented on this pull request.



In db/migrate/20251206194643_create_support_notices.rb:
> @@ -0,0 +1,12 @@
+class CreateSupportNotices < ActiveRecord::Migration[7.2]
+  def change
+    create_table :support_notices do |t|
+      t.text :notice, null: false

I would much prefer not using a column name that we'll have to migrate in the future because column rename migrations are quite painful, they either require a lot of steps like https://otwarchive.atlassian.net/browse/AO3-5994 or very careful timing around the deploy. We already have https://otwarchive.atlassian.net/browse/AO3-7158 to update the admin banner field name. So I think we should go with the desired name here (like marcus did), not an old, incorrect name that we'll have to painfully migrate away from
If you are set on matching the field name to the admin banner field name here, I would prefer if you could adjust the admin banner Jira issue to the desired field name and we'd already use the desired field name here. However, I think it's a bad idea to use the same field names here and in admin banners, since it would limit us again if we ever want to have different behaviours between them

db/migrate/20251206194643_create_support_notices.rb
Bilka2: I've updated the banner issue to `banner_content`, could you switch to `notice_content` here like Sarken said?

app/models/support_notice.rb
Bilka2: Good question. Looking at how we do this in existing code made me have a different question: Should we maybe be using https://api.rubyonrails.org/classes/ActiveRecord/Enum.html instead?

app/views/admin/support_notices/_navigation.html.erb
Bilka2: Usually when I find these checks in code I expect that there will be a difference and that I need to look at the policy, which is relevant for example for writing testing instructions or documentation. So I have a slight preference towards not having them. But it's only slight - they're still not breaking anything so you can keep them. Though note sarken's comment about the ul in`_support_notice.html.erb` if you keep them

app/models/support_notice.rb
```diff
@@ -5,15 +5,15 @@ class SupportNotice < ApplicationRecord
...
   end
 
   private
 
   def ensure_single_active_notice
     return unless self.active?
 
-    SupportNotice.where(active: true).where.not(id: self.id).update_all(active: false)
+    SupportNotice.where(active: true).where.not(id: self.id).update(active: false)
```
Bilka2, line 15: For your consideration, if you haven't already: `update_attribute`

features/step_definitions/support_notice_steps.rb
```diff
@@ -0,0 +1,84 @@
...
+  when "caution"
+    choose("support_notice_support_notice_type_caution")
+  when "error"
+    choose("support_notice_support_notice_type_error")
+  else
+    choose("support_notice_support_notice_type_notice")
+  end
+  check("support_notice_active") if active.present?
+  click_button("Create Support Notice")
+  step %{I should see "Support Notice successfully created."} if active.present?
```
Bilka2:
```suggestion
  step %{I should see "Support Notice successfully created."}
```

features/step_definitions/support_notice_steps.rb
```diff
@@ -0,0 +1,84 @@
...
+  end
+  step %{I should see "This is some support notice text"}
+end
+
+Then "I should see the edited active support notice" do
+  visit(feedbacks_path)
+  step %{I should see "This is some edited support notice text"}
+end
+
+Then "I should not see a support notice" do
```
Bilka2, line 71: Should this maybe also check for "support notice text"?

features/other_b/support_notice.feature
```diff
@@ -0,0 +1,57 @@
...
+    When an admin edits the active support notice
+    Then I should see the edited active support notice
+
+  Scenario: Admin can deactivate a support notice
+    Given there are no support notices
+      And an admin creates an active support notice
+    When an admin deactivates the support notice
+    Then I should not see a support notice
+
+  Scenario: Admin can delete a support notice and it will no longer be shown to users
```
Bilka2, line 35: Could this do create active -> check -> deactivate -> delete -> check instead? Right now this name is a bit misleading because the notice in question is never shown to users

features/step_definitions/support_notice_steps.rb
```diff
@@ -0,0 +1,84 @@
...
+  else
+    choose("support_notice_support_notice_type_notice")
+  end
+  check("support_notice_active") if active.present?
+  click_button("Create Support Notice")
+  step %{I should see "Support Notice successfully created."} if active.present?
+end
+
+When "an admin deactivates the support notice" do
+  step %{I am logged in as a "support" admin}
```
Bilka2: Logging in steps like this slow down the tests. could you remove them where they're not needed?

features/step_definitions/support_notice_steps.rb
```diff
@@ -0,0 +1,84 @@
...
+
+Given "there are no support notices" do
+  SupportNotice.delete_all
+end
+
+### WHEN
+
+When /^an admin creates an?( active)?(?: "([^"]*)")? support notice$/ do |active, notice_type| # rubocop:disable Cucumber/RegexStepName
+  step %{I am logged in as a "support" admin}
+  visit(new_admin_support_notice_path)
```
Bilka2, line 10: Could this go via the UI to the page instead? Just so that we test that at least once

spec/models/support_notice_spec.rb
Bilka2, line 1: Should we test for missing type?

spec/models/support_notice_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  context "multiple notices" do
+    let(:first_support_notice) { create(:support_notice, :active) }
+    let(:second_support_notice) { build(:support_notice, :active) }
+
+    it "deactivates other notices upon activation" do
+      expect(first_support_notice.active).to be_truthy
+      expect(second_support_notice.save).to be_truthy
+
+      expect(second_support_notice.active).to be_truthy
```
Bilka2: Just for safety, could this reload here as well?

Bilka2 (changes requested):
Reviewed your changes and the tests this time. So in combination with Sarken's front-end review that should make for a full review and I'm marking action needed :tada:

app/views/admin/support_notices/_navigation.html.erb
marcus8448: I've removed the checks

app/views/admin/_header.html.erb
marcus8448: :+1: I've updated the issue to match.

app/views/admin/support_notices/index.html.erb
marcus8448: I've updated to use `support banner` (and renamed the banner class accordingly).

features/step_definitions/support_notice_steps.rb
marcus8448: I've added the navigation to the "newer active support notice" step.

spec/models/support_notice_spec.rb
marcus8448: In theory, having a missing type is the same error as an invalid type (it's the same validation), but I don't see any harm from having the extra check.

app/views/admin/support_notices/_support_notice.html.erb
marcus8448: Reusing the support notice partial on the feedback page required a couple changes:

* The notice is no longer within the general `userstuff` information container since `blockquote`s within `userstuff` get additional styling/indentation.
  * The issue specifically asked for it to be in the `userstuff` `div`, so I'd just like to check if there's any reason(s) we need to keep it in there.
* I dropped the outer `div` on the notices, instead applying all the classes directly to the `blockquote`.

app/models/support_notice.rb
```diff
@@ -0,0 +1,17 @@
+class SupportNotice < ApplicationRecord
+  enum :support_notice_type, {
+    notice: 0,
+    caution: 1,
+    error: 2
+  }, default: 0, validate: { message: :invalid_type }
```
Bilka2: Thoughts on giving this a prefix or suffix? I think `supportnotice.notice?` being a thing as it is right now may be a bit confusing

app/models/support_notice.rb
```diff
@@ -0,0 +1,17 @@
...
+  validates :notice_content, presence: true
+  after_save_commit :ensure_single_active_notice
+
+  private
+
+  def ensure_single_active_notice
+    return unless self.active?
+
+    SupportNotice.where(active: true).where.not(id: self.id).find_each { it.update_attribute(:active, false) }
```
Bilka2: looking at this I think I might prefer `update` just because it acted on the collection? But if you like this better then that's fine too

Bilka2 (approved):
Looks good from my side of things (backend, tests)! Leaving frontend to sarken again

app/models/support_notice.rb
marcus8448: I've added a prefix, but how do you feel about renaming the types?
I think using `information`, `warning`, and `alert` makes more sense (maybe even "notice level" instead of type?). But that might just be because of logging libraries.

app/models/support_notice.rb
marcus8448: I do not like it better :sweat_smile: I've reverted it.


If we really want to skip validations then an `update_all` that sets `updated_at` could also work.

app/models/support_notice.rb
Bilka2: I think it's good to match the CSS class names, which I don't think we could rename

app/views/admin/support_notices/show.html.erb
```diff
@@ -0,0 +1,15 @@
...
+  <!--/descriptions-->
+
+  <!--subnav-->
+  <%= render "navigation" %>
+  <!--/subnav-->
+
+  <!--main content-->
+  <ul class="support banner index group">
+    <%= render "support_notice_blurb", support_notice: @support_notice %>
+  </ul>
```
sarken: This is just showing us one banner, so it should be `<%= render "support_notice", support_notice: @support_notice %>` now.

app/views/admin/support_notices/_support_notice.html.erb
sarken: Ah, sorry, that's my fault for not realizing it was using blockquote! (We'll need to make an issue to fix that in the admin banner at some point because using blockquote there is not even remotely correct.) It should be a div, and the notice as a whole should be set up like this:

```
<div class="<%= css_classes_for_support_notice(support_notice) %>">
  <div class="userstuff">
    <%= raw sanitize_field(support_notice, :notice_content, image_safety_mode: true) %>
  </div>
</div>
```

There are a few exceptions in older parts of the code, but we generally keep `userstuff` separate from other classes. (So it should come out of `css_classes_for_support_notice(support_notice)`.)

The main reason for putting the notice inside the main `<div class="userstuff">` was to restrict the notice's width: `.system .userstuff` has a maximum width of 72em to help with readability of text-heavy pages. However, I think moving the notice outside that div is a good call: it makes more sense for its width to match the form below it that the block of text above it. 

But to stop the width of the text inside the notice from getting constrained to 72em, we'll need to add this to [16-zone-system.css](https://github.com/otwcode/otwarchive/blob/master/public/stylesheets/site/2.0/16-zone-system.css#L19), right after `.system .userstuff h3` :

```
.system .notice .userstuff {
  max-width: initial;
}
```

marcus8448: I've applied the CSS to the `.support.banner` class (instead of `.notice`) and added a margin to the notice so that it matches the width of the form.

app/views/admin/support_notices/show.html.erb
marcus8448: I've removed the blurb styling but kept the actions (edit/delete) below the notice instead of in the navigation. I think it's nicer this way as the navigation buttons won't shift around on the show/edit pages, but it's easy to move if need be.

app/views/admin/support_notices/_support_notice.html.erb
sarken: Was `.notice` causing issues somewhere? I don't _think_ we use userstuff in most notices, but it's still a fix we'd ideally want applied everywhere -- there's never going to be a good reason to restrict the width of a notice's content to 72em while the notice itself can be bigger.

We usually let the notice be a bit wider than the form to help indicate that it encompasses the form (e.g., the required field notice on the new work form), but arguably that might not apply here, so we can leave the margin change in. However, the style should be moved to the `/* 5. ZONE: SYSTEM: SUPPORT */` section and the selector changed to `.feedbacks .banner` or similar (we try to avoid chaining class names), since it applies specifically to the banner as shown on the feedbacks page.

app/views/admin/support_notices/show.html.erb
sarken: Please put the navigation at the top. We generally only put actions under things when we're dealing with multiple items on a page. When it's one item, we want the actions in the subnavigation.

sarken:
(We probably won't get to this until after Rails and Elasticsearch, so no rush!)

app/views/admin/support_notices/_support_notice.html.erb
marcus8448: It was to try and avoid potential issues, since `notice` seems to be widely used. I've changed the selector back. (From a quick search there's no other `userstuff`s within a `notice` anyways)

public/stylesheets/site/2.0/16-zone-system.css
```diff
@@ -333,6 +337,10 @@ admin role is only admin facing, and changes normal views in role-admin.css
   margin: 0.643em;
 }
 
+.feedbacks .banner {
+  margin: 0.643em;
+}
```
sarken: Please combine this with `.feedbacks form .notice` directly above since they're applying the same style.

sarken:
Good catch on [f7238c9](https://github.com/otwcode/otwarchive/pull/5389/commits/f7238c95182b9657f390576bc089d6fec2280fa5)! Could you add a test to cover that scenario?

sarken (approved):

PR #5393, opened 2025-10-06
AO3-7104 Open doors claim external authors

anmazz:
## Issue

https://otwarchive.atlassian.net/browse/AO3-7104

## Purpose

Updates, internationalizes, and pluralizes the external authors claim page with Open Doors updates.

I've tried to reduce # of keys where the only difference was (lowercase) work/works with `work: t(".work", count: works_count)`. Any more than just work/works I made separate keys. I can revert this to make it explicitly 2 keys every time if that's more confusing or overkill.

## Credit

anna


app/views/external_authors/claim.html.erb
```diff
@@ -1,121 +1,122 @@
...
+      <%= button_to t(".add_works_to_account_button", current_user: current_user.login, count: works_count), complete_claim_path(invitation_token: @invitation.token) %>
     <% else %>
-      <%= button_to ts("Sign me up and give me my works!"), signup_path(invitation_token: @invitation.token), method: :get %>
+      <%= button_to t(".sign_up_button", count: works_count), signup_path(invitation_token: @invitation.token), method: :get %>
     <% end %>
   </p>
 </div>
 
-<h3 class="heading" id="other-options"><%= ts("Other Options") %></h3>
+<h3 class="heading" id="other-options"><%= t(".other_options") %></h3>
```
Bilka2:
```suggestion
<h3 class="heading" id="other-options"><%= t(".other_options.heading") %></h3>
```

app/views/external_authors/claim.html.erb
```diff
@@ -1,121 +1,122 @@
...
 <%= form_for @external_author do |f| %>
   <fieldset>
-    <legend><%= ts("Other Options") %></legend>
-    <h4><%= ts("Actions")%></h4>
+    <legend><%= t(".other_options") %></legend>
```
Bilka2:
```suggestion
    <legend><%= t(".other_options.legend") %></legend>
```
To differentiate

app/views/external_authors/claim.html.erb
```diff
@@ -1,121 +1,122 @@
...
       <li>
         <%= f.check_box :do_not_email %>
-        <%= f.label :do_not_email, ts("Do not email me in the future when works are imported with this email address.") %>
+        <%= f.label :do_not_email, t(".notifications.do_not_email") %>
       </li>
 
     </ul>
     <p class="important message">
-      <%= ts("Important: If you tick \"Do not email me in future\" but do not tick \"Do not import works\", works may be imported without you being told. If that's not what you want, turn off importing as well as emails.") %>
+      <%= t(".important_message") %>
```
Bilka2: Could this be a less generic key? Maybe something like "import without email"

config/locales/views/en.yml
```diff
@@ -826,6 +826,82 @@ en:
...
+      my_imported:
+        one: My Imported Work
+        other: My Imported Works
+      notes:
+        claim:
+          one: You can claim your work.
+          other: You can claim your works.
+        create_account:
+          one: You can create an account and claim your work (if you already have an account, log in to claim it)
+          other: You can create an account and claim your works (if you already have an account, log in to claim them).
```
Bilka2, line 879:
```suggestion
          other: You can create an account and claim your works (if you already have an account, log in to claim them)
```

Bilka2 (changes requested):
Thanks for that plural fix! I got a few more notes. I also asked Open Doors which FAQ they mean in "Refer to the Orphaning FAQ" (I suspect they mean https://archiveofourown.org/faq/orphaning, not the one you linked), I'll get back to you with their answer once I get it!

config/locales/views/en.yml
lydia-theda: Shouldn't it be the other way around (add the period to the singular version) because this is one of several bullets where the others all end in periods? (Admittedly, I'm only skimming the changelog here, not looking at the actual page, so I might be wrong.)

Bilka2: Sure, let's go with that

Bilka2:
Open Doors got back to me, the orphaning FAQ should point to https://archiveofourown.org/faq/orphaning please!

PR #5395, opened 2025-10-07
Add workers to docker compose config

brianjaustin:
# Pull Request Checklist

## Issue

Effectively a documentation change, so no Jira ticket

## Purpose

Add workers to the docker compose configuration so indexing (and other background/Resque tasks) work more easily.

## Testing Instructions

Does the init script still result in a working environment? Does `docker compose --profile dev up -d --build` still work as expected?

## References

https://github.com/otwcode/otwarchive/pull/5179#discussion_r2399866201

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -87,9 +87,22 @@ services:
       - redis
       - es
       - mc
+      - workers
     # Make `docker compose attach web` work for debugging
     stdin_open: true
     tty: true
+  workers:
+    profiles:
+      - dev
+    build:
+      context: .
+      dockerfile: ./config/docker/Dockerfile
+    command: bundle exec rake environment resque:work
+    volumes:
+      - .:/otwa
+    environment:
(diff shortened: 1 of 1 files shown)


Bilka2:
It works exactly as I hoped and running init.sh now results in filled elasticsearch indexes because the workers are running during the initial indexing. And since the workers are automatically started now, we also get anything that's queued with `deliver_later` automatically happening :tada:

We'll have to update https://github.com/otwcode/otwarchive/wiki/Docker and https://github.com/otwcode/otwarchive/wiki/Development-Data for the extra container to recreate on gemfile change and to make clear that now the workers should already be running, but the scheduler isn't.

Bilka2 (approved):

Bilka2:
Tested again after merging, getting those prefilled indexes from the init script is so nice :D

Bilka2:
Docs updated

PR #5397, opened 2025-10-08
AO3-7142 Pull collections count from ES

anmazz:
## Issue

https://otwarchive.atlassian.net/browse/AO3-7142

## Purpose

Pulls collection counts from ES

Need some help here (hence draft PR) - testing locally works fine, matches count. I however am massively struggling here to get the cucumber tests to work. I think it's because the collection is not being indexed but I'm not entirely sure how/why. I might have missed it but was looking for a similar test either rspec or cucumber that tested work count or something similar and wasn't able to, but if there's a reference that would be awesome. 

Few other questions about whether the cache key should be different since it's really similar to the other collections cache key and whether to use `dashboard_cache_options` or `collection_cache_options`. Help very much appreciated, thanks!

## Credit

anna

### DIFF
--- a/app/models/search/search_counts.rb
+++ b/app/models/search/search_counts.rb
@@ -72,6 +72,20 @@ module SearchCounts
   end
 
   ######################################################################
+  # COLLECTION COUNTS FOR USER/PSEUD DASHBOARD
+  ######################################################################
+
+  def collection_count_for_user(user)
+    Rails.cache.fetch(["user", user.id, "collections_count"], dashboard_cache_options) do
+      CollectionSearchForm.new(maintainer_id: user.id, sort_column: "title.keyword").search_results.scope(:for_search).total_entries
+    end
+  end
+
+  def collections_cache_key(owner)
+    "collections_count_#{owner.model_name.cache_key}_#{owner.id}_#{logged_in}"
+  end
+
+  ######################################################################
   # BOOKMARK COUNTS FOR USER/PSEUD DASHBOARD
   ######################################################################
 
--- a/features/collections/collection_participants.feature
+++ b/features/collections/collection_participants.feature
@@ -23,6 +23,22 @@
   When I click the 2nd button
   Then I should see "Removed sam from collection."
 
+  Scenario: Owner can invite own pseud to the collection
+  Given I have the collection "Such a nice collection"
+    And I am logged in as the owner of "Such a nice collection"
+    And "moderator" creates the pseud "moderator_pseud"
+  Given I debug collection count
+  When I follow "My Dashboard"
+    And the dashboard counts have expired
+  Then I should see "Collections (1)"
+  When I am on the "Such a nice collection" participants page
+    And I fill in "participants_to_invite" with "moderator_pseud (moderator)"
+    And I press "Submit"
+  Then I should see "New members invited: moderator_pseud (moderator)"
+  # When I follow "My Dashboard"
+  #   And the dashboard counts have expired
+  # Then I should see "Collections (1)"
+
   Scenario: Owner can't invite a nonexistent user to the collection
   Given I have the collection "Such a nice collection"
     And I am logged in as the owner of "Such a nice collection"
--- a/features/step_definitions/collection_steps.rb
+++ b/features/step_definitions/collection_steps.rb
@@ -1,5 +1,10 @@
 ### GIVEN
 
+Given "I debug collection count" do
+  user = User.find_by(login: "moderator")
+  puts "Collection count: #{SearchCounts.collection_count_for_user(user)}, user id #{user.id}"
+end
+
 Given /^I have no collections$/ do
   Collection.delete_all
 end
--- a/app/helpers/users_helper.rb
+++ b/app/helpers/users_helper.rb
@@ -46,6 +46,12 @@ module UsersHelper
     user.coauthors.collect(&:name).join(', ')
   end
 
+  # Prints link to collections page with user-appropriate number of collections
+  def collections_link(user)
+    total = SearchCounts.collection_count_for_user(user)
+    span_if_current t(".pitch.collections", coll_number: total.to_s), user_collections_path(@user)
+  end
+
   # Prints link to bookmarks page with user-appropriate number of bookmarks
   # (The total should reflect the number of bookmarks the user can actually see.)
   def bookmarks_link(user, pseud = nil)
--- a/app/views/users/_sidebar.html.erb
+++ b/app/views/users/_sidebar.html.erb
@@ -32,7 +32,7 @@
 	  <li><%= bookmarks_link(@user, @pseud) %></li>
 	<% end %>
 
-  <li><%= span_if_current t(".pitch.collections", coll_number: @user.maintained_collections.count), user_collections_path(@user) %></li>
+  <li><%= collections_link(@user) %></li>
 </ul>
 
 <% if @user == current_user || policy(:inbox_comment).show? %>


features/collections/collection_participants.feature
```diff
@@ -23,6 +23,22 @@
...
   Then I should see "Removed sam from collection."
 
+  Scenario: Owner can invite own pseud to the collection
+  Given I have the collection "Such a nice collection"
+    And I am logged in as the owner of "Such a nice collection"
+    And "moderator" creates the pseud "moderator_pseud"
+  Given I debug collection count
+  When I follow "My Dashboard"
+    And the dashboard counts have expired
+  Then I should see "Collections (1)"
```
sarken: Did you try adding `And all indexing jobs have been run` before expiring the counts?

anmazz: Yeah `the dashboard counts have expired` contains the indexing step

sarken: Okay, now that I have slept, I see the problem! You need to do `the dashboard counts have expired` _before_ you actually go to the dashboard. Indexing and expiring while you're already on the page won't show you the updated counts.

And just a quick note about the very end of the test, which you may or may not already know: moderator_pseud is only invited, which would not cause a test failure without your fix. The collection is only counted twice if the pseud is an owner or a moderator, so you'll need to set the correct role before checking the collection counts.

anmazz: @sarken ah thank you sm! exactly what I needed. and yes I'll make sure to update to be a collection owner - mentally made a note to do that and completely forgot!

config/locales/helpers/en.yml
```diff
@@ -37,6 +37,9 @@ en:
   inbox_helper:
     comment_link_with_chapter_number: Chapter %{position} of %{title}
   users_helper:
+    collections_link:
+      pitch:
+        collections: Collections (%{coll_number})
```
anmazz: There’s probably a better place to put this so please let me know

sarken: I think there are two options here.

First one, something like
```yml
users_helper:
  sidebar:
    collections: Collections (%{coll_number})
```

Then we could add others as we internationalize them, e.g. the text for `bookmarks_link` would be `users_helper.sidebar.bookmarks`.

However, we use some of those links in other places /o\ For example, `bookmarks_link` is also used at the bottom of the bookmarks listbox on your dashboard. So that might be weird and confusing.

Given that, maybe we should keep it simple and just use the the method names:
```yml
users_helper:
    collections_link: Collections (%{coll_number})
```

app/models/search/search_counts.rb
```diff
@@ -71,6 +71,20 @@ def work_cache_key(owner)
...
+
+  def collection_count_for_user(user)
+    Rails.cache.fetch(["user", user.id, "collections_count"], dashboard_cache_options) do
+      CollectionSearchForm.new(maintainer_id: user.id, sort_column: "title.keyword").search_results.scope(:for_search).total_entries
+    end
+  end
+
+  def collections_cache_key(owner)
+    "collections_count_#{owner.model_name.cache_key}_#{owner.id}_#{logged_in}"
+  end
```
sarken: It doesn't look like this is being used, which is fine. Unlike with bookmarks or works, we don't need to change the number based on who is looking at the page. This method can be removed.

features/collections/collection_participants.feature
```diff
@@ -23,6 +23,23 @@
...
+    And I have the collection "Woah another collection"
+    And I am logged in as the owner of "Such a nice collection"
+    And "moderator" creates the pseud "moderator_pseud"
+  When the dashboard counts have expired
+    And I follow "My Dashboard"
+  Then I should see "Collections (2)"
+  When I am on the "Such a nice collection" participants page
+    And I fill in "participants_to_invite" with "moderator_pseud (moderator)"
+    And I select "Owner" from "moderator_role"
+    And I press "Submit"
```
sarken: This isn't quite the right order of steps to make moderator_pseud an owner. If you try doing this on the page, you'll find you need to fill in the "Add new members" form and press its "Submit" button before the new member is added to the section where you can select roles. That's what we do in the test right above this: https://github.com/otwcode/otwarchive/blob/61e0363fd62c64950d7475fc113cda2025df1bf9/features/collections/collection_participants.feature#L16-L22

But to make things more complicated, the select menus for the roles are based on username, not pseud:
https://github.com/otwcode/otwarchive/blob/61e0363fd62c64950d7475fc113cda2025df1bf9/app/views/collection_participants/_participant_form.html.erb#L9

Because of that, the `id` for both pseuds' select menus is the same -- `moderator_role` -- so we can't rely on that to choose the right select menu.

That's a problem we need to fix, but it's a little outside the scope of this issue. Instead, let's define a new step for updating roles that relies on the `id` of the containing `li`, which is based on the participant ID rather than the username. 

Something like:

```cucumber
# Usage:
# - I give "user" the "Owner" role in the "Collection Title" collection
# - I give "pseud (user)" the "Member" role in the "Collection Title" collection

When "I give {string} the {string} role in the collection {string}" do |byline, role, collection|
  pseud = Pseud.parse_byline(byline)
  collection = Collection.find_by(title: collection)
  participant_id = CollectionParticipant.find_by(pseud: pseud, collection: collection).id
  selector = "#participant_#{participant_id}"
  step %{I select "#{role}" from "#{pseud.user.login}_role" within "#{selector}"}
  step %{I press "Update" within "#{selector}"}
end
```

anmazz: thank you!! that makes sense, appreciate the code sample. also used it to adapt it into a Then step.

app/models/search/search_counts.rb
```diff
@@ -71,6 +71,16 @@ def work_cache_key(owner)
     "work_count_#{owner.model_name.cache_key}_#{owner.id}_#{logged_in}"
   end
 
+  ######################################################################
+  # COLLECTION COUNTS FOR USER/PSEUD DASHBOARD
+  ######################################################################
+
+  def collection_count_for_user(user)
+    Rails.cache.fetch(["user", user.id, "collections_count"], dashboard_cache_options) do
+      CollectionSearchForm.new(maintainer_id: user.id, sort_column: "title.keyword").search_results.scope(:for_search).total_entries
```
sarken: Did you try `CollectionQuery.new(maintainer_id: @user.id).count` here, by any chance? That seems to be more in line with what the other count methods are using, most likely because it returns the count by itself rather than the documents that match the search. 

I think `total_entries` might also risk running into the fact Elasticsearch will only return 10k entries even if there are more than match the query ([AO3-5232](https://otwarchive.atlassian.net/browse/AO3-5232)), although I hope we never have a situation where that's relevant 😅

features/step_definitions/collection_steps.rb
```diff
@@ -323,3 +332,13 @@
   visit user_path(work.users.first)
   expect(page).not_to have_content(title)
 end
+
+Then "{string} has the {string} role in the collection {string}" do |byline, role, collection|
```
sarken: This reads a bit more like a setup (`Given`) step. Can you rephrase it to "{string} should have the {string} role..." so it's clearer we're checking expected results, not providing background?

app/models/search/search_counts.rb
anmazz: Great point. I just pulled the code exactly from how the h3 was displaying on the user's collection page... and looking at it more closely yeah there's a lot that's not strictly necessary here to display the count. This is outside the scope of this ticket, but just curious— would/should the h3 collection count need to be changed at some point as well since it's the same code as above (pulling `total_entries` from the search query)?

sarken: Someone might need to correct me on this, but I don't think there's any reason to change it there. 

Unlike the count in the sidebar, the counts in the header are always displayed on the same page as the results of a search. It makes sense to reuse the search we've already constructed to construct our header, especially since we are also listing the offset, e.g., 21-36 of 36 Items. And using `unlimited_total_entries` instead of just `total_entries` ensures the header lists the _real_ total even when there are more than 10k results.

PR #5400, opened 2025-10-08
AO3-5517 Use delete_all for kudos and comments for works and chapters

weeklies:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5517

## Purpose

Use `delete_all` rather than `destroy` for comments and kudos on works and chapters.

## Testing Instructions

Check that deleting a work or chapter:
- Succeeds for normal works and works with large number of comments/kudos
- Causes comment links to 404
- Clears associated inbox comments in inbox
- (chapter only) Updates comment count for the work 
- Changes user kudos count in Statistics page

### DIFF
--- a/app/models/chapter.rb
+++ b/app/models/chapter.rb
@@ -11,7 +11,7 @@ class Chapter < ApplicationRecord
   # acts_as_list scope: 'work_id = #{work_id}'
 
   acts_as_commentable
-  has_many :comments, as: :commentable
+  has_many :comments, as: :commentable, dependent: :nil # Handled in #delete_all_comments
 
   validates_length_of :title, allow_blank: true, maximum: ArchiveConfig.TITLE_MAX,
     too_long: ts("must be less than %{max} characters long.", max: ArchiveConfig.TITLE_MAX)
@@ -52,7 +52,14 @@ class Chapter < ApplicationRecord
   scope :in_order, -> { order(:position) }
   scope :posted, -> { where(posted: true) }
 
+  before_destroy :fix_positions_before_destroy, :invalidate_chapter_count, :delete_all_comments
+  after_destroy :update_work_stats
+
   after_save :fix_positions
+  after_save :invalidate_chapter_count,
+    if: Proc.new { |chapter| chapter.saved_change_to_posted? }
+  after_commit :update_series_index
+
   def fix_positions
     if work&.persisted?
       positions_changed = false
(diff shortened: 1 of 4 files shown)


app/models/chapter.rb
```diff
@@ -5,11 +5,13 @@ class Chapter < ApplicationRecord
   include WorkChapterCountCaching
   include CreationNotifier
   include Creatable
+  include Responder
 
   belongs_to :work, inverse_of: :chapters
   # acts_as_list scope: 'work_id = #{work_id}'
 
   acts_as_commentable
+  has_many :comments, as: :commentable # Handled in #delete_all_comments
```
Bilka2, line 14: Note for self that should go into a code comment if we keep this approach: This overrides the `dependent: :destroy` from `ActsAsCommentable::CommentableEntity`. Setting `dependent: nil` explicitly may be clearer. Same for the same line in work.rb

app/models/chapter.rb
```diff
@@ -50,7 +52,13 @@ def inherit_creatorships
   scope :in_order, -> { order(:position) }
   scope :posted, -> { where(posted: true) }
 
+  before_destroy :fix_positions_before_destroy, :invalidate_chapter_count, :delete_all_comments
+  after_destroy :update_work_stats
```
Bilka2, line 56: Note for self: For updating the comment count on the work

Does this mean that `expire_comments_count` on the work should also be called?

spec/controllers/works/default_rails_actions_spec.rb
```diff
@@ -900,12 +902,35 @@ def call_with_params(params)
         fake_login_known_user(work.users.first)
       end
 
-      it "deletes the work and redirects to the user's works with a notice" do
+      it "deletes the work and its associations, and redirects to the user's works with a notice" do
```
Bilka2, line 905: Could also check for the `creatorships` getting destroyed here please

Bilka2 (changes requested):
I think it would be nice to add comments near/on the destroy callbacks in comment.rb (and kudos.rb I suppose) that we're skipping the callbacks here, since otherwise we might add something in the future there and then get surprised when it doesn't execute for chapter deletions

Due to that difficulty, do you think it would be viable to go with `dependent: :destroy_async` instead? [It seems made for problems like this bug](https://guides.rubyonrails.org/association_basics.html#dependent), but possibly some other parts of the code don't like if we delete the work/chapter before delete all the comments/kudos on them...?

PR #4481, opened 2023-03-15
AO3-6478 checkbox logic includes overlaps

ciaraou:
# Pull Request Checklist

* [x] Have you read ["How to write the perfect pull request"](https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/)?
* [x] Have you read the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [x] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)?
* [x] Have you added the [Jira](https://otwarchive.atlassian.net) issue number
  as the *first* thing in your pull request title (e.g. `AO3-1234 Fix thing`)
* [x] Do you have fewer than 5 pull requests already open? If not, please wait
  until they are reviewed and merged before creating new pull requests.

## Issue

https://otwarchive.atlassian.net/browse/AO3-6478

## Purpose

When populating checkboxes in prompt memes/gift exchanges the method checkbox_section accepts a list of checked tags in string form. .include? on that string allowed tags that were substrings of checked tags to appear checked even if they were not originally. Splitting that string to an array ensures it checks equality and not substring. (Both splitting the string before it's passed in and changing ".include?" to ".eq?" result in errors elsewhere)

## Testing Instructions

Create a new collection either gift exchange or prompt meme and add a tag set where at least one tag is a substring of another (ex: "diana/bruce/clark" and "diana/bruce"). Sign up for your challenge with the larger tag selected. Edit your sign up. You should see only the tag you selected checked, and the substring tag unchecked. 

## Credit

pinkpurpleblue (she/her)

PR #5441, opened 2025-10-31
AO3-6478 Checkbox overlaps

anmazz:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6478

## Purpose

Adopts https://github.com/otwcode/otwarchive/pull/4481
Addresses PR feedback from original.

## References

https://github.com/otwcode/otwarchive/pull/4481

## Credit

pinkpurpleblue (she/her) - coder
Brian Austin - reviewer
anna


app/helpers/application_helper.rb
```diff
@@ -456,14 +456,22 @@ def checkbox_section(form, attribute, choices, options = {})
...
-      when options[:checked_method].nil?
-        []
-      else
-        form.object.send(options[:checked_method]) || []
-      end
+    already_checked = if options[:checked_method].is_a?(Array)
+                        options[:checked_method]
+                      elsif options[:checked_method].nil?
+                        []
+                      elsif attribute.is_a?(String) && attribute.include?("tagnames")
```
Bilka2, line 463: I don't like this solution, it feels very tacked on, because it's so hardcoded to tags and potentially fragile if we changed separator character.

I think it would be cleaner to modify the `is_checked = !options[:checked_method]` line to check for whether `already_checked` is a string and then check `eql?` or `==` instead of `include?`. We can probably get rid of the `already_checked.empty?` check too, since it's fine to call `include?` on an empty array/string

features/step_definitions/challenge_steps.rb
```diff
@@ -68,6 +68,14 @@
     step %{a canonical fandom "Unrequested"}
 end
 
+Given "I have overlapping challenge tags setup" do
+  step "I have standard challenge tags setup"
```
Bilka2, line 72: This seems like it does a whole lot of extra set up. I think it'd be fine to just do something like `I set up the tag set "Overlapping Challenge Tags" with the relationship tags "Sam Carter/Daniel Jackson, Daniel Jackson/Jack O'Neill, Sam Carter/Daniel Jackson/Jack O'Neill"` here and skip the next step

features/step_definitions/challenge_steps.rb
```diff
@@ -68,6 +68,14 @@
...
+  step %{I add the relationship tags "Sam Carter/Daniel Jackson, Daniel Jackson/Jack O'Neill, Sam Carter/Daniel Jackson/Jack O'Neill" to the tag set "Standard Challenge Tags"}
+  step %{a canonical relationship "Daniel Jackson/Jack O'Neill"}
+  step %{a canonical relationship "Sam Carter/Daniel Jackson"}
+  step %{a canonical relationship "Sam Carter/Daniel Jackson/Jack O'Neill"}
```
Bilka2, line 76: I think these are fine to have non-canonical, at least when I reproduced manually locally I did it without canonicals

features/step_definitions/challenge_promptmeme_steps.rb
```diff
@@ -26,6 +26,18 @@
   step "I fill in single-prompt challenge options"
 end
 
+Given "I have prompt meme {string} fully set up with overlapping tags" do |challengename|
+  step %{I am logged in as "mod1"}
+  step "I have overlapping challenge tags setup"
+  step %{I set up the collection "#{challengename}"}
+  step %{I select "Prompt Meme" from "challenge_type"}
+  step %{I submit}
+  step %{I fill in "General Sign-up Instructions" with "Here are some general tips"}
```
Bilka2, line 35: Unneeded?

features/step_definitions/challenge_gift_exchange_steps.rb
```diff
@@ -212,6 +223,12 @@
     click_button "Submit"
 end
 
+When "I sign up for {string} with combination E" do |title|
```
Bilka2, line 226: Could this and the other combination step have a more meaningful name?

Bilka2 (changes requested):
I think it would be good to also test tag set editing with overlapping names of one of ratings, warnings or categories, since it looks like those would have the same issue:
Screenshot_20251216_221806

app/helpers/application_helper.rb
anmazz: @Bilka2 yeah I'll be honest when I picked this up I didn't dig too deep into the code as it stood in the original PR since Brian had previously reviewed it (and assumed it was fine except for the test edits). Given the rest the comments I'll try to look at this one more thoroughly

features/step_definitions/challenge_gift_exchange_steps.rb
anmazz: @Bilka2 yeah I saw this and had no idea what any of the "Combination A/B/C" etc set ups actually meant. Seems like it's kind of just equivalent to "Combination 1/2/3" and has no real meaning in and of itself until you look at the step? I wasn't really clear

Bilka2: Yeah that's basically also my issue with it, it feels too abstract and hidden away. I'd also be fine with not making these steps at all and just having these signups directly in the scenario, then it'd also be more clear why we're checking the other tags. Or this could be something like "I sign up for {string} with Sam Carter/Daniel Jackson/Jack O'Neill", either works for me

Bilka2: (also it looks like the prompt and gift exchange one do the same thing, another argument imo to not have them named like this)

PR #5487, opened 2025-11-28
AO3-7143 Adding filter to user collection page

ReverM:
## Issue

https://otwarchive.atlassian.net/browse/AO3-7143

## Purpose

This adds the ability to filter collections on user pages like on the generic collections page.

## Credit

Danaël / Rever (they / he)


ReverM:
To note, I am currently working on implementing cucumber tests to test this feature, but I ran into a pickle (pun intended) while doing so. The feature is complete currently, but if you would prefer if there are unit tests for this feature, I shall work on them promptly.

anmazz:
> To note, I am currently working on implementing cucumber tests to test this feature, but I ran into a pickle (pun intended) while doing so. The feature is complete currently, but if you would prefer if there are unit tests for this feature, I shall work on them promptly.

(I'm not an official volunteer, so please defer to them if they disagree with me!)

Yes, from what I've seen volunteers want to see cucumber and/or rpsec tests for functionality changes, so I would definitely implement those here. Automated testing is in the Pull Request Checklist when you open a PR, more about it here on the wiki: https://github.com/otwcode/otwarchive/wiki/Automated-Testing.

You're probably already on it, but also make sure you're passing the Rubocop and ERB Linting checks. Unfortunately touching some older lines of code will often cause these checks to fail, but good news is you just need to fix the added/changed lines and not the whole file. (if you happen to be using VSCode as your IDE, I personally use the Ruby LSP extension by Shopify and it's really good at pointing out the Rubocop issues - just throwing that out there in case you used VSCode).

ReverM:
As per https://github.com/otwcode/otwarchive/wiki/Reviewdog-and-RuboCop the RuboCop errors regarding regex seem to be normal. Am I to be doing anything with that regard?

Let me know if more tests are needed as well

features/step_definitions/collection_steps.rb
```diff
@@ -4,20 +4,20 @@
...
+Given /^the collection "([^"]*)" is deleted$/ do |collection_title|
   step %{I am logged in as the owner of "#{collection_title}"}
   visit edit_collection_path(Collection.find_by(title: collection_title))
   click_link "Delete Collection"
   click_button "Yes, Delete Collection"
   page.should have_content("Collection was successfully deleted.")
 end
 
-When /^I am logged in as the owner of "([^\"]*)"$/ do |collection|
+When /^I am logged in as the owner of "([^"]*)"$/ do |collection|
```
anmazz: If you do want to update steps like these, this can be [updated with the new syntax:](https://github.com/cucumber/cucumber-expressions?tab=readme-ov-file#parameter-types)
`When "I am logged in as the owner of {string}" do |collection|`

I will defer to official volunteers here for specific advice and guidance on which of these methods is preferred in this case, but there are a few ways to go about the rubocop issues:
1. Revert all the step changes that remove the `\` since they're not strictly necessary changes, or
2. Update the simpler steps with the new {string} syntax, and disable rubocop on the steps that can't be updated (specifically the more complex steps like `Given /^(?:I have )?(?:a|an|the) (hidden)?(?: )?(anonymous)?...` -> these don't have a new syntax equivalent that's easy to rewrite.)

I'm not totally sure if rubocop can be failing before a PR is approved if it's just regex step names or not, so unfortunately I can't offer guidance there. For me personally, I find it easier to fix the errors where I can and disable rubocop if the regex can't be updated, that way if rubocop fails I can see the other issues that are true code smells.

ReverM: Ah I see, thank you I shall update these today

features/step_definitions/collection_steps.rb
```diff
@@ -82,6 +82,11 @@
   collection.collection_preference.update_attribute(:closed, true) if closed.present?
 end
 
+Given "{string} owns the collection {string} with name {string}" do |owner, title, name|
```
brianjaustin: The wording here makes it sound like it changes the owner, rather than create a new collection. What about the following?

```suggestion
Given "a collection {string} with name {string} owned by {string}" do |title, name, owner|
```

app/decorators/pseud_decorator.rb
```diff
@@ -76,28 +75,41 @@ def works_path
 
   def works_link
     return unless works_count > 0
+
     text = ActionController::Base.helpers.pluralize(works_count, "works")
     "<a href='#{works_path}'>#{text}</a>"
   end
 
+  def collections_path
```
brianjaustin: Do you know where these are used/why they're needed?

brianjaustin (dismissed):

app/decorators/pseud_decorator.rb
ReverM: I think I might have added those by accident. They don't look needed

features/step_definitions/collection_steps.rb
ReverM: I see yeah. I'll change it to your wording and change the relevant steps

ReverM: Done

brianjaustin (approved):
Thanks!

public/stylesheets/site/2.0/17-zone-home.css
```diff
@@ -63,11 +63,6 @@ user home, collections and challenges home, tag home, admin comms home
   float: none;
 }
 
-/* (no filters) */
-
-.dashboard.collections-index .index {
-  width: 100%;
-}
```
sarken, line 70: We actually still need this for the Subcollections page in collections.

sarken (changes requested):
Hi, Danaël!

No rush, but I wanted to give you a heads up that we've merged [AO3-7188](https://otwarchive.atlassian.net/browse/AO3-7188), which adds some new collection sorting options. It might be a good idea to merge your  branch with upstream master and make sure the new sort options are present and functional on the user collection pages.

I've also flagged an issue with the CSS that will need to be fixed even if everything is okay with AO3-7188. 

Thanks!

public/stylesheets/site/2.0/17-zone-home.css
ReverM: Could the filtering be added to the subcollection page instead?

ReverM:
> Hi, Danaël!
> 
> No rush, but I wanted to give you a heads up that we've merged [AO3-7188](https://otwarchive.atlassian.net/browse/AO3-7188), which adds some new collection sorting options. It might be a good idea to merge your branch with upstream master and make sure the new sort options are present and functional on the user collection pages.
> 
> I've also flagged an issue with the CSS that will need to be fixed even if everything is okay with AO3-7188.
> 
> Thanks!

Do we want more cucumber tests as a result?

[AO3-7188]: https://otwarchive.atlassian.net/browse/AO3-7188?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

ReverM:
I seem to have slightly messed up the commit order by rebasing. I will close the PR and reopen a new one. Apologies

PR #5494, opened 2025-12-01
AO3-7188 Allow collections to be sorted by number of works or bookmarked items

Kiyazz:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-7188

## Purpose

This PR will make it possible for collections to be sorted on their stats, specifically number of works and bookmarks. It will respect counts of works/bookmarks for private works when logged out

## References

AO3-7143 should also be able to recieve these changes if done first
EDIT: Since AO3-7143 is in review as well, there will be merge conflicts with it. Advice on how to resolve would be good

## Credit

kiyazz (he/him)


Kiyazz:
I know there are some rspec test cases failing right now. I'm not sure if this functionality is even possible to test with rspec since it's reliant on a column that only exists in elasticsearch and not in the database proper

Kiyazz:
Rspec is figured out, I'm still unsure about the one rubocop error I had to comment out. Any fix for that would cause test failures everywhere else and take tons of refactoring to solve so I thought it best to ignore that.

app/models/search/collection_search_form.rb
```diff
@@ -25,8 +25,9 @@ class CollectionSearchForm
     define_method(filterable) { options[filterable] }
   end
 
-  def initialize(opts = {})
+  def initialize(opts = {}, admin_logged_in = false) # rubocop:disable Style/OptionalBooleanParameter
```
redsummernight: You don't need to add a new parameter for the search form. Outside of controllers, to tell if a user (regular or admin) is logged in, you can check the presence of `User.current_user`, for example:

https://github.com/otwcode/otwarchive/blob/75323bbedeb0a9e33ef0621e4aaefe2eb428753c/app/decorators/collection_decorator.rb#L55-L61

app/models/search/collection_search_form.rb
```diff
@@ -64,7 +65,9 @@ def sort_direction
   def sort_options
     [
       ["Date Created", "created_at"],
-      ["Title", "title.keyword"]
+      %w[Title title.keyword],
+      ["Bookmarked Items", "public_bookmarked_items_count"],
```
redsummernight: The second value is the value of the option tag on the search page. We're gonna override it later anyway (in the query class, depending on who's logged in), so we can name them `bookmarked_items_count` and `works_count` and not imply anything about what's visible.

app/models/search/collection_query.rb
```diff
@@ -126,6 +126,7 @@ def sort_column
   end
 
   def sort
+    options[:sort_column] = sort_column.sub("public", "general") if (@user.present? || options[:admin_logged_in]) && sort_column.include?("public")
```
redsummernight: I'd avoid string substitution and write the column names in full, which is easier to read. You can see an example of changing column values in the TagQuery class:

https://github.com/otwcode/otwarchive/blob/75323bbedeb0a9e33ef0621e4aaefe2eb428753c/app/models/search/tag_query.rb#L47-L59

features/collections/collection_browse.feature
```diff
@@ -229,6 +229,52 @@ Feature: Collection
...
+    And I lock the work "Private 2"
+    And I bookmark the work "Private 1" to the collection "Publics"
+    And I bookmark the work "Private 2" to the collection "Publics"
+    And I post the work "Public 1" in the collection "Publics"
+    And I bookmark the work "Public 1" to the collection "Privates"
+    And all indexing jobs have been run
+    And I go to the collections page
+    And I select "Works" from "collection_search_sort_column"
+    And I press "Sort and Filter"
+  Then I should see the text with tags '<a href="/collections/Privates/works">2</a>'
```
redsummernight: These steps do not check whether the results are in the right order, only that they appear. If you can test sort order, you need to do something like:

https://github.com/otwcode/otwarchive/blob/75323bbedeb0a9e33ef0621e4aaefe2eb428753c/features/bookmarks/bookmark_filter.feature#L62-L65

https://github.com/otwcode/otwarchive/blob/75323bbedeb0a9e33ef0621e4aaefe2eb428753c/features/step_definitions/bookmark_steps.rb#L560-L565

features/collections/collection_browse.feature
```diff
@@ -229,6 +229,52 @@ Feature: Collection
...
+    And I should see the text with tags '<a href="/collections/Publics/works">1</a>'
```
redsummernight: In general, we prefer to refer to text that users can view, instead of checking raw HTML fragments.

redsummernight (changes requested):

app/models/search/collection_query.rb
```diff
@@ -126,12 +126,32 @@ def sort_column
   end
 
   def sort
+    column = if User.current_user.present?
```
redsummernight: Instead of 2 case statements, you can write:
```rb
case sort_column
when "bookmarked_items_count"
  User.current_user.present? ? "general_bookmarked_items_count" : "public_bookmarked_items_count"
when "works_count"
  User.current_user.present? ? "general_works_count" : "public_works_count"
else
  sort_column
end

```

features/collections/collection_browse.feature
```diff
@@ -229,6 +229,52 @@ Feature: Collection
...
+    And I lock the work "Private 1"
+    And I post the work "Private 2" in the collection "Privates"
+    And I lock the work "Private 2"
+    And I bookmark the work "Private 1" to the collection "Publics"
+    And I bookmark the work "Private 2" to the collection "Publics"
+    And I post the work "Public 1" in the collection "Publics"
+    And I bookmark the work "Public 1" to the collection "Privates"
+    And all indexing jobs have been run
+    And I go to the collections page
+    And I select "Works" from "collection_search_sort_column"
```
redsummernight: In feature tests, we refer to fields by their labels if possible. You can use "Sort by" here.

features/collections/collection_browse.feature
```diff
@@ -229,6 +229,52 @@ Feature: Collection
...
+    And I press "Sort and Filter"
+  Then the 1st collection result should contain "Works: 2"
```
redsummernight: You should make the tests stricter (and easier to read) by checking for the right collection title, next to each work/bookmark count check. For example, with bookmark sorting we have:

https://github.com/otwcode/otwarchive/blob/75323bbedeb0a9e33ef0621e4aaefe2eb428753c/features/bookmarks/bookmark_filter.feature#L70-L73

spec/models/search/collection_query_spec.rb
```diff
@@ -19,6 +19,28 @@
...
+    expect(q.generated_query[:sort]).to eq([{ "public_bookmarked_items_count" => { order: "desc" } }, { "id" => { order: "desc" } }])
+  end
+
+  it "sorts by works" do
+    q = CollectionQuery.new(sort_column: "works_count", sort_direction: "desc")
+    expect(q.generated_query[:sort]).to eq([{ "public_works_count" => { order: "desc" } }, { "id" => { order: "desc" } }])
+  end
+
+  it "sorts by general bookmarked items when logged in" do
+    User.current_user = create(:user)
```
redsummernight: You don't need to create and save an actual user here, since we only want to check that `User.current_user` exists for picking the right column. You can use [`User.current_user = build_stubbed(:user)`](https://thoughtbot.github.io/factory_bot/ref/build-strategies.html#build_stubbed) instead.

redsummernight (approved):
Looks good, thank you!

PR #5498, opened 2025-12-08
AO3 6998 - Error 500 when searching challenge sign-ups by pseud

ReverM:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6998

## Purpose

This removes unneeded sub-cases in the locales for sign-ups, which led to a 500 error.

## Credit

Danaël / Rever ( they / he )


ReverM:
I wanted to add a test for this functionality, but it seems that Scenario `challenge_giftexchange.feature:174` already does that.

Bilka2:
While this technically fixes the issue, this code is not properly i18n'ed - the pluralisation happens through `pluralize` doing a best guess, instead of being explicitly defined per locale.

Fixing this would require rewriting search_header - most likely completely because it also commits other i18n crimes, like blindly concatenating strings or simply inserting a word into another string instead of creating separate locale key variants for separate item types. The subscriptions email has some examples for how a situation like this can be properly handled in code. If you'd rather not do all these extra changes, please let me know and I will split this i18n work out into a separate Jira issue.

Speaking of Jira, please adjust either your Jira name or your credit name here on GitHub so that they match, it makes things much easier for us when compiling release notes.

As a sidenote, we recommend against picking up more Jira issue when you're already full on pull requests or have a bunch picked up, since we'd rather not have people overloading themselves. Doing so also denies other coders the chance to pick up and work on the issues they'd like if you are holding on to them, which isn't fair to them.

ReverM:
> While this technically fixes the issue, this code is not properly i18n'ed - the pluralisation happens through `pluralize` doing a best guess, instead of being explicitly defined per locale.
> 
> Fixing this would require rewriting search_header - most likely completely because it also commits other i18n crimes, like blindly concatenating strings or simply inserting a word into another string instead of creating separate locale key variants for separate item types. The subscriptions email has some examples for how a situation like this can be properly handled in code. If you'd rather not do all these extra changes, please let me know and I will split this i18n work out into a separate Jira issue.
> 
> Speaking of Jira, please adjust either your Jira name or your credit name here on GitHub so that they match, it makes things much easier for us when compiling release notes.
> 
> As a sidenote, we recommend against picking up more Jira issue when you're already full on pull requests or have a bunch picked up, since we'd rather not have people overloading themselves. Doing so also denies other coders the chance to pick up and work on the issues they'd like if you are holding on to them, which isn't fair to them.

I changed my display name on Jira, but I guess it is not my full name so it still displayed as my real name. 

Apologies for "hoarding" issues, I've unclaimed a bunch of them so that others have a chance at it. Are there other things I should be aware of regarding Jira handling? 

I shall go with Danaël Villeneuve moving forward to simplify things.

Bilka2:
> I changed my display name on Jira, but I guess it is not my full name so it still displayed as my real name.
> 
> Apologies for "hoarding" issues, I've unclaimed a bunch of them so that others have a chance at it. Are there other things I should be aware of regarding Jira handling?
> 
> I shall go with Danaël Villeneuve moving forward to simplify things.

How annoying that Jira isn't displaying that name, I see it now that I checked your profile! We can keep going with "Danaël / Rever" in the release note credits, that's no issue - we'd rather go with what you prefer. Maybe Jira will show the right name if you edit both or switch the two (if you're up for that)?

No worries about assigning the issues! Feel free to claim back the ones you already had set to in progress, since you presumably already did work on those. I just wanted to caution against picking up more since we'd rather not see you burn out (or get understandably frustrated when reviews or answering questions takes a while. We're a small volunteer team so things sometimes go slowly.)

I didn't notice anything else on Jira that stuck out to me - we try to document our processes, so generally there shouldn't be any nasty surprises, hopefully!

ReverM:
Given the fact this PR will require severe rework, I shall close it

Bilka2:
@ReverM After further consideration, I've split off the i18n work into https://otwarchive.atlassian.net/browse/AO3-7357. Feel free to reopen this PR

PR #5499, opened 2025-12-10
AO3-7214 AO3-7215 Bookmark Javascript Improvement

ReverM:
## Issue

https://otwarchive.atlassian.net/browse/AO3-7214
https://otwarchive.atlassian.net/browse/AO3-7215

## Purpose

This adresses both tickets 7214 and 7215 by making all buttons that could edit a bookmark disappear when the form is open, by allowing a closed form to be reopened (7215) and by fixing the opening and closing of multiple forms at the same time.
7214 was fixed by removing the variables and jquerying the appropriate component each time the form is closed or reopened.
7215 was fixed by changing the href of the opening buttons as part of the closing event, as well as changing the id to a class so every element is affected at the same time.

## Credit

Danaël / Rever ( they / he )

Jira account goes by Danaël Villeneuve


ReverM:
Appologies if having multiple ticket in one PR is bad etiquette. I decided to put both in the same PR, as they were extremely closely related, and would likely be waiting on each other if done separately.

slavalamp:
yeah, putting two tickets in one pr is fine when it's exact the same fix for both

though the pr name should begin like this: "AO3-7214 AO3-7215"

ReverM:
> though the pr name should begin like this: "AO3-7214 AO3-7215"

Changed the PR name accordingly

app/views/bookmarks/bookmark_form_dynamic.js.erb
```diff
@@ -2,21 +2,19 @@
...
-bookmark_open.hide();
+$j('.bookmark_form_trigger_for_<%= bookmark_form_id %>').hide();
 
-$j('#bookmark_form_close_for_<%= bookmark_form_id %>').click(function(){ 
-  bookmark_div.hide();
-  bookmark_open.show();
+$j('#bookmark_form_close_for_<%= bookmark_form_id %>').click(function(){
+  $j('#bookmark_form_placement_for_<%= bookmark_form_id %>').hide();
+  $j('.bookmark_form_trigger_for_<%= bookmark_form_id %>').show();
+  $j('.bookmark_form_trigger_for_<%= bookmark_form_id %>').attr('href', '#');
```
sarken: I was a little confused by this line. It didn't seem necessary at first, but when I removed it, I noticed that it seems to fix an issue introduced by the other changes in this file. 

Here's what I noticed:

Currently, if I go to [a bookmarks index](https://archiveofourown.org/collections/yuletide_recs/bookmarks), press "Save" on something I don't have bookmarked, enter some text in the form, and then use the "x" to close the form without saving, the form opens with my text entered the next time I press "Save." 

It seems like that behavior breaks when this line is removed: the second time I press "Save," the form gets reloaded and my text is lost. 

Is that indeed what this line is fixing? If so, do you know why that bug happens, and if there is any other way we could avoid it? Removing the original `href` means it's no longer possible to right-click "Save" and open the form in a new window, which is somewhat unfortunate. We can live with it if we have to, but it would be a good idea to have a code comment explaining why this line is needed.

sarken (changes requested):
Thanks for working on this! In addition to the question I left on the JavaScript file, I think you also need to change`id` to `class` for the buttons in the [bookmarkable_blurb](https://github.com/otwcode/otwarchive/blob/027cbf272402ca467a4c47cb7e4f0681c8342abc/app/views/bookmarks/_bookmarkable_blurb.html.erb#L16-L30), which is used for pages where we show multiple bookmarks under a single item, e.g., a tag's bookmark listing.

app/views/bookmarks/bookmark_form_dynamic.js.erb
ReverM: Which line in particular are you referring to?

As for the href, the reason why it's removed is the route of the bookmark is rendering the form as part of the controller (see line 176 of the bookmark controller). However, hiding with the x goes through the dynamic.js.erb file (above). If we don't do that, 7215 is reintroduced (since it will go through the controller to find the route of the link instead of the javascript watcher).

An alternate solution would be to delete the form instead of hiding it. That way, rendering would always work since the form would not be there already. Is the behavior you mentioned not already occurring if you close and open a bookmark on the works page having written stuff already?

sarken: >Which line in particular are you referring to?

The line removing the href, sorry! GitHub told me the comment was only on line 12 before I submitted the review, but it appears to have lied.

>If we don't do that, 7215 is reintroduced (since it will go through the controller to find the route of the link instead of the javascript watcher).

For me, 7215 remained fixed without this line. The only difference in behavior was the information I had entered in the form was lost upon re-opening it (which would be consistent with opening a new form instead of un-hiding the original). But based on your explanation, I suspect if I had re-opened the form twice instead of just once, I would've run into 7215. 

>An alternate solution would be to delete the form instead of hiding it.

That would mean losing any information you had entered in the form before closing it, so we don't want to do that. While we _could_ store that information somewhere and retrieve it when you re-open the form, it's probably not worth the extra code. If we get complaints about the removal of the right-click functionality, we can reconsider.

>Is the behavior you mentioned not already occurring if you close and open a bookmark on the works page having written stuff already?

The form currently (i.e., on production) keeps what I've written.

ReverM: I believe if you go on the works index, open a bookmark form and close it again you can't open it with a right click anymore. 

I've tried changing some stuff to make the right click work, but I'm unsure how to approach it.

sarken: I can reproduce that right click issue on production as well. I was going to say we can just leave this as-is in that case, but I _might_ have a solution.

Basically:

1. On the Save/Saved links, duplicate the path in a data attribute, e.g. `<%= link_to t(".saved"), edit_bookmark_path(current_user_bookmark), data: { path: edit_bookmark_path(current_user_bookmark) } , class: "bookmark_form_trigger_for_#{bookmarkable.id}", remote: true %>` in `bookmarkable_blurb.html.erb`
2. In the JavaScript, add `var path = $j(this).attr("data-path");`
3. Make sure we're not rendering the bookmark form if it has already been rendered: ```if (!bookmark_div.children().length) {
  bookmark_div.html("<%= escape_javascript(render "bookmarks/bookmark_form", :bookmarkable => @bookmarkable, :bookmark => @bookmark, :button_name => @button_name, :action => @action, :in_page => true, :dynamic => true) %>");
}```
4. Instead of replacing the `href` with `#`, replace it with `path`:   `$j('.bookmark_form_trigger_for_<%= bookmark_form_id %>').attr('href', path);` 

I haven't tested it thoroughly, though, and there may be a way to simplify storing the path for later use so we don't have to put it on all the links. 

If that (or that + a few small tweaks) doesn't work, then I think we can cope with the current right click behavior.

ReverM: Apart from making the path not a variable (since that would cause issues with opening and closing multiple bookmark form in a row), I believe this work.

ReverM:
I pushed changes to the gemfile.lock by accident. Disregard the Gem Updates tag

app/views/bookmarks/bookmark_form_dynamic.js.erb
```diff
@@ -2,21 +2,19 @@
 <% bookmark_form_id = (@bookmarkable.blank? ? "#{@bookmark.id}" : "#{@bookmarkable.id}") %>
 
 var bookmark_div = $j('#bookmark_form_placement_for_<%= bookmark_form_id %>');
-var bookmark_close = $j('#bookmark_form_close_for_<%= bookmark_form_id %>');
-var bookmark_open = $j('#bookmark_form_trigger_for_<%= bookmark_form_id %>');
 
-bookmark_div.html("<%= escape_javascript(render "bookmarks/bookmark_form", :bookmarkable => @bookmarkable, :bookmark => @bookmark, :button_name => @button_name, :action => @action, :in_page => true, :dynamic => true) %>");
-bookmark_open.hide();
+if (!bookmark_div.children().length) { bookmark_div.html("<%= escape_javascript(render "bookmarks/bookmark_form", :bookmarkable => @bookmarkable, :bookmark => @bookmark, :button_name => @button_name, :action => @action, :in_page => true, :dynamic => true) %>"); }
```
sarken: I didn't check -- does this work with the new `bookmarkable: @bookmarkable` syntax? If so, it'd be nice to fix that while we're touching this line.

app/views/bookmarks/bookmark_form_dynamic.js.erb
```diff
@@ -2,21 +2,19 @@
...
+$j('.bookmark_form_trigger_for_<%= bookmark_form_id %>').hide();
```
sarken: Could we switch this and the other lines we're touching to double quotes?

app/views/bookmarks/bookmark_form_dynamic.js.erb
```diff
@@ -2,21 +2,19 @@
...
-bookmark_open.attr('href', '#');
-$j("#bookmark_form_trigger_for_<%= bookmark_form_id %>").click(function(event){
-  bookmark_div.show();
-  bookmark_open.hide();
+$j(".bookmark_form_trigger_for_<%= bookmark_form_id %>").click(function(event){
+  $j('#bookmark_form_placement_for_<%= bookmark_form_id %>').show();
+  $j('.bookmark_form_trigger_for_<%= bookmark_form_id %>').hide();
   event.preventDefault();
-});
+});
```
sarken: Can you restore the newline at the end of the file?

config/locales/views/en.yml
```diff
@@ -536,15 +536,34 @@ en:
...
       from: From
       to: To
       word_count: Word count
     index:
       advanced_search: try our advanced search
       choose_fandom: choose a fandom
+      navigation:
+        bookmarks: Bookmarks
+        edit: Edit Bookmark
+        new: Boorkmark
```
sarken: Typo -- there's an "r" before the "k" (ngl, I would not have noticed this without the tests!)

features/bookmarks/bookmark_javascript.feature
```diff
@@ -0,0 +1,41 @@
...
+  As a humble user
+  I want to be able to edit bookmarks with javascript enabled
+  
+  Background:
+    Given a canonical fandom "The Bookmarks"
+      And the work "Bookmark: The Beginnings" by "bookmarker" with fandom "The Bookmarks"
+      And the work "Bookmark: The Sequel" by "bookmarker" with fandom "The Bookmarks"
+      And all indexing jobs have been run
+      And the dashboard counts have expired
+      And I am logged in as "bookmarker"
```
sarken: I think we could simplify this considerably by just logging in and creating a bookmark: we don't need a canonical fandom unless we're testing a tag bookmark page, and we don't need to update any dashboard counts because we're not testing any dashboard pages. 

We only need to run the indexing jobs if we're accessing pages that rely on Elasticsearch, but we have a workaround that lets us avoid constantly having to use `all indexing jobs have been run`. If you check out paths.rb, you'll see a lot of this: https://github.com/otwcode/otwarchive/blob/b23fe624793956ce5fbf2612dfe2e3456704797b/features/support/paths.rb#L128-L130

That way, instead of running indexing and following a link on the page, we can just say `When I go to username's bookmarks page`. (Not every page that relies on Elasticsearch is included in paths.rb, but you're always welcome to add any you need. It makes it easier for the next person who needs to test that page.)

However, I'm not sure we need any `Background` here at all. The second scenario seems to be broken in production and fixed in this branch with just one bookmark, so the bookmarks created in the background steps are unnecessary.

features/bookmarks/bookmark_javascript.feature
```diff
@@ -0,0 +1,41 @@
+@bookmarks
+Feature: Edit bookmarks with javascript enabled
```
sarken: Since there were a lot of places `id` needed to become `class`, and some of them were hard to track down, I think we need to be a bit more thorough with these tests. They should cover all of those locations, not just a few.

3a2d58f79714cf10fec0423660f3726a132953d3 has some tests to get started. I think the only thing that isn't covered there are the Save buttons in the bookmark and bookmarkable blurbs. Can you try adding tests for those? There might also be some tweaks you can make to the code in my commit, e.g., standardizing some variable names, removing some comments, moving the `Then` step definitions into the `Then` section of the file. (You'll also need to change "Boorkmark" to "Bookmark" once you fix the locale file -- I wanted to make sure the tests would pass, so I used the version with the typo.)

That said, I'm going to leave a couple comments on the tests you have here just to provide some information that might be helpful in the future.

features/bookmarks/bookmark_javascript.feature
```diff
@@ -0,0 +1,41 @@
...
+      And the dashboard counts have expired
+      And I am logged in as "bookmarker"
+
+  @javascript
+  Scenario: Opening multiple edit forms lets you close all of them (AO3-7214)
+    Given I am logged in as "recengine"
+      And I bookmark the work "Bookmark: The Beginnings"
+      And I bookmark the work "Bookmark: The Sequel"
+      And all indexing jobs have been run
+    When I follow "Bookmarks"
```
sarken: `When I go to the bookmarks page` would be better here, since it includes some cache expiration we wouldn't get by just following the link: https://github.com/otwcode/otwarchive/blob/b23fe624793956ce5fbf2612dfe2e3456704797b/features/support/paths.rb#L30-L34

app/views/bookmarks/_bookmarkable_blurb.html.erb
```diff
@@ -16,15 +16,9 @@
   <ul class="actions" role="navigation">
     <li>
       <% if current_user_bookmark ||= bookmark_if_exists(unwrapped) %>
-        <%= link_to ts("Saved"),
-                    edit_bookmark_path(current_user_bookmark),
-                    id: "bookmark_form_trigger_for_#{bookmarkable.id}",
-                    remote: true %>
+        <%= link_to t(".saved"), edit_bookmark_path(current_user_bookmark), data: { path: edit_bookmark_path(current_user_bookmark) }, class: "bookmark_form_trigger_for_#{bookmarkable.id}", remote: true %>
```
sarken: I think the multiline format was a little easier to read, if you could keep that here. (Reviewdog might have some opinions about the indenting, though. It usually takes me a few tries to figure out exactly what it wants.)

sarken (changes requested):

app/views/bookmarks/bookmark_form_dynamic.js.erb
sarken: Yay, I'm so glad to hear that!

config/locales/views/en.yml
ReverM: Oops yeah, that is not how you spell bookmark 😅

app/views/bookmarks/bookmark_form_dynamic.js.erb
ReverM: What exactly is this new syntax?

KooShnoo: I think sarken is talking about the hashmap syntax? https://ruby-doc.org/core-3.1.0/Hash.html#class-Hash-label-Hash+Data+Syntax

sarken: Right. It should be possible to use the JSON-style syntax from the example in my previous comment (`bookmarkable: @bookmarkable`) instead of the current hash rocket syntax (`:bookmarkable => @bookmarkable`).

ReverM:
Gonna put this to draft while I add the rest of the test coverage.

ReverM:
Some of the steps in the test are very inconsistent. I'm unsure what could be the cause. Is cucumber not the best friend with javascript? They overall seem very flaky. What could be explaining this?

ReverM:
I shall put this as ready for review as I do not understand why the step fails in the test.

app/views/bookmarks/bookmark_form_dynamic.js.erb
ReverM: Done

features/step_definitions/bookmark_steps.rb
```diff
@@ -425,6 +425,91 @@ def submit_bookmark_form(pseud, note, tags, collection)
...
+
+Then "the bookmark form should be open in the blurb for {word}'s bookmark of {string}" do |login, title|
+  user = User.find_by(login: login)
+  work = Work.find_by(title: title)
+  bookmark_id = user.bookmarks.find_by(bookmarkable: work).id
+  within("#bookmark_#{bookmark_id}") do
+    step %{I should see "save a bookmark!"}
+    step %{I should not see "Edit"}
+  end
+end
```
sarken: We're also using these steps to test the Save button. When we're testing Save, we should be checking `I should not see "Save"` (and we don't really need to check for an Edit button, either). What if we refactored them a bit to specify whether we're opening an edit bookmark or new bookmark form, e.g.,

```
"the {word} bookmark form should be open in the bookmarkable blurb for {string}" do |action, title|
```

Then we could adjust the test based on the value of `action`:
```
within("#bookmark_#{work_id}") do
  step %{I should see "save a bookmark!"}
  if action == "edit"
    step %{I should not see "Saved"}
    step %{I should not see "Edit"}
  elsif action == "new"
    step %{I should not see "Save"}
  end
end
```

Of course, that's a lot of code to repeat between two steps. That's not the end of the world with tests, in my opinion, so it's okay to stop here. However, we _could_ go a step further and pull it out into its own step, like:

```
Then "the {word} bookmark form should be open" do |action|
  step %{I should see "save a bookmark!"}
  if action == "edit"
  [...]
end
```

And then we'd only need to reference that in our `within` blocks, e.g.:

```
within("#bookmark_#{work_id}") do
  step %{the #{action} bookmark form should be open}
end
```

features/step_definitions/bookmark_steps.rb
```diff
@@ -425,6 +425,91 @@ def submit_bookmark_form(pseud, note, tags, collection)
...
+  user = User.find_by(login: login)
+  work = Work.find_by(title: title)
+  bookmark_id = user.bookmarks.find_by(bookmarkable: work).id
+  within("#bookmark_#{bookmark_id}") do
+    step %{I should not see "save a bookmark!"}
+    step %{I should see "Edit"}
+  end
+end
+
+Then "{word}'s bookmark form should be closed in the blurb for {word}'s bookmark of {string}" do |bookmarker, login, title|
```
sarken: I think we can remove this step definition and `"{word}'s bookmark form should be closed in the bookmarkable blurb for {string}"` in favor of adjusting the two basic `the bookmark form should be closed` step definitions. We can figure out if the bookmark creator and the currently-logged-in user are the same just by looking at the HTML.

The first thing to do would be exactly what we did for the earlier "the bookmark form should be open" steps: make it so we can specify whether we're looking at a `new` or `edit` form, and then adjust the expectations based on that:

```
  step %{I should not see "save a bookmark!"}
  if action == "edit"
    step %{I should see "Saved"}
    step %{I should see "Edit"}
  elsif action == "new"
    step %{I should see "Save"}
  end
```

The test failures that happen after that should just be for the special case you were addressing in `"{word}'s bookmark form should be closed in the blurb for {word}'s bookmark of {string}"`: the case where the bookmark blurb we're checking belongs to the currently-logged-in user. 

We can handle that by using `has_selector?` to check if our bookmark blurb contains the `own` class:

```
  blurb_id = "#bookmark_#{bookmark_id}"
  # Bookmarks owned by the current user don't have Save or Saved links, only Edit
  if find(blurb_id).has_selector?(".own")
    step %{I should see "Edit" within "#{blurb_id}"}
  else
    within(blurb_id) do
      step %{the #{action} bookmark form should be closed}
    end
  end
```

features/bookmarks/bookmark_javascript.feature
```diff
@@ -0,0 +1,141 @@
...
+    When I follow "Save" in the bookmarkable blurb for "Bookmark: The Sequel"
+    Then the bookmark form should be open in the bookmarkable blurb for "Bookmark: The Sequel"
+    # Close
+    When I exit the bookmark form for the bookmark of "Bookmark: The Beginnings"
+    Then bookmarker's bookmark form should be closed in the bookmarkable blurb for "Bookmark: The Beginnings"
+    When I exit the bookmark form for the bookmark of "Bookmark: The Sequel"
+    Then bookmarker's bookmark form should be closed in the bookmarkable blurb for "Bookmark: The Sequel"
+    # Reopen
+    When I follow "Save" in the bookmarkable blurb for "Bookmark: The Beginnings"
+    Then the bookmark form should be open in the bookmarkable blurb for "Bookmark: The Beginnings"
```
sarken: Can you add a newline at the end of the file here? (It might be possible to configure your editor to do this automatically, depending on which you use.)

features/bookmarks/bookmark_javascript.feature
```diff
@@ -0,0 +1,141 @@
+@bookmarks
+Feature: Edit bookmarks with javascript enabled
+  In order to have a good user experience
+  As a humble user
+  I want to be able to edit bookmarks with javascript enabled
```
sarken: It would probably be a good idea to change this section to talk about both creating and editing now that we have some tests for both.

sarken (changes requested):
Our JavaScript tests do tend to be flaky, unfortunately. If these continue to be a problem, we can update our [list of known test failures](https://github.com/otwcode/otwarchive/wiki/Automated-Testing#user-content-Known_Test_Failures) to mention `features/bookmarks` when they're merged.

features/bookmarks/bookmark_javascript.feature
ReverM: I'm unsure where exactly to put this, but for some reason, clicking on the bookmark link with javascript enabled doesnt work on github. It's one that's flaky on my end, but it's consistently not working on github. Is there an alternate way to setup pre-existing bookmarks with cucumber without passing by steps?

features/bookmarks/bookmark_javascript.feature
ReverM: I use VSCode, do you know how to set that up?

features/step_definitions/bookmark_steps.rb
ReverM: Hmm, given what is the scope of the test, I feel like the first one is more telling.

features/step_definitions/bookmark_steps.rb
```diff
@@ -355,6 +355,20 @@
   step %{all indexing jobs have been run}
 end
 
+# The regular method of creating a bookmark is flaky with javascript. This is an alternate method of creating a bookmark
+Given "{string} has a bookmark of the work {string}" do |user, title|
```
sarken: We can tweak the previous step (`Given "{string} has a bookmark of a work titled {string}"`) instead of adding a new one that's so similar. 

`work1 = Work.find_by(title: title) || FactoryBot...` shouldn't cause any trouble with existing tests.

features/bookmarks/bookmark_javascript.feature
sarken: I had the exact opposite problem: I was having consistent local failures, but it was working more often than not on GitHub Actions. The second scenario still doesn't pass locally for me even with the changes here, but we can live with that.

features/step_definitions/bookmark_steps.rb
```diff
@@ -425,6 +439,92 @@ def submit_bookmark_form(pseud, note, tags, collection)
...
+  work = Work.find_by(title: title)
+  bookmark_id = user.bookmarks.find_by(bookmarkable: work).id
+  find("#bookmark_#{bookmark_id}").click_link(link)
+end
+
+Then "the {word} bookmark form should be open in the bookmarkable blurb for {string}" do |action, title|
+  work_id = Work.find_by(title: title).id
+  within("#bookmark_#{work_id}") do
+    step %{I should see "save a bookmark!"}
+    if action == "edit"
```
sarken: I guess we should do what Reviewdog suggests and use `case-when`. It makes the code one line longer than the `if-elseif`, but temporarily disabling this cop would require adding two lines 😩

app/views/bookmarks/_bookmarkable_blurb.html.erb
```diff
@@ -16,15 +16,16 @@
   <ul class="actions" role="navigation">
     <li>
       <% if current_user_bookmark ||= bookmark_if_exists(unwrapped) %>
-        <%= link_to ts("Saved"),
-                    edit_bookmark_path(current_user_bookmark),
-                    id: "bookmark_form_trigger_for_#{bookmarkable.id}",
-                    remote: true %>
+        <%= link_to t(".saved"),
+        edit_bookmark_path(current_user_bookmark),
```
sarken: Can you address the Reviewdog comments here? I can never remember if it wants one level of indentation from the `<%=` or from `link_to`, but one of those options should make it stop complaining.

sarken (changes requested):
Just a few little things left, most of which are from Reviewdog! Thank you so much for working on this; I'm sorry it turned out to be so much more complicated than it looked.

features/bookmarks/bookmark_javascript.feature
ReverM: What step exactly fails?

features/step_definitions/bookmark_steps.rb
ReverM: Got it, I shall combine the two then.

features/bookmarks/bookmark_javascript.feature
sarken: `Then I should see "Bookmark" within "#main .navigation"` in `features/bookmarks/bookmark_javascript.feature:13`

ReverM: I was looking at the documentation of Capybara, and it seemed to say to use `js: true` to switch to the `Capybara.javascript_driver`. This is not something we seem to be doing, so could this explain the flakiness of the javascript tests? https://github.com/teamcapybara/capybara#using-capybara-with-rspec

We currently have the hook `before "@javascript" do` which changes the host and set the javascript variable to true (so that the website renders with javascript), but no driver change.

sarken: Unfortunately, no. That section is about RSpec, but these are Cucumber tests, which are discussed in [the previous section](https://github.com/teamcapybara/capybara?tab=readme-ov-file#using-capybara-with-cucumber). Using the `@javascript` tag in our tests automatically switches to the JavaScript driver, which is defined in our config: https://github.com/otwcode/otwarchive/blob/be7aee34664df92649948468cf8dad73917dfe58/features/support/capybara.rb#L82

ReverM: Ah I see

app/views/bookmarks/_bookmarkable_blurb.html.erb
ReverM: The dog is finally satisfied

features/bookmarks/bookmark_javascript.feature
sarken: Could you address the comment at the very top of this thread (the one about updating the description)? Then this will be ready to merge! 🎉

sarken (approved):
🎉

PR #5524, opened 2025-12-27
AO3-7219 Keep users with 'show the whole work by default' enabled in chapter by chapter mode when commenting

Dame-Zoom-A-Lot:
# Pull Request Checklist

Note on tests:

I decided not to add tests since comments_controller_spec did not have any tests for the create endpoint, nor tests that would verify redirect_to_all_comments behaviors. works_controller_spec likewise only had tests for #navigate.

I'm assuming this was a conscious decision to avoid bogging down CI since there's a robust QA team. 

I'm happy to put in some extra work to add net new tests for these endpoints if the testing gap was due to lack of resources, but that would increase the scope of this PR.

## Issue

https://otwarchive.atlassian.net/browse/AO3-7219

## Purpose

Fix the bug where users who had "show the whole work by default" enabled were getting redirected to the full work view after commenting.

This was happening because we were showing the whole work

`if params[:view_full_work] || (logged_in? && current_user.preference.try(:view_full_works))`

Since `params[:view_full_work]` is either `"true"` or `"false"` this always evaluates to `true` if it is set.
But even if `params[:view_full_work] == "false"`, if the user had their preference set to `view_full_works`, we would have shown the full work view and not the chapter by chapter view.

There was also a routing error in the comments_controller where
`if commentable.is_a?(Chapter) && (options[:view_full_work] || current_user.try(:preference).try(:view_full_works))` we set the chapter commentable to a work commentable, which also routed to the entire work view.

I added the same logic here that I added in the works controller

## Testing Instructions
Reported bug
1. Log in
2. Have the "Show the whole work by default" preference enabled
3. Open a work in chapter by chapter mode
4. Leave a comment or reply to a comment on a chapter
5. Confirm users stays in chapter by chapter mode

Ensure other behavior did not change
1. Log in
2. Have the "Show the whole work by default" preference enabled
3. Open a work in entire work mode
4. Leave a comment or reply to a comment on a chapter
5. Confirm user stays in entire work mode, and that the new comment is tied to the last chapter

**Testing Notes**
While testing, I found that when the new comment would be in page 2+, the user does not get redirected to the comment section. This is an unrelated issue, and potentially a bug. To test

1. Log in
2. Open work in whatever mode
3. Leave a comment that would show in page 2 of the comments view
4. User is redirected to the work view upon submitting a comment, but the stays at the top of the page and does not get navigated to the comments box

You can remove this section if there are already full testing instructions in the Jira issue.

## Credit

Zooms, any pronoun

they/them pronouns.*

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -384,7 +384,8 @@ class CommentsController < ApplicationController
             elsif @comment.unreviewed?
               redirect_to_all_comments(@commentable)
             else
-              redirect_to_comment(@comment, { view_full_work: (params[:view_full_work] == "true"), page: params[:page] })
+              user_chose_to_view_full_work = (current_user.try(:preference).try(:view_full_works) && params[:view_full_work] != "false")
+              redirect_to_comment(@comment, { view_full_work: (params[:view_full_work] == "true" || user_chose_to_view_full_work), page: params[:page] })
             end
           end
         end
@@ -689,7 +690,7 @@ class CommentsController < ApplicationController
                   page: options[:page],
                   anchor: options[:anchor])
     else
-      if commentable.is_a?(Chapter) && (options[:view_full_work] || current_user.try(:preference).try(:view_full_works))
+      if commentable.is_a?(Chapter) && (options[:view_full_work] || (current_user.try(:preference).try(:view_full_works) && options[:view_full_work] != false))
         commentable = commentable.work
       end
       redirect_to polymorphic_path(commentable,
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -55,8 +55,10 @@ class CommentsController < ApplicationController
 
   def check_pseud_ownership
     return unless params[:comment][:pseud_id]
+
     pseud = Pseud.find(params[:comment][:pseud_id])
     return if pseud && current_user && current_user.pseuds.include?(pseud)
+
     flash[:error] = ts("You can't comment with that pseud.")
     redirect_to root_path
   end
@@ -180,14 +182,17 @@ class CommentsController < ApplicationController
   def check_permission_to_review
     parent = find_parent
     return if logged_in_as_admin? || current_user_owns?(parent)
+
     flash[:error] = ts("Sorry, you don't have permission to see those unreviewed comments.")
     redirect_to logged_in? ? root_path : new_user_session_path(return_to: request.fullpath)
   end
 
   def check_permission_to_access_single_unreviewed
     return unless @comment.unreviewed?
+
     parent = find_parent
     return if logged_in_as_admin? || current_user_owns?(parent) || current_user_owns?(@comment)
+
     flash[:error] = ts("Sorry, that comment is currently in moderation.")
     redirect_to logged_in? ? root_path : new_user_session_path(return_to: request.fullpath)
   end
@@ -384,7 +389,9 @@ class CommentsController < ApplicationController
             elsif @comment.unreviewed?
               redirect_to_all_comments(@commentable)
             else
-              user_chose_to_view_full_work = (current_user.try(:preference).try(:view_full_works) && params[:view_full_work] != "false")
+              # keep the user in chapter by chapter view if that was the view they were in when they commented
+              # An entire work view url would look like works/:id
+              user_chose_to_view_full_work = (current_user.try(:preference).try(:view_full_works) && request.referer&.match(/chapters/).nil?)
               redirect_to_comment(@comment, { view_full_work: (params[:view_full_work] == "true" || user_chose_to_view_full_work), page: params[:page] })
             end
           end
@@ -690,8 +697,9 @@ class CommentsController < ApplicationController
                   page: options[:page],
                   anchor: options[:anchor])
     else
-      if commentable.is_a?(Chapter) && (options[:view_full_work] || (current_user.try(:preference).try(:view_full_works) && options[:view_full_work] != false))
-        commentable = commentable.work
+      if commentable.is_a?(Chapter)
+        user_chose_to_view_full_work = current_user.try(:preference).try(:view_full_works) && options[:view_full_work] != false 
+        commentable = commentable.work if options[:view_full_work] || user_chose_to_view_full_work
       end
       redirect_to polymorphic_path(commentable,
                                    options.slice(:show_comments,
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -191,7 +191,7 @@ class WorksController < ApplicationController
 
     # Users must explicitly okay viewing of entire work
     if @work.chaptered?
-      if params[:view_full_work] || (logged_in? && current_user.preference.try(:view_full_works))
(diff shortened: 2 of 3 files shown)


app/controllers/comments_controller.rb
```diff
@@ -55,8 +55,10 @@ def check_blocked
 
   def check_pseud_ownership
     return unless params[:comment][:pseud_id]
+
```
Dame-Zoom-A-Lot, line 58: added some whitespace only rubocop auto corrects

app/controllers/comments_controller.rb
```diff
@@ -689,8 +697,9 @@ def redirect_to_all_comments(commentable, options = {})
                   page: options[:page],
                   anchor: options[:anchor])
     else
-      if commentable.is_a?(Chapter) && (options[:view_full_work] || current_user.try(:preference).try(:view_full_works))
-        commentable = commentable.work
+      if commentable.is_a?(Chapter)
+        user_chose_to_view_full_work = current_user.try(:preference).try(:view_full_works) && options[:view_full_work] != false
```
Dame-Zoom-A-Lot, line 701: it would be nice to have a unified way to see when a user wanted to to view full work vs chaptered view. But all the different helper methods and endpoints had slightly different ways we would need to use to identify them.

app/controllers/works_controller.rb
```diff
@@ -191,7 +191,7 @@ def show
 
     # Users must explicitly okay viewing of entire work
     if @work.chaptered?
-      if params[:view_full_work] || (logged_in? && current_user.preference.try(:view_full_works))
+      if params[:view_full_work] == "true" || (logged_in? && current_user.preference.try(:view_full_works) && params[:view_full_works] != "false")
```
Bilka2, line 194: First thing I noticed here is that it should be `params[:view_full_work]`, singular. So right now that second check isn't doing anything

However, from what I can tell, that second check doesn't need to do anything to fix the issue here: If we're in chapter by chapter mode and want to stay there, then the comments controller redirects to the chapter directly. So we should never end up in this controller with `view_full_work` set to anything other than "true" or nothing. Because the general idea is that `view_full_work` is only set to false when the user specifically declined being in that mode, which only happens inside comments_controller which already handles all that within `redirect_to_all_comments`

So I would remove the ` && params[:view_full_works] != "false"` here

app/controllers/comments_controller.rb
```diff
@@ -689,8 +697,9 @@ def redirect_to_all_comments(commentable, options = {})
                   page: options[:page],
                   anchor: options[:anchor])
     else
-      if commentable.is_a?(Chapter) && (options[:view_full_work] || current_user.try(:preference).try(:view_full_works))
-        commentable = commentable.work
+      if commentable.is_a?(Chapter)
+        user_chose_to_view_full_work = current_user.try(:preference).try(:view_full_works) && options[:view_full_work] != false 
+        commentable = commentable.work if options[:view_full_work] || user_chose_to_view_full_work
```
Bilka2, line 702: Currently, this can leak a "view_full_work" param that is set to `false` into the URL it redirects to. That doesn't do anything, and it's ambiguous between the user having the preference disabled, coming from a chapter URL, or just `view_full_work` not being present. Since this controller handles the redirect itself, only dumping the user to the full work if they actually wanted it, I think we should set `options[:view_full_work]` to nil here if we're not redirecting to the work

Bilka2: That suggestion summarizes to the following decision tree to me:

```
if options[:view_full_work] it's "true" or true
  go to the work in full work mode (so including the param)
  The preference is irrelevant because either they manually chose full work mode or it was set to the bool true in your change above that checks the preference

if options[:view_full_work] is false
  go to the chapter, always
  because we already checked the preference in the only spot that can set the option to false (your change above)
  Don't include the param in the URL because we're going to a chapter, not a work

options[:view_full_work] can never be "false"

so if options[:view_full_work] is blank
  the user hasn't chosen anything
  Do the preference check and decide chapter/work based on that. Don't set view_full_work
```

Bilka2: The current decision tree here is

```
if options[:view_full_work] is "true" or true
  go to the work in full work mode

if options[:view_full_work] is "true" or true or blank
  if the preference is true
    go to the work
    view full work mode is set to options[:view_full_work] (so may be blank)
  
otherwise
  go to the chapter
  view full work mode may be blank or false
```

this is mostly the same as my suggested decision tree, the only difference is the view_full_work param possibly being set when redirecting to a chapter. But the current code took me ages to parse. Not sure if that is a me issue or not, so I'm leaving this analysis here and you can decide how you integrate my suggestion and whether you change how the checks here are done or not

Bilka2 (changes requested):
The comment creation tests are in `spec/controllers/comments/default_rails_actions_spec.rb`, works show specs are in `spec/controllers/works/default_rails_actions_spec.rb`. I think we should definitely cover the changes here with automated tests because they're quite intricate between the many different cases (comment from work with or without the param, from chapter, and those with and without the preference). Though hopefully a lot of that is already covered by tests, whether it's rspec or cucumber

app/controllers/works_controller.rb
Dame-Zoom-A-Lot: thanks for the careful review. It's been a while, but I think not having that second check was causing some unexpected behaviors. I'm trying to get set back up so I can test.

app/controllers/comments_controller.rb
Dame-Zoom-A-Lot: Agreed on not setting view_full_work unless it's true. I can edit. 

I also had trouble following what we were doing with view_full_work, and honestly, I'm a bit concerned that this could have been the intended behavior because the comments_redirect test was specifically checking that users reverted back to full works view.

Bilka2: > honestly, I'm a bit concerned that this could have been the intended behavior because the comments_redirect test was specifically checking that users reverted back to full works view.

That's a fair concern! I went digging into the history of that test, and it looks like [AO3-4487](https://otwarchive.atlassian.net/browse/AO3-4487) added those comments and solidified this behaviour and doesn't have any context on whether there was a discussion on whether it was wanted, it's a simple "fix test to match existing behaviour" issue. The initial PR that added this behaviour is #327 and the relevant Jira issues are [AO3-1514](https://otwarchive.atlassian.net/browse/AO3-1514),  [AO3-1515](https://otwarchive.atlassian.net/browse/AO3-1515).

Neither of those explicitly defines the behaviour here, so I think we aren't undoing a specific past decision. Besides, when this Jira issue was discussed before getting created, it was a very short "Yup, definitely a bug." discussion so even if we considered this intended approximately 10 years ago, we no longer do now.

Dame-Zoom-A-Lot:
@Bilka2 thanks for the careful review.

Agreed on both your points
1. Current logic is too convoluted
2. There aren't enough tests for what's honestly a pretty complicated flow

This PR's been open a little too long for me to be able to easily work off of it. [I started a fresh PR here](https://github.com/otwcode/otwarchive/pull/5798)

Major change is that I'm ditching the view_full_works parameter and working off of the referer. Please see the Jira tickets for all routes I tested.

PR #5541, opened 2026-01-12
AO3-7140 Changed view of Assignment to be less confusing

ReverM:
## Issue

https://otwarchive.atlassian.net/browse/AO3-7140

## Purpose

This changes how the assignment blurb look when you have a drafted work that is unposted (changing Posted to Created and adding (draft) to the title)

## Credit

Danaël / Rever ( they / he )

Danaël Villeneuve on Jira


config/locales/views/en.yml
```diff
@@ -569,6 +569,17 @@ en:
       challenge_requests:
         fandom: Fandom %{allowed_fandom_count}
   challenge_assignments:
+    assignment_blurb:
+      draft_header: " (draft)"
```
ReverM: What should be the best way to do this? should I put the space in the view instead?

Bilka2: Include the link in locale as a variable

ReverM: I tried doing so but I couldn't make it work. Something about having the link at the start made it not work? I

app/views/challenge_assignments/_assignment_blurb.html.erb
```diff
@@ -11,9 +11,13 @@
 </dt>
 <dd>
   <% # link to the work %>
-   <% if assignment.creation %>
-     <%= link_to (assignment.creation.try(:title) || assignment.creation.class.name), assignment.creation, :class => "work" %>
-   <% end %>
+    <% if assignment.creation %>
+      <% if !assignment.posted? %>
+        <%= t(".header_draft", link: (link_to (assignment.creation.try(:title) || assignment.creation.class.name), assignment.creation, class: "work")) %>
```
marcus8448: tiniest of nits: could you get rid of the parenthesis wrapping the `link_to`?

app/views/challenge_assignments/_assignment_blurb.html.erb
```diff
@@ -26,21 +30,26 @@
...
-    <dt><%= ts("Status:") %></dt>
+    <dt><%= t(".stats.status.header") %></dt>
     <dd>
       <% # status can be: unposted (no creation), unapproved (creation but it hasn't been approved), complete %>
       <% if !assignment.posted? %>
-        <%= ts("Unposted") %>
+        <%= t(".stats.status.unposted") %>
       <% elsif !assignment.fulfilled? %>
-        <strong><%= ts("Unapproved") %><!-- BACK END add approve button for mod --></strong>
+        <strong><%= t(".stats.status.unfullfilled") %><!-- BACK END add approve button for mod --></strong>
```
marcus8448:
```suggestion
        <strong><%= t(".stats.status.unfulfilled") %><!-- BACK END add approve button for mod --></strong>
```
Could also be `unapproved` to match the displayed text.

(same typo for fulfilled)

features/gift_exchanges/challenge_giftexchange.feature
```diff
@@ -639,6 +639,28 @@ Feature: Gift Exchange Challenge
     Then I should see "My Assignments"
       And I should not see "Awesome Gift Exchange"
 
+  Scenario: Information regarding the status of an assignment is not confusing (AO3-7140)
```
marcus8448: It's hard to tell what's being tested from the scenario name.
Maybe something like displaying the draft status [and using the creation date] for unposted works? (then the issue number can be omitted as well)

config/locales/views/en.yml
```diff
@@ -569,6 +569,18 @@ en:
       challenge_requests:
         fandom: Fandom %{allowed_fandom_count}
   challenge_assignments:
+    assignment_blurb:
+      header_draft: "%{link} (draft)"
```
marcus8448: Could you change the variable name to `title_link` or similar?

app/views/challenge_assignments/_assignment_blurb.html.erb
```diff
@@ -11,9 +11,13 @@
...
 <dd>
   <% # link to the work %>
-   <% if assignment.creation %>
-     <%= link_to (assignment.creation.try(:title) || assignment.creation.class.name), assignment.creation, :class => "work" %>
-   <% end %>
+    <% if assignment.creation %>
+      <% if !assignment.posted? %>
+        <%= t(".header_draft", link: (link_to (assignment.creation.try(:title) || assignment.creation.class.name), assignment.creation, class: "work")) %>
+      <% else %>
+        <%= t(".header_posted", link: (link_to (assignment.creation.try(:title) || assignment.creation.class.name), assignment.creation, class: "work")) %>
```
marcus8448: The title link can just be inserted directly here (no need for a translation).

marcus8448 (changes requested):
Just a couple wording + translation bits.
Could you add a test for the display of the `created_at` date for drafts?

features/gift_exchanges/challenge_giftexchange.feature
ReverM: "Draft Assignments are displayed as draft and use the creation date instead of the published date used for published Assignments"
Is this better?

marcus8448: How about "Draft Assignments are displayed as draft and use the creation date instead of the publication date"?

features/gift_exchanges/challenge_giftexchange.feature
```diff
@@ -639,6 +639,28 @@ Feature: Gift Exchange Challenge
     Then I should see "My Assignments"
       And I should not see "Awesome Gift Exchange"
 
+  Scenario: Draft Assignments are displayed as draft and use the creation date instead of the publication date
```
marcus8448, line 642: Could you update the test to backdate the draft and check that the correct dates are used (as a draft and when posted)?

config/locales/views/en.yml
```diff
@@ -569,6 +569,17 @@ en:
       challenge_requests:
         fandom: Fandom %{allowed_fandom_count}
   challenge_assignments:
+    assignment_blurb:
+      header_draft: "%{title_link} (draft)"
```
marcus8448: This key should have a `_html` suffix so that the link doesn't get escaped.

marcus8448 (changes requested):
Just two more things!

features/gift_exchanges/challenge_giftexchange.feature
ReverM: Should be done

features/step_definitions/work_steps.rb
```diff
@@ -551,6 +551,7 @@
   end
 end
 
+# Format is "DD [Month, english] YYYY"
```
ReverM, line 554: I've added a comment for future use

features/gift_exchanges/challenge_giftexchange.feature
```diff
@@ -639,6 +639,35 @@ Feature: Gift Exchange Challenge
...
+    When I follow "Fulfill"
+      And I fill in the basic work information for "InProgress"
+      And I set the publication date to 6 December 2012
+      And I press "Preview"
+      And I go to the assignments page for "myname1"
+    Then I should see "(draft)"
+      And I should see "Status: Unposted"
+      And I should see "Created"
+      And I should not see "Created: 2012-12-06"
+      And I should see "Created: 2016-04-13"
```
marcus8448: E.g. the "Created" check already should be covered by "Created: 2016-04-13".
I also think seeing the correct date is sufficient to prove that the incorrect date isn't shown.

features/gift_exchanges/challenge_giftexchange.feature
```diff
@@ -639,6 +639,35 @@ Feature: Gift Exchange Challenge
...
+      And I go to the assignments page for "myname1"
+      And I follow "Completed Assignments"
+    Then I should see "InProgress"
+      And I should not see "Status: Unposted"
+      And I should not see "Created"
+      And I should not see "(draft)"
+      And I should see "Status: Complete"
+      And I should see "Posted"
+      And I should not see "Posted: 2016-04-13"
+      And I should see "Posted: 2012-12-06"
```
marcus8448: Same idea here with Posted: and Status:. I would keep the check to not see (draft), though.

marcus8448 (commented):
I think it would be good to reduce the number of checks in the test. Steps in cucumber tests can be quite slow and I feel some of them might be redundant.

marcus8448 (approved):
Thank you!

PR #5550, opened 2026-01-19
AO3-5367 Bookmarks of unrevealed works appear in search results

ReverM:
## Issue

https://otwarchive.atlassian.net/browse/AO3-5367

## Purpose

This makes it so it is impossible to search for hidden (via unrevealed) information with bookmarks by making the json returned for the bookmarked item be empty if the item is unrevealed.

## Credit

Danaël / Rever ( they / he ) 
Danaël Villeneuve on jira


ReverM:
To note, this will require reindexing

Bilka2:
Could you explain why it doesn't work to use a filter in the bookmarkable query similar to hidden_filter? https://github.com/otwcode/otwarchive/blob/b23fe624793956ce5fbf2612dfe2e3456704797b/app/models/search/bookmarkable_query.rb#L282-L287

ReverM:
> Could you explain why it doesn't work to use a filter in the bookmarkable query similar to hidden_filter?
> 
> https://github.com/otwcode/otwarchive/blob/b23fe624793956ce5fbf2612dfe2e3456704797b/app/models/search/bookmarkable_query.rb#L282-L287

This will work for making sure we cannot search using the information on the work that is not visible. However, this would also make it so that it is impossible to search for the visible information on the bookmark. The second step of the added test would fail as this would just exclude every bookmark that is of an unrevealed bookmarkable.

app/models/work.rb
```diff
@@ -1203,6 +1203,9 @@ def document_json
   end
 
   def bookmarkable_json
+    # If the work is unrevealed, it should not have any information that can be searched on
+    return {} unless !unrevealed?
```
Bilka2: Let's still include `root`, `bookmarkable_type` and `bookmarkable_join` for unrevealed works

The dog will shut up if you use `if unrevealed?`

Can you remove the `unrevealed?` checks on user_ids/pseud_ids below?

Bilka2 (changes requested):
Thank you for the explanation! I reconfirmed that we indeed want to keep the bookmarks themselves searchable, so this current approach is good

For the reindexing, this doesn't change the mapping so we we don't need to rebuild the whole index. So, ideally we would update only the bookmarkables that are unrevealed works, without needing downtime. Could you add a rake task that does that? `reindex_hidden_unrevealed_tags` is an example of a similar task

Bilka2 (changes requested):
For the reindexing, this doesn't change the mapping so we don't need to rebuild the whole index. So, ideally we would update only the bookmarkables that are unrevealed works, without needing downtime. Could you add a rake task that does that? reindex_hidden_unrevealed_tags is an example of a similar task

ReverM:
> For the reindexing, this doesn't change the mapping so we don't need to rebuild the whole index. So, ideally we would update only the bookmarkables that are unrevealed works, without needing downtime. Could you add a rake task that does that? reindex_hidden_unrevealed_tags is an example of a similar task

Oh sorry about that. It totally slipped out of my mind. Going to do this today.

I see that `reindex_hidden_unrevealed_tags` is mentionned in the `.rake_taks~` file but it seems part of the git ignore. Do I need to modify this file?

Bilka2:
> I see that `reindex_hidden_unrevealed_tags` is mentionned in the `.rake_taks~` file but it seems part of the git ignore. Do I need to modify this file?

That seems like a temporary file, so you shouldn't modify it, no. The real task you can use for reference is in [after_tasks.rake](https://github.com/otwcode/otwarchive/blob/76dd189aaed57d924aa50aea311d4f45e3e9d740/lib/tasks/after_tasks.rake#L495), which is also where the task for this PR should be added

ReverM:
> > I see that `reindex_hidden_unrevealed_tags` is mentionned in the `.rake_taks~` file but it seems part of the git ignore. Do I need to modify this file?
> 
> That seems like a temporary file, so you shouldn't modify it, no. The real task you can use for reference is in [after_tasks.rake](https://github.com/otwcode/otwarchive/blob/76dd189aaed57d924aa50aea311d4f45e3e9d740/lib/tasks/after_tasks.rake#L495), which is also where the task for this PR should be added

Got it! One last question, if a work is reindexed as a bookmarkable, does it need to be reindex as a work as well?

Bilka2:
> Got it! One last question, if a work is reindexed as a bookmarkable, does it need to be reindex as a work as well?

No, only the bookmarks needs to be reindexed by this task, since we're not changing the work index. Basically, I think we can get all unrevealed works, similar to what `reindex_hidden_unrevealed_tags` did and then call `update_bookmarks_index` on them

lib/tasks/after_tasks.rake
```diff
@@ -512,6 +512,19 @@ namespace :After do
     puts "Finished reindexing tags on hidden and unrevealed works"
   end
 
+  desc "Reindex unrevealed bookmarks of unrevealed works"
+  task(reindex_unrevealed_bookmarkable: :environment) do
+    unrevealed_count = Work.unrevealed.joins(:bookmarks).distinct.count
+    unrevealed_batches = (unrevealed_count + 999) / 1_000
+    puts "Inspecting #{unrevealed_count} bookmarks of unrevealed works in #{unrevealed_batches} batches"
+    Work.unrevealed.joins(:bookmarks).distinct.find_in_batches.with_index do |batch, index|
```
Bilka2: Good idea to only go for unrevealed works with bookmarks! However, in [the past](https://otwarchive.atlassian.net/browse/AO3-6315) we found that simply queueing all relevant works is much faster than queuing only works with bookmarks. Queuing happens synchronously, but indexing happens asynchronously (and won't do anything for works without bookmarks), so we're happy to queue more for indexing instead of the queuing itself being a bit slower.

Comparing the count queries, this extra JOIN and DISTINCT appears to only reduce the affected works by about half:
```
Work.unrevealed.joins(:bookmarks).distinct.count
 => 98142 
Work.unrevealed.count
 => 198790 
```
And just the count queries are already pretty slow (6.8s for the join one, 4.7s for plain unrevealeds) and the EXPLAIN for the JOIN + DISTINCT doesn't look good, so we'd rather use the cheaper one here (`Work.unrevealed`).


The EXPLAINs, for context
```
Work.unrevealed.joins(:bookmarks).distinct.explain
 => 
EXPLAIN SELECT DISTINCT `works`.* FROM `works` INNER JOIN `bookmarks` ON `bookmarks`.`bookmarkable_type` = 'Work' AND `bookmarks`.`bookmarkable_id` = `works`.`id` WHERE `works`.`in_unrevealed_collection` = TRUE
+----+-------------+-----------+------+--------------------------+--------------------------+---------+--------------------------------------+----------+------------------------------------+
| id | select_type | table     | type | possible_keys            | key                      | key_len | ref                                  | rows     | Extra                              |
+----+-------------+-----------+------+--------------------------+--------------------------+---------+--------------------------------------+----------+------------------------------------+
|  1 | SIMPLE      | works     | ALL  | PRIMARY                  | NULL                     | NULL    | NULL                                 | 12183623 | Using where; Using temporary       |
|  1 | SIMPLE      | bookmarks | ref  | index_bookmarkable_pseud | index_bookmarkable_pseud | 66      | otwarchive_production.works.id,const | 13       | Using where; Using index; Distinct |
+----+-------------+-----------+------+--------------------------+--------------------------+---------+--------------------------------------+----------+------------------------------------+
2 rows in set (0.00 sec)
```
```
Work.unrevealed.explain
 => 
EXPLAIN SELECT `works`.* FROM `works` WHERE `works`.`in_unrevealed_collection` = TRUE
+----+-------------+-------+------+---------------+------+---------+------+----------+-------------+
| id | select_type | table | type | possible_keys | key  | key_len | ref  | rows     | Extra       |
+----+-------------+-------+------+---------------+------+---------+------+----------+-------------+
|  1 | SIMPLE      | works | ALL  | NULL          | NULL | NULL    | NULL | 12183623 | Using where |
+----+-------------+-------+------+---------------+------+---------+------+----------+-------------+
1 row in set (0.00 sec)

```

spec/lib/tasks/after_tasks.rake_spec.rb
```diff
@@ -461,6 +461,35 @@
...
+
+    it "does not reindex the bookmark" do
+      expect do
+        subject.invoke
+      end.not_to add_to_reindex_queue(bookmark, :background)
+    end
+  end
+
+  context "with a bookmark of an unrevealed work" do
+    let(:work) { create(:work) }
```
Bilka2, line 477:
```suggestion
    let(:work) { create(:work, in_unrevealed_collection: true) }
```
then you can remove the `before`

Bilka2 (changes requested):

app/models/work.rb
```diff
@@ -1203,6 +1203,15 @@ def document_json
   end
 
   def bookmarkable_json
+    # If the work is unrevealed, it should not have any information that can be searched on
+    if unrevealed?
+      return as_json(
```
Bilka2, line 1208: Let's also still set unrevealed in here

spec/lib/tasks/after_tasks.rake_spec.rb
ReverM: Doing so makes the test fail locally

lib/tasks/after_tasks.rake
ReverM: I see

spec/lib/tasks/after_tasks.rake_spec.rb
Bilka2: I see!

Bilka2 (approved):
Thank you!

PR #5555, opened 2026-01-22
AO3-5748 Split off assignments with both offer and pinch hitter

neuroalien:
Split off assignments with both offer and pinch hitter

# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-5748

## Purpose

What does this PR do?

Splits off assignments that have both an offer signup and a pinch hitter, into two separate assignments, one owned by the offering pseud, the other owned by the pinch hitter, neither leaking data about the other.

Usually, when a mod assigns a pinch-hitter through other means, a brand new ChallengeAssignment is created with a pinch_hitter_id (rather than adding a pinch_hitter_id to the existing ChallengeAssignment).

Let's do the same for the case of a pinch hitter being assigned at the same time as the offer signup.

We've chosen to do this at the point of saving the current configuration, so we can show the resulting assignments. We hope the maintainer won't double-assign again, thus causing a chain of splits.

The result, for a three-user challenge, will look something like this:

Screenshot 2026-01-22 at 03 03 16

I've chosen to do this in a separate method, invoked after `update_placeholder_assignments!` has done its thing, because otherwise our freshly created assignment would have been deleted as being a "leftover placeholder" by the "# if this signup has at least one giver now, get rid of any leftover placeholders" check. I didn't want to mess with how leftover placeholders are defined, since we don't have any open issues about them. "If it ain't broke" etc.

## Testing Instructions

The JIRA issue has detailed instructions.

## References

Yes, this should also fix https://otwarchive.atlassian.net/browse/AO3-5752.

## Credit

alien, they/them

### DIFF
--- a/app/controllers/challenge_assignments_controller.rb
+++ b/app/controllers/challenge_assignments_controller.rb
@@ -149,6 +149,7 @@ class ChallengeAssignmentsController < ApplicationController
     end
 
     ChallengeAssignment.update_placeholder_assignments!(@collection)
+    ChallengeAssignment.split_off_write_in_giver_assignments!(@collection)
     if @assignments.empty?
       flash[:notice] = "Assignments updated"
       redirect_to collection_potential_matches_path(@collection)
--- a/app/models/challenge_assignment.rb
+++ b/app/models/challenge_assignment.rb
@@ -496,4 +496,24 @@ class ChallengeAssignment < ApplicationRecord
       end
     end
   end
+
+  # create fresh new assignments for pinch hitters assigned alongside existing offer signups
+  def self.split_off_write_in_giver_assignments!(collection)
+    collection.assignments.each do |assignment|
+      assignment.split if assignment.double_assigned?
+    end
+  end
+
+  def double_assigned?
+    pinch_hitter && offer_signup && pinch_hitter.user != offering_user
+  end
+
+  def split
+    new_assignment = self.dup
+    new_assignment.offer_signup = nil
+    new_assignment.save!
+
+    self.pinch_hitter = nil
+    save!
+  end
 end
--- a/spec/controllers/challenge_assignments_controller_spec.rb
+++ b/spec/controllers/challenge_assignments_controller_spec.rb
@@ -146,6 +146,36 @@ describe ChallengeAssignmentsController do
         end
       end
 
+      context "when an assignment has both an offer and a pinch hitter" do
+        let(:params) do
+          {
+            collection_id: collection.name,
+            challenge_assignments: {
+              assignment.id => {
+                request_signup_pseud: signup1.pseud.byline,
+                offer_signup_pseud: signup2.pseud.byline,
+                pinch_hitter_byline: other_user.pseuds.first.byline
+              }
+            }
+          }
+        end
+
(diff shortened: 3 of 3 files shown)


app/controllers/challenge_assignments_controller.rb
```diff
@@ -149,6 +149,7 @@ def set
     end
 
     ChallengeAssignment.update_placeholder_assignments!(@collection)
+    ChallengeAssignment.split_off_write_in_giver_assignments!(@collection)
```
brianjaustin, line 152: I'm wondering if this should go inside `update_placeholder_assignments!`, since at the moment it seems like we'll always want to do them both at the same time?

app/models/challenge_assignment.rb
```diff
@@ -496,4 +496,24 @@ def self.update_placeholder_assignments!(collection)
...
+    collection.assignments.each do |assignment|
+      assignment.split if assignment.double_assigned?
+    end
+  end
+
+  def double_assigned?
+    pinch_hitter && offer_signup && pinch_hitter.user != offering_user
+  end
+
+  def split
```
brianjaustin, line 511: Could you add tests for these methods as well? (I'm also wondering if it's worth making them `private`, which would mean harder to test but a smaller public interface for this class 🤔 )

Update: please try making them private, and if that breaks something, add tests for them as public

spec/controllers/challenge_assignments_controller_spec.rb
```diff
@@ -146,6 +146,36 @@
...
+
+        it "creates a new assignment for the pinch hitter" do
+          PotentialMatch.create(request_signup: signup1,
+                                offer_signup: signup2,
+                                collection: collection)
+
+          put :set, params: params
+
+          assignment.reload
+          expect(assignment.pinch_hitter).to be_nil
```
brianjaustin, line 171: Could you add an assertion to make sure the offer signup is not cleared here? (Probably paranoid of me, but just in case 😅 )

app/models/challenge_assignment.rb
```diff
@@ -496,4 +496,24 @@ def self.update_placeholder_assignments!(collection)
...
+    new_assignment = self.dup
+    new_assignment.offer_signup = nil
+    new_assignment.save!
+
+    self.pinch_hitter = nil
+    save!
```
brianjaustin, line 517: Thinking about these 2 saves, I'm leaning towards wrapping it in an explicit transaction. That way, we don't "lose" signup information if there's a problem with one of these writes. But it means we _could_ still hit the case mentioned in the bug iff there's some database issue between the start and end of the assignments change. What do you think?

app/controllers/challenge_assignments_controller.rb
brianjaustin: After discussion elsewhere, two separate methods makes sense from a conceptual level. All good for now!

app/models/challenge_assignment.rb
brianjaustin: Based on what I can find, I'd say let's do the transaction -- if there's a wider issue with the database, the potential for data issues is elevated anyways

```rb
ChallengeAssignment.transaction do
  # existing method content
end
```

PR #5557, opened 2026-01-22
AO3-7251 Index past emails/usernames from purpose-specific tables

Kiyazz:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-7251

## Purpose

This PR changes the admin past-user and past-email searching feature to use the more efficient tables created in AO3-7249 instead of querying the audits table each time, since that table can get very large and can't easily be indexed

## Testing Instructions

See Jira issue

## References

Reverts AO3-7216, which was a hotfix this PR solves properly

Dependent on AO3-7249, branched off from that PR. Can't be merged before that PR is merged first. Also had a merge conflict with AO3-7060

## Credit

kiyazz (he/him)

### DIFF
--- a/config/config.yml
+++ b/config/config.yml
@@ -230,9 +230,6 @@ TAGGINGS_COUNT_MIN_CACHE_COUNT: 1000
 # For tagging changes, we only reindex tags used less than a certain number of times
 TAGGINGS_COUNT_REINDEX_LIMIT: 1000
 
-# How many rows we should get from audits for historic email and login values for users
-USER_HISTORIC_VALUES_LIMIT: 10_000
-
 # how many signups in a challenge before we move to static summaries generated hourly
 MAX_SIGNUPS_FOR_LIVE_SUMMARY: 20
 
--- a/config/config.yml
+++ b/config/config.yml
@@ -230,6 +230,9 @@ TAGGINGS_COUNT_MIN_CACHE_COUNT: 1000
 # For tagging changes, we only reindex tags used less than a certain number of times
 TAGGINGS_COUNT_REINDEX_LIMIT: 1000
 
+# How many rows we should get from audits for backfilling user history tables
+USER_HISTORIC_VALUES_LIMIT: 10_000
+
 # how many signups in a challenge before we move to static summaries generated hourly
 MAX_SIGNUPS_FOR_LIVE_SUMMARY: 20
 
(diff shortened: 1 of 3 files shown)


Kiyazz:
Note that this PR doesn't have migrations

brianjaustin:
> Note that this PR doesn't have migrations

@Kiyazz could you try using `git rebase -i master` and follow the instructions to "drop" the commits that are already on `master`? That will remove the things, like migrations, that aren't really added in this PR (and make it easier for us to review). Since rebasing can be a bit funky, please let us know if you'd like any help there. Thanks!

Kiyazz:
the rebase is done, but still probably best to wait until the dependent PR AO3-7249 is merged before reviewing this, since tests rely on the changes from there

Kiyazz:
wanted this PR to be around since those 2 should probably be deployed together. Each is useless without the other!

Bilka2:
>  since those 2 should probably be deployed together

We separated them because they need to be deployed sequentially in separate deploys so that we have enough time to fill the database tables before trying to read from them. To deploy those changes together at once we'd need downtime while backfilling the data, which we prefer avoid

Kiyazz:
Ah, makes sense

app/models/search/user_indexer.rb
```diff
@@ -66,8 +66,8 @@ def document(object)
 
   def extra_info(object)
     names = ([object.login] + object.pseuds.map(&:name)).uniq
-    past_names = object.historic_values("login")
-    past_emails = object.historic_values("email")
+    past_names = object.past_usernames.pluck(:username).compact
```
Bilka2: Is the `compact` here still needed? Inversely, don't we need `distinct` here so values don't show up multiple times if someone repeatedly changes their username back and forth?

Same applies for emails and the views. I think it may be good to then have this in the helper methods, so we don't need to remember to do this both here and in the view

config/config.yml
```diff
@@ -230,9 +230,6 @@ TAGGINGS_COUNT_MIN_CACHE_COUNT: 1000
 # For tagging changes, we only reindex tags used less than a certain number of times
 TAGGINGS_COUNT_REINDEX_LIMIT: 1000
 
-# How many rows we should get from audits for historic email and login values for users
-USER_HISTORIC_VALUES_LIMIT: 10_000
```
Bilka2, line 236: This is still required for the AuditsBackfillJob. Could you keep it and instead adjust the comment?

Bilka2 (changes requested):

Bilka2 (approved):
Thank you!

PR #5580, opened 2026-02-13
AO3-7294 Maintenance task to reset bookmark-only tags

Snoopynha:
# Pull Request Checklist

## Issue
 
https://otwarchive.atlassian.net/browse/AO3-7294

## Purpose

This PR addresses the accumulation of "bookmark-only" tags in wrangler bins. Sometimes, tags that are only used on bookmarks (and have zero works) get assigned a specific type (like Character or Relationship) or get attached to a Fandom. This clutters the wrangling interface with tags that shouldn't be wrangled yet.

**Changes implemented:**

### 1. New Rake Task (`tags:reset_bookmark_only`):

- Identifies tags that are **non-canonical, non-synonymous, and have zero works** (`taggings_count_cache: 0`).
- Checks if these tags have been assigned a `type` other than `'Tag'` (Unsorted) OR if they have parent associations (`common_taggings`).
- Resets the `type` to `'Tag'` (Unsorted) and removes all parent associations.

### 2. Scheduled Maintenance:

- Added the task to `config/schedule.rb` to run quarterly, ensuring the bins stay clean automatically over time.

### Technical Notes:

- **Performance:** Uses `find_each` (batch size 1000) to avoid memory spikes when processing large numbers of tags.
- **Database Efficiency:** Uses `delete_all` for associations and `update_columns` for the tag itself. This intentionally bypasses ActiveRecord callbacks to prevent triggering expensive Elasticsearch/Redis re-indexing for every single tag reset during bulk maintenance.
- **Testing:** Added a comprehensive spec file (`spec/lib/tasks/reset_bookmark_tags_spec.rb`) covering scenarios for incorrect tags, legitimate work tags, and canonical tags.

### Prerequisites:
You will need access to the Rails console (bundle exec rails c) or the ability to run Rake tasks.

#### 1. Create a "Dirty" Tag (The Target):

- Create a generic tag (e.g., "Test Dirty Bookmark Tag").
- Ensure it has **0 works**.
- Edit the tag and change its type to **Character**.
- Wrangle it to a Fandom (e.g., "Test Fandom").
- _Result_: This tag is now "dirty" (Bookmark-only but categorized).

#### 2. Create a "Control" Tag (The Innocent):

- Create another tag (e.g., "Test Work Tag").
- Add it to at least **1 work**.
- Edit the tag, make it a **Character**, and wrangle it to a Fandom.
- _Result_: This is a legitimate tag that should NOT be touched.

#### 3. Run the Task:
- Run: `bundle exec rake tags:reset_bookmark_only`

#### 4. Verify Results:

- Check the "Dirty" tag: It should now be listed as **Unsorted** (Type: Tag) and should **not** have any Fandom listed.
- Check the "Control" tag: It should still be a **Character** and still belong to its Fandom.

## Credit

Snoopynha
(She/Her)

*On Jira for assign the issues: Camilly Vitória Barbosa dos Santos


sarken:
Hi, Snoopynha!

I've updated the Jira issue status to In Review so no one will mistakenly create a duplicate pull request. However, Jira has been unreliable about showing us new accounts, so I wasn't able to update your permissions or assign the issue to you. If you could either reply here or send an email to [otw-coders@transformativeworks.org](mailto:otw-coders@transformativeworks.org) with your Jira account's email address, we'll get that fixed.

(One quick note about Jira names: it makes things a bit easier for us on the organizational side if the Full Name on your Jira account either closely matches the name you'd like us to credit in the release notes or includes it in parentheses, e.g. "Nickname (CREDIT NAME).")

Snoopynha:
@sarken Thank you for the tip, I really appreciate it!

I’ve updated the Full Name on my Jira account to include my preferred credit name as recommended.

For reference, my Jira account email is: [vitoriacamilly308@gmail.com]

Please let me know if anything else is needed.
Thanks again for your support!

brianjaustin (changes requested):
Thanks for the PR! We're moving away from rake tasks for scheduled jobs, though. Could you make a job like https://github.com/otwcode/otwarchive/blob/d0531a21c47e8d5348cc03c4a7a8f6b341a16741/app/jobs/invite_from_queue_job.rb and add it to the [Resque schedule configuration](https://github.com/otwcode/otwarchive/blob/d0531a21c47e8d5348cc03c4a7a8f6b341a16741/config/resque_schedule.yml) instead?

Snoopynha:
Hey @brianjaustin, I've converted the logic to a ResetBookmarkTagsJob and added it to the resque_schedule.yml configuration as suggested. 

The Rake task files have been removed. Let me know if there's anything else needed.

config/resque_schedule.yml
```diff
@@ -122,3 +122,10 @@ disable_admin_post_comments:
   description: >-
     Disables all comments on admin (news) posts older than the
     configured window.
+
+reset_bookmark_tags_job:
+  cron: "0 0 1 */3 *"
+  class: "ResetBookmarkTagsJob"
+  queue: low
+  description: "Reset bookmark-only tags to unsorted quarterly (AO3-7294)"
```
brianjaustin: Could you add a [trailing newline](https://dan-niles.medium.com/why-github-prs-shows-that-red-circle-understanding-newlines-at-end-of-files-8a4869821694) to this file please?

spec/jobs/reset_bookmark_tags_job_spec.rb
```diff
@@ -0,0 +1,52 @@
+require "spec_helper"
+
+describe ResetBookmarkTagsJob do
+  describe "#perform" do
+    it "should reset invalid bookmarks and ignore legitimate ones" do
```
brianjaustin: Per our [test style guide](https://github.com/otwcode/otwarchive/wiki/Automated-Testing#user-content-Our_RSpec_Standards), we don't use `should` in test descriptions. This would therefore need to be
```suggestion
    it "resets invalid bookmarks and ignores legitimate ones" do
```

However, we also prefer each individual test to be focused on the minimum number of conditions. Therefore, I would break this into
* it resets a non-canonical tag with no work uses
* it does not reset a non-canonical tag with work uses
* it does not reset a canonical tag

spec/jobs/reset_bookmark_tags_job_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+      # 1. Scenario: An "incorrect" tag - incorrect type, with kinship, without works - should NEVER be changed.
+      # It should revert to "Tag", with associations removed
```
brianjaustin: This comment is a bit confusing to me. What does 'NEVER be changed' mean? I would expect it _to be_ changed, AKA reset.

app/jobs/reset_bookmark_tags_job.rb
```diff
@@ -0,0 +1,19 @@
...
+  queue_as :low
+
+  def perform
+    # Searches for non-canonical tags with zero usage count (only bookmarks or nothing)
+    Tag.where(canonical: false, taggings_count_cache: 0).find_each(batch_size: 1000) do |tag|
+      # Check if have associations (parents) or categorized types
+      has_type = tag.type.present? && tag.type != "Tag"
+      has_parents = tag.common_taggings.exists?
+
+      if has_type || has_parents
```
brianjaustin: Minor stylistic thing: could you rewrite this as a guard? Something like:
```suggestion
      return unless has_type || has_parents
```

app/jobs/reset_bookmark_tags_job.rb
```diff
@@ -0,0 +1,19 @@
+class ResetBookmarkTagsJob < ApplicationJob
+  queue_as :low
+
+  def perform
+    # Searches for non-canonical tags with zero usage count (only bookmarks or nothing)
+    Tag.where(canonical: false, taggings_count_cache: 0).find_each(batch_size: 1000) do |tag|
```
brianjaustin: Any reason not to use `nonsynonymous` here? If I understand correctly, we don't want to clean up synonyms either
```suggestion
    Tag.nonsynonymous.where(taggings_count_cache: 0).find_each(batch_size: 1000) do |tag|
```

app/jobs/reset_bookmark_tags_job.rb
```diff
@@ -0,0 +1,19 @@
...
+      # Check if have associations (parents) or categorized types
+      has_type = tag.type.present? && tag.type != "Tag"
+      has_parents = tag.common_taggings.exists?
+
+      if has_type || has_parents
+        # Removes data from the association table (Wrangling)
+        tag.common_taggings.delete_all
+        # Return the type to 'Tag' (status 'Unsorted')
+        tag.update_column(:type, "Tag")
```
brianjaustin: Both of these skip callbacks when doing the deletion/updates. Is there a reason to do that here? It seems a bit unsafe

app/jobs/reset_bookmark_tags_job.rb
brianjaustin: Batching is a good idea, but I think we'll still run into performance issues because there are likely to be quite a lot of tags. What about enqueuing batches to Redis like the [user indexing job](https://github.com/otwcode/otwarchive/commit/14df98fc2516c2a8279571393051b03410132762) or tag reindexing job?

brianjaustin (changes requested):

config/resque_schedule.yml
brianjaustin: We can check this in the git history, so no need to include the ticket ID here
```suggestion
  description: Reset bookmark-only tags to unsorted quarterly
```

spec/jobs/reset_bookmark_tags_job_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+                         canonical: false,
+                         taggings_count_cache: 0)
+
+      # Valid association: Character can be a child of Fandom
+      CommonTagging.create!(common_tag: dirty_tag, filterable: fandom)
+
+      # Execute the job
+      ResetBookmarkTagsJob.perform_now(Tag.pluck(:id))
+
+      # Checks if the tag has been cleared (returned to "Tag" and lost the fandom)
```
brianjaustin: While I'm in favour of comments that explain the why of complex code, this doesn't really need an explanation and IMO interrupts the flow of reading the test (similarly for the other comments)

spec/jobs/reset_bookmark_tags_job_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+
+      # Tag with incorrect type (Character) and no works - will be cleaned up
+      dirty_tag = create(:tag,
+                         name: "Dirty Tag",
+                         type: "Character",
+                         canonical: false,
+                         taggings_count_cache: 0)
+
+      # Valid association: Character can be a child of Fandom
+      CommonTagging.create!(common_tag: dirty_tag, filterable: fandom)
```
brianjaustin, line 14: This can also be created via FactoryBot, which is preferred for tests
```suggestion
      create(:common_tagging, common_tag: dirty_tag, filterable: fandom)
```

app/jobs/reset_bookmark_tags_job.rb
```diff
@@ -0,0 +1,30 @@
+class ResetBookmarkTagsJob < ApplicationJob
+  queue_as :low
+
+  def perform(tag_ids = nil)
+    # Maintains the use of "canonical: false" to avoid clearing official tags with 0 works
+    base_scope = Tag.nonsynonymous.where(canonical: false, taggings_count_cache: 0)
+
+    # Searches for non-canonical tags with zero usage count (only bookmarks or nothing)
+    if tag_ids.nil?
```
brianjaustin: The branching logic here makes it harder to follow what the job does. Since the tasks are related, one idea would be to split this bit off into something like ` queue_tags` and specify that method specifically in the resque config. Alternately, you could split this into two separate classes. I do not have a strong preference for which

app/jobs/reset_bookmark_tags_job.rb
```diff
@@ -0,0 +1,30 @@
+class ResetBookmarkTagsJob < ApplicationJob
+  queue_as :low
+
+  def perform(tag_ids = nil)
+    # Maintains the use of "canonical: false" to avoid clearing official tags with 0 works
```
brianjaustin: Similarly to the tests, these comments focus on the what and not the how of things. For readability, please remove them

app/jobs/reset_bookmark_tags_job.rb
```diff
@@ -0,0 +1,30 @@
...
+    base_scope = Tag.nonsynonymous.where(canonical: false, taggings_count_cache: 0)
```
brianjaustin: The [nonsynonymous scope](https://github.com/otwcode/otwarchive/blob/0833c1d7d8997c8c11cbdfecf995487fb2f302c7/app/models/tag.rb#L286) already implies noncanonical
```suggestion
    base_scope = Tag.nonsynonymous.where(taggings_count_cache: 0)
```

brianjaustin (commented):
Thanks for the changes! I have some more feedback based on the latest version of the code. I also wanted to make sure you've read and are aware of our [policy on AI in relation to the project](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md#contributing-code)

app/jobs/reset_bookmark_tags_job.rb
Snoopynha: I preferred to do it separately because I tried to find out how to deal with a method within a job in the schedule and didn't quite understand it, but let me know if the way I did it might not be ideal.

app/jobs/enqueue_bookmarks_tags_job.rb
```diff
@@ -0,0 +1,11 @@
+class EnqueueBookmarksTagsJob < ApplicationJob
+  queue_as :low
+
+  def perform
+    base_scope = Tag.nonsynonymous.where(taggings_count_cache: 0)
+      
+    base_scope.find_in_batches(batch_size: 1000) do |tags|
```
brianjaustin: Noting for the other OTW vols: this will result in about 1.2k batches to start out with (maximum), which should be handleable by Redis/Resque

app/jobs/enqueue_bookmarks_tags_job.rb
```diff
@@ -0,0 +1,11 @@
+class EnqueueBookmarksTagsJob < ApplicationJob
+  queue_as :low
+
+  def perform
+    base_scope = Tag.nonsynonymous.where(taggings_count_cache: 0)
```
brianjaustin: I ran a test on our staging environment, and unfortunately `taggings_count_cache` includes bookmarks as well. Two queries I tried to get this information (below) had bad enough performance on the same staging environment that they stalled out after several hours without getting past taggings from 2016 😬 

**Option 1**
```sql
SELECT `taggings`.* FROM `taggings` WHERE `taggings`.`taggable_type` NOT IN ('Work', 'ExternalWork') AND `taggings`.`taggable_id` NOT IN (SELECT `taggings`.`taggable_id` FROM `taggings` WHERE `taggings`.`taggable_type` IN ('Work', 'ExternalWork'));
````

**Option 2**
```sql
SELECT `taggings`.* FROM `taggings` LEFT JOIN taggings AS taggings_inner 
ON taggings.taggable_id = taggings_inner.taggable_id 
AND taggings_inner.taggable_type IN ('Work', 'ExternalWork') WHERE `taggings`.`taggable_type` NOT IN ('Work', 'ExternalWork') AND `taggings_inner`.`taggable_id` IS NULL;
```

If you have thoughts on a better query, I can try running them on our staging environment (or at least an `explain` with a real-ish volume of data) to check their performance. I'm also checking with Tag Wrangling to see if there are any constraints we could add (e.g. limit to more recent tags) to make this work better.

Snoopynha:
@brianjaustin I tried to come up with some solutions, and the best I could come up with was to try replacing the "NOT IN" clauses with "NOT EXISTS" which usually performs better on very large databases, and instead of using the relationship table as a reference, which is usually very large, use the main "tags" table, which might be smaller for comparison. I’m not sure if it will work 🙃, but the query could look something like this:

```
SELECT `tags`.* FROM `tags`
WHERE `tags`.`canonical` = 0 
AND `tags`.`merger_id` IS NULL
AND NOT EXISTS (
  SELECT 1 FROM `taggings` 
  WHERE `taggings`.`tag_id` = `tags`.`id` 
  AND `taggings`.`taggable_type` IN ('Work', 'ExternalWork')
);
```

brianjaustin:
> @brianjaustin I tried to come up with some solutions, and the best I could come up with was to try replacing the "NOT IN" clauses with "NOT EXISTS" which usually performs better on very large databases, and instead of using the relationship table as a reference, which is usually very large, use the main "tags" table, which might be smaller for comparison. I’m not sure if it will work 🙃, but the query could look something like this:
> 
> ```
> SELECT `tags`.* FROM `tags`
> WHERE `tags`.`canonical` = 0 
> AND `tags`.`merger_id` IS NULL
> AND NOT EXISTS (
>   SELECT 1 FROM `taggings` 
>   WHERE `taggings`.`**tagger_id**` = `tags`.`id` 
>   AND `taggings`.`taggable_type` IN ('Work', 'ExternalWork')
> );
> ```

That looks promising! One naming error, which I've corrected and bolded in the above quote block. It did run in a semi-reasonable amount of time on a non-online database replica. I'm going to pass a sample of the results off to the tag wrangling team to make sure things look reasonable to them, and if it passes their checks I think we have a winner!

(Another note: I think we can add a limit of a few thousand to make the query even safer.)

brianjaustin:
@Snoopynha tag wrangling leadership has confirmed that the sample I provided is correct, so I think you can go ahead and try implementing with that version of the query! Thanks!

app/jobs/enqueue_bookmarks_tags_job.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  def perform
+    base_scope = Tag.nonsynonymous
+                    .where(canonical: false)
+                    .where("NOT EXISTS (
+                      SELECT 1 FROM taggings 
+                      WHERE taggings.tagger_id = tags.id 
+                      AND taggings.taggable_type IN ('Work', 'ExternalWork')
+                    )")
+                    .limit(5000)
```
brianjaustin: Thinking a bit more about this limit: will it mean we only ever clean up the first 5000 tags? I don't think there's anything in the query that will skip the tags that were cleaned up last time 🤔

spec/jobs/reset_bookmark_tags_job_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+describe ResetBookmarkTagsJob do
+  describe "#perform" do
+    it "resets a non-canonical tag with no work uses" do
+      fandom = create(:fandom, canonical: true)
+
+      dirty_tag = create(:tag,
+                         name: "Dirty Tag",
+                         type: "Character",
+                         canonical: false,
+                         taggings_count_cache: 0)
```
brianjaustin: Can you update the tests to give this a non-zero value? That way, they will reflect the reason why the query above needed to b changed

app/jobs/reset_bookmark_tags_job.rb
```diff
@@ -0,0 +1,23 @@
...
+                    )")
+    
+    base_scope.where(id: tag_ids).each do |tag|
+      has_type = tag.type.present? && tag.type != "Tag"
+      has_parents = tag.common_taggings.exists?
+      
+      next unless has_type || has_parents
+        
+      tag.common_taggings.destroy_all
+      tag.update(type: "Tag")
```
brianjaustin: I think this actually should be `UnsortedTag`?

brianjaustin (commented):
Thanks for the update! A few more comments from me, plus the rubocop issues

app/jobs/enqueue_bookmarks_tags_job.rb
Snoopynha: I was thinking that would be a way to make the question of "limit of a few thousand to make the query safer" that you have recommended. Can you think of another way I could do this? And regarding tags that have already been cleared, I thought that once the tags and their types were cleared, they wouldn't be found again by the request.

brianjaustin: The cleanup will change the tag type, but not remove any `taggings`. So since the query doesn't filter by type `!= 'Tag'` (so find `Fandom`, `Character`, etc.) it will still return things tags that were previously cleaned up. I think if you add that check to this SQL, the limit will be safe

(That was also missing from my initial draft query too, so apologies for the extra step!)

spec/jobs/reset_bookmark_tags_job_spec.rb
Snoopynha: I changed it to 1.
I ended up forgetting that 0 represents tags with no history or bookmarks 🫠 sorry.

app/jobs/enqueue_bookmarks_tags_job.rb
Snoopynha: I think that this time is right

brianjaustin (approved):
Let's give it a try!

For AD&T volunteers: make sure to check with TW before merging this to make sure there is adequate capacity on their side to test this fully.

PR #5625, opened 2026-03-10
AO3-4989 Keep filtering when redirecting after deleting bookmark from user bookmarks page

ReverM:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4989

## Purpose

This makes it so that deleting bookmark from the user page does keep the filtering, when using javascript. Behaviour from the general bookmark page stays unchanged

## Testing Instructions

See jira comment

## Credit

Danaël / Rever ( they / he )
Danaël Villeneuve on jira


app/controllers/bookmarks_controller.rb
```diff
@@ -288,7 +286,12 @@ def confirm_delete
   def destroy
     @bookmark.destroy
     flash[:notice] = ts("Bookmark was successfully deleted.")
-    redirect_to user_bookmarks_path(current_user)
+    if request.referer&.match("&user_id") || request.referer&.match(user_bookmarks_path(current_user))
```
Bilka2: This is matching too many possible URLs in my opinion, but I see that you can't really match more closely based on URL here.

What do you think about instead using a `return_to` parameter on the link and redirecting to that here (refer to e.g. what we do in the destroy session action)? Then you could set that on the delete link only when it's getting displayed by the right controller

features/bookmarks/bookmark_javascript.feature
```diff
@@ -139,3 +139,42 @@ Feature: Create and edit bookmarks with javascript enabled
...
+      And all indexing jobs have been run
+      And I follow "Delete" in the bookmarkable blurb for "Shakra"
+      And I confirm the popup
+    Then I should not see "Shakra"
+      And I should not see "Hornet"
+      And I should see "Sherma"
+
+  @javascript
+  Scenario: Deleting bookmarks from the bookmarks page does not keep the filtering
+      Given I am logged in as "PharloomEditor"
```
Bilka2:
```suggestion
    Given I am logged in as "PharloomEditor"
```
(I really like your examples :D)

features/step_definitions/web_steps.rb
```diff
@@ -95,6 +95,14 @@ def with_scope(locator)
   end
 end
 
+When "I confirm the popup" do
+  page.driver.browser.switch_to.alert.accept
+end
```
Bilka2: For this new step, could you use `page.accept_alert` instead? It's what's used by other test steps, with some additional verification, so it's probably good to copy that

For the other step, it looks like there is no dismiss_alert, but there is accept_confirm and dismiss_confirm so maybe that's better...?

Bilka2 (changes requested):

features/bookmarks/bookmark_javascript.feature
ReverM: Hehe

app/controllers/bookmarks_controller.rb
ReverM: Filtering on all bookmarks, and filtering on the bookmark of your bookmarks are handled by the same controller, and both by the same action so I'm unsure if that would work.

features/step_definitions/web_steps.rb
ReverM: Switched to accept_confirm and dismiss_confirm

app/controllers/bookmarks_controller.rb
Bilka2: I poked internally and it sounds like we'd prefer to keep the filtering and stay on the same page for basically all pages that use the bookmarks#index action, instead of redirecting to the `user_bookmarks_path` (whether it's the bookmarks listing in a tag, or the bookmarks of a work or whatever). So whichever implementation to redirect here would be fine with us, whether that's a parameter like I proposed or redirecting to the referrer if we're coming from the right url, which I think would be more standard with other places in the code? I hope that allows us to adjust the matching here to be a bit tighter

features/step_definitions/web_steps.rb
Bilka2: Could you check the return value, similar to how this code does it?
https://github.com/otwcode/otwarchive/blob/6d5bceaa831a9c5a14ae4680556e7c332b68ab67/features/step_definitions/comment_steps.rb#L395

app/controllers/bookmarks_controller.rb
ReverM: So redirect back to the page you were currently on regardless?

Bilka2: No, since that'd be confusing when for example viewing the bookmark (with show) since then you'd get redirected to a 404.

It should go back to the current page when viewing a list of bookmarks, so in the bookmarks controller index action. According to the routes, that should be these pages:
```
              tag_bookmarks GET    /tags/:tag_id/bookmarks(.:format)                              bookmarks#index
             user_bookmarks GET    /users/:user_id/bookmarks(.:format)                            bookmarks#index
       user_pseud_bookmarks GET    /users/:user_id/pseuds/:pseud_id/bookmarks(.:format)           bookmarks#index
             work_bookmarks GET    /works/:work_id/bookmarks(.:format)                            bookmarks#index
    external_work_bookmarks GET    /external_works/:external_work_id/bookmarks(.:format)          bookmarks#index
           series_bookmarks GET    /series/:series_id/bookmarks(.:format)                         bookmarks#index
       collection_bookmarks GET    /collections/:collection_id/bookmarks(.:format)                bookmarks#index
                  bookmarks GET    /bookmarks(.:format)                                           bookmarks#index
```
The other pages where you can delete bookmarks should keep redirecting to `user_bookmarks`

features/step_definitions/web_steps.rb
ReverM: Could we create a new step that generalizes all alerts?

app/controllers/bookmarks_controller.rb
ReverM: Are `show` and `index` the only two actions that can display a delete button for a bookmark?

Bilka2: I don't know, and I think it would be fragile to rely on that. The solution to store this information in the session is not desirable, we try to limit what we use cookies for. Sorry for not being clearer about this earlier! 

I think this should either be done with a parameter, which you can set where the link is created in the partials depending on `controller.controller_name` and `controller.action_name`, or you can do the solution with parsing the referer, similar to what the comment controller redirects do

features/step_definitions/web_steps.rb
Bilka2: I'd prefer to keep separate steps that check the full return value for now. That being said, these steps are unused now, please remove them

Bilka2 (changes requested):

app/controllers/bookmarks_controller.rb
```diff
@@ -288,7 +286,13 @@ def confirm_delete
   def destroy
     @bookmark.destroy
     flash[:notice] = ts("Bookmark was successfully deleted.")
-    redirect_to user_bookmarks_path(current_user)
+    referer_route = Rails.application.routes.recognize_path request.referer
```
Bilka2: This method is undocumented and from what I can find should not be used: https://github.com/rails/rails/issues/21477#issuecomment-171873537. The current usage in the Rails source code are exclusively for tests, so it does not seem intended for production

So I would prefer a regex instead

ReverM: Ah I see, I wanted to find a solution that wouldn't need to be modified if new routes that use that controller/action combination were to be introduced. I shall switch to string parsing then

ReverM: @Bilka2 I just noticed that search_bookmarks does not use the bookmarks#index action, but displays a list of bookmark. Should it also redirect back?

Bilka2: Yeah, I saw the code and went "oh sweet I didn't know that method existed, such a nice solution", opened the docs and then got sad because it's not meant to be used :(

I don't really mind either way for the search, probably better to include it, and since you have it now it's fine!

Bilka2: For the regex, from what I tested it also works without the `.*` at the start, maybe it would be better to remove that to reduce how complex it is? Also, would it be possible to replace the `$` with `\z`? the `$` means end of line and `\z` means end of string. Since you're using the parsed path, there shouldn't be any newlines, so no difference, but to be safe

But that's nitpicks. My bigger request here is to rescue `URI::InvalidURIError` so we don't error 500 if someone feeds us crap

ReverM: Could we alternatively check if the URI is valid via regexp or would it be safer to have an explicit rescue statement?

Bilka2: Making a regex is very hard (we try in other places and it doesn't catch all invalid URLs) so the rescue is better

app/controllers/bookmarks_controller.rb
```diff
@@ -288,6 +286,13 @@ def confirm_delete
   def destroy
     @bookmark.destroy
     flash[:notice] = ts("Bookmark was successfully deleted.")
+    # We check that the URL from where the bookmark was deleted is a list of bookmarks. If it is not, we cannot redirect the user back since it would cause a 404 error
+    if %r{(/bookmarks)|(bookmarks/search)\z}.match?(URI.parse(request.referer).path)
```
Bilka2:
```suggestion
    if %r{(/bookmarks\z)|(bookmarks/search\z)}.match?(URI.parse(request.referer).path)
```

Otherwise it matches /bookmarks/12345: https://rubular.com/r/Q59psgXvlIalwG (in that link I had to escape the slashes, that is _not_ needed here due to how you're defining it, which is better)

ReverM: Oh, I tried to factor out the /z, I guess I did it incorrectly

Bilka2 (approved):
Thanks for bearing with me through this!

PR #5645, opened 2026-03-18
AO3-2782 Disallow assignment of noncanonical fandom wrangling

ReverM:
## Issue

https://otwarchive.atlassian.net/browse/AO3-2782

## Purpose

This removes the ability for tag wranglers to assign non-canonical fandom to them, as well as adding a task that removes assignments of non-canonical fandoms 

## Credit

Danaël / Rever ( they / he )
Danaël Villeneuve on Jira


ReverM:
The tests are not quite ready. The task one is giving false positive
Update: tests are fixed

spec/lib/tasks/after_tasks.rake_spec.rb
```diff
@@ -143,6 +143,23 @@
   end
 end
 
+describe "rake After:remove_noncanonical_fandom_wrangling_assignments" do
+  let(:noncanonical_fandom) { create(:fandom, canonical: false) }
+  let(:user) { create(:tag_wrangler) }
+  let(:assignment1) { create(:wrangling_assignment, user_id: :user.id, fandom_id: :noncanonical_fandom.id) }
+  
+  it "deletes wrangling assignments of noncanonical fandoms" do
+    expect(WranglingAssignment.none?)
```
Bilka2: this needs a `.to be_truthy`, otherwise it doesn't check anything. Though maybe `be_empty` is useful here? I wasn't thinking about that when you asked about this code before, but it might work. Same for the other `expect`

lib/tasks/after_tasks.rake
```diff
@@ -167,6 +167,11 @@ namespace :After do
     STDOUT.flush
   end
 
+  desc "Remove wrangling assigments of non-canonical fandoms"
```
Bilka2: I think we usually add to the bottom of this file, could you move the task?

lib/tasks/after_tasks.rake
```diff
@@ -167,6 +167,11 @@ namespace :After do
...
+  task(remove_noncanonical_fandom_wrangling_assignments: :environment) do
+    WranglingAssignment.joins("LEFT JOIN tags ON (tags.id = wrangling_assignments.fandom_id)").where(tags: { canonical: false }).find_each(&:destroy)
```
Bilka2: Could you `destroy!` here so we get an error if destruction fails please

spec/lib/tasks/after_tasks.rake_spec.rb
```diff
@@ -143,6 +143,23 @@
...
+  end
+
+  let(:canonical_fandom) { create(:fandom, canonical: true) }
```
Bilka2:
```suggestion
  let(:canonical_fandom) { create(:canonical_fandom) }
```

spec/lib/tasks/after_tasks.rake_spec.rb
```diff
@@ -143,6 +143,23 @@
   end
 end
 
+describe "rake After:remove_noncanonical_fandom_wrangling_assignments" do
+  let(:noncanonical_fandom) { create(:fandom, canonical: false) }
+  let(:user) { create(:tag_wrangler) }
+  let(:assignment1) { create(:wrangling_assignment, user_id: :user.id, fandom_id: :noncanonical_fandom.id) }
```
Bilka2: hm, I've never seen the symbol used later in the let, shouldn't it be `user_id: user.id, fandom_id: noncanonical_fandom.id`?

Anyway, pretty sure you don't need to make and set a wrangler here, the wrangling assignment factory will take care of that for you (same for the other test)

spec/lib/tasks/after_tasks.rake_spec.rb
```diff
@@ -143,6 +143,23 @@
   end
 end
 
+describe "rake After:remove_noncanonical_fandom_wrangling_assignments" do
```
Bilka2: (Same thing regarding order in the file as the task itself)

app/controllers/tag_wranglers_controller.rb
```diff
@@ -52,7 +52,8 @@ def report_csv
     results = [%w[Name Last\ Updated Type Merger Fandoms Unwrangleable]]
     wrangled_tags.find_each(order: :desc) do |tag|
       merger = tag.merger&.name || ""
-      fandoms = tag.parents.filter_map { |parent| parent.name if parent.is_a?(Fandom) }.join(", ")
+      fandoms = tag.parents.filter_map { |parent| parent.name if parent.is_a?(Fandom) }
```
Bilka2, line 55: I'm squinting at the fact that this filters in ruby instead of SQL, but you didn't modify the code here so whatever, let's keep it

app/views/tags/edit.html.erb
```diff
@@ -64,7 +64,13 @@
 
       <% if @wranglers %>
         <dt><%= ts("Wranglers") %></dt>
-        <dd><%= wrangler_list(@wranglers, @tag) %></dd>
+          <% if (@tag.type.to_s == "Fandom") & !@tag.canonical? %>
```
Bilka2: Instead of this, I think it'd be better to change the `wrangler_list` method instead, it already does a bunch of checks depending on tag status. That way you also preserve the behaviour that a synned fandom lists the wranglers of its canonical, which I'm pretty sure should be kept

features/tags_and_wrangling/tag_wrangling_fandoms.feature
```diff
@@ -49,9 +49,23 @@ Scenario: fandoms wrangling - syns, mergers, autocompletes, metatags
     And I press "Post"
   Then I should see "Work was successfully posted."
 
-  # editing non-canonical fandom in order to syn it to existing canonical merger
+  # cannot sign up through edit page to a non-canonical fandom
```
Bilka2: I'd prefer if you could stick the test into a separate scenario, since it's not related to the basic wrangling actions that this one is about

app/controllers/tag_wranglers_controller.rb
```diff
@@ -65,13 +66,16 @@ def create
     unless params[:tag_fandom_string].blank?
       names = params[:tag_fandom_string].gsub(/$/, ',').split(',').map(&:strip)
       fandoms = Fandom.where('name IN (?)', names)
+      noncanonical_fandoms = []
       unless fandoms.blank?
-        for fandom in fandoms
-          unless !current_user.respond_to?(:fandoms) || current_user.fandoms.include?(fandom)
+        fandoms.each do |fandom|
+          unless !current_user.respond_to?(:fandoms) || current_user.fandoms.include?(fandom) || !fandom.canonical?
```
Bilka2: I would prefer to have the canonical check be inside of the conditional here, that way you just need a single check for both the error and the not assigning:
```
unless fandom.canonical?
  push...
  next
end
```

Bilka2: Also, while you're here, could you move the `!current_user.respond_to?(:fandoms)` check out of the loop? We really don't need to recheck that every iteration when we could just skip the whole thing...

Bilka2 (changes requested):
Thanks for working on this!

app/controllers/tag_wranglers_controller.rb
ReverM: Should be done

lib/tasks/after_tasks.rake
ReverM: Should be done

ReverM:
Seems like I did something that made the tests very unhappy 😓

app/controllers/tag_wranglers_controller.rb
```diff
@@ -65,13 +66,19 @@ def create
     unless params[:tag_fandom_string].blank?
       names = params[:tag_fandom_string].gsub(/$/, ',').split(',').map(&:strip)
       fandoms = Fandom.where('name IN (?)', names)
-      unless fandoms.blank?
-        for fandom in fandoms
-          unless !current_user.respond_to?(:fandoms) || current_user.fandoms.include?(fandom)
-            assignment = current_user.wrangling_assignments.build(fandom_id: fandom.id)
-            assignment.save!
+      noncanonical_fandoms = []
+      unless fandoms.blank? || !current_user.respond_to?(:fandoms)
```
Bilka2:
```suggestion
      if fandoms.present? && current_user.respond_to?(:fandoms)
```

app/helpers/tags_helper.rb
```diff
@@ -41,9 +41,11 @@ def tag_cloud(tags, classes)
   def wrangler_list(wranglers, tag)
     if wranglers.blank?
       if @tag[:type] == 'Fandom'
+        return t("tags_helper.edit.wrangler_noncanon_fandom") unless tag.canonical?
```
Bilka2:
```suggestion
        return t("tags_helper.wrangler_list.wrangler_noncanon_fandom") unless tag.canonical?
```
since this is in the helper method now I'd name it after the method

app/helpers/tags_helper.rb
```diff
@@ -41,9 +41,11 @@ def tag_cloud(tags, classes)
...
+        
         sign_up_fandoms = tag.name
       elsif Tag::USER_DEFINED.include?(@tag.class.name) && !tag.fandoms.blank?
-        sign_up_fandoms = tag.fandoms.collect(&:name).join(', ')
+        sign_up_fandoms = tag.fandoms.select(&:canonical?).collect(&:name).join(", ")
```
Bilka2:
```suggestion
        sign_up_fandoms = tag.fandoms.where(canonical: true).pluck(:name).join(", ")
```
This is better for performance: The `where` does the filtering in SQL instead of Ruby, and with `pluck` we end up loading only the name from the database, instead of loading the whole tag and then discarding everything besides the name

features/tags_and_wrangling/tag_wrangling_fandoms.feature
```diff
@@ -213,3 +213,31 @@ Scenario: Checking the media pages
...
     And I should see "Be a second B fandom" within "#letter-B .tags"
     And I should see "Be another thing" within "#letter-B .tags"
+
+Scenario: Cannot sign up for non-canonical fandom
+  Given the following activated tag wrangler exists
+    | login  | password    |
+    | Enigel | wrangulate! |
+    And basic tags
+    And a media exists with name: "TV Shows", canonical: true
+    And a character exists with name: "Neal Caffrey", canonical: true
```
Bilka2: I think these are unused?

Bilka2 (commented):
All working, yay! For my review comments, I only consider the select -> where one required, the rest are style changes that you can just ignore if you want. If it wasn't for the select -> where thing, I would've marked this reviewed already

Thank you so much for dealing with the flaky tests!!

app/helpers/tags_helper.rb
Bilka2: @ReverM Ugh, sorry about the suggestion that immediately went and caused a test failure. Apparently, freeforms use a different way to get their fandoms than any of the other tags. Looking at when this was changed (https://github.com/otwcode/otwarchive/commit/60916c64#diff-21bc23c49cb748aea4195c18c36ee8db9ef6dab622407537f8a11c9b0c02292c), it looks like freeform.rb should be updated to use the same `def fandoms` method like e.g. character.rb does, so `parents.by_type('Fandom').by_name`. I think it should be fine to change this as part of this PR

features/tags_and_wrangling/tag_wrangling_fandoms.feature
ReverM: I deleted a bit more. There were more unneeded stuff

Bilka2 (approved):
Thank you!

lib/tasks/after_tasks.rake
```diff
@@ -654,7 +654,7 @@ namespace :After do
     AuditsBackfillJob.spawn_jobs
     puts "Backfill started and running on resque in background"
   end
-
+
```
Bilka2:
```suggestion

```

PR #5651, opened 2026-03-21
AO3-6454 AO3-6198 Remove Akismetor and handle Akismet API calls ourselves

marcus8448:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6454
https://otwarchive.atlassian.net/browse/AO3-6198

## Purpose

Removes the unmaintained akismetor gem and replaces it with our own Akismet client implementation.

I kept the API surface the same (class methods), but it should be fairly trivial to change to be more like Zoho if wanted.
I also hardcoded the key/blog name to their config values since those are the only values we use.

## Credit

marcus8448 (he/him)


app/models/comment.rb
```diff
@@ -510,15 +508,15 @@ def skip_spamcheck?
   def spam?
     return false unless %w[staging production].include?(Rails.env)
 
-    Akismetor.spam?(akismet_attributes)
+    AkismetClient.spam?(akismet_attributes)
   end
 
   def submit_spam
-    Rails.env.production? && Akismetor.submit_spam(akismet_attributes)
+    Rails.env.production? && AkismetClient.submit_spam(akismet_attributes)
```
brianjaustin, line 515: Could you extract `Rails.env.production?` to a (private) helper method, something like `spam_checking_enabled?`? That way, we can test things (slightly) more easily in unit tests later

brianjaustin: Actually, do we need it now that the client automatically disables itself if the config values are not present?

lib/akismet_client.rb
```diff
@@ -0,0 +1,48 @@
+class AkismetClient
+  include HTTParty
+  base_uri "https://rest.akismet.com"
+  headers "User-Agent": "#{ArchiveConfig.APP_SHORT_NAME}/#{ArchiveConfig.REVISION}",
+          "Content-Type" => "application/x-www-form-urlencoded"
+
+  SUBMISSION_RESPONSE = "Thanks for making the web a better place.".freeze
+
+  def self.valid_key?
```
brianjaustin, line 9: Is this only used for testing?

lib/akismet_client.rb
```diff
@@ -0,0 +1,48 @@
+class AkismetClient
+  include HTTParty
+  base_uri "https://rest.akismet.com"
+  headers "User-Agent": "#{ArchiveConfig.APP_SHORT_NAME}/#{ArchiveConfig.REVISION}",
+          "Content-Type" => "application/x-www-form-urlencoded"
+
+  SUBMISSION_RESPONSE = "Thanks for making the web a better place.".freeze
```
brianjaustin, line 7:
```suggestion
  SUBMISSION_RESPONSE = "Thanks for making the web a better place."
```

With newer versions of Ruby, frozen is becoming the default IIUC

spec/lib/akismet_client_spec.rb
```diff
@@ -0,0 +1,108 @@
...
+
+      it "returns true" do
+        expect(AkismetClient.enabled?).to be_truthy
+      end
+    end
+  end
+
+  context "with a valid api key" do
+    before(:all) do
+      skip "Missing valid Akismet API key" unless AkismetClient.valid_key?
```
brianjaustin, line 43: I see this in the logs; do we need to add this as a secret to the tests? It would mean test coverage, but I worry about sending waaaay too many calls to Akismet as a result. And if the tests don't run in CI, is it worth including them?

Maybe another option: https://stackoverflow.com/a/53666136 (or the suggestions in other answers) 🤔

app/models/comment.rb
marcus8448: > Actually, do we need it now that the client automatically disables itself if the config values are not present?

We still need it to not send things on staging

lib/akismet_client.rb
marcus8448: I think that's Ruby 4, so we're not quite there yet. But I've changed it to use the magic `frozen_string_literal` comment since that'll be easier to find+replace when we get there.

app/models/comment.rb
marcus8448: Yeah, we still need it to not send things on staging.

spec/lib/akismet_client_spec.rb
marcus8448: I agree that adding the key to CI would be too risky. I've changed it to use WebMock to mock the API, but it's definitely less rigorous of a test.

lib/akismet_client.rb
```diff
@@ -0,0 +1,47 @@
...
+    ArchiveConfig.AKISMET_KEY.present? && ArchiveConfig.AKISMET_NAME.present?
+  end
+
+  def self.spam_submission_enabled?
+    Rails.env.production?
+  end
+
+  def self.encode_body(body = {})
+    # disable learning in non-production environments (https://akismet.com/blog/pro-tip-testing-testing/)
+    body = body.merge(is_test: true) unless Rails.env.production?
```
marcus8448, line 40: Not a part of the ticket, but I think it makes sense to not train on staging data.

spec/lib/akismet_client_spec.rb
brianjaustin: Less rigorous but enabled > more rigorous but not running in CI in my books :smile:

brianjaustin:
Switching the label back to action needed due to the failing tests, but otherwise looking good!

brianjaustin (approved):
Thanks!

PR #5653, opened 2026-03-22
AO3-4583 Fix redirect when creating top-level comment on work with multiple pages of comments

marcus8448:
## Issue

https://otwarchive.atlassian.net/browse/AO3-4583

## Purpose

Fixes the comment creation redirect to point at the page of the comment you just created instead of the page of comments you were on.

## Credit

marcus8448 (he/him)

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -427,6 +427,7 @@ class CommentsController < ApplicationController
               # rubocop:disable Metrics/BlockNesting
               view_full_work = params[:view_full_work] == "true"
               parent = view_full_work || current_user&.preference&.view_full_works? ? @comment.ultimate_parent : @comment.parent
+              # if this comment starts a new top-level thread, it will always be on the last page of comments
               page = parent.comments.reviewed.count.ceildiv(Comment.per_page) if @comment.thread == @comment.id
               redirect_to_comment(@comment, { view_full_work: view_full_work, page: page || params[:page] })
               # rubocop:enable Metrics/BlockNesting
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -424,7 +424,12 @@ class CommentsController < ApplicationController
             elsif @comment.unreviewed?
               redirect_to_all_comments(@commentable)
             else
-              redirect_to_comment(@comment, { view_full_work: (params[:view_full_work] == "true"), page: params[:page] })
+              # rubocop:disable Metrics/BlockNesting
+              view_full_work = params[:view_full_work] == "true"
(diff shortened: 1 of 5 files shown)


app/controllers/comments_controller.rb
```diff
@@ -424,7 +424,13 @@ def create
             elsif @comment.unreviewed?
               redirect_to_all_comments(@commentable)
             else
-              redirect_to_comment(@comment, { view_full_work: (params[:view_full_work] == "true"), page: params[:page] })
+              # rubocop:disable Metrics/BlockNesting
+              view_full_work = params[:view_full_work] == "true"
+              parent = view_full_work || current_user&.preference&.view_full_works? ? @comment.ultimate_parent : @comment.parent
```
brianjaustin, line 429: Nitpick: since the second condition is also a case of `view_full_work`, I think it would make more sense to lift the current user check up into that variable.
```suggestion
              view_full_work = params[:view_full_work] == "true" || current_user&.preference&.view_full_works?
              parent = view_full_work ? @comment.ultimate_parent : @comment.parent
```

features/comments_and_kudos/comments_redirect.feature
```diff
@@ -117,3 +117,55 @@ Scenario: Posting top level comment on a middle chapter, while in temporary view
...
+      And a comment "heyo" by "somebody" on the work "Welcome"
+      And I am logged in as a random user
+      And I set my preferences to View Full Work mode by default
+    When I view the work "Welcome"
+      And I post a comment "sup?"
+    Then I should see "sup?"
+      And I should see "heyo"
+      And I should not see "hi!"
+
+  Scenario: Posting a top level comment on the middle chapter of a work, when there is more than one page of comments,
```
brianjaustin, line 152: There are 3 cases for top level comments. Shouldn't we also have at least a few cases for non-top-level ones too?

app/controllers/comments_controller.rb
```diff
@@ -424,7 +424,13 @@ def create
...
+              # if this comment starts a new top-level thread, it will always be on the last page of comments
+              page = parent.comments.reviewed.count.ceildiv(Comment.per_page) if @comment.thread == @comment.id
```
Bilka2: this doesn't account for hidden by admin or spam or deleted, refer to `count_visible_comments_uncached`

Bilka2:
This code overlaps with #5524 in a non-desirable way: The user should stay in chapter by chapter if they were in chapter by chapter, regardless of preference. That PR has other issues that I'll try to dig up from my memory to properly review (I gave up because it turned out more complicated than I had patience for at the time) but generally I think these PRs will need to work around each other

PR #3951, opened 2020-12-31
AO3-6078 Unescape title HTML when notifying user about work deletion

JonathanMeans:
# Pull Request Checklist

* [X] Have you read ["How to write the perfect pull request"](https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/)?
* [X] Have you read the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [X] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)?
* [X] Have you added the [Jira](https://otwarchive.atlassian.net) issue number
  as the *first* thing in your pull request title (e.g. `AO3-1234 Fix thing`)
* [X] Do you have fewer than 5 pull requests already open? If not, please wait
  until they are reviewed and merged before creating new pull requests.

## Issue

https://otwarchive.atlassian.net/browse/AO3-6078 (please fill in issue number and remove this comment)

## Purpose

Show unescaped HTML for Title of fic in pop-up when deleted work

## Testing Instructions

How can the Archive's QA team verify that this is working as you intended?
Automated test was added

(If you have a Jira account with access, please update or comment on the issue
with any new or missing testing instructions.)

## References

Are there other relevant issues/pull requests/mailing list discussions?

## Credit

*What name do you want us to use to credit your work in the [Archive of Our Own's Release Notes](https://archiveofourown.org/admin_posts?tag=1)?*
Jonathan Means

(If you have a Jira account, please include the same name in the "Full name"
field on your Jira profile, so we can assign you the issues you're working on.)

*What pronouns do you prefer (she/her, he/him, zie/hir etc)?*
he/him

(If you do not fill in this section, we will use your GitHub account name and
they/them pronouns to avoid making assumptions about your gender.)

---

The HTML characters are still escaped in the email it sends. Not sure y'all's preference for that behavior, or if you want it included in this work if it is a bug.

Don't have permissions to assign myself stuff in JIRA, so just opening a PR and hopefully no one wastes too much time.

PR #5654, opened 2026-03-22
AO3-6876 Display & < > correctly in prompt titles

nicolacleary:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6876

## Purpose

Render prompt titles with special HTML characters correctly.
e.g. "`My title &amp; &lt; &gt;`" should now be shown correctly as "`My title & < >`"  

Considerations:
* The prompt title is sanitized by sanitize_value before being stored in the DB
  * & < > are all explicitly replaced with their escaped versions before the general HTML sanitising
* We need to call .inner_html on them for them to display correctly in the UI.
  * The value is santised, so there shouldn't be dangerous content (e.g. HTML tags) when we render it*
* This is very similar to how work titles are handled

### HTML tags in titles are NOT stripped:

Because `<` and `>` are replaced, we don't seem to strip `<script>` or similar, but as far as I can see we don't render them as tags:

For the title: `<script>alert("Foo")</script> & &amp; &lt;script&gt;alert("Foo")&lt;/script&gt;`
We store in the DB: `&lt;script&gt;alert("Foo")&lt;/script&gt; &amp; &amp; &lt;script&gt;alert("Foo")&lt;/script&gt;`
The UI shows: `<script>alert("Foo")</script> & & <script>alert("Foo")</script> by sometestuser`
```
    <h4 class="heading">
        &lt;script&gt;alert("Foo")&lt;/script&gt; &amp; &amp; &lt;script&gt;alert("Foo")&lt;/script&gt;
        by sometestuser
    </h4>
```

## References

Similar issues:
* https://github.com/otwcode/otwarchive/pull/878
* https://otwarchive.atlassian.net/browse/AO3-6550

On saving special characters in the DB
* https://github.com/otwcode/otwarchive/pull/3951#discussion_r555532707
* https://otwarchive.atlassian.net/browse/AO3-7107

## Credit

nicolacleary (she/her)


features/step_definitions/challenge_promptmeme_steps.rb
```diff
@@ -340,6 +349,16 @@
   step %{I follow "Prompts ("}
 end
 
+When "I view my claims" do
```
Bilka2: This step can be replaced with the more standard `I go to claimer's claims page`

features/step_definitions/challenge_promptmeme_steps.rb
```diff
@@ -340,6 +349,16 @@
...
+  step %{I follow "My Dashboard"}
+  step %{I follow "Claims"}
+end
+
+When "I view my claims for {string}" do |title|
```
Bilka2: Similarly, I would also expect this to be defined as a path in `features/support/paths.rb` so that it can be a "I go to ..." instead of an extra step

features/step_definitions/challenge_promptmeme_steps.rb
```diff
@@ -277,6 +277,15 @@
     click_button "Submit"
 end
 
+When "I sign up for {string} with combination ESCAPE_CHARS" do |title|
+  step %{I start signing up for "#{title}"}
+  step %{I fill in the 1st field with id matching "title" with "These & < > ' ? are some HTML escape chars"}
+  click_button "Submit"
+  step %{I should see "Sign-up was successfully created"}
+  step %{I should see "These "}
+  step %{I should see " are some HTML escape chars"}
```
Bilka2: I'm curious, why are the chars excluded here?

app/views/prompts/_prompt_form.html.erb
```diff
@@ -36,7 +36,7 @@
             <%= form.label :title, ts("Title:") %>
           <% end %>
         </dt>
-        <dd><%= form.text_field :title %></dd>
+        <dd><%= form.text_field :title, value: form.object.title.to_s %></dd>
```
Bilka2: Could you explain why this change is needed here?

app/models/prompt.rb
```diff
@@ -33,6 +33,13 @@ class Prompt < ApplicationRecord
     where("set_taggings.tag_id = ?", tag.id)
   }
 
+  # VIRTUAL ATTRIBUTES
+
+  # return title.html_safe to overcome escaping done by sanitiser
+  def title
+    self[:title].try(:html_safe)
```
Bilka2, line 40:
```suggestion
    read_attribute(:title).try(:html_safe)
```
to match the other code

Bilka2 (changes requested):
Nice work with this general applicable solution instead of sprinkling `html_safe` everywhere! I just have a few questions and nitpicks

app/models/prompt.rb
nicolacleary: I used it [here](https://github.com/otwcode/otwarchive/pull/5654/changes/f2c221b7e1c198974df9194d5619aa0c73f382ae) but rubocop says `Prefer 'self[:attr]' over 'read_attribute(:attr)'.`

As far as I can see there is already one example of this way as well:
https://github.com/otwcode/otwarchive/blob/master/app/models/collection.rb#L139-L142

Is your preference still to use `read_attribute`?

Bilka2: You're right, we should listen to rubocop

features/step_definitions/challenge_promptmeme_steps.rb
nicolacleary: Artifact from the way I set up the tests during development, I see no reason to exclude them anymore. Updated.

features/step_definitions/challenge_promptmeme_steps.rb
nicolacleary: Thank you, updated

app/views/prompts/_prompt_form.html.erb
nicolacleary: So I'm not _exactly_ sure why this is the case, but it seems like the form text_field bypasses the getter and reads the raw value. There are other models where this doesn't seem to happen (e.g. work) but it's tricky to narrow down why that is without sinking a lot of time into it.

If I add this to the form:
```
<pre>
Inspect: <%= form.object.inspect %>
Just do .title: <%= form.object.title %>
Getting via .attributes: <%= form.object.attributes["title"] %>
Getting via read_attribute: <%= form.object.read_attribute(:title) %>
</pre>
```

I can see:
```
Inspect: #<Request id: 2, collection_id: 11, challenge_signup_id: 2, ... title: "&amp; test", url: nil, ...>
Just do .title: & test
Getting via .attributes: &amp; test
Getting via read_attribute: &amp; test
```

I included the `.to_s` because _I think_ there could be issues with title=nil but I think actually `sanitize_value` turns that into `""` ([link](https://github.com/otwcode/otwarchive/blob/e468bceddad831c3621445d227b9bd2a5e4ca94b/lib/html_cleaner.rb#L47)), so likely this isn't needed.

Is there something better I could do here please?

nicolacleary:
I'll investigate why the works tests are failing - it seems like it only happens when you run the whole of the works tests, not any individual one. Doesn't seem related to my changes but may be related to the merging from main that I did earlier today

app/views/prompts/_prompt_form.html.erb
Bilka2: I dug into it. This is caused by the record attribute being marked as `came_from_user`.

That only happens in PromptsController#edit due to `@index = @challenge_signup.send(@prompt.class.name.downcase.pluralize).index(@prompt)`. For some reason, the `@challenge_signup.send(@prompt.class.name.downcase.pluralize)` causes the prompt's attributes to be marked as coming from the user.

So to fix this I found a in-SQL replacement for that line that no longer loads the requests into Ruby: `@index = @challenge_signup.send(@prompt.class.name.downcase.pluralize).where(id: ... @prompt.id).count`

Based on some poking at `EXPLAIN`s, it should be fine for the database despite the range query. And with this change in the controller the form works without the extra code to set the value. To make sure this index is correct, could you add a check that the edit prompt form displays "Request 1" to the test?

Bilka2:
I suspect it's flaky tests. I've rerun them

app/views/prompts/_prompt_form.html.erb
nicolacleary: I see, thank you for looking into that. I've updated as you requested

Bilka2 (approved):
Thank you very much!

PR #5659, opened 2026-03-23
AO3-6978 Create and fill new table for import urls

luis-pabon-tf:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6978

## Purpose
Step 1 of moving imported URLs to a new table to store all searchable variations.

## Testing Instructions
Besides the steps on the Jira, you can check that the task works by checking `ImportedUrl.last` on the Rails console.

## References

[AO3-6979](https://otwarchive.atlassian.net/browse/AO3-6979) follows up on this one
[AO3-6591](https://otwarchive.atlassian.net/browse/AO3-6591) Is the parent both tickets belong to
## Credit

Luis Pabon (they/he)

[AO3-6979]: https://otwarchive.atlassian.net/browse/AO3-6979?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


db/migrate/20260318013837_create_imported_urls.rb
```diff
@@ -0,0 +1,19 @@
...
+
+      t.string :original, null: false, index: true
+      t.string :minimal, null: false
+      t.string :minimal_no_protocol_no_www, null: false
+      t.string :no_www, null: false
+      t.string :with_www, null: false
+      t.string :with_http, null: false
+      t.string :with_https, null: false
+      t.string :encoded, null: false
+      t.string :decoded, null: false
```
brianjaustin, line 14: We may want indexes on some of these later, but since indexes aren't totally free, I'm fine not doing that for now to see how MariaDB optimises as-is (still better than a query starting with a wildcard 😅 )

spec/models/imported_url_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+    url.save
+
+    expect(url.minimal).to eq(formatter.minimal)
+    expect(url.minimal_no_protocol_no_www).to eq(formatter.minimal_no_protocol_no_www)
+    expect(url.no_www).to eq(formatter.no_www)
+    expect(url.with_www).to eq(formatter.with_www)
+    expect(url.with_http).to eq(formatter.with_http)
+    expect(url.with_https).to eq(formatter.with_https)
+    expect(url.encoded).to eq(formatter.encoded)
+    expect(url.decoded).to eq(formatter.decoded)
```
brianjaustin: To make it easier to see what each of these represents, could you make these be raw strings (so for example `expect(url.minimal).to eq("http://www.trickster.org/llwyden/misc/cracked.html")` instead of the first line)?

brianjaustin (changes requested):
Thanks! One thing for the test. Also, could you add a test for the rake task?

db/migrate/20260318013837_create_imported_urls.rb
luis-pabon-tf: Yeah, I was thinking that it would be better to add indexes once the new tables are usable to see which columns would actually benefit.

brianjaustin (approved):
Thanks!

PR #5663, opened 2026-03-24
AO3-7355 Cache user roles for user menu in main nav

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-7355

## Purpose

Cache the user's relevant roles together with their dashboard counts for the dropdown in the main nav, to massively reduce how often we query the role related tables. 

## Credit

Bilka

### DIFF
--- a/app/helpers/tags_helper.rb
+++ b/app/helpers/tags_helper.rb
@@ -152,10 +152,8 @@ module TagsHelper
   end
 
   def show_wrangling_dashboard
-    # Warning: @current_user_tag_wrangler is cached and not a 1:1 replacement for checking whether the current user is a wrangler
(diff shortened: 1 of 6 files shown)


app/helpers/tags_helper.rb
```diff
@@ -152,7 +152,8 @@ def tag_comment_link(tag)
   end
 
   def show_wrangling_dashboard
-    can_wrangle? &&
+    # Warning: @current_user_tag_wrangler is cached and not a 1:1 replacement for checking whether the current user is a wrangler
```
redsummernight: `Rails/HelperInstanceVariable` probably has a point here. It's easy to catch wrong uses of `@current_user_tag_wrangler` only in views, it's harder if it might be indirectly used through another helper method.

If we move `can_wrangle?` to last, short-circuiting would still cut down on quite a lot of unnecessary checks.

redsummernight (approved):

PR #5668, opened 2026-03-25
AO3-7356 Reduce N+1 queries on series show page

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-7356

## Purpose

Remove N+1 queries related to pseuds and users on the series show page. There are still N+1 queries for chapters, but I couldn't figure out how to remove those (refer to code comment). And this already helps a lot.

Failing test pushed first to show it's testing the right thing.

## Credit

Bilka

### DIFF
--- a/app/controllers/series_controller.rb
+++ b/app/controllers/series_controller.rb
@@ -42,14 +42,14 @@ class SeriesController < ApplicationController
   # GET /series/1
   # GET /series/1.xml
   def show
-    @works = @series.works_in_order.posted.select(&:visible?).paginate(page: params[:page])
+    @works = @series.works_in_order.posted.includes(:pseuds).select(&:visible?).paginate(page: params[:page])
 
     # sets the page title with the data for the series
     if @series.unrevealed?
       @page_subtitle = t(".unrevealed_series")
     else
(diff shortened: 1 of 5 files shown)


app/controllers/series_controller.rb
```diff
@@ -42,14 +42,14 @@ def index
...
-    @works = @series.works_in_order.posted.select(&:visible?).paginate(page: params[:page])
+    @works = @series.works_in_order.posted.includes(:pseuds).select(&:visible?).paginate(page: params[:page])
 
     # sets the page title with the data for the series
     if @series.unrevealed?
       @page_subtitle = t(".unrevealed_series")
     else
       @page_title = get_page_title(@series.fandoms.pluck(:name).join(t("support.array.words_connector")),
-                                   @series.anonymous? ? t(".anonymous") : @series.allpseuds.collect(&:byline).join(t("support.array.words_connector")),
+                                   helpers.text_byline(@series),
```
Bilka2, line 52: This somewhat changes the format, as previously the browser title only grabbed pseuds from the works, but the byline also grabs them from the series itself. But now it matches the list of creators on the actual page, which I think is better. And this code is cached and also more efficient, so it feels like a very obvious solution

redsummernight (approved):

PR #5669, opened 2026-03-26
AO3-7239 Migrate help files: series, translation link, languages help, parent works help and recipients

hunternet93:
## Issue

https://otwarchive.atlassian.net/browse/AO3-7239

## Purpose

Migrates HTML help files to `HelpController`: series, translation link, languages help, parent works help and recipients

## Credit

Hunter Ada Smith (she/they)


app/controllers/help_controller.rb
```diff
@@ -1,31 +1,33 @@
 class HelpController < ApplicationController
+  HELP_ACTIONS = %i[
```
hunternet93, line 2: This change wasn't part of the work item, so I'm happy to revert it if not desired.

I noticed that the controller and routes file had a number of redundant declarations for the migrated help files. While manageable now, there are ~70 help files in total to be migrated, so it will quickly become unwieldy.

This change makes this constant the single source of truth for the available help actions; the controller methods, routes, and redirects are then defined dynamically from it.

config/locales/views/en.yml
```diff
@@ -3155,12 +3181,20 @@ en:
...
+      recipients_help_title: Recipients
+      translations_help_title: Translations
     work_approved_children:
       inspired_by:
         restricted_html: "[Restricted Work] by %{creator_link} (Log in to access.)"
         revealed_html: "%{work_link} by %{creator_link}"
         title: Works inspired by this one
         unrevealed: A work in an unrevealed collection
+    work_form_associations_language:
+      languages_help_title: Languages
```
hunternet93: This particular i18n key gets repeated a number of times. Would it make more sense to de-duplicate it by placing the key at `help.languages_help.title` instead of having a separate copy on each view?

app/controllers/help_controller.rb
Bilka2: Keeping the actions in one array for both here and the routes seems like a good idea, thanks.

However, I'd prefer to not do the redirects like this. As you already discovered, we renamed some files so they need custom code anyway. In fact, this current code is incorrect because the tags help files were also renamed and aren't accounted for here. I also don't want to discourage renaming the files, since the old naming scheme is quite inconsistent. So I think the redirects should be kept as before

Speaking of renaming, since these links are all used on the work form(s), what do you think about prefixing them with something and unifying a bit, like `works_series`, `works_languages` and so on? The current naming with the "help" in some of the files moved here is strange since it results in path names like `help_languages_help_path`

config/locales/views/en.yml
Bilka2: Generally we prefer repetition since context can change how something is translated (e.g. editing multiple works vs editing one) over trying to de-duplicate

app/views/help/choosing_series.html.erb
```diff
@@ -0,0 +1,4 @@
+<h4><%= t(".header") %></h4>
```
Bilka2: Could you switch to `page_heading` to be consistent with the already converted files/pages in general?

config/locales/views/en.yml
Bilka2: That being said, could you please switch all of them to "Languages help" to match what's currently on production

config/locales/views/en.yml
```diff
@@ -3249,12 +3275,20 @@ en:
       no_works: You have no works or drafts to edit.
     skin:
       select: Select work skin
+    standard_form:
+      choosing_series_help_title: Choosing series
+      languages_help_title: Languages
+      parent_works_help_title: Parent works
```
Bilka2:
```suggestion
      parent_works_help_title: Parent works help
```
to match what it's currently

app/views/help/recipients.html.erb
```diff
@@ -0,0 +1,5 @@
+<h4><%= t(".header") %></h4>
+
+<p><%= t(".recipient_names_as_list") %></p>
+
+<p><%= t(".gift_recipients_html", not_strong_bold: tag.strong(t(".not_strong"))) %></p>
```
Bilka2:
```suggestion
<p><%= t(".gift_recipients_html", negation_bold: tag.strong(t(".negation"))) %></p>
```
This or `not_bold` + `not` would be more standard

config/locales/views/en.yml
```diff
@@ -3249,12 +3275,20 @@ en:
...
+      recipients_help_title: Recipients
+      translations_help_title: Translations
```
Bilka2:
```suggestion
      translations_help_title: Translation link
```
to match what it's currently

Bilka2 (changes requested):
Thanks for working on this, I just have a few nitpicks and a comment on your extra change

app/controllers/help_controller.rb
hunternet93: Got it. I reverted the redirects and renamed the help file actions with the `works_` prefix.

config/routes.rb
```diff
@@ -688,6 +670,11 @@
   get "/help/fandom-help.html", to: redirect("/help/tags_fandoms")
   get "/help/rating-help.html", to: redirect("/help/tags_ratings")
   get "/help/warning-help.html", to: redirect("/help/tags_warnings")
+  get "/help/languages-help.html", to: redirect("/help/works_languages")
+  get "/help/parent-works-help.html", to: redirect("/help/works_parents")
+  get "/help/recipients.html", to: redirect("/help/works_recipients")
+  get "/help/choosing-series.html", to: redirect("/help/works_series")
+  get "/help/languages-help.html", to: redirect("/help/works_translation_link")
```
Bilka2, line 677:
```suggestion
  get "/help/translation-link.html", to: redirect("/help/works_translation_link")
```

Bilka2 (changes requested):
Thank you, just one more thing!

config/routes.rb
hunternet93: Whoops, fixed.

Bilka2 (approved):

PR #5675, opened 2026-03-30
AO3-7243 i18n add work to assignment, add collectible to collection and backdating help files

charliewhiskee:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-7243

## Purpose

The three work creation related help HTML files add-work-to-assignment, add-collectible-to-collection, and backdating-help should be moved to erb files rendered by the HelpController and should be internationalized according to the [i18n standards](https://github.com/otwcode/otwarchive/wiki/Internationalization-%28i18n%29-Standards). All link_to_help for these modals should be converted to the more modern link_to_help_modal. 

If any files are renamed in the process of moving them, they should have a redirect from the old name to the new name added to routes.rb.

## Credit

charliewhiskee
they/them


charliewhiskee:
The remaining linting errors about using `ts` are from existing code that I only changed the linebreaks around. Fixing them seems outside the scope of this issue.

app/views/chapters/_chapter_form.html.erb
```diff
@@ -20,7 +20,10 @@
             <%= f.text_field :wip_length, class: "number" %>
           </p>
         </dd>
-        <dt id="managePublicationDate"><%= f.label :published_at, ts("Chapter Publication Date") %>  <%= link_to_help "backdating-help" %></dt>
+        <dt id="managePublicationDate">
+          <%= f.label :published_at, ts("Chapter Publication Date") %>
```
brianjaustin: Since the changes here are on the i18n theme, could you also extract this string (and the other one Rubocop flagged in the collectable form) too?

config/locales/views/en.yml
```diff
@@ -1118,6 +1123,20 @@ en:
...
+    add_work_to_assignment:
+      guidance_html: If you have any open challenge assignments, you can choose one of them here if the story you are posting is meant to fulfill one of them. This will automatically add your story to the challenge collection (or submit it for approval) and also mark your assignment as fulfilled for the challenge moderators.
+      header: Fulfilling Assignments
+      moderated_and_anonymous_collections_html: Note&colon; if the collection you are submitting to is moderated, and you aren't a member, your story will not be added to the collection automatically &#8212; it will have to be approved by one of the collection's moderators. If this is an anonymous and/or unrevealed collection your work will be anonymous and/or hidden as soon as you post it, including while it's waiting for approval. If your work is rejected, it will remain anonymous and/or unrevealed until you edit it to remove it from the collection or until the moderator removes it.
```
brianjaustin: We prefer to minimize raw HTML (including HTML symbols) in translation strings to make things easier on the translators. For the colon, you'll just need to surround this string with quotes to make the YAML valid. For the emdash, I think copying the character itself should work without needing the HTML entity (and then the `_html` suffixes shouldn't be necessary anymore either).

brianjaustin (changes requested):

app/views/help/add_collectible_to_collection.html.erb
```diff
@@ -0,0 +1,9 @@
+<h4><%= t(".header") %></h4>
```
Bilka2: Here and in the other files, could you switch to page_heading to be consistent with the already converted files/pages in general?

config/locales/views/en.yml
```diff
@@ -712,6 +712,8 @@ en:
       heading: Sign-up Summary for %{collection}
       no_tags: Tags were not used in this Challenge, so there is no summary to display here.
   chapters:
+    chapter_form:
+      backdating_help_title: Publication Date Options Help
```
Bilka2: Could you make sure the help titles (this one and the others) match the existing help titles please (it makes things a bit simpler if we don't have to review wording changes as part of this i18n work)

Bilka2 (commented):
Heyo, based on the review I left on #5669 I have some feedback here as well

For the file/action names, since these links are all used on the work form(s), what do you think about prefixing them with something and unifying a bit, like `works_assignment`, `works_backdating` and so on? The PR I linked is an example for some more standardized naming for these help files

charliewhiskee:
Thank you for the feedback! I've implemented the changes you both have requested.

For `add_collectibles_to_collection`, I set the new filename to be `collectibles_add_to_collection` as it's used in both `works/_standard_form.html.erb` and `collectibles/_collectible_form.html.erb`. But let me know if you'd like it to be `works_add_to_collection` (or similar) instead.

brianjaustin (approved):
Thanks! Nothing more from my side. It looks like there are a couple merge conflicts; please give us a shout if you need help resolving them (usually with the locales, you can just use something like the VS Code/IntelliJ conflict resolver and re-run the normalization command to fix any weirdness).

Approving now since the conflict resolution shouldn't be extensive

PR #5679, opened 2026-03-31
AO3-7336 Use Edit Work page instead of Edit Tags page for admin edits to works

not-varram:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-7336

## Purpose

Redirects admins (superadmin, support, policy_and_abuse) from the `/edit_tags` page to the full `/edit` page when editing a work. The admin options link now says "Edit Work" and points to `/edit`, where admins see a dedicated form with tags and language fields (no user subnavigation). This lays the groundwork for expanding the set of fields admins can edit.

## Credit

varram (he/him)


sarken (changes requested):
This is a good start, but at a quick glance, it looks like the form is still using the `update_tags` action instead of the `update` action. Could you change that? (I haven't done a full review because that will require a fair number of changes across many of these files.)

not-varram:
oh yeah, my bad. I have no idea how I missed that 💀. Tysm for pointing it out to me! I'm going to push a fix in a bit :)

app/controllers/works_controller.rb
```diff
@@ -6,15 +6,15 @@ class WorksController < ApplicationController
   # only registered users and NOT admin should be able to create new works
   before_action :load_collection
   before_action :load_owner, only: [:index]
-  before_action :users_only, except: [:index, :show, :navigate, :search, :collected, :edit_tags, :update_tags, :drafts, :share]
+  before_action :users_only, except: [:index, :show, :navigate, :search, :collected, :edit, :update, :edit_tags, :update_tags, :drafts, :share]
```
sarken: Admins no longer need access to edit_tags and update_tags, so they can probably be removed from here, check_ownership, and check_ownership_or_admin.

app/controllers/works_controller.rb
sarken: This can be removed since admins no longer need access to this action.

sarken: This can be removed since admins no longer need access to this page.

app/controllers/works_controller.rb
```diff
@@ -827,18 +837,36 @@ def index_page_title
 
   def log_admin_activity
     if logged_in_as_admin?
-      options = { action: params[:action] }
-
-      if params[:action] == 'update_tags'
-        summary = "Old tags: #{@work.tags.pluck(:name).join(', ')}"
-      end
+      summary = "Old tags: #{@work.tags.pluck(:name).join(', ')}" if params[:action].in?(%w[update update_tags])
```
sarken: Admins should no longer be using `update_tags`, so this should just be looking at `update`.

app/policies/work_policy.rb
```diff
@@ -1,12 +1,18 @@
...
+  # Allow admins to edit works (tags, language, and more in the future).
   # Include support admins due to AO3-4932.
+  def update?
+    user_has_roles?(WORK_TAG_ACCESS)
+  end
+
   def update_tags?
-    user_has_roles?(%w[superadmin policy_and_abuse support])
+    user_has_roles?(WORK_TAG_ACCESS)
   end
```
sarken: This method should be removed since we no longer need access to `update_tags?`

app/policies/work_policy.rb
```diff
@@ -1,12 +1,18 @@
...
 
   alias edit_tags? update_tags?
```
sarken: I think we can remove this since we're using `edit?` instead of `edit_tags?`, and `edit?` is already aliased to `update?` by default in `application_policy.rb`

config/locales/views/en.yml
```diff
@@ -3185,9 +3187,18 @@ en:
...
+      draft_label: " (Draft)"
+      edit_chapter: 'Edit Chapter:'
+      manage_chapters: Manage Chapters
+      orphan_work: Orphan Work
+      page_heading: Edit Work
+      remove_me_as_co_creator: Remove Me As Co-Creator
     edit_tags:
       page_heading:
-        edit_tags_and_language_html: Edit Work Tags and Language for %{title}
         edit_tags_html: Edit Work Tags for %{title}
```
sarken: I think we can change this to just `page_heading_html` now, since we no longer need multiple variations of the heading.

app/policies/work_policy.rb
```diff
@@ -1,12 +1,18 @@
 class WorkPolicy < UserCreationPolicy
+  WORK_TAG_ACCESS = %w[superadmin policy_and_abuse support].freeze
```
sarken: `EDIT_TAG_ROLES` would be more in line with our naming conventions. (We use "access" or "read" for roles that can see info but not change it.)

spec/controllers/works/updating_tags_spec.rb
```diff
@@ -4,28 +4,47 @@
   include LoginMacros
   include RedirectExpectationHelper
 
-  describe "POST #update_tags" do
-    let(:admin) { create(:admin) }
+  describe "GET #edit" do
```
sarken: I think we probably want the tests for the standard actions like `edit` and `update` to be in `default_rails_actions_spec.rb` (`update` is already there) -- this file is just for tag-specific actions like `update_tags`, `edit_tags`, and `preview_tags`.

spec/controllers/works/updating_tags_spec.rb
```diff
@@ -4,28 +4,47 @@
...
     end
 
+    it_behaves_like "an action only authorized admins can access", authorized_roles: %w[superadmin policy_and_abuse support] do
+      let(:work) { create(:work) }
+
+      subject { get :edit, params: { id: work } }
+      let(:success) do
+        expect(response).to render_template(:edit)
+      end
+    end
```
sarken: I think we can condense this a bit more:

```ruby
    context "when logged in as admin" do
      subject { get :edit, params: { id: work } }
      let(:success) { expect(response).to render_template(:edit) }

      it_behaves_like "an action only authorized admins can access", authorized_roles: %w[superadmin policy_and_abuse support]
    end
```

sarken (dismissed):

not-varram:
hello @sarken! I just went through all of your comments, let me now if this looks good now :)

app/controllers/works_controller.rb
```diff
@@ -827,18 +834,36 @@ def index_page_title
...
+      @preview_mode = true
+      render :preview_tags
+    elsif params[:save_button]
+      @work.save
+      flash[:notice] = t("works.update.tags_updated")
+      redirect_to(@work)
+    else
+      @work.posted = true
+      @work.minor_version = @work.minor_version + 1
+      @work.save
```
brianjaustin: To avoid silently failing to save the changes:
```suggestion
      @work.minor_version += 1
      @work.save!
```

app/views/works/_admin_edit_form.html.erb
```diff
@@ -0,0 +1,16 @@
+<%= form_for(@work, html: { id: "work-form", class: "verbose work post" }, url: { action: "update" }) do |f| %>
```
brianjaustin: Could you update this to use `form_with`? That's the preferred helper [as of Rails 5.1](https://guides.rubyonrails.org/form_helpers.html#using-form-tag-and-form-for)

app/views/works/_admin_edit_form.html.erb
```diff
@@ -0,0 +1,16 @@
...
+  <p class="required notice">* <%= t(".required_information") %></p>
+
+  <%= render "work_form_tags", include_blank: false %>
+
+  <fieldset role="article">
+    <legend><%= t("works.associations.title") %></legend>
+    <h3 class="landmark heading"><%= t("works.associations.title") %></h3>
```
brianjaustin: Our translators generally prefer duplication over reuse, because sometimes the context can vary or change later on in ways that may not show up in English

app/views/works/edit.html.erb
```diff
@@ -1,36 +1,42 @@
...
+  <!--subnav-->
+  <ul class="navigation actions" role="menu">
+    <li><%= link_to t(".add_chapter"), new_work_chapter_path(@work) %></li>
+    <% if @chapters %>
+      <li><%= t(".edit_chapter") %>
+        <% for chapter in @chapters %>
+          <% if chapter.posted? %>
+            <%= link_to h(chapter.position), edit_work_chapter_path(@work, chapter) %>
+          <% else %>
+            <%= link_to h(chapter.position) + t(".draft_label"), edit_work_chapter_path(@work, chapter) %>
```
brianjaustin: As part of the i18n, can you update the label to use interpolation? Some languages may want a space between the chapter position and the  `(`, others not
```suggestion
            <%= link_to h(chapter.position) + t(".draft_chapter_position", position: chapter.position), edit_work_chapter_path(@work, chapter) %>
```
(there may be some shuffling with `_html` needed depending on what needs escaping when)

brianjaustin (changes requested):
Taking over the review from Sarken; a few extra things from my side

app/views/works/_admin_edit_form.html.erb
```diff
@@ -1,11 +1,11 @@
-<%= form_for(@work, html: { id: "work-form", class: "verbose work post" }, url: { action: "update" }) do |f| %>
+<%= form_with(model: @work, id: "work-form", class: "verbose work post", local: true, url: { action: "update" }) do |f| %>
```
brianjaustin: Per https://apidock.com/rails/ActionView/Helpers/FormHelper/form_with, the default for Rails 6.1+ (which applies to us) implies this already
```suggestion
<%= form_with(model: @work, id: "work-form", class: "verbose work post", url: { action: "update" }) do |f| %>
```

brianjaustin (dismissed):
Thanks! Just one more thing from me, then should be good to go!

brianjaustin (approved):

PR #5685, opened 2026-04-01
AO3-7182 Hide restricted works from guests on prompts page

luis-pabon-tf:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-7182

## Purpose

Fixes a visibility bug on the meme prompt page. 

## Credit
Luis Pabon (they/he)

### DIFF
--- a/app/models/challenge_claim.rb
+++ b/app/models/challenge_claim.rb
@@ -71,6 +71,9 @@ class ChallengeClaim < ApplicationRecord
              CollectionItem.user_approval_statuses[:approved], CollectionItem.collection_approval_statuses[:approved])
   }
 
+  scope :fulfilled_restricted, -> {
+    fulfilled.where("works.restricted = 1")
+  }
 
   scope :posted, -> { joins(WORKS_JOIN).where("challenge_claims.creation_id IS NOT NULL AND works.posted = 1") }
 
--- a/app/models/challenge_claim.rb
+++ b/app/models/challenge_claim.rb
@@ -71,9 +71,7 @@ class ChallengeClaim < ApplicationRecord
              CollectionItem.user_approval_statuses[:approved], CollectionItem.collection_approval_statuses[:approved])
   }
 
-  scope :fulfilled_restricted, -> {
-    fulfilled.where("works.restricted = 1")
-  }
+  scope :fulfilled_restricted, -> { fulfilled.where("works.restricted = 1") }
 
   scope :posted, -> { joins(WORKS_JOIN).where("challenge_claims.creation_id IS NOT NULL AND works.posted = 1") }
 
--- a/app/models/challenge_claim.rb
+++ b/app/models/challenge_claim.rb
@@ -71,7 +71,8 @@ class ChallengeClaim < ApplicationRecord
              CollectionItem.user_approval_statuses[:approved], CollectionItem.collection_approval_statuses[:approved])
   }
 
-  scope :fulfilled_unrestricted, -> { fulfilled.where("works.restricted = 0") }
+  # TODO: AO3-6024 for making sure we also exclude works hidden by admin
+  scope :work_visible_to_all, -> { where("works.restricted = 0") }
 
   scope :posted, -> { joins(WORKS_JOIN).where("challenge_claims.creation_id IS NOT NULL AND works.posted = 1") }
 
--- a/app/models/challenge_claim.rb
+++ b/app/models/challenge_claim.rb
@@ -71,7 +71,7 @@ class ChallengeClaim < ApplicationRecord
              CollectionItem.user_approval_statuses[:approved], CollectionItem.collection_approval_statuses[:approved])
   }
 
-  scope :fulfilled_restricted, -> { fulfilled.where("works.restricted = 1") }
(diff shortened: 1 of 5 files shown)


sarken (changes requested):
Hi, Luis -- it's great to see you again! 

Thank you for working on this, and I'm sorry we didn't see your comment on the issue asking for clarification about the intended behavior. We'd like the existence of these restricted works completely hidden from guests, the same way we hide restricted works on, e.g., tag work pages. 

What that means is:
1. If the only fills for a prompt are restricted, the "Fulfilled By" listbox should not be displayed.
2. If there are both public and restricted fills for a prompt, the "Fulfilled By" listbox should be present, but only list the public works.

If you could make that change, we'd appreciate it!

luis-pabon-tf:
> If you could make that change, we'd appreciate it!

Thanks for the review! just made the change and am awaiting the automatic testing to verify it.

app/models/challenge_claim.rb
```diff
@@ -71,6 +71,7 @@ def self.order_by_offering_pseud(dir="ASC")
              CollectionItem.user_approval_statuses[:approved], CollectionItem.collection_approval_statuses[:approved])
   }
 
+  scope :fulfilled_unrestricted, -> { fulfilled.where("works.restricted = 0") }
```
sarken: Let's not include the `fulfilled` scope and instead make it something we can chain onto it:

```ruby
  # TODO: AO3-6024 for making sure we also exclude works hidden by admin
  scope :work_visible_to_all, -> { where("works.restricted = 0") }
```

app/models/prompt.rb
```diff
@@ -209,6 +209,10 @@ def fulfilled_claims
     self.request_claims.fulfilled
   end
 
+  def fulfilled_unrestricted_claims
+    self.request_claims.fulfilled_unrestricted
+  end
```
sarken: We can update this to use our chained scopes and also give it a name that's a little clearer (right now, it sounds like the claim rather than the fill is unrestricted):
```ruby  
def fulfilled_claims_visible_to_all
  self.request_claims.fulfilled.work_visible_to_all
end
```

app/views/prompts/_prompt_blurb.html.erb
```diff
@@ -80,13 +80,15 @@
     <% end %>
 
     <% # if prompt has been fulfilled list works %>
-    <% unless prompt.unfulfilled? %>
+    <% unless prompt.unfulfilled? || (guest? && prompt.fulfilled_unrestricted_claims.count < 1) %>
```
sarken: I think we can avoid a count here and do `prompt.fulfilled_claims_visible_to_all.empty?`

sarken (changes requested):
If you could also add a feature test, that would be great -- thank you!

sarken (commented):
The code looks good! All we need now is a test.

luis-pabon-tf:
> The code looks good! All we need now is a test.

Thanks for checking it! I just finished testing the new testand am awaiting the Github version of it to run.

features/prompt_memes_b/challenge_promptmeme_posting_fills.feature
```diff
@@ -593,3 +593,21 @@ Feature: Prompt Meme Challenge
...
+  Given I have Battle 12 prompt meme fully set up
+  When I am logged in as "myname1"
+  When I sign up for Battle 12
+  When I am logged in as "myname2"
+    And I claim a prompt from "Battle 12"
+    And I start to fulfill my claim with "Restricted Fill"
+    And I lock the work
+    And I press "Post"
+    And I go to "Battle 12" collection's page
+    And I follow "Prompts ("
```
sarken: I think you can use `When I view prompts for "Battle 12"` in place of these two steps.

features/prompt_memes_b/challenge_promptmeme_posting_fills.feature
```diff
@@ -593,3 +593,21 @@ Feature: Prompt Meme Challenge
...
+  Then I should see "Fulfilled By"
```
sarken, line 608: Let's be extra thorough and also check for the work with `And I should see "Restricted Fill"` (and do the opposite at the end, with `And I should not see "Restricted Fill"`)

features/prompt_memes_b/challenge_promptmeme_posting_fills.feature
```diff
@@ -593,3 +593,21 @@ Feature: Prompt Meme Challenge
...
+  Scenario: When a prompt is filled with a restricted work, the work should only be visible to logged-in users
+
+  Given I have Battle 12 prompt meme fully set up
+  When I am logged in as "myname1"
+  When I sign up for Battle 12
+  When I am logged in as "myname2"
+    And I claim a prompt from "Battle 12"
+    And I start to fulfill my claim with "Restricted Fill"
+    And I lock the work
+    And I press "Post"
```
sarken: These setup steps can all be `And` steps under the `Given`

sarken (changes requested):
Just a couple nitpicks, but otherwise it looks good-- thanks!

sarken (approved):
Thank you!

luis-pabon-tf:
Accidentally closed this.

PR #5694, opened 2026-04-03
AO3-6218 Allow certain admins to access all collection and challenge pages usually reserved for owners

not-varram:
# Pull Request Checklist

## Issue
  https://otwarchive.atlassian.net/browse/AO3-6218

## Additional Notes
Failed on QA. Made necessary changes as per comments on jira.

## Purpose
Allow admins with `support`, `policy_and_abuse`, or `superadmin` roles to access owner/maintainer collection and challenge pages for viewing/troubleshooting, without granting write permissions.

Implemented via shared controller helpers and action-level filter changes so read routes are opened for those roles while create/update/destroy paths remain owner/maintainer-only.

## Testing Instructions
Automated coverage was added/updated for the affected controllers to verify:
- allowed admin roles can access read pages
- other admin roles cannot
- privileged admins still cannot perform protected updates

(Functional QA flow is already documented in the Jira ticket.)

## Credit
varram (he/him)

### DIFF
--- a/features/collections/collection_admin_access.feature
+++ b/features/collections/collection_admin_access.feature
@@ -1,0 +1,236 @@
+@collection
+
+Feature: Admin access to collection pages
+  Admins with support, policy_and_abuse, or superadmin roles should be
+  able to view all collection and challenge pages that owners can access,
+  but should not be able to make changes.
+
+  # ============================================================
+  # General Collection Pages
+  # ============================================================
+
+  Scenario: Support admin can access the Manage Items page with no items
+    Given I have a collection "Owned Collection"
+      And I am logged in as a "support" admin
+    When I view the awaiting collection approval collection items page for "Owned Collection"
+    Then I should see "Items in"
+      And I should not see "You don't have permission"
+
+  Scenario: Support admin can access the Manage Items page with items
+    Given I am logged in as "author"
+      And I set my preferences to allow collection invitations
+      And I post the work "Admin Viewable Work"
+      And I have a collection "Items Collection"
+      And I am logged in as "moderator"
+      And I invite the work "Admin Viewable Work" to the collection "Items Collection"
+      And I am logged in as "author"
+      And "author" accepts the invitation for their work in the collection "Items Collection"
+      And I press "Submit"
+    When I am logged in as a "support" admin
+      And I view the approved collection items page for "Items Collection"
+    Then I should see "Admin Viewable Work"
+
+  Scenario: Support admin can access all Manage Items status pages
+    Given I am logged in as "author"
+      And I set my preferences to allow collection invitations
+      And I post the work "Test Work"
+      And I have a moderated collection "Moderated Collection"
+      And I am logged in as "moderator"
+      And I invite the work "Test Work" to the collection "Moderated Collection"
+    When I am logged in as a "support" admin
+      And I view the awaiting collection approval collection items page for "Moderated Collection"
+    Then I should not see "You don't have permission"
+    When I view the awaiting user approval collection items page for "Moderated Collection"
+    Then I should not see "You don't have permission"
+    When I view the rejected by collection collection items page for "Moderated Collection"
+    Then I should not see "You don't have permission"
+    When I view the rejected by user collection items page for "Moderated Collection"
+    Then I should not see "You don't have permission"
+
+  Scenario: Support admin can access Collection Settings page
+    Given I have a collection "Settings Collection"
+      And I am logged in as a "support" admin
+    When I go to "Settings Collection" collection edit page
+    Then I should see "Edit Collection"
+
+  Scenario: Support admin cannot update collection settings
+    Given I have a collection "No Edit Collection"
+      And I am logged in as a "support" admin
+    When I go to "No Edit Collection" collection edit page
+      And I press "Update"
+    Then I should see "Please log out of your admin account first!"
+
+  Scenario: Support admin can access Membership page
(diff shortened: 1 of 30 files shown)


features/collections/collection_admin_access.feature
```diff
@@ -0,0 +1,236 @@
...
+  Scenario: Support admin can access all Manage Items status pages
+    Given I am logged in as "author"
+      And I set my preferences to allow collection invitations
+      And I post the work "Test Work"
+      And I have a moderated collection "Moderated Collection"
+      And I am logged in as "moderator"
+      And I invite the work "Test Work" to the collection "Moderated Collection"
+    When I am logged in as a "support" admin
+      And I view the awaiting collection approval collection items page for "Moderated Collection"
+    Then I should not see "You don't have permission"
```
Bilka2: I tried this test locally without the changes in this PR and found it passing because the collection search error overrides this flash message. To get the flash message to show up, `all indexing jobs have been run` has to be done after creating the collection to get rid of the search error.

But I'd prefer to try to check for something else here, since that error also wouldn't show up if Support admins were forbidden from viewing the page like tag wrangling admins are down below. Maybe add the paths to `paths.rb` and then check with `I should be on ...`? Or check text on the page, though I'm guessing you would've gone for that if it had been viable...

features/collections/collection_admin_access.feature
```diff
@@ -0,0 +1,236 @@
...
+    When I go to the "PAB Collection" participants page
+    Then I should see "Members of"
+
+  Scenario: Superadmin can access collection pages
+    Given I have a collection "Super Collection"
+      And I am logged in as a super admin
+    When I go to "Super Collection" collection edit page
+    Then I should see "Edit Collection"
+    When I go to the "Super Collection" participants page
+    Then I should see "Members of"
```
Bilka2, line 91: This is fine to keep here since it's only two roles, but for the future: You can work with Scenario Outline and make the role a parameter, some other tests do that for example. Though I think in this case we don't have to exhaustively check every role since they behave the same and this is mostly about making sure the views don't error, so the way you organized this here is just fine

features/collections/collection_admin_access.feature
```diff
@@ -0,0 +1,236 @@
...
+      And I should see "signer1"
+
+  Scenario: Support admin can access the Requests Summary
+    Given I have created the gift exchange "GE Requests"
+      And I open signups for "GE Requests"
+      And I am logged in as "signer1"
+      And I sign up for "GE Requests" with combination A
+    When I am logged in as a "support" admin
+      And I go to the "GE Requests" requests page
+    Then I should not see "You are not allowed to view the requests summary!"
```
Bilka2: Same issue with this test not failing without the changes due to the error not showing up as above

features/collections/collection_admin_access.feature
```diff
@@ -0,0 +1,236 @@
...
+
+  Scenario: Support admin can access the Matching page
+    Given I have created the gift exchange "GE Matching"
+      And I open signups for "GE Matching"
+    When I am logged in as a "support" admin
+      And I go to "GE Matching" gift exchange matching page
+    Then I should not see "Sorry, you don't have permission"
```
Bilka2: And same here too :(

features/collections/collection_admin_access.feature
```diff
@@ -0,0 +1,236 @@
...
+    Given I have created the gift exchange "GE Admin Test"
+      And I open signups for "GE Admin Test"
+      And I am logged in as "signer1"
+      And I sign up for "GE Admin Test" with combination A
+    When I am logged in as a "support" admin
+      And I go to the "GE Admin Test" signups page
+    Then I should see "Sign-ups"
+      And I should see "signer1"
+
+  Scenario: Support admin can access an individual sign-up
```
Bilka2: (this organization is nice for review, but I also wouldn't mind if you combine this test and the previous one)

features/collections/collection_admin_access.feature
```diff
@@ -0,0 +1,236 @@
...
+      And I sign up for "GE No Match" with combination A
+      And I am logged in as "signer2"
+      And I sign up for "GE No Match" with combination B
+      And I close signups for "GE No Match"
+    When I am logged in as a "support" admin
+      And I go to "GE No Match" gift exchange matching page
+      And I press "Generate Potential Matches"
+    Then I should see "Please log out of your admin account first!"
+
+  Scenario: Support admin can access Assignments page
```
Bilka2, line 145: This is missing the test for accessing an individual assignment

features/collections/collection_admin_access.feature
```diff
@@ -0,0 +1,236 @@
...
+  # ============================================================
+
+  Scenario: Support admin can access prompt meme Unposted Claims
+    Given I have Battle 12 prompt meme fully set up
+      And I open signups for "Battle 12"
+      And I am logged in as "signer1"
+      And I sign up for Battle 12 with combination A
+    When I am logged in as a "support" admin
+      And I go to the "Battle 12" claims page
+    Then I should not see "Sorry, you don't have permission"
```
Bilka2: The error message problem again :(

Bilka2 (changes requested):
Thanks for adding the tests, I'm finally here for the updated review

features/collections/collection_admin_access.feature
not-varram: My bad! 😅

I've fixed this now :)

Bilka2 (approved):
Thank you!

PR #5695, opened 2026-04-03
AO3-7357 i18n SearchHelper#search_header

not-varram:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-7357

## Purpose
Rewrote `SearchHelper#search_header` to fix three i18n issues: pluralisation now uses explicit Rails `one`/`other` keys per item type, string concatenation isreplaced with full-phrase translation keys so locales can define their own word order, and each item type now has its own key tree instead of a word being inserted into a shared string.

  The method signature was updated to use keyword arguments; all call sites in views were updated accordingly.
  
## Testing Instructions
Automated coverage was added in `spec/helpers/search_helper_spec.rb`.

(Functional QA flow is documented in the Jira ticket.)

## Credit

varram (he/him)

### DIFF
--- a/config/locales/helpers/en.yml
+++ b/config/locales/helpers/en.yml
@@ -44,6 +44,122 @@ en:
   qr_code_helper:
     qr_code_as_svg:
       default_label: QR code
+  search_helper:
+    search_header:
+      default_html: "%{results}"
+      items:
+        bookmark:
+          count:
+            one: "%{count} Bookmark"
+            other: "%{count} Bookmarks"
+          range:
+            one: "%{start_number} - %{end_number} of %{total} Bookmark"
+            other: "%{start_number} - %{end_number} of %{total} Bookmarks"
+          recent:
+            one: Recent Bookmark
+            other: Recent Bookmarks
+        bookmarked_item:
+          count:
+            one: "%{count} Bookmarked Item"
+            other: "%{count} Bookmarked Items"
+          range:
+            one: "%{start_number} - %{end_number} of %{total} Bookmarked Item"
+            other: "%{start_number} - %{end_number} of %{total} Bookmarked Items"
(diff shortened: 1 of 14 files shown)


config/locales/helpers/en.yml
```diff
@@ -44,6 +44,122 @@ en:
...
+        collection_html: in %{collection_link}
+        combined_html: "%{first_scope} %{second_scope}"
+        fandom_html: in %{fandom_link}
+        language: in %{language}
+        pseud: by %{byline}
+        tag_html: in %{tag_link}
+        user: by %{username}
+      with_query_and_scope_html: "%{results} found %{location}"
+      with_query_html: "%{results} found"
+      with_scope_html: "%{results} %{location}"
```
Bilka2: This concatenation is another version of the second and third bullet point in the issue: Some locales may want to alter the first part (e.g. the "found" verb) based on the second part. I've talked to Translation, and they confirm this. So please make these translatable as the full sentences, so no disconnection between the item kind and it's location

config/locales/helpers/en.yml
```diff
@@ -44,6 +44,122 @@ en:
...
+            other: Recent Works
+      scope:
+        collection_html: in %{collection_link}
+        combined_html: "%{first_scope} %{second_scope}"
+        fandom_html: in %{fandom_link}
+        language: in %{language}
+        pseud: by %{byline}
+        tag_html: in %{tag_link}
+        user: by %{username}
+      with_query_and_scope_html: "%{results} found %{location}"
```
Bilka2: please use something other than "scope" here, since it's a reserved term/argument in the i18n gem/`t` method

Bilka2 (changes requested):

PR #5703, opened 2026-04-04
AO3-4693 Fix collection deletion for collections with a challenge

ayasayadi1:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-4693

## Purpose

This PR fixes deletion for collections with an associated challenge (gift exchange or prompt meme) by destroying the challenge first before the collection itself.

## Credit

Aya Sayadi

### DIFF
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -40,17 +40,6 @@ class Collection < ApplicationRecord
   has_many :assignments, class_name: "ChallengeAssignment", dependent: :destroy
   has_many :claims, class_name: "ChallengeClaim", dependent: :destroy
 
-  # We need to get rid of all of these if the challenge is destroyed
-  after_save :clean_up_challenge
-  def clean_up_challenge
-    return if self.challenge_id
-
-    assignments.each(&:destroy)
-    potential_matches.each(&:destroy)
-    signups.each(&:destroy)
-    prompts.each(&:destroy)
-  end
-
   has_many :collection_items, dependent: :destroy
   accepts_nested_attributes_for :collection_items, allow_destroy: true
   has_many :approved_collection_items, -> { approved_by_both }, class_name: "CollectionItem"
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -31,7 +31,13 @@ class Collection < ApplicationRecord
     self.collection_profile = CollectionProfile.new unless self.collection_profile
   end
 
-  belongs_to :challenge, dependent: :destroy, polymorphic: true
+  belongs_to :challenge, polymorphic: true
+
+  before_destroy :destroy_challenge
+  def destroy_challenge
+    challenge.destroy if challenge
+  end
+
   has_many :prompts, dependent: :destroy
 
   has_many :signups, class_name: "ChallengeSignup", dependent: :destroy
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -35,7 +35,7 @@ class Collection < ApplicationRecord
 
   before_destroy :destroy_challenge
   def destroy_challenge
-    challenge.destroy if challenge
+    challenge&.destroy
   end
 
   has_many :prompts, dependent: :destroy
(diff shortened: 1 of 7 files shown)


app/models/collection.rb
```diff
@@ -31,7 +31,13 @@ def ensure_associated
     self.collection_profile = CollectionProfile.new unless self.collection_profile
   end
 
-  belongs_to :challenge, dependent: :destroy, polymorphic: true
+  belongs_to :challenge, polymorphic: true
```
redsummernight: I think we can fix this just with callbacks. In collection, you can keep the association as-is:

https://github.com/otwcode/otwarchive/blob/e468bceddad831c3621445d227b9bd2a5e4ca94b/app/models/collection.rb#L34

In both challenge models, you can update the association to use `:nullify`.
```rb
has_one :collection, as: :challenge, dependent: :nullify
```
(which means you can fully remove all references of `clear_challenge_references`).

Rails 5+ should allow this bi-directional association to work (https://github.com/rails/rails/pull/21350).

spec/models/collection_spec.rb
```diff
@@ -692,4 +692,21 @@
       end
     end
   end
+
+  describe "#destroy" do
+    it "destroys the collection and its gift exchange challenge" do
+      challenge = create(:gift_exchange)
+      collection = create(:collection, challenge: challenge)
+      collection.destroy
```
redsummernight: You should use `destroy!` in specs, that way if there are errors you can easily see it in test logs.

spec/models/collection_spec.rb
redsummernight, line 1: Can you also update gift_exchange_spec.rb and prompt_meme_spec.rb to test that a collection's `challenge_id` and `challenge_type` fields are nulled when the associated challenge is deleted?

redsummernight (changes requested):

app/models/collection.rb
ayasayadi1: Hi, thank you for reviewing!

`dependent: :nullify` clears the FK. But since we now bypass ActiveRecord callbacks, `clean_up_challenge` never fires. As a result, signups, assignments, prompts etc survive the deletion. Some GE and PM feature tests caught this.

Previously, the flow was:
1. `before_destroy :clear_challenge_references `-> challenge references cleared -> `collection.save!` 
2. `after_save :clean_up_challenge` -> associated records deleted
4. challenge deleted

Now it's:
1. challenge deleted
2. `:nullify `-> challenge references cleared
3. associated records remain

I worked around this by moving `clean_up_challenge` into `ChallengeCore` and calling it via `before_destroy`, which restores the original behaviour

That said, I think this workflow is still risky: if the challenge deletion fails after `before_destroy` runs, we've already lost the associated records. Please lmk what you think :)

spec/models/gift_exchange_spec.rb
```diff
@@ -43,6 +43,23 @@
     end
   end
 
+  describe "#destroy" do
+    let!(:challenge) { create(:gift_exchange) }
+    let!(:collection) { create(:collection, challenge: challenge) }
+
+    it "does not destroy the collection" do
```
redsummernight: You can skip this test, since `collection.reload` in the second test will throw an error if the collection is destroyed.

Same for the test in prompt_meme_spec.rb.

app/models/collection.rb
redsummernight: > if the challenge deletion fails after before_destroy runs, we've already lost the associated records

Since it's not riskier than the previous callback flow, which also deletes the associated records before the challenge itself, we should be fine.

redsummernight (approved):
Looks good, thank you!

I'll let a second reviewer check our callback understanding, just in case.

spec/models/gift_exchange_spec.rb
ayasayadi1: Done!

sarken (approved):

PR #5709, opened 2026-04-05
AO3-6142 Make prompt meme claims show back up after deleting fill...

techlingyi:
...on the user's Claims page and the Post New Work form

# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6142

## Purpose

Copies the work model's after_destroy method for challenge assignments so it also destroys the work's challenge claims after the user deletes the work. This makes the claim reappear as unfulfilled on the user's claims page and reappear as an option in the Post New Work form.

## Notes

Ran into issues when trying to manually test posting two fills to one claim under the same username and then deleting one of them. After posting both fills, both the prompts page and user's claims page show only one fill, which then disappears upon deletion; the second fill doesn't show up afterwards. I think this falls under the scope of https://otwarchive.atlassian.net/browse/AO3-2407 - deleting one fill functions as expected when the two fills are from different usernames.

## Credit

Ling-Yi (any pronouns)


features/prompt_memes_b/challenge_promptmeme_posting_fills.feature
```diff
@@ -140,6 +140,34 @@ Feature: Prompt Meme Challenge
...
+  Then I should see "Fulfilled Story"
+  When I am on myname4's user page
+    And I follow "Claims"
+  Then I should not see "Request"
+    And I should not see "random SGA love"
+  When I delete the work "Fulfilled Story"
+  When I am on myname4's user page
+  Then I should see "Claims (1)"
+  When I follow "Claims"
+		And I follow "Fulfilled Claims"
```
Bilka2: some tabs happened here
```suggestion
    And I follow "Fulfilled Claims"
```

features/prompt_memes_b/challenge_promptmeme_posting_fills.feature
```diff
@@ -140,6 +140,34 @@ Feature: Prompt Meme Challenge
...
+  Given I have Battle 12 prompt meme fully set up
+  When I am logged in as "myname1"
+  When I sign up for Battle 12 with combination B
+    And I am logged in as "myname4"
+    And I claim a prompt from "Battle 12"
+  When I fulfill my claim
+  When I am on myname4's user page
+  Then I should see "Claims (0)"
+  When I follow "Claims"
+		And I follow "Fulfilled Claims"
```
Bilka2: some tabs happened here
```suggestion
    And I follow "Fulfilled Claims"
```

app/models/work.rb
```diff
@@ -379,7 +379,18 @@ def destroy_chapters_in_reverse
...
+    end
+  end
+
+  after_destroy :clean_up_claims
+  def clean_up_claims
+    self.challenge_claims.each do |a|
+      a.creation = nil
+      a.save!
+    end
   end
```
Bilka2: I believe you can achieve this more cleanly by setting `dependent: :nullify` on the `has_many :challenge_claims` association instead of adding this manual callback

Bilka2 (changes requested):
Just some nitpicks

techlingyi:
Thank you @Bilka2 ! I have two questions - should I also switch to the `dependent: :nullify` option for `has_many :challenge_assignments` or is that too out of scope, and should I merge the current master branch before pushing again or is that not necessary?

Bilka2:
> should I also switch to the dependent: :nullify option for has_many :challenge_assignments

It looks like the code isn't covered by tests (at least the test that I would expect to fail when I remove the method doesn't fail), so I'd say it's out of scope since it would involve adding/modifying a test

> should I merge the current master branch before pushing again or is that not necessary?

you don't have to since there are no merge conflicts, but it's nice if you do

Bilka2 (approved):
Thank you!

PR #5713, opened 2026-04-07
AO3-7325 Don't attach HTML of work to Abuse ticket when reporting bookmarks

ScottVenkataraman:
# Pull Request Checklist

## Issue
https://otwarchive.atlassian.net/browse/AO3-7325

## Purpose
What does this PR do?
This PR adds exclusion criteria, to prevent attaching HTML of the work to an abuse ticket when reporting a bookmark. 

## Testing Instructions

I'm not sure how to test in development with Zoho ticket, open to guidance on how to test that functionality. 

## Credit

Scott Venkataraman he/him

### DIFF
--- a/app/models/abuse_report.rb
+++ b/app/models/abuse_report.rb
@@ -182,10 +182,11 @@ class AbuseReport < ApplicationRecord
     end
   end
 
-  # ID of the reported work, unless the report is about comment(s) on the work
+  # ID of the reported work, unless the report is about comment(s) or bookmark(s) on the work
   def reported_work_id
     comments = url[%r{/comments/}, 0]
-    url[%r{/works/(\d+)}, 1] if comments.nil?
+    bookmarks = url[%r{/bookmarks}, 0]
+    url[%r{/works/(\d+)}, 1] if comments.nil? && bookmarks.nil?
   end
 
   # ID of the reported comment
--- a/app/models/abuse_report.rb
+++ b/app/models/abuse_report.rb
@@ -185,7 +185,7 @@ class AbuseReport < ApplicationRecord
   # ID of the reported work, unless the report is about comment(s) or bookmark(s) on the work
   def reported_work_id
     comments = url[%r{/comments/}, 0]
-    bookmarks = url[%r{/bookmarks}, 0]
+    bookmarks = url[%r{/bookmarks/}, 0]
     url[%r{/works/(\d+)}, 1] if comments.nil? && bookmarks.nil?
   end
 
--- a/spec/models/abuse_report_spec.rb
+++ b/spec/models/abuse_report_spec.rb
@@ -693,7 +693,7 @@ describe AbuseReport do
 
     context "for a work URL with bookmarks" do
       it "returns nil" do
-        allow(subject).to receive(:url).and_return("http://archiveofourown.org/works/123/bookmarks")
+        allow(subject).to receive(:url).and_return("http://archiveofourown.org/works/123/bookmarks/")
         expect(subject.reported_work_id).to be_nil
       end
     end
--- a/spec/models/abuse_report_spec.rb
+++ b/spec/models/abuse_report_spec.rb
@@ -669,6 +669,43 @@ describe AbuseReport do
     end
   end
 
+  describe "#reported_work_id" do
+    context "for a plain work URL" do
+      it "returns the work id" do
+        allow(subject).to receive(:url).and_return("http://archiveofourown.org/works/123/")
+        expect(subject.reported_work_id).to eq("123")
+      end
+    end
+
+    context "for a work URL with bookmark-form anchor" do
+      it "returns the work id" do
+        allow(subject).to receive(:url).and_return("http://archiveofourown.org/works/123#bookmark-form")
+        expect(subject.reported_work_id).to eq("123")
+      end
+    end
+
+    context "for a work URL with comments" do
+      it "returns nil" do
+        allow(subject).to receive(:url).and_return("http://archiveofourown.org/works/123/comments/")
+        expect(subject.reported_work_id).to be_nil
+      end
+    end
+
+    context "for a work URL with bookmarks" do
+      it "returns nil" do
+        allow(subject).to receive(:url).and_return("http://archiveofourown.org/works/123/bookmarks")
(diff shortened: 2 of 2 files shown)


brianjaustin (changes requested):
Could you add tests for this method (including the comments url check too while you're there)? Thanks!

app/models/abuse_report.rb
```diff
@@ -182,10 +182,11 @@ def creator_ids
     end
   end
 
-  # ID of the reported work, unless the report is about comment(s) on the work
+  # ID of the reported work, unless the report is about comment(s) or bookmark(s) on the work
   def reported_work_id
     comments = url[%r{/comments/}, 0]
-    url[%r{/works/(\d+)}, 1] if comments.nil?
+    bookmarks = url[%r{/bookmarks}, 0]
```
Bilka2: Could you also add the slash after `bookmarks` here to make sure we don't accidentally capture unrelated URLs, similar to how it's done for comments? It will still apply for pages like ao3.org/works/000/bookmarks because that gets the trailing slash added in `standardize_url`

spec/models/abuse_report_spec.rb
```diff
@@ -669,6 +669,43 @@ def queue_adapter_for_test
...
+    context "for a work URL with bookmark-form anchor" do
+      it "returns the work id" do
+        allow(subject).to receive(:url).and_return("http://archiveofourown.org/works/123#bookmark-form")
+        expect(subject.reported_work_id).to eq("123")
+      end
+    end
+
+    context "for a work URL with comments" do
+      it "returns nil" do
+        allow(subject).to receive(:url).and_return("http://archiveofourown.org/works/123/comments/")
```
redsummernight: Does setting the field without mocks work?
```rb
subject.url = "http://archiveofourown.org/works/123/comments/"
```
If it does, I'd prefer that since it's simpler (you wouldn't need to change all the existing tests though, just the newly added ones).

spec/models/abuse_report_spec.rb
```diff
@@ -669,6 +669,44 @@ def queue_adapter_for_test
...
+    context "for a work URL with comments" do
+      it "returns nil" do
+        subject.url = "http://archiveofourown.org/works/123/comments/"
+        expect(subject.reported_work_id).to be_nil
+      end
+    end
+
+    context "for a work URL with bookmarks" do
+      it "returns nil" do
+        subject.url = "http://archiveofourown.org/works/123/comments/"
```
brianjaustin: Looks like a copy-paste error?
```suggestion
        subject.url = "http://archiveofourown.org/works/123/bookmarks/"
```

spec/models/abuse_report_spec.rb
```diff
@@ -669,6 +669,44 @@ def queue_adapter_for_test
...
+        expect(subject.reported_work_id).to be_nil
+      end
+    end
+
+    context "for a work URL with a specific bookmark" do
+      it "returns nil" do
+        subject.url = "http://archiveofourown.org/works/123/comments/"
```
brianjaustin: Same here

spec/models/abuse_report_spec.rb
```diff
@@ -669,6 +669,44 @@ def queue_adapter_for_test
     end
   end
 
+  describe "#reported_work_id" do
+    context "for a plain work URL" do
+      it "returns the work id" do
+        # allow(subject).to receive(:url).and_return("http://archiveofourown.org/works/123/")
```
brianjaustin:
```suggestion
```

brianjaustin (changes requested):

spec/models/abuse_report_spec.rb
ScottVenkataraman: Thanks for catching these!

spec/models/abuse_report_spec.rb
```diff
@@ -693,7 +693,7 @@ def queue_adapter_for_test
 
     context "for a work URL with bookmarks" do
       it "returns nil" do
-        subject.url = "http://archiveofourown.org/works/123/bookmarks"
+        subject.url = "http://archiveofourown.org/works/123/bookmarks/"
```
brianjaustin, line 696: 🤔 I would expect this to work with or without the trailing slash

ScottVenkataraman: My understanding (which may be wrong) is that the trailing slash gets added in production, because of `standardize_url` running `before_validation`. But that's not happening in testing, hence the trailing slash for both bookmarks and comments. Please correct me if that's wrong.

brianjaustin: It looks like that may be a side-effect, but it's not documented in https://github.com/ScottVenkataraman/otwarchive/blob/b88545131d0a775fa9c18e4d68cca5a424619fcb/app/models/abuse_report.rb#L58. To make the code defensive to other changes, IMO It would be better to support either case (with or without the `/`)

Bilka2: it's part of clean_url and documented there

ScottVenkataraman: So with that said, where does that leave us? Should I leave as is? 

While `clean_url` makes this redundant in production, I could update the `reported_work_id` method regex to account for either case:
```
# ID of the reported work, unless the report is about comment(s) or bookmark(s) on the work
  def reported_work_id
    comments = url[%r{/comments/}, 0]
    bookmarks = url[%r{/bookmarks(/|$)}, 0]
    url[%r{/works/(\d+)}, 1] if comments.nil? && bookmarks.nil?
  end
```

Then add a second test to `abuse_report_spec.rb`, so one accounts for trailing slash and one doesn't:
```
context "for a work URL with bookmarks" do
  it "returns nil" do
    subject.url = "http://archiveofourown.org/works/123/bookmarks"
    expect(subject.reported_work_id).to be_nil
  end
end

context "for a work URL with bookmarks and trailing slash" do
  it "returns nil" do
    subject.url = "http://archiveofourown.org/works/123/bookmarks/"
    expect(subject.reported_work_id).to be_nil
  end
end
```

Thanks!

brianjaustin: Let's leave it as is for now (so no new test or change to the regex)

brianjaustin (approved):

PR #5719, opened 2026-04-10
AO3-6531 Fix comment decorator 500 error for orphaned comments

pmonfort:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6531

## Purpose

Fixes a 500 error in `CommentDecorator.from_thread_ids` caused by orphaned comments — replies whose parent comment was physically removed from
the database (via `.delete`) rather than soft-deleted.

When building the comment tree, the decorator looks up each reply's parent in a hash by `commentable_id`. If the parent was physically deleted,
the lookup returns `nil`, causing a `NoMethodError`. The fix adds a nil check so that orphaned replies are silently skipped when building the
tree, rather than crashing the page.

Additionally, this PR includes a rake task `After:create_placeholders_for_orphaned_comments` that finds all orphaned comments in the
database and creates soft-deleted placeholder comments in their place. The task:
- Uses a LEFT JOIN to efficiently find comments whose parent no longer exists
- Processes orphans top-down using the nested set (`threaded_left`) to handle cases where multiple comments were deleted in the same branch
- Creates each placeholder as a child of the closest existing ancestor, preserving the tree structure

## Testing Instructions

1. Create a comment thread with at least 3 levels (root → reply → nested reply)
2. Physically delete the middle comment from the database (e.g., `Comment.find(id).delete`)
3. Visit the page that displays the thread
4. Verify no 500 error occurs and the root comment still displays
5. Run the rake task: `bundle exec rake After:create_placeholders_for_orphaned_comments`
6. Verify a soft-deleted placeholder was created for the deleted comment
7. Visit the thread again and verify the placeholder appears as "[deleted comment]"

## Credit

Pablo Monfort (he/him)

### DIFF
--- a/lib/tasks/after_tasks.rake
+++ b/lib/tasks/after_tasks.rake
@@ -657,26 +657,26 @@ namespace :After do
 
   desc "Create placeholder comments for orphaned replies whose parent was physically deleted"
   task(create_placeholders_for_orphaned_comments: :environment) do
-    orphaned_commentable_ids = Comment
+    missing_commentable_ids = Comment
       .joins("LEFT JOIN comments AS commentable_comments ON comments.commentable_id = commentable_comments.id")
       .where(commentable_type: "Comment")
       .where(commentable_comments: { id: nil })
       .distinct
       .pluck(:commentable_id)
 
-    if orphaned_commentable_ids.empty?
+    if missing_commentable_ids.empty?
       puts "No orphaned comments found."
       next
     end
 
-    puts "Found #{orphaned_commentable_ids.size} missing parent comment(s). Creating placeholders..."
+    puts "Found #{missing_commentable_ids.size} missing parent comment(s). Creating placeholders..."
     $stdout.flush
 
     # Collect orphan info and sort by position in tree (top-down).
     # This ensures that when multiple comments are deleted in the same branch,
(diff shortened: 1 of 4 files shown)


lib/tasks/after_tasks.rake
```diff
@@ -654,5 +654,70 @@ namespace :After do
     AuditsBackfillJob.spawn_jobs
     puts "Backfill started and running on resque in background"
   end
+
+  desc "Create placeholder comments for orphaned replies whose parent was physically deleted"
+  task(create_placeholders_for_orphaned_comments: :environment) do
+    orphaned_commentable_ids = Comment
```
brianjaustin: Naming nitpick:
```suggestion
    missing_commentable_ids = Comment
```
(assuming I understand correctly and this list represents the list of comment IDs we need to make, rather than the IDs of existing comments whose parent is missing?)

brianjaustin (approved):
Just one naming thing, otherwise looking good! (I've approved already because I don't think a small nitpick is blocking either way)

lib/tasks/after_tasks.rake
pmonfort: Great catch Brian! Thanks!

PR #5728, opened 2026-04-14
AO3-7130 Hide non-interactive submit terms in tag work filters

noahbravo:
# Pull Request Checklist

* [X] Have you added the [Jira](https://otwarchive.atlassian.net) issue number as the *first* thing in your pull request title (e.g. `AO3-1234 Fix thing`)?
* [X] Do you have fewer than 5 pull requests already open? If not, please wait until they are reviewed and merged before creating new pull requests.

## Issue

https://otwarchive.atlassian.net/browse/AO3-7130

## Purpose

Fixes a screen reader issue on the tag works filter page where a visually hidden non-interactive submit landmark (`<dt class="landmark">Submit</dt>`) was still exposed to assistive technologies.

Changes made:
- Added `aria-hidden="true"` to the two submit landmark `<dt>` elements in `app/views/works/_filters.html.erb`.
- Replaced ts("Submit") with t(".submit_landmark") for those two lines to satisfy i18n lint.
- Added en.works.filters.submit_landmark in config/locales/views/en.yml and normalized locale file order (no behavior change).
- Kept existing structure and styling unchanged (minimal, low-risk fix scoped to the reported page).

I noticed that this pattern appears in other parts of the codebase as well, but limited this change to the reported page to keep the fix minimal and low-risk.

## Testing Instructions

Follow the reproduction steps in AO3-7130:
https://otwarchive.atlassian.net/browse/AO3-7130

Expected behavior:
- NVDA / TalkBack should no longer announce a stray non-interactive “Submit” term in the filter area.
- The actual “Sort and Filter” submit buttons should remain announced and operable.
- Visual layout should remain unchanged.

Automated tests:
- No new automated tests were added for this change.
- Existing automated checks required an i18n key and locale normalization update, both included in this PR.
This issue affects screen reader behavior on hidden semantic elements and was validated with manual accessibility testing.
- This issue affects screen reader behavior on hidden semantic elements and was validated with manual accessibility testing.

## Screenshots

Before:
ao3_7130_before

After:
ao3_7130_after

## Credit

Name: Noah Bravo  
Pronouns: they/them


app/views/works/_filters.html.erb
```diff
@@ -8,7 +8,7 @@
   <h3 class="landmark heading"><%= ts("Filters") %></h3>
   <%= field_set_tag ts("Filter results:") do %>
     <dl>
-      <dt class="landmark"><%= ts("Submit") %></dt>
+      <dt class="landmark" aria-hidden="true"><%= t(".submit_landmark") %></dt>
```
sarken: The `landmark` class is for text intended to be invisible to sighted users but read by screen readers, so we should avoid using `aria-hidden="true"` in combination with it -- they're at cross purposes with each other. For things we want to hide from both screen readers and sighted users, we should use the `hidden` class.

sarken (changes requested):

app/views/works/_filters.html.erb
noahbravo: Thanks for the clarification. I've updated this to use the `hidden` class instead of combining `landmark` with `aria-hidden="true"`, so it follows the intended pattern.
I noticed similar instances in other templates as well, but kept this PR scoped to AO3-7130.

sarken: Yeah, I think sticking to the narrow scope is a good idea. It's odd that this is suddenly a problem, given it's been in widespread use on the site for years -- better to test it in just one place for now.

sarken (approved):

PR #5733, opened 2026-04-14
AO3-7144 AO3-7147 Add collection listings for tags (with corrected page headings)

techlingyi:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-7144

https://otwarchive.atlassian.net/browse/AO3-7147

## Purpose

7147: Implement collection listings for tags -- now tags have index pages for collections like they do for works and bookmarks with that tag. Links have been updated accordingly.

7144: Collections index page headings are now consistent with other page headings and i18n'd.

## References

Adds to search_header, which has a PR for internationalization here: AO3-7357 #5695 

## Credit

Ling-Yi (any pronouns)


app/models/search/collection_search_form.rb
marcus8448: It looks like this list should be alphabetical.

app/helpers/search_helper.rb
```diff
@@ -27,6 +27,9 @@ def search_header(collection, search, item_name, parent=nil)
       header << ts("by %{username}", username: parent.login)
     when Language
       header << ts("in %{language}", language: parent.name)
+    when Work
+      # the rest will also be i18n'd like this with AO3-7357
+      header << t("search_helper.search_header.scope.work", work: parent.title)
```
Bilka2: Not a full review, just a note: The code for AO3-7357 needs to change approach, so I think to avoid merge conflicts/doing extra work here, it would be better to go with `ts` for now (and just don't mind that reviewdog will complain)

techlingyi: Just done, thank you! c:

techlingyi: Actually: since the kudos page for an individual work (ao3.org/works/[id]/kudos) links back to the work in its header, should I do that for the header of the collections page (ao3.org/works/[id]/collections) too? It does look a lot better as a link when the work title is in lowercase.

app/controllers/collections_controller.rb
```diff
@@ -30,29 +31,40 @@ def check_parent_visible
...
   def index
-    if params[:work_id]
-      @work = Work.find(params[:work_id])
+    base_options = {
+      page: params[:page] || 1
+    }
+    options = params[:collection_search].present? ? collection_filter_params : {}
+    options.merge!(base_options)
+
+    @page_title = case @owner
```
Bilka2:
```suggestion
    @page_subtitle = case @owner
```
with `@page_title` it won't have the `| Archive of Our Own` at the end, but we basically always want it there

app/controllers/collections_controller.rb
```diff
@@ -30,29 +31,40 @@ def check_parent_visible
...
+                  when Collection
+                    t(".subcollection", collection_title: @collection.title)
+                  when User
+                    t(".user", user_name: @user.login)
+                  when Tag
+                    t(".tag", tag_name: @tag.name)
+                  else
+                    t(".general")
```
Bilka2: Please put all these locales keys into a `page_title` group (e.g. `".page_title.subcollection"`) instead, so that it's clearer what they are used for

app/controllers/collections_controller.rb
```diff
@@ -30,29 +31,40 @@ def check_parent_visible
...
+                  end
+
+    if params[:work_id].present?
```
Bilka2: Could this check `@work` instead so that we only touch the param in one place (`load_owner`)? 

(And it may make sense to have this be inside the `@owner.present?` branch so that the `if` is clearly split between the root index and the ones with an owner. But maybe that is less readable. Up to you.)

app/controllers/collections_controller.rb
```diff
@@ -30,29 +31,40 @@ def check_parent_visible
...
       @collections = @search.search_results.scope(:for_search)
       flash_search_warnings(@collections)
-      @page_subtitle = t(".subcollections_page_title", collection_title: @collection.title)
-    elsif params[:user_id]
-      @user = User.find_by!(login: params[:user_id])
-      @search = CollectionSearchForm.new({ maintainer_id: @user.id, sort_column: "title.keyword" }.merge(page: params[:page]))
-      @collections = @search.search_results.scope(:for_search)
-      flash_search_warnings(@collections)
-      @page_subtitle = ts("%{username} - Collections", username: @user.login)
+      @pagy = pagy_query_result(@collections) if @collections.respond_to?(:total_pages)
```
Bilka2: This and the `@pagy` down in the else branch aren't used anywhere. Could you decide on one of switching to `pagy` or continuing to use `will_paginate` instead of trying to do both?

app/views/collection_profile/show.html.erb
```diff
@@ -19,7 +19,7 @@
         <dd>
           <ul class="tags commas">
             <% @collection.tags.each do |tag| %>
-              <li><%= link_to_tag_works(tag) %></li>
+              <li><%= link_to(tag.name, tag_collections_path(tag), class: "tag") %></li>
```
Bilka2:
```suggestion
              <li><%= link_to(tag.display_name, tag_collections_path(tag), class: "tag") %></li>
```
To match the behaviour of `link_to_tag_works` (relevant here because it's overridden for archive warnings which can be used in this field)

app/views/collections/_collection_blurb.html.erb
```diff
@@ -17,7 +17,7 @@
       <% unless collection.tag_string.blank? %>
         <h5 class="tags heading">
           <span class="landmark"><%= t(".collection_tags") %></span>
-          <%= safe_join(collection.tags.collect{ |tag| link_to_tag_works(tag) }, t("support.array.words_connector")) %>
+          <%= safe_join(collection.tags.collect { |tag| link_to(tag.name, tag_collections_path(tag), class: "tag") }, t("support.array.words_connector")) %>
```
Bilka2:
```suggestion
          <%= safe_join(collection.tags.collect { |tag| link_to(tag.display_name, tag_collections_path(tag), class: "tag") }, t("support.array.words_connector")) %>
```
Same as the collection profile

app/views/tag_wranglings/_wrangler_dashboard.html.erb
```diff
@@ -30,6 +30,11 @@
               <%= span_if_current t(".use_type_and_count", use_type: t(".use_type.external_works"), count: @counts[key]),
                     { controller: :bookmarks, action: :index, tag_id: @tag } %>
             </li>
+            <% elsif key == "Collections" %>
+              <li>
+              <%= span_if_current t(".use_type_and_count", use_type: t(".use_type.collections"), count: @counts[key]),
+                    { controller: :collections, action: :index, tag_id: @tag } %>
```
Bilka2: Nitpick: You can also use `tag_collections_path` here

config/locales/controllers/en.yml
```diff
@@ -137,7 +137,10 @@ en:
       page_title: "%{collection_title} - Profile"
   collections:
     index:
-      subcollections_page_title: "%{collection_title} - Subcollections"
+      general: Collections
+      subcollection: "%{collection_title} - Subcollections"
+      tag: "%{tag_name} - Collections"
+      user: "%{user_name} - Collections"
```
Bilka2:
```suggestion
      user: "%{username} - Collections"
```
We've switched away from `user_name` where possible

config/locales/views/en.yml
```diff
@@ -2995,15 +3010,22 @@ en:
       type: Type
       wrangling_status: Wrangling status
     show:
-      canonical_html: It's a %{canonical_tag_link}. You can use it to %{filter_works_link} and to %{filter_bookmarks_link}.
+      canonical_html: It's a %{canonical_tag_link}. You can use it to %{filter_works_link} and to %{filter_bookmarks_link}, as well as %{filter_collections_link}.
```
Bilka2: This change is not in the Jira issue and should be reverted.

config/locales/views/en.yml
```diff
@@ -3323,6 +3345,13 @@ en:
     index:
       advanced_search: try our advanced search
       choose_fandom: choose a fandom
+      navigation:
+        tag_bookmarks: Bookmarks
+        tag_collections: Collections
+        tag_works: Works
```
Bilka2: Nitpick: Consider grouping this and the user ones via a subkey instead of a prefix. I currently instinctively expect the translation to be "Tag Works", not "Works" inside the group of links for tags (so like tag.bookmarks in the erb)

app/controllers/collections_controller.rb
```diff
@@ -191,10 +203,28 @@ def destroy
...
 
+  protected
+
+  def load_owner
+    @user = User.find_by!(login: params[:user_id]) if params[:user_id].present?
+    @work = Work.find(params[:work_id]) if params[:work_id].present?
+    @collection = Collection.find_by!(name: params[:collection_id]) if params[:collection_id].present?
+    if params[:tag_id].present?
+      @tag = Tag.find_by_name(params[:tag_id])
+      raise ActiveRecord::RecordNotFound, "Couldn't find tag named '#{params[:tag_id]}'" unless @tag.is_a?(Tag)
```
Bilka2:
```suggestion
      @tag = Tag.find_by_name!(params[:tag_id])
```

config/routes.rb
```diff
@@ -132,6 +132,7 @@
     end
     resources :works
     resources :bookmarks
+    resources :collections
```
Bilka2: I guess you copied what works and bookmarks do? Since we only want the index here though, could you edit to this please:
```suggestion
    resources :collections, only: [:index]
```

app/models/search/collection_query.rb
```diff
@@ -44,6 +45,24 @@ def queries
     ].flatten.compact
   end
 
+  def add_owner
```
Bilka2: Could you move this to the collections controller instead? It's pretty specific to there, especially with the enforced sort options, and not really related to generic queries for collections

app/models/search/collection_search_form.rb
```diff
@@ -7,10 +7,11 @@ class CollectionSearchForm
     challenge_type
     closed
     maintainer_id
+    filter_ids
+    parent_id
```
Bilka2: Nitpick: Can you sort these alphabetically again?

app/helpers/search_helper.rb
```diff
@@ -61,4 +63,13 @@ def bookmarks_original_path
       **params.slice(:tag_id, :collection_id, :pseud_id, :user_id).permit!
     )
   end
+
+  def collections_original_path
```
Bilka2, line 64: Could you use this for the Clear Filters link in `_filters`? (Merging with master should be enough for that, since another PR has done that change)

app/helpers/search_helper.rb
```diff
@@ -61,4 +63,13 @@ def bookmarks_original_path
...
+    url_for(
+      controller: :collections,
+      action: :index,
+      only_path: true,
+      **params.slice(:tag_id, :collection_id, :user_id, :work_id).permit!
```
Bilka2: This doesn't match https://github.com/otwcode/otwarchive/pull/5788 - which one is wrong?

Looks like the other PR is wrong, but it still works because of  how url_for [treats missing route keys](https://api.rubyonrails.org/classes/ActionDispatch/Routing/UrlFor.html#method-i-url_for):
> Missing routes keys may be filled in from the current request’s parameters (e.g. :controller, :action, :id, and any other parameters that are placed in the path).

So not something we need to fix for #5788 right now, but when you merge master into this PR, please use this version. Or preferably experiment with whether we can remove the hidden user_id etc from _filters.html.erb and remove the param slice here and still have the clear filter link automatically going to the tag's collections / user's collections (with an automated test to verify behaviour)

app/controllers/collections_controller.rb
```diff
@@ -30,29 +31,40 @@ def check_parent_visible
...
       @collections = @work.approved_collections
         .by_title
         .for_blurb
         .paginate(page: params[:page])
-    elsif params[:collection_id]
-      @collection = Collection.find_by!(name: params[:collection_id])
-      @search = CollectionSearchForm.new({ parent_id: @collection.id, sort_column: "title.keyword" }.merge(page: params[:page]))
+    elsif @owner.present?
+      @sort_and_filter = @tag.present?
+      @search = CollectionSearchForm.new(options.merge(parent: @owner))
```
Bilka2: Due to how `options` is set up with always grabbing the search params from the URL, this allows filtering the subcollections page with URL parameters, which is not part of this PR. (It also allows it for the user page, but that's now in master so that's fine). Please make sure the subcollection page can still not be filtered with URL parameters

app/controllers/collections_controller.rb
```diff
@@ -30,29 +31,40 @@ def check_parent_visible
...
+    if params[:work_id].present?
       @collections = @work.approved_collections
         .by_title
         .for_blurb
         .paginate(page: params[:page])
-    elsif params[:collection_id]
-      @collection = Collection.find_by!(name: params[:collection_id])
-      @search = CollectionSearchForm.new({ parent_id: @collection.id, sort_column: "title.keyword" }.merge(page: params[:page]))
+    elsif @owner.present?
+      @sort_and_filter = @tag.present?
```
Bilka2: Warning: Merge conflict with master, `@user` should also be checked here

app/controllers/collections_controller.rb
```diff
@@ -191,10 +203,28 @@ def destroy
...
+    @work = Work.find(params[:work_id]) if params[:work_id].present?
+    @collection = Collection.find_by!(name: params[:collection_id]) if params[:collection_id].present?
+    if params[:tag_id].present?
+      @tag = Tag.find_by_name(params[:tag_id])
+      raise ActiveRecord::RecordNotFound, "Couldn't find tag named '#{params[:tag_id]}'" unless @tag.is_a?(Tag)
+
+      unless @tag.canonical?
+        redirect_path = @tag.merger.present? ? tag_collections_path(@tag.merger) : tag_path(@tag)
+        redirect_to redirect_path and return
+      end
```
Bilka2, line 226: Since this mostly a copy of what is done in `load_owner` for bookmarks (and partially works), do you think this could be extracted into a shared method? Maybe `url_for`s behaviour regarding missing route keys is also helpful here for easily making the path for the redirecting to the merger

But if that method would be really messy, don't bother

app/controllers/collections_controller.rb
```diff
@@ -191,10 +203,28 @@ def destroy
...
+    end
+    @owner = @user || @work || @collection || @tag
+  end
+
   private
 
   def collection_filter_params
-    params.permit(:commit, collection_search: [
+    params.permit(:commit, :tag_id, collection_search: [
```
Bilka2: Ref https://github.com/otwcode/otwarchive/pull/5788#discussion_r3411924654, you can probably just pick the version from master when you merge

app/helpers/search_helper.rb
```diff
@@ -27,6 +27,8 @@ def search_header(collection, search, item_name, parent=nil)
       header << ts("by %{username}", username: parent.login)
     when Language
       header << ts("in %{language}", language: parent.name)
+    when Work
+      header << ts("including %{work}", work: parent.title)
```
Bilka2: Regarding your comment about linking the work: Since that search header isn't in the list of pages in the issue to change, I'd keep it the same and not link. So this is fine

app/views/collections/_filters.html.erb
Bilka2, line 1: I believe you have to optionally use the tags_collections_path in the `form_for` if this is on a tag collections page. But merge master first, since that also modifies the `form_for`

(Now I'm wondering... would the `url_for` missing route key magic also help here to make this cleaner)

features/collections/collection_browse.feature
```diff
@@ -179,16 +179,16 @@ Feature: Collection
       And I should not see "Some Other Collection"
       And I should not see "Surprise Presents"
 
-  Scenario: Filter collections by non-canonical and non-existent collection tags
+  Scenario: Filter collections by canonicals, synonyms, noncanonicals, and non-existent collection tags
```
Bilka2: The scenario specifically tested _not_ `filter_ids`, so please separate the test for that into a different scenario (and please don't make any of the tags used here canonical)

app/models/search/collection_search_form.rb
```diff
@@ -7,10 +7,11 @@ class CollectionSearchForm
     challenge_type
     closed
     maintainer_id
+    filter_ids
```
Bilka2: From what I can tell, since the search form doesn't use this, this addition is not needed. Does everything still work without it?

app/views/collections/index.html.erb
```diff
@@ -1,49 +1,56 @@
...
 </h2>
 
-<h3 class="heading">
-  <% if @collections.empty? %>
-    <%= ts("Sorry, there were no collections found.") %>
-  <% else %>
-    <%= search_header @collections, @query, ts("Collection") %>
-  <% end %>
-</h3>
+<%# TODO: Do we still need this now that there's a 0 in the search header? %>
```
Bilka2: No, please remove it

app/views/collections/index.html.erb
```diff
@@ -1,49 +1,56 @@
...
 <ul class="navigation actions" role="navigation">
+  <% if @tag && !@collection %>
+    <li><%= span_if_current t(".navigation.works"), tag_works_path(@tag) %></li>
+    <li><%= span_if_current t(".navigation.bookmarks"), tag_bookmarks_path(@tag) %></li>
+    <li><%= span_if_current t(".navigation.collections"), tag_collections_path(@tag) %></li>
+  <% end %>
+  <% if @facets.present? %>
+    <%# Filters button for narrow screens jumps to filters when JavaScript is disabled and opens filters when JavaScript is enabled %>
+    <li class="narrow-shown hidden"><a href="#work-filters" id="go_to_filters"><%= t(".navigation.filters") %></a></li>
+  <% end %>
```
Bilka2: This button is already added below if `@sort_and_filter` is set, which it will be on the new page. So I think this addition is redundant (the controller doesn't set `@facets`), could you remove it?

app/views/collections/index.html.erb
Bilka2, line 1: For the `navigation` i18n in this file, please generally go with with whatever is in master when merging - I didn't review it here since master already has a reviewed version

features/support/paths.rb
```diff
@@ -197,6 +197,9 @@ def path_to(page_name)
     when /^the bookmarks (?:tagged|page for the tag) "(.*)"$/i
       step %{all indexing jobs have been run}
       tag_bookmarks_path(Tag.find_by_name(Regexp.last_match(1)))
+    when /^the collections "(.*)"$/i
```
Bilka2: I think this is too ambiguous, could you go with something like "the collections tagged ..." to match the name for `tag_works_path`?

app/helpers/search_helper.rb
Bilka2: Also, I believe this adds "including WORK" to the search header for works/.../bookmarks, which should not be the case

Bilka2 (changes requested):

Recommendation for how to action this review:
1. Read/skim the comments for what I say regarding merges - there's a few spots where master and this PR disagree and specifically the master or this PR's version need to be picked
2. Merge master according to that
3. Action the rest of the review

I want to note that I didn't yet review the newly added test file - there are already a lot of comments so I think it will be simpler to do that as a second round

config/locales/views/en.yml
techlingyi: Oh, I thought it fell under

> 10c. A “Collections” link should be added after “Bookmarks” in the subnavigation on tag landing pages for canonical tags and syns

So just the extra Collections button next to the Bookmarks one but no addition to the description paragraph, then?

app/helpers/search_helper.rb
techlingyi: ~~After some manual testing, it looks like the hidden tag_id needs to stay - when you filter from a tag page for collections the URL goes from /tags/`tag_id`/collections to /collections?collection_search[blahblah]=blahblah...&tag_id=`tag_id`, like it does for a tag page for works, and when I delete the hidden tag_id the search just doesn't look for collections with that tag anymore. The hidden user_id and collection_id are removable, though! I'll get rid of those.~~

Edit: NEVERMIND I realized one of your other comments on the form_for fixes this issue! 🎉 All removable then! (Now I'm just wondering why a tag page for works/bookmarks doesn't work this way…)

app/helpers/search_helper.rb
techlingyi: I've reverted it back to the original, but unfortunately because the h3 was removed for the sake of all the other pages in [AO3-7144](https://otwarchive.atlassian.net/browse/AO3-7144), it now reads just "Collections including [work]" without any count of how many collections there are below that - is that acceptable, or should I add back the h3 just for this page?

techlingyi:
@Bilka2 Ready for review again, thank you! :D

app/controllers/collections_controller.rb
```diff
@@ -31,32 +32,48 @@ def check_parent_visible
...
+                       t(".page_title.general")
+                     end
+
+    if @owner.present?
+      if @work.present?
+        @sort_and_filter = false
+        @collections = @work.approved_collections
+          .by_title
+          .for_blurb
+          .paginate(base_options)
```
Bilka2, line 58: This is still using will_paginate, it should be pagy as well

app/views/collections/index.html.erb
```diff
@@ -1,58 +1,55 @@
...
+<h3 class="landmark heading"><%= t(".landmark.navigation") %></h3>
+<ul class="navigation actions" role="navigation">
+  <% if @tag && !@collection %>
+    <li><%= span_if_current t(".navigation.works"), tag_works_path(@tag) %></li>
+    <li><%= span_if_current t(".navigation.bookmarks"), tag_bookmarks_path(@tag) %></li>
+    <li><%= span_if_current t(".navigation.collections"), tag_collections_path(@tag) %></li>
+  <% end %>
   <%# Collections and Open Challenges links unless a logged in user viewing own collections page %>
-  <% unless logged_in? && @user && @user == current_user %>
-    <li><%= span_if_current t(".navigation.actions.collections"), collections_path %></li>
```
Bilka2, line 28: could you keep the i18n for existing lines the same in this file, to reduce possible merge conflicts? (and because the existing i18n has already been reviewed and I'd like to avoid the effort of redoing and re-reviewing it). Additionally, please merge master, since there have been changes to this file. I'm not really re-reviewing the rest of the file right now because I think it will get shuffled around so much by the merge that it'd be pointless

config/locales/views/en.yml
Bilka2: Yep!

app/helpers/search_helper.rb
Bilka2: I believe it should say "N Collections including WORK_TITLE", like on the main collections index described in AO3-7144 (that's what I meant with my initial comment). I suppose you can achieve that in the same way you achieved it for the main collections index, with the `collections_in_the` directly on the index page?

(I don't like having the text outside search_header (we'll need it inside to i18n it properly) but right now I can't come up with a better solution that isn't a hack and since we're not i18n'ing it now and i18n will rewrite the method anyway it's probably fine)

app/controllers/collections_controller.rb
```diff
@@ -31,32 +32,48 @@ def check_parent_visible
...
+      end
     else
       @sort_and_filter = true
-      @search = CollectionSearchForm.new(options.merge(page: params[:page]))
+      @search = CollectionSearchForm.new(options)
       @collections = @search.search_results.scope(:for_search)
-      flash_search_warnings(@collections)
     end
+
+    flash_search_warnings(@collections) if @sort_and_filter
```
Bilka2, line 75: This does not show the search warnings if the owner is a collection, but it should. It's likely simpler to move this back into the individual if branches

app/views/collections/index.html.erb
```diff
@@ -1,58 +1,55 @@
...
-  <% end %>
-</h3>
 <!--/descriptions-->
 
 <!--Subnavigation, sorting and actions-->
-<h3 class="landmark heading"><%= t(".navigation.landmark_heading") %></h3>
-<ul class="navigation actions">
+<h3 class="landmark heading"><%= t(".landmark.navigation") %></h3>
+<ul class="navigation actions" role="navigation">
+  <% if @tag && !@collection %>
```
Bilka2, line 18: when would collection **and** tag be present? From what I can tell it can only ever be either of those, not both

features/collections/collection_taglisting.feature
```diff
@@ -0,0 +1,113 @@
...
+    And I press "Update"
+    And all indexing jobs have been run
+    And I view the tag "My Beautiful Canonical Tag"
+    And I follow "Collections" within ".primary.header.module"
+  Then I should see "My Beautiful Canonical Tag" within "h2.heading"
+    And I should see "2 Collections" within "h2.heading"
+    And I should see "Multiple Tag Test Collection"
+    And I should see "Gorgeous Tag Test Collection"
+    But I should not see "Beautiful Tag Test Collection"
+    And I should not see "Noncanonical Tag Test Collection"
```
Bilka2, line 113: Could you add the missing newline at the end of the file?

features/collections/collection_taglisting.feature
```diff
@@ -0,0 +1,113 @@
+@collections
+Feature: Collection
+  In order to have an archive full of collections
+  As a humble user
+  I want to locate and browse through tag collection listings
+
+  Scenario: Tags and synonyms of the tag show up in the tag's collection listing
```
Bilka2, line 7:
```suggestion
  Scenario: Collections tagged with tags and synonyms of the tag show up in the tag's collection listing
```

features/collections/collection_taglisting.feature
```diff
@@ -0,0 +1,113 @@
...
+    And I should see "Beautiful Tag Test Collection"
+    And I should see "Multiple Tag Test Collection"
+    And I should see "Gorgeous Tag Test Collection"
+    But I should not see "Noncanonical Tag Test Collection"
+
+  Scenario: We can further filter a tag's collection listing with collection_search_tag
+  Given a set of collections for tag page searching
+  When I go to the collections page
+    And I follow "My Beautiful Canonical Tag"
+    And I fill in "collection_search_tag" with "a noncanonical tag"
```
Bilka2, line 36: For this field name and the `collection_search_challenge_type_giftexchange` usages below, could you use the name that is displayed to humans? We prefer that in cucumber tests

features/collections/collection_taglisting.feature
```diff
@@ -0,0 +1,113 @@
...
+  Given a set of collections for tag page searching
+  When I am logged in as the owner of "Beautiful Tag Test Collection"
+    And I go to "Beautiful Tag Test Collection" collection's page
+    And I follow "Profile"
+    And I follow "Collection Settings"
+    And I fill in "Collection Tags" with ""
+    And I press "Update"
+    And all indexing jobs have been run
+    And I view the tag "My Beautiful Canonical Tag"
+    And I follow "Collections" within ".primary.header.module"
```
Bilka2, line 107: In this scenario you could use the nice `the collections tagged` path you added instead of doing the extra navigation

Bilka2 (changes requested):
This is looking very good! Mostly just minor comments from me, plus could you take a look at and respond to https://github.com/otwcode/otwarchive/pull/5733/files#r3579071085 from the previous round (I assume it's just more complicated and not cleaner to merge the tag loading?)

For my own tracking, I've looked at everything except:
* subnavigation on the collection index, pending master merge for that because it's modified so heavily in master

(If the dates on the comments are a bit confusing, I had half the review done, got pulled away and then got back now to finish it off. I edited some of the earlier comments today as I went over the files again, so basically all comments are from today, even though I started reviewing over a month ago...)

PR #5735, opened 2026-04-15
AO3-7378 Add button to copy setup key to clipboard for totp

jmalark:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-7378

## Purpose

Context: an admin setting up two-step verification for their account

This PR adds a button to copy the setup key to the clipboard (as long as JavaScript is enabled, otherwise the button will not display)

## Credit
Jesse Malark he/him

### DIFF
--- a/app/views/admin/totp/reauthenticate_create.erb
+++ b/app/views/admin/totp/reauthenticate_create.erb
@@ -16,8 +16,13 @@
     <div class="image"><p class="qr"><%= qr_code_as_svg(current_admin.totp_qr_code_uri) %></p></div>
     <div class="notes">
       <h4 class="heading"><%= t(".app_setup.key.heading") %></h4>
-      <code><%= current_admin.otp_secret.scan(/.{1,4}/).join(" ") %></code>
+      <code><%= setup_key = current_admin.otp_secret.scan(/.{1,4}/).join(" ") %></code>
       <p class="footnote"><%= t(".app_setup.key.format_note") %></p>
+      <p class="actions">  
+          <button id="copy-totp-setup-key" onclick="copySetupKey()" style="visibility: hidden;">
+            <%= t(".app_setup.key.copy_to_clipboard") %>
+          </button>
+      </p>
     </div>
   </div>
 </section>
@@ -34,3 +39,15 @@
   <% end %>
 </section>
 <!--/content-->
+
+<% content_for :footer_js do %>
+  <script type="text/javascript">
+    document.getElementById("copy-totp-setup-key").style.visibility = "visible";
+
+    async function copySetupKey() {
+      await navigator.clipboard.writeText("<%= setup_key %>");
+      document.getElementById("copy-totp-setup-key").textContent = "<%= t(".app_setup.key.copied") %>";
+    }
+  </script>
+<% end %>
+
--- a/config/locales/views/en.yml
+++ b/config/locales/views/en.yml
@@ -487,6 +487,8 @@ en:
           heading: 'Step 1: Set up your authenticator app'
           instructions: Use your authenticator app to scan the QR code or enter the manual setup key. The app will give you a 6-digit code, which you'll use in the next step.
           key:
+            copied: Copied!
+            copy_to_clipboard: Copy
             format_note: Spaces and capitalization don't matter
             heading: Setup Key
         enter_code:
--- a/Gemfile
+++ b/Gemfile
@@ -61,7 +61,7 @@ gem "terrapin"
 gem 'fastimage'
 
 # Gems for authentication
-gem "devise", git: "https://github.com/otwcode/devise", ref: "fade89f"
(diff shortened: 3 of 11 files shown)


app/views/admin/totp/reauthenticate_create.erb
```diff
@@ -34,3 +39,14 @@
...
+<% content_for :footer_js do %>
+  <script type="text/javascript">
+    document.getElementById("copy-totp-setup-key").style.visibility = "visible";
+
+    async function copySetupKey() {
+      await navigator.clipboard.writeText("<%= current_admin.otp_secret.scan(/.{1,4}/).join(" ") %>");
+      document.getElementById("copy-totp-setup-key").textContent = "<%= t(".app_setup.key.copied") %>";
+    }
+  </script>
+<% end %>
```
brianjaustin: Please add a trailing newline to this file (https://stackoverflow.com/a/5813359)

app/views/admin/totp/reauthenticate_create.erb
```diff
@@ -34,3 +39,14 @@
   <% end %>
 </section>
 <!--/content-->
+
+<% content_for :footer_js do %>
+  <script type="text/javascript">
+    document.getElementById("copy-totp-setup-key").style.visibility = "visible";
+
+    async function copySetupKey() {
+      await navigator.clipboard.writeText("<%= current_admin.otp_secret.scan(/.{1,4}/).join(" ") %>");
```
brianjaustin: Since we do this scan/join in two places, could you add a variable for it?

config/locales/views/en.yml
```diff
@@ -489,6 +489,8 @@ en:
           key:
             format_note: Spaces and capitalization don't matter
             heading: Setup Key
+            copy_to_clipboard: Copy
+            copied: Copied!
```
brianjaustin: It looks like these are in the wrong order (thus the failing test). You can automatically fix it by running
```
i18n-tasks normalize -l en
```

brianjaustin (changes requested):

app/views/admin/totp/reauthenticate_create.erb
```diff
@@ -34,3 +39,15 @@
   <% end %>
 </section>
 <!--/content-->
+
+<% content_for :footer_js do %>
+  <% setup_key = current_admin.otp_secret.scan(/.{1,4}/).join(" ") %>
```
brianjaustin: This will need to be further up in the file so it's accessible to the code on line 19

brianjaustin (changes requested):

app/views/admin/totp/reauthenticate_create.erb
jmalark: Added

config/locales/views/en.yml
jmalark: I'm not ignoring you, I just literally can't run the normalization because I'm still spinning my wheels trying to get bundle install  /the gems  installed correctly. Not sure if if bundle install didn't used to run on the docker init / compose steps or if I did something locally to throw my environment out of whack, but I haven't been able to run the run code ever since. Spent most of yesterday and a chunk of Friday fighting permission issues and trying to get stuff enabled so my terminal could understand the commands that were added from the Ruby Installer for Windows, but I'm still pretty dead in the water. I guess I'm going to be emailing otw-coders@transformativeworks.org to see if anyone is willing to walk me through what I'm missing so I can get these dependencies installed (^_^;)

brianjaustin: If you're using the docker (compose) setup, I'd recommend just running the normalize command in there as well to avoid annoying windows issues. Something along the lines of

```
docker compose run --rm test i18n-tasks normalize -l en
```

should work. (Also worth noting, the docker compose setup will not do anything to your system outside the containers, so if you had Ruby installed there, it will not install any dependencies. This is intended.)

jmalark: Ok got it to run and it put them in the corrected order

sarken (changes requested):
Thanks for working on this! 

It looks like the issue didn't contain any design information, so I took a quick look. Can you please add the following CSS to the bottom of the `.annotation` styles in `22-system-messages.css`? It will make sure the button is centered like the rest of the text, and take out the whitespace that was throwing off the vertical alignment. 

```
.annotation .actions {
  float: none;
  margin-block-end: 0;
  padding-block-end: 0;
}
```

We can make a separate issue for this once this pull request is merged, but now that we have similar JavaScript in two places, we should also move the JavaScript to an external resource (most likely application.js). This makes it easier to maintain and reuse and is more in line with modern standards, particularly regarding security.

sarken (approved):
Thanks for the CSS changes! I'll leave the final approval to Brian.

brianjaustin (approved):
Thanks!

PR #5738, opened 2026-04-17
AO3-7324 List user creations by date, not by pseud

kekley:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-7324

## Purpose

List user creations on the spamban page in order from oldest to newest by creation date and then ID, rather than by pseud. 

## Credit

kekley (they/them)

### DIFF
--- a/app/controllers/admin/admin_users_controller.rb
+++ b/app/controllers/admin/admin_users_controller.rb
@@ -22,8 +22,8 @@ class Admin::AdminUsersController < Admin::BaseController
   end
 
   def load_user_creations
-    @works = @user.works.paginate(page: params[:works_page])
-    @comments = @user.comments.paginate(page: params[:comments_page])
+    @works = @user.works.reorder(created_at: :asc, id: :asc).paginate(page: params[:works_page])
+    @comments = @user.comments.reorder(created_at: :asc, id: :asc).paginate(page: params[:comments_page])
   end
 
   def index
--- a/spec/controllers/admin/admin_users_controller_spec.rb
+++ b/spec/controllers/admin/admin_users_controller_spec.rb
@@ -770,8 +770,6 @@ describe Admin::AdminUsersController do
           let!(:newer_comment) { create(:comment, pseud: user.default_pseud) }
           let!(:older_work) { create(:work, authors: [other_pseud]) }
           let!(:newer_work) { create(:work, authors: [user.default_pseud]) }
-          let!(:other_comment) { create(:comment) }
-          let!(:other_work) { create(:work) }
 
           before do
             fake_login_admin(admin)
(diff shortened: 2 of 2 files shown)


sarken:
Hi, kekley!

If you'd like the ability to comment on, assign, and transition issues in the future, you're welcome to create a Jira account! It makes things a bit easier for us on the organizational side if the Full Name on your Jira account either closely matches the name you'd like us to credit in the release notes or includes it in parentheses, e.g. "Nickname (CREDIT NAME)."

spec/controllers/admin/admin_users_controller_spec.rb
```diff
@@ -765,18 +765,26 @@ def expect_changes_made_by(admin, log_items)
       authorized_roles.each do |role|
         context "with #{role} role" do
           let(:admin) { create(:admin, roles: [role]) }
-          let!(:user_comment) { create(:comment, pseud: user.default_pseud) }
-          let!(:other_comment) { create(:comment) }
-          let!(:user_work) { create(:work, authors: [user.default_pseud]) }
-          let!(:other_work) { create(:work) }
+          let!(:other_pseud) { create(:pseud, user: user) }
+          let!(:older_comment) { create(:comment, pseud: other_pseud) }
```
brianjaustin, line 769: Rather than setting the created_at columns below, how about using the [travel_to](https://api.rubyonrails.org/classes/ActiveSupport/Testing/TimeHelpers.html) helper when creating the comments? For example:
```suggestion
          let!(:older_comment) do
            travel_to(2.days.ago) do
              create(:comment, pseud: other_pseud)
            end
          end
```

brianjaustin (commented):
Just 1 comment, otherwise looking good already!

app/controllers/admin/admin_users_controller.rb
```diff
@@ -22,8 +22,8 @@ def user_is_banned
   end
 
   def load_user_creations
-    @works = @user.works.paginate(page: params[:works_page])
-    @comments = @user.comments.paginate(page: params[:comments_page])
+    @works = @user.works.reorder(created_at: :asc, id: :asc).paginate(page: params[:works_page])
+    @comments = @user.comments.reorder(created_at: :asc, id: :asc).paginate(page: params[:comments_page])
```
Bilka2, line 26: These two additions can be simplified to `order(id: :asc)`

reorder is only needed if there was previously a different order set on the query, which is not the case here. The issue specifies that the creations should be ordered by id, and since that's unique (so a stable sorting criterion), it can be the only order we specify

spec/controllers/admin/admin_users_controller_spec.rb
Bilka2: and since this won't work after my suggestion: you can create records with a set id with create, just pick something not super low so there is less risk of overlap with things from other tests. E.g.:
https://github.com/otwcode/otwarchive/blob/4a175843439900b0a356da3fc5cab4874b369d43/spec/models/abuse_report_spec.rb#L737-L738

PR #5739, opened 2026-04-17
AO3-7405 Prevent N+1 queries on tags#feed

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-7405

## Purpose

Prevent N+1 queries on tags#feed by using `includes`. The change in the feed builder itself is helpful because the previous code would get the works twice (once with limit 1 and once with limit 25), and that query is potentially quite slow. It's much faster to only do the limit 25 query and grab the date from the first work in the loop.

Failing test for the N+1 (and passing test for the feed date) pushed first to show that they're testing the right thing.

## Credit

Bilka

### DIFF
--- a/app/controllers/tags_controller.rb
+++ b/app/controllers/tags_controller.rb
@@ -81,7 +81,6 @@ class TagsController < ApplicationController
                else
                  @tag.works
                end
-
       @works = @works.visible_to_all.order(created_at: :desc).limit(25).includes(:tags, :language, :series)
     else
       redirect_to(tag_works_path(tag_id: @tag.to_param)) && return
--- a/app/controllers/tags_controller.rb
+++ b/app/controllers/tags_controller.rb
@@ -72,19 +72,17 @@ class TagsController < ApplicationController
   end
 
   def feed
-    begin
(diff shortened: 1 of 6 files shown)


app/controllers/tags_controller.rb
```diff
@@ -72,19 +72,16 @@ def show
...
 
   def feed
-    begin
-      @tag = Tag.find(params[:id])
-    rescue ActiveRecord::RecordNotFound
-      raise ActiveRecord::RecordNotFound, "Couldn't find tag with id '#{params[:id]}'"
-    end
+    @tag = Tag.find(params[:id])
     @tag = @tag.merger if !@tag.canonical? && @tag.merger
     # Temp for testing
```
brianjaustin, line 77: This comment feels like it isn't accurate anymore?

brianjaustin (approved):
Thanks! (Just one unrelated thing that doesn't have to be addressed if you'd rather not)

app/controllers/tags_controller.rb
marcus8448: The comment is still accurate (F/F has a feed while the other categories do not).

PR #5744, opened 2026-04-18
AO3-7397 AO3-6765 Configurable page for skin previews

nicolacleary:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-7397
https://otwarchive.atlassian.net/browse/AO3-6765 (adopted)

## Purpose

Redirect to the URL specified in the config with `SKIN_PREVIEW_URL` when previewing a skin.

Remove the functionality to redirect to the works index of a random tag - this provided no guarantees on the kind of content shown and would sometimes choose tags that no longer existed (resulting in 500 errors).

This PR also adopts changes from #4934 to improve the skin preview message:
- Adds the skin ID to the preview tip (rather than making the user copy-paste it themselves)
- Adjusts wording in message

## Credit

nicolacleary (she/her)
Adopted changes from [AO3-6765](https://otwarchive.atlassian.net/browse/AO3-6765) #4934 David Bilsky/Ironskink (He/Him)

[AO3-6765]: https://otwarchive.atlassian.net/browse/AO3-6765?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/controllers/skins_controller.rb
```diff
@@ -130,12 +130,11 @@ def update
...
-    flash[:notice] << ts("You are previewing the skin %{title}. This is a randomly chosen page.", title: @skin.title)
-    flash[:notice] << ts("Go back or click any link to remove the skin.")
-    flash[:notice] << ts("Tip: You can preview any archive page you want by tacking on '?site_skin=[skin_id]' like you can see in the url above.")
-    flash[:notice] << "<a href='#{skin_path(@skin)}' class='action' role='button'>".html_safe + ts("Return To Skin To Use") + "</a>".html_safe
-    tag = FilterCount.where("public_works_count BETWEEN 10 AND 20").random_order.first.filter
-    redirect_to tag_works_path(tag, site_skin: @skin.id)
+    flash[:notice] << t(".skin_title", title: @skin.title)
+    flash[:notice] << t(".remove_skin")
+    flash[:notice] << t(".tip")
+    flash[:notice] << ("<a href='#{skin_path(@skin)}' class='action' role='button'>".html_safe + t(".return_to_skin") + "</a>".html_safe)
```
nicolacleary, line 136: Rubocop would complain on the usage of `ts` when I removed "This is a randomly chosen page" from the notice.
I decided it made sense to convert all of the lines rather than just one

config/config.yml
```diff
@@ -802,3 +802,9 @@ PSEUD_SHARDS: 5
 TAG_SHARDS: 5
 USER_SHARDS: 5
 WORKS_SHARDS: 5
+
+# The relative URL to redirect to when previewing a skin
+# e.g. /works -> http://test.archiveofourown.org/works?site_skin=[skin_id]
+# Production should pick something more suitable with roughly 1-20 works and
+# predictable content such as /users/OTW_Translation/works
+SKIN_PREVIEW_URL: '/works'
```
nicolacleary: The ticket asks for `/users/OTW_Translation/works` but this user is not present in the development data. I chose `/works` for this value since it should always be available, but this should be changed the proper value in the production config.

app/controllers/skins_controller.rb
```diff
@@ -130,12 +130,11 @@ def update
...
+    redirect_to "#{ArchiveConfig.SKIN_PREVIEW_URL}?site_skin=#{@skin.id}"
```
nicolacleary, line 137: This provides no guarantees that the url is correct (formatting or exists) - e.g. if you configure "/idonotexist" then you would get an error only when you tried to preview a skin.

config/config.yml
slavalamp: this user exists on staging (the test.ao3.org site linked in the issue, for testers who have access to it) and you can create it in your local environment for testing it yourself (see [Creating Development Data](https://github.com/otwcode/otwarchive/wiki/Development-Data) in the wiki)

nicolacleary: I understand that it's possible to create it locally, and I have successfully created users in my dev environment.

I worry that configuring a user not in the dev data could negatively impact the dev experience:
- In order for the preview function to work in dev, devs will have to manually create a user or update the config
- If this happens with multiple features the dev setup instructions become complicated and hard to follow and maintain
- It won't be clear that it's a config error and not a bug

I could:
1. Add the `OTW_Translation` user in the fixtures - I believe this would then be part of the dev/test data for new installations
1. Configure an existing dev user here - the config would still need to be changed before testing/prod
1. Override the config in `development.rb` - I think this is possible but I would have to check

I am not sure what would be the preferred approach would be here. If the config.yml file shouldn't be changed before testing then naturally I can adjust it. Cheers

slavalamp: oh yeah good point, defaulting to a page that would exist with the dev data makes sense

setting a user works page by default as an example for people making their own archives is probably a good idea, but idk if option 1 or 2 is better (probably not 3, i don't think using the .rb files to override these configs like that is done)

maybe ask bilka since he made the ticket?

Bilka2: Not a full review, but yes please change the URL here to the URL desired in the issue without further changes to dev data. It will simply 404 in dev, which isn't an issue - this is not a commonly used page so I don't expect a significant impact to the experience there.

It vastly simplifies the timing/organisation of testing and deploying if we don't have to fit in config changes between that, so we prefer to use production-ready values in this file

nicolacleary: Clear, thanks both! I have updated the config as asked

app/controllers/skins_controller.rb
Bilka2: That's fine, but thanks for noting it!

app/controllers/skins_controller.rb
Bilka2: Could you use `helpers.link_to` to create this link HTML so that we can avoid the `html_safe` call? If you need to remove the `role='button'` for that, that's fine, it should be removed per [AO3-6765](https://otwarchive.atlassian.net/browse/AO3-6765) anyway.

Bilka2: Speaking of AO3-6765, the PR for it has been in "Reviewed: Action Needed" without changes for over a year, which makes it [adoptable](https://github.com/otwcode/otwarchive/wiki/Commit-Policy#user-content-Stalled_Pull_Requests_and_Jira_Issues) for AD&T volunteers like me. However, if I adopted it now, we'd have loads of merge conflicts between the changes here and the changes for AO3-6765.

Would you be up for adding AO3-6765 into the PR here, since you're already doing the i18n it should just be a matter of changing the text in the locale file and changing the construction of the flash here to not be an array (looks like we usually do multiline flashes with `<br />`)? If yes, you have my official AD&T blessing to adopt AO3-6765

Bilka2 (changes requested):
This is a full review now, but it's just one thing :D

app/controllers/skins_controller.rb
nicolacleary: I'll look into using the helper as per your first comment.

As for [AO3-6765](https://otwarchive.atlassian.net/browse/AO3-6765) - I am happy to add that functionality into this PR or a separate one - seems like it's not that complicated in terms of what would change.


I am, however, not quite sure what the procedure is for this. I would rather avoid using the old PR since it has a lot of unrelated linting changes and, given its age, it would take some effort to bring up to date with master (of course it can be done though if necessary) .

Would it be okay to just recreate the code changes in a new commit and update the PR title and description as necessary to add the additional ticket name and credit?

Bilka2: > Would it be okay to just recreate the code changes in a new commit and update the PR title and description as necessary to add the additional ticket name and credit?

Yup!

nicolacleary:
The force push was to correct the Jira ticket in some of the commit messages since I got them mixed up. From what I can tell it is not recommended to do this after a review but I decided it would be alright since I was only targeting commits that hadn't been reviewed yet.
Please let me know if I shouldn't do that going forward. Cheers

app/controllers/skins_controller.rb
```diff
@@ -134,13 +134,13 @@ def preview
...
-    flash[:notice] << ts("Tip: You can preview any archive page you want by tacking on '?site_skin=[skin_id]' like you can see in the url above.")
-    flash[:notice] << "<a href='#{skin_path(@skin)}' class='action' role='button'>".html_safe + ts("Return To Skin To Use") + "</a>".html_safe
-    tag = FilterCount.where("public_works_count BETWEEN 10 AND 20").random_order.first.filter
-    redirect_to tag_works_path(tag, site_skin: @skin.id)
+    flash[:notice] = [
+      t(".skin_title", title: @skin.title),
+      t(".remove_skin"),
+      t(".tip", site_skin_id: @skin.id),
+      helpers.link_to(t(".return_to_skin"), skin_path(@skin), class: "action")
+    ].join('<br />')
```
nicolacleary: Changes in flash message:
<details>

Flash before:
Image

Flash using `<br />` rather than `<ul>`:
Image

Flash using `<br />` rather than `<ul>` *and* wrapping it in `<p>`:
Image

</details>

[AO3-6765](https://otwarchive.atlassian.net/browse/AO3-6765) says:
> The “Return To Skin To Use” link ... should ideally be in a `<p>`, not a `<li>`. The role="button" can also be removed, because it is not acting like a button, just a link.

AFAIK I could just do `view_context.content_tag(:p, <rest of the line>)` and it would work (see screenshots above), but it's a little unclear for me what the desired look of this flash is, given that there are quite a few changes here now. I decided to leave it is as now since it's no longer part of a list.

nicolacleary: As far as I can tell, a `.html_safe` doesn't impact what is rendered in this case, so I omitted it.

Bilka2: thanks for showing the behaviour! I think it's fine without the `<p>` for now, though we may get feedback during testing to change it

Bilka2: We're happy about every one of those we can remove, thank you!

Bilka2 (approved):
The force push is fine. Thanks for the changes!

PR #5748, opened 2026-04-21
AO3-3396 Syns of fandom are now under fandom in stats page

ReverM:
## Issue

https://otwarchive.atlassian.net/browse/AO3-3396

## Purpose

This makes synonyms of fandom be grouped with the parent fandom in the stats page of an user (i.e work A in fandom 1, work B in fandom 2, 1 is a synonym of 2, we see A and B under 2 and we don't see 1)

## Credit

Danaël / Rever ( they / he )
Danaël Villeneuve on Jira

### DIFF
--- a/app/controllers/stats_controller.rb
+++ b/app/controllers/stats_controller.rb
@@ -17,9 +17,18 @@ class StatsController < ApplicationController
     user_chapters = Chapter.joins(pseuds: :user).where(users: { id: @user.id }).where(posted: true)
     work_query = user_works
       .joins(:taggings)
-      .joins("inner join tags on ((taggings.tagger_id = tags.id AND tags.merger_id IS NULL) OR EXISTS (SELECT * FROM tags AS tags2 WHERE tags.id = tags2.merger_id AND tags2.id = taggings.tagger_id)) AND tags.type = 'Fandom'")
+      .joins(<<~SQL.squish)
+        INNER JOIN tags
+          ON ((taggings.tagger_id = tags.id AND tags.merger_id IS NULL)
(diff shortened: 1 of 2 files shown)


app/controllers/stats_controller.rb
```diff
@@ -17,7 +17,16 @@ def index
     user_chapters = Chapter.joins(pseuds: :user).where(users: { id: @user.id }).where(posted: true)
     work_query = user_works
       .joins(:taggings)
-      .joins("inner join tags on taggings.tagger_id = tags.id AND tags.type = 'Fandom'")
+      .joins(<<~SQL.squish)
+        INNER JOIN tags
+          ON ((taggings.tagger_id = tags.id AND tags.merger_id IS NULL)
+            OR EXISTS ( /* If the tag has a synonym, we take the synonym instead */
+              SELECT * FROM tags AS tags2
```
pmonfort: The query looks great! Can you change the `tags2` name to something more descriptive? Maybe something like `synonyms`, what do you think?

ReverM: Sounds good!

pmonfort:
Looks good! The synonym grouping logic is correct, and the test covers the core case well. The CI failure (cucumber features/bookmarks) is unrelated to your changes.

Thanks for renaming tags2 to synonyms, much clearer now 👍

sarken:
Thank you for taking the time to work on this! Unfortunately, AO3-3396 is a bit of an older issue, and after some discussion, we've determined it's no longer in line with our needs. We're going to decline this pull request and close the issue. We appreciate the work you've done and sincerely apologize for the inconvenience.

PR #5752, opened 2026-04-22
AO3-3764 Paginate tag set nominations review page

pmonfort:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-3764

## Purpose

When a tag set has more than 30 unreviewed nominations, moderators currently see a random subset with the message "There are too many nominations to show at once." This makes it impossible to systematically review all nominations.

This PR replaces the random selection with proper pagination using pagy:

- Paginate nominations by unique tagname, grouping duplicates (same tag nominated by multiple users) into a single entry
- Each tag type (fandom, character, relationship, freeform) paginates independently with its own page param
- Heading shows "X - Y of Z [Type]" when paginated, or "[Type] (N left to review)" when everything fits on one page
- Pagination links include anchors to scroll to the correct fieldset
- Order by `created_at` ascending for deterministic results
- Removed the random selection logic and the "too many nominations" flash message

## Testing Instructions

1. Log in as a moderator of a tag set with more than 25 unreviewed nominations
2. Go to Review Nominations
3. Verify pagination links appear and each tag type paginates independently
4. Verify that navigating pages does not affect other tag types (e.g. going to fandom page 2 keeps freeform on page 1)
5. Verify duplicate nominations (same tagname from different users) appear only once
6. Verify approving/rejecting tags updates counts correctly
7. Verify tag sets with fewer than 25 nominations show no pagination links and display the "(N left to review)" heading
8. Verify the old "There are too many nominations to show at once" message no longer appears

## Credit

Pablo Monfort (he/him)

### DIFF
--- a/app/views/tag_set_nominations/_review.html.erb
+++ b/app/views/tag_set_nominations/_review.html.erb
@@ -38,7 +38,7 @@
           <%= check_all_none("Approve All", "Approve None", "approve") %>
           <% if @paginations&.dig(tag_type) %>
             <%== pagy_nav(@paginations[tag_type],
-                   anchor_string: "id=\"#{tag_type}_nominations\"") %>
+                   fragment: "##{tag_type}_nominations") %>
           <% end %>
           <ol class="tags index group">
             <% if tag_type == "fandom" %>
@@ -53,7 +53,7 @@
           </ol>
           <% if @paginations&.dig(tag_type) %>
             <%== pagy_nav(@paginations[tag_type],
-                   anchor_string: "id=\"#{tag_type}_nominations\"") %>
+                   fragment: "##{tag_type}_nominations") %>
           <% end %>
         </fieldset>
       <% end %>
--- a/app/views/tag_set_nominations/_review.html.erb
+++ b/app/views/tag_set_nominations/_review.html.erb
@@ -30,12 +30,22 @@
 
     <% %w(cast fandom character relationship freeform).each do |tag_type| %>
       <% if @nominations[tag_type] && @nominations[tag_type].count > 0 %>
-        <fieldset class="<%= tag_type %> listbox group">
-          <% heading = tag_type == "cast" ? ts("Already Approved Fandoms") : ts("#{tag_type_label_name(tag_type).pluralize} (%{count} left to review)", :count => @tag_set.send("#{tag_type}_nominations").unreviewed.count) %>
+        <fieldset class="<%= tag_type %> listbox group" id="<%= tag_type %>_nominations">
+          <% if tag_type == "cast" %>
+            <% heading = ts("Already Approved Fandoms") %>
+          <% else %>
+            <% pagy = @paginations[tag_type] %>
+            <% if pagy.pages > 1 %>
+              <% heading = ts("%{from} - %{to} of %{total} %{type}", from: pagy.from, to: pagy.to, total: pagy.count, type: tag_type_label_name(tag_type).pluralize) %>
+            <% else %>
+              <% heading = ts("%{type} (%{count} left to review)", type: tag_type_label_name(tag_type).pluralize, count: pagy.count) %>
+            <% end %>
+          <% end %>
           <legend><%= heading %></legend>
           <h3 class="heading"><%= heading %></h3>
           <%= check_all_none("Reject All", "Reject None", "reject") %>
           <%= check_all_none("Approve All", "Approve None", "approve") %>
+          <%== pagy_nav(@paginations[tag_type], anchor_string: "id=\"#{tag_type}_nominations\"") if @paginations&.dig(tag_type) %>
           <ol class="tags index group">
             <% if tag_type == "fandom" %>
               <%= render "review_fandoms", :fandoms => @nominations[:fandom] %>
@@ -47,6 +57,7 @@
               <% end %>
             <% end %>
           </ol>
+          <%== pagy_nav(@paginations[tag_type], anchor_string: "id=\"#{tag_type}_nominations\"") if @paginations&.dig(tag_type) %>
         </fieldset>
       <% end %>
     <% end %>
--- a/app/views/tag_set_nominations/_review.html.erb
+++ b/app/views/tag_set_nominations/_review.html.erb
@@ -35,17 +35,24 @@
             <% heading = ts("Already Approved Fandoms") %>
           <% else %>
             <% pagy = @paginations[tag_type] %>
+            <% type_name = tag_type_label_name(tag_type).pluralize %>
             <% if pagy.pages > 1 %>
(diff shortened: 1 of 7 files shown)


app/views/tag_set_nominations/_review.html.erb
```diff
@@ -30,12 +30,16 @@
...
-          <% heading = tag_type == "cast" ? ts("Already Approved Fandoms") : ts("#{tag_type_label_name(tag_type).pluralize} (%{count} left to review)", :count => @tag_set.send("#{tag_type}_nominations").unreviewed.count) %>
+        <fieldset class="<%= tag_type %> listbox group" id="<%= tag_type %>_nominations">
+          <% heading = nomination_review_heading(tag_type, @paginations[tag_type]) %>
           <legend><%= heading %></legend>
           <h3 class="heading"><%= heading %></h3>
           <%= check_all_none("Reject All", "Reject None", "reject") %>
           <%= check_all_none("Approve All", "Approve None", "approve") %>
+          <% if @paginations&.dig(tag_type) %>
+            <%== pagy_nav(@paginations[tag_type],
+                   anchor_string: "id=\"#{tag_type}_nominations\"") %>
```
sarken: Question! Does the `anchor_string` here mean that the pagination links should be like `?fandom_page=3#fandom_nominations`?

I swear I saw that behavior when I first pulled this code to try it out, but it has disappeared on me, so I'm confused. 😆

pmonfort: Yes, that's the intended behavior! The anchor_string parameter name is a bit misleading, it refers to the HTML anchor element (<a> tag) not URL anchors/fragments. I've pushed a fix using the correct fragment parameter.
Good catch! Thanks!

features/tag_sets/tag_set_nominations.feature
```diff
@@ -145,20 +145,39 @@ Feature: Nominating and reviewing nominations for a tag set
...
+  When I review nominations for "Nominated Tags"
+  Then I should see "Next" within "fieldset.freeform .pagination"
+
+  Scenario: Owner of a tag set with over 30 nominations does not see the randomized selection message
+  Given I am logged in as "tagsetter"
+    And I set up the nominated tag set "Nominated Tags" with 6 fandom noms and 6 character noms
+  When there are 36 unreviewed nominations
+  Given I am logged in as "tagsetter"
+    And I review nominations for "Nominated Tags"
+  Then I should not see "There are too many nominations to show at once"
```
sarken: I'd be okay with removing this test. We can just confirm that the message has been removed during manual testing; we don't need to test it every time we run the tests for the rest of time. 😆

spec/controllers/tag_set_nominations_controller_spec.rb
```diff
@@ -284,6 +267,62 @@ def add_unreviewed_freeform_nominations(num)
...
+                  let(:fandom_nom_num) { 0 }
+                  let(:fandom_nom_status) { :unreviewed }
+                  let(:noms_per_page) { ArchiveConfig.ITEMS_PER_PAGE }
+
+                  before do
+                    unique_tagnames = noms_per_page * 3
+                    unique_tagnames.times do |i|
+                      nominator = create(:user)
+                      tsn = TagSetNomination.create!(owned_tag_set: owned_tag_set, pseud: nominator.default_pseud)
+                      FandomNomination.create!(tag_set_nomination: tsn, tagname: "Paginated Fandom #{i}")
```
sarken: Could we use factories for the tag set nominations and tag nominations? I think it would be something like:
```
                      tsn = create(:tag_set_nomination, owned_tag_set: owned_tag_set, pseud: nominator.default_pseud)
                      create(:tag_nomination, tag_set_nomination: tsn, tagname: "Paginated Fandom #{i}", type: "FandomNomination")
```

spec/controllers/tag_set_nominations_controller_spec.rb
```diff
@@ -174,38 +167,28 @@ def add_unreviewed_freeform_nominations(num)
                     get :index, params: { tag_set_id: owned_tag_set.id }
                   end
 
-                  context 'unreviewed fandom nominations <= 30' do
-                    let(:fandom_nom_num) { 30 }
+                  context "unreviewed fandom nominations within per page limit" do
+                    let(:fandom_nom_num) { 20 }
```
sarken: Is there a reason it needs to be 20? If so, it would be a good idea to document why we do that instead of something like the `per_page + 1` we do elsewhere.

sarken (changes requested):
Just a couple of nitpicks/questions about the tests! Aside from that, this looks good and I'm excited. 🎉

features/tag_sets/tag_set_nominations.feature
pmonfort: That makes sense, I tend to over testing, TDD habit, my bad. Will remove it.

spec/controllers/tag_set_nominations_controller_spec.rb
pmonfort: Make sense, I will use Factories 👍

spec/controllers/tag_set_nominations_controller_spec.rb
pmonfort: Good call, updated to use per_page to match the pattern, thanks!

spec/controllers/tag_set_nominations_controller_spec.rb
```diff
@@ -284,16 +280,85 @@ def add_unreviewed_freeform_nominations(num)
...
+                    expect(page1_tagnames & page2_tagnames).to be_empty
+                  end
+
+                  it "page 3 returns the remaining unique tagnames" do
+                    get :index, params: { tag_set_id: owned_tag_set.id, fandom_page: 3 }
+                    page3_tagnames = assigns(:nominations)[:fandom].map(&:tagname)
+
+                    expect(page3_tagnames.length).to eq(noms_per_page)
+                    expect(page3_tagnames).to eq(page3_tagnames.uniq)
+                  end
```
sarken: I might be missing something here. How does this tell us these are definitely the unique tagnames for page 3 and not the same tags from page 1 or 2? (It's okay if it doesn't. In my opinion, we can remove it instead of expanding it, since the test above already demonstrates the tagnames are unique. If we do that, it might also mean we can go from `noms_per_page * 3` to `2`.)

sarken (commented):
Just one last question!

spec/controllers/tag_set_nominations_controller_spec.rb
pmonfort: Make sense, 2 pages is enough to validate what we need.

sarken (approved):
🎉

PR #5756, opened 2026-04-24
AO3-7345 Update success message for setting a site skin for a session

jmalark:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-7345

## Purpose

Context: When you successfully set a site skin for a session 

The success message has been expanded to add a link to the skin page and clarify how long the skin will stay/that it will survive closing the browser.

## Credit
Jesse Malark he/him


config/locales/controllers/en.yml
```diff
@@ -309,6 +309,9 @@ en:
     index:
       public_site_page_title: Public Site Skins
       public_work_page_title: Public Work Skins
+    set:
+      failure: Sorry, but only certain skins can be used this way (for performance reasons). Please drop a support request if you'd like %{title} to be added!
+      success: You're now using the %{title} skin. This will last for 2 weeks even if you close your browser. If you'd like to keep it longer, go to the <a href="/skins/%{id}">%{title} skin page</a> and select the "Use" button at the bottom of the page.
```
jmalark: Questions for whoever reviews this: can you give me any context as to when we should be using the skin_path function? It seemed inefficient to search by title when we already had access to the id, but I wasn't sure if maybe it's intended to like stop stuff from breaking if we ever change the file path structure or if maybe sometimes we don't actually have the id at this point or if it prevents some programmers from mixing up relative vs root based links, etc.            Also it's probably a pretty basic question but can you explain what leads us to using .html_safe ? Is it just links we are treating as buttons or when might I need/not need to use that? Thanks :)

slavalamp: i don't see why `skin_path` would do an extra search by title? you already have the `@skin` object here with the id, so you'd pass it to `skin_path` and it should use that id

`.html_safe` marks that the string has trusted html that should be parsed as html (= no user-inputted stuff), but in translation strings you should instead add `_html` to the name of the string (e.g. `success_html`). this link works without that here because flash messages also all get marked as `html_safe` on top of that in `application_helper.rb`, but it's better to mark it anyway for consistency and in case the flash message rendering code gets updated (honestly i feel it's kinda weird that it just marks all flash messages as safe and not the individual strings, but seems like that line is 15 years old which would explain it - the archive code was less polished back then)

also avoid putting html tags in locale files - here you can make the link a variable (named `skin_page_link`), with `"%{title} skin page"` being its own translation string (`skin_page`)

(see the [HTML in translations](https://github.com/otwcode/otwarchive/wiki/Internationalization-%28i18n%29-Standards#html-in-translations) i18n standards)

features/other_b/skin_public.feature
```diff
@@ -14,7 +14,7 @@ Feature: Public skins
     And the skin "public skin" is in the chooser
   When I am logged in as "skinner"
     And I press "public skin"
-  Then I should see "The skin public skin has been set. This will last for your current session."
+  Then I see "You're now using the public skin skin. This will last for 2 weeks even if you close your browser. If you'd like to keep it longer, go to the public skin skin page and select the \"Use\" button at the bottom of the page." on the page
```
slavalamp: i don't think there's a reason to add a new step here, `I should see` works fine here. were you confused about it having the selector part in the step definition? the selector is optional and it looks in the whole page by default

(and if you were to add a new step, the steps for things that should happen should be phrased with "should")

also you don't need to use the full text of a long string in tests, i'm not sure if there's exactly an guideline for that, but long lines get harder to read, and if accidental changes to the text happened, they should get caught by humans (and you're also testing for this notice in the spec test anyway). this test is more to confirm that the page has a notice with text in it. i think just the beginning, "You're now using the public skin skin. This will last for 2 weeks" would be enough here

slavalamp (changes requested):
hi, friendly review from a fellow contributor!

config/locales/controllers/en.yml
jmalark: Hi slavalamp. Thanks for taking a look at this.

So I had mistakenly thought skin_path was getting defined on lines 160-161 in paths.rb 
`    when /^"(.*)" skin page/
      skin_path(Skin.find_by(title: $1))`

but it's really just getting used there. I'm very new to Ruby. Where does something like skin_paths get defined?


I tried to add _html now so it's .success_html  and I started using the link_to helper and skin_path. Does the new naming style follow what you were saying or was I misunderstanding something?

features/other_b/skin_public.feature
jmalark: So when I had been trying to match on that really long string, the old step wouldn't catch it. So I had made a new step with that {string} part. But since I don't have to check that full string, the old step catches it fine and I've removed the step I had added.

config/locales/controllers/en.yml
techlingyi: Also a newcomer but I think most paths are defined in config/routes.rb through the Ruby on Rails syntax - there's a good guide [here](https://guides.rubyonrails.org/routing.html):

> For example, adding `resources :photos` to the route file will generate these `_path` helpers:
> Path Helper | Returns URL
> `photos_path` | `/photos`
> `new_photo_path` | `/photos/new`
> `edit_photo_path(:id)` | `/photos/:id/edit`
> `photo_path(:id)` | `/photos/:id`

So `skin_path(:id)` is defined implicitly through the "resources :skins" line in config/routes.rb:604.

jmalark: Awesome thank you :)

slavalamp: (sorry it took me a bit to get back to this)

the translation string keys for the messages are good, though i meant for the skin page link variable to be named specifically skin_**page**_link, since it has the words "skin page" in it. besides that, you just need to make the `"#{@skin.title} skin page"` text a translatable string now

also i'm pretty sure you should be able to use `link_to` without adding the `helpers.` prefix

i'm handwriting this example so i could've made a typo and it's definitely not formatted right - you can use rubocop to figure out the indents - but it will likely look approximately like this:

```rb
flash[:notice] = t(".success_html",
skin_title: @skin.title,
skin_page_link: link_to(t(".skin_page", skin_title: @skin.title), skin_path(@skin))
)
```

```yml
set:
  skin_page: "%{skin_title} skin page"
```

slavalamp: ...i'm seeing that i was wrong about `link_to`, this has now taught me to double check things before saying them >< sorry about that

jmalark: No worries :)   I should have the changes squashed soon.

slavalamp:
this looks good to me now, thank you 👍🏻

(no idea what's up with the failing test but looks unrelated)

sarken (approved):

PR #5765, opened 2026-04-28
AO3-7393 AO3-7395 Fix blank/invalid pseud editing behavior

techlingyi:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-7393 - Trying to change a pseud's name to an invalid one makes it so you can't update the pseud again

https://otwarchive.atlassian.net/browse/AO3-7395 - Trying to change a pseud name to an empty one results in error 500

## Purpose

Reloads the pseud's original ID if the new pseud is invalid in some way, so that the Show and Update buttons on the edit pseud form are no longer broken upon entering an invalid pseud. This also fixes the 500 error that appears with a blank pseud. An instance variable is used to keep the user's attempted pseud name in the text box (because it would revert back to the original otherwise).

## Credit

Ling-Yi (any pronouns)

### DIFF
--- a/app/controllers/pseuds_controller.rb
+++ b/app/controllers/pseuds_controller.rb
@@ -63,6 +63,7 @@ class PseudsController < ApplicationController
 
   # GET /pseuds/1/edit
   def edit
+    @textbox_name = params[:id] # default value in the form's name textfield
     @pseud = @user.pseuds.find_by!(name: params[:id])
     authorize @pseud if logged_in_as_admin?
   end
@@ -108,6 +109,8 @@ class PseudsController < ApplicationController
       flash[:notice] = t(".successfully_updated")
       redirect_to([@user, @pseud])
     else
+      @textbox_name = @pseud.name
+      @pseud.assign_attributes({ name: params[:id] })
       render action: "edit"
     end
   end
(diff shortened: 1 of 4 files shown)


techlingyi:
Wow, I haven't seen _that_ Rubocop error before! (A fluke, maybe? If anyone could rerun it that would be nice — it only flagged one error when I ran it the first time, on an identical branch I closed because the commit history was getting confusing.)

Edit: Thank you! :D

pmonfort:
This is a good workaround to fix the issue, but I think there might be a simpler solution. Rails is using the Pseud [to_param](https://github.com/otwcode/otwarchive/blob/4a69ddd5c032ce0d483b65c74c7aa46cb58b38c1/app/models/pseud.rb#L103) method to generate the form action and the "Show" link based on the pseud `name`. The problem is that after a failed update, it returns the in-memory (invalid) name instead of the persisted one. If you update that method to use `name_in_database` (an ActiveRecord method that returns the value currently stored in the database), that would fix both issues without needing the extra `@textbox_name` variable or any view changes. Thanks for your work, let me know your thoughts!

techlingyi:
@pmonfort That worked perfectly with the tests, thank you! Should be good to go now :)

pmonfort (approved):
@techlingyi this is looking great, thanks for addressing the comments 🚀

PR #5768, opened 2026-04-29
AO3-7179 Collection participant messages should list both pseud and username

jmalark:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-7179

## Purpose
I updated collection_participants_controller to use byline so that message alerts put both pseud name and username in the messages (instead of just the pseud)

## Credit
Jesse Malark he/him

### DIFF
--- a/app/controllers/collection_participants_controller.rb
+++ b/app/controllers/collection_participants_controller.rb
@@ -73,16 +73,16 @@ class CollectionParticipantsController < ApplicationController
 
   def update
     if @participant.update(collection_participant_params)
-      flash[:notice] = t('collection_participants.update_success', default: "Updated %{participant}.", participant: @participant.pseud.name)
+      flash[:notice] = t(".success", participant: @participant.pseud.byline)
     else
-      flash[:error] = t(".failure", participant: @participant.pseud.name)
+      flash[:error] = t(".failure", participant: @participant.pseud.byline)
     end
     redirect_to collection_participants_path(@collection)
   end
 
   def destroy
     @participant.destroy
-    flash[:notice] = t('collection_participants.destroy', default: "Removed %{participant} from collection.", participant: @participant.pseud.name)
+    flash[:notice] = t(".success", participant: @participant.pseud.byline)
     redirect_back_or_to root_path
   end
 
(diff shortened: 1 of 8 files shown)


app/controllers/collection_participants_controller.rb
```diff
@@ -73,16 +73,16 @@ def index
...
-      flash[:error] = t(".failure", participant: @participant.pseud.name)
+      flash[:error] = t(".failure", participant: @participant.pseud.byline)
     end
     redirect_to collection_participants_path(@collection)
   end
 
   def destroy
     @participant.destroy
-    flash[:notice] = t('collection_participants.destroy', default: "Removed %{participant} from collection.", participant: @participant.pseud.name)
+    flash[:notice] = t(".success", participant: @participant.pseud.byline)
```
jmalark, line 85: So I did go very slightly beyond the scope of the Jira issue to also fix the messages for removing a participant and failing to update a participant. They were both also only using the "pseud name" and not the "pseud name (username)" style and were both in this same file/were similar messages. Do we need to update the Jira testing instructions so those two message fixes also get tested as part of this or how do we handle slight scope creep like this?

sarken: Good catch! I'll update the testing instructions to cover removing a participant. I think we can rely on automated tests for failing to update a participant. 

However, since we're updating the removal success message, we should also update the confirmation prompt for removal: https://github.com/otwcode/otwarchive/blob/52738054e3bbfe4829cad1493d2eace9cce0caa1/app/views/collection_participants/_participant_form.html.erb#L13-L14

Could you make that say, "Are you certain you want to remove pseud (username)?" (We usually say "sure" instead of "certain," so we might as well fix that.)

sarken (changes requested):

app/controllers/collection_participants_controller.rb
jmalark: Thanks :) Updated the confirmation prompt to say "sure" instead of "certain" and use pseud (username) instead of just pseud

sarken (approved):
Thank you!

PR #5771, opened 2026-04-29
AO3-7189 Allow admins to change user skin via the Preferences page

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-7189

## Purpose

Allow support and superadmins to edit a user's skin via their preferences.

## Credit

Bilka


app/controllers/preferences_controller.rb
```diff
@@ -22,7 +22,7 @@ def update
     @available_skins = (@user.skins.site_skins + Skin.approved_skins.site_skins).uniq
     @available_locales = Locale.where(email_enabled: true)
 
-    @user.preference.attributes = preference_params
+    @user.preference.attributes = permitted_attributes(@preference)
```
redsummernight, line 25: With this change, we're moving the parameter list for regular users into a Pundit policy class. We do this in one other place in the code base, [PseudPolicy](https://github.com/otwcode/otwarchive/blob/52738054e3bbfe4829cad1493d2eace9cce0caa1/app/policies/pseud_policy.rb#L15-L22).

I'm surprised this isn't more common, but it makes sense because so far we either deal with objects that are only editable by admins (like admin settings) or have separate controllers to edit user data (users, user creations).

redsummernight (approved):

app/controllers/preferences_controller.rb
Bilka2: Yes, it's based on https://github.com/varvet/pundit/tree/v2.3.1#strong-parameters. In the current version of pundit the method names have slightly changed but the general idea of having the parameters in the policy is the same

PR #5776, opened 2026-05-01
AO3-7388 Redirect to user skin page with error if trying to preview a skin you can't use

zrei:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-7388

## Purpose

Redirects to the user skins page when trying to preview a skin that can't be used but can be viewed while logged in as a user. This includes parent-only site skins and work skins. For skins that a logged-in user cannot view (not owned and not public), they are redirected to their dashboard with an error.

Changed preview to a user-only action. Redirects to the login page with an error when trying to preview any skin while logged out. Redirects to the root page with an error when trying to preview any skin while logged in as an admin.

## Testing Instructions
For site skins, refer to Jira.

For work skins:
1. Log in
2. Make a work skin
    a. Hi, username! > My Dashboard > Skins > Create Work Skin
    b. Fill in required information
    c. Press “Submit” to save
3. In your browser address bar, add /preview to the end of the skin URL and press Return

For not owned, not public skins:
1. Log in
2. Make a site skin
    a. Hi, username! > My Dashboard > Skins > Create Site Skin
    b. Fill in required information
    c. Press “Submit” to save
3. Copy the site skin URL
4. Log into another account and paste the site skin URL into the browser address bar. Add /preview to the end of the URL and press Return

## Credit

zrei (she/they)

### DIFF
--- a/app/controllers/skins_controller.rb
+++ b/app/controllers/skins_controller.rb
@@ -3,7 +3,7 @@ class SkinsController < ApplicationController
   before_action :load_skin, except: [:index, :new, :create, :unset]
   before_action :check_ownership_or_admin, only: [:edit, :update]
   before_action :check_ownership, only: [:confirm_delete, :destroy]
-  before_action :check_visibility, only: [:show]
+  before_action :check_visibility, only: [:show, :preview]
   before_action :check_editability, only: [:edit, :update, :confirm_delete, :destroy]
 
   #### ACTIONS
@@ -129,9 +129,10 @@ class SkinsController < ApplicationController
 
   # Get /skins/1/preview
   def preview
-    if (@skin.is_a?(WorkSkin) || !@skin.approved_or_owned_by?(current_user) || @skin.unusable?)
+    if (@skin.is_a?(WorkSkin) || @skin.unusable?)
       flash[:error] = t("skins.preview.cannot_preview")
-      redirect_to user_skins_path(current_user) and return
+
+      redirect_to skin_path(@skin) and return
     end
 
     flash[:notice] = []
--- a/app/controllers/skins_controller.rb
+++ b/app/controllers/skins_controller.rb
@@ -129,6 +129,11 @@ class SkinsController < ApplicationController
 
   # Get /skins/1/preview
   def preview
+    if (@skin.is_a?(WorkSkin) || !@skin.approved_or_owned_by?(current_user) || @skin.unusable?)
+      flash[:error] = t("skins.preview.cannot_preview")
+      redirect_to user_skins_path(current_user) and return
+    end
+
     flash[:notice] = []
     flash[:notice] << ts("You are previewing the skin %{title}. This is a randomly chosen page.", title: @skin.title)
     flash[:notice] << ts("Go back or click any link to remove the skin.")
--- a/app/controllers/skins_controller.rb
+++ b/app/controllers/skins_controller.rb
@@ -129,7 +129,7 @@ class SkinsController < ApplicationController
 
   # Get /skins/1/preview
   def preview
-    if (@skin.is_a?(WorkSkin) || @skin.unusable?)
+    if @skin.is_a?(WorkSkin) || @skin.unusable?
       flash[:error] = t("skins.preview.cannot_preview")
 
       redirect_to skin_path(@skin) and return
--- a/app/controllers/skins_controller.rb
+++ b/app/controllers/skins_controller.rb
@@ -135,7 +135,6 @@ class SkinsController < ApplicationController
       redirect_to skins_path(skin_type: "Site") and return if current_user.nil?
 
       redirect_to user_skins_path(current_user) and return
-      end
     end
 
     flash[:notice] = []
--- a/app/controllers/skins_controller.rb
+++ b/app/controllers/skins_controller.rb
@@ -130,9 +130,13 @@ class SkinsController < ApplicationController
   # Get /skins/1/preview
   def preview
     if @skin.is_a?(WorkSkin) || @skin.unusable?
-      flash[:error] = t("skins.preview.cannot_preview")
+      flash[:error] = t(".cannot_preview")
 
-      redirect_to skin_path(@skin) and return
+      if current_user.nil?
+        redirect_to skins_path(skin_type: "Site") and return
+      else
+        redirect_to user_skins_path(current_user) and return
+      end
     end
 
     flash[:notice] = []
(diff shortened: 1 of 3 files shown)


zrei:
I added a `check_visibility` call before `preview` to catch users trying to preview skins they can't view. If I should instead extend the check I've added to the `preview` action in the skins controller, do let me know!

app/controllers/skins_controller.rb
```diff
@@ -129,6 +129,12 @@ def update
 
   # Get /skins/1/preview
   def preview
+    if @skin.is_a?(WorkSkin) || @skin.unusable?
+      flash[:error] = t("skins.preview.cannot_preview")
+
+      redirect_to skin_path(@skin) and return
```
Bilka2: I believe this should redirect to the user skin page per the Jira issue (so the user's skins index page), not the page of that specific skin

app/controllers/skins_controller.rb
```diff
@@ -129,6 +129,12 @@ def update
 
   # Get /skins/1/preview
   def preview
+    if @skin.is_a?(WorkSkin) || @skin.unusable?
+      flash[:error] = t("skins.preview.cannot_preview")
```
Bilka2: Please use [lazy lookup](https://guides.rubyonrails.org/i18n.html#lazy-lookup) for the translation here

spec/controllers/skins_controller_spec.rb
```diff
@@ -622,4 +622,92 @@
       end
     end
   end
+
+  describe "GET #preview" do
+    let(:skin_creator) { create(:user) }
+    subject { get :preview, params: { id: skin.id } }
+    let(:success) { it_redirects_to_with_error(skin_path(skin), "Sorry, you can't preview that skin.") }
```
Bilka2: I'm a bit hesitant calling this a success when it's an error, it feels a bit like "task failed successfully", which yes it did, but that's a bit confusing. Maybe `cant_preview` or something is better? Though I also wouldn't mind if you decided to copy pasted the error instead of making it a variable since I find that a bit clearer, whatever you prefer

spec/controllers/skins_controller_spec.rb
```diff
@@ -622,4 +622,92 @@
...
+      end
+
+      context "when workskin is not public" do
+        let(:skin) { create(:work_skin, title: "Work Skin", author: skin_creator) }   
+        
+        it_behaves_like "a non-public skin that cannot be previewed"
+      end
+    end
+
+    context "with unusable site skin" do
```
Bilka2:
```suggestion
    context "with parent only site skin" do
```
this is purely personal preference because I don't know what `unusable` means, but I sure know what parent only means, and this way the test mentions both

spec/controllers/skins_controller_spec.rb
```diff
@@ -622,4 +622,92 @@
...
+        it_behaves_like "a public skin that cannot be previewed"
+      end
+
+      context "when site skin is not public" do
+        let(:skin) { create(:skin, title: "Unusable Site Skin", unusable: true, author: skin_creator) }
+
+        it_behaves_like "a non-public skin that cannot be previewed"
+      end
+    end
+  end
```
Bilka2, line 821: To cover the (very welcome!) change regarding visibility for the preview action, could you add a variant for an official site skin that is fully accessible and one for a previewable skin by a user with `it_behaves_like "an action only the skin author can access"`? That way we know that the happy path still works

spec/controllers/skins_controller_spec.rb
```diff
@@ -622,4 +622,92 @@
...
   end
+
+  describe "GET #preview" do
+    let(:skin_creator) { create(:user) }
+    subject { get :preview, params: { id: skin.id } }
+    let(:success) { it_redirects_to_with_error(skin_path(skin), "Sorry, you can't preview that skin.") }
+  
+    shared_examples "a public skin that cannot be previewed" do
+      
+      context "when logged in as the skin creator" do
```
Bilka2, line 669:
```suggestion
    shared_examples "a public skin that cannot be previewed" do
      context "when logged in as the skin creator" do
```
as a style thing, here and for the other shared example

Bilka2 (changes requested):
Thank you for working on this! Two minor things about the code and then some requests for testing additions

app/controllers/skins_controller.rb
zrei: Updated! For logged-out users, it redirects to the public site skins page. Will update the PR description.

spec/controllers/skins_controller_spec.rb
zrei: The redirect action for the shared example `"an action only the skin author can access"` is different from the redirect for `preview`, so I recreated it for the previewable skin by a user.

Bilka2: Missed that detail, thanks for the better test changes!

spec/controllers/skins_controller_spec.rb
```diff
@@ -622,4 +622,164 @@
...
+      context "when site skin is not public" do
+        let(:skin) { create(:skin, title: "Parent Only Site Skin", unusable: true, author: skin_creator) }
+
+        it_behaves_like "a non-public skin that cannot be previewed"
+      end
+    end
+
+    context "with accessible site skin" do
+      let(:notice) { ["You are previewing the skin #{skin.title}. This is a randomly chosen page.", "Go back or click any link to remove the skin.", "Tip: You can preview any archive page you want by tacking on '?site_skin=[skin_id]' like you can see in the url above.", "<a href='#{skin_path(skin)}' class='action' role='button'>Return To Skin To Use</a>"] }
+      let(:success) { it_redirects_to_with_notice(tag_works_path(tag, site_skin: skin.id), notice) }
```
Bilka2: if you wanted to simplify this, you could skip checking the notice with `it_redirects_to_simple`. It's checked in a cucumber test, so we'd be fine skipping it here

app/controllers/skins_controller.rb
```diff
@@ -129,6 +129,14 @@ def update
 
   # Get /skins/1/preview
   def preview
+    if @skin.is_a?(WorkSkin) || @skin.unusable?
+      flash[:error] = t(".cannot_preview")
+
+      redirect_to skins_path(skin_type: "Site") and return if current_user.nil?
```
Bilka2: Sorry to change the requirements under you, but you bringing up that we can't redirect to a user's page if they're not logged in made me ask whether we even want to allow guests and admins to preview any skins at all. Turns out we don't! I've updated the Jira issue accordingly, so please update the code to make this action `users_only`

Bilka2 (changes requested):

sarken (approved):
Thank you!

PR #5780, opened 2026-05-03
AO3-7409 Adjust header and page subtitle for managing invitations

jmalark:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-7409

## Purpose

On the manage invitations page, adjusts the browser subtitle and the header

## Credit
Jesse Malark he/him

### DIFF
--- a/app/controllers/invitations_controller.rb
+++ b/app/controllers/invitations_controller.rb
@@ -20,6 +20,7 @@ class InvitationsController < ApplicationController
   end
 
   def manage
+    @page_subtitle = t(".page_title", username: @user.login)
     status = params[:status]
     @invitations = @user.invitations
     if %w(unsent unredeemed redeemed).include?(status)
--- a/app/views/invitations/manage.html.erb
+++ b/app/views/invitations/manage.html.erb
@@ -1,5 +1,12 @@
 <!--Descriptive page name, messages and instructions-->
-<h2 class="heading"><%= logged_in_as_admin? ? (@user.login + "'s") : "Your" %> Invitations</h2>
+<h2 class="heading">
+  <% if logged_in_as_admin? %>
+    <%= t(".page_heading.admin", username: @user.login) %>
+  <% else %>
+    <%= t(".page_heading.user") %>
+  <% end %>
+</h2>
+
 <!--/descriptions-->
 
 <!--subnav -->
--- a/spec/controllers/invitations_controller_spec.rb
+++ b/spec/controllers/invitations_controller_spec.rb
@@ -48,6 +48,23 @@ describe InvitationsController do
 
         expect(response).to render_template("manage")
       end
+      
+      it "assigns subtitle with username in it" do
+        admin.update!(roles: ["policy_and_abuse"])
+        fake_login_admin(admin)
+        get :manage, params: { user_id: user.login }
+
+        expect(assigns[:page_subtitle]).to eq("#{user.login} - Manage Invitations")
+      end
+    end
+
+    context "when not an admin" do
+      it "assigns subtitle with username in it" do
+        fake_login_known_user(user)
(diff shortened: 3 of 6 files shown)


app/controllers/invitations_controller.rb
```diff
@@ -20,6 +20,7 @@ def index
   end
 
   def manage
+    @page_subtitle = t(".page_subtitle", username: @user.login)
```
sarken: Let's stick to our normal naming convention here and use `page_title`.

app/views/invitations/manage.html.erb
```diff
@@ -1,5 +1,5 @@
 <!--Descriptive page name, messages and instructions-->
-<h2 class="heading"><%= logged_in_as_admin? ? (@user.login + "'s") : "Your" %> Invitations</h2>
+<h2 class="heading"><%= t(logged_in_as_admin? ? ".admin_header" : ".header", username: @user.login) %></h2>
```
sarken: For admins, let's go with "Manage Invitations for %{usernames}". I'll update the issue.

To keep close to our usual naming convention, let's use the keys `page_heading.admin` and `page_heading.user`

sarken (changes requested):

app/controllers/invitations_controller.rb
jmalark: renamed :)

app/views/invitations/manage.html.erb
jmalark: Updated

spec/controllers/invitations_controller_spec.rb
```diff
@@ -48,6 +48,22 @@
 
         expect(response).to render_template("manage")
       end
+      it "assigns subtitle with username in it" do
```
sarken, line 52: Tiny nitpick, but as long as we need another commit to finish getting the heading key straightened out, could you add a blank line above this?

app/views/invitations/manage.html.erb
```diff
@@ -1,5 +1,5 @@
 <!--Descriptive page name, messages and instructions-->
-<h2 class="heading"><%= logged_in_as_admin? ? (@user.login + "'s") : "Your" %> Invitations</h2>
+<h2 class="heading"><%= t(logged_in_as_admin? ? ".page_heading_admin" : ".page_heading_user", username: @user.login) %></h2>
```
sarken: Can you make these page_heading.admin and page_heading.user, please? It should look like this in the locale file:
```
manage:
  page_heading:
    admin: text
    user: text
```

And apologies for not catching this before, but the current code will define the `username` variable even when we don't need it. It's not terrible, but it would be better to do something like 
```
<h2 class="heading">
  <% if logged_in_as_admin? %>
    <%= t(".page_heading.admin", username: @user.login) %>
  <% else %>
    [etc]
```

That would also mean `i18n-tasks unused -l en` should pass without needing to add anything to `ignore_unused`.

sarken (changes requested):

app/views/invitations/manage.html.erb
jmalark: If we are adding an if / else condition, should I also be making the text more like displaying base text for user and base text + admin text for admin or would trying to name those things make the code less clear?

sarken (approved):
Thank you!

app/views/invitations/manage.html.erb
sarken: Whoops, sorry, didn't see this question before I approved the pull request! But in general, it's a good idea to avoid doing that. Just because "Manage Invitations" and "for username" combine nicely into "Manage Invitations for username" English doesn't mean it will work across all languages. It's easier for the translation team to have the versions entirely separate like we've done here.

PR #5721, opened 2026-04-11
AO3-5100  List all Open Challenges with pagination and filtering

ayasayadi1:
# Pull Request Checklist

* [X] Have you read ["How to write the perfect pull request"](https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/)?
* [X] Have you read the [contributing guidelines](https://github.com/otwcode/otwarchive/blob/master/CONTRIBUTING.md)?
* [X] Have you added [tests for any changed functionality](https://github.com/otwcode/otwarchive/wiki/Automated-Testing)?
* [X] Have you added the [Jira](https://otwarchive.atlassian.net) issue number
  as the *first* thing in your pull request title (e.g. `AO3-1234 Fix thing`)
* [X] Do you have fewer than 5 pull requests already open? If not, please wait
  until they are reviewed and merged before creating new pull requests.

## Issue

https://otwarchive.atlassian.net/browse/AO3-5100

## Purpose

**Merge Gift Exchange and Prompt Meme lists**

`list_challenges` runs two separate queries (one for Gift Exchanges and one for Prompt Memes), concatenates and caps the results at 15. It now runs a single query across both challenge types, paginated and sorted by closing date.

**Include challenges with no closing date**

The open challenges query filters by `signups_close_at > now`, which silently excludes challenges with no closing date. The filter now matches challenges closing in the future or challenges with no closing date set.

**Filter Open Challenges**

Open Challenges can now be filtered by _title_, _tag_, _multifandom_, _closed_, _moderated_, and _collection type_ (the "_No Challenge_" option is hidden since it's all challenges)

Gift Exchange Challenges and Prompt Meme Challenges support the same filters, except _collection type_.

## Credit

Aya Sayadi

PR #5788, opened 2026-05-05
AO3-7143 Adding filter to user collection page

ReverM:
## Issue

https://otwarchive.atlassian.net/browse/AO3-7143

## Purpose

This adds the ability to filter collections on user pages like on the generic collections page.

## Credit

Danaël / Rever (they / he)

### DIFF
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -195,7 +195,7 @@ class CollectionsController < ApplicationController
   private
 
   def collection_filter_params
-    params.permit(:commit, collection_search: [
+    params.permit(:commit, :user_id, collection_search: [
       :title, :challenge_type, :moderated, :multifandom, :closed, :tag,
       :sort_column, :sort_direction
     ])[:collection_search] || {}
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -43,8 +43,9 @@ class CollectionsController < ApplicationController
       flash_search_warnings(@collections)
       @page_subtitle = t(".subcollections_page_title", collection_title: @collection.title)
     elsif params[:user_id]
+      @sort_and_filter = true
       @user = User.find_by!(login: params[:user_id])
-      @search = CollectionSearchForm.new({ maintainer_id: @user.id, sort_column: "title.keyword" }.merge(page: params[:page]))
+      @search = CollectionSearchForm.new(collection_filter_params.merge({ maintainer_id: @user.id, sort_column: "title.keyword" }.merge(page: params[:page])))
       @collections = @search.search_results.scope(:for_search)
       flash_search_warnings(@collections)
       @page_subtitle = ts("%{username} - Collections", username: @user.login)
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -45,7 +45,7 @@ class CollectionsController < ApplicationController
     elsif params[:user_id]
       @sort_and_filter = true
       @user = User.find_by!(login: params[:user_id])
-      @search = CollectionSearchForm.new(collection_filter_params.merge({ maintainer_id: @user.id, sort_column: "title.keyword" }.merge(page: params[:page])))
+      @search = CollectionSearchForm.new(collection_filter_params.merge({ maintainer_id: @user.id }.merge(page: params[:page])))
       @collections = @search.search_results.scope(:for_search)
       flash_search_warnings(@collections)
       @page_subtitle = ts("%{username} - Collections", username: @user.login)
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -30,6 +30,7 @@ class CollectionsController < ApplicationController
   end
 
   def index
+    options = params[:collection_search].present? ? collection_filter_params : {}
     if params[:work_id]
       @work = Work.find(params[:work_id])
       @collections = @work.approved_collections
@@ -45,13 +46,13 @@ class CollectionsController < ApplicationController
     elsif params[:user_id]
       @sort_and_filter = true
       @user = User.find_by!(login: params[:user_id])
-      @search = CollectionSearchForm.new(collection_filter_params.merge({ maintainer_id: @user.id }.merge(page: params[:page])))
+      @search = CollectionSearchForm.new(options.merge({ maintainer_id: @user.id }.merge(page: params[:page])))
       @collections = @search.search_results.scope(:for_search)
       flash_search_warnings(@collections)
       @page_subtitle = ts("%{username} - Collections", username: @user.login)
     else
       @sort_and_filter = true
-      @search = CollectionSearchForm.new(collection_filter_params.merge(page: params[:page]))
+      @search = CollectionSearchForm.new(options.merge(page: params[:page]))
       @collections = @search.search_results.scope(:for_search)
       flash_search_warnings(@collections)
     end
@@ -195,10 +196,10 @@ class CollectionsController < ApplicationController
   private
 
   def collection_filter_params
-    params.permit(:commit, :user_id, collection_search: [
+    params.require(:collection_search).permit(
       :title, :challenge_type, :moderated, :multifandom, :closed, :tag,
       :sort_column, :sort_direction
-    ])[:collection_search] || {}
+    )
   end
 
   def collection_params
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -45,7 +45,7 @@ class CollectionsController < ApplicationController
     elsif params[:user_id]
       @sort_and_filter = true
       @user = User.find_by!(login: params[:user_id])
-      @search = CollectionSearchForm.new(collection_filter_params.merge({ maintainer_id: @user.id }.merge(page: params[:page])))
+      @search = CollectionSearchForm.new(collection_filter_params.merge({ maintainer_id: @user.id }.merge(page: params[:page]), default_sort_column: "title.keyword"))
       @collections = @search.search_results.scope(:for_search)
       flash_search_warnings(@collections)
       @page_subtitle = ts("%{username} - Collections", username: @user.login)
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -45,7 +45,7 @@ class CollectionsController < ApplicationController
     elsif params[:user_id]
       @sort_and_filter = true
       @user = User.find_by!(login: params[:user_id])
-      @search = CollectionSearchForm.new(collection_filter_params.merge({ maintainer_id: @user.id }.merge(page: params[:page]), default_sort_column: "title.keyword"))
+      @search = CollectionSearchForm.new(collection_filter_params.merge({ maintainer_id: @user.id, default_sort_column: "title.keyword" }.merge(page: params[:page])))
       @collections = @search.search_results.scope(:for_search)
       flash_search_warnings(@collections)
       @page_subtitle = ts("%{username} - Collections", username: @user.login)
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -45,7 +45,7 @@ class CollectionsController < ApplicationController
     elsif params[:user_id]
       @sort_and_filter = true
       @user = User.find_by!(login: params[:user_id])
-      @search = CollectionSearchForm.new(collection_filter_params.merge({ maintainer_id: @user.id, default_sort_column: "title.keyword" }.merge(page: params[:page])))
+      @search = CollectionSearchForm.new(collection_filter_params.merge({ maintainer_id: @user.id }.merge(page: params[:page])))
       @collections = @search.search_results.scope(:for_search)
       flash_search_warnings(@collections)
       @page_subtitle = ts("%{username} - Collections", username: @user.login)
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -58,7 +58,7 @@ module ApplicationHelper
   end
 
   def page_has_filters?
-    @facets.present? || (controller.controller_name == "collections" && @sort_and_filter) || (controller.action_name == 'unassigned' && controller.controller_name == 'fandoms')
+    @facets.present? || (controller.controller_name == "collections" && @sort_and_filter) || (controller.action_name == "unassigned" && controller.controller_name == "fandoms")
   end
 
   # This is used to make the current page we're on (determined by the path or by the specified condition) a span with class "current" and it allows us to add a title attribute to the link or the span
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -19,10 +19,11 @@ module ApplicationHelper
   def classes_for_main
     class_names = controller.controller_name + '-' + controller.action_name
 
-    show_sidebar = (!@hide_dashboard && (@user || @admin_posts || @collection || show_wrangling_dashboard))
-    class_names += " dashboard" if show_sidebar
-
-    class_names += " filtered" if page_has_filters?
+    # Rails/HelperInstanceVariable is disabled for the two following lines given this warning is raised for reusing helpers. This function is only used once in application.html.erb
+    # Whether or not the sidebar should be shown
+    class_names += " dashboard" if !@hide_dashboard && (@user || @admin_posts || @collection || show_wrangling_dashboard) # rubocop:disable Rails/HelperInstanceVariable
+    # Whether or not the page will have filters
+    class_names += " filtered" if @facets.present? || (controller.controller_name == "collections" && @sort_and_filter) || (controller.action_name == "unassigned" && controller.controller_name == "fandoms") # rubocop:disable Rails/HelperInstanceVariable
 
     case controller.controller_name
     when "abuse_reports", "feedbacks", "known_issues"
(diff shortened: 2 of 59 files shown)


ReverM:
https://github.com/otwcode/otwarchive/pull/5487 original PR. Reopened the PR since merging upstream made it quite angry. Apologies for the added complexity.

ReverM:
@sarken pinging you since you commented in the previous PR

ReverM:
It seems that the spec controller failure is not expected behaviour after this PR?

app/views/collections/index.html.erb
```diff
@@ -13,47 +13,49 @@
...
-  <% unless logged_in? && @user && @user == current_user %>
-    <li><%= span_if_current ts("Collections"), collections_path %></li>
-    <li><%= link_to ts("Open Challenges"), list_challenges_collections_path %></li>
-  <% end %>
-  <% if logged_in? %>
-    <% # Logged in user on own collections index gets links for user Collections and Manage Collection Items %>
-    <% if @user && @user == current_user %>
-      <li><%= span_if_current ts("Collections"), user_collections_path(@user) %></li>
-      <li><%= link_to ts("Manage Collection Items"), user_collection_items_path(@user) %></li>
+<div class="navigation actions module">
```
sarken: We only put our subnavigation in `div` like this in rare instances where we have complex navigation with multiple separate sets of actions, like two lists, or a list and a form outside the list. The `div` on `works/index.html.erb` is there because we can have both a form and a list... but I think it's wrong, actually, because they shouldn't be there at the same time.

I'm pretty sure the reason you've added this is to make the navigation take up the full width of `#main` and push the filters into the right position on the page. Interestingly, the bookmarks index doesn't have this `div`, nor does it have that problem -- and after some investigating, I figured out why. Both bookmark and work index pages leave the empty `ol.index` on the page, while the collection index removes it with the `unless @collections.blank?` logic on line 52. Empty list elements aren't great for accessibility, so we don't want to replicate that here, either.

I think the solution we need here is to add a `clear` to `form.filters`. That should keep the filters in the right position without either the empty list or the `div`.

app/views/collections/index.html.erb
```diff
@@ -13,47 +13,49 @@
...
+  <h3 class="landmark heading"><%= t(".navigation.landmark_heading") %></h3>
+  <ul class="navigation actions" role="navigation">
```
sarken: You can remove the `role="navigation"` while you're here -- it's not valid on list elements.

app/views/collections/index.html.erb
```diff
@@ -13,47 +13,49 @@
...
 
 <% unless @collections.blank? %>
   <!--pagination here-->
   <%= will_paginate @collections %>
 
   <!--main content-->
-  <h3 class="landmark heading"><%= ts("List of Collections") %></h3>
-  <ul class="collection picture index group">
+  <h3 class="landmark heading"><%= t(".main_content.landmark_heading") %></h3>
+  <ul class="collection picture index group <%= "no-filter" unless @sort_and_filter %>">
```
sarken: Instead of adding a class, let's take a look at what's causing the problem here. (But just a quick note for the future that we have [some rules for adding classes](https://otwcode.github.io/docs/classes.html#rules).) 

It looks like all collection index pages get the `filtered` class, regardless of whether they have filters, thanks to this:

https://github.com/otwcode/otwarchive/blob/7f6d5b133269aa5fa27fd064214e1fb95095403d/app/helpers/application_helper.rb#L60-L62

https://github.com/otwcode/otwarchive/blob/7f6d5b133269aa5fa27fd064214e1fb95095403d/app/helpers/application_helper.rb#L19-L25

Unfortunately, collections don't use `@facets`, so we can't just remove the collections logic from `page_has_filters?`. But we can tweak it to `(controller.controller_name == "collections" && @sort_and_filter)` or even just `@sort_and_filter`, since collections are the only place we use that variable, and I think it's reasonable to reserve it so it's only ever used to determine whether we display filters. #5721 also uses that variable for filtered challenge pages.

As a bonus, this approach should also fix the issue where `/works/:id/collections` pages

sarken (changes requested):
I checked on the failing spec, and it looks like it's failing because we lost alphabetical sorting on user collection pages. That's something we definitely want to keep; users weren't happy when we accidentally switched to created_at sorting.

app/views/collections/index.html.erb
ReverM: I was trying to find the root cause but couldn't quite find it. Thanks!

app/views/collections/index.html.erb
ReverM: It seems that stylesheet 17 is taking priority over stylesheet 18. Regardless of if the index has filtered or not. The `width: 100%` seems to be the cause of the issue in that case? It is defined to be 75% in sheet 18 within line 9 but the one in style sheet 17 is applied first. Adding `!important` to the one in sheet 18 seems to fix it? But I wasn't sure if such a solution was desired.

sarken: Which page and device/browser, and which `width: 100%`?  After actioning [this comment](https://github.com/otwcode/otwarchive/pull/5788#discussion_r3193523160), the only `width: 100%` I can find in 17 is `.dashboard.works-index h4.landmark, .dashboard.works-index ol.pagination`, and I'm not running into issues on either the user collection index or work index, but I've only checked in desktop Safari.

 https://github.com/otwcode/otwarchive/commit/0c99f89e959180e7fa4ca86a278edae61f1c0974 has the changes I'm testing with.

ReverM: Oh I see, I got confused and reintroduced the class in page 17.

ReverM:
> I checked on the failing spec, and it looks like it's failing because we lost alphabetical sorting on user collection pages. That's something we definitely want to keep; users weren't happy when we accidentally switched to created_at sorting.

It seems that the default sort column is defined to be created_at in  `collection_search_form.rb`. Should we switch to `title.keyword`?

sarken:
We just want to change the sort column on the user collection pages, not everywhere, so I don't think that's the right place to do it. Does restoring `sort_column: "title.keyword"` in the controller not work here?

sarken:
No rush, just noting that I put the label back to Reviewed: Action Needed since there were still test failures.

ReverM:
I found the issue for the cucumber test, waiting for response for the rubocop failure.

app/helpers/application_helper.rb
```diff
@@ -19,10 +19,11 @@ def link_to_function(name, *args, &block)
...
 
-    show_sidebar = (!@hide_dashboard && (@user || @admin_posts || @collection || show_wrangling_dashboard))
-    class_names += " dashboard" if show_sidebar
-
-    class_names += " filtered" if page_has_filters?
+    # Rails/HelperInstanceVariable is disabled for the two following lines given this warning is raised for reusing helpers. This function is only used once in application.html.erb
+    # Whether or not the sidebar should be shown
+    class_names += " dashboard" if !@hide_dashboard && (@user || @admin_posts || @collection || show_wrangling_dashboard) # rubocop:disable Rails/HelperInstanceVariable
+    # Whether or not the page will have filters
+    class_names += " filtered" if @facets.present? || (controller.controller_name == "collections" && @sort_and_filter) || (controller.action_name == "unassigned" && controller.controller_name == "fandoms") # rubocop:disable Rails/HelperInstanceVariable
```
sarken: Please keep the `page_has_filters?` helper. It's much more readable even if it means having a method that isn't currently reused. We also have some other pull requests that add additional logic to the `page_has_filters?` method, and incorporating that would make this line even longer and harder to read.

features/support/paths.rb
```diff
@@ -169,6 +169,8 @@ def path_to(page_name)
       edit_skin_path(Skin.find_by(title: $1), wizard: true)
     when /^the new collection page/
       new_collection_path
+    when /^(.*?)(?:'s)? collections page$/i
```
sarken, line 172: When we define a new path for an index that draws from Elasticsearch, we also like to include `step %{all indexing jobs have been run}` to make sure the index is fully up to date when we access it. That way, we don't have to remember to add so many `all indexing jobs have been run` steps throughout the feature files themselves.

features/collections/collection_browse.feature
```diff
@@ -287,6 +287,79 @@ Feature: Collection
...
+    And I post the work "Private 1" in the collection "Privates"
+    And I lock the work "Private 1"
+    And I post the work "Private 2" in the collection "Privates"
+    And I lock the work "Private 2"
+    And I bookmark the work "Private 1" to the collection "Publics"
+    And I bookmark the work "Private 2" to the collection "Publics"
+    And I post the work "Public 1" in the collection "Publics"
+    And I bookmark the work "Public 1" to the collection "Privates"
+    And all indexing jobs have been run
+    And I change my username to "Owner"
```
sarken: Why are we changing the username? It should be fine to use the collection owner's existing name (which is `moderator`).

features/collections/collection_browse.feature
```diff
@@ -287,6 +287,79 @@ Feature: Collection
     And the 2nd collection result should contain "Privates"
     And the 2nd collection result should contain "Bookmarked Items: 1"
 
+Scenario: Sort collections by Works and Bookmarks from user's collection
+
+  Given I have a collection "Privates"
+    And I have a collection "Publics"
+    And a set of collections for searching
```
sarken: What is this data being used for?

features/collections/collection_browse.feature
```diff
@@ -287,6 +287,79 @@ Feature: Collection
...
+  When I go to the collections page
+    Then I should not see "2 Collections"
```
sarken: I think these steps can be removed. They seem more like troubleshooting the set up for this scenario than something we should be testing here.

features/collections/collection_browse.feature
```diff
@@ -287,6 +287,79 @@ Feature: Collection
...
+    And I bookmark the work "Private 2" to the collection "Publics"
+    And I post the work "Public 1" in the collection "Publics"
+    And I bookmark the work "Public 1" to the collection "Privates"
+    And all indexing jobs have been run
+    And I change my username to "Owner"
+    And I go to Owner's collections page
+    And I select "Works" from "Sort by"
+    And I select "Descending" from "Sort direction"
+    And I press "Sort and Filter"
+  Then the 1st collection result should contain "Privates"
```
sarken, line 250: While we're checking the results of our filtering, it would be a good idea to check that the select menus display the proper options, e.g. `Then "Works" should be selected within "Sort by"`. I think it's fine to just do this once and not for logged out, logged in, and admin.

features/collections/collection_browse.feature
```diff
@@ -287,6 +287,79 @@ Feature: Collection
     And the 2nd collection result should contain "Privates"
     And the 2nd collection result should contain "Bookmarked Items: 1"
 
+Scenario: Sort collections by Works and Bookmarks from user's collection
```
sarken: I've left a couple of comments on lines in this scenario, but it looks like it's basically a duplicate of the scenario above it; the main difference appears to be that we are explicitly setting the sort direction to descending. 

I'd suggest converting the previous test to a scenario outline, adding steps to specify the sort direction, and using the examples `the collection page` and `moderator's collection page`. It should cover the same things while adding fewer lines of code.

features/search/filters.feature
```diff
@@ -363,3 +363,21 @@ Feature: Filters
     When I fill in "work_search_query" with "bad~query!!!"
       And I press "Sort and Filter"
     Then I should see "Your search failed because of a syntax error"
+
+  @collections
```
sarken: We don't usually put tags before individual scenarios unless the tag is pulling double duty as a hook (e.g., `@javascript` and `@load-default-skin`).

features/search/filters.feature
```diff
@@ -363,3 +363,21 @@ Feature: Filters
...
+  Scenario: Filtering on a user collection page should only return collections by that user.
```
sarken: I think this scenario would be better in `collection_browse.feature` since it doesn't need the `Background` steps at the top of this file. 

However, like redsummernight said, I think we can also just live without this.

features/collections/collection_browse.feature
redsummernight: Since this issue adds the user collection page, we want to test that such a page only shows collections belonging to the user. You covered this in the other feature file, with "Filtering on a user collection page should only return collections by that user".

On the other hand, I think we can skip this scenario. It's the same as "Scenario: Sort collections by Works and Bookmarks". It also doesn't deal with multiple collection owners, just the same one that got renamed.

features/search/filters.feature
```diff
@@ -363,3 +363,21 @@ Feature: Filters
...
+    Given a collection "Duplicate Name" with name "collection1" owned by "meatloaf"
+      And a collection "Duplicate Name" with name "collection2" owned by "recengine"
+      And a collection "The Hobbits" with name "collectionhobbit" owned by "recengine"
+      And a collection "Not Mine!" with name "yours" owned by "iminvisible"
+      And all indexing jobs have been run
+    When I go to recengine's collections page
+      And I fill in "Filter by title" with "Duplicate Name"
+      And I press "Sort and Filter"
+    Then I should see "1 Collection"
```
redsummernight: You can add checks for the titles that should show up, "1nd collection result should contain" and so on.

features/search/filters.feature
```diff
@@ -363,3 +363,21 @@ Feature: Filters
...
+    When I follow "Clear Filters"
+    Then I should see "2 Collections"
+    When I fill in "Filter by title" with "Not Mine!"
+      And I press "Sort and Filter"
+    Then I should not see "iminvisible"
```
redsummernight: This "should not see" step can be skipped since there's already a better "no collections found" check that covers the same thing.

app/controllers/collections_controller.rb
```diff
@@ -43,8 +43,9 @@ def index
       flash_search_warnings(@collections)
       @page_subtitle = t(".subcollections_page_title", collection_title: @collection.title)
     elsif params[:user_id]
+      @sort_and_filter = true
       @user = User.find_by!(login: params[:user_id])
-      @search = CollectionSearchForm.new({ maintainer_id: @user.id, sort_column: "title.keyword" }.merge(page: params[:page]))
+      @search = CollectionSearchForm.new(collection_filter_params.merge({ maintainer_id: @user.id, default_sort_column: "title.keyword" }.merge(page: params[:page])))
```
redsummernight: Can we use `sort_column` instead of a new parameter?

sarken (changes requested):

app/controllers/collections_controller.rb
ReverM: `sort_column` would override whatever sorting the user tries to make. It cannot be used

app/helpers/application_helper.rb
ReverM: Got it, I will edit the comment accordingly too

features/collections/collection_browse.feature
ReverM: The sort direction is specified in this one because the collection page default to date_updated + descending, while the user collection is title.keyword + ascending

features/search/filters.feature
ReverM: Should this scenario be removed then?

app/models/search/collection_search_form.rb
```diff
@@ -71,7 +71,7 @@ def sort_options
   end
 
   def default_sort_column
-    "created_at"
+    options[:default_sort_column] ||= "created_at"
```
sarken: To red's point about not adding a new parameter, would something like `options[:maintainer_id] ? "title.keyword" : "created_at"` work here instead of setting the parameter in the controller?

app/controllers/collections_controller.rb
```diff
@@ -194,7 +195,7 @@ def destroy
   private
 
   def collection_filter_params
-    params.permit(:commit, collection_search: [
+    params.permit(:commit, :user_id, collection_search: [
```
sarken: Do you know when the `user_id` parameter is used? I didn't notice it in the URL and it seemed to work without this addition, but it was only a quick test while I was looking into red's suggestion.

app/models/search/collection_search_form.rb
ReverM: That does work

app/controllers/collections_controller.rb
ReverM: I'm not certain, but from my understanding, it seems to be used on initialization? Since `user_id` is present if we are on a specific user's collection page, it will be passed for initialization.

app/controllers/collections_controller.rb
ReverM: I have adressed this by sarken's suggestion

app/controllers/collections_controller.rb
redsummernight: In other places, for example the work controller, we use `.permit` only for the search form parameters:

https://github.com/otwcode/otwarchive/blob/19d32d9aa73615509c222a443b20e1ec2347595c/app/controllers/works_controller.rb#L932-L935

Here we use `.permit` at the root `params`, which means everything we use needs to be explicitly permitted:

- `user_id`, for the user lookup (`params[:user_id`),
- `commit`, which is the form submit button.

For the sake of consistency, let's follow the work controller example and use `params.require(:collection_search).permit()`, omitting the "outer" params.

(On Rails 8, we probably [should be using `.expect` instead](https://api.rubyonrails.org/v8.1/classes/ActionController/Parameters.html), but that's for another time.)

ReverM: Done

redsummernight (approved):
Looks good to me! I'd wait for sarken's review before we set the "Ready to Merge" label.

app/controllers/collections_controller.rb
sarken: This cleared up the last of my questions/concerns, so I will mark this ready to merge. Thank you both!

sarken (approved):

PR #5791, opened 2026-05-06
AO3-7307 Created Resque job to clean up old user audits

hunternet93:
## Issue

https://otwarchive.atlassian.net/browse/AO3-7307

## Purpose

This adds a Resque job that will delete audit records older than a configurable limit. A new admin setting is added to exempt particular usernames from deletion.

## Credit

Hunter Ada Smith (she/they)


app/jobs/audits_cleanup_job.rb
```diff
@@ -0,0 +1,42 @@
...
+          created_at: ..ArchiveConfig.USER_KEEP_AUDIT_CREATES_DESTROYS_DAYS.days.ago
+        ).and(
+          Audited.audit_class.where.not(
+            auditable_id: preserve_user_ids
+          )
+        )
+      )
+    end
+
+    query.limit(DELETE_LIMIT).delete_all
```
pmonfort: The job runs monthly and only deletes 10,000 records per run. Since the ticket mentions the audits table is very large, the initial cleanup could take many months to catch up. Would it make sense to delete in batches with a small sleep between them to avoid blocking the queue, maybe including a max batch count as a safeguard? It could also be worth increasing the DELETE_LIMIT, maybe to 50,000. What do you think?

app/policies/admin_setting_policy.rb
```diff
@@ -5,6 +5,7 @@ class AdminSettingPolicy < ApplicationPolicy
   # Define which roles can update which settings.
   ALLOWED_SETTINGS_BY_ROLES = {
     "policy_and_abuse" => %i[
+      preserve_audit_records_usernames
```
pmonfort: The ticket mentions this setting should be accessible to "superadmins, Policy & Abuse admins, and Legal admins." It's added to `policy_and_abuse` and `superadmin` here, but legal is missing.

pmonfort (commented):
Code looks good, great work! Left a couple of comments

db/migrate/20260506200220_add_preserve_audit_records_usernames_admin_setting.rb
```diff
@@ -0,0 +1,5 @@
+class AddPreserveAuditRecordsUsernamesAdminSetting < ActiveRecord::Migration[8.1]
+  def change
+    add_column :admin_settings, :preserve_audit_records_usernames, :string
```
Bilka2: Based on https://github.com/otwcode/otwarchive/pull/4492#discussion_r1196605638, this means that the cache key for the admin settings should be updated, or there will be issues during deploy

app/views/admin/settings/index.html.erb
```diff
@@ -28,6 +28,9 @@
 
       <dt><%= f.label :days_to_purge_unactivated, t(".fields.days_to_purge_unactivated") %></dt>
       <dd><%= admin_setting_text_field(f, :days_to_purge_unactivated, size: "3") %></dd>
+
+      <dt><%= f.label :preserve_audit_records_usernames, t(".fields.preserve_audit_records_usernames") %></dt>
+      <dd><%= admin_setting_text_field(f, :preserve_audit_records_usernames, size: "20") %></dd>
```
Bilka2: As per the updated Jira issue description, please hide the setting itself from admin roles who do not have access to it, so that they can't access what it is set to.

app/jobs/audits_cleanup_job.rb
hunternet93: I updated the job to delete in batches of 5,000 up to a limit of 50,000. I don't have much DBA experience, but hopefully that should avoid locking up the table during execution. The limits are now configurable at runtime, so the job could be manually enqueued with a higher limit to clear out the backlog during off hours.

pmonfort: Sounds good, thanks for the update! The batching approach looks solid 👏

pmonfort:
Thanks for addressing my comments, the PR looks good to me. Now we only need Bilka2's approval 👍

app/jobs/audits_cleanup_job.rb
```diff
@@ -0,0 +1,54 @@
...
+    query_delete_count = 0
+    job_delete_count = 0
+
+    loop do
+      query_delete_count = query.delete_all
+      break unless query_delete_count == query_delete_limit
+
+      job_delete_count += query_delete_count
+      break unless job_delete_count < job_delete_limit
+    end
```
Bilka2, line 51: Instead of doing this loop, please make this job a batched job. We already have existing structures to support this, for example the recent backfill task for user_past_emails and user_past_usernames which took its data from the audits is based on the RedisSetJob class. Additionally, please use a separate new queue for these jobs. Since this PR will be a difficult deploy and the jobs will run regularly, a separate queue allows us to throttle the jobs by how many workers we assign to the queue - or even pause them completely by not assigning any workers. Additionally, the task and batch size for the job should be configurable in config.yml (for the defaults, the ones that are in RedisSetJob are fine)

Generally, this loop here would tie up a single worker for a long time, which would be a disaster for the utilities queue which is generally expected to be processed very quickly. One of its purposes is that it enqueues jobs for other queues (scheduled reindex jobs). So if we put something long running in the utilities queue, we risk backing up other asynchronous processes that are vital for the user-facing parts of the archive, like work reindexing

app/models/admin_setting.rb
```diff
@@ -19,6 +21,7 @@ class AdminSetting < ApplicationRecord
     invite_from_queue_frequency: ArchiveConfig.INVITE_FROM_QUEUE_FREQUENCY,
     account_creation_enabled?: ArchiveConfig.ACCOUNT_CREATION_ENABLED,
     days_to_purge_unactivated: 2,
+    preserve_audit_records_usernames: nil,
```
Bilka2: Is this required?

app/views/admin/settings/index.html.erb
```diff
@@ -28,6 +28,11 @@
 
       <dt><%= f.label :days_to_purge_unactivated, t(".fields.days_to_purge_unactivated") %></dt>
       <dd><%= admin_setting_text_field(f, :days_to_purge_unactivated, size: "3") %></dd>
+
+      <% unless admin_setting_disabled?(:preserve_audit_records_usernames) %>
+        <dt><%= f.label :preserve_audit_records_usernames, t(".fields.preserve_audit_records_usernames") %></dt>
+        <dd><%= admin_setting_text_field(f, :preserve_audit_records_usernames, size: "20") %></dd>
```
Bilka2: PAC requested a few adjustments for this exclusion list. I will use this comment here to list them all, even though they're not all related to the view itself:
* The list should be a list of user ids, not usernames, to prevent issues when a user renames
* If the list is changed, saving the settings should validate that it results in valid users (so no ids that have no associated user, no incorrect separators etc) and should an appropriate error if it's invalid. If the list is not changed, the validation should not run, to prevent issues where an admin wants to change another setting but this list has become invalid because a user deleted their account and then the admin is stuck trying to figure out how to fix the list. Please add an automated test to validate this, since it's quite critical that we can adjust the admin settings in a pinch
* parsing of the comma separation of the input should allow using a comma both with or without a space after it (`123,654` and `123, 654` should both be valid)
* here in the view there should be an extra line above the setting that lists currently excluded users
  * generally the users should be listed as `USERNAME (ID), USERNAME (ID)`. the username should link to the user's admin dashboard
  * if the list includes a user that has been deleted, the user should be listed as `Deleted user (ID)` and not be linked

With those considerations, the UI should be this:

Users currently excluded from audit record cleanup: username (id), username (id), Deleted user (id), username (id)

Exclude these user ids from audit record cleanup (comma separated): [field containing id, id, id, id]

config/resque_schedule.yml
```diff
@@ -122,3 +122,9 @@ disable_admin_post_comments:
   description: >-
     Disables all comments on admin (news) posts older than the
     configured window.
+
+cleanup_audits:
+  cron: "0 0 1 * *"
```
Bilka2: I'm not a cron expert, but if this could be defined to run at always the same time, that would be nice, as that makes things more predictable for PAC. we have a number of tasks around 6-7-something am (refer to schedule.rb) so a time sometime before or after that would probably work

app/policies/admin_setting_policy.rb
```diff
@@ -1,10 +1,11 @@
 class AdminSettingPolicy < ApplicationPolicy
   # Defines the roles that allow admins to view all settings.
-  SETTINGS_ROLES = %w[policy_and_abuse superadmin support tag_wrangling].freeze
+  SETTINGS_ROLES = %w[policy_and_abuse superadmin support tag_wrangling legal].freeze
```
Bilka2, line 3: I am quite surprised no tests had to be changed for this addition - please add rspec tests for the controller that confirm that legal now has access to the settings and that the three roles can adjust preserve_audit_records_usernames (and potentially that it is not cleared when an unauthorized admin changes other settings)

app/views/admin/settings/index.html.erb
Bilka2: Apologies for changing all these requirements under you, I hope it's still doable for you. If not, please let us know and we can take over

Bilka2 (changes requested):

app/jobs/audits_cleanup_job.rb
hunternet93: I looked at the `RedisSetJob` class but I'm not sure how it would work for this use case. It seems to take a set of items and split it into batches for jobs to act on. However, the audits cleanup job doesn't have an efficient way to get that list of IDs.

This job currently runs a `DELETE FROM audits WHERE ... LIMIT n` query to delete old records from the audit table. Each individual query has a limit of 5,000 rows to keep query runtime down, and the job has a 50,000-row limit to keep the job runtime down as well.

To batch using `RedisSetJob`, it'd instead need to run a `SELECT FROM audits WHERE ...` query without a limit to get a set of all row IDs to be deleted. Given the table's massive size, this initial query would probably be quite slow and cause the batching part of the job to be the longest part of this process. It would also be somewhat inefficient compared to a `DELETE WHERE` query that deletes matching rows as they are found.

Would a tail recursion setup work better? Each job run could delete up to `JOB_DELETE_LIMIT` rows, then enqueue another instance of the job and terminate. If a job runs out of rows to delete before hitting its limit, it will terminate without recursing.

Alternately, I could refactor the job to use `User.find_in_batches` and then query and delete audits for each user separately, similar to `AuditsBackfillJob`. However, I think that would end up being pretty inefficient for a scheduled job, since after the first full run it would end up processing lots of users that don't have any old audit table entries.

app/models/admin_setting.rb
hunternet93: Nope, removed.

app/views/admin/settings/index.html.erb
hunternet93: I updated the admin settings page as described, with an exception. I ended up doing simple regex validation for the new setting instead of validating user IDs. This prevents issues like you mentioned where a valid value for the setting becomes invalid due to a user deletion even if not changed by an admin. It also lets admins specify deleted user IDs, so audit records for deleted users can still be preserved.

The code handling this setting is always tolerant of invalid/deleted user IDs, so the only consequence of a deleted or incorrect user ID being entered is that it will appear as "Deleted user (123)" in the admin settings page.

The validation regex checks for a repeated sequence of digit(s), comma; then ensures that it ends with digit(s). This permits any sequence of numbers separated by commas, while rejecting non-numbers and numbers that are separated by anything other than commas. (Optional whitespace padding is permitted at all points.)

config/resque_schedule.yml
hunternet93: I updated the schedule to run on the 1st day of the month at 07:55 server time, which seems to be a time of day a bit after certain other jobs run.

app/policies/admin_setting_policy.rb
hunternet93: I added tests for the legal role as well as tests to ensure unauthorized roles cannot modify the setting.

app/jobs/audits_cleanup_job.rb
hunternet93: I ended up going with tail recursion where the job re-enqueues itself whenever it hits its deletion limit. This prevents any one execution of the job from taking too long and blocking the queue, while still completing its deletion.

PR #5795, opened 2026-05-08
AO3-6978 Create and fill new table for import urls

luis-pabon-tf:
# Pull Request Checklist

## Issue
https://otwarchive.atlassian.net/browse/AO3-6978

## Purpose

Step 1 of moving imported URLs to a new table to store all searchable variations.

## References

This is a follow-up to PR [5659](https://github.com/otwcode/otwarchive/pull/5659) as that had issues during the production backfill.
[AO3-6979](https://otwarchive.atlassian.net/browse/AO3-6979) follows up on this one
[AO3-6591](https://otwarchive.atlassian.net/browse/AO3-6591) Is the parent both tickets belong to

## Credit

Luis Pabon (they/he)

[AO3-6979]: https://otwarchive.atlassian.net/browse/AO3-6979?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/tasks/work_import_urls.rake
```diff
@@ -7,17 +7,24 @@ namespace :work_import_urls do
...
 
     puts "Backfilling #{total} work_import_urls records..."
 
     scope.find_each(batch_size: batch_size) do |work|
       next if ImportedUrl.exists?(work_id: work.id)
 
-      work.imported_url = ImportedUrl.create(original: work.imported_from_url)
+      begin
+        work.imported_url = ImportedUrl.create(original: work.imported_from_url)
+      rescue StandardError => e
```
sarken, line 19: Could you update spec/lib/tasks/work_import_urls.rake_spec.rb‎ with a test that shows the output for the error?

db/migrate/20260318013837_create_imported_urls.rb
```diff
@@ -7,10 +7,10 @@ def change
       t.string :minimal, null: false
       t.string :minimal_no_protocol_no_www, null: false
       t.string :no_www, null: false
-      t.string :with_www, null: false
+      t.string :with_www, null: false, limit: 300
```
sarken, line 13: Could you add a comment documentation why these are the chosen values for these limits?

sarken (changes requested):
Thanks for working on this!

lib/tasks/work_import_urls.rake
luis-pabon-tf: I'm trying to figure a way to do that. With the new limits in place the error won't hit since the original url can't be larger than 255 already. The largest I've gotten on the encoded column has been around 600 chars.

sarken: That's okay -- it looks like we don't test the output for fully successful runs, either, so we can live without it.

sarken (approved):

PR #5798, opened 2026-05-10
AO3-7219 persist work preference after comment

Dame-Zoom-A-Lot:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-7219

## Purpose

Fix the bug where users who had "show the whole work by default" enabled were getting redirected to the full work view after commenting.

This was happening because of this line in the `redirect_to_all_comments` function

```
if commentable.is_a?(Chapter) && (options[:view_full_work] || current_user.try(:preference).try(:view_full_works))
   commentable = commentable.work
end
```
1. Commentable is always a chapter in multi-chapter works. 
2. Regardless of options[:view_full_work], if user's preference has view_full_works set to true, (options[:view_full_work] || current_user.try(:preference).try(:view_full_works)) is always true

We also had bugs earlier in the code when setting `options[:view_full_work]`

In `show_comments` and `add_comment_reply`, `options[:view_full_work]` always evaluates as truthy regardless of whether it's 'true', or 'false', since params get evaluated as string values.

I simplified the logic by basing redirects on the previous page (referer) rather than the view_full_work param.

Based on manual test + looking at the code, I'm assuming that there's 3 ways to create / edit / delete a comment
1. From work view (`/works/:id`)
2. From chapter view (`/works/:id/chapters/:id`)
3. From comment view (`/comments/:id`)

I was concerned that by updating the logic so we no longer rely on `params[:view_full_work]`, I'd break the flow from when a user deletes a comment from the comment view. 

But I confirmed that we never set the `view_full_work` param when a user is navigated to to the `comments` path in both of the below cases:
1. User clicks `Thread` on a comment box, which navigates to the `comments/:id` path
2. User creates a reply beyond maximum comment thread depth, which navigates to the `comments/:id` path

We also did not set the `view_full_work` param when a user switched from entire work view to chapter view.

Since the `view_full_work` param is set unpredictably, and inconsistently, I believe it would be more maintainable if we consolidate all the logic to the redirect function and go off of where the user last was instead.

## Testing Instructions
[Please see Jira ticket](https://otwarchive.atlassian.net/browse/AO3-7219) - [most recent test instructions are here](https://otwarchive.atlassian.net/browse/AO3-7219?focusedCommentId=381818) with screenshots
 
## References
This replaces an original PR: https://github.com/otwcode/otwarchive/pull/5524

@Bilka2 thank you for your careful review. I agreed with your comment that the routing logic in the original PR was too convoluted, and that the simplest solution is to take the user back to where they were to begin with.

I was going to work off of my old PR, but it was so old that just starting from a clean slate was easier.

## Credit

Zooms (any)

they/them pronouns.*

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -436,7 +436,7 @@ class CommentsController < ApplicationController
             elsif @comment.unreviewed?
               redirect_to_all_comments(@commentable)
             else
-              redirect_to_comment(@comment, { view_full_work: (params[:view_full_work] == "true"), page: params[:page] })
+              redirect_to_comment(@comment, {  page: params[:page] })
             end
           end
         end
@@ -733,6 +733,10 @@ class CommentsController < ApplicationController
   def redirect_to_all_comments(commentable, options = {})
     default_options = {anchor: "comments"}
     options = default_options.merge(options)
+    # If the user is coming from the chapter view (i.e. referrer is /chapters/:id)
+    # then stay in chapter view. Otherwise, go to user preference
+    is_coming_from_chapter_view = request.referer&.match(/chapter/).present?
+    view_full_work = !is_coming_from_chapter_view || current_user.try(:preference).try(:view_full_work)
 
     if commentable.is_a?(Tag)
       redirect_to comments_path(tag_id: commentable.to_param,
@@ -741,16 +745,16 @@ class CommentsController < ApplicationController
                   page: options[:page],
                   anchor: options[:anchor])
     else
-      if commentable.is_a?(Chapter) && (options[:view_full_work] || current_user.try(:preference).try(:view_full_works))
+      if commentable.is_a?(Chapter) && view_full_work
         commentable = commentable.work
       end
       redirect_to polymorphic_path(commentable,
                                    options.slice(:show_comments,
                                                  :add_comment_reply_id,
                                                  :delete_comment_id,
-                                                 :view_full_work,
                                                  :anchor,
-                                                 :page))
+                                                 :page)
+                                          .merge({view_full_work:}))
     end
   end
 
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -614,7 +614,6 @@ class CommentsController < ApplicationController
         # so we're being extra-nice and preserving any intention to comment along with the show comments option
         options = {show_comments: true}
         options[:add_comment_reply_id] = params[:add_comment_reply_id] if params[:add_comment_reply_id]
-        options[:view_full_work] = params[:view_full_work] if params[:view_full_work]
         options[:page] = params[:page]
         redirect_to_all_comments(@commentable, options)
       end
@@ -644,7 +643,6 @@ class CommentsController < ApplicationController
         options[:controller] = @commentable.class.to_s.underscore.pluralize
         options[:anchor] = "comment_#{params[:id]}"
         options[:page] = params[:page]
-        options[:view_full_work] = params[:view_full_work]
         if @thread_view
           options[:id] = @thread_root
           options[:add_comment_reply_id] = params[:id]
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -733,10 +733,6 @@ class CommentsController < ApplicationController
   def redirect_to_all_comments(commentable, options = {})
     default_options = {anchor: "comments"}
     options = default_options.merge(options)
-    # If the user is coming from the chapter view (i.e. referrer is /chapters/:id)
-    # then stay in chapter view. Otherwise, go to user preference
-    is_coming_from_chapter_view = request.referer&.match(/chapter/).present?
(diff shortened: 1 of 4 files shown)


app/controllers/comments_controller.rb
```diff
@@ -706,51 +704,61 @@ def cancel_comment_delete
...
-        }
-      else
-        default_options = {
-           controller: comment.commentable.class.to_s.underscore.pluralize,
-           action: :show,
-           id: (comment.commentable.is_a?(Tag) ? comment.commentable.to_param : comment.commentable.id),
-           anchor: "comment_#{comment.id}"
-        }
-      end
+      default_options = if comment.ultimate_parent.is_a?(Tag)
```
Dame-Zoom-A-Lot, line 715: rubocop formatting changes

Dame-Zoom-A-Lot:
Setting to draft mode while I add tests to default_rails_actions_spec

app/controllers/comments_controller.rb
```diff
@@ -337,7 +342,7 @@ def set_page_subtitle
   def index
     return raise_not_found if @commentable.blank?
 
-    return unless @commentable.class == Comment
+    return unless @commentable.instance_of?(Comment)
```
Dame-Zoom-A-Lot, line 345: rubocop fix

app/controllers/comments_controller.rb
```diff
@@ -436,7 +441,7 @@ def create
             elsif @comment.unreviewed?
               redirect_to_all_comments(@commentable)
             else
-              redirect_to_comment(@comment, { view_full_work: (params[:view_full_work] == "true"), page: params[:page] })
+              redirect_to_comment(@comment, { view_full_work: (params[:view_full_work] && params[:view_full_work] == "true"), page: params[:page] })
```
Dame-Zoom-A-Lot, line 444: former syntax was setting `view_full_work = false` when `params[:view_full_work]` was not set, and we should have been navigating the user based on their preference

This only sets view_full_work if params[:view_full_work] was defined to begin with

app/controllers/comments_controller.rb
```diff
@@ -706,44 +712,57 @@ def cancel_comment_delete
...
     else
-      if commentable.is_a?(Chapter) && (options[:view_full_work] || current_user.try(:preference).try(:view_full_works))
-        commentable = commentable.work
-      end
+      is_coming_from_chapter_view = request.referer&.match(/chapters/).present?
+      is_not_coming_from_work_view = request.referer&.match(/works/).nil?
+      view_full_work = if is_coming_from_chapter_view
+                         false
+                       elsif is_not_coming_from_work_view
+                         if params[:view_full_work] == "true"
```
Dame-Zoom-A-Lot: We may want to set this as a helper, because I think this is how we'd actually want the logic to be for full work vs chapter view:

1. If user is coming from chapter by chapter view --> stay in chapter by chapter view
2. If user is coming from full work view -> stay in full work view
3. If user is coming from a third location, such as the comment page, _if_ we have a param set, then follow param, otherwise, stick to the user's preference

spec/controllers/comments/comments_controller_spec.rb
```diff
@@ -298,7 +388,8 @@
         delete :destroy, params: { id: comment.id }
         expect(flash[:comment_notice]).to eq "Comment deleted."
         it_redirects_to_simple(work_path(work, show_comments: true, anchor: :comments))
-        expect { comment.reload }.to raise_exception(ActiveRecord::RecordNotFound)
+        expect { comment.reload }
```
Dame-Zoom-A-Lot, line 391: rubocop changes

spec/controllers/comments/default_rails_actions_spec.rb
```diff
@@ -520,7 +620,7 @@
             post :create, params: { work_id: work.id, comment: comment_attributes }
             comment = Comment.last
             expect(flash[:error]).to be_nil
-            expect(response).to redirect_to(chapter_path(comment.commentable, show_comments: true, view_full_work: false, anchor: "comment_#{comment.id}"))
+            expect(response).to redirect_to(chapter_path(comment.commentable, show_comments: true, anchor: "comment_#{comment.id}"))
```
Dame-Zoom-A-Lot, line 623: We were translating params[:view_full_work] = nil to params[:view_full_work] = false, which would have overridden user preference.

I'm guessing that this is why we might have had that workaround that introduced the original bug, where we were setting view_full_work = params[:view_full_work == true || <user preference>

spec/controllers/comments/default_rails_actions_spec.rb
```diff
@@ -325,6 +325,106 @@
         end
       end
 
+      context "when work is not restricted for a logged in user" do
```
Dame-Zoom-A-Lot, line 328: I considered adding tests for delete as well, since that also calls redirect_to_all_comments under the hood. 

But I want to be mindful of taking up too much CI resources. Since both #create and #delete are using `redirect_to_all_comments` to redirect to either chapter view, work view, etc, I'm assuming just testing in the create route should be enough coverage.

Dame-Zoom-A-Lot:
@Bilka2 this is ready for review again. Sorry about the churn

PR #5800, opened 2026-05-10
AO3-6860 Allow CSS aspect-ratio property in skins

Taxen99:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6860

## Purpose

This allows the `aspect-ratio` CSS property to be used in skins, and updates the help popup for skins accordingly.

## Credit

Taxen99 (he/him)

### DIFF
--- a/lib/css_cleaner.rb
+++ b/lib/css_cleaner.rb
@@ -13,8 +13,6 @@ module CssCleaner
   PAREN_NUMBER_REGEX = Regexp.new('\(\s*' + NUMBER_WITH_UNIT_REGEX.to_s + '+\s*\)')
   PREFIX_REGEX = Regexp.new('moz|ms|o|webkit')
 
-  GLOBAL_VALUE_REGEX = Regexp.new("inherit|initial|revert|revert-layer|unset", Regexp::IGNORECASE)
-
   FUNCTION_NAME_REGEX = Regexp.new('scalex?y?|translatex?y?|skewx?y?|rotatex?y?|matrix', Regexp::IGNORECASE)
   TRANSFORM_FUNCTION_REGEX = Regexp.new("#{FUNCTION_NAME_REGEX}#{PAREN_NUMBER_REGEX}")
 
@@ -140,9 +138,7 @@ module CssCleaner
       # preserve the original capitalization
       clean = value if sanitize_css_font(value).present?
     elsif property == "aspect-ratio"
-      # Allow either global values such as "inherit", "revert", etc; "auto"; a ratio;
-      # "auto" followed by a ratio; or a ratio followed by "auto".
-      clean = value if value.match(/^(#{GLOBAL_VALUE_REGEX}|auto|(auto\s+)?#{NUMBER_OR_RATIO_REGEX}(\s+auto)?)$/i) && !value.match(/^auto.*auto$/i)
+      clean = value if value.match(/\A(#{ALPHA_REGEX}|(auto\s+)?#{NUMBER_OR_RATIO_REGEX}(\s+auto)?)\z/i)
     elsif property == "content"
       # don't allow var() function
       clean = value.match(/\bvar\b/i) ? "" : sanitize_css_content(value)
--- a/lib/css_cleaner.rb
+++ b/lib/css_cleaner.rb
@@ -13,7 +13,7 @@ module CssCleaner
   PAREN_NUMBER_REGEX = Regexp.new('\(\s*' + NUMBER_WITH_UNIT_REGEX.to_s + '+\s*\)')
   PREFIX_REGEX = Regexp.new('moz|ms|o|webkit')
 
-  GLOBAL_VALUE_REGEX = Regexp.new('inherit|initial|revert|revert-layer|unset', Regexp::IGNORECASE)
+  GLOBAL_VALUE_REGEX = Regexp.new("inherit|initial|revert|revert-layer|unset", Regexp::IGNORECASE)
 
   FUNCTION_NAME_REGEX = Regexp.new('scalex?y?|translatex?y?|skewx?y?|rotatex?y?|matrix', Regexp::IGNORECASE)
   TRANSFORM_FUNCTION_REGEX = Regexp.new("#{FUNCTION_NAME_REGEX}#{PAREN_NUMBER_REGEX}")
--- a/lib/css_cleaner.rb
+++ b/lib/css_cleaner.rb
@@ -8,9 +8,12 @@ module CssCleaner
   ALPHA_REGEX = Regexp.new('[a-z\-]+')
   UNITS_REGEX = Regexp.new('deg|cm|em|ex|in|mm|pc|pt|px|s|%', Regexp::IGNORECASE)
   NUMBER_REGEX = Regexp.new('-?\.?\d{1,3}\.?\d{0,3}')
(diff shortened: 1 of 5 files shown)


sarken:
Hi, Taxen99!

Thank you so much for this pull request. Someone will be along soon to do a thorough review, but I noticed some automated test failures related to the i18n changes and wanted to give you a quick heads up that our [i18n-tasks documentation](https://github.com/otwcode/otwarchive/wiki/Internationalization-%28i18n%29-Standards#user-content-i18ntasks) might help you fix those up.

In the meantime, we've updated the Jira issue status to In Review so no one mistakenly creates a duplicate pull request. If you'd like the ability to comment on, assign, and transition issues in the future, you're welcome to create a Jira account! It makes things a bit easier for us on the organizational side if the Full Name on your Jira account either closely matches the name you'd like us to credit in the release notes or includes it in parentheses, e.g. "Nickname (CREDIT NAME)."

lib/css_cleaner.rb
```diff
@@ -8,10 +8,13 @@ module CssCleaner
   ALPHA_REGEX = Regexp.new('[a-z\-]+')
   UNITS_REGEX = Regexp.new('deg|cm|em|ex|in|mm|pc|pt|px|s|%', Regexp::IGNORECASE)
   NUMBER_REGEX = Regexp.new('-?\.?\d{1,3}\.?\d{0,3}')
+  NUMBER_OR_RATIO_REGEX = Regexp.new("#{NUMBER_REGEX}(\s*\/\s*#{NUMBER_REGEX})?")
   NUMBER_WITH_UNIT_REGEX = Regexp.new("#{NUMBER_REGEX}\s*#{UNITS_REGEX}?\s*,?\s*")
   PAREN_NUMBER_REGEX = Regexp.new('\(\s*' + NUMBER_WITH_UNIT_REGEX.to_s + '+\s*\)')
   PREFIX_REGEX = Regexp.new('moz|ms|o|webkit')
 
+  GLOBAL_VALUE_REGEX = Regexp.new("inherit|initial|revert|revert-layer|unset", Regexp::IGNORECASE)
```
pmonfort: We can drop this new constant. The aspect-ratio condition can use the existing `ALPHA_REGEX` (`[a-z\-]+`) instead, it matches `auto`, `inherit`, `initial`, `revert`, `revert-layer`, `unset`, and similar keyword values. `aspect-ratio` is pretty low-risk for unexpected keywords, so we don't need a separate `GLOBAL_VALUE_REGEX` here.

lib/css_cleaner.rb
```diff
@@ -136,6 +139,10 @@ def sanitize_css_declaration_value(property, value)
     if property == "font-family"
       # preserve the original capitalization
       clean = value if sanitize_css_font(value).present?
+    elsif property == "aspect-ratio"
+      # Allow either global values such as "inherit", "revert", etc; "auto"; a ratio;
+      # "auto" followed by a ratio; or a ratio followed by "auto".
+      clean = value if value.match(/^(#{GLOBAL_VALUE_REGEX}|auto|(auto\s+)?#{NUMBER_OR_RATIO_REGEX}(\s+auto)?)$/i) && !value.match(/^auto.*auto$/i)
```
pmonfort: We need to make a few changes here:

1. Please use \A and \z instead of `^` and `$`. In Ruby, `^/$` match the start/end of a line, not the whole string. Newlines aren't stripped before this runs, so something like `1\nurl(http://...)` could currently pass. Using `\A/\z` fixes that.

2. Replace `GLOBAL_VALUE_REGEX|auto` with `ALPHA_REGEX`. That covers auto, global values like inherit, etc., and keeps things simpler. (With that change, values like banana would also be allowed, which we're okay with for this property.)

3. Drop the `!value.match?("auto.+auto")` check. It doesn't work reliably, and we don't usually check "is this CSS right" anyway.

pmonfort (commented):
Thanks for working on this! I left a few comments

lib/css_cleaner.rb
Taxen99: Thanks! Should be fixed now.

Taxen99:
Should be fixed now.

spec/lib/css_cleaner_spec.rb
```diff
@@ -243,6 +243,38 @@
         skin = create(:work_skin, css: "p { color: red; }")
         expect(skin.reload.css).to include("#workskin p")
       end
+
+      context "with aspect-ratio property" do
```
pmonfort: Since `aspect-ratio` behaves the same for both site skins and work skins, can you move these tests to the context "when cleaning Skin CSS" block (line 7) using `build(:skin, ...)` instead? The "when cleaning WorkSkin CSS" context is for work-skin-specific behavior.

Also, could you add a blank line between the two it blocks?

Taxen99: Should be good now

spec/lib/css_cleaner_spec.rb
```diff
@@ -197,6 +197,39 @@
         expect(skin.save).to be_falsy
         expect(skin.css).to eq("")
       end
+
+      context "with aspect-ratio property" do
+        it "allows permitted values" do
+          skin = build(:work_skin, css: [
```
pmonfort: Thanks for moving the tests! This should use `build(:skin, ...)` to match the context.

```suggestion
          skin = build(:skin, css: [
```

spec/lib/css_cleaner_spec.rb
```diff
@@ -197,6 +197,39 @@
...
+            "div { aspect-ratio: auto 4/3 }",
+            "div { aspect-ratio: 1 auto }",
+            "div { aspect-ratio: 0.25 auto }",
+            "div { aspect-ratio: 0.5/ 1 auto }"
+          ].join("\n"))
+          expect(skin.save).to be_truthy
+        end
+
+        it "strips nonsensical values" do
+          skin = build(:work_skin, css: [
```
pmonfort: Idem

```suggestion
          skin = build(:skin, css: [
```

pmonfort (approved):
Code looks great, thanks for addressing the comments!

lib/css_cleaner.rb
sarken, line 56: Just one more quick thing! Could you add this comment here for future reference? If I've done this right, you should be able to just accept this suggestion without any more work.
```suggestion
  # Unless we have a reason to allow a new value syntax for all or most
  # properties, sanitization should be handled on a per-property basis in 
  # sanitize_css_declaration_value. While invalid values are typically harmless,
  # including new syntax in VALUE_REGEX risks allowing unwanted valid values.
  VALUE_REGEX = Regexp.new("#{TRANSFORM_FUNCTION_REGEX}|#{URL_FUNCTION_REGEX}|#{COLOR_STOP_FUNCTION_REGEX}|#{COLOR_REGEX}|#{NUMBER_WITH_UNIT_REGEX}|#{ALPHA_REGEX}|#{SHAPE_FUNCTION_REGEX}|#{FILTER_FUNCTION_REGEX}|#{DROP_SHADOW_FUNCTION_REGEX}|#{VAR_FUNCTION_REGEX}")
```

sarken (changes requested):

Taxen99:
Should hopefully actually be good now.

sarken (approved):
Thank you so much!

PR #5802, opened 2026-05-10
AO3-2373 Adding tests to ensure related work emails are always sent

ASalbenblatt:

## Issue

https://otwarchive.atlassian.net/browse/AO3-2373

## Purpose

The issue appears to have been fixed at some point by unrelated code so this PR just adds tests to ensure that the issue stays fixed and is easily caught if it happens again.

## Testing Instructions

1. Create a fic
2. Mark it as a translation or remix (I've only tested those) of an existing work of yours
3. Post it without previewing it
4. See if you get an email notification (you should)

Note: even though one comment says to "Add 'inspired by' to existing work" and then look for a notification, that is actually covered by issue AO3-1506 and so is not included in this PR

## Credit

FlyingFalcon they/them

### DIFF
--- a/features/works/work_related.feature
+++ b/features/works/work_related.feature
@@ -11,6 +11,29 @@ Scenario: Posting a remix / related work emails the creator of the original work
   Then a parent related work should be seen
     And the original author should be emailed
 
+Scenario: Posting a remix in a variatey of ways should always send related work emails the creator of the original work and list the parent work in the proper location on the remix / related work
+
+  Given I have related works setup
+    And all emails have been delivered
+  When I post a related work as remixer without previewing it
+  Then a parent related work should be seen
+    And the original author should be emailed
+
+  Given all emails have been delivered
+  When I post a related work as remixer after saving it as a draft
+  Then a parent related work should be seen
+    And the original author should be emailed
+
+  Given all emails have been delivered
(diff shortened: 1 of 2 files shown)


features/works/work_related.feature
```diff
@@ -11,6 +11,29 @@ Scenario: Posting a remix / related work emails the creator of the original work
   Then a parent related work should be seen
     And the original author should be emailed
 
+Scenario: Posting a remix in a variatey of ways should always send related work emails the creator of the original work and list the parent work in the proper location on the remix / related work
+
+  Given I have related works setup
+    And all emails have been delivered
+  When I post a related work as remixer without previewing it
+  Then a parent related work should be seen
+    And the original author should be emailed
```
Bilka2: Just one request from me: Could you make the definition of this step check the subject of the email to make sure it's the related work email? I'm worried that we might accidentally break these tests and not notice because the email is replaced with some other email. And it looks like `the original author should be emailed` is only used for related work tests should it be simple enough to add the subject check to it

(Thank you very much for the note about AO3-1506, saved me some Jira searching!)

Bilka2 (changes requested):

Bilka2 (approved):
Thank you!

PR #5805, opened 2026-05-12
AO3-7465 Allow Rails to manage YJIT default

nateberkopec:
https://otwarchive.atlassian.net/browse/AO3-7465

This will turn YJIT on. This will increase memory consumption by ~10-20%.


brianjaustin (approved):
# Deployment Notes

1. This should go in its own release
1. We want to try it on a few machines as an ad-hoc deploy first, then do the normal deployment process

PR #5806, opened 2026-05-12
AO3-7443 Remove Google from embed allowlist

sarken:
## Issue

https://otwarchive.atlassian.net/browse/AO3-7443

## Purpose

What it says on the tin.

## Testing Instructions

Refer to Jira.


brianjaustin (approved):
Thanks!

PR #5817, opened 2026-05-15
AO3-7472 Hide hidden works in the related work notes

slavalamp:
## Pull Request Checklist

## Issue

[AO3-7472](https://otwarchive.atlassian.net/browse/AO3-7472)

## Purpose

Hides the works hidden by admins in the related work notes (inspired by, translation of, etc) in work notes, work edit pages and work downloads

## References

Please merge this either before or together with #5818, since AO3-6022 will make it impossible to unapprove a related work while it's hidden

## Credit

slavalamp

[AO3-7472]: https://otwarchive.atlassian.net/browse/AO3-7472?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/helpers/works_helper.rb
```diff
@@ -148,6 +148,11 @@ def related_work_note(related_work, relation, download: false)
         language: language,
         locale: default_locale,
         creator_link: creator_link)
+    elsif related_work.respond_to?(:hidden_by_admin?) && related_work.hidden_by_admin?
+      t(".#{relation}.hidden_by_admin_html",
```
slavalamp: ...just realized i should've just called the translation key `hidden_html`, so this is a reminder to myself to change it later

app/controllers/related_works_controller.rb
```diff
@@ -6,10 +6,10 @@ class RelatedWorksController < ApplicationController
 
   def index
     @page_subtitle = t(".page_title", login: @user.login)
-    @translations_of_user = @user.related_works.posted.where(translation: true)
-    @remixes_of_user = @user.related_works.posted.where(translation: false)
-    @translations_by_user = @user.parent_work_relationships.posted.where(translation: true)
-    @remixes_by_user = @user.parent_work_relationships.posted.where(translation: false)
+    @translations_of_user = @user.related_works.posted_children.where(translation: true)
```
slavalamp: i copied some scopes from my other pr (AO3-5550 AO3-6022), so the scopes being renamed in this file and `_sidebar.html.erb` are just temporary workarounds to make them work without the rest of the changes in those prs

app/models/related_work.rb
```diff
@@ -7,9 +7,41 @@ class RelatedWork < ActiveRecord::Base
...
+  scope :reciprocal, -> { where(reciprocal: true) }
+
+  scope :posted_children, -> { joins(:work).where(work: { posted: true }) }
+  scope :unhidden_children, -> { joins(:work).where(work: { hidden_by_admin: false }) }
+  scope :unrestricted_children, -> { joins(:work).where(work: { restricted: false }) }
+
+  scope :join_parents, lambda {
+    joins("LEFT JOIN works parent_works ON (related_works.parent_type = 'Work' AND parent_works.id = related_works.parent_id)")
+      .joins("LEFT JOIN external_works parent_external_works ON (related_works.parent_type = 'ExternalWork' AND parent_external_works.id = related_works.parent_id)")
+  }
```
slavalamp, line 25: same new scopes as in #5818

app/controllers/related_works_controller.rb
slavalamp: just realized i could've just kept the old `posted` scope... it'd probably be less annoying to merge then

app/models/related_work.rb
slavalamp: *except for `:join_parents` because i now replaced it with an activerecord `where().or(where())` query since it doesn't need to include deleted works on the related works page (the other pr), and it seems like work pages would need to have a `where().or(where()).or(where())` kind of query here in order to include them, which seems noticeably slower than sql joins, so i'm leaving it as is. but maybe there still are better solutions 🤷🏻

app/models/related_work.rb
```diff
@@ -11,6 +11,42 @@ class RelatedWork < ActiveRecord::Base
...
+
+  scope :posted_or_deleted_parents, lambda {
+    join_parents.where("parent_works.posted = true OR parent_works.id IS NULL")
+  }
+
+  scope :unhidden_or_deleted_parents, lambda {
+    join_parents.where("parent_works.hidden_by_admin = false OR parent_works.id IS NULL")
```
marcus8448: It looks like this misses hidden external works.

app/models/related_work.rb
```diff
@@ -11,6 +11,42 @@ class RelatedWork < ActiveRecord::Base
...
+  }
```
marcus8448, line 33: Could the `join_parents` call be pulled out of these scopes? (so that the user has to call `join_parents` once before using either of them)
I'm not sure if rails is guaranteed to deduplicate the joins.

marcus8448 (changes requested):
Thanks so much for working on this!

app/models/related_work.rb
slavalamp: it can be pulled out, but it deduplicated them when i checked and i figured the code would be more readable this way

slavalamp:
i realized i also wasn't testing whether the hidden works were hidden on edit pages and whether admins could still see them, so i've added steps for that

this also feels like way too many tests now. i was referencing other existing tests when i wrote mine, but i still doubt:
- whether it's actually needed to test both remixes and translations every time (they're using all the exact same scopes underneath, and i'm guessing more than these tests would break if specifically the translations broke)
- whether those `I log out` steps are really needed for downloading works (or, i guess i'm pretty sure they aren't, so it's more, is it bad to leave them in?)
- and whether it's needed to test that downloads update when a work's hidden status changes (i'd actually already asked sarken some time ago and she'd said "probably" but i still don't see why they wouldn't update in this case specifically...)

app/models/related_work.rb
marcus8448: ~~Although it does look slightly nicer, I'd prefer it pulled out just in case the behaviour does change in the future (since it's not documented anywhere).~~

marcus8448 (commented):
> whether it's actually needed to test both remixes and translations every time (they're using all the exact same scopes underneath, and i'm guessing more than these tests would break if specifically the translations broke)

Yeah, I think coverage of one is enough. Since these relations are filtered in the association and totally removed (unlike mystery/restricted works) there's really no difference in how we handle them.

> whether those I log out steps are really needed for downloading works (or, i guess i'm pretty sure they aren't, so it's more, is it bad to leave them in?)

We shouldn't be changing anything about the content of downloads based on who's logged in, so it shouldn't matter in this case. In the interest of test runtime, it might as well go. 

> and whether it's needed to test that downloads update when a work's hidden status changes (i'd actually already asked sarken some time ago and she'd said "probably" but i still don't see why they wouldn't update in this case specifically...)

Looking at what we do for downloads, the caching is on the systems side so we wouldn't really catch anything here. I think it can go as well.

app/models/related_work.rb
marcus8448: Actually, scratch that. If Rails changes its behaviour it should fail spectacularly and the tests should catch it. I can't think of a scenario where an complete duplicate string join would be wanted either.

slavalamp:
`work_related.feature` and `work_download.feature` both still have similar redundant tests for anonymous and restricted works and possibly other cases, so maybe it'd be a good idea to make a ticket to remove those later

slavalamp:
apparently downloads do in fact look different depending on who generates them

at least when generating them in dev and in tests 

(i would've caught this by running the tests locally but apparently i forgot to actually run the tests locally at first 😔)

features/step_definitions/admin_steps.rb
```diff
@@ -414,6 +414,18 @@
   step %{I follow "Make Work Visible"}
 end
 
+When "I hide the external work {string}" do |title|
+  external_work = ExternalWork.find_by(title: title)
+  visit external_work_path(external_work)
+  step %{I follow "Hide External Work"}
+end
+
+When "I unhide the external work {string}" do |title|
```
slavalamp: this step definition technically isn't needed anymore but i forgot to remove it twice now and don't feel like making an extra commit&push just for this...

marcus8448 (changes requested):
> apparently downloads do in fact look different depending on who generates them

Actually, I think this needs to be fixed then. After the initial generation, we cache the download result and serve it publicly, so we need to only have public information in there. We don't want these details revealed if an owner/admin is the first one to download a work.

I think it should be pretty doable by passing an argument (defaulting to current user) to the `visible`s and adding it to the scopes?

PR #5820, opened 2026-05-15
AO3-5556 Don't change collections on preview - Follow up PR

pmonfort:
# Pull Request Checklist

* [X] Have you added the [Jira](https://otwarchive.atlassian.net) issue number as the *first* thing in your pull request title (e.g. `AO3-1234 Fix thing`)
* [X] Do you have fewer than 5 pull requests already open? If not, please wait  until they are reviewed and merged before creating new pull requests.

## Issue

https://otwarchive.atlassian.net/browse/AO3-5556

## Purpose

Follow-up to #4501 by @tickinginstant, adopting and completing the work from that PR.

### Original PR (#4501)

- Generalize the function assign_tags_of_type to a function assign_through_association defined in ApplicationRecord that can be used for any type of association, not just tags.
- Use that function to assign collections, so that setting collection_names will only set up the changes to be performed on save instead of performing them immediately. Similarly with `collections_to_add` and `collections_to_remove`.
- Change the approve_automatically and set_anonymous_and_unrevealed callbacks in the CollectionItem class to run before_validation instead of before_save, so that we get the correct values for anonymous, unrevealed, user_approval_status, and collection_approval_status after validation.
- Make the Collectible module a concern and move it to the correct directory.
- Rewrite the Collectible function set_anon_unrevealed so that it uses the status of the collection items to compute whether the work should be anonymous/unrevealed, and make it run as an after_validation callback so that it has access to the collection item statuses set before validation.
- Rewrite BookmarksController#update to use the collection handling from Collectible, and add a function akin to WorksController#in_moderated_collection so that the message on adding a moderated collection stays the same as it was before the change.
- Stop calling save(validate: false) so much, and replace it with a reindex_item callback in the CollectionItem class. The item association on CollectionItems has touch: true, so changes to collection items will already update the work/bookmark timestamp. And indexing is the only other real purpose that calling save(validate: false) serves.

### This PR

- Resolved merge conflicts with master (~2 years of divergence)
- Reorder callbacks in `CollectionItem` to follow Rails convention (`before_validation` → `before_save` → `after_save` → `after_create_commit` → `after_destroy` → `after_commit`)
- Fix annotation style in `Collectible` (`# Note:` → `# NOTE:`) per Hound/Rubocop convention
- Extract `assign_through_association` from `ApplicationRecord` into an `AssociationAssignment` concern, included only in `Collectible` and `Taggable`
- Extract `collections_meta_tag` helper in `WorksHelper` to replace variable assignment in `works/_meta.html.erb`
- Fix cucumber steps: Cancel is a link (`I follow`), not a button (`I press`)

### Steps to reproduce the bug

1. Post a work
2. Edit the work
3. Add a collection to the work, especially an anonymous or unrevealed collection
4. Press "Preview"
5. Note that in the preview, the collection is listed in the work meta, and the byline is updated to "Anonymous" if the collection is anonymous (there will be no
indication on the preview page if the collection is unrevealed)
6. Press "Cancel"
7. The collection should **not** be added to the work

### Automated tests

- `bundle exec rspec spec/models/collection_item_spec.rb spec/models/concerns/collectible_spec.rb spec/lib/works_owner_spec.rb` — 65 examples, 0 failures
- `bundle exec cucumber features/collections/work_preview_collections.feature` — 6 scenarios, 6 passed

## References

Follow-up to #4501

## Credit

Pablo Monfort (he/him))


app/helpers/works_helper.rb
```diff
@@ -30,6 +30,15 @@ def work_meta_list(work, chapter = nil)
     content_tag(:dl, list.to_s, class: 'stats').html_safe
   end
 
+  def collections_meta_tag(work)
+    collections_data = show_collections_data(work.approved_collections_after_saving)
+    return if collections_data.blank?
+
+    content_tag(:dt, t("works_helper.collections_meta_tag.collections"), class: "collections") +
+      "\n".html_safe +
```
brianjaustin: Is there a way to join these with a newline that doesn't need `html_safe`? (Not super blocking if there isn't a nice way, but since we're in the neighbourhood...) Maybe `content_tag(:br)`?

app/models/concerns/collectible.rb
```diff
@@ -0,0 +1,162 @@
...
+             through: :collection_items,
+             source: :collection,
+             dependent: :destroy
+
+    # NOTE: this scope includes the items in the children of the specified collection
+    scope :in_collection, lambda { |collection|
+      distinct.joins(:approved_collection_items).merge(collection.all_items)
+    }
+
+    after_validation :set_anon_unrevealed, if: -> { collection_items.loaded? }
```
brianjaustin, line 37: To make sure I understand the flow here, why do we need the `if` condition?

app/models/collection_item.rb
```diff
@@ -47,46 +54,110 @@ def self.for_user(user=User.current_user)
...
+    self.anonymous = collection.anonymous?
+  end
+
+  before_validation :approve_automatically, on: :create
+  def approve_automatically
+    return unless item && collection
+
+    # approve with the current user, who is the person who has just
+    # added this item -- might be either moderator or owner
+    approve(User.current_user)
```
brianjaustin, line 71: Do we still need to check if this is `:false`? iirc there's some legacy code that sets it to that 🥲

spec/lib/works_owner_spec.rb
```diff
@@ -67,8 +67,6 @@
     shared_examples_for "an owner collection" do
       it "should change after a new work is created" do
         FactoryBot.create(:work, collection_names: @owner.name)
-        @owner.collection_items.each {|ci| ci.approve(nil); ci.save}
-        @child.collection_items.each {|ci| ci.approve(nil); ci.save} if @child
         expect(@original_cache_key).not_to eq(@owner.works_index_cache_key)
```
brianjaustin, line 70: Are these deletions because the approval is automatic now? Or some other reason?

spec/controllers/bookmarks_controller_spec.rb
```diff
@@ -217,10 +217,7 @@
       it "shows the collection warning message" do
         put :update, params: { id: bookmark.id, bookmark: { collection_names: collection.name } }
 
-        success_msg = "Bookmark was successfully updated. " \
-                      "Added to collection(s): #{collection.title}. " \
-                      "Please note: private bookmarks are not listed in collections."
-        it_redirects_to_with_notice(bookmark_path(bookmark), success_msg)
+        it_redirects_to_with_notice(bookmark_path(bookmark), "Bookmark was successfully updated.")
```
brianjaustin, line 220: This doesn't seem to match the description of the test -- is the test useful anymore?

config/locales/controllers/en.yml
```diff
@@ -103,6 +103,11 @@ en:
       success: Bookmark was successfully created. It should appear in bookmark listings within the next few minutes.
       warnings:
         private_bookmark_added_to_collection: " Please note: private bookmarks are not listed in collections."
+    new_moderated_collections_message_html:
```
brianjaustin, line 106: Why does this (and the other key) need `_html`? I don't see any HTML getting injected 🤔

app/controllers/bookmarks_controller.rb
```diff
@@ -351,6 +304,27 @@ def set_own_bookmarks
...
+
+    return if new_moderated_collections.blank?
+
+    links = new_moderated_collections.map do |collection|
+      view_context.link_to(collection.title, collection_path(collection))
+    end
+
+    message = t("bookmarks.new_moderated_collections_message_html",
+                count: new_moderated_collections.length,
+                collections: view_context.safe_join(links, ", "))
```
brianjaustin, line 323: Does this need to be `collections_html`? (Related to my question about the parent key name)

app/controllers/works_controller.rb
```diff
@@ -566,22 +566,23 @@ def send_external_invites(works)
...
-        next unless collection_item.collection == collection
-        if collection_item.approved_by_user? && collection_item.unreviewed_by_collection?
-          moderated_collections << collection
-        end
-      end
-    end
-    if moderated_collections.present?
-      flash[:notice] ||= ''
-      flash[:notice] += ts(" You have submitted your work to #{moderated_collections.size > 1 ? 'moderated collections (%{all_collections}). It will not become a part of those collections' : "the moderated collection '%{all_collections}'. It will not become a part of the collection"} until it has been approved by a moderator.", all_collections: moderated_collections.map(&:title).join(', '))
+  def in_moderated_collection(flash = self.flash)
```
brianjaustin: `in_moderated_collection` sounds like it would return a boolean to me; maybe `show_moderated_collection_warning` would be more accurate?

app/helpers/works_helper.rb
pmonfort: Good point, the `html_safe` is probably overkill for something like this. We don't really need the `\n` I'll remove it. Thanks!"

app/models/concerns/collectible.rb
pmonfort: It's a performance guard, without it, every save (even unrelated ones like editing tags) would trigger a query. If `collection_items` is loaded, there's a chance it was updated. It may be confusing though, I can remove it if you'd like.

app/models/collection_item.rb
pmonfort: I've been digging trying to find where current_user could be set to `:false`, and the only thing I found in the project is on line 171 of `bookmark.rb`, but that's a local parameter default, so it shouldn't affect User.current_user. We should be safe, but if you feel more comfortable keeping the `:false` check just in case, I can restore the previous code. What do you think?

```
git grep -n ":false"
app/models/bookmark.rb:150:    unless current_user == :false || !current_user
app/models/bookmark.rb:171:  def self.count_visible_bookmarks(bookmarkable, current_user=:false)
```

spec/lib/works_owner_spec.rb
pmonfort: Yes, exactly, the approval is now handled automatically in the `before_validation :approve_automatically` callback on `CollectionItem`, so the manual `approve(nil)` save calls are no longer needed.

spec/controllers/bookmarks_controller_spec.rb
pmonfort: You're right, the test description no longer matches the assertion. I'll rename it to something like "successfully updates when adding to a collection" to reflect what it's actually testing. I think the test is still valuable to cover the case of updating collection_names. What do you think?

app/controllers/bookmarks_controller.rb
pmonfort: No, the parameter name doesn't need `_html` only the parent key does.

config/locales/controllers/en.yml
pmonfort: The HTML comes from the %{collections} interpolation

app/controllers/works_controller.rb
pmonfort: Good call, I'll rename it to show_moderated_collection_warning.

app/helpers/works_helper.rb
pmonfort: Actually, we do need a space between the `<dt>` and `<dd>` I've added a trailing space to the `<dt>` content instead cleaner than `html_safe` let me know your thoughts

pmonfort:
@brianjaustin thanks for your code review! The code is ready for a second review, let me know your thoughts.

app/models/concerns/collectible.rb
brianjaustin: Nah, I think a comment for clarity would be enough :+1:

brianjaustin (approved):
Thanks! Nothing blocking from my side

PR #5822, opened 2026-05-17
AO3-6960 Admin post drafts and previews

marcus8448:
## Issue

https://otwarchive.atlassian.net/browse/AO3-6960

## Purpose

Allows admins to save unpublished (draft) admin posts and preview changes before publication.

## Testing Instructions

See Jira.

## Credit

marcus8448 (he/him)


app/views/admin_posts/_posting_fieldset.html.erb
```diff
@@ -0,0 +1,26 @@
...
+<fieldset>
+  <legend><%= t(".legend") %></legend>
+
+  <ul class="actions">
+    <% unless @admin_post.persisted? && @admin_post.posted? %>
+      <li><%= form.submit t(".actions.save_draft"), name: "save_button" %></li>
+    <% end %>
+
+    <% if @preview_mode %>
+      <li><%= submit_tag t(".actions.edit"), name: "edit_button" %>
```
marcus8448: ns

app/controllers/admin_posts_controller.rb
```diff
@@ -1,44 +1,51 @@
...
-    if params[:language_id].present? && (@language = Language.find_by(short: params[:language_id]))
-      @admin_posts = @admin_posts.where(language_id: @language.id)
-      @tags = AdminPostTag.distinct.joins(:admin_posts).where(admin_posts: { language_id: @language.id }).order(:name)
-    else
-      @admin_posts = @admin_posts.non_translated
-      @tags = AdminPostTag.order(:name)
-    end
-    @admin_posts = @admin_posts.order('created_at DESC').page(params[:page])
+    @page_subtitle = t(".page_title")
+    @admin_posts = @admin_posts.posted.order(published_at: :desc).page(params[:page])
```
pmonfort: This uses will_paginate, but drafts below uses pagy. Should this also use pagy for consistency?

app/controllers/admin_posts_controller.rb
```diff
@@ -55,10 +62,15 @@ def edit
   # POST /admin_posts
   def create
     @admin_post = AdminPost.new(admin_post_params)
+    @admin_post.posted = true if params[:post_button] && !@admin_post&.translated_post&.draft?
+
     authorize @admin_post
-    if @admin_post.save
-      flash[:notice] = ts("Admin Post was successfully created.")
-      redirect_to(@admin_post)
+    if params[:preview_button]
```
pmonfort: Should we validate that the post is valid before previewing? Something similar to what's done in `update` where `@admin_post.valid?` is checked before previewing the post.

app/controllers/admin_posts_controller.rb
```diff
@@ -67,12 +79,36 @@ def create
...
+
+    render action: "edit"
+  end
+
+  # PUT /admin_posts/1/post
+  def post
+    @admin_post = AdminPost.find(params[:id])
+    authorize @admin_post
+
+    @admin_post.posted = true
```
pmonfort, line 107: Should we also check `@admin_post&.translated_post&.draft?` like we do in create and update to avoid potential bugs?

app/views/admin/_admin_nav.html.erb
```diff
@@ -3,10 +3,17 @@
   <li>
     <%= span_if_current t(".ao3_news"), admin_posts_path %>
   </li>
+  <li>
+    <%= span_if_current t(".ao3_news_drafts"), drafts_admin_posts_path %>
```
pmonfort: The "AO3 News Drafts" link is shown without a permission check. The ticket says it should be visible to "any admin with a role listed in the AdminPostPolicy section." Should this be wrapped in `<% if policy(AdminPost).can_draft? %>` like the other links?

app/views/admin_posts/_posting_fieldset.html.erb
```diff
@@ -0,0 +1,26 @@
...
+<fieldset>
+  <legend><%= t(".legend") %></legend>
+
+  <ul class="actions">
+    <% unless @admin_post.persisted? && @admin_post.posted? %>
+      <li><%= form.submit t(".actions.save_draft"), name: "save_button" %></li>
+    <% end %>
+
+    <% if @preview_mode %>
+      <li><%= form.submit t(".actions.edit"), name: "edit_button" %>
```
pmonfort: Missing `</li>` closing tag here.

features/admins/admin_post_draft_news.feature
```diff
@@ -0,0 +1,162 @@
+@admin @comments
+Feature: Admin Actions to Draft News
+  In order to post news items
+  As an an admin
```
pmonfort: Nitpick: double "an"

pmonfort (commented):
Great work! The code looks good. Tested locally and the core flow works well. Left a few comments, mostly small things.

app/controllers/admin_posts_controller.rb
marcus8448: I've added a validation to the model to prevent translations from being posted first, which should cover this and also prevent already published posts from turned into translations of a draft post.

pmonfort: Great approach, thanks for taking care of that!

db/migrate/20260509172710_add_posted_status_and_publication_date_to_admin_posts.rb
```diff
@@ -0,0 +1,20 @@
...
+  uses_departure! if Rails.env.staging? || Rails.env.production?
+
+  def up
+    return if column_exists?(:admin_posts, :posted)
+
+    change_table :admin_posts do |t|
+      t.boolean :posted, default: false, null: false
+      t.datetime :published_at, default: nil, null: true
+    end
+    AdminPost.update_all("published_at = created_at, posted = 1")
```
brianjaustin, line 11: Won't this need to happen before the migration happens to avoid the migration failing due to the null columns? Although, I'd be more in favour of moving this to a before rake tasks (iirc the file doesn't exist but it might after a recent PR of mine) since we can test it with rspec that way

spec/helpers/admin_post_helper_spec.rb
```diff
@@ -1,19 +1,42 @@
...
       indonesian_post = create(:admin_post, language: indonesian, translated_post: english_post)
 
       expect(sorted_translations(english_post.reload)).to eq([indonesian_post, german_post, finnish_post])
     end
+
+    context "when the news is posted" do
+      it "does not include draft translations" do
+        _german_post = create(:admin_post, :draft, language: german, translated_post: english_post)
+        finnish_post = create(:admin_post, language: finnish, translated_post: english_post)
+        _indonesian_post = create(:admin_post, :draft, language: indonesian, translated_post: english_post)
```
brianjaustin: Small nitpick: you can omit the variable assignment for german/indonesian, since they are not used

spec/models/admin_post_spec.rb
```diff
@@ -46,4 +46,30 @@
       expect(new_post.reload.disable_all_comments?).to be(false)
     end
   end
+
+  describe "#set_published_at" do
```
brianjaustin, line 62: Could you also add a test for the validation that posting a translation isn't possible when the parent post is in draft?

app/models/admin_post.rb
```diff
@@ -137,4 +154,18 @@ def update_translation_tags
...
+    self.published_at = Time.current if self.posted && !self.published_at
+  end
+
+  def post_translations
+    return if translations.blank? || !self.posted
+
+    transaction do
+      translations.find_each do |post|
+        post.update(posted: true, published_at: self.published_at)
+      end
```
brianjaustin: Any reason not to use `update_all` here, so we have a single batch-update query?

app/views/admin_posts/_admin_post.html.erb
```diff
@@ -7,20 +7,25 @@
 <h4 class="landmark heading"><%= ts("Post Header") %></h4>
 <div class="wrapper">
   <dl class="meta">
-    <dt class="published"><%= ts("Published:") %></dt>
-    <dd class="published"><%= admin_post.created_at %></dd>
+    <% if admin_post.posted? %>
+      <dt class="published"><%= t(".published") %></dt>
+      <dd class="published"><%= admin_post.published_at %></dd>
```
brianjaustin: Shouldn't this be wrapped in `<%= l(...) %>` so it can be localized?

app/views/admin_posts/preview.html.erb
```diff
@@ -0,0 +1,21 @@
+<div class="admin">
+  <h2 class="heading"><%= t(".page_heading") %></h2>
+  <div class="draft news">
+    <%= render "admin_post", admin_post: @admin_post %>
+  </div>
+  <div class="clear"><!-- presentational --></div>
+
+  <%= form_for(@admin_post) do |f| %>
```
brianjaustin: Could you replace this with `form_with`? Per https://guides.rubyonrails.org/form_helpers.html#using-form-tag-and-form-for, that's the standard for Rails 5.1+

app/views/admin_posts/_admin_post.html.erb
marcus8448: Yeah, it should. Do we want the default `Sun, 17 May 2026 03:53:38 +0000` or should I define a format to match the old one? (`2026-05-17 03:53:38 UTC`)

db/migrate/20260509172710_add_posted_status_and_publication_date_to_admin_posts.rb
marcus8448: I'm not sure if I understand what you mean. Posted should be false be default, and published should be ok as null?

app/models/admin_post.rb
marcus8448: I was thinking validations/callbacks, since the other callbacks do it the same way. But it looks fine in this case since the translated post should have already updated everything itself.

app/views/admin_posts/_admin_post.html.erb
brianjaustin: As discussed internally, we'll stick with the default for now

brianjaustin (approved):

app/views/admin_posts/_admin_post.html.erb
```diff
@@ -31,8 +31,12 @@
...
       <dd class="tags">
         <ul class="tags commas">
-          <% for tag in admin_post.tags %>
-            <li><%= link_to tag.name, admin_posts_path(tag: tag.id), class: "tag" %></li>
+          <% admin_post.tags.each do |tag| %>
+          <li><%= link_to tag.name, if tag.new_record?
+                                      "#"
+                                    else
+                                      admin_post.draft? ? drafts_admin_posts_path(tag: tag.id) : admin_posts_path(tag: tag.id)
+                                    end, class: "tag" %></li>
```
brianjaustin: Small styling nitpick: the `<li>` element should be indented 1 extra level since it's inside the loop

brianjaustin (approved):

PR #5823, opened 2026-05-18
AO3-7442 Add TOTP 2FA to users

EchoEkhi:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-7442

## Credit

EchoEkhi (He/Him)


app/models/user.rb
```diff
@@ -175,6 +179,43 @@ class User < ApplicationRecord
   has_many :log_items, dependent: :destroy
   validates_associated :log_items
 
+  #####################################
+  # TOTP 2FA
+  #####################################
+
+  serialize :otp_backup_codes, type: Array, coder: YAML, yaml: { permitted_classes: [String] }
+  
+  # Generate a TOTP secret it it does not already exist
```
lydia-theda:
```suggestion
  # Generate a TOTP secret if it does not already exist
```

config/locales/controllers/en.yml
```diff
@@ -384,9 +384,37 @@ en:
...
+      already_disabled: TOTP two-step verification is already disabled.
+      already_enabled: TOTP two-step verification is already enabled.
+      confirm_disable:
+        page_title: Disable Two-Step Verification
+      confirm_reset_backup_codes:
+        page_title: Two-Step Verification Backup Codes
+      create:
+        incorrect_code: Incorrect verification code. Your code may have expired, or you may need to set up your authenticator app again.
+        page_title: Two-Step Verification Backup Codes
+        success: Successfully enabled two-step verification; please make note of your backup codes.
```
lydia-theda: In the interest of being explicitly clear to users, I think we should change "make note of" to more closely mirror the language used elsewhere about saving the backup codes. Something more like this?

(I haven't checked whether .write_codes_down appears on this same page, or if this particular phrasing will conflict with anything else. But a word like "save" is more clear than "note".)
```suggestion
        success: Two-step verification was successfully enabled. Please save your backup codes in case you lose access to your authenticator app.
```

config/locales/controllers/en.yml
```diff
@@ -384,9 +384,37 @@ en:
...
+        success: Successfully disabled two-step verification.
+      new:
+        page_title: Confirm Password
+      reauthenticate_create:
+        incorrect_password: Your password was incorrect. Please try again or, if you've forgotten your password, log out and reset your password via the link on the login form.
+        page_title: Enable Two-Step Verification
+      reset_backup_codes:
+        not_enabled: Please enable two-step verification first.
+        page_title: Two-Step Verification Backup Codes
+        success: Successfully reset your backup codes; please make note of your new backup codes, your old backup codes will no longer work.
```
lydia-theda: An initial attempt at aligning the wording with other messages across the site:
```suggestion
        success: Your backup codes have been successfully regenerated. Please save your new backup codes, as your old backup codes will no longer work.
```

config/locales/views/en.yml
```diff
@@ -3230,6 +3244,48 @@ en:
...
+        required_before_enabling: Before enabling two-step verification, you need to verify your password.
+        submit: Continue
+      reauthenticate_create:
+        about:
+          effects_of_enabling: When two-step verification is enabled, logging in to your %{app_name} account will require both your password and a verification code from your authenticator app.
+          general_info: Two-step verification, also called two-factor authentication, helps prevent other people from accessing your account even if they know your password. All you need to set it up is an authenticator app such as 1Password, Google Authenticator, or Twilio Authy.
+          heading: About Two-Step Verification
+        app_setup:
+          heading: 'Step 1: Set up your authenticator app'
+          instructions: Use your authenticator app to scan the QR code or enter the manual setup key. If you are using the same device the app is installed on, click the import settings button. The app will give you a 6-digit code, which you'll use in the next step.
```
lydia-theda: Let's try to avoid the use of plain "app", because there are many people who use that word to refer to AO3.

```suggestion
          instructions: Use your authenticator app to scan the QR code or enter the manual setup key. If you are using the same device your authenticator app is installed on, click the import settings button. The authenticator app will give you a 6-digit code, which you'll use in the next step.
```

config/locales/views/en.yml
```diff
@@ -3230,6 +3244,48 @@ en:
...
+          key:
+            copied: Copied!
+            copy_to_clipboard: Copy
+            format_note: Spaces and capitalization don't matter
+            heading: Setup Key
+          qr_scan_note: Scan with your authenticator app
+        enter_code:
+          heading: 'Step 2: Enter your 6-digit code'
+          instructions: Enter the 6-digit code from your app to enable two-step verification.
```
lydia-theda:
```suggestion
          instructions: Enter the 6-digit code from your authenticator app to enable two-step verification.
```

app/controllers/users/sessions_controller.rb
```diff
@@ -28,4 +32,78 @@ def destroy
...
+
+  def user_params
+    params.require(:user).permit(:login, :password, :remember_me)
+  end
+
+  def find_user
+    if session[:otp_user_id]
+      User.find(session[:otp_user_id])
+    elsif user_params[:login]
+      User.find_by(login: user_params[:login])
```
pmonfort: You should use [find_first_by_auth_conditions](https://github.com/otwcode/otwarchive/blob/master/app/models/user.rb#L262) it's a method that search both username and email.

```suggestion
      User.find_first_by_auth_conditions(login: user_params[:login])
```

app/controllers/users/sessions_controller.rb
```diff
@@ -28,4 +32,78 @@ def destroy
...
+    end
+  end
+
+  def user_params
+    params.require(:user).permit(:login, :password, :remember_me)
+  end
+
+  def find_user
+    if session[:otp_user_id]
+      User.find(session[:otp_user_id])
```
pmonfort: Nit: `User.find` raises `RecordNotFound` if the session has a stale ID, while the elsif branch returns `nil` It would be nice to have both paths behave the same way, `find_by` would return nil instead of a 500.


```suggestion
      User.find_by(id: session[:otp_user_id])
```

app/controllers/users/sessions_controller.rb
```diff
@@ -28,4 +32,78 @@ def destroy
...
+      flash.now[:error] = t("users.sessions.invalid_totp")
+      prompt_for_otp_two_factor(user)
+    end
+  end
+
+  def user_params
+    params.require(:user).permit(:login, :password, :remember_me)
+  end
+
+  def find_user
```
pmonfort, line 105: Nit: `find_user` is called twice per request, once in `totp_two_factor_enabled?` (line 107) and once in `authenticate_with_totp_two_factor` (line 38). Could memoize to avoid the extra query.

app/views/users/totp/reauthenticate_create.html.erb
```diff
@@ -0,0 +1,59 @@
...
+    <div class="notes">
+      <h4 class="heading"><%= t(".app_setup.key.heading") %></h4>
+      <code><%= setup_key = current_user.otp_secret.scan(/.{1,4}/).join(" ") %></code>
+      <p class="footnote"><%= t(".app_setup.key.format_note") %></p>
+      <p class="actions">
+        <button id="copy-totp-setup-key" onclick="copySetupKey()" style="visibility: hidden;">
+          <%= t(".app_setup.key.copy_to_clipboard") %>
+        </button>
+      </p>
+      <h4 class="heading">Setup on This Device</h4>
```
pmonfort: This string should use `t(...)` for i18n consistency

app/views/users/totp/reauthenticate_create.html.erb
```diff
@@ -0,0 +1,59 @@
...
+      <p class="actions">
+        <a href="<%= current_user.totp_qr_code_uri %>">Import Settings</a>
```
pmonfort: This string should use `t(...)` for i18n consistency

spec/controllers/users/totp_controller_spec.rb
```diff
@@ -0,0 +1,316 @@
...
+      it "allows enabling TOTP" do
+        fake_login_known_user(user)
+        post :create, params: { user_id: user.login, totp_attempt: user.current_otp }
+        expect(user.reload.totp_enabled?).to be_truthy
+        expect(flash[:notice]).to eq("Successfully enabled two-step verification; please make note of your backup codes.")
+      end
+
+      it "audits and redacts otp_secret" do
+        fake_login_known_user(user)
+        post :reset_backup_codes, params: { user_id: user.login }
```
pmonfort: This test is under describe "POST #create" but calls `post :reset_backup_codes` this should be call `post :create`

app/controllers/users/totp_controller.rb
```diff
@@ -0,0 +1,103 @@
...
+    else
+      flash[:error] = t(".failure")
+    end
+
+    redirect_to user_preferences_path
+  end
+
+  private
+
+  def require_user_owner
```
pmonfort: This method is defined but never used, was it left over from an earlier approach?

app/controllers/users/sessions_controller.rb
```diff
@@ -28,4 +32,78 @@ def destroy
...
+    user.validate_and_consume_otp!(params[:totp_attempt]) ||
+      user.invalidate_otp_backup_code!(params[:totp_attempt])
+  end
+
+  def prompt_for_otp_two_factor(user)
+    @user = user
+
+    session[:otp_user_id] = user.id
+    
+    session[:pwned] = user.respond_to?(:password_pwned?) && user.password_pwned?(user_params[:password]) if params[:user] && params[:user][:password]
```
pmonfort: Could you try to keep lines to 80 characters max? There are several lines over 100 characters in the PR.

pmonfort (commented):
Thanks for working on this! I left a few comments

app/views/users/totp/reauthenticate_create.html.erb
sarken: Also, just to check, should this be "Set up"? I can't tell if this heading is using it as a noun or a verb.

app/views/users/totp/reauthenticate_create.html.erb
sarken: Is there a reason it can't use `link_to`? It might be good to have a comment if so.

lydia-theda (commented):
Back in May I took a quick look at the message wording so here are those comments.

app/views/users/totp/reauthenticate_create.html.erb
lydia-theda: I concur, it should be Set Up

app/controllers/users/totp_controller.rb
```diff
@@ -0,0 +1,96 @@
...
+
+      render action: :reset_backup_codes and return
+    else
+      flash.now[:error] = t(".incorrect_code")
+      render action: :new and return
+    end
+  end
+
+  # GET /users/<login>/totp/confirm_reset_backup_codes
+  def confirm_reset_backup_codes
```
pmonfort, line 44: Maybe this action should have the same guard as `reset_backup_codes` below, redirecting with an error unless `current_user.totp_enabled?`, using a before_action for both, same pattern as `confirm_disable` and `disable` with `check_totp_enabled`. Otherwise the confirmation page renders for users without TOTP and they only get bounced after submitting. What do you think?

PR #5827, opened 2026-05-20
AO3-3934 Footer not sitting at bottom of page

forceofcalm:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-3934

## Purpose

The footer does not sit at the bottom of the page on pages where the site content does not fill the full height. This is especially obvious on larger screens and pages with short content like /about.

Fixing this was not easily, if at all, possible with only CSS2. To remedy this, we use minimal CSS3 flexbox rules that should not affect the current views in older browsers without CSS3, while fixing the issue in all browsers that do support CSS3. Unfortunately, because this is entirely a cosmetic change and due to the limitations in CSS2, this may be the best solution we have, though I am open to any and all suggestions.

This change also supports other device layouts and is mobile-responsive.

## Testing Instructions

I've tested this change as thoroughly as I could for regressions, but there may be pages or edge cases I am unaware of that could be affected. If this CSS3 approach is accepted, I can update the Jira ticket.

## Credit

calm (they/them)

### DIFF
--- a/public/stylesheets/site/2.0/10-types-groups.css
+++ b/public/stylesheets/site/2.0/10-types-groups.css
@@ -103,11 +103,9 @@ https://otwcode.github.io/docs/classes-taxonomy.html#grouping-and-spacing
 #outer.wrapper {
   float: left;
   width: 100%;
-
     display: flex;
     display: -webkit-flex;
     flex-direction: column;
-    -webkit-flex-direction: column;
     min-height: 100vh;
 }
 
@@ -116,7 +114,6 @@ https://otwcode.github.io/docs/classes-taxonomy.html#grouping-and-spacing
 }
 
 #inner.wrapper {
-    -webkit-flex: 1;
     flex: 1;
 }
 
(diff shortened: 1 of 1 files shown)


public/stylesheets/site/2.0/10-types-groups.css
```diff
@@ -103,12 +103,23 @@ https://otwcode.github.io/docs/classes-taxonomy.html#grouping-and-spacing
 #outer.wrapper {
   float: left;
   width: 100%;
+
```
sarken: Can you remove this extra blank line?

sarken (changes requested):
This is just a minor cosmetic issue, so the CSS3 approach is fine by me! In the name of simplicity, and since this is a noncritical fix, could you remove the `-webkit` prefixed properties? We don't use them elsewhere.

public/stylesheets/site/2.0/10-types-groups.css
```diff
@@ -103,12 +103,20 @@ https://otwcode.github.io/docs/classes-taxonomy.html#grouping-and-spacing
 #outer.wrapper {
   float: left;
   width: 100%;
+    display: flex;
+    display: -webkit-flex;
```
sarken: This one can go too! Sorry for not saying "properties and values"

sarken (changes requested):

sarken (approved):

PR #5831, opened 2026-05-20
AO3-7476 Fix that text-only bylines for anon creations could contain HTML

Bilka2:
## Issue

https://otwarchive.atlassian.net/browse/AO3-7476

## Purpose

Fix that text only bylines, like they are currently used in the series browser title, could contain HTML for admins viewing anon series or creators viewing their own anon series.

This bug did not get found earlier because all other spots using text-only bylines deal with logged-out scenarios (share modal).

## Credit

Bilka


brianjaustin (approved):
Thanks!

PR #5832, opened 2026-05-20
AO3-6576 Published date in news post blurb is unnecessarily complex to screen readers 

forceofcalm:
# Pull Request Checklist

## Issue

https://otwarchive.atlassian.net/browse/AO3-6576

## Purpose

This removes the extraneous `<span>`s wrapped around the date/time elements on news posts, work blurbs, and anything else that uses `date_in_zone` to improve screenreader accessibility. A comment on the Jira ticket also noted that the time zone's `title` seemed strange, so it has been updated to use the unabbreviated time zone name instead of the Country/City it previously used.

## Credit

calm (they/them)

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -257,7 +257,7 @@ module ApplicationHelper
     unabbreviated_zone_name = TZInfo::Timezone.get(zone).friendly_identifier(true) rescue zone
     
     time_parts = [
-      localized_time.strftime('<abbr class="day" title="%A">%a</abbr> %d <abbr class="month" title="%B">%b</abbr> %Y %I:%M%p'),
+      localized_time.strftime('<abbr class="day" title="%A">%a</abbr> %d <abbr class="month" title="%B">%b</abbr> %Y %I:%M%p').html_safe,
       " ",
       content_tag(:abbr, localized_time.zone, class: "timezone", title: unabbreviated_zone_name)
     ]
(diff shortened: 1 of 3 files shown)


app/helpers/application_helper.rb
```diff
@@ -253,24 +253,37 @@ def time_in_zone(time, zone = nil, user = User.current_user)
     return ts("(no time specified)") if time.blank?
```
Bilka2, line 253: Could you be so nice and change this over to `t` based on the [i18n standards](https://github.com/otwcode/otwarchive/wiki/Internationalization-(i18n)-Standards)?

Bilka2 (changes requested):
After some internal discussion, we'd like to remove the `abbr` tags from the output as well, not just the span tags

In connection with that making the method and output at lot less complicated, could you investigate whether this could use [localize](https://www.rubydoc.info/gems/i18n/I18n/Backend/Base:localize) (usually shortened to `l`) for the formatting instead of `strftime`? [Rails' i18n guide](https://guides.rubyonrails.org/i18n.html#adding-date-time-formats) has some information on how to add a format, in this case I expect that we'd want a completely different format from the ones that Rails provides, to match the previous output formatting, similar to how you added `date_blurb_html`

app/helpers/date_helper.rb
```diff
@@ -6,7 +6,7 @@ def set_format_for_date(datetime)
     if datetime > 30.days.ago && !AdminSetting.enable_test_caching?
       time_ago_in_words(datetime)
     else
-      date_in_user_time_zone(datetime).to_date.to_formatted_s(:rfc822)
+      date_in_zone(date_in_user_time_zone(datetime), format: :date_blurb_html)
```
Bilka2, line 9: Since this already sets the time zone with `date_in_user_time_zone`, isn't the `date_in_zone` call redundant, so this could be something like this?
```suggestion
      I18n.l(date_in_user_time_zone(datetime), format: :date_blurb_html)
```

If yes then the changes to `date_in_zone` aren't needed

